PR #3110, opened 2021-09-01
torch_pin

miaoshasha:
To be used in build_torch_wheels.sh here: https://github.com/pytorch/xla/blob/master/scripts/build_torch_wheels.sh#L179


JackCaoG (approved):
Thanks!

zcain117 (approved):

PR #3128, opened 2021-09-16
match an optional 3rd digit for version number

miaoshasha:
tested:
~$ release_version="1.2.1"
~$ echo "${release_version}" | grep -oP '\d+.\d+(.\d+)?'
1.2.1
~$ release_version="1.2"
~$ echo "${release_version}" | grep -oP '\d+.\d+(.\d+)?'
1.2
~$ release_version="1.10.1"
~$ echo "${release_version}" | grep -oP '\d+.\d+(.\d+)?'
1.10.1
~$ release_version="1.10.11"
~$ echo "${release_version}" | grep -oP '\d+.\d+(.\d+)?'
1.10.11

### DIFF
--- a/docker/common.sh
+++ b/docker/common.sh
@@ -16,7 +16,7 @@ function collect_wheels() {
   release_version=$1
   wheel_version="${release_version}"
   if [ "${release_version}" != "nightly" ]; then
-    wheel_version=$( echo "${release_version}" | grep -oP '\d+.\d+' )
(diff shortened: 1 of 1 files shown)


JackCaoG (approved):
Thanks @miaoshasha !

jysohn23 (approved):
If the wheel paths/names are changing we may need to update places we tell users how to install (tutorials / colab).

PR #3133, opened 2021-09-20
Updated python env to python3

ultrons:
Earlier version of the scripts specified #!/usr/bin/env python for debug_run.py and related scripts.
This led to module not found errors for modules such as matplotlib, PIL etc, even though these modules were already installed on TPU VM.

The PR is to change env reference to python3.
With this change, the debug_run.py script runs without errors.

Also updated for other scripts using env python.

### DIFF
--- a/scripts/bench_tensor_io.py
+++ b/scripts/bench_tensor_io.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function
 
--- a/scripts/cond_patch.py
+++ b/scripts/cond_patch.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function
 
--- a/scripts/debug_run.py
+++ b/scripts/debug_run.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function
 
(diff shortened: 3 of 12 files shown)


jysohn23:
Alternatively you can just update the default python version to be py3 with
```
sudo update-alternatives --install /usr/bin/python python /usr/bin/python2.7 1
sudo update-alternatives --install /usr/bin/python python /usr/bin/python3 2
```

ultrons:
@jysohn23 , thanks, yes it's a workaround that would help.

Although out of the box what we have pytorch/xla will error out for those scripts. So it would nice to fix something.

Perhaps we can update:
the [documentation] (https://github.com/pytorch/xla/blob/master/TROUBLESHOOTING.md)
And put a check in the scripts for python version that points users to do update-alternatives or we can simply update the env lines as proposed.
Or we can include the update-alternatives lines in the default TPU VM image itself.
No change is needed on the TPU Node side since it's by default python 3.x only.

Happy to close this request or keep it open.
Please advise.

JackCaoG:
seem to be no harm of doing this. I can merge this one.

JackCaoG (approved):

yeounoh:
I actually realized the change on master and traced back to this PR -- we should control the python version and installed dependencies through the box's Python setting (or and preferably via conda or venv), not hard-coding a particular version on every python script. I would suggest reverting this PR for maintainability and mark the original issue as `stale` ? cc @JackCaoG

PR #3143, opened 2021-09-22
torch pin

miaoshasha:
checkout pytorch release/1.9 branch


miaoshasha:
torch_pin file already exists in r1.9.... no need to do this again.

Issue #3141, opened 2021-09-22
[PT_BREAK] nll_backward for negative weights


JackCaoG:
Corresponding pytorch pr https://github.com/pytorch/pytorch/pull/64572

```
Sep 21 18:37:08 ======================================================================
Sep 21 18:37:08 FAIL [0.062s]: test_nll_loss_all_ignored_xla (__main__.TestNNDeviceTypeXLA)
Sep 21 18:37:08 ----------------------------------------------------------------------
Sep 21 18:37:08 Traceback (most recent call last):
Sep 21 18:37:08   File "/opt/conda/lib/python3.6/site-packages/torch/testing/_internal/common_device_type.py", line 368, in instantiated_test
Sep 21 18:37:08     result = test(self, **param_kwargs)
Sep 21 18:37:08   File "/var/lib/jenkins/workspace/xla/test/../../test/test_nn.py", line 17359, in test_nll_loss_all_ignored
Sep 21 18:37:08     helper([2, 3])
Sep 21 18:37:08   File "/var/lib/jenkins/workspace/xla/test/../../test/test_nn.py", line 17356, in helper
Sep 21 18:37:08     self.assertEqual(F.nll_loss(input, target, ignore_index=0, reduction="mean").item(), float("nan"))
Sep 21 18:37:08   File "/var/lib/jenkins/workspace/xla/test/pytorch_test_base.py", line 607, in assertEqual
Sep 21 18:37:08     return DeviceTypeTestBase.assertEqual(self, x, y, *args, **kwargs)
Sep 21 18:37:08   File "/opt/conda/lib/python3.6/site-packages/torch/testing/_internal/common_utils.py", line 1955, in assertEqual
Sep 21 18:37:08     super().assertTrue(result, msg=self._get_assert_msg(msg, debug_msg=debug_msg))
Sep 21 18:37:08 AssertionError: False is not true : Scalars failed to compare as equal! Comparing 0.0 and nan gives a difference of nan, but the allowed difference with rtol=0.001 and atol=0.001 is only nan!
Sep 21 18:37:08 
Sep 21 18:37:08 ======================================================================
Sep 21 18:37:08 FAIL [0.052s]: test_nll_loss_total_weight_is_zero_xla (__main__.TestNNDeviceTypeXLA)
Sep 21 18:37:08 ----------------------------------------------------------------------
Sep 21 18:37:08 Traceback (most recent call last):
[...]

PR #3144, opened 2021-09-23
Nll loss update

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3141. 

For `nll_loss` when reduction = "mean" and whole target is equal to ignore_index. It should returns NaN.
For `nll_loss_backward` when reduction = "mean" and whole target is equal to ignore_index. It should returns 0.

Modify the cpp util to handle the `nan` in the input case.


JackCaoG:
```
[  FAILED  ] AtenXlaTensorTest.TestSign
[  FAILED  ] AtenXlaTensorTest.TestAbs
[  FAILED  ] AtenXlaTensorTest.TestNllLossBackward
[  FAILED  ] AtenXlaTensorTest.TestLerpInplace
[  FAILED  ] AtenXlaTensorTest.TestLerpScalarInplace
```

Need to look into these failures

miladm (approved):
LGTM - thank you @JackCaoG

Issue #3120, opened 2021-09-10
[RFC] Introducing the zero gradients optimization for AMP


ymwangg:
## Introduction
Based on our discussions in #3043, we propose a solution (PR: #3119) to remove the extra host-device sync for NaN checking in GradScaler. We saw ~20% speedup in the `test_train_mp_imagenet.py` example and some improvements in other models with this optimization. However, this method may be controversial as it changes the training algorithm. So, we'd like to get some feedbacks from the community.
The assumption of this optimization is that the following algorithm:
```python
if not has_nan:
    optimizer.step()
```
is equivalent to:
```python
if has_nan:
    optimizer.zero_grad()
optimizer.step()
```
Then we can use the following trick to perform the gradient update:
```python
found_inf = torch.stack(
     tuple(optimizer_state["found_inf_per_device"].values())).sum()
scaling_factor = 1.0 - found_inf.to(torch.bool).to(torch.float)
for grad in optimizer.gradients:
     grad.nan_to_num_()
     grad.mul_(scaling_factor)
optimizer.step()
```
This is a reasonable approximation as `has_nan=True` happens infrequently. We have tested several models and haven't found any counterexample that this method changes the training behavior.
## Caveats
1. One major concern is that this method changes the training algorithm and may pose unknown risks to users. We also have some debates on it.
2. We noticed increased memory consumption when using Adam optimizer investigated separately in #3102 due to the removal of the mark_step in GradScaler, though this issue is not specific to this optimization. One workaround is to add a barrier before calling `optimizer.step`.

PS: Special thanks to @codeislife99 for extensive testing/discussions and @trevor-m for helpful discussions.

CC: @JackCaoG @anijain2305 @byronyi

## Bechmark Results
Here are some benchmark results running on Nvidia Tesla V100 GPU.
To reproduce the numbers, run the following scripts in `xla/test` with different batch sizes:
```sh
GPU_NUM_DEVICES=1 python test_train_mp_mnist_amp.py --fake_data --batch_size 128
GPU_NUM_DEVICES=1 python test_train_mp_mnist_amp.py --fake_data --batch_size 128 --use_zero_grad
GPU_NUM_DEVICES=1 python test_train_mp_imagenet.py --fake_data --amp --batch_size 64
GPU_NUM_DEVICES=1 python test_train_mp_imagenet.py --fake_data --amp --batch_size 64 --use_zero_grad
```

| Model      | original | with zero_grad | speedup |
[...]

PR #3145, opened 2021-09-24
Syncfree SGD optimizer for AMP

ymwangg:
This PR adds the syncfree SGD optimizer as discussed in #3120 for AMP mode, which utilizes a scalar tensor to indicate whether the optimizer step should be skipped or not. We plan to add Adam optimizer in the future.

The syncfree SGD optimizer delivers similar performance improvement comparing to the zero gradients optimization while it does not change the optimization algorithm.

The following is the performance comparison chart collected on Nvidia Tesla V100 GPU.
| Model      | original | with zero_grad | with syncfree_optimizer |
| ------------------------- | ----------- | ----------- | ----------- |
| Mnist (batch_size=128)    | 19791.77 | 34146.73 | 37732.39 |
| Mnist (batch_size=256)    | 32025.77 | 54305.00 | 53677.90 |
| Mnist (batch_size=512)    | 50905.95 | 72757.84 | 82433.28 |
| ImageNet (batch_size=64)  | 720.41  | 980.22  | 947.10  |
| ImageNet (batch_size=128) | 975.15 | 1197.09 | 1178.05 |
| ImageNet (batch_size=256) | 1191.62 | 1356.52 | 1345.13 |

Unit: samples/second

To reproduce the numbers, run the following scripts in `xla/test` with different batch sizes:
```sh
GPU_NUM_DEVICES=1 python test_train_mp_mnist_amp.py --fake_data --batch_size 128
GPU_NUM_DEVICES=1 python test_train_mp_mnist_amp.py --fake_data --batch_size 128 --use_syncfree_optim
GPU_NUM_DEVICES=1 python test_train_mp_imagenet.py --fake_data --amp --batch_size 64
GPU_NUM_DEVICES=1 python test_train_mp_imagenet.py --fake_data --amp --batch_size 64 --use_syncfree_optim
```


JackCaoG:
Sorry for the delay, will take a look tmr.

test/test_syncfree_optimizers.py
```diff
@@ -0,0 +1,120 @@
...
+      xm.optimizer_step(
+          syncfree_optimizer, optimizer_args={"found_inf": found_inf})
+      xm.mark_step()
+      # reference step
+      ref_optimizer.zero_grad()
+      ref_output = ref_model(data)
+      ref_loss = loss_fn(ref_output, target)
+      ref_loss.backward()
+      if i % 2 != 0:
+        xm.optimizer_step(ref_optimizer)
```
JackCaoG, line 83: Can we add a comment about `i%2 != 0` is mock the `found_inf` tensor above?

torch_xla/amp/syncfree/grad_scaler.py
```diff
@@ -0,0 +1,11 @@
+import torch
+
+
+class GradScaler(torch.cuda.amp.GradScaler):
```
JackCaoG, line 4: Question, does `GradScaler` really needs to be in a separate class under `syncfree`? Would be make sense to add a `sync_free` option to existing amp `GradScalar`?

torch_xla/amp/syncfree/sgd.py
```diff
@@ -0,0 +1,199 @@
...
+               weight_decay=0,
+               nesterov=False):
+    if lr is not required and lr < 0.0:
+      raise ValueError("Invalid learning rate: {}".format(lr))
+    if momentum < 0.0:
+      raise ValueError("Invalid momentum value: {}".format(momentum))
+    if weight_decay < 0.0:
+      raise ValueError("Invalid weight_decay value: {}".format(weight_decay))
+
+    defaults = dict(
```
JackCaoG, line 103: I am less familiar with this part, I am guessing it is mostly the same as the pytorch sgd? The only difference is it calls pt/xla's  `_xla_sgd_optimizer_step` in the end.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -385,6 +385,49 @@ XLATensor XLATensor::get_dimensions_size(const XLATensor& input,
...
+  param.SetInPlaceIrValue(
+      ir::ops::Where(found_inf.GetIrValue(), param.GetIrValue(),
+                     param.GetIrValue() - d_p_value * lr_value));
+  // update step counter
+  ir::Value one_value =
+      GetIrValueForScalar(1.0, step.shape(), step.GetDevice());
+  step.SetInPlaceIrValue(ir::ops::Where(found_inf.GetIrValue(),
+                                        step.GetIrValue(),
+                                        step.GetIrValue() + one_value));
+}
```
JackCaoG, line 429: I think we should move this logic to a new `sgd_optimizer_step` Node under `torch_xla/csrc/ops/sgd_optimizer_step.h` and `torch_xla/csrc/ops/sgd_optimizer_step.cpp`. There are 2 reasons

1. It is more clear when user dump the IR graph this way. They will see a `xla::sgd_optimizer_step` in the graph instead of a bunch of `add`, `where` etc
2. We are trying to move all computation logic in the IR level (tensor_method.cpp) either to pytorch dispatcher or XLA lowering in the process of Lazy Tenor Upstream. 

You can use [logsumexp](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/logsumexp.h) Node as an example.

JackCaoG (changes requested):
Thanks for the contribution and detail benchmark!

torch_xla/amp/syncfree/sgd.py
ymwangg: That's right, the only difference is using `_xla_sgd_optimizer_step` to replace `F.sgd` and adding a `step` counter.
https://github.com/pytorch/pytorch/blob/master/torch/optim/sgd.py

torch_xla/amp/syncfree/grad_scaler.py
ymwangg: Using one `GradScaler` is a good idea. There are two options: 
1. Add a `sync_free` option to existing `amp.GradScaler.__init__`.
2. Add a check like`is_syncfree_optim = "found_inf" in inspect.signature(optimizer.step).parameters`.

I think probably option 2 is better as it automatically detects if the optimizer is sync_free optimizer.

torch_xla/csrc/tensor_methods.cpp
JackCaoG: I think a better example would be [collectivepermute](https://github.com/pytorch/xla/commit/39d22ee6ee6af6b022ba90b72cc2cca071316136), since it is also a op that doesn't exist from pytorch's perspective.

ymwangg: Thanks for the pointer, I've updated the PR with the following changes:
1. Incorporate `amp.syncfree.GradScaler` into the original `amp.GradScaler`.
2. Make `amp.syncfree.SGD` a subclass of `torch.optim.SGD`.
3. Add the new `SgdOptimizerStep` Node.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -385,6 +386,29 @@ XLATensor XLATensor::get_dimensions_size(const XLATensor& input,
...
+  ir::Value momentum_value =
+      GetIrValueForScalar(momentum, param.shape(), param.GetDevice());
+  ir::Value lr_value =
+      GetIrValueForScalar(lr, param.shape(), param.GetDevice());
+  ir::Value dampening_value =
+      GetIrValueForScalar(dampening, param.shape(), param.GetDevice());
+  ir::NodePtr node = ir::MakeNode<ir::ops::SgdOptimizerStep>(
+      found_inf.GetIrValue(), step.GetIrValue(), param.GetIrValue(),
+      d_p.GetIrValue(), buf.GetIrValue(), weight_decay_value, momentum_value,
+      lr_value, dampening_value, /*use_weight_decay*/ weight_decay != 0,
```
JackCaoG, line 405: nit `/*use_weight_decay=*/` (missing `=` at the end).

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -820,4 +820,45 @@ std::vector<xla::XlaOp> BuildAmpUpdateScale(const xla::XlaOp& current_scale,
...
 
+std::vector<xla::XlaOp> BuildSgdOptimizerStep(
+    const xla::XlaOp& found_inf, const xla::XlaOp& step,
+    const xla::XlaOp& param, const xla::XlaOp& d_p, const xla::XlaOp& buf,
+    const xla::XlaOp& weight_decay, const xla::XlaOp& momentum,
+    const xla::XlaOp& lr, const xla::XlaOp& dampening, bool use_weight_decay,
+    bool use_momentum, bool use_nesterov) {
+  /* XLA version of the SGD algorithm
+   * https://github.com/pytorch/pytorch/blob/master/torch/optim/_functional.py#L162-L180
+   */
```
JackCaoG, line 831: nit, it seems like we always use `//` instead of `/**/` for text comments. Could you follow this style? Thnaks!

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -820,4 +820,45 @@ std::vector<xla::XlaOp> BuildAmpUpdateScale(const xla::XlaOp& current_scale,
...
+  auto dtype = XlaHelpers::ShapeOfXlaOp(param).element_type();
```
JackCaoG, line 832: We usually use `dtype` to refer to `at::ScalarType` which is a type from pytorch core. What you have here is a `XlaPrimitiveType` so you can just call it `type` or `para_type`.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -820,4 +820,45 @@ std::vector<xla::XlaOp> BuildAmpUpdateScale(const xla::XlaOp& current_scale,
...
+    xla::XlaOp buf_compute = xla::Select(
+        is_initialized_cond, buf * momentum + d_p_compute * (one - dampening),
+        d_p_compute);
+    d_p_compute =
+        use_nesterov ? d_p_compute + buf_compute * momentum : buf_compute;
+    new_buf = xla::Select(found_inf_cond, buf, buf_compute);
+  }
+  // update param
+  xla::XlaOp new_param =
+      xla::Select(found_inf_cond, param, param - d_p_compute * lr);
```
JackCaoG, line 853: it might be more efficient to do 
```
param -= d_p_compute * lr * xla::Not(found_inf_cond);
```
XLA should handle boolean to int/double promotion pretty well. You can also save a `not_found_inf_cond` since you need that for below computation too.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -820,4 +820,45 @@ std::vector<xla::XlaOp> BuildAmpUpdateScale(const xla::XlaOp& current_scale,
...
+  // update step counter
+  xla::XlaOp new_step = xla::Select(found_inf_cond, step, step + one);
```
JackCaoG, line 855: ditto, we can avoid `xla::Select` here. arithmetic operation usually have more room to optimize for the compiler.

JackCaoG (changes requested):
Mostly LGTM, minor nits.

torch_xla/csrc/xla_lower_util.cpp
ymwangg: changed.

torch_xla/csrc/xla_lower_util.cpp
ymwangg: changed

ymwangg: I found this trick is buggy when `d_p_compute` has `nan` and multiply it by 0 does not affect those `nan` elements. I've reverted this optimization and enhanced the test to cover this case.

.circleci/common.sh
```diff
@@ -109,6 +109,11 @@ function run_torch_xla_tests() {
       # python test/test_train_mp_mnist_amp.py --fake_data
     # fi
 
+    echo "Running Syncfree Optimizer Test"
+    if [ -x "$(command -v nvidia-smi)" ]; then
+      python test/test_syncfree_optimizers.py
```
JackCaoG, line 114: nit, can you remove 
```
    # if [ -x "$(command -v nvidia-smi)" ]; then
      # python test/test_train_mp_mnist_amp.py --fake_data
    # fi
```
above and add a 
```
# python test/test_train_mp_mnist_amp.py --fake_data
```
after your test. This way we group all gpu tests together(although it is commented out).

torch_xla/amp/syncfree/sgd.py
```diff
@@ -0,0 +1,118 @@
+import torch
+import torch_xla
+from torch import Tensor
+from typing import List, Optional
+
+
+class SGD(torch.optim.SGD):
```
JackCaoG, line 7: Can you add the link to pytorch's SGD github link in the front of the comment since this code came from there.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -385,6 +386,29 @@ XLATensor XLATensor::get_dimensions_size(const XLATensor& input,
                           at::ScalarType::Int);
 }
 
+void XLATensor::sgd_optimizer_step(const XLATensor& found_inf, XLATensor& step,
```
JackCaoG, line 389: Nit, Let's group `step`, `param` and `buf` in the front since `sgd_optimizer_step` is actually an inplace operations. It is more clear to put all tensors you want to update in the front and that's what we have done for most of our inplace ops.

I was thinking about renaming this function to `sgd_optimizer_step_` since it is an inplace op but I guess `step` is an indication of inplace update so it is fine.

JackCaoG (commented):
Thanks, mostly LGTM. After you fix the coupe nits and all test passed, we can start landing this pr.

JackCaoG (approved):
@ymwangg Thanks a lot for contributing, amazing job!

PR #3142, opened 2021-09-22
torch pin file

miaoshasha:
for checking out pytorch release/1.10 branch

PR #3147, opened 2021-09-30
torch pin file

miaoshasha:
torch pin file


JackCaoG:
Question, why is this commit needed? When I check r1.10 branch, there is already a torch pin commit https://github.com/pytorch/xla/pull/3142

miaoshasha:
> Question, why is this commit needed? When I check r1.10 branch, there is already a torch pin commit #3142

Good point, I think it's because after I merged in #3142, I did not update my local r1.10 branch, therefore created this pr. So this pr won't make changes to r1.10.

Issue #3152, opened 2021-10-04
[PT_BREAK] AtenXlaTensorTest.TestMatmul_1x2 failed


JackCaoG:
## 🐛 Bug

With https://github.com/pytorch/pytorch/pull/64387, TestMatmul_1x2 failed. My guess is that test was checking `mm` but with the optimization a different op is being called when `torch::matmul` is being called.

PR #3154, opened 2021-10-05
Update matmul cpp test to check a different counter

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3152, pytorch pr in https://github.com/pytorch/pytorch/pull/64387


miladm (approved):
LGTM. Thanks @JackCaoG.
Probably we need a rebase for this PR since it's been a while. Sorry it fell off from my radar.

PR #3160, opened 2021-10-12
Add support for monitor

kat25230:


kat25230:
> Thanks for the contribution, can we add a test in https://github.com/pytorch/xla/blob/master/test/test_profiler.py to cover this feature?

Yeah! will do it soon!

Issue #3184, opened 2021-10-26
[PT_BREAK] Changed how dispatching works for advanced indexing operations


zou3519:
Hi, https://github.com/pytorch/pytorch/pull/66506 is failing XLA CI and the failure seems real. For some more context, that PR fixes a bug in the dispatcher where the indices passed to advanced indexing (`tensor[idx0, idx1, ...]`) are not considered for dispatching; previously the dispatcher always dispatches on the dispatch keys in `tensor` but now the dispatcher considers all of the keys in `{tensor, idx, idx1, ...}`.

Please let me know if there's anything I can do to help. Thanks.

PR #3188, opened 2021-10-29
Advance index

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3184. After the pytorch/pytorch#66506, index and base of an index can be on different device and pt/xla needs to handle this case by moving all tensors to the xla device.

FYI: @zou3519


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1643,11 +1649,19 @@ at::Tensor& XLANativeFunctions::index_put_(
   XLA_CHECK(self.scalar_type() == values.scalar_type());
   CanonicalIndexInfo canonical_index_info =
       GetCanonicalIndexInfo(self, indices);
-  XLATensor self_tensor = bridge::GetXlaTensor(self);
+  c10::optional<Device> device =
+      bridge::GetXlaDevice(canonical_index_info.base);
+  if (!device.has_value()) {
```
miladm, line 1654: Why don't we apply this `if/else` logic in `GetCAnonicalIndexInfo()`? Is that because we need `device` for `bridge::GetOrCreateXlaTensor(values, *device)`? If yes, the next question is, why do we believe an upstream `value` tensor "can" be on a different device?

(My understanding is `base` and `index` are the root cause of this failure; correct me if I am wrong.)

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1591,10 +1591,16 @@ at::Tensor XLANativeFunctions::index(
...
-      XLATensor::index(bridge::GetXlaTensor(canonical_index_info.base),
-                       bridge::GetXlaTensors(canonical_index_info.indices),
-                       canonical_index_info.start_dim));
+  c10::optional<Device> device =
+      bridge::GetXlaDevice(canonical_index_info.base);
+  if (!device.has_value()) {
+    device = bridge::GetXlaDevice(canonical_index_info.indices);
+  }
+  XLA_CHECK(device.has_value());
+  return bridge::AtenFromXlaTensor(XLATensor::index(
```
miladm, line 1600: Can we test this scenario? (i.e. having `base` and `index` on diff devices)

miladm (changes requested):
Thanks @JackCaoG. I put in a few comments. Let me know what you think.

torch_patches/.torch_pin
```diff
@@ -0,0 +1 @@
+#66506
```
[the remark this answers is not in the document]
JackCaoG: https://github.com/pytorch/xla/issues/3184 has a good summary.

torch_xla/csrc/aten_xla_type.cpp
JackCaoG: Pytorch side has a test for this use case, https://github.com/pytorch/pytorch/pull/66506/files

torch_xla/csrc/aten_xla_type.cpp
JackCaoG: I put it here mostly because we also handle the binary op tensor [device mismatch](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/aten_xla_type.cpp#L115) in this file. 

Index is of type `c10::List<c10::optional<at::Tensor>>` which is something `GetOrCreateXlaTensor` can't handle. We generally not trying to change the `at::tensor` device type but create the `XLATensor` on the correct device. `at::tensor` -> `XLATensor` happens in this file.

miladm: Got it. Given it's used in three spots, why don't we do this logic as a separate function call?

JackCaoG: It is a bit awkward to add a helper function like

```
bridge::GetXlaDevice(at::Tensor tensor, vector<at::Tensor> tensors);
```
one cleaner way of doing this is to append the single tensor to the end of the vector and use the vector version of the `GetXlaDevice`, but that felt like an overkill since we either need to create a new vector or manually pop the `base` after this call.

miladm: Yeah, maybe this implementation is the most appropriate of all.

miladm (approved):
LGTM - Thanks @JackCaoG

JackCaoG:
@zou3519 Yup, I will merge this one once pytorch one merged

PR #3197, opened 2021-11-02
update libtpu link

miaoshasha:
correct link to libtpu


setup.py
```diff
@@ -54,7 +54,7 @@
 third_party_path = os.path.join(base_dir, 'third_party')
 
 _libtpu_version = '0.1.dev20210615'
```
JackCaoG, line 56: Can you also update this link to `dev20211015 `? Same for the 1.10 since `dev20211013` wheel has some issue.

JackCaoG (changes requested):

setup.py
miaoshasha: done

PR #3195, opened 2021-11-02
Silu backward

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3192. Need to backport to 1.10 but `silu_backward` symbol is missing on pytorch 1.10.

Pytorch pr to add symbol https://github.com/pytorch/pytorch/pull/67665

PR #3202, opened 2021-11-06
Silu backward r1 10

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3192, backporting https://github.com/pytorch/xla/pull/3195 to r1.10. Only difference being using `OpKind(at::aten::silu)` for `silu_backward` since `silu_backward` string is missing in pytorch 1.10.


bdhirsh (approved):
LGTM

Issue #3105, opened 2021-08-30
[PT_BREAK] Test crash after `torch.where` change


JackCaoG:
Pytroch pr in https://github.com/pytorch/pytorch/pull/62084, failure log
```
Aug 30 13:03:21 ERROR [0.037s]: test_where_scalar_valid_combination_xla_uint8 (__main__.TestTorchDeviceTypeXLA)
Aug 30 13:03:21 ----------------------------------------------------------------------
Aug 30 13:03:21 Traceback (most recent call last):
Aug 30 13:03:21   File "/opt/conda/lib/python3.6/site-packages/torch/testing/_internal/common_device_type.py", line 378, in instantiated_test
Aug 30 13:03:21     raise rte
Aug 30 13:03:21   File "/opt/conda/lib/python3.6/site-packages/torch/testing/_internal/common_device_type.py", line 373, in instantiated_test
Aug 30 13:03:21     result = test(self, **param_kwargs)
Aug 30 13:03:21   File "/var/lib/jenkins/workspace/xla/test/../../test/test_torch.py", line 7694, in test_where_scalar_valid_combination
Aug 30 13:03:21     self._test_where_scalar_template(device, dtype, checkResult)
Aug 30 13:03:21   File "/var/lib/jenkins/workspace/xla/test/../../test/test_torch.py", line 7659, in _test_where_scalar_template
Aug 30 13:03:21     exec_fn(scalar_type, dtype, condition, x, scalar_1)
Aug 30 13:03:21   File "/var/lib/jenkins/workspace/xla/test/../../test/test_torch.py", line 7683, in checkResult
Aug 30 13:03:21     self.assertEqual(scalar_out, tensor_out)
Aug 30 13:03:21   File "/var/lib/jenkins/workspace/xla/test/pytorch_test_base.py", line 606, in assertEqual
Aug 30 13:03:21     x, y = self.prepare_for_compare(x, y)
Aug 30 13:03:21   File "/var/lib/jenkins/workspace/xla/test/pytorch_test_base.py", line 549, in prepare_for_compare
Aug 30 13:03:21     x = tx.to(device='cpu')
Aug 30 13:03:21 RuntimeError: Error while lowering: u8[0]{0} aten::where
Aug 30 13:03:21 Error: /var/lib/jenkins/workspace/xla/torch_xla/csrc/helpers.cpp:584 : Check failed: xla::ShapeUtil::SameElementType(shape1, shape2) 
[...]

PR #3208, opened 2021-11-10
Fix torch.where(bool, int64, 0.5)

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3105.

Case this failed is 
```
bool_t = torch.tensor([True, False], device=device)
true_t = torch.tensor([10, 20], device=device)

t1 = torch.where(bool_t, true_t, 0.5)
```


wonjoolee95 (approved):
LGTM 👍

JackCaoG:
Need to resolve the conflict

miladm (changes requested):
Thanks @JackCaoG!
LTGM

nit: shall we add this specific condition as a new test scenario to `TestWhere`?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2916,9 +2916,19 @@ void XLATensor::zero_(XLATensor& input) {
...
-      condition.GetIrValue(), input.GetIrValue(), other.GetIrValue()));
+                           const XLATensor& other,
+                           c10::optional<at::ScalarType> logical_element_type) {
+  if (logical_element_type.has_value()) {
+    return input.CreateFrom(
+        ir::ops::Where(
+            condition.GetIrValue(), input.GetIrValue(), other.GetIrValue(),
+            MakeXlaPrimitiveType(*logical_element_type, &input.GetDevice())),
+        logical_element_type);
+  } else {
```
miladm, line 2927: nit: you can drop this else block given the previous return.

PR #3219, opened 2021-11-19
update colab torch version

miaoshasha:
custom wheel for colab is hard to build and colab has set 1.10 as default torch version
so installing 1.9 as default for our colabs.


JackCaoG (approved):
Thanks @miaoshasha

Issue #3210, opened 2021-11-11
Index_add with alpha


JackCaoG:
## 🚀 Feature
Support `Index_add` with `alpha` parameter. More context in https://github.com/pytorch/xla/pull/3209

miladm:
@yeounoh can you please help with this issue?

PR #3227, opened 2021-11-24
Enabling alpha scaling for index_add op lowering

yeounoh:
We previously lowered index_add op without alpha kwarg support (https://github.com/pytorch/pytorch/pull/65993): 

> Currently, XLA doesn't support an alpha kwarg, which is interesting to me as it diverges from what we have in PyTorch core. Should we also think of adding an alpha kwarg in XLA?

This PR addresses this per #3210


miladm:
Thanks @yeounoh! 
For matching the style, please review [C++ Style Guide](https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md#c-style-guide) and [Python Style Guide](https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md#python-style-guide)

yeounoh:
Note that this PR is branched off of #3209 which is still yet to be merged. We can merge this one after merging #3209  WDYT @JackCaoG

JackCaoG:
Yea, we need to merge my pr first, which is blocked by another pytorch pr

test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -5448,12 +5448,13 @@ TEST_F(AtenXlaTensorTest, TestIndexAdd) {
     }
   }
   ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
-  ExpectCounterChanged("xla::index_add_", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::index_add", cpp_test::GetIgnoredCounters());
```
JackCaoG, line 5451: Hmm, this is mixing the stuff in my pr in this pr. I guess it is due to you try to merge to master directly. This will make review a bit tricky

test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -5448,12 +5448,13 @@ TEST_F(AtenXlaTensorTest, TestIndexAdd) {
...
 }
 
 TEST_F(AtenXlaTensorTest, TestIndexAddInPlace) {
   int index_size = 10;
   int rank = 3;
+  double alpha = 1.0;
```
JackCaoG, line 5457: For you case, you should use different alpha

miladm: Consider using a `vector` of alpha values, containing, `{0, 1, <another-int>}`

test/cpp/test_aten_xla_tensor.cpp
miladm: I wonder if a rebase of `master` can help simplify things here?

xla_native_functions.yaml
```diff
@@ -139,7 +139,7 @@ supported:
   - hardtanh
   - hardtanh_backward
   - index.Tensor
-  - index_add_
+  - index_add
```
miladm, line 142: + What's the reason we moved to `index_add` instead of `index_add_`?

+ Given the PT dispatcher identifies `index_add_` as the reference (see [this](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/native_functions.yaml#L5773)), isn't it more appropriate to stay with `index_add_`?

CC @JackCaoG for more context

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1355,18 +1355,14 @@ XLATensor XLATensor::index(const XLATensor& input,
...
-      XlaHelpers::GetCanonicalDimensionIndex(dim, input.shape().get().rank());
-  return input.CreateFrom(IndexAdd(input, canonical_dim, index, source));
-}
-
-void XLATensor::index_add_(XLATensor& input, xla::int64_t dim,
-                           const XLATensor& index, const XLATensor& source) {
+                               const XLATensor& index, const XLATensor& source,
+                               const at::Scalar& alpha) {
+  ir::Value constant = GetIrValueForScalar(
+      alpha, sourcee.shape().get().element_type(), input.GetDevice());
```
miladm, line 1361: nit: `source`

miladm (commented):
Thanks for the initial PR push @yeounoh. Hope the comments below help.

test/cpp/test_aten_xla_tensor.cpp
yeounoh: @JackCaoG if you are planning on merging your PR, we can wait until the landing. It automatically resolve, an alternative would be to create a PR against your branch as well. WDYT?

xla_native_functions.yaml
yeounoh: The reference (PT native functions) will be updated once the underpinning PT PR lands; afterwards, it would be more appropriate to manually implement the out-of-place version.

test/cpp/test_aten_xla_tensor.cpp
JackCaoG: you can wait when my pr merged.

JackCaoG:
@yeounoh You should be able to rebase and pass the ci build now

test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -5500,38 +5500,41 @@ TEST_F(AtenXlaTensorTest, TestIndexAdd) {
...
+                                           base.sizes().end());
+          int canonical_dim = dim < 0 ? dim + rank : dim;
+          value_sizes[canonical_dim] = index_size;
+          torch::Tensor value =
+              isFloatingType(scalar_type)
+                  ? torch::rand(value_sizes, torch::TensorOptions(scalar_type))
+                  : torch::randint(100, value_sizes,
+                                   torch::TensorOptions(scalar_type));
+          torch::Tensor xla_base = CopyToDevice(base.clone(), device);
+          torch::Tensor result = base.index_add_(dim, index, value, alpha);
```
yeounoh, line 5529: This (native torch implementation, bypassing lowering) fails when `scalar_type=kByte` with:
```
++ exception with description "value cannot be converted to type uint8_t without overflow" thrown in the test body.
```
cc @JackCaoG any suggestion?

JackCaoG: I am guessing it is because when `alpha = 2`, `value = 100` and `base = 100` it will overflow. Since we have alpha scaling, we can just make `value` with upper_bound of `50` instead of `100`.

yeounoh: Awesome, and it's unsigned. I think the real culprit might be `alpha=-1.0` which will cause overflow. Let me try to limit the `randint` range as well as the `alphas`.

Issue #3224, opened 2021-11-23
Nightly build test backward compatibility issue with syncfree


yeounoh:
## 🐛 Bug

Some of nightly build (master) test cases break with latest stable releases (e.g., `r1.10`, `r.1.9`, etc.) with syncfree AMP package import.

## To Reproduce

1. Create a torch-xla VM (use `--image-family=torch-xla`)
2. Activate torch-xla-1.10 virtualenv: `conda activate torch-xla-1.10`
3. Checkout pytorch/xla git repo `master` branch
4. Run: `python test/test_train_mp_mnist_amp.py`

Error traceback:
```
raceback (most recent call last):
  File "test/test_train_mp_mnist_amp.py", line 40, in <module>
    from torch_xla.amp import autocast, GradScaler, syncfree
ImportError: cannot import name 'syncfree' from 'torch_xla.amp' (/anaconda3/envs/torch-xla-1.10/lib/python3.7/site-
packages/torch_xla/amp/__init__.py)
```

## Expected behavior

Nightly build should work with the latest stable release(s). In order to resolve this issue, we should:
- Testing the newly introduced syncfree SGD optimizer within its own test case.
- HOT-FIX/ implement a conditional import and/or syncfree test case if it's not available in the runtime.

yeounoh:
Hi @ymwangg, thank you for the contribution #3145 . As I was investigating the backward compatibility issue here, I've noticed that the following test cases were expanded for different amp optimizer options:

- `test_train_mp_mnist_amp.py`
- `test_train_mp_imagenet.py`

and we have the following test cases as well:

- `test_syncfree_optimizers.py` (using mnist)
- `test_train_mp_mnist.py`

Normally, it would be leaner and more test target specific to:

- have an appropriate/ simple tests for a test target, like you do in `test_syncfree_optimizers.py`
- have a evaluation script wrapping around individual test cases for performance evaluation & comparison
- also, it would be better to have test cases as mutually exclusive as possible, without much overlap.

Going forward, this will become critical as we introduce more optimization options.

As a small step toward these and to fix the backward compatibility issue, I was thinking about reverting:

- `test_train_mp_imagenet.py`,

so it doesn't test different optimizer options; and we can keep:

- `test_syncfree_optimizers.py` (for syncfree specific tests)
- `test_train_mp_mnist_amp.py` (for testing different amp optimizer optinos); also possibly rename this to `test_amp_optimizers.py`

[...]

PR #3228, opened 2021-11-27
Unbreak syncfree import

yeounoh:
This PR addresses #3224 , until we release a new stable ver >=1.11

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -111,10 +111,12 @@ function run_torch_xla_tests() {
 
     # GPU tests
     if [ -x "$(command -v nvidia-smi)" ]; then
-      echo "Running Syncfree Optimizer Test"
-      python test/test_syncfree_optimizers.py
-      # echo "Running MNIST Test"
-      # python test/test_train_mp_mnist_amp.py --fake_data
+      # Syncfree SGD optimizer tests
+      if [ -d ./amp/torch_xla/amp/syncfree]
+        echo "Running Syncfree Optimizer Test"
+        python test/test_syncfree_optimizers.py
+        # echo "Running MNIST Test"
+        # python test/test_train_mp_mnist_amp.py --fake_data
     fi
 
     pushd test/cpp
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -56,6 +56,9 @@ function install_deps_pytorch_xla() {
   pip install hypothesis
   pip install cloud-tpu-client
 
+  # Using the Ninja generator requires CMake version 3.13 or greater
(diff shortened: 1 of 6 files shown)


test/test_train_mp_imagenet.py
```diff
@@ -27,18 +27,6 @@
     '--test_only_at_end': {
         'action': 'store_true',
     },
-    # AMP only works with XLA:GPU
```
JackCaoG, line 30: Let's create a `test_train_mp_imagenet_amp.py` and only run the test if it is on GPU.

.circleci/common.sh
```diff
@@ -56,6 +56,9 @@ function install_deps_pytorch_xla() {
   pip install hypothesis
   pip install cloud-tpu-client
 
+  # Using the Ninja generator requires CMake version 3.13 or greater
+  pip install cmake>=3.13 --upgrade
```
yeounoh, line 60: cc @miladm FYI

test/test_ops.py
```diff
@@ -252,6 +249,9 @@ def __new__(cls, name, variant_test_name=""):
             # Failing Ops
             # Refer for more info : https://github.com/pytorch/xla/pull/3019#issuecomment-877132385
             # AllowedOpInfoEntry('cdist'),  // precision issue on TPU
+            # AllowedOpInfoEntry('linalg.multi_dot'),  // failing on CPU
```
yeounoh, line 252: #3232 cc @JackCaoG

.circleci/common.sh
```diff
@@ -111,10 +114,19 @@ function run_torch_xla_tests() {
...
+        echo "Running Syncfree Optimizer Test"
+        python test/test_syncfree_optimizers.py
+
+        # Following test scripts are mainly useful for
+        # performance evaluation & comparison among different
+        # amp optimizers.
+        # echo "Running MNIST Test"
+        # python test/test_train_mp_mnist_amp.py --fake_data
+        # echo "Running ImageNet Test"
+        # python etst/test_train_mp_imagenet_amp.py --fake_data
```
JackCaoG, line 128: mnist is disabled for a different reason, we can run `imagenet_amp` when it is gpu setup

JackCaoG (approved):
Minor comment, not going to block the merge. You can fix the issue in a separate pr

.circleci/common.sh
JackCaoG: Make sure you only run 1 epoch through, otherwise it might take too long

yeounoh: Created #3233

PR #3229, opened 2021-11-29
Fix f64 div(f32, f64) issue

ymwangg:
This PR fixes the f64 div issue when a f32 tensor is divided by a f64 scalar. This issue significantly slows down some huggingface models on Nvidia T4 GPU which has poor f64 performance.

Minimal code to reproduce this issue:
```python
import torch, torch_xla

device = xm.xla_device()

a = torch.rand(10,10, dtype=torch.float32).to(device)
b = a / 2.0
print(torch_xla._XLAC._get_xla_tensors_text([b]))
```
Before this fix:
```llvm
IR {
  %0 = f64[] xla::device_data(), device=GPU:0
  %1 = f32[10,10]{1,0} xla::device_data(), device=GPU:0
  %2 = f64[10,10]{1,0} aten::div(%1, %0)
  %3 = f32[10,10]{1,0} xla::cast(%2), type=f32, ROOT=0
}
```
After this fix:
```llvm
IR {
  %0 = f32[] xla::device_data(), device=GPU:0
  %1 = f32[10,10]{1,0} xla::device_data(), device=GPU:0
  %2 = f32[10,10]{1,0} aten::div(%1, %0), ROOT=0
}
```

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1245,7 +1245,7 @@ at::Tensor XLANativeFunctions::div(
     c10::optional<c10::string_view> rounding_mode) {
   XLA_FN_COUNTER("xla::");
   at::ScalarType dtype = at::result_type(self, other);
-  auto operands = GetBinaryOperands(self, other);
+  auto operands = GetBinaryOperands(self, UnwrapNumber(other, dtype));
   return bridge::AtenFromXlaTensor(
       XLATensor::div(operands.first, operands.second, rounding_mode, dtype));
 }
--- a/test/test_data_type.py
+++ b/test/test_data_type.py
@@ -48,6 +48,13 @@ class XlaDataTypeTest(unittest.TestCase):
     else:
       assert 'f64' in device_data_hlo, device_data_hlo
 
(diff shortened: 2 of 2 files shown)


JackCaoG:
Thanks @ymwangg , I will take a look

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1245,7 +1245,7 @@ at::Tensor XLANativeFunctions::div(
     c10::optional<c10::string_view> rounding_mode) {
   XLA_FN_COUNTER("xla::");
   at::ScalarType dtype = at::result_type(self, other);
-  auto operands = GetBinaryOperands(self, other);
+  auto operands = GetBinaryOperands(self, UnwrapNumber(other, dtype));
```
JackCaoG, line 1248: Ah.. nice catch. All of the binary ops should have this `UnwrapNumber` logic baked in. However `div` now we have `rounding_mode` in its signature, we don't call `DoBinaryOp`.

JackCaoG (approved):
Thanks!

JackCaoG:
@ymwangg Can you rebase to get pass the build error?

Issue #3237, opened 2021-12-06
Move `softplus` op out of tensor_ops


wonjoolee95:
As discussed in this PR https://github.com/pytorch/xla/pull/3236: 

"We should move the Softplus out of the tensor_ops and make it an regular IR in the future. We should create a BuildSoftPlus and BuildSoftPlusBackward function and use that for Mish lowering.

For the context, as we are moving toward the Lazy Tensor world, we are trying to get rid of this kind of IR level computation. It is better to have a Mish IR op."

PR #3250, opened 2021-12-10
Move softplus out of tensor_ops and make it an regular op

wonjoolee95:
This PR partly addresses https://github.com/pytorch/xla/issues/3237. 

This PR moves `softplus` out of tensor_ops and makes it an regular op. Following this, we'll do the same for `softplus_backward` and `mish`, which are currently both in tensor_ops.


JackCaoG (approved):
Thanks!

PR #3263, opened 2021-12-21
Move softplus_backward out of tensor_ops and make it an regular op

wonjoolee95:
This PR is a follow-up to https://github.com/pytorch/xla/pull/3250 and partly addresses #3237.

This PR moves softplus_backward out of tensor_ops and makes it an regular op. Previously, we've done this for softplus. Following this, we'll do the same for mish.


JackCaoG (approved):
Thanks!

JackCaoG:
Can you take a look at the test failure and remove the `tensor_op` version if it is not being used by other op?

PR #3255, opened 2021-12-16
Update CircleCI to use cuda11

wonjoolee95:
Draft PR to update CircleCI to use cuda11. Posting draft PR to see if CircleCI tests will fail.

PR #3269, opened 2021-12-27
.circleci: Only unlink if file exists

seemethere:
/usr/bin/bazel doesn't exist in upstream pytorch/pytorch so
make it so that unlink doesn't run if it doesn't exist

Fixes https://github.com/pytorch/xla/issues/3270 introduced in https://github.com/pytorch/xla/pull/3255

Signed-off-by: Eli Uriegas <eliuriegas@fb.com>


seemethere:
Going to go ahead and merge this to unbreak upstream CI

Issue #3279, opened 2022-01-05
PyTorch test_upsamplingBicubic2d_correctness_xla failing


wonjoolee95:
As noted in https://github.com/pytorch/xla/pull/3276, After pulling in latest changes for PyTorch, test `test_upsamplingBicubic2d_correctness_xla` fails with error:
```
======================================================================
FAIL: test_upsamplingBicubic2d_correctness_xla (__main__.TestNNDeviceTypeXLA)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/wonjoo/anaconda3/lib/python3.8/site-packages/torch/testing/_internal/common_device_type.py", line 376, in instantiated_test
    result = test(self, **param_kwargs)
  File "test/test_nn.py", line 15449, in test_upsamplingBicubic2d_correctness
    self.assertEqual(out_t, expected_out_t, atol=1e-5, rtol=0)
  File "/usr/local/google/home/wonjoo/pytorch/xla/test/pytorch_test_base.py", line 611, in assertEqual
    return DeviceTypeTestBase.assertEqual(self, x, y, *args, **kwargs)
  File "/usr/local/google/home/wonjoo/anaconda3/lib/python3.8/site-packages/torch/testing/_internal/common_utils.py", line 2088, in assertEqual
    super().assertTrue(result, msg=self._get_assert_msg(msg, debug_msg=debug_msg))
AssertionError: False is not true : Tensors failed to compare as equal!Attempted to compare equality of tensors with different dtypes. Got dtypes torch.float32 and torch.float64.

----------------------------------------------------------------------
Ran 1 test in 0.206s

FAILED (failures=1)
```

This test has been pushed a week ago from PyTorch -- https://github.com/pytorch/pytorch/pull/68819.

PR #3280, opened 2022-01-06
Downgrade to cuda10 and update softplus_backward op parameter

wonjoolee95:
In our previous attempt to fix the PT/xla head build failure (https://github.com/pytorch/xla/pull/3278) due to PyTorch's change on `softplus_backward`, we saw that the CircleCI build is failing with message:
```
+ sudo apt-get -qq install clang-9
E: Unable to locate package clang-9

Exited with code exit status 100
CircleCI received exit code 100
```

This seems odd given that this exact CircleCi commands have been succeeding before. However, to unblock the head as soon as possible, posting a new PR that does two things:
1. Fix the `softplus_backward` issue.
2. Revert the cuda11 change.

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -62,10 +62,7 @@ function install_deps_pytorch_xla() {
   # Bazel doesn't work with sccache gcc. https://github.com/bazelbuild/bazel/issues/3642
   sudo apt-get -qq update
 
-  # Install npm, nodejs, and necessary dependencies
-  sudo apt-get -qq install -y aptitude
-  sudo aptitude -y install npm
-  sudo aptitude -y install nodejs
+  sudo apt-get -qq install npm nodejs
 
   # XLA build requires Bazel
   # We use bazelisk to avoid updating Bazel version manually.
@@ -74,6 +71,7 @@ function install_deps_pytorch_xla() {
   if [[ -e /usr/bin/bazel ]]; then
     sudo unlink /usr/bin/bazel
   fi
+
   sudo ln -s "$(command -v bazelisk)" /usr/bin/bazel
 
   # Symnlink the missing cuda headers if exists
@@ -97,7 +95,7 @@ function install_deps_pytorch_xla() {
 function build_torch_xla() {
   XLA_DIR=$1
   pushd "$XLA_DIR"
-  CC=gcc-7 CXX=gcc-7 python setup.py install
+  CC=clang-9 CXX=clang++-9 python setup.py install
   popd
 }
 
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -8,10 +8,6 @@ source .circleci/common.sh
 PYTORCH_DIR=/tmp/pytorch
 XLA_DIR=$PYTORCH_DIR/xla
 
-sudo apt-get -qq update
-sudo apt-get -qq install clang-9
-sudo apt-get -qq install clang++-9
-
 source "$PYTORCH_DIR/.jenkins/pytorch/common_utils.sh"
 install_torchvision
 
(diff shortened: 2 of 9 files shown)


wonjoolee95:
Created a new PR instead of updating the PR https://github.com/pytorch/xla/pull/3278 with a new branch in case something messes up. I can merge the revert commit in the other PR if necessary.

wonjoolee95:
Seems like PyTorch has deprecated python3.6 and therefore removed the `pytorch-linux-bionic-cuda10.2-cudnn7-py3.6-clang9` image that we were using before - https://github.com/pytorch/pytorch/commit/066c9ff08f49acb9763381e4bdee6185dfc61871#diff-400f0e59b296557a196974ef08e11442d0bf24cabd8568f8e57643981f261843.

Updated to use `pytorch-linux-bionic-cuda10.2-cudnn7-py3.7-clang9` instead.

.circleci/common.sh
```diff
@@ -62,10 +62,7 @@ function install_deps_pytorch_xla() {
   # Bazel doesn't work with sccache gcc. https://github.com/bazelbuild/bazel/issues/3642
   sudo apt-get -qq update
 
-  # Install npm, nodejs, and necessary dependencies
-  sudo apt-get -qq install -y aptitude
```
yeounoh, line 66: We don't need this anymore?

.circleci/common.sh
```diff
@@ -97,7 +95,7 @@ function install_deps_pytorch_xla() {
 function build_torch_xla() {
   XLA_DIR=$1
   pushd "$XLA_DIR"
-  CC=gcc-7 CXX=gcc-7 python setup.py install
+  CC=clang-9 CXX=clang++-9 python setup.py install
```
yeounoh, line 98: Let's leave a comment for future reference here, saying that the base image declared in `./circleci/config.yml` should have clang-9 installed. This is a critical info because different images would have different c compilers.

.circleci/test.sh
```diff
@@ -8,10 +8,6 @@ source .circleci/common.sh
 PYTORCH_DIR=/tmp/pytorch
 XLA_DIR=$PYTORCH_DIR/xla
 
-sudo apt-get -qq update
```
yeounoh, line 11: Better to leave these to indicate the clang-9 dependencies. These commends would be no-ops anyway?

test/pytorch_test_base.py
```diff
@@ -263,6 +263,7 @@
         'test_softshrink_inplace_overlap_xla',  # doesn't raise
         'test_Conv2d_backward_depthwise_xla_float64',  # slow compilation
         'test_leaky_relu_inplace_with_neg_slope_xla',  # expecting a specific error message
+        'test_upsamplingBicubic2d_correctness_xla',  # FIXME! Got dtypes torch.float32 and torch.float64
```
yeounoh, line 266: Do we have an issue tracking this/fix? We should make sure that we clear all the regressions at some point in the near future.

yeounoh:
I have some minor comments; otherwise looks good to me. Thank you @wonjoolee95

test/pytorch_test_base.py
wonjoolee95: Yep, have an open issue here for tracking -- https://github.com/pytorch/xla/issues/3279.

.circleci/common.sh
wonjoolee95: So all the CircleCI related changes were part of the `revert` commit. Whatever that's deleted/removed, we're just reverting this one commit -- https://github.com/pytorch/xla/commit/91b5b55575fd249edd7b3cb852c659fc8655a35f.

.circleci/common.sh
wonjoolee95: Same comment as above -- this was part of the `revert` so we're just going back to the previous state. 

This is a good point though. I'll be working again to upgrade to cuda11, and I'll keep this in mind.

.circleci/test.sh
wonjoolee95: Good point! Again, just part of the `revert` commit 😄 I'll be adding these back when I upgrade back to cuda11.

test/pytorch_test_base.py
yeounoh: Awesome, thank you!

yeounoh (approved):
LGTM

PR #3281, opened 2022-01-07
Disable XLA tensors do not have storage

JackCaoG:
Disable the test since it tried to access the storage and XLA tensor does not have storage


yeounoh (approved):
LGTM

PR #3292, opened 2022-01-13
Async RNG seed upload with env var `XLA_TRANSFER_SEED_ASYNC=1`

JackCaoG:
Making PyTorch/XLA RNG seed sending Async with an env var `XLA_TRANSFER_SEED_ASYNC=1`

I tested imagenet with fake data and TPU and saw following
```
master:
| Training Device=xla:1/0 Epoch=1 Step=1160 Loss=0.00136 Rate=562.05 GlobalRate=456.88 Time=05:57:48
| Training Device=xla:0/2 Epoch=1 Step=1160 Loss=0.00136 Rate=562.00 GlobalRate=472.92 Time=05:57:48
| Training Device=xla:0/4 Epoch=1 Step=1160 Loss=0.00136 Rate=562.00 GlobalRate=469.03 Time=05:57:48
| Training Device=xla:0/1 Epoch=1 Step=1160 Loss=0.00136 Rate=561.84 GlobalRate=473.10 Time=05:57:48
| Training Device=xla:0/7 Epoch=1 Step=1160 Loss=0.00136 Rate=561.76 GlobalRate=471.02 Time=05:57:48
| Training Device=xla:0/3 Epoch=1 Step=1160 Loss=0.00136 Rate=562.27 GlobalRate=468.53 Time=05:57:48
| Training Device=xla:0/5 Epoch=1 Step=1160 Loss=0.00136 Rate=561.73 GlobalRate=470.90 Time=05:57:48
| Training Device=xla:0/6 Epoch=1 Step=1160 Loss=0.00136 Rate=561.44 GlobalRate=471.37 Time=05:57:48
```

```
async seed sending
| Training Device=xla:1/0 Epoch=1 Step=1160 Loss=0.00136 Rate=560.20 GlobalRate=458.10 Time=06:42:28
| Training Device=xla:0/6 Epoch=1 Step=1160 Loss=0.00136 Rate=560.28 GlobalRate=463.02 Time=06:42:28
| Training Device=xla:0/3 Epoch=1 Step=1160 Loss=0.00136 Rate=560.16 GlobalRate=464.99 Time=06:42:28
| Training Device=xla:0/1 Epoch=1 Step=1160 Loss=0.00136 Rate=560.19 GlobalRate=464.19 Time=06:42:28
| Training Device=xla:0/5 Epoch=1 Step=1160 Loss=0.00136 Rate=560.17 GlobalRate=464.80 Time=06:42:28
| Training Device=xla:0/2 Epoch=1 Step=1160 Loss=0.00136 Rate=559.96 GlobalRate=468.75 Time=06:42:28
| Training Device=xla:0/4 Epoch=1 Step=1160 Loss=0.00136 Rate=560.00 GlobalRate=467.79 Time=06:42:28
| Training Device=xla:0/7 Epoch=1 Step=1160 Loss=0.00136 Rate=559.95 GlobalRate=464.98 Time=06:42:28
```

```
async on
| Training Device=xla:0/4 Epoch=1 Step=1160 Loss=0.00136 Rate=556.90 GlobalRate=452.86 Time=06:11:50
| Training Device=xla:1/0 Epoch=1 Step=1160 Loss=0.00136 Rate=556.64 GlobalRate=446.03 Time=06:11:50
| Training Device=xla:0/3 Epoch=1 Step=1160 Loss=0.00136 Rate=556.73 GlobalRate=453.02 Time=06:11:50
| Training Device=xla:0/5 Epoch=1 Step=1160 Loss=0.00136 Rate=556.52 GlobalRate=445.45 Time=06:11:50
| Training Device=xla:0/1 Epoch=1 Step=1160 Loss=0.00136 Rate=556.18 GlobalRate=454.10 Time=06:11:50
| Training Device=xla:0/2 Epoch=1 Step=1160 Loss=0.00136 Rate=556.41 GlobalRate=452.99 Time=06:11:50
| Training Device=xla:0/7 Epoch=1 Step=1160 Loss=0.00136 Rate=556.06 GlobalRate=454.70 Time=06:11:50
| Training Device=xla:0/6 Epoch=1 Step=1160 Loss=0.00136 Rate=556.31 GlobalRate=455.49 Time=06:11:50
```
I saw a slight slowdown when async in turning on but no difference if I only make seed sending async

Looking at the metric
```
master:
Metric: TransferToServerTime
  TotalSamples: 1629
  Accumulator: 14s223ms363.480us
  ValueRate: 047ms365.107us / second
  Rate: 4.73877 / second
  Percentiles: 1%=806.312us; 5%=859.181us; 10%=918.245us; 20%=001ms008.134us; 50%=001ms232.361us; 80%=034ms453.229us; 90%=038ms059.089us; 95%=041ms195.293us; 99%=060ms169.396us
```
```
async seed sending
Metric: TransferToServerTime
  TotalSamples: 1629
  Accumulator: 14s608ms076.264us
  ValueRate: 045ms082.908us / second
  Rate: 4.7666 / second
  Percentiles: 1%=786.733us; 5%=852.923us; 10%=896.930us; 20%=972.145us; 50%=001ms213.824us; 80%=034ms370.254us; 90%=036ms150.783us; 95%=037ms233.854us; 99%=045ms016.930us
Metric: TensorToData
  TotalSamples: 1629
  Accumulator: 14s620ms038.231us
  ValueRate: 045ms123.516us / second
  Rate: 4.7666 / second
  Percentiles: 1%=790.766us; 5%=858.115us; 10%=901.730us; 20%=977.066us; 50%=001ms219.743us; 80%=034ms388.311us; 90%=036ms167.578us; 95%=037ms251.818us; 99%=045ms043.966us
```
```
async on:
Metric: TransferToServerTime
  TotalSamples: 1629
  Accumulator: 56s549ms003.270us
  ValueRate: 046ms281.348us / second
  Rate: 4.65329 / second
  Percentiles: 1%=840.060us; 5%=928.485us; 10%=979.762us; 20%=001ms082.171us; 50%=001ms327.852us; 80%=034ms852.131us; 90%=037ms024.033us; 95%=040ms629.738us; 99%=064ms862.448us
Metric: HandleBarrierWait
  TotalSamples: 581928
  Accumulator: 393ms361.102us
  ValueRate: 002ms782.456us / second
  Rate: 104.882 / second
  Percentiles: 1%=000.064us; 5%=000.064us; 10%=000.064us; 20%=000.065us; 50%=000.065us; 80%=000.225us; 90%=000.738us; 95%=000.963us; 99%=001.278us
Metric: TensorToData
  TotalSamples: 1629
  Accumulator: 08s652ms635.361us
  ValueRate: 023ms428.008us / second
  Rate: 4.65362 / second
  Percentiles: 1%=113.021us; 5%=126.192us; 10%=136.218us; 20%=150.262us; 50%=224.224us; 80%=016ms438.278us; 90%=019ms209.621us; 95%=022ms570.061us; 99%=045ms415.144us
```

Here is how to understand this data. For some reason `TransferToServerTime` increase significantly when turning on async for all data uploads. Since it is a background thread, it is actually fine. `HandleBarrierWait` is the amount of time that pt/xla is waiting for the handle to be assigned(data upload finish) and it only takes 393ms which means slower `TransferToServer` didn't block the graph tracing and graph execution. `TensorToData` is recorded in the parent of the `TransferToServerTime` and it only takes 8 s. This means we save ~6s in the device upload(foreground).

When making `rng seed sending` async, there is no speed implication... which is expected because random seed is actually not used on base `imagenet.py`. I verified using a small code that async handle is created and output the correct value. 

Conclusion:
Making all device data async seems unnecessary at this point and have speed impact. I will only make rng seed sending async by default.


JackCaoG:
error message
```
RuntimeError: tensorflow/compiler/xla/xla_client/xrt_computation_client.cc:470 : Check failed: async->session->session()->Run( async->session_work.feed_inputs, async->session_work.outputs_handles, &outputs) == ::tensorflow::Status::OK() (INVALID_ARGUMENT: GraphDef argument to Extend includes node 'XRTAllocateFromTensor_110', which was created by a previous call to Create or Extend in this session. vs. OK)
```

I have run into this error when running my model, suspecting it is a clean up problem. Looking.

JackCaoG:
Was able to repo this error on resnet if I add more debugging messages.  I am seeing some overlapping device uploading to the same device and then errors like 

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  tensorflow/compiler/xla/xla_client/xrt_computation_client.cc:468 : Check failed: async->session->session()->Run( async->session_work.feed_inputs, async->session_work.outputs_handles, &outputs) == ::tensorflow::Status::OK() (NOT_FOUND: FeedInputs: unable to find feed output Placeholder_23:0 vs. OK)
```

Here is my guess, although it is OK to make device uploading async, there can be at most 1 (chunk of)allocation happens at the same time. Execution of the graph actually carry the same constrain where it is async but the next execution has to be wait for the previous execution to finish. I am going to add a global device allocation lock and see if that solves the issue.

JackCaoG:
implemented the device lock which force the 1 alloc for each device at the same time. However I still see the error message
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  tensorflow/compiler/xla/xla_client/xrt_computation_client.cc:485 : Check failed: async->session->session()->Run( async->session_work.feed_inputs, async->session_work.outputs_handles, &outputs) == ::tensorflow::Status::OK() (NOT_FOUND: FeedInputs: unable to find feed output Placeholder_17:0 vs. OK)
```
looking.

JackCaoG:
I started to think it is a better idea to implement this async behavior in pt/xla layer instead of xrt. My guess is that tf grpc server doesn't really like messing up with feedinput/output node async. Logic will be slightly more complicated but at least we can keep it after we do the runtime replacement.

JackCaoG:
I think my implementation has a bug in the in place operation. Was able to repo if I add a sleep in the transfer function.

JackCaoG:
I think I fixed the bug, confirmed locally and will let ci confirm it again. The bug is in `CreateTensorsData` `at::Tensor` is passed as reference and can be modified inplace before async upload happened. We need to make a deep copy before passing it to the async function. There are some cleanup and optimization I want to do but I think the core part is finished. I need to redo the benchmark through.

JackCaoG:
Code is cleaned up, I will run the benchmark tmr.

JackCaoG:
@hjm-aws Could you help me to test this feature on your workload and let me know the performance? One alternative is to only transfer async for RNG seed.

@wconstab  for visibility.

JackCaoG:
Switched to only make `rng seed sending` aysnc by default. It will has much smaller impact to existing workload but still solve the rng seed slowness issue.

JackCaoG:
While waiting for @hjm-aws giving it a try. @miladm @yeounoh Could you guys take a look at this pr if you have cycle? I can provide more info regarding design and motivation if needed.

miladm:
Will review it in the morning @JackCaoG

third_party/xla_client/xrt_computation_client.cc
```diff
@@ -319,11 +349,24 @@ std::vector<size_t> XrtComputationClient::PartitionTransferToServer(
 
 std::vector<ComputationClient::DataPtr> XrtComputationClient::TransferToServer(
     absl::Span<const TensorSource> tensors) {
+  return TransferToServerHelper(tensors, {});
+}
+
+void XrtComputationClient::TransferToServer(
+    absl::Span<const TensorSource> tensors, absl::Span<const DataPtr> datas) {
+  XLA_CHECK_EQ(tensors.size(), datas.size());
```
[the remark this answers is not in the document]
miladm: on a second thought, it seems line 459 of this file (in `TransferToServerInternal`) creates an empty tensor when the `size()` is zero.

torch_xla/csrc/tensor_util.cpp
```diff
@@ -608,21 +650,46 @@ void PopulateTensorBuffer(const at::Tensor& tensor,
...
-          void* dest_buffer, size_t dest_buffer_size) {
-        PopulateTensorBuffer(tensor, source_tensor.shape, dest_buffer,
-                             dest_buffer_size, device);
-      };
+                                                const Device& device,
+                                                bool transfer_async) {
+  XLA_TIMED("TensorToData");
+  if (transfer_async) {
+    std::shared_ptr<DataAsync> async = std::make_shared<DataAsync>();
+    auto populate_mwait = std::make_shared<xla::util::MultiWait>(1);
```
miladm, line 658: can we replace the hardcoded `1` with an defined paramter for more code clarity?

torch_xla/csrc/tensor_util.cpp
```diff
@@ -608,21 +650,46 @@ void PopulateTensorBuffer(const at::Tensor& tensor,
...
+                               dest_buffer_size, device);
+          populate_mwait->Done();
+        };
 
-  std::vector<xla::ComputationClient::TensorSource> source_tensors;
-  source_tensors.emplace_back(shape, device.ToString(), std::move(populate_fn));
+    async->source_tensors.emplace_back(shape, device.ToString(),
+                                       std::move(populate_fn));
+    TransferToServerAsync(async, {device.ToString()});
+    XLA_CHECK_EQ(async->async_datas.size(), 1);
```
miladm, line 670: ditto

torch_xla/csrc/tensor_util.cpp
```diff
@@ -608,21 +650,46 @@ void PopulateTensorBuffer(const at::Tensor& tensor,
...
+    source_tensors.emplace_back(shape, device.ToString(),
+                                std::move(populate_fn));
 
-  auto handles =
-      xla::ComputationClient::Get()->TransferToServer(source_tensors);
-  XLA_CHECK_EQ(handles.size(), 1);
-  return std::move(handles.front());
+    auto handles =
+        xla::ComputationClient::Get()->TransferToServer(source_tensors);
+    XLA_CHECK_EQ(handles.size(), 1);
```
miladm, line 690: ditto

torch_xla/csrc/tensor_util.cpp
```diff
@@ -746,30 +813,59 @@ bool TensorCompare(const at::Tensor& t1, const at::Tensor& t2) {
 }
 
 xla::ComputationClient::DataPtr TensorToXlaData(const at::Tensor& tensor,
-                                                const Device& device) {
-  return TensorToXlaData(
-      tensor, CreateComputationShapeFromTensor(tensor, &device), device);
+                                                const Device& device,
```
miladm, line 816: I suggest we add metrics measurements to compute the latency of this code. Wdyt?

miladm (changes requested):
Thanks so much @JackCaoG! Awesome work.

+ Added a few comments
+ Can you provision a few test scenarios as unit-test?

third_party/xla_client/computation_client.h
```diff
@@ -166,10 +167,25 @@ class ComputationClient {
   // will be populated in an asynchrounous fashion.
   virtual DataPtr CreateDataPlaceholder(std::string device, Shape shape) = 0;
 
+  // Create DataPtr that only has dummy information which can be filled in
+  // later.
+  virtual std::vector<DataPtr> CreateAsyncDatas(
+      absl::Span<const TensorSource> tensors) = 0;
+
+  // Lock the DataPtr
+  virtual std::vector<xla::util::ExceptionCleanup> LockAsyncDatas(
```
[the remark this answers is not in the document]
JackCaoG: hmm, we already have `CreateAsyncDataHandlesCounter`. Each counter should only be locked once(during upload), rest of the test are just `barrier` call.

third_party/xla_client/xrt_computation_client.cc
```diff
@@ -295,6 +295,36 @@ ComputationClient::DataPtr XrtComputationClient::CreateDataPlaceholder(
...
+    XrtHandlePtr handle_ptr = std::make_shared<XrtHandle>(
+        DataHandleLocker::dummy_handle,
+        [this, device = tensors[i].device](int64_t handle) {
+          this->ReleaseXrtData(device, handle);
+        },
+        /*async=*/true);
+    results[i] = std::make_shared<XrtData>(this, tensors[i].device,
+                                           tensors[i].shape, handle_ptr);
+  }
+  CreateAsyncDataHandlesCounter()->AddValue(results.size());
```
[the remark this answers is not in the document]
JackCaoG: `ReleaseXrtData` call is a lambda function that will be called when Async data being garbage collected. check https://github.com/pytorch/xla/blob/9ddc27d88c1e974a5fe3261acbf53101d43238a8/third_party/xla_client/xrt_computation_client.h#L47

third_party/xla_client/xrt_computation_client.cc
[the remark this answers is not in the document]
JackCaoG: This function should only be called for the async case. So datas should always be non-empty. 

For the normal blocking upload case, `TransferToServer` at line 328 should be called.

torch_xla/csrc/tensor_util.cpp
```diff
@@ -18,14 +18,56 @@
...
 namespace torch_xla {
 namespace {
 
+struct DataAsync {
+  std::vector<xla::ComputationClient::TensorSource> source_tensors;
+  std::vector<xla::ComputationClient::DataPtr> async_datas;
+  std::vector<xla::util::ExceptionCleanup> handle_unlockers;
+};
+
+void TransferToServerAsync(std::shared_ptr<DataAsync> async,
```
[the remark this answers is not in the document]
JackCaoG: we already have metric on the caller of this function(`TensorToXlaData` and `TensorToXlaData`) which might be a better measurement of the total e2e transfer time in the async upload case.

torch_xla/csrc/tensor_util.cpp
JackCaoG: Hmm, the reason we set it to 1 is that there is only one function we expect to run async, which is the `populate_fn` below. What kind of defined paramter are you thinking about?

Other usecase of of this function are in https://github.com/pytorch/xla/blob/e0eb2a7bbc1e7dcefa4d83c88548a9dad7c3a387/third_party/xla_client/xrt_computation_client.cc#L361

torch_xla/csrc/tensor_util.cpp
JackCaoG: We expect one async data to come out of this async call because this function only take one tensor. If we created an var called `num_tensor_upload` it might be misleading because we always expect one in this case.

torch_xla/csrc/tensor_util.cpp
JackCaoG: sure, this function currently will never be called with  `transfer_async=True` because we only do that for seed upload. We can add a counter for future use through.

JackCaoG:
In https://github.com/pytorch/xla/issues/3352, user reported that by making rng seed sending async, it is possible to have some negative speed impact.  My analysis is https://github.com/pytorch/xla/issues/3352#issuecomment-1041079612. I think I will make RNG seed sending async only with an env var `XLA_TRANSFER_SEED_ASYNC=1` hence we won't change the default behavior. @aws-rhsoln FYI.

torch_xla/csrc/tensor_util.cpp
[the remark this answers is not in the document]
miladm: Correct, I am thinking about exploring more a drilled down measure of performance across our code stack. Hence, this suggestion.

torch_xla/csrc/tensor_util.cpp
miladm: Something like this:

`#define NUM_ASYNC_FUNCTIONS 1`

JackCaoG: Maybe something like
```
std::make_shared<xla::util::MultiWait>(/*num_async_func=*/1)
```
?

torch_xla/csrc/tensor_util.cpp
miladm: If this is always expected to be `1` without an exception, then the input parameter is redundant. Otherwise, resisting to hardcode numbers is my preferred way of passing information. I gave a suggestion on what `define` can look like in the previous comment.

torch_xla/csrc/tensor_util.cpp
miladm: Thanks.

miladm:
@JackCaoG how are we unit-testing this code?

JackCaoG:
> @JackCaoG how are we unit-testing this code?

Since I am going to make it control by an env var, I am going to add a line in `run_test.sh` and run our `test_operations.py` with this. I should also just add a unit test to `test_operations.py` to test the rng seed..

torch_xla/csrc/tensor_util.cpp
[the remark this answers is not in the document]
JackCaoG: done

torch_xla/csrc/tensor_util.cpp
JackCaoG: updated to `std::make_shared<xla::util::MultiWait>(/*num_wait=*/1);`

torch_xla/csrc/tensor_util.cpp
JackCaoG: updated

torch_xla/csrc/tensor_util.cpp
JackCaoG: oh wait, this function is just a mask and will call the `TensorToXlaData` with a different signature. We already have metrics there

JackCaoG:
@miladm I am planning to land this change tmr once all test passed and backport to 1.11. Let me know if there is anything else you want me to change.

miladm:
Thanks @JackCaoG

PR #3299, opened 2022-01-14
Update numpy version

JackCaoG:
Head currently failed with `RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd`. Update numpy version


JackCaoG:
It might be better to ping to a specified version, instead of always chasing for nightly.

miladm (approved):

PR #3306, opened 2022-01-19
Enable eager execution

aws-rhsoln:
ML scientists during their initial model development phase tend to use tools like PDB to debug their models. They would step through their models and print arbitary tensors to check their values and see if they look correct. Currently, if the user wants to debug issues in their model by printing output of intermediate layers or use debugger like pdb to step and investigate the tensors, the users incur an expensive intermediate graph compilation and execution. Adding the ability to run operations eagerly allows users to debug operations before running their expensive training jobs. Consider the example
```
import pdb
pdb.set_trace()
dev = xm.xla_device()
linear1 = torch.nn.Linear(10,30).to(dev)
linear2 = torch.nn.Linear(30,20).to(dev)
linear3 = torch.nn.Linear(20,10).to(dev)
linear4 = torch.nn.Linear(10,1).to(dev)
inp = torch.randn(2,10).to(dev)
output1 = linear1(inp)
output2 = linear2(output1)
PDB ---> xm.mark_step() # Users puts a xm.mark_step in pdb to investigate tensors 
output3 = linear3(output2)
output4 = linear4(output3)
xm.mark_step()
print(output4.to('cpu'))
print(output3.to('cpu')) 
```
In the above example, there are two graphs, both having a graphsize of 2. In larger models these intermediate graphsizes can be large and incur high compile and execution times. Also, in the next iteration if the user puts the mark_step at different location it would result in a different graph resulting in a cache miss. This would force the user to have fixed breakpoints, thereby putting a constraint. If each op is compiled and executed independently, the position of breakpoint doesn't matter anymore and there would be fixed number of compilations and executions. Moreover, with per op execution, the chances of hitting the cache increases as the layers in large models keep on repeating, thereyby avoiding a graph compile. With the `xrt_server` preserving the compilation cache across training runs, compiling per op can result in reusable cache, thereby further reducing the compilation time during debug.
 

## Pitch

We would like users to be able to enable eager execution by calling an api. This api needs to be called before any xla_tensor is created. The api is similar to what tensorflow had in 1.15 for enabling eager execution.

```
import os

dev = xm.xla_device()
# Enabling the eager exection. This API needs to be called before
# any XLA tensor is created. The rest of the script can remain 
# unchanged.
xm.enable_eager_execution() 
linear1 = torch.nn.Linear(10,30).to(dev)
linear2 = torch.nn.Linear(30,20).to(dev)
linear3 = torch.nn.Linear(20,10).to(dev)
linear4 = torch.nn.Linear(10,1).to(dev)
inp = torch.randn(2,10).to(dev)
output1 = linear1(inp)
output2 = linear2(output1)
PDB ---> output1.to('cpu')
output3 = linear3(output2)
output4 = linear4(output3)
PDB ---> output4.to('cpu')
xm.mark_step()
```
The number of graph compilation and executions remain the same and does not depend on where the `mark_step` is inserted.

Note: Doing per-op execution would result in higher e2e execution time when compared to lazy mode, however for initial development when the number of tensor prints by the user is going to be high, not doing intermediate graph compile should offset this time.


JackCaoG:
Thanks for contributing. I didn't look at the change in detail yet but what you are trying to achieved is very similar to `OpByOp` execution model we already support. Take a look at https://github.com/pytorch/xla/blob/master/TROUBLESHOOTING.md#environment-variables.

What `OpByOp` does is to skip the optimization when doing the compilation and execute the program in a op by op manner. Compilation time will be very fast but execution will be slower.

aws-rhsoln:
> Thanks for contributing. I didn't look at the change in detail yet but what you are trying to achieved is very similar to `OpByOp` execution model we already support. Take a look at https://github.com/pytorch/xla/blob/master/TROUBLESHOOTING.md#environment-variables.
> 
> What `OpByOp` does is to skip the optimization when doing the compilation and execute the program in a op by op manner. Compilation time will be very fast but execution will be slower.

Yeah it is similar, except that this would help users execute eagerly which they are used to doing in frameworks like PT. The issue with OpByOp mode is that it would break the lazily collected graph into ops when the user wants to print or when a xm.mark_step is encountered. This would result in re-executions in cases like these:
```
import os

os.environ["XLA_GET_TENSORS_OPBYOP"] = "1"
os.environ["XLA_SYNC_TENSORS_OPBYOP"] = "1"
dev = xm.xla_device()
linear1 = torch.nn.Linear(10,30).to(dev)
linear2 = torch.nn.Linear(30,20).to(dev)
linear3 = torch.nn.Linear(20,10).to(dev)
linear4 = torch.nn.Linear(10,1).to(dev)
inp = torch.randn(2,10).to(dev)
output1 = linear1(inp)
output2 = linear2(output1)
output3 = linear3(output2)
output4 = linear4(output3)
PDB ---> output4.to('cpu')
PDB ---> output3.to('cpu')
```
It would cut 2 graphs, one having 8 nodes (each linear layer creats 2 nodes) and other having 6 nodes and each graph would then be executed op by op. This would result in duplicate executions.

JackCaoG:
I felt like what you want is essentially least amount of wait time when user want to inspect a tensor value. Can we just spawn a sub process that call `xm.mark_step()` every 0.x seconds and set the execution mode to `OpByOp`. This way it will actively execute the pending graph.

In LTC  @wconstab  has plan to make `Lazy` a mode and `eager` would just means native pytorch eager mode. I don't think this pr fit our long term vision and it adds overhead in our critical path (tensor creation). I am also a bit unsure about the how much user experience improvement this pr will bring. PyTorch/XLA program can emit slightly different result if the kernel is not fused and running program in this `eager` mode will for sure damaged the performance and memory usage by a lot.

torch_xla/csrc/tensor.cpp
```diff
@@ -622,6 +632,10 @@ void XLATensor::SetIrValue(ir::Value ir_value) {
     AssignIrValue(std::move(ir_value));
     TryLimitGraphSize();
   }
+  if (ExecutionContext::Get()->is_eager_execution() && !copy_from_aten_tensor) {
```
JackCaoG, line 635: a bit confuse about `copy_from_aten_tensor`, why not just check `data()->view` if the intention is not to sync the view tensor?

torch_xla/csrc/tensor.h
```diff
@@ -24,6 +24,38 @@
 
 namespace torch_xla {
 
+class ExecutionContext {
```
JackCaoG, line 27: As discussed this should be a debug only mode. Would it make sense to check a simple `static bool debug_egaer_mode` within a function? I want to avoid introducing new Context class unless we have to.

JackCaoG (changes requested):

torch_xla/csrc/tensor.cpp
```diff
@@ -274,7 +274,13 @@ class XLATensor::DeviceContextArena {
 
  public:
   static DeviceContextArena* Get() {
-    static DeviceContextArena* arena = new DeviceContextArena();
+    static DeviceContextArena* arena = []() {
+      auto* ret = new DeviceContextArena();
+      // Setting a flag in ExecutionContext to alert that the device
+      // context is created and the execution mode cannot be changed now
+      ExecutionContext::Get()->set_device_context_created();
```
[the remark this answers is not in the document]
aws-rhsoln: we can use an env variable. The only con would be: It introduces another env variable for users to be aware of. If that's okay, I can change it to using an env variable instead of api call.

torch_xla/csrc/tensor.cpp
aws-rhsoln: The reason for the `copy_from_aten_tensor` was to avoid a sync that happens when a copy happens via this [path](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/aten_xla_type.cpp#L414) . That is when an aten tensor is copied to xla tensor. That is not an aten op, but just a copy and hence avoiding a sync here.

torch_xla/csrc/tensor.h
aws-rhsoln: Yeah, if go to using an env variable, it would make sense to just use a simple static bool and avoid the api thing and context creation.

wconstab:
I think this approach would work fine for lazy-tensor-core too.  We could discuss on the lazy_tensor design doc what kind of user api (env var or otherwise) we want to shoot for.  Env var seems simplest for starters for torch-xla.

torch_xla/csrc/tensor.cpp
JackCaoG: I think there are two cases here
1. the copied IR is a `device_data`, this is most likely the case since `eager_debug_mode` is on. Every IR will be materized to a `device_data`(pretty much just means a data on the device side). In this case we can just check the IR type
2. The copied IR is not a `device_data`. It is very unlikely but in this case we actually want to execute this ir.

I think it is better to check if the `IR` is a `device_data` instead of `copy_from_aten_tensor`.

Device Data reference https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/device_data.h

aws-rhsoln: Makes sense. Will make the change

torch_xla/csrc/tensor.cpp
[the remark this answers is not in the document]
aws-rhsoln: I think we can keep it as an api for the following reasons:
1. One less env variable that the user needs to be aware of.
2. In future, if we want to provide the flexibility to switch between lazy and eager_debug in the middle of training runs, it can be made possible with this design.
3. It would be synonymous to the tensorflow 1.15 api for enable eager execution.
Let me know if this sounds right.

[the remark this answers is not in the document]
JackCaoG: I am fine with it being a API, but I don't find it useful to inject it to the DeviceContext. I can't think of a case where we only want to enable debug mode on a certain device. It could just be a process wide static boolean variable.

IMO it is a debugging flag we can enable, so I am hesitant to make it a mode(from the code structure pov). TF eager mode AFAIK has some reasonable performance and people can run their model eagerly. Eager debug mode this pr propsed is a purely for debugging. It shouldn't be in the same level as Lazy.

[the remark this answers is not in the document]
aws-rhsoln: The DeviceContextArena here is only for alerting the ExecutionContext class that the contxt arena has been created and any further attempt to change the execution mode will result in an error. This is just one mechanism to alert the user that they need to set the debug mode before the tensor creation.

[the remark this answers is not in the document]
JackCaoG: Isn't this with your vision of `provide the flexibility to switch between lazy and eager_debug in the middle of training runs`? If we want to guarantee this being set in the beginning, imo env var is a better approach.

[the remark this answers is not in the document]
aws-rhsoln: Yeah that is the vision and wanted to keep that door open. Since, we haven't felt the need to do that rn, kept a check to make sure that the setting of debug mode doesn't happen after tensor creation.
Creating an env variable would be the easiest way rn, but it would make it difficult to enable and disable the debug mode between training runs in future.
Moreover, it adds another env variable which I think might not be a great idea :)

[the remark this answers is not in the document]
aws-rhsoln: The vision was to have the switch sometime in the future. However, didn't have a strong reason yet to have that switch. Hence, was trying to make sure that the change doesn't happen after the tensors are registered. 
I agree the env variable would be a better and simpler approach if the purpose is to just set it at the beginning. Hence, have made the change to set a static variable that reads from an env variable.

JackCaoG:
@aws-rhsoln There seems to be still some kind of merge conflict

aws-rhsoln:
> @aws-rhsoln There seems to be still some kind of merge conflict

Will try to resolve it and send in a revision

aws-rhsoln:
Ran the BERT-large model for 10 iterations. Here are the metrics:
```
EAGER DEBUG MODE:
Metric: XrtCompile
  TotalSamples: 358
  Accumulator: 28s857ms591.389us
  Mean: 078ms811.708us
  StdDev: 420ms665.013us
  Rate: 3.69262 / second
  Percentiles: 25%=022ms367.483us; 50%=029ms283.217us; 80%=079ms168.842us; 90%=089ms117.083us; 95%=094ms097.250us; 99%=541ms433.761us
Metric: XrtExecute
  TotalSamples: 68993
  Accumulator: 03m10s751ms891.251us
  Mean: 004ms876.244us
  StdDev: 111ms658.587us
  Rate: 200.191 / second
  Percentiles: 25%=193.963us; 50%=214.762us; 80%=261.520us; 90%=548.940us; 95%=001ms035.784us; 99%=004ms343.308us
  
  LAZY MODE:
Metric: XrtCompile
  TotalSamples: 2
  Accumulator: 16m56s565ms618.863us
  Mean: 08m58s782ms309.432us
  StdDev: 19s235ms806.571us
  Rate: 0.00393881 / second
  Percentiles: 25%=08m39s548ms502.861us; 50%=08m17s017ms116.002us; 80%=08m17s017ms116.002us; 90%=08m17s017ms116.002us; 95%=08m17s017ms116.002us; 99%=08m17s017ms116.002us
Metric: XrtExecute
  TotalSamples: 10
  Accumulator: 02m32s168ms546.572us
  Mean: 09s217ms754.657us
  StdDev: 356ms037.205us
  Rate: 0.0171977 / second
  Percentiles: 25%=09s051ms592.705us; 50%=09s182ms245.882us; 80%=10s600ms077.524us; 90%=10s037ms299.425us; 95%=10s037ms299.425us; 99%=10s037ms299.425us
```

For 2-layer BERT:
```
EAGER DEBUG MODE:
Metric: XrtCompile
  TotalSamples: 292
  Accumulator: 11s296ms214.594us
  Mean: 039ms685.666us
  StdDev: 053ms055.274us
  Rate: 9.62879 / second
  Percentiles: 25%=021ms109.436us; 50%=027ms160.091us; 80%=039ms982.515us; 90%=051ms787.316us; 95%=086ms160.113us; 99%=355ms206.725us
Metric: XrtExecute
  TotalSamples: 8537
  Accumulator: 39s930ms102.810us
  Mean: 005ms704.853us
  StdDev: 042ms841.875us
  Rate: 164.159 / second
  Percentiles: 25%=220.880us; 50%=306.633us; 80%=002ms632.986us; 90%=007ms759.590us; 95%=011ms172.809us; 99%=047ms932.984us

LAZY MODE:
Metric: XrtCompile
  TotalSamples: 2
  Accumulator: 19s739ms186.742us
  Mean: 09s370ms593.371us
  StdDev: 371ms599.849us
  Rate: 0.175261 / second
  Percentiles: 25%=09s999ms993.522us; 50%=10s740ms193.220us; 80%=10s740ms193.220us; 90%=10s740ms193.220us; 95%=10s740ms193.220us; 99%=10s740ms193.220us
Metric: XrtExecute
  TotalSamples: 10
  Accumulator: 13s060ms102.048us
  Mean: 01s306ms010.205us
  StdDev: 061ms283.828us
  Rate: 0.453409 / second
  Percentiles: 25%=01s272ms396.594us; 50%=01s282ms046.295us; 80%=01s342ms692.091us; 90%=01s473ms744.644us; 95%=01s473ms744.644us; 99%=01s473ms744.644us
```

test/run_tests.sh
```diff
@@ -60,6 +60,14 @@ function run_dynamic {
   fi
 }
 
+function run_eager_debug {
+  echo "Running in Eager Debug mode: $@"
+  IR_DEBUG=ir_debug_file
+  XLA_USE_EAGER_DEBUG_MODE=1 XLA_IR_DEBUG=1 XLA_SAVE_TENSORS_FILE=$IR_DEBUG run_test "$@"
```
JackCaoG, line 66: let's remove `XLA_IR_DEBUG=1 XLA_SAVE_TENSORS_FILE=$IR_DEBUG`, we can skip the `test_pow_constant` in eager debug mode

test/test_operations.py
```diff
@@ -1768,7 +1768,17 @@ def test_binaryop_order(self):
   def test_pow_constant(self):
     xla_device = xm.xla_device()
     t1 = torch.pow(torch.tensor([2.0, 3.0], device=xm.xla_device()), 5)
-    hlo_text = torch_xla._XLAC._get_xla_tensors_text([t1])
+    if os.environ.get('XLA_USE_EAGER_DEBUG_MODE'):
```
JackCaoG, line 1771: I think we can skip this test if `XLA_USE_EAGER_DEBUG_MODE` is on.

torch_xla/csrc/tensor.h
```diff
@@ -100,6 +100,26 @@ class XLATensor {
   // Applies the queue of operations in preparation for using the data.
   void ApplyPendingGraph();
 
+  // This method just Syncs the tensors passed as argument. This method is
```
JackCaoG, line 103: nit `Syncs` -> `syncs`

JackCaoG (commented):
Mostly lgtm, minor nits

torch_xla/csrc/tensor.cpp
```diff
@@ -1671,4 +1683,17 @@ xla::uint64 XLATensor::GetRunningSeed(const Device& device) {
   return DeviceContextArena::Get()->GetRunningSeed(device);
 }
 
+bool XLATensor::UseEagerDebugMode() {
```
JackCaoG, line 1686: hmm.. this doesn't needs to be a class method. I don't want to block you because of this but maybe I will fix it later.

JackCaoG (approved):
Thanks @aws-rhsoln !

PR #3319, opened 2022-01-28
disable TestNuclearNorm

JackCaoG:
disable the test to unblock pytorch head
https://github.com/pytorch/pytorch/issues/71964


malfet (approved):
LGTM, but arent `DISABLED_` prefix in front of the test name would achieve the same?

JackCaoG:
@malfet you are right.. I am planning to fix it tmr so it should be fine 😄

PR #3320, opened 2022-01-28
Output the same dtype for aten::div.Scalar as input if input is float

byronyi:
So lowering of `div.Scalar` is consistent with `div.Tensor`, which also uses input dtype if it is floating point:

https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor_methods.cpp#L1154-L1164


JackCaoG (approved):
Thanks @byronyi !

PR #3332, opened 2022-02-02
Disable failing tests on TPU

wonjoolee95:
We've had op tests failing for a while in TPU. Since our tests are always failing, it's hard to distinguish new potential bugs coming in. This PR attempts to make our op tests in TPU green again by disabling currently failing tests.

With these disabled tests, I can confirm that these tests are correctly getting skipped while testing on TPU. 

Issue tracking the disabled tests: https://github.com/pytorch/xla/issues/3333.

Pending: run all tests via `test/run_tests.sh`.

### DIFF
--- a/test/pytorch_test_base.py
+++ b/test/pytorch_test_base.py
@@ -360,6 +360,18 @@ DISABLED_TORCH_TESTS_TPU_ONLY = {
         'test_EmbeddingBag_per_sample_weights_and_offsets_xla',  # server side crash
         'test_upsamplingBilinear2d_xla',  # precision
         'test_upsamplingNearest2d_xla',  # precision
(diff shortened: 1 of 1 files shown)


JackCaoG:
@wonjoolee95 Are these enough to unblock our TPU op tests?

wonjoolee95:
> @wonjoolee95 Are these enough to unblock our TPU op tests?

Now it should be enough :) All passing at least on my TPUVM.

JackCaoG (approved):
Wow.. we need to someday revisit these failures.

wonjoolee95:
> Wow.. we need to someday revisit these failures.

Yeah, definitely. I'll merge these in now without waiting for the CircleCI since they only impact the TPU tests.

PR #3335, opened 2022-02-03
Lower `torch.linspace()`

### LINKED ISSUE #3327: Lower `linspace`
## 🚀 Feature

Lower [torch.linspace](https://pytorch.org/docs/stable/generated/torch.linspace.html)

<details>
<summary>Starting point from @JackCaoG:</summary>
<br>

```
diff --git a/torch_xla/csrc/aten_xla_type.cpp b/torch_xla/csrc/aten_xla_type.cpp
index c6621ab1..716a44bc 100644
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1789,6 +1789,25 @@ at::Tensor XLANativeFunctions::lerp(const at::Tensor& self,
       bridge::GetXlaTensor(self), bridge::GetXlaTensor(end), weight));
 }
 
+static at::Tensor linspace(const at::Scalar& start, const at::Scalar& end,
+                           c10::optional<int64_t> steps,
+                           c10::optional<at::ScalarType> dtype,
+                           c10::optional<at::Layout> layout,
+                           c10::optional<at::Device> device,
+                           c10::optional<bool> pin_memory) {
+  XLA_FN_COUNTER("xla::");
+  if (IsDefined(layout) || IsDefined(device) || IsDefined(pin_memory) ||
+      !IsDefined(steps)) {
+    return at::native::call_fallback_fn<&xla_cpu_fallback,
+                                        ATEN_OP(linspace)>::call(start, end,
+                                                                 steps, dtype,
+                                                                 layout, device,
+                                                                 pin_memory);
+  }
+  return bridge::AtenFromXlaTensor(XLATensor::linspace(
+      bridge::GetXlaTensor(start), bridge::GetXlaTensor(end), *steps, dtype));
+}
+
 at::Tensor XLANativeFunctions::log(const at::Tensor& self) {
   XLA_FN_COUNTER("xla::");
   return bridge::AtenFromXlaTensor(XLATensor::log(bridge::GetXlaTensor(self)));
diff --git a/torch_xla/csrc/tensor.h b/torch_xla/csrc/tensor.h
index a7069ca1..10202134 100644
--- a/torch_xla/csrc/tensor.h
+++ b/torch_xla/csrc/tensor.h
@@ -696,6 +696,9 @@ class XLATensor {
   static XLATensor lerp(const XLATensor& input, const XLATensor& end,
                         const at::Scalar& weight);
 
+  static XLATensor linspace(const XLATensor& start, const XLATensor& end,
+                            xla::int64_t steps, c10::optional<at::ScalarType> logical_element_type);
+
   static XLATensor log(const XLATensor& input);
 
   static XLATensor log_base(const XLATensor& input, ir::OpKind op, double base);
diff --git a/torch_xla/csrc/tensor_methods.cpp b/torch_xla/csrc/tensor_methods.cpp
index 292fca56..2590e7e4 100644
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -1605,6 +1605,14 @@ XLATensor XLATensor::lerp(const XLATensor& input, const XLATensor& end,
       ir::ops::Lerp(input.GetIrValue(), end.GetIrValue(), weight_val));
 }
 
+XLATensor linspace(const XLATensor& start, const XLATensor& end,
+                   xla::int64_t steps,
+                   c10::optional<at::ScalarType> logical_element_type) {
+  return input.CreateFrom(
+      ir::ops::Linspace(start.GetIrValue(), end.GetIrValue(), steps),
+      logical_element_type);
+}
+
 XLATensor XLATensor::log(const XLATensor& input) {
   // Here we explictly pass c10::nullopt as logical_element_type because
   // otherwise result will inherit the input's logical_element_type. In the
diff --git a/xla_native_functions.yaml b/xla_native_functions.yaml
index 1c340f6c..cc5d114c 100644
--- a/xla_native_functions.yaml
+++ b/xla_native_functions.yaml
@@ -158,6 +158,7 @@ supported:
   - leaky_relu_backward
   - lerp.Scalar
   - lerp.Tensor
+  - linspace
   - log
   - log1p
   - log2
```

</details>

will-cromar:
@JackCaoG FYI

- [ ] Add tests
- [ ] Handle `layout`, `pin_memory`, and null `steps` in `aten_xla_type.cpp`
- [ ] Handle `steps == 1` in `ops.cpp`

Fixes #3327

### DIFF
--- a/torch_xla/csrc/ops/ops.cpp
+++ b/torch_xla/csrc/ops/ops.cpp
@@ -21,6 +21,7 @@
 #include "torch_xla/csrc/ops/infer_output_shape.h"
 #include "torch_xla/csrc/ops/log_softmax_backward.h"
 #include "torch_xla/csrc/ops/permute.h"
+#include "torch_xla/csrc/ops/put.h"
 #include "torch_xla/csrc/ops/softmax_backward.h"
 #include "torch_xla/csrc/ops/sum.h"
 #include "torch_xla/csrc/pooling.h"
@@ -806,6 +807,16 @@ NodePtr BaddBmm(const Value& lhs, const Value& rhs, const Value& bias,
 NodePtr Lerp(const Value& start, const Value& end, const Value& weight) {
   ScopePusher ir_scope(at::aten::lerp.toQualString());
   return start + weight * (end - start);
+}
+
+NodePtr Linspace(const at::Scalar& start, const at::Scalar& end, const int64_t steps) {
+  // TODO: dangerous when steps == 1 !
+  double step = (end.toDouble() - start.toDouble()) / (steps - 1);
+  // Add extra step to end because arange upper bound is exclusive
+  NodePtr res = ARange(start, end.toDouble() + step, step, at::ScalarType::Float);
+  auto last_idx = ScalarOp(steps - 1, xla::PrimitiveType::S32);
+  auto end_val = ScalarOp(end, xla::PrimitiveType::F32);
+  return MakeNode<Put>(res, last_idx, end_val, /*accumulate=*/false);
 }
 
 NodePtr LogicalNot(const Value& input) {
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1789,6 +1789,13 @@ at::Tensor XLANativeFunctions::lerp(const at::Tensor& self,
       bridge::GetXlaTensor(self), bridge::GetXlaTensor(end), weight));
 }
 
+at::Tensor XLANativeFunctions::linspace(const at::Scalar& start, const at::Scalar& end, c10::optional<int64_t> steps, c10::optional<at::ScalarType> dtype, c10::optional<at::Layout> layout, c10::optional<at::Device> device, c10::optional<bool> pin_memory) {
+  XLA_FN_COUNTER("xla::");
+  // TODO: deal with layout and pin_memory
+  return bridge::AtenFromXlaTensor(XLATensor::linspace(
+    start, end, *steps, GetScalarTypeOrFloat(dtype), GetXlaDeviceOrCurrent(device)));
+}
+
 at::Tensor XLANativeFunctions::log(const at::Tensor& self) {
   XLA_FN_COUNTER("xla::");
   return bridge::AtenFromXlaTensor(XLATensor::log(bridge::GetXlaTensor(self)));
--- a/torch_xla/csrc/ops/ops.h
+++ b/torch_xla/csrc/ops/ops.h
@@ -226,6 +226,8 @@ NodePtr BaddBmm(const Value& lhs, const Value& rhs, const Value& bias,
 
 NodePtr Lerp(const Value& start, const Value& end, const Value& weight);
 
+NodePtr Linspace(const at::Scalar& start, const at::Scalar& end, const int64_t steps);
+
 NodePtr LogicalNot(const Value& input);
 
 NodePtr LogicalXor(const Value& input, const Value& other);
--- a/torch_xla/csrc/tensor.h
+++ b/torch_xla/csrc/tensor.h
@@ -696,6 +696,10 @@ class XLATensor {
   static XLATensor lerp(const XLATensor& input, const XLATensor& end,
                         const at::Scalar& weight);
 
+  static XLATensor linspace(const at::Scalar& start, const at::Scalar& end,
+                            const int64_t steps, at::ScalarType element_type,
+                            const Device& device);
+
   static XLATensor log(const XLATensor& input);
 
   static XLATensor log_base(const XLATensor& input, ir::OpKind op, double base);
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -1605,6 +1605,13 @@ XLATensor XLATensor::lerp(const XLATensor& input, const XLATensor& end,
       ir::ops::Lerp(input.GetIrValue(), end.GetIrValue(), weight_val));
 }
 
+XLATensor XLATensor::linspace(const at::Scalar& start, const at::Scalar& end,
+                              const int64_t steps, at::ScalarType element_type, const Device& device) {
+  return XLATensor::Create(
+    ir::ops::Linspace(start, end, steps),
+    device, element_type);
+}
+
 XLATensor XLATensor::log(const XLATensor& input) {
   // Here we explictly pass c10::nullopt as logical_element_type because
   // otherwise result will inherit the input's logical_element_type. In the
--- a/xla_native_functions.yaml
+++ b/xla_native_functions.yaml
@@ -158,6 +158,7 @@ supported:
   - leaky_relu_backward
   - lerp.Scalar
   - lerp.Tensor
+  - linspace
   - log
   - log1p
   - log2


torch_xla/csrc/ops/ops.cpp
```diff
@@ -809,6 +810,22 @@ NodePtr Lerp(const Value& start, const Value& end, const Value& weight) {
   return start + weight * (end - start);
 }
 
+NodePtr Linspace(const Value& start, const Value& end, const int64_t steps) {
```
JackCaoG, line 813: This will work, the only "issue" is that it is not creating a Node call `Linspace`, it is returning a combination of other IR nodes. When user dump the IR graph he might not realized this `Put` IR was a result of `Linspace` op.

This for the most part is fine, and we do this a lot in early days to speed up the op lowering process, but we are trying to move away from it for the better debuggility. Could you create a `Linspace` op under ops dir like [kth_value](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/kth_value.h) and add a lower function(which prettry much follow the same logic you have here, you might want to check how `Put` is lowered and call that lower function directly).

I think one of the reason you lower this way is to make use of the `Arange` Node. I think it is a good time for us to actually refactor that a bit too. We can refacotr https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/ops.cpp#L489-L550 into a helper function to return a `xla::Literal` and use that helper function in both `Arange` and `Linspace`

JackCaoG (changes requested):
Thanks @will-cromar ! mostly lgtm.

JackCaoG:
GPU test failure is a glitch, rebase should fix that.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -694,6 +694,24 @@ xla::XlaOp CreatePut(const Device& device, xla::XlaOp input, xla::XlaOp index,
   return XlaHelpers::DynamicReshapeAs(r1_scatter, input_shape);
 }
 
+xla::XlaOp CreateLinspace(const Device& device, xla::XlaOp start,
+                          xla::XlaOp end, int64_t steps) {
+  std::tie(start, end) = XlaHelpers::PromoteValues(start, end);
+  if (steps == 1) {
+    return BuildExpand(start, {1});
```
JackCaoG, line 701: question, if steps == 1, should we still promote the `start`? We should check pytorch behavior for this.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -694,6 +694,24 @@ xla::XlaOp CreatePut(const Device& device, xla::XlaOp input, xla::XlaOp index,
...
+  }
+  xla::XlaOp indices = xla::ConstantLiteral(
+      start.builder(), XlaHelpers::Range<int64_t>(0, steps, 1));
+
+  xla::XlaOp last_index = XlaHelpers::ScalarValue(
+      steps - 1, xla::PrimitiveType::S64, start.builder());
+  xla::XlaOp step_val = XlaHelpers::PromotedDiv(end - start, last_index);
```
JackCaoG, line 708: no need to call `PromotedDiv ` here. `start`, `end` and `last_index` should have the same type here.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -694,6 +694,24 @@ xla::XlaOp CreatePut(const Device& device, xla::XlaOp input, xla::XlaOp index,
...
+                          xla::XlaOp end, int64_t steps) {
+  std::tie(start, end) = XlaHelpers::PromoteValues(start, end);
+  if (steps == 1) {
+    return BuildExpand(start, {1});
+  }
+  xla::XlaOp indices = xla::ConstantLiteral(
+      start.builder(), XlaHelpers::Range<int64_t>(0, steps, 1));
+
+  xla::XlaOp last_index = XlaHelpers::ScalarValue(
+      steps - 1, xla::PrimitiveType::S64, start.builder());
```
JackCaoG, line 707: let's make `last_index` has the same type as `start` and `end` for the sake of type promotion.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -694,6 +694,24 @@ xla::XlaOp CreatePut(const Device& device, xla::XlaOp input, xla::XlaOp index,
...
 }
 
+xla::XlaOp CreateLinspace(const Device& device, xla::XlaOp start,
+                          xla::XlaOp end, int64_t steps) {
+  std::tie(start, end) = XlaHelpers::PromoteValues(start, end);
+  if (steps == 1) {
+    return BuildExpand(start, {1});
+  }
+  xla::XlaOp indices = xla::ConstantLiteral(
+      start.builder(), XlaHelpers::Range<int64_t>(0, steps, 1));
```
JackCaoG, line 704: Use `xla::ConvertElementType` to convert the `indices` to the same type as `start` and `end` for the sake of type promotion

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -694,6 +694,24 @@ xla::XlaOp CreatePut(const Device& device, xla::XlaOp input, xla::XlaOp index,
...
+
+  xla::XlaOp last_index = XlaHelpers::ScalarValue(
+      steps - 1, xla::PrimitiveType::S64, start.builder());
+  xla::XlaOp step_val = XlaHelpers::PromotedDiv(end - start, last_index);
+
+  xla::XlaOp res = XlaHelpers::PromotedMul(indices, step_val) + start;
```
JackCaoG, line 710: ditto

JackCaoG (commented):
Mostly lgtm

torch_xla/csrc/xla_lower_util.cpp
will-cromar: We use `last_index` as the index argument into `Put`, so it actually must be integral.

torch_xla/csrc/xla_lower_util.cpp
will-cromar: See above.

torch_xla/csrc/xla_lower_util.cpp
JackCaoG: when we refer to dtpye, we usually mean `at::ScalarType`. There are currently two type for a XLATensor.

`at::ScalarType` is the type we inherit/set from the pytorch frontend. It represents how pytorch think/want the type of certain tensor should be.

`xla::PrimitveType` are the actual type we store in the XLA device side. If user didn't set `at::ScalarType` explictly, we will convert the `xla::PrimitveType` to the corresponding `at::scalartype` if pytorch frontend ask for it.


For the most part these type should match, but for example if user do
```
a = torch.tensor(int32)
b = torch.tensor(int32)
c = torch.add(a, b, dtype = int64)
```

what we ended up doing is `xla::PrimitiveType` of `c` will still be `int32`(inferred by xla compiler) but the `ScalarType` are `int64`. We do this to make pytorch frontend happy.


What I want to say here is that `dtype` should not affect the xla compiler backend lowering logic. In this case we want to make the result of the `linspace`'s `xla::Primitivetype` respect the `start` and `end`. Instead of being promoitved to int64 due to index is of type int64.

will-cromar: I converted `start` and `end` to F32 when creating the `ir::Value`s in `tensor_methods.cpp`, since the PyTorch docs say `start` and `end` are taken as `float`s. If I kept the type of start and end as `int` and the float division result of `(end - start) / steps` was not the same as the int division result, I got a different output compared to `torch.linspace` without XLA. So PyTorch definitely promotes the types to floats internally.

 I mainly added the promotion as a sanity check to avoid a crash in case we change something at a higher level or another function calls `CreateLinspace`. It will be a no-op if you're actually calling `torch.linspace`. Since we're passing in start and end as floats, IIUC they can not be promoted to int64; index and steps will instead be promoted from int to float.

torch_xla/csrc/ops/linspace.cpp
```diff
@@ -0,0 +1,39 @@
...
+      steps_(steps) {}
+
+NodePtr Linspace::Clone(OpList operands) const {
+  return MakeNode<Linspace>(operands.at(0), operands.at(1), steps_);
+}
+
+XlaOpVector Linspace::Lower(LoweringContext* loctx) const {
+  xla::XlaOp start = loctx->GetOutputOp(operand(0));
+  xla::XlaOp end = loctx->GetOutputOp(operand(1));
+  return ReturnOp(CreateLinspace(loctx->device(), start, end, steps_), loctx);
```
JackCaoG, line 28: nit, `CreateLinspace ` -> `BuildLinspace` to match our usual format for the lowering helper function

JackCaoG (approved):
Thanks @will-cromar ! Minor naming convention nit. Otherwise LGTM. Feel free to merge after CI pass again.

PR #3345, opened 2022-02-04
TF version bump to resolve v2-nightly, libtpu-nightly runtime issues

yeounoh:
Pinning to the (more) recent TF version that works with `v2-nightly` and `libtpu-nightly`, in preparation for `1.11` release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ import torch
 base_dir = os.path.dirname(os.path.abspath(__file__))
 third_party_path = os.path.join(base_dir, 'third_party')
 
-_libtpu_version = '0.1.dev20220118'
+_libtpu_version = '0.1.dev20220128'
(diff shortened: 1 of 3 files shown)


JackCaoG:
If I remember correctly, this is to avoid the `TPUEmbedding` op failure? We still need to use `continuous` to make it work on TPUVM right?

yeounoh:
> If I remember correctly, this is to avoid the `TPUEmbedding` op failure? We still need to use `continuous` to make it work on TPUVM right?

Correct, `TPUEmbedding` and also `devcie access` issues. The `continuation` fix will have to happen outside of this PR: 1) if we modify XLA, re-pin Tensorflow; 2) if we don't modify XLA, then we will set the `no-continuation` flag in `v2-nightly`.

JackCaoG:
Let's wait a couple days and see if we can also fix other tf/libtpu issues before merging. My guess is that change might be in the libtpu end but we might want to update tf to match that.

JackCaoG (approved):

JackCaoG:
Merge this to unblock cloud build

PR #3353, opened 2022-02-09
Lower `torch._linalg_svd`

will-cromar:
@JackCaoG FYI

- Update `some` to `full_matrices` in `svd.cpp/h` to be consistent with `torch.linalg.svd*`
- Add `deprecated_svd` flag to `SVD` to determine output shape. See the [PyTorch docs](https://pytorch.org/docs/stable/generated/torch.svd.html) for an explanation of the differences between `torch.svd` (deprecated) and `torch.linalg.svd`.
- Fall back to CPU for complex numbers, since we need to return the conjugate transpose of `V` (i.e. `Vh`), but conjugate is not yet lowered.
- Re-enable `TestNuclearNorm`

### DIFF
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -847,7 +847,7 @@ TEST_F(AtenXlaTensorTest, TestSVD) {
 }
 
 TEST_F(AtenXlaTensorTest, TestLinalgSVD) {
-  static const int dims[] = {4, 7};
+  static const int dims[] = {0, 4, 7};
   for (auto m : dims) {
     for (auto n : dims) {
       torch::Tensor a =
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -852,10 +852,12 @@ TEST_F(AtenXlaTensorTest, TestLinalgSVD) {
     for (auto n : dims) {
       torch::Tensor a =
           torch::rand({m, n}, torch::TensorOptions(torch::kFloat));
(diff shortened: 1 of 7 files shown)


JackCaoG:
There is a failure on `triangular_solve`. To run the test, I usually commented out every entry except the one I want run. In you case it is https://github.com/pytorch/xla/blob/master/test/test_ops.py#L184 then
```
python test/test_ops.py
```
should work.

will-cromar:
I ran `python setup.py install` and `python test/test_ops.py` locally (without commenting anything out) and didn't see any failures. Will re-run CI.

will-cromar:
I'm not sure why I can't reproduce the CI error by running the test myself, but I can reproduce it this way:

```
A = torch.randn(0, 0, device=xm.xla_device())
print(torch._linalg_svd(A))

[...]
RuntimeError: Error while lowering: (f32[0,0]{1,0}, f32[0]{0}, f32[0,0]{1,0}) aten::svd, num_outputs=3, full_matrices=0, compute_uv=1, deprecated_svd=0
[...]
```

Dimensions [1, 0] and [0, 1] also lead to an error. The existing deprecated SVD function `torch.svd` has the same problem.

will-cromar:
Coming back to this PR. I wasn't able to completely remove all of the code to explicitly compute the expected shapes of `u` and `v`, since we need to compute it when `full_matrices=false` and when the input has a dimension of size 0. I combined that code in a helper `UVShapes` instead of computing the shapes in both `LowerSVD` and `NodeOutputShape`. This PR will also make `torch.svd` behave correctly with dimensions of size 0. Will mark ready for review once I confirm CI tests pass.

PR #3358, opened 2022-02-10
Update gather, scatter, and scatter_add ops to accept unique_indices parameter

wonjoolee95:
In an open PyTorch PR https://github.com/pytorch/pytorch/pull/70873/files, they have added a new parameter `unique_indices` into `gather`, `scatter`, and `scatter_add` ops. Updated these ops to receive this new parameter.

Following PyTorch https://github.com/pytorch/pytorch/pull/70873/files#diff-4e77adb023a4e15f75009c08294251dfe9c3b7bb6947e7f7b4609c65edb53b20, we will fall back to the CPU implementation if `unique_indices` is set to `True`. @JackCaoG, please let me know if you think this is okay.

With this change, build succeeds locally with PyTorch branched on the mentioned PR above.


JackCaoG:
Reading the upstream pr, seems like `unique_indices ` is only used to throw warnings. We need to do 2 things
1. Verify if xla `scatter` and `scatter_add` shares the same `nondeterministic` behavior when index is not unique (most likely yes)
2. Considering throwing the similar warning if needed.

wonjoolee95:
Thanks for the comment, @JackCaoG!

1. I can confirm that PT/XLA `scatter` and `scatter_add` are non-deterministic when indices are not unique. Both `scatter` and `scatter_add` call `xla::Scatter` -- https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_lower_util.cpp#L670. And according to XLA's documentation (https://www.tensorflow.org/xla/operation_semantics#scatter), `scatter` is not deterministic when indices are not unique:
```
The order in which updates are applied is non-deterministic. So, when multiple indices in updates refer to the same index in operand, the corresponding value in output will be non-deterministic.
```

2. I've went ahead and just threw a runtime error instead of a warning. I tried to follow PT's behavior (https://pytorch.org/docs/master/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms):
```
When enabled, operations will use deterministic algorithms when available, and if only nondeterministic algorithms are available they will throw a [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError) when called.
```

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1486,7 +1486,8 @@ at::Tensor XLANativeFunctions::frac(const at::Tensor& self) {
 
 at::Tensor XLANativeFunctions::gather(const at::Tensor& self, int64_t dim,
                                       const at::Tensor& index,
-                                      bool /* sparse_grad */) {
+                                      bool /* sparse_grad */,
+                                      bool /* unique_indices */) {
```
wonjoolee95, line 1490: PyTorch added this parameter to `gather` since their implementation of `gather` calls their `scatter`. However, PT/XLA does not. PT/XLA directly calls `xla::Gather`.

JackCaoG: hmm, does this means pt/xla gather can handle `non-unique-index`?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2876,38 +2877,53 @@ at::Tensor scatter_reduce_helper(const at::Tensor& self, int64_t dim,
...
 }
 
 at::Tensor XLANativeFunctions::scatter(const at::Tensor& self, int64_t dim,
                                        const at::Tensor& index,
-                                       const at::Scalar& value) {
+                                       const at::Scalar& value,
+                                       bool unique_indices) {
   XLA_FN_COUNTER("xla::");
+  XLA_CHECK(unique_indices)
+      << "scatter not have a deterministic implementation";
```
JackCaoG, line 2894: Can we make it a bit more explicit, like
```
Scatter will have a nondeterministic behavior when indices are not unqiue.
```

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2876,38 +2877,53 @@ at::Tensor scatter_reduce_helper(const at::Tensor& self, int64_t dim,
...
   return scatter_reduce_helper(self, dim, index, src, reduce);
 }
 
 at::Tensor XLANativeFunctions::scatter(const at::Tensor& self, int64_t dim,
                                        const at::Tensor& index,
                                        const at::Scalar& value,
-                                       c10::string_view reduce) {
+                                       c10::string_view reduce,
+                                       bool unique_indices) {
```
JackCaoG, line 2913: Do we also need to update our c++ test to pass `unique_indices=true`?

torch_xla/csrc/aten_xla_type.cpp
wonjoolee95: Good point, I was looking into this before a bit. According to XLA's documentation, there is no mention of non-deterministic behavior (https://www.tensorflow.org/xla/operation_semantics#gather). However, to be safe, I've also added the `XLA` check to `gather` in the new commit.

torch_xla/csrc/aten_xla_type.cpp
wonjoolee95: Good point, updated to:
> scatter will have a non-deterministic behavior when indices are not unqiue

torch_xla/csrc/aten_xla_type.cpp
wonjoolee95: Yep, updated in the new commit.

wonjoolee95:
The build succeeds locally with PyTorch based on the branch mentioned above. Also can confirm that all `gather` and `scatter`, `scatter_add` tests succeeds. These include:
```
source xlaCppTest.sh Gather
source xlaCppTest.sh Scatter
source xlaCppTest.sh ScatterR1
source xlaCppTest.sh ScatterR3
source xlaCppTest.sh ScatterBiggerSource
source xlaCppTest.sh ScatterScalar
source xlaCppTest.sh ScatterReduceAdd
source xlaCppTest.sh ScatterAdd
source xlaCppTest.sh ScatterAddInPlace
```

wonjoolee95:
Thanks for the comments again, @JackCaoG. Updated the reflect the comments.

One more question -- how do we usually go about merging these changes? I assume our CircleCI failure is expected because it's trying to pull the master PyTorch branch. 

When we're ready to merge this change on our side, should we message the PyTorch folks to say it's okay to merge their changes?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2849,7 +2852,7 @@ at::Tensor scatter_reduce_helper(const at::Tensor& self, int64_t dim,
     // TODO: implement scatter_mul
     return at::native::call_fallback_fn<
         &xla_cpu_fallback, ATEN_OP2(scatter, reduce)>::call(self, dim, index,
-                                                            src, *reduce);
+                                                            src, *reduce, true);
```
JackCaoG, line 2855: nit 
```
/* unique_indices=*/true
```

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2870,44 +2873,60 @@ at::Tensor scatter_reduce_helper(const at::Tensor& self, int64_t dim,
     return at::native::call_fallback_fn<
         &xla_cpu_fallback, ATEN_OP2(scatter, value_reduce)>::call(self, dim,
                                                                   index, value,
-                                                                  *reduce);
+                                                                  *reduce,
+                                                                  true);
```
JackCaoG, line 2877: ditto

JackCaoG (approved):
Thanks @wonjoolee95 !

JackCaoG:
Can you add a torch pin like https://github.com/pytorch/xla/pull/3303/files? The content of the pin are just uptream pr number. CI will checkout the corresponding upstream branch.

Remember to remove the pin before meging your pr (once pytorch once merged)

wonjoolee95:
> Can you add a torch pin like https://github.com/pytorch/xla/pull/3303/files? The content of the pin are just uptream pr number. CI will checkout the corresponding upstream branch.
> 
> Remember to remove the pin before meging your pr (once pytorch once merged)

Done! Now waiting for our tests to pass..

wonjoolee95:
Hmm, it seems like related PyTorch tests are failing because they are expecting a specific warning. In cases, they expect a warning and expect the op to continue.

JackCaoG:
We can disable the test if it is only expecting a pytorch specified warning message. We have done that in the past.

wonjoolee95:
> We can disable the test if it is only expecting a pytorch specified warning message. We have done that in the past.

The problem was that PyTorch expects a just "warning" message and expects the program to continue, but this change in PT/XLA throws a runtime error.

I've updated the PR to print a warning message instead of throwing a runtime error. This is how it'll look like:
```
>>> import torch
>>> import torch_xla
>>> device = "xla:0"
>>> x = torch.tensor([[True, True, True, True, True], [True, True, True, True, True]], device=device)
>>> res = torch.zeros(3, 5, dtype=torch.bool, device=device)
>>> res = res.scatter_add_(0, torch.tensor([[0, 1, 2, 0, 0], [2, 0, 0, 1, 2]], device=device), x)
2022-02-14 07:15:42.761410: W  365033 /usr/local/google/home/wonjoo/pytorch/xla/torch_xla/csrc/aten_xla_type.cpp:2961] scatter will have a non-deterministic behavior when indices are not unqiue
>>> res
tensor([[ True,  True,  True,  True,  True],
        [False,  True, False,  True, False],
        [ True, False,  True, False,  True]], device='xla:0')
```

wonjoolee95:
Disabled PyTorch tests (due to expecting PyTorch `non-deterministic` warning) include:
```
        'test_nondeterministic_alert_scatter_add_autograd_xla',  # expects non-deterministic warning
        'test_nondeterministic_alert_scatter_add_xla',  # expects non-deterministic warning
        'test_nondeterministic_alert_scatter_autograd_xla',  # expects non-deterministic warning
        'test_nondeterministic_alert_scatter_xla',  # expects non-deterministic warning
```

JackCaoG (approved):

wonjoolee95:
Currently, waiting for PyTorch PR to be merged https://github.com/pytorch/pytorch/pull/70873/files. Once this merges, I'll remove the torch pin and merge this PR.

wonjoolee95:
Upstream PR (https://github.com/pytorch/pytorch/pull/70873) got closed without merging, closing this.

PR #3359, opened 2022-02-10
Add all_gather_out and reduce_scatter_out

JackCaoG:
This is to enable us to support torch distributed api.

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -284,8 +284,8 @@ def xla_replication_devices(local_devices):
   if len(kind_devices) != len(local_devices):
     # Replication can only happen among all devices of one kind.
     raise RuntimeError(
-        'Cannot replicate if number of devices ({}) is different from {}'.
-        format(len(local_devices), len(kind_devices)))
+        'Cannot replicate if number of devices ({}) is different from {}'
+        .format(len(local_devices), len(kind_devices)))
   replication_devices = []
   for device in torch_xla._XLAC._xla_get_all_devices():
     xdev = parse_xla_device(device)
@@ -304,8 +304,8 @@ def unlazy(tensors):
   This API is for benchmarking, don't use it in real models.
 
   Args:
-    tensors (List[torch.Tensor]): List of `torch.Tensor`s to materialize.
-      For each Tensor `t` in the list, `t.device` must be an `xla` device.
+    tensors (List[torch.Tensor]): List of `torch.Tensor`s to materialize. For
+      each Tensor `t` in the list, `t.device` must be an `xla` device.
   """
   torch_xla._XLAC._xla_sync_multi(tensors, devices=[], wait=True)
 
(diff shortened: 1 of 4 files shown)


torch_xla/core/xla_model.py
```diff
@@ -600,10 +604,16 @@ def all_gather(value, dim=0, groups=None):
...
-  result = torch_xla._XLAC._xla_all_gather(value, token, dim, shard_count,
-                                           groups or [])
-  devctx.all_reduce_token = result[1]
-  return result[0]
+  if output != None:
+    new_token = torch_xla._XLAC._xla_all_gather_out(output, value, token, dim,
+                                                    shard_count, groups or [])
+    devctx.all_reduce_token = new_token
+    return output
+  else:
```
hjm-aws, line 612: nit: This `else:` is not needed because of the `return` on the previous line.

torch_xla/core/xla_model.py
```diff
@@ -684,17 +696,28 @@ def reduce_scatter(reduce_type,
...
-  return result[0]
+  if output != None:
+    new_token = torch_xla._XLAC._xla_reduce_scatter_out(reduce_type, output,
+                                                        input, token, scale,
+                                                        scatter_dim,
+                                                        shard_count, groups or
+                                                        [])
+    devctx.all_reduce_token = new_token
+    return output
+  else:
```
hjm-aws, line 715: nit: This `else:` is not needed because of the `return` on the previous line.

torch_xla/csrc/tensor.h
```diff
@@ -225,6 +225,12 @@ class XLATensor {
       double scale, int64_t scatter_dim, int64_t shard_count,
       std::vector<std::vector<int64_t>> groups);
 
+  static ir::Value reduce_scatter_out(XLATensor& output, const XLATensor& input,
```
hjm-aws, line 228: Do you need to follow Google C++ style here? https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs requires you to put output params at the end of the list.

hjm-aws (approved):

JackCaoG:
@hjm-aws I will merge this pr once all test passed.

miladm (approved):
Thanks @JackCaoG!
LGTM. Some nits are below.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -212,6 +212,19 @@ std::pair<at::Tensor, std::shared_ptr<ir::Value>> ReduceScatter(
       std::make_shared<ir::Value>(new_token));
 }
 
+std::shared_ptr<ir::Value> ReduceScatterOut(
+    const std::string& reduce_type, at::Tensor& output, const at::Tensor& input,
+    const std::shared_ptr<ir::Value>& token, double scale, int64_t scatter_dim,
+    int64_t shard_count,
+    const std::vector<std::vector<int64_t>>& replica_groups) {
+  XLATensor out = bridge::GetXlaTensor(output);
```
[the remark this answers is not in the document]
hjm-aws: I came here for inspiration of how to set the `ir::Value` for an output tensor. Here I agree with Milad: I don't believe we need to leave the `out` variable outside of the `reduce_scatter_out` call. Leaving `out` outside doesn't do anything -- `out` will be destructed immediately after the return statement anyway.

`XLATensor` is a holder for an `shared_ptr<Data>`, so it can be used as a temp. `XLATensor::SetIrValue` (and any other non-const method on `XLATensor`) manipulates the `shared_ptr<Data>` member inside `XLATensor`, so passing `bridge::GetXlaTensor(output)` is fine.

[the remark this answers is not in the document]
hjm-aws: NVM, I found the compiler doesn't like it :D

PR #3373, opened 2022-02-15
Force tzdata installtion to be non-interactive

JackCaoG:


JackCaoG:
Doc push is only enabled on master branch. merge this pr after linter check passed

PR #3375, opened 2022-02-15
Fix doc push

JackCaoG:


JackCaoG:
Verified it on circleCI that doc push no longer hang, merge to master to unblock doc push

PR #3339, opened 2022-02-04
Add XLA backend for torch.distributed

hjm-aws:

PR #3378, opened 2022-02-16
Add XLA backend for torch.distributed (#3339)

JackCaoG:
Cherry picking https://github.com/pytorch/xla/pull/3339 to the release 1.11 branch

### DIFF
--- a/test/test_torch_distributed_all_gather_xla_backend.py
+++ b/test/test_torch_distributed_all_gather_xla_backend.py
@@ -1,0 +1,37 @@
+import os
+import sys
+import torch
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla.distributed.xla_backend
+import torch.distributed as dist
+
+
+def _mp_fn(index):
+  device = xm.xla_device()
+  if xm.xla_device_hw(device) in ('TPU', 'GPU'):
+    world_size = xm.xrt_world_size()
+    rank = xm.get_ordinal()
+
+    dist.init_process_group('xla', world_size=world_size, rank=rank)
+
+    input = torch.ones((2, 3)) * rank
+    outputs = [torch.zeros_like(input)] * world_size
+    xinput = input.to(device)
+    xoutputs = [o.to(device) for o in outputs]
+    xoutput0 = xoutputs[0]
+    dist.all_gather(xoutputs, xinput)
+    for i, o in enumerate(xoutputs):
+      expected = torch.ones((2, 3)) * i
+      assert torch.all(o.cpu() == expected), f'{o} != {expected}'
+    expected0 = torch.zeros_like(input)
+    assert torch.all(xoutput0.cpu() == expected0), f'{xoutput0} != {expected0}'
+  else:
+    print(
+        'Default device {} is not a TPU or GPU device'.format(device),
+        file=sys.stderr)
+
+
+if __name__ == '__main__':
+  xmp.spawn(_mp_fn, args=())
--- a/test/test_torch_distributed_multi_all_reduce_xla_backend.py
+++ b/test/test_torch_distributed_multi_all_reduce_xla_backend.py
@@ -1,0 +1,38 @@
+import os
+import sys
+import torch
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla.distributed.xla_backend
+import torch.distributed as dist
+
+
+def _mp_fn(index):
+  device = xm.xla_device()
+  if xm.xla_device_hw(device) in ('TPU', 'GPU'):
+    world_size = xm.xrt_world_size()
+    rank = xm.get_ordinal()
+
+    dist.init_process_group('xla', world_size=world_size, rank=rank)
+
+    num_all_reduces = 20
+    xinputs_list = []
+    for i in range(num_all_reduces):
+      inputs = torch.ones((2, 3)) * i
+      xinputs = inputs.to(device)
+      xinputs_list.append(xinputs)
+      dist.all_reduce(xinputs)
+    for i in range(num_all_reduces):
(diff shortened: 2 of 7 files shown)


miladm (approved):
LGTM - Thanks @JackCaoG 
(pending the pass of currently running CI tests)

test/test_torch_distributed_all_gather_xla_backend.py
```diff
@@ -0,0 +1,37 @@
...
+    outputs = [torch.zeros_like(input)] * world_size
+    xinput = input.to(device)
+    xoutputs = [o.to(device) for o in outputs]
+    xoutput0 = xoutputs[0]
+    dist.all_gather(xoutputs, xinput)
+    for i, o in enumerate(xoutputs):
+      expected = torch.ones((2, 3)) * i
+      assert torch.all(o.cpu() == expected), f'{o} != {expected}'
+    expected0 = torch.zeros_like(input)
+    assert torch.all(xoutput0.cpu() == expected0), f'{xoutput0} != {expected0}'
```
yeounoh, line 29: Not sure if this is needed?

test/test_torch_distributed_multi_all_reduce_xla_backend.py
```diff
@@ -0,0 +1,38 @@
+import os
```
yeounoh, line 1: Probably not easy in the current test structure, but would be more consistent if we can group all_reduce tests in a single test file using unit test framework. Also, it might be better to have a separate `distributed` test folder under `test/`.

test/test_torch_distributed_all_gather_xla_backend.py
JackCaoG: we need to make sure after allgather tensor has the expected value.

test/test_torch_distributed_multi_all_reduce_xla_backend.py
JackCaoG: yea.. we should eventually have a separate repo for distributed tests

torch_xla/distributed/xla_backend.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+_register_xla_backend()
+
+
+class ProcessGroupXla(ProcessGroup):
+  '''ProcessGroup for XLA devices. See ProcessGroup for doc.
+
+    Here we are implementing only a Python subclass. For implementing a
+    C++/Python extension, see
```
yeounoh, line 29: Do we have the C++ binding in a separate PR?

torch_xla/distributed/xla_backend.py
```diff
@@ -0,0 +1,290 @@
...
+          scatter_dim=0,
+          shard_count=shard_count,
+          scale=1,
+          groups=groups,
+          output=output_tensor)
+
+    return WorkXla(output_tensors)
+
+  # Call site:
+  # https://github.com/pytorch/pytorch/blob/70f57bcb1e45d21532bdb1c44d3aab018d1cbe88/torch/distributed/distributed_c10d.py#L2683
```
yeounoh, line 110: Replace the reference link with `https://github.com/pytorch/pytorch/blob/release/1.11/torch/distributed/distributed_c10d.py#L2774`

torch_xla/distributed/xla_backend.py
```diff
@@ -0,0 +1,290 @@
...
+    elif reduce_op == dist.ReduceOp.MIN:
+      return xm.REDUCE_MIN
+    elif reduce_op == dist.ReduceOp.MAX:
+      return xm.REDUCE_MAX
+    elif reduce_op == dist.ReduceOp.BXOR:
+      raise NotImplementedError(f'reduce op {reduce_op}')
+    else:
+      raise ValueError(f'Invalid reduce op {reduce_op}')
+
+  def allreduce(self, tensors, all_reduce_options):
```
yeounoh, line 57: nit: inconsistent naming convention, should use `all_reduce()`, `all_gather()` like `reduce_scatter` below.

torch_xla/distributed/xla_backend.py
```diff
@@ -0,0 +1,290 @@
...
+    raise NotImplementedError
+
+  def monitored_barrier(self, *args):
+    raise NotImplementedError
+
+  def Options(self, *args):
+    raise NotImplementedError
+
+
+class WorkXla(Work):
```
yeounoh, line 154: nit. should comment/describe the class better, as it's not obvious from the name.

yeounoh (approved):
Minor comments mostly, approving as the original PR was already approved.

test/test_torch_distributed_all_gather_xla_backend.py
yeounoh: Yes, then shouldn't we reverse the order here
```
xoutput0 = xoutputs[0] # copy
dist.all_gather(xoutputs, xinput)
```

JackCaoG: oh, it is a special use case we need to support. We need to make sure `xoutput0` also get updated when we do `all_gather` on `xoutputs`

torch_xla/distributed/xla_backend.py
JackCaoG: not sure if aws plans to do that, it is also unclear to me how C++ binding is going to help

torch_xla/distributed/xla_backend.py
JackCaoG: This api if I am not mistaken is inherit from torch distributed, so it has to be `allreduce`. Example https://github.com/pytorch/pytorch/blob/release/1.10/torch/distributed/distributed_c10d.py#L1217

torch_xla/distributed/xla_backend.py
JackCaoG: I think it is OK, this pr was already submitted to master and we don't need to replace links for every release branch.

torch_xla/distributed/xla_backend.py
JackCaoG: We should fix that in the master.

Issue #3386, opened 2022-02-18
No able to set `GLIBCXX_USE_CXX11_ABI` in the cpp test


JackCaoG:
## 🐛 Bug

I was working on the release wheel building and I found that cpp test build will fail at linking due to the `CXX_ABI` issue. During the release we set `CXX_ABI` to 0 and that was done by a check in [here](https://github.com/pytorch/xla/blob/master/test/cpp/CMakeLists.txt#L19). To be even more sure, I manually set [this line](https://github.com/pytorch/xla/blob/master/test/cpp/CMakeLists.txt#L66) and [this line](https://github.com/pytorch/xla/blob/master/test/cpp/CMakeLists.txt#L44) to 0 but I found something really weird.

under `test/cpp/build/CMakeFiles/test_ptxla.dir/flags.make` I saw

```
CXX_FLAGS =-g   -D_GLIBCXX_USE_CXX11_ABI=0 -Wno-sign-compare -Wno-deprecated-declarations -Wno-return-type -Wno-macro-redefined -Wno-return-std-move -fsized-deallocation -D_GLIBCXX_USE_CXX11_ABI=1
```

I commented out the
```
target_compile_options(test_ptxla PRIVATE ${TGT_OPTS})
```

rerun the build and then saw

```
CXX_FLAGS =-g   -D_GLIBCXX_USE_CXX11_ABI=1
```

This pretty much means I have no control over `D_GLIBCXX_USE_CXX11_ABI` because at somehow there will always be an  `-D_GLIBCXX_USE_CXX11_ABI=1` in the end. I also tried to move 

```
target_compile_options(test_ptxla PRIVATE ${TGT_OPTS})
```

to the end of the CMake file but those options we config still appears before `D_GLIBCXX_USE_CXX11_ABI=1`. 

@davidel Do you have any idea where are these two config coming fron?

JackCaoG:
by manually removing the `-D_GLIBCXX_USE_CXX11_ABI=1` in `test/cpp/build/CMakeFiles/test_ptxla.dir/flags.make` I was able to build cpp test with `CXX_ABI=0`

The workaround I am thinking right now is
1. don't build CPP test for the release wheel
2. run the cmake, have a script to update `test/cpp/build/CMakeFiles/test_ptxla.dir/flags.make` then redo the make

Both solutions are not ideal through..

davidel:
Hi @JackCaoG , you have to match that flag to everything that links to PT, otherwise you are going to have linking issues.
[...]

PR #3388, opened 2022-02-22
Set _GLIBCXX_USE_CXX11_ABI in docker file

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3386

Previously this ENV_VAR was not used by pytorch CMAKE (require by pytorch/xla c++ test CMAKE) when compiler is clang. After https://github.com/pytorch/pytorch/pull/72081 pytorch/xla also need to set this env var during our docker/wheel build to make CXX_ABI complied with pytorch


JackCaoG:
Merge it to unblock the build

yeounoh (commented):
LGTM, thank you @JackCaoG

PR #3395, opened 2022-02-24
Use Cloud Build private pool

will-cromar:
Fixes timeout in CUDA build by using larger machine with more build jobs.

I will update the triggers in our project after I submit.

Private pool docs: https://cloud.google.com/build/docs/private-pools/run-builds-in-private-pool

[Successful test build link](https://console.cloud.google.com/cloud-build/builds;region=us-central1/8f00200a-c1a9-4ea6-90dd-13ec30070e72?project=tpu-pytorch)

### DIFF
--- a/build_torch_xla_libs.sh
+++ b/build_torch_xla_libs.sh
@@ -40,8 +40,8 @@ if [[ "$TPUVM_MODE" == "1" ]]; then
 fi
 
 MAX_JOBS=
-if [[ "$XLA_CUDA" == "1" ]] && [[ "$CLOUD_BUILD" == "true" ]]; then
-  MAX_JOBS="--jobs=16"
+if [[ ! -z "$BAZEL_JOBS" ]]; then
+  MAX_JOBS="--jobs=$BAZEL_JOBS"
 fi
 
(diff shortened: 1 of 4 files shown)


JackCaoG:
This update will affect all of our build including TPUVM, 2vm right?

will-cromar:
Right. The only difference is that the build will run on a different/larger build machine (`e2-standard-32`), which should make the other builds faster too. There doesn't seem to be a way to make this change for just CUDA unless we create two cloudbuild.yaml configs.

JackCaoG (approved):
Thanks!

malfet:
Hmm, this would hide build failures for non-Google employees, which makes quick fixes/contributions from non-core-XLA devs harder, wouldn't it? 
For example, I can not follow link posted in PR description.

will-cromar:
As far as I know, this doesn't actually change the visibility of our build statuses -- that's controlled by the same GCP project permissions. We are just using a custom pool with larger workers. For example, can you see [this build](https://console.cloud.google.com/cloud-build/builds;region=global/cc2b26a5-b1ad-490e-8a8c-37c6d397bc06?project=tpu-pytorch) that happened outside of a private pool?

PR #3385, opened 2022-02-18
Migrate to upstream torch::lazy OpKind, Node, Value, and Output

wonjoolee95:
[WIP] Migrate to upstream torch::lazy OpKind, Node, Value, and Output.

This PR has gotten a bit big due to `torch::lazy::OpKind` migration that was required for the IR related classes. The most important changes are in the `ir.h` file, specifically the new `Node`, `Value` and `Output` classes.

PR #3398, opened 2022-02-28
Migrate to upstream OpKind

wonjoolee95:
Migrates existing `torch_xla::ir::OpKind` to `torch::lazy::OpKind`, which is required to migrate `Node`, `Value`, and `Output` classes. This PR only does simple replacement (no inheritance, logic change, etc), since the existing `torch_xla::ir::OpKind` and the upstream `torch:lazy::OpKind` are the same. This PR replaces all the calls and initializations to `torch::lazy::OpKind`. 

Initially was part of https://github.com/pytorch/xla/pull/3385, separated into the new PR to make the original PR bit smaller and easier to review.


wonjoolee95:
> @wonjoolee95 Could you add a more detail description? Like which class you are migrating to, what's the difference between existing class and upstream etc

Sure thing! This is more of an attempt to make the original PR https://github.com/pytorch/xla/pull/3385 smaller. Added in some info to the PR summary on top.

JackCaoG (approved):
Thanks! Could you fix the conflict?

wonjoolee95:
Fixed some merge conflicts. Will merge as CI tests pass.

PR #3400, opened 2022-03-01
Bind torch and torch_xla build option, USE_CXX11_ABI, to _GLIBCXX_USE…

### LINKED ISSUE #3390: Clean up CXX_ABI setup in pytorch/XLA
## 🚀 Feature
Test if we can remove
```
  if [ "$CXX_ABI" == "0" ]; then
    export CFLAGS="${CFLAGS} -D_GLIBCXX_USE_CXX11_ABI=0"
    export CXXFLAGS="${CXXFLAGS} -D_GLIBCXX_USE_CXX11_ABI=0"
  fi
```
in `scripts/build_torch_wheels.sh` since _GLIBCXX_USE_CXX11_ABI should be enough to control pytorch CXXABI setting. PyTorch/XLA should just read pytorch setting by import torch; print(int(torch._C._GLIBCXX_USE_CXX11_ABI)) in setup.py.

yeounoh:
Resolve #3390 . This change will simplify and make sure that torch and torch-xla use the same GLIBCXX_USE_ABI flag.

### DIFF
--- a/.circleci/docker/Dockerfile
+++ b/.circleci/docker/Dockerfile
@@ -30,13 +30,13 @@ ENV TORCH_CUDA_ARCH_LIST Maxwell
 ENV TORCH_NVCC_FLAGS "-Xfatbin -compress-all"
 ENV CUDA_PATH /usr/local/cuda
 
-# Whether to build torch and torch_xla libraries with CXX ABI
+# Set C/C++ compilers
 ENV CC "${cc}"
 ENV CXX "${cxx}"
-ENV CXX_ABI "${cxx_abi}"
+
+# Whether to build torch and torch_xla libraries with CXX ABI
 ENV _GLIBCXX_USE_CXX11_ABI "${cxx_abi}"
-ENV CFLAGS "${CFLAGS} -D_GLIBCXX_USE_CXX11_ABI=${CXX_ABI}"
-ENV CXXFLAGS="${CXXFLAGS} -D_GLIBCXX_USE_CXX11_ABI=${CXX_ABI}"
+ENV CXXFLAGS "${CXXFLAGS} -D_GLIBCXX_USE_CXX11_ABI=${cxx_abi}"
 
 # Whether to build for TPUVM mode
 ENV TPUVM_MODE "${tpuvm}"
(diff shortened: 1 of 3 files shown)


JackCaoG:
Have we verified this works by building with the script?

JackCaoG:
lol, what is 
```
yeounoh-test-build (tpu-pytorch) In progress — Summary
```

yeounoh:
> lol, what is
> 
> ```
> yeounoh-test-build (tpu-pytorch) In progress — Summary
> ```

I think the new GCB app is automatically linking my test trigger to this PR. Quite impressive lol. I will remove the linkage after the test is done.

.circleci/docker/Dockerfile
```diff
@@ -30,13 +30,13 @@ ENV TORCH_CUDA_ARCH_LIST Maxwell
...
+# Set C/C++ compilers
 ENV CC "${cc}"
 ENV CXX "${cxx}"
-ENV CXX_ABI "${cxx_abi}"
+
+# Whether to build torch and torch_xla libraries with CXX ABI
 ENV _GLIBCXX_USE_CXX11_ABI "${cxx_abi}"
-ENV CFLAGS "${CFLAGS} -D_GLIBCXX_USE_CXX11_ABI=${CXX_ABI}"
-ENV CXXFLAGS="${CXXFLAGS} -D_GLIBCXX_USE_CXX11_ABI=${CXX_ABI}"
+ENV CXXFLAGS "-D_GLIBCXX_USE_CXX11_ABI=${cxx_abi}"
```
JackCaoG, line 39: hmm, do we still need to manually set `"-D_GLIBCXX_USE_CXX11_ABI` here? I was hoping by setting `_GLIBCXX_USE_CXX11_ABI` env var above will build pytorch with the correct CXXABI. and then pytorch/xla will just read pytorch setting in our `setup.py`

yeounoh: Unfortunately, we do -- if we want to use `torch._C._GLIBCXX_USE_CXX11_ABI`, then we should set `CXXFLAGS`, too -- otherwise, it always returns 1 (True) alwyas -- [whelp :(](https://discuss.pytorch.org/t/setting-glibcxx-use-cxx11-abi-to-0/98054)

JackCaoG: any reason we overwrite the `CXXFLAGS` instead of pending to it? I guess it is base env of the docker so there shouldn't be any `CXXFLAGS` before this?

yeounoh: Good catch, it worked -- but to be future-proof, let me append :) thanks!

JackCaoG (approved):
Thanks!

PR #3402, opened 2022-03-02
Migrate to upstream Node and Output

wonjoolee95:
Migrate the existing `torch_xla::ir::Node` to upstream `torch::lazy::Node`. This is attempt #2 to make https://github.com/pytorch/xla/pull/3385 smaller. 

As a part of this migration, we need to extend our existing `torch_xla::ir::Node` with the upstream `torch::lazy::Node`. The only difference between the existing and upstream `Node` class is that our existing `torch_xla::ir::Node` contains information about `xla::Shape` with functions `torch_xla::ir::Node::shape()` whereas the upstream `torch::lazy::Node` doesn't hold any information about shape. 

The `torch::lazy::Node` defines virtual functions `std::vector<Output>& operands()` and `Output& operand(size_t i)` that return `torch::lazy::Output`. However, since PT/XLA requires `xla::shape` information in `Output`, this PR also introduces two new functions `operands_wish_shape()` and `operands_with_shape(size_t i)` that return our existing `torch_xla::ir::Output` which holds information about `xla::shape`.


wonjoolee95:
FYI was able to fix the run time issue: `RuntimeError: vector::_M_range_check: __n (which is 1489162288) >= this->size() (which is 2452)` through using `gdb`:
```
(gdb) catch throw
(gdb) run wonjoo.py
(gdb) where
#0  0x00007fffd68482c2 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007fffd683f278 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007fffd7ce6c9b in std::vector<c10::InternedStrings::SymbolInfo, std::allocator<c10::InternedStrings::SymbolInfo> >::_M_range_check (this=0x7fffe79a2840 <c10::globalStrings()::s+56>, 
    __n=1489121488) at /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_vector.h:1073
#3  0x00007fffd7ce60cf in std::vector<c10::InternedStrings::SymbolInfo, std::allocator<c10::InternedStrings::SymbolInfo> >::at (this=0x7fffe79a2840 <c10::globalStrings()::s+56>, __n=1489121488)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_vector.h:1094
#4  0x00007fffd7cde28a in c10::InternedStrings::customString (this=0x7fffe79a2808 <c10::globalStrings()::s>, sym=...) at ../aten/src/ATen/core/interned_strings.cpp:91
#5  0x00007fffd7cde20f in c10::InternedStrings::string (this=0x7fffe79a2808 <c10::globalStrings()::s>, sym=...) at ../aten/src/ATen/core/interned_strings.cpp:45
#6  0x00007fffd7ce59ad in c10::Symbol::toQualString (this=0x7fffffff8558) at ../aten/src/ATen/core/interned_strings.cpp:109
#7  0x00007fffdc520055 in torch::lazy::OpKind::hash (this=0x7fffffff8558) at ../torch/csrc/lazy/core/ir.cpp:41
```
The `torch::lazy::hash` call was throwing this error and saw that the constructor did not initialize the `op` correctly.

wonjoolee95:
Most of the tests are passing but still many are failing with:
```
test_multinomial_invalid_xla (__main__.TestTorchDeviceTypeXLA) ... 2022-03-02 10:03:12.179622: W   16931 tensorflow/core/framework/op_kernel.cc:1745] OP_REQUIRES failed at xrt_execute_op.cc:434 : INVALID_ARGUMENT: Run-time shape mismatch for XRTExecute argument[0] (3524830195124450). Expected element_type: F32
dimensions: 1
layout {
  minor_to_major: 0
  format: DENSE
}
is_dynamic_dimension: false
; got element_type: F32
dimensions: 3
layout {
  minor_to_major: 0
  format: DENSE
}
is_dynamic_dimension: false
```

wonjoolee95:
Was able to narrow down the issue a bit more by repro'ing `python test/test_torch.py TestTorchDeviceTypeXLA.test_cdist_empty_xla` locally. It seems like manually running the tests one by one for `torch.dist` succeeds:
```
# Test 1
x = torch.randn((0, 5), device=device)
y = torch.randn((4, 5), device=device)
print(torch.cdist(x, y))
```
On a new python session, the below also succeeds:
```
# Test 2
x = torch.randn((2, 5), device=device)
y = torch.randn((0, 5), device=device)
print(torch.cdist(x, y))
```
However when I override the variables in the same python session, they fail:
```
x = torch.randn((0, 5), device=device)
# Overriding variable x
x = torch.randn((2, 5), device=device)
y = torch.randn((4, 5), device=device)
print(torch.cdist(x, y)) # Fails with the error message in the thread above
```
And since the `test_cdist_empty_xla` is organized in such a way where the variable are overridden constatly for different test cases, it's causing this failure. Seems like this PR introduced a bug related to when the Node is being replaced.

torch_xla/csrc/ir.h
```diff
@@ -235,7 +247,8 @@ class Node {
   std::vector<NodePtr> operands_;
   // Outputs do not hold references on the nodes, and neither do the uses, since
   // otherwise we get into circular reference counting.
-  std::vector<Output> operands_as_outputs_;
+  std::vector<torch::lazy::Output> operands_as_outputs_;
+  std::vector<Output> operands_as_outputs_with_shape_;
```
JackCaoG, line 251: WE shouldn't introduce a new vector just because our `Output` is different than `torch::lazy::Output` (I think the only difference is our version has `xla::shape()`).

@wconstab I remembered we had a conversation regarding this, what's the reason you removed `shape`  within the `Output`? I remembered you mentioned that it was not being used anywhere or something.

wonjoolee95:
Looking back at PT/XLA, seems like we can remove `shape` calls from `torch_xla::ir::Output`. There are calls to `shape()` but we can call `Node.shape()` instead. Working on a quick fix to remove the existing `torch_xla::ir::Output` completely.

torch_xla/csrc/ir.h
wconstab: I don't mind adding shape back to output, although, I also want to discuss Output vs Value with you guys and see if we can merge the two.

I also intended to put shape() api back on Node base class.  I'm surprised you didn't remind me, i thought you needed it.  If it helps, I can do it soon.

JackCaoG: Ah OK. XLANode needs to have `xla::shape`, our hope is that we can use `lazy::shape` in the `IR Node and XLATensor` layer and `xla::shape` from lowering part and below. Since we hasn't start using codegen, no `lazy::shape` is being provided(from formula or meta tensor) so we hasn't run into this issue yet.

The long term plan is still to use `lazy::shape` in the non-lowering layer but until all of the op we currently supported today has a meta kernel or formula, we can't do this switch. This will become a larger issue when we do the phase 3 migration since LazyTensorBase class will always use `lazy::shape`.

wconstab: OK, well as long as it is generally helpful / not harmful for you, I will move shape to Node base soon.

wonjoolee95:
I was able see that the cache incorrectly gets the nodes. In the function `SyncTensorsGraphInternal` in `tensor.cpp`, I've added a `cout` line to see when we hit the cache:
```
  std::shared_ptr<Async> async = TryRunCachedSync(tensors, &coll, &po_data);
  if (async != nullptr) {
    std::cout << "[WONJOO] SaveTensorsGraphInfo found cache" << std::endl;
    return async;
  }
```

When I run the following code:
```
>>> x = torch.randn((0, 5), device=device)
>>> y = torch.randn((4, 5), device=device)
>>> z = torch.cdist(x, y)
>>> z
tensor([], device='xla:0', size=(0, 4))
>>> x = torch.randn((2, 5), device=device)
>>> z2 = torch.cdist(x, y)
>>> z2
[WONJOO] SaveTensorsGraphInfo found cache
```
I can see that printing `z2` will hit the cache, but we shouldn't since we've updated `x` to a new tensor. Seeing it act this way, we thought if it was the hashing of the node that has an issue (especially since `torch::lazy::Node` does not include and hash `xla::shape`), but seems like our existing hashing already correctly includes the shape through the `GetOpHash` function:
```
torch::lazy::hash_t Node::GetOpHash(torch::lazy::OpKind op,
                                    const xla::Shape& shape,
                                    torch::lazy::hash_t hash_seed) {
  torch::lazy::hash_t h =
      torch::lazy::HashCombine(op.hash(), torch::lazy::Hash(shape.ToString()));
  return torch::lazy::HashCombine(h, hash_seed);
}
```
```

**Comment by shunting314** [comment]
Here is the PR Will mentioned: https://github.com/pytorch/pytorch/pull/72561 . If dynamic shape is not enabled, then the PR is basically a nop. But if dynamic shape is enabled, the PR make sure the hash used to lookup compiled graph does not have shape information, but the hash used to lookup the shape cache contains the shape information.

The behavior from your tests indeed seems like dynamic shape is enabled.

**Comment by JackCaoG** [comment]
@wonjoolee95 Can we figure out 
1. Which Node construcotr is called for tensor `x` and `y`
2. Dump the hash computed for these two node for both test cases, they should be different
3. Given `x` and `y` should have different hash seed, `torch.cdist(x, y)` nodes should also have different hash

**Comment by wonjoolee95** [comment]
@shunting314 Thanks for the comment. In our case, it seems like dynamic shape is disabled. When I add some print statements in our xla `Node` constructor, I can see that the outputs are all `false`.

**Comment by wonjoolee95** [comment]
@shunting314 Just wanted to clarify one thing on the PR you mentioned above. If we want to disable dynamic shape, should we have `bakeInSizes` to `false`?

**Comment by JackCaoG** [comment]
> bakeInSizes

I think it controls whether `shape` hash includes the actual size(`std::vector<int64_t> sizes_`) value. If it is set to false only the `sizes_.size()`(number of dimension of this shape) is included in the hash.

**Comment by wonjoolee95** [comment]
After the most recent commit with updating the hash logic, the test cases above with `cdist` is passing. I'll run some other tests that were failing and add more details.

**Comment by wonjoolee95** [comment]
All CI tests passing now. The latest commit fixes the way we generate the upstream `torch::lazy::Node` hash by combing the hashes for the operands like below:
```
torch::lazy::hash_t OperandHashes(const OpList& operands,
                                  const torch::lazy::hash_t& seed,
                                  bool bakeInSizes) {
  torch::lazy::hash_t hash = seed;
  for (auto& operand : operands) {
    if (!operand) {
      hash = torch::lazy::HashCombine(
          hash, static_cast<uint64_t>(torch::lazy::kNullOpt));
      continue;
    }
    auto operand_hash =
        bakeInSizes ? operand.hash_with_sizes() : operand.hash_without_sizes();
    hash = torch::lazy::HashCombine(hash, operand_hash);
  }
  return hash;
}
```

With this method, our existing 3 constructors look like this:
1. Creates a new node with the given op name.
```
Node::Node(torch::lazy::OpKind op, OpList operands, xla::Shape shape,
           size_t num_outputs, torch::lazy::hash_t hash_seed)
    : torch::lazy::Node(
          op, num_outputs,
          /* node_hash */ torch::lazy::HashCombine(op.hash(), hash_seed),
          /* hash_func */
          [&](bool bakeInSizes) -> torch::lazy::hash_t {
            return OperandHashes(operands,
                                 torch::lazy::HashCombine(op.hash(), hash_seed),
                                 bakeInSizes);
          }),
```
2. Same as the constructor above, but the shape is generated by a function, only if needed (shape cache miss).
```
Node::Node(torch::lazy::OpKind op, OpList operands,
           const std::function<xla::Shape()>& shape_fn, size_t num_outputs,
           torch::lazy::hash_t hash_seed)
    : Node(std::move(op), operands, xla::Shape(), num_outputs, hash_seed) {
```
3. Contructor used to create leaf nodes.
```
Node::Node(torch::lazy::OpKind op, xla::Shape shape, size_t num_outputs,
           torch::lazy::hash_t hash_seed)
    : torch::lazy::Node(op, num_outputs, /* hash_func */
                        [&](bool /*bakeInSizes*/) -> torch::lazy::hash_t {
                          return GetOpHash(op, shape, hash_seed);
                        }),
```

There is some clean up left to do (commented code, some unused functions), but I'll include these in a clean-up PR next that cleans up all the phase 1 related code.

**Comment by shunting314** [comment]
> @shunting314 Just wanted to clarify one thing on the PR you mentioned above. What is the use of the newly introduced boolean flag `bakeInSizes`? If we want to disable dynamic shape, should we have `bakeInSizes` to `false`?

yes, if dynamic shape is disabled, a false value of `bakeInSizes` will be passed in to relevant methods.

**Comment by JackCaoG** [comment]
A bunch of view ops failed with
```
RuntimeError: Error while lowering: f32[5,1,5]{2,1,0} aten::view, output_size=(5, 1, 5)
Error: vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
```

**Comment by wonjoolee95** [comment]
Updated to fix the `view_ops` related tests.

File: torch_xla/csrc/ir.cpp [hunk 2]
```diff
@@ -100,45 +117,40 @@ std::string Use::ToString() const {
...
 Node::Node(torch::lazy::OpKind op, OpList operands, xla::Shape shape,
            size_t num_outputs, torch::lazy::hash_t hash_seed)
-    : op_(std::move(op)),
-      num_outputs_(num_outputs),
-      shape_(std::move(shape)),
-      node_hash_(torch::lazy::HashCombine(op_.hash(), hash_seed)),
-      hash_(node_hash_) {
+    : torch::lazy::Node(
+          op, num_outputs,
+          /* node_hash */ torch::lazy::HashCombine(op.hash(), hash_seed),
```
JackCaoG, line 142: /* node_hash= */ (no space in between I think)

torch_xla/csrc/ir.cpp
```diff
@@ -81,6 +81,23 @@ void EmitShortFrameInfo(std::ostream& stream,
...
+                                  bool bakeInSizes) {
+  torch::lazy::hash_t hash = seed;
+  for (auto& operand : operands) {
+    if (!operand) {
+      hash = torch::lazy::HashCombine(
+          hash, static_cast<uint64_t>(torch::lazy::kNullOpt));
+      continue;
+    }
+    auto operand_hash =
+        bakeInSizes ? operand.hash_with_sizes() : operand.hash_without_sizes();
```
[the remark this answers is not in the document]
JackCaoG: On a second thought, we should always ignore `bakeInSizes` and use the `operand_hash` with `xla::shape`

torch_xla/csrc/ir.cpp
```diff
@@ -100,45 +117,40 @@ std::string Use::ToString() const {
...
-      shape_(std::move(shape)),
-      node_hash_(torch::lazy::HashCombine(op_.hash(), hash_seed)),
-      hash_(node_hash_) {
+    : torch::lazy::Node(
+          op, num_outputs,
+          /* node_hash */ torch::lazy::HashCombine(op.hash(), hash_seed),
+          /* hash_func */
+          [&](bool bakeInSizes) -> torch::lazy::hash_t {
+            return OperandHashes(operands,
+                                 torch::lazy::HashCombine(op.hash(), hash_seed),
```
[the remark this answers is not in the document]
JackCaoG: we should just ignore `bakeInSizes`

torch_xla/csrc/ir.cpp
```diff
@@ -153,11 +165,11 @@ Node::Node(torch::lazy::OpKind op, OpList operands,
 
 Node::Node(torch::lazy::OpKind op, xla::Shape shape, size_t num_outputs,
            torch::lazy::hash_t hash_seed)
-    : op_(std::move(op)),
-      num_outputs_(num_outputs),
-      shape_(std::move(shape)),
-      node_hash_(GetOpHash(op_, shape_, hash_seed)),
-      hash_(node_hash_) {
+    : torch::lazy::Node(op, num_outputs, /* hash_func */
```
JackCaoG, line 168: /* hash_func= */

torch_xla/csrc/ir.cpp
```diff
@@ -153,11 +165,11 @@ Node::Node(torch::lazy::OpKind op, OpList operands,
...
+                        [&](bool /*bakeInSizes*/) -> torch::lazy::hash_t {
+                          return GetOpHash(op, shape, hash_seed);
```
JackCaoG, line 170: OK it seem like you already ignore `bakeInSizes` here, you should also update other Node constructors to do the same.

torch_xla/csrc/ir.cpp
```diff
@@ -179,16 +191,17 @@ const xla::Shape& Node::shape(size_t output_index) const {
 void Node::AddOperand(NodePtr node, size_t index) {
```
JackCaoG, line 191: I think we should refined `NodePtr` to be a pointer of base Node class instead of pointer of `xla::Node`. Is there anything blocking us from doing that?

torch_xla/csrc/ir_util.cpp
```diff
@@ -85,16 +88,16 @@ std::vector<Value> Util::Clone(absl::Span<const Value> values,
 }
 
 std::vector<Value> Util::Clone(absl::Span<const Value> values) {
```
JackCaoG, line 90: Hmm, this can be done in a separate pr, can we verify if `Clone` is being used anywhere in our code base today? I think it is not. If that's the case we should consider removing it.

I think the motivation was we clone the IR graph and do some IR level optimization before lowering but that optimization was never done.

torch_xla/csrc/ir_util.h
```diff
@@ -13,40 +14,42 @@ class Util {
  public:
   // Tracks the emission status of the nodes during the post-order generation.
   // It helps tracking loops within the computation graphs.
-  enum EmitStatus {
-    kNotEmitted,
-    kEmitting,
-    kEmitted,
-  };
+  // enum EmitStatus {
```
JackCaoG, line 17: let's remove

torch_xla/csrc/lowering_context.cpp
```diff
@@ -45,7 +47,7 @@ class HloMetadataSetter {
     std::string op_type =
         absl::StrReplaceAll(node->op().ToString(), {{":", "_"}});
     metadata.set_op_type(op_type);
-    const ir::MetaData& nmeta = node->metadata();
+    const torch::lazy::MetaData& nmeta = node->metadata();
```
JackCaoG, line 50: This looks like a bug. You kept the xla version of the `MetaData` in Node class but in here `node->metadata()` will return the `torch::lazy::MetaData`. I am surprised this code actually compiles.

JackCaoG (changes requested):
Thanks @wonjoolee95 

@wconstab Can we move the `lazy::shape` to base IR class and move the `hash_with_size` and `hash_without_size` to the `TsNode`?

torch_xla/csrc/ir.cpp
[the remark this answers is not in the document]
wonjoolee95: Updated to ignore `bakeInSizes`. Also this PR actually deletes the `node_hash_` in `torch_xla::ir::Node` since we're using the parent class' hash.

torch_xla/csrc/ir.cpp
wonjoolee95: Good point, I'll include this in a new commit in this PR.

torch_xla/csrc/ir.h
```diff
@@ -176,16 +135,16 @@ class Node {
...
+    return operands_as_outputs_.at(i);
+  }
 
   const std::set<Use>& uses() const { return uses_; }
 
-  torch::lazy::hash_t node_hash() const { return node_hash_; }
-
-  torch::lazy::hash_t hash() const { return hash_; }
-
   const MetaData& metadata() const { return metadata_; }
```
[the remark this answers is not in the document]
wonjoolee95: Updated PR with a new commit that migrates to upstream `MetaData` and related functions.

torch_xla/csrc/ir.cpp
wonjoolee95: Updated in the latest commit.

wonjoolee95:
> Is this pr ready for the next round of review?

Yep, the latest commits addressed the comments and should be good for next round of review.

A small issue I faced while migrating to upstream `MetaData` was that in this code `torch::lazy::MetaData metadata = torch::lazy::GetMetaDataIfDebugging()` in `torch_xla/csrc/ir.cpp` throws the error:
```
import _XLAC
ImportError: /opt/conda/lib/python3.7/site-packages/torch_xla-1.11-py3.7-linux-x86_64.egg/_XLAC.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZN5torch4lazy22GetMetaDataIfDebuggingEv
```
Which seems weird because the function `torch::lazy::GetMetaDataIfDebugging` does exist and is implemented here -- `https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/ir_metadata.h#L46`. 

The only place this is called is in the `Node::String()` function for debugging, so I've commented out this code for now with a `TODO`. Other than this, this PR is updated to address all the previous comments.

torch_xla/csrc/tensor.cpp
```diff
@@ -1573,7 +1575,8 @@ XLATensor::CompilationResult XLATensor::Compile(
                                    std::move(po_data->emission_map));
   for (auto index : coll.indices) {
     ir::Value ir_value = tensors[index].CurrentIrValue();
-    xla::XlaOp root = lowering_ctx.GetOutputOp(ir_value);
+    xla::XlaOp root = lowering_ctx.GetOutputOp(
+        torch::lazy::Output(ir_value.node.get(), ir_value.index));
```
JackCaoG, line 1579: hmm.. seems like we used to can implicitly convert `Value` to `Output` but that is no longer true?

JackCaoG (changes requested):
Thanks @wonjoolee95 . We shouldn't merge this pr if it will break debugging metadata print. We should also add some test to make sure it works..

torch_xla/csrc/ir.cpp
```diff
@@ -224,10 +210,12 @@ std::string Node::ToString() const {
...
     ss << ", num_outputs=" << num_outputs();
   }
-  if (!metadata_.scope.empty()) {
-    ss << ", scope=" << metadata_.scope;
-  }
-  EmitShortFrameInfo(ss, metadata_.frame_info);
+  // TODO @wonjoo Fix the import _XLAC ImportError
+  // torch::lazy::MetaData metadata = torch::lazy::GetMetaDataIfDebugging();
+  // if (!metadata.scope.empty()) {
+  //   ss << ", scope=" << metadata.scope;
```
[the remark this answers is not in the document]
wonjoolee95: Hey @wconstab, the error is same as the one pasted in the conversation above, it is:
```
import _XLAC
ImportError: /opt/conda/lib/python3.7/site-packages/torch_xla-1.11-py3.7-linux-x86_64.egg/_XLAC.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZN5torch4lazy22GetMetaDataIfDebuggingEv
```
Which usually means that it cannot find the function `torch::lazy::GetMetaDataIfDebugging`. Initially I thought it's the same issue as the import path issue on the `backend_device.h` but it seems like `orch::lazy::GetMetaDataIfDebugging` is already included in part of the `setup.py` as `'include/torch/csrc/lazy/core/*.h'.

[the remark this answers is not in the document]
wconstab: ahh ok- looks like we missed something. Here is a PR... #74784

torch_xla/csrc/ir.cpp
```diff
@@ -66,19 +67,19 @@ ShapeCache* GetShapeCache() {
...
+  torch::lazy::hash_t hash = node_hash;
+  for (auto& operand : operands) {
+    if (!operand) {
+      hash = torch::lazy::HashCombine(
+          hash, static_cast<uint64_t>(torch::lazy::kNullOpt));
+      continue;
     }
-    stream << ", location=" << frame.function << "@" << frame.file.substr(pos)
-           << ":" << frame.line;
+    auto operand_hash = operand.hash_with_sizes();
```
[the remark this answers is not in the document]
wonjoolee95: It can, since our `bakeInSizes` are defaulted to `false`. Updated to call `operand.hash()` directly.

torch_xla/csrc/ir.cpp
```diff
@@ -127,18 +109,24 @@ torch::lazy::hash_t Value::hash() const {
   return torch::lazy::HashCombine(node->hash(), index);
```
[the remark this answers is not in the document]
wonjoolee95: It's not different in anymore, since `bakeInSizes` is defaulted. Removed it.

torch_xla/csrc/ir.cpp
wonjoolee95: Yep that is correct.

torch_xla/csrc/tensor.cpp
wonjoolee95: Yep, due to the `Output` now being of type `torch::lazy` currently after this migration.

torch_xla/csrc/ir_util.cpp
wonjoolee95: Thanks! I'll keep this into a separate PR to make this PR concise as possible.

torch_xla/csrc/ir.cpp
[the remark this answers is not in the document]
wconstab: I just landed the fix, it failed to land Friday.

Please ping me if there are other issues/blockers!

JackCaoG:
@wonjoolee95 Build is currently failing, can you take a look?

torch_xla/csrc/ir.cpp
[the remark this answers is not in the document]
wonjoolee95: > @wonjoolee95 Build is currently failing, can you take a look?

Failing due to upstream having `shape()` function. Latest commit renames our `ir::Node::shape` to `xla_shape`.

wonjoolee95:
> @wonjoolee95 Build is currently failing, can you take a look?

Failing due to upstream having shape() function. Latest commit renames our ir::Node::shape to xla_shape.

JackCaoG (commented):
Mostly LGTM, I think the only missing part is the `registerGetFrameInfo`

torch_xla/csrc/ir.cpp
```diff
@@ -100,45 +101,27 @@ std::string Use::ToString() const {
   return ss.str();
 }
 
-size_t Output::Hasher::operator()(const Output& output) const {
-  return torch::lazy::StdHashCombine(
-      reinterpret_cast<std::ptrdiff_t>(output.node), output.index);
-}
+const xla::Shape& Value::shape() const { return node->xla_shape(index); }
```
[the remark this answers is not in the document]
wonjoolee95: Sounds good, I'll follow-up with this in the coming `torch::lazy::Value` PR.

torch_xla/csrc/ir.cpp
```diff
@@ -100,45 +101,27 @@ std::string Use::ToString() const {
...
-      num_outputs_(num_outputs),
-      shape_(std::move(shape)),
-      node_hash_(torch::lazy::HashCombine(op_.hash(), hash_seed)),
-      hash_(node_hash_) {
-  metadata_.scope = GetCurrentScope();
-  metadata_.frame_info = GetFrameInfo();
+    : torch::lazy::Node(
+          op, num_outputs,
+          /*node_hash=*/torch::lazy::HashCombine(op.hash(), hash_seed),
+          /*hash_func=*/
```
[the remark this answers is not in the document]
wonjoolee95: Updated

torch_xla/csrc/ir.cpp
```diff
@@ -148,47 +131,49 @@ Node::Node(torch::lazy::OpKind op, OpList operands,
...
            torch::lazy::hash_t hash_seed)
-    : op_(std::move(op)),
-      num_outputs_(num_outputs),
-      shape_(std::move(shape)),
-      node_hash_(GetOpHash(op_, shape_, hash_seed)),
-      hash_(node_hash_) {
-  metadata_.scope = GetCurrentScope();
-  metadata_.frame_info = GetFrameInfo();
-}
+    : torch::lazy::Node(op, num_outputs, /*hash_func=*/
```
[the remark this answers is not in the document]
wonjoolee95: Updated

wonjoolee95:
> Mostly LGTM, I think the only missing part is the registerGetFrameInfo

Addressed rest of the comments. The PR should be good, except I don't see a good way on our end to set the upstream `metadata`. 

The upstream `torch::lazy::Node` sets `metadata` only in its constructor:
```
Node::Node(OpKind op, size_t num_outputs, hash_t node_hash, std::function<hash_t(bool)> dag_hash_fn)
    : ...
      metadata_(GetMetaDataIfDebugging()) {} 
```
with `GetMetaDataIfDebugging()` is defined as:
```
MetaData GetMetaDataIfDebugging() {
  if (!FLAGS_torch_lazy_ir_debug) {
    return MetaData();
  }
  MetaData meta;
  meta.scope = GetCurrentScope();
  meta.frame_info = GetFrameInfo(); // registered through RegisterGetFrameInfo()
  return meta;
}
```

In our `torch_xla::ir::Node` constructor, we need a way to set the upstream `metadata` after we call `RegisterGetFrameInfo()`. We want to do something like:
```
Node::Node(torch::lazy::OpKind op, OpList operands, xla::Shape shape,
           size_t num_outputs, torch::lazy::hash_t hash_seed)
    : torch::lazy::Node(
        // the upstream constructor sets metadata
        )
    torch::lazy::RegisterGetFrameInfo(...);
    // we need some function like SetMetaData to set the metadata after registering our GetFrameInfo function
    torch::lazy::SetMetaData(GetMetaDataIfDebugging()) 
}
```

We can continue to keep our own set of `metadata` and related functions to fix this problem, but @JackCaoG, please let me know if this makes sense and if you see other ways for us to do this.

JackCaoG:
> > Mostly LGTM, I think the only missing part is the registerGetFrameInfo
> 
> Addressed rest of the comments. The PR should be good, except I don't see a good way on our end to set the upstream `metadata`.
> 
> The upstream `torch::lazy::Node` sets `metadata` only in its constructor:
> 
> ```
> Node::Node(OpKind op, size_t num_outputs, hash_t node_hash, std::function<hash_t(bool)> dag_hash_fn)
>     : ...
>       metadata_(GetMetaDataIfDebugging()) {} 
> ```
> 
> with `GetMetaDataIfDebugging()` is defined as:
> 
> ```
> MetaData GetMetaDataIfDebugging() {
>   if (!FLAGS_torch_lazy_ir_debug) {
>     return MetaData();
>   }
>   MetaData meta;
>   meta.scope = GetCurrentScope();
>   meta.frame_info = GetFrameInfo(); // registered through RegisterGetFrameInfo()
>   return meta;
> }
> ```
> 
> In our `torch_xla::ir::Node` constructor, we need a way to set the upstream `metadata` after we call `RegisterGetFrameInfo()`. We want to do something like:
> 
> ```
> Node::Node(torch::lazy::OpKind op, OpList operands, xla::Shape shape,
>            size_t num_outputs, torch::lazy::hash_t hash_seed)
>     : torch::lazy::Node(
>         // the upstream constructor sets metadata
>         )
>     torch::lazy::RegisterGetFrameInfo(...);
>     // we need some function like SetMetaData to set the metadata after registering our GetFrameInfo function
>     torch::lazy::SetMetaData(GetMetaDataIfDebugging()) 
> }
> ```
> 
> We can continue to keep our own set of `metadata` and related functions to fix this problem, but @JackCaoG, please let me know if this makes sense and if you see other ways for us to do this.

Can we call `torch::lazy::RegisterGetFrameInfo(...); ` outside of the Node constructor? This function should be called once per process. We don't need to call it on every Node construction.

wonjoolee95:
> Can we call `torch::lazy::RegisterGetFrameInfo(...); ` outside of the Node constructor? This function should be called once per process. We don't need to call it on every Node construction.

I see. I can add it back to the end of the constructor after we call the parent class constructor.

wonjoolee95:
I can't seem to get the the metadata set with having `RegisterGetFrameInfo` outside the functions in `ir.cpp`. Instead I've added to the `Node` constructor for now.

JackCaoG:
hmm @wconstab What's the intended way of calling `RegisterGetFrameInfo`? I think we only need to call it once. One idea I had was in `ir.h` we do 

```
static bool registed = RegisterGetFrameInfo(GetPythonFrames);
```

but  `RegisterGetFrameInfo` returns void so we can't assign it to any variable. I also don't know if we can call a function on the file scope without assigning it to any variable..

wconstab:
> but RegisterGetFrameInfo returns void so we can't assign it to any variable. I also don't know if we can call a function on the file scope without assigning it to any variable..
I'm fine adding a return bool to enable this use case

But on the TS side we actually have an explicit initialization API we have to call to initialize the TS backend, so we can just call it there.  I guess you could also go that route, where you create an 'InitXLABackend' function which returns a bool and then you call that at library load time if you want it to be called by default, or you bind it to python if you want more control.

LMK- happy to add bool return to RegisterGetFrameInfo

wonjoolee95:
> > but RegisterGetFrameInfo returns void so we can't assign it to any variable. I also don't know if we can call a function on the file scope without assigning it to any variable..
> > I'm fine adding a return bool to enable this use case
> 
> But on the TS side we actually have an explicit initialization API we have to call to initialize the TS backend, so we can just call it there. I guess you could also go that route, where you create an 'InitXLABackend' function which returns a bool and then you call that at library load time if you want it to be called by default, or you bind it to python if you want more control.
> 
> LMK- happy to add bool return to RegisterGetFrameInfo

Thanks for the information. For now, I think adding the bool return value for `RegisterGetFrameInfo` seems to be the quickest fix. After that, we can also discuss on our end if we want to adapt something similar. We can also keep our own sets of `metadata` and related functions if we want to get through this issue for now. @JackCaoG, what do you think?

Also, @wconstab if adding the bool return value is all there is, I can open a quick PR on the PyTorch's end as well.

JackCaoG:
@wonjoolee95 Let me quickly review this pr and get it merge without `RegisterGetFrameInfo` if possible. Don't want to block pytorch team for too long. Let's have a follow up pr to fix `RegisterGetFrameInfo` soon.

torch_xla/csrc/python_util.cpp
```diff
@@ -1,53 +1,54 @@
...
+// #include "torch_xla/csrc/python_util.h"
 
-#include <Python.h>
-#include <frameobject.h>
-#include <pybind11/pybind11.h>
-#include <torch/csrc/utils/python_strings.h>
+// #include <Python.h>
+// #include <frameobject.h>
+// #include <pybind11/pybind11.h>
+// #include <torch/csrc/utils/python_strings.h>
```
JackCaoG, line 7: please clean up this file

torch_xla/csrc/python_util.h
```diff
@@ -1,24 +1,20 @@
-#pragma once
+// #pragma once
 
-#include <iostream>
```
JackCaoG, line 3: ditto, just remove this file

torch_xla/csrc/ir.cpp
```diff
@@ -66,21 +68,23 @@ ShapeCache* GetShapeCache() {
...
-    stream << ", location=" << frame.function << "@" << frame.file.substr(pos)
-           << ":" << frame.line;
+    hash = torch::lazy::HashCombine(hash, operand.hash());
   }
+  return hash;
 }
 
+// Register GetFrameInfo function
+// torch::lazy::RegisterGetFrameInfo()
+
```
JackCaoG, line 87: please delete these comments

JackCaoG (commented):
@wonjoolee95 Can you clean up the commented lines and I think we can merge this pr with a follow up pr for `RegisterGetFrameInfo`

wonjoolee95:
> @wonjoolee95 Can you clean up the commented lines and I think we can merge this pr with a follow up pr for `RegisterGetFrameInfo`

Thanks @JackCaoG. Just removed the two commented out files (`torch_xla/csrc/python_util.cpp` and `torch_xla/csrc/python_util.h`) with the latest commit.

wconstab:
You could still land this PR with xla/csrc/python_util frameinfo function, but whenever you do make the change to use ours, see this fix:
https://github.com/pytorch/pytorch/pull/74979

I had left some dead code in ir_metadata.cpp and there is already a new 'frame info' func in torch/csrc/lazy/core/debug_util.cpp which my fix above hooks up to ir_metadata.  Once that fix is landed (iiuc) then you will be able to just use our frame func without xla/csrc/debug_util

wonjoolee95:
Removed calls to upstream `RegisterGetFrameInfo`. Local testing still TBD, having some issues with my local machine, now trying to test on a VM.

wonjoolee95:
Also FYI our CI having some issues, tracking in https://github.com/pytorch/xla/issues/3458.

JackCaoG:
@wonjoolee95 I merged https://github.com/pytorch/xla/pull/3459, can you rebase?

wonjoolee95:
> @wonjoolee95 I merged #3459, can you rebase?

Thanks, just rebased with master.

JackCaoG (approved):
Thanks! Let's verify if python frame is working after Will's change merged

PR #3413, opened 2022-03-08
Fix xm.all_gather for non-empty groups

### LINKED ISSUE #3412: xm.all_gather issue with non-empty groups
## 🐛 Bug

`torch_xla.core.xla_model.all_gather` function fails when its `groups` arg is non-empty. This is causing our data parallel tests to fail with the following error:
```
 result = torch_xla._XLAC._xla_all_gather(value, token, dim, shard_count,
      6                                            groups or [])

TypeError: _xla_all_gather(): incompatible function arguments. The following argument types are supported:
    1. (arg0: at::Tensor, arg1: _XLAC.IrValue, arg2: int, arg3: int, arg4: list) -> tuple
```

The non-empty `groups` forces `shard_count=None` ([code](https://github.com/pytorch/xla/blob/e39c40506bc87e5d291f458e5cc078885302e07d/torch_xla/core/xla_model.py#L606)), and it's breaking the call.

## To Reproduce
Run a call to 
```
def f(rank):
  ...
  value = torch.tensor([rank], dtype=torch.float).to(device)
  groups = [[0, 1, 2, 3], [4, 5, 6, 7]]
  result = xm.all_gather(value, dim=0, groups=groups)
  print(result)

xmp.spawn(f, args=())
```

## Expected behavior

It should perform 4-way concatenation on the two groups:
```
tensor([0., 1., 2., 3.])
tensor([0., 1., 2., 3.])
tensor([0., 1., 2., 3.])
tensor([0., 1., 2., 3.])
tensor([4., 5., 6., 7.])
tensor([4., 5., 6., 7.])
tensor([4., 5., 6., 7.])
tensor([4., 5., 6., 7.])
```

## Environment

 - Reproducible on XLA backend [CPU/TPU]: TPU
 - torch_xla version: `r1.11`

yeounoh:
Resolve #3412

### DIFF
--- a/test/test_mp_all_gather.py
+++ b/test/test_mp_all_gather.py
@@ -8,21 +8,41 @@ import torch_xla.distributed.xla_multiprocessing as xmp
 
 def _mp_fn(index):
   device = xm.xla_device()
+  world_size = xm.xrt_world_size()
   if xm.xla_device_hw(device) in ('TPU', 'GPU'):
-    world_size = xm.xrt_world_size()
+    # Testing with a single replica group
     ordinal_tensor = torch.tensor([index], dtype=torch.float).to(device)
-    result = xm.all_gather(ordinal_tensor)
+    result = xm.all_gather(ordinal_tensor, dim=0)
 
     cpu_result = result.cpu()
     expected = torch.arange(0, world_size, dtype=torch.float)
     if not cpu_result.allclose(expected):
       print('xm.all_gather() produced wrong reductions', file=sys.stderr)
-      print('[{}] {}'.format(index, cpu_result), file=sys.stderr)
+      print(f'[{index}] {cpu_result}', file=sys.stderr)
       sys.exit(1)
+
+    # Testing with two replica groups
+    if world_size % 2 == 0 and world_size > 1:
+      mp_groups = [[n for n in range(world_size) if n % 2 == 0],
+                   [n for n in range(world_size) if n % 2 == 1]]
+      group_size = len(mp_groups[0])
+      replica_id = int(index % 2 == 1)
+
+      result = xm.all_gather(ordinal_tensor, dim=0, groups=mp_groups)
+
+      cpu_result = result.cpu()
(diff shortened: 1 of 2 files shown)


yeounoh:
I am still testing, DO_NOT_MERGE_YET

yeounoh:
Finished testing, removing DO_NOT_MERGE_YET label. This needs to be picked up by `r1.11`

test/test_mp_all_gather.py
```diff
@@ -8,21 +8,41 @@
 
 def _mp_fn(index):
   device = xm.xla_device()
+  world_size = xm.xrt_world_size()
   if xm.xla_device_hw(device) in ('TPU', 'GPU'):
-    world_size = xm.xrt_world_size()
+    # Testing with a single replica group
     ordinal_tensor = torch.tensor([index], dtype=torch.float).to(device)
-    result = xm.all_gather(ordinal_tensor)
+    result = xm.all_gather(ordinal_tensor, dim=0)
```
JackCaoG, line 15: why do we add `dim=0` here, is it the default?

JackCaoG (approved):
Thanks! Please verify that this change fixed the DLRM test.

test/test_mp_all_gather.py
yeounoh: It's default, but to be explicit.

test/test_mp_all_gather.py
```diff
@@ -8,21 +8,41 @@
...
+    else:
+      print(
+          f'Failed to create two replica groups with {world_size} replicas',
+          file=sys.stderr)
+
   else:
-    print(
-        'Default device {} is not a TPU or GPU device'.format(device),
-        file=sys.stderr)
+    print('{} is not a TPU or GPU device'.format(device), file=sys.stderr)
```
hjm-aws, line 45: Nit: you can use a shorter version: `f'{device} is not a TPU or GPU device'`

hjm-aws (approved):
Thank you Yeounoh for spotting and fixing this!

PR #3417, opened 2022-03-10
Update circleci ubuntu image version

yeounoh:
Our Circle CI machines use Ubuntu 16.04 as base OS version, and it's being deprecated: 
```
This job is either using an Ubuntu 14.04 image, an Ubuntu 16.04 image, or the default Ubuntu 14.04-based image by specifying machine:true in config. These images (including the default image) will be deprecated May 31, 2022. Upgrade your image from Ubuntu 14.04 or 16.04 to 20.04.
```

This shouldn't affect the actual build & test workloads, as we run our customer docker image inside the host machine; but we will need to update the host machine OS and make sure nvidia-smi docker-nvidia run for GPU testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,7 +100,8 @@ run_build: &run_build
   - run:
       name: "Setup custom environment variables"
       command: |
-        pyenv global 3.6.5
+        pyenv install 3.7.0
+        pyenv global 3.7.0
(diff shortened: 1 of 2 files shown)


JackCaoG (approved):
Thanks!

JackCaoG:
A bit hesitant to merge it today.. Let's do that Monday morning :D. WDYT?

PR #3428, opened 2022-03-17
Updated colab tutorials to PyTorch 1.11

chandrasekhard2:
Updated colab tutorials to PyTorch 1.11. Added flexibility to install GPU PyTorch wheel and dependencies.

### DIFF
--- a/contrib/colab/DC-GAN.ipynb
+++ b/contrib/colab/DC-GAN.ipynb
@@ -66,21 +66,38 @@
     {
       "cell_type": "code",
       "metadata": {
-        "id": "OApBOAe1fpH_",
-        "colab_type": "code",
-        "colab": {}
+        "id": "OApBOAe1fpH_"
       },
       "source": [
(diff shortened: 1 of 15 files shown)


yeounoh (approved):
LGTM

JackCaoG:
Do we manually tested it? For some context we should work with colab team to preinstall pt/xla instead of we asked user to manually install upon init.

We did that for 1.9 because we do not have a working version of wheel for 1.10. I am OK with we do it now but we should remove the manual installation after wheel is preinstalled.

yeounoh:
> Do we manually tested it? For some context we should work with colab team to preinstall pt/xla instead of we asked user to manually install upon init.
> 
> We did that for 1.9 because we do not have a working version of wheel for 1.10. I am OK with we do it now but we should remove the manual installation after wheel is preinstalled.

@chandrasekhard2 manually tested the tutorials with the new wheels & instructions. I think we can first unblock colab tutorials with this PR, and make another pass once pt/xla is pre-installed (e.g., remove `!pip install ... torch-xla ...` step at the beginning).

PR #3439, opened 2022-03-23
Patch view change into 1.10

JackCaoG:
This is to backport https://github.com/pytorch/xla/pull/3411 into 1.10, the goal is to unblock researcher from keep using 1.10 given 1.11 currently has some speed regression.


JackCaoG:
oh well.. this is a bit awkward. The base ci image is removed.. I tested the build locally through.

JackCaoG:
I will just use this branch to build a patched wheel, not going to merge this into 1.10

PR #3442, opened 2022-03-23
Rephrase `all_reduce` docstring to clarify the condition for in-place op

thisisalbertliang:
Currently, the docstring for `torch_xla.core.xla_model.all_reduce` might be a bit misleading. In particular, it does not make it clear that the all-reduce op is NOT in-place when `inputs` is a single tensor.

For PyTorch users who have been using [`torch.distributed.all_reduce`](https://github.com/pytorch/pytorch/blob/d583f9c9d2bd0cc5fb5d369cd498bd62b62c43fe/torch/distributed/distributed_c10d.py#L1253) in the past, this could be especially confusing because `torch.distributed.all_reduce` always operates in-place, which leads to programmer bugs such as b/226225366.

This PR rephrases the docstring for `all_reduce` to make it explicit that the all-reduce op is NOT inplace when a single `torch.Tensor` is passed.

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -527,7 +527,7 @@ def _host_all_reduce(reduce_type, inputs, cctx, scale=None):
 
 
 def all_reduce(reduce_type, inputs, scale=1.0, groups=None, cctx=None):
-  """Performs an inplace reduce operation on the input tensor(s).
+  """Performs an all-reduce operation on the input tensor(s).
 
   Args:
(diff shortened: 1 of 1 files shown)


JackCaoG:
failure is unrelated, should be fixed by https://github.com/pytorch/xla/commit/e848a472f0897435f67559a969e658c8af658372

miladm:
Looks like we no longer need this PR. @thisisalbertliang?

thisisalbertliang:
@miladm 

I think we should still merge this PR. It's a small change to make the `xm.all_reduce` docstring less confusing.

PR #3450, opened 2022-03-25
Fix aten::gather

ymwangg:
This PR fixed a bug in the lowering of aten::gather that caused the deberta model to be 6.8x slower than PyTorch.

It looks like the [non sparse implementation](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/client/lib/slicing.cc#L171-L202) of xla::TorchGather is very inefficient. For a typical deberta workload `input=f32[10,10,512,512]` and `index=u32[10,10,512,512]`, this algorithm will first broadcast `index` to `u32[10,10,512,512,512]` followed by a reduction to `f32[10,10,512,512]`.

I tested the performance of gather with  `input f32[10,10,N,N]` and `index u32[10,10,N,N]` for `N=64,128,256,512` on GPU. In all cases, the original sparse implementation is orders of magnitude better than the non-sparse implementation. So maybe the default sparse implementation should always be used, though it's possible the non-sparse version is better on TPU.

| N          | sparse (us) | non-sparse (us) |
| ------  | ------ | ------  |
| 512      | 251     | 29664 |
| 256     | 81        | 5360   |
| 128      | 27       | 830     |
| 64       | 12        | 252     |

Below shows the performance of `microsoft/deberta-base` model with a batch_size of 10 on Nvidia V100.
|   Test   |  Throughput (it/s) |
| ---- | ------ |
| torch | 3.22 |
| torch_xla |  0.47 |
| torch_xla + this PR | 4.84 |

### DIFF
--- a/torch_xla/csrc/ops/gather.cpp
+++ b/torch_xla/csrc/ops/gather.cpp
@@ -9,24 +9,11 @@
 namespace torch_xla {
 namespace ir {
 namespace ops {
-namespace {
-
-xla::Shape NodeOutputShape(const Value& input, const Value& index,
-                           int64_t dim) {
-  auto lower_for_shape_fn =
-      [&](absl::Span<const xla::XlaOp> operands) -> xla::XlaOp {
-    return xla::TorchGather(operands[0], operands[1], dim,
-                            IsSparseGather(operands[0], operands[1], dim));
-  };
-  return InferOutputShape({input.xla_shape(), index.xla_shape()},
-                          lower_for_shape_fn);
-}
-
-}  // namespace
 
 Gather::Gather(const Value& input, int64_t dim, const Value& index)
     : Node(torch::lazy::OpKind(at::aten::gather), {input, index},
-           [&]() { return NodeOutputShape(input, index, dim); },
+           xla::ShapeUtil::MakeShape(input.xla_shape().element_type(),
+                                     index.xla_shape().dimensions()),
            /*num_outputs=*/1, torch::lazy::MHash(dim)),
       dim_(dim) {}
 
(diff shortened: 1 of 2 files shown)


ronghanghu:
This seems great! I wonder if it makes any performance difference on TPUs? (We are using `torch.gather` heavily in our internal codebase for TPUs and it would be great to see if it also gets better on TPUs.)

JackCaoG:
@miladm maybe we can test the performance on TPU while reviewing this pr?

miladm:
Sounds good @JackCaoG. I will take a pass at the performance aspect as well.

JackCaoG:
Hey sorry to let this pr to stale for a long time, we will try to get back to this one soon.

miladm:
Thanks @ymwangg for your contribution!

I observe a `3.3x` speedup on `TPUv3` when running the following micro-benchmark:

```
import torch
import torch_xla
import time

def _mp_fn(rank: int):
  for x in range(1000):
    t1 = torch.randn(500, device=xm.xla_device())
    t2 = xm.all_gather(t1).flatten()
    t3 = t2.sum()
    xm.mark_step()

t1 = time.time()
xmp.spawn(_mp_fn, nprocs=8, start_method='fork')
print("time elapsed: ", (time.time() - t1))
```

**Test** | Runtime
--- | ---
torch_xla `master` | 117.12s
torch_xla + this PR | 35.07s

This test shows a **3.3x** speedup.

cc @JackCaoG @shauheen

miladm (approved):
Thanks @ymwangg! 

LGTM.

JackCaoG:
@ymwangg Can you resolve the conflict and rerun the CI? otherwise this pr is ready to merge

ymwangg:
Ok, rebase is done.

JackCaoG:
Thanks, I will merge the pr once all test passed.

PR #3454, opened 2022-03-28
Migrate to upstream torch::lazy::Value

wonjoolee95:
Migrate to upstream torch::lazy::Value. 

The only difference from the upstream `torch::lazy::Value` is that we have `xla::Shape`. Since we still need this, we've renamed the existing `shape()` functions to `xla_shape()` (similar to https://github.com/pytorch/xla/pull/3402).

### DIFF
--- a/torch_xla/csrc/ir.h
+++ b/torch_xla/csrc/ir.h
@@ -62,7 +62,7 @@ using OutputMap =
 // Represents an input/operand for a Node object.
 struct Value : public torch::lazy::Value {
   Value() = default;
-  Value(NodePtr node, size_t index = 0) : node(std::move(node)), index(index) {}
+  Value(NodePtr node, size_t index = 0) : torch::lazy::Value(std::dynamic_pointer_cast<torch::lazy::Node>(node), index), node(std::move(node)), index(index) {}
 
   // Retrieves the shape of this value. If the IR Node generating the value is a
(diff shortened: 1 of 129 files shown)


torch_xla/csrc/ir.h
```diff
@@ -60,16 +60,20 @@ using OutputMap =
     std::unordered_map<torch::lazy::Output, T, torch::lazy::Output::Hasher>;
 
 // Represents an input/operand for a Node object.
-struct Value {
+struct Value : public torch::lazy::Value {
   Value() = default;
-  Value(NodePtr node, size_t index = 0) : node(std::move(node)), index(index) {}
+  Value(NodePtr node, size_t index = 0)
+      : torch::lazy::Value(std::dynamic_pointer_cast<torch::lazy::Node>(node),
```
wonjoolee95, line 66: The `dynamic_pointer_cast` here is needed because our `torch_xla::ir::NodePtr` is still pointing to `torch_xla::ir::Node`. 

I can follow-up with migrating to the upstream `torch::lazy::NodePtr` (was pushed off from the `Node` PR since it required many file changes) if we think it's necessary. Doing so will also help us to remove many of the functions in our `torch_xla:::ir::Value`.

wonjoolee95: Agreed, sounds good. I'll merge this PR for now and follow-up with the `NodePtr`.

Issue #3434, opened 2022-03-21
RunPostOrder and FetchTensorData methods adds overhead and prevents complete pipelining of two executions


aws-rhsoln:
## 🐛 Bug

When training model with multiple workers, two `ExecuteComputation` are not completely pipelined. The two `ExecuteComputation` are separated by execution of two methods RunPostOrder and FetchTensorData. These two methods end up taking around 5-7% of the total execution. Here is the profile from the MNIST script ran on GPU instance with 8 workers.
![Screen Shot 2022-03-21 at 1 21 12 PM](https://user-images.githubusercontent.com/73188662/159359570-905a1966-186e-4318-a4d3-e2d4c9752722.png)

Here the highlighted area corresponds to the time spent in [RunPostOrder](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor.cpp#L1272) and [FetchTensorData](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor.cpp#L1312) methods. Each ExecuteComputation takes around 1.8ms.

I tried increasing the batchsize so that more time is spent on ExecutionComputation, and see if I can overlap the ExecuteComputation with the RunPostOrder and FetchTensorData, however, that did not help. The two methods seems to run only after the previous execution is completed. This could be because of the [device lock present in CollectSyncTensors](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor.cpp#L1180) (not completely sure though).

## To Reproduce

1. Apply the below patch on [MNIST training script](https://github.com/pytorch/xla/blob/master/test/test_train_mp_mnist.py) . The patch is mainly to add logic to simulate gradient accumulation and add profiling.
```
From d05f582cdf0b8ec97399c04fdcd1769bc44b2821 Mon Sep 17 00:00:00 2001
From: EC2 Default User <ec2-user@ip-172-31-36-175.ec2.internal>
Date: Mon, 21 Mar 2022 21:29:29 +0000
Subject: [PATCH] profiling

---
 test/test_train_mp_mnist.py | 18 ++++++++++--------
 1 file changed, 10 insertions(+), 8 deletions(-)

diff --git a/test/test_train_mp_mnist.py b/test/test_train_mp_mnist.py
index f4a2e8e4..ecf09309 100644
--- a/test/test_train_mp_mnist.py
+++ b/test/test_train_mp_mnist.py
@@ -24,7 +24,7 @@ import torch_xla.utils.utils as xu
[...]

PR #3457, opened 2022-03-30
Move device lock before the execution instead of tensor gathering

JackCaoG:
This pr ports https://github.com/pytorch/pytorch/pull/74864 to pytorch/xla and should help https://github.com/pytorch/xla/issues/3434. This will need some more testing but the concept seem right to me.

I think this optimization is based on the assumption that even if previous execution does not finish, we still truncate the pending IR of the Tensor to become a BackendData with the correct shape. The actual handle assignment to the BackendData wont' happen until the previous execution finish but that would be OK since we only need true data handle in next execution.

There is a catch that PlaceHolder and Ir truncating only happened when `force_xla_data` is true. AFAICT GetTensor will have `force_xla_data == false` and Synctensor will have `force_xla_data == true`. I think this is fine. If we consider a case of

```
a = torch.tensor(1)
b = torch.tensor(1)
c = a + b
print (c) ---> GetTensor, won't truncate c's IR
d = c + a
xm.mark_step() --> Synctensor, will truncate IR
```

we expect print not changing the IR of c anyway so there is no need to wait for that execution to finish before we finalize our second graph.

### DIFF
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -1316,7 +1316,7 @@ XLATensor::PostOrderData XLATensor::RunPostOrder(
   for (auto node : po_data.post_order) {
     const ir::ops::DeviceData* device_data = ir::ops::DeviceData::Cast(node);
     if (device_data != nullptr) {
-      if (device_data->data()->HasValue() == 0) {
+      if (!device_data->data()->HasValue()) {
         TensorCollectionBarrier(coll);
       }
       xla::ComputationClient::Data::OpaqueHandle handle =
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -1200,14 +1200,6 @@ XLATensor::SyncTensorCollection XLATensor::CollectSyncTensors(
   coll.config = config;
   coll.device = *unique_device;
   coll.indices.reserve(tensors.size());
-  TF_VLOG(4) << "Waiting on device barrier for device " << coll.device
-             << " ...";
-  {
-    XLA_TIMED("DeviceLockWait");
-    coll.unlocker = LockDevices(unique_device.AsSet());
-  }
-  TF_VLOG(4) << "Waiting on device barrier for device " << coll.device
-             << " done!";
   for (size_t i = 0; i < tensors.size(); ++i) {
     if (tensor_ids.insert(tensors[i].GetUniqueId()).second &&
         tensors[i].CurrentXlaData() == nullptr) {
@@ -1369,6 +1361,14 @@ std::shared_ptr<XLATensor::Async> XLATensor::ScheduleSyncTensorsGraph(
     ComputationCache::TypePtr cached_computation) {
   tensorflow::profiler::TraceMe activity(
       "ScheduleSyncTensorsGraph", tensorflow::profiler::TraceMeLevel::kInfo);
+  TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
+             << " ...";
+  {
+    XLA_TIMED("DeviceLockWait");
+    coll->unlocker = LockDevices({coll->device});
+  }
+  TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
+             << " done!";
   std::shared_ptr<Async> async = std::make_shared<Async>(
       coll, std::move(parameters_data), std::move(tensors_data),
       std::move(cached_computation));
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -996,7 +996,8 @@ std::vector<xla::ComputationClient::DataPtr> XLATensor::GatherTensorsXlaData(
 void XLATensor::TensorCollectionBarrier(SyncTensorCollection* coll) {
   std::string invalid_device(
       "Unknown0"); /* Temp solution to idetify unassigned devices */
-  if (coll->device.ToString().compare(invalid_device) == 0 || coll->unlocker.size() > 0) {
+  if (coll->device.ToString().compare(invalid_device) == 0 ||
+      coll->unlocker.size() > 0) {
     return;
   }
   TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -1302,7 +1302,7 @@ XLATensor::ComputationCache* XLATensor::GetComputationCache() {
 XLATensor::PostOrderData XLATensor::RunPostOrder(
     const std::vector<XLATensor>& tensors, SyncTensorCollection* coll) {
   absl::Span<const size_t> indices = coll->indices;
-  std::vector<const ir::Node*> roots;
+  std::vector<const torch::lazy::Node*> roots;
   roots.reserve(indices.size());
   for (auto index : indices) {
     ir::XlaValue ir_value = tensors.at(index).CurrentIrValue();
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -996,19 +996,17 @@ std::vector<xla::ComputationClient::DataPtr> XLATensor::GatherTensorsXlaData(
 void XLATensor::TensorCollectionBarrier(SyncTensorCollection* coll) {
   std::string invalid_device(
       "Unknown0"); /* Temp solution to idetify unassigned devices */
-  if (coll->device.ToString().compare(invalid_device) == 0) {
+  if (coll->device.ToString().compare(invalid_device) == 0 || coll->unlocker.size() > 0) {
     return;
   }
-  if (coll->unlocker.size() == 0) {
-    TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
-               << " ...";
-    {
-      XLA_TIMED("DeviceLockWait");
-      coll->unlocker = LockDevices({coll->device});
-    }
-    TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
-               << " done!";
+  TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
+             << " ...";
+  {
+    XLA_TIMED("DeviceLockWait");
+    coll->unlocker = LockDevices({coll->device});
   }
+  TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
+             << " done!";
 }
 
 std::vector<at::Tensor> XLATensor::GetTensorsOpByOp(
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -994,7 +994,7 @@ std::vector<xla::ComputationClient::DataPtr> XLATensor::GatherTensorsXlaData(
 }
 
 void XLATensor::TensorCollectionBarrier(SyncTensorCollection* coll) {
-  if (coll->barrier_applied == false) { /* I wonder if there can be a race condition on this variable */
+  if (coll->barrier_applied == false) {
     TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
                << " ...";
     {
@@ -1287,7 +1287,6 @@ std::shared_ptr<XLATensor::Async> XLATensor::TryRunCachedSync(
   XLA_VALUE_METRIC("TensorsGraphSize", po_data->post_order.size());
   TF_VLOG(5) << "TensorsGraphSize=" << po_data->post_order.size();
 
-  TensorCollectionBarrier(&coll); /* Temp location. Will eventually move inside ScheduleSyncTensorsGraph() */
   return ScheduleSyncTensorsGraph(
       tensors, coll, std::move(po_data->parameters_data),
       coll->device.toString(), std::move(cached_computation));
@@ -1314,29 +1313,12 @@ XLATensor::PostOrderData XLATensor::RunPostOrder(
   std::unordered_map<xla::ComputationClient::Data::OpaqueHandle, size_t>
       data_handles;
 
-  //TensorCollectionBarrier(coll);
-  /* Condtionally Run TensorCollectionBarrier() */
-  for (auto node : po_data.post_order) {
-    const ir::ops::DeviceData* device_data = ir::ops::DeviceData::Cast(node);
-    if (device_data != nullptr && device_data->data()->HasValue() == 0) {
-      TensorCollectionBarrier(coll);
-      std::cout << "barrier at 0: " << device_data->data()->HasValue() << std::endl; /* If this line returns a 0, we have a problem. */
-      break;
-    } else if (device_data == nullptr { /* Ideally, we don't want this block, though it causes a segfault for `TestAtenXlaTensor.test_bfloat16_float32_cast` */
-      TensorCollectionBarrier(coll);
-      std::cout << "barrier at 0 - b/c of nullptr" << std::endl;
-      break;
-    }
-  }
-  /* Added for testing - most likely will remove */
-  if (po_data.post_order.size() == 0) {
-      std::cout << "empty po_data.post_order" << std::endl;
-      TensorCollectionBarrier(coll);
-  }
-
   for (auto node : po_data.post_order) {
     const ir::ops::DeviceData* device_data = ir::ops::DeviceData::Cast(node);
     if (device_data != nullptr) {
+      if (device_data->data()->HasValue() == 0) {
+        TensorCollectionBarrier(coll);
+      }
       xla::ComputationClient::Data::OpaqueHandle handle =
           device_data->data()->GetOpaqueHandle();
       auto it = data_handles.find(handle);
@@ -1399,6 +1381,7 @@ std::shared_ptr<XLATensor::Async> XLATensor::ScheduleSyncTensorsGraph(
     ComputationCache::TypePtr cached_computation) {
   tensorflow::profiler::TraceMe activity(
       "ScheduleSyncTensorsGraph", tensorflow::profiler::TraceMeLevel::kInfo);
+  TensorCollectionBarrier(coll);
   std::shared_ptr<Async> async = std::make_shared<Async>(
       coll, std::move(parameters_data), std::move(tensors_data),
       std::move(cached_computation));
@@ -1724,7 +1707,6 @@ std::shared_ptr<XLATensor::Async> XLATensor::SyncTensorsGraphInternal(
       std::move(compile_result.computation));
   GetComputationCache()->Add(coll.hash, cached_computation);
 
-  TensorCollectionBarrier(&coll); /* Temp location. Will eventually move inside ScheduleSyncTensorsGraph() */
   return ScheduleSyncTensorsGraph(
       tensors, &coll, std::move(compile_result.parameters_data),
       compile_result.device.toString(), std::move(cached_computation));
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -1301,6 +1301,8 @@ XLATensor::ComputationCache* XLATensor::GetComputationCache() {
 
 XLATensor::PostOrderData XLATensor::RunPostOrder(
     const std::vector<XLATensor>& tensors, SyncTensorCollection* coll) {
+  tensorflow::profiler::TraceMe activity(
+      "RunPostOrder", tensorflow::profiler::TraceMeLevel::kInfo);
   absl::Span<const size_t> indices = coll->indices;
   std::vector<const torch::lazy::Node*> roots;
   roots.reserve(indices.size());
@@ -1316,6 +1318,8 @@ XLATensor::PostOrderData XLATensor::RunPostOrder(
   for (auto node : po_data.post_order) {
     const ir::ops::DeviceData* device_data = ir::ops::DeviceData::Cast(node);
     if (device_data != nullptr) {
+      /* Acceptable race condition: HasValue may return falsse. This is OK
+       * since the conditional barrier is a performance optimization. */
       if (!device_data->data()->HasValue()) {
         TensorCollectionBarrier(coll);
       }
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -996,7 +996,7 @@ std::vector<xla::ComputationClient::DataPtr> XLATensor::GatherTensorsXlaData(
 void XLATensor::TensorCollectionBarrier(SyncTensorCollection* coll) {
   static const std::string invalid_device(
       "Unknown0"); /* Temp solution to idetify unassigned devices */
-  if (coll->device.ToString().compare(invalid_device) == 0 ||
+  if (coll->device.toString().compare(invalid_device) == 0 ||
       coll->unlocker.size() > 0) {
     return;
   }
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -1287,6 +1287,7 @@ std::shared_ptr<XLATensor::Async> XLATensor::TryRunCachedSync(
   XLA_VALUE_METRIC("TensorsGraphSize", po_data->post_order.size());
   TF_VLOG(5) << "TensorsGraphSize=" << po_data->post_order.size();
 
+  TensorCollectionBarrier(&coll); /* Temp location. Will eventually move inside ScheduleSyncTensorsGraph() */
   return ScheduleSyncTensorsGraph(
       tensors, coll, std::move(po_data->parameters_data),
       coll->device.toString(), std::move(cached_computation));
@@ -1711,7 +1712,6 @@ std::shared_ptr<XLATensor::Async> XLATensor::SyncTensorsGraphInternal(
              << torch::lazy::HashToString(coll.hash);
   std::shared_ptr<Async> async = TryRunCachedSync(tensors, &coll, &po_data);
   if (async != nullptr) {
-    TensorCollectionBarrier(&coll);
     return async;
   }
 
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -993,6 +993,20 @@ std::vector<xla::ComputationClient::DataPtr> XLATensor::GatherTensorsXlaData(
   return result_tensors_data;
 }
 
+void XLATensor::TensorCollectionBarrier(SyncTensorCollection* coll) {
+  if (coll->indices.empty()) {
+    return;
+  }
+  TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
+             << " ...";
+  {
+    XLA_TIMED("DeviceLockWait");
+    coll->unlocker = LockDevices({coll->device});
+  }
+  TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
+             << " done!";
+}
+
 std::vector<at::Tensor> XLATensor::GetTensorsOpByOp(
     std::vector<XLATensor>* tensors) {
   SyncTensorsConfig config;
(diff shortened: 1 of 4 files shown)


wconstab:
> There is a catch that PlaceHolder and Ir truncating only happened when force_xla_data is true. AFAICT GetTensor will have force_xla_data == false and Synctensor will have force_xla_data == true. I think this is fine. If we consider a case of

Regarding this, I think the behavior is OK/correct but the naming could be improved so it is more obvious from the API names what is happening.

wconstab (approved):
LGTM, although I am also waiting to hear your results to confirm correctness/performance across your benchmark workloads

JackCaoG:
@wconstab @desertfire CPU and GPU both failed at test `TestParallelTensorMNIST`. This is likely a real error, I will take a look tmr. Let's hold on merging ltc pr for now.

JackCaoG:
@desertfire the failure was coming from opbyop executor, I fixed by adding the barrier. However there is one change I think we need to apply to ltc code, we need to check if `coll` is not empty since it might return early in https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/lazy_graph_executor.cpp#L615. Accessing device in this case will segfault.

JackCaoG:
@desertfire I found any bug that I don't have a very good solution. in `RunPostOrder` we actually access the real data handle 
https://github.com/pytorch/xla/blob/81f0845767d6d60350c69ed2660d953e0d386da8/torch_xla/csrc/tensor.cpp#L1290

This step is to make sure `po_data.parameters_data` has unique `xla_data`. The issue here is that true data handle won't be assigned until the previous execution finish in https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor.cpp#L1370

desertfire:
> GetOpaqueHandle

Hmm, the LTC behavior diverges here.
[GetHandle](https://github.com/pytorch/pytorch/blob/7aaa75af05aa0174ddeea370684aff38708d114e/torch/csrc/lazy/ts_backend/ts_backend_impl.h#L24) still works, because [CreateDataPlaceHolder](https://github.com/pytorch/pytorch/blob/7aaa75af05aa0174ddeea370684aff38708d114e/torch/csrc/lazy/ts_backend/ts_backend_impl.cpp#L162) would make sure that TSData is there.

If this change doesn't work for XLA, we can either add an option for control or derive a subclass of LazyGraphExecutor. Given how small this change is, I would say add an option and XLA can override its value. How do you think?

wconstab:
> If this change doesn't work for XLA, we can either add an option for control or derive a subclass of LazyGraphExecutor. Given how small this change is, I would say add an option and XLA can override its value. How do you think?

Well, we really want a solution for XLA since AWS is the first reporter of this issue.  (Even though you're right that we could land it for TS only and it would strictly be an improvement).

> This step is to make sure po_data.parameters_data has unique xla_data. The issue here is that true data handle won't be assigned until the previous execution finish

Hmm, you don't have a way before launching the computation to know how many data handles will be returned, or of them will be unique?

JackCaoG:
@desertfire the placeholder does not have a valid handle for xla at least, it will get filled later. the RunPostOrder want to access the real handle is the issue here.

JackCaoG:
Let me run some experiments. I think when two graph does not have any dependency, we can run `RunPostOrder` and access the real handle.

for example
```
a = torch.tensor(1)
b = torch.tensor(1)
c = a + b
xm.mark_step() 
d = c + a.           -----> c's handle is not ready yet, we need to block RunPostOrder
xm.mark_step() 
```

```
a = torch.tensor(1)
b = torch.tensor(1)
c = a + b
xm.mark_step() 
d = a*x.            -------> we can proceed
xm.mark_step() 
```

@wconstab the graph is being run in a post-order fashion, same tensor might be the input to multiple result. For example
```
a = torch.tensor(1)
b = torch.tensor(1)
c = a * 2 + b
d = a * 4
```

In this case a will show up in graph twice.. I think.. need to verify. I think checking the device handle is the easiest way since if the device handle is the same, we know they are the same (from runtime point of view). I don't know if there is ever a case that two tensor have the same handle. If this is never the case, we can probably do the dedup at Tensor level/Data level not at the handle level.

miladm:
Update:

I am trying the above test scenarios from @JackCaoG + the tests in `test_operations.py` to verify this implementation.

It turns out this implementation leads to a couple of segfault scenarios.

1. A source of failure is calling `GetOpaqueHandle` under `RunPostOrder` leads to a segfault. 
2. Another source of failure is caused by moving `TensorCollectionBarrier` past `SaveTensorsGraphInfo` in `SyncTensorsGraphInternal`. I am pin pointing the line that causes the error.

miladm:
It turns out when `coll.indices.empty()` is `true`, `XLATensor::SyncTensorsGraphInternal` returns a `nullptr`. Before doing so, I need to call `TensorCollectionBarrier(&coll)`. The current implementation calls `TensorCollectionBarrier()` in `RunPostOrder()`.I have a working implementation that passes all our existing tests. 

The `OpByOp` code path runs into similar seg-faults I am currently addressing.

torch_xla/csrc/tensor.h
```diff
@@ -1426,7 +1430,7 @@ class XLATensor {
       std::string device, ComputationCache::TypePtr cached_computation);
 
   static PostOrderData RunPostOrder(const std::vector<XLATensor>& tensors,
-                                    absl::Span<const size_t> indices);
+                                    SyncTensorCollection* coll);
```
wonjoolee95, line 1433: @miladm The build error seems maybe due to this header change. I think we need to update all the places where `RunPostOrder` is called in our code.

torch_xla/csrc/tensor.cpp
```diff
@@ -1273,8 +1280,9 @@ XLATensor::ComputationCache* XLATensor::GetComputationCache() {
 }
 
 XLATensor::PostOrderData XLATensor::RunPostOrder(
-    const std::vector<XLATensor>& tensors, absl::Span<const size_t> indices) {
-  std::vector<const torch::lazy::Node*> roots;
+    const std::vector<XLATensor>& tensors, SyncTensorCollection* coll) {
+  absl::Span<const size_t> indices = coll->indices;
+  std::vector<const ir::Node*> roots;
```
wonjoolee95, line 1285: This should be `std::vector<const torch::lazy::Node*> roots;`. I think this might've been the cause of the build error. Can you try updating it and build locally?

torch_xla/csrc/tensor.h
wonjoolee95: Disregard this, the comment above should fix the build failure I think.

torch_xla/csrc/tensor.cpp
miladm: Good catch @wonjoolee95. Thanks!

wconstab:
@miladm just catching up, sounds like you fixed at least one of the major failures- is there still something else failing or is it just a matter of cleaning up the PR at this point?

torch_xla/csrc/tensor.cpp
```diff
@@ -1649,12 +1662,13 @@ std::shared_ptr<XLATensor::Async> XLATensor::SyncTensorsGraphInternal(
...
   SyncTensorCollection coll = CollectSyncTensors(*tensors, config);
   if (coll.indices.empty()) {
+    TensorCollectionBarrier(&coll);
     return nullptr;
   }
+  PostOrderData po_data = RunPostOrder(*tensors, &coll);
   DebugUtil::SaveTensorsGraphInfo("ScheduleSyncTensorsGraph", *tensors,
                                   &coll.indices);
 
-  PostOrderData po_data = RunPostOrder(*tensors, coll.indices);
```
JackCaoG, line 1657: why do we move this line?

torch_xla/csrc/tensor.cpp
```diff
@@ -975,6 +975,20 @@ std::vector<xla::ComputationClient::DataPtr> XLATensor::GatherTensorsXlaData(
   return result_tensors_data;
 }
 
+void XLATensor::TensorCollectionBarrier(SyncTensorCollection* coll) {
+  if (coll->barrier_applied == false) {
```
JackCaoG, line 979: I think there was a check here about `coll-> indices `. I have a comment around https://github.com/pytorch/xla/pull/3457#issuecomment-1084100839, why is it removed?

miladm: Yup, I removed that check early on as it was causing seg faults. Now that we have a working implementation, I can revisit the check and confirm if it works as you had intended.

torch_xla/csrc/tensor.cpp
miladm: This line leads to a seg fault in cases where the conditional lock needed to be called. I moved it after the `RunPostOrder`. Do you have any concerns to share here?

miladm:
To compare the performance gain of this implementation, I ran it against [ResNet50](https://github.com/pytorch/xla/blob/master/test/test_train_mp_imagenet.py) and [MNIST](https://github.com/pytorch/xla/blob/master/test/test_train_mp_mnist.py) models. Neither model shows performance gains compared to the current PyTorch/XLA `master` commit. This is rather surprising because it suggests there has been no independent graphs that would run back to back. I'd like to hear other's thoughts on this. CC @JackCaoG @wconstab 

In light of the above observation, we can consider drop the conditional barriers approach, and replace it with an implementation that doesn't require dependency tracking via [`GetOpaqueHandle` API](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor.cpp#L1290-L1298). If so, to confirm we will have the expected gain in advance, I suggest we discuss a test scheme that verifies this solution closes the expected 7% MNIST performance gap.

wconstab:
> To compare the performance gain of this implementation, I ran it against [ResNet50](https://github.com/pytorch/xla/blob/master/test/test_train_mp_imagenet.py) and [MNIST](https://github.com/pytorch/xla/blob/master/test/test_train_mp_mnist.py) models.

Are you able to confirm with profiling that either of these models on your HW are showing bubbles in-between kernel executions on the accelerator due to this run-post-order locking issue?

JackCaoG:
`This is rather surprising because it suggests there has been no independent graphs that would run back to back` 

I think this is an incorrect assumption. It is possible that this `RunPostOrder` and `FetchTensorData` lag does not impact some model as much. You can add print statement to the conditional lock to confirm, or do a profile.

miladm:
Yup, I will extract the profile to confirm if the same performance observation made on GPU ([here](https://github.com/pytorch/xla/issues/3434)) holds true for TPU devices.

miladm:
The profiling numbers for this branch and `master` are attached below. 
+ Turns out the gap is similar between the two branch runs. Each `ExecuteComputation` takes around `4.4ms`.
+ The profiles suggest the observed gap on TPU is about `>2x` wider than GPU ([see here](https://github.com/pytorch/xla/issues/3434#issue-1175965260)). 
+ The profiles "seem to suggest" `CollectSyncTensors` takes less relative time on the TPU than GPU. 
+ For the TPU, a `MarkStep` takes around `0.4ms`.

**`master` branch profile:**

image

**`move_sync_lock` branch profile:** 

I was not able to capture the ExecutorComputation diagram as it was plotted many rows away from this profile row. Though, as expected, it started immediately after `StepMarker`.

image

miladm:
Update on the results shared [above](https://github.com/pytorch/xla/pull/3457#issuecomment-1102183456): 

+ Turns out the reason I observed no perf difference between `master` and this branch was because all `ExecuteComputation` triggers completed well before the next `SyncTensorsGraphsInternal` call. See below for fix.

To verify this PR gives the intended performance gain + respond to @will-cromar's questions above:

+ Raising `--batch_size=2048` on MNIST: this shows `ExecuteComputation` overlap with the next `SyncTensorsGraphsInternal` call. For this training batch size, I observe **8.4%** speedup
+ Raising `--batch_size=8192` on MNIST: I observe **16.3%** speedup.

As intended, the conditional barrier enables an `ExecuteComputation` call to overlap with the next graph processing cycle (e.g. with `RunPostOrder` as shown below).

image

torch_xla/csrc/tensor.cpp
JackCaoG: Can you verify `XLA_SAVE_TENSORS_FILE` still works after you switch the order?

torch_xla/csrc/tensor.cpp
```diff
@@ -975,6 +975,20 @@ std::vector<xla::ComputationClient::DataPtr> XLATensor::GatherTensorsXlaData(
...
+    TF_VLOG(4) << "Waiting on device barrier for device " << coll->device
+               << " ...";
+    {
+      XLA_TIMED("DeviceLockWait");
+      coll->unlocker = LockDevices({coll->device});
+      coll->barrier_applied = true;
```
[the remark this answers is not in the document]
JackCaoG: still want to avoid additional boolean if possible, since ltc will also want to make similar changes.

JackCaoG:
Thanks @miladm, this is great to know. If we can clean up the pr a bit based on my previous comments, I think we this one is ready to go

torch_xla/csrc/tensor.cpp
JackCaoG: the check is trying to avoid accessing empty `coll->device`, if we can prove that this will never be the case we don't need to check here.

miladm:
A PR is on its way.

torch_xla/csrc/tensor.cpp
miladm: + I tested `XLA_SAVE_TENSORS_FILE` against `test_operations.py`. All tests successfully pass, and the saved file contains several HLO graphs corresponding to `ScheduleSyncTensorsGraph`.
+ However, combining this flag with the OpByOp flags fails; this is a known [issue](https://github.com/pytorch/xla/issues/3485) however.

```
XLA_GET_TENSORS_OPBYOP=1 XLA_SYNC_TENSORS_OPBYOP=1 XLA_SAVE_TENSORS_FILE=/tmp/graphs.hlo XLA_SAVE_TENSORS_FMT=hlo python test_operations.py
```

torch_xla/csrc/tensor.cpp
miladm: It turns out `coll->indices.empty()` leads to seg-faults. The correct check is to skip the barrier when:

+ The device is unassigned
+ The barrier was previously held

torch_xla/csrc/tensor.cpp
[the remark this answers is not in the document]
miladm: Removed the boolean.

torch_xla/csrc/tensor.cpp
```diff
@@ -975,6 +975,23 @@ std::vector<xla::ComputationClient::DataPtr> XLATensor::GatherTensorsXlaData(
   return result_tensors_data;
 }
 
+void XLATensor::TensorCollectionBarrier(SyncTensorCollection* coll) {
+  std::string invalid_device(
```
JackCaoG, line 979: nit: `static const std::string`;

torch_xla/csrc/tensor.cpp
```diff
@@ -1284,9 +1297,15 @@ XLATensor::PostOrderData XLATensor::RunPostOrder(
   po_data.post_order = ir::Util::ComputePostOrder(roots, &po_data.emission_map);
   std::unordered_map<xla::ComputationClient::Data::OpaqueHandle, size_t>
       data_handles;
+
   for (auto node : po_data.post_order) {
     const ir::ops::DeviceData* device_data = ir::ops::DeviceData::Cast(node);
     if (device_data != nullptr) {
+      /* Acceptable race condition: HasValue may return falsse. This is OK
```
JackCaoG, line 1304: typo: `false`

torch_xla/csrc/tensor.cpp
```diff
@@ -1649,12 +1669,13 @@ std::shared_ptr<XLATensor::Async> XLATensor::SyncTensorsGraphInternal(
       "SyncTensorsGraphInternal", tensorflow::profiler::TraceMeLevel::kInfo);
   SyncTensorCollection coll = CollectSyncTensors(*tensors, config);
   if (coll.indices.empty()) {
+    TensorCollectionBarrier(&coll);
```
JackCaoG, line 1672: Do we know why the barrier here is necessary?

JackCaoG (commented):
mostly lgtm

torch_xla/csrc/tensor.cpp
JackCaoG: Please add a comment here explaining why it is necessary, otherwise lgtm.

JackCaoG:
Well, I can't approve my own pr, but lgtm.

torch_xla/csrc/tensor.cpp
miladm: Context: 
This check is necessary to ensure previous execution completes before exiting `SyncTensorsGraphInternal`. 

Specifically, this check prevents a corner case scenario that appeared in the eager mode tests in which  `get_handle()` causes a seg fault via the following call sequence: `GetTensors` -> `GetTensorsFused` -> `SyncTensorsGraphInternal` then `xla::ComputationClient::Get()->TransferFromServer()`. Although the issue is captured in our eager mode tests, it can happen in the default execution mode too.

JackCaoG:
@miladm can you rerun the CI? I am suspecting failure is related to https://github.com/pytorch/pytorch/pull/76264 which is already reverted.

philipturner:
This seems related to https://github.com/s4tf/s4tf/issues/14, which I am trying to fix. Swift for TensorFlow and PyTorch [share the same code base](https://arxiv.org/abs/2102.13267) for LazyTensor/XLA execution. The [pytorch/xla](https://github.com/pytorch/xla) directory `third_party/xla_client` matches the [s4tf/s4tf](https://github.com/s4tf/s4tf) directory `Sources/xla/xla_client`. The [pytorch/xla](https://github.com/pytorch/xla) directory `torch_xla/csrc` matches the [s4tf/s4tf](https://github.com/s4tf/s4tf) directory `Sources/xla/xla_tensor`. I'm having a crash when creating a `LazyTensorBarrier` (equal to `mark.step()` in PyTorch). It fetches `xla_data->HasValue()` somewhere and asserts that it isn't a `nullptr`. Unfortunately, it is a null pointer and I can't trace why that happens. The crash did not appear when compiled against TensorFlow 2.4, but it now crashes when I compile against TensorFlow 2.9.

JackCaoG:
@philipturner This might be a bit tricky to debug. I don't see why tf would affect `XrtData`. My best guess is empty XrtData was created using 
https://github.com/pytorch/xla/blob/f0efaf51df5b96dddff13cd6deabce34926fcba3/third_party/xla_client/xrt_computation_client.h#L148-L150

as a placeholder in 
https://github.com/pytorch/xla/blob/f0efaf51df5b96dddff13cd6deabce34926fcba3/torch_xla/csrc/tensor.cpp#L1405-L1408

but it never get an assignment in 
https://github.com/pytorch/xla/blob/f0efaf51df5b96dddff13cd6deabce34926fcba3/torch_xla/csrc/tensor.cpp#L1440-L1445


That being said.. this part of the logic is purely lazy tensor and XRT related.. not sure why tf will affect this. This might also be a race condition.

philipturner:
I synchronized S4TF's `x10/xla_client` directory up to this commit, although `x10/xla_tensor` (equivalent to `torch_xla/csrc`) might be a bit behind. This PR did not fix the crash. In S4TF, we don't need to wrap/unwrap the XLA data because it's always a `xla::ComputationClient::DataPtr`. I don't have any functions called `WrapXlaData` or `UnwrapXlaData`, and I don't think I need any such functions.

I located the third section of code you mentioned above. It's inside `XLATensor::ScheduleSyncTensorsGraph`. I modified it so it logged the state of tensors during each loop iteration. It's a bit messy, but the print output gets the job done.

```c++
      for (size_t i = 0; i < results.size(); ++i) {
        TF_VLOG(0) << "Starter ============";
        if (async->tensors_data[i] != nullptr) {
          TF_VLOG(0) << "Case 1";
          if (async->tensors_data[i]->HasValue()) {
            TF_VLOG(0) << "Case 1.1";
          } else {
            TF_VLOG(0) << "Case 1.2";
          }
        } else {
          TF_VLOG(0) << "Case 2";
          if (async->tensors_data[i]->HasValue()) {
            TF_VLOG(0) << "Case 2.1";
          } else {
            TF_VLOG(0) << "Case 2.2";
          }
        }

        if (async->tensors_data[i] != nullptr) {
          async->tensors_data[i]->Assign(*results[i]);
        } else {
          async->tensors_data[i] = std::move(results[i]);
        }
      }
```

This is all I found in the tutorial, up to the point where it crashed. It's "Case 1.2" every time - not a single 1.1, 2.1, or 2.2. That means the `xla::ComputationClient::DataPtr` comes in as a non-null pointer, but without any data. It returns `false` for `HasValue()`, then gets assigned some actual data in `->Assign(*results[i])`.

```
Test Case '-[x10Tests.XLATensorTests testLazyTensorBarrier]' started.
2022-07-08 15:19:00.977198: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1234] Starter ============
2022-07-08 15:19:00.977209: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1236] Case 1
2022-07-08 15:19:00.977211: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1240] Case 1.2
2022-07-08 15:19:00.977213: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1234] Starter ============
2022-07-08 15:19:00.977214: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1236] Case 1
2022-07-08 15:19:00.977252: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1240] Case 1.2
2022-07-08 15:19:00.977269: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1234] Starter ============
2022-07-08 15:19:00.977271: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1236] Case 1
2022-07-08 15:19:00.977273: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1240] Case 1.2
2022-07-08 15:19:00.977274: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1234] Starter ============
2022-07-08 15:19:00.977276: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1236] Case 1
2022-07-08 15:19:00.977277: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1240] Case 1.2
2022-07-08 15:19:00.977279: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1234] Starter ============
2022-07-08 15:19:00.977280: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1236] Case 1
2022-07-08 15:19:00.977281: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1240] Case 1.2
2022-07-08 15:19:00.977283: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1234] Starter ============
2022-07-08 15:19:00.977284: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1236] Case 1
2022-07-08 15:19:00.977285: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1240] Case 1.2
2022-07-08 15:19:00.977287: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1234] Starter ============
2022-07-08 15:19:00.977288: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1236] Case 1
2022-07-08 15:19:00.977290: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1240] Case 1.2
2022-07-08 15:19:00.977292: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1234] Starter ============
2022-07-08 15:19:00.977293: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1236] Case 1
2022-07-08 15:19:00.977295: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:1240] Case 1.2
2022-07-08 15:19:00.984498: F tensorflow/compiler/xla/xla_client/tf_logging.cc:26] tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:585 : Check failed: xla_data->HasValue() 
*** Begin stack trace ***
	tensorflow::CurrentStackTrace()
	swift_xla::XLATensor::GetXlaData()
	swift_xla::XLATensor::ToTensor(bool)
	XLATensor_materialize
...
*** End stack trace ***
Trying to access XLA data while an async operation is in flight: f32[]
error: Exited with signal code 6
```

For the first/second sections of code you mentioned, we initialize data a bit differently. But the overall result is the same, creating a `nullptr` for `handle_ptr`.

Code section 2:
```c++
      xla_data = xla::GetX10Device(tensor_device)
                     ->CreateDataPlaceholder(std::move(shape));
      tensor.SetXlaData(xla_data, config.sync_xla_data);
```

`CreateDataPlaceholder` is defined in a `xrt_computation_client.cc`, but we refactored a lot of things two years ago. In PyTorch, the definition does not appear in this C++ file.

```c++
class XrtComputationClient::XrtDevice : public ComputationClient::Device {
 public:
  ...
  DataPtr CreateDataPlaceholder(Shape shape) override {
    return std::make_shared<XrtData>(this, std::move(shape));
  }
}
```

S4TF version of your first code section from the previous comment:

```c++
  struct XrtData : public Data {
    XrtData(Device* device, Shape device_shape)
        : Data(device, std::move(device_shape)),
          handle_ptr(nullptr) {}
    XrtData(XrtDevice* device, Shape device_shape, int64_t handle);

    XrtData(XrtDevice* device, Shape device_shape, XrtHandlePtr handle);
```

philipturner:
Here's a more precise look at the crash site.

```c++
xla::ComputationClient::DataPtr XLATensor::GetXlaData() {
  bool up_to_date = true;
  xla::ComputationClient::DataPtr xla_data = CurrentXlaData();
  // My additions to existing code begin here.
  if ((xla_data != nullptr) && (xla_data->HasValue() == false)) {
    TF_VLOG(0) << "Encountered something that would cause the crash.";
    ApplyPendingGraph();
  }
  // My additions to existing code end here.

  ir::Value ir_value;
  if (up_to_date) {
    xla::ComputationClient::DataPtr xla_data = CurrentXlaData();
    if (xla_data != nullptr) {
      // This still fails, even though I applied the pending graph.
      XLA_CHECK(xla_data->HasValue())
          << "Trying to access XLA data while an async operation is in flight: "
          << xla_data->shape();
      return xla_data;
    }
  }
  if (ir_value) {
    AssignIrValue(std::move(ir_value));
  }
  if (data()->ir_value) {
    ApplyPendingGraph();
  } else {
    XLA_CHECK(data()->tensor_data);
    data()->xla_data = TensorToXlaData(*data()->tensor_data, GetDevice());
  }
  return data()->xla_data;
}
```

Terminal output

```
Test Case '-[x10Tests.XLATensorTests testLazyTensorBarrier]' started.
2022-07-08 15:59:11.153362: I tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:584] Encountered something that would cause the crash.
2022-07-08 15:59:11.160997: F tensorflow/compiler/xla/xla_client/tf_logging.cc:26] tensorflow/compiler/tf2xla/xla_tensor/tensor.cpp:593 : Check failed: xla_data->HasValue() 
*** Begin stack trace ***
	tensorflow::CurrentStackTrace()
	swift_xla::XLATensor::GetXlaData()
	swift_xla::XLATensor::ToTensor(bool)
	XLATensor_materialize
```

Repository branch: [philipturner/s4tf:fix-xla-bug](https://github.com/philipturner/s4tf/tree/fix-xla-bug)

`ApplyPendingGraph` does nothing because it only checks whether the `DataPtr` is null, not whether the handle is null. That's why it crashes anyways. Is there supposed to be a scenario where the `DataPtr` itself is non-null, but the underlying `handle_ptr` is null?

```c++
void XLATensor::ApplyPendingGraph() {
  DeviceBarrier(GetDevice());
  // This method is called to ensure that the tensor data is available on
  // device, so that a call to CurrentXlaData() returns a valid pointer.
  if (CurrentXlaData() == nullptr) {
    std::vector<XLATensor> tensors({*this});
    SyncTensorsGraph(&tensors, {}, /*wait=*/true, /*sync_xla_data=*/false);
  }
}
```

PR #3463, opened 2022-03-31
Start implementation of PjRtComputationClient.

will-cromar:
Add a very experimental integration with PjRt. PT/XLA will still use XRT unless you set the `PJRT_DEVICE=CPU` environment variable.

Currently only implemented with CPU client. More devices coming soon.

FYI @JackCaoG @skye


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -0,0 +1,182 @@
...
+}
+
+std::vector<std::string> PjRtComputationClient::GetAllDevices() const {
+  return PjRtDevicesToString(client->devices());
+}
+
+std::shared_ptr<std::vector<std::string>>
+PjRtComputationClient::GetReplicationDevices() {
+  return std::make_shared<std::vector<std::string>>(
+      PjRtDevicesToString(client->addressable_devices()));
```
JackCaoG, line 179: what' the difference between `addressable_devices` and `devices`?

third_party/xla_client/pjrt_computation_client.h
```diff
@@ -0,0 +1,157 @@
...
+  };
+
+  struct PjRtComputation : public Computation {
+    PjRtComputation(PjRtClient* self, XlaComputation computation,
+                    ProgramShape program_shape,
+                    std::vector<std::string> devices,
+                    xla::CompileOptions options)
+        : Computation(std::move(computation), std::move(program_shape),
+                      std::move(devices)),
+          executable(self->Compile(this->computation(), options).ValueOrDie()) {
```
JackCaoG, line 47: what's the motivation here of moving the compilation to the `PjRtComputation` constructor? I felt like leaving it out and explicitly taking a  `executable` might be more obvious. In the Xrt world, computation is return as the result of the `Compile` call.

JackCaoG (commented):
Thanks Will, we should add a line in https://github.com/pytorch/xla/blob/master/test/run_tests.sh#L95 to test the `test_operations.py` with the pjrt cpu client and see if it would work.

will-cromar:
I just ran `test_operations.py`, which uses methods I haven't implemented yet, e.g.

```
RuntimeError: ./tensorflow/compiler/xla/xla_client/pjrt_computation_client.h:134 : SetReplicationDevices not implemented
```

Do you mean to add a flag to the script to test with PjRt? I don't think we want to add these tests to our CI until I have fleshed out the implementation more.

will-cromar:
Actually, if I avoid throwing an error in `SetReplicationDevices`, the tests pass. Added an optional flag into `test_operations.py` to use PjRt. Do you want to enable it on the CI?

will-cromar:
Is it better to have a flag in `test_operations.py` to control the runtime, or just set environment variables (`PJRT_DEVICE=CPU` in this case) before we call it?

JackCaoG:
we should leave `test_operations.py` as it is and control the env var in `run_test.sh` like we do for other special modes.

third_party/xla_client/computation_client.cc
```diff
@@ -265,16 +266,27 @@ bool ParseEnvDevices(XrtComputationClient::Options* options) {
...
-  XrtComputationClient::Options options;
-  std::unique_ptr<tensorflow::tpu::TopologyProto> topology_proto;
-  if (!ParseEnvBasedTpuClusterConfig(&options) &&
-      !ParseEnvDeviceCounts(&options) && !ParseEnvDevices(&options) &&
-      !ParseMeshConfig(&options, &topology_proto)) {
-    XLA_ERROR() << "Missing XLA configuration";
+  std::unique_ptr<ComputationClient> client;
+
+  if (sys_util::GetEnvString(env::kEnvPjRtDevice, "") != "") {
+    PjRtComputationClient::Options options;
```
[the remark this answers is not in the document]
JackCaoG: OK, that's good. I don't expect us to support multi cpu since it is mostly for the test.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -0,0 +1,182 @@
...
+  datas.reserve(tensors.size());
+  for (auto& tensor : tensors) {
+    xla::Shape shape = tensor.shape;
+    xla::Literal literal(shape);
+    tensor.populate_fn(tensor, literal.untyped_data(), literal.size_bytes());
+
+    std::shared_ptr<xla::PjRtBuffer> buffer =
+        client->BufferFromHostLiteral(literal, client->addressable_devices()[0])
+            .ValueOrDie();
+    buffer->BlockHostUntilReady();
```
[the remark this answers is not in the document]
JackCaoG: sg.. I think it is OK to leave it right now. We can update the API upon next tf update.

third_party/xla_client/pjrt_computation_client.h
JackCaoG: maybe, I saw in the Xrt Compile it took the `vecvtor<CompileInstance>` and it is passed by value. Worse case we can also pass `CompileInstance` by value. It would also be helpful to get the stack trace when it segfault so we know the exact issue.

will-cromar: I made this constructor less clunky as you suggested and am not getting an error now. There must have been a mistake somewhere else.

JackCaoG (approved):
Thanks!

PR #3469, opened 2022-04-04
Add function in ir.cpp to update upstream env vars

wonjoolee95:
Fixes regression introduced by https://github.com/pytorch/xla/pull/3402 with python frames when `XLA_IR_DEBUG` flag is set. 

With this change, can confirm that python frame is added to the IR:
```
Python 3.8.10 (default, Mar 15 2022, 12:22:08) 
[GCC 9.4.0] on linux
>>> import torch
>>> import torch_xla
>>> a = torch.tensor([1, 2], device="xla:0")
>>> print(torch_xla._XLAC._get_xla_tensors_text([a]))
IR {
  %0 = s64[2]{0} xla::device_data(), location=<module>@<stdin>:1, device=CPU:0, ROOT=0
}
```

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -107,6 +107,10 @@ def _prepare_to_exit():
   if int(os.environ.get('PT_XLA_DEBUG', '0')):
     _summarize_fn_tracker()
 
+def _map_xla_env_vars_to_lazy():
+  _XLAC._map_xla_env_vars_to_lazy()
+
 
 atexit.register(_prepare_to_exit)
 _apply_patches()
+_map_xla_env_vars_to_lazy()
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -107,6 +107,7 @@ def _prepare_to_exit():
   if int(os.environ.get('PT_XLA_DEBUG', '0')):
     _summarize_fn_tracker()
 
+
 def _map_xla_env_vars_to_lazy():
   _XLAC._map_xla_env_vars_to_lazy()
 
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -33,6 +33,7 @@
 #include "torch/csrc/autograd/utils/wrap_outputs.h"
 #include "torch/csrc/autograd/variable.h"
 #include "torch/csrc/jit/python/pybind.h"
+#include "torch/csrc/lazy/core/config.h"
 #include "torch/csrc/lazy/core/ir_util.h"
 #include "torch_xla/csrc/XLANativeFunctions.h"
 #include "torch_xla/csrc/aten_xla_bridge.h"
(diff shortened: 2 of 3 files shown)


JackCaoG (commented):
Thanks Wonjoo. This is great.  I am thinking about making it a bit more generic.

What we can do is 
1. move the `SetXlaEnvVars` to `torch_xla/csrc/init_python_bindings.cpp`
2. Make a python binding for `SetXlaEnvVars`, check example [here](torch_xla/csrc/init_python_bindings.cpp)
3. Call the python binding in [here](https://github.com/pytorch/xla/blob/master/torch_xla/__init__.py#L66)
This way we auto map all future xla env var to pytorch one upon pt/xla init

wonjoolee95:
> Thanks Wonjoo. This is great. I am thinking about making it a bit more generic.
> 
> What we can do is
> 
> 1. move the `SetXlaEnvVars` to `torch_xla/csrc/init_python_bindings.cpp`
> 2. Make a python binding for `SetXlaEnvVars`, check example [here](torch_xla/csrc/init_python_bindings.cpp)
> 3. Call the python binding in [here](https://github.com/pytorch/xla/blob/master/torch_xla/__init__.py#L66)
>    This way we auto map all future xla env var to pytorch one upon pt/xla init

Thanks for the comments. Updated the function name to `MapXlaEnvVarsToLazy` and moved it to `torch_xla/csrc/init_python_bindings.cpp`. 

One thing, though, is that the `_XLAC` import happens after `_setup_debug_env`, so it seems like I can't add our new function there. So I just defined a new function `_map_xla_env_vars_to_lazy` and called it at the end of the file. Let me know if you think this is ok, @JackCaoG.

Also, can confirm that the python frame info gets correctly attached to the IR with this change:
```
Python 3.8.10 (default, Mar 15 2022, 12:22:08) 
[GCC 9.4.0] on linux
>>> import torch
>>> import torch_xla
>>> a = torch.tensor([1, 2], device="xla:0")
>>> print(torch_xla._XLAC._get_xla_tensors_text([a]))
IR {
  %0 = s64[2]{0} xla::device_data(), location=<module>@<stdin>:1, device=CPU:0, ROOT=0
}

```

JackCaoG (approved):
FYI @wconstab . I think we can update this the config flag setting mechanism. 

@wonjoolee95 I will approve and merge this pr for now. I think we should have a small test like https://github.com/pytorch/xla/blob/master/test/test_data_type.py to verify frame if there when the env var is set.

wconstab:
> FYI @wconstab . I think we can update this the config flag setting mechanism.

Are you proposing to use this 'map_' function to handle all the envs basically? Or something else, maybe I misunderstood.

JackCaoG:
@wconstab It is just a workaround for now. We need to maintain the `XLA_xxx` env var while LTC has a different set of `LAZY_xxx` env var. Doing a env var mapping seems unavoidable, wdyt?

wconstab:
Yea seems fine to me to have the legacy mapping, and it can live for as long as needed..

If we update our env var infra on the lazy core side, it would just provide a new way to set here, so that would have to be updated, i think that's fine too.  (e.g. Instead of calling `FLAGS_blah = xla_value` you might call `get_blah() = xla_value` )

PR #3473, opened 2022-04-05
Add test_env_var_mapper.py test

wonjoolee95:
As a follow-up to https://github.com/pytorch/xla/pull/3469, this PR adds a test to verify frame info when the `XLA_IR_DEBUG` env var is set. 

Created a new test `test/test_env_var_mapper.py` that uses `xla_tensors_report = torch_xla._XLAC._xla_tensors_report(0, str(xla_device))` to determine if frame info is attached to the IR when `XLA_IR_DEBUG` is set. For now, we only test the case when `XLA_IR_DEBUG` is enabled (we don't test when disabled, but this can be added as well if needed). This python test `test/test_env_var_mapper.py` will be extended as we map more XLA env vars to upstream. 

The `frame_info` is included in the `torch_xla._XLAC._xla_tensors_report` only if it's not empty, as defined in https://github.com/pytorch/xla/blob/master/torch_xla/csrc/init_python_bindings.cpp#L346. And the keyword `Frame` is attached to the `cout` by the upstream `torch::lazy` frames, as defined in https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/ir_metadata.cpp#L28. So if the keyward `Frame` is included in the `xla_tensors_report`, we can verify that `XLA_IR_DEBUG` env var has correctly been mapped to the upstream torch `FLAGS_torch_lazy_ir_debug` and that the frame info has been attached correctly. 

Here is what the `torch_xla._XLAC._xla_tensors_report` output looks like `frame_info` is not empty:
```
Tensor: id=1, shape=f32[2]{0}, device=Unknown0, ir_nodes=1
Frames:
  test_xla_ir_debug (/home/wonjoo/pytorch/xla/test/test_env_var_mapper.py:18)
  _callTestMethod (/usr/lib/python3.8/unittest/case.py:633)
  run (/usr/lib/python3.8/unittest/case.py:676)
  __call__ (/usr/lib/python3.8/unittest/case.py:736)
  run (/usr/lib/python3.8/unittest/suite.py:122)
  __call__ (/usr/lib/python3.8/unittest/suite.py:84)
  run (/usr/lib/python3.8/unittest/suite.py:122)
  __call__ (/usr/lib/python3.8/unittest/suite.py:84)
  run (/usr/lib/python3.8/unittest/runner.py:176)
  runTests (/usr/lib/python3.8/unittest/main.py:271)
  __init__ (/usr/lib/python3.8/unittest/main.py:101)
  <module> (/home/wonjoo/pytorch/xla/test/test_env_var_mapper.py:28)
IR {
  %0 = f32[2]{0} xla::device_data(), location=test_xla_ir_debug@test_env_var_mapper.py:19, device=TPU:0, ROOT=0
}
```

Test output:
```
Ran 1 test in 32.852s

OK

```

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -51,8 +51,8 @@ function run_downcast_bf16 {
   XLA_DOWNCAST_BF16=1 run_test "$@"
 }
 
-function run_xla_ir_debug_enabled {
-  echo "Running with XLA_IR_DEBUG enabled: $@"
+function run_xla_ir_debug {
+  echo "Running with XLA_IR_DEBUG: $@"
   XLA_IR_DEBUG=1 run_test "$@"
 }
 
@@ -114,7 +114,7 @@ function run_all_tests {
   run_xla_backend_mp python3 "$CDIR/test_torch_distributed_all_reduce_xla_backend.py"
   run_xla_backend_mp python3 "$CDIR/test_torch_distributed_multi_all_reduce_xla_backend.py"
   run_xla_backend_mp python3 "$CDIR/test_torch_distributed_reduce_scatter_xla_backend.py"
-  run_xla_ir_debug_enabled python3 "$CDIR/test_env_var_mapper.py"
+  run_xla_ir_debug python3 "$CDIR/test_env_var_mapper.py"
 }
 
 if [ "$LOGFILE" != "" ]; then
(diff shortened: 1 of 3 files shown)


test/run_tests.sh
```diff
@@ -51,6 +51,11 @@ function run_downcast_bf16 {
   XLA_DOWNCAST_BF16=1 run_test "$@"
 }
 
+function run_xla_ir_debug_enabled {
```
JackCaoG, line 54: nit, maybe just `run_xla_ir_debug`?

JackCaoG (approved):
Thanks @wonjoolee95 !

Issue #3423, opened 2022-03-15
new `all_gather` together w/ `reduce_scatter` causes GRPC error on v3-128 and v3-256 (nightly 20220308)


ronghanghu:
## 🐛 Bug

The nightly PyTorch XLA build (20220308 for torch, torchvision, torch_xla, and libtpu) gives an unexpected GRPC error when `all_gather` is used together with `reduced_scatter`.

After some debugging, I think this GRPC error is related to the all_gather update in https://github.com/pytorch/xla/pull/3275. The error goes away after I revert to the older all_gather implementation via all_reduce (in https://github.com/pytorch/xla/blob/v1.10.0/torch_xla/core/xla_model.py#L583-L615). Also, the weird thing is that the GRPC error happens only when all_gather is used together with reduce_scatter -- it doesn't happen when either op is used alone.

Specifically, when I run `xm.all_gather` on a v3-128 pod followed by `xm.reduce_scatter`, it consistently gives me the following error message:
```
2022-03-09 06:20:03 10.164.15.223 [8] 2022-03-09 06:20:03.763131: W tensorflow/core/distributed_runtime/rpc/grpc_remote_master.cc:157] RPC failed with status = "UNAVAILABLE: Connection reset by peer" and grpc_error_string = "{"created":"@1646806803.763047345","description":"Error received from peer ipv4:10.164.15.223:51011","file":"external/com_github_grpc_grpc/src/core/lib/surface/call.cc","file_line":1056,"grpc_message":"Connection reset by peer","grpc_status":14}", maybe retrying the RPC
```
The training process hangs and later crashes after this error happens.

**This error happens on both v3-128 and v3-256.** However, it doesn't happen on v3-8 or v3-32 and doesn't happen when using the older all_gather implementation via all_reduce. See details below.

## To Reproduce

1. Allocate a **v3-128 TPU VM pod** (e.g. `ronghanghu-v3-128` below) from `tpu-vm-pt-1.10` runtime and install the nightly 20220308 build **on all VM nodes** (e.g. through `gcloud alpha compute tpus tpu-vm ssh --worker all`) via
```
sudo pip3 uninstall -y torch torchvision torch_xla libtpu_nightly
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch-nightly+20220308-cp38-cp38-linux_x86_64.whl
[...]

PR #3474, opened 2022-04-06
All gather layout

JackCaoG:
Ping the layout for `all-gather`, this should solve https://github.com/pytorch/xla/issues/3423. I will let gpu test to verify that this does not break anything and test the pod with new wheel.

FYI @ronghanghu


ronghanghu:
Awesome, I'll try it out. Thanks for the fix!

hjm-aws (approved):

JackCaoG:
ehh.. test failed at gpu, I will try to take a look today

JackCaoG:
test actually passed on TPU front, I will try to repro it on gpu vm and debug a bit more

JackCaoG:
Chatted with Blake from the XLA team, TPU emitter actually does not supporting pinnging the layout for both `all_gather` and `reduce_scatter` then use them side by side. We should only pin the layout for `all_reduce` and leave everything else not pinned. I will start a separate pr to remove the layout pining for `reduce_scatter`.

PR #3475, opened 2022-04-06
Removing Duplicate Ostream Definitions for at::Scalar

Gamrix:
The duplicate definition of the the Ostream function in <ATen/core/Formatting.h> and in XLA was causing errors in the PR
https://github.com/pytorch/pytorch/pull/75050


JackCaoG (approved):
Thanks, I will let CI to finish but otherwise lgtm

PR #3478, opened 2022-04-07
Fix Linear Interpolation

antoniojkim:
Prior implementation was out-of-date and did not match [latest docs](https://pytorch.org/docs/stable/generated/torch.lerp.html). Fixing to match expected behaviour according to docs.

CC: @JackCaoG @will-cromar

### DIFF
--- a/torch_xla/csrc/helpers.cpp
+++ b/torch_xla/csrc/helpers.cpp
@@ -368,10 +368,9 @@ xla::XlaOp XlaHelpers::FlattenDimRange(xla::XlaOp input, int64_t start,
 xla::XlaOp XlaHelpers::LinearInterpolation(xla::XlaOp value0, xla::XlaOp value1,
                                            double alpha) {
   const xla::Shape& shape = XlaHelpers::ShapeOfXlaOp(value0);
-  xla::XlaOp one = xla::One(value0.builder(), shape.element_type());
   xla::XlaOp alpha_value =
       ScalarValue(alpha, shape.element_type(), value0.builder());
-  return value0 * alpha_value + value1 * (one - alpha_value);
+  return value0 + alpha_value * (value1 - value0);
 }
 
 xla::PrimitiveType XlaHelpers::PromoteType(xla::PrimitiveType type1,
--- a/torch_xla/csrc/ops/linear_interpolation.cpp
+++ b/torch_xla/csrc/ops/linear_interpolation.cpp
@@ -9,7 +9,7 @@ namespace ops {
 
 LinearInterpolation::LinearInterpolation(const Value& value,
                                          const Value& new_value, double alpha)
-    : Node(ir::OpKind(at::aten::lerp), {value, new_value}, value.xla_shape(),
+    : Node(torch::lazy::OpKind(at::aten::lerp), {value, new_value}, value.xla_shape(),
            /*num_outputs=*/1, torch::lazy::MHash(alpha)),
       alpha_(alpha) {}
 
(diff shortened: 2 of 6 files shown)


torch_xla/csrc/helpers.cpp
```diff
@@ -368,10 +368,9 @@ xla::XlaOp XlaHelpers::FlattenDimRange(xla::XlaOp input, int64_t start,
 xla::XlaOp XlaHelpers::LinearInterpolation(xla::XlaOp value0, xla::XlaOp value1,
                                            double alpha) {
   const xla::Shape& shape = XlaHelpers::ShapeOfXlaOp(value0);
-  xla::XlaOp one = xla::One(value0.builder(), shape.element_type());
   xla::XlaOp alpha_value =
       ScalarValue(alpha, shape.element_type(), value0.builder());
-  return value0 * alpha_value + value1 * (one - alpha_value);
+  return value0 + alpha_value * (value1 - value0);
```
will-cromar, line 373: Thanks for the fix! Can you update the header as well?

https://github.com/pytorch/xla/blob/cc43bf117fd795443116a0b13a467f7f2fb35a91/torch_xla/csrc/helpers.h#L110

torch_xla/csrc/ops/linear_interpolation.cpp
```diff
@@ -2,15 +2,14 @@
...
 #include "torch_xla/csrc/lowering_context.h"
-#include "torch_xla/csrc/ops/xla_ops.h"
 
 namespace torch_xla {
 namespace ir {
 namespace ops {
 
 LinearInterpolation::LinearInterpolation(const Value& value,
                                          const Value& new_value, double alpha)
-    : Node(xla_moving_average, {value, new_value}, value.xla_shape(),
```
JackCaoG, line 13: I think we should leave the opkind here as it is . `LinearInterpolation` seems to be doing very similar thing as lerp but we have a lerp op implemented separted. Having 2 node with same op kind is confusing. Maybe one day we can refactor the `LinearInterpolation` node completely but until then it is probably better to leave it as it is.

JackCaoG (approved):
Thanks!

JackCaoG:
`test_batchnorm_simple_average_xla_float32` failed on CPU, seems relevant for this change

antoniojkim:
> `test_batchnorm_simple_average_xla_float32` failed on CPU, seems relevant for this change

It looks like this is a test coming from PyTorch itself. So, the test needs to be fixed before this change can be merged in, but the test can't be fixed without merging this change in 😅. We have ourselves a circular dependency...

JackCaoG:
The failure seems to be coming from pr. Do you know which part of this pr break the upstream test? It is usually a signal of regression.

antoniojkim:
> The failure seems to be coming from pr. Do you know which part of this pr break the upstream test? It is usually a signal of regression.

I'm guessing that the test in question is specifically testing for the previous incorrect values. So, the change here is breaking that.

JackCaoG:
The test itself is not xla specified, it is created by pytorch team and runs on pytorch cpu and gpu too. I check https://raw.githubusercontent.com/pytorch/pytorch/master/test/test_nn.py and the error seem to be
```
self.assertNotEqual(running_mean1, zeros)
```

My guess it is a real error caused by different way of computing average. It seems like our old way of doing things although is not align with the document, but produce an expected result. I don't have cycle to repro this on my end so if you are also busy we can leave liner interpolation as it is for now.

antoniojkim:
> The test itself is not xla specified, it is created by pytorch team and runs on pytorch cpu and gpu too. I check https://raw.githubusercontent.com/pytorch/pytorch/master/test/test_nn.py and the error seem to be
> 
> ```
> self.assertNotEqual(running_mean1, zeros)
> ```
> 
> My guess it is a real error caused by different way of computing average. It seems like our old way of doing things although is not align with the document, but produce an expected result. I don't have cycle to repro this on my end so if you are also busy we can leave liner interpolation as it is for now.

Yeah, I'm okay with leaving it as is. Doesn't really affect me either way.

PR #3479, opened 2022-04-07
Only include `test/cpp/build/test_ptxla` when `BUILD_CPP_TESTS=1`

will-cromar:
Otherwise you get this error:

```
warning: install_data: setup script did not provide a directory for 'test/cpp/build/test_ptxla' -- installing right in 'build/bdist.linux-x86_64/egg'

error: can't copy 'test/cpp/build/test_ptxla': doesn't exist or not a regular file
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -347,9 +347,10 @@ setup(
         ],
     },
     data_files=[
-        'test/cpp/build/test_ptxla',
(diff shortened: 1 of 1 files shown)


JackCaoG (approved):
Thanks

PR #3481, opened 2022-04-07
Use upstream torch::lazy::Node in xla NodePtr

wonjoolee95:
Use `torch::lazy::Node` in `NodePtr` instead of the existing `torch_xla::ir::Node`.

### DIFF
--- a/test/cpp/test_ir.cpp
+++ b/test/cpp/test_ir.cpp
@@ -20,36 +20,46 @@ TEST(IrTest, TestScalarCreate) {
 TEST(IrTest, TestReplace) {
   ir::NodePtr scalar1 = ir::ops::ScalarOp(1.0, xla::F32);
   ir::NodePtr scalar2 = ir::ops::ScalarOp(2.0, xla::F32);
-  ir::Value add = scalar1 + scalar2;
-  EXPECT_EQ(scalar1->uses().size(), 1);
-  EXPECT_EQ(scalar2->uses().size(), 1);
+  ir::Value value1 = ir::Value(scalar1, 0);
+  ir::Value value2 = ir::Value(scalar2, 0);
+  ir::NodePtr add = value1 + value2;
+
+  EXPECT_EQ(dynamic_cast<ir::Node*>(scalar1.get())->uses().size(), 1);
+  EXPECT_EQ(dynamic_cast<ir::Node*>(scalar2.get())->uses().size(), 1);
 
   ir::NodePtr scalar3 = ir::ops::ScalarOp(3.0, xla::F32);
-  scalar1->ReplaceAllUsesWith(scalar3);
-  EXPECT_EQ(scalar1->uses().size(), 0);
-  EXPECT_EQ(scalar3->uses().size(), 1);
+  dynamic_cast<ir::Node*>(scalar1.get())->ReplaceAllUsesWith(scalar3);
 
-  add->ReplaceOperand(0, scalar1);
-  EXPECT_EQ(scalar1->uses().size(), 1);
+  EXPECT_EQ(dynamic_cast<ir::Node*>(scalar1.get())->uses().size(), 0);
+  EXPECT_EQ(dynamic_cast<ir::Node*>(scalar3.get())->uses().size(), 1);
+
+  dynamic_cast<ir::Node*>(add.get())->ReplaceOperand(0, scalar1);
+  EXPECT_EQ(dynamic_cast<ir::Node*>(scalar1.get())->uses().size(), 1);
 }
 
 TEST(IrTest, TestHash) {
   ir::NodePtr scalar1 = ir::ops::ScalarOp(1.0, xla::F32);
   ir::NodePtr scalar2 = ir::ops::ScalarOp(2.0, xla::F32);
-  ir::Value add1 = scalar1 + scalar2;
+  ir::Value value1 = ir::Value(scalar1, 0);
+  ir::Value value2 = ir::Value(scalar2, 0);
+  ir::Value add1 = value1 + value2;
 
   ir::NodePtr scalar3 = ir::ops::ScalarOp(1.0, xla::F32);
   ir::NodePtr scalar4 = ir::ops::ScalarOp(2.0, xla::F32);
-  ir::Value add2 = scalar1 + scalar2;
+  ir::Value value3 = ir::Value(scalar3, 0);
+  ir::Value value4 = ir::Value(scalar4, 0);
+  ir::Value add2 = value3 + value4;
 
   ir::NodePtr scalar5 = ir::ops::ScalarOp(11.0, xla::F32);
   ir::NodePtr scalar6 = ir::ops::ScalarOp(22.0, xla::F32);
-  ir::Value add3 = scalar5 + scalar6;
+  ir::Value value5 = ir::Value(scalar5, 0);
+  ir::Value value6 = ir::Value(scalar6, 0);
+  ir::Value add3 = value5 + value6;
 
   EXPECT_EQ(add1->hash(), add2->hash());
   EXPECT_NE(add1->hash(), add3->hash());
 
-  ir::Value sub = scalar1 - scalar2;
+  ir::Value sub = value1 - value2;
 
   EXPECT_NE(add1->hash(), sub->hash());
 }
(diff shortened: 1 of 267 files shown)


wonjoolee95:
Building the test is currently failing, specifically the `test_ir.cpp` test case:
```
TEST(IrTest, TestReplace) {
  ir::NodePtr scalar1 = ir::ops::ScalarOp(1.0, xla::F32);
  ir::NodePtr scalar2 = ir::ops::ScalarOp(2.0, xla::F32);
  ir::Value add = scalar1 + scalar2;
  EXPECT_EQ(scalar1->uses().size(), 1);
  EXPECT_EQ(scalar2->uses().size(), 1);
```
with the error message:
```
/workspace/pytorch/xla/test/cpp/test_ir.cpp:23:27: error: invalid operands to binary
      expression ('ir::NodePtr' (aka 'shared_ptr<torch::lazy::Node>') and 'ir::NodePtr')
  ir::Value add = scalar1 + scalar2;
                  ~~~~~~~ ^ ~~~~~~~
```
I wonder how this used to work before since we don't seem to have an `operator+` overloaded for even `ir::NodePtr`. I see that we have a `arithmetic_ir_ops.h` that overloads `operator+` here https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/arithmetic_ir_ops.h#L8, but this takes in `Value` as inputs and not `NodePtr`.

JackCaoG:
You need to rebase to bypass the build error. As the error above, it is due to https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/arithmetic_ir_ops.h#L8.

It is a bit weird since if you can compile that means the operator must works for `torch::xla::Value`.  I am unsure how operator on `torch::xla::Value` transferred to original `NodePtr` here.  Is it the only test failing?

wonjoolee95:
> You need to rebase to bypass the build error. As the error above, it is due to https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/arithmetic_ir_ops.h#L8.
> 
> It is a bit weird since if you can compile that means the operator must works for `torch::xla::Value`. I am unsure how operator on `torch::xla::Value` transferred to original `NodePtr` here. Is it the only test failing?

Just rebased with latest but still seeing the same error above. The test doesn't get ran here, since the building of the tests fail due to the error .

wonjoolee95:
Trying to run this on master without this change, I can confirm that `NodePtr operator+(const Value& node1, const Value& node2)` is getting called when we add two `NodePtr`. 

Added log to the `operator+`:
```
NodePtr operator+(const Value& node1, const Value& node2) {
  std::cout << "WONJOO: operator+" << std::endl;
  std::cout << "WONJOO: node1=" << node1->ToString() << std::endl;
  std::cout << "WONJOO: node2=" << node2->ToString() << std::endl;
```
And the output:
```
(base) jenkins@1ec624a6af03:/workspace/scripts$ source xlaIrTest.sh Replace     
2022-04-08 21:55:38.010002: I 1643924 tensorflow/core/tpu/tpu_api_dlsym_initializer.cc:116] Libtpu path is: libtpu.so
Note: Google Test filter = IrTest.TestReplace
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from IrTest
[ RUN      ] IrTest.TestReplace
WONJOO: operator+
WONJOO: node1=[] prim::Constant, value=1
WONJOO: node2=[] prim::Constant, value=2
[       OK ] IrTest.TestReplace (2 ms)
[----------] 1 test from IrTest (2 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (2 ms total)
[  PASSED  ] 1 test.
/workspace/scripts
```

JackCaoG:
If this is the only place, feel free to casted the NodePtr to `XlaNodePtr`. If it affects real pt/xla code we should look into how to resolve that.

wonjoolee95:
> If this is the only place, feel free to casted the NodePtr to `XlaNodePtr`. If it affects real pt/xla code we should look into how to resolve that.

@JackCaoG, thanks for the comment. Currently, this `test_ir.cpp` is the only place that explicitly defines `NodePtr` and does operations on it. However, casting this to xla `Node` ptr just for this test would require changing the return type of the `operator+` which I believe will affect real pt/xla code. I'm trying if I can update the existing `operator+` to accept `torch::lazy:Value` to solve this issue without us having to do the explicit cast in the test.

JackCaoG:
other thing we could try is
1. casted the NodePtr to XlaNodePtr in the test
2. do operation on  `XlaNodePtr` instead

If this does not work, wrap `NodePtr` in a `torch_xla::ir:: Value` and then do the operation.

We don't necessary need to keep support NodePtr arithmetic if it is only used in the test. Workaround it in the test as much as possible.

wonjoolee95:
This should be ready for review as the CI pass, locally tests succeeded.

test/cpp/test_ir.cpp
```diff
@@ -20,36 +20,46 @@ TEST(IrTest, TestScalarCreate) {
...
+  EXPECT_EQ(dynamic_cast<ir::Node*>(scalar1.get())->uses().size(), 1);
 }
 
 TEST(IrTest, TestHash) {
   ir::NodePtr scalar1 = ir::ops::ScalarOp(1.0, xla::F32);
   ir::NodePtr scalar2 = ir::ops::ScalarOp(2.0, xla::F32);
-  ir::Value add1 = scalar1 + scalar2;
+  ir::Value value1 = ir::Value(scalar1, 0);
+  ir::Value value2 = ir::Value(scalar2, 0);
+  ir::Value add1 = value1 + value2;
```
JackCaoG, line 45: nit, maybe just 
```
ir::Value add1 = ir::Value(scalar1, 0) + ir::Value(scalar2, 0);
```

same for other places.

torch_xla/csrc/ir.cpp
```diff
@@ -102,14 +102,18 @@ std::string Use::ToString() const {
   return ss.str();
 }
 
-const xla::Shape& Value::xla_shape() const { return node->xla_shape(index); }
-
-const xla::Shape& Value::xla_node_shape() const { return node->xla_shape(); }
+const xla::Shape& Value::xla_shape() const {
+  Node* casted = dynamic_cast<Node*>(node.get());
```
JackCaoG, line 106: Let's have a separate pr to rename all `torch_xla::Node` to `torch_xla::XlaNode` so it is not as confusing(given there is a `torch::lazy::Node` already)

torch_xla/csrc/ir.cpp
```diff
@@ -102,14 +102,18 @@ std::string Use::ToString() const {
...
+  return casted->xla_shape(index);
+}
 
-torch::lazy::hash_t Value::hash() const {
-  return torch::lazy::HashCombine(node->hash(), index);
+const xla::Shape& Value::xla_node_shape() const {
+  Node* casted = dynamic_cast<Node*>(node.get());
```
JackCaoG, line 111: this will likely incur some cost but I expect it to be small. Let's verify that on benchmark

torch_xla/csrc/ir.cpp
```diff
@@ -102,14 +102,18 @@ std::string Use::ToString() const {
...
+  return casted->xla_shape();
 }
 
+Node* Value::operator->() const { return dynamic_cast<Node*>(node.get()); }
```
JackCaoG, line 115: why is this necessary?

torch_xla/csrc/ir.h
```diff
@@ -155,7 +146,7 @@ class Node : public torch::lazy::Node {
 
   xla::Shape xla_shape_;
   // A node holds a real reference to its operands.
-  std::vector<NodePtr> operands_;
+  // std::vector<NodePtr> operands_;
```
JackCaoG, line 149: remove it if we no longer needs it

JackCaoG (changes requested):
Thanks Wonjoo! couple nits

torch_xla/csrc/ir.cpp
wonjoolee95: Sounds good, I'll follow-up with a separate PR.

torch_xla/csrc/ir.cpp
wonjoolee95: This was more of a convenience change because `operator->()` is often used in our code to call the `xla_shape()` from our `Value` like . Adding this will prevent us from manually doing the dynamic cast each time.

But after looking again, I noticed we only do this call couple times in our ops. So I've removed this operator and did manually dynamic casting instead.

wonjoolee95:
> Thanks Wonjoo! couple nits

Thanks for the comments, @JackCaoG ! Address all the comments with the newest commits, this should be ready for another round of review. Thanks!

JackCaoG (approved):
Thanks! Feel free to merge after all test passed.

Issue #3455, opened 2022-03-29
gradient checkpointing isn't effectively triggered and fails to save memory


ronghanghu:
## 🐛 Bug

Gradient checkpointing (activation checkpointing) via [`torch.utils.checkpoint.checkpoint`](https://pytorch.org/docs/stable/checkpoint.html#torch.utils.checkpoint.checkpoint) is supposed to save memory by recomputing the intermediate activations in the backward pass (rather than storing all intermediate activations of the entire computation graph) and has been working well in the vanilla PyTorch. However, I find that it fails to work in PyTorch XLA.

Specifically, gradient checkpointing via `torch.utils.checkpoint.checkpoint` does not result in any reduced memory consumption when running on TPUs -- The memory consumption is almost identical with and without checkpointing when measured via `xm.get_memory_info`. **And the more weird thing is that their runtime speed is also nearly identical, although gradient checkpointing is supposed to make the speed slower by trading off compute for memory.** (In a typical case, gradient checkpointing would lead to 1.33X slower runtime as the backward pass usually takes double the time of the forward pass, and gradient checkpointing results in calling the forward pass twice).

When monitoring the XLA metrics from `met.metrics_report()`, the `DestroyDataHandles` metrics are identical between the two cases with or without gradient checkpointing, but their `DestroyXlaTensor` metrics are notably different (see details below). Similarly, their `CreateDataHandles` metrics are the same but their `CreateXlaTensor` metrics are different. And their `ExecuteTime` metrics are also (nearly) identical.

And in practical use cases in PyTorch XLA on vision transformers, `torch.utils.checkpoint.checkpoint` also doesn't save any TPU memory for a large model or allow using a larger batch size. I suspect the XLA compiler somehow ignored the gradient checkpointing.

(A similar issue was previously reported in https://github.com/pytorch/xla/issues/1571.)

## To Reproduce

1. Allocate a v3-8 TPU VM with `tpu-vm-pt-1.10` runtime environment.
2. Save the following content to a file (e.g. `~/test_grad_checkpoint.py`). *It runs a dummy 64-layer network where each layer has two Conv2d + ReLU sub-modules, which should be able to benefit from gradient checkpointing in principle (since its inner conv activations no longer need to be stored when using checkpointing).*
```
import argparse
import torch
import torch.utils.checkpoint
import torch_xla.core.xla_model as xm
import torch_xla.debug.metrics as met


def run(grad_checkpoint):
    device = xm.xla_device()
[...]

PR #3482, opened 2022-04-08
Add optimization barrier to enable the gradient checkpointing

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3455. 

This pr introduced a new api `xm.optimization_barrier`, the usage is pretty straightforward, you call this api on the output of the forward pass. For example

```
        dummy_data = torch.zeros(64, 1024, 14, 14, device=device)
        optimizer.zero_grad()
        x = dummy_data
        for n_l, layer in enumerate(model):
            if n_l > 0 and grad_checkpoint:
                x = torch.utils.checkpoint.checkpoint(layer, x)
            else:
                x = layer(x)

        (x,) = xm.optimization_barrier([x])

        dummy_loss = x.sum()
        dummy_loss.backward()
        optimizer.step()
```

this ensures that all of its inputs(in this case the result of the forward pass) are evaluated before any operators that depend on the `optimization_barrier`'s outputs, and XLA is forbidden from eliding or moving non-trivial computations across `optimization_barrier` operators.

running a mini repo on my end, without `optimization_barrier`, I see
```
step 198, free memory after forward = 11514304
step 198, free memory = 11514304
step 199, free memory after forward = 11514304
step 199, free memory = 11514304
Metric: ExecuteTime
  TotalSamples: 200
  Accumulator: 41s329ms857.374us
  ValueRate: 590ms547.571us / second
  Rate: 2.85296 / second
  Percentiles: 1%=204ms379.962us; 5%=205ms526.820us; 10%=205ms608.515us; 20%=205ms785.190us; 50%=205ms188.236us; 80%=206ms065.475us; 90%=207ms667.218us; 95%=218ms647.765us; 99%=259ms730.178us
```

with `optimization barrier` I saw
```
step 198, free memory after forward = 15487456
step 198, free memory = 15487456
step 199, free memory after forward = 15487456
step 199, free memory = 15487456
Metric: ExecuteTime
  TotalSamples: 200
  Accumulator: 12s261ms783.098us
  ValueRate: 763ms711.836us / second
  Rate: 12.4415 / second
  Percentiles: 1%=061ms649.505us; 5%=061ms756.333us; 10%=061ms832.865us; 20%=061ms923.435us; 50%=061ms077.979us; 80%=061ms220.387us; 90%=061ms284.215us; 95%=061ms348.149us; 99%=081ms290.276us
```

What is a bit puzzling is that execute time improved. I would expect execute time to increase since cse was not able to collapse the two forward pass computation. It is a nice surprise so I will take it.

fyi @ronghanghu

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -748,6 +748,19 @@ class TestDynamicShape(XlaTestCase):
     self.assertEqual(x_dim0_shape.item(), 3)
 
 
+class TestOptimizationBarrier(XlaTestCase):
+
+  def test_optimization_barrier_correctness(self):
+    device = xm.xla_device()
+    # only test optimization_barrier on TPU
+    if xm.xla_device_hw(device) != 'TPU':
+      return
+    x = torch.randn(5, 5, device=device)
+    y = torch.randn(5, 5, device=device)
+    (x1, y1) = xm.optimization_barrier([x, y])
+    self.assertEqual(x + y, x1 + y1)
+
+
 class TestDataType(XlaTestCase):
 
   def test_mixed_dtype_tuple(self):
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -1032,6 +1032,6 @@ def optimization_barrier(tensors):
   the gradient checkpointing.
 
   Args:
-    tensors (torch.Tensor): `torch.Tensor`s to add barrier to.
+    tensors (List[torch.Tensor]): List of `torch.Tensor` to add barrier to.
   """
   return torch_xla._XLAC._xla_optimization_barrier(tensors)
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -1024,3 +1024,14 @@ def get_memory_info(device):
     memory in KB) keys.
   """
   return torch_xla._XLAC._xla_memory_info(str(device))
+
+
+def optimization_barrier(tensors):
+  """Blocks xla compiler from moving computations across this barrier. The common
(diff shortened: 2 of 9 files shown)


JackCaoG:
It is a bit tricky to test op as it does not work on XLA:CPU, I want to unblock user first so I might merged this pr with only manual testing.

ronghanghu:
Thanks for the new API! A question regarding the usage

Suppose we have a model with a forward pass that takes inputs a, b, and c, and returns two variables x and y
```
x, y = torch.utils.checkpoint.checkpoint(layer, a, b, c)
```

To use this API and avoid fusing x and y with the backward pass, should I separately run
```
x = xm.optimization_barrier(x)
y = xm.optimization_barrier(y)
```
before my backward pass, or do I only need to call this API on one of x and y, or should I use it in another way?

JackCaoG:
You should do
```
x = xm.optimization_barrier(x)
y = xm.optimization_barrier(y)
```

That being said, It might be a good idea for me to make this api to accept a list of tensor instead of a single tensor.

ronghanghu:
I guess either interface is okay. If it takes a single tensor, I can also call it in one line like
```
x, y = map(xm.optimization_barrier, (x, y))
```
and apply similarly it to other cases (e.g. when the model output is a dictionary)

JackCaoG:
manually verified that test passed on TPU, will merge this today(after build passed)

JackCaoG:
failure is related to https://github.com/pytorch/xla/commit/3e3255e413ef01516fbc7112a8e53882ac0fb5f7 not being included, merge the pr to unblock user.

PR #3484, opened 2022-04-09
Remove reduce_scaterr layout pining

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3423. Currently we don't support pinging `reduce_scatter` layout follow by a `reduce_scatter`. We also can not pin the `all_gather` layout. 

FYI @ronghanghu


JackCaoG:
I am going to merge this pr now since it should not break anything and passed gpu test.

PR #3489, opened 2022-04-11
Update disabled test name for test_upsamplingBicubic2d_xla

wonjoolee95:
Tests `test_upsamplingBicubic2d_xla` have been part of our skip list but was incorrectly named previously. These tests were causing timeouts in the `python-ops` nightly tests.


JackCaoG (approved):
Thanks!

Issue #3138, opened 2021-09-21
[RFC] Exposing additional XLA collective communication primitives.


hjm-aws:
## 🚀 Feature
Add the ability to translate the following Collective Communication ops to native
 XLA instructions:
 * `all_gather`
 * `reduce_scatter`
 * `collective_permute`
 * `send`
 * `recv`

At the moment `all_gather` is implemented with `all_reduce` in PyTorch/XLA. The other 4 
operators are not yet implemented. We are proposing to change the current `all_gather` 
implementation to generate native XLA `all_gather` instruction, and to add native 
XLA support for `reduce_scatter`, `collective_permute`, `send`, and `recv` operations.

*We are intending to submit PRs for this proposal.*

## Motivation

Adding the ability to distributed training libraries (such as DeepSpeed) and 
Apps (such as MegatronLM) to utilize underlying XLA devices/accelerators to 
train very large models by calling `torch.distributed` API.

Currently MegatronLM and DeepSpeed use `torch.distributed` API to execute Collective 
Communication primitives, but `torch.distributed` does not support XLA. To enable 
MegatronLM and DeepSpeed to run on XLA, we are proposing to utilize PT/XLA, which is 
a bridge from PT to XLA, to provide the ability to translate the collective 
communication primitives listed above to native XLA instructions.

## Pitch

We would like users to be able to call the following APIs to insert the corresponding 
XLA instructions:

```
import torch_xla.core.xla_model as xm

xm.all_gather(tensor)
xm.reduce_scatter(xm.REDUCE_SUM, tensor, groups=[[0,1]])
xm.collective_permute(tensor, pairs=[[0,2],[1,3]])
xm.send(tensor, channel)
tensor = xm.recv(shape, channel)
```
## Alternatives

User could implement all these ops with `all_reduce`, but the efficiency would be too
low for them to be practical.

jysohn23:
FYI, we had added support for `ReduceScatter` collective (https://github.com/pytorch/xla/pull/3075) but had to roll back due to new TF version breaking our internal builds. We should be rolling forward soon.

jysohn23:
Also, we do expose the `CollectivePermute` collective: https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/collective_permute.cpp

hjm-aws:
Thank you Daniel for your comment.

[...]

PR #3493, opened 2022-04-13
Make Optimization Barrier in place

JackCaoG:
This will make `optimization barrier` less awkward. However I tested this implementation and see that it does not help the memory saving. 

Looking at the HLO graph. I think the issue is that
```
 %opt-barrier.9085 = f32[64,1024,14,14]{3,2,1,0} opt-barrier(f32[64,1024,14,14]{3,2,1,0} %maximum.9084)                                                                                                                                        
  %constant.9086 = f32[] constant(0)                                                                                                                                                                                                            
  %reduce.9092 = f32[] reduce(f32[64,1024,14,14]{3,2,1,0} %opt-barrier.9085, f32[] %constant.9086), dimensions={0,1,2,3}, to_apply=%AddComputation.9088 
```

optimization barrier HLO happened at the very end. It should be between forward and backward which will prevent xla compiler to merge forward and backward graph. I need to dig a bit more into this..

fyi @ronghanghu

PR #3494, opened 2022-04-13
Add send/recv support for https://github.com/pytorch/xla/issues/3138

hjm-aws:
I don't have e2e tests for TPU/GPU because I don't know how to compose channel_id for them. Any suggestions are welcome!


JackCaoG:
Thanks for the contribution. I will follow up with xla team to figure out how to set the channel id.

torch_xla/distributed/xla_backend.py
```diff
@@ -148,6 +148,9 @@ def send(self, tensors, dst_rank, tag=0):
     for t in tensors:
       channel_id = self.make_send_channel_id(dst_rank, tag)
       token = xm.send(t, channel_id)
+      # Make the sent tensor depend on the token, such that the `send`
+      # op can actually be built into the computation graph.
+      t.data.data = token
```
JackCaoG, line 153: so the assumption here is that this process will not use the `tensors` after it is being sent? Is this also specified in the pytorch ddp?

I find it a bit confusing that after the `send` the tensor becomes a token. If you want to force the dependency another thing you could try is use `xla::Tuple` and then untuple to bind the `token` and `input` (like what I do in https://github.com/pytorch/xla/pull/3493/files#diff-c561cfa200169af27f0f5ff1db5e5449e1398fce312de49e0a4c867efeea55bdR38). This will also force `send` and `token` show up in the HLO as long as one of the two is used in later graph.

JackCaoG: for example you can do something like
```
BuildSendWithToken(){
  xla::XlaOp result_token = xla::SendWithToken(input, token, channel_handle);

    xla::XlaOp tuple_res = xla::Tuple(inputs[0].builder(), {result_token, input});
    xla::XlaOp new_input = xla::GetTupleElement(tuple_res, 1)

  return {new_input, result_token};
}
```

then I think you can do 
```
t.data.data = new_input
```

JackCaoG (commented):
mostly lgtm, some nits

JackCaoG:
@hjm-aws Can you try to rebase and see if test will pass?

JackCaoG (approved):
Thanks @hjm-aws !

hjm-aws:
Thank you Jack!

Issue #3427, opened 2022-03-16
Lower torch.roll


rwightman:
## 🚀 Feature
A lowering of torch.roll would be helpful so code model code using it can be common across gpu/tpu. 

```
pt-xla-profiler: Op(s) not lowered: aten::roll,  Please open a GitHub issue with the above op lowering requests.
pt-xla-profiler: TransferFromServerTime too frequent: 20 counts during 1 steps
```

## Additional context

Roll can be implemented fairly easily with cat across relevant dimensions, however, torch.roll can be specified across multiple dim, which are done in the same CUDA kernel so making code common by reverting to cat will slow down models on GPU.

JackCaoG:
Thanks Ross, we will take a look.

cicirori:
@rwightman 

I implemented a version of roll's lower mainly refer to [tf2xla roll op](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tf2xla/kernels/roll_op.cc), and in the meantime I read [torch roll_cuda implementation]( https://github.com/pytorch/pytorch/blob/344ecb2e7103b4d92ec3859318643a66dd50a4ed/aten/src/ATen/native/cuda/TensorTransformations.cu#L102)

Unfortunately with multiple dimensional rolls, torch also seems to run a separate cuda kernel for each shift/dim pair? 

I haven't analyzed the tf2xla implementation in detail, but it looks like it's not very efficient in terms of memory usage, I'm not sure.

JackCaoG:
roll lowering is merged, closing this issue.

PR #3505, opened 2022-04-15
lowering roll op

cicirori:
To make better use of timm, roll op needs to be supported. https://github.com/pytorch/xla/issues/3427

I implemented a version of roll's lower mainly refer to [tf2xla roll op](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tf2xla/kernels/roll_op.cc), and in the meantime I read [torch roll_cuda implementation](https://github.com/pytorch/pytorch/blob/344ecb2e7103b4d92ec3859318643a66dd50a4ed/aten/src/ATen/native/cuda/TensorTransformations.cu#L102)

The main difference between the tf and pytorch versions is the support for flatten mode, which I have modified accordingly.


JackCaoG:
@cicirori Thanks, give me ping when all test passed and I can take a pass.

torch_xla/csrc/ops/roll.cpp
```diff
@@ -0,0 +1,81 @@
...
+    offset = ((offset % dim_size) + dim_size) % dim_size;
+
+    // Stack two copies of the dimension, then slice from the calculated
+    // offset.
+    xla::XlaOp concat =
+        xla::ConcatInDim(loctx->builder(), {input, input}, cur_dim);
+    std::vector<int64_t> start_indices(
+        need_flatten ? 1 : input_shape.dimensions_size(), 0);
+    start_indices[cur_dim] = dim_size - offset;
+    input = BuildSlice(concat, start_indices,
```
cicirori, line 62: The tf2xla version uses dynamic slice here, and I'm not quite sure what the difference is between it and naive slice. 

Since the arguments to the roll are passed as python constants, I understand that naive slice will work just fine here. 

But I'm not sure about that.

cicirori:
@JackCaoG 
Since I implemented it on the old code, porting to the community version would have some compilation errors, they have been fixed for now.

torch_xla/csrc/ops/roll.cpp
codego7250: Dynamic version has more flexibility with out any performance penalty. And Indices may be dynamic.  So keep it the same as tf's implementation.

cicirori: Thanks for your comment!  Already changed to dynamic slice.

test/test_operations.py
```diff
@@ -1401,6 +1401,106 @@ def test_flip_empty_tensor(self):
     data = torch.tensor([])
     self.assertEqual(data, data.flip(0))
 
+  def test_roll(self):
+    device = xm.xla_device()
+    numbers = torch.arange(1, 9, device=device)
+
+    single_roll = numbers.roll(1, 0)
+    expected = torch.tensor([8, 1, 2, 3, 4, 5, 6, 7], device=device)
+    self.assertEqual(
```
JackCaoG, line 1410: can we write test using `runAtenTest` util which will call the op on cpu and xla tensor then compare the output.

torch_xla/csrc/ops/roll.cpp
```diff
@@ -0,0 +1,83 @@
...
+    : Node(torch::lazy::OpKind(at::aten::roll), {input}, input.xla_shape(),
+           /*num_outputs=*/1, torch::lazy::MHash(shifts, dims)),
+      shifts_(std::move(shifts)),
+      dims_(std::move(dims)) {}
+
+torch::lazy::NodePtr Roll::Clone(OpList operands) const {
+  return ir::MakeNode<Roll>(operands.at(0), shifts_, dims_);
+}
+
+XlaOpVector Roll::Lower(LoweringContext* loctx) const {
```
JackCaoG, line 27: let's create a helper function called `BuildLower` and add it to the https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_lower_util.h. `Roll::Lower` can just call the helper function.

test/test_operations.py
JackCaoG: actually I think we need a C++ test instead of the python test. Most likely pytorch already have a pytorch test for roll and we currently fall back to CPU.

Can you write a C++ test similar to https://github.com/pytorch/xla/blob/master/test/cpp/test_aten_xla_tensor.cpp#L327 ?

JackCaoG (changes requested):
Mostly LGTM, thanks for the contribution!

test/test_operations.py
cicirori: done

torch_xla/csrc/ops/roll.cpp
cicirori: done

cicirori:
@JackCaoG It looks like there are some problems with CI build?

JackCaoG:
@cicirori Can you try to rebase, we recently made some changes regarding IR Node class.

cicirori:
> @cicirori Can you try to rebase, we recently made some changes regarding IR Node class.

thanks! it passed after rebasing.

JackCaoG (approved):
Thanks @cicirori !

cicirori:
> Thanks @cicirori !

@JackCaoG Thank you all the same. This is my first Github PR haha.

Issue #3506, opened 2022-04-15
All-reduce together w/ reduce-scatter causes crash on nightly 20220413 build


ronghanghu:
## 🐛 Bug

The nightly PyTorch XLA build (20220413 for torch, torchvision, torch_xla) gives an unexpected error when `all_reduce` is used together with `reduced_scatter` as follows
```
2022-04-15 06:35:51.117279: E tensorflow/core/tpu/kernels/tpu_compilation_cache_external.cc:113] during context [post-optimization]: HloModule has a mix of layout constrained and unconstrained AllReduce instructions.                           
2022-04-15 06:35:51.117346: F tensorflow/core/tpu/kernels/tpu_program_group.cc:86] Check failed: xla_tpu_programs.size() > 0 (0 vs. 0)
```

This error doesn't happen on the 20220408 build. This is likely a side effect of https://github.com/pytorch/xla/pull/3484 that removes `reduce_scatter`'s layout pining.

## To Reproduce

1. Allocate a v3-8 TPU VM from `tpu-vm-pt-1.10` runtime and install `20220413` version of `torch`, `torchvision`, and `torch_xla`, while keeping `20220408` version of libtpu (since the newer `20220413` version was reported bad in https://github.com/pytorch/xla/issues/3502#issuecomment-1099777942).
```
# torch, torchvision and torch_xla 20220413
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch-nightly+20220413-cp38-cp38-linux_x86_64.whl
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torchvision-nightly+20220413-cp38-cp38-linux_x86_64.whl
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch_xla-nightly+20220413-cp38-cp38-linux_x86_64.whl

# libtpu 20220408
sudo pip3 install https://storage.googleapis.com/cloud-tpu-tpuvm-artifacts/wheels/libtpu-nightly/libtpu_nightly-0.1.dev20220408-py3-none-any.whl
```

2. Save the following content to a python file (e.g. `/home/ronghanghu/test_all_gather_all_reduce_reduce_scatter.py` below).
```
import torch
import torch_xla.core.xla_model as xm
import torch_xla.distributed.xla_multiprocessing as xmp

def _mp_fn(index):
    world_size = xm.xrt_world_size()
[...]

PR #3511, opened 2022-04-19
Make layout pining optional for cross core communication

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/3506. I verified that with all `pin_layout=True`(default) or all `pin_layout=False`, test will passed.

PyTorch/xla currently compile graph separately for each TPU core(or GPU core). It is possible that graph being generated in each core is slightly different due to 
1. slightly different input shape
2. different embedding table size
3. ...

XLA compiler generate can tolerate small different among different cores but this will be a problem for communication ops. If the input shape difference ended up resulting in a layout difference among tensors that user want to call communication op, there can be a data corruption.

To overcome this problem we introduce the layout pining, which guarantee that all cores that participate in communication has the same layout for input tensor. However in some corner cases, pinging all layout will not work. For example `all_gather(pin)` + `reduce_scatter(pin)` might fail in some case. 

This pr aim to provide a workaround when such failure happened. PyTorch/XLA will pin all communcation op layout by defualt, but if there is a compilation error with message `HloModule has a mix of layout constrained` user can choose to unpin all layout.

FYI @ronghanghu @hjm-aws

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -526,7 +526,12 @@ def _host_all_reduce(reduce_type, inputs, cctx, scale=None):
         REDUCE_SUM, inputs, token, 1.0, [])
 
 
-def all_reduce(reduce_type, inputs, scale=1.0, groups=None, cctx=None):
+def all_reduce(reduce_type,
+               inputs,
+               scale=1.0,
+               groups=None,
+               cctx=None,
+               pin_layout=True):
   """Performs an inplace reduce operation on the input tensor(s).
 
   Args:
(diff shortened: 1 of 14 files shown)


JackCaoG:
I will merge this pr when all test passed to unblock the user, but I want someone to review it too. I can have a follow up pr to fix the review comment. @yeounoh

ronghanghu:
Thanks! Just to double-check, before this PR, we currently have the following behavior:
* all_reduce: pinned
* all_to_all: pined
* all_gather: unpinned
* reduce_scatter: unpinned

Is this right?

JackCaoG:
@ronghanghu yea, your statement is correct.

JackCaoG:
I modify the default parameter and only pin the `all_reduce` layout by default to make `test_mp_distributed_mm.py` work. This is also what Blake suggested.

PR #3514, opened 2022-04-20
Node hash new

JackCaoG:
This is to fix the build error in https://github.com/pytorch/pytorch/pull/75324. Pytorch/xla needs to implement two hash function and maintain the hash object.

From PyTorch/XLA's perspective, `hash` and `shape_hash` is the same thing.


JackCaoG:
FYI @wconstab @antoniojkim

torch_xla/csrc/ir.cpp
```diff
@@ -114,15 +114,11 @@ const xla::Shape& Value::xla_node_shape() const {
...
-          [&](bool /*bakeInSizes*/) -> torch::lazy::hash_t {
-            return GetOperandHashes(
-                operands, torch::lazy::HashCombine(op.hash(), hash_seed));
-          }),
-      xla_shape_(std::move(shape)) {
+    : torch::lazy::Node(op, num_outputs),
+      xla_shape_(std::move(shape)),
+      node_hash_(torch::lazy::HashCombine(op.hash(), hash_seed)),
+      dag_hash_(GetOperandHashes(
+          operands, torch::lazy::HashCombine(op.hash(), hash_seed))) {
```
antoniojkim, line 121: If we want to minimize computation, I don't think these _need_ to be in the MIL. Doing
```suggestion
      xla_shape_(std::move(shape)) {
      node_hash_ = torch::lazy::HashCombine(op.hash(), hash_seed);
      dag_hash_ = GetOperandHashes(operands, node_hash_);
    }
```
will save one hash calculation per node construction which can add up.

wconstab (approved):
this lgtm. 

i see you can torch-pin from this PR to #75324, but does that run enough CI steps to be sure?

Let me know what the best sequence for landing is.  Or if we should rebase/pin #75324 somehow.

torch_xla/csrc/ir.cpp
JackCaoG: oh, nice catch

antoniojkim (approved):

JackCaoG:
@wconstab pin to #75324 will checkout that pytorch branch when building in pt/xla CI. We will run both CPU and GPU test so it should be enough to cover :D.

wonjoolee95 (approved):
Thanks, @JackCaoG!

PR #3537, opened 2022-04-28
Remove unpool backward

JackCaoG:
This is to fix https://github.com/pytorch/pytorch/pull/68625, we don't need to directly lower unpool backward anymore


wonjoolee95 (approved):
Thanks! LGTM, will let CI do the rest.

JackCaoG:
The actual failure is unrelated, it is about
```
Invalid device specification: Unknown
```

albanD:
@JackCaoG this should be good to go!

seemethere (approved):

PR #3550, opened 2022-05-03
Experimental support for single-host TPU in PjRt

will-cromar:
This PR adds experimental support for execution on a single TPU host. To use PjRt instead of XRT, set `PJRT_DEVICE=TPU`. This will not work with `xla_multiprocessing`, as only one process can open the TPU device at a time. I will look into workarounds and solutions in a future PR.

Also clean up some rough edges from #3463:

- Execute computation on specified device, rather than executing on all addressable devices and returning the first result
- Propagate `explode_tuple` execute option to PjRt
- Construct device string from platform name and device id instead of using debug string (e.g. "TPU:5", "CPU:0")
- Give incoming tensors the default layout before transferring them to device
- Implement `ComputationClient::SetReplicationDevices`

### DIFF
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -80,15 +80,13 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
   for (auto& tensor : tensors) {
     xla::Literal literal(tensor.shape);
     tensor.populate_fn(tensor, literal.untyped_data(), literal.size_bytes());
-    xla::Shape shape = LayoutUtil::GetWithDefaultLayout(tensor.shape);
-    literal = literal.Relayout(shape);
 
     PjRtDevice* pjrt_device = StringToPjRtDevice(tensor.device);
     std::shared_ptr<xla::PjRtBuffer> buffer =
         client_->BufferFromHostLiteral(literal, pjrt_device).ValueOrDie();
     buffer->GetReadyFuture().Await();
     ComputationClient::DataPtr data =
-        std::make_shared<PjRtData>(tensor.device, shape, buffer);
+        std::make_shared<PjRtData>(tensor.device, tensor.shape, buffer);
     datas.push_back(data);
   }
 
@@ -120,6 +118,7 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
     xla::ProgramShape program_shape =
         instance.computation.GetProgramShape().ValueOrDie();
     xla::CompileOptions compile_options;
+    // TODO(wcromar): set compile_options.argument_layouts, enable strict shapes
     compile_options.executable_build_options.set_num_replicas(
         client_->addressable_device_count());
     compile_options.executable_build_options.set_device_ordinal(
@@ -162,6 +161,7 @@ PjRtComputationClient::ExecuteComputation(
 
   xla::ExecuteOptions execute_options;
   execute_options.untuple_result = options.explode_tuple;
+  execute_options.strict_shape_checking = false;
   std::vector<std::unique_ptr<xla::PjRtBuffer>> results =
       pjrt_computation.executable
           ->ExecuteSharded(buffers, pjrt_device, execute_options)
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 
+#include "absl/strings/ascii.h"
 #include "absl/types/span.h"
 #include "tensorflow/compiler/xla/client/xla_computation.h"
 #include "tensorflow/compiler/xla/layout_util.h"
@@ -20,8 +21,8 @@ namespace xla {
 namespace {
 
 std::string PjRtDeviceToString(PjRtDevice* const device) {
-  std::string platform(device->client()->platform_name());
-  std::transform(platform.begin(), platform.end(), platform.begin(), ::toupper);
+  std::string platform =
+      absl::AsciiStrToUpper(device->client()->platform_name());
   std::string str = absl::StrFormat("%s:%d", platform, device->id());
   return str;
 }
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -55,7 +55,7 @@ PjRtComputationClient::PjRtComputationClient() {
 
   XLA_CHECK(client_.get() != nullptr);
 
-  for (auto* device : client->addressable_devices()) {
+  for (auto* device : client_->addressable_devices()) {
     std::string device_str = PjRtDeviceToString(device);
     string_to_device.emplace(device_str, device);
   }
@@ -120,8 +120,10 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
     xla::ProgramShape program_shape =
         instance.computation.GetProgramShape().ValueOrDie();
     xla::CompileOptions compile_options;
-    compile_options.executable_build_options.set_num_replicas(client->addressable_device_count());
-    compile_options.executable_build_options.set_device_ordinal(pjrt_device->id());
+    compile_options.executable_build_options.set_num_replicas(
+        client_->addressable_device_count());
+    compile_options.executable_build_options.set_device_ordinal(
+        pjrt_device->id());
     std::unique_ptr<xla::PjRtExecutable> executable =
         client_->Compile(instance.computation, compile_options).ValueOrDie();
     std::shared_ptr<PjRtComputation> pjrt_computation =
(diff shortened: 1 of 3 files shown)


JackCaoG:
@will-cromar Did you have chance to test the `test_operations.py` on TPU single core? Want to make sure that this pass our basic test on TPU.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -42,13 +44,21 @@ PjRtComputationClient::PjRtComputationClient() {
   std::string device_type = sys_util::GetEnvString(env::kEnvPjRtDevice, "");
   if (device_type == "CPU") {
     TF_VLOG(1) << "Initializing PjRt CPU client...";
-    client_ = xla::GetCpuClient(/*asynchronous=*/false).ValueOrDie();
+    client_ = std::move(xla::GetCpuClient(/*asynchronous=*/false).ValueOrDie());
+  } else if (device_type == "TPU") {
+    TF_VLOG(1) << "Initializing PjRt TPU client...";
+    client_ = xla::GetTpuClient(/*max_inflight_computations=*/1).ValueOrDie();
```
JackCaoG, line 50: Is `max_inflight_computations=1` a short term thing or long term thing? What's it for?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -68,17 +78,17 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
   std::vector<ComputationClient::DataPtr> datas;
   datas.reserve(tensors.size());
   for (auto& tensor : tensors) {
-    xla::Shape shape = tensor.shape;
-    xla::Literal literal(shape);
+    xla::Literal literal(tensor.shape);
     tensor.populate_fn(tensor, literal.untyped_data(), literal.size_bytes());
+    xla::Shape shape = LayoutUtil::GetWithDefaultLayout(tensor.shape);
```
JackCaoG, line 83: regarding layout I think we specified the layout if it is not optimal for TPU in 
https://github.com/pytorch/xla/blob/3c83269f8b163243263078c2f1e93d7617321712/torch_xla/csrc/layout_manager.cpp#L137-L152.

Always set the layout to default before sending to server seems incorrect to me.

JackCaoG (commented):
Thanks! Mostly LGTM.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: I agree. I looked at the `XrtComputationClient` to understand how the shape/layout was handled there, but it uses [`ShapeUtil::MakeShapeWithDescendingLayoutAndSamePhysicalLayout`](https://github.com/tensorflow/tensorflow/blob/cacc32bdef7f7f3d80e0780b63c76a9b75a3f3d4/tensorflow/compiler/xla/shape_util.h#L377), which actually changes the shape. I assume the shape is converted back somewhere else, but I couldn't find it.

https://github.com/pytorch/xla/blob/3c83269f8b163243263078c2f1e93d7617321712/third_party/xla_client/xrt_computation_client.cc#L1941

I assigned the default layout here because otherwise I would need to know the argument shapes at compile time (see [`CompileOptions.argument_layout`](https://github.com/tensorflow/tensorflow/blob/cacc32bdef7f7f3d80e0780b63c76a9b75a3f3d4/tensorflow/compiler/xla/pjrt/pjrt_client.h#L218)), and I don't think that information is passed to `ComputationClient::Compile` right now.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: Short term. I can tune a better value once I get some non-trivial models running across all 8 cores.

will-cromar:
Yes, `test_operations.py` passes on one TPU core. I can add an automated test after we merge this PR.

JackCaoG:
Rebase should fix the build issue, let me know when you want me to take another pass on this pr.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: I avoided the relayout here by disabling [`strict_shape_checking`](https://github.com/tensorflow/tensorflow/blob/b6725018e7ddc0258922998e9496d170b6485b68/tensorflow/compiler/xla/pjrt/pjrt_client.h#L855) in `Execute()` and added a TODO to set the argument layout explicitly.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: As I understand it, this is the maximum number of asynchronous operations allowed at a time. For JAX this defaults to 32: [JAX_TPU_MAX_INFLIGHT_COMPUTATIONS](https://github.com/tensorflow/tensorflow/blob/44e84cebef5a89a0840f8d401819c25af41ec3db/tensorflow/compiler/xla/python/xla_client.py#L101)

third_party/xla_client/pjrt_computation_client.cc
JackCaoG: sounds good, if the real layout is required during compilation we should pass it down to the compile..

JackCaoG (approved):
Thanks @will-cromar ! Feel free to merge after CI is green

third_party/xla_client/pjrt_computation_client.h
```diff
@@ -146,7 +145,11 @@ class PjRtComputationClient : public ComputationClient {
   };
 
  private:
-  std::unique_ptr<PjRtClient> client_;
+  std::shared_ptr<PjRtClient> client_;
+  std::unordered_map<std::string, xla::PjRtDevice* const> string_to_device_;
+  std::shared_ptr<std::vector<std::string>> replication_devices_;
+
```
yeounoh, line 151: nit. I think we could move `PjRtData` and `PjRtComputation` to private member section?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -106,9 +115,15 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
   std::vector<ComputationClient::ComputationPtr> computations;
 
   for (auto& instance : instances) {
+    PjRtDevice* pjrt_device = StringToPjRtDevice(instance.compilation_device);
     xla::ProgramShape program_shape =
         instance.computation.GetProgramShape().ValueOrDie();
     xla::CompileOptions compile_options;
+    // TODO(wcromar): set compile_options.argument_layouts, enable strict shapes
+    compile_options.executable_build_options.set_num_replicas(
```
yeounoh, line 123: This is not necessary as it default to 1 anyway, but might be useful to explicitly set `num_partitions` to be specific about the compile & execution mode -- I think these two always go hand-to-hand.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -106,9 +115,15 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
...
+        client_->addressable_device_count());
+    compile_options.executable_build_options.set_device_ordinal(
+        pjrt_device->id());
     std::unique_ptr<xla::PjRtExecutable> executable =
```
yeounoh, line 127: Would this be shared if we compile once and execute on multiple devices? I guess for the current `Execute` fn paradigm we don't have to, but would this assumption change later?

third_party/xla_client/pjrt_computation_client.h
will-cromar: Good catch. I was following ComputationClient, but users should not be using PjRtData/Computation directly, so these should be private instead. I'll move them in a follow-up PR.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -131,22 +146,31 @@ PjRtComputationClient::ExecuteComputation(
...
+        << pjrt_device->DebugString() << " vs "
+        << pjrt_data->buffer->device()->DebugString();
     buffers.push_back(pjrt_data->buffer.get());
   }
 
   xla::ExecuteOptions execute_options;
-  execute_options.untuple_result = true;
-  std::vector<std::vector<std::unique_ptr<xla::PjRtBuffer>>> results =
-      pjrt_computation.executable->Execute({buffers}, execute_options)
+  execute_options.untuple_result = options.explode_tuple;
```
yeounoh, line 164: What does this option do? 
 Maybe I am looking at a wrong struct: http://google3/third_party/tensorflow/compiler/xla/pjrt/pjrt_client.h;l=839;rcl=448855448

Could it be useful to leave a short comment, what & why we do this?

third_party/xla_client/pjrt_computation_client.cc
will-cromar: That's a good point. I'll hard-code one partition to be explicit about the assumption here.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -131,22 +146,31 @@ PjRtComputationClient::ExecuteComputation(
...
+  execute_options.strict_shape_checking = false;
+  std::vector<std::unique_ptr<xla::PjRtBuffer>> results =
+      pjrt_computation.executable
+          ->ExecuteSharded(buffers, pjrt_device, execute_options)
```
yeounoh, line 168: You've already explained to me, @will-cromar could you help me understand again why we are switching from Execute to ExecuteSharded here?

third_party/xla_client/pjrt_computation_client.cc
will-cromar: This option corresponds 1:1 with the execute option we use in XRT to return a vector of tensors rather than a vector containing a tuple of tensors. We should probably document it there, though.

https://github.com/pytorch/xla/blob/1ab841877eaaa7a0acd6774172bcc52315f04073/third_party/xla_client/computation_client.h#L120-L122

third_party/xla_client/pjrt_computation_client.cc
will-cromar: Execute = execute one computation with a list of inputs on a list of devices (corresponds to [ComputationClient::ExecuteReplicated](https://github.com/pytorch/xla/blob/1ab841877eaaa7a0acd6774172bcc52315f04073/third_party/xla_client/computation_client.h#L217)]). Note how down below I had to fetch the 0th output vector before iterating through the outputs. If I left Execute here, every thread would execute every argument on every device, rather than executing the _i_-th set of arguments on the _i_-th device (or just crash).

ExecuteSharded = execute one computation with one set of inputs on one device, which is the same API as [ExecuteComputation](https://github.com/pytorch/xla/blob/1ab841877eaaa7a0acd6774172bcc52315f04073/third_party/xla_client/computation_client.h#L202)

third_party/xla_client/pjrt_computation_client.cc
will-cromar: This option is explained here:

https://github.com/pytorch/xla/blob/1e7b2b25c5573e4e3dc06d54d02d9ae809ae79f2/third_party/xla_client/computation_client.h#L200-L201

PR #3556, opened 2022-05-08
allow disabling layout pinning in optimizer_step and reduce_gradients

ronghanghu:
Following https://github.com/pytorch/xla/pull/3511, we should also allow users to specify layout pinning in `optimizer_step` and `reduce_gradients`, e.g. if they want to use these two functions together with `xm.all_gather` and `xm.reduce_scatter`. This PR is a short-term solution to such use cases.

As a long-term solution, it would be great to allow pinning layout in all collective ops simultaneously.

cc: @JackCaoG


JackCaoG (approved):
Thanks!

PR #3568, opened 2022-05-14
Pin all layout by default and use all_reduce based all_gather when la…

JackCaoG:
…yout is pin

In PyTorch/XLA we need to pin all layouts for cross-core-communciation ops since program is built separately for every core. However there is not an easy way to pin the layout for `all-gather` on TPU. In this pr I changed the `all-gather` is to use `all-reduce` when user actually wants to pin the layout. I also changed the `layout-pinning` to `True` by defualt. It was set to False because I can't pin `all-gather`.

@hjm-aws I am guessing you guys still prefer to use `all_gather` directly so I set the `pin_layout` to False for all xla_process_group. Let me know if that works for you.

FYI @hjm-aws @ronghanghu

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -593,7 +593,49 @@ def all_reduce(reduce_type,
   return results[0] if isinstance(inputs, torch.Tensor) else results
 
 
-def all_gather(value, dim=0, groups=None, output=None, pin_layout=False):
+def _all_gather_using_all_reduce(value, dim=0, groups=None, pin_layout=True):
+  """Performs an all-gather operation using all-reduce along a given dimension.
+
+  Args:
+    value (torch.Tensor): The input tensor.
+    dim (int): The gather dimension.
+      Default: 0
+    groups (list, optional): A list of list, representing the replica groups for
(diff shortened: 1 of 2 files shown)


torch_xla/core/xla_model.py
```diff
@@ -616,6 +658,11 @@ def all_gather(value, dim=0, groups=None, output=None, pin_layout=False):
     A tensor which has, in the ``dim`` dimension, all the values from the
     participating replicas.
   """
+  if pin_layout == True and output == None:
```
hjm-aws, line 661: Since this difficulty is specific to TPU and GPU, should we spell out the condition here in the `if` test?

hjm-aws, line 661: Also maybe you want to use `pin_layout` instead of `pin_layout == True`?

JackCaoG: That's fair, let me see if there is an easy way to determine the current device type.

JackCaoG:
@hjm-aws If this looks OK to you I will merge it today.

hjm-aws (approved):

PR #3571, opened 2022-05-17
Partially implement LTC Backend interface for PyTorch/XLA.

JackCaoG:
Partially implement LTC Backend interface for PyTorch/XLA.

### DIFF
--- a/torch_xla/csrc/xla_backend_impl.cpp
+++ b/torch_xla/csrc/xla_backend_impl.cpp
@@ -1,0 +1,171 @@
+#include "torch_xla/csrc/xla_backend_impl.h"
+
+#include <ATen/ScalarOps.h>
+
+#include "third_party/xla_client/debug_macros.h"
+#include "torch_xla/csrc/computation.h"
+#include "torch_xla/csrc/device.h"
+#include "torch_xla/csrc/lowering_context.h"
+#include "torch_xla/csrc/ops/device_data.h"
+#include "torch_xla/csrc/tensor_util.h"
+
+namespace torch_xla {
+class XlaBackendImpl : public torch::lazy::BackendImplInterface {
+ public:
+  XlaBackendImpl() {}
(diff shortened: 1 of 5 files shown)


JackCaoG:
@wonjoolee95 I think this pr is ready to be review. This pr only partially implement LTC Backend interface. I will have follow up pr(s) for the remaining prs.

torch_xla/csrc/xla_backend_impl.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  // GetCompilationDevices always returns devices today because
+  // 1. devices passed to GetCompilationDevices is always empty
+  // 2. ComputationClient's replication device is never set since we always have
+  // one replica per process
+  // We should revisit this API for SPMD
+  std::vector<std::string> GetCompilationDevices(
+      const std::string& device,
+      c10::ArrayRef<std::string> devices) const override {
+    return xla::ComputationClient::Get()->GetCompilationDevices(device,
```
JackCaoG, line 89: fyi @yeounoh, we currently call it in https://github.com/pytorch/xla/blob/185dd3cd45e4a1b9dcbbf88a3c604ce8c5e752f3/torch_xla/csrc/tensor.cpp#L1666.

wonjoolee95 (approved):
Thanks! So this `XlaBackendImpl` won't be used yet in XLA until we implement the `InitXlaBackend` function, correct?

JackCaoG:
@wonjoolee95 It won't be used even if we implement `InitXlaBackend` for the most part. Some functions needs to call `getBackend()->xxxx` to execute the function we defined here. Currently most of the `Backend` function is used the the ltc's LazyGraphExecutor.cpp. Until we fully migrated to that this file won't be very relevant.

One thing we can do incrementally is to start using the function in `XlaBackend` in our `torch_xla/csrc/tensor.cpp`. That way we can get some early signal of the correctness of these methods.

PR #3588, opened 2022-05-19
Allowing `float16` and `bfloat16` computation type and sharding groups in XLA FSDP

ronghanghu:
Following an offline discussion with @hjm-aws, this PR provides an enhancement to the XLA FSDP implementation with the following new options
* Allow specifying the `compute_dtype` argument to be `torch.float16` or `torch.bfloat16` instead of always `torch.float32` to support gathering full parameters and performing forward and backward in half-precision data types.
  * Also, the `fp32_reduce_scatter` argument is provided to allow always performing reduce-scatter in full-precision, and `buffer_dtype` is provided to allow casting module buffers to other data types than `compute_dtype`.
* Allow specifying the `sharding_groups` argument (along with `sharding_rank` and `sharding_world_size`) in FSDP's all_gather and reduce_scatter collective ops to support potential mixed parallelism, such as FSDP + Megatron. Also, allow groups in full gradient clipping.

---

The MNIST and ImageNet examples are updated to allow `--compute_dtype` (default to `"float32"`) and `--fp32_reduce_scatter` (default to `False`) arguments, which were tested on v3-8 TPU VM and working well.

Specifically, running ImageNet ResNet-50 with `bfloat16` compute_dtype as follows
```
python3 -u ~/workspace/xla_fsdp_clean_up/test/test_train_mp_imagenet_fsdp.py \
  --datadir /datasets/imagenet-1k --drop_last \
  --model resnet50 --test_set_batch_size 64 --eval_interval 10 \
  --lr 0.4 --batch_size 128 --num_warmup_epochs 5 --lr_scheduler_divide_every_n_epochs 30 --lr_scheduler_divisor 10 --num_epochs 100 \
  --use_nested_fsdp --compute_dtype bfloat16 \
  2>&1 | tee ~/workspace/r50_bfloat16/bfloat16_stdout_stderr_$(date +%Y-%m-%d_%H-%M-%S).log
```
gives
```
Max Accuracy: 75.69%
```
which is quite close to the 75.99% max accuracy using `float32` compute_dtype as in our [previous tests](https://github.com/pytorch/xla/pull/3431#issuecomment-1119737644).

cc: @hjm-aws

### DIFF
--- a/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
+++ b/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
@@ -433,7 +433,7 @@ class XlaFullyShardedDataParallel(nn.Module):
         norm_type (float or int): type of the used p-norm. Can be ``'inf'``
             for infinity norm.
         groups (list, optional): A list of list, representing the replica
-            groups for the `all_reduce()` operation to compute global norms.
+            groups for the all-reduce operation to compute global norms.
             See `xm.all_reduce` for details.
 
     Returns:
@@ -611,8 +611,9 @@ class XlaFullyShardedDataParallel(nn.Module):
             continue
           if torch.is_floating_point(buf):
             orig_dtype = buf.dtype
-            if orig_dtype != self.buffer_dtype:
-              buf = buf.to(dtype=dtype or self.buffer_dtype)
+            cast_dtype = dtype or self.buffer_dtype
+            if orig_dtype != cast_dtype:
+              buf = buf.to(cast_dtype)
               buf._orig_dtype = orig_dtype
           setattr(module, name, buf)
 
@@ -1495,8 +1496,8 @@ def _calc_grad_norm(parameters: List[torch.nn.Parameter],
 def _cast_floats_tensors(dtype: torch.dtype, *args: Any,
                          **kwargs: Any) -> Tuple[Any, Any]:
   """
-    Cast floating point Tensors in *args or **kwargs to dtype if they are not.
-    """
+  Cast floating point Tensors in *args or **kwargs to dtype if they are not.
+  """
 
   def fn(t):
     if t.dtype != dtype and torch.is_floating_point(t):
--- a/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
+++ b/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
@@ -623,8 +623,8 @@ class XlaFullyShardedDataParallel(nn.Module):
             f"compute_dtype={self.compute_dtype}, "
             f"buffer_dtype={self.buffer_dtype}, "
             f"flatten_parameters={self.flatten_parameters}, "
-            f"reshard_after_forward={self.reshard_after_forward}",
-            f"sharding_groups={self.sharding_groups}, ")
+            f"reshard_after_forward={self.reshard_after_forward}, "
+            f"sharding_groups={self.sharding_groups}")
     return repr
 
   def __getattr__(self, name: str) -> Any:
--- a/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
+++ b/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
@@ -36,6 +36,8 @@ import torch_xla.core.xla_model as xm
 from .xla_flatten_params_wrapper import XlaFlattenParamsWrapper
 from .utils import dummy_all_gather, dummy_all_reduce, dummy_reduce_scatter
 
+FLOAT_DTYPES = [torch.float32, torch.float16, torch.bfloat16]
+
 
 class TrainingState(Enum):
   """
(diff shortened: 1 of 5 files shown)


JackCaoG:
Do we need this change for the 1.12 release(it will most likely be cut today or tmr)?

ronghanghu:
@JackCaoG Thanks, I'm thinking this change can go into nightly releases in the future. The existing version was already working well under the test cases in https://github.com/pytorch/xla/pull/3431#issuecomment-1119737644, so I think it was already good enough for 1.12 and we don't need to merge this one immediately.

We probably also want to ship other enhancements to FSDP based on our profiling results (such as overlaying compute ops with collective ops to trade-off memory for speed), which we are developing and could provide in subsequent PRs.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -938,12 +1055,17 @@ def _post_backward_hook(self, param: Parameter, *unused: Any) -> None:
...
     reduced_grad = self.reduce_scatter_op(
         xm.REDUCE_SUM,
         grad_flat.detach(),
         scale=1.0,
         scatter_dim=0,
-        shard_count=self.world_size)
+        shard_count=self.world_size,
+        groups=self.sharding_groups)
+    if reduced_grad.dtype != torch.float32:
+      reduced_grad = reduced_grad.to(torch.float32)
```
hjm-aws, line 1068: Are these two lines necessary?

If `fp32_reduce_scatter` is false, should we just keep a 16-bit data type and let the optimizer decide what to do? Or do existing prevailing optimizers don't like 16-bit grads?

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -351,6 +421,7 @@ def clip_grad_norm_(
       self,
```
hjm-aws, line 421: Sorry I didn't notice `clip_grad_norm_` in the last PR. When is this method called?

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
ronghanghu: I think it would be better to cast the gradient back to fp32 here, because:
1) The model parameters are fp32, so one would naturally expect that the backward pass of a module would yield the same gradient dtype as the parameters. This is also how the MNIST and ImageNet examples can run with a simple change of `compute_dtype` flag -- the optimizer still sees fp32 gradients and can do the same optimization as before.
2) bfloat16 is very bad for gradient accumulation, e.g. 4096 + 1 = 4096 in bfloat16, and float16 also has less precision, so one would like to cast the gradients back to fp32 for accumulation (which is addressed in this cast `reduced_grad.to(torch.float32)`).
3) most existing optimizers like Adam needs to keep their 1st and 2nd momentum in fp32 (or at least the 2nd momentum in Adam) to work well and expect full precision gradients to accumulate the momentum.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
ronghanghu: This method is called to apply gradient clipping similar to [torch.nn.utils.clip_grad_norm_](https://pytorch.org/docs/stable/generated/torch.nn.utils.clip_grad_norm_.html), but based on the full gradient norm (rather than the norm of the sharded gradients in each worker as in FSDP)

hjm-aws: Sorry, I meant I didn't see it been called within FSDP. Do you mean it's supposed to be called outside of FSDP?

hjm-aws:
Thanks Ronghang for the PR!

Wanted to confirm with you whether the accuracy number `75.69%` was obtained with `fp32_reduce_scatter` set to `True` or `False`?

hjm-aws (approved):

JackCaoG:
@miladm Can you take a look at this one? If this is low risk maybe we can make it to 1.12.

ronghanghu:
Thanks for the review! @hjm-aws 

> Wanted to confirm with you whether the accuracy number 75.69% was obtained with fp32_reduce_scatter set to True or False?

It's using `fp32_reduce_scatter=False` on v3-8 TPU VM with bfloat16 compute_dtype. (`fp32_reduce_scatter` is default to `False` in the MNIST and ImageNet examples when the `--fp32_reduce_scatter` flag is not set.)

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
ronghanghu: Yes, it should be called outside of FSDP (it's not part of the model, but rather a part of the optimization strategy) and before the gradients are consumed by the optimizer, similar to calling torch.nn.utils.clip_grad_norm_ outside the model. An example usage is in https://github.com/ronghanghu/vit_10b_fsdp_example/blob/master/run_vit_training.py#L264-L273 (changing `torch.nn.utils.clip_grad_norm_` to `model.clip_grad_norm_` in the training loop for the FSDP case)

ronghanghu:
> If this is low risk maybe we can make it to 1.12.

I also verified that the ViT FSDP example with 10 billion parameters in https://github.com/ronghanghu/vit_10b_fsdp_example works well with this PR. So I'm OK with merging it into the 1.12 release.

JackCaoG:
I will take another pass of this pr and try to merge it to 1.12 before we build our rc1.

JackCaoG (approved):
Thanks @ronghanghu !

JackCaoG:
@wonjoolee95 Can you backport this pr to the 1.12 branch?

ronghanghu:
Thanks, @JackCaoG and @wonjoolee95 !

PR #3590, opened 2022-05-19
Check if XLA_SANDBOX_BUIDL is unset before applying the bazel patch

yeounoh:
The original checks worked for CI workflows, but had a bug with a local build. `XLA_SANDBOX_BUILD` may not be set in case of local build -- and it should be treated as a non-sandboxed build, and thus we apply the patch.

### DIFF
--- a/build_torch_xla_libs.sh
+++ b/build_torch_xla_libs.sh
@@ -35,16 +35,15 @@ if [[ "$XLA_BAZEL_VERBOSE" == "1" ]]; then
 fi
 
 BUILD_STRATEGY="standalone"
-if [[ -z "$XLA_SANDBOX_BUILD" ]] || [[ "$XLA_SANDBOX_BUILD" == "0" ]]; then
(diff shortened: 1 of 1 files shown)


build_torch_xla_libs.sh
```diff
@@ -35,7 +35,7 @@ if [[ "$XLA_BAZEL_VERBOSE" == "1" ]]; then
 fi
 
 BUILD_STRATEGY="standalone"
-if [[ "$XLA_SANDBOX_BUILD" == "0" ]]; then
```
JackCaoG, line 38: Maybe we can just do `if [[ "$XLA_SANDBOX_BUILD" == "1" ]]` and then a `else`？

yeounoh: Lol sounds good, done.

JackCaoG (approved):
Feel free to merge after all test is green

Issue #3589, opened 2022-05-19
PyTorch/XLA `gtest` needs to be updated


miladm:
I am running into the following error when building PyTorch/XLA CPP tests using `python setup.py install`.

**Error report from: `xla/test/cpp/build/gtest/src/googletest-stamp/googletest-build-err.log`**
```
In file included from xla/test/cpp/build/gtest/src/googletest-src/googletest/src/gtest-all.cc:42:
xla/test/cpp/build/gtest/src/googletest-src/googletest/src/gtest-death-test.cc: In function 'bool testing::internal::StackGrowsDown()':
xla/test/cpp/build/gtest/src/googletest-src/googletest/src/gtest-death-test.cc:1301:24: error: 'dummy' may be used uninitialized [-Werror=maybe-uninitialized]
 1301 |   StackLowerThanAddress(&dummy, &result);
      |   ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
xla/test/cpp/build/gtest/src/googletest-src/googletest/src/gtest-death-test.cc:1290:13: note: by argument 1 of type 'const void*' to 'void testing::internal::StackLowerThanAddress(const void*, bool*)' declared here
 1290 | static void StackLowerThanAddress(const void* ptr, bool* result) {
      |             ^~~~~~~~~~~~~~~~~~~~~
xla/test/cpp/build/gtest/src/googletest-src/googletest/src/gtest-death-test.cc:1299:7: note: 'dummy' declared here
 1299 |   int dummy;
      |       ^~~~~
cc1plus: all warnings being treated as errors
make[5]: *** [googletest/CMakeFiles/gtest.dir/build.make:76: googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o] Error 1
make[4]: *** [CMakeFiles/Makefile2:172: googletest/CMakeFiles/gtest.dir/all] Error 2
make[3]: *** [Makefile:146: all] Error 2
```

Please investigate.

---

**Blocker:** 
- [ ] https://github.com/tensorflow/tensorflow/issues/56021

miladm:
Turns out this problem has existed with the current pin (i.e. commit `6f5fd0d7199b9a19faa` in googletest version `1.10`) for quite some time ([ref](https://stackoverflow.com/questions/69935158/dummy-may-be-used-uninitialized)). The solutions is to upgrade to version `1.11.0`. Doing it now.

miladm:
[...]

PR #3591, opened 2022-05-19
`gtest` pin update

miladm:
Issue: https://github.com/pytorch/xla/issues/3589


setup.py
```diff
@@ -53,7 +53,7 @@
 base_dir = os.path.dirname(os.path.abspath(__file__))
 third_party_path = os.path.join(base_dir, 'third_party')
 
-_libtpu_version = '0.1.dev20220413'
+_libtpu_version = '0.1.dev20220518'
```
JackCaoG, line 56: this seems to be a error with commit history. It is already `dev20220518` on master

JackCaoG:
commit history seems to be messed up

setup.py
miladm: lol... logs happened at the same time.... I started a fresh pr to avoid headaches. closing this one.

PR #3595, opened 2022-05-21
Hotfix cloudbuild bazel update

yeounoh:
This is to replace the bazel download link with npm install. In particular, bazel `6e54699884cfad49d4e8f6dd59a4050bc95c4edf` (download link is broken) url returns 404, which is causing cloudbuild triggers to fail.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,14 +4,14 @@
 ARG base_image="debian:buster"
 FROM "${base_image}"
 
-ARG python_version="3.6"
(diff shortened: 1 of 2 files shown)


yeounoh:
Tested locally to confirm that this works. This is not circleci related, we can merge and continue test (or wait for the next nightly triggers) in cloudbuild. cc @JackCaoG

JackCaoG (approved):
Thanks @yeounoh !

PR #3605, opened 2022-05-24
Make MP_OP_TEST optional

JackCaoG:
This is the xla prt of pr to fix https://github.com/pytorch/xla/issues/3513

we have ~8 cc op tests, on CPU CI each of them will start 4 grpc server(since we config test to use 4 cpu) with random port. Every time we have a port conflict test will fail. This pr is to provide an option to limit these cc ops tests on cpu to only be run on pytorch/xla CI. It is very unlikely upstream will break pt/xla cc op. Most of the cc op test is not supported on CPU so it will bring up the xrt server and then directly quit which only creates churn.

I will set the env var on upstream to disable these cc op tests on CPU CI.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -85,7 +85,7 @@ function run_pjrt {
   PJRT_DEVICE=CPU run_test "$@"
 }
 
-function run_all_tests {
+function run_op_tests {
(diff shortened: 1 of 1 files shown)


JackCaoG:
@yeounoh gentle ping

yeounoh (approved):
LGTM, feel free to tag me along with the upstream CI env var change PR later.

PR #3607, opened 2022-05-24
Add PjRt tests for MNIST and ImageNet.

will-cromar:
Add temporary `run_all_chips.sh` to split processes across multiple chips. Because libtpu is loaded when `torch_xla` is imported, we have to set environment variables before running the script. When libtpu is loaded only when the ComputationClient is initialized (bug filed internally), we can set the environment variables and run the 4 processes from the same python script.

The main difference with the original tests is that we spin up 4 processes (1 process per chip) with 2 threads each (1 thread per core). The diffs with the original tests are below to make reviewing this PR easier.

<details>
  <summary>MNIST</summary>

```
< def train_mnist(flags, **kwargs):
<   torch.manual_seed(1)
---
> def train_mnist(flags, index, state_dict, **kwargs):
>   rank = int(os.getenv('CLOUD_TPU_TASK_ID', 0)) + index
77c77
<         os.path.join(flags.datadir, str(xm.get_ordinal())),
---
>         os.path.join(flags.datadir, str(rank)),
84c84
<         os.path.join(flags.datadir, str(xm.get_ordinal())),
---
>         os.path.join(flags.datadir, str(rank)),
95c95
<           rank=xm.get_ordinal(),
---
>           rank=rank,
114,115c114,117
<   device = xm.xla_device()
<   model = MNIST().to(device)
---
>   device = xm.xla_device(index)
>   model = MNIST()
>   model.load_state_dict(state_dict)
>   model = model.to(device)
149c151
<     accuracy = xm.mesh_reduce('test_accuracy', accuracy, np.mean)
---
>     # accuracy = xm.mesh_reduce('test_accuracy', accuracy, np.mean)
177c179
< def _mp_fn(index, flags):
---
> def _mt_fn(index, flags, model):
179c181,182
<   accuracy = train_mnist(flags)
---
>   torch.manual_seed(1)
>   accuracy = train_mnist(flags, index, model.state_dict())
185d187
<     sys.exit(21)
188,189c190,202
< if __name__ == '__main__':
<   xmp.spawn(_mp_fn, args=(FLAGS,), nprocs=FLAGS.num_cores)
---
> import concurrent.futures
> 
> xm.set_replication(xm.xla_device(), xm.get_xla_supported_devices())
> 
> with concurrent.futures.ThreadPoolExecutor(
>     max_workers=len(xm.get_xla_supported_devices())) as executor:
>   model = MNIST()
>   futures = [
>       executor.submit(_mt_fn, i, FLAGS, model)
>       for i in range(len(xm.get_xla_supported_devices()))
>   ]
>   for f in concurrent.futures.as_completed(futures):
>     f.result()
```
</details>

<details>

<summary>ImageNet</summary>

```
< def train_imagenet():
---
> def train_imagenet(index, state_dict):
116a117
>   rank = int(os.getenv('CLOUD_TPU_TASK_ID', 0)) + index
158c159
<           rank=xm.get_ordinal(),
---
>           rank=rank,
163c164
<           rank=xm.get_ordinal(),
---
>           rank=rank,
180,183c181,184
<   torch.manual_seed(42)
< 
<   device = xm.xla_device()
<   model = get_model_property('model_fn')().to(device)
---
>   device = xm.xla_device(index)
>   model = get_model_property('model_fn')()
>   model.load_state_dict(state_dict)
>   model = model.to(device)
232c233
<     accuracy = xm.mesh_reduce('test_accuracy', accuracy, np.mean)
---
>     # accuracy = xm.mesh_reduce('test_accuracy', accuracy, np.mean)
260c261
< def _mp_fn(index, flags):
---
> def _mt_fn(index, flags, model):
264c265
<   accuracy = train_imagenet()
---
>   accuracy = train_imagenet(index, model.state_dict())
271,272c272,285
< if __name__ == '__main__':
<   xmp.spawn(_mp_fn, args=(FLAGS,), nprocs=FLAGS.num_cores)
---
> import concurrent.futures
> 
> xm.set_replication(xm.xla_device(), xm.get_xla_supported_devices())
> 
> with concurrent.futures.ThreadPoolExecutor(
>     max_workers=len(xm.get_xla_supported_devices())) as executor:
>   torch.manual_seed(42)
>   model = get_model_property('model_fn')()
>   futures = [
>       executor.submit(_mt_fn, i, FLAGS, model)
>       for i in range(len(xm.get_xla_supported_devices()))
>   ]
>   for f in concurrent.futures.as_completed(futures):
>     f.result()
```
</details>

These tests should be temporary. Long-term, we want to implement tests that work for both XRT and PjRt and/or create new tests that use `torch.distributed` instead.

Tested manually on v2-8 TPU. Will add automated tests after this is merged.


JackCaoG:
why do we need to manually save and load `state_dict`? I felt like we have talked about that but I don't remember 😄 . I am guessing `xm.get_ordinal()` won't work for pjrt?

will-cromar:
`get_ordinal` doesn't work in a threaded context, because the ordinal comes from an environment variable, which can't be thread-local. 

https://github.com/pytorch/xla/blob/bb43e8a7048c008e5379f3d8ef7027496fbce0ef/torch_xla/core/xla_model.py#L175

When I was initializing the model separately in each thread, each copy ended up with different initial parameters. To be honest, I am not actually sure why each copy ends up with the same parameters in XRT. My guess is that each process can separately set the same seed, whereas in a multithreaded context, each thread in a process will share the same random generator.

will-cromar:
`torch.manual_seed` is global. When directly using random functions directly, you can [pass in a generator](https://discuss.pytorch.org/t/is-there-a-randomstate-equivalent-in-pytorch-for-local-random-generator-seeding/37131/3). Do you know how to do something similar for Module initialization?

I believe DDP will synchronize parameters, so this might not be an issue at all once we move to standard `torch.distributed` APIs.

JackCaoG:
I don't think generator works too well for us. PyTorch/XLA relies on RNG generator from the XLA. We will passed a base seed and pass that seed to the XLA compiler to generate random number (instead of generating that on CPU and transfer the tensor to the device). We will also increment the random seed before every use.

by `each thread in a process will share the same random generator.` what do you mean? Are two process share the same base seed? or when one thread incremented the RNG seed it will affect other thread as well?

will-cromar:
As I understand it, our examples are initializing the model on CPU with the `torch.manual_seed` (when `MNIST()` is called) and moving to the device after (when `.to()` is called). I don't think XLA's RNG is involved during initialization.

https://github.com/pytorch/xla/blob/e2357ba8d8d128ed9fbbc92bad62dd5df91d1240/test/test_train_mp_mnist.py#L115

Since the seed is global, when one thread in a process increments the RNG, it will affect the other thread. Since the model initialization is run in parallel and both threads are incrementing the RNG at the same time, the resulting initial parameters are not deterministic between threads or across processes.

JackCaoG:
I am a bit confusing, looking at the code

https://github.com/pytorch/xla/blob/e2357ba8d8d128ed9fbbc92bad62dd5df91d1240/torch_xla/csrc/tensor.cpp#L322-L328

our rng seed IR Value is per device, so different thread shouldn't matter here?

will-cromar:
Right, but both initializations are done on one device (CPU) and then the copies are moved to the XLA devices. I haven't had to implement SetRngSeed in ComputationClient to make these examples work, so I don't think the XLA devices' seeds are used at all.

https://github.com/pytorch/xla/blob/e2357ba8d8d128ed9fbbc92bad62dd5df91d1240/third_party/xla_client/pjrt_computation_client.h#L100-L102

JackCaoG:
Oh ok, we are talking about 2 different random seed.. I see. This one process 2 thread sounds really hacky now lol

JackCaoG:
XLA devices' seeds are used when you called `randn` or some random op on XLA tensor.  that should work as it is pre device.

JackCaoG (approved):
Thanks!

will-cromar:
For sure. The 4 process x 2 thread setup is an unfortunate hack to get around the constraint of one process per chip, and I want to try some ways of getting around it.

For what it's worth, the original examples also are initializing the model on CPU and moving it to each device. It's just in this case, there's no clean way I know of in PyTorch to actually copy a model onto two devices, so I have this clunky hack where I pass the state_dict from the CPU model into each XLA/TPU model.

JackCaoG:
I am ok with landing this as experimental, we can have more detail discussion regarding long term solution later.

test/pjrt/test_train_pjrt_mnist.py
```diff
@@ -0,0 +1,202 @@
...
+      device,
+      x,
+      loss.item(),
+      tracker.rate(),
+      tracker.global_rate(),
+      summary_writer=writer)
+
+
+def train_mnist(flags, index, state_dict, **kwargs):
+  rank = int(os.getenv('CLOUD_TPU_TASK_ID', 0)) + index
```
ronghanghu, line 62: Should this line be
```
rank = int(os.getenv('CLOUD_TPU_TASK_ID', 0)) * 2 + index
```
on TPU v3 instead (that is, adding `*2` to `CLOUD_TPU_TASK_ID`)? Otherwise, `rank` is one of 0, 1, 2, 3, 4 (**including 4**) on v3-8, which seems like a bug to me. I think it should be one of 0, 1, 2, 3, 4, 5, 6, 7 instead on v3-8.

Similarly for the ImageNet example

JackCaoG: I think you are right, down below we set 
```
          num_replicas=xm.xrt_world_size(),

```

which I think `xm.xrt_world_size()` should return 8, so our rank should be 0-7. This might explain why we see some accuracy gap on v3-8 using pjrt. @ronghanghu nice catch!

will-cromar: Thanks for flagging this line. It is wrong, but actually not for this reason. On TPU v2 and v3, we have 4 processes (`CLOUD_TPU_TASK_ID` 0 through 3) with two visible devices each (two threads with `index` 0 or 1). So on process 0, we have ranks `0 * 2 + 0 = 0` and `0 * 2 + 1 = 1`, and on process 3, we have `3 * 2 + 0 = 6` and `3 * 2 + 1 = 7`. This gives us the range 0 through 7 that we expect.

I didn't fix this line when I added the v4 tests, which have 4 processes with one visible device each (i.e. index is always 0). These will have ranks 0, 2, 4, and 6 but have world size 4. This is just a clunky expression anyway, so I think the right thing to do is update functions in `xla_model` like `xm.get_ordinal()` and `xm.get_local_ordinal()` for PjRt so we don't need to compute the rank like this at all. I see you've already filed a bug to do that (#3692)

ronghanghu: Thanks @will-cromar! Yeah, I also feel the right thing to do is update functions in xla_model like xm.get_ordinal() and xm.get_local_ordinal().

PR #3609, opened 2022-05-25
Adding support for torchrun in xla backend

amithrm:
This pull request enables the code needed to integrate torchrun launcher with xla backend.


JackCaoG:
@amithrm Can you provide a test for this new feature?

amithrm:
@JackCaoG sure..will add tests

JackCaoG:
I have to admit that I am not an export of torchrun, let me read up some documentations first lol. Looping in @will-cromar to make sure this does not conflict with our future pjrt runtime.

amithrm:
we did some internal testing. It appears that at scale, we see issues with the set up of GRPC channels. We should understand if you see similar issues at your end too.

amithrm:
@JackCaoG 
A simple test that you can run on GPU-XLA:

```
import sys
import torch
import torch_xla

import os


def _mp_fn(index):
  print('XRT_LOCAL_WORKER:{}'.format(os.environ['XRT_LOCAL_WORKER']))
  print('XRT_DEVICE_MAP:{}'.format(os.environ['XRT_DEVICE_MAP']))
  print('XRT_WORKERS:{}'.format(os.environ['XRT_WORKERS']))
  print('XRT_HOST_WORLD_SIZE:{}'.format(os.environ['XRT_HOST_WORLD_SIZE']))
  device = xm.xla_device()
  ordinal_tensor = torch.tensor([index], dtype=torch.float).to(device)
  print('rank:{}, value:{}'.format(index, ordinal_tensor))
  result = xm.all_reduce('sum', ordinal_tensor)

  cpu_result = result.cpu()
  print('rank:{}, value:{}'.format(index, cpu_result))


if __name__ == '__main__':
  xmp.spawn(_mp_fn, args=(), nprocs=2, join=True)
```

amithrm:
Run command:

GPU_NUM_DEVICES=2 python3 allreduce_xla.py

This will output:

XRT_LOCAL_WORKER:localservice:0
XRT_DEVICE_MAP:GPU:0;/job:localservice/replica:0/task:0/device:XLA_GPU:0|GPU:1;/job:localservice/replica:0/task:1/device:XLA_GPU:0
XRT_WORKERS:localservice:0;grpc://dfda805bbe4b:49887|localservice:1;grpc://dfda805bbe4b:33097
XRT_LOCAL_WORKER:localservice:1
XRT_DEVICE_MAP:GPU:0;/job:localservice/replica:0/task:0/device:XLA_GPU:0|GPU:1;/job:localservice/replica:0/task:1/device:XLA_GPU:0
XRT_WORKERS:localservice:0;grpc://dfda805bbe4b:49887|localservice:1;grpc://dfda805bbe4b:33097

If you look for XRT_WORKERS, this has the grpc string for each worker. This won't scale with number of workers.

JackCaoG:
We want to support torch run with new PJRT run time, in the mean time if this torch run utility can unblock aws folks we can also take it. 

I am a bit hesitant whether claim official support for XRT:TPU + torch run. @will-cromar Let's invesgate what's the gap here, if it is free we might as well just take it.

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
...
+        gpus_to_use_list = gpus_to_use.split(',')
+        os.environ['CUDA_VISIBLE_DEVICES'] = gpus_to_use_list[local_rank]
+    else:
+        # If no explicit visible devices are provided, local_rank is used to identify
+        # the gpu used by this process
+        os.environ['CUDA_VISIBLE_DEVICES'] = str(local_rank)
+
+
+def _init_xrt_context(master_addr=None, master_port=None, store=None):
+    '''
```
will-cromar, line 157: nit: Can you update this docstring to match the format used in other files? e.g. https://github.com/pytorch/xla/blob/590dee5d08ff4cf3ba72eab306b470e22bd08e16/torch_xla/core/xla_model.py#L187-L197

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
...
+        # comma separated string containing numbered gpu devies
+        gpus_to_use_list = gpus_to_use.split(',')
+        os.environ['CUDA_VISIBLE_DEVICES'] = gpus_to_use_list[local_rank]
+    else:
+        # If no explicit visible devices are provided, local_rank is used to identify
+        # the gpu used by this process
+        os.environ['CUDA_VISIBLE_DEVICES'] = str(local_rank)
+
+
+def _init_xrt_context(master_addr=None, master_port=None, store=None):
```
will-cromar, line 156: Remove the leading underscore, since this looks like the main public API of this module.

test/test_allreduce_torchrun.py
```diff
@@ -0,0 +1,29 @@
...
+
+    return_code = proc0.wait()
+    assert return_code == 0
+
+def test_local_launch_allreduce_cores():
+    # This test launches a allreduce using torchrun launcher
+    cmd = "torchrun --nproc_per_node=2  --master_addr=127.0.0.1 --master_port=2020 allreduce_torchrun.py "
+
+    new_env0 = os.environ.copy()
+    new_env0['NEURON_RT_VISIBLE_CORES'] = '0,1'
```
will-cromar, line 18: I'm not familiar with Neuron at all. Do we need to install anything extra beyond PyTorch and PyTorch/XLA to run this test? If not, can you add a GPU CI test like this? https://github.com/pytorch/xla/blob/590dee5d08ff4cf3ba72eab306b470e22bd08e16/test/run_tests.sh#L85-L90

torch_xla/distributed/xla_backend.py
```diff
@@ -41,6 +42,14 @@ def __init__(self, prefix_store, rank, size, timeout):
     self.prefix_store = prefix_store  # reserved for future use.
     self.timeout = timeout
     self._mesh = []
+    # Initialize xrt neuron environment
+    # Passes in the store created by torch.distributed to avoid
+    # creating two TCP stores. We only want to call this
+    # when the user is using torchrun and not xmp.spawn()
+    # or some other flow.
+    if os.getenv('TORCHELASTIC_RUN_ID') != None:
```
will-cromar, line 50: Is `TORCHELASTIC_RUN_ID` ever set by anything other than torchrun?

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
...
+            [str(i) for i in range(local_world_size)])
+
+
+def _setup_nccl_service(rank):
+    # Set up NCCL COMM ID required for NCCL communicator IDs
+    address = _get_address_from_store('nccl_info', rank)
+    os.environ['NEURON_RT_ROOT_COMM_ID'] = address
+
+
+def set_xrt_envs(dev_kind, world_size, rank, local_rank):
```
will-cromar, line 136: Should this name match the form of `_set_tpu_xrt_env`? i.e. `_set_gpu_xrt_env`

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
...
+        assert master_addr is not None
+        assert master_port is not None
+        _TCP_STORE = dist.TCPStore(master_addr, int(master_port), world_size,
+                                   is_server)
+    else:
+        _TCP_STORE = store
+
+    node_list = None
+
+    if dev_kind == 'TPU':
```
will-cromar, line 212: Why are the two `if dev_kind == 'TPU'` blocks separated here? Does part need to run before `_set_mesh_config` and half need to run after?

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
+# Copyright Amazon Web Services and its Affiliates. All Rights Reserved.
+# ==============================================================================
+
```
will-cromar, line 3: I don't think `yapf` is catching this, but can you also adjust the indent to two spaces to match the rest of the repository?

test/test_allreduce_torchrun.py
```diff
@@ -0,0 +1,29 @@
+import os
+import subprocess
+
+def test_local_launch_allreduce():
+    # This test launches a allreduce using torchrun launcher
+    cmd = "torchrun --nproc_per_node=2  --master_addr=127.0.0.1 --master_port=2020 allreduce_torchrun.py "
+
+    proc0 = subprocess.Popen(cmd, shell=True)
```
will-cromar, line 8: Here and in the test below, can this just be named `proc`?

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
...
+    _set_mesh_config(rank)
+
+    if dev_kind == 'TPU':
+        _setup_nccl_service(rank)
+        _set_neuron_envs(rank, world_size, local_world_size)
+
+        total_nodes = world_size // local_world_size
+        if local_rank == 0:
+            local_env = os.environ.copy()
+            subprocess.Popen([
```
will-cromar, line 229: Do you need to use subprocess here? Can you call [`xrt_run_server.run_service`](https://github.com/pytorch/xla/blob/590dee5d08ff4cf3ba72eab306b470e22bd08e16/torch_xla/core/xrt_run_server.py#L25) directly with [`multiprocessing`](https://docs.python.org/3/library/multiprocessing.html) instead?

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
...
+import torch.distributed as dist
+import torch_xla.core.xla_model as xm
+import torch_xla.core.xla_env_vars as xenv
+from torch_xla.utils.utils import get_free_tcp_ports
+from torch_xla.distributed.xla_multiprocessing import _get_devices_per_worker
+
+_NODE_LIST = None
+_TCP_STORE = None
+_INIT_XRT_ALREADY_CALLED = False
+XRT_SERVER_REGEX = 'torch_neuronx.distributed._xrt_run_server'
```
will-cromar, line 16: I assume most of our users don't have this installed. Should this be configurable so they can select the default server in PT/XLA instead?

https://github.com/pytorch/xla/blob/master/torch_xla/core/xrt_run_server.py

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
...
+import os
+import socket
+import subprocess
+import torch.distributed as dist
+import torch_xla.core.xla_model as xm
+import torch_xla.core.xla_env_vars as xenv
+from torch_xla.utils.utils import get_free_tcp_ports
+from torch_xla.distributed.xla_multiprocessing import _get_devices_per_worker
+
+_NODE_LIST = None
```
will-cromar, line 13: `_NODE_LIST` looks unused, and `_TCP_STORE` and `_INIT_XRT_ALREADY_CALLED` are already marked `global`, so they don't need to be declared here

will-cromar (changes requested):
This doesn't conflict with PJRT, but we will have to add a new code path for it. Runtime configuration is much simpler under PJRT, so I expect that we can simplify a lot of this.

@JackCaoG what do you think of moving this under `torch_xla.experimental` for now?

test/allreduce_torchrun.py
```diff
@@ -0,0 +1,29 @@
+import torch
```
JackCaoG, line 1: test won't get run in CI unless you added it to https://github.com/pytorch/xla/blob/master/test/run_tests.sh#L151

JackCaoG:
in terms of moving the code to experimental I guess it is related to how do we want to set the user expection. Is there any known cavet for this feature? If it works well we don't need to put it in experimental. However we should add a README similar to https://github.com/pytorch/xla/blob/master/docs/ddp.md

torch_xla/distributed/xrt_init.py
```diff
@@ -0,0 +1,241 @@
+# Copyright Amazon Web Services and its Affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
amithrm: No, not intentional. I will consult with internal folks and get their feedback!

torch_xla/distributed/xrt_init.py
amithrm: Sure, no problem!

torch_xla/distributed/xrt_init.py
amithrm: Yes

test/test_allreduce_torchrun.py
amithrm: Yes, these are neuron specific env variables. Will add the GPU specific ones. But, we have PJRT_DEVICE listed above, will this interfere with the torchrun?

torch_xla/distributed/xla_backend.py
amithrm: No, its only specific to torchrun.

torch_xla/distributed/xrt_init.py
amithrm: will do ...thanks!

test/allreduce_torchrun.py
amithrm: Sounds good! will add it here

test/test_allreduce_torchrun.py
amithrm: sure..

torch_xla/distributed/xrt_init.py
amithrm: Will, we basically follow the same template as done in xrt_run_server.py using subprocess.Popen. Essentially, its the same server invocation. However, what we do extra is spawn up an extra "watch" process that polls if the original process is around or got killed(This happens when Slurm terminates the workers and Slurm is not aware of this server, So, the watch process does this job).

torch_xla/distributed/xrt_init.py
amithrm: The same server described in this module is run by rank 0 worker on a node. This is spawned by SLURM. So, as described above, in addition to the subprocess.Popen call in this module, we also spawn another "Watch" /"Polling" process. That is the only difference. (Forgot to add the module here). But pasting the main contents below:
```
polling_process = multiprocessing.Process(target=_polling,
                                              args=(int(args.pid_to_track), ))
server_process = multiprocessing.Process(
        target=torch_xla._XLAC._run_xrt_local_service, args=(int(args.port), ))
```

torch_xla/distributed/xrt_init.py
amithrm: yes, worked on this and other modifications suggested by you. Resolving the final issue mentioned above related to starting of server processes.

JackCaoG:
@will-cromar can you take another pass of this pr when you have some time?

test/test_allreduce_torchrun.py
will-cromar: Nope, there shouldn't be an issue. PJRT won't work with this PR, but you can add a new function to `run_tests.sh` that sets up anything neuron related and skips setting `PJRT_DEVICE` (i.e. make a `run_neuron` function like `run_pjrt`)

will-cromar (commented):
Thanks! Overall LGTM. Please add a CI test (run_tests.sh) and then we can merge this

test/test_allreduce_torchrun.py
amithrm: Hi @will-cromar Looks like in the CI tests, the file allreduce_torchrun.py is not picked up. What is a good way to fix this? Is there a global prefix that I can add before the file name?

amithrm:
Looks like the file need to test (allreduce_torchrun.py) is not getting picked up. Checking with @will-cromar on how to fix this. And some yapf fixes are pending in one file.

amithrm:
@will-cromar I see build failure: NameError: name 'sympy' is not defined

JackCaoG:
weird, head is green right now https://github.com/pytorch/xla/commits/master

JackCaoG:
Ah ok https://github.com/pytorch/xla/pull/4313/files should fix it, can you rebase again?

amithrm:
@JackCaoG Alll the 4 pass @will-cromar is there anything else needed?

will-cromar (approved):
LGTM

JackCaoG (approved):
Thanks!

jeffhataws:
Thanks @JackCaoG and @amithrm !

Issue #3587, opened 2022-05-19
Update XLA torch.gather heuristics for multi-platforms


yeounoh:
## 🚀 Feature
There are two code paths for composing and executing `torch.gather` in PyTorch/XLA, that relies on a custom heuristic to decide between the two. The heuristic requires a major update to better serve various platforms and use cases. cc @ymwangg @ronghanghu 

## Motivation

Based on the recent [study](https://github.com/pytorch/xla/pull/3566#issuecomment-1126592037) for GPU/CPU, and also experiencing some related performance regression [issue](https://github.com/pytorch/xla/issues/3441) on TPU, we feel strongly that we would need a better heuristics to serve different platforms. 

## Pitch

There were a couple of different strategies employed to address the issue:
- forcing one type of `torch.gather` implementation always,
- making the decision based on the number of input tensor elements (w.r.t. the indices) and allow user to change/try a different threshold via setting an env var.

Both approaches had clear shortcomings, and had a diverging behaviors on different platforms. We propose that we improve our heuristics in the following ways:
- Pre-condition on the platform type
- Update the heuristics with a more conservative default threshold (advanced user can still change/try their own via setting an env var)
- Use the threshold-based heuristics only for TPU platform, for others

## Alternatives

One of the two approaches that we've tried already.

yeounoh:
Hi @ymwangg I will follow up with you once I create a PR to address this -- stay tuned!

JackCaoG:
@yeounoh Can this be closed?

yeounoh:
This was fixed with https://github.com/pytorch/xla/pull/3629, closing.

PR #3610, opened 2022-05-25
Update sparsity check heuristics used in TorchGather

yeounoh:
Addresses #3587

### DIFF
--- a/torch_xla/csrc/ops/gather.cpp
+++ b/torch_xla/csrc/ops/gather.cpp
@@ -25,10 +25,9 @@ xla::Shape NodeOutputShape(const torch::lazy::Value& input,
 
 Gather::Gather(const torch::lazy::Value& input, int64_t dim,
                const torch::lazy::Value& index)
-    : XlaNode(
-          torch::lazy::OpKind(at::aten::gather), {input, index},
-          [&]() { return NodeOutputShape(input, index, dim); },
-          /*num_outputs=*/1, torch::lazy::MHash(dim)),
+    : XlaNode(torch::lazy::OpKind(at::aten::gather), {input, index},
+              [&]() { return NodeOutputShape(input, index, dim); },
+              /*num_outputs=*/1, torch::lazy::MHash(dim)),
       dim_(dim) {}
 
 torch::lazy::NodePtr Gather::Clone(torch::lazy::OpList operands) const {
--- a/torch_xla/csrc/ops/gather.cpp
+++ b/torch_xla/csrc/ops/gather.cpp
@@ -7,31 +7,45 @@
 #include "torch_xla/csrc/ops/infer_output_shape.h"
 
 namespace torch_xla {
+namespace {
+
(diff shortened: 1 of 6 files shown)


JackCaoG:
@wonjoolee95 FYI, we will need this pr in the 1.12. You can still play with trigger for now but we will likely need to rebuild the rc(maybe rc1?) with this fix.

@yeounoh Can you make sure this change does not regress vit model?

yeounoh:
> @wonjoolee95 FYI, we will need this pr in the 1.12. You can still play with trigger for now but we will likely need to rebuild the rc(maybe rc1?) with this fix.
> 
> @yeounoh Can you make sure this change does not regress vit model?

Ok, just verified the new heuristics with the ViT model from @ronghanghu . 

@ymwangg , this would allow sparse gather on GPU and CPU -- thank you again for your detailed analysis.

torch_xla/csrc/ops/gather.cpp
```diff
@@ -7,31 +7,44 @@
 #include "torch_xla/csrc/ops/infer_output_shape.h"
 
 namespace torch_xla {
+namespace {
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input,
```
JackCaoG, line 12: hmm, will `IsSparseGather` change the output shape? Why do we need a `NodeOutputShape` here?

JackCaoG (commented):
Mostly lGTM, one minor question.

torch_xla/csrc/ops/gather.cpp
yeounoh: It does, since dense gather actually uses reduce instead of gather.

JackCaoG (approved):
Thanks!

torch_xla/csrc/ops/gather.cpp
ymwangg: FYI https://pytorch.org/docs/stable/generated/torch.gather.html. The output tensor of `torch.gather` has the same shape as `index`.

An issue I found with XLA dense gather is that it does not allow `index.size(d) < input.size(d)` for dimension `d != dim` so that the following code does not work with dense gather. Perhaps this type of input is not typical since nobody has complained.
```
a = torch.randint(0, 10, size=(8, 8), dtype=torch.int64, device=device)
b = torch.ones(10, 10).to(device)
c = torch.gather(b, dim=1, index=a)
```

yeounoh: I see, @ymwangg let me check with XLA folks about this before landing, and see if they could modify the implementation or the check(s).

yeounoh: @ymwangg I think it's a bug and will file a separate bug to track internally. Thanks!

yeounoh:
Merging cc @JackCaoG @ymwangg

JackCaoG:
@yeounoh Thanks, Can you backport this pr to r1.12?

PR #3611, opened 2022-05-25
Include missing dependencies in setup.py

will-cromar:
`absl-py` and `cloud-tpu-client` are both required for `xla_dist`.

Bonus: fix typo `litbpu` -> `libtpu`


JackCaoG (approved):

JackCaoG:
Do we need this to be backported to 1.12?

will-cromar:
This issue would have been going on through several releases, but it should be safe to cherry-pick this into 1.12

yeounoh (approved):
LGTM, thanks @will-cromar

PR #3613, opened 2022-05-26
Porting #3610 to r1.12

yeounoh:
Porting #3610 into `r1.12`.


wonjoolee95 (approved):
Thanks!

JackCaoG (approved):

JackCaoG:
oh, seems like every merge to r1.12 will trigger on demand rebuild of wheel.

wonjoolee95:
> oh, seems like every merge to r1.12 will trigger on demand rebuild of wheel.

Yeah, I've set the trigger to run anytime there is a push to the `r1.12` branch for now. But I don't think we need to wait for these for this PR.

PR #3614, opened 2022-05-27
Wrap ComputationClient::Computation inside of torch_xla::Computation

JackCaoG:
Refactor  `torch_xla::Computation` to unify it with the `ComputationClient::Computation`. The goal of this exercise is that
1. LTC upstream needs a way to represent a computation
2. runtime needs a way to represent a computation
3. Pytorch and TF(where our runtime resides) can not take dependency with each other

so we have to wrap a runtime Computation inside a `torch_xla::Computation` which inherits from `torch::lazy::Computation`.

another thing LTC does is to replace `CompileInstance` in https://github.com/pytorch/xla/blob/73aa5c5ef583291938f3251e899adc25498c68d9/third_party/xla_client/computation_client.h#L105
with `torch::lazy::Computation` too. Compare https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/lazy_graph_executor.cpp#L818 with https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor.cpp#L1675

FYI @will-cromar since `Computation` is used by pjrt as well.

### DIFF
--- a/torch_xla/csrc/computation.h
+++ b/torch_xla/csrc/computation.h
@@ -6,6 +6,7 @@
 #include "tensorflow/compiler/xla/client/xla_builder.h"
 #include "tensorflow/compiler/xla/client/xla_computation.h"
 #include "tensorflow/compiler/xla/service/hlo_module.h"
+#include "tensorflow/compiler/xla/xla_client/computation_client.h"
 #include "tensorflow/compiler/xla/xla_client/debug_macros.h"
 #include "tensorflow/compiler/xla/xla_client/types.h"
 #include "torch/csrc/lazy/backend/lowering_context.h"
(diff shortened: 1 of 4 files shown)


torch_xla/csrc/computation.h
```diff
@@ -6,66 +6,96 @@
...
+// In this case hash_ and name_ are not required. Computation would be a wrapper
+// around a xla::ComputationClient::Computation.
+// xla::ComputationClient::Computation::devices_ would be empty.
+// 3. To represent a computation that is already compiled. In this case name_
+// and hash_ are not required in this case. Computation will be a wrapper around
+// xla::XrtComputationClient::XrtComputation or
+// xla::PjRtComputationClient::PjRtComputation which contains a handle to
+// represent the compiled program.
+// It is not ideal to use same class for 3 different purposes but this is the
+// path took by upstream ltc.
```
wonjoolee95, line 33: 👍 👍 👍

wonjoolee95 (approved):
Thank you, this is awesome! Great comments 😄

Issue #3352, opened 2022-02-08
[Questions] long bernoulli time usage


cicirori:
## ❓ Questions and Help
I'm using torch xla to train some transformer-like models, and I got the timeline using PyTorch profiler and saw something very strange.
code:
```
import argparse
import time
import torch
import os
import time
from datasets import load_from_disk
from datetime import datetime as dt
from transformers import AutoModelForSequenceClassification, AutoTokenizer, DataCollatorWithPadding


parser = argparse.ArgumentParser()
parser.add_argument('--xla_enabled', action='store_true')
parser.add_argument('--amp_enabled', action='store_true')
parser.add_argument('--pytorch_profile', action='store_true')
parser.add_argument('--syncfree_optimizer', action='store_true')
parser.add_argument('--xla_debug_metrics_enabled', action='store_true')
parser.add_argument('--profile_folder', type=str, default='./profile_folder')
parser.add_argument('--dataset_path', type=str, default='./sst_data/train')
parser.add_argument('--num_epochs', type=int, default=3)
parser.add_argument('--batch_size', type=int, default=11)
parser.add_argument('--max_seq_length', type=int, default=512)
parser.add_argument('--break_step_for_profiling', type=int, default=25)
parser.add_argument('--model_name', type=str, default='gpt2')
args = parser.parse_args()
print("Job running args: ", args)

if args.xla_enabled:
  import torch_xla.core.xla_model as xm


def get_autocast_and_scaler():
  if args.xla_enabled:
    from torch_xla.amp import autocast, GradScaler
    return autocast, GradScaler()

  from torch.cuda.amp import autocast, GradScaler
  return autocast, GradScaler()

def metsumm(step=None):
    rep = met.metrics_report().split('\n')
    # print(rep)
    out = []
    prefix = '\tSTEP {}: '.format(step) if step else ''
    for i, line in enumerate(rep):
        if 'ExecuteTime' in line or 'TransferToServerTime' in line:
            out.extend([line,rep[i+1], rep[i+2]])
    for i in range(3):
        x = '{}{}'.format(prefix, out[i])
        x = x + ' '*(45-len(x)) + out[i+3].split(':')[-1]
        print(x)
    print('')

def loop_with_amp(model, inputs, optimizer, autocast, scaler):
  with autocast():
    outputs = model(**inputs)
    loss = outputs["loss"]

  scaler.scale(loss).backward()
  scaler.step(optimizer)
  scaler.update()

  return loss, optimizer


[...]

PR #3618, opened 2022-05-30
code for async scalar transfer

aws-rhsoln:
This PR is built on top of #3292 . Here we transfer the scalars to device in async mode. 
Consider the example in this comment: https://github.com/pytorch/xla/issues/3352#issuecomment-1042077820
```
denom = (exp_avg_sq.sqrt() / math.sqrt(bias_correction2)).add_(eps)
```
This step occurs in the adam optimizer (might be in other places too). In this case, the `bias_correction2` is a scalar that needs to be transferred to the device. Since this transfer happens synchronously, if a previous execution is running, this transfer would be blocked. This would block the model tracing and hence the model_tracing of the current step does not overlap with the execution of the previous step.
Here is what the profile looks like for a 2-layer bert on CPU with the transfer blocked.
Screen Shot 2022-05-29 at 10 09 43 PM

The optim_loop block is just xp.Trace over the optimizer step. It is blocked by transfer. 

By tracing the scalars asynchronously, the model tracing is no longer blocked and the execute computations are overlapping with the model tracing (this can be seen by reduced gap between the two executes and optim_loop not been stretched.

Screen Shot 2022-05-29 at 10 10 40 PM

### DIFF
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -207,14 +207,6 @@ class XlaDataCacheArena {
       device_caches_;
 };
 
-bool ShouldTransferScalarAsync() {
-  // TODO: Test the transfer_async with different models, and if it helps
-  // make it default
-  static const bool transfer_async =
-      xla::sys_util::GetEnvBool("XLA_TRANSFER_SCALAR_ASYNC", false);
-  return transfer_async;
-}
-
 XlaDataCacheArena::XlaDataCache* GetXlaDataCache(
     const torch::lazy::BackendDevice& device) {
   static const size_t kMaxCacheSize =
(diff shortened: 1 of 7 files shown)


JackCaoG:
If I remember correctly this "execution blocks transfer" behavior only happens in your asic? I remembered I tried to repro this behavior but failed.

torch_xla/csrc/tensor.cpp
```diff
@@ -219,18 +227,19 @@ torch::lazy::Value IrValueFromScalar(const at::Scalar& value,
...
 torch::lazy::BackendDataPtr GetDeviceData(
-    const at::Tensor& tensor, const torch::lazy::BackendDevice& device) {
+    const at::Tensor& tensor, const torch::lazy::BackendDevice& device,
+    bool transfer_async = false) {
   XlaDataCacheArena::XlaDataCache* cache = GetXlaDataCache(device);
   torch::lazy::BackendDataPtr device_data = cache->Get(tensor);
   if (device_data == nullptr) {
     at::Tensor tensor_copy = torch::lazy::CopyTensor(tensor);
-    device_data = TensorToXlaData(tensor_copy, device);
+    device_data = TensorToXlaData(tensor_copy, device, transfer_async);
```
JackCaoG, line 242: Let's move all of these logic (check env var) into `TensorToXlaData` instead of methods in `tensor.cpp`. The only reason I do it this way is I can't tell which call to `TensorToXlaData` is from RNG seed but if we want to transfer all scalar async then this additional layer is not necessary.

JackCaoG: I also don't see any particular scalar check(check whether a tensor is scalar before deciding it should be send async)

JackCaoG: We should also kill the RNG option if this `XLA_TRANSFER_SCALAR_ASYNC` covers everything you need.

aws-rhsoln:
> If I remember correctly this "execution blocks transfer" behavior only happens in your asic? I remembered I tried to repro this behavior but failed.

The above behavior is observed on CPU. So I think this behavior is visible on other devices too. It will become a bottleneck only when the execution time is comparable to model tracing time. In the CPU example above, the Execution time is very high and hence, this tracing cost is not a problem

torch_xla/csrc/tensor.cpp
aws-rhsoln: If we move the check to TensorToXLAData, we can kill RNG option and have two checks: 1) Env variable set check 2) If scalar check.
I can try to make those work

torch_xla/csrc/tensor_util.cpp
```diff
@@ -651,9 +651,11 @@ void PopulateTensorBuffer(const at::Tensor& tensor,
 
 torch::lazy::BackendDataPtr TensorToXlaData(
     const at::Tensor& tensor, const xla::Shape& shape,
-    const torch::lazy::BackendDevice& device, bool transfer_async) {
+    const torch::lazy::BackendDevice& device) {
   XLA_TIMED("TensorToData");
-  if (transfer_async) {
+  static const bool transfer_async =
+      xla::sys_util::GetEnvBool("XLA_TRANSFER_SCALAR_ASYNC", false);
+  if (tensor.dim() == 0 && tensor.numel() == 1 && transfer_async) {
```
JackCaoG, line 658: nit, can we move `transfer_async` check to first since we expect this to be false in may cases.

JackCaoG (commented):
mostly lgtm, can we look into why test failed? You can also rebase first to make sure it is a real issue.

torch_xla/csrc/tensor_util.cpp
aws-rhsoln: Fixed!

aws-rhsoln:
> mostly lgtm, can we look into why test failed? You can also rebase first to make sure it is a real issue.

Yeah let me take a look. Its coming from `TestLayerNormBackward`, I guess it might require a rebase. Will check.

JackCaoG (approved):
LGTM. our CI is currently broken so I would rather merge this one after CI issue is fixed to reduce the noise (I will take care of the merge).

JackCaoG (approved):
Thanks! I filed an internal issue to create a document to explain all of the environment variable we have today. It is kind of getting out of controls now days lol.

PR #3616, opened 2022-05-27
Load libtpu when ComputationClient is created.

will-cromar:
This lets us set topology env vars between after importing `torch_xla` and before creating the computation client. Example of restricting a process to one chip:

```
root@845104b51fc1:/root# PJRT_DEVICE=TPU python
Python 3.8.13 (default, Apr 20 2022, 18:59:55) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch_xla.core.xla_model as xm
2022-05-27 16:34:42.423356: I tensorflow/core/tpu/tpu_initializer_helper.cc:253] Libtpu path is: libtpu.so # libtpu load fails here
>>> import os
>>> os.environ["TPU_CHIPS_PER_PROCESS_BOUNDS"] = "1,1,1"
>>> os.environ["TPU_PROCESS_BOUNDS"] = "1,1,1"
>>> os.environ["TPU_VISIBLE_DEVICES"] = "0"
>>> xm.xla_device()
2022-05-27 16:35:53.310894: I tensorflow/compiler/xla/xla_client/pjrt_computation_client.cc:50] Initializing PjRt TPU client...
2022-05-27 16:35:53.310960: I tensorflow/core/tpu/tpu_initializer_helper.cc:253] Libtpu path is: /home/ptxla/.local/lib/python3.8/site-packages/libtpu/libtpu.so # actual libtpu path
2022-05-27 16:35:54.823864: I tensorflow/compiler/xla/service/service.cc:174] XLA service 0x559faa3f5d30 initialized for platform TPU (this does not guarantee that XLA will be used). Devices:
2022-05-27 16:35:54.823921: I tensorflow/compiler/xla/service/service.cc:182]   StreamExecutor device (0): TPU, 2a886c8
2022-05-27 16:35:54.823930: I tensorflow/compiler/xla/service/service.cc:182]   StreamExecutor device (1): TPU, 2a886c8
device(type='xla', index=0)
>>>
```

Previously, libtpu was loaded by [`tpu_api_dlsym_initializer`](https://github.com/tensorflow/tensorflow/blob/8c78a5a5f6d6e510f5bbc327b2efaeeaea4f3796/tensorflow/core/tpu/tpu_api_dlsym_initializer.cc#L25) in TensorFlow.

- Remove our direct dependency on `tpu_api_dlsym_initializer`
- Implement dynamic loading for `XrtComputationClient`
[...]

PR #3628, opened 2022-06-03
Refactor PjRt tests to take advantage of dynamic libtpu loading

will-cromar:
With the dynamic loading of libtpu implemented in #3616, we can more easily launch processes from the Python test script itself.

- Create `test/pjrt/common.py` with utilities for launching multiprocess+multithreaded workloads
  - Set the `XRT_SHARD_WORLD_SIZE` based on the number of visible devices (unblocks. TPU v4)
- Remove run_all_chips.sh
- Symlink `args_parse.py` from parent folder to `test/pjrt`
- Simplify MNIST and ImageNet PjRt tests

Tested with 1 TPU chip, 4 TPU chips, and CPU on v2-8 and v4-8

Will send a separate PR to update tests in ml-testing-accelerators.


test/pjrt/common.py
```diff
@@ -0,0 +1,92 @@
...
+def configure_tpu_topology(rank: int, processes: int, base_port=8476):
+  '''Set default TPU topology environment variables for a single TPU host.'''
+  ports = list(range(base_port, base_port + processes))
+  os.environ.setdefault('TPU_CHIPS_PER_PROCESS_BOUNDS', '1,1,1')
+  os.environ.setdefault('TPU_PROCESS_BOUNDS', '2,2,1')
+  os.environ.setdefault('TPU_PROCESS_ADDRESSES',
+                        ','.join(f'localhost:{port}' for port in ports))
+
+  os.environ.setdefault('TPU_VISIBLE_DEVICES', str(rank))
+  os.environ.setdefault('TPU_PROCESS_PORT', str(ports[rank]))
```
JackCaoG, line 28: what is this `TPU_PROCESS_PORT` used for?

test/pjrt/common.py
```diff
@@ -0,0 +1,92 @@
...
+    kwargs: kwargs to pass to `fn`
+
+  Returns:
+    Dict of the form {process_rank: {thread_rank: return_value}}, where
+    return_value is the result of calling `fn`.
+  '''
+  if pjrt_device() == 'TPU':
+    processes = num_visible_tpu_chips()
+  else:
+    processes = 1
```
JackCaoG, line 78: how many process are we going to have for multi gpu?

JackCaoG (commented):
mostly lgtm

test/pjrt/common.py
will-cromar: One process per GPU will almost certainly be best option. I haven't tested GPU at all yet, though

test/pjrt/common.py
will-cromar: It's an internal setting in libtpu that's only used when there are multiple communicating processes per TPU host. It corresponds to one of the ports in `TPU_PROCESS_ADDRESSES`

JackCaoG (approved):
Thanks!

JackCaoG:
Hopefully we can just use ddp for pjrt instead of introducing a new way for people to run pt/xla model. 😄

PR #3629, opened 2022-06-03
More conservative IsSparseGather check to favor dense gather on TPU

yeounoh:
Change the `IsSparseGather` check so that it favors dense gather on TPU more.


JackCaoG (approved):
Thanks!

PR #3634, opened 2022-06-07
Cleanup README

JackCaoG:
We should make it very clear that TPU Node is legacy and people should not use it anymore.


miladm (approved):
LGTM. Thanks.

PR #3636, opened 2022-06-07
delegate sym_sizes to sizes since Pytorch is now using sym_sizes

Krovatkin:
@JackCaoG  We are switching `THPVariable_size` to use `sym_sizes` here https://github.com/pytorch/pytorch/blob/b8382eb3c1d25c8d6aafd4bea0351db07c045f12/tools/autograd/templates/python_variable_methods.cpp#L98 in https://github.com/pytorch/pytorch/pull/78135 . I think to make XLA tests pass we need to implement `sym_sizes_custom`

### DIFF
--- a/torch_xla/csrc/tensor_impl.cpp
+++ b/torch_xla/csrc/tensor_impl.cpp
@@ -109,7 +109,8 @@ at::IntArrayRef XLATensorImpl::sizes_custom() const {
 }
 
 c10::SymIntArrayRef XLATensorImpl::sym_sizes_custom() const {
-  return sizes_custom();
+  auto sizes = sizes_custom();
+  return c10::SymIntArrayRef(sizes.data(), sizes.size());
 }
 
 at::IntArrayRef XLATensorImpl::strides_custom() const {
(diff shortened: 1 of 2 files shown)


JackCaoG:
Hey @Krovatkin , I am not as familiar to this part of the code. Can you add some context regarding  what `sym_sizes_custom` is being used for? It seems like it just wrap the current size(which is static) and wrap it inside a symint? Does this means we are going to always use symint going forward?

JackCaoG:
FYI @miladm

Krovatkin:
@JackCaoG Yes, we will be switching to using symints in Pytorch. It will happen gradually since we can't just switch all `sizes()` calls to `sym_sizes()`. 
@miladm will eventually need to reimplement `sym_sizes_custom` to return `SymbolicIntNode`s that wrap over lazy IR similar to 
https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/tensor_impl.cpp#L140 Note, this implementation returns `sym_sizes_` that are created in https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/tensor_impl.cpp#L89-L93

JackCaoG:
oh OK, I guess this is pretty much a no-op for pytorch/xla now, since our symint just a wrapper around static size.

Krovatkin:
> oh OK, I guess this is pretty much a no-op for pytorch/xla now, since our symint just a wrapper around static size

correct!

JackCaoG (approved):
Thanks!

Krovatkin:
@JackCaoG  thank you so much!!!

PR #3649, opened 2022-06-14
Register XLA lazy backend

JackCaoG:
In this pr we register the `XLABackend` to LTC and also register the `XLANativeFunctions` and `AutogradXLANativeFunctions` generated by LTC codegen. This helped us to get rid of the redundant codegen we have.


JackCaoG:
@Krovatkin This pr also fixed the CI issue you had in your `ExecuteComputation` pr.

JackCaoG:
Some CPP test seems to fail, will take a look tmr.

.gitignore
```diff
@@ -10,11 +10,6 @@ torch_xla/csrc/version.cpp
 *.pyc
 *.so
 
-# Files autogenerated by scripts/generate_code.sh
```
wonjoolee95, line 13: Just curious -- why are we able to remove this? IIRC, we generate our xla ops each build. Do our existing ops still get generated if we remove this?

torch_xla/csrc/xla_backend_impl.cpp
```diff
@@ -151,6 +158,10 @@ class XlaBackendImpl : public torch::lazy::BackendImplInterface {
     return at::DeviceType::CPU;
   }
 
+  void SetDefaultDeviceType(std::string type) override {
+    default_device_type_ = XlaDeviceType(c10::Device(type).type());
```
wonjoolee95, line 162: nit: is the `c10::Device(type).type()` always same as our own definition of `XlaDeviceType` as defined in https://github.com/pytorch/xla/blob/master/torch_xla/csrc/device.h#L13?

wonjoolee95 (approved):
Thanks! Had some nit questions, but LGTM.

torch_xla/csrc/xla_backend_impl.cpp
JackCaoG: Ah OK, I guess the right thing to do here is to get a defualt device and set the default type correctly. In the TPU setup we actually want it to be TPU. I will update this.

JackCaoG: hmm, wait. This function already takes string as type. What's the right thing to do here? How do I convert `string` to `XlaDeviceType`?

Maybe we need to convert  https://github.com/pytorch/xla/blob/dbc67604cfff305a37544513c20e377b254054ef/torch_xla/csrc/device.cpp#L53-L64

to a helper function? @wonjoolee95

wonjoolee95: What does the `c10::Device(type).type()` return? Is it just the type of the device in string format like `TPU` or `CPU`? If so, I think you're right we can move this to a helper function.

PR #3656, opened 2022-06-17
Implement IRBuilder

JackCaoG:
IRBuilder is being used by the LTC common code path. Sometimes we need to create some IR Node like `Scalar` or `DeviceData` in LTC. However different backends has different implementation of these nodes. IRBuilder provide a way for each backend to return the same logical IR.

### DIFF
--- a/torch_xla/csrc/ir_builder.h
+++ b/torch_xla/csrc/ir_builder.h
@@ -1,3 +1,4 @@
+#include "tensorflow/compiler/xla/xla_client/debug_macros.h"
 #include "torch/csrc/lazy/core/ir.h"
 #include "torch/csrc/lazy/core/ir_builder.h"
 #include "torch_xla/csrc/ops/as_strided.h"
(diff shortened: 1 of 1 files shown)


JackCaoG:
I think it might be better to add `XLA_ERROR` on all things we haven't implemented yet.. Working on that.

torch_xla/csrc/ir_builder.h
```diff
@@ -0,0 +1,143 @@
...
+#include "torch_xla/csrc/ops/expand.h"
+#include "torch_xla/csrc/ops/generic.h"
+#include "torch_xla/csrc/ops/ops.h"
+
+namespace torch_xla {
+
+struct XLAIrBuilder : IrBuilder {
+  NodePtr MakeDeviceData(
+      const std::shared_ptr<BackendData>& data) const override {
+    return torch::lazy::MakeNode<DeviceData>(data);
```
wonjoolee95, line 16: nit: regarding explicit namespaces, should we explicitly define namespaces for `torch::lazy` objects such as `NodePtr` and `Value` like `torch::lazy::MakeNode`?

wonjoolee95 (approved):
Thanks! One tiny nit, but overall LGMT.

torch_xla/csrc/ir_builder.h
JackCaoG: Oh right, yea. I should use `torch::lazy::NodePtr` instead.

JackCaoG:
I will merge this one if test passed.

PR #3542, opened 2022-04-29
[xla][cpu] Add Arm Compute Library (ACL) build option for torch-xla

snadampal:
ACL build support has been added under the build flag: XLA_CPU_USE_ACL

This will enable the acl runtime for xla once the below tensorflow/xla PR is merged.
https://github.com/tensorflow/tensorflow/pull/55534/commits

PR #3660, opened 2022-06-22
add aarch64 build scripts for torch and xla wheel building

snadampal:


snadampal:
Hi @JackCaoG, recently I have extended the TensorFlow xla aarch64 cpu backend with Arm Compute Library runtime (https://github.com/tensorflow/tensorflow/pull/55534) and also  opened  pytorch xla PR (https://github.com/pytorch/xla/pull/3542) to enable the same in PyTorch/xla.

context for the below PRs is... to provide one step build option on aarch64 cpus to enable the above optimizations. 
https://github.com/pytorch/xla/pull/3660 --> the current one for wheel builder
https://github.com/pytorch/xla/pull/3666 --> next one for the docker builder

JackCaoG:
Thanks, seems like tf pr is approved. We will need to update our tf pins after that and then we will review these prs.

snadampal:
I agree with waiting on Tensorflow for the Torch xla PR (https://github.com/pytorch/xla/pull/3542). However, the below two are not dependent on them. These two will enable the current torch xla on aarch64 cpus, though not fully optimized, still a good start to get torch xla working on aarch64 cpus. Sorry, if this point was not clear earlier.

https://github.com/pytorch/xla/pull/3660 --> the current one for wheel builder
https://github.com/pytorch/xla/pull/3666 --> next one for the docker builder

snadampal:
Hi @JackCaoG , TensorFlow xla [PR](https://github.com/tensorflow/tensorflow/pull/55534) got merged.

JackCaoG:
Thanks... before I go into details like refactor the build script with existing one, don't pin bazel version etc.. Let's discuss what's the use case of this script. We had the other script because we(PyTorch/XLA team)  publish nightly wheel and release wheel. I can understand the need to add a build command flag in case you want to build with aarch64, but upstreaming a script to build docker and wheel seems a lot(especially we don't plan to publish aarch64 wheel).

Do you mind sharing a bit more info regarding what this change will enable? Is it using the xla:cpu for aarch64 cpu? To us xla:cpu is just for the testing and developing purpose, we don't plan to use xla:cpu for any real workload.

snadampal:
Hi @JackCaoG , yes, this change will enable xla:cpu for aarch64 cpu.  I've observed ~3x performance improvements for natural language processing workloads with xla+cpu enhancements compared to the default runtime+openBLAS backend.
Please let me know the constraints you see for enabling xla:cpu for real workloads...

snadampal:
Hi @JackCaoG , could you please let me know what the next steps are for this PR. thank you!

JackCaoG:
I don't think we should create a new script, can you reuse the existing build script we have and add your option there>?

snadampal:
@JackCaoG , though this new script is derived from the existing, it has got too many differences: arm compute library, onednn version and build command, bazel, and torch-xla build commands. I thought it might make the script too convoluted to incorporate the changes, thats why I went with aarch64 specific which would be easier to maintain as well going forward.
I will give another attempt to merge this into the existing and see.

JackCaoG:
@will-cromar want to get your input in this one. Do you think it is better to keep a separate script for arm build? I know you are working on reducing our docker image size, not sure if this pr will have any consequence to that.

will-cromar:
I'm working on a broader refactoring of our build process in #4044, especially with an eye toward reducing the size of the final release image. I don't see anything in this PR that would be incompatible with the new build, unless you specifically need something from `conda`, which I removed. Can you help clarify the specific things you are changing for aarch64 and why they are necessary? I see the following:

- Skip installing `anaconda` base packages
- Skip installing `mkl-include`
- Set `XLA_CPU_USE_ACL=1` while building `torch_xla`

Regarding one script vs two scripts possible, I would rather keep the common parts of the build script together in one place.

snadampal:
Hi @will-cromar , I wouldn't need anything from conda in specific, since my work is based on the current scripts I used the conda path.  Is there any pre-requisite on the host os and versions, like ubuntu, centos etc? As I mentioned earlier, I'm hoping to get these scripts part of 1.13 release. Is the new build process targeted for 1.13? if yes, I'm happy to rework this PR to comply with new scripts.

will-cromar:
We'll keep the current build scripts through 1.13. I believe the current build process uses `buster` like my WIP script (except for the current GPU build), so I don't expect an incompatibility there. This change LGTM, I don't see any changes that I expect to be incompatible with the new build process.

JackCaoG (approved):

snadampal:
thanks, @JackCaoG and @will-cromar  for merging the pr.

miladm:
Thanks for pushing this PR.

Currently, we run our CI unit tests on x86 backend. Here, I don't see a discussion future CI test plan for aarch64. Please keep in mind that going forward this may turn into an issue we want to address - unless you already have a solution to share.

CC @JackCaoG @will-cromar @snadampal

PR #3664, opened 2022-06-24
Overlay secondary llvm-raw remote cache without changing OSS tensorflow & llvm p…

yeounoh:
Addresses https://github.com/pytorch/pytorch/issues/80034


yeounoh:
The build is blocked on [comment](https://github.com/pytorch/pytorch/issues/80034#issuecomment-1165972591)

JackCaoG:
The change lgtm, I will approve it once test is green

build_torch_xla_libs.sh
```diff
@@ -71,6 +71,13 @@ if [ "$CMD" == "clean" ]; then
   bazel clean
   popd
 else
+  # Overlay llvm-raw secondary cache. The remote cache should be updated
+  # nightly with the pinned llvm archive. Note, this commands will be NO-OP if there is no match.
+  sed -i '/.*github.com\/llvm.*,/a "https://storage.googleapis.com/tpu-pytorch/llvm-raw/{commit}.tar.gz".format(commit = LLVM_COMMIT),' \
+    $THIRD_PARTY_DIR/tensorflow/third_party/llvm/workspace.bzl
```
JackCaoG, line 77: so this only get updated when pytorch/xla update its tf pin?

yeounoh: This is local update to the build file, it needs to be done before every build. It's local because the change is temporary and specific to each CI run.

yeounoh:
All tests have passed, and the upstream build looks good, too. We have to land this first, and then re-enable the upstream XLA workflow.

JackCaoG (approved):

PR #3666, opened 2022-06-26
add aarch64 docker for torch and xla wheels

snadampal:


snadampal:
This PR is dependent on https://github.com/pytorch/xla/pull/3660

snadampal:
Looks like new docker file is not required for aarch64. I'm making the default Dockerfile to work, fixing the wheel builder script which gets invoked directly from the Dockerfile for the actual build.

snadampal:
Hi @JackCaoG ,  I've updated the default Dockerfile and tested it on aarch64 with ubuntu:focal base image.
docker build -t --build-arg base_image=ubuntu:focal torch-xla -f docker/Dockerfile .

snadampal:
Hi @JackCaoG, @will-cromar , please let me know if there is any feedback on this.

docker/Dockerfile
```diff
@@ -14,9 +14,12 @@ ARG bazel_jobs=""
 ARG git_clone="true"
 
 RUN apt-get update
-RUN apt-get install -y git sudo python-pip python3-pip
+RUN apt-get install -y git sudo python3-pip
 RUN git clone https://github.com/pytorch/pytorch
 
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
+RUN sudo apt-get install -y -q
```
JackCaoG, line 21: Could you elaborate a bit on what this line is trying to do?

snadampal: I was getting debian front end config issues, it was going into interactive mode and expecting user response. 
I see you aleady have this for the wheel script (https://github.com/pytorch/xla/blob/master/scripts/build_torch_wheels.sh#L9) , but still it was going to interactive mode

> DEBIAN_FRONTEND=noninteractive

When I looked around, this is where I found the above solution
https://github.com/moby/moby/issues/27988#issuecomment-462809153

PR #3667, opened 2022-06-27
Disable ECR image update

yeounoh:
This is to temporarily disable ECR image update until we stabilize the upstream build [issue](https://github.com/pytorch/pytorch/issues/80034)


JackCaoG (commented):
Could you explain a bit why this will fix the issue? I understand that something has changed and while we try to re-tag we get the error in CI. However I am don't understand what has been changed. AFAIK your pr regarding llvm cache has not merged.

yeounoh:
This will unblock us since the issue now is that the upstream ECR repo is immutable (i.e., `:latest:` tag is not really tracking the latest image). On the other hand,  our latest image in GCR has the local llvm archive in it, so the hash has changed -- I've updated it after testing it locally and to test in the upstream.

PR #3671, opened 2022-06-29
Verify Slogdet

JackCaoG:
This is to verify whether all test passed for https://github.com/pytorch/pytorch/pull/79742


JackCaoG:
Able to repo error with 
```
(Pdb) t_inp
tensor([], device='xla:0', size=(0, 0))
(Pdb) torch.logdet(t_inp)
*** RuntimeError: Error while lowering: [Float[]] aten::logdet, location=<module>@<stdin>:1
XLA builder error: INVALID_ARGUMENT: limit index (-1) must be greater or equal to start index (0) in slice with positive stride in slice operation; argument shape: f32[0]; starts: {0}; limits: {-1}; strides: {1}.: 
```

JackCaoG:
So this is a bit interesting and annoying
```
(Pdb) torch.logdet(torch.randn(0,0))
tensor(0.)
(Pdb) torch.logdet(torch.randn(0,0,0))
tensor([])
(Pdb) torch.logdet(torch.randn(0,0,0,0))
tensor([], size=(0, 0))
(Pdb) torch.logdet(torch.randn(0,0,0,0,0))
tensor([], size=(0, 0, 0))
(Pdb) torch.logdet(torch.randn(0,0,0,0,0,0))
tensor([], size=(0, 0, 0, 0))
```

`xla::logdet` does not handle empty input. I tried to workaround it by always return a 0, but it would be hard to mimic all these edge cases.
```
 torch_xla::XlaOpVector Logdet::Lower(LoweringContext* loctx) const {
   xla::XlaOp xla_input = loctx->GetOutputOp(operand(0));
-  return ReturnOp(xla::LogDet(xla_input), loctx);
+  const xla::Shape& input_shape = XlaHelpers::ShapeOfXlaOp(xla_input);
+  bool input_not_empty = false;
+  for (int i = 0; i < input_shape.dimensions_size(); i++) {
+    if (input_shape.dimensions(i) > 0) {
+      input_not_empty = true;
+      break;
+    }
+  }
+  // xla::LogDet does not handle empty input
+  if (input_not_empty) {
+    return ReturnOp(xla::LogDet(xla_input), loctx);
+  } else {
+    return ReturnOp(
+        xla::Zero(xla_input.builder(), XlaHelpers::TypeOfXlaOp(xla_input)),
+        loctx);
+  }
```

This is not a regression, my guess is that somehow test used to fallback to cpu and was never run against xla device. I might disable the test for now instead of introducing more hacks to map the pytorch behavior.

JackCaoG:
@Lezcano since this one is only about disabling the test, I can merge without waiting for pytorch pr. Let me know when you are ready on your end, I will then merge this pr.

Lezcano:
I'm quite impressed that there's no `xla::Shape::numel()` to get the number of elements and a `xla::Shape::empty()`...

But yes, please, merge this and I'll do the usual dance of changing the pin and all that in my PR

wonjoolee95 (approved):

JackCaoG:
@Lezcano I think tonight pytorch bot will update xla hash pass this pr, so you actually don't need to do anything this time. 😄

Lezcano:
what a time to be alive.
Since I don't know what's "tonight" (EU / East / Pacific?) I'll just merge the PR tomorrow or on Monday.
Thank you!

PR #3672, opened 2022-06-29
Make computation to take string when used for compilation

JackCaoG:
Currently `Computation` is used for both represent a `CompileInstance` and `CompiledProgram`. In this change, we make `XlaBuilder` to inject the `device` into the `torch_xla::Computation`. this way during the compilation, we know the device hw type.


wonjoolee95 (approved):
👍

JackCaoG:
This one should be low risk, merge it for now. TODO: add test.

torch_xla/csrc/xla_backend_impl.cpp
```diff
@@ -125,7 +125,8 @@ class XlaBackendImpl : public torch::lazy::BackendImplInterface {
       // TODO(JackCaoG): Verify this with GPU, we might only have 1 process with
       // multiple GPU as replica.
       compile_instances.push_back(xla::ComputationClient::CompileInstance(
-          torch_xla_computation->move_computation(), current_device.toString(),
+          torch_xla_computation->move_computation(),
+          torch_xla_computation->get_device_string(),
```
yeounoh, line 129: shoudn't this be,
```
(
  torch_xla_computation->move_computation(), 
  current_device.toString(),
  {torch_xla_computation->get_device_string()}, &shape
)
```
since there is one compilation device, and we can remove the 1-device assumption in `get_device_string()`.

yeounoh: cc @JackCaoG

PR #3681, opened 2022-07-04
Fix typo `partecipating` -> `participating`

philipturner:
Synchronizing the code base with S4TF: https://github.com/philipturner/s4tf/commit/2c4d12e60682f879fdce8ad3c333b34e98a35b17. This commit is to a branch involved with https://github.com/s4tf/s4tf/pull/17; philipturner/s4tf is not the head repository of Swift for TensorFlow.

Quick question: when do you plan on upgrading from TF 2.8 to TF 2.9? I have done that already.

### DIFF
--- a/torch_xla/csrc/tensor.h
+++ b/torch_xla/csrc/tensor.h
@@ -180,7 +180,7 @@ class XLATensor : public c10::intrusive_ptr_target {
   // Applies all the pending IR operations queued over the input tensors. All
   // the tensors must be on the same device. If wait is true, the sync operation
   // will be run synchronously. The devices argument, if not empty, tells the
-  // devices which should be partecipating into the replicated computation.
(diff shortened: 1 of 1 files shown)


JackCaoG:
we don't sync on tf releases, we just update our tf pin rough every month.

JackCaoG:
It is OK if you want to modify these comments, but most of these will be gone since majority of the lazy tensor system has been upstreamed to https://github.com/pytorch/pytorch/tree/master/torch/csrc/lazy

JackCaoG:
pt/xla will not be archived, we still maintain the xla lowering, runtime, some api etc...It is just the lazy tracing part will mostly lived in upstream pytorch so other pytorch backend can also make use of it.

JackCaoG (approved):

PR #3684, opened 2022-07-06
[SPMD][PoC] compile & execute with PjRt

yeounoh:
This is a follow-up to #3476 and contributes to #3871. The changes include:
- `Compile` partitioned HLO computation graph with sharding annotations.
- `PjRtComputationClient` integration to support `SPMD` sharded operations.
- `PjRtShardedData` struct to represent sharded `Data`.
- `InputHandler` for parameter sharding and sharded data transfer.
- Remove duplicate copies of sharding annotations.
- `ExecuteReplicated` for partitioned computation.

The PoC implementation supports `replicated` and `tiled` sharding annotations, and single-host `xla:tpu` backend. This enables a simple sharded computation on v3-8, like
```
    t1 = torch.randn(1, 128, device='cpu')
    t2 = torch.randn(1, 128, device='cpu')
    expected = t1 @ t2.T

    xt1 = t1.to(xm.xla_device())
    xt2 = t2.to(xm.xla_device())
    xs.mark_sharding(xt1, (1, 8), (0, 1))
    self.assertEqual('{devices=[1,8]0,1,2,3,4,5,6,7}',
                     torch_xla._XLAC._get_xla_sharding_spec(xt1))

    actual = (xt1 @ xt2.T).cpu()
```


yeounoh:
Going to breakdown this PR into smaller ones and land them separately. Some of the unit tests need to land in the end, after landing all the changes cc @JackCaoG

yeounoh:
cc @jonb377 FYI

yeounoh:
@JackCaoG could you help review this PR, this would be the last one of the three PoC PRs. We need this to unblock @steventk-g for his virtual device work. I will create a separate follow-up PR to address any comments.

JackCaoG:
you need to update this pr a bit to account for new tf update, should not be too much code changes through

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -267,6 +267,7 @@ PjRtComputationClient::ExecuteComputation(
   xla::ExecuteOptions execute_options;
   execute_options.untuple_result = options.explode_tuple;
   execute_options.strict_shape_checking = false;
+
```
JackCaoG, line 270: is this one typo or intentation?

third_party/xla_client/xrt_computation_client.h
```diff
@@ -252,15 +252,15 @@ class XrtComputationClient : public ComputationClient {
...
-  void TransferToServer(absl::Span<const TensorSource> tensors,
-                        absl::Span<const DataPtr> datas) override;
-
   DataPtr TransferShardsToServer(absl::Span<const TensorSource> tensor_shards,
                                  std::string device,
                                  xla::Shape shape) override {
     XLA_ERROR() << __FUNCTION__ << " not implemented";
   }
 
+  void TransferToServer(absl::Span<const TensorSource> tensors,
```
JackCaoG, line 261: why the move?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1443,55 +1402,39 @@ void InitXlaModuleBindings(py::module m) {
...
+         bool bidirectional_windowed_einsum = false) -> std::string {
+        xla::HloModuleConfig config;
+        config.set_use_spmd_partitioning(true);
+        config.set_replica_count(num_replicas);
+        config.set_num_partitions(num_devices);
+
+        std::string hlo_text = GetTensorsHloGraph(tensors);
+        auto hlo_module_error =
+            xla::ParseAndReturnUnverifiedModule(hlo_text, config);
+        if (!hlo_module_error.ok()) {
```
JackCaoG, line 1423: should we just error out here with `XLA_CHECK_OK`?

torch_xla/csrc/tensor.cpp
```diff
@@ -699,6 +703,13 @@ void XLATensor::SetInPlaceIrValue(torch::lazy::Value ir_value) {
 }
 
 void XLATensor::AssignIrValue(torch::lazy::Value ir_value) const {
+  // Sharding annotation should be persisted, since the xla_data() is sharded.
```
JackCaoG, line 706: I don't fully get this part, can you explain a bit more. In what case `sharding` will be non-empty? What if it is already being set?

torch_xla/csrc/tensor.cpp
```diff
@@ -1430,31 +1448,58 @@ std::shared_ptr<XLATensor::Async> XLATensor::ScheduleSyncTensorsGraph(
     ComputationCache::TypePtr cached_computation) {
   tensorflow::profiler::TraceMe activity(
       "ScheduleSyncTensorsGraph", tensorflow::profiler::TraceMeLevel::kInfo);
+  bool is_sharded = cached_computation->is_sharded;
+  std::cout << "ScheduleSyncTensorsGraph, is_sharded: " << is_sharded
```
JackCaoG, line 1452: remove the cout

torch_xla/csrc/tensor.cpp
```diff
@@ -1430,31 +1448,58 @@ std::shared_ptr<XLATensor::Async> XLATensor::ScheduleSyncTensorsGraph(
...
+            xla::ComputationClient::Get()->GetAllDevices();
+        std::vector<std::vector<xla::ComputationClient::DataPtr>>
+            device_arguments = torch_xla::ShardingUtil::InputHandler(
+                UnwrapXlaData(async->parameters_data), devices);
+        xla::ComputationClient::ExecuteReplicatedOptions execute_options;
+        execute_options.explode_tuple = options.explode_tuple;
+
+        TF_VLOG(3) << "Executing IR graph hash "
+                   << torch::lazy::HashToString(hash) << " on all devices.";
+        results = xla::ComputationClient::Get()->ExecuteReplicated(
```
JackCaoG, line 1478: hmm, `results` here is `virtual_data_ptr`? I saw done below we do `UnwrapXlaData(async->tensors_data[i])->Assign(*results[i])` so each element in `result` here should represent a tensor in a virtual device, hence should be a `virtual_data_ptr`, is my understanding correct?

torch_xla/csrc/tensor.cpp
```diff
@@ -1754,7 +1802,8 @@ XLATensor::CompilationResult XLATensor::Compile(
   return {/*device=*/coll.device,
           /*emitted_nodes=*/lowering_ctx.GetEmittedNodeCount(),
           /*computation=*/std::move(computations.front()),
-          /*parameters_data=*/std::move(po_data->parameters_data)};
+          /*parameters_data=*/std::move(po_data->parameters_data),
+          /*is_sharded=*/is_sharded};
```
JackCaoG, line 1806: @alanwaketan I guess we need to add `is_sharded ` to `torch_xla_computation` in https://github.com/pytorch/xla/blob/master/torch_xla/csrc/computation.h#L119

torch_xla/csrc/tensor.h
```diff
@@ -1248,14 +1249,20 @@ class XLATensor : public c10::intrusive_ptr_target {
...
     std::shared_ptr<xla::ComputationClient::Computation> computation;
     std::vector<torch::lazy::BackendDataPtr> parameters_data;
+    bool is_sharded = false;
   };
 
   struct CachedComputation {
     CachedComputation(
         std::shared_ptr<xla::ComputationClient::Computation> computation)
         : computation(std::move(computation)) {}
+    CachedComputation(
```
JackCaoG, line 1259: can we just use the old constructor and make `is_sharded` default to false?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -21,6 +21,26 @@ namespace {
 
 using xla::internal::XlaBuilderFriend;
 
+std::vector<int64_t> TileAssignmentDimensions(
```
JackCaoG, line 24: can we add a comment here?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -157,15 +234,12 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
...
-    // TODO(yeounoh) support higher-order topology
-    XLA_CHECK(sharding.tile_shape().dimensions_size() <= 2);
-    XLA_CHECK(tensor.sizes().size() >= sharding.tile_shape().dimensions_size())
-        << "Data dimensions should be at least as large as the tile shape "
-           "dimensions.";
+    CHECK(sharding.tile_shape().dimensions_size() <= 2);
+    CHECK(tensor.sizes().size() >= sharding.tile_shape().dimensions_size());
 
     auto tile_shape = sharding.tile_assignment_dimensions();
     for (int64_t core : sharding.tile_assignment_devices()) {
```
JackCaoG, line 245: let's not use core here, on v4 going forward, we use chip. I think it is better to just use `device` here to make it general.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -38,14 +58,71 @@ bool ShardingUtil::SetHloSharding(LoweringContext* lowering_ctx) {
...
+        break;
+      case 4:
+        tile_array.Each([&](absl::Span<const int64_t> indices, int64_t* v) {
+          auto r = tile_assignment[indices[0]].cast<py::list>();
+          r = r[indices[1]].cast<py::list>();
+          r = r[indices[2]].cast<py::list>();
+          *v = r[indices[3]].cast<int64_t>();
+        });
+        break;
+      case 5:
```
JackCaoG, line 103: we need test for all these cases.

JackCaoG (changes requested):
This pr is still too big and hard to review. I would argue that we don't have enough test to confidently land this. It is OK to make an exception this time given the timeline but we should add more tests to cover all new features adding in this pr.

torch_xla/csrc/tensor.cpp
alanwaketan: You mean add to upstream? I guess we can add that.

third_party/xla_client/xrt_computation_client.cc
```diff
@@ -1113,7 +1113,7 @@ std::unique_ptr<xrt::XLAComputation> XrtComputationClient::CreateXrtComputation(
         output_shape->ToProto();
   }
   *xrt_computation->mutable_hlo_snapshot() =
-      std::move(*computation.Snapshot().value());
+      std::move(*computation.Snapshot().ConsumeValueOrDie());
```
steventk-g, line 1116: It looks like this is breaking the build
```
tensorflow/compiler/xla/xla_client/xrt_computation_client.cc:1116:41: error: ‘class tensorflow::StatusOr<std::unique_ptr<xla::HloSnapshot> >’ has no member named ‘ConsumeValueOrDie’; did you mean ‘ValueOrDie’?
       std::move(*computation.Snapshot().ConsumeValueOrDie());
                                         ^~~~~~~~~~~~~~~~~
                                         ValueOrDie
```

third_party/xla_client/xrt_computation_client.h
yeounoh: To put them in the same order as defined in the parent class. Come to think of it, it would look cleaner if we re-order the parent methods instead.

torch_xla/csrc/device.cpp
```diff
@@ -72,6 +77,11 @@ const torch::lazy::BackendDevice* GetDefaultDevice() {
   return &default_device;
 }
 
+torch::lazy::BackendDevice GetVirtualDevice() {
+  // TODO(yeounoh) tensor.device().type() == kSPMD
```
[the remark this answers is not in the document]
yeounoh: Define kSPMD device type within PyTorch/XLA. We don't need this, since Steven is going to expand/ re-define.

torch_xla/csrc/init_python_bindings.cpp
yeounoh: I like the idea, ack.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: It's intentional, to separate the options/parameters declaration and the usage in a function call.

torch_xla/csrc/tensor.cpp
yeounoh: Done

torch_xla/csrc/tensor.cpp
yeounoh: Sorry for the confusion, rephrasing:
```
// Sharding annotation it not null, if xla_data() is sharded.
// TODO(yeounoh): Sharding annotation must be removed by explicit call to
// ClearSharding.
  ```

File: torch_xla/csrc/tensor.cpp [hunk 14]
```diff
@@ -1430,31 +1448,58 @@ std::shared_ptr<XLATensor::Async> XLATensor::ScheduleSyncTensorsGraph(
...
+            << std::endl;
+  auto computation = Computation(cached_computation->computation);
+  std::cout << "- " << computation.to_string() << std::endl;
   TensorCollectionBarrier(coll);
   std::shared_ptr<Async> async = std::make_shared<Async>(
       coll, std::move(parameters_data), std::move(tensors_data),
       std::move(cached_computation));
 
-  auto syncfn = [async, hash = coll->hash]() {
+  auto syncfn = [async, is_sharded, hash = coll->hash]() {
```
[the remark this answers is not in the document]
yeounoh: Good point, thanks -- done.

torch_xla/csrc/tensor.h
yeounoh: Good point, thanks -- done.

torch_xla/csrc/tensor.cpp
yeounoh: Executed results are not virtual and reside on actual devices. The scope of virtual device is to defer physical device data creation until we are certain about the sharding strategies, so we can avoid sending the full unpartitioned data before sending partitioned data again. Do we want to expand the notion of "virtual" device and its usage in our codebase?

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: Ack

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Done

third_party/xla_client/xrt_computation_client.cc
yeounoh: Thanks, let me look into this -- I was able to build this before my leave for the demo, but it might have something to do with the new tf update.

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: Good point. Ack

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: Good point, we only test 1~3 dimensional sharding. Let me add a couple of test cases to cover 4th and 5th dimensions, and we should be good.

yeounoh:
> This pr is still too big and hard to review. I would argue that we don't have enough test to confidently land this. It is OK to make an exception this time given the timeline but we should add more tests to cover all new features adding in this pr.

Thanks for the review @JackCaoG . Though it's crucial to land this to unblock Steven for his work, I want to make sure that we go over the changes here and flag any concerns, and possibly address all of them, especially the test coverage. It shouldn't take too long.

yeounoh:
> you need to update this pr a bit to account for new tf update, should not be too much code changes through

Do we have a reference PR in handy?

JackCaoG:
> > you need to update this pr a bit to account for new tf update, should not be too much code changes through
> 
> Do we have a reference PR in handy?

https://github.com/pytorch/xla/pull/4036 I think you just need to update `ConsumeValueOrDie` to the new function call.

test/cpp/test_xla_sharding.cpp
```diff
@@ -56,12 +57,32 @@ TEST_F(XLAShardingTest, ShardTensor) {
...
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 1, 2}));
+
+  // Replicated, all shards should be identical.
   sharding = xla::HloSharding::Replicate().ToProto();
   shards = ShardingUtil::ShardTensor(tensor, sharding, devices);
   EXPECT_EQ(shards.size(), 8);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({8, 7, 4}));
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 7, 4}));
+
+  // 4D tiled, the first and second dims are replicated and the last halved. The
```
yeounoh, line 76: @steventk-g I've also added the test case for 4-dimension tensor input. Follow these examples (e.g., cute and tesseract cases), and it should be straightforward to add a case for 5-dimension input (let's do that in a separate PR, but this is lower priority than virtual device work). Writing the test case should be quick, but try to spend some extra time studying `ShardingUtil::ShardTensor` -- but let's box our time no more than half a day.  cc @JackCaoG

yeounoh:
@JackCaoG I addressed the comments, thanks again for the review. I was able to build locally and test.

JackCaoG:
Thanks, I will take a look later today

JackCaoG:
Seems like test strict out crashed. @yeounoh were you able to run a simple pytorch/xla program locally?

yeounoh:
> Seems like test strict out crashed. @yeounoh were you able to run a simple pytorch/xla program locally?

Thanks @JackCaoG , the tests were green at some point (or not??). Let me debug this sometime today.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1375,59 +1373,20 @@ void InitXlaModuleBindings(py::module m) {
...
+    // TODO(yeounoh) use `SPMD` device to avoid input at::Tensor uploading.
+    // For now, we move the pre-uploaded data back to cpu_tensor for sharding.
     XLATensorPtr xtensor = bridge::GetXlaTensor(input);
+    std::vector<XLATensorPtr> xla_tensors{xtensor};
+    at::Tensor cpu_tensor = XLATensor::GetTensors(&xla_tensors)[0];
     auto sharding_spec = std::make_shared<XLATensor::ShardingSpec>(sharding);
+    auto xla_data = CreateTensorsData(
+        std::vector<at::Tensor>{cpu_tensor},
+        std::vector<XLATensor::ShardingSpecPtr>{sharding_spec},
+        std::vector<std::string>{GetVirtualDevice().toString()})[0];
```
steventk-g, line 1388: After virtual device is implemented, we'll use it to get the full list of device strings and pass that here, instead of passing a virtual device string itself.

No action needed

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1375,59 +1373,20 @@ void InitXlaModuleBindings(py::module m) {
...
-      }
-    }
+    xla::OpSharding sharding =
+        ShardingUtil::CreateOpSharding(tile_assignment, replicated, manual);
 
+    // TODO(yeounoh) use `SPMD` device to avoid input at::Tensor uploading.
+    // For now, we move the pre-uploaded data back to cpu_tensor for sharding.
     XLATensorPtr xtensor = bridge::GetXlaTensor(input);
+    std::vector<XLATensorPtr> xla_tensors{xtensor};
+    at::Tensor cpu_tensor = XLATensor::GetTensors(&xla_tensors)[0];
```
steventk-g, line 1383: With virtual device, we will no longer to cast the input back and forth to get a cpu_tensor. The input itself will live on the cpu at the time this method is called.

No action needed

torch_xla/csrc/device.cpp
```diff
@@ -72,6 +77,10 @@ const torch::lazy::BackendDevice* GetDefaultDevice() {
   return &default_device;
 }
 
+torch::lazy::BackendDevice GetVirtualDevice() {
+  return ParseDeviceString("SPMD:0");
```
steventk-g, line 81: We won't want to use device strings for virtual device. Instead, once virtual device is implemented, this will return a global variable, like `GetCurrentDevice()` below.

No action needed.

steventk-g:
Here's a link to the virtual device draft PR, it's still an untested work-in-progress
https://github.com/pytorch/xla/pull/4091

yeounoh:
CPU test passes, but the GPU fails with the following somewhat unrelated (at least on the outset) error:
```
*** Begin stack trace ***
	
	
	gsignal
	abort
	
	xla::XrtLocalService::XrtLocalService(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int)
	xla::XrtComputationClient::MaybeCreateLocalService(xla::XrtComputationClient::Options const&)
	xla::XrtComputationClient::XrtComputationClient(xla::XrtComputationClient::Options, std::unique_ptr<tensorflow::tpu::TopologyProto, std::default_delete<tensorflow::tpu::TopologyProto> >)
	xla::ComputationClient::Create()
	
	
	xla::ComputationClient::Get()
	
	
	_PyMethodDef_RawFastCallKeywords
	
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallKeywords
	
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallDict
	
	_PyObject_GenericGetAttrWithDict
	_PyEval_EvalFrameDefault
	_PyEval_EvalCodeWithName
	_PyFunction_FastCallKeywords
	
	_PyEval_EvalFrameDefault
	
	
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallKeywords
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallKeywords
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallDict
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallDict
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallKeywords
	
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallKeywords
	
	_PyEval_EvalFrameDefault
	_PyFunction_FastCallKeywords
	_PyEval_EvalFrameDefault
	_PyEval_EvalCodeWithName
	_PyFunction_FastCallKeywords
	
	_PyEval_EvalFrameDefault
	_PyEval_EvalCodeWithName
	PyEval_EvalCode
	
	PyRun_StringFlags
	PyRun_SimpleStringFlags
	
	_Py_UnixMain
	__libc_start_main
	
*** End stack trace ***
  File "/tmp/pytorch/xla/test/test_torch_distributed_multi_all_reduce_xla_backend.py", line 38, in <module>
    xmp.spawn(_mp_fn, args=())
  File "/opt/conda/lib/python3.7/site-packages/torch_xla-1.14-py3.7-linux-x86_64.egg/torch_xla/distributed/xla_multiprocessing.py", line 399, in spawn
    start_method=start_method)
  File "/opt/conda/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 198, in start_processes
    while not context.join():
  File "/opt/conda/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 146, in join
    signal_name=name
torch.multiprocessing.spawn.ProcessExitedException: process 0 terminated with signal SIGABRT

```

The mater branch is green, though. cc @JackCaoG

JackCaoG:
Seesm irrelevant, let me just restart the gpu ci

yeounoh:
> Seesm irrelevant, let me just restart the gpu ci

Yea, this [one](https://app.circleci.com/pipelines/github/pytorch/xla/13795/workflows/7206b828-64a7-4a6c-8307-cee3a8b54e06) succeeded. Thanks @JackCaoG

test/cpp/test_xla_sharding.cpp
JackCaoG: @steventk-g Please open a github issue or bug to track this so we don't forget :D

test/test_xla_sharding.py
```diff
@@ -55,6 +54,21 @@ def test_mark_sharding(self):
     actual = (xt1 @ xt2.T).cpu()
     self.assertTrue(torch.allclose(expected, actual))
 
+  def test_mark_sharding_4d(self):
+    t = torch.randn(2, 4, 8, 16, device='cpu')
+    expected = t + t
+
+    xt = t.to(xm.xla_device())
+    n_devices = xm.xrt_world_size()
```
JackCaoG, line 62: Does CI run this test or we only run it on TPU?

third_party/xla_client/computation_client.h
```diff
@@ -220,18 +220,18 @@ class ComputationClient {
...
-      absl::Span<const TensorSource> tensor_shards, std::string device,
-      xla::Shape shape) = 0;
-
   // Transfers local tensor values to the TPU servers and fetches the handles.
   // Update the handles associated with DataPtrs passed instead of creating new
   // datas.
   virtual void TransferToServer(absl::Span<const TensorSource> tensors,
                                 absl::Span<const DataPtr> datas) = 0;
 
+  // Transfers local sharded tensor values to the TPU servers and returns a
```
JackCaoG, line 229: I would use TPU Device instead of TPU Server, there is no server in PJRT context.

test/cpp/test_xla_sharding.cpp
steventk-g: Done https://github.com/pytorch/xla/issues/4093

torch_xla/csrc/tensor.cpp
```diff
@@ -632,22 +635,29 @@ std::string XLATensor::DumpHloComputation(
...
+    return std::make_shared<ShardingSpec>(*sharding);
   }
-  return std::make_shared<ShardingSpec>(*sharding);
+  return nullptr;
 }
+
 void XLATensor::SetShardingSpec(const ShardingSpec& sharding_spec) {
   XLA_CHECK(GetIrValue().node != nullptr) << "Tyring to access a null cursor";
-  dynamic_cast<XlaNode*>(data()->ir_value.node.get())
+  dynamic_cast<XlaNode*>(GetIrValue().node.get())
```
JackCaoG, line 652: hmm, we should add a `XlaNodeCast` to replace `dynamic_cast<XlaNode*>` so it is cleaner

torch_xla/csrc/tensor.cpp
```diff
@@ -699,6 +709,14 @@ void XLATensor::SetInPlaceIrValue(torch::lazy::Value ir_value) {
 }
 
 void XLATensor::AssignIrValue(torch::lazy::Value ir_value) const {
+  // Sharding annotation it not null, if xla_data() is sharded.
+  // TODO(yeounoh): Sharding annotation must be removed by explicit call to
+  // ClearSharding.
+  ShardingSpecPtr sharding = sharding_spec();
+  if (sharding != nullptr) {
```
JackCaoG, line 716: we need a test for this. For example when we deep copy a tensor with sharding, the result tensor should also have sharding. Something similar to https://github.com/pytorch/xla/blob/b3f79ccf1c2e85f76c737eeeab0e4ca512d180b4/test/test_operations.py#L1670 

@steventk-g can you add a test case?

torch_xla/csrc/tensor.cpp
```diff
@@ -1788,7 +1837,7 @@ std::shared_ptr<XLATensor::Async> XLATensor::SyncTensorsGraphInternal(
   TF_VLOG(5) << "TensorsGraphSize=" << compile_result.emitted_nodes;
 
   auto cached_computation = std::make_shared<CachedComputation>(
-      std::move(compile_result.computation));
+      std::move(compile_result.computation), compile_result.is_sharded);
```
JackCaoG, line 1840: why do we need `is_sharded` separatelly in `CachedComputation`?

torch_xla/csrc/tensor.cpp
JackCaoG: I am confuse, `ExecuteReplicated` returns a ` virtual std::vector<std::vector<DataPtr>>` according to 
https://github.com/pytorch/xla/blob/3db9008f92a203c202299355d13b9f2c8bba2175/third_party/xla_client/computation_client.h#L263-L267

but you are assign it to a `std::vector<xla::ComputationClient::DataPtr>`. How does that works?

JackCaoG (commented):
Mostly LGTM, I had a question regarding `ExecuteReplicated`. If we can align on that this pr is ready to merge.

torch_xla/csrc/tensor.cpp
steventk-g: Done https://github.com/pytorch/xla/issues/4095

torch_xla/csrc/tensor.cpp
yeounoh: It takes the first (arbitrary, since it's replicated) element of the nested vector output.

JackCaoG: oh ok..

JackCaoG: can you add a log? It is OK to do that in a separate pr.

JackCaoG (approved):
Thanks @yeounoh ! I will merge this pr to unblock @steventk-g

test/cpp/test_xla_sharding.cpp
steventk-g: This task is blocked because there is no `Array5D` class in XLA. It only goes up to 4D right now.

JackCaoG: @yeounoh do you know if we really need 5d or it is a good to have?

third_party/xla_client/computation_client.h
yeounoh: Done

torch_xla/csrc/tensor.cpp
yeounoh: I see, I normally prefer more explicit type identifiers especially for casting (similar to avoid using auto too much).

torch_xla/csrc/tensor.cpp
yeounoh: Good point, @steventk-g let me handle this if you haven't already started.

torch_xla/csrc/tensor.cpp
yeounoh: We could pass around is_sharded between APIs, or wrap it inside the CachedComputation. Is sharded is later needed for the execution (will be associated with the cached computation only), and the latter doesn't require changing the function APIs here and there.

test/test_xla_sharding.py
yeounoh: We only run cpp tests -- covers the internal changes that affects the non-spmd code paths -- and the python API tests are disabled [link](https://github.com/pytorch/xla/blob/e0dd3e08fe393fc64f769a1fc30861e7e793b019/test/run_tests.sh#L130). I will re-enable it after debugging/ adding the API unit tests.

test/cpp/test_xla_sharding.cpp
yeounoh: it's a good to have. The ask was up to 3d.

yeounoh: @steventk-g I think we could use the generic Array to represent the 5d case, but a low prio for now. Let me look into it later.

PR #3687, opened 2022-07-06
Create OP_CODEGEN_GUIDE.md

wonjoolee95:
Preview: https://github.com/pytorch/xla/blob/cb73ec34ea53f443e3516e6159692adb2a3e7bc9/CODEGEN_MIGRATION_GUIDE.md

---

Migrate the internal codegen guide by @JackCaoG to README. Re-worded and re-structured the README so it can be shared publicly.

### DIFF
--- a/OP_CODEGEN_GUIDE.md
+++ b/OP_CODEGEN_GUIDE.md
@@ -74,7 +74,7 @@ at::Tensor XLANativeFunctions::abs(const at::Tensor& self) {
 ```
 
 ### 2. Codegen the op and inspect the generated file
-Find the op in  `xla/xla_native_functions.yaml` and move it to the full_codegen column and run the build again. The build will fail (reason explained later in this guide) but you can still see the generated file. The code snippets below uses `abs` as an example.
+Find the op in  `xla/xla_native_functions.yaml` and move it to the full_codegen column and run `python setup.py install` under xla directory again. The build will fail (reason explained later in this guide) but you can still see the generated file. The code snippets below uses `abs` as an example.
 #### XLANativeFunctions.cpp
 ```
 at::Tensor XLANativeFunctions::abs(const at::Tensor & self) {
@@ -206,4 +206,7 @@ to
 ```
 
 ## Run the test and verify the result
-Run the C++ op test or a simple test that only involves the generated ops. As usual, two things to verify are the correctness and the xla counter being incremented correctly.
+Run the C++ op test or a simple test that only involves the generated ops. 
+- 
+
+As usual, two things to verify are the correctness and the xla counter being incremented correctly.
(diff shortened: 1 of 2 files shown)


OP_CODEGEN_GUIDE.md
```diff
@@ -0,0 +1,209 @@
...
+An good example of a "simple" op would be something like `abs`:
+```
+at::Tensor XLANativeFunctions::abs(const at::Tensor& self) {
+  XLA_FN_COUNTER("xla::");
+  return bridge::AtenFromXlaTensor(XLATensor::abs(bridge::GetXlaTensor(self)));
+}
+```
+
+### 2. Codegen the op and inspect the generated file
+Find the op in  `xla/xla_native_functions.yaml` and move it to the full_codegen column and run the build again. The build will fail (reason explained later in this guide) but you can still see the generated file. The code snippets below uses `abs` as an example.
```
JackCaoG, line 77: instead of `run the build again`, maybe make it more concrate like run `python setup.py install under xla dir`

OP_CODEGEN_GUIDE.md
```diff
@@ -0,0 +1,209 @@
...
+```
+to
+```
+  torch::lazy::NodePtr exp =
+      Pow(torch::lazy::MakeNode<Abs>(input, std::vector<torch::lazy::Shape>()),
+          norm_exp);
+```
+
+## Run the test and verify the result
+Run the C++ op test or a simple test that only involves the generated ops. As usual, two things to verify are the correctness and the xla counter being incremented correctly.
```
JackCaoG, line 209: Please provide an example of how to run c++ test.

JackCaoG (commented):
Thanks @wonjoolee95 ! Mostly lgtm

JackCaoG:
also rebase should fix the test issue

JackCaoG:
It is also more of a `CODE_GEN_MIGRATION`  guide since it only covers how to convert an existing op to use codegen. If we want to call it `CODEGEN_GUIDE`, we want to add a section at the end to explain how to start from scratch(which should be pretty similar to what we already have in this doc)

OP_CODEGEN_GUIDE.md
wonjoolee95: Updated

wonjoolee95:
> It is also more of a `CODE_GEN_MIGRATION` guide since it only covers how to convert an existing op to use codegen. If we want to call it `CODEGEN_GUIDE`, we want to add a section at the end to explain how to start from scratch(which should be pretty similar to what we already have in this doc)

Makes sense, updated to `CODEGEN_MIGRATION_GUIDE` since we are focused on the migration for now.

JackCaoG (approved):
Thanks!

JackCaoG:
@wonjoolee95 one thing I forget mentioned is can you add a couple example pr to this doc?

wonjoolee95:
> @wonjoolee95 one thing I forget mentioned is can you add a couple example pr to this doc?

Sure thing! Will follow-up with a PR.

PR #3689, opened 2022-07-06
Update TestCatTypePromotion test to remove tests for torch::kFloat and torch::kDouble

wonjoolee95:
Update TestCatTypePromotion test to remove test case for torch::kHalf

---

Related to the recent PR to https://github.com/pytorch/xla/pull/3608, which adds type promotion for cat op. Since TPU does not support f16, remove test case for torch::kHalf.


JackCaoG:
hmm I am a bit confuse. f16 is not supported on TPU but why do we remove test case for f32 and f64?

wonjoolee95:
> hmm I am a bit confuse. f16 is not supported on TPU but why do we remove test case for f32 and f64?

Updated the commit to properly remove the f16 `torch::kHalf` dtype test case.

JackCaoG (approved):
Thanks @wonjoolee95 !

JackCaoG:
Feel free to merge after test passed

PR #3696, opened 2022-07-08
Implement ir builder

JackCaoG:


JackCaoG:
@wonjoolee95 I think this one is ready for review. GPU failure is something else.

torch_xla/csrc/ir_builder.h
```diff
@@ -30,27 +33,31 @@ struct XLAIrBuilder : IrBuilder {
   torch::lazy::NodePtr MakeView(
       const torch::lazy::Value& input0,
       const std::vector<int64_t>& output_size) const override {
-    return torch::lazy::MakeNode<ViewOp>(input0, output_size);
+    // TODO(JackCAoG): use functionization pass instead
```
wonjoolee95, line 36: nit: why are we changing this to functionization pass?

wonjoolee95 (approved):
Thanks!

torch_xla/csrc/ir_builder.h
JackCaoG: so after functionization pass, we no longer see the view operator, all of the view op becomes a new `view_copy` op. The complicated part of view is that they should share the same memory space. This is not possible for pytorch/xla right now because XLA does not expose its memory space to user. What we did is we create two node and whenever an in place operaton being done on one of the op, we replay that operation to the other node(after reshape).

after functionization pass, we no longer need to maintain this relationship  in pytorch/xla layer. We can treat view op like any other op. PyTorch core will handle to apply the same in_place operation to the Node that are views of each other.

PR #3558, opened 2022-05-10
dynamic expand with symint implementation POC

miladm:
This is the POC implementation of `torch.Tensor.expand` op based on the PyTorch `SymInt` POC implementation [PR](https://github.com/pytorch/pytorch/pull/75759).

PR #3702, opened 2022-07-12
Dynamic `expand` with `SymInt` implementation

miladm:
Rebased version of issue #3558.


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2310,11 +2331,17 @@ std::tuple<at::Tensor, at::Tensor> XLANativeFunctions::nll_loss_forward(
 at::Tensor XLANativeFunctions::nonzero(const at::Tensor& self) {
   XLA_FN_COUNTER("xla::");
   XLATensorPtr self_tensor = bridge::GetXlaTensor(self);
-  // Initially make XLA handled nonzero() handling experimental, and opt-in.
-  if (!DebugUtil::ExperimentEnabled("nonzero")) {
-    return at::native::call_fallback_fn<&xla_cpu_fallback,
-                                        ATEN_OP(nonzero)>::call(self);
-  }
+  /*
```
miladm, line 2334: To avoid breaking pt/xla for users without dynamic shape needs, I plan to uncomment this block after getting LGTM from reviewers. I continue to rely on the `experimental` flags for a little longer.

We should discuss the process to remove this code soon.

miladm: Will looking adopting a temp approach like `LTC_ENABLE_DYNAMIC_SHAPES`.

Issue #3692, opened 2022-07-07
allowing `xm.get_ordinal()` and default device in `xm.xla_device()` in PJRT


ronghanghu:
## 🚀 Feature

**Context**: The new PRJT runtime provides many advantages over the XRT runtime and is a really great new feature in PyTorch/XLA. On the other hand, the current [PJRT MNIST](https://github.com/pytorch/xla/blob/master/test/pjrt/test_train_pjrt_mnist.py) and [PJRT ImageNet](https://github.com/pytorch/xla/blob/master/test/pjrt/test_train_pjrt_imagenet.py) examples require
* manually computing the rank (from `CLOUD_TPU_TASK_ID` and `index`) instead of directly getting the rank using `xm.get_ordinal()`, and
* manually specifying the thread index in `xm.xla_device(index)` instead of directly using `xm.xla_device()`.

These differences will likely break a lot of existing user code that directly uses `xm.get_ordinal()` and `xm.xla_device()` for rank and XLA devices (e.g. the [FSDP implementation](https://github.com/pytorch/xla/blob/cc19c3abcbb3f702d5f468ee08549edd926ef549/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py#L290-L291) and the [MoCo v3 implementation](https://github.com/ronghanghu/moco_v3_tpu/blob/global_avg_pool/distributed.py#L35) and many other codebases) and introduces bugs and overheads in transiting to PJRT.

**Proposed new feature**:

For compatibility with existing code, it would be great to allow directly getting the rank via `xm.get_ordinal()` (which should return a value between 0 and `xm.get_world_size()`) and directly setting the XLA device via `xm.xla_device()` when using PJRT.

This can be implemented perhaps by storing the index as a thread-local variable and reading it in `xm.get_ordinal()` and `xm.xla_device()` when using PJRT.

Also, a related bug fix to the PJRT MNIST (and ImageNet) example is that we should add `* 2` to `CLOUD_TPU_TASK_ID` when computing rank on TPU v3, as mentioned in https://github.com/pytorch/xla/pull/3607#discussion_r915347367.

---

Also, below are a few of my throughs and discussions related to PJRT.

#### Regarding random ops in PyTorch (and e.g. NumPy)

[...]

PR #3704, opened 2022-07-12
Add `torch_xla.experimental.pjrt`

will-cromar:
- Move functions from `test/pjrt/common.py` into `torch.experimental.pjrt`, improve type annotations
- Add functions to set local and global ordinal for a thread
- Return correct local and global ordinal from `xla_model` (fixes internal calls to `get_ordinal()` and `get_local_ordinal()`, see #3692)
- Remove manually computed `index` and `rank` from PJRT examples
- Add TPU topology variables to `xenv`
- Create unit tests and add to CI

TODO in future PR: Add multiprocess tests for this module that use CPU and TPU


will-cromar:
Tested manually on TPU v3 and v4

ronghanghu:
Thanks for this implementation!

A quick question: I wonder would the `get_ordinal()` and `get_local_ordinal()` in this PR also work on a TPU pod? For example, when running PJRT on v3-32 pod, does `get_ordinal()` return a global rank in 0, ..., 31 and `get_local_ordinal()` return a local rank on each TPU VM in 0, ..., 8?

will-cromar:
`get_ordinal()` won't work on pod yet, but it will. Right now `configure_tpu_topology` assumes a single host. Setting up the topology variables on pods will involve querying the TPU metadata and setting `TPU_PROCESS_ADDRESSES` with the correct IPs, setting `TPU_PROCESS_BOUNDS` to the correct shape, and finding the right `CLOUD_TPU_TASK_ID` for each process (which will actually be the same as ordinal on TPU v4). I experimented with setting these variables by hand and didn't run up against any hard blockers. I'm trying to get an example with DDP working right now, but after that and once this PR is merged, my next priority will be pod support.

I think I misunderstood what `get_local_ordinal` is supposed to do. From the docstring: `Retrieves the replication local ordinal of the current process. The local ordinals range from 0 to the number of local devices minus 1.` So, my understanding is that the local ordinal would be the index of the current device among currently visible devices. Don't each of the worker processes in XRT have only one visible TPU device? Looking at our automated test logs from XRT, each process prints something like (`Device=xla:0/7`), i.e. each process is using `xla:0`. Based on my understanding of the docstring, then the local ordinal in each process would be `0` and not `0, ..., 8`. Is the local ordinal actually supposed to be the ordinal of the current process within a host?

CC @JackCaoG

ronghanghu:
> I'm trying to get an example with DDP working right now, but after that and once this PR is merged, my next priority will be pod support.

Thanks! Looking forward to it as this is ultimately what we need in our use cases.


> Is the local ordinal actually supposed to be the ordinal of the current process within a host?

In XRT this is the process index within a host (so ranging from 0, ..., 7 on TPU v3) since this is usually what we need when we set up distributed training details and check whether a process is a master ordinal on a TPU VM host as in https://github.com/pytorch/xla/blob/master/torch_xla/core/xla_model.py#L200-L204. Also, I guess an API that always returns 0 is not very helpful :) So I believe it would be great to preserve this behavior of `get_local_ordinal` in PJRT as well.

will-cromar:
Thanks for clarifying. I'll update `get_local_ordinal` to match XRT.

So, as I understand it, we need to keep track of four indices: the process rank (corresponds to `rank` in `torch.distributed` and `CLOUD_TPU_TASK_ID`), the global ordinal (not the same as process rank for TPU v3, since there will be 2 devices per process), the local ordinal, and the device index (i.e. what I thought local ordinal was; will always be 0 only for TPU v4, but can be 0 or 1 for TPU v3).

To add an additional layer of confusion, `ComputationClient::GetLocalDevices` returns the devices visible to a process, while `get_local_ordinal` returns the ordinal in a host (not process) context. PjRt uses "addressable" rather than "local" to refer to the devices available to a process. I will try to make the same distinction for clarity.

ronghanghu:
Yeah, I see this is a little confusing in PJRT. I guess the high-level goal here should be to make threads in PJRT mimic what processes do in XRT at the user interface level, so that people can port their XRT code to PJRT hopefully without modification.

So a "rank" in PJRT (as what a user sees) should be the global ordinal (i.e. a rank among all the threads as opposed to the processes). But since this is different from how the PJRT works under the hood, I guess one needs to keep track of all these things four indices above

will-cromar:
For sure, we want to keep `xm.*` as compatible as possible with XRT to give users the shortest path for an initial migration.

Luckily the story is simpler for TPU v4. Device index will always be 0 and the global device ordinal will match the CLOUD_TPU_TASK_ID and process rank, leaving us only having to keep track of two indices. Unfortunately, TPU v3 breaks the assumption of 1 process per device and adds additional complexity.

torch_xla/experimental/pjrt.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def num_visible_tpu_chips(default: int = 4) -> int:
+  """Returns number of TPU chips visible to current process."""
+  visible_devices = xu.getenv_as(xenv.TPU_VISIBLE_DEVICES, str)
+
+  return len(visible_devices.split(',')) if visible_devices else default
+
+
+def configure_tpu_topology(rank: int, processes: int, base_port=8476) -> None:
```
JackCaoG, line 47: how is the single process PJRT example looks like? Would user needs to set any of below env vars?

JackCaoG (commented):
Thanks @will-cromar !

Is it fair to say that to migrate from XRT to PJRT on multi-process + single-host, they only need to change the spawn part(and passing state dict) and config? Any where else user needs to change the model code?

test/pjrt/test_experimental_pjrt.py
```diff
@@ -0,0 +1,88 @@
...
+  def test_num_local_devices(self):
+    self.assertLen(xm.get_xla_supported_devices(),
+                   pjrt.addressable_device_count())
+
+  def test_num_global_devices(self):
+    self.assertLen(torch_xla._XLAC._xla_get_all_devices(),
+                   pjrt.global_device_count())
+
+  def test_world_size(self):
+    self.assertEqual(xm.xrt_world_size(), pjrt.global_device_count())
```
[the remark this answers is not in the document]
will-cromar: `xrt_world_size` and `len(num_global_devices)` will be the same.

I'm not entirely clear if, when we're using DDP with TPU v3, we need to call `init_process_group` 4 or 8 times. I suspect it will be 4 because it sets the global default process group. If that's true, then the world size in the `torch.dist` sense will be different from the number of devices.

torch_xla/experimental/pjrt.py
will-cromar: Yes. If the user sets `TPU_VISIBLE_DEVICES=[0 .. 3]` and `TPU_PROCESS_BOUNDS=1,1,1`, the other variables will be inferred.

torch_xla/experimental/pjrt.py
```diff
@@ -0,0 +1,216 @@
...
+import torch_xla.core.xla_model as xm
+import torch_xla.utils.utils as xu
+
+_PJRT_ORDINALS = threading.local()
+
+R = TypeVar('R')
+FN = TypeVar('FN')
+
+
+def set_device_type(pjrt_device: str) -> None:
```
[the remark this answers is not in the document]
will-cromar: I haven't used it anywhere else yet. Later, I want to add a check here to ensure the client has not been initialized when this is called and surface an error.

test/pjrt/test_experimental_pjrt.py
```diff
@@ -0,0 +1,88 @@
...
+      pjrt.xla_device()
+
+  def test_default_ordinals(self):
+    global_ordinal = xm.get_ordinal()
+    self.assertEqual(global_ordinal, 0)
+
+    local_ordinal = xm.get_local_ordinal()
+    self.assertEqual(local_ordinal, 0)
+
+  def test_num_local_devices(self):
```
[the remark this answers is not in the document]
will-cromar: `get_supported_devices()` returns the set of devices that are visible to the current process. Each thread in the process can "see" all of the addressable devices because they share the same client.

torch_xla/experimental/pjrt.py
will-cromar: Yes, you can still just set `PJRT_DEVICE=TPU` and use one core, but the client will be initialized with 8 devices, preventing any other process from using them.

torch_xla/core/xla_model.py
```diff
@@ -172,6 +176,9 @@ def get_ordinal(defval=0):
   Returns:
     The replication ordinal of the current process.
   """
+  if pjrt.using_pjrt():
```
JackCaoG, line 179: nit, should we update the docString to this function? I think the PJRT case we want to replication ordinal of the current thread? Not sure if here is the right place to explain the PJRT + v3 complication. same for the get_local_ordinal.

will-cromar: Good catch, updated these docstrings.

JackCaoG (approved):
Thanks!

JackCaoG:
@will-cromar Rebase should solve your problem.

test/pjrt/test_experimental_pjrt.py
```diff
@@ -0,0 +1,88 @@
...
+    self.assertLen(torch_xla._XLAC._xla_get_all_devices(),
+                   pjrt.global_device_count())
+
+  def test_world_size(self):
+    self.assertEqual(xm.xrt_world_size(), pjrt.global_device_count())
+
+  @parameterized.named_parameters(('single_thread', [0], [0]),
+                                  ('1_host_x_3_threads', [1, 1, 1], [2, 2, 2]),
+                                  ('3_hosts_x_1_thread', [0, 0, 0], [0, 1, 2]))
+  def test_set_ordinals(self, local_ordinals, global_ordinals):
```
yeounoh, line 55: Nit. Maybe adding a comment describing the test setup and what we are trying to setup might be healpful for readability.

test/pjrt/test_experimental_pjrt.py
```diff
@@ -0,0 +1,88 @@
...
+    num_threads = len(local_ordinals)
+    self.assertLen(global_ordinals, num_threads)
```
yeounoh, line 57: Would this be true only in a single host case?

test/pjrt/test_experimental_pjrt.py
```diff
@@ -0,0 +1,88 @@
...
+    with concurrent.futures.ThreadPoolExecutor(max_workers=num_threads) as e:
+      results = e.map(_thread_fn, local_ordinals, global_ordinals)
+      for result, local_ordinal, global_ordinal in zip(results, local_ordinals,
+                                                       global_ordinals):
+        expected = (local_ordinal, global_ordinal)
+        self.assertEqual(result, expected)
+
+  def test_xla_device_default(self):
+    device = xm.xla_device()
+    self.assertEqual(device, torch.device('xla:0'))
```
yeounoh, line 76: Could be redundant, but adding another statement asserting that the default device is not equal to 'xla:1' could make it more complete -- would be even better we had a parameterized option to change the default device, too.

test/pjrt/test_experimental_pjrt.py
```diff
@@ -0,0 +1,88 @@
...
+
+  def test_xla_device_error(self):
+    with self.assertRaises(IndexError):
+      xm.xla_device(10)
```
yeounoh, line 80: Question, xm.xla_device is indexing into the global ordinal?

torch_xla/core/xla_env_vars.py
```diff
@@ -20,3 +20,9 @@
 TPU_CHIPS_PER_HOST_BOUNDS = 'TPU_CHIPS_PER_HOST_BOUNDS'
 TPU_MESH_CTLER_ADDR = 'TPU_MESH_CONTROLLER_ADDRESS'
 TPU_MESH_CTLER_PORT = 'TPU_MESH_CONTROLLER_PORT'
+PJRT_DEVICE = 'PJRT_DEVICE'
```
yeounoh, line 23: We should start adding comments describing the xla env vars -- maybe in a separate clean-up PR in the future.

torch_xla/csrc/ir_builder.h
```diff
@@ -1,18 +1,21 @@
...
 #include "torch_xla/csrc/ops/diagonal.h"
 #include "torch_xla/csrc/ops/expand.h"
 #include "torch_xla/csrc/ops/generic.h"
 #include "torch_xla/csrc/ops/ops.h"
+#include "torch_xla/csrc/tensor_util.h"
 
 namespace torch_xla {
 
-struct XLAIrBuilder : IrBuilder {
+struct XLAIrBuilder : torch::lazy::IrBuilder {
```
yeounoh, line 16: We should rebase after landing this LTC change?

torch_xla/experimental/pjrt.py
```diff
@@ -0,0 +1,216 @@
...
+  Returns:
+    A `torch.device` representing an XLA device.
+  """
+  devices = xm.get_xla_supported_devices(devkind=devkind)
+  device_index = n or (local_ordinal(default=0) % addressable_device_count())
+  if device_index > len(devices):
+    raise IndexError('Device index {} out of range in {}'.format(
+        device_index, devices))
+
+  torch_xla._XLAC._xla_set_default_device(devices[device_index])
```
yeounoh, line 125: Nice

yeounoh (approved):
LGTM -- left some minor comments and questions. Also rebasing would help us get rid of some changes from earlier PRs.

torch_xla/csrc/ir_builder.h
will-cromar: Oops, my bad. Rebased this branch.

torch_xla/core/xla_env_vars.py
will-cromar: Other than `PJRT_DEVICE`, the rest of these topology variables apply to other TPU VM generally. I'll see if we can add a description to the Cloud TPU documentation and link it here.

test/pjrt/test_experimental_pjrt.py
will-cromar: Not in general. It indexes into the _i_th addressable device, which does not necessarily correspond to the global or even local ordinal.

See my comment here: https://github.com/pytorch/xla/pull/3704#issuecomment-1187712330

test/pjrt/test_experimental_pjrt.py
will-cromar: Good idea. Added a docstring. Also one of these test cases didn't make any sense, so I fixed it.

test/pjrt/test_experimental_pjrt.py
will-cromar: I'm basically just checking that the test inputs have equal length here. Rewrote this line to be more explicit.

test/pjrt/test_experimental_pjrt.py
will-cromar: I think checking equality to `xla:0` is sufficient here. Would `torch.device('xla:1')` throw an error if there was only one device?

Added a TODO to rewrite this test with multiple devices. I like the idea of using a parameterized test for this.

yeounoh: ^ yea, i think that would resonate with the challenge you had to solve with the default device to some extent.

Issue #3638, opened 2022-06-08
gradient checkpoint cause bigger memory usage on GPU


cicirori:
## ❓ Questions and Help

Recently I started testing GC performance on the GPU on the master version of pytorch and torch xla.

Unfortunately in consistent with my previous conclusions(https://github.com/pytorch/xla/issues/3455#issuecomment-1101056839), the current torch xla GC still seems have difficulty achieving  the desired results, even with memory gains on some very simple cases.

Simple test program:

```
import argparse
import torch
import torch_xla.utils.checkpoint
import torch.utils.checkpoint
import torch_xla.core.xla_model as xm
import torch_xla.debug.metrics as met


def run(grad_checkpoint, use_cuda=False):
    if use_cuda:
        device = 'cuda'
    else:
        device = xm.xla_device()

    model = torch.nn.ModuleList(
        [
            torch.nn.Sequential(
                torch.nn.Conv2d(1024, 1024, 1),
                torch.nn.ReLU(),
                torch.nn.Conv2d(1024, 1024, 1),
                torch.nn.ReLU(),
            )
            for _ in range(64)
        ]
    ).to(device)
    optimizer = torch.optim.SGD(model.parameters(), lr=0.0)

    for step in range(200):
        dummy_data = torch.zeros(64, 1024, 14, 14, device=device)
        optimizer.zero_grad()
        x = dummy_data
        for n_l, layer in enumerate(model):
            if n_l > 0 and grad_checkpoint:
                if use_cuda:
                    x = torch.utils.checkpoint.checkpoint(layer, x)
                else:
                    x = torch_xla.utils.checkpoint.checkpoint(layer, x)
            else:
                x = layer(x)
        dummy_loss = x.sum()
        dummy_loss.backward()
        optimizer.step()
        if not use_cuda:
            xm.mark_step()
            mem_info = xm.get_memory_info(device)
            mem_info['kb_used'] = mem_info['kb_total'] - mem_info['kb_free']
            print(f"step {step}, memory = {mem_info['kb_used']}")
        else:
            print(f"step {step}, memory = {torch.cuda.memory_summary()}")
    if not use_cuda:
        print(met.metrics_report())


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--grad_checkpoint", type=int, required=True)
    parser.add_argument("--use_cuda", type=int, required=True)
    args = parser.parse_args()
    run(args.grad_checkpoint, args.use_cuda)
```

[...]

PR #3721, opened 2022-07-16
Apply Optimization Barrier to grad input and saved input

JackCaoG:
Prior to this change, gc was applied to input and output of the forward. Which will cause the problem described in https://github.com/pytorch/xla/issues/3638#issuecomment-1184976269.

Using test script with 1.12 pytorch/xla wheel
```
import argparse
import torch
import torch_xla.utils.checkpoint
import torch.utils.checkpoint


def run(grad_checkpoint):
    device = xm.xla_device()
    model = torch.nn.ModuleList(
        [
            torch.nn.Sequential(
                torch.nn.Conv2d(1024, 1024, 1),
                torch.nn.ReLU(),
                torch.nn.Conv2d(1024, 1024, 1),
                torch.nn.ReLU(),
            )
            for _ in range(64)
        ]
    ).to(device)
    optimizer = torch.optim.SGD(model.parameters(), lr=0.0)

    for step in range(200):
        dummy_data = torch.zeros(64, 1024, 14, 14, device=device)
        optimizer.zero_grad()
        x = dummy_data
        for n_l, layer in enumerate(model):
            if n_l > 0 and grad_checkpoint:
                x = torch_xla.utils.checkpoint.checkpoint(layer, x)
            else:
                x = layer(x)
        dummy_loss = x.sum()
        dummy_loss.backward()
        optimizer.step()
        xm.mark_step()
        print(f"step {step}, free memory = {xm.get_memory_info(device)['kb_free']}")

    print(met.metrics_report())


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--grad_checkpoint", type=int, required=True)
    args = parser.parse_args()
    run(args.grad_checkpoint)
```

without gc
```
step 197, free memory = 11514304
step 198, free memory = 11514304
step 199, free memory = 11514304
Metric: ExecuteTime
  TotalSamples: 199
  Accumulator: 41s307ms151.043us
  ValueRate: 796ms539.909us / second
  Rate: 3.83257 / second
  Percentiles: 1%=205ms104.269us; 5%=205ms230.546us; 10%=205ms348.668us; 20%=205ms477.548us; 50%=206ms688.449us; 80%=206ms063.679us; 90%=206ms474.142us; 95%=227ms397.762us; 99%=257ms511.905us
```

with old gc
```
step 197, free memory = 11288560
step 198, free memory = 11288560
step 199, free memory = 11288560
Metric: ExecuteTime
  TotalSamples: 199
  Accumulator: 54s516ms401.018us
  ValueRate: 833ms287.457us / second
  Rate: 3.09857 / second
  Percentiles: 1%=267ms587.570us; 5%=267ms809.318us; 10%=267ms870.413us; 20%=267ms004.902us; 50%=267ms260.097us; 80%=268ms556.363us; 90%=269ms010.364us; 95%=284ms677.477us; 99%=316ms706.940us
```

with this new gc (1.12 wheel + this pr)
```
step 197, free memory = 13003104
step 198, free memory = 13003104
step 199, free memory = 13003104
Metric: ExecuteTime
  TotalSamples: 199
  Accumulator: 56s391ms471.602us
  ValueRate: 837ms002.569us / second
  Rate: 2.9537 / second
  Percentiles: 1%=280ms340.983us; 5%=281ms885.300us; 10%=281ms958.868us; 20%=281ms135.443us; 50%=281ms439.287us; 80%=282ms822.307us; 90%=282ms277.625us; 95%=303ms060.574us; 99%=337ms030.345us
```

### DIFF
--- a/torch_xla/utils/checkpoint.py
+++ b/torch_xla/utils/checkpoint.py
@@ -8,6 +8,18 @@ from torch.utils.checkpoint import detach_variable, check_backward_validity, get
 from typing import Any, Iterable, List, Tuple, Union
 
 
+def extract_tensors_from_list(inputs):
+  tensor_inputs = []
+  if torch.is_tensor(inputs):
+    tensor_inputs.append(inputs)
+  # tensor is Iterable so we need to avoid iterating through tensor
+  elif isinstance(inputs, Iterable):
+    for input in inputs:
+      if torch.is_tensor(input):
+        tensor_inputs.append(input)
+  return tensor_inputs
+
+
 class CheckpointFunction(torch.autograd.Function):
 
   @staticmethod
@@ -51,18 +63,10 @@ class CheckpointFunction(torch.autograd.Function):
       else:
         ctx.inputs.append(arg)
 
+    ctx.save_for_backward(*tensor_inputs)
+
     with torch.no_grad():
       outputs = run_function(*args)
-    if torch.is_tensor(outputs):
-      tensor_outputs.append(outputs)
-    # tensor is Iterable so we need to avoid iterating through tensor
-    elif isinstance(outputs, Iterable):
-      for output in outputs:
-        if torch.is_tensor(output):
-          tensor_outputs.append(output)
-
-    xm.optimization_barrier_(tensor_inputs + tensor_outputs)
-    ctx.save_for_backward(*tensor_inputs)
 
     return outputs
 
@@ -88,6 +92,7 @@ class CheckpointFunction(torch.autograd.Function):
     rng_devices = []
     if ctx.preserve_rng_state and ctx.had_cuda_in_fwd:
       rng_devices = ctx.fwd_gpu_devices
+    xm.optimization_barrier_(extract_tensors_from_list(inputs + list(args)))
     with torch.random.fork_rng(
         devices=rng_devices, enabled=ctx.preserve_rng_state):
       if ctx.preserve_rng_state:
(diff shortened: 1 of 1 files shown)


JackCaoG:
It is somewhat confusing that with the old gc, memory usage actually increased(which was not the case when I first implemented it, I think it happened after a libtpu update). @ronghanghu I think you mentioned you observed the same behavior but old gc can still reduce the peak memory usage and helped you to scale up your model. I think old gc still takes affect because execution time increased as expected.

For what's worth, new implementation of the gc seems better.

ronghanghu:
I tested this PR and it worked well for me. I tried a few of our internal models on TPUs (they are quite large models with 5B to 20B parameters and cannot run without a valid gradient checkpointing implementation).

With this PR, the performance (speed, memory usage returned by `xm.get_memory_info`, etc) is **exactly the same as before**, so I think both this new and the previously existing gradient checkpointing implementation are valid for our use cases. I guess previously the XLA compiler didn't go too far against us on TPUs so the optimization barrier on the forward-pass inputs and outputs was sufficient there.

JackCaoG:
Oh OK. Thanks @ronghanghu for testing this out. I think what we have in this pr is a better approach so I will go with it if this does not break your existing use case.

torch_xla/utils/checkpoint.py
```diff
@@ -8,6 +8,18 @@
 from typing import Any, Iterable, List, Tuple, Union
 
 
+def extract_tensors_from_list(inputs):
```
yeounoh, line 11: nit, would it make sense to move this inside the CheckpointFunction class? We don't have other use cases, yet -- and the function name is quite general.

JackCaoG: oh sure, will do.

yeounoh (approved):
LGTM

PR #3735, opened 2022-07-20
Add `dynamic_ir`

miladm:
A key building block for supporting dynamic ops is `dynamic_ir`. This module inherits from LTC core lazy dimensions to create IR nodes for tensor `size` dimensions.

### DIFF
--- a/torch_xla/csrc/ops/dynamic_ir.cpp
+++ b/torch_xla/csrc/ops/dynamic_ir.cpp
@@ -23,12 +23,11 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
 }
 
 int64_t SizeNode::getStaticValue() const {
-  return dynamic_cast<const XlaNode*>(operand(0).node)->shape(0).size(dim_);
+  return operand(0).node->shape(0).size(dim_);
 }
 
 bool SizeNode::isDynamic() const {
-  auto symbolic_vec =
-      dynamic_cast<const XlaNode*>(operand(0).node)->shape(0).is_symbolic();
+  auto symbolic_vec = operand(0).node->shape(0).is_symbolic();
   if (!symbolic_vec.has_value()) {
     return true;
   }
--- a/torch_xla/csrc/ops/dynamic_ir.cpp
+++ b/torch_xla/csrc/ops/dynamic_ir.cpp
@@ -1,0 +1,135 @@
+#include "torch_xla/csrc/ops/dynamic_ir.h"
+
+#include "absl/strings/str_join.h"
+#include "tensorflow/compiler/xla/xla_client/debug_macros.h"
+#include "torch_xla/csrc/lowering_context.h"
+#include "torch_xla/csrc/ops/infer_output_shape.h"
+
+static const torch::lazy::DimensionNode* DimCast(torch::lazy::Output output) {
+  return dynamic_cast<const torch::lazy::DimensionNode*>(output.node);
+}
+
+namespace torch_xla {
+
+SizeNode::SizeNode(torch::lazy::Value input, size_t dim)
+    : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::size")},
+              {input}, [&]() {  if (input.node->shapes().size() > 0) {
+                                  torch::lazy::Shape sh_ = input.node->shape(0);
+                                  torch::lazy::Shape sh__ = sh_.with_symbolic_dims(sh_.is_symbolic());
+                                  std::cout << "sh_ vs. sh__" << sh_.is_symbolic().value()[0] << " " << sh__.is_symbolic().value()[0] << std::endl;
+                                  std::cout << "sh_ vs. sh__" << sh_.is_symbolic().value()[1] << " " << sh__.is_symbolic().value()[1] << std::endl;
+                                  return sh__;
+                                } else {
+                                  return torch::lazy::Shape();
+                                }
+                             },
+                  [&]() { return xla::ShapeUtil::MakeShape(xla::S32, {}); }, 1,
+              torch::lazy::MHash(dim)),
+      dim_(dim){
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/ops/dynamic_ir.h
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+// Represents the result of calling `size` on a Tensor
+class SizeNode : public XlaNode, public torch::lazy::DimensionNode {
+ public:
+  SizeNode(torch::lazy::Value input, size_t dim);
+  int64_t getStaticValue() const override;
+  bool isDynamic() const override;
+  std::string ToString() const override;
+  size_t dim_ = 0;
```
JackCaoG, line 44: any reason `dim_` is not private?

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -0,0 +1,118 @@
...
+              torch::lazy::MHash(dim)),
+      dim_(dim){};
+
+XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
+  auto input = loctx->GetOutputOp(operand(0));
+  return ReturnOp(xla::GetDimensionSize(input, this->dim_), loctx);
+}
+
+int64_t SizeNode::getStaticValue() const {
+  return dynamic_cast<const XlaNode*>(operand(0).node)->shape(0).size(dim_);
```
JackCaoG, line 26: `shape` will return a `torch::lazy::shape` I think? If that's case why not just leave it as `torch::lazy::Node`, `shape()` also existed on upstream base class.

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+bool SizeNode::isDynamic() const {
+  auto symbolic_vec =
+      dynamic_cast<const XlaNode*>(operand(0).node)->shape(0).is_symbolic();
```
JackCaoG, line 31: ditto

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -0,0 +1,118 @@
...
+    return true;
+  }
+  return symbolic_vec->at(dim_);
+}
+
+std::string SizeNode::ToString() const { return "SizeNode"; }
+
+SizeAdd::SizeAdd(torch::lazy::Value a, torch::lazy::Value b)
+    : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::add")},
+              {a, b}, xla::ShapeUtil::MakeShape(xla::S32, {}), 1){};
```
JackCaoG, line 42: shouldn't the shape of `SizeAdd` to be input dependent?

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -0,0 +1,118 @@
...
+  return DimCast(operand(0))->isDynamic() || DimCast(operand(1))->isDynamic();
+}
+
+std::string SizeAdd::ToString() const { return "SizeAdd"; }
+
+XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
+  auto input1 = loctx->GetOutputOp(operand(0));
+  auto input2 = loctx->GetOutputOp(operand(1));
+  return ReturnOp(
+      (xla::GetDimensionSize(input1, 0) + xla::GetDimensionSize(input2, 0)),
```
JackCaoG, line 61: why do we only add the first dimension size?

JackCaoG (commented):
Thanks, some minor comments

torch_xla/csrc/ops/dynamic_ir.cpp
miladm: because it's guaranteed that the shape is one dimensional. each node represents one dimension.

torch_xla/csrc/ops/dynamic_ir.cpp
miladm: Well, either way works. 

This is mainly a design choice that provides explicit clarity that `SizeNode` delivers a staticValue, implying the value is not necessarily the true shape.

torch_xla/csrc/ops/dynamic_ir.cpp
miladm: same as above.

torch_xla/csrc/ops/dynamic_ir.cpp
JackCaoG: I guess my question is more of why not use `operand(0).node->shape(0)` directly? What's the type of `operand(0).node`?

torch_xla/csrc/ops/dynamic_ir.cpp
JackCaoG: on a second thought, does `SizeAdd` always applied to a single dimension? If that's the case it will always be a single s32?

torch_xla/csrc/ops/dynamic_ir.cpp
miladm: Thanks for the clarification. Yes, I call `shape()` directly - change coming up now. The reason it was initially set like this is due to a legacy reason - we didn't use to have multiple inheritance for `SizeNode` initially; now that we do, we can do what you suggested.

torch_xla/csrc/ops/dynamic_ir.h
miladm: I made it private.

torch_xla/csrc/ops/dynamic_ir.cpp
miladm: My understanding is yes.

JackCaoG (approved):
Thanks!

PR #3744, opened 2022-07-21
Add `SymIntElements` to enable unpacking `SymIntArrayRef`

miladm:
Add `SymIntElements` to enable unpacking `SymIntArrayRef`

### DIFF
--- a/torch_xla/csrc/torch_util.cpp
+++ b/torch_xla/csrc/torch_util.cpp
@@ -15,18 +15,19 @@ void SymIntElements::SetSymIntNodeElements(c10::SymInt& size) {
         std::dynamic_pointer_cast<torch::lazy::SymbolicIntNode>(
             symbolicIntNode);
     auto size_node = lazySymIntNode->node_;
-    size_nodes.push_back(size_node);
-    upper_bounds.push_back(
+    size_nodes_.push_back(size_node);
+    upper_bounds_.push_back(
         std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node)
             ->getStaticValue());
-    dynamic_dims.push_back(
+    dynamic_dims_.push_back(
         std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node)
             ->isDynamic());
   } else {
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/torch_util.cpp
```diff
@@ -5,6 +5,29 @@
...
+    size_nodes.push_back(size_node);
+    upper_bounds.push_back(
+        std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node)
+            ->getStaticValue());
+    dynamic_dims.push_back(
+        std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node)
+            ->isDynamic());
+  } else {
+    auto size_node = torch::lazy::MakeNode<Constant>(std::move(
+        XlaHelpers::ScalarLiteral(size.expect_int(), xla::PrimitiveType::F64)));
```
JackCaoG, line 25: why f64? if it is a size then we should store it as s32.

torch_xla/csrc/torch_util.h
```diff
@@ -5,9 +5,33 @@
...
+  SymIntElements(c10::SymIntArrayRef& size) {
+    std::vector<c10::SymInt> _sizes = torch::lazy::ToVector<c10::SymInt>(size);
+    for (auto& _size : _sizes) {
+      SetSymIntNodeElements(_size);
+    }
+  }
+
+  std::vector<torch::lazy::NodePtr> size_nodes;
+  std::vector<int64_t> upper_bounds;
+  std::vector<bool> dynamic_dims;
```
JackCaoG, line 29: Can we renamed name with `_` at the end and move all of them to private then provide a getter(I don't think we need a setter).

JackCaoG (changes requested):
nits

torch_xla/csrc/torch_util.h
miladm: We likely need setters. I will push setters at a later stage when the use for them is evident.

JackCaoG (approved):
Thanks!

PR #3745, opened 2022-07-21
TensorFlow June pin update & PyTorch/XLA `c++17` migration

miladm:
Update the TF pin to a June commit. 

Because TensorFlow has recently moved to `c++17`, this PR requires that we move the TF dependency, the PyTorch/XLA source code, PyTorch/XLA tests to `c++17`. Keep in mind that in this implementation, we expect PyTorch to remain at `c++14`.

Related issue: https://github.com/pytorch/pytorch/issues/81415

----
**Blockers**
- [x] https://github.com/pytorch/pytorch/pull/82032
- [x] https://github.com/pytorch/pytorch/pull/81904


miladm:
This issue runs into upstream pytorch issues that's describe at https://github.com/pytorch/pytorch/issues/81415

miladm:
@wconstab this is my TF PR from last week. 
Let's use this PR to test against your PR: https://github.com/pytorch/pytorch/pull/81904

JackCaoG:
@miladm You can pin upstream pr using `torch_path/.torch_pin`

JackCaoG:
It seems like we built pytorch/xla in c++14 (to match pytorch), however when we include headers from tf, c++17 stuff will break.. I think you mentioned you run into a different set of issues when you try to compile pytorch/xla with c++17.

miladm:
When compiling pytorch/xla with c++14 and tensorflow with c++17:
* I see 98 errors. The common trend I observe has the following breaking pattern.

```
sw/dupe/pytorch5/xla/third_party/tensorflow/bazel-tensorflow/tensorflow/compiler/xla/shape_util.h:789:24: error: ‘nullopt’ is not a member of ‘std’; did you mean ‘absl::lts_20211102::nullopt’?
  789 |       if (pool != std::nullopt) {
      |                        ^~~~~~~
```

```
sw/dupe/pytorch5/xla/third_party/tensorflow/bazel-tensorflow/tensorflow/compiler/xla/literal.h:138:8: error: ‘optional’ in namespace ‘std’ does not name a template type
  138 |   std::optional<int64_t> GetFirstInteger() const;
      |        ^~~~~~~~
```

```
sw/dupe/pytorch5/xla/third_party/tensorflow/bazel-tensorflow/tensorflow/compiler/xla/literal.h:472:46: error: expected primary-expression before ‘>’ token
  472 |       DCHECK(!std::holds_alternative<TupleRep>(rep_));
      |                                              ^
```

These errors suggest some continued inconsistencies between compiler versions. As a next step, I will make an attempt to upgrade the pytorch/xla compiler version to c++17.

miladm:
Turns out transition to c++17 does lead to successful compilation of Tensorflow dependency while it remains consistent with the upstream pytorch code (which is mostly in C++14 and c++11). Let's confirm everything works in CI after the https://github.com/pytorch/pytorch/pull/82032 is merged.

miladm:
Somehow the CI runs into a bunch of errors like below. I will dig into it.

```
/tmp/pytorch/xla/third_party/tensorflow/bazel-tensorflow/tensorflow/compiler/xla/literal.h:138:8: error: no template named 'optional' in namespace 'std'
  std::optional<int64_t> GetFirstInteger() const;
```

JackCaoG:
@miladm It is when building the test, can you modify the https://github.com/pytorch/xla/blob/master/test/cpp/CMakeLists.txt to use c++17 as well?

miladm:
That's right. Moving my repo to docker to enable local test build. lol

miladm:
All tests passed. Removing the torch pin to prepare for merge.

JackCaoG (approved):
yay

build_torch_xla_libs.sh
```diff
@@ -54,9 +54,9 @@ if [[ ! -z "$BAZEL_JOBS" ]]; then
   MAX_JOBS="--jobs=$BAZEL_JOBS"
 fi
 
-OPTS+=(--cxxopt="-std=c++14")
+OPTS+=(--action_env=BAZEL_CXXOPTS="-std=c++17")
```
yeounoh, line 57: Does passing `--cxxopt=-"std=c++17"` instead work, too?

miladm: Yes, that works. This format seems to be preferred according to a blog post I read at the time. Don't have the reference handy unfortunately.

test/cpp/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ ExternalProject_Add(
   CMAKE_ARGS
     "-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=${PT_CXX_ABI}")
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
```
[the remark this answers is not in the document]
miladm: Didn't try, but I'd expect either to work.

PR #3753, opened 2022-07-22
Add `SetDimensionSizes` API to enable lowering Dynamic Shape in PyTorch/XLA

miladm:
Add `SetDimensionSizes` API to enable lowering Dynamic Shape in PyTorch/XLA

### DIFF
--- a/torch_xla/csrc/data_ops.cpp
+++ b/torch_xla/csrc/data_ops.cpp
@@ -94,6 +94,7 @@ xla::XlaOp BuildView(xla::XlaOp input, absl::Span<const int64_t> output_sizes) {
 
 xla::XlaOp SetDimensionSizes(xla::XlaOp input,
                              absl::Span<const xla::XlaOp> output_sizes) {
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/data_ops.cpp
```diff
@@ -92,6 +92,17 @@ xla::XlaOp BuildView(xla::XlaOp input, absl::Span<const int64_t> output_sizes) {
   return XlaHelpers::DynamicReshape(input, complete_output_sizes);
 }
 
+xla::XlaOp SetDimensionSizes(xla::XlaOp input,
+                             absl::Span<const xla::XlaOp> output_sizes) {
+  for (int i = 0; i < output_sizes.size(); i++) {
```
JackCaoG, line 97: do we need to check whether input size and `output_sizes` has same length?

miladm: Makes sense. Adding the protection.

miladm:
As discussed offline. We will land this feature as part of the a PR where the code path is tested. 
This is the candidate PR that lands this code: https://github.com/pytorch/xla/pull/3889

CC @JackCaoG

PR #3754, opened 2022-07-22
Add `--all-tags` to `docker push`

will-cromar:
See https://docs.docker.com/engine/reference/commandline/push/#push-all-tags-of-an-image


yeounoh (approved):
LGTM -- let's merge and see if this fixes the nightly builds.

PR #3762, opened 2022-07-26
Partially codegen adaptive

JackCaoG:
Fix https://github.com/pytorch/xla/issues/3763 and https://github.com/pytorch/xla/issues/3764

For ops that has special logic like `_adaptive_avg_pool2d`

```
at::Tensor XLANativeFunctions::_adaptive_avg_pool2d(
    const at::Tensor& self, at::IntArrayRef output_size) {
  XLA_FN_COUNTER("xla::");
  auto output_size_list = XlaHelpers::I64List(output_size);
  if (!IsSupportedAdaptivePool(XlaHelpers::I64List(self.sizes()),
                               output_size_list, /*pool_dim=*/2)) {
    return at::native::call_fallback_fn<
        &xla_cpu_fallback, ATEN_OP(_adaptive_avg_pool2d)>::call(self,
                                                                output_size);
  }
  auto shapes =
      torch::lazy::compute_shape__adaptive_avg_pool2d(self, output_size);
  return bridge::AtenFromXlaTensor(XLATensor::_adaptive_avg_pool2d(
      bridge::GetXlaTensor(self), output_size_list, std::move(shapes)));
}
```

we only support some of the output size. We can still codegen the IR using the new `ir_gen`.

In this pr, I update the codegen to make `{OP}OutputShapeFn` to also take vector input, it used to only take `torch::lazy::Value` but sometime we need `vector<>` input to determine the output shape.


JackCaoG:
I should also update migration guide.. will do that in a separate pr.

wonjoolee95 (approved):
Nice! 👍

PR #3773, opened 2022-07-27
Track changes for https://github.com/pytorch/pytorch/pull/82238

ezyang:
We're making factory functions have more accurate dispatch keys
which means that XLA has to re-opt into getting the mutating
variants and getting functionalization to fix it.


ezyang:
I changed my mind and I'm doing it differently so this isn't necessary

PR #3784, opened 2022-07-27
implementing sym_numel_custom

Krovatkin:
We are introducing sym_numel that should be returning the symbolic number of elements, unfortunately the default implementation will give the wrong result even for static shapes, so they will give XLA a new one.

### DIFF
--- a/torch_xla/csrc/tensor_impl.cpp
+++ b/torch_xla/csrc/tensor_impl.cpp
@@ -120,6 +120,15 @@ c10::SymIntArrayRef XLATensorImpl::sym_sizes_custom() const {
                              sizes.size());
 }
 
+c10::SymInt XLATensorImpl::sym_numel_custom() const {
+  auto sym_sizes = sym_sizes_custom();
+  c10::SymInt prod{1};
+  for (auto s : sym_sizes) {
+    prod *= s;
+  }
+  return prod;
+}
+
 c10::SymIntArrayRef XLATensorImpl::sym_sizes() const {
   // it isn't strictly necessary to delegate to `sym_sizes_custom`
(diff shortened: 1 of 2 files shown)


Krovatkin:
@JackCaoG @miladm

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -120,6 +120,15 @@ c10::SymIntArrayRef XLATensorImpl::sym_sizes_custom() const {
                              sizes.size());
 }
 
+c10::SymInt XLATensorImpl::sym_numel_custom() const {
+  auto sym_sizes = sym_sizes_custom();
+  c10::SymInt prod{1};
+  for (auto s : sym_sizes) {
+    prod *= s;
```
JackCaoG, line 127: what's the behavior here if symint is actually dynamic? Is it going to use upper bound?

Krovatkin: it will construct an expression. How we evaluate it (using an upper bound or using an exact value will be determined by how this expression would be used)

miladm (commented):
Thanks @Krovatkin 
I wonder if you need unit tests for `sym_sizes`. Wdyt?

torch_xla/csrc/tensor_impl.cpp
Krovatkin: actually right now it will break on real symints tbh 🤡  since we are missing larger parts of LazySymInt implementation but the above is how things should work in the future

miladm:
sounds good to me. lol

miladm (approved):
LGTM. Thanks.
Waiting for CI to pass.

PR #3785, opened 2022-07-27
Add regression test for torch.expand

ymwangg:
This PR adds test to capture the regression due to aten::expand's fallback to aten::as_strided reported in https://github.com/pytorch/pytorch/pull/82010.

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -58,11 +58,11 @@ function install_deps_pytorch_xla() {
   pip install hypothesis
   pip install cloud-tpu-client
   pip install absl-py
-  pip install --upgrade numpy>=1.18.5
+  pip install --upgrade "numpy>=1.18.5"
   pip install --upgrade numba
 
   # Using the Ninja generator requires CMake version 3.13 or greater
-  pip install cmake>=3.13 --upgrade
+  pip install "cmake>=3.13" --upgrade
 
   sudo apt-get -qq update
 
--- a/.circleci/docker/install_conda.sh
+++ b/.circleci/docker/install_conda.sh
@@ -39,7 +39,7 @@ function install_and_setup_conda() {
   /usr/bin/yes | pip install cloud-tpu-client
   /usr/bin/yes | pip install expecttest==0.1.3
   /usr/bin/yes | pip install ninja  # Install ninja to speedup the build
-  /usr/bin/yes | pip install cmake>=3.13 --upgrade  # Using Ninja requires CMake>=3.13
+  /usr/bin/yes | pip install "cmake>=3.13" --upgrade  # Using Ninja requires CMake>=3.13
   /usr/bin/yes | pip install absl-py
   # Additional PyTorch requirements
   /usr/bin/yes | pip install scikit-image scipy==1.1.0  # >1.1.0 breaks PyTorch tests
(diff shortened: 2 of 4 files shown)


JackCaoG:
You might want to add a torch pin similar to https://github.com/pytorch/xla/pull/3358/files#diff-e71f55b9deb410fad9d5945c4c30878a93369434520520987ac03c59bfa5280d to pin to your pytorch pr

ymwangg:
Interesting, the test was supposed to pass and it passed in my local environment. I though the issue has been fixed by https://github.com/pytorch/pytorch/pull/82189 and my PR https://github.com/pytorch/pytorch/pull/82010 is not required. Let me double check what's going on.

ymwangg:
It looks like the previous CI failure is due to eager debug mode. So to get around the `run_eager_debug` issue, I crated a new file.

This PR is ready to review. I've verified this new test can catch the regression with `torch_pin= 30e74be784ad6c4d5691cb49dbba8327943ee9cb`, the commit ahead of the fix https://github.com/pytorch/pytorch/pull/82189. No other actions on pytorch side are needed.

I also fixed bugs like `pip install --upgrade numpy>=1.18.5`, which will write a file to `=1.18.5` when debugging the CI container.

cc @JackCaoG @miladm

.circleci/common.sh
```diff
@@ -58,11 +58,11 @@ function install_deps_pytorch_xla() {
   pip install hypothesis
   pip install cloud-tpu-client
   pip install absl-py
-  pip install --upgrade numpy>=1.18.5
+  pip install --upgrade "numpy>=1.18.5"
   pip install --upgrade numba
 
   # Using the Ninja generator requires CMake version 3.13 or greater
-  pip install cmake>=3.13 --upgrade
+  pip install "cmake>=3.13" --upgrade
```
JackCaoG, line 65: any reason for these changes?

test/test_operations_fallback.py
```diff
@@ -0,0 +1,47 @@
...
+# Normal imports section starts here.
+import torch
+import torch_xla
+import torch_xla.utils.utils as xu
+import torch_xla.core.xla_model as xm
+import torch_xla.debug.metrics as met
+import unittest
+
+
+class TestOperationsFallback(unittest.TestCase):
```
JackCaoG, line 24: hmm, is fallback a correct name? Fallback means a ops fallback to pytorch native CPU. I think the purpose of this test is more of we want to check we still get the whole op instead some decompositied ops. Maybe `TestOperationsHLO`?

.circleci/common.sh
ymwangg: I accidentally found this bug while debugging ci container. `pip install cmake>=3.13` is equivalent to `pip install cmake > "=3.13"` and creates a file named `=3.13`.

test/test_operations_fallback.py
ymwangg: Sure, I'll rename `fallback` to `hlo`.

JackCaoG:
@ymwangg Sorry I forgot about this one, can you rebase and rerun ci? I think this one is ready to merge

ymwangg:
@JackCaoG This one is ready.

JackCaoG (approved):

PR #3790, opened 2022-07-28
partially codegen adaptive_avgpool3d and backward

JackCaoG:
Fix https://github.com/pytorch/xla/issues/3788 and https://github.com/pytorch/xla/issues/3787

We have some special fallback logic in `adaptive_avgpool3d` so I can't codegen it. However during the development, I still codegen the NativeFunction and copy the parts that is relevant to the `aten_xla_type.cpp`. This prevent me from writing the shape fn logic from scratch. I also get rid of corresponding entry in the `tensor_method.cpp` since in the long run we don't want to keep those.

LazyIr
```
class AdaptiveAvgPool3d : public XlaNode {
 public:
    return torch::lazy::OpKind(at::aten::_adaptive_avg_pool3d);
  }

  AdaptiveAvgPool3d(const torch::lazy::Value& self,
                    const ::std::vector<int64_t>& output_size,
                    std::vector<torch::lazy::Shape>&& shapes)
      : XlaNode(
            torch::lazy::OpKind(at::aten::_adaptive_avg_pool3d), {self},
            std::move(shapes),
            [&]() { return AdaptiveAvgPool3dOutputShape(self, output_size); },
            /* num_outputs */ 1, torch::lazy::MHash(output_size)),
        output_size(output_size) {}

    std::stringstream ss;
    ss << XlaNode::ToString();
    ss << ", output_size=" << output_size;
    return ss.str();
  }

  bool CanBeReused(const torch::lazy::Value& self,
                   const ::std::vector<int64_t>& output_size) const {
    return false;
  }

  ::std::vector<int64_t> output_size;
};

class AdaptiveAvgPool3dBackward : public XlaNode {
 public:
    return torch::lazy::OpKind(at::aten::_adaptive_avg_pool3d_backward);
  }

  AdaptiveAvgPool3dBackward(const torch::lazy::Value& grad_output,
                            const torch::lazy::Value& self,
                            std::vector<torch::lazy::Shape>&& shapes)
      : XlaNode(torch::lazy::OpKind(at::aten::_adaptive_avg_pool3d_backward),
                {grad_output, self}, std::move(shapes),
                [&]() {
                  return AdaptiveAvgPool3dBackwardOutputShape(grad_output,
                                                              self);
                },
                /* num_outputs */ 1, torch::lazy::MHash()) {}

    std::stringstream ss;
    ss << XlaNode::ToString();

    return ss.str();
  }

  bool CanBeReused(const torch::lazy::Value& grad_output,
                   const torch::lazy::Value& self) const {
    return false;
  }

};

```


wonjoolee95 (approved):
Nice! Thanks!

PR #3791, opened 2022-07-28
Use two existing mirrors for download_llvm_raw

yeounoh:
Try contacting both of the existing mirrors for llvm raw download. I tested locally and it seems that we can't rely on just one mirror as the two mirrors may have different (old) archives left in the storage.


JackCaoG (approved):
Thanks!

yeounoh:
Build passed. Merging.

PR #3798, opened 2022-07-29
Respect default dtype

peterbell10:
Prior to pytorch/pytorch#82241, the python API never actually passed nullopt
and so the `OrFloat` path was never tested.


peterbell10:
@JackCaoG ping

JackCaoG (approved):
Thanks!

JackCaoG:
@peterbell10 Test seems to be passing, I guess I can directly merge this pr?

peterbell10:
Yes this makes change sense on its own. I was only using the pin to test that it fixed the xla problems in the main pytorch PR.

PR #3825, opened 2022-08-04
Add env var to disable the neumeric token that is used to string CC o…

hjm-aws:
…ps togather.

### DIFF
--- a/torch_xla/csrc/token_handler.cpp
+++ b/torch_xla/csrc/token_handler.cpp
@@ -31,15 +31,10 @@ xla::XlaOp SliceOneToken(xla::XlaOp input) {
 
 }  // namespace
 
-#ifndef PT_XLA_DISABLE_NUMERIC_CC_TOKEN_DEFAULT_VALUE
-#define PT_XLA_DISABLE_NUMERIC_CC_TOKEN_DEFAULT_VALUE false
-#endif
-
 xla::XlaOp TokenHandler::GetInput(xla::XlaOp input,
                                   const xla::Shape* input_shape) {
   static bool disable_numeric_token =
-      xla::sys_util::GetEnvBool("PT_XLA_DISABLE_NUMERIC_CC_TOKEN",
-                                PT_XLA_DISABLE_NUMERIC_CC_TOKEN_DEFAULT_VALUE);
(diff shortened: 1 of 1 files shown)


torch_xla/csrc/token_handler.cpp
```diff
@@ -30,8 +31,19 @@ xla::XlaOp SliceOneToken(xla::XlaOp input) {
...
 }  // namespace
 
+#ifndef PT_XLA_DISABLE_NUMERIC_CC_TOKEN_DEFAULT_VALUE
+#define PT_XLA_DISABLE_NUMERIC_CC_TOKEN_DEFAULT_VALUE false
+#endif
+
 xla::XlaOp TokenHandler::GetInput(xla::XlaOp input,
                                   const xla::Shape* input_shape) {
+  static bool disable_numeric_token =
+      xla::sys_util::GetEnvBool("PT_XLA_DISABLE_NUMERIC_CC_TOKEN",
```
JackCaoG, line 41: we usually don't prefix our env var with `PT_XLA`, maybe `DISABLE_NUMERIC_CC_TOKEN_DEFAULT_VALUE` directly?

torch_xla/csrc/token_handler.cpp
```diff
@@ -30,8 +31,19 @@ xla::XlaOp SliceOneToken(xla::XlaOp input) {
...
+                                PT_XLA_DISABLE_NUMERIC_CC_TOKEN_DEFAULT_VALUE);
```
JackCaoG, line 42: I think it is OK to hard code `false` directly here. We generally don't use `ifdef` in pt/xla.

JackCaoG:
rebase should solve the build issue.

torch_xla/csrc/token_handler.cpp
hjm-aws: OK, we'll append the following to the end of `torch_xla/__init__.py`:
```
_set_missing_env('DISABLE_NUMERIC_CC_TOKEN', 'true')
```

torch_xla/csrc/token_handler.cpp
hjm-aws: Done

JackCaoG (approved):

JackCaoG:
Thanks, I will merge the pr one all test passed.

PR #3833, opened 2022-08-04
Refactor PjRt device transfers to use `BufferFromHostBuffer`

will-cromar:
- Remove call to `Await()` while still keeping `literal` alive for the duration of the transfer.
- Bump `max_inflight_computations` to the [same default value as JAX](https://github.com/tensorflow/tensorflow/blob/44e84cebef5a89a0840f8d401819c25af41ec3db/tensorflow/compiler/xla/python/xla_client.py#L101). Otherwise, performance drops slightly on v4-8 based on my testing.

Based on [`CopyToDevice` implementation in PjRt itself](https://github.com/tensorflow/tensorflow/blob/a37ac96c6eefe809003f29a815b915bfed5ad36b/tensorflow/compiler/xla/pjrt/pjrt_stream_executor_client.cc#L1491-L1501).

Tested with ResNet50+fake data on v4-8 and got comparable (if not slightly better) performance.

### DIFF
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -96,7 +96,8 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
                 literal_pointer->untyped_data(),
                 literal_pointer->shape().element_type(),
                 literal_pointer->shape().dimensions(), byte_strides,
-                PjRtClient::HostBufferSemantics::kZeroCopy,
+                PjRtClient::HostBufferSemantics::
+                    kImmutableUntilTransferCompletes,
                 [literal{std::move(literal)}]() { /* frees literal */ },
                 pjrt_device)
             .ValueOrDie());
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -48,7 +48,9 @@ PjRtComputationClient::PjRtComputationClient() {
     client_ = std::move(xla::GetCpuClient(/*asynchronous=*/false).ValueOrDie());
   } else if (device_type == "TPU") {
     TF_VLOG(1) << "Initializing PjRt TPU client...";
-    client_ = xla::GetTpuClient(/*max_inflight_computations=*/1).ValueOrDie();
+    int64_t max_inflight_computations = sys_util::GetEnvInt(
+        env::kEnvPjRtTpuMaxInflightComputations, /*defval=*/32);
+    client_ = xla::GetTpuClient(max_inflight_computations).ValueOrDie();
   } else {
     XLA_ERROR() << absl::StrFormat("Unknown %s '%s'", env::kEnvPjRtDevice,
                                    device_type);
@@ -79,13 +81,26 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
   std::vector<ComputationClient::DataPtr> datas;
   datas.reserve(tensors.size());
   for (auto& tensor : tensors) {
-    xla::Literal literal(tensor.shape);
-    tensor.populate_fn(tensor, literal.untyped_data(), literal.size_bytes());
-
     PjRtDevice* pjrt_device = StringToPjRtDevice(tensor.device);
-    std::shared_ptr<xla::PjRtBuffer> buffer =
-        client_->BufferFromHostLiteral(literal, pjrt_device).ValueOrDie();
-    buffer->GetReadyFuture().Await();
+
+    auto literal = std::make_shared<xla::Literal>(tensor.shape);
+    tensor.populate_fn(tensor, literal->untyped_data(), literal->size_bytes());
+    absl::InlinedVector<int64_t, 4> byte_strides(
+        literal->shape().dimensions_size());
+    ShapeUtil::ByteStrides(literal->shape(), absl::MakeSpan(byte_strides));
+
+    // Avoid use-after-free on `literal` due to unsequenced move and use.
+    xla::Literal* literal_pointer = literal.get();
+    std::shared_ptr<xla::PjRtBuffer> buffer = std::move(
+        client_
+            ->BufferFromHostBuffer(
+                literal_pointer->untyped_data(),
+                literal_pointer->shape().element_type(),
+                literal_pointer->shape().dimensions(), byte_strides,
+                PjRtClient::HostBufferSemantics::kZeroCopy,
+                [literal{std::move(literal)}]() { /* frees literal */ },
+                pjrt_device)
+            .ValueOrDie());
     ComputationClient::DataPtr data =
         std::make_shared<PjRtData>(tensor.device, tensor.shape, buffer);
     datas.push_back(data);
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -85,8 +85,7 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
 
     auto literal = std::make_shared<xla::Literal>(tensor.shape);
     tensor.populate_fn(tensor, literal->untyped_data(), literal->size_bytes());
-    absl::InlinedVector<int64_t, 4> byte_strides(
-        literal->shape().dimensions_size());
+    std::vector<int64_t> byte_strides(literal->shape().dimensions_size());
     ShapeUtil::ByteStrides(literal->shape(), absl::MakeSpan(byte_strides));
 
     // Avoid use-after-free on `literal` due to unsequenced move and use.
@@ -101,6 +100,7 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
                 [literal{std::move(literal)}]() { /* frees literal */ },
                 pjrt_device)
             .ValueOrDie());
+
     ComputationClient::DataPtr data =
         std::make_shared<PjRtData>(tensor.device, tensor.shape, buffer);
     datas.push_back(data);
--- a/third_party/xla_client/env_vars.cc
+++ b/third_party/xla_client/env_vars.cc
@@ -18,6 +18,8 @@ const char* const kEnvShardOrdinal = "XRT_SHARD_ORDINAL";
 const char* const kEnvStartService = "XRT_START_LOCAL_SERVER";
 const char* const kEnvTpuvmMode = "TPUVM_MODE";
 const char* const kEnvPjRtDevice = "PJRT_DEVICE";
+const char* const kEnvPjRtTpuMaxInflightComputations =
+    "PJRT_TPU_MAX_INFLIGHT_COMPUTATIONS";
 
 }  // namespace env
 }  // namespace xla
--- a/third_party/xla_client/env_vars.h
+++ b/third_party/xla_client/env_vars.h
@@ -19,6 +19,7 @@ extern const char* const kEnvShardOrdinal;
 extern const char* const kEnvStartService;
 extern const char* const kEnvTpuvmMode;
 extern const char* const kEnvPjRtDevice;
+extern const char* const kEnvPjRtTpuMaxInflightComputations;
 
 }  // namespace env
 }  // namespace xla


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -79,13 +81,26 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
...
   for (auto& tensor : tensors) {
-    xla::Literal literal(tensor.shape);
-    tensor.populate_fn(tensor, literal.untyped_data(), literal.size_bytes());
-
     PjRtDevice* pjrt_device = StringToPjRtDevice(tensor.device);
-    std::shared_ptr<xla::PjRtBuffer> buffer =
-        client_->BufferFromHostLiteral(literal, pjrt_device).ValueOrDie();
-    buffer->GetReadyFuture().Await();
+
+    auto literal = std::make_shared<xla::Literal>(tensor.shape);
```
miladm, line 86: why are you adding `make_shared` here?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -79,13 +81,26 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
...
+    tensor.populate_fn(tensor, literal->untyped_data(), literal->size_bytes());
+    absl::InlinedVector<int64_t, 4> byte_strides(
+        literal->shape().dimensions_size());
+    ShapeUtil::ByteStrides(literal->shape(), absl::MakeSpan(byte_strides));
```
miladm, line 90: side question: can you confirm this works as expected for dynamic shape? I assume it should.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -79,13 +81,26 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
...
+
+    // Avoid use-after-free on `literal` due to unsequenced move and use.
+    xla::Literal* literal_pointer = literal.get();
+    std::shared_ptr<xla::PjRtBuffer> buffer = std::move(
+        client_
+            ->BufferFromHostBuffer(
```
miladm, line 96: How do you intend to unit test this approach?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -79,13 +81,26 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
   std::vector<ComputationClient::DataPtr> datas;
   datas.reserve(tensors.size());
   for (auto& tensor : tensors) {
-    xla::Literal literal(tensor.shape);
-    tensor.populate_fn(tensor, literal.untyped_data(), literal.size_bytes());
-
     PjRtDevice* pjrt_device = StringToPjRtDevice(tensor.device);
-    std::shared_ptr<xla::PjRtBuffer> buffer =
-        client_->BufferFromHostLiteral(literal, pjrt_device).ValueOrDie();
-    buffer->GetReadyFuture().Await();
```
miladm, line 88: Why is this step no longer needed under `BufferFromHostBuffer`?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -48,7 +48,9 @@ PjRtComputationClient::PjRtComputationClient() {
     client_ = std::move(xla::GetCpuClient(/*asynchronous=*/false).ValueOrDie());
   } else if (device_type == "TPU") {
     TF_VLOG(1) << "Initializing PjRt TPU client...";
-    client_ = xla::GetTpuClient(/*max_inflight_computations=*/1).ValueOrDie();
+    int64_t max_inflight_computations = sys_util::GetEnvInt(
+        env::kEnvPjRtTpuMaxInflightComputations, /*defval=*/32);
```
miladm, line 52: Why is 32 the right `defval`?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -79,13 +81,26 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
...
-    tensor.populate_fn(tensor, literal.untyped_data(), literal.size_bytes());
-
     PjRtDevice* pjrt_device = StringToPjRtDevice(tensor.device);
-    std::shared_ptr<xla::PjRtBuffer> buffer =
-        client_->BufferFromHostLiteral(literal, pjrt_device).ValueOrDie();
-    buffer->GetReadyFuture().Await();
+
+    auto literal = std::make_shared<xla::Literal>(tensor.shape);
+    tensor.populate_fn(tensor, literal->untyped_data(), literal->size_bytes());
+    absl::InlinedVector<int64_t, 4> byte_strides(
```
miladm, line 88: What does `byte_strides` hold? 
Is it the number of bytes consumed by `tensor`?

miladm (commented):
Thanks @will-cromar. Added a few comments and questions to align.

third_party/xla_client/pjrt_computation_client.cc
miladm: I see you did this to match that of JAX. Does it make sense to verify PT/XLA will benefit from the same config or something slightly different?

will-cromar: I copied the default value from JAX. I made it configurable, so we can try a few different values once we have a better set of benchmarks.

third_party/xla_client/pjrt_computation_client.cc
miladm: This makes sense now.

will-cromar: `Await` was used before to ensure that `literal` (which was on the stack) was kept alive for the duration of the transfer. Using this new implementation, a callback takes ownership of `literal`, and after that callback fires (after the transfer is complete), the `literal` gets freed. So, basically, that means we can perform the transfer asynchronously without the risk of `literal` getting cleaned up in the middle of the transfer.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: See above for how/why we pass ownership of `literal` to the `on_done_with_host_buffer` callback. To be honest, I couldn't figure out how to make this compile with `unique_ptr` or `literal` as a stack variable, so I used `shared_ptr` and "borrowed" the [logic used in `pjrt_stream_executor_client`](https://github.com/tensorflow/tensorflow/blob/a37ac96c6eefe809003f29a815b915bfed5ad36b/tensorflow/compiler/xla/pjrt/pjrt_stream_executor_client.cc#L1491-L1501).

Not sure if this is the best way. I'm definitely open to alternate ways of writing this.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: As I understand it, stride is used in the same sense as in [PyTorch](https://pytorch.org/docs/stable/generated/torch.Tensor.stride.html):

```
Stride is the jump necessary to go from one element to the next one in the specified dimension [dim](https://pytorch.org/docs/stable/generated/torch.Tensor.dim.html#torch.Tensor.dim).
```

Not sure what the magic number `4` is here, this was copied in from PjRt. Will look into it an clarify.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: How can I confirm this works with dynamic shape?

third_party/xla_client/pjrt_computation_client.cc
will-cromar: We don't have good unit tests for either ComputationClient, unfortunately. This does get a sanity check as part of the PjRt CPU integration tests that run in the CI, and I manually ran tests on TPU.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: Read the doc for `InlinedVector`, not sure it was actually doing anything useful here. Replaced it with `std::vector` to remove the magic number.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: Since none of this code has been tested with dynamic shapes, can you file an issue to me internally to look into it?

third_party/xla_client/pjrt_computation_client.cc
miladm: Sounds good. The reason I brought up this question is to know if we are using `byte_strides` as a storage for the "overall size of the tensor" or for determining the "number of bytes to traverse on every stride taken"? Looks like it's the latter.

third_party/xla_client/pjrt_computation_client.cc
miladm: Yeah, it'd great to track this as an internal issue to determine the optimal size for PyTorch/XLA.

CC @JackCaoG

third_party/xla_client/pjrt_computation_client.cc
miladm: Sounds good. I suggest incorporating unit tests for the main building blocks of PjRt integration in a future set of PRs.

third_party/xla_client/pjrt_computation_client.cc
miladm: Done. I sent you a ticket. Thanks.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: Yeah, stride is the incremental step size, not the overall size

miladm: Ok, in that case, are you passing the "max" incremental step size in each dimension as the input parameter to this method? If so, I understand what you are doing.

will-cromar: ByteStrides fills the input `strides` vector with the stride in each dimension 

https://github.com/tensorflow/tensorflow/blob/be29b5d32fd132d7f6d75c0ba2e37b7b0347444e/tensorflow/compiler/xla/shape_util.cc#L1729-L1740

third_party/xla_client/pjrt_computation_client.cc
will-cromar: Agreed. Filed a cleanup ticket internally.

JackCaoG:
General logic make sense to me. Both CPU and GPU CI failed with `SIGSEGV` which
```
means that the program accessed a memory location that was not assigned to it
```
I think this is a real error. By making the literal as a shared_ptr, we keep the address space valid until the transfer to device is done. However, how do we guarantee that when we trying to execute the graph, the transfer has completed? This might be the cause of the `SIGSEGV`?(executor trying to access the device memory while host to device transfer is not completed).

will-cromar:
Thanks for flagging the error. To be honest, I've just tuned out our CI at this point and hadn't looked yet 😬 

My understanding based on this comment is that this should be safe: https://github.com/tensorflow/tensorflow/blob/a37ac96c6eefe809003f29a815b915bfed5ad36b/tensorflow/compiler/xla/pjrt/pjrt_stream_executor_client.cc#L21-L24

However, I think you're right that this is a real error this time, although I'm surprised that I didn't see a crash in end-to-end model training. I'll try `Await`ing each buffer before we execute the graph and see what happens

JackCaoG:
ops tests runs more smaller graph, which will make race condition easier to surface.

will-cromar:
I got a crash in a different place within the same set of tests (not that surprising for a memory issue).

Adding `Await` before execution (or even immediately after the `BufferFromHostBuffer`) call made no difference. But changing `host_buffer_semantics` from `kZeroCopy` to `kImmutableUntilTransferCompletes` made everything work as expected on my development machine.

The [documentation](https://github.com/tensorflow/tensorflow/blob/93d566f53864121f6f2b71a7fad4d260dea0b144/tensorflow/compiler/xla/pjrt/pjrt_client.h#L643) for this enum notes that `On non-CPU platforms [kZeroCopy] acts identically to kImmutableUntilTransferCompletes`, which could explain why I saw no issues on TPU.

It's not clear to me why `kZeroCopy` would cause issues based on the documentation: ``The caller promises to keep `data` alive and not to mutate its contents as long as the buffer is alive; to notify the caller that the buffer may be freed, the runtime will call `on_done_with_host_buffer` when the PjRtBuffer is freed.`` The `on_done_with_host_buffer` callback shouldn't fire until the runtime frees the buffer, so I don't know what is accessing the PjRtBuffer after it is freed in this case.

However, I think it's safer to err on the side of having the same behavior across platforms (i.e. `kImmutableUntilTransferCompletes`), especially since we're mainly using CPU for testing.

will-cromar:
The latest errors don't look PJRT related. Will rebase a try again...

JackCaoG (approved):
Thanks!

miladm (approved):
LGTM. Thanks @will-cromar.

PR #3836, opened 2022-08-05
Add parameter broadcasting to PJRT examples.

darisoy:
Add the [this snippet](https://github.com/pytorch/xla/issues/3692#issue-1298080951) to `torch_xla` to initialize all PJRT process with the same parameters.

Update the PJRT examples to use this change and add unit test for the functionality.

The unit test, mnist and imagenet tests have been tested on both v3-8 and v4-8 TPU VMs.

### DIFF
--- a/test/pjrt/test_parameter_broadcast_tpu.py
+++ b/test/pjrt/test_parameter_broadcast_tpu.py
@@ -1,0 +1,31 @@
+from numpy.testing import assert_array_equal, assert_raises
+import torch
+import torch.nn as nn
+import torch_xla.core.xla_model as xm
+from torch_xla.experimental import pjrt
+from absl.testing import absltest, parameterized
+
+def broadcast(sync):
+    torch.manual_seed(xm.get_ordinal())
+    device = xm.xla_device()
+    model = nn.Linear(5, 5).to(device)
+    if sync:
+        pjrt.broadcast_xla_master_model_param(model)
+    return next(model.parameters()).detach().cpu().numpy()
+    
+class TestBroadcastParametersPjrt(parameterized.TestCase):
+    
+    @parameterized.named_parameters(('synchronized_parameters', True), 
+                                    ('unsynchronized_parameters', False))
+    def test_broadcast_parameter_sync(self, sync):
+        torch.set_default_tensor_type('torch.FloatTensor')
+        results = pjrt.run_multiprocess(broadcast, sync)
+        for key in results:
+            a, b = results[0][0], results[key][0]
+            if sync:
+                assert_array_equal(a, b)
+            elif key != 0:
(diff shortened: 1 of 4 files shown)


test/pjrt/test_parameter_broadcast_tpu.py
```diff
@@ -0,0 +1,31 @@
...
+    return next(model.parameters()).detach().cpu().numpy()
+    
+class TestBroadcastParametersPjrt(parameterized.TestCase):
+    
+    @parameterized.named_parameters(('synchronized_parameters', True), 
+                                    ('unsynchronized_parameters', False))
+    def test_broadcast_parameter_sync(self, sync):
+        torch.set_default_tensor_type('torch.FloatTensor')
+        results = pjrt.run_multiprocess(broadcast, sync)
+        for key in results:
```
will-cromar, line 23: It might not be obvious here on first reading what `0` represents here. I suggest changing the iteration variable here to `process_key` or `process_id`, and then maybe change `a` and `b` to `master_params` and `worker_params` to make clear why `results[0]` is significant.

torch_xla/experimental/pjrt.py
```diff
@@ -214,3 +215,20 @@ def run_multiprocess(fn: Callable[..., R], *args,
     }
 
   return results
+
+@requires_pjrt
+def broadcast_xla_master_model_param(model: xm) -> None:
```
will-cromar, line 220: Nit: I think this name should be more terse. I think `broadcast_master_parameters`. `xla` and `model` are already apparent from the signature and context.

Also, the argument should be type `nn.Module`. `xm` is a module in this context

torch_xla/experimental/pjrt.py
```diff
@@ -214,3 +215,20 @@ def run_multiprocess(fn: Callable[..., R], *args,
     }
 
   return results
+
+@requires_pjrt
```
will-cromar, line 219: Nit: this doesn't _actually_ require `pjrt`. Although this will be the recommended way to synchronize model parameters under PJRT, this is still compatible with XRT

will-cromar (changes requested):
Thanks! Just a few minor suggestions

will-cromar (approved):
LGTM. Thanks!

JackCaoG (approved):
LGTM. Does `pjrt.run_multiprocess` supports single core training? In the old `xmp.spawn` world, we can pass the num_process as 1 and thins still works.

will-cromar:
That's possible on v4-8. You'll need to pick one chip and set the process bounds (e.g. `TPU_VISIBLE_DEVICES=0 TPU_PROCESS_BOUNDS=1,1,1`). On v3, this will still give you 2 threads. It's worth having a `num_process` argument that works in both cases though. Can you file a feature request to me internally?

PR #3855, opened 2022-08-10
Remove Ger lowering as it is not needed

JackCaoG:
Fix https://github.com/pytorch/xla/issues/3852

this op will be deprecated based on https://pytorch.org/docs/stable/generated/torch.ger.html


wonjoolee95:
Should we lower `torch.outer` as a follow-up item to this?

wonjoolee95 (approved):

JackCaoG:
`torch.outer` is dispatched to `mul` + `view`, we get it for free. I think now `torch.ger` and `torch.outer` has the same dispatch logic.

wonjoolee95:
> `torch.outer` is dispatched to `mul` + `view`, we get it for free. I think now `torch.ger` and `torch.outer` has the same dispatch logic.

Nice!

Issue #3446, opened 2022-03-24
XLA profiler cannot capture TPU device trace when running on a pod


ronghanghu:
## 🐛 Bug / 🚀 Feature request

Currently, the XLA profiler can capture the TPU device trace when running on a v3-8 TPU VM, but **cannot capture the device trace in trace_viewer when running on a TPU pod** (e.g. a v3-128 pod). This is unlike the [TensorFlow profiler](https://cloud.google.com/tpu/docs/cloud-tpu-tools#trace_viewer_interface) which is able to capture TPU device traces when running on a pod.

As we have frequently observed a bigger overhead from pod training (compared to training on a v3-8 under the same per-TPU batch size, e.g. https://github.com/pytorch/xla/issues/3441), it would be great if the TPU device trace can also be captured in the pod case to help understand the performance bottlenecks.

(Not sure whether this should be a bug report or a feature request. Since many practical TPU use cases involve training in pods, it would be great if the XLA profiler could also work for the pod case.)

## To Reproduce

While the XLA profiler can capture CPU host traces when doing TPU pod training (e.g. v3-128), it cannot capture the TPU device trace. In particular, no device trace shows up on the `trace_viewer` page in the profiler, as shown in the screenshot below (only CPU traces are captured).

![v3_128](https://user-images.githubusercontent.com/6997335/159957219-82ef40da-9c09-45e8-99d8-13e1429df01e.png)

The tensorboard output (including the profiler results) for this case is also uploaded to https://drive.google.com/file/d/108GRRqndJJyEQEhmICx1u4aaiF1vkQ_F/view?usp=sharing.

---

Note that on the other hand, the profiler successfully captures TPU device traces on a v3-8 (as in the screenshot below).

![v3_8](https://user-images.githubusercontent.com/6997335/159961053-101f50c0-a1b5-444c-88a6-b1b354b309b3.png)

---

#### To reproduce the failure case above on v3-128

1. Allocate a v3-128 TPU VM pod (e.g. with name `tpu-debug-128`) from the `tpu-vm-pt-1.10` environment
2. Clone the PyTorch XLA repo the TPU VM to download the profiler script (do this on all the nodes in the TPU VM, e.g. through`gcloud alpha compute tpus tpu-vm ssh --worker all`):
``` 
mkdir -p /home/ronghanghu/workspace && git clone https://github.com/pytorch/xla /home/ronghanghu/workspace/xla
```
[...]

PR #3866, opened 2022-08-10
Add profiler traces to `PjRtComputationClient`

will-cromar:
The existing profiler works out of the box because the traces work the same way. Tested manually with TensorBoard:

image

### DIFF
--- a/test/pjrt/test_profiler.py
+++ b/test/pjrt/test_profiler.py
@@ -24,7 +24,10 @@ def _profile(logdir: str, port: int = 9012):
 class TestPjRtProfiler(absltest.TestCase):
 
   def setUp(self):
-    pjrt.set_device_type('CPU')
+    assert pjrt.using_pjrt()
+
+    # HACK: ensure libtpu is loaded if using TPU
+    xm.xla_device()
 
   def test_profiler_output(self):
     tempdir = self.create_tempdir().full_path
--- a/test/pjrt/test_profiler.py
+++ b/test/pjrt/test_profiler.py
@@ -2,6 +2,7 @@ import codecs
 import contextlib
 import glob
 import os
+import time
 import threading
 
 from absl.testing import absltest
@@ -18,8 +19,13 @@ def _profile(logdir: str, port: int = 9012):
   tracer = threading.Thread(target=xp.trace, args=(f'localhost:{port}', logdir))
   tracer.setDaemon(True)
   tracer.start()
+
+  # HACK: Give tracer time to start before we return control
+  time.sleep(.5)
+
   yield
 
+  del server
 
 class TestPjRtProfiler(absltest.TestCase):
 
@@ -32,10 +38,9 @@ class TestPjRtProfiler(absltest.TestCase):
   def test_profiler_output(self):
     tempdir = self.create_tempdir().full_path
 
+    device = xm.xla_device()
+    ones = torch.ones([5])
     with _profile(tempdir):
-      device = xm.xla_device()
-      ones = torch.ones([5])
-
       xones = ones.to(device)
       xtwos = xones + xones
       xm.mark_step()
--- a/test/pjrt/test_profiler.py
+++ b/test/pjrt/test_profiler.py
@@ -27,6 +27,7 @@ def _profile(logdir: str, port: int = 9012):
 
   del server
 
+
 class TestPjRtProfiler(absltest.TestCase):
 
   def setUp(self):
--- a/test/pjrt/test_profiler.py
+++ b/test/pjrt/test_profiler.py
@@ -1,0 +1,56 @@
+import codecs
+import contextlib
+import glob
+import os
+import threading
+
+from absl.testing import absltest
+import torch
+import torch_xla.core.xla_model as xm
+import torch_xla.experimental.pjrt as pjrt
+import torch_xla.debug.profiler as xp
+
+
+@contextlib.contextmanager
+def _profile(logdir: str, port: int = 9012):
+  server = xp.start_server(port)
+
+  tracer = threading.Thread(target=xp.trace, args=(f'localhost:{port}', logdir))
+  tracer.setDaemon(True)
+  tracer.start()
+  yield
+
+
+class TestPjRtProfiler(absltest.TestCase):
+
+  def setUp(self):
+    pjrt.set_device_type('CPU')
+
+  def test_profiler_output(self):
+    tempdir = self.create_tempdir().full_path
+
+    with _profile(tempdir):
+      device = xm.xla_device()
+      ones = torch.ones([5])
+
+      xones = ones.to(device)
+      xtwos = xones + xones
+      xm.mark_step()
+
+    profiles = glob.glob(os.path.join(tempdir, "plugins/profile/*/*.xplane.pb"))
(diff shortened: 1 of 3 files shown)


test/pjrt/test_profiler.py
```diff
@@ -0,0 +1,59 @@
...
+  yield
+
+
+class TestPjRtProfiler(absltest.TestCase):
+
+  def setUp(self):
+    assert pjrt.using_pjrt()
+
+    # HACK: ensure libtpu is loaded if using TPU
+    xm.xla_device()
```
will-cromar, line 30: Is there a more explicit way to initialize the ComputationClient?

will-cromar:
The new test passed on CPU but flaked on GPU:

```
2022-08-10 23:44:46.343597: W  562435 tensorflow/core/profiler/lib/profiler_session.cc:107] Profiling is late by 1594605 nanoseconds and will start immediately.
2022-08-10 23:44:48.736212: W  562436 tensorflow/core/profiler/rpc/client/capture_profile.cc:133] No trace event is collected from localhost:9012
```

darisoy (approved):

will-cromar:
The GPU test has `Profiling is late by 1594605 nanoseconds` vs CPU with `Profiling is late by 760266 nanoseconds`. Maybe the XLA execution had finished by the time tracing had started?

will-cromar:
There is some lag between when the tracer thread starts and when it actually starts tracing, long enough that XLA execution can finish before it starts. This test was only passing when `xm.xla_device()` and `torch.ones()` were sufficiently slow to let the tracer get a head start, and I was able to reproduce the error consistently by moving them before the tracer thread started.

Adding a half-second delay makes the test pass consistently on my machine. Hopefully that holds up on the CI 🤞

will-cromar:
Test still flaked on GPU (even though it's using CPU) and I can't reproduce the error locally. Removing it from the CI tests since I'll have to add a TPU version outside of the CI anyway.

test/pjrt/test_profiler.py
JackCaoG: I think this is fine, this is how I usually trigger computationClient creation too..

JackCaoG: would this test work on non-tpu platform?

will-cromar: Yes. I tested it on CPU and it works, although it's too flaky to add to the CI (see my comments below)

JackCaoG:
How is this test different from https://github.com/pytorch/xla/blob/master/test/test_profiler.py we already have? It seem like that tests runs mnist. If you switch the runtime from xrt to pjrt would `test_profiler.py` still pass?

For the TPU test, can you make sure device side trace also shows up?

will-cromar:
This test specifically checks for the PJRT traces I added in this PR. This test is also more isolated (no dependency on another test) and uses a much smaller graph, so it's actually possible to read what is in the trace file if there's an error. The other test is more of an end-to-end test, while I tried to write a minimal integration test here.

I tried running the other test with PJRT and it did not work. I would need to add a call to `pjrt.run_multiprocess` to set up the TPU environment variables, set the world size, etc.

I did manually check that the TPU device traces are in the output.

JackCaoG (approved):
Thanks!

miladm (approved):
Great. Thanks.

ronghanghu:
@will-cromar Thanks for adding this feature! We tried using the profiler under PJRT in our applications and observed a few issues:

1) Here in PJRT when starting a profiling server with `xp.start_server` on v3-8 running the PJRT ImageNet example (by adding `server = xp.start_server(9012)` at the beginning of `train_imagenet`), it only captures the device traces of 2 TPU cores. This is unlike on XRT where we could get the device traces of all 8 TPU cores.

2) Unlike in the XRT cases where we propagate the `xp.Trace` annotations to TPU device traces with `XLA_HLO_DEBUG=1` (as described in [Annotation propagation](https://cloud.google.com/blog/topics/developers-practitioners/pytorchxla-performance-debugging-cloud-tpu-vm-part-iii)), it seems that the TPU device traces captured under PJRT cannot propagate the annotations to TPU device traces even with `XLA_HLO_DEBUG=1`.

3) We are happy to find that the profiler could work on a TPU pod (unlike the XRT cases where the profiler doesn't work on a pod as we reported in https://github.com/pytorch/xla/issues/3446)! Meanwhile, on a v3-128 pod, it still only captures the device traces of 2 TPU cores.

4) A related bug is that `xm.get_local_ordinal` only returns a value in 0,1,2,3 on TPU v3 (and two TPU v3 cores that are on the different chips in the TPU v3 board could have the same `xm.get_local_ordinal`), whereas it is expected to return a value in 0,1,2,3,4,5,6,7 as in XRT. I believe we should module by 8 instead of 4 as `local_world_size` on TPU v3 in https://github.com/pytorch/xla/blob/5059c3b13b5c190d2871ccd004caff89117ae2af/torch_xla/experimental/pjrt.py#L160

Below is a screenshot illustrating the traces we captured with `XLA_HLO_DEBUG=1` under PJRT on a v3-128 TPU, running our [ImageNet FSDP test case](https://gist.github.com/ronghanghu/d5d88c3072a2ca54cc1778122e3f79ee#file-test_train_pjrt_imagenet_fsdp_with_profiler-py).

pjrt_on_pod

JackCaoG:
I have a feeling that only seeing one core is due to
1. In XRT there is a single process(XRT_SERVER) driving the whole TPU device execution 
2. In PJRT on v3 each process only drive the TPU execution for 1 chip(2 core)

There should be a way around this through.

will-cromar:
1. @JackCaoG is correct. Each process only has two cores, so this is expected. I also expect starting multiple profiler servers and tracing all of them to show more cores, although I think we need a better API to make this use case easier in practice.
2. I wasn't aware of this use case. Can you file an issue to me and I can follow up?
3. Awesome! I'm glad this works without having to implement anything extra (except a better profiler API that I alluded to in 1) 😁 
4. Good catch. I'm mainly doing my testing on TPU v4 so this slipped by. I need to refactor how we store these ordinals in #3949 anyway, so I'll make sure to fix this issue as part of that change.

ronghanghu:
> @JackCaoG is correct. Each process only has two cores, so this is expected. I also expect starting multiple profiler servers and tracing all of them to show more cores, although I think we need a better API to make this use case easier in practice.

I see, this makes sense to me. (I was thinking about the case in JAX PJRT where its `jax.profiler.trace` could capture the device traces of all 8 TPU v3 cores, but I guess the difference is that JAX runs in an SPMD manner :)

> I wasn't aware of this use case. Can you file an issue for me and I can follow up?

Sounds good, I'll file a new issue on this. This is mostly to propagate the `xp.Trace` annotations (similar to `jax.profiler.TraceAnnotation`) onto the "TensorFlow Name Scope" annotations on the TPU device trace. It was working under XRT when we set `XLA_HLO_DEBUG=1`.

Below is a TPU trace that we captured earlier with the XRT runtime in PT/XLA, where the `xp.Trace` annotations such as "Forward model" can be propagated to TPU device traces under "TensorFlow Name Scope".

Screen Shot 2022-09-01 at 12 54 51 PM
Screen Shot 2022-09-01 at 12 55 03 PM

PR #3889, opened 2022-08-15
Support `expand.symint` with real symbolic symint

miladm:
Support `expand.symint` 
This op supports dynamic `SymInt` tests after other required components land in the `master`.

### DIFF
--- a/test/cpp/cpp_test_util.cpp
+++ b/test/cpp/cpp_test_util.cpp
@@ -398,8 +398,8 @@ torch::lazy::NodePtr CreateNonZeroNode2d(int64_t num_non_zero_element,
   //   if j == num_col:
   //     j = 0
   //     i += 1
+  //     count += 1
   // res = t1.non_zero()
-  // ASSERT_TRUE(num_non_zero_element <= (num_row * num_col));
   torch::lazy::Value scalar_value =
       torch::lazy::Value(ScalarOp(0.0, xla::F32), 0);
   torch::lazy::Value scalar_value_1 =
--- a/test/cpp/cpp_test_util.cpp
+++ b/test/cpp/cpp_test_util.cpp
@@ -9,6 +9,10 @@
 #include "torch_xla/csrc/ir_dump_util.h"
 #include "torch_xla/csrc/lowering_context.h"
 #include "torch_xla/csrc/ops/device_data.h"
+#include "torch_xla/csrc/ops/expand.h"
+#include "torch_xla/csrc/ops/nonzero.h"
+#include "torch_xla/csrc/ops/ops.h"
+#include "torch_xla/csrc/ops/update_slice.h"
 #include "torch_xla/csrc/tensor_impl.h"
 #include "torch_xla/csrc/tensor_util.h"
 #include "torch_xla/csrc/torch_util.h"
@@ -380,6 +384,48 @@ void TestBackward(
       }
     }
   }
+}
+
+torch::lazy::NodePtr CreateNonZeroNode2d(int64_t num_non_zero_element,
+                                         int64_t num_row, int64_t num_col) {
+  // t1 = torch.tensor(0).expand(row,col)
+  // i = 0
+  // j = 0
+  // count = 0
+  // while count < num_non_zero_element:
+  //   t1[i][j] = 1
+  //   j += 1
+  //   if j == num_col:
+  //     j = 0
+  //     i += 1
+  // res = t1.non_zero()
+  // ASSERT_TRUE(num_non_zero_element <= (num_row * num_col));
+  torch::lazy::Value scalar_value =
+      torch::lazy::Value(ScalarOp(0.0, xla::F32), 0);
+  torch::lazy::Value scalar_value_1 =
+      torch::lazy::Value(ScalarOp(1.0, xla::F32), 0);
+  std::vector<int64_t> target_size = {num_row, num_col};
+  torch::lazy::NodePtr expand_node =
+      torch::lazy::MakeNode<Expand>(scalar_value, target_size);
+  torch::lazy::Value expand_value = torch::lazy::Value(expand_node, 0);
+  int64_t count = 0;
+  int64_t i = 0;
+  int64_t j = 0;
+  torch::lazy::Value slice_value = expand_value;
+  while (count++ < num_non_zero_element) {
+    std::vector<int64_t> base_indeies = {i, j++};
+    // Use Slice to do element update
+    torch::lazy::NodePtr slice_node = torch::lazy::MakeNode<UpdateSlice>(
+        slice_value, scalar_value_1, base_indeies);
+    slice_value = torch::lazy::Value(slice_node, 0);
+    if (j == num_col) {
+      j = 0;
+      i++;
+    }
+  }
+  torch::lazy::NodePtr nonzero_node =
+      torch::lazy::MakeNode<NonZero>(slice_value);
+  return nonzero_node;
 }
 
 }  // namespace cpp_test
--- a/test/cpp/cpp_test_util.cpp
+++ b/test/cpp/cpp_test_util.cpp
@@ -413,10 +413,10 @@ torch::lazy::NodePtr CreateNonZeroNode2d(int64_t num_non_zero_element,
   int64_t j = 0;
   torch::lazy::Value slice_value = expand_value;
   while (count++ < num_non_zero_element) {
-    std::vector<int64_t> base_indeies = {i, j++};
+    std::vector<int64_t> base_indices = {i, j++};
     // Use Slice to do element update
     torch::lazy::NodePtr slice_node = torch::lazy::MakeNode<UpdateSlice>(
-        slice_value, scalar_value_1, base_indeies);
+        slice_value, scalar_value_1, base_indices);
     slice_value = torch::lazy::Value(slice_node, 0);
     if (j == num_col) {
       j = 0;
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -4945,7 +4945,8 @@ TEST_F(AtenXlaTensorTest, TestExpandSymIntDynamic) {
 
   torch::lazy::NodePtr size_node_nonzero_0 =
       torch::lazy::MakeNode<SizeNode>(nonzero_node, 0);
-  auto symint_node = c10::make_intrusive<XLASymIntNodeImpl>(size_node_nonzero_0);
+  auto symint_node =
+      c10::make_intrusive<XLASymIntNodeImpl>(size_node_nonzero_0);
   c10::SymInt dynamic_symint = symint_node->toSymInt();
 
   ForEachDevice([&](const torch::Device& device) {
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -4907,7 +4907,7 @@ TEST_F(AtenXlaTensorTest, TestExpandSymIntSymbolic) {
   // symbolic(since it wraps around a sizeNode) but not really dynamic.
   torch::lazy::Value scalar_value =
       torch::lazy::Value(ScalarOp(1.0, xla::F32), 0);
-  std::vector<int64_t> target_size = {2, 3, 5};
+  std::vector<int64_t> target_size = {2, 3, 4};
   torch::lazy::NodePtr expand_node =
       torch::lazy::MakeNode<Expand>(scalar_value, target_size);
   torch::lazy::Value expand_value = torch::lazy::Value(expand_node, 0);
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -9,6 +9,9 @@
 #include "tensorflow/compiler/xla/xla_client/metrics.h"
 #include "torch_xla/csrc/aten_xla_bridge.h"
 #include "torch_xla/csrc/helpers.h"
+#include "torch_xla/csrc/ops/dynamic_ir.h"
+#include "torch_xla/csrc/ops/expand.h"
+#include "torch_xla/csrc/ops/ops.h"
 #include "torch_xla/csrc/torch_util.h"
 #include "torch_xla_test.h"
 
@@ -4890,6 +4893,37 @@ TEST_F(AtenXlaTensorTest, TestExpandSymIntStatic) {
         c10::SymIntArrayRef({c10::SymInt(2), c10::SymInt(3), c10::SymInt(4)}),
         /*implicit=*/false);
     AllClose(b, xla_b);
+  });
+
+  ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::expand_symint", cpp_test::GetIgnoredCounters());
+}
+
+TEST_F(AtenXlaTensorTest, TestExpandSymIntSymbolic) {
+  torch::Tensor a = torch::ones({3, 4}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor b = a.expand({2, 3, 4}, /*implicit=*/false);
+
+  // create a symbolic symint with value 2.
+  // TODO: use `non_zero` to creat real dynamic dimension
+  torch::lazy::Value scalar_value =
+      torch::lazy::Value(ScalarOp(1.0, xla::F32), 0);
+  std::vector<int64_t> target_size = {2, 3, 5};
+  torch::lazy::NodePtr expand_node =
+      torch::lazy::MakeNode<Expand>(scalar_value, target_size);
+  torch::lazy::Value expand_value = torch::lazy::Value(expand_node, 0);
+  torch::lazy::NodePtr size_node =
+      torch::lazy::MakeNode<SizeNode>(expand_value, /*dim=*/0);
+  auto symint_node = c10::make_intrusive<XLASymIntNodeImpl>(size_node);
+  // This is not a dynamic size from xla perspective but it is a symint that
+  // wraps around a SizeNode instead of a scalar.
+  c10::SymInt dynamic_symint = symint_node->toSymInt();
+
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = xla_a.expand_symint(
+        c10::SymIntArrayRef({dynamic_symint, c10::SymInt(3), c10::SymInt(4)}),
+        /*implicit=*/false);
+    EXPECT_EQ(ToCpuTensor(xla_b).sum().item().toInt(), 24);
   });
 
   ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -4903,8 +4903,8 @@ TEST_F(AtenXlaTensorTest, TestExpandSymIntSymbolic) {
   torch::Tensor a = torch::ones({3, 4}, torch::TensorOptions(torch::kFloat));
   torch::Tensor b = a.expand({2, 3, 4}, /*implicit=*/false);
 
-  // create a symbolic symint with value 2.
-  // TODO: use `non_zero` to creat real dynamic dimension
+  // create a symbolic symint with value 2 and upperbound 2. This symint is
+  // symbolic(since it wraps around a sizeNode) but not really dynamic.
   torch::lazy::Value scalar_value =
       torch::lazy::Value(ScalarOp(1.0, xla::F32), 0);
   std::vector<int64_t> target_size = {2, 3, 5};
@@ -4916,6 +4916,36 @@ TEST_F(AtenXlaTensorTest, TestExpandSymIntSymbolic) {
   auto symint_node = c10::make_intrusive<XLASymIntNodeImpl>(size_node);
   // This is not a dynamic size from xla perspective but it is a symint that
   // wraps around a SizeNode instead of a scalar.
+  c10::SymInt dynamic_symint = symint_node->toSymInt();
+
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = xla_a.expand_symint(
+        c10::SymIntArrayRef({dynamic_symint, c10::SymInt(3), c10::SymInt(4)}),
+        /*implicit=*/false);
+    EXPECT_EQ(ToCpuTensor(xla_b).sum().item().toInt(), 24);
+    AllClose(b, xla_b);
+  });
+
+  ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::expand_symint", cpp_test::GetIgnoredCounters());
+}
+
+TEST_F(AtenXlaTensorTest, TestExpandSymIntDynamic) {
+  torch::Tensor a = torch::ones({3, 4}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor b = a.expand({2, 3, 4}, /*implicit=*/false);
+
+  // use non_zero to create a symbolic symint with upper bound 100 and real
+  // value 2
+  int64_t num_non_zero_element = 2;
+  int64_t num_row = 10;
+  int64_t num_col = 10;
+  torch::lazy::NodePtr nonzero_node =
+      CreateNonZeroNode2d(num_non_zero_element, num_row, num_col);
+
+  torch::lazy::NodePtr size_node_nonzero_0 =
+      torch::lazy::MakeNode<SizeNode>(nonzero_node, 0);
+  auto symint_node = c10::make_intrusive<XLASymIntNodeImpl>(size_node_nonzero_0);
   c10::SymInt dynamic_symint = symint_node->toSymInt();
 
   ForEachDevice([&](const torch::Device& device) {
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -4924,6 +4924,7 @@ TEST_F(AtenXlaTensorTest, TestExpandSymIntSymbolic) {
         c10::SymIntArrayRef({dynamic_symint, c10::SymInt(3), c10::SymInt(4)}),
         /*implicit=*/false);
     EXPECT_EQ(ToCpuTensor(xla_b).sum().item().toInt(), 24);
+    AllClose(b, xla_b);
   });
 
   ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1145,19 +1145,19 @@ at::Tensor XLANativeFunctions::expand_symint(const at::Tensor& self,
                                              at::SymIntArrayRef sym_size,
                                              bool implicit) {
   XLA_FN_COUNTER("xla::");
-  auto size = c10::asIntArrayRefSlow(sym_size);
-  return bridge::AtenFromXlaTensor(XLATensor::expand(
-      bridge::GetXlaTensor(self), torch::lazy::ToVector<int64_t>(size)));
-}
+  c10::optional<at::IntArrayRef> size = c10::asIntArrayRefSlowOpt(sym_size);
+  if (size.has_value()) {
+    return bridge::AtenFromXlaTensor(XLATensor::expand(
+        bridge::GetXlaTensor(self), torch::lazy::ToVector<int64_t>(*size)));
+  } else {
+    // at least one of the dimension is symbolic, use the sym_int version of the
+    // node
 
-at::Tensor XLANativeFunctions::expand_symint(const at::Tensor& self,
-                                             c10::SymIntArrayRef size,
-                                             bool implicit) {
-  XLA_FN_COUNTER("xla::");
-  torch::lazy::Shape shape =
-      torch::lazy::compute_shape_expand(self, size, implicit)[0];
-  return bridge::AtenFromXlaTensor(
-      XLATensor::expand_symint(bridge::GetXlaTensor(self), size, shape));
+    // torch::lazy::Shape shape =
+    //     torch::lazy::compute_shape_expand(self, size, implicit)[0];
+    return bridge::AtenFromXlaTensor(
+        XLATensor::expand_symint(bridge::GetXlaTensor(self), sym_size));
+  }
 }
 
 at::Tensor& XLANativeFunctions::exponential_(
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1166,10 +1166,12 @@ at::Tensor XLANativeFunctions::expand_symint(const at::Tensor& self,
       size_elements.SetUpperBound(idx, padded_self[idx]);
     }
   }
-  std::vector<torch::lazy::Shape> dynamic_shapes_ = torch::lazy::compute_shape_expand(self, size, implicit);
+  std::vector<torch::lazy::Shape> dynamic_shapes_ =
+      torch::lazy::compute_shape_expand(self, size, implicit);
   return bridge::AtenFromXlaTensor(XLATensor::expand_symint(
       bridge::GetXlaTensor(self), size_elements.GetNodes(),
-      size_elements.GetUpperBounds(), size_elements.GetDynamicDims(), dynamic_shapes_[0]));
+      size_elements.GetUpperBounds(), size_elements.GetDynamicDims(),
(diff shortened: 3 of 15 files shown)


test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -4658,6 +4658,20 @@ TEST_F(AtenXlaTensorTest, TestExpandAs) {
...
+  torch::Tensor a = torch::rand({3, 4}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor b = a.expand({2, 3, 4}, /*implicit=*/false);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = xla_a.expand_symint(
+        c10::SymIntArrayRef({c10::SymInt(2), c10::SymInt(3), c10::SymInt(4)}),
+        /*implicit=*/false);
+    AllClose(b, xla_b);
+    ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
+    ExpectCounterChanged("xla::expand_symint", cpp_test::GetIgnoredCounters());
```
JackCaoG, line 4671: nit, move this out of `ForEachDevice`, we only need to check it once for all device.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1306,6 +1306,30 @@ at::Tensor XLANativeFunctions::expand(const at::Tensor& self,
       bridge::GetXlaTensor(self), torch::lazy::ToVector<int64_t>(size)));
 }
 
+at::Tensor XLANativeFunctions::expand_symint(const at::Tensor& self,
+                                             c10::SymIntArrayRef size,
+                                             bool implicit) {
+  XLA_FN_COUNTER("xla::");
+  SymIntElements size_elements = SymIntElements(size);
+  // Replace -1 concrete int dim with the true shape value
+  std::vector<c10::SymInt> _sizes = torch::lazy::ToVector<c10::SymInt>(size);
```
JackCaoG, line 1315: I don't think we use `_xxx` naming convention, we only used `xxx_` for class private members. I think it is best to keep the new code align with old codes style. :D

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1306,6 +1306,30 @@ at::Tensor XLANativeFunctions::expand(const at::Tensor& self,
...
+  int64_t num_new_dimensions = _sizes.size() - self.dim();
+  std::vector<int64_t> padded_self(num_new_dimensions, 0);
+  padded_self.insert(padded_self.end(), self.sizes().begin(),
+                     self.sizes().end());
+  for (const auto idx : c10::irange(_sizes.size())) {
+    if (!_sizes[idx].is_symbolic() && _sizes[idx].expect_int() == -1) {
```
JackCaoG, line 1321: can you add a comment explaining what `_sizes[idx].expect_int()` means?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1306,6 +1306,30 @@ at::Tensor XLANativeFunctions::expand(const at::Tensor& self,
...
+      size_elements.SetUpperBound(idx, padded_self[idx]);
+    }
+  }
+  std::vector<torch::lazy::Shape> dynamic_shapes_ =
+      torch::lazy::compute_shape_expand(self, size, implicit);
+  return bridge::AtenFromXlaTensor(XLATensor::expand_symint(
+      bridge::GetXlaTensor(self), size_elements.GetNodes(),
+      size_elements.GetUpperBounds(), size_elements.GetDynamicDims(),
```
JackCaoG, line 1329: question: any reason not to pass `size_elements` directly?

torch_xla/csrc/data_ops.h
```diff
@@ -25,6 +25,10 @@ std::vector<int64_t> GetCompleteShape(absl::Span<const int64_t> output_sizes,
 // output size.
 xla::XlaOp BuildView(xla::XlaOp input, absl::Span<const int64_t> output_sizes);
 
+// Update Output Dynamic Dimensions based on input size()
```
JackCaoG, line 28: this comment is somewhat confusing, the parameter is called `output_sizes` and the comment refer to `input size()`. Maybe just something along the line of `Return a new XlaOp that reflects pass in size`?

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input,
+                           const std::vector<int64_t> upper_bounds,
+                           const std::vector<bool> dynamic_dims) {
+  return xla::ShapeUtil::MakeShape(GetXlaShape(input).element_type(),
+                                   {upper_bounds}, {dynamic_dims});
+}
+
+std::vector<torch::lazy::Value> GetValues(
```
JackCaoG, line 35: maybe rename to `GetOperandList` which matches similar function in https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/amp_foreach_non_finite_check_and_unscale.cpp#L24

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "torch/csrc/lazy/core/helpers.h"
+#include "torch/csrc/lazy/core/util.h"
+#include "torch_xla/csrc/data_ops.h"
+#include "torch_xla/csrc/lowering_context.h"
+#include "torch_xla/csrc/ops/infer_output_shape.h"
+
+namespace torch_xla {
+namespace {
+
+std::vector<xla::Shape> GetShapes(const torch::lazy::Value& input,
```
JackCaoG, line 15: Maybe `GetOperandShapes`?

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+xla::Shape NodeOutputShape(const torch::lazy::Value& input,
+                           const std::vector<int64_t> upper_bounds,
+                           const std::vector<bool> dynamic_dims) {
+  return xla::ShapeUtil::MakeShape(GetXlaShape(input).element_type(),
+                                   {upper_bounds}, {dynamic_dims});
+}
+
+std::vector<torch::lazy::Value> GetValues(
+    const torch::lazy::Value& input,
+    const std::vector<torch::lazy::Value> dimensions) {
```
JackCaoG, line 37: either pass `dimensions` as `const &` or pass it as value (drop the const). First one might be a better choice. If you choose the second one, no need to make a new vector.

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+  return values;
+}
+
+}  // namespace
+
+ExpandDynamic::ExpandDynamic(const torch::lazy::Value& input,
+                             const std::vector<torch::lazy::Value>& dimensions,
+                             const std::vector<int64_t> upper_bounds,
+                             const std::vector<bool> dynamic_dims,
+                             const torch::lazy::Shape& dynamic_shapes)
```
JackCaoG, line 49: Maybe pass this as `vector<torch::lazy::Shape>` to match codegended ops

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+    : XlaNode(
+          torch::lazy::OpKind(at::aten::expand), GetValues(input, dimensions),
+          {dynamic_shapes},
+          [&]() { return NodeOutputShape(input, upper_bounds, dynamic_dims); },
+          /*num_outputs=*/1, torch::lazy::MHash(upper_bounds)),
+      dynamic_shapes_(dynamic_shapes),
```
JackCaoG, line 55: I think you can access `dynamic_shapes` from base class directly

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+  values.insert(values.begin(), input);
+  return values;
+}
+
+}  // namespace
+
+ExpandDynamic::ExpandDynamic(const torch::lazy::Value& input,
+                             const std::vector<torch::lazy::Value>& dimensions,
+                             const std::vector<int64_t> upper_bounds,
+                             const std::vector<bool> dynamic_dims,
```
JackCaoG, line 48: I think you can drop the const here, you will `std::move` them later anyway.

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+                             const torch::lazy::Shape& dynamic_shapes)
+    : XlaNode(
+          torch::lazy::OpKind(at::aten::expand), GetValues(input, dimensions),
+          {dynamic_shapes},
+          [&]() { return NodeOutputShape(input, upper_bounds, dynamic_dims); },
+          /*num_outputs=*/1, torch::lazy::MHash(upper_bounds)),
+      dynamic_shapes_(dynamic_shapes),
+      upper_bounds_(std::move(upper_bounds)),
+      dynamic_dims_(std::move(dynamic_dims)) {
+  shapes_ = GetShapes(input, upper_bounds_, dynamic_dims_);
```
JackCaoG, line 58: What's the reason of including `input` in this vector? The below logic starts from `shapes_[1]` anyway. I actually think we should not init `shapes_` as class member, we can just compute it during lowering time.

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+ExpandDynamic::ExpandDynamic(const torch::lazy::Value& input,
+                             const std::vector<torch::lazy::Value>& dimensions,
+                             const std::vector<int64_t> upper_bounds,
+                             const std::vector<bool> dynamic_dims,
+                             const torch::lazy::Shape& dynamic_shapes)
+    : XlaNode(
+          torch::lazy::OpKind(at::aten::expand), GetValues(input, dimensions),
+          {dynamic_shapes},
+          [&]() { return NodeOutputShape(input, upper_bounds, dynamic_dims); },
+          /*num_outputs=*/1, torch::lazy::MHash(upper_bounds)),
```
JackCaoG, line 54: don't we need to include `dynamic_dims` in hashes too?

torch_xla/csrc/ops/expand_dynamic.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+XlaOpVector ExpandDynamic::Lower(LoweringContext* loctx) const {
+  xla::XlaOp input = loctx->GetOutputOp(operand(0));
+  std::vector<xla::XlaOp> size_ops;
+  for (int i = 1; i < shapes_.size(); i++) {
+    xla::XlaOp size_op = xla::Zeros(loctx->builder(), shapes_[i]);
+    size_ops.push_back(size_op);
+  }
+  xla::XlaOp output =
+      SetDimensionSizes(BuildExpand(input, upper_bounds_), size_ops);
```
JackCaoG, line 69: I think this part of logic is incorrect. You are actually set the dynamic size of input to a bunch of `0`. I think you mix the 
1. real size value (result of `getDimensionSize` for example, always a scalar)
2. whether dimension is dynamic

instead you pass a bunch of different shapes of `0` to the `SetDimensionSizes`.


Think of
```
t1 = tensor[10, 10, <=10]
size0 = getDimensionSize(t1, 0)
size1 = getDimensionSize(t1, 1)
size2 = getDimensionSize(t1, 2)

t2 = [1]
t2.expand(t1.static_size)
t2.setDimensionSize(size0, 0)         --> unnecessary
t2.setDimensionSize(size1, 1)          --> unnecessary
t2.setDimensionSize(size2, 2)
```

If you want to avoid unnecessary `setDimensionSize`, check whether `size0`, `size1` or `size2` are real size Node or just s int wrapped by symint (In the real code you always wrap `size` in a symint.)

JackCaoG (changes requested):
Thanks! Check my comment regarding lower function, let me know what you think.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1306,6 +1306,30 @@ at::Tensor XLANativeFunctions::expand(const at::Tensor& self,
       bridge::GetXlaTensor(self), torch::lazy::ToVector<int64_t>(size)));
 }
 
+at::Tensor XLANativeFunctions::expand_symint(const at::Tensor& self,
+                                             c10::SymIntArrayRef size,
```
JackCaoG, line 1310: We aim to have all of the complicated logic in the Node instead of `aten_xla_type.cpp` since we want to codegen `aten_xla_type.cpp`.

we need to have Node with different size SymInt to has different hash, we can either
1. manually expand `Symint` to upper_bound + is_dynamic + Node
2. Implement a hash functions for `SymInt` that account the hash for above 3 componenets.


I think it is cleaner to have `aten_xla_type` to pass `c10::SymIntArrayRef` to `ExapndSymint` Node and handle all of the complicated logic in lowering. This way we can codegen this op later and handle all of the expand logic in place.

miladm: I agree. I see three solutions here:

+ move the logic to `ExpandSymInt`. Moving the entire logic to this calss poses a bit of challenge because I need to pass the `const std::vector<torch::lazy::Value>& dimensions` to the constructor initialize the corresponding `XlaNode`. `dimensions` is the IRValue object from upstream Node. 

+ move the logic to `tensor_methods`. This is not the cleanest approach, but the easiest.

+ Something in between: produce the dimensions IR value in `tensor_methods` and unpack the rest (i.e. upper_bound, is_dynamic) in `ExpandSymInt`

I prefer the last choice. Wdyt @JackCaoG?

torch_xla/csrc/ops/expand_dynamic.cpp
miladm: I'd say this it's optional. We have not found a concrete reason why it's needed.

torch_xla/csrc/ops/expand_dynamic.cpp
miladm: Let's discuss this offline.

torch_xla/csrc/ops/expand_dynamic.cpp
miladm: I may not need these fields if I bring the upstream computation into this class.

torch_xla/csrc/ops/expand_dynamic.cpp
miladm: What exactly are you referring to?

torch_xla/csrc/aten_xla_type.cpp
miladm: I can as discussed below.

torch_xla/csrc/aten_xla_type.cpp
miladm: The node is only useful in its' IR Value form. So, I decided to abstract the logic and make tensor_methods leaner. I will push the new change shortly.

torch_xla/csrc/ops/expand_dynamic.cpp
miladm: I addressed this note.

torch_xla/csrc/ops/expand_dynamic.cpp
miladm: I no longer need `GetShapes`

torch_xla/csrc/aten_xla_type.cpp
miladm: I addressed this as per our discussion offline.

torch_xla/csrc/data_ops.cpp
```diff
@@ -92,6 +92,18 @@ xla::XlaOp BuildView(xla::XlaOp input, absl::Span<const int64_t> output_sizes) {
   return XlaHelpers::DynamicReshape(input, complete_output_sizes);
 }
 
+xla::XlaOp SetDimensionSizes(xla::XlaOp input,
+                             absl::Span<const xla::XlaOp> output_sizes) {
+  XLA_CHECK_EQ(output_sizes.size(), XlaHelpers::ShapeOfXlaOp(input).rank());
+  for (int i = 0; i < output_sizes.size(); i++) {
+    xla::Shape dim_shape = XlaHelpers::ShapeOfXlaOp(output_sizes[i]);
+    if (dim_shape.is_dynamic()) {
```
JackCaoG, line 100: will this ever be true through? for example 
```
size = getDimenSize(tensor_a, 0);
```

even if dimension 0 of tensor_a is dynamic, size would still just be a scalar has a static shape. The value of it represent a dynamic shape but its own shape is static.

torch_xla/csrc/ops/expand_symint.cpp
```diff
@@ -0,0 +1,66 @@
...
+      dynamic_dims_(std::move(size_elements.GetDynamicDims())) {}
+
+XlaOpVector ExpandSymInt::Lower(LoweringContext* loctx) const {
+  xla::XlaOp input = loctx->GetOutputOp(operand(0));
+  std::vector<xla::XlaOp> size_ops;
+  for (int i = 1; i < operands().size(); i++) {
+    size_ops.push_back(loctx->GetOutputOp(operand(i)));
+  }
+  xla::XlaOp output =
+      SetDimensionSizes(BuildExpand(input, upper_bounds_), size_ops);
```
JackCaoG, line 54: I think you want to pass `dynamic_dims_` and use that to determine if you want to call `SetDimensionSizes`

torch_xla/csrc/data_ops.cpp
JackCaoG: Oh, I looked at your `SetSymIntNodeElements ` call, you can just check whether output_sizes is a `xla::Constant`. I think if it is a dynamic dimension it will be something else(need to verify..).

xla_native_functions.yaml
```diff
@@ -162,6 +162,7 @@ supported:
   - empty.SymInt
   - empty_strided
   - expand
+  - expand.SymInt
```
Krovatkin, line 165: hmm, this would probably break since we switched to a single overload for `expand`.

torch_xla/csrc/ops/expand_symint.cpp
```diff
@@ -0,0 +1,66 @@
...
+                           const std::vector<bool> dynamic_dims) {
+  return xla::ShapeUtil::MakeShape(GetXlaShape(input).element_type(),
+                                   {upper_bounds}, {dynamic_dims});
+}
+
+std::vector<torch::lazy::Value> GetValues(
+    const torch::lazy::Value& input,
+    const std::vector<torch::lazy::Value>& dimensions) {
+  std::vector<torch::lazy::Value> values = dimensions;
+  values.insert(values.begin(), input);
```
Krovatkin, line 26: this might be more efficient 

```cpp
std::vector<torch::lazy::Value> values;
values.reserve(dimensions.size() + 1); // to avoid reallocs
values.insert(dimensions.begin(), dimensions.end(), values.end();
```

torch_xla/csrc/data_ops.cpp
Krovatkin: I think you are right here. `output_sizes` could be either constants or `SizeNode`s.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1164,6 +1165,28 @@ XLATensorPtr XLATensor::expand(const XLATensorPtr& input,
       GetExpandDimensions(input_shape.get(), std::move(size))));
 }
 
+XLATensorPtr XLATensor::expand_symint(const XLATensorPtr& input,
+                                      c10::SymIntArrayRef size,
+                                      const torch::lazy::Shape& shape) {
```
Krovatkin, line 1170: @JackCaoG I'd appreciate if you could check this part a bit more closely.

JackCaoG:
expand with static symint now works out of the box. I will repurpose this pr to support dynamic symints.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1128,6 +1129,29 @@ XLATensorPtr XLATensor::expand(const XLATensorPtr& input,
   return output;
 }
 
+XLATensorPtr XLATensor::expand_symint(const XLATensorPtr& input,
+                                      c10::SymIntArrayRef sym_size) {
+  SymIntElements size_elements = SymIntElements(sym_size);
+  // // Replace -1 concrete int dim with the true shape value
```
Krovatkin, line 1135: can we nuke the commented out code?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1128,6 +1129,29 @@ XLATensorPtr XLATensor::expand(const XLATensorPtr& input,
...
+  // for (const auto idx : c10::irange(sizes_.size())) {
+  //   // Passing -1 as the size for a dimension means not changing the size of
+  //   // that dimension.
+  //   if (!sizes_[idx].is_symbolic() && sizes_[idx].expect_int() == -1) {
+  //     size_elements.SetUpperBound(idx, padded_input[idx]);
+  //   }
+  // }
+  XLATensorPtr output = input->CreateFrom(
+      torch::lazy::MakeNode<ExpandSymInt>(input->GetIrValue(), size_elements));
+  output->storage_ = input->Storage();
```
Krovatkin, line 1151: argh, do you guys always do this?

torch_xla/csrc/data_ops.cpp
```diff
@@ -92,6 +92,31 @@ xla::XlaOp BuildView(xla::XlaOp input, absl::Span<const int64_t> output_sizes) {
   return XlaHelpers::DynamicReshape(input, complete_output_sizes);
 }
 
+xla::XlaOp SetDimensionSizes(xla::XlaOp input,
+                             absl::Span<const xla::XlaOp> symbolic_output_sizes,
+                             std::vector<bool> dynamic_dims) {
+  // XLA_CHECK_EQ(output_sizes.size(), XlaHelpers::ShapeOfXlaOp(input).rank());
```
Krovatkin, line 98: nit: please nuke the commented out code

torch_xla/csrc/ops/expand_symint.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace {
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input,
+                           const std::vector<int64_t> upper_bounds,
+                           const std::vector<bool> dynamic_dims) {
+  return xla::ShapeUtil::MakeShape(GetXlaShape(input).element_type(),
+                                   {upper_bounds}, {dynamic_dims});
+}
+
+std::vector<torch::lazy::Value> GetValues(
```
Krovatkin, line 22: looks like this is a private (to this file method)? can we mark it `static` or put it in a anonymous namespace?

torch_xla/csrc/ops/expand_symint.cpp
```diff
@@ -0,0 +1,72 @@
...
+    const torch::lazy::Value& input,
+    const std::vector<torch::lazy::NodePtr>& dimensions) {
+  std::vector<torch::lazy::Value> values;
+  values.reserve(dimensions.size() + 1);
+  values.push_back(input);
+  for (torch::lazy::NodePtr dim : dimensions) {
+    if (dim != nullptr) {
```
Krovatkin, line 29: nit: `if (dim)` maybe?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1128,6 +1129,29 @@ XLATensorPtr XLATensor::expand(const XLATensorPtr& input,
...
+  //                     input->shape().get().dimensions().end());
+  // for (const auto idx : c10::irange(sizes_.size())) {
+  //   // Passing -1 as the size for a dimension means not changing the size of
+  //   // that dimension.
+  //   if (!sizes_[idx].is_symbolic() && sizes_[idx].expect_int() == -1) {
+  //     size_elements.SetUpperBound(idx, padded_input[idx]);
+  //   }
+  // }
+  XLATensorPtr output = input->CreateFrom(
+      torch::lazy::MakeNode<ExpandSymInt>(input->GetIrValue(), size_elements));
```
Krovatkin, line 1150: size_elements only contain dynamic dims?
it looks a bit convoluted that we need to remember which dimensions are dynamic and then apply them.
Do you think after my `SizeConstant` change we can just create values for static dimensions as well and just **directly** call `xla::SetDimensionSize` that should be able to support both static and dynamic expressions? 

@JackCaoG @miladm

Krovatkin (approved):
:shipit:

torch_xla/csrc/tensor_methods.cpp
JackCaoG: we do, it is something Jiewen added to unblock the ddp for us.

torch_xla/csrc/ops/expand_symint.cpp
JackCaoG: I can put it in anonymous namespace

torch_xla/csrc/tensor_methods.cpp
JackCaoG: so `symSizeElement` only going to store the symbolic size if it is actually dynamic. We don't want to call `setDimensionSize` with a `Constant` on static dimension since it is not very useful.

miladm: +1 to @JackCaoG's comment

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1152,9 +1152,19 @@ at::Tensor XLANativeFunctions::expand_symint(const at::Tensor& self,
                                              at::SymIntArrayRef sym_size,
                                              bool implicit) {
   XLA_FN_COUNTER("xla::");
-  auto size = c10::asIntArrayRefSlow(sym_size);
-  return bridge::AtenFromXlaTensor(XLATensor::expand(
-      bridge::GetXlaTensor(self), torch::lazy::ToVector<int64_t>(size)));
+  c10::optional<at::IntArrayRef> size = c10::asIntArrayRefSlowOpt(sym_size);
+  if (size.has_value()) {
```
vanbasten23, line 1156: It seems `size.has_value()` implies none of the dimensions is symbolic, right?

JackCaoG: right, then we can fall back to static expand

test/cpp/cpp_test_util.cpp
```diff
@@ -382,5 +386,47 @@ void TestBackward(
...
+  std::vector<int64_t> target_size = {num_row, num_col};
+  torch::lazy::NodePtr expand_node =
+      torch::lazy::MakeNode<Expand>(scalar_value, target_size);
+  torch::lazy::Value expand_value = torch::lazy::Value(expand_node, 0);
+  int64_t count = 0;
+  int64_t i = 0;
+  int64_t j = 0;
+  torch::lazy::Value slice_value = expand_value;
+  while (count++ < num_non_zero_element) {
+    std::vector<int64_t> base_indeies = {i, j++};
```
miladm, line 416: `base_indices`

torch_xla/csrc/ops/expand_symint.cpp
JackCaoG: oh lol, it is already in a anonymous namespace, check line 13

torch_xla/csrc/tensor_methods.cpp
JackCaoG: @miladm I will remove these `-1` handling logic for now. If we want to support it we should add corresponding cpp tests. I am still unsure whether pytorch will convert those -1 for us, my guess is no

test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -4896,6 +4899,69 @@ TEST_F(AtenXlaTensorTest, TestExpandSymIntStatic) {
...
 
+TEST_F(AtenXlaTensorTest, TestExpandSymIntSymbolic) {
+  torch::Tensor a = torch::ones({3, 4}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor b = a.expand({2, 3, 4}, /*implicit=*/false);
+
+  // create a symbolic symint with value 2 and upperbound 2. This symint is
+  // symbolic(since it wraps around a sizeNode) but not really dynamic.
+  torch::lazy::Value scalar_value =
+      torch::lazy::Value(ScalarOp(1.0, xla::F32), 0);
+  std::vector<int64_t> target_size = {2, 3, 5};
```
miladm, line 4910: why is the last array element `5` and not `4`?

torch_xla/csrc/aten_xla_type.cpp
Krovatkin: yes, it's a pretty common idiom in PyTorch for asXXX to return std::nullopt if a conversion was unsuccessful

test/cpp/cpp_test_util.cpp
```diff
@@ -382,5 +386,47 @@ void TestBackward(
...
+  // i = 0
+  // j = 0
+  // count = 0
+  // while count < num_non_zero_element:
+  //   t1[i][j] = 1
+  //   j += 1
+  //   if j == num_col:
+  //     j = 0
+  //     i += 1
+  // res = t1.non_zero()
```
miladm, line 401: you want to add `count += 1`

test/cpp/test_aten_xla_tensor.cpp
JackCaoG: I think it is a typo.. I will fix it

torch_xla/csrc/data_ops.cpp
```diff
@@ -92,6 +92,25 @@ xla::XlaOp BuildView(xla::XlaOp input, absl::Span<const int64_t> output_sizes) {
...
+                             absl::Span<const xla::XlaOp> symbolic_output_sizes,
+                             std::vector<bool> dynamic_dims) {
+  size_t current_output_size_index = 0;
+  size_t symbolic_output_sizes_len = symbolic_output_sizes.size();
+  for (size_t i = 0; i < dynamic_dims.size(); i++) {
+    if (dynamic_dims[i]) {
+      // Dimension i is dynamic
+      XLA_CHECK_LT(current_output_size_index, symbolic_output_sizes_len);
+      input = xla::SetDimensionSize(
+          input, symbolic_output_sizes[current_output_size_index++], i);
```
miladm, line 105: In what scenario `current_output_size_index` and `i` deviate in value?

torch_xla/csrc/ops/expand_symint.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+std::vector<torch::lazy::Value> GetValues(
+    const torch::lazy::Value& input,
+    const std::vector<torch::lazy::NodePtr>& dimensions) {
+  std::vector<torch::lazy::Value> values;
+  values.reserve(dimensions.size() + 1);
+  values.push_back(input);
+  for (torch::lazy::NodePtr dim : dimensions) {
+    if (dim) {
+      // Dimension Node only exist for symbolic dimension.
```
miladm, line 30: @JackCaoG is this due a different PR you recently pushed while I was away?

torch_xla/csrc/data_ops.cpp
JackCaoG: static dimension. `symbolic_output_sizes` only contains dynamic dimensions, where `dynamic_dims` is a boolean array that also includes static dimensions

miladm (commented):
Thanks for helping make all the changes @JackCaoG. Looks super close to final. Left a bunch of comments. Will take another pass this evening.

torch_xla/csrc/ops/expand_symint.cpp
JackCaoG: I would recommend checking https://github.com/pytorch/xla/pulls?q=is%3Apr+author%3AJackCaoG+label%3Adynamism . It is one of those pr I added to refactor the symintelements.

miladm: I see this [PR](https://github.com/pytorch/xla/pull/3944) covering the topic. Thanks.

torch_xla/csrc/data_ops.cpp
miladm: Yup, this [PR](https://github.com/pytorch/xla/pull/3944) covers the context on why we may deviate.

miladm:
All tests seem to be passing and approval is in. I can't think of more feedback now. @JackCaoG?

torch_xla/csrc/data_ops.cpp
Krovatkin: @miladm hehe I did warn @JackCaoG it could get confusing (discussion here: https://github.com/pytorch/xla/pull/3889/files#r975882060 ) 

There's quite a bit of logic that tries really hard to separate dynamic dimensions from static ones whereas the only thing we really seem to care about is if there's at least one dynamic dimension (so we need to use `ExpandSymInt::Lower`) assuming of course XLA can optimize `xla::SetDimensionSize(Scalar(X))` *wink* *wink*

JackCaoG (approved):

miladm:
Merging the PR now.

PR #3902, opened 2022-08-18
Set PJRT device assignment in order of device ID

### LINKED ISSUE #3824: Mismatched rank in collective ops (all-gather, reduce-scatter, and all-to-all) in PJRT runtime
## 🐛 Bug

For the PJRT runtime, currently `xm.all_reduce` works well on v3-8 after https://github.com/pytorch/xla/pull/3704 is merged, but `xm.reduce_scatter` and `xm.all_to_all` still still do not work well, and the `xm.all_gather` under "pin_layout=False" doesn't work well (`all_gather` under "pin_layout=True" is actually [using `all_reduce`](https://github.com/pytorch/xla/blob/ff2e0ea559b5f8785c7f577834af2b5af84b6849/torch_xla/core/xla_model.py#L608-L647)).

This issue is submitted based on the discussions in https://github.com/pytorch/xla/pull/3813#issuecomment-1204215319.

## To Reproduce

The example below contains a test case for each collective op (all_gather, reduce_scatter, and all_to_all), which doesn't fully work under PJRT for now.

1. Allocate a new v3-8 TPU VM with `tpu-vm-pt-1.12` runtime and install the nightly PyTorch/XLA wheels that contain https://github.com/pytorch/xla/pull/3704
```
# torch, torchvision and torch_xla
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch-nightly+20220802-cp38-cp38-linux_x86_64.whl
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torchvision-nightly+20220802-cp38-cp38-linux_x86_64.whl
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch_xla-nightly+20220802-cp38-cp38-linux_x86_64.whl

# libtpu
sudo pip3 install https://storage.googleapis.com/cloud-tpu-tpuvm-artifacts/wheels/libtpu-nightly/libtpu_nightly-0.1.dev20220518-py3-none-any.whl
```
2. Save the following content to a file `test_pjrt_collective_ops.py`
```python3
import torch


def test_pjrt_collective_ops():
    rank = xm.get_ordinal()
    device = xm.xla_device()
    pin_layout = False

    xm.mark_step()

    # all-gather -- expected output for rank i on v3-8:
    #   [0, 1, 2, 3, 4, 5, 6, 7]
    t1 = (torch.ones(1) * rank).to(device)
    all_gather_out = xm.all_gather(t1, dim=0, pin_layout=pin_layout)

    # reduce-scatter -- expected output for rank i on v3-8:
    #   [-i]
    t2 = -torch.arange(world_size, dtype=torch.float32).to(device)
    reduce_scatter_out = xm.reduce_scatter(
        xm.REDUCE_SUM,
        t2,
        scale=1.0 / world_size,
        scatter_dim=0,
        shard_count=world_size,
        pin_layout=pin_layout,
    )

    # all-to-all -- expected output for rank i on v3-8:
    #   [[[-i, -i, -i, -i, -i, -i, -i, -i],
    #       0,  1,  2,  3,  4,  5,  6,  7]]]
    t3 = torch.cat(
        [
            -torch.arange(world_size, dtype=torch.float32).view(-1, 1, 1),
            torch.ones(world_size, 1, 1) * rank,
        ],
        dim=1,
    ).to(device)
    all_to_all_out = xm.all_to_all(
        t3,
        split_dimension=0,
        concat_dimension=2,
        split_count=world_size,
        pin_layout=pin_layout,
    )

    xm.mark_step()
    print(
        f"\n[rank {rank} of {world_size} (pin_layout={pin_layout})]\n"
        f"* all_gather_out:\n{all_gather_out}\n"
        f"* reduce_scatter_out:\n{reduce_scatter_out}\n"
        f"* all_to_all_out:\n{all_to_all_out}\n",
        end="",
        flush=True,
    )


if __name__ == "__main__":
    pjrt.run_multiprocess(test_pjrt_collective_ops)
```
3. Run this file under PJRT via `PJRT_DEVICE=TPU python3 test_pjrt_collective_ops.py`, which shows that these 3 collective ops have problematic outputs that do not match their rank from `xm.get_ordinal()`, printing the following
```
[rank 0 of 8 (pin_layout=False)]
* all_gather_out:
tensor([0., 1., 4., 5., 6., 7., 2., 3.], device='xla:0')
* reduce_scatter_out:
tensor([0.], device='xla:0')
* all_to_all_out:
tensor([[[0., -0., -0., -0., -0., -0., -0., -0.],
         [0., 1., 4., 5., 6., 7., 2., 3.]]], device='xla:0')

[rank 1 of 8 (pin_layout=False)]
* all_gather_out:
tensor([0., 1., 4., 5., 6., 7., 2., 3.], device='xla:1')
* reduce_scatter_out:
tensor([-1.], device='xla:1')
* all_to_all_out:
tensor([[[-1., -1., -1., -1., -1., -1., -1., -1.],
         [ 0.,  1.,  4.,  5.,  6.,  7.,  2.,  3.]]], device='xla:1')

[rank 2 of 8 (pin_layout=False)]
* all_gather_out:
tensor([0., 1., 4., 5., 6., 7., 2., 3.], device='xla:0')
* reduce_scatter_out:
tensor([-6.], device='xla:0')
* all_to_all_out:
tensor([[[-6., -6., -6., -6., -6., -6., -6., -6.],
         [ 0.,  1.,  4.,  5.,  6.,  7.,  2.,  3.]]], device='xla:0')

[rank 3 of 8 (pin_layout=False)]
* all_gather_out:
tensor([0., 1., 4., 5., 6., 7., 2., 3.], device='xla:1')
* reduce_scatter_out:
tensor([-7.], device='xla:1')
* all_to_all_out:
tensor([[[-7., -7., -7., -7., -7., -7., -7., -7.],
         [ 0.,  1.,  4.,  5.,  6.,  7.,  2.,  3.]]], device='xla:1')

[rank 4 of 8 (pin_layout=False)]
* all_gather_out:
tensor([0., 1., 4., 5., 6., 7., 2., 3.], device='xla:0')
* reduce_scatter_out:
tensor([-2.], device='xla:0')
* all_to_all_out:
tensor([[[-2., -2., -2., -2., -2., -2., -2., -2.],
         [ 0.,  1.,  4.,  5.,  6.,  7.,  2.,  3.]]], device='xla:0')

[rank 5 of 8 (pin_layout=False)]
* all_gather_out:
tensor([0., 1., 4., 5., 6., 7., 2., 3.], device='xla:1')
* reduce_scatter_out:
tensor([-3.], device='xla:1')
* all_to_all_out:
tensor([[[-3., -3., -3., -3., -3., -3., -3., -3.],
         [ 0.,  1.,  4.,  5.,  6.,  7.,  2.,  3.]]], device='xla:1')

[rank 6 of 8 (pin_layout=False)]
* all_gather_out:
tensor([0., 1., 4., 5., 6., 7., 2., 3.], device='xla:0')
* reduce_scatter_out:
tensor([-4.], device='xla:0')
* all_to_all_out:
tensor([[[-4., -4., -4., -4., -4., -4., -4., -4.],
         [ 0.,  1.,  4.,  5.,  6.,  7.,  2.,  3.]]], device='xla:0')

[rank 7 of 8 (pin_layout=False)]
* all_gather_out:
tensor([0., 1., 4., 5., 6., 7., 2., 3.], device='xla:1')
* reduce_scatter_out:
tensor([-5.], device='xla:1')
* all_to_all_out:
tensor([[[-5., -5., -5., -5., -5., -5., -5., -5.],
         [ 0.,  1.,  4.,  5.,  6.,  7.,  2.,  3.]]], device='xla:1')
```

For example, in the output above, `reduce_scatter` seems to do the "reduce" part correctly, but the "scatter" part has a rank mismatching the rank from "xm.get_ordinal".

## Expected behavior

For each rank `i` (in 0, ..., 7 on v3-8), it should have the following output:
```
all-gather -- expected output for rank i on v3-8:
   [0, 1, 2, 3, 4, 5, 6, 7]
reduce-scatter -- expected output for rank i on v3-8:
   [-i]
all-to-all -- expected output for rank i on v3-8:
   [[[-i, -i, -i, -i, -i, -i, -i, -i],
       0,  1,  2,  3,  4,  5,  6,  7]]]
```

## Environment

 - Reproducible on XLA backend [CPU/TPU]: v3-8 TPU VM with `tpu-vm-pt-1.12` runtime
 - torch_xla version: nightly 20220802 (see above)

## Additional context

* Another problem is that `xm.rendezvous` is not working under PJRT yet -- it doesn't actually introduce a barrier across all ranks. I think once these issues are resolved (and https://github.com/pytorch/xla/pull/3813 is merged), it should be possible to port most XRT codebases to PJRT.
* The `xm.all_reduce` op seems to work well under PJRT, perhaps because it's not affected by the rank order.
* The examples above use `pin_layout=False`, but they should be mostly the same if one switches to `pin_layout=True` (note that `xm.all_gather` only performs an actual all-gather under `pin_layout=False`. Under `pin_layout=True`, the API `xm.all_gather` doesn't actually perform all-gather, but instead, it [first pads the inputs with zeros and then performs all_reduce](https://github.com/pytorch/xla/blob/ff2e0ea559b5f8785c7f577834af2b5af84b6849/torch_xla/core/xla_model.py#L673-L678)). 
* Also, it would be great to tuplify the input during compilation and execution to get rid of the size limit in https://github.com/pytorch/xla/issues/3453#issuecomment-1121490148 and allow training larger models in PyTorch/XLA.

cc: @will-cromar

will-cromar:
Fixes #3824

## Context

The relationship between the TPU option `TPU_VISIBLE_DEVICES` and PjRt's `device->id()` is not deterministic, and you may get a different order on each new TPU host. Unfortunately, we can't know the device ID for a value of `TPU_VISIBLE_DEVICES`, until we've already initialized the TPU client in the current process.

The [default device assignment order](https://github.com/tensorflow/tensorflow/blob/73faab182e0a0b77b63585f45409b1e91b816097/tensorflow/compiler/xla/pjrt/pjrt_client.h#L478-L481) given by PjRt is not always in order of device ID. This order determines the order of results for collectives.

## Changes

- When the TPU client is initialized, we can use the device ID of the device assigned to each replica to set the global and local ordinals.
- Since we can no longer use `local_ordinal` to find the PyTorch device index, lean on `torch_xla`'s internal thread-local default device to reduce the number of indices we have to juggle.
- Don't set the thread default device in `xla_device`.
- Add new collective unit tests based on examples in #3824 (thanks @ronghanghu!)
- Move broadcast test into `test_experimental_pjrt_tpu.py`.

## Notes

- This will make results consistent across different TPU hosts, but it may cause issues if someone relies on the process rank explicitly, e.g. if someone uses the `RANK` variable from torchrun or references `CLOUD_TPU_TASK_ID`.
- Tested manually on v4-8 and v3-32. I saw a potential small (<2%) performance impact on ResNet50 on v4-8 with fake data. If this sticks in our automated tests, I can follow up. However, I think consistency with XRT should be the first priority.
-  The return value for `pjrt.run_multiprocess` (`{process_rank: {thread_rank: return_val}}`) doesn't make sense anymore now that process rank has no little relationship to the actual ordinal of a replica. I will follow-up with a PR to either remove the return value (like `xmp.spawn`) or return the replica IDs instead.


JackCaoG:
Thanks, I will take a look today.

torch_xla/experimental/pjrt.py
```diff
@@ -99,14 +99,14 @@ def xla_device(n: Optional[int] = None,
...
+
   devices = xm.get_xla_supported_devices(devkind=devkind)
-  device_index = n or (local_ordinal(default=0) % addressable_device_count())
-  if device_index > len(devices):
-    raise IndexError('Device index {} out of range in {}'.format(
-        device_index, devices))
+  if n > len(devices):
+    raise IndexError('Device index {} out of range in {}'.format(n, devices))
 
-  torch_xla._XLAC._xla_set_default_device(devices[device_index])
```
JackCaoG, line 108: any reason we remove this set default device logic?

JackCaoG (commented):
mostly lgtm , left some minor comments.

torch_xla/experimental/pjrt.py
will-cromar: I set the default device in `run_thread_per_device` already, so it's not necessary to do it again here. IMO it's not intuitive for `xla_device` to more than return the _n_th or default device and actually affect thread-local state.

JackCaoG: was this line introduced by the pjrt or it existed in the xrt world too?

will-cromar: It exists in XRT too:

https://github.com/pytorch/xla/blob/1715ccd0c11553c0e7ab6550715ae83769165b55/torch_xla/core/xla_model.py#L252

JackCaoG: I would say leave it as it is otherwise we might need to deal with weird xrt error(unless it is something blocking pjrt). If we want to refactor this from xrt it should be in a separate pr with corresponding test.

JackCaoG: `defualt_device` is useful in a lot of c++ code as well, we don't want to ask user to manually set default device so adding it to `xla_device` call is a compromise.

will-cromar: Done

JackCaoG (approved):
Thanks!

JackCaoG:
Rebase should solve the build issue.

PR #3715, opened 2022-07-14
Extend `nonzero` to provide dynamism to its `torch::lazy::shape` member

miladm:

PR #3909, opened 2022-08-19
Add `sym_sizes` with dynamic shape support

miladm:
Add `sym_sizes` with dynamic shape support

### DIFF
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -279,6 +279,16 @@ TEST_F(AtenXlaTensorTest, TestSubScalarInPlace) {
   });
 }
 
+TEST_F(AtenXlaTensorTest, TestSymSizes) {
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor a = torch::rand({2, 3}, torch::TensorOptions(torch::kFloat));
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor c = a.sym_sizes(0);
+    AllClose(a.sym_sizes(0), 2);
+    AllClose(a.sym_sizes(0).is_symbolic(), false);
+  });
+}
+
 TEST_F(AtenXlaTensorTest, TestMul) {
   torch::Tensor a = torch::rand({2, 2}, torch::TensorOptions(torch::kFloat));
   torch::Tensor b = torch::rand({2, 2}, torch::TensorOptions(torch::kFloat));
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -283,9 +283,8 @@ TEST_F(AtenXlaTensorTest, TestSymSizes) {
   ForEachDevice([&](const torch::Device& device) {
     torch::Tensor a = torch::rand({2, 3}, torch::TensorOptions(torch::kFloat));
     torch::Tensor xla_a = CopyToDevice(a, device);
-    torch::Tensor c = a.sym_sizes(0);
-    AllClose(a.sym_sizes(0), 2);
-    AllClose(a.sym_sizes(0).is_symbolic(), false);
+    ASSERT_EQ(a.sym_sizes().at(0).expect_int(), 2);
+    ASSERT_EQ(a.sym_sizes().at(0).is_symbolic(), false);
   });
 }
 
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -285,6 +285,20 @@ TEST_F(AtenXlaTensorTest, TestSymSizes) {
     torch::Tensor xla_a = CopyToDevice(a, device);
     ASSERT_EQ(a.sym_sizes().at(0).expect_int(), 2);
     ASSERT_EQ(a.sym_sizes().at(0).is_symbolic(), false);
+
+    torch::Tensor b = torch::tensor({{0.0, 1.0}, {0.0, 0.0}},
+                                    torch::TensorOptions(torch::kFloat));
+    torch::Tensor xla_b = CopyToDevice(b, device);
+    xla_b = torch::nonzero(xla_b);
+    auto s0 = xla_b.sym_sizes().at(0);
+    ASSERT_EQ(s0.is_symbolic(), true);
+    auto sininode =
+        dynamic_cast<XLASymIntNodeImpl*>(s0.toSymIntNodeImpl().get());
+    auto snode =
+        std::dynamic_pointer_cast<torch_xla::SizeNode>(sininode->node());
+    ASSERT_TRUE(snode);
+    ASSERT_EQ(snode->getStaticValue(), 4);
+    ASSERT_EQ(snode->getDynamicValue(), 1);
   });
 }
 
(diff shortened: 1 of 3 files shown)


miladm:
This [PR](https://github.com/pytorch/xla/pull/3715/files) shows that this implementation must be correct. However, there are other issues we want to resolves in the mentioned PR that may require extra time. This PR proposes the landing of the `sym_sizes` independently given all we know about its implementation methodology.

Krovatkin:
@JackCaoG I remember you had some doubts about this PR?

Krovatkin:
never mind, I'm seeing real failures now, need to investigate @JackCaoG

JackCaoG:
seems OK from high level, I am not sure what does this pr enabled us through. is it `.sym_size` for `XLATensorImpl`?

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -115,9 +119,8 @@ at::IntArrayRef XLATensorImpl::sizes_custom() const {
 }
 
 c10::SymIntArrayRef XLATensorImpl::sym_sizes_custom() const {
```
vanbasten23, line 121: Wondering what the "custom" in `sym_sizes_custom()` implies

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -172,6 +175,35 @@ void XLATensorImpl::SetupSizeProperties() {
...
+    for (auto i : c10::irange(rank)) {
+      if (tensor_->shape().get().is_dynamic_dimension(i)) {
+        auto dim_node = a.MakeSizeNode(tensor_->GetIrValue(), i);
+        auto symint_node =
+            c10::make_intrusive<torch::lazy::SymIntNodeImpl>(dim_node);
+        auto sn = symint_node->toSymInt();
+        sym_sizes_.push_back(sn);
+        /*TODO(miladm): verify numel_ calculation after adding a dynamic op
+         */
+        numel_ *= dynamic_cast<SizeNode*>(dim_node.get())->getStaticValue();
```
vanbasten23, line 197: I've read the definition of numel_ in the base class but it's not clear to me what it means. Do you know the meaning of this member?

JackCaoG: number of elements

torch_xla/csrc/tensor_impl.cpp
Krovatkin: `custom` means that a given `TensorImpl` can implement `sym_sizes` in whichever way they see fit. Note, `sym_sizes_custom` is a virtual method whereas `sym_sizes()` isn't. `sym_sizes()` checks if a given tensor impl implements `CustomSizes` policy and if so calls a virtual sym_sizes. This is kinda convoluted and was done to make sure we don't pay a virtual call penalty on the fast path: cpu and cuda tensors don't need the custom_sym_sizes implementation.

Krovatkin:
> seems OK from high level, I am not sure what does this pr enabled us through. is it .sym_size for XLATensorImpl?

This PR finally opens the Pandora box, hehe. For ops like `nonzero` size() or shape in python will now return symints and for any code in C++ that calls `sym_sizes()`.

test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -276,6 +276,15 @@ TEST_F(AtenXlaTensorTest, TestSubScalarInPlace) {
   });
 }
 
+TEST_F(AtenXlaTensorTest, TestSymSizes) {
```
JackCaoG, line 279: nit, `TestSymSizesStatic`?

test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -276,6 +276,15 @@ TEST_F(AtenXlaTensorTest, TestSubScalarInPlace) {
...
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor a = torch::rand({2, 3}, torch::TensorOptions(torch::kFloat));
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    ASSERT_EQ(a.sym_sizes().at(0).expect_int(), 2);
+    ASSERT_EQ(a.sym_sizes().at(0).is_symbolic(), false);
+  });
+}
+
```
JackCaoG, line 287: should we add a dynamic test? You can call `non_zero` but you would need to make sure https://github.com/pytorch/xla/blob/master/test/cpp/run_tests.sh#L11 is set so `non_zero` won't fallback to cpu.

torch_xla/csrc/tensor_impl.cpp
vanbasten23: Thanks

torch_xla/csrc/tensor_impl.cpp
vanbasten23: Thanks

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -168,10 +172,31 @@ void XLATensorImpl::SetupSizeProperties() {
     for (int i = 0; i < updated_strides.size(); i++) {
       sizes_and_strides_.stride_at_unchecked(i) = updated_strides[i];
     }
+    SetupSymSizeProperties();
     generation_ = generation;
```
miladm, line 176: Ditto

test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -279,6 +279,29 @@ TEST_F(AtenXlaTensorTest, TestSubScalarInPlace) {
...
+    ASSERT_EQ(a.sym_sizes().at(0).expect_int(), 2);
+    ASSERT_EQ(a.sym_sizes().at(0).is_symbolic(), false);
+
+    torch::Tensor b = torch::tensor({{0.0, 1.0}, {0.0, 0.0}},
+                                    torch::TensorOptions(torch::kFloat));
+    torch::Tensor xla_b = CopyToDevice(b, device);
+    xla_b = torch::nonzero(xla_b);
+    auto s0 = xla_b.sym_sizes().at(0);
+    ASSERT_EQ(s0.is_symbolic(), true);
+    auto sininode =
```
miladm, line 295: nit: is the variable name meant to be `sininode`?

test/cpp/test_aten_xla_tensor.cpp
miladm: Name is ok now given it tests both dynamic and static `SymSizes`

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -168,10 +172,31 @@ void XLATensorImpl::SetupSizeProperties() {
     for (int i = 0; i < updated_strides.size(); i++) {
```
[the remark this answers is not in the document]
Krovatkin: if I understand your point, you are suggesting to duplicate the strides logic inside `SetupSymSizeProperties` as well, so it updates both sym_sizes and sym_strides the same way `SetupSizeProperties`

If this is your concern, I believe it's fine for now, since `sym_strides` is still returning static strides now.
We should figure out if we need support for dynamic strides as well or whether it's okay for sym_strides to continue returning static strides.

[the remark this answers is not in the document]
miladm: @ezyang: based on our offline conversation about `sym_strides` under functionalization, it seems we assume no need for dynamic strides; correct?

JackCaoG (approved):

PR #3912, opened 2022-08-20
Fix upstream break 83586

JackCaoG:
This is to fix https://github.com/pytorch/pytorch/pull/83586


JackCaoG:
Force merge this pr to unblock master.

PR #3914, opened 2022-08-22
Companion PR for https://github.com/pytorch/pytorch/pull/83628

ezyang:


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -3199,7 +3195,8 @@ at::Tensor XLANativeFunctions::new_empty_strided(
 }
 
 at::Tensor XLANativeFunctions::narrow_copy(const at::Tensor& self, int64_t dim,
-                                           int64_t start, int64_t length) {
+                                           c10::SymInt start,
```
Krovatkin, line 3198: OoO, why did we also change `start` to `SymInt`. Do we have to switch all int arguments to `SymInt` for symint kernels?

Krovatkin (commented):
@JackCaoG this is the change I mentioned last dynamic shapes sync, Ed was going to make. It's pretty boiler-platy.

torch_xla/csrc/aten_xla_type.cpp
ezyang: No. But start should be a SymInt as it is a "size-y" int. Notice that dim doesn't get changed.

JackCaoG (approved):
lgtm, thanks!

PR #3923, opened 2022-08-24
Add lowring for floor_divide

ymwangg:
This op is seen in DETR model. https://pytorch.org/docs/stable/generated/torch.floor_divide.html.

### DIFF
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -3011,6 +3011,32 @@ TEST_F(AtenXlaTensorTest, TestFloor) {
   });
 }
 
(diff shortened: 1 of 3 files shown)


JackCaoG (approved):
It is one of those ops that will be deprecated but I guess we need to lower it for the time being.

JackCaoG:
seems like op test is failing, this test used to always fallback to CPU or passed. after this pr floor_divide will be run on xla device and had some mismatch behavior compared to pytorch one.

ymwangg:
Ok, I see what's wrong here. The pytorch doc I checked (1.12 stable release) is outdated and pytorch master has changed the rounding mode from "trunc" to "floor". I've fixed this issue and updated the C++ tests.

JackCaoG (approved):
Thanks!

PR #3929, opened 2022-08-24
[PjRt] Implement `xm.rendezvous` for TPU v4.

will-cromar:
Use `torch.distributed` to implement `xm.rendezvous` and replace the XRT mesh server. Two caveats:

- `tag` is effectively ignored. Processes will join collective operations in the order that they are reached.
- Each replica must call into `xm.rendezvous`, even if they are not part of the rendezvous, as required by [`torch.distributed.new_group`](https://pytorch.org/docs/stable/distributed.html#torch.distributed.new_group). I confirmed with @JackCaoG that rendezvous with all other replicas is by far the most common case.

`xm.rendezvous` will probably not give the expected results on TPU v3, since we have multiple replicas per process. I will find a workaround in a future PR.

### DIFF
--- a/test/pjrt/test_mesh_service.py
+++ b/test/pjrt/test_mesh_service.py
@@ -1,0 +1,50 @@
+import itertools
+import os
+from typing import List, Optional
+from absl.testing import absltest, parameterized
+
+import torch.distributed as dist
+import torch_xla.core.xla_model as xm
+from torch_xla.experimental import pjrt
+
+
+class PjRtMeshServiceTest(parameterized.TestCase):
+
+  @staticmethod
+  def _rendezvous_default(xrt_mesh_addr: Optional[str], replicas: List[int]):
+    if xrt_mesh_addr:
+      os.environ['XRT_MESH_SERVICE_ADDRESS'] = xrt_mesh_addr
+
+    payload = b'message %d' % xm.get_ordinal()
+    return xm.get_ordinal(), xm.rendezvous("test rendezvous", payload, replicas)
+
+  @parameterized.named_parameters(
+      ('defaults', None, []), ('xrt_address', 'localhost:9477', []),
+      ('four_replicas', None, [0, 1, 2, 3]), ('two_replicas', None, [0, 1]))
+  def test_rendezvous(self, xrt_mesh_addr, replicas):
+    results = pjrt.run_multiprocess(self._rendezvous_default, xrt_mesh_addr,
+                                    replicas)
+    values = list(
+        itertools.chain.from_iterable(row.values() for row in results.values()))
+    replicas = replicas or list(range(len(values)))
+
+    for ordinal, value in values:
+      if ordinal in replicas or not replicas:
+        self.assertEqual(value, [b'message %d' % r for r in replicas])
+
+  @staticmethod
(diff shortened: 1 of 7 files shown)


JackCaoG:
rebase should fix the build issue

test/pjrt/test_mesh_service.py
```diff
@@ -0,0 +1,50 @@
...
+
+    for ordinal, value in values:
+      if ordinal in replicas or not replicas:
+        self.assertEqual(value, [b'message %d' % r for r in replicas])
+
+  @staticmethod
+  def _mesh_reduce():
+    return xm.mesh_reduce('test mesh reduce', xm.get_ordinal(), sum)
+
+  def test_mesh_reduce(self):
```
JackCaoG, line 39: Would mesh_reduce just work? I thought we need to implement something else given that we don't use mesh server in PJRT world.

torch_xla/experimental/pjrt.py
```diff
@@ -224,3 +226,42 @@ def broadcast_master_param(model: nn.Module) -> None:
...
+  Returns:
+    List of bytes from other replicas.
+  """
+  if not dist.is_initialized():
+    logging.warning(
+        'Default process group not initialized. Creating XLA process group...')
+    mesh_master = xu.getenv_as("XRT_MESH_SERVICE_ADDRESS", str)
+
+    import torch_xla.distributed.xla_backend
+    dist.init_process_group(
```
JackCaoG, line 253: would this conflict with ddp in any way?

JackCaoG (commented):
mostly lgtm

test/pjrt/test_mesh_service.py
will-cromar: I was surprised this just works too. `mesh_reduce` uses `rendezvous` to communicate serialized objects between replicas:

https://github.com/pytorch/xla/blob/227ac13d7102f4e30540df7ef50d09a02a985290/torch_xla/core/xla_model.py#L1118

torch_xla/experimental/pjrt.py
will-cromar: It could cause an issue if someone were to run rendezvous before they tried to initialize the main process group with another configuration. Otherwise, I added a check here to make sure we don't initialize the process group twice.

JackCaoG: hmm on a second thought, this backend is gloo, while on dpp we use the `xla_backend`. In a ddp setup I would expect two of them not conflict with each other?

JackCaoG (approved):
Thanks!

torch_xla/experimental/pjrt.py
JackCaoG: @alanwaketan to confirm my statement above.

will-cromar: I followed the example in the PyTorch docs that initialized a `nccl` default group and a `gloo` subgroup to use `barrier`, so I assume this is safe: https://pytorch.org/docs/stable/distributed.html#debugging-torch-distributed-applications.

The default group that this function creates also uses the `xla` backend, which does seem to work fine in the tests.

PR #3944, opened 2022-08-27
Clean up SymIntElements

JackCaoG:
We don't need to create constant `size_node` for static dimensions. I don't think we ever want to call `SetDimensionSize` with a static dimension.

TODO
- [x] add cpp test

### DIFF
--- a/torch_xla/csrc/torch_util.h
+++ b/torch_xla/csrc/torch_util.h
@@ -16,26 +16,26 @@ namespace torch_xla {
 // Unpack SymInt objects into their building blocks
 struct SymIntElements {
  public:
-  SymIntElements(c10::SymInt& size) { SetSymIntNodeElements(size); }
+  SymIntElements(c10::SymInt& size) { AddSymIntNodeElements(size); }
   SymIntElements(c10::SymIntArrayRef& size) {
     std::vector<c10::SymInt> _sizes = torch::lazy::ToVector<c10::SymInt>(size);
     for (auto& _size : _sizes) {
-      SetSymIntNodeElements(_size);
+      AddSymIntNodeElements(_size);
     }
   }
-  std::unordered_map<size_t, torch::lazy::NodePtr> GetNodeMap() {
+  std::unordered_map<size_t, torch::lazy::NodePtr> GetNodeMap() const {
     return size_node_map_;
   }
-  std::vector<int64_t> GetUpperBounds() { return upper_bounds_; }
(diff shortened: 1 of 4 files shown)


JackCaoG:
@Krovatkin I think this one is ready for review.

torch_xla/csrc/torch_util.h
```diff
@@ -16,24 +16,27 @@ namespace torch_xla {
...
   void SetUpperBound(int64_t index, int64_t upper_bound) {
     XLA_CHECK_GT(upper_bounds_.size(), index);
     upper_bounds_[index] = upper_bound;
   }
 
  private:
-  void SetSymIntNodeElements(c10::SymInt& size);
-  std::vector<torch::lazy::NodePtr> size_nodes_;
+  void AddSymIntNodeElements(c10::SymInt& size);
+  std::unordered_map<size_t, torch::lazy::NodePtr> size_node_map_;
```
Krovatkin, line 39: I think we **need** a comment here. I believe @miladm 's intention here was to only keep track of dimension indices that map to size nodes.

Krovatkin: we could probably get away with using a dense vector with nullptr gaps since we always know a rank ahead of time to allocate it. it would probably make `SymIntElements::GetNode` simpler, but no strong preferences.

torch_xla/csrc/torch_util.h
```diff
@@ -16,24 +16,27 @@ namespace torch_xla {
 // Unpack SymInt objects into their building blocks
 struct SymIntElements {
  public:
-  SymIntElements(c10::SymInt& size) { SetSymIntNodeElements(size); }
+  SymIntElements(c10::SymInt& size) { AddSymIntNodeElements(size); }
   SymIntElements(c10::SymIntArrayRef& size) {
     std::vector<c10::SymInt> _sizes = torch::lazy::ToVector<c10::SymInt>(size);
```
Krovatkin, line 21: why don't we just iterate over `size` directly? It doesn't look like we are storing `_sizes`

Krovatkin (approved):
:shipit: 

I love the tests! Very comprehensive!

torch_xla/csrc/torch_util.h
JackCaoG: not sure if I understand the suggestion. `AddSymIntNodeElements` currently only handle `c10::SymInt` so we iterate over `c10::SymIntArrayRef` to reuse the helper function.

Issue #3868, opened 2022-08-11
Unexpected performance  of dropout


ang868:
hi, recently I've been doing some torch xla performance analysis on gpu.  I came across with a problem that the default random algorithm with poor performance. 

I use bert base written by huggingface transformers as my test algorithm and all of the following tests were done on nvidia 16G V100.

The througput results training with torch xla are as follows:

- bert base without dropout: 150.38 samples/s
- bert base with dropout:       106.67 samples/s

There are 29.1% performance drop. The magnitude of the performance degradation does not meet expectations. I tested bert base running with pytorch native:

- bert base without dropout: 99.62 samples/s
- bert base  with dropout:      91.23 samples/s

There are only 8.4% performance drop.

I found there are three options to select different  random algorithm in torch xla: default, philox and three_fry. So I tested the performance impact of the three different algorithms:

- ThreeFry generator(torch xla):  129.87 samples/s
- Philox generator (torch xla):      102.56 samples/s
- Default generator (torch xla):    106.67 samples/s

I think the philox generator will eventually be selected if you choose default generator option. So they're basically the same performance. 

So my questions are:

1. As described in the tensorflow document(https://www.tensorflow.org/guide/random_numbers#xla_devices), ThreeFry is fast on TPU but slow on CPU/GPU compared to Philox. Is it as expected that three fry generator has better throughput performance than philox? 
2. Although the performance of three fry is better than philox, it is still worse than expected. Is there any way to further optimize this performance?

JackCaoG:
Thanks for raising this issue, I will try to find someone from xla:gpu team to answer this question.

ymwangg:
@ang868 I saw very similar patterns before. XLA implements these RNGs using other XLA ops (e.g. [here](https://github.com/tensorflow/tensorflow/blob/490246643af1043ead0f5584158c062a63004012/tensorflow/compiler/xla/client/lib/prng.cc#L46-L49)) and the generated cuda kernel may not be as performant as those pytorch handwritten kernels that utilizes curand library. I'm not sure why XLA does not have a curand version of RNG but my guess is XLA wants a stateless RNG. One optimization I came up with is to replace the torch_xla RNG with the pytorch version using XLA custom_call, but the main issue I found is using XLA custom_call may reduce fusion opportunities and thus deliver limited speedup while increases memory usage.

BTW the performance of three_fry looks very reasonable and I think we may consider making it default on GPU.

ang868:
[...]

PR #3951, opened 2022-08-30
Set three_fry as the default RNG for GPU

ymwangg:
As discussed in https://github.com/pytorch/xla/issues/3868, three_fry demonstrates better performance than the default (philox) RNG on GPU. So we should consider making three_fry the default RNG on GPU.

Below are some models I've tested:
model | default | three_fry | speedup
-- | -- | -- | --
bert-base-uncased | 85.1333625 | 93.148043 | 1.09414265
roberta-base | 77.8481562 | 84.241191 | 1.08212185
facebook/bart-base | 63.1630113 | 69.2708869 | 1.0967002
gpt2 | 86.7388357 | 89.7944364 | 1.0352276
google/mt5-small | 53.6285788 | 54.9645302 | 1.02491118
t5-small | 86.5723865 | 86.4317218 | 0.99837518
microsoft/deberta-base | 52.1680586 | 53.5818296 | 1.02710032
google/long-t5-local-base | 26.8973489 | 28.0964403 | 1.04458028
resnet50 | 1067.582 | 1070.72111 | 1.00294039

cc @JackCaoG @ang868

### DIFF
--- a/test/pytorch_test_base.py
+++ b/test/pytorch_test_base.py
@@ -87,6 +87,7 @@ DISABLED_TORCH_TESTS_ANY = {
         'test_cdist_large_batch',
         'test_cdist_non_contiguous',
         'test_cdist_non_contiguous_batch',
+        'test_cov',  # XLA generates inf while nan is expected
         'test_broadcast_batched_matmul',  # incorrect Size
         'test_bincount',
         'test_view_all_dtypes_and_devices',  # uses half
--- a/torch_xla/csrc/random.cpp
+++ b/torch_xla/csrc/random.cpp
@@ -8,14 +8,26 @@
 #include "tensorflow/compiler/xla/xla_client/debug_macros.h"
 #include "tensorflow/compiler/xla/xla_client/sys_util.h"
 #include "torch_xla/csrc/convert_ops.h"
+#include "torch_xla/csrc/device.h"
 #include "torch_xla/csrc/helpers.h"
 
 namespace torch_xla {
 namespace {
 
+std::string GetDefaultGitGeneratorName() {
+  XlaDeviceType hw_type = static_cast<XlaDeviceType>(GetCurrentDevice().type());
+  switch (hw_type) {
+    case XlaDeviceType::GPU:
+      return "three_fry";
+    default:
+      return "default";
+  }
+}
+
 xla::BitGeneratorTy GetBitGenerator() {
-  static const std::string* bit_generator = new std::string(
-      xla::sys_util::GetEnvString("XLA_RNG_BIT_GENERATOR", "default"));
+  static const std::string* bit_generator =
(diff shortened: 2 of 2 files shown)


JackCaoG (approved):
There were some ci error before, not sure if it is related.

ymwangg:
I'm able to reproduce the CI error locally.
```
res = tensor([[nan, nan], [nan, inf]], device='xla:1')
ref = [[nan nan], [nan nan]]
```
It looks like changing RNG happened to generate invalid inputs for `torch.cov` and the assertEqual treats `inf` and `nan` as not equal.

JackCaoG:
Hmm, can you give me a bit more context, I assume that there is a input which `torch.conv` generate `nan` on cpu and `inf` on xla:gpu. If that's the case, we can just disable that test from our end. I just want to make sure this is not a regression.

ymwangg:
Below is the minimum code to reproduce:
```py
import numpy as np
import torch
import os

os.environ['GPU_NUM_DEVICES'] = '1'
os.environ['XLA_RNG_BIT_GENERATOR'] = 'default'

device = xm.xla_device()
count = 1
x = torch.testing.make_tensor((2,1),dtype=torch.float, device=device)
y = torch.randint(1, 3, (1,), device=device)
z = torch.testing.make_tensor((1,), dtype=torch.float, device=device, low=1)

x = torch.testing.make_tensor((2,1),dtype=torch.float, device=device)
y = torch.tensor([count], dtype=torch.int64, device=device)
z = torch.testing.make_tensor((1,), dtype=torch.float, device=device, low=1)

xm.mark_step()
w = torch.cov(x,correction=2,fweights=y,aweights=z)
xm.mark_step()

print(w)
print(np.cov(x.cpu().numpy(), ddof=2, fweights=y.cpu().numpy(), aweights=z.cpu().numpy()))
```
With outputs:
```
xla: tensor([[nan, nan], [nan, inf]], device='xla:1')
numpy: [[nan nan] [nan nan]]
```
If I change `count` to 3, I got:
```
xla: tensor([[0.0000e+00, 0.0000e+00], [0.0000e+00, 1.7053e-13]], device='xla:1')
numpy: [[0. 0.], [0. 0.]]
```

I checked the pytorch implementation and it looks `torch.cov` will force a divide by zero if the degree of freedom is below 1 ([here](https://github.com/pytorch/pytorch/blob/65ea3d062161f3aa5c8969b62ca322b0518300ae/aten/src/ATen/native/Correlation.cpp#L101)).

So my guess is that the discrepancy is due to `1.7053e-13/0.0 = inf` while `0.0/0.0 = nan`.

I've disabled this test since this op is not officially supported by pytorch/xla (due to lack of torch.equal support) and the error appears to be irrelevant.

JackCaoG (approved):
Thanks!

PR #3915, opened 2022-08-22
[WIP] Reproduce `nan` loss with DDP and XLA backend

will-cromar:
See #3625

- Add [`opts` argument](https://github.com/pytorch/pytorch/blob/a85d1f0bcdd02cf18d3b0517337458cb51a18cdb/torch/_C/_distributed_c10d.pyi#L212-L217) to `allgather` to fix `TypeError: allgather() takes 3 positional arguments but 4 were given` during initialization
- Refactor `broadcast` to produce the same graph on all replicas to avoid hanging on `REDUCE_SUM`.
- Add patch for PyTorch to avoid `RuntimeError: .../tensor_impl.cpp:181 : XLA tensors do not have storage` when it calls `is_alias_of`

Although I can get a model to initialize with DDP, the loss consistently goes to `nan`. I found that the gradient increased a little more than 4x after each iteration on a v4-8 (4 devices), suggesting that the gradient is not getting reset after each iteration and is just accumulating infinitely. Once the parameters become unstable, the loss becomes `nan`. It is unclear if this is related to the hack I'm using to fix the crash in `is_alias_of`.

@JackCaoG performed some experiments and found that the IR changed after the first steps:

<details>

<summary>IR for the first 4 steps and comments by @JackCaoG</summary>

```
IR {
  %0 = f32[] xla::device_data(), device=TPU:0
  %1 = s64[1]{0} xla::device_data(), device=TPU:0
  %2 = (s64[4]{0}, f32[]) xla::all_gather(%1, %0), num_outputs=2, dim=0, shard_count=4, pin_layout=0, groups=()
  %3 = s64[] prim::Constant(), value=1
  %4 = s64[6]{0} xla::device_data(), device=TPU:0
  %5 = s64[6]{0} aten::mul(%4, %3)
  %6 = (s64[6]{0}, f32[]) xla::cross_replica_sum(%5, %2.1), num_outputs=2, reduce_type=0, scale=1, pin_layout=0, groups=()
  %7 = f32[] prim::Constant(), value=1
  %8 = f32[10]{0} xla::device_data(), device=TPU:0
  %9 = f32[10]{0} aten::view(%8), output_size=(10)
  %10 = f32[10,10]{1,0} xla::device_data(), device=TPU:0
  %11 = f32[100]{0} aten::view(%10), output_size=(100)
  %12 = f32[110]{0} aten::cat(%11, %9), dim=0, dtype=Float
  %13 = f32[110]{0} aten::mul(%12, %7)
[...]

PR #3952, opened 2022-08-30
Fix DDP initialization with XLA backend

will-cromar:
Merging parts from #3915 that work.

- Add [`opts` argument](https://github.com/pytorch/pytorch/blob/a85d1f0bcdd02cf18d3b0517337458cb51a18cdb/torch/_C/_distributed_c10d.pyi#L212-L217) to `allgather` to fix `TypeError: allgather() takes 3 positional arguments but 4 were given during initialization`
- Refactor `broadcast` to produce the same graph on all replicas to avoid hanging on `REDUCE_SUM`.
- Add CI test to confirm initialization works on CPU

Manually tested on TPU v4.

### DIFF
--- a/test/pjrt/test_ddp.py
+++ b/test/pjrt/test_ddp.py
@@ -1,0 +1,37 @@
+from absl.testing import absltest, parameterized
+import torch.distributed as dist
+import torch.nn as nn
+from torch.nn.parallel import DistributedDataParallel as DDP
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.xla_backend
+from torch_xla.experimental import pjrt
+
+
+def _init_xla_backend(init_file: str):
+  rank = xm.get_ordinal()
+  world_size = xm.xrt_world_size()
+
+  dist.init_process_group(
+    "xla",
+    init_method=f"file://{init_file}",
+    rank=rank,
+    world_size=world_size)
+
+class TestPjRtDistributedDataParallel(parameterized.TestCase):
+
+  @staticmethod
+  def _ddp_init(init_file: str):
+    _init_xla_backend(init_file)
+
+    device = xm.xla_device()
+    model = nn.Linear(10, 10).to(device)
+    ddp_model = DDP(model)
+
+  def test_ddp_init(self):
+    pjrt.run_multiprocess(
+      self._ddp_step,
+      self.create_tempfile().full_path)
+
+
+if __name__ == "__main__":
+  absltest.main()
--- a/test/pjrt/test_ddp.py
+++ b/test/pjrt/test_ddp.py
@@ -12,10 +12,11 @@ def _init_xla_backend(init_file: str):
   world_size = xm.xrt_world_size()
 
   dist.init_process_group(
-    "xla",
-    init_method=f"file://{init_file}",
-    rank=rank,
-    world_size=world_size)
+      "xla",
+      init_method=f"file://{init_file}",
+      rank=rank,
+      world_size=world_size)
+
 
 class TestPjRtDistributedDataParallel(parameterized.TestCase):
 
@@ -28,9 +29,7 @@ class TestPjRtDistributedDataParallel(parameterized.TestCase):
     ddp_model = DDP(model)
 
   def test_ddp_init(self):
-    pjrt.run_multiprocess(
-      self._ddp_step,
-      self.create_tempfile().full_path)
+    pjrt.run_multiprocess(self._ddp_init, self.create_tempfile().full_path)
 
 
 if __name__ == "__main__":
--- a/test/test_torch_distributed_xla_backend.py
+++ b/test/test_torch_distributed_xla_backend.py
@@ -155,12 +155,19 @@ class XlaBackendTest(unittest.TestCase):
     opts = dist.BroadcastOptions()
     opts.rootRank = 0
     opts.rootTensor = 0
+
+    # Sync value of `tensor` to remove constants 1 and 0 from graph
+    xm.mark_step()
+
     # xla doesn't have broadcast. We use all_reduce to implement broadcast.
     all_reduce_pattern = r'%all\-reduce\.\d+ = .+ all\-reduce\('
     with xm_cc_op_intercepted('all_reduce'):
       pg_xla.broadcast([tensor], opts)
     hlo = torch_xla._XLAC._get_xla_tensors_hlo([tensor])
     hlo_matches(hlo, all_reduce_pattern)
+
+    assert 'constant' not in hlo, hlo
+
     # purge all computations attached the device.
     xm.mark_step()
 
--- a/torch_xla/distributed/xla_backend.py
+++ b/torch_xla/distributed/xla_backend.py
@@ -82,12 +82,10 @@ class ProcessGroupXla(ProcessGroup):
   # https://github.com/pytorch/pytorch/blob/release/1.10/torch/distributed/distributed_c10d.py#L1129
   def broadcast(self, tensors, opts):
     root_tensor = tensors[opts.rootTensor]
-    with torch.no_grad():
-      scale = 1 if opts.rootRank != self.rank() else 0
-      root_tensor = root_tensor * scale
-
-    xm.all_reduce(
-        xm.REDUCE_SUM, [root_tensor], groups=self._mesh, pin_layout=False)
+    xm.collective_broadcast([root_tensor],
+                            opts.rootRank,
+                            groups=self._mesh,
+                            pin_layout=False)
 
     return _ret_work([root_tensor])
 
--- a/torch_xla/distributed/xla_backend.py
+++ b/torch_xla/distributed/xla_backend.py
@@ -70,7 +70,7 @@ class ProcessGroupXla(ProcessGroup):
     xm.all_reduce(reduce_type, tensors, groups=self._mesh, pin_layout=False)
     return _ret_work(tensors)
 
-  def allgather(self, output_tensors_list, input_tensors):
+  def allgather(self, output_tensors_list, input_tensors, opts=None):
     for input_tensor, output_tensors in zip(input_tensors, output_tensors_list):
       result = xm.all_gather(input_tensor, groups=self._mesh, pin_layout=False)
       for i, slice in enumerate(torch.split(result, input_tensor.shape[0])):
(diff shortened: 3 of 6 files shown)


test/pjrt/test_ddp.py
```diff
@@ -0,0 +1,37 @@
...
+    rank=rank,
+    world_size=world_size)
+
+class TestPjRtDistributedDataParallel(parameterized.TestCase):
+
+  @staticmethod
+  def _ddp_init(init_file: str):
+    _init_xla_backend(init_file)
+
+    device = xm.xla_device()
```
alanwaketan, line 26: Do we need to specify the device ordinal here? I guess PJRT takes care of mapping the default xla device to the actual physical TPU core. And then during the lowering, the XLA IR will be able to extract the correct physical device ordinal.

test/pjrt/test_ddp.py
```diff
@@ -0,0 +1,37 @@
...
+    model = nn.Linear(10, 10).to(device)
+    ddp_model = DDP(model)
+
+  def test_ddp_init(self):
+    pjrt.run_multiprocess(
+      self._ddp_step,
```
alanwaketan, line 32: Where is this method? I guess it should be `self. _ddp_init`?

torch_xla/distributed/xla_backend.py
```diff
@@ -82,10 +82,8 @@ def allgather(self, output_tensors_list, input_tensors):
   # https://github.com/pytorch/pytorch/blob/release/1.10/torch/distributed/distributed_c10d.py#L1129
   def broadcast(self, tensors, opts):
     root_tensor = tensors[opts.rootTensor]
-    root_rank = opts.rootRank
-    if root_rank != self.rank():
-      with torch.no_grad():
-        root_tensor.zero_()
+    scale = 1 if opts.rootRank != self.rank() else 0
+    root_tensor = root_tensor * scale
```
alanwaketan, line 86: Should we have torch.no_grad() here to omit these steps from backward?

will-cromar: Done

test/pjrt/test_ddp.py
will-cromar: 🤦  fixed

test/pjrt/test_ddp.py
will-cromar: Correct. If I omit the `to(device)` and pass `device_ids` into DDP I get this error:

```
ValueError: DistributedDataParallel device_ids and output_device arguments only work with single-device/multiple-device GPU modules or CPU modules, but got device_ids [1], output_device None, and module parameters {device(type='cpu')}.
```

alanwaketan: Oh, `device_ids` is more or less a CUDA concept: https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/distributed.py#L436.

What I was referring to is something possibly like `device = xm.xla_device(rank)`.

will-cromar: Gotcha. The initialization in `xmp.spawn` for XRT and `pjrt.run_multiprocess` in PJRT take care of assigning the default device to each process/thread such that `xm.xla_device()` will always return the device assigned to a replica.

alanwaketan: Sounds good. Thank you, Will.

JackCaoG (approved):
Thanks!

JackCaoG:
failure seems real, through I don't know why it only fails on CPU. Maybe multiprocess on CPU is actually not working(which is somewhat expected..)

will-cromar:
Talked offline with @JackCaoG. We need to use an in-place multiply here, otherwise the unit test will have a different copy of the input tensor:

https://github.com/pytorch/xla/blob/ef539a8d920be4f141e392c4bb321cefa89c4a0f/torch_xla/distributed/xla_backend.py#L87

If I change this line to an in-place multiply (`root_tensor.mul_(scale)`), then the failing unit test passes again. However, then the new test in this PR (`test_ddp_init`) starts hanging forever again. I checked that the HLO is the same between them:

<details>
<summary>HLO in process 0</summary>
0 HloModule IrToHlo.29, entry_computation_layout={(f32[],s64[1]{0},s64[6]{0})->(s64[6]{0})}

%AddComputation.20 (x.21: s64[], y.22: s64[]) -> s64[] {
  %x.21 = s64[] parameter(0)
  %y.22 = s64[] parameter(1)
  ROOT %add.23 = s64[] add(s64[] %x.21, s64[] %y.22)
}

ENTRY %IrToHlo.29 (p0.1: f32[], p1.2: s64[1], p2.13: s64[6]) -> (s64[6]) {
  %p2.13 = s64[6]{0} parameter(2)
  %constant.12 = s64[] constant(1)
  %broadcast.14 = s64[6]{0} broadcast(s64[] %constant.12), dimensions={}
  %multiply.15 = s64[6]{0} multiply(s64[6]{0} %p2.13, s64[6]{0} %broadcast.14)
  %p0.1 = f32[] parameter(0)
  %p1.2 = s64[1]{0} parameter(1)
  %convert.3 = s64[] convert(f32[] %p0.1)
  %broadcast.4 = s64[1]{0} broadcast(s64[] %convert.3), dimensions={}
  %add.5 = s64[1]{0} add(s64[1]{0} %p1.2, s64[1]{0} %broadcast.4)
  %all-gather.6 = s64[4]{0} all-gather(s64[1]{0} %add.5), replica_groups={}, dimensions={0}
  %constant.7 = s32[] constant(0)
  %broadcast.8 = s32[1]{0} broadcast(s32[] %constant.7), dimensions={}
  %gather.9 = s64[] gather(s64[4]{0} %all-gather.6, s32[1]{0} %broadcast.8), offset_dims={}, collapsed_slice_dims={0}, start_index_map={0}, index_vector_dim=0, slice_sizes={1}
  %convert.10 = f32[] convert(s64[] %gather.9)
  %multiply.11 = f32[] multiply(f32[] %p0.1, f32[] %convert.10)
  %convert.16 = s64[] convert(f32[] %multiply.11)
  %tuple.17 = (s64[6]{0}, s64[]) tuple(s64[6]{0} %multiply.15, s64[] %convert.16)
  %get-tuple-element.18 = s64[6]{0} get-tuple-element((s64[6]{0}, s64[]) %tuple.17), index=0
  %get-tuple-element.19 = s64[] get-tuple-element((s64[6]{0}, s64[]) %tuple.17), index=1
  %all-reduce.24 = (s64[6]{0}, s64[]) all-reduce(s64[6]{0} %get-tuple-element.18, s64[] %get-tuple-element.19), replica_groups={}, to_apply=%AddComputation.20
  %get-tuple-element.26 = s64[] get-tuple-element((s64[6]{0}, s64[]) %all-reduce.24), index=1
  %convert.27 = f32[] convert(s64[] %get-tuple-element.26)
  %get-tuple-element.25 = s64[6]{0} get-tuple-element((s64[6]{0}, s64[]) %all-reduce.24), index=0
  ROOT %tuple.28 = (s64[6]{0}) tuple(s64[6]{0} %get-tuple-element.25)
}
</details>

Note: this unit test is skipped on GPUs, which is why the GPU CI tests pass.

will-cromar:
If I set `XLA_NO_SPECIAL_SCALARS=1` (so we don't use constants), the test doesn't hang.

torch_xla/distributed/xla_backend.py
```diff
@@ -82,10 +82,10 @@ def allgather(self, output_tensors_list, input_tensors):
   # https://github.com/pytorch/pytorch/blob/release/1.10/torch/distributed/distributed_c10d.py#L1129
   def broadcast(self, tensors, opts):
     root_tensor = tensors[opts.rootTensor]
-    root_rank = opts.rootRank
-    if root_rank != self.rank():
-      with torch.no_grad():
-        root_tensor.zero_()
+    with torch.no_grad():
```
JackCaoG, line 85: Can we add a comment here to explain why this is prefered over a if condition.

yeounoh:
> XLA_NO_SPECIAL_SCALARS

This is currently set to 0 by default. Should we make 1 a default value? @JackCaoG  I noticed that we have a few new flags being read and used in the code -- I think we should start listing/documenting them unless we plan to deprecate (e.g., README within the repo, or a comment block inside the torch_xla __init__.py)?

torch_xla/distributed/xla_backend.py
```diff
@@ -70,7 +70,7 @@ def allreduce(self, tensors, all_reduce_options):
     xm.all_reduce(reduce_type, tensors, groups=self._mesh, pin_layout=False)
     return _ret_work(tensors)
 
-  def allgather(self, output_tensors_list, input_tensors):
+  def allgather(self, output_tensors_list, input_tensors, opts=None):
```
yeounoh, line 73: @will-cromar I think we need to add some comment around `opts=None`. We might think that this unused argument needs to be remove in the future -- some linter would complain, too.

will-cromar:
Added a new `xm.collective_broadcast` operation that uses `scale` as device data (instead of a constant) in both `dist.broadcast` and `pjrt.broadcast_master_param`.

Manually ran `test_ddp.py` and `test_experimental_pjrt_tpu.py` on TPU to confirm these still pass.

torch_xla/distributed/xla_backend.py
will-cromar: Done.

will-cromar:
The default for `XLA_NO_SPECIAL_SCALARS` should still be 1 to enable XLA optimizations of operations using `0` and `1`. Ideally, there really should be some more explicit way of configuring this option (e.g. with some context manager), but we can work around it by explicitly sending `1` or `0` to the XLA device from CPU.

torch_xla/distributed/xla_backend.py
will-cromar: Is a comment here necessary? This is consistent with the Python interface it is implementing here: https://github.com/pytorch/pytorch/blob/a85d1f0bcdd02cf18d3b0517337458cb51a18cdb/torch/_C/_distributed_c10d.pyi#L212-L217

Do you think we should like to this `pyi` file somewhere in the code?

yeounoh: I see, if we are implementing the abstract class, then we don't need to -- had to view the entire file to see `class ProcessGroupXla(ProcessGroup)` lol. Thanks @will-cromar

test/test_torch_distributed_xla_backend.py
```diff
@@ -155,12 +155,19 @@ def test_broadcast(self):
     opts = dist.BroadcastOptions()
     opts.rootRank = 0
     opts.rootTensor = 0
+
+    # Sync value of `tensor` to remove constants 1 and 0 from graph
+    xm.mark_step()
```
alanwaketan, line 160: Just curios, which part of the above program will set constants 1 and 0 in the graph? Or this is just a safeguard?

will-cromar: This makes the HLO more readable in the case of a failure, since it's just taking `tensor` as an input and multiplying it by 1 after this. I also saw some constants in the graph from creating `tensor`, definitely from the `+ 1` but I think also from `arange(2)` because the result is `[0, 1]`. In any case, we don't want to depend on the implementation details of `arange` at all, so I think it's best to sync the graph here.

alanwaketan (approved):
LGTM.

PR #3954, opened 2022-08-31
Make SizeNode to use XlaShape and add test

JackCaoG:
Not all of the Ops has `torch::lazy::shape`, fallback to use `xla_shape` to check for upperbound and is_dynamic. 

TODO:
- [x] add test for other SizeNodes

### DIFF
--- a/test/cpp/test_ir.cpp
+++ b/test/cpp/test_ir.cpp
@@ -109,8 +109,6 @@ TEST(IrTest, TestSizeNode) {
 
   EXPECT_EQ(dim_node_0->getStaticValue(), 3);
   EXPECT_EQ(dim_node_1->getStaticValue(), 4);
-  EXPECT_FALSE(dim_node_0->isSymbolic());
-  EXPECT_FALSE(dim_node_1->isSymbolic());
 
   ForEachDevice([&](const torch::lazy::BackendDevice& device) {
     // Lower the SizeNode and execute the GetDimensionSize.
@@ -133,7 +131,6 @@ TEST(IrTest, TestSizeAddNode) {
       std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node_add);
 
   EXPECT_EQ(dim_node_add->getStaticValue(), 7);
-  EXPECT_FALSE(dim_node_add->isSymbolic());
 
   ForEachDevice([&](const torch::lazy::BackendDevice& device) {
     // Lower the SizeAddNode and execute the GetDimensionSize.
@@ -155,7 +152,6 @@ TEST(IrTest, TestSizeMulNode) {
       std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node_mul);
 
   EXPECT_EQ(dim_node_mul->getStaticValue(), 12);
-  EXPECT_FALSE(dim_node_mul->isSymbolic());
 
   ForEachDevice([&](const torch::lazy::BackendDevice& device) {
     // Lower the SizeAddNode and execute the GetDimensionSize.
@@ -177,7 +173,6 @@ TEST(IrTest, TestSizeDivNode) {
       std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node_div);
 
   EXPECT_EQ(dim_node_div->getStaticValue(), 2);
-  EXPECT_FALSE(dim_node_div->isSymbolic());
 
   ForEachDevice([&](const torch::lazy::BackendDevice& device) {
     // Lower the SizeAddNode and execute the GetDimensionSize.
--- a/test/cpp/test_ir.cpp
+++ b/test/cpp/test_ir.cpp
@@ -120,27 +120,27 @@ TEST(IrTest, TestSizeNode) {
   });
 }
 
-// TEST(IrTest, TestSizeAddNode) {
-//   torch::lazy::NodePtr scalar_node =
-//       ScalarOp(1.0, xla::ShapeUtil::MakeShape(xla::F32, {3, 4}));
-//   torch::lazy::NodePtr size_node_0 =
-//       torch::lazy::MakeNode<SizeNode>(scalar_node, 0);
-//   torch::lazy::NodePtr size_node_1 =
-//       torch::lazy::MakeNode<SizeNode>(scalar_node, 1);
-//   torch::lazy::NodePtr size_node_add =
-//       torch::lazy::MakeNode<SizeAdd>(size_node_0, size_node_1);
-//   std::shared_ptr<torch::lazy::DimensionNode> dim_node_add =
-//       std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node_add);
+TEST(IrTest, TestSizeAddNode) {
+  torch::lazy::NodePtr scalar_node =
+      ScalarOp(1.0, xla::ShapeUtil::MakeShape(xla::F32, {3, 4}));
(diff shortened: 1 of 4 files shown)


torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -23,28 +23,34 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
...
 
 bool SizeNode::isSymbolic() const {
-  auto symbolic_vec = operand(0).node->shape(0).is_symbolic();
-  if (!symbolic_vec.has_value()) {
-    return true;
-  }
-  return symbolic_vec->at(dim_);
+  // Not all IR has torch::lazy::shape now, use xla::shape to unblock
+  // the development.
+  return dynamic_cast<const XlaNode*>(operand(0).node)
```
JackCaoG, line 36: @Krovatkin @miladm  Something is off here. Let me explain my understanding.

from pytorch symint perspective
1. `isSymbolic == true` symint wraps around a Node
2. `isSymbolic == false` symint wraps around a int

This pr aims to provide `isSymbolic` using xla_shape, what I did was
1. `isSymbolic == true` this dimension is dynamic
2. `isSymbolic == false` this dimension is static

Noted that `Symbolic` != `Dynamic`, for example we can call `xla::GetDimensionSize(static_tensor, 0)` and it will return a Node represents size of `static_tensor` but `static_tensor` is not dynamic.

I guess the question is more of why do we ever need `isSymbolic` in the SizeNode. According to the defination, if we ever trigger the SizeNode computation, it will always be Symbolic. If none of the input is `Symbolic`, the regular python math operator will be used anyway.

JackCaoG: chatted with @Krovatkin , we agree that `isSymbolic` should be removed from SizeNode since it should always return true. `isSymbolic` should also equal to `isDynamic` for most real context. I will make it always return true in this pr and remove it once upstream remove it.

JackCaoG:
For some reason this pr breaks `SymintTest.TestDynamicSymint` looking into it

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -23,95 +23,78 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
...
-  return ReturnOp(
-      (xla::GetDimensionSize(input1, 0) + xla::GetDimensionSize(input2, 0)),
-      loctx);
+  return ReturnOp((input1 + input2), loctx);
 }
 
 SizeMul::SizeMul(torch::lazy::Value a, torch::lazy::Value b)
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::mul")},
-              {a, b}, xla::ShapeUtil::MakeShape(xla::S32, {}), 1){};
+              {a, b}, xla::ShapeUtil::MakeShape(xla::S32, {}), 1) {
```
Krovatkin, line 58: sorry, this was already in the original code but shouldn't this be `xla::S64`?

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -23,95 +23,78 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
...
-  auto symbolic_vec = operand(0).node->shape(0).is_symbolic();
-  if (!symbolic_vec.has_value()) {
-    return true;
-  }
-  return symbolic_vec->at(dim_);
+  // Not all IR has torch::lazy::shape now, use xla::shape to unblock
+  // the development.
+  return dynamic_cast<const XlaNode*>(operand(0).node)
+      ->xla_shape(operand(0).index)
+      .dimensions(dim_);
```
Krovatkin, line 30: 😢 very sad

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -23,95 +23,78 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
...
 
 std::string SizeAdd::ToString() const { return "SizeAdd"; }
 
 XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
   auto input1 = loctx->GetOutputOp(operand(0));
   auto input2 = loctx->GetOutputOp(operand(1));
-  return ReturnOp(
-      (xla::GetDimensionSize(input1, 0) + xla::GetDimensionSize(input2, 0)),
-      loctx);
+  return ReturnOp((input1 + input2), loctx);
```
Krovatkin, line 53: okay this makes more sense to me since I wasn't sure why we needed to do `GetDimensionSize` on scalars 😸

test/cpp/test_ir.cpp
```diff
@@ -94,5 +95,91 @@ TEST(IrTest, TestScopePusherWithDebugging) {
...
+  torch::lazy::NodePtr size_node_add =
+      torch::lazy::MakeNode<SizeAdd>(size_node_0, size_node_1);
+  std::shared_ptr<torch::lazy::DimensionNode> dim_node_add =
+      std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node_add);
+
+  EXPECT_EQ(dim_node_add->getStaticValue(), 7);
+
+  ForEachDevice([&](const torch::lazy::BackendDevice& device) {
+    // Lower the SizeAddNode and execute the GetDimensionSize.
+    auto results = ExecuteAndFetch({size_node_add}, device);
```
Krovatkin, line 137: do you think `ExecuteAndFetch` will be used to implement for `getDynamicValue()` or is this for testing?

Krovatkin (approved):
:shipit:

test/cpp/test_ir.cpp
```diff
@@ -94,5 +95,91 @@ TEST(IrTest, TestScopePusherWithDebugging) {
   FLAGS_torch_lazy_ir_debug = restore_FLAGS_torch_lazy_ir_debug;
 }
 
+TEST(IrTest, TestSizeNode) {
+  torch::lazy::NodePtr scalar_node =
+      ScalarOp(1.0, xla::ShapeUtil::MakeShape(xla::F32, {3, 4}));
```
vanbasten23, line 100: Perhaps a silly question, for the "op" in "ScalarOp", is it operator or operand? Because it's a scalar, so it looks like an operand. But the torch::lazy::Node is a "operation a Node can be associated to".

test/cpp/test_ir.cpp
JackCaoG: `ExecuteAndFetch` calls `GetTensors` which is what I used to implement `getDynamicValue`

test/cpp/test_ir.cpp
JackCaoG: It is a Node, check https://github.com/pytorch/xla/blob/227ac13d7102f4e30540df7ef50d09a02a985290/torch_xla/csrc/ops/ops.h#L17

torch_xla/csrc/ops/dynamic_ir.cpp
JackCaoG: good catch, let me fix it.

JackCaoG:
I will merge this pr once all test passed.

JackCaoG:
Ok I need to rebase..

miladm:
Looked into this PR. Makes sense. Thanks @JackCaoG.

PR #3959, opened 2022-09-01
Fix the location of libxla_computation_client.so

alanwaketan:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3961
* #3960
* __->__ #3959

Summary:
Update the location of libxla_computation_client.so to the correct place
while building the cpp test binary.

Test Plan:
CI for builiding the cpp test binary.

PR #3960, opened 2022-09-01
Support Tensor.is_alias_of

alanwaketan:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3961
* __->__ #3960
* #3959

Summary:
Tensor.is_alias_of relies on Storage to perform. However, XLATensorImpl was
not implemented with that in mind. This commit adds a fake storage to XLATensor
as a marker to mark XLATensor that point to the same storage. The reason
why it's not done at XLATensorImpl is that XLATensor maintains the view ops/alias
logic in XLATensor class instead of relying on XLATensorImpl to do the check.

Test Plan:
./test/cpp/build/test_ptxla --gtest_filter=AtenXlaTensorTest.TestViewIsAliasOf

### DIFF
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -11212,5 +11212,25 @@ TEST_F(AtenXlaTensorTest, TestRoll) {
   ExpectCounterChanged("xla::roll", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestViewIsAliasOf) {
+  auto a = torch::empty(4, torch::TensorOptions(torch::kFloat));
+  auto b = torch::empty(4, torch::TensorOptions(torch::kFloat));
+
+  ForEachDevice([&](const torch::Device& device) {
+    auto xla_a = CopyToDevice(a, device);
+    auto xla_b = CopyToDevice(b, device);
+    EXPECT_EQ(!a.is_alias_of(b), !xla_a.is_alias_of(xla_b));
+
+    auto c = a.view({2, 2});
+    auto xla_c = xla_a.view({2, 2});
+    EXPECT_EQ(a.is_alias_of(c), xla_a.is_alias_of(xla_c));
+
+    auto d = c.view({1, 4});
+    auto lazy_d = xla_c.view({1, 4});
+    EXPECT_EQ(d.is_alias_of(c), lazy_d.is_alias_of(xla_c));
+    EXPECT_EQ(d.is_alias_of(a), lazy_d.is_alias_of(xla_a));
+  });
+}
+
 }  // namespace cpp_test
 }  // namespace torch_xla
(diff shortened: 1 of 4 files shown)


test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -11212,5 +11212,25 @@ TEST_F(AtenXlaTensorTest, TestRoll) {
   ExpectCounterChanged("xla::roll", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestViewIsAliasOf) {
+  auto a = torch::empty(4, torch::TensorOptions(torch::kFloat));
```
JackCaoG, line 11216: can we spell out `torch::tensor` in the test? I found `auto` hard to read sometime 😄

torch_xla/csrc/tensor.cpp
```diff
@@ -492,28 +492,32 @@ XLATensorPtr XLATensor::Create(
...
-    : data_(std::make_shared<Data>(std::move(view), device,
-                                   logical_element_type)) {}
+    : XLATensor(std::make_shared<Data>(std::move(view), device,
+                                       logical_element_type)) {}
 
-XLATensor::XLATensor(std::shared_ptr<Data> data) : data_(std::move(data)) {}
+XLATensor::XLATensor(std::shared_ptr<Data> data)
+    : data_(std::move(data)),
+      storage_(c10::Storage(
+          {}, 0,
```
JackCaoG, line 519: I am a bit confuse, if we always init `Storage` the same way, how can we tell each tensor's storage is different? (I mean it works, but I am not sure why...)

JackCaoG (commented):
mostly lgtm

torch_xla/csrc/tensor.cpp
alanwaketan: It's a pointer. We compare the memory address instead of the values.

test/cpp/test_aten_xla_tensor.cpp
alanwaketan: Sure. I'm a lazy person so I'm a big fans of `auto`.

torch_xla/csrc/tensor.cpp
JackCaoG: If I understand correctly, you are comparing two `c10::DataPtr` which is a pointer to decide if two XLATensor has the same storage?

alanwaketan: That's correct. The actual value stored in the pointer doesn't matter.

JackCaoG (approved):
Feel free to merge after auto change

alanwaketan:
It looks like pt/xla repo doesn't allow people to push to master without a PR (GH UI?). Here is the error log from ghstack: https://paste.googleplex.com/6155066363871232. I guess we just can't use ghstack...

PR #3961, opened 2022-09-01
Make expand to support is_alias_of

alanwaketan:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #3961
* #3960
* #3959

Summary:
Expand is diffcult to express by using view tensors, and therefore the
storage bit or is_alias_of bit will not be correctly set. Set that manually.

Test Plan:
./test/cpp/build/test_ptxla --gtest_filter=AtenXlaTensorTest.TestExpand


alanwaketan:
There are two other ops we probably need to tweak a bit too: split and split_with_sizes. Here is a complete list of view ops and whether we need tweaks or not: http://shortn/_NJBOu5TaRe.

alanwaketan:
Since ghstack is not compatible with our repo's current setting, let's discard this pull request. And use #3967 instead.

Issue #3871, opened 2022-08-11
[RFC] XLAShardedTensor & Sharding Annotation API


yeounoh:
## 🚀 Feature

We propose `XLAShardedTensor` to represent a sharded tensor that wraps around `torch.Tensor`, and `mark_sharding()` API for tensor sharding annotation. XLAShardedTensor allows annotating tensors with sharding specs and dispatching the annotations to the XLA backend for XLA GSPMD support in PyTorch/XLA. 

### Usage Example
```python
import torch
import torch_xla.core.xla_model as xm
import torch_xla.distributed.xla_sharding as xs
from torch_xla.distributed.xla_sharding import Mesh

mesh_shape = (2, 4)  # device mesh 
num_devices = len(xm.get_xla_supported_devices())
device_ids = np.array(range(num_devices))
mesh = Mesh(device_ids, mesh_shape, ('x', 'y'))

t = torch.randn(8, 4).to(xm.xla_device())

# Mesh partitioning, each device holds 1/8-th of the input
partition_spec = (0, 1)
m1_sharded = xs.mark_sharding(t, mesh, partition_spec)

# XLAShardedTensor behaves like a unpartitioned native tensor:
# - we can apply native torch.Tensor ops and nn.layers
# - print the unpartitioned tensor when sent to CPU
# - provides access to the list of local shards after lazy execution
assert isinstance(m1_sharded, XLAShardedTensor) == True
```

## Motivation

Our goal is to support GSPMD model sharding in PyTorch -- this would allow a user to bring in PyTorch models implemented (as if) on a single device and annotate a few tensors with desired sharding specs to run efficient model parallelism. Such an automated approach to model sharding allows the XLA compiler to optimize the entire computation graph end-to-end and frees up the user from implementing sharded version of ops with proper collectives in place. 

The current PyTorch `ShardedTensor` abstraction [RFC](https://github.com/pytorch/pytorch/issues/55207) provides simple primitives to express sharded tensor also, and the `ShardedTensor` API provides a set of convenience helper functions to shard tensor or model parameters with sharding specs. Under the hood, it requires manual/explicit implementation of sharded ops (e.g., sharded version of `torch.nn.functional.linear`) and careful injection of collective comms. And the abstraction represents sharded tensors directly, not the sharding annotation for XLA compiler-based sharding that will take place lazily and support the `xla` backend.

## Pitch

[...]

PR #3963, opened 2022-09-01
[SPMD][PoC] InputHandler for input sharding

yeounoh:
This is taken from #3684 and addresses #3871 . The original PR will be broken down into 3 PRs, and this will be the second after/following #3962 .  

The changes include:
- Add `CreateTensorsData` to shard & transfer tensor data
- Add `PjRtShardedData` & sharded data helper functions
- Add `ExecuteReplicated` implemenation in PjRtComputationClient
- Add `is_sharded` parameter to `CompileInstance`
- Add `InputHandler` for input sharding.
- Refactor sharding APIs in XLATensor and XlaNode

### DIFF
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -88,7 +88,7 @@ TEST_F(XLAShardingTest, CreateTensorsData) {
   EXPECT_TRUE(xla::Shape::Equal().IgnoreLayout()(xla_data->shape(),
                                                  shards[0]->shape()));
   EXPECT_TRUE(
-      TwoEqualXlaData(tensors_data[0], WrapXlaData(shards[0]), at::kFloat));
+      XlaDataValuesEqual(tensors_data[0], WrapXlaData(shards[0]), at::kFloat));
 
   // Returns multiple input shards, replicated
   int64_t n_devices = xla::ComputationClient::Get()->GetLocalDevices().size();
@@ -99,8 +99,8 @@ TEST_F(XLAShardingTest, CreateTensorsData) {
     EXPECT_EQ(shards.size(), n_devices);
     EXPECT_TRUE(xla::Shape::Equal().IgnoreLayout()(sharded_xla_data->shape(),
                                                    shards[0]->shape()));
-    EXPECT_TRUE(TwoEqualXlaData(WrapXlaData(shards[0]), WrapXlaData(shards[1]),
-                                at::kFloat));
+    EXPECT_TRUE(XlaDataValuesEqual(WrapXlaData(shards[0]),
+                                   WrapXlaData(shards[1]), at::kFloat));
   }
 }
 
@@ -131,13 +131,13 @@ TEST_F(XLAShardingTest, InputHandler) {
 
   auto arg0_dev0 = arguments_by_device[0][0];
   auto arg0_dev1 = arguments_by_device[1][0];
-  EXPECT_TRUE(TwoEqualXlaData(WrapXlaData(arg0_dev0), WrapXlaData(arg0_dev1),
-                              at::kFloat));
+  EXPECT_TRUE(XlaDataValuesEqual(WrapXlaData(arg0_dev0), WrapXlaData(arg0_dev1),
+                                 at::kFloat));
 
   auto arg1_dev0 = arguments_by_device[0][1];
   auto arg1_dev1 = arguments_by_device[1][1];
-  EXPECT_TRUE(TwoEqualXlaData(WrapXlaData(arg1_dev0), WrapXlaData(arg1_dev1),
-                              at::kFloat));
+  EXPECT_TRUE(XlaDataValuesEqual(WrapXlaData(arg1_dev0), WrapXlaData(arg1_dev1),
+                                 at::kFloat));
 }
 
 }  // namespace cpp_test
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -64,5 +64,81 @@ TEST_F(XLAShardingTest, ShardTensor) {
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 7, 4}));
 }
 
+TEST_F(XLAShardingTest, CreateTensorsData) {
+  if (xla::sys_util::GetEnvString(xla::env::kEnvPjRtDevice, "") == "") {
+    GTEST_SKIP() << "`PJRT_DEVICE` is not set.";
+  }
+
+  std::vector<at::Tensor> tensors(2);
+  std::fill_n(tensors.begin(), tensors.size(),
+              at::ones({8, 8}, at::TensorOptions(at::kFloat)));
+  std::vector<std::string> devices(2);
+  std::fill_n(devices.begin(), devices.size(), GetDefaultDevice()->toString());
+  std::vector<XLATensor::ShardingSpecPtr> shardings = {
+      nullptr, std::make_shared<XLATensor::ShardingSpec>(
+                   xla::HloSharding::Replicate().ToProto())};
+  std::vector<torch::lazy::BackendDataPtr> tensors_data =
+      CreateTensorsData(tensors, shardings, devices);
+
+  // Returns the input without sharding
+  auto xla_data = dynamic_cast<XLAData*>(tensors_data[0].get())->xla_data();
+  std::vector<xla::ComputationClient::DataPtr> shards =
+      xla::ComputationClient::Get()->GetDataShards(xla_data);
+  EXPECT_EQ(shards.size(), 1);
+  EXPECT_TRUE(xla::Shape::Equal().IgnoreLayout()(xla_data->shape(),
+                                                 shards[0]->shape()));
+  EXPECT_TRUE(
+      TwoEqualXlaData(tensors_data[0], WrapXlaData(shards[0]), at::kFloat));
+
+  // Returns multiple input shards, replicated
+  int64_t n_devices = xla::ComputationClient::Get()->GetLocalDevices().size();
+  if (n_devices > 1) {
+    auto sharded_xla_data =
+        dynamic_cast<XLAData*>(tensors_data[1].get())->xla_data();
+    shards = xla::ComputationClient::Get()->GetDataShards(sharded_xla_data);
+    EXPECT_EQ(shards.size(), n_devices);
+    EXPECT_TRUE(xla::Shape::Equal().IgnoreLayout()(sharded_xla_data->shape(),
+                                                   shards[0]->shape()));
+    EXPECT_TRUE(TwoEqualXlaData(WrapXlaData(shards[0]), WrapXlaData(shards[1]),
+                                at::kFloat));
+  }
+}
+
+TEST_F(XLAShardingTest, InputHandler) {
+  if ((xla::sys_util::GetEnvString(xla::env::kEnvPjRtDevice, "") == "") ||
+      (xla::ComputationClient::Get()->GetLocalDevices().size() < 2)) {
+    GTEST_SKIP()
+        << "`PJRT_DEVICE` is not set, with more than 2 local devices, ("
+        << xla::ComputationClient::Get()->GetLocalDevices().size()
+        << " local devices detected).";
+  }
+
+  std::vector<at::Tensor> tensors(2);
+  std::fill_n(tensors.begin(), tensors.size(),
+              at::ones({8, 8}, at::TensorOptions(at::kFloat)));
+  std::vector<std::string> devices(2);
+  std::fill_n(devices.begin(), devices.size(), GetDefaultDevice()->toString());
+  std::vector<XLATensor::ShardingSpecPtr> shardings = {
+      nullptr, std::make_shared<XLATensor::ShardingSpec>(
+                   xla::HloSharding::Replicate().ToProto())};
+  std::vector<torch::lazy::BackendDataPtr> tensors_data =
+      CreateTensorsData(tensors, shardings, devices);
+
+  devices = xla::ComputationClient::Get()->GetLocalDevices();
+  std::vector<xla::ComputationClient::DataPtr> arguments =
+      UnwrapXlaData(tensors_data);
+  auto arguments_by_device = ShardingUtil::InputHandler(arguments, devices);
+
+  auto arg0_dev0 = arguments_by_device[0][0];
+  auto arg0_dev1 = arguments_by_device[1][0];
+  EXPECT_TRUE(TwoEqualXlaData(WrapXlaData(arg0_dev0), WrapXlaData(arg0_dev1),
+                              at::kFloat));
+
+  auto arg1_dev0 = arguments_by_device[0][1];
+  auto arg1_dev1 = arguments_by_device[1][1];
+  EXPECT_TRUE(TwoEqualXlaData(WrapXlaData(arg1_dev0), WrapXlaData(arg1_dev1),
+                              at::kFloat));
+}
+
 }  // namespace cpp_test
 }  // namespace torch_xla
--- a/third_party/xla_client/BUILD
+++ b/third_party/xla_client/BUILD
@@ -51,7 +51,7 @@ tf_cc_shared_object(
     }),
     visibility = ["//visibility:public"],
     deps = [
-        "computation_client_impl",
+        ":computation_client_impl",
         "//tensorflow/compiler/xla:literal_util",
         "//tensorflow/compiler/xla/client",
         "//tensorflow/compiler/xla/client:global_data",
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -313,9 +313,10 @@ PjRtComputationClient::ExecuteReplicated(
       pjrt_computation.executable->Execute(argument_handles, execute_options)
           .ValueOrDie();
 
-  std::vector<std::vector<ComputationClient::DataPtr>> data_handles;
+  std::vector<std::vector<ComputationClient::DataPtr>> data_handles(
+      results.size());
   for (auto& result : results) {
-    std::vector<ComputationClient::DataPtr> datas;
+    std::vector<ComputationClient::DataPtr> datas(result.size());
     for (int32_t i = 0; i < result.size(); ++i) {
       std::unique_ptr<xla::PjRtBuffer> buffer = std::move(result[i]);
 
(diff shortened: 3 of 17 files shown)


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -118,6 +133,21 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
   return datas;
 }
 
+ComputationClient::DataPtr PjRtComputationClient::TransferShardsToServer(
+    absl::Span<const TensorSource> tensor_shards, std::string device,
```
will-cromar, line 137: Do you need the `device` and `shape` arguments here? They're contained in `TensorSource` for each tensor: https://github.com/pytorch/xla/blob/6e42e7cb3af01d9f8909e112a3be0148a87acad0/third_party/xla_client/computation_client.h#L129-L130

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -145,26 +175,49 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
...
+      xla::DeviceAssignment device_assignment(client_->device_count(), 1);
+      device_assignment.FillIota(0);
+      compile_options.executable_build_options.set_device_assignment(
+          device_assignment);
+    }
+
+    PjRtDevice* pjrt_device = StringToPjRtDevice(instance.compilation_device);
     std::unique_ptr<xla::PjRtExecutable> executable =
-        client_->Compile(instance.computation, compile_options).ValueOrDie();
+        ConsumeValue(client_->Compile(instance.computation, compile_options));
```
will-cromar, line 210: Is there any difference in behavior between `ConsumeValue` and `ValueOrDie`? It seems to use a deprecated reference to `ValueOrDie` anyway:

https://github.com/pytorch/xla/blob/b7f54864a0933d6401a023e12ec6ccac1982e2d9/third_party/xla_client/debug_macros.h#L25-L29

`ConsumeValueOrDie`: 
 https://github.com/tensorflow/tensorflow/blob/57d35fb585de65a34cd6871b8ca3a6501183f98a/tensorflow/core/platform/statusor.h#L234-L235

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -145,26 +175,49 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
...
+    HloComputation* hlo_computation = hlo_modules[0]->entry_computation();
+    xla::ProgramShape program_shape =
+        xla::ProgramShape(hlo_computation->ToProto().program_shape());
+
     std::shared_ptr<PjRtComputation> pjrt_computation =
-        std::make_shared<PjRtComputation>(std::move(instance.computation),
-                                          program_shape, instance.devices,
-                                          std::move(executable));
+        std::make_shared<PjRtComputation>(
+            std::move(xla::XlaComputation(hlo_modules[0]->ToProto())),
```
will-cromar, line 219: How is this new `XlaComputation` different than `instance.computation`? Does compilation change the program shape?

torch_xla/experimental/xla_sharded_tensor.py
```diff
@@ -48,6 +48,20 @@ class XLAShardedTensor(torch.Tensor):
   # Note: we can consider returning a callback or even define
   # sharding at XLAShardedTensor construction after pjrt migration.
   local_shards: List[XLAShard] = None
+  # A logical device topology, each element describes
+  # a number of devices in the corresponding axis.
+  # NOTE: we could use more specific device-rank mapping, e.g., ShardingSpec,
+  # if needed. The change shouldn't be difficult, or create another constructor.
+  mesh_shape: Tuple[int]  # TODO: create a wrapper for named axes
```
will-cromar, line 55: Regarding your TODO, this might be helpful: https://github.com/pytorch/xla/blob/b7f54864a0933d6401a023e12ec6ccac1982e2d9/torch_xla/experimental/tpu.py#L32-L51

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Yes, because they are the original (unpartitioned) shape and device. Each tensor source is a shard and their source and device are specific to the shard.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: There is an important distinction between `ConsumeValueOrDie` and `ValueOrDie` -- the former deletes the old copy, which we don't need anymore. It's better to "consume".

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Yes, we should take the program_shape after the compilation -- it worked before because it didn't change before and after, but it can change depending on the compilation options, including sharding.

torch_xla/experimental/xla_sharded_tensor.py
yeounoh: +1 thanks :)

torch_xla/csrc/tensor_util.cpp
```diff
@@ -904,6 +904,70 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
...
+  std::vector<xla::ComputationClient::DataPtr> handles;
+  for (size_t i = 0; i < tensors.size(); ++i) {
+    torch::lazy::BackendDevice device = ParseDeviceString(devices[i]);
+    xla::Shape shape = CreateComputationShapeFromTensor(tensors[i], &device);
+
+    std::vector<xla::ComputationClient::TensorSource> source_tensors;  // in
+    std::vector<xla::ComputationClient::DataPtr> new_handles;          // out
+    if (shardings[i] != nullptr) {
+      xla::OpSharding sharding = shardings[i]->sharding;
+      // TODO(yeounoh) PJRT runs a process per host and without cross host
```
will-cromar, line 924: Can you clarify in this comment that you're talking about the SPMD case? Without SPMD, we still run 4 processes per host by default

will-cromar (approved):

test/cpp/test_xla_sharding.cpp
```diff
@@ -64,5 +64,81 @@ TEST_F(XLAShardingTest, ShardTensor) {
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 7, 4}));
 }
 
+TEST_F(XLAShardingTest, CreateTensorsData) {
+  if (xla::sys_util::GetEnvString(xla::env::kEnvPjRtDevice, "") == "") {
+    GTEST_SKIP() << "`PJRT_DEVICE` is not set.";
```
JackCaoG, line 69: does this work on both PJRT:CPU and PJRT:TPU?

third_party/xla_client/BUILD
```diff
@@ -51,7 +51,7 @@ tf_cc_shared_object(
     }),
     visibility = ["//visibility:public"],
     deps = [
-        "computation_client_impl",
+        ":computation_client_impl",
```
JackCaoG, line 54: why is this needed?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -222,6 +275,63 @@ PjRtComputationClient::ExecuteComputation(
...
+  // TODO(yeounoh) currently only support single-slice execution
+  execute_options.multi_slice_config = nullptr;
+  std::vector<std::vector<std::unique_ptr<PjRtBuffer>>> results =
+      pjrt_computation.executable->Execute(argument_handles, execute_options)
+          .ValueOrDie();
+
+  std::vector<std::vector<ComputationClient::DataPtr>> data_handles;
+  for (auto& result : results) {
+    std::vector<ComputationClient::DataPtr> datas;
+    for (int32_t i = 0; i < result.size(); ++i) {
```
JackCaoG, line 319: nit
```
datas.reserve(result.size());
```

third_party/xla_client/pjrt_computation_client.h
```diff
@@ -140,6 +143,35 @@ class PjRtComputationClient : public ComputationClient {
     std::shared_ptr<PjRtBuffer> buffer;
   };
 
+  struct PjRtShardedData : public Data {
+    PjRtShardedData(std::string device, Shape shape)
+        : Data(std::move(device), std::move(shape)) {}
```
JackCaoG, line 148: what's this constructor for?

torch_xla/csrc/tensor.cpp
```diff
@@ -630,29 +630,24 @@ std::string XLATensor::DumpHloComputation(
...
-  XLA_CHECK(data()->ir_value.node != nullptr)
-      << "Tyring to access a null cursor";
+XLATensor::ShardingSpecPtr XLATensor::sharding_spec() const {
+  XLA_CHECK(GetIrValue().node != nullptr) << "Tyring to access a null cursor";
+  auto* sharding =
+      dynamic_cast<XlaNode*>(data()->ir_value.node.get())->GetSharding();
+  if (sharding == nullptr) {
+    return nullptr;
+  }
+  return std::make_shared<ShardingSpec>(*sharding);
```
JackCaoG, line 640: why do you need shardingspec to be a shared_ptr when you store it as raw_ptr in ir?

torch_xla/csrc/tensor_util.cpp
```diff
@@ -904,6 +904,70 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
...
+  XLA_CHECK_EQ(tensors.size(), devices.size());
+
+  std::vector<xla::ComputationClient::DataPtr> handles;
+  for (size_t i = 0; i < tensors.size(); ++i) {
+    torch::lazy::BackendDevice device = ParseDeviceString(devices[i]);
+    xla::Shape shape = CreateComputationShapeFromTensor(tensors[i], &device);
+
+    std::vector<xla::ComputationClient::TensorSource> source_tensors;  // in
+    std::vector<xla::ComputationClient::DataPtr> new_handles;          // out
+    if (shardings[i] != nullptr) {
```
JackCaoG, line 922: I think we should just assert on `shardings`. If there is no sharding, this function should not be called. User can user the other helper function

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -86,6 +86,65 @@ xla::HloModuleProto ShardingUtil::SpmdPartitioningPass(
   return module.get()->ToProto();
 }
 
+std::vector<std::vector<xla::ComputationClient::DataPtr>>
+ShardingUtil::InputHandler(
```
JackCaoG, line 90: can you add a comment to this function? Name doesn't tell me much regarding how I should use this function

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -86,6 +86,65 @@ xla::HloModuleProto ShardingUtil::SpmdPartitioningPass(
...
+    } else {
+      // Input is replicated across addressable devices
+      int64_t source_device_i =
+          ParseDeviceString(shards[0]->device()).ordinal();
+      arguments_by_device[source_device_i][argument_i] = shards[0];
+
+      auto literal = std::make_shared<xla::Literal>(std::move(
+          xla::ComputationClient::Get()->TransferFromServer(shards)[0]));
+      std::vector<xla::ComputationClient::TensorSource> source_tensors;
+      for (int64_t device_i = 0; device_i < devices.size(); ++device_i) {
```
JackCaoG, line 114: This part of the code seems redundant, can we reuse the `CreateTensorsData` logic above. Also please add a comment here explaining why we delay the replicated data uploading until here.

torch_xla/csrc/xla_sharding_util.cpp
JackCaoG: nvm, say it in .h

torch_xla/csrc/tensor.cpp
yeounoh: We store it as a shared ptr in IR, and I think we should prefer returning a shared ptr instead of the raw. It's implicitly being moved here with the return statement here.

torch_xla/csrc/tensor_util.cpp
yeounoh: We should also handle the implicitly replicated cases here, in which case the sharding[i] is null. I think it's easier to understand to handle these two cases in one place, as they are both related to SPMD mode. I could consider refactoring this a little bit, in the next PR?

third_party/xla_client/pjrt_computation_client.h
yeounoh: It's a base constructor we should implement because we are implementing Data.

third_party/xla_client/BUILD
yeounoh: Not needed, but good to have.

test/cpp/test_xla_sharding.cpp
yeounoh: Yea, that's the goal.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Ack.

torch_xla/csrc/tensor_util.cpp
yeounoh: Ack.

test/cpp/test_xla_sharding.cpp
JackCaoG: we are trying to merge this pr and will run this test as part of the CI. Would be nice if it works on PJRT:CPU when we merge otherwise we have no way of knowing other pr breaks it

third_party/xla_client/pjrt_computation_client.h
JackCaoG: It seems wrong that we construct a `PjRtShardedData` without shards

torch_xla/csrc/tensor_util.cpp
```diff
@@ -904,6 +904,70 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
...
+            };
+        source_tensors.emplace_back(std::move(shard_shape),
+                                    shard_device.toString(),
+                                    std::move(populate_fn));
+      }
+      new_handles.push_back(
+          xla::ComputationClient::Get()->TransferShardsToServer(
+              source_tensors, devices[i], shape));
+    } else {
+      auto populate_fn =
```
JackCaoG, line 954: can you add a comment here explain the implict replicatation?

third_party/xla_client/pjrt_computation_client.h
yeounoh: Right, also we don't use it. Let me make a note of this, and see if we can avoid declaring it (e.g., delete) -- in the next PR.

torch_xla/csrc/tensor.cpp
yeounoh: ^ sorry @JackCaoG it's possible that this (2nd) PR doesn't have that change, and I used to save the raw_ptr before.

torch_xla/csrc/tensor_util.cpp
yeounoh: Ack

test/cpp/test_xla_sharding.cpp
yeounoh: Just confirmed, this works on both PJRT:CPU and PJRT:TPU cc @JackCaoG

JackCaoG: can we make sure tests runs on PJRT:CPU on CI?

third_party/xla_client/pjrt_computation_client.h
yeounoh: Ok, delete works.

torch_xla/csrc/tensor_util.cpp
yeounoh: Done.

torch_xla/csrc/tensor_util.cpp
yeounoh: Done.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Done

test/cpp/test_xla_sharding.cpp
```diff
@@ -64,5 +64,81 @@ TEST_F(XLAShardingTest, ShardTensor) {
...
+  }
+
+  std::vector<at::Tensor> tensors(2);
+  std::fill_n(tensors.begin(), tensors.size(),
+              at::ones({8, 8}, at::TensorOptions(at::kFloat)));
+  std::vector<std::string> devices(2);
+  std::fill_n(devices.begin(), devices.size(), GetDefaultDevice()->toString());
+  std::vector<XLATensor::ShardingSpecPtr> shardings = {
```
JackCaoG, line 77: I really think we should break these into 2 tests.

test/cpp/test_xla_sharding.cpp
```diff
@@ -64,5 +64,81 @@ TEST_F(XLAShardingTest, ShardTensor) {
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 7, 4}));
 }
 
+TEST_F(XLAShardingTest, CreateTensorsData) {
```
JackCaoG, line 67: nit: `TestCreateTensorsData`  instead of `CreateTensorsData`. same for other tests

JackCaoG (approved):
Mostly lgtm, I am ok with merging this one. I will circle back and fix some of my own comments.

test/cpp/test_xla_sharding.cpp
yeounoh: I think {TEST_NAME}.{FUNCTION_IN_TEST}, e.g., `XLAShardingTest.CreateTensorsData` better, as-is?

test/cpp/test_xla_sharding.cpp
yeounoh: I see, could you elaborate more? If it's the FUNCTION_IN_TEST related and a couple of cases (sharding & replicated), then we can keep it as one. I think optionally we can break this into two, per a style preference. But I could be missing something.

JackCaoG: I think we should have one `TestNullSharding` and one `TestReplicated`. Each test should only test one behavior.

test/cpp/test_xla_sharding.cpp
JackCaoG: It is to be consistent with existing c++ test https://github.com/pytorch/xla/blob/master/test/cpp/test_aten_xla_tensor.cpp

yeounoh: +1 Thanks, consistency is more important

PR #3966, opened 2022-09-02
Support Tensor.is_alias_of

alanwaketan:
Summary:
Tensor.is_alias_of relies on Storage to perform. However, XLATensorImpl was
not implemented with that in mind. This commit adds a fake storage to XLATensor
as a marker to mark XLATensor that point to the same storage. The reason
why it's not done at XLATensorImpl is that XLATensor maintains the view ops/alias
logic in XLATensor class instead of relying on XLATensorImpl to do the check.

Test Plan:
./test/cpp/build/test_ptxla --gtest_filter=AtenXlaTensorTest.TestViewIsAliasOf


alanwaketan:
This is the same as #3960.

JackCaoG (approved):

PR #3967, opened 2022-09-03
Make expand to support is_alias_of

alanwaketan:
Summary:
Expand is diffcult to express by using view tensors, and therefore the
storage bit or is_alias_of bit will not be correctly set. Set that manually.

Test Plan:
./test/cpp/build/test_ptxla --gtest_filter=AtenXlaTensorTest.TestExpand

### DIFF
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -1087,9 +1087,11 @@ XLATensorPtr XLATensor::exp(const XLATensorPtr& input) {
 XLATensorPtr XLATensor::expand(const XLATensorPtr& input,
                                std::vector<int64_t> size) {
   auto input_shape = input->shape();
-  return input->CreateFrom(torch::lazy::MakeNode<Expand>(
+  auto output = input->CreateFrom(torch::lazy::MakeNode<Expand>(
       input->GetIrValue(),
       GetExpandDimensions(input_shape.get(), std::move(size))));
+  output->storage_ = input->Storage();
+  return output;
 }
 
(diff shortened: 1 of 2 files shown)


alanwaketan:
There are two other ops we probably need to tweak a bit too: split and split_with_sizes. Here is a complete list of view ops and whether we need tweaks or not: http://shortn/_NJBOu5TaRe. So far, I haven't encountered any issues while testing them. expand however will hit an assertion without this change.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1087,9 +1087,11 @@ XLATensorPtr XLATensor::exp(const XLATensorPtr& input) {
 XLATensorPtr XLATensor::expand(const XLATensorPtr& input,
                                std::vector<int64_t> size) {
   auto input_shape = input->shape();
```
JackCaoG, line 1089: I am a bit hesitant because we actually did not lower the expand as view(and yes this is a bug). I felt like in this case we should not trying to make `is_alias` works for expand tensor.

alanwaketan: For view ops with operands that have storages, PyTorch asserts the results should be alias of the inputs. Here is the assertion: https://paste.googleplex.com/4898491405959168. Nick tried once making expand as a proper view op and failed. I guess it's probably hard to implement. Ultimately we can leverage functionalization to properly fix this kinds of view problems all at once. I guess we can just patch the assertion and wait until LTC migration fully completes, and leave a small window for potential bugs.

Do you anticipate any issues we might have?

alanwaketan:
Thanks for approving this pull request, Jack.

Issue #3949, opened 2022-08-29
PJRT `xm.get_ordinal` and `xm.all_gather` fails in backward pass due to thread locals not accessible from autograd


ronghanghu:
## 🐛 Bug

https://github.com/pytorch/xla/pull/3902 addressed the issue of mismatched ranks in PJRT collective ops and made `xm.get_ordinal` return the corresponding ranks used by collective ops such as `xm.reduce_scatter`, which resolves many PJRT use cases.

However, further tests show that the thread locals `pjrt._PJRT_ORDINALS` (which `xm.get_ordinal` and consequently `xm.all_gather` are building upon) can only be read correctly during the forward pass, but cannot be read during the backward pass when invoked by PyTorch's autograd engine.

This breaks several use cases where one needs to call `xm.get_ordinal` or `xm.all_gather` in the autograd backward pass, such as in a custom-defined `torch.autograd.Function` or in more complicated backward hooks such as FSDP or other cases.

(This issue was discovered when testing FSDP under PJRT, which involves calling `xm.all_gather` from a backward hook of a network module.)

## To Reproduce

The example below contains a simple test case with a PyTorch autograd function `MyCustomFunction` that defines a dummy forward and backward method. It shows that `xm.get_ordinal` always returns 0 when invoked in the backward pass. This is because the thread locals `pjrt._PJRT_ORDINALS` cannot be accessed from the autograd engine that calls the backward functions or hooks.

1. Allocate a new v3-8 TPU VM with `tpu-vm-pt-1.12` runtime and install the nightly PyTorch/XLA wheels that contain https://github.com/pytorch/xla/pull/3704
```
# torch, torchvision and torch_xla
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch-nightly+20220829-cp38-cp38-linux_x86_64.whl
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torchvision-nightly+20220829-cp38-cp38-linux_x86_64.whl
sudo pip3 install https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch_xla-nightly+20220829-cp38-cp38-linux_x86_64.whl

# libtpu
sudo pip3 install https://storage.googleapis.com/cloud-tpu-tpuvm-artifacts/wheels/libtpu-nightly/libtpu_nightly-0.1.dev20220623-py3-none-any.whl
```
[...]

PR #3973, opened 2022-09-06
[PJRT] Remove thread-local replica ordinals

will-cromar:
See #3949 for context

- Replace thread-local ordinals with the default `BackendDevice`'s ordinal.
- Save local world size (i.e. the number of processes on a host) in the `LOCAL_WORLD_SIZE` environment variable. This should be consistent with [`torchrun`](https://pytorch.org/docs/stable/elastic/run.html#environment-variables), although I haven't tested it.

Ran `xla/test/pjrt/test_experimental_pjrt_tpu.py` manually on v3-8 and v4-8

### DIFF
--- a/test/pjrt/test_experimental_pjrt_tpu.py
+++ b/test/pjrt/test_experimental_pjrt_tpu.py
@@ -144,6 +144,45 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
         devices,
         [torch.device(f'xla:{i}') for i in range(expected_num_devices)])
 
+
+  @parameterized.named_parameters(('xla_model', xm.get_ordinal),
+                                  ('pjrt', pjrt.global_ordinal))
+  def test_global_ordinal(self, ordinal_func):
+    accelerator_num_devices = {
+        'v3-8': 8,
+        'v4-8': 4,
+    }
+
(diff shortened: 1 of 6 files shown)


test/pjrt/test_experimental_pjrt_tpu.py
```diff
@@ -144,6 +144,42 @@ def test_default_xla_devices(self):
         devices,
         [torch.device(f'xla:{i}') for i in range(expected_num_devices)])
 
+  @parameterized.named_parameters(('xla_model', xm.get_ordinal),
+                                  ('pjrt', pjrt.global_ordinal))
```
JackCaoG, line 148: I guess functionally `xm.get_ordinal` just calls `pjrt.global_ordinal` so they are the same?

JackCaoG (approved):
I don't think we run `test/pjrt/test_experimental_pjrt_tpu.py` on TPU nightly, maybe we should consider adding a set of PJRT TPU test on xlml test separately.

test/pjrt/test_experimental_pjrt_tpu.py
will-cromar: Right. This is confirming we get the same results from both endpoints in case the implementation changes (on purpose or by accident) in the future.

PR #3974, opened 2022-09-07
Companion PR for pytorch/pytorch#84579 

ezyang:
pytorch/pytorch#84579


JackCaoG:
If I understand correctly, now we have to opt-in symint support in a op based. If we put a op under `symint`, we can define the `op_symint` variant that have takes size as `symint`?

ezyang:
> If I understand correctly, now we have to opt-in symint support in a op based. If we put a op under symint, we can define the op_symint variant that have takes size as symint?

That's right. And if you don't opt in, no change on XLA side is needed.

JackCaoG (approved):
Thanks!

ezyang:
subsumed by https://github.com/pytorch/xla/pull/3978

PR #3978, opened 2022-09-07
Make `ExpectCounterChanged` to account for `_symint` ops without fixing the tests or ops explicitly

Krovatkin:
Make ExpectCounterChanged to account for _symint ops without fiMake ExpectCounterChanged to account for _symint ops without fixing the tests or ops explicitlyxing the tests or ops explicitly.

we are hitting test failures w/ counters after we symintified a few kernel and added the "_symint" suffix. Unfortunately, the failures are bit cryptic and requires some investigation: https://github.com/pytorch/pytorch/runs/8231927032?check_suite_focus=true . These failures would happen every time we rename an op. 
This PR tweaks `ExpectCounterChanged` to account for possible renames of ops to `_symint`.


Krovatkin:
@JackCaoG  lets see if this'd do it

ezyang:
Looks like it worked. So I can land this companion diff instead of mine when we finally land.

JackCaoG (approved):
Thanks Nick!

test/cpp/torch_xla_test.cpp
```diff
@@ -51,13 +47,43 @@ void XlaTest::ExpectCounterNotChanged(
...
+      start_msnap_->CounterChanged(counter_regex, *end_msnap_, ignore_set);
+
+  ExpectCounterNotChanged_(changed);
+
+  // Some operators could've been renamed to `opName_symint`, yet the tests are
+  // using the old names. We modify `ExpectCounterNotChanged` to also check
+  // `opName_symint` counters. When we finish migrating the ops to symints, we
+  // would remove this logic and fix all the tests
+  auto changed_symint =
+      start_msnap_->CounterChanged(counter_regex, *end_msnap_, ignore_set);
```
miladm, line 64: Why is this line not saying `counter_regex + "_symint"`? I am trying to understand the diff between line `54` and `63` here.

CC @JackCaoG @Krovatkin

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -3035,8 +3035,8 @@ std::tuple<at::Tensor, at::Tensor> XLANativeFunctions::var_mean(
                          bridge::AtenFromXlaTensor(std::get<1>(results)));
 }
 
-at::Tensor XLANativeFunctions::view(const at::Tensor& self,
-                                    at::SymIntArrayRef sym_size) {
+at::Tensor XLANativeFunctions::view_symint(const at::Tensor& self,
```
miladm, line 3038: To catch up: 

+ I understand this PR only enables `SymInt` for `view`
+ I don't expect `view` is ready for Dynamic Shape through this change - i.e. more work would be needed for DS enablement. The same applies to `empty`, `narrow_copy`.

@Krovatkin @JackCaoG please confirm (or correct me if inaccurate).

ezyang: Yeah, all this does is change the calling convention; but the insides still assert that the int is not symbolic. Once the IR side is ready, you can just modify this function to track it.

Future SymInt'ified schemas will not get updates like this, because we subsequently added an interlock (the `symint:` field in the yaml) which controls which calling convention you get. But there's a few stragglers like this from before we had this interlock.

miladm: > Once the IR side is ready, you can just modify this function to track it.

Is this part of an existing PR already? If so, please point me to it so I follow your changes accordingly.

@ezyang

ezyang: no, that was up to you guys :)

test/cpp/torch_xla_test.cpp
miladm: I will push a PR shortly.

PR #3979, opened 2022-09-07
switching to XLASymIntNodeImpl + test : part I

Krovatkin:


torch_xla/csrc/tensor.h
```diff
@@ -26,6 +27,13 @@
 
 namespace torch_xla {
 
+class TORCH_API XLASymIntNodeImpl : public c10::SymIntNodeImpl {
+ public:
+  XLASymIntNodeImpl(torch::lazy::NodePtr ptr) : node_(std::move(ptr)){};
+  c10::SymIntNode mul(const c10::SymIntNode& other) override;
+  torch::lazy::NodePtr node_;
```
JackCaoG, line 34: random question, do we want `node_` to be private and provide a getter? I am OK if upstream also mark it as a public member.

JackCaoG (approved):

JackCaoG:
You might need to rebase(and fix the linter)

torch_xla/csrc/tensor.h
Krovatkin: Made it private and provided a getter. That's indeed better.

JackCaoG:
seems like you need to rebase again, there were too many pr went into recently to update the ds bits of things lol

test/cpp/test_symint.cpp
```diff
@@ -129,5 +129,40 @@ TEST(SymintTest, TestDynamicSymints) {
...
 }
 
+TEST(SymintTest, TestDynamicSymintArithmetic) {
+  torch::lazy::Value scalar_value =
+      torch::lazy::Value(ScalarOp(1.0, xla::F32), 0);
+  // Assign a incorrect 3d shape for the test purpose
+  std::vector<torch::lazy::Shape> abs_lazy_shapes = {
+      torch::lazy::Shape(torch::kFloat, {10, 20, 30})};
+  torch::lazy::NodePtr abs_node =
+      torch::lazy::MakeNode<Abs>(scalar_value, std::move(abs_lazy_shapes));
```
JackCaoG, line 139: fyi, this will break after I merge https://github.com/pytorch/xla/pull/3954 since now SizeNode checks xla::shape. check my test example in that pr. Using `expand` directly should be simpler.

JackCaoG (commented):
mostly lgtm, I think you need to update test after https://github.com/pytorch/xla/pull/3954

miladm (commented):
Reviewed this PR. Thanks @Krovatkin.

PR #3980, opened 2022-09-08
[DDP] Add a correctness test case

alanwaketan:
Summary:
This pull request adds a correctness test case that compare the parameters after each training step between a DDP + TPU setup and a single CPU setup. If the inputs and labels are sharded correctly, the stepped parameters should match each.

Test Plan:
PJRT_DEVICE=TPU python test/pjrt/test_ddp.py TestPjRtDistributedDataParallel.test_ddp_correctness

### DIFF
--- a/test/pjrt/test_ddp.py
+++ b/test/pjrt/test_ddp.py
@@ -1,6 +1,9 @@
 from absl.testing import absltest, parameterized
+import copy
+import torch
 import torch.distributed as dist
 import torch.nn as nn
+import torch.optim as optim
 from torch.nn.parallel import DistributedDataParallel as DDP
 import torch_xla.core.xla_model as xm
 import torch_xla.distributed.xla_backend
@@ -16,7 +19,23 @@ def _init_xla_backend(init_file: str):
       init_method=f"file://{init_file}",
       rank=rank,
       world_size=world_size)
+  return rank, world_size
 
+def _assert_all_close(parameters_a, parameters_b):
+  for param_a, param_b in zip(parameters_a, parameters_b):
+    assert torch.allclose(param_a.cpu(), param_b.cpu())
+
+def _train_step(model, inputs, labels, optimizer, loss_fn):
+  optimizer.zero_grad()
+
+  outputs = model(inputs)
+  loss = loss_fn(outputs, labels)
+  loss.backward()
+  optimizer.step()
+
+  xm.mark_step()
+
+  return loss
 
 class TestPjRtDistributedDataParallel(parameterized.TestCase):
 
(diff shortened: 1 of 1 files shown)


test/pjrt/test_ddp.py
```diff
@@ -16,7 +19,23 @@ def _init_xla_backend(init_file: str):
       init_method=f"file://{init_file}",
       rank=rank,
       world_size=world_size)
+  return rank, world_size
 
+def _assert_all_close(parameters_a, parameters_b):
```
will-cromar, line 25: Will [`numpy.testing.assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) or [`absltest`'s `assertSequenceAlmostEqual`](https://github.com/abseil/abseil-py/blob/a0ae31683e6cf3667886c500327f292c893a1740/absl/testing/absltest.py#L1000-L1021) work here instead?

test/pjrt/test_ddp.py
```diff
@@ -31,6 +50,49 @@ def _ddp_init(init_file: str):
...
+      ddp_labels = copy.deepcopy(cpu_labels[offset : offset + local_batch_size]).to(device)
+      ddp_loss = _train_step(ddp_model, ddp_inputs, ddp_labels, ddp_optimizer, loss_fn)
+      with torch.no_grad():
+        ddp_loss = ddp_loss / world_size
+        dist.all_reduce(ddp_loss)
+
+      # TODO(@alanwaketan): Investigate why the atol here is this low.
+      assert torch.allclose(cpu_loss, ddp_loss, atol=1e-02)
+      _assert_all_close(cpu_model.parameters(), ddp_model.parameters())
+      print(f"iteration {step}: cpu_loss = {cpu_loss}, ddp_loss = {ddp_loss}, cpu_model.parameters() ~= ddp_model.parameters()")
```
will-cromar: Will this print 100 times during the unit test? Can you instead print only the final value or guard this behind some verbosity flag?

JackCaoG (commented):
On CI, this test is being run as 1 CPU, I felt like that won't do real dpp check? https://github.com/pytorch/xla/blob/b7f54864a0933d6401a023e12ec6ccac1982e2d9/test/run_tests.sh#L105

@will-cromar I think PJRT multi CPU does not work?

will-cromar (approved):
Otherwise LGTM. Thanks!

will-cromar:
The CI only runs one CPU with PjRt until we work out how to use PjRt CPU with multiprocessing. The manual test on TPU is more representative of a real use case, but I think this is still useful to run in the CI as a sanity check until then.

JackCaoG (approved):

test/pjrt/test_ddp.py
alanwaketan: `nn.module.parameters()` is a generator. Therefore, normal array check helpers won't work. https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.parameters.

test/pjrt/test_ddp.py
alanwaketan: Good point. I was too lazy. Let me add that.

alanwaketan:
Thanks Jack and Will for approving this pull request.

PR #3982, opened 2022-09-08
Companion fix to process group creation after pytorch PR73164

esaliya:
PyTorch PR https://github.com/pytorch/pytorch/pull/73164 adds changes to extend the 3rd party distributed backend registration.  As part of this the internal data structure is changed to a named tuple of {creator_fn, bool}. This causes some of the tests in xla to fail as it expects a callable being returned for a particular backend name. 

This PR fixes this issue by calling the creator_fn of the namedtuple instead of the namedtuple itself.

PR #3983, opened 2022-09-09
[Distributed] Companion fix after pytorch PR73164

alanwaketan:
Summary:
Author: @esaliya

PyTorch PR pytorch/pytorch#73164 adds changes to extend the 3rd party
distributed backend registration. As part of this the internal data
structure is changed to a named tuple of {creator_fn, bool}. This
causes some of the tests in xla to fail as it expects a callable
being returned for a particular backend name.

This PR fixes this issue by calling the creator_fn of the namedtuple
instead of the namedtuple itself.

Test Plan:
python test/test_torch_distributed_xla_backend.py


alanwaketan:
Cc @esaliya. This is the effort to land #3982.

JackCaoG (approved):

kumpera:
Could we extend PTD's Backend to avoid this in the future?

alanwaketan:
> Could we extend PTD's Backend API to avoid this in the future?

Avoid what?

torch_patches/.torch_pin
```diff
@@ -0,0 +1 @@
+#73164
```
vanbasten23, line 1: This means this PR works only with the change in https://github.com/pytorch/pytorch/pull/73164. Then would you check in this file when you merge this PR, or would you wait until https://github.com/pytorch/pytorch/pull/73164 merges first and then discard change in this torch_patches/.torch_pin?

alanwaketan: I believe the Later. You can check out this guide for more details. https://github.com/pytorch/xla/blob/master/.circleci/README.md.

vanbasten23: Thanks!

PR #3984, opened 2022-09-09
Add ComputationClient metrics to PjRtComputationClient

jonb377:
This change adds ComputationClient metrics to functions which have been implemented in PjRtComputationClient.

### DIFF
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -83,11 +83,13 @@ ComputationClient::DataPtr PjRtComputationClient::CreateDataPlaceholder(
 
 std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
     absl::Span<const TensorSource> tensors) {
+  metrics::TimedSection timed(TransferToServerMetric());
   tensorflow::profiler::TraceMe activity(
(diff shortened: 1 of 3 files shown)


JackCaoG:
rebase should fix the build issue

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -114,31 +117,38 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
         std::make_shared<PjRtData>(tensor.device, tensor.shape, buffer);
     datas.push_back(data);
   }
+  OutboundDataMetric()->AddSample(total_size);
+  CreateDataHandlesCounter()->AddValue(datas.size());
```
jonb377, line 121: @JackCaoG It looks like PjRt doesn't have the same notion of "handles" as XRT. I'm not familiar with handles in XRT, so I wanted to get some clarification - do these "handle" metrics still make sense to use in PjRt, or should we make a new counter?

JackCaoG: yea handle is same as buffer in the `TransferToServer` context.

will-cromar (approved):
LGTM. Thanks!

Issue #3965, opened 2022-09-02
XLA profiler issues: 1) TPU device trace does not show trace annotations, and 2) PRJT only captures 2 cores


ronghanghu:
## 🐛 Bug

The XLA profiler has been a crucial tool for performance analysis on TPUs. In https://github.com/pytorch/xla/pull/3866, the profiler was also integrated into the PJRT runtime. However, a few problems as mentioned below still hinder applying the XLA profiler to real-world machine learning models for performance debugging. (This issue is submitted as a follow-up to the discussion in https://github.com/pytorch/xla/pull/3866#issuecomment-1234716910, and is part of the challenges we're facing for TPU profiling and performance debugging in benchmark models and our internal use cases.)

Specifically
* The XLA profiler is expected to propagate the `xp.Trace` annotations to "TensorFlow Name Scope" on the TPU device traces when running with `XLA_HLO_DEBUG=1` (as described in [Annotation propagation](https://cloud.google.com/blog/topics/developers-practitioners/pytorchxla-performance-debugging-cloud-tpu-vm-part-iii)) -- This feature is important for performance analysis of a neural network model (e.g. to show the time consumption of different submodules), and was working well under PT/XLA 1.10 (with `tpu-vm-pt-1.10` TPU VM environment). 
  * However, with the nightly PyTorch/XLA wheels, **under both PJRT and XRT runtime, the profiler now fails to propagate the trace annotations from `xp.Trace` to TPU device traces with `XLA_HLO_DEBUG=1`**. This suggests that something related to tracing and HLO annotation could be broken due to changes between PT/XLA 1.10 and the current nightly version.
* **Under the PJRT runtime, the XLA profiler can only capture the device traces of 2 cores (instead of 8 cores) on TPU v3**, whereas earlier in XRT the traces of all 8 cores can be captured on a v3-8 TPU VM.
  * Meanwhile, it's really good to see that under PJRT runtime the XLA profiler can also capture TPU device traces on a pod now (although it still captures only 2 TPU v3 cores on e.g. a v3-128 pod).

(Apart from these two problems described above, there is also a long-standing problem that under XRT runtime the XLA profiler cannot capture TPU device traces when running on a pod, as mentioned in https://github.com/pytorch/xla/issues/3446. But I guess that problem is less important if we are committed to transiting to PJRT :)

## To Reproduce

The example below contains a simple XLA profiler test case for both XRT and PJRT on a v3-8 TPU VM. The observation is that 

[...]

PR #3985, opened 2022-09-09
Making sure XLA_HLO_DEBUG populate the scope metadata

JackCaoG:
Fix the issue mentioned in https://github.com/pytorch/xla/issues/3965#issuecomment-1241384829, this is not the full fix.

We used to always populate the scope information in IR metadata but now it is defualt off. I think it is OK but we need to make sure `XLA_HLO_DEBUG` also populate scope metadata in IR.


JackCaoG:
@ronghanghu FYI

wonjoolee95 (approved):
Thanks!

PR #3990, opened 2022-09-12
Add upsample_nearest/bilinear implementation for CPU and GPU

ymwangg:
This PR adds lowering for upsample_nearest/bilinear on CPU and GPU.  The implementation is adapted from tensorflow tf2xla with additional tests. Note I need to lower atol and rtol from 1e-5 to 1e-4 in the upsample_bilinear C++ test to make it pass.

BTW do we know why TPU does not use this implementation and uses a custom_call implementation instead (maybe for performance reason)?

### DIFF
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -3992,6 +3992,7 @@ TEST_F(AtenXlaTensorTest, TestUpsampleNearest2D) {
     torch::Tensor xla_result = torch::upsample_nearest2d(xla_input, {uh, uw});
     AllClose(result, xla_result);
   });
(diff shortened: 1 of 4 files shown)


JackCaoG:
Thansk, let me know when this one is ready for review.

ymwangg:
@JackCaoG This PR is ready for review.

JackCaoG (approved):
Thanks!

cicirori:
Nice! Upsample ops were being used in so many models I've seen.
Coincidentally, I recently implemented a custom-call that wraps the torch cuda kernel to support upsample_bilinear2d. I'll compare the two implementations when I have time.

ymwangg:
@cicirori Cool! I'm very curious about the performance comparison of these two implementations.

PR #3991, opened 2022-09-13
Don't load libtpu on deployment_tests

JackCaoG:
This is to bypass the IPV6 libtpu issue in docker


JackCaoG:
I am going to force merge this one to unblock nightly wheel build

PR #3993, opened 2022-09-13
Remove "apply PyTorch source patches"

ezyang:
These days, there are really never any source patches.


JackCaoG:
we actually about to have one for tf. That script handles both tf patches and pytorch patches. We haven't have pytorch patches for years but we need tf patches to workaround the build issue from time to time.

ezyang:
sgtm

PR #3994, opened 2022-09-13
Adding SizeConstant

Krovatkin:


torch_xla/csrc/ops/dynamic_ir.h
```diff
@@ -74,4 +75,18 @@ class SizeDiv : public XlaNode, public torch::lazy::DimensionNode {
...
+                     public torch::lazy::DimensionNode {
+ public:
+  SizeConstant(int64_t val);
+  int64_t getStaticValue() const override { return value().to<int64_t>(); };
+  bool isSymbolic() const override { return false; };
+  std::string ToString() const override {
+    this->torch_xla::Scalar::ToString();
+  };
+  virtual XlaOpVector Lower(LoweringContext* loctx) const override {
+    TORCH_CHECK(false, "NYI");
```
Krovatkin, line 88: @JackCaoG 
I don't think we necessarily need to implement it if we aren't going to use XLA to evaluate size expressions? If we do, I can just call `Scalar::Lower` I believe?

JackCaoG: I think we would need lower and yes we can use the `Scalar::Lower`.

When we do `SizeAddNode` = `SizeConstant` + `SizeNode`, we want `SizeAddNode->lower` gives the correct result.

test/cpp/test_symint.cpp
```diff
@@ -85,6 +85,31 @@ TEST(SymintTest, TestDynamicSymint) {
...
+  // TODO: test getDynamicValue()
+  torch::lazy::Value scalar_value =
+      torch::lazy::Value(ScalarOp(1.0, xla::F32), 0);
+  // Manully assign the torch::lazy::shape to avoid calling shape fn in this
+  // test. Note that we have to use one of those codegen ops so they take
+  // lazy::shape in constructor.
+  std::vector<torch::lazy::Shape> abs_lazy_shapes = {
+      torch::lazy::Shape(torch::kFloat, {9})};
+  torch::lazy::NodePtr abs_node =
+      torch::lazy::MakeNode<Abs>(scalar_value, std::move(abs_lazy_shapes));
```
JackCaoG, line 104: ditto, we will use `xla::shape` for `SizeNode` so you might want to just use `torch::lazy::MakeNode<Expand>` so it will be expand to the real dimension and has a correct `xla::shape`

torch_xla/csrc/ops/dynamic_ir.h
Krovatkin: got it! so you are planning to use XLA to evaluate those expressions?

test/cpp/test_symint.cpp
Krovatkin: right! thanks!

torch_xla/csrc/ops/dynamic_ir.h
```diff
@@ -74,4 +75,18 @@ class SizeDiv : public XlaNode, public torch::lazy::DimensionNode {
   virtual XlaOpVector Lower(LoweringContext* loctx) const override;
 };
 
+class SizeConstant : public torch_xla::Scalar,
```
vanbasten23, line 78: Can you point me to the spec of this class? I was looking at the design doc "Bounded Dynamic Shape Design for PyTorch" by Milad but didn't find it.

JackCaoG: it is new.. well, it just represent a constant, so you can do something like `a.size + 5` where a is a tensor. You need a ir representation of this computation.

torch_xla/csrc/ops/dynamic_ir.h
JackCaoG: Yea, I think we have to otherwise we violoate the contract of the Node class. For every node, we expect after the lowered HLO represents its real value.

It is kind of awkward that we say a `SizeAddNode`'s lowering does not always reflect its real size.

JackCaoG: I am not sure if we are planning to support

```
t1 = torch.tensor(1)
t2 = some_dynamic_tensor
t1 *= (t2.size(0) + 5)
```

in which case the result of `SizeAddNode` in last line needs to be correct.

test/cpp/test_symint.cpp
```diff
@@ -85,6 +85,31 @@ TEST(SymintTest, TestDynamicSymint) {
...
+  torch::lazy::Value abs_value = torch::lazy::Value(abs_node, 0);
+  torch::lazy::NodePtr size_node =
+      torch::lazy::MakeNode<SizeNode>(abs_value, /*dim=*/0);
+
+  torch::lazy::NodePtr add19 = torch::lazy::MakeNode<SizeAdd>(sc10, size_node);
+  EXPECT_EQ(add25->getStaticValue(), 19);
```
vanbasten23, line 110: Do you mean add19->getStaticValue()?

test/cpp/test_symint.cpp
```diff
@@ -85,6 +85,31 @@ TEST(SymintTest, TestDynamicSymint) {
...
+  // lazy::shape in constructor.
+  std::vector<torch::lazy::Shape> abs_lazy_shapes = {
+      torch::lazy::Shape(torch::kFloat, {9})};
+  torch::lazy::NodePtr abs_node =
+      torch::lazy::MakeNode<Abs>(scalar_value, std::move(abs_lazy_shapes));
+  torch::lazy::Value abs_value = torch::lazy::Value(abs_node, 0);
+  torch::lazy::NodePtr size_node =
+      torch::lazy::MakeNode<SizeNode>(abs_value, /*dim=*/0);
+
+  torch::lazy::NodePtr add19 = torch::lazy::MakeNode<SizeAdd>(sc10, size_node);
```
vanbasten23, line 109: Wondering why we need to test `torch::lazy::MakeNode<SizeAdd>(sc10, size_node)` or what is it testing.

JackCaoG:
Rebase should fix the error. On a separate note, we recently update TF, before rebuilding tf, please apply the patch by `./script/apply_patch.sh` since we recently introduced a patch in https://github.com/pytorch/xla/tree/master/tf_patches

test/cpp/test_symint.cpp
vanbasten23: Never mind. Please disregard the comment.

miladm (approved):
LGTM

PR #3996, opened 2022-09-13
[PJRT] Flatten output of `pjrt.run_multiprocess`

will-cromar:
Previously, the output of `pjrt.run_multiprocess` was a dict like `{process_rank: {thread_rank: result}}`. PR #3902 removed the connection between the result of `xm.get_ordinal()` and the process or thread rank, so this output structure is no longer useful. Instead, return a dict like `{replica_ordinal: result}`.

- Update tests to use new output structure.
- Integrate with `xmp.spawn`. I was able to run the XRT MNIST example on v4-8 successfully without any changes.
   - Use the `torch.multiprocessing` context to match `xmp.spawn`.
- Improve type annotations in `pjrt.py`

### DIFF
--- a/test/pjrt/test_experimental_pjrt_tpu.py
+++ b/test/pjrt/test_experimental_pjrt_tpu.py
@@ -64,12 +64,8 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
 
   def test_xla_devices_single_process_all_chips(self):
     accelerator_devices = {
-        'v3-8': {
-            i: torch.device(f'xla:{i}') for i in range(8)
-        },
-        'v4-8': {
-            i: torch.device(f'xla:{i}') for i in range(4)
-        },
+        'v3-8': {i: torch.device(f'xla:{i}') for i in range(8)},
+        'v4-8': {i: torch.device(f'xla:{i}') for i in range(4)},
     }
 
     if self.accelerator_type not in accelerator_devices:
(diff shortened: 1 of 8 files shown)


JackCaoG:
rebase should fix the build issue.

test/pjrt/test_experimental_pjrt_tpu.py
```diff
@@ -241,16 +211,14 @@ def _reduce_scatter(pin_layout):
     )
     xm.mark_step()
 
-    return xm.get_ordinal(), out.cpu().numpy()
+    return out.cpu().numpy()
```
JackCaoG, line 214: why don't we need to return ` xm.get_ordinal()` here?

JackCaoG (commented):
the change mostly lgtm, I have some high level questions
1. Have we tested this with pod?
2. Do we need the PJRT.multiprocess api anymore if `xmp.spawn` works

test/pjrt/test_experimental_pjrt_tpu.py
will-cromar: `run_multiprocess` already gives the replica ordinal for each result, so we don't need to return it from this function.

will-cromar:
@JackCaoG 

1. I tested this on a v4-16.
2. Discussed offline. I can follow up with a CL that makes `run_multiprocess` private and replaces their usage in our examples.

JackCaoG (approved):
Thanks!

PR #3998, opened 2022-09-13
Make alias return a fresh TensorImpl.

ezyang:
This is needed to appease https://github.com/pytorch/pytorch/pull/84893
where we need to actually return fresh TensorImpl in the kernel
implementation; the AD machinery won't do this for you.


ezyang:
A pin shouldn't be necessary as this change is BC

JackCaoG (approved):
linter is not happy, otherwise lgtm

PR #4006, opened 2022-09-14
DimCast Helpers

Krovatkin:


torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -6,8 +6,17 @@
...
-  return dynamic_cast<const torch::lazy::DimensionNode*>(output.node);
+const torch::lazy::DimensionNode* DimCast(const torch::lazy::Node* node) {
+  return dynamic_cast<const torch::lazy::DimensionNode*>(node);
+}
+
+const torch::lazy::DimensionNode* DimCast(torch::lazy::Output output) {
+  return DimCast(output.node);
+}
+
+const std::shared_ptr<torch::lazy::DimensionNode> DimCast(
```
JackCaoG, line 17: I am slightly worried that function has the same name but different return type will create trouble/confusion. Maybe `DimPtrCast` (it is a bad name, but `DimSharedPtrCast` sounds weird...)

Krovatkin: If you feel strongly about it, I can rename it to `DimPtrCast`. Usually, though, overloading makes things much easier since we don't have to think about which function to call. It's always `Dimcast(X)->getStaticValue()`. From user's perspective, it doesn't really matter if they are getting a raw pointer or a smart one since she can call `getStaticValue()/getDynamicValue()` on both.
Please let me know what you think!

JackCaoG: ok, if the main use case for `DimCast` is to do `DimCast->` directly I am fine with it.

JackCaoG:
error seems real.

torch_xla/csrc/ops/dynamic_ir.cpp
Krovatkin: I think so!
`DimensionNode` is a very simple interface.

```
class TORCH_API DimensionNode {
 public:
  virtual bool isSymbolic() const {
    return false;
  };
  virtual int64_t getDynamicValue() const {
    TORCH_CHECK(false, "NYI");
  };
  virtual int64_t getStaticValue() const {
    TORCH_CHECK(false, "NYI");
  };
  virtual ~DimensionNode() = default;
};
```
We should be able to get rid of `isSymbolic` as well soonish.

Krovatkin:
@JackCaoG yeah I copied the implementations from the other PR but I pasted it in the wrong place, lol :)

JackCaoG (approved):
Thanks!

miladm (commented):
Reviewed this PR. Thanks @Krovatkin.

PR #4007, opened 2022-09-14
Fixing a race where processes other than ones saving can remove chkpt directory

amithrm:
Guarding the creation of the chkpt directories is required as we have noticed it can be accidentally deleted by workers who do not save checkpoints.


JackCaoG:
@amithrm can you rebase this pr and explain a bit more why checking on `saved_tensor` will help here?

amithrm:
Hi Jack, rebased..And the issue this patch tries to solve is that if only few processes try to write to data, the checkpoint path deletion/creation is not guarded. So, all the workers (even if they save or not) try to do it and this creates a race where the worker trying to save can get into a state where the directory is removed by other worker.

JackCaoG (approved):
Thanks!

PR #4011, opened 2022-09-15
reduce the overhead incurred in FetchTensorData

aws-rhsoln:
When training model with multiple workers, two ExecuteComputation are not completely pipelined. We had an issue listed here: https://github.com/pytorch/xla/issues/3434 . This issue talks about two methods- RunPostOrder and FetchTensorData. Most part of RunPostOrder method was overlapped with previous execution with the help of this PR: https://github.com/pytorch/xla/pull/3457 . However the FetchTensorData method still remained as a bottleneck. The reason for this was: FetchTensorData runs after previous execution is complete and it runs before the next execution. In this method, it does two things: Reset the IR value and create xla_data. 
From our experiments resetting these IR values and creating xla_data for the tensors in the graph can be expensive as the graph grows in size. In such cases, it becomes a bottleneck. 
To avoid such scenarios, this PR creates a copy of the IR values and xla_data for all the tensors when the previous execution is on. It then uses that copy of ir_values for creating a graph for compilation for the next execution.


JackCaoG:
Out of curiosity, how much speed up does this pr bring.

JackCaoG (commented):
This pr makes already pretty complicated lazy execution code more confusing, I want to understand how much speed up(and in what circustance, always, small model, large model etc) are we talking about here.

aws-rhsoln:
> This pr makes already pretty complicated lazy execution code more confusing, I want to understand how much speed up(and in what circustance, always, small model, large model etc) are we talking about here.

This PR brings 5-6% speedup on BERT large.

JackCaoG:
Thanks, I will try to review it today. If you have the profile of before and after that would be super helpful.

aws-rhsoln:
> Thanks, I will try to review it today. If you have the profile of before and after that would be super helpful.

I didn't get a chance to get it on GPU.. Will try to get it this week!

torch_xla/csrc/debug_util.cpp
```diff
@@ -53,15 +53,20 @@ DebugUtil::GraphFormat DebugUtil::GetDefaultGraphFormat() {
 
 std::string DebugUtil::GetTensorsGraphInfo(
     absl::Span<const XLATensorPtr> tensors, const std::vector<size_t>* indices,
```
JackCaoG, line 55: we don't really need `tensors ` and `indices` if we always pass`ir_values`, let's just clean this up

torch_xla/csrc/debug_util.h
```diff
@@ -25,6 +25,7 @@ class DebugUtil {
   static std::string GetTensorsGraphInfo(
       absl::Span<const XLATensorPtr> tensors,
```
JackCaoG, line 26: ditto

torch_xla/csrc/debug_util.h
```diff
@@ -25,6 +25,7 @@ class DebugUtil {
...
       const std::vector<size_t>* indices,
+      std::vector<torch::lazy::Value>* ir_values = nullptr,
```
JackCaoG, line 28: make this function always take `ir_values` is better

torch_xla/csrc/tensor.cpp
```diff
@@ -1769,20 +1802,26 @@ std::shared_ptr<XLATensor::Async> XLATensor::SyncTensorsGraphInternal(
     TensorCollectionBarrier(&coll);
     return nullptr;
   }
-  PostOrderData po_data = RunPostOrder(*tensors, &coll);
+  std::vector<torch::lazy::Value> ir_values;
+  std::vector<torch::lazy::BackendDataPtr> tensor_data_vec;
+  FetchTensorData(tensors, coll.config, coll.indices, ir_values,
```
JackCaoG, line 1807: I think `FetchTensorData` is no longer an accurate name. Can you rename it to something along the line of `ExtractIRAndPrepareXlaData`

JackCaoG:
I think I understand the intention of this pr.

FetchData is called after the lock, because it will Remove all Tensor's IR Value and then replace it with a placeholder. These placeholder will be updated after the device execution.

The problem here is that Creating PlaceHolder can be expensive when graph size is large, what this pr does is pre-create those `XLAData` so the expensive operation can be overlapped.

torch_xla/csrc/tensor.cpp
```diff
@@ -1769,20 +1802,26 @@ std::shared_ptr<XLATensor::Async> XLATensor::SyncTensorsGraphInternal(
...
+                  tensor_data_vec);
+  PostOrderData po_data = RunPostOrder(ir_values, &coll);
   DebugUtil::SaveTensorsGraphInfo("ScheduleSyncTensorsGraph", *tensors,
-                                  &coll.indices);
+                                  &coll.indices, &ir_values);
```
JackCaoG, line 1811: why do we have to do this change through, wouldn't using `tensors` and `coll.indices` still works?

torch_xla/csrc/tensor.cpp
```diff
@@ -1410,13 +1440,10 @@ std::vector<torch::lazy::BackendDataPtr> XLATensor::FetchTensorData(
...
     torch::lazy::BackendDataPtr xla_data = tensor->CurrentXlaData();
     if (xla_data == nullptr && config.force_xla_data) {
-      const torch::lazy::BackendDevice& tensor_device = tensor->GetDevice();
-      xla::Shape shape = MakeShapeWithDeviceLayout(
-          tensor->shape(), static_cast<XlaDeviceType>(tensor_device.type()));
-      xla_data =
-          WrapXlaData(xla::ComputationClient::Get()->CreateDataPlaceholder(
-              tensor_device.toString(), std::move(shape)));
-      tensor->SetXlaData(xla_data, config.sync_xla_data);
+      xla_data = tensor_data_vec[i];
```
JackCaoG, line 1443: why not use `tensor->SetXlaData` here?

JackCaoG:
@aws-rhsoln Unless you have a strong preference I will leave this pr out of the 1.13 release since that will be cut tmr. I will be out for a week so I won't be able to review it next week.

aws-rhsoln:
> @aws-rhsoln Unless you have a strong preference I will leave this pr out of the 1.13 release since that will be cut tmr. I will be out for a week so I won't be able to review it next week.

Yeah should be fine, we can put this in the next release.

torch_xla/csrc/tensor.cpp
aws-rhsoln: It may not work actually, since by this time, the tensor's ir_value is reset. Hence, we have to pass the ir_values along with tensors and indices. Currently tensors are required just to get the device of the tensor.

torch_xla/csrc/tensor.cpp
aws-rhsoln: SetXlaData also creates a new ir_value. We don't want to create ir value here because this is run after previous execution and before next execution.

torch_xla/csrc/tensor.cpp
aws-rhsoln: Fair point. Let me rename. Thanks!

torch_xla/csrc/debug_util.cpp
aws-rhsoln: The only reason for having tensors is - we get device from it

torch_xla/csrc/tensor.cpp
aws-rhsoln: Actually, you are right. Revisiting this code made me realize I can just call the SaveTensorsGraph before the reset. Actually we don't need to change this piece of code. Reverting it back

aws-rhsoln:
> > This pr makes already pretty complicated lazy execution code more confusing, I want to understand how much speed up(and in what circustance, always, small model, large model etc) are we talking about here.
> 
> This PR brings 5-6% speedup on BERT large. This should give a significant boost when the execution is comparable to model tracing and other overheads.

I tested this change using BERT large on GPU device. Seeing a 3% improvement in throughput.
Previously the gap between two executeComputations was about 9ms. This has now reduced to roughly 3ms
Screen Shot 2022-10-05 at 9 55 16 AM
Screen Shot 2022-10-05 at 4 11 51 PM

torch_xla/csrc/tensor.cpp
JackCaoG: can we add a comment here?

aws-rhsoln: yeah, let me add a comment!

JackCaoG:
This mostly lgtm, @alanwaketan can you take another pass? This pr will help us to shrink the gap between executions but it will change a bunch of API. What's the implication to the LTC migration we have?

alanwaketan:
Changing method signature will definitely affect the LTC migration plan. Wonder if this change can be applied to the upstream LTC as well? @aws-rhsoln

JackCaoG:
@alanwaketan yea, I think this change should also be upstreamed to LTC.

aws-rhsoln:
> @alanwaketan yea, I think this change should also be upstreamed to LTC.

Can you point me to the LTC code base, I can take a look

alanwaketan:
> > @alanwaketan yea, I think this change should also be upstreamed to LTC.
> 
> Can you point me to the LTC code base, I can take a look

Here we are: https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/lazy_graph_executor.h. It should be pretty similar.

aws-rhsoln:
> @alanwaketan yea, I think this change should also be upstreamed to LTC.

I haven't looked at the LTC code base, can you point me to that.

aws-rhsoln:
> > > @alanwaketan yea, I think this change should also be upstreamed to LTC.
> > 
> > 
> > Can you point me to the LTC code base, I can take a look
> 
> Here we are: https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/lazy_graph_executor.h. It should be pretty similar.

Oh cool! Taking a look

aws-rhsoln:
> > > > @alanwaketan yea, I think this change should also be upstreamed to LTC.
> > > 
> > > 
> > > Can you point me to the LTC code base, I can take a look
> > 
> > 
> > Here we are: https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/lazy_graph_executor.h. It should be pretty similar.
> 
> Oh cool! Taking a look

Okay looks like it is very similar to the one we have here. One question: Do we need to first merge LTC changes and then come here and merge this one? Or both are decoupled?

alanwaketan:
I guess we can see if the same technique can be applied to the upstream first, i.e., having a pull request first. If the upstream PR works, it shouldn't matter which pull request get merged first. We can review upstream LTC PRs too.

aws-rhsoln:
Here is the PR for lazy tensor core: https://github.com/pytorch/pytorch/pull/87119/files
@alanwaketan

aws-rhsoln:
If this PR looks good, can we merge it? @JackCaoG

JackCaoG:
@alanwaketan does the upstream one lgty? If upstream pr align the api with this one I can merge this pr to unblock aws team.

alanwaketan:
> @alanwaketan does the upstream one lgty? If upstream pr align the api with this one I can merge this pr to unblock aws team.

Reviewing it.

alanwaketan (approved):
Yea, it seems matching the upstreamed one. @JackCaoG Can you approve the upstream one?

JackCaoG:
There were still some unresolved comments on upsteram, @aws-rhsoln can you resolve those. Let me take last pass of this pr then I will merge it.

torch_xla/csrc/tensor.cpp
```diff
@@ -1570,7 +1602,12 @@ XLATensor::OpByOpAsync XLATensor::SyncTensorsGraphOpByOp(
                                   &coll.indices);
 
   std::vector<torch::lazy::Value> roots = CollectRoots(*tensors, coll.indices);
-  auto tensors_data = FetchTensorData(tensors, coll.config, coll.indices);
+  std::vector<torch::lazy::Value> ir_values;
+  std::vector<torch::lazy::BackendDataPtr> tensor_data_vec;
+  ExtractIRAndPrepareXlaData(tensors, coll.config, coll.indices, ir_values,
```
JackCaoG, line 1607: nit `ExtractIRAndPrepareXlaData_` instead since it will inplace update `ir_values` and `tensor_data_vec`

torch_xla/csrc/tensor.cpp
```diff
@@ -1769,20 +1805,26 @@ std::shared_ptr<XLATensor::Async> XLATensor::SyncTensorsGraphInternal(
     TensorCollectionBarrier(&coll);
     return nullptr;
   }
-  PostOrderData po_data = RunPostOrder(*tensors, &coll);
   DebugUtil::SaveTensorsGraphInfo("ScheduleSyncTensorsGraph", *tensors,
                                   &coll.indices);
+  std::vector<torch::lazy::Value> ir_values;
+  std::vector<torch::lazy::BackendDataPtr> tensor_data_vec;
+  ExtractIRAndPrepareXlaData(tensors, coll.config, coll.indices, ir_values,
```
JackCaoG, line 1812: ditto

JackCaoG (approved):
Thanks @aws-rhsoln ! I only have one minor nit. If you can rebase and update the function name. I will merge it when all test passed.

torch_xla/csrc/tensor.cpp
aws-rhsoln: Just curious, what's the reasoning behind `_`, is it for indicating in-place updates ?

aws-rhsoln: Oh nvm, I see it!

JackCaoG:
Thanks @aws-rhsoln !

miladm:
Thanks for the PR. 
@JackCaoG did we do any performance analysis on Google side to verify perf gains here?
CC @chandrasekhard2

JackCaoG:
no, I expect the gain is similar. This is not really device dependent.

PR #4014, opened 2022-09-16
[DDP] Add a test case for XRT

alanwaketan:
Summary:
This pull request extends the existing PJRT DDP correctness test to support XRT as well.

Test Plan:
MASTER_ADDR=localhost MASTER_PORT=6001 python test/test_ddp.py TestXrtDistributedDataParallel


will-cromar (approved):

test/pjrt/distributed_util.py
```diff
@@ -0,0 +1,88 @@
+from absl import logging
```
JackCaoG, line 1: why is there an empty file `test/pjrt/__init__.py`?

test/run_tests.sh
```diff
@@ -139,6 +139,7 @@ function run_mp_op_tests {
   run_xla_backend_mp python3 "$CDIR/test_torch_distributed_all_reduce_xla_backend.py"
   run_xla_backend_mp python3 "$CDIR/test_torch_distributed_multi_all_reduce_xla_backend.py"
   run_xla_backend_mp python3 "$CDIR/test_torch_distributed_reduce_scatter_xla_backend.py"
+  run_xla_backend_mp python3 "$CDIR/test_ddp.py"
```
JackCaoG, line 142: I guess on CPU this test is not doing anything since by default we have 1 CPU, but on GPU we can test it on 2 gpu at least.

test/pjrt/distributed_util.py
JackCaoG, line 1: can we move this file under `test/` instead of leaving it inside `test/pjrt`?

JackCaoG (approved):
minor nits, mostly lgtm

test/run_tests.sh
alanwaketan: Hmmm, why are we running these tests in cpu env then? I saw the tests do early exit if the device is CPU...

test/pjrt/distributed_util.py
alanwaketan: That's just more formal way of making a directory as a package.

alanwaketan:
Let me create some GPU nodes to reproduce the test failure. Was not able to reproduce it in my TPUVM.

alanwaketan:
Cannot reproduce with local 4 x v100 setup...

alanwaketan:
Sad, I cannot run the docker image directly as the compiled binary (nccl version) doesn't match my local V100's architecture. The CI is running T4...

I guess I will need some T4 machines. Maybe just let me file a FIXME and investigate that later.

JackCaoG:
it is not worth it to get a T4.. We do want this to be tested through. Given that you already verified it on V100 I would suggest

1. enable this test(or any ddp gpu test) on v100 on xlml test

This test should be run on TPU if it is part of `run_test.sh` so we will get coverage too.

alanwaketan:
Yea, it's part of `run_test.sh`. Okay, let me migrate some of the xlml tests to use the DDP wrapper.

alanwaketan:
Well, it passes this time. Let's observe if it's a flaky test in the CI.

PR #4018, opened 2022-09-19
Add setup for CI on TPUs

steventk-g:
In this PR:
* Cloudbuild and kubernetes file to run TPU tests for PJRT
* Using new Dockerfile and cuda installation script from https://github.com/pytorch/xla/pull/4044

### DIFF
--- a/test/pjrt/test_experimental_pjrt_tpu.py
+++ b/test/pjrt/test_experimental_pjrt_tpu.py
@@ -36,20 +36,11 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
     os.environ.pop(xenv.TPU_VISIBLE_CHIPS, None)
     os.environ.pop(xenv.TPU_PROCESS_BOUNDS, None)
 
-  @absltest.skipIf(tpu.get_tpu_env()['ACCELERATOR_TYPE'] == 'v3-8',
-                   'This test is not currently supported on v3 TPUVMs.')
+  @absltest.skipIf(
+      tpu.version() <= 3,
+      'This test is not currently supported on v3 TPUVMs or earlier.')
   def test_xla_devices_multiprocess(self):
     accelerator_devices = {
-        'v3-8': {
-            0: torch.device('xla:0'),
-            1: torch.device('xla:1'),
-            3: torch.device('xla:0'),
-            4: torch.device('xla:1'),
-            5: torch.device('xla:0'),
-            6: torch.device('xla:1'),
-            7: torch.device('xla:0'),
-            8: torch.device('xla:1'),
-        },
         'v4-8': {
             0: torch.device('xla:0'),
             1: torch.device('xla:0'),
@@ -66,6 +57,9 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
     devices_per_process = pjrt._run_multiprocess(xm.xla_device)
     self.assertDictEqual(devices_per_process, expected)
 
+  @absltest.skipIf(
+      tpu.version() <= 2,
+      'This test is not currently supported on v2 TPUVMs or earlier.')
   def test_xla_devices_single_process_all_chips(self):
     accelerator_devices = {
         'v3-8': {i: torch.device(f'xla:{i}') for i in range(8)},
@@ -83,6 +77,9 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
     devices = pjrt._run_multiprocess(xm.xla_device)
     self.assertDictEqual(devices, expected)
 
+  @absltest.skipIf(
+      tpu.version() <= 2,
+      'This test is not currently supported on v2 TPUVMs or earlier.')
   def test_xla_devices_single_process_one_chip(self):
     accelerator_devices = {
         'v3-8': {
@@ -113,6 +110,9 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
 
     xmp.spawn(_assert, nprocs=1)
 
+  @absltest.skipIf(
+      tpu.version() <= 2,
+      'This test is not currently supported on v2 TPUVMs or earlier.')
   def test_xla_devices_single_process_one_chip_one_device_spawn(self):
     accelerators = ['v3-8', 'v4-8']
 
@@ -124,6 +124,9 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
     with concurrent.futures.ProcessPoolExecutor(max_workers=1) as executor:
       executor.submit(self._fail_on_nonfirst_device).result()
 
+  @absltest.skipIf(
+      tpu.version() <= 2,
+      'This test is not currently supported on v2 TPUVMs or earlier.')
   def test_default_xla_devices(self):
     accelerator_num_devices = {
         'v3-8': 8,
@@ -145,6 +148,9 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
 
   @parameterized.named_parameters(('xla_model', xm.get_ordinal),
                                   ('pjrt', pjrt.global_ordinal))
+  @absltest.skipIf(
+      tpu.version() <= 2,
+      'This test is not currently supported on v2 TPUVMs or earlier.')
   def test_global_ordinal(self, ordinal_func):
     accelerator_num_devices = {
         'v3-8': 8,
@@ -162,6 +168,9 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
 
   @parameterized.named_parameters(('xla_model', xm.get_local_ordinal),
                                   ('pjrt', pjrt.local_ordinal))
+  @absltest.skipIf(
+      tpu.version() <= 2,
+      'This test is not currently supported on v2 TPUVMs or earlier.')
   def test_local_ordinal(self, ordinal_func):
     accelerator_num_devices = {
         'v3-8': 8,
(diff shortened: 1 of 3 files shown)


test/pjrt/test_experimental_pjrt_tpu.py
```diff
@@ -34,6 +34,8 @@ def setUp(self):
     # sub-processes will not be able to initialize the client witht the correct
     # settings.
 
+  @absltest.skipIf(tpu.get_tpu_env()['ACCELERATOR_TYPE'] == 'v3-8',
+                   'This test is not currently supported on v3 TPUVMs.')
   def test_xla_devices_multiprocess(self):
     accelerator_devices = {
         'v3-8': {
```
vanbasten23, line 41: do you need to remove 'v3-8' here since the test is not supported on v3?

test/tpu/pjrt_test_job.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: v1
```
vanbasten23, line 1: i wonder what this yaml file does

steventk-g: This defines the kubernetes workload that the cloud build executes

steventk-g: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

test/pjrt/test_experimental_pjrt_tpu.py
steventk-g: @will-cromar Do you think we will ever support this test on v3? If not, we may as well remove it from `accelerator_devices` here. If we do plan on supporting it eventually, we should probably keep the code.

will-cromar: Go ahead and remove the `v3-8` test config -- it's based on a bad assumption of mine. Then change the `NotImplementedError`s to `skip`s instead so we don't fail out when tests are not implemented yet.

test/tpu/Dockerfile
```diff
@@ -0,0 +1,127 @@
+ARG python_version=3.8
```
will-cromar, line 1: Can this be removed in favor of the experimental Dockerfile? Does it need any changes to support tests? https://github.com/pytorch/xla/blob/master/docker/experimental/Dockerfile

steventk-g: Yeah I am using that dockerfile locally to test right now. This is an old push, the PR is still in draft

steventk-g: Done

test/pjrt/test_experimental_pjrt_tpu.py
steventk-g: Done

test/pjrt/test_experimental_pjrt_tpu.py
```diff
@@ -33,18 +33,10 @@ def setUp(self):
     # sub-processes will not be able to initialize the client witht the correct
     # settings.
 
+  @absltest.skipIf(tpu.get_tpu_env()['ACCELERATOR_TYPE'] == 'v3-8',
```
miladm, line 36: from resourcing standpoint, is it not safer to target `v2-8` devices as a starting point?

will-cromar: If the tests pass on `v2-8`, I support this. Most of our usage is on `v3`, but `v2` is very similar and generally much more available.

I recently added [`tpu.version`](https://github.com/pytorch/xla/blob/2fd2ffeeb303245ccab76cd5287d9818565aaad2/torch_xla/experimental/tpu.py#L91-L98), which will make it easier to skip tests that only work on v4.

test/tpu/pjrt_test_job.yaml
```diff
@@ -0,0 +1,44 @@
...
+    securityContext:
+      privileged: true
+    image: gcr.io/$PROJECT_ID/pytorch-xla-test:$BUILD_ID
+    command:
+    - bash
+    - -c
+    - |
+      ls /root/anaconda3/envs/pytorch/lib/python3.8/site-packages/ | grep libtpu
+      sudo rm -rf /usr/local/lib/python3.8/dist-packages/libtpu*
+      sudo pip3 install torch_xla[tpuvm]
```
will-cromar, line 37: You can skip the steps to install libtpu if `BUNDLE_LIBTPU=1` is set during the build

test/tpu/pjrt_test_job.yaml
```diff
@@ -0,0 +1,44 @@
...
+  affinity:
+    nodeAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        nodeSelectorTerms:
+        - matchExpressions:
+          - key: tpu.googleapis.com/name
+            operator: In
+            values:
+            # Update with your TPU name
+            - gke-tpu
```
will-cromar, line 15: Do you need the `name` constraint here? These tests should work on any TPU with the correct `type`

test/tpu/cloudbuild.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - |
+    set -u
+    set -e
+    set -x
+
+    apt-get update
+    apt-get -y install gettext
+
+    gcloud container clusters get-credentials --zone $_CLUSTER_ZONE $_CLUSTER_NAME
+    export BUILD_ID=$BUILD_ID
```
will-cromar, line 52: Add a comment that you're just adding these variables to the shell environment, since it's probably not clear at first to someone who doesn't know GCB that well

test/pjrt/test_experimental_pjrt_tpu.py
miladm: Let's confirm and transition.

test/tpu/cloudbuild.yaml
```diff
@@ -0,0 +1,60 @@
+substitutions:
+  # GCS location to write temporary checkpoints and summaries.
+  _GCS_BUCKET: 'gs://pytorch-xla-tpu-ci-tests/'
+  # Name of GKE cluster in which to run Job.
+  _CLUSTER_NAME: 'tpu-cluster'
+  # Location of GKE cluster.
+  _CLUSTER_ZONE: 'europe-west4-a'
```
miladm, line 7: if a particular zone is stocked out, what happens. do we have a way to try another zone?

miladm (commented):
A general concern to consider:
If a user decides to use this script to run a CI test against their local branch, what happens? This is particularly interesting if there is an active CI test (against `master`) already in flight. 

+ What's the intended behavior then?
+ Do we have a queuing mechanism to support this scenario?
+ If we want to disallow this behavior, how to we block it? If we want to allow it, can we do it such that CI tests can be triggered via github (as opposed to a local dev node)?

I had a conversation about this with @steventk-g. It'd be worth thinking about this topic and address it at some point - if not as part of this PR.

test/tpu/cloudbuild.yaml
will-cromar: Stockouts won't be an issue because the TPUs are created in advance. If all of the TPUs in our pool are occupied, the request will have to wait for available resources. It will take some experimentation to get the TPU pool size right.

test/pjrt/test_experimental_pjrt_tpu.py
steventk-g: Migrated to v2. Leaving in this annotation because we still want to skip this test in the case that it ever does get run on TPU v3.

test/tpu/cloudbuild.yaml
steventk-g: Done

test/pjrt/test_experimental_pjrt_tpu.py
will-cromar: This test will still fail on v2, right? Can you change the skip condition to `tpu.version() <= 3`?

test/tpu/pjrt_test_job.yaml
steventk-g: Done

test/tpu/pjrt_test_job.yaml
steventk-g: Removed

test/pjrt/test_experimental_pjrt_tpu.py
steventk-g: Corrected this skip and added a bunch for tests that failed on v2

test/pjrt/test_experimental_pjrt_tpu.py
```diff
@@ -160,6 +168,9 @@ def test_global_ordinal(self, ordinal_func):
 
   @parameterized.named_parameters(('xla_model', xm.get_local_ordinal),
                                   ('pjrt', pjrt.local_ordinal))
+  @absltest.skipIf(
+      tpu.version() <= 2,
+      'This test is not currently supported on v2 TPUVMs or earlier.')
   def test_local_ordinal(self, ordinal_func):
     accelerator_num_devices = {
         'v3-8': 8,
```
will-cromar, line 176: This version of the test should also work on v2, but it would be clunky to specify `v2-8` and `v3-8` separately for each of these tests. Can you file a follow-up bug to add a better utility to specify the correct output for these tests?

will-cromar (approved):
LGTM. Great work!

test/pjrt/test_experimental_pjrt_tpu.py
steventk-g: Created https://github.com/pytorch/xla/issues/4383

PR #3922, opened 2022-08-24
Update TensorFlow pin (08-2022)

wonjoolee95:
[WIP] Update TensorFlow pin (08-2022)

Changes include:
- Updated path compiler/mlir/hlo/WORKSPACE to compiler/xla/mlir_hlo/WORKSPACE.
  - Reason: TF xla/hlo directory was moved (https://github.com/tensorflow/tensorflow/commit/6ed5af7784f3f1eaaea77cf1224b588e2521cf73). 
- Updated path tensorflow/core/platform/default to tensorflow/tsl/platform/default
  - Reason: TF //tensorflow/core/platform/default was moved to //tensorflow/tsl/platform/default as of https://github.com/tensorflow/tensorflow/commit/d727fc2b1ef78961b2eba7dbbfba0094ea2c8a0f
- Updated ConsumeValueOrDie() to value() in xla_client
  - Reason: `ConsumeValueOrDie()` was deprecated in favor of value() in TF as of https://github.com/tensorflow/tensorflow/commit/9dcc8d0cfa870090753a8108ef0d16ff1ebde049
- Update PjRtExecutable to PjRtLoadedExecutable
  - Reason: PjRtExecutable was introduced as a new base calss for PjRtLoadedExecutable as of https://github.com/tensorflow/tensorflow/commit/a9701644665aa06f8314ea1ce03c0240a7705f50
- Updated parameter type tensorflow::profiler::ToolOptions 
- Added tf_patch for tensorflow::profiler to bypass the profiler related import dependency error

PR #4030, opened 2022-09-21
Update libtpu to align with latest TF pin update

wonjoolee95:
Update libtpu to align with latest TF pin update (https://github.com/pytorch/xla/pull/3922)

---
Tested on TPUVM with XRT, PJRT through sanity tests and `resnet` with `--fake-data`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ import torch
 base_dir = os.path.dirname(os.path.abspath(__file__))
 third_party_path = os.path.join(base_dir, 'third_party')
 
-_libtpu_version = '0.1.dev20220623'
+_libtpu_version = '0.1.dev20220824'
 _libtpu_storage_path = f'https://storage.googleapis.com/cloud-tpu-tpuvm-artifacts/wheels/libtpu-nightly/libtpu_nightly-{_libtpu_version}-py3-none-any.whl'
 
 


JackCaoG:
hmm, this is a bit weird, I think our tf is ~ 08/13. Did we use 08/24 libtpu because libtpu build was failing in 08/13 ?

JackCaoG:
What's the resnet throughput? With fake data on XRT v3-8 it should be ~550 imag/sec per core

wonjoolee95:
With 08/24 libtpu, I can see ~480. The full output is: https://paste.googleplex.com/5788741690982400. I can re-try with 08/13 libtpu.

JackCaoG:
This is low, can we make sure we don't introduce speed regression. I should of check this before merging the tf pin update ...

JackCaoG:
ok, `Rate` is 570 after stable, this is about the same as last ones. I think we are good.

wonjoolee95:
Okay, everything seems also okay with the 08/16 libtpu. I can see rate upto ~580. Full output: https://paste.googleplex.com/6149501730422784. I'll update to 08/16 libtpu which is the date of last commit on the new TF pin to keep things aligned.

JackCaoG (approved):

JackCaoG:
no need to waiting for CI I guess, I will force merge this one

PR #4039, opened 2022-09-23
Allow einsum to take optional parameter to align with torch

janeyx99:
This is to allow https://github.com/pytorch/pytorch/pull/84890 to allow an einsum optimization, which currently breaks, see https://github.com/pytorch/pytorch/actions/runs/3114441739/jobs/5050779137.

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1056,7 +1056,8 @@ at::Tensor XLANativeFunctions::dot(const at::Tensor& self,
 }
 
 at::Tensor XLANativeFunctions::einsum(c10::string_view equation,
-                                      at::TensorList tensors) {
+                                      at::TensorList tensors,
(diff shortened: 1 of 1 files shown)


steventk-g:
Thanks for catching this! Please format the file so that the linter will pass 
`clang-format-7 -i torch_xla/csrc/aten_xla_type.cpp`

JackCaoG:
let's see if build pass, if that works, I can take over pr and fix linter. We need a branch in pytorch/xla directly to pin it from pytorch side. This pr is currently on a fork which make this process slightly complicated.

janeyx99:
Will the build pass when my pytorch PR hasn't landed yet (so the extra argument's not there)?

janeyx99:
@JackCaoG I may be able to give myself access and create a branch on xla directly--shall I try that instead?

janeyx99:
Closing in favor of https://github.com/pytorch/xla/pull/4040

(btw i fixed lint)

PR #4041, opened 2022-09-23
Make lint err messaging clearer with `git diff`

janeyx99:
This would help those not working on Linux to see the error messages better. I used git diff to see what I should change manually since my change was so small, and it would look like:
image

I modified the error message a bit in this commit so it makes more sense, though I'd also understand not wanting super verbose logs potentially.


JackCaoG (approved):
This is great thanks!

JackCaoG:
I think you had a typo through, it should be `at::` instead of `at:`

janeyx99:
Ya, I realized that after I saw the error message 😛

PR #4043, opened 2022-09-23
[PJRT] Broadcast worker ID from `TPU:0`

will-cromar:
This lets other hosts discover the IP of the host with `TPU:0` in a pod context.

Tested manually on a v4-16 where `TPU:0` was on worker 1.

### DIFF
--- a/torch_xla/experimental/pjrt.py
+++ b/torch_xla/experimental/pjrt.py
@@ -267,8 +267,11 @@ def broadcast_master_param(model: nn.Module) -> None:
   xm.mark_step()
 
 
-def rendezvous(tag: str, payload: bytes,
-               ordinals: Optional[List[int]]) -> List[bytes]:
+def rendezvous(tag: str,
+               payload: bytes,
+               ordinals: Optional[List[int]],
+               *,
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/pjrt.py
```diff
@@ -267,8 +267,11 @@ def broadcast_master_param(model: nn.Module) -> None:
   xm.mark_step()
 
 
-def rendezvous(tag: str, payload: bytes,
-               ordinals: Optional[List[int]]) -> List[bytes]:
+def rendezvous(tag: str,
+               payload: bytes,
+               ordinals: Optional[List[int]],
+               *,
```
JackCaoG, line 273: python noob, what does this * mean?

JackCaoG (approved):

JackCaoG:
@AlexWertheim FYI, with this pr, you don't need to manually set mesh port anymore

torch_xla/experimental/pjrt.py
will-cromar: This makes `mesh_port` a [keyword-only argument](https://peps.python.org/pep-3102/)

PR #4044, opened 2022-09-23
Significantly reduce Docker image size

will-cromar:
This PR brings the size of the GPU image down to `6.24 GB` from `17.2 GB`

- Use local `xla` directory instead of cloning from github. This allows building an image from a local working copy
- Separate build image from release image so we don't release build artifacts
- Break build steps up to take better advantage of caching to make debugging less painful. In particular, build TF in its own step because it takes forever
- Remove unnecessary dependencies
- Use smaller `python:X.Y-slim` base images

Needs follow-up:

- Create new Google Cloud Build config for this image
- Collect wheels and upload to GCS
- Start using in e2e tests
- Example of using `builder` image for development

This image still ended up being messier than I wanted, although there are still advantages compared to the old build that make it more reasonable to use for CI and development. We should merge this to unblock #4018 and keep working on it as part of our broader build refactoring work.

Tested manually on TPU VM and GPU

### DIFF
--- a/docker/experimental/Dockerfile
+++ b/docker/experimental/Dockerfile
@@ -6,11 +6,12 @@ ARG cuda_version=11.2
 ARG cudnn_version=8.1.1.33
 # Debian repo doesn't have libcudnn
 ARG cuda_repo=ubuntu1804
-ARG tf_cuda_compute_capabilities="7.0,8.0"
+ARG tf_cuda_compute_capabilities="7.0"
 
 ARG tpuvm=1
 ARG build_cpp_tests=0
 
+# Contains all build dependencies for PT/XLA
 FROM python:${python_version}-${debian_version} AS builder
 
 RUN apt-get update
@@ -56,12 +57,14 @@ RUN chmod +x /usr/local/bin/bazel
 RUN mkdir xla
 WORKDIR /pytorch/xla/
 
-COPY tf_patches/
-COPY third_party/ third_party/
-WORKDIR /pytorch/xla/tensorflow
-RUN git apply /pytorch/xla/tf_patches/patch.diff
+# Contains actual build artifacts
+FROM builder AS build
 
-WORKDIR /pytorch/xla
+COPY tf_patches/ tf_patches/
+COPY third_party/ third_party/
+
+RUN patch -d third_party/tensorflow -N -p1 < /pytorch/xla/tf_patches/patch.diff
+
 COPY build_torch_xla_libs.sh .
 
 ARG tpuvm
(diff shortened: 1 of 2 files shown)


docker/experimental/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+ARG tf_cuda_compute_capabilities
+RUN TPUVM_MODE=${tpuvm} XLA_CUDA=${cuda} TF_CUDA_COMPUTE_CAPABILITIES=${tf_cuda_compute_capabilities} bash build_torch_xla_libs.sh -O -D_GLIBCXX_USE_CXX11_ABI=1
+
+COPY torch_xla/ torch_xla/
+COPY scripts/ scripts/
+COPY setup.py .
+COPY xla_native_functions.yaml .
+
+ARG build_cpp_tests
+RUN BUILD_CPP_TESTS=${build_cpp_tests} TPUVM_MODE=${tpuvm} XLA_CUDA=${cuda} TF_CUDA_COMPUTE_CAPABILITIES=${tf_cuda_compute_capabilities} python setup.py bdist_wheel
```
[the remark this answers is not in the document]
will-cromar: Yeah, this should be possible, and we have a partner team looking into exactly that. In this image, I run the TF build separately so it ends up cached if successful, which is a marginal improvement

docker/experimental/Dockerfile
```diff
@@ -0,0 +1,122 @@
+ARG python_version=3.8
+ARG debian_version=buster
+
+ARG cuda=0
+ARG cuda_version=11.2
+ARG cudnn_version=8.1.1.33
+# Debian repo doesn't have libcudnn
+ARG cuda_repo=ubuntu1804
+ARG tf_cuda_compute_capabilities="7.0"
```
JackCaoG, line 9: given https://github.com/pytorch/xla/issues/4089, should we support more `tf_cuda_compute_capabilities` ?

docker/experimental/Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+
+RUN USE_CUDA=0 python setup.py bdist_wheel
+
+RUN pip install dist/*.whl
+
+RUN curl -LO https://github.com/bazelbuild/bazelisk/releases/download/v1.11.0/bazelisk-linux-amd64
+RUN mv bazelisk-linux-amd64 /usr/local/bin/bazel
+RUN chmod +x /usr/local/bin/bazel
+
+RUN mkdir xla
```
JackCaoG, line 57: I a, a bit lost here, seems like we didn't git clone xla repo, so where would the code come from?

docker/experimental/Dockerfile
will-cromar: Good idea. I'll merge that change.

docker/experimental/Dockerfile
will-cromar: You run this build from a working copy of PyTorch/XLA, so each of the files/directories needed for the build are copied in from the working directory in the `COPY` statements below. Each `COPY` in the `build` step only brings in what is needed to make the build incremental -- if any of the files in the `COPY` are different, all of the `RUN` steps have to run again and can't be cached. If you copied in everything at once, changes in `torch_xla` would force you to rebuild TensorFlow, for example.

The `builder` image doesn't have any PT/XLA checked out. My idea here is that you can check out PT/XLA on the host machine and mount it at `/pytorch/xla` in the container to have just the build dependencies preinstalled.

The whole working copy is copied into the `release` image at the end in `COPY . /pytorch/xla`.

JackCaoG (approved):
Thanks @will-cromar ! This is super exciting!

JackCaoG:
you can rebase to get over the build issue.

PR #4045, opened 2022-09-24
Minor clean up for expand_symint

JackCaoG:


torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -41,6 +41,9 @@ int64_t SizeNode::getDynamicValue() const {
   }
   torch::lazy::NodePtr cloned =
```
vanbasten23, line 42: What's the reason we need to clone a sizeNode, as opposed to using the current `this` SizeNode?

JackCaoG: wrapping `self` inside a new object(dummy_size_tensors) and pass it to another function is kind of weird. I also need to deal with const_cast if I do it that way. I think it is just easier to make a clone and no worry about poetical consequence.

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -41,6 +41,9 @@ int64_t SizeNode::getDynamicValue() const {
...
       torch::lazy::MakeNode<SizeNode>(operands_[0], dim_);
+  // Wrap the IR of SizeNode into a dummy tensor and execute/fetch the value
```
vanbasten23, line 44: By "IR of SizeNode", it seems you mean the variable `cloned` which is a `torch::lazy::NodePtr` object. I thought `torch::lazy::NodePtr` denote the SizeNode. So I'm a bit confused. What's the difference between the node and IR of the node?

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -41,6 +41,9 @@ int64_t SizeNode::getDynamicValue() const {
...
+  // of this tensor. GetTensors will return a cpu at::Tensor so we can just
+  // extract the value of it.
   std::vector<XLATensorPtr> dummy_size_tensors = {
```
vanbasten23, line 47: It seems we're using 
```
XLATensorPtr XLATensor::Create(
    torch::lazy::Value ir_value, const torch::lazy::BackendDevice& device,
    c10::optional<at::ScalarType> logical_element_type)
```
but `clone` is a torch::lazy::NodePtr object. How is the `torch::lazy::NodePtr` converted to a `torch::lazy:Value`? Is it by using torch::lazy:Value's implicit constructor
```
Value(NodePtr&& node, size_t index = 0)
      : node(std::move(node)), index(index) {}
```
?

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -41,6 +41,9 @@ int64_t SizeNode::getDynamicValue() const {
   }
   torch::lazy::NodePtr cloned =
       torch::lazy::MakeNode<SizeNode>(operands_[0], dim_);
+  // Wrap the IR of SizeNode into a dummy tensor and execute/fetch the value
+  // of this tensor. GetTensors will return a cpu at::Tensor so we can just
```
vanbasten23, line 45: If it is a TPU at::Tensor, we cannot extract the value of it?

JackCaoG: if it is a TPU tensor, it must be a `XLATensor`

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: Never mind, I think so. It uses the conversion constructor.

PR #4051, opened 2022-09-27
Fix a divide-by-0 error for SizeDiv node.

vanbasten23:
Also added a test.

### DIFF
--- a/test/cpp/test_ir.cpp
+++ b/test/cpp/test_ir.cpp
@@ -1,4 +1,5 @@
 #include <gtest/gtest.h>
+#include <stdexcept>
 
 #include "cpp_test_util.h"
 #include "torch_xla/csrc/ir.h"
@@ -329,6 +330,32 @@ TEST_F(IrTest, TestSizeDivNodeDynamic) {
       std::dynamic_pointer_cast<torch::lazy::DimensionNode>(node_div);
   EXPECT_EQ(dim_node_div->getStaticValue(), 50);
   EXPECT_EQ(dim_node_div->getDynamicValue(), 0);
+}
+
+TEST_F(IrTest, TestSizeDivNodeDynamicByZero) {
+  torch::lazy::Value scalar_value = torch::lazy::Value(ScalarOp(0.0, xla::F32), 0);
+  std::vector<int64_t> target_size = {2, 2};
+  torch::lazy::Value node = torch::lazy::Value(torch::lazy::MakeNode<Expand>(scalar_value, target_size));
+  torch::lazy::NodePtr nonzero_node = torch::lazy::MakeNode<NonZero>(node);
+
+  // static value = 4, dynamic value = 0
+  torch::lazy::NodePtr size_node_nonzero_0 =
+      torch::lazy::MakeNode<SizeNode>(nonzero_node, 0);
+  std::shared_ptr<torch::lazy::DimensionNode> dim_size_node_nonzero_0 =
+      std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node_nonzero_0);
+  // static value = 2, dynamic value = 2
+  torch::lazy::NodePtr size_node_nonzero_1 =
+      torch::lazy::MakeNode<SizeNode>(nonzero_node, 1);
+  std::shared_ptr<torch::lazy::DimensionNode> dim_size_node_nonzero_1 =
+      std::dynamic_pointer_cast<torch::lazy::DimensionNode>(size_node_nonzero_1);
+
+  torch::lazy::NodePtr node_div = torch::lazy::MakeNode<SizeDiv>(
+      torch::lazy::Value(size_node_nonzero_1, 0),
+      torch::lazy::Value(size_node_nonzero_0, 0));
+  std::shared_ptr<torch::lazy::DimensionNode> dim_node_div =
+      std::dynamic_pointer_cast<torch::lazy::DimensionNode>(node_div);
+
+  EXPECT_THROW(dim_node_div->getDynamicValue(), std::runtime_error);
 }
 
 }  // namespace cpp_test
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -137,6 +137,8 @@ int64_t SizeDiv::getDynamicValue() const {
   const torch::lazy::DimensionNode* dim_node_1 = DimCast(operand(1));
   XLA_CHECK(dim_node_0);
   XLA_CHECK(dim_node_1);
+  XLA_CHECK(dim_node_1->getDynamicValue() != 0)
```
JackCaoG, line 140: nit, save the value of `getDynamicValue` here. I am still debating whether we should cache the value of `GetDynamicValue` here. Each `getDynamicValue` is a recursive call and can be expensive depending on how many layers it has.

JackCaoG:
rebase should solve the build issue

torch_xla/csrc/ops/dynamic_ir.cpp
miladm: @JackCaoG according to this [line](https://github.com/pytorch/xla/blob/c2d85a006885a7718affa52feb891e753e811670/torch_xla/csrc/ops/dynamic_ir.cpp#L38-L41), the cache is already in place. Sounds like you have additional compute concern here?

JackCaoG: I mean the cache for `SizeAddNode`, we currently only have cache for `SizeNode`. `SizeAddNode` can poteianlly cause a regression call but it might be a overkill to cache it here since cpu compute is pretty quick now days.

JackCaoG: what I mean by save the value of `getDynamicValue` is don't call `dim_node_1->getDynamicValue()` twice here.

Krovatkin: ideally we should probably have `getDynamicValue()_` that does actual computation and `getDynamicValue()` (in a base class) that calls `getDynamicValue()_`, caches  the value and returns it. wdyt @JackCaoG  ?

test/cpp/test_ir.cpp
```diff
@@ -331,5 +333,35 @@ TEST_F(IrTest, TestSizeDivNodeDynamic) {
...
+      torch::lazy::Value(ScalarOp(0.0, xla::F32), 0);
+  std::vector<int64_t> target_size = {2, 2};
+  torch::lazy::Value node = torch::lazy::Value(
+      torch::lazy::MakeNode<Expand>(scalar_value, target_size));
+  torch::lazy::NodePtr nonzero_node = torch::lazy::MakeNode<NonZero>(node);
+
+  // static value = 4, dynamic value = 0
+  torch::lazy::NodePtr size_node_nonzero_0 =
+      torch::lazy::MakeNode<SizeNode>(nonzero_node, 0);
+  std::shared_ptr<torch::lazy::DimensionNode> dim_size_node_nonzero_0 =
```
Krovatkin, line 347: this is ded or were you going to use it?

test/cpp/test_ir.cpp
```diff
@@ -331,5 +333,35 @@ TEST_F(IrTest, TestSizeDivNodeDynamic) {
...
+      std::dynamic_pointer_cast<torch::lazy::DimensionNode>(
+          size_node_nonzero_0);
+  // static value = 2, dynamic value = 2
+  torch::lazy::NodePtr size_node_nonzero_1 =
+      torch::lazy::MakeNode<SizeNode>(nonzero_node, 1);
+  std::shared_ptr<torch::lazy::DimensionNode> dim_size_node_nonzero_1 =
```
Krovatkin, line 353: ded

test/cpp/test_ir.cpp
vanbasten23: good catch.

test/cpp/test_ir.cpp
vanbasten23: done

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: @Krovatkin It sounds creating a base class with a `getDynamicValue()` would dedup the cache implementation, right? How many such subclasses are there besides SizeAdd/SizeMul/SizeDiv?

JackCaoG: so you mean something similar to
```
DimensionNode::getDynamicValue_() {
 // everyNode need to implement it
}
DimensionNode::getDynamicValue() {
  if (!cached) {
     cached_val = getDynamicValue_();
  }
  return cached_val?
}
```

yea I guess it works and make things cleaner.

vanbasten23: SG. Let me add the cache to torch::lazy::DimensionNode then

JackCaoG (approved):

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: Will do the caching a different PR.

Krovatkin: > yea I guess it works and make things cleaner.

exactly!!!

Krovatkin: > How many such subclasses are there besides SizeAdd/SizeMul/SizeDiv?

Potentially as many as methods in SymIntNodeImpl?
20-ish something, I think

Krovatkin: > @Krovatkin It sounds creating a base class with a getDynamicValue() would dedup the cache implementation, right?

yes, that's correct! it's just a suggestion though, please feel to ignore it

Issue #4052, opened 2022-09-27
Python Ops Tests Fail for Einsum


steventk-g:
Seeing the following errors in python ops test, related to both `s64` and `f32` types. It looks like there are other tests failing with similar errors as well (https://github.com/pytorch/xla/blob/78e809245f95784b294f766267120d81c477e7e4/test/test_ops.py#L251)

```
======================================================================
2022-09-27 08:26:09.501 PDT

ERROR: test_reference_eager_einsum_xla_int64 (__main__.TestOpInfoXLA)
2022-09-27 08:26:09.501 PDT

----------------------------------------------------------------------
2022-09-27 08:26:09.501 PDT

Traceback (most recent call last): File "/root/anaconda3/envs/pytorch/lib/python3.7/site-packages/torch/testing/_internal/common_device_type.py", line 391, in instantiated_test raise rte File "/root/anaconda3/envs/pytorch/lib/python3.7/site-packages/torch/testing/_internal/common_device_type.py", line 378, in instantiated_test result = test(self, **param_kwargs) File "/root/anaconda3/envs/pytorch/lib/python3.7/site-packages/torch/testing/_internal/common_device_type.py", line 815, in test_wrapper return test(*args, **kwargs) File "/pytorch/xla/test/test_ops.py", line 415, in test_reference_eager self.compare_with_eager_reference(op, sample_input) File "/pytorch/xla/test/test_ops.py", line 406, in compare_with_eager_reference self.assertEqual(actual, expected, exact_dtype=True, exact_device=False) File "/pytorch/xla/test/pytorch_test_base.py", line 639, in assertEqual x, y = self.prepare_for_compare(x, y) File "/pytorch/xla/test/pytorch_test_base.py", line 582, in prepare_for_compare x = tx.to(device='cpu') RuntimeError: UNIMPLEMENTED: From /job:tpu_worker/replica:0/task:0:
2022-09-27 08:26:09.501 PDT

2 root error(s) found.
2022-09-27 08:26:09.501 PDT

 (0) UNIMPLEMENTED: While rewriting computation to not contain X64 element types, XLA encountered an HLO for which this rewriting is not implemented: %dot.3 = s64[3,4]{1,0} dot(s64[3]{0} %p1.2, s64[4]{0} %p0.1), lhs_contracting_dims={}, rhs_contracting_dims={}
2022-09-27 08:26:09.501 PDT

 [[{{node XRTCompile}}]]
2022-09-27 08:26:09.501 PDT

[...]

PR #4053, opened 2022-09-27
Move einsum to failing tests

steventk-g:
To resolve https://github.com/pytorch/xla/issues/4052, we are moving einsum to the comment section with the other failing ops.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -192,7 +192,6 @@ allowed_opinfo = set(
             AllowedOpInfoEntry('angle'),
             AllowedOpInfoEntry('linalg.solve'),
             AllowedOpInfoEntry('linalg.matrix_rank'),
-            AllowedOpInfoEntry('einsum'),
             AllowedOpInfoEntry('linalg.svd'),
             AllowedOpInfoEntry('linalg.svdvals'),
             AllowedOpInfoEntry('polar'),
@@ -250,6 +249,7 @@ allowed_opinfo = set(
 
             # Failing Ops
             # Refer for more info : https://github.com/pytorch/xla/pull/3019#issuecomment-877132385
+            # AllowedOpInfoEntry('einsum'), https://github.com/pytorch/xla/issues/4052
             # AllowedOpInfoEntry('cdist'),  // precision issue on TPU
             # AllowedOpInfoEntry('linalg.multi_dot'),  // failing on CPU
             # AllowedOpInfoEntry('matmul'),            // failing on CPU
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -79,6 +79,16 @@ bool IsOperationOnType(const c10::optional<at::ScalarType>& opt_dtype,
   return tensor_type == type;
 }
 
+bool TensorsAreOfType(std::vector<XLATensorPtr> tensors, at::ScalarType type) {
+  for (const XLATensorPtr& tensor : tensors) {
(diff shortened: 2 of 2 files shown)


test/test_ops.py
```diff
@@ -285,6 +284,7 @@ def __new__(cls, name, variant_test_name=""):
             # AllowedOpInfoEntry('exp'),
             # AllowedOpInfoEntry('diag_embed'),
             # AllowedOpInfoEntry('diagonal'),
+            # AllowedOpInfoEntry('einsum'),
```
wonjoolee95, line 287: nit: can you add the link to the github issue in the comment?

wonjoolee95:
@JackCaoG, the `test_reference_eager_einsum` tests continue to fail on TPUVM (succeeds on CPU/GPU). We're disabling these tests for now. Any thoughts on this?

JackCaoG:
we can just fallback when input type is 64 or complex for now, that way we can keep the test

steventk-g:
> we can just fallback when input type is 64 or complex for now, that way we can keep the test

The op fails for both `s64` and `f32`, both of the types we currently test. Since the test only runs on XLA, what's the point of falling back?

```
  @ops(ops_to_test, allowed_dtypes=(torch.float32, torch.long))
  def test_reference_eager(self, device, dtype, op):
    if self.device_type != 'xla':
      self.skipTest("This test runs only on XLA")

    sample_inputs = op.sample_inputs(device, dtype)
    for sample_input in sample_inputs:
      self.compare_with_eager_reference(op, sample_input)
```

JackCaoG:
I mean in the einsum implemantion side. What kind of type combination will crash on TPU?

JackCaoG:
Bottom line is, if without the new `einsum` lowering, test will not crash, we should fix the test by fallback to `at::native::einsum` for those unsupported case.

test/test_ops.py
```diff
@@ -25,220 +25,220 @@ def __new__(cls, name, variant_test_name=""):
...
-            AllowedOpInfoEntry('linalg.solve'),
-            AllowedOpInfoEntry('linalg.matrix_rank'),
+            # AllowedOpInfoEntry('abs'),
+            # AllowedOpInfoEntry('add'),
+            # AllowedOpInfoEntry('mul'),
+            # AllowedOpInfoEntry('sub'),
+            # AllowedOpInfoEntry('addmm'),
+            # AllowedOpInfoEntry('addmm', 'decomposed'),
+            # AllowedOpInfoEntry('addmv'),
+            # AllowedOpInfoEntry('addbmm'),
```
JackCaoG, line 35: Please uncomment these op tests

steventk-g: Done

test/test_ops.py
steventk-g: Done

test/test_ops.py
```diff
@@ -250,6 +249,7 @@ def __new__(cls, name, variant_test_name=""):
 
             # Failing Ops
             # Refer for more info : https://github.com/pytorch/xla/pull/3019#issuecomment-877132385
+            # AllowedOpInfoEntry('einsum'), https://github.com/pytorch/xla/issues/4052
```
wonjoolee95, line 252: @steventk-g, wasn't the intention of falling back to cpu so that we don't have to disable the `einsum` tests?

steventk-g: We still need to remove the test because precision tests for float32 are also failing

wonjoolee95: Then can we also fall back on those types? If this test was passing before the new `einsum` lowering, we should still expect this test to pass with the lowering implementation.

PR #4061, opened 2022-09-28
Codegen Softshrink

vanbasten23:


torch_xla/csrc/elementwise.cpp
```diff
@@ -128,16 +128,17 @@ xla::XlaOp BuildHardSwishBackward(xla::XlaOp grad_output, xla::XlaOp input) {
...
-      xla::Select(xla::Lt(input, Neg(xla_lambd)), input + xla_lambd, zero);
-  return xla::Select(xla::Le(input, xla_lambd), le_lambda_branch,
-                     input - xla_lambd);
+xla::XlaOp BuildSoftshrink(xla::XlaOp input, xla::XlaOp lambda) {
+  const xla::Shape& input_shape = XlaHelpers::ShapeOfXlaOp(input);
+  xla::PrimitiveType input_element_type = input_shape.element_type();
+  lambda = MaybeConvertTo(lambda, input_element_type);
+
+  xla::XlaOp zero =
+      XlaHelpers::ScalarValue<int64_t>(0, input_element_type, input.builder());
```
vanbasten23, line 137: Just wondering if we can use `lambda.builder()` to replace `input.builder()` this line?

torch_xla/csrc/elementwise.cpp
```diff
@@ -128,16 +128,17 @@ xla::XlaOp BuildHardSwishBackward(xla::XlaOp grad_output, xla::XlaOp input) {
...
-  const xla::Shape& shape = XlaHelpers::ShapeOfXlaOp(input);
-  xla::XlaOp zero = xla::Zero(input.builder(), shape.element_type());
-  xla::XlaOp xla_lambd =
-      XlaHelpers::ScalarBroadcast(lambda.to<double>(), shape, builder);
-  xla::XlaOp le_lambda_branch =
-      xla::Select(xla::Lt(input, Neg(xla_lambd)), input + xla_lambd, zero);
-  return xla::Select(xla::Le(input, xla_lambd), le_lambda_branch,
-                     input - xla_lambd);
+xla::XlaOp BuildSoftshrink(xla::XlaOp input, xla::XlaOp lambda) {
+  const xla::Shape& input_shape = XlaHelpers::ShapeOfXlaOp(input);
```
vanbasten23, line 132: Curious how I can convert a xla::XlaOp object to a scalar? I asked because the lambda for Softshrink needs to be nonnegative (https://pytorch.org/docs/stable/generated/torch.nn.Softshrink.html). How can I do the check here? Also, if I can manage to check lambda here, the check will be done before we actually execute the graph, right?

torch_xla/csrc/elementwise.cpp
wonjoolee95: .builder() is a method that belongs to xla::XlaOp so it should be okay, but is there a specific need to do that?

also nit: for zero XlaOp, we can just do:
```
xla::XlaOp zero = xla::Zero(input.builder(), shape.element_type());
```

vanbasten23: `.builder() is a method that belongs to xla::XlaOp so it should be okay, but is there a specific need to do that?`
There is no specific need. It's just both `input` and `lambda` can provide a builder and I am not sure if there is a requirement of which build we have to use.

wonjoolee95 (approved):
Thanks!

PR #4069, opened 2022-10-03
Move torch::einsum to torch::_einsum

janeyx99:
Will unblock https://github.com/pytorch/pytorch/pull/85908.


steventk-g:
It looks like this is breaking our einsum tests

janeyx99:
closing as this does not look necessary anymore—we’ve taken a different approach.

PR #4073, opened 2022-10-06
implementing XLASymIntNodeImpl::str

Krovatkin:


test/test_str.py
```diff
@@ -0,0 +1,6 @@
+import torch, torch_xla, torch_xla.core.xla_model as xm
```
JackCaoG, line 1: Let's move this to `test_dynamic_shape` that you added on another pr

PR #4074, opened 2022-10-06
Dynamic Shape Simple NN Model Test

miladm:
Testing dynamic shape functionality using a simple model. This PR is meant to identify the gaps in dynamic shape op support. We will keep track of the bugs we address as part of the investigation in this PR.

### DIFF
--- a/test/dynamic_shape_model_test.py
+++ b/test/dynamic_shape_model_test.py
@@ -1,0 +1,67 @@
+import torch, torch_xla
+import torch_xla.core.xla_model as xm
+import numpy
+
+dev = xm.xla_device()
+
+class Feedforward(torch.nn.Module):
+        def __init__(self, input_size, hidden_size):
+            super(Feedforward, self).__init__()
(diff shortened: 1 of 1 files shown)


test/dynamic_shape_model_test.py
```diff
@@ -0,0 +1,79 @@
+from sklearn.datasets import make_blobs
+import torch, torch_xla
+import torch_xla.core.xla_model as xm
+import numpy
+
+pd = torch._C._EnablePythonDispatcher()
```
vanbasten23, line 6: I wonder what this line does

Krovatkin: it enables us to run python implementations of `CompositeAutogradImplicit` ops.
`CompositeAutogradImplicit` means we don't have an explicit backward formula for an op instead an op is composed of a bunch of ops that do have backward formulas and combines this formulas is equivalent to differentiating the op explicitly.

Issue #4075, opened 2022-10-06
Simple NN model test with dynamic shape breaks `torch.nn.functional`


miladm:
Looks like [this code](https://github.com/pytorch/pytorch/blob/master/torch/nn/functional.py#L3081) breaks in pytorch when running the test in this [PR](https://github.com/pytorch/xla/pull/4074).

Error message:
```
Traceback (most recent call last):
  File "test.py", line 63, in <module>
    before_train = criterion(y_pred.squeeze(), y_test)
  File "/opt/conda/lib/python3.7/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/torch/nn/modules/loss.py", line 619, in forward
    return F.binary_cross_entropy(input, target, weight=self.weight, reduction=self.reduction)
  File "/opt/conda/lib/python3.7/site-packages/torch/nn/functional.py", line 3081, in binary_cross_entropy
    if target.size() != input.size():
RuntimeError: NYI
```

CC @Krovatkin

Krovatkin:
could you please paste the command and/or test you are running? @miladm

Krovatkin:
Interesting! We actually failed on a different unimplemented function than I expected. 

```
Thread 1 "python" hit Catchpoint 1 (exception thrown), __cxxabiv1::__cxa_throw (obj=0x55555b7b8160,
    tinfo=0x7fff4e324728 <typeinfo for c10::Error>, dest=0x7fffd54e0ba0 <c10::Error::~Error()>)
    at /opt/conda/conda-bld/gcc-compiler_1654084175708/work/build/gcc-final/x86_64-conda-linux-gnu/libstdc++-v3/libsupc++/eh_throw.cc:77
77      in /opt/conda/conda-bld/gcc-compiler_1654084175708/work/build/gcc-final/x86_64-conda-linux-gnu/libstdc++-v3/libsupc++/eh_throw.cc
(gdb) where
#0  __cxxabiv1::__cxa_throw (obj=0x55555b7b8160, tinfo=0x7fff4e324728 <typeinfo for c10::Error>,
    dest=0x7fffd54e0ba0 <c10::Error::~Error()>)
    at /opt/conda/conda-bld/gcc-compiler_1654084175708/work/build/gcc-final/x86_64-conda-linux-gnu/libstdc++-v3/libsupc++/eh_throw.cc:77
[...]

PR #4076, opened 2022-10-07
implement XLASymIntNodeImpl::wrap

Krovatkin:
This is the first step to fix https://github.com/pytorch/xla/issues/4075

### DIFF
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -5,6 +5,7 @@ import unittest
 
 dev = xm.xla_device()
 
+
 class TestDynamicShapes(unittest.TestCase):
 
   def test_wrap(self):
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -15,6 +15,7 @@ class TestDynamicShapes(unittest.TestCase):
     a3 = a2.shape[0] + 3  # tests wrap
     self.assertIsInstance(a3, torch.SymIntNode)
 
+
 if __name__ == '__main__':
   test = unittest.main()
   sys.exit(0 if test.result.wasSuccessful() else 1)
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -5,20 +5,14 @@ import unittest
 
 dev = xm.xla_device()
 
-
 class TestDynamicShapes(unittest.TestCase):
 
   def test_wrap(self):
     a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
     a2 = torch.nonzero(a1)
+    self.assertTrue(a2.shape[0] == 3)
     a3 = a2.shape[0] + 3  # tests wrap
     self.assertIsInstance(a3, torch.SymIntNode)
-
-  def test_wrap(self):
-    a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
-    a2 = torch.nonzero(a1)
-    self.assertTrue(a2.shape[0] == 3)
-
 
 if __name__ == '__main__':
   test = unittest.main()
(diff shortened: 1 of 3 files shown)


torch_xla/csrc/tensor.cpp
```diff
@@ -1840,6 +1840,11 @@ c10::SymIntNode XLASymIntNodeImpl::mul(const c10::SymIntNode& other) {
   return c10::make_intrusive<XLASymIntNodeImpl>(nmul);
 }
 
+c10::SymIntNode XLASymIntNodeImpl::wrap(int64_t num) {
```
vanbasten23, line 1843: What's the expected behavior for "wrap"? I couldn't find the documentation in the design doc.

test/test_dynamic_shapes.py
```diff
@@ -0,0 +1,20 @@
...
+
+dev = xm.xla_device()
+
+
+class TestDynamicShapes(unittest.TestCase):
+
+  def test_wrap(self):
+    a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
+    a2 = torch.nonzero(a1)
+    a3 = a2.shape[0] + 3  # tests wrap
```
vanbasten23, line 14: Can you explain why `a2.shape[0]+3` tests wrap?

torch_xla/csrc/tensor.cpp
Krovatkin: I think you also reviewed `SizeConstant` PR as well. We need `SizeConstants` or constants since we may have constants in size expressions. For example, `x.shape[0] * 2` or `x.shape[0] / 2`.
The easiest way to make it work is to use the other argument of "+" x.shape[0] to turn 2 into `SizeConstant(2)` since the other argument is already a proper `SymInt`. This is why we have `wrap`

test/test_dynamic_shapes.py
Krovatkin: what happens here is that a2.shape invokes XlaSymIntNodeImpl::wrap and that produces another SymInt(XlaSymIntNodeImpl(SizeConstant(2)) which we can that use freely in addition.

torch_xla/csrc/tensor.cpp
vanbasten23: I see. So `wrap` does a cast. Thanks for explaining.

test/test_dynamic_shapes.py
vanbasten23: I see. One question about how python's `a2.shape` maps to c++'s `XlaTensorImpl::sym_sizes`, you mentioned the magic happens at https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/python/init.cpp#L1409-L1413. That line reference (1409-1413) seem a bit off. Do you mean https://github.com/pytorch/pytorch/blob/e2a4dfa468330c0587849bea4896ff5fffb33010/torch/csrc/jit/python/init.cpp#L1367-L1370 ?

torch_xla/csrc/tensor.cpp
vanbasten23, line 1843: I wonder why this is a member function of the class `XLASymIntNodeImpl` as opposed to a standalone util function. This function doesn't seem to use any information from the class itself.

vanbasten23:
LGTM

miladm (approved):
Looks like this [line](https://github.com/pytorch/xla/blob/master/test/test_operations.py#L638) fails on GPU tests. Otherwise, LGTM. Thanks!

test/test_dynamic_shapes.py
```diff
@@ -11,15 +11,10 @@ class TestDynamicShapes(unittest.TestCase):
   def test_wrap(self):
     a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
     a2 = torch.nonzero(a1)
+    self.assertTrue(a2.shape[0] == 3)
```
miladm, line 14: @Krovatkin why is this line not set to test the upper bound?

miladm: nvm... I misread the line :)

miladm:
The PR is up to date. All tests passed this morning. Thus, merging the PR before tests complete.

PR #4077, opened 2022-10-07
[PJRT] Enable PJRT C API option

will-cromar:
[`PjRtCApiClient` in upstream TensorFlow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.h)

The new PjRt C API will be used to enable the new TPU runtime that we will support long-term. Note: The `PjRtCApiClient` currently requires a Google-internal build of libtpu. The `TPU_C_API` option will _not_ work with the public `libtpu-nightly` builds at the time of writing.

- Create a `PjRtCApiClient` when `PJRT_DEVICE=TPU_C_API`
  - Still return `"TPU"` from `pjrt.device_type()` because this should be interchangeable with `PjRtTpuClient`.
- Avoid calling `executable->GetHloModules()` except for SPMD-sharded executables because it's not yet supported in the C API. This also means we won't be able to test SPMD with the C API until `GetHloModules` is supported upstream.
- Use `xla::PjRtClient::HostBufferSemantics::kZeroCopy` with C API because it does not yet support `kImmutableUntilTransferCompletes`. 
- Avoid calling `AcquireExternalReference` to get an `OpaqueHandle`. [`AcquireExternalReference`](https://github.com/tensorflow/tensorflow/blob/3899ff4b6e6018a5333e2cf000b826a11411c2b3/tensorflow/compiler/xla/pjrt/pjrt_client.h#L762-L779) is for sharing memory with an external framework and is not necessary in this case. We only need a unique int to represent the underlying buffer (analogous to a "handle" in XRT) to use in [`RunPostOrder`](https://github.com/pytorch/xla/blob/e515d31d199215409277981f7474a6cce1df77b6/torch_xla/csrc/tensor.cpp#L1369-L1370). Use the buffer address directly instead because it's not trivial to get aliased `PjRtBuffer`s and PjRt doesn't have the same notion of a "handle" or a unique buffer ID. The C API does not support `AcquireExternalReference`, but we didn't actually need it anyway.
- Don't modify `TPU_LIBRARY_PATH` in `__init__.py`, since that needs to be set to the custom internal `libtpu` build for testing this PR. Instead, use `TPU_LOAD_LIBRARY=0` to prevent the TF TPU runtime from initializing the TPU when we're not using XRT. This expresses our intent (i.e. don't load libtpu yet) more clearly anyway.
- Patch https://github.com/tensorflow/tensorflow/commit/9a4502ab470177c259865b5b58598be690bdd86e until we update our TF pin past the fix.

Tested manually with ResNet50 on a v4-8. Performance was within ~2% of the `PjRtTpuClient` baseline.

CC @yeounoh @skye

### DIFF
--- a/tf_patches/patch.diff
+++ b/tf_patches/patch.diff
@@ -3,9 +3,9 @@ index 9fe020d5937..32774c2f3c0 100644
 --- a/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
 +++ b/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
 @@ -18,12 +18,7 @@ limitations under the License.
- 
+
  #include <stdint.h>
- 
+
 -// _Float16 always gets us the correct ABI type, so use that if available.
 -// AArch64 GCC defines __FLT16_MANT_DIG__ even when _Float16 is not available.
 -#if defined(__FLT16_MANT_DIG__) && \
@@ -32,3 +32,15 @@ index b70ea8af5df..b54a895e306 100644
    } else if (tool_name == "tool_names") {
      return GetAvailableToolNames(session_snapshot);
    } else {
+diff --git a/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.cc b/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.cc
+index e6fda37012b..5d10dc157e2 100644
+--- a/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.cc
++++ b/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.cc
+@@ -787,7 +787,6 @@ PjRtCApiExecutable::ExecuteSharded(
+
+   args.execute_device =
+       tensorflow::down_cast<PjRtCApiDevice*>(device)->c_device();
+-  args.execute_device->device = device;
+
+   RETURN_STATUS_IF_ERROR(pjrt_c_api()->PJRT_Executable_Execute(&args),
+                          pjrt_c_api());
--- a/tf_patches/patch.diff
+++ b/tf_patches/patch.diff
@@ -3,9 +3,9 @@ index 9fe020d5937..32774c2f3c0 100644
 --- a/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
 +++ b/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
 @@ -18,12 +18,7 @@ limitations under the License.
-
+ 
  #include <stdint.h>
-
+ 
 -// _Float16 always gets us the correct ABI type, so use that if available.
 -// AArch64 GCC defines __FLT16_MANT_DIG__ even when _Float16 is not available.
 -#if defined(__FLT16_MANT_DIG__) && \
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -298,8 +298,13 @@ PjRtComputationClient::ExecuteComputation(
     std::unique_ptr<xla::PjRtBuffer> buffer = std::move(result);
 
     std::shared_ptr<PjRtData> data = std::make_shared<PjRtData>(
-        // TODO(wcromar): do we need `logical_on_device_shape` here?
-        device, buffer->on_device_shape(), std::move(buffer));
+        device,
+        // TODO(wcromar): just use `logical_on_device_shape` when it's supported
+        // in C API
+        client_->runtime_type() == xla::PjRtRuntimeType::kTfrt
+            ? buffer->on_device_shape()
+            : buffer->logical_on_device_shape().ValueOrDie(),
+        std::move(buffer));
 
     datas.push_back(data);
   }
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -8,8 +8,9 @@
 #include "tensorflow/compiler/xla/layout_util.h"
 #include "tensorflow/compiler/xla/literal.h"
 #include "tensorflow/compiler/xla/pjrt/cpu_device.h"
-#include "tensorflow/compiler/xla/pjrt/pjrt_executable.h"
+#include "tensorflow/compiler/xla/pjrt/pjrt_c_api_client.h"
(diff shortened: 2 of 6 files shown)


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -278,8 +298,8 @@ PjRtComputationClient::ExecuteComputation(
     std::unique_ptr<xla::PjRtBuffer> buffer = std::move(result);
 
     std::shared_ptr<PjRtData> data = std::make_shared<PjRtData>(
-        device, buffer->logical_on_device_shape().ValueOrDie(),
-        std::move(buffer));
+        // TODO(wcromar): do we need `logical_on_device_shape` here?
+        device, buffer->on_device_shape(), std::move(buffer));
```
will-cromar, line 302: @JackCaoG Do you know if we actually need to use `logical_on_device_shape` here instead of `on_device_shape`? Here's the difference from the [doc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/xla/pjrt/pjrt_client.h#L751-L757):

```
  // Same as on_device_shape when the shape is static. When the shape is
  // dynamic, it gathers the metadata from the device and returns a static shape
  // representing the logical shape of the data. This approach is identical to
  // how tensorflow and xrt setup the output buffer in the graph.
  //
  // Since this method actually acquires locks and communicate with the device,
  // it does not have the const qualifier, similar to what ToLiteral does.
  ```
  
  To play it safe and avoid breaking SPMD, I left `logical_on_device_shape` in `ExecuteReplicated`. If we need `logical_on_device_shape`, I can file a feature request to PJRT to implement this in the C API.

Comment (JackCaoG) [reply third_party/xla_client/pjrt_computation_client.cc:302]: From the comments it seems like we need `logical_on_device_shape` for dynamic shape tensors. We should file a feature request. 

The more concerning part of this thing is when I look up the tfrt implementation of this function I found
```
Dynamic shapes are not implemented in the TFRT backend
```

I think we need to identify the gap between streamexecutor and tfrt regarding the dynamic shape support. @miladm

Comment (will-cromar) [reply third_party/xla_client/pjrt_computation_client.cc]: For this PR, I modified this line to only use `on_device_shape` when we're using TFRT. Otherwise, we'll continue to use `logical_on_device_shape`.

I also filed a feature request to implement `logical_on_device_shape` in the C API, but the underlying TFRT TPU implementation of this function might not support dynamic shapes.

File: tf_patches/patch.diff [hunk 2]
```diff
@@ -32,3 +32,15 @@ index b70ea8af5df..b54a895e306 100644
    } else if (tool_name == "tool_names") {
      return GetAvailableToolNames(session_snapshot);
    } else {
+diff --git a/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.cc b/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.cc
+index e6fda37012b..5d10dc157e2 100644
+--- a/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.cc
++++ b/tensorflow/compiler/xla/pjrt/pjrt_c_api_client.cc
```
JackCaoG, line 38: we should be able to remove this once we update tf again right?

will-cromar: Yeah, this should get removed next time we update the pin.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -220,17 +227,30 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
     std::unique_ptr<xla::PjRtLoadedExecutable> executable =
         ConsumeValue(client_->Compile(instance.computation, compile_options));
 
-    const auto& hlo_modules = ConsumeValue(executable->GetHloModules());
-    HloComputation* hlo_computation = hlo_modules[0]->entry_computation();
-    xla::ProgramShape program_shape =
-        xla::ProgramShape(hlo_computation->ToProto().program_shape());
+    if (instance.is_sharded) {
```
JackCaoG, line 230: I guess this means CAPI won't support our spmd case yet?

will-cromar: Right. Not until `GetHloModules` is implemented (which is WIP)

third_party/xla_client/pjrt_computation_client.h
```diff
@@ -126,13 +130,8 @@ class PjRtComputationClient : public ComputationClient {
              std::shared_ptr<PjRtBuffer> buffer)
         : Data(std::move(device), std::move(device_shape)), buffer(buffer) {}
 
-    void* get_handle() const {
```
JackCaoG, line 129: whyb is `get_handle` removed? Is it not being used in any place?

tf_patches/patch.diff
```diff
@@ -32,3 +32,15 @@ index b70ea8af5df..b54a895e306 100644
    } else if (tool_name == "tool_names") {
```
JackCaoG, line 32: General question, how would user config to use `TPU_CAPI` instead of `SE`?

third_party/xla_client/pjrt_computation_client.h
will-cromar: `get_handle` was only used by `GetOpaqueHandle` since PJRT doesn't actually have handles and it's not exposed by `Data`. I just rolled it into `GetOpaqueHandle` instead.

tf_patches/patch.diff
will-cromar: For now, we can specify `PJRT_DEVICE=TPU_C_API` instead of `PJRT_DEVICE=TPU` to use the `PjRtCApiClient`. We'll want to flip the default around once the C API is in public libtpu builds.

JackCaoG (approved):
You can update `GetOpaqueHandle ` in a separate pr after this one merged

third_party/xla_client/pjrt_computation_client.cc
miladm: Sounds like TFRT does not adopt DS yet; we should certainly work on landing it as part of PjRt adoption. How do we start this line of discussion? @will-cromar @JackCaoG

yeounoh: FYI Regarding SPMD, the output shape will always be static. So using `on_device_shape` won't break. It does sound like we need the support, though, for dynamic shape.

PR #4079, opened 2022-10-07
Add an equality node

Krovatkin:


torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -88,6 +88,25 @@ XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
...
+  XLA_CHECK(dim_node_0);
+  XLA_CHECK(dim_node_1);
+};
+
+int64_t SizeEq::getDynamicValue() const {
+  const torch::lazy::DimensionNode* dim_node_0 = DimCast(operand(0));
+  const torch::lazy::DimensionNode* dim_node_1 = DimCast(operand(1));
+  XLA_CHECK(dim_node_0);
+  XLA_CHECK(dim_node_1);
+  return dim_node_0->getDynamicValue() == dim_node_1->getDynamicValue() ? 1 : 0;
```
JackCaoG, line 105: would return `dim_node_0->getDynamicValue() == dim_node_1->getDynamicValue()` caused a weird cast issue?

JackCaoG:
Can you rebase and kick off CI again?

torch_xla/csrc/ops/dynamic_ir.h
```diff
@@ -53,6 +53,21 @@ class SizeNode : public XlaNode, public torch::lazy::DimensionNode {
   mutable int64_t runtime_size_;
 };
 
+class SizeEq : public XlaNode, public torch::lazy::DimensionNode {
+ public:
+  SizeEq(torch::lazy::Value a, torch::lazy::Value b);
+  int64_t getDynamicValue() const override;
+  int64_t getStaticValue() {
+    TORCH_CHECK(false, "Comparison operators should be using getDynamicValue");
```
vanbasten23, line 61: I guess the use case for the SizeEq node is that users does a `tensor1.shape[0] == tensor2.shape[0]` or `tensor1.shape[0] == tensor1.shape[1]` or something similar, right?

vanbasten23, line 61: Intuitively, I thought 2 nodes are equal when both static value and dynamic value are equal respectively. I wonder why we are only comparing dynamic values here.

torch_xla/csrc/tensor.cpp
```diff
@@ -1828,6 +1828,12 @@ bool XLATensor::ShouldSyncIrNode() {
   return this->data()->ir_value->op() != xla_device_data;
 }
 
+c10::SymIntNode XLASymIntNodeImpl::eq(const c10::SymIntNode& other) {
```
vanbasten23, line 1831: Do you need to add a test?

torch_xla/csrc/ops/dynamic_ir.h
miladm: That's because conditional operations work off of "true" dynamic tensor values. Look under conditional operations in the design doc for more detail.

miladm: any two tensor shapes can be compared.

torch_xla/csrc/ops/dynamic_ir.h
```diff
@@ -53,6 +53,21 @@ class SizeNode : public XlaNode, public torch::lazy::DimensionNode {
...
+  }
+  bool isSymbolic() const override { return true; }
+  std::string ToString() const override;
+  virtual XlaOpVector Lower(LoweringContext* loctx) const override {
+    // TODO: not sure we will ever need it?
+    TORCH_CHECK(false, "Lowering comparison nodes isn't supported yet!");
```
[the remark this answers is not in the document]
miladm: Yes, according to the design, we can leave this unimplemented. The reason is we materialize tensor results on conditional events.

miladm (commented):
Awesome. Took an initial pass. Reran the build.

miladm (approved):
Thanks @Krovatkin

PR #4081, opened 2022-10-09
Add ConvertXlaShapeToLazy helper function

wonjoolee95:
Add ConvertXlaShapeToLazy helper function that converts `xla::Shape` to `torch::lazy::Shape`. Also add unit tests for `lazy` related functions. 

PyTorch's `at::ScalarType` is defined at https://github.com/pytorch/pytorch/blob/041edeeecb75f3c110605d7311fa46abe1c62ea9/c10/core/ScalarType.h#L28. 

XLA's `xla::PrimitiveType` is defined at https://github.com/tensorflow/tensorflow/blob/b751126f998f5f5ef023f7817511a01c42efe99a/tensorflow/compiler/xla/xla_data.proto#L27. 

---
TODO
- Decide what to do with incompatible types
- ~Unit tests~

### DIFF
--- a/test/cpp/test_lazy.cpp
+++ b/test/cpp/test_lazy.cpp
@@ -11,32 +11,15 @@ namespace cpp_test {
 class LazyTest : public TorchXlaTest {};
 
 TEST_F(LazyTest, TestXlaShapeToLazy) {
-/*
-  Shape(PrimitiveType element_type, absl::Span<const int64_t> dimensions,
-        absl::Span<const bool> dynamic_dimensions,
-        std::vector<Shape> tuple_shapes)
-*/    
     int64_t dimensions[] = {1, 2};
     bool dynamic_dimensions[] = {false, false};
     absl::Span<const int64_t> xla_dimensions = absl::Span<const int64_t>(dimensions);
     absl::Span<const bool> xla_dynamic_dimensions = absl::Span<const bool>(dynamic_dimensions);
     std::vector<xla::Shape> xla_tuple_shapes = std::vector<xla::Shape>();
-
-    // std::cout << "WONJOO cpptest: dimensions.size()=" << dimensions.size() << std::endl;
-    std::cout << "WONJOO cpptest: xla_dimensions.size()=" << xla_dimensions.size() << std::endl;
-    for (const auto& elem : xla_dimensions) {
-      std::cout << "WONJOO cpptest: xla_dimensions[i]=" << elem << std::endl;;
-    } 
-
     xla::Shape xla_shape = xla::Shape(xla::PrimitiveType::F32, xla_dimensions, xla_dynamic_dimensions, xla_tuple_shapes);
 
     torch::lazy::Shape lazy_shape = XlaHelpers::ConvertXlaShapeToLazy(xla_shape);
-
-    std::cout << "WONJOO cpptest: xla::Shape=" << xla_shape << std::endl;
-    std::cout << "WONJOO cpptest: torch::lazy::Shape=" << lazy_shape << std::endl;
-
     std::vector<int64_t> lazy_dimensions = xla::util::ToVector<int64_t>(lazy_shape.sizes());
-    std::cout << "WONJOO cpptest: lazy_dimensions=" << lazy_dimensions << std::endl;
     const c10::optional<std::vector<bool>>& lazy_dynamic_dimensions = lazy_shape.is_symbolic();
     EXPECT_EQ(lazy_shape.scalar_type(), at::ScalarType::Float);
     EXPECT_EQ(lazy_shape.dim(), 2);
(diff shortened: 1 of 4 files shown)


wonjoolee95:
@alanwaketan, this should mostly be ready for review. Still need to decide what we should do with the incompatible data types.

torch_xla/csrc/helpers.cpp
```diff
@@ -619,4 +619,62 @@ xla::StatusOr<xla::XlaComputation> XlaHelpers::WrapXlaComputation(
   return builder.Build(orig_result);
 }
 
+torch::lazy::Shape XlaHelpers::ConvertXlaShapeToLazy(const xla::Shape& shape) {
```
alanwaketan, line 622: It looks like we already have similar utils in `TensorTypeFromXlaType`: https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor_util.cpp#L1077.

Also, there are actually multiple similar utils in that file: `TensorTypeToRawXlaType`, `XlaTypeFromTensorType`, and `MakeXlaPrimitiveType`. @JackCaoG Any thoughts on why? and should we clean that up?

wonjoolee95: Good catch, the existing `TensorTypeToRawXlaType` function is sufficient at least for this PR's use case. So I removed the existing `switch` case to this util function call.

torch_xla/csrc/helpers.cpp
```diff
@@ -619,4 +619,18 @@ xla::StatusOr<xla::XlaComputation> XlaHelpers::WrapXlaComputation(
...
+  at::ScalarType scalar_type = TensorTypeToRawXlaType(shape.element_type());
```
alanwaketan, line 623: I guess it should be `TensorTypeFromXlaType`?

torch_xla/csrc/helpers.cpp
```diff
@@ -619,4 +619,18 @@ xla::StatusOr<xla::XlaComputation> XlaHelpers::WrapXlaComputation(
...
+  c10::optional<std::vector<bool>> is_symbolic = c10::nullopt;
+  if (shape.is_dynamic()) {
+    std::vector<bool> xla_dynamic_dimensions =
+        xla::util::ToVector<bool>(shape.dynamic_dimensions());
+    is_symbolic = c10::make_optional(xla_dynamic_dimensions);
+  }
+
+  return torch::lazy::Shape(scalar_type,
+                            xla::util::ToVector<int64_t>(shape.dimensions()),
+                            is_symbolic);
```
alanwaketan, line 633: `std::move(is_symbolic)`?

test/cpp/test_lazy.cpp
```diff
@@ -0,0 +1,61 @@
...
+  std::vector<xla::Shape> xla_tuple_shapes = std::vector<xla::Shape>();
+  xla::Shape xla_shape = xla::Shape(xla::PrimitiveType::F32, xla_dimensions,
+                                    xla_dynamic_dimensions, xla_tuple_shapes);
+
+  torch::lazy::Shape lazy_shape = XlaHelpers::ConvertXlaShapeToLazy(xla_shape);
+  std::vector<int64_t> lazy_dimensions =
+      xla::util::ToVector<int64_t>(lazy_shape.sizes());
+  const c10::optional<std::vector<bool>>& lazy_dynamic_dimensions =
+      lazy_shape.is_symbolic();
+  EXPECT_EQ(lazy_shape.scalar_type(), at::ScalarType::Float);
```
alanwaketan, line 51: I guess we can have different sclar_type and dimensions in this test case to increase the coverage a bit.

torch_xla/csrc/helpers.cpp
wonjoolee95: Yep, my local dev is rebuilding tf so it was taking too long to rebuild and test lol

torch_xla/csrc/helpers.cpp
wonjoolee95: Done

test/cpp/test_lazy.cpp
wonjoolee95: Done

wonjoolee95:
@alanwaketan, this should be ready for another round of review. Thanks!

test/cpp/test_lazy.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include "torch/csrc/lazy/core/shape.h"
+#include "torch_xla/csrc/helpers.h"
+#include "torch_xla_test.h"
+
+namespace torch_xla {
+namespace cpp_test {
+
+class LazyTest : public TorchXlaTest {};
+
+TEST_F(LazyTest, TestXlaShapeToLazyWithF64) {
```
alanwaketan, line 13: Oops. Thanks for adding more test cases. I didn't mean to ask you to add more tests, just to change a few parameters in the previous test. Hope this didn't create too much overhead for you.

test/cpp/test_lazy.cpp
```diff
@@ -0,0 +1,122 @@
...
+  int64_t dimensions[] = {1};
+  bool dynamic_dimensions[] = {false, false};
```
alanwaketan, line 15: Is this supposed to be the same size as `dimensions`? Just took a look over the XLA implementation. It doesn't seem they have any checks to ensure that. Just a comment. lol.

test/cpp/test_lazy.cpp
```diff
@@ -0,0 +1,122 @@
...
+  EXPECT_EQ(lazy_shape.scalar_type(), at::ScalarType::Double);
+  EXPECT_EQ(lazy_dimensions, xla::util::ToVector<int64_t>(xla_dimensions));
+  EXPECT_EQ(lazy_dynamic_dimensions.has_value(), true);
+  EXPECT_EQ(lazy_dynamic_dimensions.value(),
+            std::vector<bool>(std::begin(dynamic_dimensions),
+                              std::end(dynamic_dimensions)));
+}
+
+}  // namespace cpp_test
+}  // namespace torch_xla
```
alanwaketan, line 122: Nit: add a newline at the end.

torch_xla/csrc/helpers.cpp
```diff
@@ -619,4 +619,18 @@ xla::StatusOr<xla::XlaComputation> XlaHelpers::WrapXlaComputation(
   return builder.Build(orig_result);
 }
 
+torch::lazy::Shape XlaHelpers::ConvertXlaShapeToLazy(const xla::Shape& shape) {
+  at::ScalarType scalar_type = TensorTypeFromXlaType(shape.element_type());
```
alanwaketan, line 623: It looks like `TensorTypeFromXlaType` could fail. Should we have a test case for that?

test/cpp/test_lazy.cpp
wonjoolee95: Good point, yeah. It seems like they don't have any checks but it should be the same size.

torch_xla/csrc/helpers.cpp
wonjoolee95: Good point, that was also on my mind but I thought that kind of testing should be unit-tested on that `TensorTypeFromXlaType` instead of our helper method. But looking back, it seems like we don't have any unit tests on that `TensorTypeFromXlaType` helper method. I'll add a test for the failure case.

wonjoolee95:
Thanks! Addressed comments. Will wait for CI to be green again and then merge.

test/cpp/test_lazy.cpp
wonjoolee95: Updated the tests to have same sizes.

PR #4085, opened 2022-10-10
[DDP] Add a test case to test a larger model

alanwaketan:
Summary:
This commit adds a test case to test a larger model that can trigger multiple all_reduces instead of one.

Test Plan:
XRT:
MASTER_ADDR=localhost MASTER_PORT=6000 python test/test_ddp.py TestXrtDistributedDataParallel.test_ddp_correctness_large_net PJRT:
PJRT_DEVICE=TPU python test/pjrt/test_ddp.py TestPjRtDistributedDataParallel.test_ddp_correctness_large_net

### DIFF
--- a/test/test_ddp.py
+++ b/test/test_ddp.py
@@ -8,6 +8,7 @@ import distributed_util as util
 
 FLAGS = args_parse.parse_common_options()
 
+
 class TestXrtDistributedDataParallel(parameterized.TestCase):
 
   @staticmethod
--- a/test/test_ddp.py
+++ b/test/test_ddp.py
@@ -1,3 +1,4 @@
+from absl import flags
 from absl.testing import absltest, parameterized
 import sys
 import torch_xla.core.xla_model as xm
@@ -5,11 +6,13 @@ import torch_xla.distributed.xla_multiprocessing as xmp
 
 import distributed_util as util
 
+FLAGS = flags.FLAGS
+
 
 class TestXrtDistributedDataParallel(parameterized.TestCase):
 
   @staticmethod
-  def _ddp_correctness(rank):
+  def _ddp_correctness(rank, use_large_net: bool, debug: bool):
     # We cannot run this guard before XMP,
     # see API_GUIDE.md#running-on-multiple-xla-devices-with-multi-processing.
     device = xm.xla_device()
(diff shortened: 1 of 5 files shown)


test/test_ddp.py
```diff
@@ -19,10 +22,13 @@ def _ddp_correctness(rank):
...
       return
-    util.ddp_correctness(None)
+    util.ddp_correctness(None, use_large_net=use_large_net, debug=debug)
 
   def test_ddp_correctness(self):
-    xmp.spawn(self._ddp_correctness, args=())
+    xmp.spawn(self._ddp_correctness, args=(False, FLAGS.debug))
+
+  def test_ddp_correctness_large_net(self):
+    xmp.spawn(self._ddp_correctness, args=(True, FLAGS.debug))
```
will-cromar, line 31: I'm a little concerned that `FLAGS.debug` is defined and used in `distributed_util.py`, but it's referenced here. But the command line args don't seem to get propagated to child processes with `spawn`, so you can't reference `FLAGS.debug` there. Is it possible to pass the whole `FLAGS` object to the child processes instead? That's what we do in `test_train_mnist_mp.py` for example.

alanwaketan: Good point. Let me fix that.

alanwaketan: It looks like `absl.flags` can't be pickled... Let me try using `args_parse`.

will-cromar (approved):
LGTM. Thanks!

alanwaketan:
Thanks Will for approving it. I will fix all the CI issues before merging.

torch_xla/distributed/xla_multiprocessing.py
```diff
@@ -32,6 +32,10 @@ def _is_xla_config():
   return False
 
 
+# TODO: Some usages of this function are to caculate the number of hosts/worlds,
```
alanwaketan, line 35: @will-cromar I made a few changes here to fix the GPU CI failures. Can you take a look at these new changes?

will-cromar: Going by the [torchrun documentation](https://pytorch.org/docs/stable/elastic/run.html#environment-variables), doesn't world size mean the number of processes and local world size mean the number of processes on the current host?

torch_xla/distributed/xla_multiprocessing.py
```diff
@@ -387,17 +393,27 @@ def spawn(fn,
...
-    return torch.multiprocessing.start_processes(
+    result = torch.multiprocessing.start_processes(
         _mp_start_fn,
         args=(pf_cfg, fn, args),
         nprocs=pf_cfg.num_devices,
         join=join,
         daemon=daemon,
         start_method=start_method)
 
+  # For GPU, xenv.WORKERS are set in the launcher and then get carried to the children.
```
will-cromar, line 408: :grimacing: this sounds like a nightmare to debug. Thanks for tracking it down. I'll make sure we don't make the same sort of mistake in PJRT

torch_xla/distributed/xla_multiprocessing.py
alanwaketan: Does that contradict to what I say? I can improve the TODO to reduce the confusion.

torch_xla/distributed/xla_multiprocessing.py
alanwaketan: yea, it is...

torch_xla/distributed/xla_multiprocessing.py
alanwaketan: Oh yea, I guess I shouldn't use `host/world` here.

PR #4090, opened 2022-10-12
Make sure XrtData is not empty before accessing handle

JackCaoG:
Before the graph execution, we might create empty data_place_holder in 
https://github.com/pytorch/xla/blob/79c433a066f3757a39693869c19f22d2366ace7f/torch_xla/csrc/tensor.cpp#L1416-L1418

In this case `XrtData` is empty and will have empty handlePtr. We want to guard this to avoid weird segfault.

### DIFF
--- a/third_party/xla_client/pjrt_computation_client.h
+++ b/third_party/xla_client/pjrt_computation_client.h
@@ -127,7 +127,6 @@ class PjRtComputationClient : public ComputationClient {
         : Data(std::move(device), std::move(device_shape)), buffer(buffer) {}
 
     void* get_handle() const {
-      XLA_CHECK(HasValue());
       return buffer->AcquireExternalReference()
           .ValueOrDie()
           ->OpaqueDeviceMemoryDataPointer();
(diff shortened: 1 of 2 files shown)


will-cromar:
Do we need a similar change in PJRT? I can update this in #4077

https://github.com/pytorch/xla/blob/79c433a066f3757a39693869c19f22d2366ace7f/third_party/xla_client/pjrt_computation_client.h#L129-L136

JackCaoG:
Let me quickly update this in this pr, I was working on a separate change requires this change lol.

JackCaoG:
@will-cromar I saw you actually removed `get_handle`, I guess I will just remove my PJRT change and let you handle it

alanwaketan:
So we still want to crash the program but differently?

alanwaketan (approved):

JackCaoG:
for the most part we should never hit this corner case. We had a lock to prevent user access palceholder Data. I hit this because I added a pybind to directly query handle value, this is for dynamo.

PR #4091, opened 2022-10-12
Introduce virtual device

steventk-g:
Changes in this PR 

* Expose virtual device via a flag `XLA_USE_SPMD`
* Use virtual device to conditionally delay the data transfer of a tensor. This is accomplished by setting the device on the backend based on the flag, so the PJRT computation client can check for the virtual device before transferring data (NOTE: We need to use the device on the backend rather than checking for the flag directly, so that we still have a way to transfer sharded data later on).
* When the flag is enabled, transfer sharded data _without_ redownloading from an xla device. This is done in `_xla_mark_sharding`. The re-downloading path is preserved so that `XLA_USE_SPMD=0` still works as well.
* When the flag is enabled, ensure that the user gets `xla:0` from `xm.xla_device()`. At this point, users should expect all tensors to be treated as if they are on the virtual device when SPMD is enabled.


yeounoh:
Let's make sure that we cover the explict sharded cases, where we want to avoid the initial unpartitioned data transfer. We will have to double-check, but `Modify XLATensor::Compile to begin data transfer on implicitly sharded tensors.` this may not be needed.

steventk-g:
Notes after chat with Yeounoh:
* We need to locate the place where data transfer is initiated to backend device. This is probably in upstream code. This is where we can check the device type and potentially skip the data transfer.
* We need to determine how to check the device type of an `at::Tensor` or `XLATensor`. The XlaDeviceType of tensors to shard will be "SPMD", and the device type will be XLA (like physical XLA devices: TPU, CPU, GPU).
* Explicitly sharded tensors on an SPMD device will be transferred to the backend device by a call to `CreateTensorData` in `_xla_mark_sharding`. 
* We need to decide when to transfer data for implicitly sharded tensors, if not in `XLATensor::Compile`

JackCaoG:
`We need to locate the place where data transfer is initiated to backend device` add a log to https://github.com/pytorch/xla/blob/4a267b43712c1c726b59a32bd7b74fa108033df1/third_party/xla_client/pjrt_computation_client.cc#L107

This is the only entry for the transfer data to device.

steventk-g:
Remaining implementation details before I can start testing:
* Determine how to filter tensors and devices in CreateTensorsData methods. We want all devices passed into the sharded method to be SPMD, and we want to stop data transfer to backend devices when a tensor with an SPMD device is passed into the non-sharded method. Can we simply remove the non-SPMD tensors in the first case, and remove the SPMD tensors in the second case?
* Figure out what to return from TensorToXlaData when we don't transfer data to a real backend.

steventk-g:
Next steps:
* It looks like there are a few places where we initiate data transfer to device, by calling `CreateTensorsData` or `TensorToXlaData`. 
* The current plan is to guard calls to `CreateTensorsData` so that we only transfer data to devices once sharding is specified or after tracing begins (that is, in the case when SPMD is enabled). To do that we can guard the cases that call the `CreateTensorData` without a sharding spec argument. When SPMD is enabled, we will instead wrap `at::Tensor`s as `XLATensors`, we will leave the handles to backend data as null, and we can make sure that these handles are not accessed before the data is transferred. 
* We can ignore calls to `TensorToXlaData` as they occur during tracing/compilation, after sharding would have been set.

steventk-g:
Update: `TensorToXlaData` gets called when we run `tensor.to(device)`, so we need to handle that path. Right now, I'm looking into how we can create a `DataPtr` that wraps the CPU data so we can return that as a `BackendDataPtr` from this method. However, I'm running into some complications creating the `DataPtr`.

test/test_virtual_device.py
```diff
@@ -0,0 +1,84 @@
...
+    partition_spec = (0, 1)
+    t1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                      dtype=torch.float,
+                      device=xm.xla_device())
+    t2 = torch.tensor([[8, 7, 6, 5, 4, 3, 2, 1]],
+                      dtype=torch.float,
+                      device=xm.xla_device())
+    xst1 = xs.mark_sharding(t1, self._get_mesh((1, self.n_devices)),
+                            partition_spec)
+    t3 = xst1 + t2
```
yeounoh, line 63: Can we do `t3=t1+t2` instead? We may make the `mark_sharding` API in-place in the future instead of returning `xst1`. Also, we should compare the result for correctness in value -- e.g., comparing with `[9, 9, 9, 9, 9,9 ,9 9]` and with or without the virtual device optimzation in this case. 

This applies to the other test cases here.  As a side note, we could place the entire test class in test_xla_sharding.py and reuse some of the test constructs, so we can group all the spmd related unit tests in one file?

jonb377 (commented):
Nice work Steven!

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -124,30 +124,35 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
...
-    PjRtDevice* pjrt_device = StringToPjRtDevice(tensor.device);
-
-    auto literal = std::make_shared<xla::Literal>(tensor.shape);
-    tensor.populate_fn(tensor, literal->untyped_data(), literal->size_bytes());
-    std::vector<int64_t> byte_strides(literal->shape().dimensions_size());
-    XLA_CHECK_OK(
-        ShapeUtil::ByteStrides(literal->shape(), absl::MakeSpan(byte_strides)));
-    total_size += literal->size_bytes();
+    ComputationClient::DataPtr data;
+    if (tensor.device == "SPMD:0") {
```
yeounoh, line 128: Do we need the placeholder data for the current flag interface as well? In any case, let's clarify + comment some more about this route.

yeounoh: Not sure if this is straightforward to do, but it could be cleaner if we can do the check and call or not call the TransferToServer --then we don't need to change the runtime client APIs.

torch_xla/core/xla_model.py
```diff
@@ -238,6 +238,11 @@ def xla_device(n=None, devkind=None):
   Returns:
     A `torch.device` with the requested instance.
   """
+  if xu.check_env_flag('XLA_USE_SPMD'):
```
yeounoh, line 241: Let's also add some comment here, explaining why this override is needed.

torch_xla/core/xla_model.py
```diff
@@ -137,7 +137,7 @@ def get_xla_supported_devices(devkind=None, max_devices=None):
     The list of device strings.
   """
   xla_devices = _DEVICES.value
-  devkind = [devkind] if devkind else ['TPU', 'GPU', 'CPU']
+  devkind = [devkind] if devkind else ['TPU', 'GPU', 'CPU', 'SPMD']
```
yeounoh, line 140: For now, add a TODO comment to clarify that the flag interface doesn't need/use this -- if that's the case.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1392,19 +1392,28 @@ void InitXlaModuleBindings(py::module m) {
...
     // confusion/error.
     XLA_CHECK(xtensor->sharding_spec() == nullptr)
         << "Existing annotation must be cleared first.";
 
+    at::Tensor cpu_tensor;
+    if (xla::sys_util::GetEnvBool("XLA_USE_SPMD", false)) {
+      // When virtual device is enabled, we expect tensor data to live on the
+      // CPU when it's passed into mark_sharding.
+      XLA_CHECK(xtensor->CurrentTensorData().has_value())
+          << "Cannot shard tensor. Data missing from CPU.";
```
yeounoh, line 1408: A bit more details, "When virtual device is enabled for SPMD, we defer the initial data transfer to the device and retain the original data on the host, until the sharded data transfer."

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1392,19 +1392,28 @@ void InitXlaModuleBindings(py::module m) {
...
+      cpu_tensor = xtensor->CurrentTensorData().value();
+    } else {
```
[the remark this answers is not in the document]
yeounoh: can we also check if the host tensor data is not present, or the xla_data is present instead?

torch_xla/utils/utils.py
```diff
@@ -313,3 +313,7 @@ def __exit__(self, type, value, traceback):
 
   def set_error(self, error):
     self._error = error
+
+
+def check_env_flag(name, default=''):
```
yeounoh, line 318: +1

yeounoh (changes requested):
I added some comments, looking good -- great work @steventk-g  :)

torch_xla/core/xla_model.py
```diff
@@ -238,6 +238,11 @@ def xla_device(n=None, devkind=None):
   Returns:
     A `torch.device` with the requested instance.
   """
+  if xu.check_env_flag('XLA_USE_SPMD'):
+    device = 'xla:0'
+    torch_xla._XLAC._xla_set_default_device(device)
```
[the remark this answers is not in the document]
steventk-g: Yeah - the idea is that to the user the device will be set to `xla:0` and that in this case, it represents the virtual device. It's a bit complicated because `xla:0` also refers to the physical device, e.g. when defining a mesh. Eventually, we probably want to either (1) treat spmd as it's own kind of device, or (2) explicitly add it as an xla device. Both of these would require a lot of code to change, possibly upstream, which is beyond the scope of this POC.

torch_xla/core/xla_model.py
steventk-g: Done

test/test_virtual_device.py
steventk-g: Done

torch_xla/utils/utils.py
```diff
@@ -313,3 +313,7 @@ def __exit__(self, type, value, traceback):
...
+  return os.getenv(name, default).upper() in ['ON', '1', 'YES', 'TRUE', 'Y']
```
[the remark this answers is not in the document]
steventk-g: Done

torch_xla/core/xla_model.py
steventk-g: I just removed this for now - since we may not even go this route.

torch_xla/csrc/init_python_bindings.cpp
steventk-g: Added

third_party/xla_client/pjrt_computation_client.cc
steventk-g: Done - added a comment and moved this up to tensor_util.cpp. We still need to use the placeholder data, even with the flag, but we can set the "SPMD:0" device internally based on the flag so that it's 100% transparent to the user.

test/test_xla_sharding.py
```diff
@@ -103,6 +103,27 @@ def test_deep_copy(self):
...
+
+  def test_mark_sharding(self):
+    partition_spec = (0, 1)
+    xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                       dtype=torch.float,
+                       device=xm.xla_device())
+    xs.mark_sharding(xt1, self._get_mesh((1, self.n_devices)), partition_spec)
+    self.assertTrue(
+        torch.allclose(
+            xt1 + 0,
```
yeounoh, line 121: Hey @steventk-g could we actually replace this with t2 = t1 + 1 and assert on t2?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -459,7 +459,8 @@ at::Tensor XLANativeFunctions::_copy_from(const at::Tensor& self,
   auto self_tensor = bridge::TryGetXlaTensor(self);
   if (!self_tensor) {
     static bool sync_update =
-        xla::sys_util::GetEnvBool("XLA_TENSOR_UPDATE_SYNC", true);
+        xla::sys_util::GetEnvBool("XLA_TENSOR_UPDATE_SYNC", true) &&
```
yeounoh, line 462: It would be helpful, actually important note here why XLA_USE_SPMD condition is added, what it does.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1392,19 +1392,32 @@ void InitXlaModuleBindings(py::module m) {
...
 
+    at::Tensor cpu_tensor;
+    if (xla::sys_util::GetEnvBool("XLA_USE_SPMD", false) &&
+        xtensor->CurrentTensorData().has_value()) {
+      // When virtual device is enabled for SPMD, we defer the initial data
+      // transfer to the device and retain the original data on the host, until
+      // the sharded data transfer.
+      cpu_tensor = xtensor->CurrentTensorData().value();
+    } else {
+      // If the at::Tensor data is not present, we need to re-download the
```
yeounoh, line 1411: A note to ourselves, in case XLA_USE_SPMD, we would expect that at::Tensor data to be present.

torch_xla/csrc/tensor_util.cpp
```diff
@@ -653,6 +653,16 @@ torch::lazy::BackendDataPtr TensorToXlaData(
     const at::Tensor& tensor, const xla::Shape& shape,
     const torch::lazy::BackendDevice& device) {
   XLA_TIMED("TensorToData");
+  if (device.toString() == "SPMD:0") {
```
yeounoh, line 656: We should check if xla_data is present and return it instead of the placeholder. This can be called after the sharding too.

test/test_xla_sharding.py
yeounoh: TODO: adding more unit tests, except the one with a known issue.

torch_xla/csrc/tensor_util.cpp
steventk-g: If the XLA data is present, the device would be something other than SPMD:0. The "SPMD:0" device is only used in the codepath where we update a tensor from an `at::Tensor`, so it should always be _before_ the data is sharded

torch_xla/csrc/aten_xla_type.cpp
steventk-g: Done

torch_xla/csrc/init_python_bindings.cpp
steventk-g: Done

test/test_xla_sharding.py
steventk-g: Added some tests here
https://github.com/pytorch/xla/pull/4214
https://github.com/pytorch/xla/pull/4215

Opened an issue
https://github.com/pytorch/xla/issues/4220

torch_xla/csrc/tensor_util.cpp
```diff
@@ -653,6 +653,16 @@ torch::lazy::BackendDataPtr TensorToXlaData(
...
+    // When SPMD is enabled, we want to delay the data transfer for XLA
+    // tensors until the data is sharded. So, we skip the data transfer
+    // here and simply return a placeholder for the backend data ptr.
+    // Data will only be transferred via CreateTensorsData, when users
+    // call the mark_sharding API.
+    return WrapXlaData(
+        xla::ComputationClient::Get()->CreateDataPlaceholder("SPMD:0", shape));
```
yeounoh, line 663: nit. Use `device` instead of "SPMD:0" and also check for device `type()` is `XlaDeviceType::SPMD`.

yeounoh (approved):
LGTM, thank you @steventk-g 👍

torch_xla/csrc/tensor_util.cpp
steventk-g: Done

PR #4098, opened 2022-10-14
Scope test

JackCaoG:
Upstream https://github.com/pytorch/pytorch/pull/87005


alanwaketan (commented):
The test case itself looks good to me.

However, any chances we can have this test case in the upstream? It seems weird that we have the implementation and test case in different places.

alanwaketan (approved):

JackCaoG:
> The test case itself looks good to me.
> 
> However, any chances we can have this test case in the upstream? It seems weird that we have the implementation and test case in different places.

I didn't find where python lazy test is. I also don't know if upstream has equivalant of `xp.trace`.

alanwaketan:
LTC has its python test here: https://github.com/pytorch/pytorch/blob/master/test/lazy/test_reuse_ir.py#L1. I don't think it has profiler though...

JackCaoG:
alternative solution is to write a CPP test for LTC

alanwaketan:
If it's too hard, then I will say just forgo it.

PR #4100, opened 2022-10-14
Add pin to upstream pytorch 1.13

vanbasten23:


.torch_pin
```diff
@@ -0,0 +1,2 @@
+release/1.13
+
```
JackCaoG, line 2: remove extra line

vanbasten23: done

JackCaoG (approved):
Do we expect CI to fail without your other pr? If that's the case you can force merge this one.

vanbasten23:
> Do we expect CI to fail without your other pr? If that's the case you can force merge this one.

I don't expect the CI to fail.
Thanks for the quick review!

PR #4101, opened 2022-10-17
TF/libtpu Pin Update 10/17/2022

steventk-g:


JackCaoG:
Thanks Steven, let do a couple checks here
1. open a pytorch pr similar to https://github.com/pytorch/pytorch/pull/85948 which will point their CI to this pr. This will make sure pytorch CI can build with our new pin
2. Build this pr on TPUVM and run the Resnet50 test then compare it with the current nightly. This will make sure we don't see any speed regression.

steventk-g:
1. I've created https://github.com/pytorch/pytorch/pull/87636. It looks like one check is failing, but it also failed on https://github.com/pytorch/pytorch/pull/85948 and it looks like it's because the commit can't be accessed from that platform for some reason.
2. Here are the stats for the build with the pin updated, I'll try the nightly build next
```
Epoch 1 train begin 21:24:56
```
```
Epoch 18 test end 23:12:35, Accuracy=100.00
Max Accuracy: 100.00%
```

steventk-g:
```
Epoch 1 train begin 15:01:46
```
```
Epoch 18 test end 16:48:43, Accuracy=100.00
Max Accuracy: 100.00%
```

2hr 12min vs 1hr 47min. 

Looks like a perf regression but is this significant?

steventk-g:
About 1hr 48min both times. Looks like no regression

JackCaoG:
Ci failure should go away after rebase, we need to follow up internally regarding 0de6ecda97e261528b51709c11a4e7e22a39ca33 . Worse case we need to make that threadlocal thing to be conditional so we can turn it off on our end. Carrying a patch will cause longer term pain.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -71,9 +69,11 @@ PjRtComputationClient::PjRtComputationClient() {
     bool async = sys_util::GetEnvBool(env::kEnvPjrtAsyncGpuClient, true);
     /* TODO(jonbolin): Set allowed_devices based on local ordinal */
     auto allowed_devices = std::make_optional<std::set<int>>(std::set{0});
-    client_ = xla::GetGpuClient(/*asynchronous=*/async, GpuAllocatorConfig{},
-                           /*distributed_client=*/nullptr, /*node_id=*/0,
-                           allowed_devices=allowed_devices).ValueOrDie();
+    client_ = xla::GetStreamExecutorGpuClient(
```
JackCaoG, line 72: @will-cromar I guess this is fine since TFRT does not support GPU yet?

will-cromar: Yes, that's correct

JackCaoG (approved):
LGTM, waiting for Will to confirm the GPU line. Let's follow up on the random seed internally.

JackCaoG:
@steventk-g Thanks for driving this, let's rebase and then merge this one.

PR #4105, opened 2022-10-19
Set default xla_gpu_force_compilation_parallelism

ymwangg:
This PR sets `XLA_FLAGS: --xla_gpu_force_compilation_parallelism=8` by default. We choose this number because most GPU instances have `num_cpus >= 8*num_gpus`  and the marginal benefit of using more than 8 threads is small.

Below is the compilation time benchmark of a few models tested on p3dn.24xlarge instance.
| Model | 1 thread | 8 threads | 16 threads|
| --- | ----------- | ----------| -----------|
| bert-base-uncased | 67s | 31s| 27s|
| bert-large-uncased | 217s | 80s| 72s|
| resnet-50 | 17s | 11s | 9s|


JackCaoG (approved):
Thanks!

PR #4109, opened 2022-10-20
Add an api to clear counter and metrics

JackCaoG:
Upstream version at https://github.com/pytorch/pytorch/blob/d9ff56ccc03cd5c2f95112ef3ec2316557f5699c/torch/csrc/lazy/core/metrics.h#L92 actually reset the counter instead of clearing it. 

However I found it is more intuitive to clear the counters and there is no way to resert metrics(metrics can not be empty). I think I will submit a pr to update upstream later.


alanwaketan (approved):
LGTM.

JackCaoG:
I think I will have better luck to put this test in a separate file. It is a bad idea to clear all metrics for the `test_operations.py`

Issue #4093, opened 2022-10-13
Add 5D tiled test for ShardingUtil::ShardTensor


steventk-g:
Adds to tests introduced by https://github.com/pytorch/xla/pull/3684

Add the test to `XLAShardingTest.ShardTensor` in `test/cpp/test_xla_sharding.cpp`

steventk-g:
This task is blocked because there is no `Array5D` class in XLA. It only goes up to 4D right now.

yeounoh:
Thanks @steventk-g , I think we can use the generic `xla::Array<T>` to implement the hypercube mesh. Let me take a look.

PR #4110, opened 2022-10-20
Add 5d sharding test case

steventk-g:
https://github.com/pytorch/xla/issues/4093


steventk-g:
Current blocker is that there is not Array5D class in xla right now, only up to 4D is supported.

JackCaoG:
hmm, all test seemed to pass?

steventk-g:
> hmm, all test seemed to pass?

The code I've added doesn't exercise the 5D sharding codepath, it's still using `Array4D`

PR #3815, opened 2022-08-01
Codegen for bitwise and, or, xor, and not 

steventk-g:
## PyTorch PR

https://github.com/pytorch/pytorch/pull/82617

## LazyIr.h
```
class BitwiseAndTensor : public XlaNode {
 public:
  static torch::lazy::OpKind ClassOpKind() {
    return torch::lazy::OpKind(at::aten::bitwise_and);
  }

  BitwiseAndTensor(const torch::lazy::Value& self, const torch::lazy::Value& other, std::vector<torch::lazy::Shape>&& shapes)
      : XlaNode(torch::lazy::OpKind(at::aten::bitwise_and),
              {self, other}, std::move(shapes),
              [&]() { return BitwiseAndTensorOutputShape(self, other); },
              /* num_outputs */ 1,
              torch::lazy::MHash())
  {
    
  }

  std::string ToString() const override {
    std::stringstream ss;
    ss << XlaNode::ToString();
    
    return ss.str();
  }

  

  bool CanBeReused(const torch::lazy::Value& self, const torch::lazy::Value& other) const {
    return false;
    }

  torch_xla::XlaOpVector Lower(LoweringContext* loctx) const override;

  
  

};

class BitwiseNot : public XlaNode {
 public:
  static torch::lazy::OpKind ClassOpKind() {
    return torch::lazy::OpKind(at::aten::bitwise_not);
  }

  BitwiseNot(const torch::lazy::Value& self, std::vector<torch::lazy::Shape>&& shapes)
      : XlaNode(torch::lazy::OpKind(at::aten::bitwise_not),
              {self}, std::move(shapes),
              [&]() { return BitwiseNotOutputShape(self); },
              /* num_outputs */ 1,
              torch::lazy::MHash())
  {
    
  }

  std::string ToString() const override {
    std::stringstream ss;
    ss << XlaNode::ToString();
    
    return ss.str();
  }

  

  bool CanBeReused(const torch::lazy::Value& self) const {
    return false;
    }

  torch_xla::XlaOpVector Lower(LoweringContext* loctx) const override;

  
  

};

class BitwiseOrTensor : public XlaNode {
 public:
  static torch::lazy::OpKind ClassOpKind() {
    return torch::lazy::OpKind(at::aten::bitwise_or);
  }

  BitwiseOrTensor(const torch::lazy::Value& self, const torch::lazy::Value& other, std::vector<torch::lazy::Shape>&& shapes)
[...]

PR #4111, opened 2022-10-20
Update codegen to use xla::Shape

wonjoolee95:
Update codegen to use xla::Shape

---
This PR updates codegen to use `xla::Shape` instead of `torch::lazy::Shape` by overriding the upstream `GenLazyNativeFuncDefinition` and `GenLazyIr` generators. It overrides `GenLazyNativeFuncDefinition.shape_inference()` and `GenLazyNativeFuncDefinition.build_ir_node()` functions to remove `torch::lazy::Shape` related shape inference and constructor codegen. It also overrides `GenXlaLazyIR.gen()` to remove `std::vector<torch::lazy::Shape>` in `Node` constructor`.  

This depends on an open upstream PyTorch PR https://github.com/pytorch/pytorch/pull/87823 to make `GenLazyNativeFuncDefinition` customizable. 

---
New `LazyIr.h` example:
```
class Frac : public XlaNode {
public:
  return torch::lazy::OpKind(at::aten::frac);
}

Frac(const torch::lazy::Value& self)
    : XlaNode(torch::lazy::OpKind(at::aten::frac),
              {self},
              [&]() { return FracOutputShape(self); },
              /* num_outputs */ 1,
              torch::lazy::MHash())
{
  
}

  std::stringstream ss;
  ss << XlaNode::ToString();
  
  return ss.str();
}



bool CanBeReused(const torch::lazy::Value& self) const {
    return false;
    }

};
```
Changes
- Removed `std::vector<torch::lazy::Shape>` argument to the constructor
- Updated to call the `XlaNode` constructor without the `torch::lazy::Shape`

Old `LazyIr.h` example: https://github.com/pytorch/xla/pull/3930

---
New `XlaNativeFunctions.cpp` example:
```
    at::Tensor XLANativeFunctions::frac(const at::Tensor & self) {
        
        XLA_FN_COUNTER("xla::");
        auto common_device = torch_xla::bridge::GetXlaDevice(self);
        
        torch::lazy::NodePtr node = torch::lazy::ReuseNode<Frac>(lazy_self->GetIrValue());
      if (!node) {
          
          node = torch::lazy::MakeNode<Frac>(lazy_self->GetIrValue());
          CacheNode(node);
      }
      
        return result;
    }
```
Changes:
- Removed `torch::lazy::Shape` related creation/assertion logic
- Removed `symbolicShapeEnabled` logic (need to double check if this is okay)

Old `XlaNativeFunctions.cpp` example: https://github.com/pytorch/xla/pull/3930


wonjoolee95:
@alanwaketan, this should be ready for a review. Thanks!

JackCaoG:
Let's also remove https://github.com/pytorch/xla/blob/647a3d5a0996994e2efb1c4452882b4876a949a1/torch_xla/csrc/tensor.cpp#L466-L468

I added it a while back to use `lazy shape` to construct `logical_element _type`

scripts/gen_lazy_tensor.py
```diff
@@ -61,11 +61,113 @@ def node_base_ctor_call(self, schema: LazyIrSchema) -> str:
     scalar_hashes = ", ".join([f"{a.name}" for a in scalar_args])
 
     return f"""{self.node_base}(torch::lazy::OpKind({aten_symbol(schema)}),
-              {{{base_ctor_value_args}}}, std::move(shapes),
+              {{{base_ctor_value_args}}},
               [&]() {{ return {schema.node_name}OutputShape({shape_fn_inputs}); }},
               /* num_outputs */ {len(schema.returns)},
               torch::lazy::MHash({scalar_hashes}))"""
 
+  def gen(self, schema: LazyIrSchema) -> List[str]:
```
wonjoolee95, line 69: Most of this function is same as upstream codegen. This is updated to remove `torch::lazy::Shape` related code.

alanwaketan: It looks like maybe we can make the parts we don't need as a method in the `GenLazyIR` class such that we can substitute them with things that output empty strings for instances.

scripts/gen_lazy_tensor.py
```diff
@@ -61,11 +61,113 @@ def node_base_ctor_call(self, schema: LazyIrSchema) -> str:
...
+{has_optional_decls}
+
+}};
+
+""",
+    ]
+
+
+@dataclass(frozen=True)
+class GenXlaLazyNativeFuncDefinition(GenLazyNativeFuncDefinition):
```
alanwaketan, line 156: A few words to explain why we override the default one would be great. So other persons don't need to eyeball the gigantic f string below to find the differences.

scripts/gen_lazy_tensor.py
alanwaketan: You did a good job on GenXlaLazyNativeFuncDefinition. We can repeat that.

wonjoolee95: Thanks for the comment. It seems to be a bit harder to do make this overriding clean for the `GenLazyIR` class. The upstream `GenLazyNativeFuncDefinition` class has specific function named `shape_inference()` that we can override to remove the `torch::lazy::Shape` related logic. But for the upstream `GenLazyIR` class, the shape inference logic is part of a big `gen()` function so we are unable to override only the specific `torch:lazy:Shape` related codegen. I've added a comment to this function describing this, let me know if you think this is okay.

scripts/gen_lazy_tensor.py
wonjoolee95: Added comment.

scripts/gen_lazy_tensor.py
wonjoolee95: Sorry, accidentally deleted one of your original comments on this thread while I was trying to delete one of my older replies lol. Just for future reference, Jiewen's original comment was:
```
It looks like maybe we can make the parts we don't need as a method in the GenLazyIR class such that we can substitute them with things that output empty strings for instances.
```

wonjoolee95:
Building cpp tests now succeeding locally, will let CI to verify the remaining tests.

wonjoolee95:
Some unit test failures, looking into it.

wonjoolee95:
`bitwise_scalar` tests were failing: `TestBitwiseAndScalar`, `TestBitwiseOrScalar`, and `TestBitwiseXorScalar` due to `shape` check:
```
int [4, 2]
-vs-
long [4, 2]
```
The resulting shape was casted/promoted to `s64`, causing an assertion failure. Previously when using `torch::lazy::Shape`, the `XLANativeFunctions.cpp` populated `torch::lazy::Shape::logical_element_type` with the `torch::lazy::Shape`'s `scalar_type` as such: https://github.com/pytorch/xla/blob/647a3d5a0996994e2efb1c4452882b4876a949a1/torch_xla/csrc/tensor.cpp#L466-L468. 
Now since this logic and `torch::lazy::Shape` are both removed, codegen'ed `bitwise` ops casted/promoted the resulting shape to `long`. The last commit moves these ops to `ir_codegen` only, so we can explicitly call `DoBinaryOpWithoutPromo` in `aten_xla_type.cpp`. This does not introduce any new behavior or regression for these ops.

wonjoolee95:
@alanwaketan this should be ready for another review, thanks!

scripts/gen_lazy_tensor.py
```diff
@@ -61,11 +61,118 @@ def node_base_ctor_call(self, schema: LazyIrSchema) -> str:
...
+  def gen(self, schema: LazyIrSchema) -> List[str]:
+    opkind = schema.opkind or aten_symbol(schema)
+
+    # for now, we just want one IR class decl and soon after also the method defs
+    # and we use the functional version not out/inplace.
+    all_args = schema.filtered_args()
+    value_args = schema.filtered_args(values=True, scalars=False)
+    scalar_args = schema.filtered_args(values=False, scalars=True)
+
+    ctor_args = [f"const {i.lazy_type.cpp_type()}& {i.name}" for i in all_args]
```
alanwaketan, line 80: I guess you can provide a method in the base class probably called: `def should_use_lazy_shape(self, schema: LazyIrSchema) -> bool:` to replace this line. The default implementation should include the exact same logic as the line specified below, and then you can override it to false. Then you don't need to copy & paste the whole `gen` here.
https://github.com/pytorch/pytorch/blob/cda0d5a57b9126c6d244fdd5b02198f05c742615/torchgen/dest/lazy_ir.py#L255

test/cpp/test_symint.cpp
```diff
@@ -169,10 +169,8 @@ TEST(SymintTest, TestDynamicSymintArithmetic) {
       std::vector<torch::lazy::Shape>{
           torch::lazy::Shape(torch::kFloat, {10, 20, 30})};
 
-  torch::lazy::NodePtr abs_node =
-      torch::lazy::MakeNode<Abs>(expand_value, std::move(abs_lazy_shapes));
-  torch::lazy::NodePtr relu_node =
-      torch::lazy::MakeNode<Relu>(expand_value, std::move(relu_lazy_shapes));
+  torch::lazy::NodePtr abs_node = torch::lazy::MakeNode<Abs>(expand_value);
```
alanwaketan, line 172: Should we remove `abs_lazy_shapes` and `relu_lazy_shapes` as well?

alanwaketan:
> `bitwise_scalar` tests were failing: `TestBitwiseAndScalar`, `TestBitwiseOrScalar`, and `TestBitwiseXorScalar` due to `shape` check:
> 
> ```
> int [4, 2]
> -vs-
> long [4, 2]
> ```
> 
> The resulting shape was casted/promoted to `s64`, causing an assertion failure. Previously when using `torch::lazy::Shape`, the `XLANativeFunctions.cpp` populated `torch::lazy::Shape::logical_element_type` with the `torch::lazy::Shape`'s `scalar_type` as such:
> 
> https://github.com/pytorch/xla/blob/647a3d5a0996994e2efb1c4452882b4876a949a1/torch_xla/csrc/tensor.cpp#L466-L468
> 
> 
> Now since this logic and `torch::lazy::Shape` are both removed, codegen'ed `bitwise` ops casted/promoted the resulting shape to `long`. The last commit moves these ops to `ir_codegen` only, so we can explicitly call `DoBinaryOpWithoutPromo` in `aten_xla_type.cpp`. This does not introduce any new behavior or regression for these ops.

Just for my education, can you point me to where this promotion happens?

scripts/gen_lazy_tensor.py
wonjoolee95: Sounds good, I'll can update that. There is an ongoing related discussion regarding upstream codegen changes at https://github.com/pytorch/pytorch/pull/87823. I'll wait for that discussion to conclude and update this PR accordingly.

wonjoolee95:
> Just for my education, can you point me to where this promotion happens?

Sure thing! In the current codegen'ed `bitwise` operators, they call `PromotedBinaryOp` as shown at https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/ops_lower_fn.cpp#L158. This helper function defined at https://github.com/pytorch/xla/blob/master/torch_xla/csrc/helpers.cpp#L557 is responsible the promotion. FYI the PR that codegens `bitwise` operators is https://github.com/pytorch/xla/pull/3815/files.

alanwaketan:
Thanks, @wonjoolee95.

It looks like we don't want to promote scalar tensors. Because of decomposition, all the scalar variants of binary ops will be first converted to tensor version before reaching us (wrap the scalars into scalar tensors). Therefore, we need a way to detect scalar tensors and don't promote them.

On the other hand, it looks like we do want to promote scalars? Or because scalars are annotated differently in xla so they don't get promoted by xla as well.

Can you help me better understand the situation here?

wonjoolee95:
The problem here seems be the lack of `DoBinaryOpWithoutPromo` calls in the codegen'ed `bitwise` ops. Previous to being codegen'ed, all the `bitwise` ops were wrapped around `DoBinaryOpWithoutPromo` for **both** .scalar and .tensor variations: https://github.com/pytorch/xla/blob/1715ccd0c11553c0e7ab6550715ae83769165b55/torch_xla/csrc/aten_xla_type.cpp#L903-L910
So the old behavior looks like we don't want to promote scalars or tensors. And when we codegen'ed these ops at https://github.com/pytorch/xla/pull/3815/files, these `DoBinaryOpWithoutPromo` were omitted when we removed the bitwise ops from `aten_xla_type.cpp`.

alanwaketan:
Yea, if we dig into tensor variant of `DoBinaryOpWithoutPromo`, what it does is to change the dtype of the tensor if it's a scalar tensor. It does nothing to tensors. Correct me if I'm wrong.

For the scalar variant, it does nothing special. That's why I'm confused. @JackCaoG Do you have any insights?

test/cpp/test_symint.cpp
wonjoolee95: Updated

scripts/gen_lazy_tensor.py
wonjoolee95: Updated upstream lazy codegen to have boolean flag

wonjoolee95:
Since the original upstream PyTorch PR https://github.com/pytorch/pytorch/pull/87823 got merged, opened a new one to add the `use_lazy_shape` flag, which allows us to not override the entire `GenLazyIr.gen()` function -- https://github.com/pytorch/pytorch/pull/88444.

alanwaketan:
> Yea, if we dig into tensor variant of `DoBinaryOpWithoutPromo`, what it does is to change the dtype of the tensor if it's a scalar tensor. It does nothing to tensors. Correct me if I'm wrong.
> 
> For the scalar variant, it does nothing special. That's why I'm confused. @JackCaoG Do you have any insights?

Just talked to Jack. It looks like the rule here is just to pass all the tests.

alanwaketan (approved):
LGTM. Thanks!

wonjoolee95:
Upstream PR is merged, will merge this PR without the torch_pin.

PR #4116, opened 2022-10-22
Disable fallback on xla::dot int inputs

JackCaoG:
confirmed that `xla:dot` can handle int type now.


wonjoolee95 (approved):
👍 should be good to merge after CIs are green!

PR #4119, opened 2022-10-25
Dynamo Poc

JackCaoG:
This pr is based on @shunting314 branch at https://github.com/shunting314/torchdynamo/tree/dynamo-torchxla-integration. I am helping him to add test and landing this pr to pytorch/xla.

We have tested it in an old version and was able to get inference running. 

Upstream pytorch pr https://github.com/pytorch/pytorch/pull/87741


JackCaoG:
Test should pass now, want to take another pass to review some of the cpp implementation and will request for review.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+   * Return tensor ids and tensors for DeviceData nodes.
+   */
+  m.def("_get_tensors_xla_device_data_node",
+        [](const std::vector<at::Tensor>& tensors)
+            -> std::pair<std::vector<int64_t>, std::vector<at::IValue>> {
+          std::vector<int64_t> tensor_ids;
+          std::vector<at::IValue> ivalues;
+          std::vector<torch::lazy::Node*> roots;
+          for (auto& tensor : tensors) {
+            auto xtensor = bridge::TryGetXlaTensor(tensor);
```
JackCaoG, line 1467: we should check for this value before appending to the list, this might segfault if tensor is not an xlaTensor.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+              data_handles;
+
+          for (auto nodeptr : post_order) {
+            const torch_xla::DeviceData* device_data =
+                torch_xla::DeviceData::Cast(nodeptr);
+            if (!device_data) {
+              continue;
+            }
+
+            const auto backend_data = device_data->data();
```
JackCaoG, line 1481: I am going to get rid of these autos since it make code reading a bit hard.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+
+            // follow the implementation of XLATensor::CreateTensorNode
+            auto dataptr =
+                ((torch_xla::XLAData*)backend_data.get())->xla_data();
+            TORCH_CHECK(dataptr);
```
JackCaoG, line 1486: TODO: use XLA_CHECK

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+              continue;
+            }
+
+            // add tensor_id after we make sure the handle does not exist yet.
+            tensor_ids.push_back(infoptr->tensor_id);
+            // This will create a new XLATensor with different Id which is not
+            // ideal. However there is not an easy way to get the origional
+            // tensor with the tensor_id.
+            // TODO(JackCaoG): invesgate if we have idle XLATensor.
+            auto tensor = bridge::AtenFromXlaTensor(
```
JackCaoG, line 1502: TODO, I think we can just create a map at the beginning of this method and reuse the pass in `at::Tensor` instead of creating new ones.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+                      xtensor->CurrentIrValue().node.get()) != nullptr) {
+                need_materialization.push_back(false);
+              } else {
+                // input tensor is an IR other than DeviceData which means a
+                // compuation is required to get the value of this tensor.
+                need_materialization.push_back(true);
+              }
+            } else {
+              // TODO: maybe also handle it is a XLATensor with tensor_data case
+              XLA_CHECK(false);
```
JackCaoG, line 1534: add more detail error messages.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+
+  m.def("_run_cached_graph",
+        [](const std::string& hash_str,
+           const std::vector<at::IValue>& graph_inputs)
+            -> std::vector<at::Tensor> {
+          TORCH_CHECK(hash_str.size() == sizeof(torch::lazy::hash_t));
+          torch::lazy::hash_t hash = *(torch::lazy::hash_t*)(hash_str.c_str());
+          auto cachedComputation = XLATensor::GetComputationCache()->Get(hash);
+          // TODO implement a fallback mechanism, or make sure those entries
+          // never get kicked out
```
JackCaoG, line 1559: @shunting314 I think we can have a larger default cache size for dynamo, but a fallback sounds more robust solution. How does non-xla backend handles it? Do they never clear the cache?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+          XLA_CHECK(cachedComputation) << "Failed to get computation by hash "
+                                       << torch::lazy::HashToString(hash)
+                                       << ". Maybe the entry get "
+                                          "kicked out of the LRU cache";
+          auto start_prep_input = std::chrono::high_resolution_clock::now();
```
JackCaoG, line 1564: todo, add a proper XLA counter for this.

torch_xla/csrc/init_python_bindings.cpp
shunting314: One way to handle this is to propagate a python exception up to dynamo so it can fall back to eager / or trigger recompile. But it's not there yet.

I'm thinking maybe we can have some statistics for how many entries a normal model need first. If the number is usually quite small, then a fallback here should be something good to have but not on the critical path. What do you think?

shunting314: The ts backend does similar thing. We never clear cache explicitly.

JackCaoG: for the most part cache = CompileTime. It is usually < 50, for the most working models, it is < 25. However for dynamo I am not entirely sure since the tracing mechanism is slight different but I don't expect this to be a huge number (otherwise compilation will take forever).

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+
+          // setup the parameters_data
+          std::vector<xla::ComputationClient::DataPtr> parameters_data;
+          auto device = torch_xla::GetCurrentDevice();
```
JackCaoG, line 1568: This is OK for now, but I felt like it should be something passed to this function from python. @shunting314 How do you think?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,184 @@ void InitXlaModuleBindings(py::module m) {
...
+            auto xlaTensorOpt = bridge::TryGetXlaTensor(tensor);
+            xla::ComputationClient::DataPtr dataptr;
+            if (xlaTensorOpt) {
+              torch::lazy::BackendDataPtr backend_data_ptr =
+                  xlaTensorOpt->GetXlaData();
+              dataptr =
+                  dynamic_cast<torch_xla::XLAData*>(backend_data_ptr.get())
+                      ->xla_data();
+            } else {
+              auto backend_data =
```
JackCaoG, line 1581: TODO: add counter

JackCaoG:
I will fix comments I raised then I think this pr is good to go. FYI @wconstab

test/test_dynamo_integrations_util.py
```diff
@@ -0,0 +1,96 @@
...
+    xla_out += 1
+    assert (hash != torch_xla._XLAC._get_graph_hash([xla_out]))
+
+    # Repeat the same computation on input with same shape, hash should be the same
+    xla_input2 = torch.randn(20, 1, 14, 14).to(xla_device)
+    xla_out_2 = xla_dummy_model(xla_input)
+    assert (hash == torch_xla._XLAC._get_graph_hash([xla_out_2]))
+
+  def test_run_cached_graph(self):
+    xla_device = xm.xla_device()
```
JackCaoG, line 74: TODO: add test where input is on CPU

JackCaoG, line 74: TODO: add test where input is on CPU

torch_xla/csrc/init_python_bindings.cpp
JackCaoG: the default cache size is 1024 according to https://github.com/pytorch/xla/blob/d3b1a7db6098b9181d2251f015f5992218d4c38b/torch_xla/csrc/tensor.cpp#L1366

I think we should be fine now, but I will leave this TODO here.

torch_xla/csrc/init_python_bindings.cpp
JackCaoG: OK I am wrong. The tensor here might not be part of the input tensors. For example
```
    t1 = torch.randn(20, 5).to(xla_device)
    t2 = torch.randn(20, 5).to(xla_device)
    t3 = t2 + t1
    t4 = t3 * t2
    res_pair = torch_xla._XLAC._get_tensors_xla_device_data_node([t4])
```

both `t1` and `t2` are device_datas but they are not part of input passed to `_get_tensors_xla_device_data_node`

JackCaoG: actually this is OK, when I look at `RegisterTensor`, when it sees same `XlaData` it will keep the TensorID

JackCaoG:
Ok I think this one is ready for review now

shunting314 (commented):
Thanks you very much for helping merging this!

LGTM. But I'll let other people in your team to stamp in case they have other feedbacks

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1441,8 +1442,176 @@ void InitXlaModuleBindings(py::module m) {
...
 
   BuildProfilerSubmodule(&m);
-}
 
+  m.def("_get_tensors_handle",
+        [](const std::vector<at::Tensor>& tensors) -> std::vector<int64_t> {
+          std::vector<int64_t> handles;
+          handles.reserve(tensors.size());
+          for (auto& tensor : tensors) {
+            handles.push_back(bridge::GetXlaTensor(tensor)->GetOpaqueHandle());
```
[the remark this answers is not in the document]
JackCaoG: I used it in test to make sure `_get_tensors_xla_device_data_node` do not create new datas.

byronyi:
That's a great update. 

Nit: I'm wondering why you implement the cache in XLA instead of passing back an `XlaComputation` for dynamo to keep the object (and purge when re-compile) in its own cache.

PR #4121, opened 2022-10-25
Release document for PJRT

will-cromar:
Tested manually with r1.13 pre-release image.

### DIFF
--- a/docs/pjrt.md
+++ b/docs/pjrt.md
@@ -53,6 +53,8 @@ By default, PjRt will use all TPU chips. To use only one TPU chip, configure
 TPU_PROCESS_BOUNDS=1,1,1 TPU_VISIBLE_CHIPS=0 PJRT_DEVICE=TPU python3 xla/test/test_train_mp_imagenet.py --fake_data --batch_size=256 --num_epochs=1
 ```
 
+Note: `xmp.spawn`'s `nprocs` argument is not implemented for PjRt.
+
 #### Pods
 
 On TPU Pods, use `gcloud` to run your command on each TPU in parallel:
--- a/docs/pjrt.md
+++ b/docs/pjrt.md
@@ -33,17 +33,24 @@ PJRT_DEVICE=CPU python3 xla/test/test_train_mp_mnist.py --fake_data
 
 ### v4 TPU
 
+To create a new TPU with PyTorch/XLA r1.13 installed:
+
+```
+gcloud alpha compute tpus tpu-vm create $USER-pjrt --accelerator-type=v4-8 --version=tpu-vm-v4-pt-1.13 --zone=us-central2-b --project=$PROJECT
+```
+
 On a v4-8, you can run our ResNet50 example like this:
 
 ```
-PJRT_DEVICE=TPU python3 xla/test/test_train_mp_imagenet.py --fake_data
+git clone --depth=1 --branch r1.13 https://github.com/pytorch/xla.git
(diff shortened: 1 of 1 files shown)


docs/pjrt.md
```diff
@@ -0,0 +1,142 @@
...
+```
+git clone --depth=1 --branch r1.13 https://github.com/pytorch/xla.git
+PJRT_DEVICE=TPU python3 xla/test/test_train_mp_imagenet.py --fake_data --batch_size=256 --num_epochs=1
+```
+
+By default, PjRt will use all TPU chips. To use only one TPU chip, configure
+`TPU_PROCESS_BOUNDS` and `TPU_VISIBLE_CHIPS`:
+
+```
+TPU_PROCESS_BOUNDS=1,1,1 TPU_VISIBLE_CHIPS=0 PJRT_DEVICE=TPU python3 xla/test/test_train_mp_imagenet.py --fake_data --batch_size=256 --num_epochs=1
```
JackCaoG, line 53: what if user changed https://github.com/pytorch/xla/blob/master/test/test_train_mp_imagenet.py#L272 `nprocs` to 1?

docs/pjrt.md
```diff
@@ -0,0 +1,142 @@
...
+per TPU host. See the [Cloud TPU documentation](https://cloud.google.com/tpu/docs/system-architecture-tpu-vm)
+for more information about TPU architecture.
+
+- Performance gains are possible for workloads constrained by data transfer
+  speeds.
+- Under XRT, the server process is the only process that interacts with the TPU
+  devices, and client processes don't have direct access to the TPU devices.
+  When profiling a single-host TPU (e.g. v3-8 or v4-8), you would normally see 8
+  device traces (one for each TPU core). With PjRt, each process has one chip,
+  and a profile from that process will show only 2 TPU cores.
```
JackCaoG, line 89: on v3 right? I thought we are only going to see 1 device trace on v4.

docs/pjrt.md
will-cromar: `nprocs` is actually ignored in the PJRT case right now. There's a feature request on me internally to implement a special case for `nprocs=1`

docs/pjrt.md
will-cromar: I can't find any screenshots, but I'm pretty sure that v4 also shows both cores in the device trace.

docs/pjrt.md
JackCaoG: can we update the doc to reflect that? User might try to set it and will be confused. Ideally we should throw an error now if user use PJRT and set `nprocs` to 1 on TPU, but we don't need to cherry pick this change now.

docs/pjrt.md
```diff
@@ -0,0 +1,142 @@
...
+- `torch.distributed` uses a global process group and does not support
+  multi-threading, so the `xla` `torch.distributed` backend will not work with
+  PjRt and TPU v2 and v3 at this time.
+- Because the current implementation of `xm.rendezvous` for PjRt relies on
+  `torch.distributed`, `xm.rendezvous` is not supported with PjRt on TPU v2 and
+  v3.
+
+### Compatible examples
+
+For an overview of the changes required to migrate from TPU v2/v3 to v4, compare
```
JackCaoG, line 126: Is this part still necessary? I think the old imagenet example works on PJRT right? or does it only works on v4?

will-cromar: It only works on v4

JackCaoG (approved):
Can we also cherry-pick this into 1.13 release branch?

docs/pjrt.md
will-cromar: Done.

PR #4125, opened 2022-10-26
node.js v18 gets installed by default since 10/25/2022 due to new lts version available. Lock the version to v16 on bionic

weiwangmeta:
Node.js v18 requires glibc 2.28 per release notes here (https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#toolchain-and-compiler-upgrades). (Bionic does not have glibc 2.28 support by default). May be able to avoid https://github.com/pytorch/pytorch/pull/87737

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -66,7 +66,7 @@ function install_deps_pytorch_xla() {
 
   sudo apt-get -qq update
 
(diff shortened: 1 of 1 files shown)


JackCaoG:
@wonjoolee95 Can you rebase after this one merged and see if cache can be reenabled?

PR #4126, opened 2022-10-26
node.js v18 gets installed by default since 10/25/2022 due to new lts…

vanbasten23:
… version available. Lock the version to v16 on bionic (#4125)

* node.js v18 gets installed by default. Lock the version to v16 on bionic (ubuntu 18) as node.js v18 requires glibc 2.28 per release notes. (Bionic does not have glibc 2.28 support by default). May be able to avoid https://github.com/pytorch/pytorch/pull/87737

* supply v16.18.0 at the correct location


JackCaoG (approved):
Thanks!

PR #4139, opened 2022-10-28
[SPMD] Use on_device_shape in ExecuteReplicated

yeounoh:
Use `buffer->on_device_shape` to support tfrt in ExecuteReplicated mode. Also fixes some minor bugs introduced during the latest SPMD refactoring.


yeounoh:
Verified on a TPUVM with simple matmul example.

PR #4143, opened 2022-11-01
Fix clear counter

JackCaoG:
My previous implementation clear the `vector<CounterData>` and `vector<MetricsData>` directly, but this is incorrect. Counters and Metrics and declared as static variable in different places of code, and won't get recreated if being deleted. What we can do is to reset the value of those counters and make sure they don't show up in report if value is zero. Added additional test to account for this case.


alanwaketan (approved):
LGTM.

Issue #4095, opened 2022-10-14
Test tensor deep copy with sharding


steventk-g:
The Pr https://github.com/pytorch/xla/pull/3684 introduces a sharding spec property on tensors. We want to test that a deep copy preserves the sharding spec.

yeounoh:
This was addressed in #4144

PR #4144, opened 2022-11-01
[SPMD] Test sharding_spec with clear_sharding, deepcopy

yeounoh:
This addresses the follwoing:
- Implement `xla_sharding.clear_sharding` API to reset sharding annotation.
- Test/verify `deepcopy` preserves sharding annotation #4095 
- Refactoring based on the comments from #3684

### DIFF
--- a/test/test_xla_sharding.py
+++ b/test/test_xla_sharding.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import copy
 
 import unittest
 import numpy as np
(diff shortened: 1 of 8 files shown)


test/test_xla_sharding.py
```diff
@@ -18,24 +19,17 @@
...
-    partition_spec = (1,)
-    t1 = torch.tensor([2.0, 3.0], dtype=torch.float, device=xm.xla_device())
-    t1_sharded = XLAShardedTensor(t1, mesh_shape, partition_spec)
-    t2 = torch.tensor([2.0, 3.0], dtype=torch.float, device=xm.xla_device())
-    t3 = torch.add(t1_sharded, t2)
+    partition_spec = (0, 1)
+    xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                       dtype=torch.float,
+                       device=xm.xla_device())
+    xst1 = xs.mark_sharding(xt1, mesh_shape, partition_spec)
```
JackCaoG, line 29: so xt1 stays as `torch.Tensor` but under the hood sharding spec is already being inserted? From here on, should people use `xt1` or `xst1`? This part is a bit confusing.

torch_xla/experimental/xla_sharding.py
```diff
@@ -74,4 +74,7 @@ def mark_sharding(t: Union[torch.Tensor,
 
 def clear_sharding(t: Union[torch.Tensor, XLAShardedTensor]) -> torch.Tensor:
   """Clear sharding annotation from the input tensor and return a `cpu` casted tensor."""
-  return NotImplemented
+  torch_xla._XLAC._xla_clear_sharding(t)
+  if isinstance(t, XLAShardedTensor):
```
JackCaoG, line 78: in what case we will call `clear_sharding` on a non-XLAShardedTensor?

test/test_xla_sharding.py
yeounoh: Yes, annotation doesn't change the type of tensor. Annotation is attached to the backing IR. people can use both `xt1` and `xst1` because `xst1` is just a wrapper that dispatches to `xt1` and it's XLA backend. This reason for this is that the user can simply take their existing code and don't change/replace tensors with sharded tensors explicitly (so the true annotating); also returned handle to the sharded tensor provides an access to XLAShardedTensor specific APIs (e.g., sharding_spec, accessing shards, etc.).

torch_xla/experimental/xla_sharding.py
yeounoh: In the example above, one may call `xs.claer_sharding` on `xt1` just like `mark_sharding`.

steventk-g (approved):
LGTM!

jonb377 (approved):
LGTM! Nice catch on the shared_ptr issue 🙂

torch_xla/experimental/xla_sharding.py
JackCaoG: ok sg.

test/test_xla_sharding.py
JackCaoG: should we consider rename `mark_sharding` to `mark_sharding_` since it is actully an inplace op?

an op that both inplace update the op and returns it is a bit counterintuitive. IMO we should at least make it clear that it is inplace.

JackCaoG (approved):
lgtm, thanks

PR #4156, opened 2022-11-04
Enable `padded` option for even tensor sharding

yeounoh:
This fixes the inconsistent shard size issue I faced while testing convolution layer sharding. This has been the issue, but wasn't surfaced during simple linear testing. Also fixed a minor bug, where `xrt_world_size()` returns all the devices, not just `xla` devices.

### DIFF
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -39,7 +39,8 @@ TEST_F(XLAShardingTest, ShardTensor) {
           CreateComputationShapeFromTensor(tensor, GetDefaultDevice()),
           devices.size())
           .ToProto();
-  auto shards = ShardingUtil::ShardTensor(tensor, sharding, devices);
+  auto shards =
+      ShardingUtil::ShardTensor(tensor, sharding, devices, /*padded=*/false);
   EXPECT_EQ(shards.size(), 8);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({1}));
   EXPECT_EQ(shards[1].sizes(), c10::ArrayRef<long>({1}));
@@ -52,7 +53,8 @@ TEST_F(XLAShardingTest, ShardTensor) {
       {4, 5, 6, 7},
   });
   sharding = xla::HloSharding::Tile(mesh).ToProto();
-  shards = ShardingUtil::ShardTensor(tensor, sharding, devices);
+  shards =
+      ShardingUtil::ShardTensor(tensor, sharding, devices, /*padded=*/false);
   EXPECT_EQ(shards.size(), 8);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({4, 2, 4}));
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({4, 1, 4}));
@@ -61,14 +63,16 @@ TEST_F(XLAShardingTest, ShardTensor) {
   // size should be smaller in dim=1 because it's not evenly divisible.
   xla::Array3D<int64_t> cube({{{0, 1}, {2, 3}, {4, 5}, {6, 7}}});
   sharding = xla::HloSharding::Tile(cube).ToProto();
-  shards = ShardingUtil::ShardTensor(tensor, sharding, devices);
+  shards =
+      ShardingUtil::ShardTensor(tensor, sharding, devices, /*padded=*/false);
   EXPECT_EQ(shards.size(), 8);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({8, 2, 2}));
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 1, 2}));
 
   // Replicated, all shards should be identical.
   sharding = xla::HloSharding::Replicate().ToProto();
-  shards = ShardingUtil::ShardTensor(tensor, sharding, devices);
+  shards =
+      ShardingUtil::ShardTensor(tensor, sharding, devices, /*padded=*/false);
   EXPECT_EQ(shards.size(), 8);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({8, 7, 4}));
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 7, 4}));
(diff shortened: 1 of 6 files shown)


test/cpp/test_xla_sharding.cpp
```diff
@@ -83,6 +83,13 @@ TEST_F(XLAShardingTest, ShardTensor) {
   EXPECT_EQ(shards.size(), 8);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({1, 8, 2, 2}));
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({1, 8, 1, 2}));
+
+  // 4D tiled and padded, all shard sizes should be idential.
+  shards =
+      ShardingUtil::ShardTensor(tensor, sharding, devices, /*padded=*/true);
```
JackCaoG, line 89: in what case should we pad?

torch_xla/csrc/tensor_util.cpp
```diff
@@ -926,8 +926,8 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
       // across global devices for multi-host training.
       std::vector<std::string> local_devices =
           xla::ComputationClient::Get()->GetLocalDevices();
-      std::vector<at::Tensor> shards =
-          ShardingUtil::ShardTensor(tensors[i], sharding, local_devices);
+      std::vector<at::Tensor> shards = ShardingUtil::ShardTensor(
+          tensors[i], sharding, local_devices, /*padded=*/true);
```
JackCaoG, line 930: can you leave a comment explain why a padding is needed here?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -271,12 +271,12 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
                               (tensor.sizes()[2] % tile_shape[2] != 0);
         shard = tensor.index(
             {at::indexing::Slice(
-                 (core / tile_shape[1] / tile_shape[2]) * x_partition,
-                 (core / tile_shape[1] / tile_shape[2] + 1) * x_partition),
-             at::indexing::Slice((core / tile_shape[2]) * y_partition,
-                                 (core / tile_shape[2] + 1) * y_partition),
-             at::indexing::Slice((core % tile_shape[2]) * z_partition,
-                                 (core % tile_shape[2] + 1) * z_partition)});
+                 (i / tile_shape[1] / tile_shape[2]) * x_partition,
```
JackCaoG, line 274: why change from `core` to `i` here?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -311,32 +311,49 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
...
+      int64_t core = sharding.tile_assignment_devices()[i];
       shards[core] = shard.contiguous(at::MemoryFormat::Contiguous);
     }
   } else if ((sharding.type() == xla::OpSharding::MANUAL) ||
              (sharding.type() == xla::OpSharding::TUPLE)) {
     TF_LOG(ERROR) << "Unsupported OpSharidng type " << sharding.type();
   }
+
+  // Zero-pad to the right to ensure the sizes are even
+  if (shards.size() > 0 && padded) {
```
JackCaoG, line 343: I have to admit I don't fully understand what this pad is doing, if you can give an example input and example output that will be very helpful.

torch_xla/experimental/xla_sharded_tensor.py
```diff
@@ -58,7 +58,7 @@ class XLAShardedTensor(torch.Tensor):
   # 4-way row-wise, and replicate column-wise.
   # >> input = torch.randn(8, 10)
   # >> mesh_shape = (4, 2)
-  # >> assert np.prod(mesh_shape) == xm.xrt_world_size()
+  # >> assert np.prod(mesh_shape) == len(xm.get_xla_supported_devices())
```
JackCaoG, line 61: would it also include `xla:cpu`?

test/cpp/test_xla_sharding.cpp
yeounoh: We should pad always -- I was debating whether to pad outside this util function, leaving it more generic tensor sharding. Instead, I introduced a padded (default true) arg for the util function.

torch_xla/csrc/tensor_util.cpp
yeounoh: Done

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: To enable custom device assignment, so the i-th shard is indexed by core, which can be a custom device order.

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: When we split a tensor to shards, this zero-padding ensures that we always pad the last dimension(s) of any smaller shards, so that all the shards have the same size.
Suppose we want to make an 1x4 shard to 2x8 shard,

`{{0, 1, 2, 3}}` will become `{{0, 1, 2, 3, 0, 0, 0, 0}, {0, 0, 0, 0, 0, 0, 0, 0}}`.

i.e., zero-pad to the right (from the last dimensions).

torch_xla/experimental/xla_sharded_tensor.py
yeounoh: Yes

JackCaoG: oh ok nvm. I am thinking about the case of XRT where we will have 1 CPU and 8 TPU. I think the PJRT case we will only have TPU as XLA device unless we specified `PJRT_DEVICE` to be CPU.

JackCaoG (approved):

JackCaoG:
can you fix the test error?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -311,32 +311,49 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
...
+    for (size_t i = 1; i < shards.size(); ++i) {
+      std::vector<long> pads;
+      for (size_t j = 0; j < shards[i].sizes().size(); ++j) {
+        XLA_CHECK_GE(shards[0].sizes().at(j), shards[i].sizes().at(j));
+        pads.push_back(shards[0].sizes().at(j) - shards[i].sizes().at(j));
+        pads.push_back(0);  // no padding on lhs
+      }
+      // Padding starts from the last dimension
+      std::reverse(pads.begin(), pads.end());
+      shards[i] = at::constant_pad_nd(
```
jonb377, line 353: Just for my understanding - if we apply padding to the shards, does the padding need to be removed when reassembling the tensor?

jonb377 (approved):
LGTM, thanks!

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: Good question, we don't need to reassemble for the computation execution, but we will need to account for the padding when we do (e.g., provide access to the shards, etc.). Let's create a task to track.

PR #4176, opened 2022-11-09
[LTC] Employ more Computation

alanwaketan:
Summary:
Replace std::shared_ptr<xla::ComputationClient::Computation> with ComputationPtr in CompilationResult and CachedComputation.

Test Plan:
CI.


JackCaoG (approved):
This computation is really message lol. I guess it is the pain we have to pay for tf and pytorch can't depend on each other.

alanwaketan:
> This computation is really message lol. I guess it is the pain we have to pay for tf and pytorch can't depend on each other.

I guess we can go over the history of ComputationClient in more details offline. I hope that we can refactor it a bit such that we can get rid of these complexity.

PR #4177, opened 2022-11-09
Make ExecuteComputationOptions as default

alanwaketan:
Summary:
ExecuteComputationOptions has always been used to pass the default members. Therefore, let's make it as a default parameter of ComputationClient::ExecuteComputation().

Test Plan:
CI.


JackCaoG (approved):

alanwaketan:
Thanks, Jack.

PR #4178, opened 2022-11-09
[LTC] Use torch::lazy::Util::ComputePostOrder

alanwaketan:
Summary:
This pull request replaces all Util::ComputePostOrder with torch::lazy::Util::ComputePostOrder. In order, it also removes the xla implementation plus GetGraphSize.

Test Plan:
CI

### DIFF
--- a/torch_xla/csrc/debug_util.cpp
+++ b/torch_xla/csrc/debug_util.cpp
@@ -54,7 +54,7 @@ DebugUtil::GraphFormat DebugUtil::GetDefaultGraphFormat() {
 std::string DebugUtil::GetTensorsGraphInfo(
     absl::Span<const XLATensorPtr> tensors, const std::vector<size_t>* indices,
     GraphFormat format) {
-  std::vector<torch::lazy::Node*> root_nodes;
+  std::vector<const torch::lazy::Node*> root_nodes;
   std::vector<torch::lazy::Value> root_values;
   std::vector<torch::lazy::hash_t> root_hashes;
   xla::util::Unique<torch::lazy::BackendDevice> unique_device;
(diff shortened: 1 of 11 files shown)


torch_xla/csrc/debug_util.cpp
```diff
@@ -54,7 +54,7 @@ DebugUtil::GraphFormat DebugUtil::GetDefaultGraphFormat() {
 std::string DebugUtil::GetTensorsGraphInfo(
     absl::Span<const XLATensorPtr> tensors, const std::vector<size_t>* indices,
     GraphFormat format) {
-  std::vector<const torch::lazy::Node*> root_nodes;
+  std::vector<torch::lazy::Node*> root_nodes;
```
JackCaoG, line 57: I think I run into some issue before, but can you remind me why do we have to drop the const here?

alanwaketan: Just the upstream `ComputePostOrder` doesn't take const type pointers so that every argument being eventually passed in that function need to drop the const qualifier. Here the `root_nodes` will eventually be passed to `DumpUtil::ToTex` or `DumpUtil::ToDot` which eventually use the `ComputePostOrder`.

JackCaoG: would it be better to have upstream to add the const qualifier? `ComputePostOrder` shouldn't modify the `root_nodes` AFAIK.

alanwaketan: I suspect the reason they dropped the const qualifier, they must did some quirky things to modify the roots. Yea, I can figure that out.

alanwaketan: Looks like it is due to the fact that c10::ArrayRef didn't support c10::ArrayRef<const T>. I think now it actually supports it. Let me make another PR in upstream.
https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/ir_util.cpp#L11-L12

alanwaketan: Okay, I think the patch is ready for another round of reviews.

JackCaoG (approved):

alanwaketan:
Thanks, Jack.

PR #4179, opened 2022-11-09
[WIP] part codegen atan2

ManfeiBai:
partially codegen atan2

---

Generated `LazyIr.h`:

```
class Atan2 : public XlaNode {
 public:
    return torch::lazy::OpKind(at::aten::atan2);
  }

  Atan2(const torch::lazy::Value& self, const torch::lazy::Value& other)
      : XlaNode(torch::lazy::OpKind(at::aten::atan2),
              {self, other},
              [&]() { return Atan2OutputShape(self, other); },
              /* num_outputs */ 1,
              torch::lazy::MHash())
  {
    
  }

    std::stringstream ss;
    ss << XlaNode::ToString();
    
    return ss.str();
  }

  

  bool CanBeReused(const torch::lazy::Value& self, const torch::lazy::Value& other) const {
    return false;
    }

};
```

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -163,20 +163,16 @@ std::pair<XLATensorPtr, XLATensorPtr> GetBinaryOperands(
 
 // The input is in format of {N, C, H, W} and the output will be {H, W}.
 std::vector<int64_t> GetOutputSizeWithScale(
-    absl::Span<const int64_t> input_size,
-    const c10::optional<at::ArrayRef<double>>& scale_factors,
-    const at::OptionalIntArrayRef& output_size) {
-  if (!output_size) {
-    XLA_CHECK(scale_factors);
-    XLA_CHECK_EQ(scale_factors->size(), 2);
-    // Calculate the output size from input_shape and scale_factors
-    XLA_CHECK_EQ(input_size.size(), 4);
-    int64_t output_h = input_size[2] * (*scale_factors)[0];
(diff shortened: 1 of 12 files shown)


wonjoolee95:
@ManfeiBai, for `ir_gen` ops, we need to list them both under `ir_gen` and `supported`. The `ir_gen` just generates the `LazyIr.h` for the listed op, but we still need to list them under `supported`. Try this, it should take care of the build failure. Take a look at `bitwise_and.Tensor` op example.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -692,11 +688,16 @@ at::Tensor XLANativeFunctions::atan2(const at::Tensor& self,
...
-                        at::ScalarType dtype) {
-                      return XLATensor::atan2(xself, xother, dtype);
-                    });
+
+  auto common_device = torch_xla::bridge::GetXlaDevice(self, other);
+  XLA_CHECK(common_device);
+  torch::lazy::NodePtr node = torch::lazy::MakeNode<Atan2>(
+      bridge::GetXlaTensor(self)->GetIrValue(),
+      bridge::GetXlaTensor(other)->GetIrValue(),
+      bridge::GetXlaTensor(at::result_type(self, other))->GetIrValue());
```
wonjoolee95, line 697: The `MakeNode` calls the constructor for `Atan2` node defined at `LazyIr.h`:
```
Atan2(const torch::lazy::Value& self, const torch::lazy::Value& other)
```
It should only take `self` and `other` tensors as parameters.

PR #4181, opened 2022-11-09
Implement `xm.rendezvous` with XLA collective communication

will-cromar:
We have found that `gloo` doesn't scale effectively to large pod sizes, and it's not easily possible to use `torch.distributed` in a multithreaded context such as TPU v3.

- `xm.rendezvous` will now call `xm.mark_step` to sync results from XLA.
- Support multithreaded contexts like TPU v2/v3.
- Don't initialize an XLA process group in `xm.rendezvous`. Also remove initialization based on `XRT_MESH_SERVICE_ADDRESS` since host 0 is not predictable anyway.
- Require that user calls `xm.rendezvous` from _all_ replicas per [XLA requirements](https://www.tensorflow.org/xla/operation_semantics#allreduce): `Computing the result of AllReduce requires having one input from each replica, so if one replica executes a AllReduce node more times than another, then the former replica will wait forever.` This covers the vast majority of the usage of `rendezvous` in our experience.

Tested manually on a TPU v4-8 with 1 process and 4 threads to simulate a v3.

### DIFF
--- a/test/pjrt/test_mesh_service.py
+++ b/test/pjrt/test_mesh_service.py
@@ -1,9 +1,6 @@
-import itertools
-import os
-from typing import List, Optional
+import functools
 from absl.testing import absltest, parameterized
 
-import torch.distributed as dist
 import torch_xla.core.xla_model as xm
 from torch_xla.experimental import pjrt
 
@@ -42,6 +39,19 @@ class PjRtMeshServiceTest(parameterized.TestCase):
 
     expected = [b'message'] * len(results)
     self.assertDictEqual(results, {r: expected for r in results})
+
+  def test_rendezvous_empty_payload(self):
+    test_fn = functools.partial(xm.rendezvous, 'test rendezvous', b'')
+    results = pjrt._run_multiprocess(test_fn)
+
+    expected = [b''] * len(results)
+    self.assertDictEqual(results, {r: expected for r in results})
+
+  def test_rendezvous_string_payload(self):
+    test_fn = functools.partial(xm.rendezvous, 'test rendezvous', "")
+
+    with self.assertRaises(TypeError):
+      pjrt._run_multiprocess(test_fn)
 
   @staticmethod
   def _mesh_reduce():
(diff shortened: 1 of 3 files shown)


torch_xla/experimental/pjrt.py
```diff
@@ -259,57 +259,44 @@ def broadcast_master_param(model: nn.Module) -> None:
...
-    import torch_xla.distributed.xla_backend
-    dist.init_process_group(
-        "xla",
-        init_method=init_method,
-        world_size=global_device_count(),
-        rank=global_ordinal())
+  assert not ordinals or len(
+      ordinals) == global_device_count(), 'Only global rendezvous is supported'
+  assert isinstance(payload, bytes)
+  data = torch.tensor(list(payload), dtype=torch.uint8)
```
ronghanghu, line 284: Perhaps we should first call `xm.mark_step()` at the beginning of this rendezvous function before this line to finish all execution of previous graphs? Otherwise, the collective ops below might be added as part of the HLO of earlier model training or inference execution and trigger a recompilation of the model (that could take several minutes).

will-cromar: Good idea! Done.

AlexWertheim:
I tested this on v4-8 and v4-4096 and it seems to work successfully on both accelerator types. Using `gloo` on `v4-4096` resulted in `connection refused` and `connection closed by peer` errors when calling `xm.rendezvous`; after discussion with @will-cromar and @JackCaoG, we suspect there are limits with the number of active tcp connections between devices.

test/pjrt/test_mesh_service.py
```diff
@@ -1,34 +1,57 @@
-import itertools
-import os
-from typing import List, Optional
+import functools
```
JackCaoG, line 1: is this test being run as part of the CI? I can't seems to find it.

will-cromar: This is only good as a sanity check since we can't run XLA collective ops on CPU and we don't support multi-GPU yet, but I just added it anyway.

JackCaoG (approved):
Anything in the PJRT README we should update?

Issue #2508, opened 2020-09-23
Crash using XLA GPU


zcain117:
## 🐛 Bug

I'm trying to use pytorch/xla + GPU. I tried running on 4 GPUs (v100) using:

```
unset XRT_TPU_CONFIG
export GPU_NUM_DEVICES=4
python3 pytorch/xla/test/test_train_mp_imagenet.py \
  ...
```

The test ran 1 epoch of training and 1 eval loop successfully and then fails with:

`Traceback (most recent call last): File "/root/anaconda3/envs/pytorch/lib/python3.6/site-packages/torch_xla/distributed/xla_multiprocessing.py", line 330, in _mp_start_fn _start_fn(index, pf_cfg, fn, args) File "/root/anaconda3/envs/pytorch/lib/python3.6/site-packages/torch_xla/distributed/xla_multiprocessing.py", line 324, in _start_fn fn(gindex, *args) File "/pytorch/xla/test/test_train_mp_imagenet.py", line 259, in _mp_fn accuracy = train_imagenet() File "/pytorch/xla/test/test_train_mp_imagenet.py", line 246, in train_imagenet write_xla_metrics=True) File "/root/anaconda3/envs/pytorch/lib/python3.6/site-packages/torch_xla/test/test_utils.py", line 72, in write_to_summary metrics = mcu.parse_metrics_report(met.metrics_report()) File "/root/anaconda3/envs/pytorch/lib/python3.6/site-packages/torch_xla/debug/metrics.py", line 47, in metrics_report return torch_xla._XLAC._xla_metrics_report() RuntimeError: tensorflow/compiler/xla/xla_client/xrt_computation_client.cc:1478 : Check failed: session.Run({result}, &outputs) == ::tensorflow::Status::OK() (Invalid argument: The ClusterSpec names the job and task index to be the same names that were provided when the server booted. This is currently not allowed. Job: localservice, task index: 0 vs. OK)`
 

this was on the latest nightly image

zcain117:
I ran the same code on a machine with 1 GPU instead of 4 GPUs and the error did not happen. 1 GPU case trained successfully for 2 epochs and 2 eval loops.

So this error seems specific the the multi-GPU case

zcain117:
I tried on a GCE VM that has 4 GPUs. I am running using the latest `nightly_3.6_cuda` docker image from pytorch/xla

This version succeeds:
```
[...]

PR #4187, opened 2022-11-10
Fix multi-gpu metrics report

ymwangg:
This PR fixed https://github.com/pytorch/xla/issues/2508. I'm not sure why GPU device does not allow reusing the same ClusterSpec but leaving it blank seems to work.

Tested with `GPU_NUM_DEVICES=8 python test_train_mp_mnist.py --metrics_debug --fake_data --num_epochs 1`.

Single-gpu metrics report: https://gist.github.com/ymwangg/609b79ed783a1797f87d55480d23bc8d
Multi-gpu metrics report: https://gist.github.com/ymwangg/8218ddf4ed8ce2d1ac35758effabdf6f

cc @YangFei1990

### DIFF
--- a/third_party/xla_client/xrt_computation_client.cc
+++ b/third_party/xla_client/xrt_computation_client.cc
@@ -1591,7 +1591,13 @@ std::map<std::string, Metric> XrtComputationClient::GetMetrics() const {
     tensorflow::SessionOptions session_options;
     session_options.env = tensorflow::Env::Default();
     session_options.target = worker_target.second;
-    session_options.config = session_cache_->GetConfig();
+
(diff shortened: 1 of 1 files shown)


YangFei1990:
Great, thanks a lot @ymwangg! A few questions for my understanding:
If we do not use the cached session config will it generate some default config? How are we sure that such change won't affect the metrics generation?

ymwangg:
@YangFei1990 My understanding is that the same graph will produce the same results regardless of different session configs. It could be wrong and Google folks may have a better answer.

JackCaoG:
Thanks! I think rebase should fix the build issue.

JackCaoG (approved):
Thanks @ymwangg !

PR #4188, opened 2022-11-10
Create new Cloud Build workflow for nightly images/packages

will-cromar:
- Use [Kaniko](https://github.com/GoogleContainerTools/kaniko) to cache intermediate build results from Docker build.
- Add `_BUILD_ARGS` field for Docker build args so we don't have to have multiple copies of each build arg (one in Dockerfile, one in `substitutions`, and one `--build-arg`). We just need to provide the build args where we want to override the default.
- Tag wheel versions consistent with [`torch` nightly builds](https://download.pytorch.org/whl/nightly/torch/). (Still need support adding `+cpu`, `+cu102`, etc. Will add in a follow-up PR.)
- Use `${PROJECT_ID}` instead of our project name so it's easier to do test runs in our own private test projects.
- Use [Artifact Registry](https://cloud.google.com/artifact-registry/docs/overview) to store images and wheels, which seems to be the recommended replacement for GCR based on the documentation. If this works well, then we can talk about migrating to it for releases. Otherwise we can move these packages into GCR/GCS once we are more confident in them. I'm mainly just trying to keep the experimental build artifacts private/separate at this point. (Although, it looks like repository setup for end-users is [unreasonably tedious](https://cloud.google.com/artifact-registry/docs/python/authentication), so I'm expecting to move back to GCR/GCS for public releases unless there's a better way.)
- Make `.gitignore` paths more specific, because [Cloud Build relies on `.gitignore` paths](https://cloud.google.com/sdk/gcloud/reference/topic/gcloudignore). Remove TF submodule from gitignore because [submodules don't need to be in the `.gitignore`](https://stackoverflow.com/a/7912101). Otherwise, Cloud Build ignores your local TF directory for local builds.

Successful test build: 
image


.gitignore
```diff
@@ -17,12 +17,10 @@ torch_xla/csrc/generated/
 #
 # Below files are not deleted by "setup.py clean".
 
-third_party/tensorflow/
```
JackCaoG, line 20: I think we currently still have patch for `tf`, would those show up if we remove `tf` from ignore list?

docker/experimental/Dockerfile
```diff
@@ -46,7 +48,8 @@ WORKDIR /pytorch
 COPY torch_patches/ torch_patches/
 RUN bash -c "torch_pin=$(cat torch_patches/.torch_pin &); git fetch origin ${torch_pin:-master}; git checkout origin/${torch_pin:-master}"
 
-RUN USE_CUDA=0 python setup.py bdist_wheel
```
JackCaoG, line 49: do we want to set `USE_CUDA=0` as a defualt config? otherwise we might accidently build with `USE_CUDA=1` which is slower.

.gitignore
will-cromar: Nope, just tested this. In both cases, `git status` for pytorch/xla doesn't show the TF changes and `git status` in the TF submodule does.

docker/experimental/Dockerfile
will-cromar: What do you mean? `USE_CUDA=0` is there already. Do you mean this should be configurable? Do we ever want to build the PyTorch GPU wheel ourselves?

JackCaoG: oh I mean this `USE_CUDA=0` would only take effect in this command right? Say someone download this docker and want to do development within this docker, when they do `python setup.py instlall` for pytorch, it will compile with `USE_CUDA=1`. I am wondering if we should do
https://github.com/pytorch/xla/blob/fd4151bc9707b38b24bf10b81a95f5a4e33b196d/docker/Dockerfile#L21

will-cromar: Yeah, that makes sense. Done.

JackCaoG (approved):
Thanks! This is exciting!

docker/experimental/Dockerfile
```diff
@@ -10,6 +10,9 @@ ARG tf_cuda_compute_capabilities="7.0,7.5,8.0"
```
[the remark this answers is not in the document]
will-cromar: Mainly our triggers invoke the build, but we also have to run it manually sometimes to debug it. The build is broken into multiple steps to take better advantage of caching and make debugging faster. Currently, you have to rerun the whole build, and you can't easily run the build with local changes in your repository.

docker/experimental/Dockerfile
```diff
@@ -10,6 +10,9 @@ ARG tf_cuda_compute_capabilities="7.0,7.5,8.0"
 
 ARG tpuvm=1
 ARG build_cpp_tests=0
```
steventk-g, line 12: Are you able to build cpp tests inside of this dockerfile? We're running into issues (the missing MKL library problem, see the failures on https://xl-ml-test.appspot.com/dashboard) with the image we currently use for cpp ops test, and I'm wondering if this can be used instead.

docker/experimental/cloudbuild.yaml
```diff
@@ -0,0 +1,120 @@
...
+  _IMAGE_REPOSITORY: ${_ARTIFACT_REGISTRY_REGION}-docker.pkg.dev/${PROJECT_ID}/torch-xla-images
+  _RELEASE_IMAGE_URL: ${_IMAGE_REPOSITORY}/torch-xla:${_IMAGE_TAG}
+  _ARTIFACTS_IMAGE_URL: ${_IMAGE_REPOSITORY}/artifacts:${_IMAGE_TAG}
+  _CACHE_IMAGE_URL: ${_IMAGE_REPOSITORY}/cache
+  _BAZEL_JOBS: '16'
+  _CACHE: 'true'
+  _CACHE_TTL: '18h'
+  _BUILD_ARGS: tpuvm=1,cuda=0
+options:
+  machineType: E2_HIGHCPU_32
```
steventk-g, line 117: Any reason to choose e2 over n1? https://cloud.google.com/build/docs/api/reference/rest/v1/projects.builds#machinetype

steventk-g (approved):

docker/experimental/Dockerfile
will-cromar: Yes. However, it will increase the image size if we include the CPP tests. Do you know how large they are? If it's not huge, we can include them in the nightly TPU Node image and resolve that test failure.

docker/experimental/cloudbuild.yaml
will-cromar: E2 is cheaper in GCE IIRC, so I defaulted to it here.

steventk-g: Makes sense, thanks!

docker/experimental/Dockerfile
steventk-g: Created an issue to follow up https://github.com/pytorch/xla/issues/4197

docker/experimental/Dockerfile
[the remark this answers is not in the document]
vanbasten23: Got it. Thanks.

PR #4193, opened 2022-11-11
Experimental TPU implementation of DistributedDataParallel

will-cromar:
- Adds experimental implementation of DDP that works on TPU v3 and gives moderately better performance on TPU v4 (1805 ex/sec/replica vs 1710 ex/sec/replica in my experiments).
- Add C++ methods to get PJRT process index and process count (similar to JAX)
- Initialize `torch.distributed` in `pjrt.spawn`, since the master address and process ranks are not predictable on TPU.
- Tweak learning rate and `atol` in DDP tests. The LR was so small that the parameter comparison always passed, even when I had errors in my code.
- Don't broadcast buffers in ResNet50 example to match native XLA distributed implementation.
- Test PJRT with both upstream and experimental DDP.

Tested with PJRT and XRT:

```
XRT_TPU_CONFIG="localservice;0;localhost:51011" TPU_NUM_DEVICES=4 MASTER_ADDR=localhost MASTER_PORT=12355 python pytorch/xla/test/test_ddp.py
PJRT_DEVICE=TPU python pytorch/xla/test/pjrt/test_ddp.py
```

### DIFF
--- a/test/distributed_util.py
+++ b/test/distributed_util.py
@@ -1,11 +1,10 @@
 import copy
+from typing import Optional
 import torch
 import torch.distributed as dist
 import torch.nn as nn
 import torch.optim as optim
-# TODO: fix this
-# from torch.nn.parallel import DistributedDataParallel as DDP
-from torch_xla.experimental.pjrt import DistributedDataParallel as DDP
+import torch.nn.parallel
 import torch_xla.core.xla_model as xm
 import torch_xla.distributed.xla_backend
 
@@ -95,11 +94,12 @@ def train_step(model, inputs, labels, optimizer, loss_fn):
   return loss
 
 
-def ddp_correctness(init_file: str,
+def ddp_correctness(init_file: Optional[str] = None,
                     *,
+                    ddp: type = torch.nn.parallel.DistributedDataParallel,
                     use_large_net: bool = False,
                     debug: bool = False):
-  if not dist.is_initialized():
+  if init_file:
     rank, world_size = init_xla_backend(init_file)
   else:
     rank, world_size = xm.get_ordinal(), xm.xrt_world_size()
@@ -119,7 +119,7 @@ def ddp_correctness(init_file: str,
   # bucket_cap_mb is set to 1 mb such that we can still have multiple all_reduces while avoiding
   # using models that are too larger (25 mb).
   # To be noted, DDP currently uses one bucket for the first iteration. See pytorch#73732.
-  ddp_model = DDP(
+  ddp_model = ddp(
       copy.deepcopy(cpu_model).to(device),
       gradient_as_bucket_view=True,
       bucket_cap_mb=1)
--- a/test/distributed_util.py
+++ b/test/distributed_util.py
@@ -3,7 +3,9 @@ import torch
 import torch.distributed as dist
 import torch.nn as nn
 import torch.optim as optim
-from torch.nn.parallel import DistributedDataParallel as DDP
+# TODO: fix this
+# from torch.nn.parallel import DistributedDataParallel as DDP
+from torch_xla.experimental.pjrt import DistributedDataParallel as DDP
 import torch_xla.core.xla_model as xm
 import torch_xla.distributed.xla_backend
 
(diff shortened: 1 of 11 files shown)


test/distributed_util.py
```diff
@@ -93,11 +90,13 @@ def train_step(model, inputs, labels, optimizer, loss_fn):
...
-def ddp_correctness(init_file: str,
-                    *,
+def ddp_correctness(ddp: type = torch.nn.parallel.DistributedDataParallel,
                     use_large_net: bool = False,
                     debug: bool = False):
-  rank, world_size = init_xla_backend(init_file)
+  if not dist.is_initialized():
+    rank, world_size = init_xla_backend()
+  else:
+    rank, world_size = xm.get_ordinal(), xm.xrt_world_size()
```
JackCaoG, line 99: one customer asked if they can use `torch.distributed` to get the ordinal and world size, do you know if that will work if we init the xla backend?

torch_xla/experimental/pjrt.py
```diff
@@ -179,9 +192,25 @@ def _thread_fn(device: torch.device):
...
+
+    if os.getenv('PJRT_INIT_TORCH_DISTRIBUTED', '0') == '1':
+      if device_type() == 'TPU':
+        # HACK: need to call with each device since it relies on an XLA collective
+        master_ip = next(executor.map(_discover_tpu_master_worker_ip, devices))
+        init_method = f'tcp://{master_ip}:{master_port}'
+      else:
+        init_method = None
+
+      init_pjrt_process_group(init_method=init_method)
```
JackCaoG, line 212: hmm, so user don't need to init the process group if they are using our wrapper?

test/distributed_util.py
JackCaoG: if I understand correctly
1. if user use the upstream ddp, they need to manually `init_process_group`
2. if user use our custom wrapper, they don't.

If this a fair understanding?

will-cromar: Ordinal and world size will be correct in `torch.distributed` on TPU v4 if the process group is initialized with `init_pjrt_process_group`, but not on v2/v3, because we'll have two ordinals per rank on v2/v3.

I added the option to initialize the torch.distributed group through `xmp.spawn` for three reasons, in order of importance:

1. The master IP is not predictable in a TPU slice.
2. If on TPU v2/v3, only one thread can initialize the `torch.distributed` group. It's easier to do it here than tell the user to manage that themselves.
3. It may be convenient for the PJRT process ID to match each process' rank so we have one fewer index to keep track of.

So, with either DDP implementation, the user can initialize the main process group themselves if they solve the three problems above.

The default is to not initialize the process group since we found that TCP process group initialization is flaky on large pod slices.

torch_xla/experimental/pjrt.py
will-cromar: See above.

JackCaoG (approved):
Can we update README and add corresponding tests on xlml? I think we don't have either native ddp nor this new custom ddp e2e test running right now.

test/test_train_mp_imagenet.py
```diff
@@ -193,7 +200,7 @@ def train_imagenet():
   device = xm.xla_device()
   model = get_model_property('model_fn')().to(device)
   if FLAGS.ddp:
-    model = DDP(model, gradient_as_bucket_view=True)
+    model = DDP(model, gradient_as_bucket_view=True, broadcast_buffers=False)
```
alanwaketan, line 203: Why do we disable this option?

torch_xla/experimental/pjrt.py
```diff
@@ -311,3 +340,61 @@ def rendezvous(tag: str, payload: bytes,
...
+
+  @staticmethod
+  def _reduce_grad(grad: torch.Tensor) -> torch.Tensor:
+    """Average gradients across replicas."""
+    return xm.all_reduce(xm.REDUCE_SUM, grad, scale=1. / global_device_count())
+
+  def __init__(self,
+               module: nn.Module,
+               *,
+               broadcast_buffers: bool = True,
```
alanwaketan, line 381: I guess we don't need to implement this logic?

alanwaketan (approved):
LGTM. Thanks, Will. This looks awesome.

torch_xla/experimental/pjrt.py
will-cromar: The `broadcast_buffers` is implemented in `forward`

will-cromar:
Filed bugs to update documentation and add tests.

torch_xla/experimental/pjrt.py
alanwaketan: I guess my confusion is like if the previous optimizer_step approach doesn't demand this, what's the benefits of providing such option here? And then it doesn't seem like you test this option.

PR #4195, opened 2022-11-12
Run cached graph api should grab device lock first

JackCaoG:
We need to grab the device lock before execute the computation in `_run_cached_graph`. This is to prevent more than 1 graph being executed in the same time.

Current implementation we can only grab lock within `tensor.cpp` hence this change.

FYI @shunting314 @wconstab


alanwaketan (approved):
LGTM.

Issue #4169, opened 2022-11-08
[Dynamo] Training Poc issue `infoptr` is missing


JackCaoG:
## 🐛 Bug

for more detail check comments in https://github.com/pytorch/pytorch/pull/88449

I was using
```
TF_CPP_MIN_LOG_LEVEL=0 TF_CPP_VMODULE="pjrt_computation_client=5" python benchmarks/dynamo/torchbench.py --performance --training --use-xla-baseline --randomize-input --backend=torchxla_trace_once --only mobilenet_v2 2>&1 | tee /tmp/dynamo.txt
```
on shunting's branch with some debugging flag, I am able to see
```
device data with handle 93913242061056 does not have info
```

if I search for this handle, it is in the very beginning
```
2022-11-08 01:31:33.305976: I   86150 tensorflow/compiler/xla/service/service.cc:173] XLA service 0x556968434f00 initialized for platform TPU (this does not guarantee that XLA will be used). Devices:                   
2022-11-08 01:31:33.306071: I   86150 tensorflow/compiler/xla/service/service.cc:181]   StreamExecutor device (0): TPU, 2a886c8                                                                                           
2022-11-08 01:31:33.306078: I   86150 tensorflow/compiler/xla/service/service.cc:181]   StreamExecutor device (1): TPU, 2a886c8                                                                                           
2022-11-08 01:31:33.306083: I   86150 tensorflow/compiler/xla/service/service.cc:181]   StreamExecutor device (2): TPU, 2a886c8                                                                                           
2022-11-08 01:31:33.306090: I   86150 tensorflow/compiler/xla/service/service.cc:181]   StreamExecutor device (3): TPU, 2a886c8                                                                                           
2022-11-08 01:31:33.331434: I   86150 ./tensorflow/compiler/xla/xla_client/pjrt_computation_client.h:133] create data with handle 93913242061056 
```

which makes me believe this is the random seed ir.

FYI @shunting314 @wconstab

JackCaoG:
I able to print the ir of the `res` of the model and confirmed that random seed is presented.

```
(Pdb) print(torch_xla._XLAC._get_xla_tensors_text(args_and_out))
IR {
  %0 = f32[96,3,224,224]{3,2,1,0} xla::device_data(), device=TPU:0, ROOT=0
  %1 = f32[1000]{0} xla::device_data(), device=TPU:0
  %2 = f32[1000,1280]{1,0} xla::device_data(), device=TPU:0
[...]

PR #4196, opened 2022-11-12
Add a api to reset and extract RNG Seed

JackCaoG:
This pr is the pytorch/xla side of fix for https://github.com/pytorch/xla/issues/4169.

In this pr we provided a way to check whether a device_data is a IR seed and add an api to get the `device_data` corresponding to the random seed. Remaining part is to add a check in graphInputMatcher https://github.com/pytorch/pytorch/blob/e4a8661ab84022c1bff622c6d2f6e679180b1df5/torch/_dynamo/optimizations/torchxla_integration.py#L36

something similar to 
```
if tensor_id == seed_info_id:
  inp = torch_xla._XLAC._get_rng_seed_as_tensor(str(traced_ivalue.device), True)
```

Whether to `Reset` the IR seed depends on if we do a `mark_step` across different dynamo runs. (mark_step will increment base seed and we want different base seed for every step).

FYI @shunting314 @wconstab

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1513,6 +1513,7 @@ void InitXlaModuleBindings(py::module m) {
             if (infoptr) {
               tensor_ids.push_back(infoptr->tensor_id);
             } else {
+              // TODO(JackCaoG): Make sure this device data is actually seed.
               tensor_ids.push_back(seed_info_id);
             }
             at::Tensor tensor = bridge::AtenFromXlaTensor(
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1527,12 +1527,12 @@ void InitXlaModuleBindings(py::module m) {
 
   m.def("_get_seed_info_id", []() -> int64_t { return seed_info_id; });
 
-  m.def("_reset_and_get_rng_seed",
-        [](const std::string& device_str) -> at::IValue {
+  m.def("_get_rng_seed_as_tensor",
+        [](const std::string& device_str, bool reset) -> at::IValue {
           torch::lazy::BackendDevice device =
               bridge::AtenDeviceToXlaDevice(c10::Device(device_str));
           return bridge::AtenFromXlaTensor(torch_xla::XLATensor::Create(
-              XLATensor::ResetAndGetRngSeedData(device)));
+              XLATensor::GetRngSeedData(device, reset)));
         });
 
   // Return true if value of the tensor requires a computation.
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -58,6 +58,8 @@
 namespace torch_xla {
 namespace {
 
+static int64_t seed_info_id = -127389;
+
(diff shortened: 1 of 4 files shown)


shunting314:
```
if tensor_id == seed_info_id:
  inp = torch_xla._XLAC._get_rng_seed_as_tensor(str(traced_ivalue.device), True)
```

This should be done at inference/training time rather than compiling time, right?

JackCaoG:
> ```
> if tensor_id == seed_info_id:
>   inp = torch_xla._XLAC._get_rng_seed_as_tensor(str(traced_ivalue.device), True)
> ```
> 
> This should be done at inference/training time rather than compiling time, right?

right, we want to use the new seed for every execution.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1498,27 +1500,39 @@ void InitXlaModuleBindings(py::module m) {
...
-
-            // add tensor_id after we make sure the handle does not exist yet.
-            tensor_ids.push_back(infoptr->tensor_id);
+            auto* infoptr =
+                static_cast<torch::lazy::LazyGraphExecutor::DeviceDataInfo*>(
+                    backend_data->info());
+            if (infoptr) {
+              tensor_ids.push_back(infoptr->tensor_id);
+            } else {
+              // TODO(JackCaoG): Make sure this device data is actually seed.
```
shunting314, line 1516: yep, I'm also worried if there would be DeviceData node other than seed that has null InfoPtr

torch_xla/csrc/tensor.cpp
```diff
@@ -367,6 +367,27 @@ class XLATensor::DeviceContextArena {
     devctx->seed_ir_value = torch::lazy::Value();
   }
 
+  torch::lazy::BackendDataPtr GetRngSeedData(
+      const torch::lazy::BackendDevice& device, bool reset) {
```
shunting314, line 371: Can you explain a bit how this 'reset' arg works? In dynamo if we want to replace the seed placeholder with the real seed, should we call this method with reset to be true?

shunting314 (approved):
Thank you for adding this. Adding comments for my understanding.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -58,6 +58,8 @@
 namespace torch_xla {
 namespace {
 
+static int64_t seed_info_id = -127389;
```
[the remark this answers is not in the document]
JackCaoG: nope, I just randomly pick a negative number since regular tensor ID can not be negative. (scalar tensor id would be -1)

torch_xla/csrc/tensor.cpp
JackCaoG: if you pass `reset == false`, and keep calling `GetRngSeedData` it will always return you the same seed data. If you pass `reset==true`, it will give you a new seed.

However `mark_step` will also increment the base seed, and I believe we call `mark_step` for every `optim_mode` now. If that's the case we don't need to `reset` the seed in this API.

torch_xla/csrc/init_python_bindings.cpp
JackCaoG: I will add the check in a following pr.

PR #4204, opened 2022-11-15
Optionally bundle libtpu into TPU VM wheel

will-cromar:
Add build flag `BUNDLE_LIBTPU` to optionally copy `libtpu.so` into the final wheel. If `BUNDLE_LIBTPU=1`, first try to copy the locally installed `libtpu.so`. If `libtpu` is not installed, extract it from the pinned libtpu wheel. 

At run time, we will use the following precedence order:

- `TPU_LIBRARY_PATH` for custom libtpu builds
- Bundled `libtpu.so`
- `libtpu` pip package

This lets us install `torch_xla` side by side with JAX. Note that you cannot use the TPU from both simultaneously. From the same machine:

```
wcromar@t1v-n-bf2f726f-w-0:~$ PJRT_DEVICE=TPU TF_CPP_MIN_LOG_LEVEL=0 python3 -c 'import torch_xla.core.xla_model as xm; print(xm.xla_device())'
INFO:torch_xla:Letting libtpu.so load fail during _XLAC import. libtpu.so will be loaded from `libtpu` Python package when the ComputationClient is created.
2022-11-15 19:15:24.408415: I tensorflow/core/tpu/tpu_initializer_helper.cc:275] Libtpu path is: libtpu.so
INFO:torch_xla:Using bundled libtpu.so (/home/wcromar/.local/lib/python3.8/site-packages/torch_xla/lib/libtpu.so)
2022-11-15 19:15:24.509817: I tensorflow/core/tpu/tpu_initializer_helper.cc:275] Libtpu path is: /home/wcromar/.local/lib/python3.8/site-packages/torch_xla/lib/libtpu.so
[...]
xla:0
wcromar@t1v-n-bf2f726f-w-0:~$ TF_CPP_MIN_LOG_LEVEL=0 python3 -c "import jax; print(jax.local_devices())"
2022-11-15 19:23:55.104045: I external/org_tensorflow/tensorflow/compiler/xla/stream_executor/tpu/tpu_initializer_helper.cc:275] Libtpu path is: /home/wcromar/.local/lib/python3.8/site-packages/libtpu/libtpu.so
[...]
[TpuDevice(id=0, process_index=0, coords=(0,0,0), core_on_chip=0), TpuDevice(id=1, process_index=0, coords=(1,0,0), core_on_chip=0), TpuDevice(id=2, process_index=0, coords=(0,1,0), core_on_chip=0), TpuDevice(id=3, process_index=0, coords=(1,1,0), core_on_chip=0)]
[...]
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,9 @@
 #   XLA_CUDA=0
 #     build the xla/xrt client with CUDA enabled
 #
+#   BUNDLE_LIBTPU=1
+#     include libtpu in final wheel
+#
 
 from __future__ import print_function
 
@@ -45,10 +48,13 @@ import multiprocessing.pool
 import os
 import platform
 import re
+import requests
 import shutil
 import subprocess
 import sys
+import tempfile
 import torch
+import zipfile
 
 base_dir = os.path.dirname(os.path.abspath(__file__))
 third_party_path = os.path.join(base_dir, 'third_party')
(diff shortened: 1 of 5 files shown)


will-cromar:
@JackCaoG should this be the default for TPU VM wheels?

setup.py
```diff
@@ -31,11 +31,15 @@
 #   XLA_CUDA=0
 #     build the xla/xrt client with CUDA enabled
 #
+#   BUNDLE_LIBTPU=1
```
JackCaoG, line 34: current default is 0, so we should put `BUNDLE_LIBTPU=0` here

JackCaoG:
@will-cromar yea I think we should make it default for the TPUVM wheel. As of developer, I think we should never manually download libtpu right? We should expect `setup.py` script download the correct libtpu and bundle it with the wheel. The issue I can see here is if `setup.py` can't find the libtpu installed, it will continually download it as a tmp file everytime we build.

setup.py
will-cromar: Oops, fixed.

setup.py
```diff
@@ -139,6 +146,34 @@ def build_extra_libraries(base_dir, build_mode=None):
     sys.exit(1)
 
 
+def maybe_bundle_libtpu(base_dir):
+  libtpu_path = os.path.join(base_dir, 'torch_xla', 'lib', 'libtpu.so')
```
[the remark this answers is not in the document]
will-cromar: Everything in `lib` is included. See `package_data` at the bottom. `maybe_bundle_libtpu` moves libtpu into `lib` so it's included in the final package.

will-cromar:
Made bundling the default when `TPUVM_MODE` is set in both Docker builds. Developers should install libtpu locally still with `pip install torch_xla[tpuvm]` so they don't download it repeatedly. We should be able to cache it in a smarter way with Bazel in the future.

JackCaoG:
can you update the README section that includes
```
pip3 install torch_xla[tpuvm]
```

With this change, user don't need to do this anymroe with future wheels.

will-cromar:
Done

JackCaoG (approved):
Thanks!

PR #4212, opened 2022-11-17
fix casting SymInt to XLASymNodeImpl

Krovatkin:
fix casting SymInt to XLASymNodeImpl

### DIFF
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -6,7 +6,18 @@ dev = xm.xla_device()
 
 
 class TestDynamicShapes(unittest.TestCase):
-
+  def test_simple_expand(self):
+    size1 = 5
+    size2 = 2
+    t1 = torch.zeros([size1, size2], device=dev)
+    t1[3][0] = 1
+    # t2 has size [<=10, 2]
+    t2 = torch.nonzero(t1)
+    t5 = torch.ones(1, device=dev)
+    t5.expand(t2.size(0))
+    t5_cpu = t5.cpu()
+    self.assertEqual(t5_cpu.shape[0], 1)
+  
   def test_wrap(self):
     a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
     a2 = torch.nonzero(a1)
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -6,6 +6,7 @@ dev = xm.xla_device()
 
 
 class TestDynamicShapes(unittest.TestCase):
+
   def test_simple_expand(self):
     size1 = 5
     size2 = 2
@@ -17,7 +18,7 @@ class TestDynamicShapes(unittest.TestCase):
     t5.expand(t2.size(0))
     t5_cpu = t5.cpu()
     self.assertEqual(t5_cpu.shape[0], 1)
-  
+
   def test_wrap(self):
     a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
     a2 = torch.nonzero(a1)


torch_xla/csrc/torch_util.cpp
```diff
@@ -11,9 +11,17 @@ namespace torch_xla {
 void SymIntElements::AddSymIntNodeElements(c10::SymInt& size) {
   if (size.is_symbolic()) {
     // c10::SymInt --(convert)--> c10::SymIntNode --(cast)-->
-    // lazy::SymIntNodeImpl
+    // PythonSymNodeImpl --> getPyObj() --> XLASymIntNodeImpl
     // --(get)--> lazy::NodePtr --(cast)--> lazy::DimensionNode
     c10::SymNode symbolicIntNode = size.toSymNodeImpl();
+    auto* pySymNode =
+        dynamic_cast<torch::impl::PythonSymNodeImpl*>(symbolicIntNode.get());
+    auto symNode = py::cast<c10::SymNode>(py::handle(pySymNode->getPyObj()));
```
vanbasten23, line 19: Thanks for the fix! I wonder if there is an simple way to avoid doing the multiple cast operations whenever we want a `XLASymNodeImpl` from a `c10::SymNode`.

torch_xla/csrc/torch_util.cpp
```diff
@@ -11,10 +12,13 @@ namespace torch_xla {
...
   if (size.is_symbolic()) {
     // c10::SymInt --(convert)--> c10::SymIntNode --(cast)-->
-    // lazy::SymIntNodeImpl
+    // PythonSymNodeImpl --> getPyObj() --> XLASymIntNodeImpl
     // --(get)--> lazy::NodePtr --(cast)--> lazy::DimensionNode
     c10::SymNode symbolicIntNode = size.toSymNodeImpl();
-    auto* lazySymNode = dynamic_cast<XLASymNodeImpl*>(symbolicIntNode.get());
+    auto* pySymNode =
+        dynamic_cast<torch::impl::PythonSymNodeImpl*>(symbolicIntNode.get());
+    auto symNode = py::cast<c10::SymNode>(py::handle(pySymNode->getPyObj()));
```
vanbasten23, line 20: Would you mind explaining a bit what this part `py::cast<c10::SymNode>(py::handle(pySymNode->getPyObj()))` is doing?

Additionally, since this file is completely in c++, I wonder why there is python object involved.

torch_xla/csrc/torch_util.cpp
Krovatkin: I don't think there is, but we can ask @ezyang to confirm. I believe the invariant Ed wanted is that any `SymNode` that comes from python must be a python object and is wrapped in `PythonSymNodeImpl`. Ed added bindings for `SymNode`, so `XLASymNodeImpl` can turn into a python object when we do `.shape[0]` in python, thanks to pybind.
`PythonSymNodeImpl` in this design is simply a container for that python object.

torch_xla/csrc/torch_util.cpp
Krovatkin: As I mentioned in the comment above, `PythonSymNodeImpl` is a container for python `SymNode`. In case of XLA, that python object is created by pybind when we return a XLASymNodeImpl to python because XLASymNodeImpl extends SymNode and SymNode has python bindings defined here :  https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/python/init.cpp#L1154 

`pySymNode->getPyObj` is a `PyObject*` for some `XLASymNodeImpl`. First we need to cast this `PyObject*` to `SymNode`  and we can do that because pybind11 automatically adds casters for any bound C++ classes : https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/python/init.cpp#L1154 

Now we finally have a c++ SymNode but it's still of the base class rather than the derived type (XLASymNodeImpl) that we need.

so we cast again `SymNode` to  `XLASymNodeImpl` so we can finally access `node()` and get the IR SizeNode out!

JackCaoG:
Hmm, seems like it crashed the `TestExpandSymIntSymbolic`, maybe we should also update the c++ test to reflect the new pySymNode.

Krovatkin:
life is hard :'( linters, tests....

Krovatkin:
I actually need to think how to make C++ tests work :( since we need a python object to wrap into `PythonSymNodeImpl`. Grrrrrrr

JackCaoG:
> I actually need to think how to make C++ tests work :( since we need a python object to wrap into `PythonSymNodeImpl`. Grrrrrrr

if it is really hard we can also get rid of it.. as long as we have python one running.

torch_xla/csrc/torch_util.cpp
vanbasten23: Thanks for explaining.

ezyang:
This is not right. The real problem is the binding code is buggy; it should not unconditionally test for pythonsymintnodeimpl; it should check if it can cast to symintnodeimpl direct via pybind11

Krovatkin:
Okay, I'm trying to fix it on the core side now ( https://github.com/pytorch/pytorch/pull/89295 )  as Ed suggested.

vanbasten23 (approved):

test/test_dynamic_shapes.py
```diff
@@ -7,6 +7,18 @@
...
 
+  def test_simple_expand(self):
+    size1 = 5
+    size2 = 2
+    t1 = torch.zeros([size1, size2], device=dev)
+    t1[3][0] = 1
+    # t2 has size [<=10, 2]
+    t2 = torch.nonzero(t1)
+    t5 = torch.ones(1, device=dev)
+    t5.expand(t2.size(0))
```
JackCaoG, line 18: I felt like we should add a check here to test `t5.shape()`, this can be down in a follow up pr.

JackCaoG (approved):
Thanks Nick!

test/test_dynamic_shapes.py
vanbasten23: I can add a test later.

PR #4215, opened 2022-11-17
Add unsharded computation test for virtual device

steventk-g:
Without sharding, we expect tensors defined on the virtual device to be transferred to the default xla device (e.g. `TPU:0`) for computation execution. When there is sharding involved, the tensors may be sharded explicitly, or implicitly replicated across devices.

### DIFF
--- a/test/test_xla_sharding.py
+++ b/test/test_xla_sharding.py
@@ -175,6 +175,17 @@ class VirtualDeviceTest(XlaShardingTest):
     self.assertIn("VirtualDeviceUsage", met.counter_names())
     self.assertNotEqual(met.counter_value("VirtualDeviceUsage"), 0)
 
+  def test_no_sharding(self):
+    t1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                      dtype=torch.float,
+                      device=xm.xla_device())
+    t2 = torch.tensor([[8, 7, 6, 5, 4, 3, 2, 1]],
+                      dtype=torch.float,
+                      device=xm.xla_device())
+    t3 = t1 + t2
+    t3_expected = [9.0, 9.0, 9.0, 9.0, 9.0, 9.0, 9.0, 9.0]
+    self.assertEqual(t3.tolist()[0], t3_expected)
+
 
 if __name__ == '__main__':
   test = unittest.main()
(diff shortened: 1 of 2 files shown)


test/test_xla_sharding.py
```diff
@@ -175,6 +175,18 @@ def test_model_weight_metrics(self):
     self.assertIn("VirtualDeviceUsage", met.counter_names())
     self.assertNotEqual(met.counter_value("VirtualDeviceUsage"), 0)
 
+  def test_no_sharding(self):
+    partition_spec = (0, 1)
```
JackCaoG, line 179: partition_spec is not being used here?

steventk-g: Removed

test/test_xla_sharding.py
```diff
@@ -175,6 +175,17 @@ def test_model_weight_metrics(self):
     self.assertIn("VirtualDeviceUsage", met.counter_names())
     self.assertNotEqual(met.counter_value("VirtualDeviceUsage"), 0)
 
+  def test_no_sharding(self):
```
JackCaoG, line 178: I need to do some catch up, so when `XLA_USE_SPMD` is set, will `xla_device` return a virtual device? Is this test trying to test operations on real device or virtual device without sharding?

steventk-g: Yeah that flag enables the virtual device.

The goal of this test is to define tensors on the virtual device, and then do an operation without applying sharding, i.e. without transferring data to the TPU. We want to make sure that tensors on the virtual device can be used without being sharded, and that there aren't any data transfer issues.

JackCaoG: make sense, would the computation
```
t3 = t1 + t2
```
being run on all devices but replicated?

steventk-g: Confirmed that the computation happens on device `TPU:0` observing logs in `third_party/xla_client/pjrt_computation_client.cc`, but there is no replication or sharding going on.

```
start no sharding test
2022-12-18 22:41:59.286360: I   81440 tensorflow/compiler/xla/xla_client/pjrt_computation_client.cc:279] Executing PjRt computation on TPU:0
2022-12-18 22:41:59.286452: I   81440 tensorflow/compiler/xla/xla_client/pjrt_computation_client.cc:283] Executing PjRt computation on TPU:0
2022-12-18 22:41:59.287176: I   81440 tensorflow/compiler/xla/xla_client/pjrt_computation_client.cc:326] Returning 1 results
end no sharding test
```

yeounoh: Virtual device abstraction won't trigger SPMD (sharded computation). This is correct.

yeounoh (approved):
LGTM

JackCaoG (approved):

PR #4208, opened 2022-11-17
Replace XLA_COUNTER with TORCH_LAZY_COUNTER

alanwaketan:
Summary:
This pull request replaces most of XLA_COUNTER with the pytorch/xla lib with upstream TORCH_LAZY_COUNTER. However, XLA_COUNTER is still kept to be used in the xla_client lib.

Test Plan:
python test/test_metrics.py MetricsTest.test_metrics_report
and CI.

PR #4217, opened 2022-11-18
[LTC] Replace XLA_FN_COUNTER with TORCH_LAZY_FN_COUNTER

alanwaketan:
Summary:
TORCH_LAZY_FN_COUNTER supports prefix op names with any string. Therefore, we can just reuse it to add xla prefix to xla ops.

This is  paired with pytorch/pytorch#89608.

Test Plan:
CI.


alanwaketan:
To be noted, this PR depends on #4208. Will need a rebase before landing.

scripts/gen_lazy_tensor.py
```diff
@@ -112,7 +112,7 @@ def build_ir_node(self, func: NativeFunction, schema: LazyIrSchema) -> str:
       get_tensorlist="GetTensorList",
       get_tensor_or_wrap_number="bridge::GetXlaTensorOrCreateForWrappedNumber",
       try_get_tensor="bridge::TryGetXlaTensor",
-      metrics_counter='XLA_FN_COUNTER("xla::")',
+      metrics_counter='TORCH_LAZY_FN_COUNTER("xla::")',
```
JackCaoG, line 115: I think `TORCH_LAZY_FN_COUNTER` is default so maybe we don't need to pass it? I am not sure if `xla::` will also be there through.

JackCaoG (approved):
mostly lgtm, thanks!

scripts/gen_lazy_tensor.py
alanwaketan: The default is "lazy::". We need to replace that with "xla::".

PR #4218, opened 2022-11-18
Add python test for sizeAdd node and implement `int_` for XLASymNodeImpl.

vanbasten23:


vanbasten23:
@JackCaoG @Krovatkin @miladm in my current python test in this PR, it couldn't exercise `SizeAdd::Lower` and `SizeAdd::getDynamicValue`. Do you know what are the corresponding python operations that could exercise those 2 functions?

Krovatkin:
try `int(t4)` that should trigger `getDynamicValue()`.

Krovatkin:
I really hope this test should pass even without any core changes :) already

vanbasten23:
No luck. Running the script failed at `int(t4)`:
```
  File "pytorch/xla/test/test_dynamic_shapes.py", line 32, in test_sizeAdd
    int(t4)
  File "/home/ptxla/.local/lib/python3.8/site-packages/torch/__init__.py", line 215, in __int__
    return self.node.int_()
RuntimeError: NYI
```
I think it might be because `XLASymNodeImpl` does overwrite `int_` function.

What about `SizeAdd::Lower`? I was under the impression that either `print(t4.cpu())` or `xm.mark_step()` would trigger `SizeAdd::Lower` but unfortunately the breakpoint there was not hit either.

Krovatkin:
AFAICT, `t4` is a symbolic shape and not a tensor, so `t4.cpu()` should trigger an error if anything :) 
`print(t4)` will trigger `<=X`.

I think I implemented `bool_()` so you could try `bool(t4 == the_right_answer)`

if you use`t4` in `t5 = expand` and then do `t5.cpu()` that would trigger execution as well, but it will error out since we are still fixing SymNode casting on the pytorch core side.
but really we should implement `int_`.

Krovatkin:
cc: @JackCaoG I know you suggested we should test the stuff we already implemented but implementing `int_()` would really make testing much easier for @vanbasten23

JackCaoG:
@Krovatkin sg, let's implement that.

vanbasten23:
Thanks all for the input. Let me implement `int_`.

vanbasten23:
Actually, I can do `self.assertEqual(t4, 3)` to exercise `getDynamicValue()`.

vanbasten23:
For `SizeAdd::isSymbolic` and `SizeAdd::Lower`, we can test indirectly by testing `expand_symint`. Is there a way to test `SizeAdd::ToString` from python's perspective?

test/test_dynamic_shapes.py
```diff
@@ -14,6 +14,18 @@ def test_wrap(self):
...
 
+  def test_sizeAdd(self):
+    size1 = 5
+    size2 = 2
+    t1 = torch.zeros([size1, size2], device=dev)
+    t1[3][0] = 1
+    # t2 has size [<=10, 2]
+    t2 = torch.nonzero(t1)
+    t4 = t2.shape[0] + t2.shape[1]  # Creates a SizeAdd node.
+    dynamic_size = int(t4)  # Exercises SizeAdd::getDynamicValue
```
JackCaoG, line 25: let's move comments above, we don't use comment at the end so it is better to be consistent across whole repo.

JackCaoG (approved):
minor nits, otherwise lgtm.

test/test_dynamic_shapes.py
vanbasten23: Will do. 

For `SizeAdd::isSymbolic` and `SizeAdd::Lower`, we can test indirectly by testing `expand_symint`. Is there a way to test `SizeAdd::ToString` from python's perspective?

JackCaoG: expand a tensor(t5) to t4, and then do 

`print(torch_xla._XLAC._get_xla_tensors_text([t5]))`.

btw don't call it t4, it is not a tensor, but a size. maybe something like `dyn_size` is better

PR #4221, opened 2022-11-18
Fixed two bugs: 1. all TPU cores are used in the program compilation …

xsi-aws:
…in multi-instance case. 2. Invalid TPU mesh coord indices causing the fingerprint to be different in each compilation for multi-instance, resulting in extra memory usage.

### DIFF
--- a/third_party/xla_client/xrt_computation_client.cc
+++ b/third_party/xla_client/xrt_computation_client.cc
@@ -1083,7 +1083,16 @@ std::unique_ptr<xrt::XLAComputation> XrtComputationClient::CreateXrtComputation(
   if (devices.size() > 1) {
     auto device_assignment = config->mutable_device_assignment();
     auto computation_device = device_assignment->add_computation_devices();
-    for (int64_t i = 0; i < devices.size(); ++i) {
+    auto local_device = GetLocalDevices();
+    int64_t local_tpu;
+    for (auto& device : local_device) {
+      if (device.compare(0, 4, "CPU:") == 0 && devices.size() > 1) {
+        continue;
+      }
+      local_tpu = std::stoi(device.substr(4));
+    }
+    int64_t local_ordinal_start = 32 * (local_tpu / 32);
+    for (int64_t i = local_ordinal_start; i < local_ordinal_start + 32; ++i) {
       Device device(devices[i]);
       auto replica_device = computation_device->add_replica_devices();
       if (device.kind == "TPU") {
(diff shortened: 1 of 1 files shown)


third_party/xla_client/xrt_computation_client.cc
```diff
@@ -1083,7 +1083,16 @@ std::unique_ptr<xrt::XLAComputation> XrtComputationClient::CreateXrtComputation(
...
     auto device_assignment = config->mutable_device_assignment();
     auto computation_device = device_assignment->add_computation_devices();
-    for (int64_t i = 0; i < devices.size(); ++i) {
+    auto local_device = GetLocalDevices();
+    int64_t local_tpu;
+    for (auto& device : local_device) {
+      if (device.compare(0, 4, "CPU:") == 0 && devices.size() > 1) {
+        continue;
+      }
+      local_tpu = std::stoi(device.substr(4));
```
hjm-aws, line 1092: break out of loop here.

third_party/xla_client/xrt_computation_client.cc
```diff
@@ -1083,7 +1083,16 @@ std::unique_ptr<xrt::XLAComputation> XrtComputationClient::CreateXrtComputation(
   if (devices.size() > 1) {
     auto device_assignment = config->mutable_device_assignment();
     auto computation_device = device_assignment->add_computation_devices();
-    for (int64_t i = 0; i < devices.size(); ++i) {
+    auto local_device = GetLocalDevices();
+    int64_t local_tpu;
```
hjm-aws, line 1087: Initialize this to some invalid value, like -1. Then do a `XLA_CHECK` after the for loop to ensure it's assigned a valid value.

third_party/xla_client/xrt_computation_client.cc
```diff
@@ -1083,7 +1083,16 @@ std::unique_ptr<xrt::XLAComputation> XrtComputationClient::CreateXrtComputation(
...
+    for (auto& device : local_device) {
+      if (device.compare(0, 4, "CPU:") == 0 && devices.size() > 1) {
+        continue;
+      }
+      local_tpu = std::stoi(device.substr(4));
+    }
+    int64_t local_ordinal_start = 32 * (local_tpu / 32);
```
hjm-aws, line 1094: make it const

third_party/xla_client/xrt_computation_client.cc
```diff
@@ -1405,19 +1414,20 @@ void XrtComputationClient::InitializeDevices(
...
                              topology_proto->num_tpu_devices_per_task() *
-                             topology_proto->mesh_shape_size() +
-                         parsed_device.id * topology_proto->mesh_shape_size();
+                             topology_proto->mesh_shape_size();
+    int64_t base_index = task_index + parsed_device.id * topology_proto->mesh_shape_size();
     std::vector<int> device_mesh_coords(topology_proto->mesh_shape_size());
     for (int i = 0; i < topology_proto->mesh_shape_size(); ++i) {
       device_mesh_coords[i] =
-          topology_proto->device_coordinates(base_index + i);
+          topology_proto->device_coordinates(base_index - task_index + i);
```
hjm-aws, line 1424: Since 
```
base_index = task_index + parsed_device.id * topology_proto->mesh_shape_size()
```

so 
```
base_index - task_index = parsed_device.id * topology_proto->mesh_shape_size()
```

So here it's just 
```
parsed_device.id * topology_proto->mesh_shape_size() + i
```

Then can we simply redefine base_index as 
```
const int64_t base_index = parsed_device.id * topology_proto->mesh_shape_size();
```
and use it here?

third_party/xla_client/xrt_computation_client.cc
```diff
@@ -1405,19 +1414,20 @@ void XrtComputationClient::InitializeDevices(
...
     }
     device_mesh_coords_.insert(
         {dev_target.second, std::move(device_mesh_coords)});
   }
 
+
```
hjm-aws, line 1430: avoid unnecessary whitespace changes

hjm-aws (changes requested):

third_party/xla_client/xrt_computation_client.cc
hjm-aws: Also we can't use 32 here. It at least needs to be an env var because other hardware will not necessarily have 32 devices per node.

PR #4223, opened 2022-11-19
Prevent multi register Lazy key

JackCaoG:
Doing a quick fix for 1.13 since it is less intrsuive. I will do a proper fix using `std::call_once` in master

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -129,4 +129,5 @@ def _init_xla_lazy_backend():
 
 atexit.register(_prepare_to_exit)
 _apply_patches()
-_init_xla_lazy_backend()
+if os.environ.get('INIT_LAZY_BACKEND', '1') != '0':
+  _init_xla_lazy_backend()
--- a/torch_xla/core/xrt_run_server.py
+++ b/torch_xla/core/xrt_run_server.py
@@ -6,6 +6,7 @@ import subprocess
 import sys
 
 from pathlib import Path
+os.environ['INIT_LAZY_BACKEND'] = '0'
 from torch_xla.__init__ import server_is_alive, XRT_RUN_SERVER_PROCESS, XRT_SERVER_REGEX
 
 


will-cromar (approved):

torch_xla/__init__.py
```diff
@@ -129,4 +129,5 @@ def _init_xla_lazy_backend():
 
 atexit.register(_prepare_to_exit)
 _apply_patches()
-_init_xla_lazy_backend()
+if os.environ.get('INIT_LAZY_BACKEND', '1') != '0':
```
yeounoh, line 132: @JackCaoG do you think we could add some comments here explaining why and what this change / flag does for future reference?

JackCaoG: this is a 1.13 only quick hack, I will fix it properly in the master.

torch_xla/core/xrt_run_server.py
```diff
@@ -6,6 +6,7 @@
 import sys
 
 from pathlib import Path
+os.environ['INIT_LAZY_BACKEND'] = '0'
```
yeounoh, line 9: Similar to the earlier comment -- also because I don't know why this wasn't done before and needed now. Maybe something like, `// disable INIT_LAZY_BACKEND in xrt_run_server because ...` could help us, otherwise, this could become a tribal knowledge.

torch_xla/__init__.py
JackCaoG: ok added a quick comment.

yeounoh (approved):
Approve, I added some minor comments asking for some comments/contexts -- but, we can handle this later when you fix it with `execute_once` semantic.

torch_xla/__init__.py
yeounoh: Gotcha, thanks @JackCaoG

torch_xla/core/xrt_run_server.py
JackCaoG: well the right fix is to make sure `_init_lazy_backend`'s corresponding key registration function to only trigger once per process. This fix is not enough, but a quick workaround since it is less intrusive.

yeounoh: > this is a 1.13 only quick hack, I will fix it properly in the master.

ditto

PR #4225, opened 2022-11-19
move all patches to different files and add comments

JackCaoG:
@steventk-g I don't know why we need `tf_patches/stream_executor.diff`. If you can add a comment that would be nice. We want to get rid of these patches eventually, hence we need to remember why they failed and better organize them.


yeounoh (approved):
LGTM thanks!

yeounoh:
> @steventk-g I don't know why we need `tf_patches/stream_executor.diff`. If you can add a comment that would be nice. We want to get rid of these patches eventually, hence we need to remember why they failed and better organize them.

`make_default_target_header_only` is deprecated (unused) -- I will see if we can skip this while updating the pin cc @ManfeiBai @JackCaoG

vanbasten23:
The last line in each of the new patch files doesn't end with a newline. I am not sure if it will cause some problem.

JackCaoG:
I think it is fine, since patches was applied.

steventk-g:
Saw the following error without `tf_patches/stream_executor.diff`
```
ERROR: /workspace/pytorch/xla/third_party/tensorflow/tensorflow/cc/BUILD:775:22: Linking tensorflow/cc/ops/resource_variable_ops_gen_cc failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc @bazel-out/k8-opt/bin/tensorflow/cc/ops/resource_variable_ops_gen_cc-2.params
bazel-out/k8-opt/bin/tensorflow/core/protobuf/_objs/autotuning_proto_cc_impl/autotuning.pb.o:autotuning.pb.cc:function InitDefaultsscc_info_AutotuneResult_FailureResult_tensorflow_2fcore_2fprotobuf_2fautotuning_2eproto(): error: undefined reference to 'stream_executor::dnn::_AlgorithmProto_default_instance_'
```
It looks like keeping this header only means that our build won't link the actual object files at the right time to get the dependencies it needs to build, so we can't use header only for our build.

JackCaoG:
> Saw the following error without `tf_patches/stream_executor.diff`
> 
> ```
> ERROR: /workspace/pytorch/xla/third_party/tensorflow/tensorflow/cc/BUILD:775:22: Linking tensorflow/cc/ops/resource_variable_ops_gen_cc failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc @bazel-out/k8-opt/bin/tensorflow/cc/ops/resource_variable_ops_gen_cc-2.params
> bazel-out/k8-opt/bin/tensorflow/core/protobuf/_objs/autotuning_proto_cc_impl/autotuning.pb.o:autotuning.pb.cc:function InitDefaultsscc_info_AutotuneResult_FailureResult_tensorflow_2fcore_2fprotobuf_2fautotuning_2eproto(): error: undefined reference to 'stream_executor::dnn::_AlgorithmProto_default_instance_'
> ```
> 
> It looks like keeping this header only means that our build won't link the actual object files at the right time to get the dependencies it needs to build, so we can't use header only for our build.

sg, can you add some comment's to the diff?

PR #4228, opened 2022-11-21
XLA Op Key Registration should only happen once.

JackCaoG:
This is a more robust fix compared to 1.13 quick fix in https://github.com/pytorch/xla/pull/4223


wonjoolee95 (approved):
👍

PR #4231, opened 2022-11-21
Disable failing take python op tests on TPU

wonjoolee95:
Python op tests `test_take_xla_int16` and `test_take_xla_float64` have been failing on TPUs, disabling them for now. 

---
These tests fail due to `dtype` mismatch, `'dtype' do not match: torch.int32 != torch.int16` for the `int16` test and `'dtype' do not match: torch.float32 != torch.float64` for the `float64` test. 

For example, for the `int16` test with gdb triggered at the `ops_xla_shape_fn.cpp`, I can see that the `shape` of `input` at `TakeOutputShape` (at `ops_xla_shape_fn.cpp`) is wrongly set to `s32` (it should be `s16` since `test_torch.py` sets `dtype=torch.int16`). The outcome tensors' values are equal, but just their `dtype`s mismatch. The error only happens on TPU. 

@JackCaoG, let me know if you think disabling these would be fine for now. 

---
With these changes, the failing `take` tests get skipped:
```
wonjoo@t1v-n-0dd7892e-w-0:~/pytorch$ python test/test_torch.py TestTorchDeviceTypeXLA.test_take_xla_int16
s
Ran 1 test in 0.003s

OK (skipped=1)

wonjoo@t1v-n-0dd7892e-w-0:~/pytorch$ python test/test_torch.py TestTorchDeviceTypeXLA.test_take_xla_float64
s
Ran 1 test in 0.003s

OK (skipped=1)
wonjoo@t1v-n-0dd7892e-w-0:~/pytorch$ 
```

### DIFF
--- a/test/pytorch_test_base.py
+++ b/test/pytorch_test_base.py
@@ -366,6 +366,8 @@ DISABLED_TORCH_TESTS_TPU_ONLY = {
         'test_diff_xla_float64',  # expected instruction to have shape equal
         'test_nullary_op_mem_overlap_xla',  # core dumped
         'test_index_reduce',  # takes too long
+        'test_take_xla_float64',  # 'dtype' do not match: torch.float32 != torch.float64
+        'test_take_xla_int16',  # 'dtype' do not match: torch.int32 != torch.int16
     },
 
     # test_indexing.py


wonjoolee95:
Debugging this a little more, I can see that for the `int16` test case, the `dtype` for the tensor initially passed to `GetXlaTensorOrCreateForWrappedNumber` at `XLANativeFunctions.cpp` like:
```
torch_xla::XLATensorPtr lazy_index = torch_xla::bridge::GetXlaTensorOrCreateForWrappedNumber(index, *common_device);
```
is actually of `XLAShortType`, which I believe is `int16`. But later when `TakeOutputShape` is called, the `dtype` for the `input` tensor becomes `int32`, so somewhere in between the `dtype` is being incorrectly set.

JackCaoG:
I suspect it is from 
https://github.com/pytorch/xla/blob/b2bd721cf18b496feec02f12313bb532484eaab9/torch_xla/csrc/tensor_util.cpp#L1181-L1190

can you confirm? Pretty much TPU used to can't handle certain types so we cast them implicitly.

wonjoolee95:
> I suspect it is from
> 
> https://github.com/pytorch/xla/blob/b2bd721cf18b496feec02f12313bb532484eaab9/torch_xla/csrc/tensor_util.cpp#L1181-L1190
> 
> can you confirm? Pretty much TPU used to can't handle certain types so we cast them implicitly.

This definitely looks suspicious. But oddly, when I put debugger/print lines in this function, it doesn't seem to get triggered/printed. Also I tried modifying this code to directly return `U16` as such:
```
 case xla::PrimitiveType::U16: 
   return xla::PrimitiveType::U16;
```
but is still failing with the same error.

JackCaoG:
something is off, please open a separate bug to track this. We can disable the test for now, but the goal of test is to catch errors on our end.

JackCaoG (approved):
Thanks!

wonjoolee95:
> something is off, please open a separate bug to track this. We can disable the test for now, but the goal of test is to catch errors on our end.

Makes sense, opened https://github.com/pytorch/xla/issues/4238 and assigned it to myself. Thanks!

PR #4235, opened 2022-11-22
inplace operation within no_grad context for all_gather

aws-rhsoln:
Currently the allgather performs inplace copy which results in errors that look like this: https://discuss.pytorch.org/t/leaf-variable-was-used-in-an-inplace-operation/308

Hence performing the inplace copy within `no_grad` context.


JackCaoG:
@ronghanghu I am not an expert here, but I think cc op under `no_grad` context is fine?

ronghanghu:
Yes, I think it's right that `torch.distributed.all_gather` does not handle back propagation, so it's fine to put it under `no_grad` context (there is a separate version of `torch.distributed.nn.functional.all_gather` that handles back-propagation, although there are issues with it such as https://github.com/pytorch/pytorch/pull/63100).

JackCaoG (approved):
Thanks!

PR #4226, opened 2022-11-21
Companion PR for pytorch/pytorch#89326

nkaretnikov:

PR #4240, opened 2022-11-22
Companion PR for pytorch/pytorch#89326

nkaretnikov:


nkaretnikov:
@JackCaoG I now have push access to the XLA repo, so had to update. This is the same as https://github.com/pytorch/xla/pull/4226 (updated on current master). Apologies for the noise!

JackCaoG (approved):
Thanks! almost forgot about the pin pr has to be created on xla branch directly lol.

PR #4227, opened 2022-11-21
Companion PR for pytorch/pytorch#89327

nkaretnikov:

PR #4242, opened 2022-11-23
Companion PR for pytorch/pytorch#89327

nkaretnikov:


JackCaoG (approved):

nkaretnikov:
This is the same as https://github.com/pytorch/xla/pull/4227, but rebased on current master and pushed directly to xla repo. Had to re-create because I didn't have push access before.

PR #4243, opened 2022-11-23
Fix `non_zero` cast issue

JackCaoG:
This will fix https://github.com/pytorch/xla/issues/4155

TODO: verify if `getDimensionSize` actually return a `s64`, if not I might need to do a cast to `s64`.

FYI @Krovatkin @vanbasten23

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -751,7 +751,9 @@ class TestDynamicShape(XlaTestCase):
 
   def test_nonzero_cast(self):
     t1 = torch.ones(5, 2, device=xm.xla_device())
-    # this cast should not fail
+    # Result of the nonzero should be the index type. Currently
+    # index type is s64 on cpu and gpu, but s32 on TPU. We should be
+    # able to cast it to any other type without error.
     t2 = torch.nonzero(t1.int()).float()
     xm.mark_step()
 
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -749,6 +749,12 @@ class TestDynamicShape(XlaTestCase):
         torch.masked_select(x, mask), 0)
     self.assertEqual(x_dim0_shape.item(), 3)
 
+  def test_nonzero_cast(self):
+    t1 = torch.ones(5, 2, device=xm.xla_device())
+    # this cast should not fail
+    t2 = torch.nonzero(t1.int()).float()
+    xm.mark_step()
+
 
 class TestOptimizationBarrier(XlaTestCase):
 
--- a/torch_xla/csrc/ops/dynamic_ir.cpp
+++ b/torch_xla/csrc/ops/dynamic_ir.cpp
@@ -5,6 +5,7 @@
 #include "torch_xla/csrc/lowering_context.h"
 #include "torch_xla/csrc/ops/infer_output_shape.h"
 #include "torch_xla/csrc/tensor.h"
+#include "torch_xla/csrc/tensor_util.h"
 
 namespace torch_xla {
 
@@ -23,8 +24,10 @@ const std::shared_ptr<torch::lazy::DimensionNode> DimCast(
 
 SizeNode::SizeNode(torch::lazy::Value input, size_t dim)
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::size")},
-              {input}, xla::ShapeUtil::MakeShape(xla::S64, {}), 1,
-              torch::lazy::MHash(dim)),
+              {input},
+              xla::ShapeUtil::MakeShape(
+                  GetShapeDimensionType(/*device=*/nullptr), {}),
+              1, torch::lazy::MHash(dim)),
       dim_(dim) {
   // Not all IR has torch::lazy::shape now, use xla::shape to unblock
   // the development.
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -1876,7 +1876,7 @@ std::pair<XLATensorPtr, XLATensorPtr> XLATensor::nms(
 XLATensorPtr XLATensor::nonzero(const XLATensorPtr& input) {
   torch::lazy::NodePtr node =
       torch::lazy::MakeNode<NonZero>(input->GetIrValue());
-  return input->CreateFrom(torch::lazy::Value(node, 0), at::ScalarType::Long);
+  return input->CreateFrom(torch::lazy::Value(node, 0));
 }
 
 XLATensorPtr XLATensor::norm(const XLATensorPtr& input,
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -1876,7 +1876,9 @@ std::pair<XLATensorPtr, XLATensorPtr> XLATensor::nms(
 XLATensorPtr XLATensor::nonzero(const XLATensorPtr& input) {
   torch::lazy::NodePtr node =
       torch::lazy::MakeNode<NonZero>(input->GetIrValue());
-  return input->CreateFrom(torch::lazy::Value(node, 0));
+  // Nonzero result type should not depend on input type, hence we shouldn't
+  // use input->CreateFrom which will inherit the logical_element_type.
+  return Create(torch::lazy::Value(node, 0), input->GetDevice());
 }
 
 XLATensorPtr XLATensor::norm(const XLATensorPtr& input,
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -5140,7 +5140,7 @@ TEST_F(AtenXlaTensorTest, TestNonzero) {
   ForEachDevice([&](const torch::Device& device) {
     torch::Tensor xla_a = CopyToDevice(a, device);
     torch::Tensor xla_b = torch::nonzero(xla_a);
-    AllClose(b, xla_b);
+    AllClose(b, torch::_cast_Long(xla_b));
 
     if (DebugUtil::ExperimentEnabled("nonzero")) {
       // If the nonzero support is enabled, we must not see any aten:: calls.


JackCaoG:
Hmm, ok. This fix seems incorrect because `setDimensionSize` and `getDimensionSize` take/returns `s32` for TPU. This means TPU expects `index` and `size` to be s32
```
(Pdb) torch_xla._XLAC._get_xla_tensor_dimension_size(t2, 0)
tensor(10, device='xla:0', dtype=torch.int32)
```

this conflicts with the pytorch assumption where `index` and `size` should be s64
```
(Pdb) torch.nonzero(t1.cpu()).dtype
torch.int64
```

While I can cast the `s32` to `s64` on TPU, it doesn't make much sense TPU just can't represent `s64` size and I doubt if we ever encounter a dimension that's larger than s32-max. My next step would be try to force `non_zero` on TPU also returns a logical element type of `s32`

JackCaoG:
confirmed that in HLO, `get-dimension-size` returns a `s32`
```
(Pdb) size = torch_xla._XLAC._get_xla_tensor_dimension_size(t2, 0)
(Pdb) print(torch_xla._XLAC._get_xla_tensors_hlo([size]))
HloModule IrToHlo.4, entry_computation_layout={(s64[<=10,2]{0,1})->(s32[])}

ENTRY %IrToHlo.4 (p0.1: s64[<=10,2]) -> (s32[]) {
  %p0.1 = s64[<=10,2]{0,1} parameter(0)
  %get-dimension-size.2 = s32[] get-dimension-size(s64[<=10,2]{0,1} %p0.1), dimensions={0}
  ROOT %tuple.3 = (s32[]) tuple(s32[] %get-dimension-size.2)
}
```

JackCaoG:
ok now on TPU, `nonzero`'s real xla type and dtype will be s32, I am sure some test will fail on TPU.. Let me give it a try.

vanbasten23:
Quick question: in your investigation, why did you do `(Pdb) torch_xla._XLAC._get_xla_tensor_dimension_size(t2, 0)
` as opposed to `t2.dtype`? Is it because `torch_xla._XLAC._get_xla_tensor_dimension_size` returns the size from XLA layer?

JackCaoG:
dtype was overwritten by `logical_element_type`
https://github.com/pytorch/xla/blob/f8581559272003d8828f15e2207e794e462ad463/torch_xla/csrc/tensor.cpp#L552-L556

test/test_operations.py
```diff
@@ -749,6 +749,12 @@ def test_masked_select_shape(self):
         torch.masked_select(x, mask), 0)
     self.assertEqual(x_dim0_shape.item(), 3)
 
+  def test_nonzero_cast(self):
+    t1 = torch.ones(5, 2, device=xm.xla_device())
+    # this cast should not fail
```
vanbasten23, line 754: mind adding a comment about where the cast happens, such as https://github.com/pytorch/xla/issues/4155#issuecomment-1325562170.?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1876,7 +1876,7 @@ std::pair<XLATensorPtr, XLATensorPtr> XLATensor::nms(
 XLATensorPtr XLATensor::nonzero(const XLATensorPtr& input) {
   torch::lazy::NodePtr node =
       torch::lazy::MakeNode<NonZero>(input->GetIrValue());
-  return input->CreateFrom(torch::lazy::Value(node, 0), at::ScalarType::Long);
```
vanbasten23, line 1879: I wonder why we removed the `at::ScalarType::Long`?

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -23,8 +24,10 @@ const std::shared_ptr<torch::lazy::DimensionNode> DimCast(
 
 SizeNode::SizeNode(torch::lazy::Value input, size_t dim)
```
vanbasten23, line 25: i see. So the SizeNode is created when we do `.float()`.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1876,7 +1876,7 @@ std::pair<XLATensorPtr, XLATensorPtr> XLATensor::nms(
...
+  return input->CreateFrom(torch::lazy::Value(node, 0));
```
vanbasten23, line 1879: I assume what this `CreateFrom` does is: `input->GetIrValue()`(line1878) contains the IR graph before we add the `NonZero` node; then `input->CreateFrom` insert the `NonZero` node to the existing IR graph. Is my understanding correct?

torch_xla/csrc/tensor_methods.cpp
JackCaoG: this way we don't overwrite the dtype of the result of the nonzero

vanbasten23 (approved):

torch_xla/csrc/tensor_methods.cpp
JackCaoG: `CreateFrom` will create a new XLATensor using the IR node provided but it will use the same logical_element_type and device from the input tensor. I used `Create` here instead.

JackCaoG:
@vanbasten23 Can you test if dynamic input liner model will works on PJRT + TPU? @will-cromar had a bug open for supporting dynamic shape + PJRT so we want to figure out the gap here.

vanbasten23:
> @vanbasten23 Can you test if dynamic input liner model will works on PJRT + TPU? @will-cromar had a bug open for supporting dynamic shape + PJRT so we want to figure out the gap here.

How can I test it on PJRT+TPU? Is it by setting `PJRT_DEVICE=TPU` and run the model on my TPU VM?

will-cromar:
Right. All you need to do is set `PJRT_DEVICE=TPU` on TPU VM https://github.com/pytorch/xla/blob/master/docs/pjrt.md

JackCaoG:
I am able to repo the GPU failure with
```
XLA_EXPERIMENTAL="nonzero:masked_select" ./build/test_ptxla --gtest_filter=AtenXlaTensorTest.TestNonzero
```
and the error is expected
```
Different shape:
long [3, 2]
-vs-
int [3, 2]
```
I guess I will cast both to s64 before doing the `allClose`

JackCaoG:
adding a `_cast_Long` seems to resolve the error
```
AllClose(b, torch::_cast_Long(xla_b));
```
I will post a fix.

vanbasten23:
> > @vanbasten23 Can you test if dynamic input liner model will works on PJRT + TPU? @will-cromar had a bug open for supporting dynamic shape + PJRT so we want to figure out the gap here.
> 
> How can I test it on PJRT+TPU? Is it by setting `PJRT_DEVICE=TPU` and run the model on my TPU VM?

I tested the forward pass model on PJRT+TPU and it works fine.

The model returns a different loss each time I run. But this problem exists for both XRT and PJRT, so I think it might be an issue with the [model](https://github.com/pytorch/xla/pull/4114/files) itself.

JackCaoG:
hmm @vanbasten23 if we fixed the input we should see the same result every time. Can you look into why that's not the case? We need to confirm that `dynamic_shape` on xla return the same result as on CPU.

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -23,8 +24,10 @@ const std::shared_ptr<torch::lazy::DimensionNode> DimCast(
...
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::size")},
-              {input}, xla::ShapeUtil::MakeShape(xla::S64, {}), 1,
-              torch::lazy::MHash(dim)),
+              {input},
+              xla::ShapeUtil::MakeShape(
```
vanbasten23, line 28: Just want to confirm that the change in this file is not directly related to the cast failure. And it is more of making the output type correct by making it depend on the input device type, right?

JackCaoG: yes, that's correct.

vanbasten23:
> hmm @vanbasten23 if we fixed the input we should see the same result every time. Can you look into why that's not the case? We need to confirm that `dynamic_shape` on xla return the same result as on CPU.

Yeah, I'm already looking into it. Will let you know if I find anything.

vanbasten23:
Not directly related to this pr but do you know why `NonZero` has 2 outputs: https://github.com/pytorch/xla/blob/d668f2caf43c799a8165c541d3c7d4196489d850/torch_xla/csrc/ops/nonzero.cpp#L27?

JackCaoG:
> Not directly related to this pr but do you know why `NonZero` has 2 outputs:
> 
> https://github.com/pytorch/xla/blob/d668f2caf43c799a8165c541d3c7d4196489d850/torch_xla/csrc/ops/nonzero.cpp#L27
> 
> ?

if you look at the lowering, it returns 2 XLAOP. However second XLAOP was discarded. I am not sure why that's the case.

vanbasten23:
> > hmm @vanbasten23 if we fixed the input we should see the same result every time. Can you look into why that's not the case? We need to confirm that `dynamic_shape` on xla return the same result as on CPU.
> 
> Yeah, I'm already looking into it. Will let you know if I find anything.

I figured out why the loss varies each time. By default, the weights in each layer has some randomization. So to make the loss deterministic and make it easy to write test, I need to initialize the weights explicitly by myself. I've updated the model [here](https://github.com/pytorch/xla/pull/4114/files).

vanbasten23:
@JackCaoG in the first comment of this pr, you mentioned `TODO: verify if getDimensionSize actually return a s64, if not I might need to do a cast to s64`. I wonder if this TODO is still valid.

I ran the test in my [pr](https://github.com/pytorch/xla/pull/4256) and dump the HLO graph:
```
%set-dimension-size.39 = s32[<=10,2]{1,0} set-dimension-size(s32[10,2]{1,0} %concatenate.38, s32[] %reduce.17), dimensions={0}
..
%get-dimension-size.130 = s32[] get-dimension-size(f32[<=10]{0} %multiply.123), dimensions={0}
```
It looks getDimensionSize returns s32.

JackCaoG:
yea, I decided to make the SizeNode to be s32 on TPU instead, so we are honest about the real underlying type and avoid additional casting.

PR #4244, opened 2022-11-24
Try to make op tests runs on TPUVM

JackCaoG:
We really need to make python op test green on TPU since dynamic shape can not be easily tested on CPU. Here is what I think we need to do 
1. Create the TPUVM version of xlml test, current one is on 2vm
2. Make python op to pass on TPUVM
3. investigate some of these weird crash that does not happen on 2vm
4. Move op test to use PJRT on TPU

I tried to disable all of the tests that failed but I didn't have time to finish all of them. @wonjoolee95 can you maybe onboard @ManfeiBai on op test and how to debug them then finish the goal I listed above? Happy to schedule a meeting to go over the steps.

### DIFF
--- a/test/pytorch_test_base.py
+++ b/test/pytorch_test_base.py
@@ -307,6 +307,7 @@ DISABLED_TORCH_TESTS_TPU_ONLY = {
     # test_torch.py
     'TestDevicePrecisionXLA': {
         'test_digamma',  # Precision issue at the first assert, then NAN handling (both on TPU)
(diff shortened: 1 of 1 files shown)


JackCaoG:
@wonjoolee95 I will merge this pr for now, you guys can build on top of this one.

wonjoolee95:
Thanks. FYI @ManfeiBai and I will be tracking this work at https://docs.google.com/document/d/1OzGNibtIclMRrpLGhK9z-Gp4HGvaye53jT6XjI7-EbY/edit#.

Issue #4234, opened 2022-11-22
Behavior of `F.linear` differs between vanilla Pytorch and XLA/CPU


pscollins:
## 🐛 Bug

I hit a `RuntimeError` when trying to port an [existing model](https://github.com/facebookresearch/unbiased-teacher-v2) to XLA. I'm running against a locally-built copy of pytorch/XLA using XLA-CPU.

## To Reproduce

```
In [1]: import torch

In [2]: import torch.nn.functional as F

In [3]: F.linear(torch.ones([54400, 17]), torch.ones([17]))
Out[3]: tensor([17., 17., 17.,  ..., 17., 17., 17.])

In [4]: import torch_xla

In [5]: import torch_xla.core.xla_model as xm

In [6]: device = xm.xla_device()

In [7]: F.linear(torch.ones([54400, 17], device=device), torch.ones([17], device=device))
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In [7], line 1
----> 1 F.linear(torch.ones([54400, 17], device=device), torch.ones([17], device=device))

RuntimeError: Value out of range (expected to be in range of [-1, 0], but got 1)
```

See snippet above

## Expected behavior

Behavior should match between vanilla pytorch and pytorch/XLA (either throw an exception in both cases or don't throw an exception in both cases).

## Environment

 - Reproducible on XLA backend [CPU/TPU]: CPU, did not try TPU
 - torch_xla version: I built torch_xla from HEAD using the "build from source" workflow described in CONTRIBUTING.md around 11/20, `torch_xla/version.py` shows:
```
# Autogenerated file, do not edit!
__version__ = '1.14'
__xla_gitrev__ = 'f790bc8ac411a8e6903b89adf7610b812996537b'
__torch_gitrev__ = '7b0d577c226fae78f377b26feab4122c4203ad59'
```

JackCaoG:
@Liyang90 Do you have cycle to take a look at this one?

Liyang90:
Hi,

I can confirm the issue.
According to [torch.nn.functional.linear](https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html)
The second argument `weight` can be 2D or 1D: "Weight: (out\_features, in\_features) or (in\_features)", but in PyTorch/XLA, the `weight` can only be 2D. If a 1D tensor is passed in, this error will happen.

@JackCaoG Should we update the implementation to match PyTorch's `torch.nn.functional.linear` behavior?

[...]

PR #4246, opened 2022-11-24
Fix 1d transpose

JackCaoG:
This is to fix https://github.com/pytorch/xla/issues/4234.

@wonjoolee95 let's remove the view part in transpose after functionization, you can just do
```
return input->CreateFrom(ir_value);
```
instead.

FYI @pscollins  @Liyang90

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1864,8 +1864,14 @@ class TestAtenXlaTensor(XlaTestCase):
   def test_transpose_1d(self):
 
     def test_fn(t1):
-      t1.t_()
       return t1.t()
+
+    self.runAtenTest([torch.arange(15, dtype=torch.int32)], test_fn)
+
+  def test_transpose_1d_inplace(self):
+
+    def test_fn(t1):
+      return t1.t_()
 
     self.runAtenTest([torch.arange(15, dtype=torch.int32)], test_fn)
 
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1861,6 +1861,14 @@ class TestAtenXlaTensor(XlaTestCase):
     emb_out = emb(index)
     assert emb_out.dtype == torch.bfloat16
 
+  def test_transpose_1d(self):
+
+    def test_fn(t1):
+      t1.t_()
+      return t1.t()
+
+    self.runAtenTest([torch.arange(15, dtype=torch.int32)], test_fn)
+
 
 class MNISTComparator(nn.Module):
 
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -2528,14 +2528,26 @@ XLATensorPtr XLATensor::trace(const XLATensorPtr& input) {
 XLATensorPtr XLATensor::transpose(const XLATensorPtr& input, int64_t dim0,
                                   int64_t dim1) {
   auto input_shape = input->shape();
-  auto permute_dims = torch::lazy::MakeTransposePermutation(
-      /*dim0=*/dim0, /*dim1=*/dim1, /*rank=*/input_shape.get().rank());
-  ViewInfo view_info(ViewInfo::Type::kPermute, input_shape, permute_dims);
+  ViewInfo view_info;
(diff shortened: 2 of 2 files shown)


test/test_operations.py
```diff
@@ -1861,6 +1861,14 @@ def test_emb_bf16(self):
     emb_out = emb(index)
     assert emb_out.dtype == torch.bfloat16
 
+  def test_transpose_1d(self):
+
+    def test_fn(t1):
+      t1.t_()
```
alanwaketan, line 1867: Is it better to test `t1.t_()` and `t1.t()` in two separate tests?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2528,14 +2528,26 @@ XLATensorPtr XLATensor::trace(const XLATensorPtr& input) {
 XLATensorPtr XLATensor::transpose(const XLATensorPtr& input, int64_t dim0,
                                   int64_t dim1) {
   auto input_shape = input->shape();
-  auto permute_dims = torch::lazy::MakeTransposePermutation(
-      /*dim0=*/dim0, /*dim1=*/dim1, /*rank=*/input_shape.get().rank());
-  ViewInfo view_info(ViewInfo::Type::kPermute, input_shape, permute_dims);
+  ViewInfo view_info;
+  if (input_shape.get().rank() <= 1) {
+    // return a view of self if input rank <=1
+    torch::lazy::Value ir_value = input->GetIrValue();
```
alanwaketan, line 2534: Just curious what's the benefits of not doing it within `MakeTransposePermutation`?

alanwaketan (commented):
Generally, LGTM. Some comments.

test/test_operations.py
JackCaoG: yea, I will separate tests.

torch_xla/csrc/tensor_methods.cpp
JackCaoG: I am not sure if `kPermute` can handle two same dimension. We are in the middle of the functionization refactoring so I don't want to touch the view code too much hence this quick fix. We can revisit after refactoring is done. @wonjoolee95 fyi.

alanwaketan (approved):
LGTM.

wonjoolee95 (approved):
Thanks for the heads up.

test/test_operations.py
```diff
@@ -1861,6 +1861,20 @@ def test_emb_bf16(self):
     emb_out = emb(index)
     assert emb_out.dtype == torch.bfloat16
 
+  def test_transpose_1d(self):
+
+    def test_fn(t1):
+      return t1.t()
```
vanbasten23, line 1867: Is it true that the non-in-place operation will create a copy of tensor and transpose it?

vanbasten23 (approved):

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2528,14 +2528,26 @@ XLATensorPtr XLATensor::trace(const XLATensorPtr& input) {
 XLATensorPtr XLATensor::transpose(const XLATensorPtr& input, int64_t dim0,
                                   int64_t dim1) {
   auto input_shape = input->shape();
-  auto permute_dims = torch::lazy::MakeTransposePermutation(
-      /*dim0=*/dim0, /*dim1=*/dim1, /*rank=*/input_shape.get().rank());
-  ViewInfo view_info(ViewInfo::Type::kPermute, input_shape, permute_dims);
+  ViewInfo view_info;
+  if (input_shape.get().rank() <= 1) {
+    // return a view of self if input rank <=1
```
vanbasten23, line 2533: Since we want to make it a noop if the input tensor 1d, I wonder why we can't just return `input` here.

vanbasten23: oh I see. It's not a in-place operation so we need to create a copy of the tensor, such as the ViewInfo, right?

test/test_operations.py
JackCaoG: yes

PR #4255, opened 2022-11-30
TF pin for 11/14/2022

ManfeiBai:
Tensorflow pin update to
```
commit 8b2e0ca8e3fed42abfaa7baeabe25fc8dfd1c2ab (HEAD)
Author: Rahul Batra <rahbatra@amd.com>
Date:   Mon Nov 14 15:22:48 2022 +0000

    [ROCm]: Updated CI files to ROCm5.3
```

### DIFF
--- a/tf_patches/f16_abi_clang.diff
+++ b/tf_patches/f16_abi_clang.diff
@@ -1,12 +1,11 @@
-upstream CI will fail without this
 diff --git a/tensorflow/compiler/xla/service/cpu/runtime_fp16.h b/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
 index 9fe020d5937..32774c2f3c0 100644
 --- a/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
 +++ b/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
 @@ -18,12 +18,7 @@ limitations under the License.
- 
+
  #include <stdint.h>
- 
+
 -// _Float16 always gets us the correct ABI type, so use that if available.
 -// AArch64 GCC defines __FLT16_MANT_DIG__ even when _Float16 is not available.
 -#if defined(__FLT16_MANT_DIG__) && \
@@ -17,3 +16,5 @@ index 9fe020d5937..32774c2f3c0 100644
  // Older versions of Clang don't have _Float16. Since both float and _Float16
  // are passed in the same register we can use the wider type and careful casting
  // to conform to x86_64 psABI. This only works with the assumption that we're
+
+
--- a/tf_patches/f16_abi_clang.diff
+++ b/tf_patches/f16_abi_clang.diff
@@ -1,3 +1,4 @@
+upstream CI will fail without this
 diff --git a/tensorflow/compiler/xla/service/cpu/runtime_fp16.h b/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
 index 9fe020d5937..32774c2f3c0 100644
 --- a/tensorflow/compiler/xla/service/cpu/runtime_fp16.h
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -60,7 +60,7 @@ PjRtComputationClient::PjRtComputationClient() {
     client_ = xla::GetTpuClient(max_inflight_computations).value();
   } else if (device_type == "TPU_C_API") {
     TF_VLOG(1) << "Initializing PjRt C API client...";
-    client_ = std::move(xla::GetCApiClient(device_type).value());
+    client_ = std::move(xla::GetCApiClient("TPU").value());
     // TODO(wcromar): remove this when C API supports
     // kImmutableUntilTransferCompletes
     host_buffer_semantics_ = xla::PjRtClient::HostBufferSemantics::kZeroCopy;
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -60,7 +60,7 @@ PjRtComputationClient::PjRtComputationClient() {
     client_ = xla::GetTpuClient(max_inflight_computations).value();
   } else if (device_type == "TPU_C_API") {
     TF_VLOG(1) << "Initializing PjRt C API client...";
-    client_ = std::move(xla::GetCApiClient().value());
+    client_ = std::move(xla::GetCApiClient(device_type).value());
     // TODO(wcromar): remove this when C API supports
     // kImmutableUntilTransferCompletes
     host_buffer_semantics_ = xla::PjRtClient::HostBufferSemantics::kZeroCopy;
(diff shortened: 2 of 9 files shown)


torch_xla/csrc/layout_manager.cpp
```diff
@@ -120,7 +120,7 @@ xla::Shape MakeShapeWithSortedLayout(absl::Span<const int64_t> dimensions,
   std::sort(layout.begin(), layout.end(), [&](int64_t a, int64_t b) {
     return dimensions[a] > dimensions[b];
   });
-  return xla::ShapeUtil::MakeShapeWithLayout(type, dimensions, layout);
+  return xla::ShapeUtil::MakeShapeWithDenseLayout(type, dimensions, layout);
```
yeounoh, line 123: This is correct in our case.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -60,7 +60,7 @@ PjRtComputationClient::PjRtComputationClient() {
     client_ = xla::GetTpuClient(max_inflight_computations).value();
   } else if (device_type == "TPU_C_API") {
     TF_VLOG(1) << "Initializing PjRt C API client...";
-    client_ = std::move(xla::GetCApiClient().value());
+    client_ = std::move(xla::GetCApiClient(device_type).value());
```
yeounoh, line 63: I think the canonical device type should be "TPU" instead of "TPU_C_API"

yeounoh (changes requested):
Mostly LGTM -- thanks and great work @ManfeiBai . Left some comments for you to take a look. We also need to update setup.py for libtpu ver. Once we can do a quick training test on a TPUVM we can merge this :)

third_party/xla_client/pjrt_computation_client.cc
JackCaoG: `TPU_C_API` is a special mode to test the new CAPI bridge.

tf_patches/f16_abi_clang.diff
```diff
@@ -1,12 +1,11 @@
-upstream CI will fail without this
```
JackCaoG, line 1: can we restore this line.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: `device_type` check should be against "TPU_C_API", but the call to `xla::GetCApiClient()` I think we should pass "TPU".

JackCaoG:
we need to find a way to drop all these diffs, maybe a good topic for next week's fixit.

yeounoh:
> we need to find a way to drop all these diffs, maybe a good topic for next week's fixit.

Yea, it would be nice if our build works straight with the TF without us making any local changes. Having said that, I reviewed the diffs with Manfei, they are all needed this time. But, I share the same concern with you and +1.

torch_xla/csrc/layout_manager.cpp
ManfeiBai: Thanks, Yeounoh!

JackCaoG:
> > we need to find a way to drop all these diffs, maybe a good topic for next week's fixit.
> 
> Yea, it would be nice if our build works straight with the TF without us making any local changes. Having said that, I reviewed the diffs with Manfei, they are all needed this time. But, I share the same concern with you and +1.

I think we need to resolve this from the google3 tf side, basically to upstream some of the diffs.

third_party/xla_client/pjrt_computation_client.cc
ManfeiBai: Thanks, updated to "TPU"

tf_patches/f16_abi_clang.diff
ManfeiBai: Thanks, updated ~

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Yea, for a future reference, tpu initializer would do something like,
```
stream_executor::tpu::SetPjrtApi("TPU", GetTpuPjrtApi()).IgnoreError();
```
and the "TPU" is used as a key (lower-cased) to map the C API to the device type. cc @will-cromar @JackCaoG

ManfeiBai: should we lower-case "TPU" right now or keep "TPU" and lower-cased by function?

yeounoh: It's done by the function.

ManfeiBai: Thanks, Yeounoh!

ManfeiBai:
I ran `git reset --hard 59e77b15649dbf54d6eb26de44f4da0185431033` in TPUVM to pull the newest modify in this pr, but the newest tf commit have not been pulled automatically in TPUVM; However the tensorflow newest commit has been pushed, and is visiable under "File Changed" tab now;
How do we make sure the final "third_party/tensorflow" is the newest commit in github?

JackCaoG:
your brnach's tf is pin to ~14 days ago https://github.com/tensorflow/tensorflow/tree/8b2e0ca8e3fed42abfaa7baeabe25fc8dfd1c2ab.

If that's where you want to pin, you just need to do a `git submodule sync` and `git submodule update --init --recursive` on your TPUVM. @yeounoh Are we doing another one for Dec? If not maybe it is a good idea to sync to Dec 1's head since next update will be Jan.

miladm (approved):
LGTM after the conflict resolution and retest of the CI.

Did you verify the upstream pytorch tests pass with this branch?

yeounoh:
> your brnach's tf is pin to ~14 days ago https://github.com/tensorflow/tensorflow/tree/8b2e0ca8e3fed42abfaa7baeabe25fc8dfd1c2ab. (by clicking https://github.com/ManfeiBai/PyTorchXLA/tree/fortfpin11222022/third_party)
> 
> If that's where you want to pin, you just need to do a `git submodule sync` and `git submodule update --init --recursive` on your TPUVM. @yeounoh Are we doing another one for Dec? If not maybe it is a good idea to sync to Dec 1's head since next update will be Jan.

Yea, @ManfeiBai is repurposing this with a Dec commit (from 12/2/22). @miladm yea, we will test this on TPUVM & upstream. This is still WIP.

miladm:
FWIW, there was a recent dynamic shape refactor in tensorflow repo that may influence our DS initiative. Wondering if we should squeeze that in - if not too late. If so, happy to share more context. Wdyt?

yeounoh:
> FWIW, there was a recent dynamic shape refactor in tensorflow repo that may influence our DS initiative. Wondering if we should squeeze that in - if not too late. If so, happy to share more context. Wdyt?

Yea, let me know of the date, @miladm I will work with @ManfeiBai to squeeze that in.

ManfeiBai:
Thanks for the comments, @yeounoh , @miladm , @JackCaoG ,
The new TF pin update with the Dec commit (from 12/2/22) is in https://github.com/pytorch/xla/pull/4324,
We close this PR because we will skip 11/14/22 and update TF pin to 12/2/22 directly, and welcome to continue our discuss in #4324 too!

PR #4258, opened 2022-12-01
Add more tests for sizeAdd node.

vanbasten23:
- Test SizeAdd::Lower function.
- Fix an issue with another existing test "test_simple_expand".

### DIFF
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -33,13 +33,16 @@ class TestDynamicShapes(unittest.TestCase):
     t1[3][0] = 1
     # t2 has size [<=10, 2]
     t2 = torch.nonzero(t1)
-    # Exercise SizeAdd::getStaticValue
+    # Create a SizeAdd IR node.
     dyn_size = t2.shape[0] + t2.shape[1]
-    # Exercises SizeAdd::getDynamicValue
+    # Exercises SizeAdd::getDynamicValue.
     dynamic_size = int(dyn_size)
     self.assertEqual(dynamic_size, 3)
-    # Exercise SizeAdd::getStaticValue
+    # Exercise SizeAdd::getStaticValue.
     self.assertEqual(str(dyn_size), '<=12')
+    t3 = torch.ones(1, device=dev)
+    t4 = t3.expand(dyn_size)
+    self.assertEqual(t4.size(0), 3)
 
 
 if __name__ == '__main__':
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -12,12 +12,13 @@ class TestDynamicShapes(unittest.TestCase):
     size2 = 2
     t1 = torch.zeros([size1, size2], device=dev)
     t1[3][0] = 1
+    t1[3][1] = 1
     # t2 has size [<=10, 2]
     t2 = torch.nonzero(t1)
     t5 = torch.ones(1, device=dev)
-    t5.expand(t2.size(0))
-    t5_cpu = t5.cpu()
-    self.assertEqual(t5_cpu.shape[0], 1)
+    t6 = t5.expand(t2.size(0))
+    t6_cpu = t6.cpu()
+    self.assertEqual(t6_cpu.shape[0], 2)
 
   def test_wrap(self):
     a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
@@ -34,6 +35,7 @@ class TestDynamicShapes(unittest.TestCase):
     # t2 has size [<=10, 2]
     t2 = torch.nonzero(t1)
     # Create a SizeAdd IR node.
+    # t2.shape[1] generates a SizeConstant node.
     dyn_size = t2.shape[0] + t2.shape[1]
     # Exercises SizeAdd::getDynamicValue.
     dynamic_size = int(dyn_size)
@@ -41,8 +43,10 @@ class TestDynamicShapes(unittest.TestCase):
     # Exercise SizeAdd::getStaticValue.
     self.assertEqual(str(dyn_size), '<=12')
     t3 = torch.ones(1, device=dev)
+    # Exercise SizeAdd::Lower.
     t4 = t3.expand(dyn_size)
     self.assertEqual(t4.size(0), 3)
+    print(torch_xla._XLAC._get_xla_tensors_text([t4]))
 
 
 if __name__ == '__main__':
--- a/torch_xla/csrc/ops/dynamic_ir.cpp
+++ b/torch_xla/csrc/ops/dynamic_ir.cpp
@@ -86,6 +86,7 @@ int64_t SizeAdd::getDynamicValue() const {
 std::string SizeAdd::ToString() const { return "SizeAdd"; }
 
 XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
+  std::cerr << "xw32, file=" << __FILE__ << ", line=" << __LINE__ << "function=" << __FUNCTION__ << ": " << std::endl;
   auto input1 = loctx->GetOutputOp(operand(0));
   auto input2 = loctx->GetOutputOp(operand(1));
   return ReturnOp((input1 + input2), loctx);
--- a/torch_xla/csrc/ops/dynamic_ir.cpp
+++ b/torch_xla/csrc/ops/dynamic_ir.cpp
@@ -60,11 +60,12 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
   return ReturnOp(xla::GetDimensionSize(input, this->dim_), loctx);
 }
 
-std::string SizeNode::ToString() const { return "SizeNode"; }
+std::string SizeNode::ToString() const { return "SizeNode for op aten::size"; }
 
 SizeAdd::SizeAdd(torch::lazy::Value a, torch::lazy::Value b)
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::add")},
               {a, b}, xla::ShapeUtil::MakeShape(xla::S64, {}), 1) {
+  std::cout << "xw32, file=" << __FILE__ << ", line=" << __LINE__ << "function=" << __FUNCTION__ << ": " << std::endl;
   const torch::lazy::DimensionNode* dim_node_0 = DimCast(operand(0));
   const torch::lazy::DimensionNode* dim_node_1 = DimCast(operand(1));
   // SizeAdd can only be perfomed between two DimensionNode
@@ -83,7 +84,7 @@ int64_t SizeAdd::getDynamicValue() const {
   return dim_node_0->getDynamicValue() + dim_node_1->getDynamicValue();
 }
 
-std::string SizeAdd::ToString() const { return "SizeAdd"; }
+std::string SizeAdd::ToString() const { return "SizeAdd for op aten::add"; }
 
 XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
   std::cerr << "xw32, file=" << __FILE__ << ", line=" << __LINE__ << "function=" << __FUNCTION__ << ": " << std::endl;
@@ -111,7 +112,7 @@ int64_t SizeEq::getDynamicValue() const {
 
 std::string SizeEq::ToString() const { return "SizeEq"; }
 
-SizeConstant::SizeConstant(int64_t val) : Scalar(c10::Scalar{val}, xla::S64){};
+SizeConstant::SizeConstant(int64_t val) : Scalar(c10::Scalar{val}, xla::ShapeUtil::MakeShape(GetShapeDimensionType(/*device=*/nullptr), {})){};
 
 SizeMul::SizeMul(torch::lazy::Value a, torch::lazy::Value b)
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::mul")},
@@ -174,5 +175,9 @@ XlaOpVector SizeDiv::Lower(LoweringContext* loctx) const {
   auto input2 = loctx->GetOutputOp(operand(1));
   return ReturnOp(xla::Div(input1, input2), loctx);
 }
+
+// XlaOpVector SizeConstant::Lower(LoweringContext* loctx) const {
+//   return torch_xla::Scalar::Lower(loctx);
+// }
 
 }  // namespace torch_xla
--- a/torch_xla/csrc/ops/dynamic_ir.h
+++ b/torch_xla/csrc/ops/dynamic_ir.h
@@ -120,10 +120,10 @@ class SizeConstant : public torch_xla::Scalar,
   int64_t getDynamicValue() const override { return getStaticValue(); };
   bool isSymbolic() const override { return false; };
   std::string ToString() const override {
-    this->torch_xla::Scalar::ToString();
+    return this->torch_xla::Scalar::ToString();
   };
   virtual XlaOpVector Lower(LoweringContext* loctx) const override {
-    this->torch_xla::Scalar::Lower(loctx);
+    return torch_xla::Scalar::Lower(loctx);
   };
 };
 


test/test_dynamic_shapes.py
```diff
@@ -12,12 +12,13 @@ def test_simple_expand(self):
...
     t1[3][0] = 1
+    t1[3][1] = 1
     # t2 has size [<=10, 2]
     t2 = torch.nonzero(t1)
     t5 = torch.ones(1, device=dev)
-    t5.expand(t2.size(0))
-    t5_cpu = t5.cpu()
-    self.assertEqual(t5_cpu.shape[0], 1)
+    t6 = t5.expand(t2.size(0))
+    t6_cpu = t6.cpu()
```
vanbasten23, line 20: I wonder why we need to do `.cpu()` first

test/test_dynamic_shapes.py
```diff
@@ -33,13 +34,19 @@ def test_sizeAdd(self):
...
     dynamic_size = int(dyn_size)
     self.assertEqual(dynamic_size, 3)
-    # Exercise SizeAdd::getStaticValue
+    # Exercise SizeAdd::getStaticValue.
     self.assertEqual(str(dyn_size), '<=12')
+    t3 = torch.ones(1, device=dev)
+    # Exercise SizeAdd::Lower.
+    t4 = t3.expand(dyn_size)
+    self.assertEqual(t4.size(0), 3)
+    print(torch_xla._XLAC._get_xla_tensors_text([t4]))
```
vanbasten23, line 49: I wasn't able to test SizeAdd::ToString() method. Let me try to look further.

test/test_dynamic_shapes.py
```diff
@@ -12,12 +12,13 @@ def test_simple_expand(self):
...
     t1 = torch.zeros([size1, size2], device=dev)
     t1[3][0] = 1
+    t1[3][1] = 1
     # t2 has size [<=10, 2]
     t2 = torch.nonzero(t1)
     t5 = torch.ones(1, device=dev)
-    t5.expand(t2.size(0))
-    t5_cpu = t5.cpu()
-    self.assertEqual(t5_cpu.shape[0], 1)
+    t6 = t5.expand(t2.size(0))
```
vanbasten23, line 19: @JackCaoG you made a comment in https://github.com/pytorch/xla/pull/4212/files "I felt like we should add a check here to test t5.shape(), this can be down in a follow up pr.".

Regarding the comment, I think you want to check t6.shape() according to this PR. Can you explain why you want to check t4.shape() and how is it different from what we are doing right now `self.assertEqual(t6_cpu.shape[0], 2)`?

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -83,7 +83,7 @@ int64_t SizeAdd::getDynamicValue() const {
   return dim_node_0->getDynamicValue() + dim_node_1->getDynamicValue();
 }
 
-std::string SizeAdd::ToString() const { return "SizeAdd"; }
+std::string SizeAdd::ToString() const { return "SizeAdd for op aten::add"; }
```
JackCaoG, line 86: are these debugging message that can be removed? It is a long tostring lol

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -110,7 +110,10 @@ int64_t SizeEq::getDynamicValue() const {
 
 std::string SizeEq::ToString() const { return "SizeEq"; }
 
-SizeConstant::SizeConstant(int64_t val) : Scalar(c10::Scalar{val}, xla::S64){};
+SizeConstant::SizeConstant(int64_t val)
+    : Scalar(c10::Scalar{val},
+             xla::ShapeUtil::MakeShape(
+                 GetShapeDimensionType(/*device=*/nullptr), {})){};
```
JackCaoG, line 116: can you update this to all size nodes?

test/test_dynamic_shapes.py
JackCaoG: without the `.cpu` I expect the size to be a symint. Moving the tensor to cpu will trigger execution and materialize the size.

test/test_dynamic_shapes.py
JackCaoG: yea, t6 is find, I thought expand is a inplace operation.

JackCaoG: can you also add a check for `t6`'s ir, by using `print(torch_xla._XLAC._get_xla_tensors_text([t6]))`, I think you should see something like `<=10` in the IR.

test/test_dynamic_shapes.py
```diff
@@ -12,12 +12,13 @@ def test_simple_expand(self):
     size2 = 2
```
JackCaoG, line 12: can we add a check in the beginning of this test to make sure `XLA_EXPERIMENTAL` is set?

test/test_dynamic_shapes.py
vanbasten23: Yeah, I added the print after `self.assertEqual(t6_cpu.shape[0], 2)` and it prints 
```
IR {
  %0 = f32[<=10]{0} xla::device_data(), device=TPU:0, ROOT=0
}
```
Is there any you want to check here?

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: This string has to contain the op name `aten::add` otherwise it will fail to output the IR https://github.com/pytorch/xla/blob/ae35e1e2a808b1920123d76e71871a3a49f1cb58/torch_xla/csrc/ir_dump_util.cpp#L114.

Maybe I can just use `aten::add`. WDYT?

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: I'm planning to add other missing python test for other node in later PR and at that time I'd have to update it. How about I update it in the later PR?

test/test_dynamic_shapes.py
vanbasten23: done

test/test_dynamic_shapes.py
JackCaoG: what's the output of `torch_xla._XLAC._get_xla_tensors_text([t4])` here?  we want to do a `self.assertIn` for the IR here.

torch_xla/csrc/ops/dynamic_ir.cpp
JackCaoG: let's just update the check 
```
 XLA_CHECK_NE(pos, std::string::npos) << node_string << " : " << op_string; 
```
to exclude the `SizeXXX` nodes. Make the check more flexible that it first check whether node is a SizeNode.

test/test_dynamic_shapes.py
vanbasten23: The output of `torch_xla._XLAC._get_xla_tensors_text([t4])` here is
```
IR {
  %0 = s32[] prim::Constant(), value=2
  %1 = f32[] prim::Constant(), value=1
  %2 = f32[1]{0} aten::view(%1), output_size=(1)
  %3 = f32[] prim::Constant(), value=0
  %4 = f32[5,2]{1,0} aten::expand(%3), size=(5, 2)
  %5 = f32[1,2]{1,0} xla::generic_slice(%4), base_indices=(3, 0), sizes=(1, 2)
  %6 = f32[2]{0} aten::view(%5), output_size=(2)
  %7 = f32[2]{0} xla::update_slice(%6, %2), base_indices=(0)
  %8 = f32[1,2]{1,0} aten::view(%7), output_size=(1, 2)
  %9 = f32[5,2]{1,0} xla::update_slice(%4, %8), base_indices=(3, 0)
  %10 = (s32[<=10,2]{1,0}, s32[]) aten::nonzero(%9), num_outputs=2
  %11 = s32[] aten::size(%10.0)
  %12 = s64[] aten::add(%11, %0)
  %13 = f32[] prim::Constant(), value=1
  %14 = f32[1]{0} aten::expand(%13), size=(1)
  %15 = f32[<=12]{0} aten::expand(%14, %12), size=(12), dynamic_dims=(1), ROOT=0
}
```
It seems hard to test in the python code. But how about let me test in https://github.com/pytorch/xla/blob/1685a28821c6a00350bac39dd93faa6797c3c297/test/cpp/test_ir.cpp#L165

test/test_dynamic_shapes.py
vanbasten23: Or do you want to check `<=10` is in `torch_xla._XLAC._get_xla_tensors_text([t6])`?

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: Wouldn't it make the code hacky? It looks to me we don't use the "ToString()" much because it is not in the IR text, as you can see in the comment above. It seems the string in "ToString()" only shows up in the error message.

vanbasten23:
Ok, now the test that I added `TestDynamicShapes.test_sizeAdd` passes on TPU but fails on CPU. I was able to repro the failure on my local cloudtop machine and the failure is:
```
RuntimeError: Error while lowering: SizeAdd for op aten::add
XLA builder error: INVALID_ARGUMENT: Binary op add with different element types: s32[] and s64[].: 
```
The [hlo](https://paste.googleplex.com/4576327268040704) and the [ir graph](https://paste.googleplex.com/6261432990564352). The [full error message](https://paste.googleplex.com/6023607699374080) also doesn't show anymore useful info.

Via gdb, I was able to break at https://github.com/pytorch/xla/blob/1685a28821c6a00350bac39dd93faa6797c3c297/torch_xla/csrc/ops/dynamic_ir.cpp#L90 and find out `input1` has type s32 and `input2` has type s64, hence the type mismatch. However, I couldn't find where `input1` and `input2` come from because `p input1` doesn't reveal any useful info.

So my question is:
1. Why didn't the ir graph or the hlo graph show the type mismatch?
2. What do you suggest to debug further?
@JackCaoG

test/test_dynamic_shapes.py
JackCaoG:
```
Or do you want to check <=10 is in torch_xla._XLAC._get_xla_tensors_text([t6])?
```
yea, I want to make sure that `<=10` works

test/test_dynamic_shapes.py
JackCaoG: You can just assert if `f32[<=12]{0} aten::expand` is in the `torch_xla._XLAC._get_xla_tensors_text([t4])`

torch_xla/csrc/ops/dynamic_ir.cpp
JackCaoG: ok then `aten::add_size` maybe?

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -60,7 +60,7 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
   return ReturnOp(xla::GetDimensionSize(input, this->dim_), loctx);
 }
 
-std::string SizeNode::ToString() const { return "SizeNode"; }
+std::string SizeNode::ToString() const { return "SizeNode for op aten::size"; }
```
JackCaoG, line 63: this one I guess you can just use `aten::size`

JackCaoG (commented):
Let's dig a bit into the
```
ERROR: test_sizeAdd (__main__.TestDynamicShapes)
  File "/tmp/pytorch/xla/test/test_dynamic_shapes.py", line 48, in test_sizeAdd
    self.assertEqual(t4.size(0), 3)
  File "/opt/conda/lib/python3.7/unittest/case.py", line 852, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/opt/conda/lib/python3.7/unittest/case.py", line 842, in _baseAssertEqual
    if not first == second:
  File "/opt/conda/lib/python3.7/site-packages/torch/__init__.py", line 212, in __bool__
    return self.node.bool_()
RuntimeError: Error while lowering: SizeAdd for op aten::add
XLA builder error: INVALID_ARGUMENT: Binary op add with different element types: s32[] and s64[].:
```

We need to identify where the s32 coming from. Dumping the IR of `t4` should help. Maybe we should make `SizeAdd` to handle the casting.

test/test_dynamic_shapes.py
vanbasten23: done

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: SG. Done

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: done

test/test_dynamic_shapes.py
vanbasten23: What I am trying to test is `SizeAdd::ToString()` instead of `expand`. I set `SizeAdd::String()` to return "SizeAdd for op aten::add" but as you can see the string does not appear in `torch_xla._XLAC._get_xla_tensors_text([t4])`.

So I don't think `torch_xla._XLAC._get_xla_tensors_text([t4])` will exercise `SizeAdd::ToString()`. That's why I suggest to test in `test/cpp/test_ir.cpp` in this pr.

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -83,7 +83,7 @@ int64_t SizeAdd::getDynamicValue() const {
   return dim_node_0->getDynamicValue() + dim_node_1->getDynamicValue();
 }
 
-std::string SizeAdd::ToString() const { return "SizeAdd"; }
+std::string SizeAdd::ToString() const { return "aten::add"; }
```
JackCaoG, line 86: `aten::add_size`

test/cpp/test_ir.cpp
```diff
@@ -171,6 +171,7 @@ TEST_F(IrTest, TestSizeAddNode) {
       torch::lazy::MakeNode<SizeNode>(scalar_node, 1);
   torch::lazy::NodePtr size_node_add =
       torch::lazy::MakeNode<SizeAdd>(size_node_0, size_node_1);
+  EXPECT_EQ(size_node_add->ToString(), "SizeAdd for op aten::add");
```
JackCaoG, line 174: update the check here

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -60,7 +60,7 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
   return ReturnOp(xla::GetDimensionSize(input, this->dim_), loctx);
 }
 
-std::string SizeNode::ToString() const { return "SizeNode"; }
+std::string SizeNode::ToString() const { return "aten::size"; }
```
JackCaoG, line 63: Update the `xla::ShapeUtil::MakeShape(xla::S64,` to `GetShapeDimensionType`

vanbasten23: What do you mean? "SizeNode" is already using your fix:
```
SizeNode::SizeNode(torch::lazy::Value input, size_t dim)
    : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::size")},
              {input},
              xla::ShapeUtil::MakeShape(
                  GetShapeDimensionType(/*device=*/nullptr), {}),
```
For other nodes, I plan to update in follow-up PRs.

test/cpp/test_ir.cpp
vanbasten23: done

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: done

torch_xla/csrc/ops/dynamic_ir.cpp
JackCaoG: If test can pass I am ok, but I think one of the cpu test failed because SizeAdd hardcode the s64 as type.

vanbasten23:
@JackCaoG I followed your suggestion to create a shorter repo `test_sizeAdd_shortened` and try to examine the HLO dump, in order to find where the s32/s64 mismatch comes from. But I couldn't get an HLO graph. I tried `print(torch_xla._XLAC._get_xla_tensors_hlo([t4]))` and `$ debug_flag save_ir XLA_DUMP_HLO_GRAPH=1  python3 pytorch/xla/test/test_dynamic_shapes.py TestDynamicShapes.test_sizeAdd_shortened`. Neither of them generate the HLO for me. Do you know if I'm missing anything?

JackCaoG:
I will try to repo sometime today

JackCaoG:
I am able to repo the error, will post my debugging steps tonight.

JackCaoG:
After verify that expand is the source of the error, I dump the IR of the `t4`
```
(Pdb) print(torch_xla._XLAC._get_xla_tensors_text([t4]))
IR {
  %0 = s64[] prim::Constant(), location=to_node@symbolic_shapes.py:139, value=2
  %1 = f32[] prim::Constant(), location=test_sizeAdd_shortened@test_dynamic_shapes.py:94, value=1
  %2 = f32[1]{0} aten::view(%1), location=test_sizeAdd_shortened@test_dynamic_shapes.py:96, output_size=(1)
  %3 = f32[] prim::Constant(), location=test_sizeAdd_shortened@test_dynamic_shapes.py:93, value=0
  %4 = f32[5,2]{1,0} aten::expand(%3), location=test_sizeAdd_shortened@test_dynamic_shapes.py:93, size=(5, 2)
  %5 = f32[1,2]{1,0} xla::generic_slice(%4), location=test_sizeAdd_shortened@test_dynamic_shapes.py:96, base_indices=(3, 0), sizes=(1, 2)
  %6 = f32[2]{0} aten::view(%5), location=test_sizeAdd_shortened@test_dynamic_shapes.py:96, output_size=(2)
  %7 = f32[2]{0} xla::update_slice(%6, %2), location=test_sizeAdd_shortened@test_dynamic_shapes.py:96, base_indices=(0)
  %8 = f32[1,2]{1,0} aten::view(%7), location=test_sizeAdd_shortened@test_dynamic_shapes.py:96, output_size=(1, 2)
  %9 = f32[5,2]{1,0} xla::update_slice(%4, %8), location=test_sizeAdd_shortened@test_dynamic_shapes.py:96, base_indices=(3, 0)
  %10 = (s64[<=10,2]{1,0}, s64[]) aten::nonzero(%9), num_outputs=2, location=test_sizeAdd_shortened@test_dynamic_shapes.py:96
  %11 = s64[] aten::size(%10.0)
  %12 = s64[] aten::add(%11, %0)
  %13 = f32[] prim::Constant(), location=test_sizeAdd_shortened@test_dynamic_shapes.py:104, value=1
  %14 = f32[1]{0} aten::expand(%13), location=test_sizeAdd_shortened@test_dynamic_shapes.py:104, size=(1)
  %15 = f32[<=12]{0} aten::expand(%14, %12), location=test_sizeAdd_shortened@test_dynamic_shapes.py:106, size=(12), dynamic_dims=(1), ROOT=0
}
```
weirdly both input seems to be s64, and this is the only `add` explicitly in the IR. 
```
(Pdb) t4
*** RuntimeError: Error while lowering: aten::add_size
XLA builder error: INVALID_ARGUMENT: Binary op add with different element types: s32[] and s64[].: 
```
One possibility is that `add` that error out here not come from `aten::add` explictly but is part of the lowering of some other op, for example `expand`. I am just going to use gdb to verify.


```
(pytorch) root@65d79ef78e4c:/pytorch/xla# dynamic cgdb python


(gdb) catch throw
Catchpoint 1 (throw)
(gdb) r test/test_dynamic_shapes.py TestDynamicShapes.test_sizeAdd_shortened
```
and I saw
```
 82│ LoweringContext::LoweringContext(
 83│     const std::string& name, torch::lazy::BackendDevice device,
 84│     c10::ArrayRef<const torch::lazy::Node*> post_order,
 85│     torch::lazy::Util::EmissionMap emit_status)
 86│     : torch::lazy::LoweringContext(name, device, {}, emit_status),
 87│       builder_(name) {
 88│   for (auto node : post_order) {
 89├>    LowerNode(node);
 90│   }
 91│ }
```
and if I do
```
(gdb) p node->ToString()
$1 = "aten::add_size"
```
so I was wrong, it is actually coming from `aten::add_size`. Now I need to go back to the lowering code of the SizeAdd and figure out why type at IR level matches, but failed during the actual lowering.

JackCaoG:
```
  %11 = s64[] aten::size(%10.0)
```
looks suspious,  if I have to guess, `GetDimensionSize` also returns `s32`, but we incorrectly flag it to be s64.

JackCaoG:
ok confirmed, if I do (which avoids `SizeAdd` and use `SizeNode` directly which won't crash)
```
(Pdb) t4 = t3.expand(shape0)
(Pdb) print(torch_xla._XLAC._get_xla_tensors_hlo([t4]))
```

I am able to see
```
...
  %get-dimension-size.53 = s32[] get-dimension-size(s64[<=10,2]{1,0} %set-dimension-size.52), dimensions={0}, metadata={op_type="aten__size" op_name="aten__size" source_file="test_sizeAdd_shortened@test_dynamic_shapes.py" source_line=100}
...
```

which suggest Size is actually a s32 even on CPU, but we force it to be S64 in https://github.com/pytorch/xla/blob/d7d04792aab101779ee567b8d2f4659dd0a7c418/torch_xla/csrc/tensor_util.cpp#L1253-L1259

JackCaoG:
with
```
diff --git a/torch_xla/csrc/tensor_util.cpp b/torch_xla/csrc/tensor_util.cpp
index 4329fa0f..27e30982 100644
--- a/torch_xla/csrc/tensor_util.cpp
+++ b/torch_xla/csrc/tensor_util.cpp
@@ -1252,10 +1252,7 @@ bool RequiresRawTypeCasting(at::ScalarType scalar_type,
 
 xla::PrimitiveType GetShapeDimensionType(
     const torch::lazy::BackendDevice* device) {
-  torch::lazy::BackendDevice xla_device = GetDeviceOrCurrent(device);
-  XlaDeviceType hw_type = static_cast<XlaDeviceType>(xla_device.type());
-  return hw_type == XlaDeviceType::CPU ? xla::PrimitiveType::S64
-                                       : xla::PrimitiveType::S32;
+  return xla::PrimitiveType::S32;
 }
```
issue is gone. That being said `SizeAdd` needs to be able to handle type casting. Please change the lowering to 
https://github.com/pytorch/xla/blob/d7d04792aab101779ee567b8d2f4659dd0a7c418/torch_xla/csrc/ops/arithmetic_ir_ops.cpp#L17

JackCaoG:
Here is the full diff
```
diff --git a/torch_xla/csrc/ops/dynamic_ir.cpp b/torch_xla/csrc/ops/dynamic_ir.cpp
index 01c20462..f1de87f0 100644
--- a/torch_xla/csrc/ops/dynamic_ir.cpp
+++ b/torch_xla/csrc/ops/dynamic_ir.cpp
@@ -7,6 +7,9 @@
 #include "torch_xla/csrc/tensor.h"
 #include "torch_xla/csrc/tensor_util.h"
 #include "torch_xla/csrc/xla_graph_executor.h"
+#include "torch_xla/csrc/convert_ops.h"
+
 
 namespace torch_xla {
 
@@ -88,9 +91,11 @@ int64_t SizeAdd::getDynamicValue() const {
 std::string SizeAdd::ToString() const { return "aten::add_size"; }
 
 XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
-  auto input1 = loctx->GetOutputOp(operand(0));
-  auto input2 = loctx->GetOutputOp(operand(1));
-  return ReturnOp((input1 + input2), loctx);
+  static xla::PrimitiveType dim_type = GetShapeDimensionType(/*device=*/nullptr);
+  xla::XlaOp input1 = loctx->GetOutputOp(operand(0));
+  xla::XlaOp input2 = loctx->GetOutputOp(operand(1));
+  xla::XlaOp res = XlaHelpers::PromotedAdd(input1, input2);
+  return ReturnOp(MaybeConvertTo(res, dim_type), loctx);
 }
 
 SizeEq::SizeEq(torch::lazy::Value a, torch::lazy::Value b)
diff --git a/torch_xla/csrc/tensor_util.cpp b/torch_xla/csrc/tensor_util.cpp
index 4329fa0f..27e30982 100644
--- a/torch_xla/csrc/tensor_util.cpp
+++ b/torch_xla/csrc/tensor_util.cpp
@@ -1252,10 +1252,7 @@ bool RequiresRawTypeCasting(at::ScalarType scalar_type,
 
 xla::PrimitiveType GetShapeDimensionType(
     const torch::lazy::BackendDevice* device) {
-  torch::lazy::BackendDevice xla_device = GetDeviceOrCurrent(device);
-  XlaDeviceType hw_type = static_cast<XlaDeviceType>(xla_device.type());
-  return hw_type == XlaDeviceType::CPU ? xla::PrimitiveType::S64
-                                       : xla::PrimitiveType::S32;
+  return xla::PrimitiveType::S32;
 }
 
 }  // namespace torch_xla

```

vanbasten23:
Thanks for looking into it. So to confirm, the `s32` comes from SizeNode and `s64` comes from `%0 = s64[] prim::Constant(), location=to_node@symbolic_shapes.py:139, value=2`. By `looks suspious, if I have to guess, GetDimensionSize also returns s32, but we incorrectly flag it to be s64.`, we don't actually need to do anything here. Is it correct?

JackCaoG:
my patch will force all size to be s32, as long as make sure `SizeAdd` and other size ops will cast result to s32 we should be good.

vanbasten23:
> looks suspious, if I have to guess, GetDimensionSize also returns s32, but we incorrectly flag it to be s64.

But regarding [your comment](https://github.com/pytorch/xla/pull/4258#issuecomment-1360834735) "looks suspious, if I have to guess, GetDimensionSize also returns s32, but we incorrectly flag it to be s64.", we don't have to change the part that flags it to be s64, right?

JackCaoG:
no, we need to make sure dimensionType is always s32.

However for the case of SizeAdd, if we have a `s32 + s64`, I think result will be s64, hence we need `MaybeConvertTo` to cast the result back to s32.

vanbasten23:
Recap:
The failing test is 
```
  def test_sizeAdd_shortened(self):
    self.assertNotEqual(os.environ['XLA_EXPERIMENTAL'], '')
    size1 = 5
    size2 = 2
    t1 = torch.zeros([size1, size2], device=dev)
    t1[3][0] = 1
    # t2 has size [<=10, 2]
    t2 = torch.nonzero(t1)

    shape0 = t2.shape[0]
    shape1 = t2.shape[1]
    dyn_size = shape0 + shape1
    t3 = torch.ones(1, device=dev)

    # Exercise SizeAdd::Lower.
    t4 = t3.expand(dyn_size)
    print(torch_xla._XLAC._get_xla_tensors_hlo([t4]))
    self.assertEqual(t4.size(0), 3) # exception throws here.
```
What we concluded is "Size is actually a s32 even on CPU, but we force it to be S64 in [here](https://github.com/pytorch/xla/blob/d7d04792aab101779ee567b8d2f4659dd0a7c418/torch_xla/csrc/tensor_util.cpp#L1253-L1259). 

To confirm:
1. To fix: 
- Change SizeAdd's Lower() to be `return node.ReturnOp(XlaHelpers::PromotedAdd(op0, op1), loctx);`.
- Cast SizeAdd's result to s32. Is it by changing this [line](https://github.com/pytorch/xla/blob/0986ec50d49ddd208424aa4d1119f4faa4e71e3b/torch_xla/csrc/ops/dynamic_ir.cpp#L30) to be s32?
1. Since you mentioned `Size is actually a s32 even on CPU, but we force it to be S64` in XLA in your [comment](https://github.com/pytorch/xla/pull/4258#issuecomment-1360837552), should we also change the return tup of [getDynamicValue](https://github.com/pytorch/xla/blob/0986ec50d49ddd208424aa4d1119f4faa4e71e3b/torch_xla/csrc/ops/dynamic_ir.cpp#L41) from int64 to int32 as well?
Also should we change 
```
 xla::PrimitiveType GetShapeDimensionType( 
     const torch::lazy::BackendDevice* device) { 
   torch::lazy::BackendDevice xla_device = GetDeviceOrCurrent(device); 
   XlaDeviceType hw_type = static_cast<XlaDeviceType>(xla_device.type()); 
   return hw_type == XlaDeviceType::CPU ? xla::PrimitiveType::S64 
                                        : xla::PrimitiveType::S32; 
 } 
```
just so that it'll always return s32 for size operation?

torch_xla/csrc/ops/dynamic_ir.cpp
vanbasten23: this is fixed.

test/test_dynamic_shapes.py
```diff
@@ -50,26 +55,33 @@ def test_simple_expand_on_2d_tensor(self):
...
   def test_wrap(self):
+    self.assertNotEqual(os.environ['XLA_EXPERIMENTAL'], '')
     a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
     a2 = torch.nonzero(a1)
     self.assertTrue(a2.shape[0] == 3)
     a3 = a2.shape[0] + 3  # tests wrap
     self.assertIsInstance(a3, torch.SymInt)
 
   def test_sizeAdd(self):
+    self.assertNotEqual(os.environ['XLA_EXPERIMENTAL'], '')
```
JackCaoG, line 66: move this check under `if __name__ == '__main__':`

JackCaoG (approved):
mostly lgtm, minor nits

PR #4259, opened 2022-12-01
Enable dynamic shape tests on TPUVM

JackCaoG:
There was a bug to prevent dynamic_shape to be used on TPUVM but that has been solved long ago.


vanbasten23 (approved):
Thanks Jack!

PR #4260, opened 2022-12-01
Move tensor methods out from XLATensor

alanwaketan:
Summary:
This is to simplify the interfaces in XLATensor and match what upstream does. Keeping the code base similar will be easier for us to maintain both at the same time.

A few thing to notice:
1. All tensor methods are now under namespace tensor_methods. The reason to have a distinct namespace is to distinguish from the XLANativeFunctions ones that share the same name.
2. Also some methods and data structures in XLATensor are moved from private to public such that tensor methods can still access them, including `struct Data`, `CreateFrom()`, `SetIrValue()`, `SetInPlaceIrValue()`, `MakeOutputTensors()`, `SetSubView()`, `ModifyCurrentView()`, `CreateViewTensor()`, `CopyTensorToDevice()` and `data()`.
3. `DispatchComparisonOp` is moved to the anonymous namespace within tensor_methods.cpp, and becomes proper helpers.

Test Plan:
CI.


wonjoolee95 (approved):
Thanks! LGTM pending CI. I guess we should eventually update our codegen/op-lowering guides to reflect this. Maybe I can quickly update those as part of the fixit week 😄

alanwaketan:
Thanks Wonjoo for approving the change. And good catch on the guides. That will be a fantastic fixit task.

PR #4262, opened 2022-12-01
Fix the documentation link syntax for PJRT docs.

stgpetrovic:
Markdown requires no spaces between the [description] and the (url).


will-cromar (approved):
Thanks!

PR #4263, opened 2022-12-01
[PJRT] Create function to spawn threads instead of processes.

will-cromar:
This unblocks some of @jonb377's experiments that don't support multiprocessing yet.

### DIFF
--- a/test/pjrt/test_experimental_pjrt_tpu.py
+++ b/test/pjrt/test_experimental_pjrt_tpu.py
@@ -1,12 +1,11 @@
 import concurrent.futures
-import itertools
 import os
+from typing import Dict
 import requests
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


jonb377 (approved):
LGTM, thanks Will!

JackCaoG (commented):
Do we expect to keep this `spawn thread` long term? If not I would say add comment to make it very clear that this is an internal temporary helper. This will prevent us from forgetting deleting this method and some user takes dependency on this method.

will-cromar:
It's not clear whether we need this long-term yet or if we can easily make @jonb377's workload multiprocess-friendly. Since this is under `experimental`, users should assume that we may not merge this into the stable API.

JackCaoG (approved):
Thanks!

PR #4269, opened 2022-12-02
[PJRT] Update PJRT release notes with latest changes.

will-cromar:
Also including some notes on using Docker.

### DIFF
--- a/docs/pjrt.md
+++ b/docs/pjrt.md
@@ -1,11 +1,13 @@
 # Experimental PjRt Runtime Support
 
+_This document reflects the current state of PjRt support in current nightly
+builds_. See the [same document on the r1.13 branch](https://github.com/pytorch/xla/blob/r1.13/docs/pjrt.md)
+for the status in the latest stable release.
+
 The PyTorch/XLA team is currently migrating from the currently-supported XRT
 runtime to the [PjRt
 runtime](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/compiler/xla/pjrt)
-used by [JAX](https://github.com/google/jax). Although PjRt may work on TPU v2
-and v3, we plan on making PjRt the officially supported runtime for PyTorch/XLA
-on TPU v4 and future generations of TPU.
+used by [JAX](https://github.com/google/jax).
 
 PjRt is available as an _experimental preview_ in PyTorch/XLA r1.13. The
 PyTorch/XLA team will provide limited support on a best-effort basis during this
@@ -31,7 +33,7 @@ like this:
 PJRT_DEVICE=CPU python3 xla/test/test_train_mp_mnist.py --fake_data
 ```
 
-### v4 TPU
+### TPU
 
 To create a new TPU with PyTorch/XLA r1.13 installed:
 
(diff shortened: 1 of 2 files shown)


docs/pjrt.md
```diff
@@ -100,6 +122,47 @@ for more information about TPU architecture.
...
+or [`torch.distributed.all_gather_object`](https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_gather_object)
+with a `gloo` process group. If you are also using the `xla` `torch.distributed`
+backend, you can use `torch.new_group` to create a `gloo` subgroup. See [this
+example](https://pytorch.org/docs/stable/distributed.html#torch.distributed.barrier)
+from the PyTorch documentation. Keep in mind these constraints:
+
+- `torch.distributed` is not fully supported on TPU v2/v3 at this time. Only a
+  subset of operations with the `xla` backend are tested, and `gloo` will likely
+  not work as expected in a multiprocessing context.
+- In our experiments, `gloo` does not scale well to thousands of chips, so
```
alanwaketan, line 160: Is there any documentation to read more about gloo with XLA? Just curious why gloo works.

test/test_train_mp_imagenet.py
```diff
@@ -199,6 +200,9 @@ def train_imagenet():
 
   device = xm.xla_device()
   model = get_model_property('model_fn')().to(device)
+  if pjrt.using_pjrt():
```
alanwaketan, line 203: What's this change about? To mimic DDP behavior?

docs/pjrt.md
will-cromar: `gloo` works out of the box in this case because it doesn't interact with XLA at all. It's PyTorch's recommended backend for CPU-only training: https://pytorch.org/docs/stable/distributed.html#which-backend-to-use

test/test_train_mp_imagenet.py
will-cromar: Added a comment here to explain. Because threads share the same random generator while initializing, it's impossible to set the same initialization seed across replicas in PJRT with TPU v2/v3, so you have to synchronize the parameters manually. DDP does not broadcast parameters according to the [documentation](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel).

I included this change in this PR so I can use it as an example in `pjrt.md`.

docs/pjrt.md
alanwaketan: So you mean we recommend users to do cpu training or recommend users to move tensors to cpu and then use gloo?

will-cromar: This section is talking specifically about `xm.rendezvous`, which is used for communicating data that's already on CPU. `gloo` won't work with XLA tensors directly, so you couldn't use it e.g. as the backend for DDP on TPUs. But you can still use `torch.distributed` utilities like `barrier` and `all_gather_object` in some cases as long as the input tensors aren't XLA tensors

test/test_train_mp_imagenet.py
will-cromar: The issue here is that with TPU v2/v3, each process manages multiple replicas, which all share the same global RNG. For example, rank 0 will have ordinals 0 and 1 concurrently initializing and advancing the random generator, and you'll end up with different parameters between the two threads. On TPU v4, you can set the same seed in each process since each process runs initialization only once (assuming you are running with one TPU chip per process, which is not always true).

We could work around the limitation for TPU v2/v3 if we could 1) initialize the parameters directly on the device (we could set both TPU cores' seeds to the same thing) or 2) directly supply a separate RNG with the same seed for each thread. As far as I know, neither of these are possible.

My understanding from the DDP documentation is that parameters are not broadcast between replicas, but you can optionally broadcast buffers on each forward pass:

```
Parameters are never broadcast between processes. The module performs an all-reduce step on gradients and assumes that they will be modified by the optimizer in all processes in the same way. Buffers (e.g. BatchNorm stats) are broadcast from the module in process of rank 0, to all other replicas in the system in every iteration.
```

If DDP does broadcast parameters at intitialization, then I can move this line into the experimental implementation in `pjrt.py` to maintain parity.

JackCaoG (approved):
Thanks!

docs/pjrt.md
alanwaketan: Oh, it's complicated but it makes sense.

alanwaketan (approved):
LGTM.

PR #4273, opened 2022-12-05
Document environment variables

steventk-g:


configuration.yaml
```diff
@@ -0,0 +1,490 @@
...
+  XLA_DUMP_HLO_GRAPH:
+    description:
+      - If set to true in case of a compilation or execution error the
+        offending HLO graph will be dumped as part of the runtime error raised
+        by xla_util.cc.
+    type: bool
+    default_value: false
+  XLA_TEST_DUMP_METRICS:
+    description:
+      - Controls whether or not metrics are dumped in test tear down.
```
JackCaoG, line 57: I think this is only for cpp test? If that's the case let's specified `cpp test`

configuration.yaml
```diff
@@ -0,0 +1,490 @@
...
+    type: bool
+    default_value: false
+  XLA_NO_SPECIAL_SCALARS:
+    description:
+      - When set to false, this will enable a compiler optimization to route
```
JackCaoG, line 62: it is not really a compiler optimization, it is more of we decided to pass scale value 0 and 1 as constant rather than device data to enable some compiler optimization.

configuration.yaml
```diff
@@ -0,0 +1,490 @@
...
+    type: bool
+    default_value: false
+  BUNDLE_LIBTPU:
+    description:
+      - Whether or not to include libtpu in the final wheel.
+    type: bool
+    default_value: false
+  ALLOW_MULTIPLE_LIBTPU_LOAD:
+    description:
+      - Allows for multiple Libtpu load. Set to true by default.
```
JackCaoG, line 120: allow for multiple process to load libtpu at the same time.

configuration.yaml
```diff
@@ -0,0 +1,490 @@
...
+    type: bool
+    default_value: true
+  PT_XLA_DEBUG:
+    description:
+      - Used to automatically analyze the metrics report and provide a summary.
+    type: bool
+    default_value: false
+  TPUVM_MODE:
+    description:
+      - Set to true when building wheel for a TPUVM.
```
JackCaoG, line 130: something along the line of `include some additional TPUVM bits when building the third party tensorflow`

configuration.yaml
```diff
@@ -0,0 +1,490 @@
...
+    description:
+      - List of metrics percentiles to record.
+    type: string
+    default_value: "0.01:0.05:0.1:0.2:0.5:0.8:0.9:0.95:0.99"
+  GRPC_VERBOSITY:
+    description:
+      - Verbosity level for GRPC, e.g. INFO, ERROR, etc.
+    type: string
+    default_value: "ERROR"
+  TPU_ML_PLATFORM:
```
JackCaoG, line 443: lol I felt like we should not make this configurable

JackCaoG (commented):
Mostly LGTM. Thanks for putting this together! Is it possible to group similar env var for better reading expericence?

For example we can split them into 
1. XRT related
2. Build related
3. Debug related
4. Feature enablement
5. Device configuration
6. Distributed related
7. ...

configuration.yaml
steventk-g: I wonder if we should include it here for the sake of documentation, and remark that it shouldn't be changed. My intention in including it was just to cover every environment variable that shows up in this repo.
https://github.com/pytorch/xla/blob/f8581559272003d8828f15e2207e794e462ad463/torch_xla/__init__.py#L66

configuration.yaml
```diff
@@ -0,0 +1,490 @@
...
+    type: bool
+    default_value: false
+  XLA_BAZEL_VERBOSE:
+    description:
+      - Turn on verbose messages during the bazel build of the xla/xrt client.
+    type: bool
+    default_value: false
+  XLA_CUDA:
+    description:
+      - Build the xla/xrt client with CUDA enabled.
```
will-cromar, line 110: Change to `XLA client` since this applies to both XRT and PJRT

will-cromar (commented):
Wow! Thanks for laying all of this out. LGTM at least for the variables I know about

I agree with Jack's comment that it would be clearer if you categorized the variables, especially separating the build-time variables from the run-time variables and separating the XRT configuration from the rest.

@JackCaoG Do we also want some designation of which variables are stable and which might get removed in the future? Or are we committed to actually supporting all of these?

JackCaoG:
yea I think in this or I am fine with only group them by cateratory. We can update it based on stable/to_be_removed in a follow up pr.

configuration.yaml
steventk-g: Done

configuration.yaml
steventk-g: Done

configuration.yaml
steventk-g: Done

configuration.yaml
steventk-g: Done

configuration.yaml
steventk-g: Done

PR #4278, opened 2022-12-05
Fix tf_patches/f16_abi_clang.diff ill-format.

vanbasten23:
Without this fix, `git apply` command failed with
```
(base) jenkins@ecaf13a10ec0:/workspace/pytorch/xla/third_party/tensorflow$ git apply ../../tf_patches/f16_abi_clang.diff
error: corrupt patch at line 19
```
The file needs a newline at the end.


JackCaoG (approved):

JackCaoG:
wait, does it mean this `git apply` never worked?

If that's the case, and our CI is passing, we can just get rid of this patch.

vanbasten23:
> 

I think CI uses `patch` https://github.com/pytorch/xla/blob/f2c87eec1dee2b59739d5553a2920dca47068482/scripts/cond_patch.py#L37
and it could deal with files without ending in newlines. But `git apply` keeps failing on my docker container.

PR #4282, opened 2022-12-06
Lower _native_batch_norm_legit

JackCaoG:
New op is being added in https://github.com/pytorch/pytorch/pull/88697/files and has similar sematics as `native_batch_norm`. These ops will only be used in dynamo and currently blocks https://github.com/pytorch/pytorch/pull/88449

I don't think this op will be triggered in c++ or enable the opinfo test.


alanwaketan (approved):
LGTM.

PR #4284, opened 2022-12-06
Add note [Note: Re-using upstream TensorImpl]

wonjoolee95:
Add note [Note: Re-using upstream TensorImpl]

---
We decided not to re-use the upstream LTCTensorImpl. Adding a note for future reference. The LTC design doc is updated to reflect this as well.


alanwaketan (approved):
LGTM. Wonjoo, thanks for the investigation.

JackCaoG (approved):

PR #4285, opened 2022-12-06
Correctly configure PJRT execution when nprocs=1.

stgpetrovic:
Currently there is not easy way to tweak a model written with pjrt.run_multiprocess to run with single core/device easily. This is a useful feature when debugging some model failure.

In the XRT world, we can pass the `num_process` as 1 to `xmp.spawn`, and this should be mirrored in PJRT.

### DIFF
--- a/torch_xla/experimental/pjrt.py
+++ b/torch_xla/experimental/pjrt.py
@@ -251,18 +251,10 @@ def _run_singleprocess(fn: Callable[..., R],
   if device_type() == 'TPU':
     tpu.configure_one_chip_topology()
 
-  device = xm.get_xla_supported_devices()[0]
   xm.set_replication(xm.xla_device(), [])
-
-  @functools.wraps(fn)
-  def _thread_fn(device: torch.device):
-    torch_xla._XLAC._xla_set_default_device(device)
-
-    return fn()
-
   init_pjrt_process_group()
 
-  return _thread_fn(device)
+  return fn()
 
 
 @requires_pjrt
@@ -344,10 +336,9 @@ def spawn(fn: Callable,
   spawn_fn = _SpawnFn(fn, *args)
 
   if nprocs == 1:
-    tpu.configure_one_chip_topology()
     return _run_singleprocess(spawn_fn, start_method=start_method)
   elif nprocs is not None:
-    print('Unsupported nprocs (%d), ignoring...' % nprocs, file=sys.stderr)
+    logging.warning('Unsupported nprocs (%d), ignoring...' % nprocs)
 
   _run_multiprocess(spawn_fn, start_method=start_method)
 
--- a/torch_xla/experimental/pjrt.py
+++ b/torch_xla/experimental/pjrt.py
@@ -263,8 +263,10 @@ def _run_singleprocess(fn: Callable[..., R],
 
       return fn()
 
-    init_pjrt_process_group()
-    return executor.submit(_thread_fn, xm.xla_device()).result()
+    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
+      init_pjrt_process_group()
+
+      return executor.submit(_thread_fn, xm.xla_device()).result()
 
 
 @requires_pjrt
--- a/torch_xla/experimental/pjrt.py
+++ b/torch_xla/experimental/pjrt.py
@@ -103,9 +103,9 @@ def global_device_count() -> int:
 @requires_pjrt
 def world_size() -> int:
   """Returns the total number of configured logical devices."""
-  if not torch_xla._XLAC._xla_get_replication_devices_count():
+  if torch_xla._XLAC._xla_get_replication_devices_count() == 0:
     return 1
-  return len(torch_xla._XLAC._xla_get_all_devices())
+  return global_device_count()
 
 
 @requires_pjrt
@@ -170,25 +170,6 @@ def _merge_replica_results(
   return dict(replica_results)
 
 
-# TODO(wcromar): remove this when the TPU master IP becomes predictable
-def _discover_tpu_master_worker_ip(device: torch.device):
-  torch_xla._XLAC._xla_set_default_device(device)
-
-  return tpu.discover_master_worker_ip()
-
-
-def _maybe_init_distributed_torch(executor, devices, master_port):
-  if os.getenv('PJRT_INIT_TORCH_DISTRIBUTED', '0') == '1':
-    if device_type() == 'TPU':
-      # HACK: need to call with each device since it relies on an XLA collective
-      master_ip = next(executor.map(_discover_tpu_master_worker_ip, devices))
-      init_method = f'tcp://{master_ip}:{master_port}'
-    else:
-      init_method = None
-
-    init_pjrt_process_group(init_method=init_method)
-
-
 @requires_pjrt
 def _run_thread_per_device(local_rank: int,
                            local_world_size: int,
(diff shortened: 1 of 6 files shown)


torch_xla/experimental/pjrt.py
```diff
@@ -161,6 +170,25 @@ def _merge_replica_results(
...
 
 
+# TODO(wcromar): remove this when the TPU master IP becomes predictable
+def _discover_tpu_master_worker_ip(device: torch.device):
+  torch_xla._XLAC._xla_set_default_device(device)
+
+  return tpu.discover_master_worker_ip()
+
+
+def _maybe_init_distributed_torch(executor, devices, master_port):
```
will-cromar, line 180: nit: either move back into `_run_thread_per_device` (see below) or add [type hints](https://docs.python.org/3/library/typing.html) here

torch_xla/experimental/pjrt.py
```diff
@@ -220,6 +233,49 @@ def _discover_tpu_master_worker_ip(device: torch.device):
...
+  Returns:
+    the result of calling `fn`.
+  """
+  with concurrent.futures.ProcessPoolExecutor(
+      max_workers=1,
+      mp_context=torch.multiprocessing.get_context(start_method)) as executor:
+    os.environ.setdefault('LOCAL_WORLD_SIZE', '1')
+
+    if device_type() == 'TPU':
+      tpu.configure_topology(0, 1)
```
will-cromar, line 262: Should this be `tpu.configure_one_chip_topology`? That will set all of the necessary env vars.

torch_xla/experimental/pjrt.py
```diff
@@ -220,6 +233,49 @@ def _discover_tpu_master_worker_ip(device: torch.device):
...
+    fn: Function to run on the device devices
+    args: args to pass to `fn`
+    start_method: The Python `multiprocessing` process creation method.
+      Default: `spawn`
+    kwargs: kwargs to pass to `fn`
+
+  Returns:
+    the result of calling `fn`.
+  """
+  with concurrent.futures.ProcessPoolExecutor(
```
will-cromar, line 256: This outer `executor` isn't used at all

I think that's actually okay though. The implementation that we're mirroring also does not create any child processes for a single-device run.

https://github.com/pytorch/xla/blob/a1c054dadfa44aa320fe58b73883ad3c23c0912a/torch_xla/distributed/xla_multiprocessing.py#L397-L398

So I think you can just remove this context and follow the logic of `xmp.spawn`. The main reason we would want to create a child process is so any state (especially TPU topology env vars and the TPU client itself) we alter while running `fn` effectively gets reset when the child exits. I guess a user might also reasonably expect `xmp.spawn` to _spawn_ something.

@JackCaoG what do you think is the better option?

torch_xla/experimental/pjrt.py
```diff
@@ -220,6 +233,49 @@ def _discover_tpu_master_worker_ip(device: torch.device):
...
+    devices = xm.get_xla_supported_devices()[:1]
+    xm.set_replication(xm.xla_device(), [])
+
+    @functools.wraps(fn)
+    def _thread_fn(device: torch.device):
+      torch_xla._XLAC._xla_set_default_device(devices[0])
+
+      return fn()
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
```
will-cromar, line 273: You don't need to create a thread pool executor here, since we're only running one instance of `fn`.

The only reason you would need a thread pool that I see is for `_maybe_init_torch_distributed`. However, the only reason we need a thread pool there is we need to have to use a hack to discover the TPU master IP. We actually don't need to contact the master over TCP at all to initialize torch.distributed in this case, since there's only one process.

There's a default in `init_pjrt_process_group` to use a random file to initialize when in a single process context. So, you can actually just directly call `init_pjrt_process_group` here and remove the `ThreadPoolExecutor` entirely.

will-cromar (changes requested):

torch_xla/experimental/pjrt.py
```diff
@@ -99,6 +100,14 @@ def global_device_count() -> int:
   return len(torch_xla._XLAC._xla_get_all_devices())
 
 
+@requires_pjrt
+def world_size() -> int:
+  """Returns the total number of configured logical devices."""
+  if not torch_xla._XLAC._xla_get_replication_devices_count():
+    return 1
+  return len(torch_xla._XLAC._xla_get_all_devices())
```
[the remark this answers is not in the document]
stgpetrovic: Done.

torch_xla/experimental/pjrt.py
stgpetrovic: Done.

torch_xla/experimental/pjrt.py
stgpetrovic: Cool! Done.

torch_xla/experimental/pjrt.py
stgpetrovic: Done.

torch_xla/experimental/pjrt.py
JackCaoG: I don't see a need to create child process for a single-device. If this is significantly easy to implement this way I am also ok with it but it does seems a bit redundant.

will-cromar: Sounds good. @stgpetrovic can you remove both executors?

torch_xla/experimental/pjrt.py
```diff
@@ -218,6 +226,45 @@ def _discover_tpu_master_worker_ip(device: torch.device):
...
+  os.environ.setdefault('LOCAL_WORLD_SIZE', '1')
+
+  if device_type() == 'TPU':
+    tpu.configure_one_chip_topology()
+
+  device = xm.get_xla_supported_devices()[0]
+  xm.set_replication(xm.xla_device(), [])
+
+  @functools.wraps(fn)
+  def _thread_fn(device: torch.device):
```
will-cromar, line 258: You can remove `thread_fn` and call `fn()` directly. `set_replication` already sets the default device correctly in this case: https://github.com/pytorch/xla/blob/4c7c343c8482df80c21a365e2a945713ff0cf04e/torch_xla/core/xla_model.py#L345

torch_xla/experimental/pjrt.py
```diff
@@ -280,16 +327,28 @@ def __call__(self) -> None:
...
+    nprocs (int): The number of processes/devices for the replication. At the
+      moment, if specified, can be either 1 or the maximum number of devices.
     args: args to pass to `fn`
     start_method: The Python `multiprocessing` process creation method.
       Default: `spawn`
   """
   spawn_fn = _SpawnFn(fn, *args)
+
+  if nprocs == 1:
+    tpu.configure_one_chip_topology()
```
will-cromar, line 347: Remove this since it's already called in `_run_single_process`

torch_xla/experimental/pjrt.py
```diff
@@ -280,16 +327,28 @@ def __call__(self) -> None:
...
+    return _run_singleprocess(spawn_fn, start_method=start_method)
+  elif nprocs is not None:
+    print('Unsupported nprocs (%d), ignoring...' % nprocs, file=sys.stderr)
```
will-cromar, line 350: Use [`logging.warning`](https://docs.python.org/3/library/logging.html) instead

torch_xla/experimental/pjrt.py
```diff
@@ -218,6 +226,45 @@ def _discover_tpu_master_worker_ip(device: torch.device):
...
+
+  Returns:
+    the result of calling `fn`.
+  """
+  os.environ.setdefault('LOCAL_WORLD_SIZE', '1')
+
+  if device_type() == 'TPU':
+    tpu.configure_one_chip_topology()
+
+  device = xm.get_xla_supported_devices()[0]
```
will-cromar, line 254: Use `device = xm.xla_device()` instead. That will still default to `xla:0` but leave us some flexibility to change the default device before calling `_run_singleprocess`

will-cromar (approved):
Just some final minor nits. Otherwise LGTM once the CI passes.

PR #4286, opened 2022-12-06
Add sharding hash to IR nodes

steventk-g:


torch_xla/csrc/ir.h
```diff
@@ -113,15 +113,28 @@ class XlaNode : public torch::lazy::Node {
   torch::lazy::hash_t hash() const override { return dag_hash_; }
 
   torch::lazy::hash_t shapeHash() const override { return dag_hash_; }
+
+  // TODO(steventk) Do we need to add this to the parent class
+  // (torch::lazy::Node)?
+  torch::lazy::hash_t shardingHash() const { return sharding_hash_; }
```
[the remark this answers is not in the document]
JackCaoG: we want to keep this in the downstream XLANode.

torch_xla/csrc/ir.h
```diff
@@ -132,9 +138,13 @@ class XlaNode : public torch::lazy::Node {
 
   static std::vector<torch::lazy::SourceLocation> GetFrameInfo();
 
+  static torch::lazy::hash_t CreateShardingHash(
+      std::shared_ptr<xla::OpSharding> sharding, torch::lazy::hash_t hash_seed);
+
   xla::Shape xla_shape_;
   torch::lazy::hash_t node_hash_ = 0;
   torch::lazy::hash_t dag_hash_;
+  torch::lazy::hash_t sharding_hash_ = 0;
```
yeounoh, line 147: How does sharding_hash_ get used, is it combined to node_hash_ to represent the final node hash?

steventk-g: In the subsequent PR (https://github.com/pytorch/xla/pull/4287), we combine it with the dag hash when `hash()` is called.

JackCaoG: I think we should just combined it with `dag_hash_` during construction and not maintaining it as a separate state.

torch_xla/csrc/ir.h
```diff
@@ -113,15 +113,21 @@ class XlaNode : public torch::lazy::Node {
...
     return output_sharding_;
   }
-  void SetSharding(const xla::OpSharding& sharding) {
-    output_sharding_ = std::make_shared<xla::OpSharding>(sharding);
+
+  void SetSharding(const xla::OpSharding& sharding);
+
+  void ClearSharding() {
+    output_sharding_ = nullptr;
+    sharding_hash_ = 0;
```
JackCaoG, line 129: I felt like we should not provide `ClearSharding` and `SetSharding`, we need to create new IR node if we want to modify the sharding.

torch_xla/csrc/ir.h
steventk-g: In subsequent PRs I combine it with dag hash and then move that logic into the constructor, this was just meant to be an intermediate state.

yeounoh (approved):
LGTM, these will be tested in the subsequent PR #4287

PR #4287, opened 2022-12-06
Use sharding hash

steventk-g:


test/test_xla_sharding.py
```diff
@@ -116,6 +116,18 @@ def test_inplace_add_with_sharding(self):
     xt.add_(1)  # inplace update should preserve the sharding
     self.assertEqual(sharding_spec, torch_xla._XLAC._get_xla_sharding_spec(xt))
 
+  def test_shard_hashing(self):
+    xt1 = torch.ones(2, 2).to(xm.xla_device())
```
yeounoh, line 120: There is a function called, `_get_graph_hash` in `_XLAC`. Let's use it to explicitly check (assertions) before and after sharding hashes, and keep the final execution to make sure both the original computation and the new shasrded computation yield the same result.

torch_xla/csrc/ir.cpp
```diff
@@ -165,4 +172,40 @@ const xla::Shape& GetXlaShape(const torch::lazy::Value& value) {
   return casted->xla_shape(value.index);
 }
 
+torch::lazy::hash_t XlaNode::CreateShardingHash(
+    std::shared_ptr<xla::OpSharding> sharding, torch::lazy::hash_t hash_seed) {
```
yeounoh, line 176: Let's leave a comment explaining that we are skipping some fields, like layout.

yeounoh (changes requested):
Mostly LGTM. Some minor changes needed.

torch_xla/csrc/ir.h
```diff
@@ -110,18 +110,29 @@ class XlaNode : public torch::lazy::Node {
...
   // TODO: test multi-output example.
   const std::shared_ptr<xla::OpSharding> GetSharding() const {
     return output_sharding_;
   }
-  void SetSharding(const xla::OpSharding& sharding) {
-    output_sharding_ = std::make_shared<xla::OpSharding>(sharding);
+
+  void SetSharding(const xla::OpSharding& sharding);
+
+  void ClearSharding() {
```
yeounoh, line 132: For a future reference, #4288 , we may need to create a new IR when we clear sharding; the new IR node should return the same hash as the original pre-sharded IR node. cc @JackCaoG

yeounoh: Can we implement and test this in #4288 ?

yeounoh (approved):
LGTM, this will be the basis of #4288 . Let's land this PR after e2e testing with @jonb377 's benchmark script. Specifically, we should watch out for # compilation (cache miss) regression.

test/test_xla_sharding.py
```diff
@@ -118,6 +118,23 @@ def test_inplace_add_with_sharding(self):
...
+    xt2 = torch.ones(2, 2).to(xm.xla_device())
+
+    # Add sharding to xt1, this should result in the hashes being different for
+    # xt1 and xt2
+    xs.mark_sharding(xt1, self._get_mesh((1, self.n_devices)), (0, 1))
```
JackCaoG, line 127:
```
t1 = torch.ones(2, 2).to(xm.xla_device())
t2 = t1 * 2
xs.mark_sharding(xt1, self._get_mesh((1, self.n_devices)), (0, 1))
t3 = t1 * 2
```

I think with the current implemenation, `t1` will give different hash after `mark_sharding` will give different hash, but `t2` and `t3` will point to same IR. I felt like that's wrong? I think `SetShardingSpec` should create new IR with sharding and we should not maintain both `sharding_hash_` and `dag_hash_`

yeounoh: https://github.com/pytorch/xla/pull/4287#discussion_r1046625399

PR #4288, opened 2022-12-06
Introduce CreateShardedIrValue API

steventk-g:
We replace SetSharding/ClearSharding APIs with CreateShardedIRValue, so that we can create hashes for the sharded node upon it's creation rather than updating it after it's been created.

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1501,7 +1501,7 @@ void InitXlaModuleBindings(py::module m) {
         std::vector<XLATensor::ShardingSpecPtr>{new_sharding_spec},
         std::vector<std::string>{GetVirtualDevice().toString()})[0];
     xtensor->SetXlaData(xla_data);
-    xtensor->SetShardingSpec(*new_sharding_spec);
+    xtensor->CreateShardedIrValue(sharding);
   });
   m.def("_xla_clear_sharding", [](const at::Tensor& input) {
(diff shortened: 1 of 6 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1488,7 +1488,7 @@ void InitXlaModuleBindings(py::module m) {
         std::vector<XLATensor::ShardingSpecPtr>{sharding_spec},
         std::vector<std::string>{GetVirtualDevice().toString()})[0];
     xtensor->SetXlaData(xla_data);
-    xtensor->SetShardingSpec(*sharding_spec);
+    xtensor->CreateShardedIrValue(*sharding_spec);
```
yeounoh, line 1491: Let's also do e2e training test, and verify that this doesn't affect the compilation & caching to be cold throughout the loop.

torch_xla/csrc/ir.h
```diff
@@ -110,18 +115,22 @@ class XlaNode : public torch::lazy::Node {
 
   torch::lazy::hash_t node_hash() const { return node_hash_; }
 
-  torch::lazy::hash_t hash() const override { return dag_hash_; }
+  torch::lazy::hash_t hash() const override {
+    if (sharding_hash_ != 0) {
+      return torch::lazy::HashCombine(dag_hash_, sharding_hash_);
```
JackCaoG, line 120: why don't we always combine `dag_hash_` and `sharding_hash_`? From you implementation above `dag_hash_` already include `sharding_hash_`

Issue #4202, opened 2022-11-15
[Dynamic Shape] `torch.exapnd` does not work with dynamic target size from python


JackCaoG:
## 🐛 Bug

In https://github.com/pytorch/xla/pull/3889 we added the support for `exapnd` with `symint`. We have a test at https://github.com/pytorch/xla/blob/a843d8bf523d68e49ac2976b92a131c207cd1117/test/cpp/test_aten_xla_tensor.cpp#L4972-L4978 
to verify that this works at C++ front. However when I try to test this on python it does not work and result into a crash

```
import torch
import torch_xla
import torch_xla.core.xla_model as xm

import os
os.environ["XLA_EXPERIMENTAL"] = "nonzero:masked_select:masked_scatter"

dev = xm.xla_device()
size1 = 5
size2 = 2
t1 = torch.zeros([size1, size2], device=dev)
t1[3][0] = 1
# t2 has size [<=10, 2]
t2 = torch.nonzero(t1)
t5 = torch.tensor(10, device=dev)
t5.expand(t2.size(0))
print(t5.cpu())
```

error
```
(pytorch) root@65d79ef78e4c:/pytorch/xla# python test/test_dynamic_fail.py 
Segmentation fault (core dumped)
```

@miladm @vanbasten23 @Krovatkin @wconstab

vanbasten23:
It sounds we also need to add python tests. Let me fix it and also add your python test to the CI.

JackCaoG:
I think
https://github.com/pytorch/xla/blob/a843d8bf523d68e49ac2976b92a131c207cd1117/test/cpp/test_aten_xla_tensor.cpp#L5003-L5015
is a better example since it is truly dynamic.

vanbasten23:
I followed the example you provided (https://github.com/pytorch/xla/blob/a843d8bf523d68e49ac2976b92a131c207cd1117/test/cpp/test_aten_xla_tensor.cpp#L5003-L5015) and wrote a corresponding python script:
```
import torch
import torch_xla
import torch_xla.core.xla_model as xm

import os
os.environ["XLA_EXPERIMENTAL"] = "nonzero:masked_select:masked_scatter"

dev = xm.xla_device()
size1 = 5
size2 = 2
t1 = torch.zeros([size1, size2], device=dev)
t1[3][0] = 1
t1[3][1] = 1
# t2 has size [<=10, 2]
t2 = torch.nonzero(t1)
t3 = torch.ones(size1, size2)
[...]

PR #4291, opened 2022-12-07
Add one more symint expand test.

vanbasten23:
Inspired by [GH issue](https://github.com/pytorch/xla/issues/4202), I'm adding another symint expand test in this PR.

### DIFF
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -19,6 +19,18 @@ class TestDynamicShapes(unittest.TestCase):
     t5_cpu = t5.cpu()
     self.assertEqual(t5_cpu.shape[0], 1)
 
+  def test_simple_expand_on_2d_tensor(self):
+    size1 = 5
+    size2 = 2
+    t1 = torch.zeros([size1, size2], device=dev)
+    t1[3][0] = 1
+    t1[3][1] = 1
+    # t2 has size [<=10, 2]
+    t2 = torch.nonzero(t1)
+    t3 = torch.ones(1, size1, size2)
+    t4 = t3.expand(t2.shape[0], size1, size2)
+    print(t4.cpu())
+
   def test_wrap(self):
     a1 = torch.tensor([[1, 0, 0, 5, 0, 6]], device=dev)
     a2 = torch.nonzero(a1)
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -27,9 +27,27 @@ class TestDynamicShapes(unittest.TestCase):
     t1[3][1] = 1
     # t2 has size [<=10, 2]
     t2 = torch.nonzero(t1)
-    t3 = torch.ones(1, size1, size2)
-    t4 = t3.expand(t2.shape[0], size1, size2)
-    print(t4.cpu())
+    t3 = torch.ones(1, size2, device=dev)
+
+    # varargs
+    t4 = t3.expand(t2.shape[0], t2.shape[1])
+    self.assertEqual(t4.shape[0], 2)
+    self.assertEqual(t4.shape[1], size2)
+
+    # shape list
+    t4 = t3.expand((t2.shape[0], t2.shape[1]))
+    self.assertEqual(t4.shape[0], 2)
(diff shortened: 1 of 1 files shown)


vanbasten23:
Right now, the test fails with `RuntimeError: SymIntArrayRef expected to contain only concrete integers`.

test/test_dynamic_shapes.py
```diff
@@ -19,6 +19,18 @@ def test_simple_expand(self):
...
+  def test_simple_expand_on_2d_tensor(self):
+    size1 = 5
+    size2 = 2
+    t1 = torch.zeros([size1, size2], device=dev)
+    t1[3][0] = 1
+    t1[3][1] = 1
+    # t2 has size [<=10, 2]
+    t2 = torch.nonzero(t1)
+    t3 = torch.ones(1, size1, size2)
+    t4 = t3.expand(t2.shape[0], size1, size2)
```
JackCaoG, line 31: @ezyang Does pytorch supports one of the expand dimension is dynamic and others are static?

ezyang:
You can nail this down precisely if you get a c++ backtrace but usually this just means there isn't a _symint registration for the op (expand, I guess?)

vanbasten23:
> You can nail this down precisely if you get a c++ backtrace but usually this just means there isn't a _symint registration for the op (expand, I guess?)

Thanks Ed. I've gotten the [c++ backtrace](https://gist.github.com/vanbasten23/a5b097ce557f2c0f01a7815aa4ea02b9). Can you show me where I can find the _symint registration for an op?

JackCaoG:
@vanbasten23 Did you compile pytorch and pytorch/xla with `DEBUG=1`? A lot of the frame seems to be missing

JackCaoG:
I expect the code to hit https://github.com/pytorch/xla/blob/653df694f7fafe5cda39fbc90e14e0347fde07ae/torch_xla/csrc/aten_xla_type.cpp#L1156-L1170

@vanbasten23 can you confirm?

vanbasten23:
> @vanbasten23 can you confirm?

re @JackCaoG , ah I see where the problem is. In the current test, `t1`(`t1 = torch.zeros([size1, size2], device=dev)`) tensor is on xla device but `t3 = torch.ones(1, size1, size2)` is not. Perhaps that's why we see `frame #21: at::_ops::expand::call(at::Tensor const&, c10::ArrayRef<c10::SymInt>, bool) ` in the stacktrace, in which PyTorch dispatch the call to aten I guess?

Btw, I did compile pytorch and pt/xla with DEBUG=1 and that's all the stacktrace.

After I set the device for `t3` tensor to be XLA, the error goes away. Thanks for checking!

miladm (approved):
LGTM. Thanks.

miladm:
UX question:
Turns out we got this `RuntimeError: SymIntArrayRef expected to contain only concrete integers` when developers don't have their tensors on the right device. Wondering if we can improve the error message quality to improve developer experience - before we hit [this line](https://github.com/pytorch/pytorch/blob/master/c10/core/SymIntArrayRef.h#L33-L36). @wconstab wdyt?

wconstab:
> UX question:
Turns out we get this RuntimeError: SymIntArrayRef expected to contain only concrete integers when developers don't have their tensors on the right device. Wondering if we can improve the error message quality to improve developer experience - before we hit [this line](https://github.com/pytorch/pytorch/blob/master/c10/core/SymIntArrayRef.h#L33-L36). @wconstab wdyt?

What did you have in mind @miladm?  That runtime check you linked is really general- someone asked to convert a symint arrayref to an intarrayref, and we're safety checking that none of them are indeed symbolic (which would make the conversion an error).  This conversion gets called from all over.  It's not guaranteed that the issue is due to not using the XLA device.

ezyang:
this PR shoudl help https://github.com/pytorch/pytorch/pull/90675 someone please review it lol

ezyang:
this PR shoudl help https://github.com/pytorch/pytorch/pull/90675 someone please review it lol

miladm:
Awesome! Thanks @ezyang.

PR #4292, opened 2022-12-07
Sharding annotation for empty node

yeounoh:
This PR includes the following changes:
- Check & skip sharding annotation if the same sharding already exists.
- Create a tensor node or re-use the current IR in case of empty IR node.


steventk-g (approved):

torch_xla/csrc/xla_sharding_util.h
```diff
@@ -18,6 +18,10 @@ class ShardingUtil {
   // building the computation; otherwise, this is a no-op.
   static bool SetHloSharding(LoweringContext* lowering_ctx);
 
+  // Returns true if two sharding specs are the same.
+  static bool EqualShardingSpecs(const XLATensor::ShardingSpec& a,
```
JackCaoG, line 22: nit, `ShardingSpecsAreEqual`

torch_xla/csrc/tensor.cpp
```diff
@@ -228,17 +228,23 @@ XLATensor::ShardingSpecPtr XLATensor::sharding_spec() const {
...
 
-void XLATensor::SetShardingSpec(const ShardingSpec& sharding_spec) {
-  XLA_CHECK(GetIrValue().node != nullptr) << "Tyring to access a null cursor";
-  dynamic_cast<XlaNode*>(GetIrValue().node.get())
-      ->SetSharding(sharding_spec.sharding);
+void XLATensor::SetShardingSpec(const ShardingSpec& sharding) {
+  // Existing annotation must be cleared explicitly. We do not clear and
+  // overwrite the existing sharding on the user's behalf. This is a no-op if
+  // the same sharding already applied.
+  if (sharding_spec() == nullptr ||
```
JackCaoG, line 235: shouldn't we throw an error or log something if we call `SetShardingSpec` on a tensor with different sharding?

JackCaoG: silently no-op seems dangerous.

PR #4293, opened 2022-12-07
api to grab running seed as device data

shunting314:
First of all, I'd like to reflect my understanding of how random number is used in torchxla:

When tracing modules that uses random numbers like DropOut, torchxla will need call `XLAGraphExecutor::GetRngSeed` to grab the current seed and put the seed into the traced graph. The API has a side effect of updating some DeviceContext object so next time when it's called, the next (different) seed is returned.  Here is an example of the bernoulli op used by Dropout: https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor_methods.cpp#L782 . In this example, we can see that the current seed is grabbed as an IR node and used to create the Bernoulli IR node.

In our dynamo/torchxla bridge we reused the traced graph. To make sure we can maintain the numerical correctness (for benchmarking), the bridge should have the same behavior as regular torchxla. This PR change `GetRngSeedData` method to do the same thing as `GetRngSeed` but guarantee to return a DeviceData node. In the bridge whenever we need prepare a graph input which is identified as a seed, we should call `GetRngSeedData`.

### DIFF
--- a/torch_xla/csrc/xla_graph_executor.cpp
+++ b/torch_xla/csrc/xla_graph_executor.cpp
@@ -339,6 +339,20 @@ class DeviceContextArena {
     return devctx->seed_ir_value;
   }
 
+  torch::lazy::BackendDataPtr GetBaseSeedData(
+      const torch::lazy::BackendDevice& device) {
+    static const at::ScalarType kSeedType = at::ScalarType::Long;
+    DeviceContext* devctx = GetDeviceContext(device);
+    std::lock_guard<std::mutex> lock(devctx->lock);
+    at::Tensor tensor = at::scalar_tensor(MakeIntScalar(devctx->seed),
+                                          at::TensorOptions(kSeedType));
+    torch::lazy::BackendDataPtr device_data = TensorToXlaData(tensor, device);
+    devctx->seed_ir_value = torch::lazy::MakeNode<DeviceData>(device_data);
+    devctx->running_seed = devctx->seed;
+    return torch_xla::DeviceData::Cast(devctx->seed_ir_value.node.get())
+        ->data();
+  }
+
   uint64_t GetRunningSeed(const torch::lazy::BackendDevice& device) {
     DeviceContext* devctx = GetDeviceContext(device);
     std::lock_guard<std::mutex> lock(devctx->lock);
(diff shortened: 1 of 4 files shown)


shunting314:
I confirmed the fix on all the 9 models we are looking (exclude squeezenet since it fall back on avg_pool2d with a workaroud for the max_pool2d issue.)

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -354,21 +354,21 @@ class DeviceContextArena {
...
-      at::Tensor tensor = at::scalar_tensor(MakeIntScalar(devctx->seed),
-                                            at::TensorOptions(kSeedType));
-      torch::lazy::BackendDataPtr device_data = TensorToXlaData(tensor, device);
-      devctx->seed_ir_value = torch::lazy::MakeNode<DeviceData>(device_data);
-    } else if (!devctx->seed_ir_value) {
-      devctx->seed_ir_value =
-          IrValueFromScalar(MakeIntScalar(devctx->seed), kSeedType, device);
-    }
+
+    devctx->running_seed = kSeedAdd + kSeedMul * devctx->running_seed;
```
JackCaoG, line 366: we force the running seed and real seed divergeg here, which sounds like a bad thing..

shunting314: This is to be equivalent to what regular torchxla does. In regular torchxla, when a seed is needed, it calls [XLAGraphExecutor::GetRngSeed](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_graph_executor.cpp#L319) to grab the seed IR node. We need do the same thing for the dynamo/torchxla bridge. The only difference is we need a DeviceData node for the bridge.

Increment the running seed so that the next call will get a different seed. This also what `XLAGraphExecutor::GetRngSeed` does.

Let me know if there is more clarification needed

JackCaoG: In the `GetRngSeed`, we incremented both `devctx->running_seed` and `devctx->seed_ir_value`. However in this function we only increment `devctx->running_seed`.

Imagine a user call this `GetRngSeedData` function multiple times, it will always get the same result because `devctx->seed` is not being updated. 

If the purpose of this function is to get the current base seed without incremented it(since we expect `mark_step` to do so) , we should not touch the `running_seed` and rename this function to `GetBaseRngSeedData`. We also should overwrite `devctx->seed_ir_value` but just return the base seed as an data ptr

shunting314: I got what you mean. I think the issue is at line 368, we should use devctx->running_seed rather than devctx->seed. Will send an update and some test case soon.

JackCaoG (approved):
Thanks!

PR #4294, opened 2022-12-07
[LTC] Reuse some methods from LazyTensor

alanwaketan:
Summary:
This patch tweaks XLATensor class to use some identical methods from LazyTensor. Here are a few things worth noting:
1. It renames CurrentXlaData() to CurrentDataHandle() such that we can reuse the latter.
2. It replaces data_ptr() with data() which now returns a const shared_ptr<Data>& type.
3. It tweaks the XLATensor::Data class to inherits from LazyTensor::Data. In order to provide accesses to this core data member, here we store two shared_ptrs. One in the LazyTensor class as the LazyTensor::Data type for base class method to access. One in the XLATensor class as the derived XLATensor::Data type such that it's easier to access XLA extra fields within.
4. Methods being removed from XLATensor are: generation(), alias(), GetDevice(), GetUniqueId(), SetTensorData(), and GetNextTensorId().

Test Plan:
CI.


torch_xla/csrc/tensor.cpp
```diff
@@ -120,14 +120,15 @@ XLATensor::XLATensor(std::shared_ptr<View> view,
...
+    : torch::lazy::LazyTensor(data),
+      data_(std::move(data)),
       storage_(c10::Storage(
           {}, 0,
           c10::DataPtr(nullptr, backendDeviceToAtenDevice(data_->device)))) {}
 
-XLATensor::Data* XLATensor::data() const {
-  XLA_CHECK(data_ != nullptr) << "Trying to access a null cursor";
-  return data_.get();
+auto XLATensor::data() const -> const std::shared_ptr<Data>& {
```
wonjoolee95, line 129: nit: just curious why function declaration is used as opposed to `const std::shared_ptr<Data>& 
XLATensor::data() const`?

wonjoolee95 (approved):
Thanks! Awesome 👍

torch_xla/csrc/tensor.cpp
```diff
@@ -120,14 +120,15 @@ XLATensor::XLATensor(std::shared_ptr<View> view,
...
+  TORCH_CHECK(data_ != nullptr, "Trying to access a null cursor");
```
JackCaoG, line 130: I think we should use `XLA_CHECK` within our repo. Unless we want to replace all `XLA_CHECK` with `TORCH_CHECK`

torch_xla/csrc/tensor.cpp
alanwaketan: The latter case you then have to specify `XLATensor::Data` instead of just using `Data`. `auto` is shorter than `XLATensor::`, and shorter is better. hahaha.

torch_xla/csrc/tensor.cpp
alanwaketan: Good catch. Let me fix that.

torch_xla/csrc/tensor.h
```diff
@@ -270,12 +242,15 @@ class XLATensor : public torch::lazy::LazyTensor {
...
-
   static bool UseEagerDebugMode();
 
   bool ShouldSyncIrNode();
 
+  // We store two shared_ptr of Data in a XLATensor.
+  // One in the LazyTensor class as the LazyTensor::Data type
+  // for base class method to access. One here as the derived
+  // XLATensor::Data type such that it's easier to access XLA
+  // extra fields.
```
JackCaoG, line 253: shouldn't `XLATensor::Data` inherits `torch::lazy::Data` and we only need to access `XLATensor::Data`?

alanwaketan: Base class has methods like GetUniqueId() which will need to access data_, and therefore we need to set the data_ field there. However, we also need to provide a data() accessor that returns a const reference to the data_ field but with XLATensor::Data type. That's why we also need to store another pointer within XLATensor.

JackCaoG: I guess after we remove the view, `logical_element_type` is the only thing unique in our `Data`?

JackCaoG: OK so, it seems like we only create one `Data` object which is a `XLATensor::Data` but we have two pointers of different type point to this object. Is this a fair understanding?

alanwaketan: That's correct!

alanwaketan:
Thanks, Jack and Wonjoo.

PR #4297, opened 2022-12-07
Support multiple GPUs on a single host with PjRt

jonb377:


JackCaoG:
@ymwangg I think PJRT is a much nicer runtime than XRT and we should transition to PJRT on GPU.

jonb377:
@JackCaoG @will-cromar Please hold on the review for this, I want to test one more configuration to see if we can avoid using the distributed runtime client for single-host. I'll reopen the PR when it's ready.

jonb377:
Replication groups are still not working (e.g. in `test_mp_all_gather.py`, only one group completes while the other remains blocked).

docs/pjrt.md
```diff
@@ -86,7 +86,24 @@ for more information.
...
+```
+PJRT_DEVICE=GPU GPU_NUM_DEVICES=4 python3 xla/test/test_train_mp_imagenet.py --fake_data --batch_size=128 --num_epochs=1
+```
+
+Currently, only a single host is supported, and multi-host GPU cluster support
+will be added in an future release.
+
+#### Known Issues
+
+The GPU integration has issues with replica groups. If the replica groups
```
will-cromar, line 103: Can you clarify that you mean the `groups` arguments of XLA collective ops?

will-cromar (approved):
Great work! Thanks!

torch_xla/experimental/gpu.py
```diff
@@ -0,0 +1,43 @@
...
+    # TODO(jonbolin): For multi-host, this needs to be consistent across hosts
+    os.environ.setdefault(xenv.PJRT_DIST_SERVICE_ADDR, '127.0.0.1:8547')
+    global distributed_service
+    if distributed_service is None:
+      num_nodes = global_world_size
+      distributed_service = torch_xla._XLAC._xla_get_distributed_runtime_service(
+          num_nodes)
+
+
+def shutdown_distributed_runtime() -> None:
```
steventk-g, line 38: What happens if we crash before shutting down the runtime?

steventk-g (approved):

JackCaoG:
yay, we should use PJRT:GPU in CI instead of XRT. I don't see any reason for people to use XRT:GPU instead of PJRT:GPU now.

will-cromar:
Should we wait until the issue with replica groups is fixed before switching the CI?

JackCaoG:
yea if there is a feature gap let's wait

torch_xla/experimental/gpu.py
jonb377: By default, the runtime is shut down in the destructor, but since we need to recreate the service for each `xmp.spawn` call, I decided to make an explicit `shutdown` call here instead of relying on the garbage collector to avoid potential race conditions.

In a crash of the main Python process (which hosts the DistributedRuntimeService), the individual per-device subprocesses should be terminated as children.

If one of the subprocesses crashes, the server will become aware of the crash through missed heartbeats. This is actually what happens with the current collectives issue, since the subprocesses crash after the replica groups are changed. In that case, I've observed the Python process will hang for a bit before terminating with a BrokenProcessPool exception in Python.

steventk-g: Thanks for explanation Jon!

PR #4307, opened 2022-12-09
[LTC] Inherits LazyGraphExecutor::DeviceContextArena

alanwaketan:
Summary:
This patch inherits LazyGraphExecutor::DeviceContextArena and overrides the follow methods:
1. GetLiveTensors to return XLATensorPtrs.
2. GetRngSeed to use our own + and * for torch::lazy::Value.
3. IrValueFromScalar to use TensorToXlaData.

In addition, it has an extra method: GetBaseSeedData that is used by dynamo.

This patch needs pytorch/pytorch#90531 to function.

Test Plan:
CI.


wonjoolee95 (approved):

alanwaketan:
Thanks Jack and Wonjoo.

PR #4311, opened 2022-12-10
Improve numerical stability of torch.sigmoid

ymwangg:
We recently found the torch_xla lowering of `torch.sigmoid` is not numerically stable on GPU. One common use-case of `torch.sigmoid` is to force the output value to be within [0,1].
For example, the following code failed with nan loss because `x = -5.9604645e-08`.
```py
x = torch.sigmoid(torch.tensor([-16.740633],device=device))
y = torch.tensor([1.0],device=device)
print(torch.nn.functional.binary_cross_entropy(x,y)) # print tensor(nan, device='xla:1')
```
Are there any special reasons for torch_xla to use `sigmoid(x) = 0.5+0.5*tanh(0.5*x)` instead of `sigmoid(x) = 1 / (1 + exp(-x))`?

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1883,6 +1883,16 @@ class TestAtenXlaTensor(XlaTestCase):
 
     self.runAtenTest([torch.arange(15, dtype=torch.int32)], test_fn)
 
+  def test_sigmoid_bounds(self):
+    torch.manual_seed(0)
+    xla_device = xm.xla_device()
+    for _ in range(100):
+      x = torch.rand(1000).to(xla_device)
+      lower_bound = torch.sigmoid(x * (-100.0))
+      upper_bound = torch.sigmoid(x * (100.0))
+      assert torch.all(lower_bound >= 0.0)
+      assert torch.all(upper_bound <= 1.0)
+
 
 class MNISTComparator(nn.Module):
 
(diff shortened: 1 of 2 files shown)


ymwangg:
Ok it looks this issue exists on both cpu and gpu:
cpu: https://app.circleci.com/pipelines/github/pytorch/xla/14953/workflows/946bc70e-a4ba-4e89-be01-0d7ea7f653cb/jobs/34704
gpu: https://app.circleci.com/pipelines/github/pytorch/xla/14953/workflows/946bc70e-a4ba-4e89-be01-0d7ea7f653cb/jobs/34703

JackCaoG:
I have a feeling that it might be because `sigmoid(x) = 0.5+0.5*tanh(0.5*x)` is faster.. let me double check.

ymwangg:
Yes, the `tanh` implementation is slightly faster on GPU.
Using the following script:
```py
x = torch.rand(1000000000,device=device)
xm.mark_step()
t0 = time.time()
for _ in range(100):
    for _ in range(100):
        y = torch.sigmoid(x)
    xm.mark_step()
t1 = time.time()
print(t1-t0)
```
I'm getting `1.2621409893035889` with `tanh` implementation (with clamp) and `1.301847219467163` with normal implementation.

If we want to keep the `tanh` implementation, one way is to wrap it with `xla::Clamp(zero, half + half * xla::Tanh(half * input), one)`.

JackCaoG:
I talked with Blake. Speed was the main reason we used `tanh` and TPU does not have this numerical instability issue. He suggested us to lower `sigmod` using `XlaOp Logistic(XlaOp operand);` which will have different TPU and GPU implementation in the backend to handle the subtle difference in accelerators.

ymwangg:
Updated and thanks for the info. I just realize `xla::Logistic` is equivalent to `torch.sigmoid`.

JackCaoG (approved):
Thanks!

PR #4314, opened 2022-12-11
[LTC] Adopt more LazyGraphExecutor virtual interfaces

alanwaketan:
Summary:
This patch tries to adopt even more LazyGraphExecutor virtual interfaces:
1. LazyGraphExecutor::Async.
2. TensorCollectionBarrier
3. SyncLiveTensorsGraph

And comments on methods that we don't adopt.

This depends on https://github.com/pytorch/pytorch/pull/90650.

Test Plan:
CI.


torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -567,14 +541,8 @@ XLAGraphExecutor::SyncTensorCollection XLAGraphExecutor::CollectSyncTensors(
 }
 
 void XLAGraphExecutor::TensorCollectionBarrier(SyncTensorCollection* coll) {
-  static const std::string invalid_device(
-      "Unknown0"); /* Temp solution to identify unassigned devices */
-  if (coll->device.toString().compare(invalid_device) == 0 ||
-      coll->unlocker.size() > 0) {
-    return;
-  }
+  torch::lazy::LazyGraphExecutor::TensorCollectionBarrier(coll);
```
JackCaoG, line 544: hmm so the reason we need to overwrite this function is for future SPMD use case?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -956,8 +924,7 @@ XLAGraphExecutor::PostOrderData XLAGraphExecutor::RunPostOrder(
 }
 
 XLAGraphExecutor::ComputationCache::TypePtr
-XLAGraphExecutor::LookupCachedCompile(const std::vector<XLATensorPtr>& tensors,
```
JackCaoG, line 959: haha, I guess at some point we dropped the use of `tensors` without noticing it.

JackCaoG (approved):

torch_xla/csrc/xla_graph_executor.cpp
alanwaketan: Yea, or I can add the todo to the upstream. 😂

PR #4316, opened 2022-12-12
Add experimental test coverage feature

yeounoh:
PoC Python test coverage report to CIrcleCI pipeline, limited to python tests.
- Example [report](https://output.circle-artifacts.com/output/job/fe08da23-1406-4193-bf7c-88fdc41d7cd0/artifacts/0/htmlcov/index.html) in circleci, with a select few test cases. This is also available in json.

I've extended the coverage to most of Python tests, we can add more tests later. I would keep the coverage script outside the main `run_test.sh` since we may want to run tests without `coverage` locally.

We can follow up with an investigation with `ng3-metrics` or to push this to GCS or Spanner cc @miladm 

cc @steventk-g would this `test/utils/run_test_coverage.sh` work for your case, TPU CI with GHA?


vanbasten23:
very cool! How does the eventual test coverage report look like?

yeounoh:
> very cool! How does the eventual test coverage report look like?
Here is an example,
https://output.circle-artifacts.com/output/job/fe08da23-1406-4193-bf7c-88fdc41d7cd0/artifacts/0/htmlcov/index.html

yeounoh:
I am thinking about enabling test coverage run only for master branch. I will remove `ci_test_coverage` from whitelist before landing.
```
filters:
    branches:
      only:
        - master
        - ci_test_coverage
 ```

File: test/utils/run_test_coverage.sh [hunk 1]
```diff
@@ -0,0 +1,132 @@
+#!/bin/bash
```
steventk-g, line 1: This should integrate seamlessly with GHA. Thanks Yeounoh!

steventk-g (approved):

steventk-g:
Just wondering - it looks like the reports generated give information about the test files themselves. Is there information about the source code coverage % breakdown by file and line?

test/utils/run_test_coverage.sh
```diff
@@ -0,0 +1,132 @@
...
+set -exo pipefail
+CDIR="$(cd "$(dirname "$0")/.." ; pwd -P)"
+VERBOSITY=2
+
+export TORCH_TEST_DEVICES="$CDIR/pytorch_test_base.py"
+export CPU_NUM_DEVICES=4
+
+function run_test {
+  coverage run -p "$@"
```
steventk-g, line 10: I think we need to direct coverage to the `--source` here

steventk-g: https://stackoverflow.com/questions/67980243/coverage-py-only-shows-test-files

yeounoh: Thanks for poiting this out, pointing it to `torch_xla/` let's see what it does.

yeounoh: It didn't produce something meaningful (0% coverage), reverting -- let's investigate different types of coverage in the next round.

PR #4321, opened 2022-12-13
Specified vague "Any" type definitions

gkroiz:


torch_xla/distributed/fsdp/xla_flatten_params_wrapper.py
```diff
@@ -440,14 +444,15 @@ def unflatten_params(self,
...
     """Forward missing attributes to wrapped module."""
     try:
       return super().__getattr__(name)  # defer to nn.Module's logic
     except AttributeError:
       return getattr(self.module, name)  # fallback to wrapped module
-
-  def __getitem__(self, key: int) -> Any:
+      
+  T = TypeVar('T', bound=Module)
+  def __getitem__(self, key: int) -> Union[nn.Sequential, T]:
```
ronghanghu, line 455: I think one would need to add `from typing import TypeVar` in this file?

Also, this method is just to return the `nn.Module` instances in case the base module is a `nn.Sequential`, so I think we can just use `def __getitem__(self, key: int) -> nn.Module:` as the signature here.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -652,14 +652,15 @@ def extra_repr(self) -> str:
...
+  def __getattr__(self, name: str) -> Union[Tensor, nn.Module]:
     """Forward missing attributes to wrapped module."""
     try:
       return super().__getattr__(name)  # defer to nn.Module's logic
     except AttributeError:
       return getattr(self.module, name)
 
-  def __getitem__(self, key: int) -> Any:
+  T = TypeVar('T', bound=Module)
+  def __getitem__(self, key: int) -> Union[nn.Sequential, T]:
```
ronghanghu, line 663: same here, I think we can just use `def __getitem__(self, key: int) -> nn.Module:` as the signature here.

ronghanghu (commented):
Thanks! Apart from the two comments above, everything looks good to me

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
gkroiz: Changed

torch_xla/distributed/fsdp/xla_flatten_params_wrapper.py
gkroiz: Changed

gkroiz (commented):
Changed function return type for `__getitem__()`

JackCaoG:
please run the linter per https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md#python-style-guide

ronghanghu (approved):

PR #4289, opened 2022-12-06
[stale] Test backward pass nn model with dynamic input

vanbasten23:

PR #4322, opened 2022-12-13
Add dynamic shape support to sigmoidbackward

vanbasten23:
I messed up a previous [branch](https://github.com/pytorch/xla/pull/4289). So I'll just copy the a few comments here.


vanbasten23:
Succeeds when we use dynamic test data but static training data: https://github.com/pytorch/xla/commit/f82efbcf7058d589377b751891d6938ca3b2d4e5. It outputs:
```
Finished training. Got loss: 0.686253547668457
Finished testing, loss= 0.6257358193397522
```

vanbasten23:
Once I made the test data dynamic (https://github.com/pytorch/xla/commit/0048f3b0da9c7b1f6760269f767ac87883db188d), the test failed with error:
```
root@t1v-n-2a2b95ef-w-0:/workspaces/work# python3 pytorch/xla/test/test_dynamic_shape_models.py
  File "pytorch/xla/test/test_dynamic_shape_models.py", line 78, in <module>
    train(model, loss_fn=criterion, optimizer=optimizer)
  File "pytorch/xla/test/test_dynamic_shape_models.py", line 65, in train
    loss.backward()
  File "/home/ptxla/.local/lib/python3.8/site-packages/torch/_tensor.py", line 484, in backward
    torch.autograd.backward(
  File "/home/ptxla/.local/lib/python3.8/site-packages/torch/autograd/__init__.py", line 197, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Function SigmoidBackward0 returned an invalid gradient at index 0 - got [80, 1] but expected shape compatible with [<=80, 1]
```
which I'm going to look into.

vanbasten23:
From milad: 
Looks like sigmoid_bakwards needs to support dynamism. Wdyt?

vanbasten23:
Yeah, Also, for the error RuntimeError: Function SigmoidBackward0 returned an invalid gradient at index 0 - got [80, 1] but expected shape compatible with [<=80, 1] , it failed at https://github.com/pytorch/pytorch/blob/912a1f7b2776c0e7ebf9038e4483a4aa709aa893/torch/csrc/autograd/engine.cpp#L812. Stacktrace: https://gist.github.com/vanbasten23/a68180922e9f4c554b92365c961c21a4

vanbasten23:
Milad: We have the stack trace pointing to the pytorch/torch/csrc/autograd/engine.cpp error. See the previous comment. @wconstab @ezyang we are wondering if the autograd engine is missing dynamism support. Wdyt?

vanbasten23:
Ed: This error is not one I've seen before. At a guess, CLA's bounded symints don't correctly implement the equality/comparison operators and a test AD engine is doing is failing. If you log symint ops should become clear

wconstab:
so, the way i would debug this is to stick some print statements into [this code in engine.cpp](https://github.com/pytorch/pytorch/blob/912a1f7b2776c0e7ebf9038e4483a4aa709aa893/torch/csrc/autograd/engine.cpp#L812) 

you're trying to figure out why 
```
    if (!metadata.is_same_shape(grad)) {
      if (metadata.is_expandable_to_shape(grad)) {
 ```
 are failing, and it'd be good to understand both (a) how `is_same_shape` works when it gets sym vs non-sym shapes input, and (b) there are shapes stored both in 'metadata' and 'grad' which are expected to match: can you reason about whether one or both is correct, or if one is missing symint?

finally, if the 2 copies of shapes looks right, there should be some calls to symint::operator(something) made by `is_same_shape` - maybe the XLA version of symint is missing support or incorrect for one of those calls (I think that's what ed was suggesting)

**Comment by vanbasten23** [comment]
I did some digging and also added print statement in the XLASymNode's virtual methods by following Will and Ed's suggestion.

1) First, I think the error message I got `got [80, 1] but expected shape compatible with [<=80, 1]` may be incorrect and it should be the other way around: `got [<=80, 1] but expected shape compatible with [80, 1]`.

It's because when we check the shape compatibility https://github.com/pytorch/pytorch/blob/cdf4a80cc111b210f9ab9448da5aeea2007a0171/torch/csrc/autograd/input_metadata.h#L109, the `grad.sym_size()` is the desired size. But when we construct the error message https://github.com/pytorch/pytorch/blob/cdf4a80cc111b210f9ab9448da5aeea2007a0171/torch/csrc/autograd/input_metadata.h#L128, it treat `grad.sym_size()` as the actual size.

2) IIUC, the real issue happens in `is_expandable_to` https://github.com/pytorch/pytorch/blob/cdf4a80cc111b210f9ab9448da5aeea2007a0171/aten/src/ATen/ExpandUtils.h#L496-L512, here I can tell `shape` is [<=80, 1] and `desired` is [80, 1], with some print statements. So when it checks the equality of the 0th element https://github.com/pytorch/pytorch/blob/cdf4a80cc111b210f9ab9448da5aeea2007a0171/aten/src/ATen/ExpandUtils.h#L507, `<=80` does not equal `80`. Coming to pytorch/XLA, when we compare 2 SymNodeImpl, XLASymNodeImpl does implement the `eq` method: https://github.com/pytorch/xla/blob/2fd2ffeeb303245ccab76cd5287d9818565aaad2/torch_xla/csrc/tensor.cpp#L612. Per my understanding, `size != target` is true as long as they don't have the same dynamic value.

So to fix, when we check expandable https://github.com/pytorch/pytorch/blob/cdf4a80cc111b210f9ab9448da5aeea2007a0171/aten/src/ATen/ExpandUtils.h#L507, should we do `size > target && size != 1` instead. WDYT? @wconstab @ezyang

**Comment by ezyang** [comment]
I agree with diagnosis but not the proposed fix. Can we introduce a new operator which denotes the semantic test we want to do here? What exactly is the test we are doing

**Comment by miladm** [comment]
Here we face a `ne` comparison operator. I don't see an implementation for `XLASymNodeImpl::ne` in [dynamic_ir](https://github.com/pytorch/xla/blob/6a792ddc87525075d8e4cea076bef70e1b8761a4/torch_xla/csrc/ops/dynamic_ir.cpp). At the same time, if `ne` were to be the source of failure, I would've expected an error message from this [code block](https://github.com/pytorch/pytorch/blob/master/c10/core/SymNodeImpl.h#L55-L57). Regardless, it's relevant we implement `ne` in XLA as a stepping stone.

@ezyang let me know if you have a different idea in mind when referring to "operator" above.

**Comment by miladm** [comment]
https://github.com/pytorch/xla/pull/4338

**Comment by miladm** [comment]
On a slightly tangential topic: DS breaks the graph at conditional operations. The expectation is not to have conditional ops on the critical path of the compiler (FWD+BKWD). What are your thoughts on finding a solution to avoid running into such scenarios in the compiler?

@wconstab @ezyang

**Comment by wconstab** [comment]
> DS breaks the graph at conditional operations

this is probably worth breaking out into another issue and getting more specific about the type of conditional and the type of symint.  Your options also depend on whether you're factoring in Dynamo or not.

> The expectation is not to have conditional ops on the critical path of the compiler (FWD+BKWD)

Not sure what you mean here.  Are you expecting to avoid these operators in model code, or find a solution that avoids graph-breaking when they are encountered?

**Comment by miladm** [comment]
Yup, happy to create a separate issue to carry the discussion. 

> this is probably worth breaking out into another issue and getting more specific about the type of conditional and the type of symint. Your options also depend on whether you're factoring in Dynamo or not.

+ IIUC, dynamo may reduce the number of conditionals on a given code path. Though, when any conditional is passed down to downstream (either from Dynamo or not), we need to trigger a graph break. Correct me if I am wrong.
+ Can you elaborate how you think the type of conditional plays a role in deciding whether or not a graph breaks occurs?

> Not sure what you mean here. Are you expecting to avoid these operators in model code, or find a solution that avoids graph-breaking when they are encountered?

+ re: model code, the user is responsible for avoiding conditionals in their implementation. The fewer conditionals they introduce, the more optimal their code will be.
+ I am looking to understand if there is a way for us to avoid having PT source code cause a graph breaks (upon conditionals) on commonly executed ops.

**Comment by miladm** [comment]
Update: the `ne` PR is good to have, but doesn't address this issue ([reference](https://github.com/pytorch/pytorch/blob/master/c10/core/SymInt.cpp#L105-L107)).

I'd like to understand why `metadata` is not dynamic. @vanbasten23 can you please take a look?

**Comment by wconstab** [comment]
> Can you elaborate how you think the 'type' of conditional plays a role in deciding whether or not a graph break occurs?

I should have said, the  inputs to the conditional plays a role... -> if the conditional can be evaluated statically, based on information from faketensors or based on constants, we can keep tracing and then build the assumptions into our hash (lazy) or guards (dynamo).

> I am looking to understand if there is a way for us to avoid having PT source code cause a graph breaks (upon conditionals) on commonly executed ops.

Yea, we're making changes to PT source all the time in the symbolic-shapes workstream, trying to make it friendlier to tracing.  Of course, for certain ops like nonzero there isn't much we can do...  But if there is a particular change you want to propose, let us know.

**Comment by ezyang** [comment]
your link 'here' is broken

**Comment by ezyang** [comment]
I think it's reasonable for grad to be dynamic size, though you kind of have a problem which is that `<=79` shouldn't necessarily compare equal to `<=79`; e.g., 68 and 70 would be valid concretizations of these types but they're not equal.

The way we get around this in core is we maintain precise shape variables so we can tell that "s0" is the same as "s0". IDK if that'll work for XLA though

**Comment by vanbasten23** [comment]
> your link 'here' is broken

Sorry, I fix the link. Correct link is https://github.com/pytorch/pytorch/blob/fdc973308bcac5ff3e1c7d91c6d85e5328011653/torch/csrc/autograd/engine.cpp#L807

**Comment by ezyang** [comment]
it's not really clear what it would mean for grad to be dynamic, it's coming from the downstream gradient calculation

**Comment by vanbasten23** [comment]
> I think it's reasonable for grad to be dynamic size, though you kind of have a problem which is that `<=79` shouldn't necessarily compare equal to `<=79`; e.g., 68 and 70 would be valid concretizations of these types but they're not equal.
> 
> The way we get around this in core is we maintain precise shape variables so we can tell that "s0" is the same as "s0". IDK if that'll work for XLA though

Currently, the way that XLA compares equal two `<=79` is by checking the real size (dynamic size) [here](https://github.com/pytorch/xla/blob/45542f7551247636d5358a079055d8d77a0abd53/torch_xla/csrc/ops/dynamic_ir.cpp#L110). In the current case, if `grad` happens to be [<=80, 1] and has dynamic size of 79, then `metadata`(has shape [<=80, 1] with dynamic size [79, 1]) should be expandable to it. So I wonder if dynamism propagate correctly to `grad`. IOW, I wonder if `grad` should be SymInts [<=80, 1] instead of static shape [80, 1].

**Comment by ezyang** [comment]
OK, then yes, it sounds like you need to propagate dynamism further

**Comment by vanbasten23** [comment]
To confirm, [here](https://github.com/pytorch/pytorch/blob/2a37ba8e81604fda5ba78fe5ee8c8662ce0c25f3/torch/csrc/autograd/engine.cpp#L807) 
```
if (!metadata.is_same_shape(grad)) {
```
`grad` is the output which has size [80, 1]. `metadata` (ie the input metadata, has size [<=80, 1]) is the expectation against which we are validating the output. Do you mean the expectation ([<=80, 1]) is correct and we need to propagate the dynamism to the output `grad` so that its size become [<=80, 1]?

**Comment by ezyang** [comment]
It is very difficult to discuss this, because you haven't laid out the exact sequence of operations that execute, and where things mismatch. Let me give a guess, but I may not be right.

What it sounds like is going on here is you have the following operator sequence:

```
y: [<= 80, 1] = dynamic_op(x)
z: [<= 80, 1] = sigmoid(y)
... some other stuff on z ...
```

When we run backwards, you end up with

```
grad_z: [<= 80, 1]
grad_y: [80, 1] = sigmoid_backward(grad_z)
... more stuff ...
```

You are producing a tensor with size [80, 1] from sigmoid backward. This is wrong. You should produce a [<= 80, 1]. 

Does this actually match what you observed, or am I misunderstanding?

**Comment by vanbasten23** [comment]
> It is very difficult to discuss this, because you haven't laid out the exact sequence of operations that execute, and where things mismatch. Let me give a guess, but I may not be right.
> 
> What it sounds like is going on here is you have the following operator sequence:
> 
> ```
> y: [<= 80, 1] = dynamic_op(x)
> z: [<= 80, 1] = sigmoid(y)
> ... some other stuff on z ...
> ```
> 
> When we run backwards, you end up with
> 
> ```
> grad_z: [<= 80, 1]
> grad_y: [80, 1] = sigmoid_backward(grad_z)
> ... more stuff ...
> ```
> 
> You are producing a tensor with size [80, 1] from sigmoid backward. This is wrong. You should produce a [<= 80, 1].
> 
> Does this actually match what you observed, or am I misunderstanding?

Yeah, what you described matches my observation and understanding.

In the [test script](https://github.com/pytorch/xla/pull/4322/files#diff-c108635f3268c766c8167dfef84f8a59345c25277f4294caa9883637a66a054c) in the pr, we generate the training data from a `dynamic_op` so x_train has size [<=80, 2] and y_train has size [<=80]. The NN model doesn't involve any  dynamic_ops.

Then when we run backwards, the output that we got is a tensor with size [80, 1]. So it sounds we need to produce a [<=80, 1] by propagating the dynamism to autograd result.

Add @miladm and @JackCaoG  to correct me if I'm wrong.

**Comment by JackCaoG** [comment]
`when we run backwards, the output that we got is a tensor with size [80, 1]` 

Can we identify where this part is happening?

**Comment by vanbasten23** [comment]
> Can we identify where this part is happening?

It failed when we do `loss.backward()`. Per the error message `RuntimeError: Function SigmoidBackward0 returned an invalid gradient at index 0 - got [80, 1] but expected shape compatible with [<=80, 1]`, it looks `SigmoidBackward0` produces [80, 1]. I can look into it to see how this `SigmoidBackward0` works.

**Comment by JackCaoG** [comment]
that `[80, 1]` must be generated somewhere, most likely some C++ code, let's identify that code path then we know how to fix it.

**Comment by vanbasten23** [comment]
The output [80, 1] is produced [here](https://github.com/pytorch/pytorch/blob/2a37ba8e81604fda5ba78fe5ee8c8662ce0c25f3/torch/csrc/autograd/engine.cpp#L907) by the `fn`. I added some print statement at the above place and printed `fn.name()`:
```
xw32, file=/workspaces/work/pytorch/torch/csrc/autograd/engine.cpp, line=916function=call_function: about to run validate_outputs on fn.name()=BinaryCrossEntropyBackward0
xw32, file=/workspaces/work/pytorch/torch/csrc/autograd/engine.cpp, line=916function=call_function: about to run validate_outputs on fn.name()=SqueezeBackward0
xw32, file=/workspaces/work/pytorch/torch/csrc/autograd/engine.cpp, line=916function=call_function: about to run validate_outputs on fn.name()=SigmoidBackward0
```
IIUC, `fn(std::move(inputs))` would invoke `variable_list SigmoidBackward0::apply(variable_list&& grads)` generated in `/workspaces/work/pytorch/torch/csrc/autograd/generated/Functions.cpp`. I'm having some trouble tracking further..

**Comment by JackCaoG** [comment]
It seems like we lower `sigmod_backward` according to https://github.com/pytorch/xla/blob/master/xla_native_functions.yaml#L283, can you check if our implementation is triggered? If this static tensor is produced by pytorch/xla then it should be relatively easy to fix.

**Comment by JackCaoG** [comment]
also, what's the shape of the `inputs` passed to `fn` ? I assume it is [<=80, 1]

**Comment by vanbasten23** [comment]
> It seems like we lower sigmod_backward according to https://github.com/pytorch/xla/blob/master/xla_native_functions.yaml#L283, can you check if our implementation is triggered?

Indeed, our `sigmoid_backward ` function is triggered. How do you know it's our `sigmoid_backward` that will be triggered, as I found multiple implementation of `sigmoid_backward` on pytorch side?

> also, what's the shape of the inputs passed to fn ? I assume it is [<=80, 1]

`inputs` is a vector of tensors. Actually, how can I know if a tensor is dynamic? So I print both `sizes()` and `sym_sizes()` via
```
for (int i=0; i<inputs.size(); i++) {
    std::cerr << "xw32, file=" << __FILE__ << ", line=" << __LINE__ << "function=" << __FUNCTION__ << ": inputs " << i << ":" << inputs[i].sizes() << ", inputs[i].sym_sizes()="<< inputs[i].sym_sizes() << std::endl;
  }
```
and I got `xw32, file=/workspaces/work/pytorch/torch/csrc/autograd/engine.cpp, line=893function=call_function: inputs 0:[80, 1], inputs[i].sym_sizes()=[<=80, 1]`.

**Comment by ezyang** [comment]
Oh, so this could be there is an API calling sizes() when it should be calling sym_sizes(). In PT2 dynamic shapes we make it a hard error to call sizes() when there are symbolic shapes involved, but it kind of looks like you're not doing that

**Comment by JackCaoG** [comment]
lowering of sigmoidbackward is in 
https://github.com/pytorch/xla/blob/68491026a6b32ec0b6f069063e0fe3499b9bf66d/torch_xla/csrc/ops/ops.cpp#L181-L184

Check the input shapes and output shapes here. We might lost the dynamism in this operation somehow.

**Comment by JackCaoG** [comment]
`SigmoidBackward` is also a special case where we don't have real lowering but reuse a bunch of IR level computation, I wonder if that has any implication to outputshape lost the dynamism(if that's the case). Anyway, print everything here you should have some ideas.

**Comment by vanbasten23** [comment]
Yeah, I'm already looking at how the output shape is generated. https://github.com/pytorch/xla/blob/3edc88f158616bf10d922aca04016c58f65a2aba/torch_xla/csrc/tensor_methods.cpp#L2204-L2208.
`grad_output->shape()` and `output->shape()` gave me `f32[<=80,1]{1,0}` and `f32[<=80,1]{1,0}` respectively.

Digging deeper, my current hypothesis is `return grad_output * (ScalarOp(1, GetXlaShape(output)) - output) * output` would invoke https://github.com/pytorch/pytorch/blob/8b617f813d86c348be368a72170ab0d319308b23/torch/csrc/lazy/core/ops/arithmetic_ir_ops.cpp#L31-L36 and maybe we need to make `GetPromotedBinaryOpShape` dynamic. But once I added some print statement in `NodePtr operator*(const Value& node1, const Value& node2)`, I couldn't see my print statements.

**Comment by vanbasten23** [comment]
Ok, `return grad_output * (ScalarOp(1, GetXlaShape(output)) - output) * output` would invoke https://github.com/pytorch/xla/blob/fccbc6a287dcb03ba3b1c7f5f61b0ef4aa2cf575/torch_xla/csrc/ops/arithmetic_ir_ops.cpp#L39-L51. The `XlaHelpers::GetPromotedBinaryOpShape` seems to determine the output shape. Looking inside https://github.com/pytorch/xla/blob/fccbc6a287dcb03ba3b1c7f5f61b0ef4aa2cf575/torch_xla/csrc/helpers.cpp#L481-L488, it looks we are treating both `shape1` and `shape2` as static tensor.

**Comment by JackCaoG** [comment]
yea, I guess we need to make `GetPromotedBinaryOpShape` understand dynamism.

**Comment by JackCaoG** [comment]
ok let's do this, instead of using `GetPromotedBinaryOpShape`, let's do something similar to 
https://github.com/pytorch/xla/blob/c4bc972be1bb77d3bfb921ab78f4922af343220c/torch_xla/csrc/ops/ops.cpp#L259-L277

which will reuse the lower function to get the xla:shape of the output

**Comment by vanbasten23** [comment]
Oops, didn't see your last message. I modified `GetPromotedBinaryOpShape` as in this naive [pytorch pr](https://github.com/pytorch/pytorch/pull/91286) and changed how pt/xla uses it in torch_xla/csrc/helpers.cpp (I uploaded the change in this pr ) and the script passed the failure.

Your suggestion looks cleaner. Let me try it. Thanks.

**Comment by vanbasten23** [comment]
@JackCaoG I tried your approach and it worked: the error goes away and is replaced with a new error. But to confirm, your proposed fix works because `InferOutputShape` uses `XlaHelpers::ShapeOfXlaOp(result)` internally and  `XlaHelpers::ShapeOfXlaOp(result)` will take dynamism into consideration, right?

**Comment by vanbasten23** [comment]
Update:

Recap: for the original SigmoidBackward error `RuntimeError: Function SigmoidBackward0 returned an invalid gradient at index 0 - got [80, 1] but expected shape compatible with [<=80, 1]`, I tried Jack's [suggestion](https://github.com/pytorch/xla/pull/4322#issuecomment-1362205330) and I think it fixed the SigmoidBackward error and we are now having another [transpose error](https://gist.github.com/vanbasten23/873b44d3b798d287a1a9378d3b09c9c5).

What I am trying to do is to create a PR and fix SigmoidBackward error but not the transpose error. I'll create another github issue/pr for the transpose error later. So, right now I'm having some trouble creating a failing test with only SigmoidBackward error but not transpose error: the test will take a dynamic tensor, call SigmoidBackward, then check if the result is also dynamic.

I have 2 options: writing python test or c++ test. 

Python tests: I have 2 [python tests](https://gist.github.com/vanbasten23/d1487e6f8ca5c86a08a31f54e9de6e8a) that could reproduce the SigmoidBackward error. The problem with these 2 tests is that both test not only reproduce the SigmoidBackward error but also reproduce the transpose error. Removing the `torch.nn.Linear` layer didn't help ([test](https://gist.github.com/vanbasten23/fa872b6c8ec3b53da6a90efe81d1aded)) and it failed with error `RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn`.

C++ test: By following the [existing backward test](https://github.com/pytorch/xla/blob/40127959f376a7a35a4734a2c295d44c58943848/test/cpp/test_aten_xla_tensor.cpp#L6215) in pytorch/xla, I came up with [mine](https://gist.github.com/vanbasten23/46b0d9101263642438c43f0ca68dd244). But it failed with [error](https://gist.github.com/vanbasten23/dc98e154926825ce19cd6a2fc0f24a69) `C++ exception with description "element 0 of tensors does not require grad and does not have a grad_fn`. I have no luck fixing it so far. If I remove the `nonzero`, I don't get the error.

**Comment by vanbasten23** [comment]
To unblock me, I think I can rely on the existing [SigmoidBackward c++ test](https://github.com/pytorch/xla/blob/40127959f376a7a35a4734a2c295d44c58943848/test/cpp/test_aten_xla_tensor.cpp#L6215) and merge the fix for SidmoidBackward. Then I'll focus on the next transpose error. After that, I'll add my python test to provide more test coverage. Wdyt? @miladm @JackCaoG

**Comment by JackCaoG** [comment]
existing sigmoid test can make sure you don;'t introudce regression to static shape test. As we discussed offline testing `nonzero` + `sigmoid` on C++ test has some issues. I think it is OK to skip the unit test this time but adding a linear model tests that would fail without this fix(after you fixed other linear model failures).

File: test/test_dynamic_shape_models.py [hunk 1]
```diff
@@ -85,6 +85,38 @@ def test_forward_pass_dynamic_input_compile_once(self):
...
+
+    # testing
+    model.eval()
+    with torch.no_grad():
+      x_test, y_test = self.create_dynamic_test_data(num_test_samples,
+                                                     num_features, xla_dev)
+      y_pred = model(x_test)
+      criterion(y_pred.squeeze(), y_test).item()
+      xm.mark_step()
+    print('Test passed.')
```
JackCaoG, line 118: remove this line

JackCaoG (approved):
I would rename this pr to something along the line of
```
Add dynamic shape support to sigmoidbackward
```
to be less confusing.

test/test_dynamic_shape_models.py
vanbasten23: sg. Do you mind if I remove it in the next pr?

JackCaoG: it is ok

PR #4323, opened 2022-12-13
Store Terraform state in shared GCS bucket

will-cromar:
GCP recommendation here: https://cloud.google.com/docs/terraform/resource-management/store-state

I tested this by deleting my local Terraform state and re-initializing everything. Other members of the team with access to `tpu-pytorch` should be able to run Terraform updates now.


docker/experimental/terraform/main.tf
```diff
@@ -2,6 +2,27 @@ provider "google" {
   project = "tpu-pytorch"
 }
 
+resource "random_id" "bucket_prefix" {
+  byte_length = 8
+}
+
+resource "google_storage_bucket" "default" {
+  name          = "${random_id.bucket_prefix.hex}-bucket-tfstate"
```
steventk-g, line 10: Would this create a new random GCS bucket every time it's executed, since the resource names are different each time? Will we end up with a bunch of unused GCS buckets with random ids?

steventk-g (approved):

docker/experimental/terraform/main.tf
will-cromar: I was worried about this too, but it doesn't look like that will happen. The bucket name is hard coded in `backend "gcs" { ... }`, so I think Terraform looks at that bucket by name, gets the previously-generated ID, and finds the existing bucket instead of creating a new one for `google_storage_bucket.default`.

This part of the config copied and pasted straight from [GCP's documentation](https://cloud.google.com/docs/terraform/resource-management/store-state), so hopefully it works

PR #4326, opened 2022-12-14
Reflect that pjrt now supports `nprocs`.

stgpetrovic:
With merging of https://github.com/pytorch/xla/pull/4285, pjrt now supports `nprocs`.


will-cromar (approved):
Thanks!

PR #4327, opened 2022-12-14
[PT/XLA] Fix timing of ExecuteComputation metrics.

stgpetrovic:
Correct the ExecuteTime metric to reflect the actual execution time and not the time it takes to call ExecuteComputation.


JackCaoG:
As a follow up, we might also want to figure out a way for `wait_device_ops` to wait for the actual device execution to finish. 

The way it works now is that `ExecuteComputation` will grab the device lock and will release the lock once the function call is completed. `wait_device_ops` will try to grab the device lock hence blocked until `ExecuteComputation` is finished. Under PJRT, the `device_lock` will be release early which make `wait_device_ops` inacurrate.

test/test_metrics.py
```diff
@@ -162,6 +163,22 @@ def test_metrics_report(self):
     report = met.metrics_report()
     self.assertIn("CachedCompile", report)
 
+  def test_execute_time_metric(self):
+    xm.xla_device()
```
will-cromar, line 167: Can you add a comment here explaining that this initializes the client before the timer starts?

test/test_metrics.py
```diff
@@ -162,6 +163,22 @@ def test_metrics_report(self):
...
+        10000, 10000, device=xm.xla_device()) * torch.randn(
+            10000, 10000, device=xm.xla_device())
+    value_mean = value.mean()
+    xm.mark_step()
+    cpu_value = value_mean.cpu()
+    wall_time_ns = time.perf_counter_ns() - begin
+    self.assertIn("ExecuteTime", met.metric_names())
+    execute_time_ns = met.metric_data('ExecuteTime')[1]
+    self.assertGreater(execute_time_ns, .5 * wall_time_ns)
+    print(execute_time_ns / 1e9, wall_time_ns / 1e9)
```
will-cromar, line 180: Remove this print statement, since the values will already get printed if this fails.

will-cromar (approved):
Otherwise LGTM. Thanks!

test/pytorch_test_base.py
```diff
@@ -464,6 +464,11 @@
     'TestIndexingXLA': {
         'test_index_put_accumulate_large_tensor_xla',  # illegal memory access was encountered
     },
+
+    # test_metrics.py
+    'MetricsTest': {
+        'test_execute_time_metric',  # timing rigged for testing on CPU
```
JackCaoG, line 470: if the intent is only to run this on CPU, we need to add this test to `DISABLED_TORCH_TESTS_TPU_ONLY` too. 

However I thought this list only works for the test from the upstream. `test_metrics` is a test we defined and I think it won't go through this file. If you want to only want to run it on CPU, you can just check the device hardware type in the test.

test/test_metrics.py
```diff
@@ -162,6 +163,22 @@ def test_metrics_report(self):
...
+    value = torch.randn(
+        10000, 10000, device=xm.xla_device()) * torch.randn(
+            10000, 10000, device=xm.xla_device())
+    value_mean = value.mean()
+    xm.mark_step()
+    cpu_value = value_mean.cpu()
+    wall_time_ns = time.perf_counter_ns() - begin
+    self.assertIn("ExecuteTime", met.metric_names())
+    execute_time_ns = met.metric_data('ExecuteTime')[1]
+    self.assertGreater(execute_time_ns, .5 * wall_time_ns)
```
JackCaoG, line 179: Maybe add a comment here explain this check a bit

PR #4331, opened 2022-12-14
Add test to verify that virtual device reduces outbound data size for…

steventk-g:
… SPMD


test/test_xla_sharding.py
```diff
@@ -147,6 +147,27 @@ def test_metrics_recorded(self):
...
+    baseline_outbound_data = met.counter_value("OutboundData")
+    partition_spec = (0, 1)
+    xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                       dtype=torch.float,
+                       device=xm.xla_device())
+    xs.mark_sharding(xt1, self._get_mesh((1, self.n_devices)), partition_spec)
+    outbound_with_virtual_device = met.counter_value(
+        "OutboundData") - baseline_outbound_data
+
+    os.environ["XLA_USE_SPMD"] = "0"
```
JackCaoG, line 160: do we need to restore this environment after this test? I think it is better to set it to 1, run `mark_sharding`, set it to 0, run again, compare the result and then restore the value of `XLA_USE_SPMD`

steventk-g: Done

JackCaoG: I saw this test always set
```
```
in the setup so this is fine.

test/test_xla_sharding.py
```diff
@@ -147,6 +147,29 @@ def test_metrics_recorded(self):
     self.assertIn("VirtualDeviceUsage", met.counter_names())
     self.assertNotEqual(met.counter_value("VirtualDeviceUsage"), 0)
 
+  def test_outbound_data_metrics(self):
+    baseline_outbound_data = met.counter_value("OutboundData")
+    os.environ["XLA_USE_SPMD"] = "1"
+    partition_spec = (0, 1)
+    xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                       dtype=torch.float,
+                       device=xm.xla_device())
```
JackCaoG, line 156: maybe add a check before `mark_sharding` and verify `OutboundData` is 0? I think it is easier to use `met.clear_counters` or `met.clear_all()` in the beginning of the test if you want to check counter value, unless you have a specified reason to keep the metrics across tests.

test/test_xla_sharding.py
```diff
@@ -147,6 +147,27 @@ def test_metrics_recorded(self):
...
+
+    os.environ["XLA_USE_SPMD"] = "0"
+    xt2 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                       dtype=torch.float,
+                       device=xm.xla_device())
+    xs.mark_sharding(xt2, self._get_mesh((1, self.n_devices)), partition_spec)
+    outbound_without_virtual_device = met.counter_value(
+        "OutboundData") - outbound_with_virtual_device
+
+    self.assertLess(outbound_with_virtual_device,
```
yeounoh, line 168: 🚀

test/test_xla_sharding.py
yeounoh: Maybe add `met.clear_all()` in `SetUp` and call it a day -- since `met.clear_all()` is assumed to be working/testsed in this scope.

yeounoh (approved):
LGTM

test/test_xla_sharding.py
steventk-g: I see `met.clear_counters()` but not `met.clear_all()`, and I don't think clearing counters will affect this metric. We can look at the baseline data transferred before anything gets done in the test, and then take the difference with snapshots after operations with/without virtual device, so that we can look at the data transfer delta after each type of operation. It should be equivalent.

JackCaoG: you can get `clear_all` by rebasing.

PR #4334, opened 2022-12-14
[PJRT] Fix TPU local ordinal calculation

will-cromar:
We discovered recently that PJRT device IDs are not contiguous on each host, which broke the previous local ordinal calculation that assumed they were. Instead, use the index provided during `spawn` to set the local rank (unless one has been set from the outside*), then use that rank to find the local ordinal based on the index of the default addressable device.

Also fix some bugs that were causing TPU client state to persist between tests (FYI @stgpetrovic).

\* I'm trying to keep compatibility with `torchrun` in mind, although I haven't actually tested it.

### DIFF
--- a/test/pjrt/test_experimental_pjrt_tpu.py
+++ b/test/pjrt/test_experimental_pjrt_tpu.py
@@ -1,6 +1,7 @@
 import concurrent.futures
 import os
 from typing import Dict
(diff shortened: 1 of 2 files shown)


JackCaoG (approved):
Thanks!

PR #4342, opened 2022-12-15
Update xla_base tag to v0.5

yeounoh:
This is needed after installing a new package to the current base image.


JackCaoG (approved):

vanbasten23:
Thanks for the fix. It's good to know we need to bump up the tag whenever we install new packages to the base image.

PR #4344, opened 2022-12-15
Prevent conda MKL installation, and install via pip

steventk-g:
We want to use a version of MKL that won't cause our tests to crash. The default version pulled by conda crashes, and it looks like conda does not have the same MKL versions available as pip. So, we can use the nomkl conda flag to ensure that we install conda packages without the MKL dependency, and we can add the proper version of MKL with pip afterwards.

Using `nomkl` conda package 
https://anaconda.org/conda-forge/nomkl

Suggested around the web, e.g. by https://stackoverflow.com/questions/71306262/conda-env-broken-after-installing-pytorch-on-m1-intel-mkl-fatal-error

Testing: Built an image on cloudtop and used it to run cpp tests.

### DIFF
--- a/.circleci/docker/install_conda.sh
+++ b/.circleci/docker/install_conda.sh
@@ -30,8 +30,11 @@ function install_and_setup_conda() {
   conda update -y -n base conda
   conda install -y python=$PYTHON_VERSION
 
-  conda install -y numpy=1.18.5 pyyaml mkl-include setuptools cmake cffi typing \
-    tqdm coverage hypothesis dataclasses cython
+  conda install -y nomkl numpy=1.18.5 pyyaml setuptools cmake \
+    cffi typing tqdm coverage hypothesis dataclasses cython
+
+  /usr/bin/yes | pip install mkl==2022.2.1
+  /usr/bin/yes | pip install mkl-include==2022.2.1
   /usr/bin/yes | pip install typing_extensions  # Required for Python<=3.9
   /usr/bin/yes | pip install --upgrade oauth2client
   /usr/bin/yes | pip install lark-parser
--- a/docker/experimental/Dockerfile
+++ b/docker/experimental/Dockerfile
@@ -41,7 +41,7 @@ RUN update-alternatives --install /usr/bin/clang++ clang++ $(which clang++-8) 70
 ENV CC=clang-8
 ENV CXX=clang++-8
 
-RUN pip install mkl mkl-include setuptools typing_extensions cmake requests
+RUN pip install mkl==2022.2.1 mkl-include=2022.2.1 setuptools typing_extensions cmake requests
 
 RUN git clone --recursive --depth=1 https://github.com/pytorch/pytorch.git
 WORKDIR /pytorch
--- a/scripts/build_torch_wheels.sh
+++ b/scripts/build_torch_wheels.sh
@@ -195,9 +195,9 @@ function install_and_setup_conda() {
   conda activate "$ENVNAME"
   export CMAKE_PREFIX_PATH="$(dirname $(which conda))/../"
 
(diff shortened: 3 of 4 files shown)


JackCaoG:
Let's rebase and reenable the cpp test in this pr. For the debugging purpose, might be easier if we download the CI docker and play with it and make sure cpp test can run.

docker/experimental/Dockerfile
```diff
@@ -41,7 +41,7 @@ RUN update-alternatives --install /usr/bin/clang++ clang++ $(which clang++-8) 70
 ENV CC=clang-8
 ENV CXX=clang++-8
 
-RUN pip install mkl mkl-include setuptools typing_extensions cmake requests
+RUN pip install setuptools typing_extensions cmake requests
```
steventk-g, line 44: May have to create a conda environment to use the conda `nomkl` option to remove conda, something like https://stackoverflow.com/questions/46656367/how-to-create-an-environment-in-anaconda-with-numpy-nomkl

JackCaoG:
Can we prioritizing fixing this? I really want to get cpp test running in CI.

yeounoh:
> Let's rebase and reenable the cpp test in this pr. For the debugging purpose, might be easier if we download the CI docker and play with it and make sure cpp test can run.

It would be good to debug locally first (running the failing tests on docker), not sure how `nomkl` would behave in the CI or with the tests.

yeounoh (approved):
This worked locally, we can merge after we pass the downstream and upstream CI tests.

.circleci/docker/install_conda.sh
```diff
@@ -30,8 +30,9 @@ function install_and_setup_conda() {
   conda update -y -n base conda
   conda install -y python=$PYTHON_VERSION
 
-  conda install -y numpy=1.18.5 pyyaml mkl-include setuptools cmake cffi typing \
-    tqdm coverage hypothesis dataclasses cython
+  conda install -y numpy=1.18.5 pyyaml mkl-include setuptools \
```
yeounoh, line 33: This should be `mkl-include==2022.2.1` as well?

scripts/build_torch_wheels.sh
```diff
@@ -195,9 +195,9 @@ function install_and_setup_conda() {
   conda activate "$ENVNAME"
   export CMAKE_PREFIX_PATH="$(dirname $(which conda))/../"
 
-  conda install -y numpy pyyaml setuptools cmake cffi typing tqdm coverage tensorboard hypothesis dataclasses
+  conda install -y nomkl numpy pyyaml setuptools cmake cffi typing tqdm coverage tensorboard hypothesis dataclasses
```
JackCaoG, line 198: do we still need this nomkl?

.circleci/docker/install_conda.sh
```diff
@@ -30,8 +30,11 @@ function install_and_setup_conda() {
   conda update -y -n base conda
   conda install -y python=$PYTHON_VERSION
 
-  conda install -y numpy=1.18.5 pyyaml mkl-include setuptools cmake cffi typing \
-    tqdm coverage hypothesis dataclasses cython
+  conda install -y nomkl numpy=1.18.5 pyyaml setuptools cmake \
```
JackCaoG, line 33: ditto, we we still need nomkl?

yeounoh: yes, to make sure that other packages don't install mkl.

yeounoh (approved):
LGTM :)

JackCaoG (approved):

JackCaoG:
weird, CI still failed with same error. I guess you can pull docker build by your CI by 
```
docker pull gcr.io/tpu-pytorch/xla_base:latest--24bc1ba0721bd19cd6c7eb0c062dd2f4a9e18250
```
and then inspect the mkl version(I find the docker name from the log). Everything is built so you should be able to run cpp test and saw it failed.

PR #4359, opened 2022-12-16
switch to pinning layout by default in FSDP to avoid data corruption in PJRT and be consistent with xm.all_reduce

ronghanghu:
This PR switches the XLA FSDP implementation to pinning layout by default to avoid data corruption in PJRT and be consistent with `xm.all_reduce`.

cc: @JackCaoG @hjm-aws @AlexWertheim

### DIFF
--- a/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
+++ b/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
@@ -213,7 +213,7 @@ class XlaFullyShardedDataParallel(nn.Module):
       sharding_rank: Optional[int] = None,
       sharding_world_size: Optional[int] = None,
       shard_param_on_dim_0: bool = False,
-      pin_layout_in_collective_ops: bool = False,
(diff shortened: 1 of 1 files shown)


JackCaoG (approved):
Thanks!

JackCaoG:
@aws-rhsoln @amithrm FYI, this might affect you guys, since now `all-gather` will become `all_reduce` based when layout is pin. We need to do this for TPU.

amithrm:
Hi @JackCaoG, is the pinning behavior because of the way compiler handles CC ops in TPUs? Or is this a requirement of FSDP?  Checking because do we need to this as a compulsory fix and hence test out all our use cases

JackCaoG:
when layout is unpin, and each process generate slightly different graph(which we do in pytorch/xla), there is a chance data corruption will happen(because of some XLA optimization assume all graph will be the same).  I don't know if this will be the case for you guys through(whether this is a TPU specified behavior), maybe check with your compiler team?

PR #4362, opened 2022-12-18
Update to clear_all, and add back test (removed in bad merge)

steventk-g:


test/test_xla_sharding.py
```diff
@@ -175,6 +175,28 @@ def test_model_weight_metrics(self):
...
+    partition_spec = (0, 1)
+
+    met.clear_all()
+    xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                       dtype=torch.float,
+                       device=xm.xla_device())
+    xs.mark_sharding(xt1, self._get_mesh((1, self.n_devices)), partition_spec)
+    outbound_with_virtual_device = met.metric_data("OutboundData")[1]
+
+    os.environ["XLA_USE_SPMD"] = "0"
```
JackCaoG, line 188: don't you need to restore the value of `XLA_USE_SPMD` after this test?

steventk-g: I believe that the test setup is run before every method, so the flag gets set before every unit test
https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp

yeounoh (approved):
LGTM

PR #4372, opened 2022-12-19
Enable FSDP test on gpu ci

JackCaoG:
We want to enable some level of FSDP test, we might need to adjust the batch size since CI GPU is not very strong.


JackCaoG:
It took 5 minutes to finished one epoch when I uses an small fake data, I think this is ok.

PR #4387, opened 2022-12-22
Fix a typo in test_input_output_aliases.py

alanwaketan:


test/test_dynamic_shape_models.py
```diff
@@ -45,6 +45,7 @@ def forward(self, x):
 )
 class TestDynamicShapeModels(unittest.TestCase):
 
+  @unittest.skip("Broke by functionalization")
   def test_forward_pass_dynamic_input_correctness(self):
```
JackCaoG, line 49: Do we need to skip this test to get functionalization working? I felt like this will block our ds effort, I will be hesitant to merge functionalization if it blocks development of other projects.

alanwaketan: Oop. A miss commit.... actually. Let me revert it.

JackCaoG (approved):
Thanks!

PR #4390, opened 2022-12-22
[SPMD] Use sharding annotation from XLATensor data as ground truth.

yeounoh:
Use `XLATensor` data to hold a copy of user specified sharding annotation. This way, the IR node truncation and reset should not accidentally remove the sharding annotation, which should be removed via explicit `clear_sharding`.
- `XLATensor::sharding_spec` returns sharding from `XLATensor::data` (source of the truth)
- `XLATensor::Data` constructors now have `sharding` (default to `nullptr`) argument
- `CollectSyncTensors` check for sharding equality if IR sharding exists, and syncs the shardings from `XLATensor::Data` to IR nodes otherwise. 
- Shallow copying `XLATensor` should copy sharding annotation
- Deep copying `XLATensor` should persist sharding annotation

Also, 
- Added a debugging probe in `AssignIrValue`.

### DIFF
--- a/test/test_xla_sharding.py
+++ b/test/test_xla_sharding.py
@@ -7,6 +7,7 @@ import numpy as np
 
 import torch
 from torch import nn
+import torch.optim as optim
 import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.debug.metrics as met
@@ -19,6 +20,18 @@ from torch_xla.experimental.pjrt import using_pjrt
 @unittest.skipIf(not using_pjrt() or xm.get_xla_supported_devices("GPU"),
                  f"Requires PJRT_DEVICE set to `TPU` or `CPU`.")
 class XlaShardingTest(unittest.TestCase):
+
+  class SimpleLinear(nn.Module):
+    def __init__(self):
+        super(SimpleLinear, self).__init__()
+        self.fc1 = nn.Linear(128, 64)
+        self.relu = nn.ReLU()
+        self.fc2 = nn.Linear(64, 1)
+
+    def forward(self, x):
+        y = self.relu(self.fc1(x))
+        z = self.fc2(y)
+        return z
 
   n_devices = 0
   device_ids = None
@@ -110,6 +123,24 @@ class XlaShardingTest(unittest.TestCase):
     sharding_spec = torch_xla._XLAC._get_xla_sharding_spec(xt)
     xm.mark_step()  # mark_step should preserve the sharding
     self.assertEqual(sharding_spec, torch_xla._XLAC._get_xla_sharding_spec(xt))
+
+  def test_optimizer_step_with_sharding(self):
+    model = self.SimpleLinear().to(xm.xla_device())
+    xs.mark_sharding(model.fc1.weight, self._get_mesh((1, self.n_devices)), (0, 1))
+    sharding_spec = torch_xla._XLAC._get_xla_sharding_spec(model.fc1.weight)
+
+    model.train()
+    optimizer = optim.SGD(model.parameters(), lr=0.1)
+    data = torch.randn(128, 128).to(xm.xla_device())
+    target = torch.zeros(128).to(xm.xla_device())
+    for i in range(5):
+      optimizer.zero_grad()
+      output = model(data)
+      loss = nn.CrossEntropy(output, target)
+      loss.backward()
+      optimizer.step()
+      xm.mark_step()
+    self.assertEqual(sharding_spec, torch_xla._XLAC._get_xla_sharding_spec(model.fc1.weight))
 
   def test_inplace_add_with_sharding(self):
     xt = torch.ones(2, 2).to(xm.xla_device())
--- a/test/test_xla_sharding.py
+++ b/test/test_xla_sharding.py
@@ -129,8 +129,7 @@ class BasicShardingTest(XlaShardingTest):
     xt2 = xt.clone()
 
     # check the original sharding spec is preserved after clone()
-    self.assertEqual(sharding_spec,
-        torch_xla._XLAC._get_xla_sharding_spec(xt))
+    self.assertEqual(sharding_spec, torch_xla._XLAC._get_xla_sharding_spec(xt))
 
     # check the cloned sharding spec is the same
     self.assertEqual(
--- a/test/test_xla_sharding.py
+++ b/test/test_xla_sharding.py
@@ -22,16 +22,17 @@ from torch_xla.experimental.pjrt import using_pjrt
 class XlaShardingTest(unittest.TestCase):
 
   class SimpleLinear(nn.Module):
+
     def __init__(self):
-        super(SimpleLinear, self).__init__()
-        self.fc1 = nn.Linear(128, 64)
-        self.relu = nn.ReLU()
-        self.fc2 = nn.Linear(64, 1)
+      super(XlaShardingTest.SimpleLinear, self).__init__()
+      self.fc1 = nn.Linear(128, 64)
+      self.relu = nn.ReLU()
+      self.fc2 = nn.Linear(64, 1)
 
     def forward(self, x):
-        y = self.relu(self.fc1(x))
-        z = self.fc2(y)
-        return z
+      y = self.relu(self.fc1(x))
+      z = self.fc2(y)
+      return z
 
   n_devices = 0
   device_ids = None
@@ -46,6 +47,9 @@ class XlaShardingTest(unittest.TestCase):
       device_ids = self.device_ids
     assert len(device_ids) == self.n_devices
     return xs.Mesh(device_ids, mesh_shape)
+
+
+class BasicShardingTest(XlaShardingTest):
 
   def test_xla_sharded_tensor(self):
     partition_spec = (0, 1)
@@ -117,6 +121,15 @@ class XlaShardingTest(unittest.TestCase):
         torch_xla._XLAC._get_xla_sharding_spec(xt),
         torch_xla._XLAC._get_xla_sharding_spec(xt2))
 
+  def test_clone(self):
+    xt = torch.randn(2, 4, 8, 16).to(xm.xla_device())
+    xs.mark_sharding(xt, self._get_mesh((1, 1, 1, self.n_devices)),
+                     (0, 1, 2, 3))
+    xt2 = xt.clone()
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(xt),
+        torch_xla._XLAC._get_xla_sharding_spec(xt2))
+
   def test_mark_step_with_sharding(self):
     xt = torch.ones(2, 2).to(xm.xla_device())
     xs.mark_sharding(xt, self._get_mesh((1, self.n_devices)), (0, 1))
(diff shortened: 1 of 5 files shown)


torch_xla/csrc/tensor.h
```diff
@@ -187,14 +191,13 @@ class XLATensor : public torch::lazy::LazyTensor {
...
   };
-  using ShardingSpecPtr = std::shared_ptr<ShardingSpec>;
 
   // Annotate the IR value with ShardingSpec.
   void SetShardingSpec(const ShardingSpec& sharding_spec);
   // Clear sharding annotation attached to the IR value and transfer sharded
   // data back to host.
   void ClearShardingSpec();
-  ShardingSpecPtr sharding_spec() const;
+  ShardingSpecPtr sharding_spec() const { return data()->sharding; }
```
JackCaoG, line 200: add an assert here to make sure `IR`'s sharding spec(if there is one) matches XLATensor's sharding spec. 

ideally we also need to force IR's shardingspec immutable, I think we will sort of do that in Steven's follow up pr where shardingspec becomes an constructor argument.

torch_xla/csrc/tensor.cpp
```diff
@@ -291,8 +287,13 @@ void XLATensor::AssignIrValue(torch::lazy::Value ir_value) const {
...
       // Create a tensor node if applicable, re-use the current IR otherwise.
-      // TODO(yeounoh) this has some performance implications for convolution.
-      ir_value = GetIrValue();
+      torch::lazy::BackendDataPtr handle = CurrentDataHandle();
+      if (handle != nullptr) {
+        ir_value = CreateTensorNode(handle, /*read_only=*/false);
+      } else {
+        ir_value = CurrentIrValue();
+      }
+      XLA_CHECK(ir_value);
```
JackCaoG, line 296: I am generally confuse about this part of the code, why would `ir_value` here? can we add more detial comments?

yeounoh: Thanks, done.

torch_xla/csrc/tensor.h
yeounoh: THanks, done!

yeounoh:
Do not merge, until I rebase --and remove the unnecessary tf pin update here.

yeounoh:
> Do not merge, until I rebase --and remove the unnecessary tf pin update here.

Can merge now.

torch_xla/csrc/tensor.cpp
```diff
@@ -210,39 +218,41 @@ torch::lazy::BackendDataPtr XLATensor::GetXlaData() {
...
 void XLATensor::SetShardingSpec(const ShardingSpec& sharding) {
   // Existing annotation must be cleared explicitly. We do not clear and
   // overwrite the existing sharding on the user's behalf. This is a no-op if
   // the same sharding already applied.
   if (sharding_spec() == nullptr ||
       !ShardingUtil::EqualShardingSpecs(sharding, *sharding_spec())) {
     TORCH_LAZY_COUNTER("SetShardingSpec", 1);
-    XLA_CHECK(GetIrValue().node != nullptr) << "Tyring to access a null cursor";
+    data()->sharding = std::make_shared<ShardingSpec>(sharding);
     dynamic_cast<XlaNode*>(GetIrValue().node.get())
```
JackCaoG, line 229: what's going to happen if `GetIrValue().node` is nullptr? I assume when there we eventually create IR for this node, we will copy the sharding_spec over?

torch_xla/csrc/tensor.cpp
```diff
@@ -294,9 +304,19 @@ void XLATensor::AssignIrValue(torch::lazy::Value ir_value) const {
...
+      // the current IR to preserve the sharding.
+      // TODO(yeounoh) this may not be needed when functionalization is fully
+      // implemented.
+      torch::lazy::BackendDataPtr handle = CurrentDataHandle();
+      if (handle != nullptr) {
+        ir_value = CreateTensorNode(handle, /*read_only=*/false);
+      } else {
+        ir_value = CurrentIrValue();
+      }
+      XLA_CHECK(ir_value);
```
JackCaoG, line 319: hmm what if current `XLATensor` already have a shardingspec and the `ir_value` passed also have a shardingspec, are we going to error out here?

yeounoh: The new ir_value as input would not have the sharding spec, unless it's provided by the user via mark_sharding. Also, once the sharding is attached to the tensor and its IR, it cannot be modified unless it's first cleared from both XLATensor and IR.

torch_xla/csrc/tensor.cpp
yeounoh: Sharded tensor sharding is attached via mark_sharding and it's accompanied with a tensor IR node, backed by sharded device data.

yeounoh:
> 

sounds good, per our offline discussion, I will add sharding as a default field to ctor, add checks for sync tensors.

test/test_xla_sharding.py
```diff
@@ -104,13 +121,43 @@ def test_deep_copy(self):
...
         torch_xla._XLAC._get_xla_sharding_spec(xt2))
 
+  def test_clone(self):
+    xt = torch.randn(2, 4, 8, 16).to(xm.xla_device())
+    xs.mark_sharding(xt, self._get_mesh((1, 1, 1, self.n_devices)),
+                     (0, 1, 2, 3))
+    xt2 = xt.clone()
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(xt),
+        torch_xla._XLAC._get_xla_sharding_spec(xt2))
```
JackCaoG, line 131: nit, this won't check the case where `clone` will incorrectly remove the sharding_spec from origional tensor and the new tensor. Maybe consider adding another check to save the old sharding_spec and make sure it does not change after `clone`

torch_xla/csrc/tensor.cpp
```diff
@@ -79,6 +79,14 @@ XLATensorPtr XLATensor::Create(
     c10::optional<at::ScalarType> logical_element_type) {
   XLATensorPtr xtensor = c10::make_intrusive<XLATensor>(
       XLATensor(std::move(ir_value), device, logical_element_type));
+  // Preserve sharding if a new tensor is created from a sharded IR node.
+  if (ir_value) {
```
JackCaoG, line 83: 1. `ir_value` is already std::moved, I think it will always be null?
2. why not do this logic in `XLATensor`'s constrcutor? `Create` message suppose to just register the created tensor.

torch_xla/csrc/tensor.cpp
JackCaoG: if `ir_value` is empty, doesn't it means we want to reset the IR value to null? If that's the case shouldn't we just leave it as it is. We can attach the sharding_spec when we create the real IR for the current XLATensor.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -896,7 +896,11 @@ XLATensorPtr clamp(const XLATensorPtr& input,
 }
 
 XLATensorPtr clone(const XLATensorPtr& input) {
-  return input->CreateFrom(input->GetIrValue());
+  XLATensorPtr cloned = input->CreateFrom(input->GetIrValue());
+  if (input->sharding_spec() != nullptr) {
+    cloned->SetShardingSpec(*input->sharding_spec());
```
JackCaoG, line 901: maybe pass `sharding_spec` to `CreateFrom` directly?

yeounoh: I'd prefer them separate, so that each API has smaller and more specific scope, thus more modular. If need to, let's address the API changes in a separate PR.

torch_xla/csrc/tensor.cpp
yeounoh: Good catch, and yes should do this using the constructor now!

torch_xla/csrc/tensor_methods.cpp
JackCaoG: ok after I did a quick grep, `CreateFrom` was used in a lot of different methods under `tensor_methods` we for the most of them we don't want to inherit the `sharding_spec`, separting them actually is the right thing to do here.

test/test_xla_sharding.py
yeounoh: Sounds good, let me extend this.

torch_xla/csrc/tensor.cpp
yeounoh: Oh, that's because GetIrValue shoulnd't be empty.

torch_xla/csrc/tensor.cpp
yeounoh: Hm, I was thinknig that the empty IR would be replaced with a tensor node here, since the sharding annotation implies sharded data handle here. WIth the new change in the `CollectSyncTensor` we may be able to simplify (skip) here, let me test though.

torch_xla/csrc/tensor.cpp
JackCaoG: Do we still need to keep `dynamic_cast<XlaNode*>(GetIrValue().node.get())->SetSharding(sharding.sharding);` through? given that in `CollectSyncTensors` we will set it anyway

yeounoh: Yea, I think we could simplify this by removing it, just like in AssignIrValue. Let me test & veirfy though.

torch_xla/csrc/tensor.cpp
yeounoh: Ok for `AssignIrValue` with an empty node, we can skip and rely on `CollectSyncTensor`.

torch_xla/csrc/tensor.cpp
yeounoh: Done.

torch_xla/csrc/tensor.cpp
yeounoh: We can't seem to skip this part, I suspect there are nodes that don't need to be synced (sync tensors). The `AssignIrValue` can be simplified though because it's used to trim after the initial annotation.

torch_xla/csrc/tensor.cpp
```diff
@@ -114,6 +114,14 @@ XLATensor::XLATensor(torch::lazy::Value ir_value,
...
     : XLATensor(std::make_shared<Data>(std::move(ir_value), device,
                                        logical_element_type)) {
+  // Preserve sharding if a new tensor is created from a sharded IR node.
+  if (CurrentIrValue()) {
+    auto* xla_node = dynamic_cast<XlaNode*>(CurrentIrValue().node.get());
+    if (xla_node->GetSharding()) {
+      ShardingSpec sharding = ShardingSpec{*xla_node->GetSharding()};
+      SetShardingSpec(sharding);
+    }
+  }
```
JackCaoG, line 124: I think I asked in a previous review but I forgot if you answered. What if passed in `ir_value` has `ShardingSpec`, is that impossible? If it is impossible can we assert on that?

JackCaoG: oh OK I misread, this happened after you assign the `ir_value` to self, so `CurrentIrValue` will return `ir_value`, this make sense now..

torch_xla/csrc/tensor.cpp
```diff
@@ -291,16 +299,6 @@ void XLATensor::SetInPlaceIrValue(torch::lazy::Value ir_value) {
...
-    if (!ir_value) {
-      // Create a tensor node if applicable, re-use the current IR otherwise.
-      // TODO(yeounoh) this has some performance implications for convolution.
-      ir_value = GetIrValue();
-    }
-    dynamic_cast<XlaNode*>(ir_value.node.get())
-        ->SetSharding(sharding->sharding);
-  }
   data()->ir_value = std::move(ir_value);
   data()->generation += 1;
```
JackCaoG, line 303: what if passed in `ir_value` has a sharding spec?

torch_xla/csrc/tensor.h
```diff
@@ -53,31 +53,43 @@ using XLATensorPtr = c10::intrusive_ptr<XLATensor>;
...
         : torch::lazy::LazyTensor::Data(device),
           view(std::move(view)),
-          logical_element_type(logical_element_type) {}
+          logical_element_type(logical_element_type),
+          sharding(sharding) {}
 
     ~Data();
 
     std::shared_ptr<View> view;
+    ShardingSpecPtr sharding;
```
JackCaoG, line 92: can we add comment here to explain this sharding is the source of truth and the ir one is just a copy?

torch_xla/csrc/tensor.cpp
yeounoh: The check will be done at the sharding_spec() call.

torch_xla/csrc/tensor.h
yeounoh: Sounds good.

torch_xla/csrc/tensor.cpp
JackCaoG: I am afraid this will make debugging a bit difficult, since we can only see that there is a mismatch but it is hard to figure out why(error message is lazy). The code is cleaner this way I guess, so I am also OK with this approach.

JackCaoG (approved):
Thanks @yeounoh !

torch_xla/csrc/tensor.cpp
yeounoh: Yea, it's cleaner this way that we have a limited (consolicated) places, where we do the checks and the syncing. I see your point, though. Let's add some debugging probes `TF_VLOG` for that instead.

JackCaoG: let's do
```
  TF_VLOG(5) << "Assign IR value " << ir_value->ToString() + sharding ? (" with sharding " + sharding->DebugString()) : "";
```
This way when `TF_VLOG` is not enabled, we don't construct the debug string.

PR #4396, opened 2022-12-28
Use table to represent wheels and dockers in README

JackCaoG:
use https://github.com/pytorch/xla/tree/JackCaoG/Update_wheel_docker_to_table#readme to view how it looks


will-cromar:
What do you think of moving the version columns to the left side? The URLs in the leftmost column are harder to read

JackCaoG:
sg, let me do that.

will-cromar (approved):

PR #4400, opened 2022-12-28
Enable PJRT C API with `libtpu-nightly`

will-cromar:
As of our last TF update, the `libtpu-nightly` package contains the new PJRT C API.

- Use `GetHloModules` now that it's implemented
- Switch `ExecuteSharded` to use `on_device_shape` to match `ExecuteReplicated`
- Use `xla::PjRtClient::HostBufferSemantics::kImmutableUntilTransferCompletes` now that it's implemented
- Temporary hack: copy `TimedSection` pointer to output buffers' `OnReady` callbacks. Futures with `ExecuteSharded`  are [already implemented](https://github.com/tensorflow/tensorflow/commit/b8f59020ea0e9e6fba0e9c5e7be88271703eaf9e) and will be available in the next libtpu update (thanks @jyingl3!)
- Add a TPU test for the `ExecuteTime` metric to make sure I'm using the callback correctly.
- Explicitly load TPU PJRT plugin to address a [TODO in TensorFlow](https://github.com/tensorflow/tensorflow/blob/a0c74d196a21b9d1230820d31e15492ec84d07a6/tensorflow/compiler/xla/stream_executor/tpu/pjrt_api.cc#L43-L51)

Tested manually with `test_experimental_pjrt_tpu.py` and `test_train_mp_imagenet.py` on TPU v4.

### DIFF
--- a/test/pjrt/test_experimental_pjrt_tpu.py
+++ b/test/pjrt/test_experimental_pjrt_tpu.py
@@ -1,5 +1,6 @@
 import concurrent.futures
 import os
+import time
 from typing import Dict
 from unittest import mock
 import requests
@@ -10,6 +11,7 @@ import torch.nn as nn
 from absl.testing import absltest, parameterized
 import torch_xla.core.xla_env_vars as xenv
 import torch_xla.core.xla_model as xm
+import torch_xla.debug.metrics as met
 from torch_xla.experimental import pjrt
 from torch_xla.experimental import tpu
 import torch_xla.distributed.xla_multiprocessing as xmp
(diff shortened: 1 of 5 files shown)


test/pjrt/test_experimental_pjrt_tpu.py
```diff
@@ -250,6 +252,33 @@ def test_device_attributes(self):
...
+    value_mean = value.mean()
+    xm.mark_step()
+    cpu_value = value_mean.cpu()
+    wall_time_ns = time.perf_counter_ns() - begin
+    _, execute_time_ns, _ = met.metric_data('ExecuteTime')
+
+    return execute_time_ns
+
+  def test_execute_time_metric(self):
+    results = pjrt._run_multiprocess(self._execute_time_metric)
```
JackCaoG, line 273: is this test the one added by Goran?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -322,29 +311,31 @@ PjRtComputationClient::ExecuteComputation(
...
-                           returned_future, /*fill_future=*/true)
+          ->ExecuteSharded(buffers, pjrt_device, execute_options)
           .value();
+
   // Signal that `ExecuteSharded` has completed for the ExecuteTime metric.
   // Copies the `timed` shared pointer into the lambda.
-  returned_future->OnReady([timed](Status unused) mutable { timed.reset(); });
+  // TODO(wcromar): Uncomment this when we update past TF commit
+  // b8f59020ea0e9e6fba0e9c5e7be88271703eaf9e
+  // returned_future->OnReady([timed](Status unused) mutable { timed.reset();
```
JackCaoG, line 324: is `returned_future->OnReady` broken with current pin?

test/pjrt/test_experimental_pjrt_tpu.py
will-cromar: Goran added a similar test for CPU here:

https://github.com/pytorch/xla/blob/9c01c3ca14f7b34f84421c6db1e102386c26cf3f/test/test_metrics.py#L166-L186

Making that test work for TPU is tricky because the overhead is way higher on TPU, so the execute time is a small fraction of the overall wall time. I did check that this new test does fail if I revert our fix to the execute time metric.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: Yes. The fix should be included in our next pin update and we can revert this change.

JackCaoG (approved):

PR #4403, opened 2022-12-28
Pjrt device attributes

darisoy:
Add device attributes to PjRt
using the computation_client (along with pjrt_computation_client and xrt_computation_client)
added unit tests for the new method and passing.

### DIFF
--- a/third_party/xla_client/computation_client.h
+++ b/third_party/xla_client/computation_client.h
@@ -8,6 +8,7 @@
 #include <string>
 #include <vector>
 
+#include "absl/container/flat_hash_map.h"
 #include "absl/types/optional.h"
 #include "absl/types/span.h"
 #include "tensorflow/compiler/xla/client/xla_computation.h"
@@ -305,11 +306,12 @@ class ComputationClient {
 
   virtual int GetNumProcesses() const = 0;
 
-  using DeviceAttributes =
+  using DeviceAttribute =
       std::variant<std::string, int64_t, std::vector<int64_t>, float>;
 
-  virtual std::map<std::string, DeviceAttributes> GetDeviceAttributes(
-      const std::string& device) = 0;
+  virtual const absl::flat_hash_map<std::string,
+                                    xla::ComputationClient::DeviceAttribute>&
+  GetDeviceAttributes(const std::string& device) = 0;
 
   virtual void SetReplicationDevices(
       std::shared_ptr<std::vector<std::string>> devices) = 0;
--- a/third_party/xla_client/computation_client.h
+++ b/third_party/xla_client/computation_client.h
@@ -305,10 +305,11 @@ class ComputationClient {
 
   virtual int GetNumProcesses() const = 0;
 
-  using DeviceAttributes = std::variant<std::string, int64_t, std::vector<int64_t>, float>;
+  using DeviceAttributes =
+      std::variant<std::string, int64_t, std::vector<int64_t>, float>;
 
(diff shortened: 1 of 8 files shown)


third_party/xla_client/computation_client.h
```diff
@@ -305,6 +305,12 @@ class ComputationClient {
 
   virtual int GetNumProcesses() const = 0;
 
+  using DeviceAttributes =
```
will-cromar, line 308: nit: this should be `DeviceAttribute` since it represents one attribute

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -443,6 +443,20 @@ int PjRtComputationClient::GetNumProcesses() const {
   return max_process_index + 1;
 };
 
+std::map<std::string, xla::PjRtDeviceAttribute>
```
will-cromar, line 446: This should be `std::map<std::string, DeviceAttribute>` to match the signature in the headers (although they're technically the same thing)

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -443,6 +443,20 @@ int PjRtComputationClient::GetNumProcesses() const {
...
+PjRtComputationClient::GetDeviceAttributes(const std::string& device) {
+  XLA_CHECK(string_to_device_.find(device) != string_to_device_.end())
+      << "Unknown device " << device;
+  const absl::flat_hash_map<std::string, xla::PjRtDeviceAttribute>& attributes =
+      string_to_device_[device]->Attributes();
+
+  std::map<std::string, xla::PjRtDeviceAttribute> map;
```
will-cromar, line 453: Do you need to do a copy here? If you're just converting to a `std::map` so callers don't have to deal with an absl container, then use the iterator constructor of `map`, e.g. the opposite of this: https://github.com/abseil/abseil-cpp/blob/6abc1958562c49d797ea23270a355caf5dc39f94/absl/container/flat_hash_map.h#L155-L158

https://en.cppreference.com/w/cpp/container/map/map

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1173,6 +1173,18 @@ void InitXlaModuleBindings(py::module m) {
   m.def("_xla_get_device_ordinal", [](const std::string& device_str) {
     return bridge::AtenDeviceToXlaDevice(device_str).ordinal();
   });
+  m.def("_xla_get_device_attributes", [](const std::string& device_str) {
+    std::map<std::string, 
+             std::variant<std::string, int64_t, std::vector<int64_t>, float>>
```
will-cromar, line 1178: nit: can you use `DeviceAttribute` here?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -443,6 +443,13 @@ int PjRtComputationClient::GetNumProcesses() const {
   return max_process_index + 1;
 };
 
+const absl::flat_hash_map<std::string, xla::ComputationClient::DeviceAttribute>&
+PjRtComputationClient::GetDeviceAttributes(const std::string& device) {
+  XLA_CHECK(string_to_device_.find(device) != string_to_device_.end())
```
will-cromar, line 448: Use `PjRtComputationClient::StringToPjRtDevice` instead

will-cromar (approved):
Otherwise LGTM

PR #4409, opened 2023-01-04
Force fail on all unimplemented methods.

vanbasten23:
This pr makes it easy to debug dynamic shape related issues.

### DIFF
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -604,9 +604,9 @@ bool XLASymNodeImpl::is_float() {
 }
 
 c10::SymNode XLASymNodeImpl::add(const c10::SymNode& other) {
-  auto pother = dynamic_cast<XLASymNodeImpl*>(other.get());
-  auto nadd = torch::lazy::MakeNode<SizeAdd>(node(), pother->node());
-  return c10::make_intrusive<XLASymNodeImpl>(nadd);
+  auto p_other = dynamic_cast<XLASymNodeImpl*>(other.get());
+  auto n_add = torch::lazy::MakeNode<SizeAdd>(node(), p_other->node());
+  return c10::make_intrusive<XLASymNodeImpl>(n_add);
 }
 
 c10::SymNode XLASymNodeImpl::sub(const c10::SymNode& other) {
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/tensor.cpp
```diff
@@ -603,14 +603,42 @@ bool XLASymNodeImpl::is_float() {
   return false;
 }
 
-bool XLASymNodeImpl::bool_() {
-  auto dn = torch_xla::DimCast(node());
-  return dn->getDynamicValue() != 0;
+c10::SymNode XLASymNodeImpl::add(const c10::SymNode& other) {
+  auto pother = dynamic_cast<XLASymNodeImpl*>(other.get());
```
JackCaoG, line 607: nit `p_other` nad `n_add`

JackCaoG (commented):
hmm, does upstream has a defualt implementation? I would expect upstream just failed if it is unimplemented.

vanbasten23:
> hmm, does upstream has a defualt implementation? I would expect upstream just failed if it is unimplemented.

I agree, I'd expect it just fails if we haven't implemented the method for XLASymNodeImpl. 

I added it here because Ed in his [comment](https://github.com/pytorch/xla/pull/4289#issuecomment-1348387865) mentioned `The easiest is to just edit each virtual method in the XLA SymNode subclass to print stuff. We don't have a builtin printer in core but it is also trivial to write for us as our SymNode is written in Python. I suppose we could also write an adapter lol`. I'm reluctant to edit each virtual methods every time we debug a dynamic shape related issue. And we may need to implement them eventually.

torch_xla/csrc/tensor.cpp
vanbasten23: done

torch_xla/csrc/tensor.cpp
```diff
@@ -603,43 +603,139 @@ bool XLASymNodeImpl::is_float() {
   return false;
 }
 
-bool XLASymNodeImpl::bool_() {
-  auto dn = torch_xla::DimCast(node());
-  return dn->getDynamicValue() != 0;
+c10::SymNode XLASymNodeImpl::add(const c10::SymNode& other) {
+  auto p_other = dynamic_cast<XLASymNodeImpl*>(other.get());
+  auto n_add = torch::lazy::MakeNode<SizeAdd>(node(), p_other->node());
+  return c10::make_intrusive<XLASymNodeImpl>(n_add);
```
JackCaoG, line 609: can we add test for this method?

vanbasten23: Another [pr](https://github.com/pytorch/xla/pull/4258) is intended to add tests for this method and the current pr just wants to force fail the unimplemented method. Do you mind if I add the test in the other pr?

JackCaoG: oh ok, you did not modify this implemention, it is just a git diff thing.

JackCaoG (approved):
Thanks!

PR #4416, opened 2023-01-05
Change all size node logical_element_type to be compatible with the underlying XLA device.

vanbasten23:
This PR also fixes an issue of [pr](https://github.com/pytorch/xla/pull/4410).

### DIFF
--- a/torch_xla/csrc/ops/dynamic_ir.cpp
+++ b/torch_xla/csrc/ops/dynamic_ir.cpp
@@ -62,11 +62,14 @@ XlaOpVector SizeNode::Lower(LoweringContext* loctx) const {
   return ReturnOp(xla::GetDimensionSize(input, this->dim_), loctx);
 }
 
-std::string SizeNode::ToString() const { return "SizeNode"; }
+std::string SizeNode::ToString() const { return "aten::size for size"; }
 
 SizeAdd::SizeAdd(torch::lazy::Value a, torch::lazy::Value b)
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::add")},
-              {a, b}, xla::ShapeUtil::MakeShape(xla::S64, {}), 1) {
+              {a, b},
+              xla::ShapeUtil::MakeShape(
+                  GetShapeDimensionType(/*device=*/nullptr), {}),
+              1) {
   const torch::lazy::DimensionNode* dim_node_0 = DimCast(operand(0));
   const torch::lazy::DimensionNode* dim_node_1 = DimCast(operand(1));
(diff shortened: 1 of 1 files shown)


torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -95,7 +98,10 @@ XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
 
 SizeEq::SizeEq(torch::lazy::Value a, torch::lazy::Value b)
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::eq")},
-              {a, b}, xla::ShapeUtil::MakeShape(xla::S64, {}), 1) {
+              {a, b},
+              xla::ShapeUtil::MakeShape(
```
vanbasten23, line 102: In my original problematic [pr](https://github.com/pytorch/xla/pull/4410), it uses `GetShapeDimensionType(/*device=*/nullptr)` which only uses a `xla::PrimitiveType` but the base class constructor needs a `shape`. So the difference here is I use `xla::ShapeUtil::MakeShape` to pass a `shape` to base class constructor.

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -116,7 +122,10 @@ SizeConstant::SizeConstant(int64_t val) : Scalar(c10::Scalar{val}, xla::S64){};
 
 SizeMul::SizeMul(torch::lazy::Value a, torch::lazy::Value b)
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::mul")},
-              {a, b}, xla::ShapeUtil::MakeShape(xla::S64, {}), 1) {
+              {a, b},
+              xla::ShapeUtil::MakeShape(
```
vanbasten23, line 126: another difference from my original problematic pr for the same reason

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -145,7 +154,10 @@ XlaOpVector SizeMul::Lower(LoweringContext* loctx) const {
 
 SizeDiv::SizeDiv(torch::lazy::Value a, torch::lazy::Value b)
     : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::div")},
-              {a, b}, xla::ShapeUtil::MakeShape(xla::S64, {}), 1) {
+              {a, b},
+              xla::ShapeUtil::MakeShape(
```
vanbasten23, line 158: another difference from my original problematic pr for the same reason

PR #4417, opened 2023-01-05
Fix worker pool name in GCB config

will-cromar:
I had the format for the pool name incorrect: https://cloud.google.com/build/docs/api/reference/rest/v1/projects.builds#Build.PoolOption

This should match the other cloud build config:

https://github.com/pytorch/xla/blob/acfb22750b39820dffb99e639d493cc633654d35/docker/cloudbuild.yaml#L45-L46

Also fix the package version


yeounoh (approved):

yeounoh:
LGTM, just to confirm -- have we tested this on GCB (manual invocation)?

will-cromar:
I was able to successfully start the build on this branch:

image

Issue #4402, opened 2022-12-28
[Dynamo] Make `_run_cached_graph` not blocking for dynamo


JackCaoG:
## 🚀 Feature
`_run_cached_graph` is the API we used to execute the cached graph for our dynamo bridge. Currently this call is blocking, which means it will wait until actual device execution to finish. That is not necessary since we can create data placeholder for the XLATensor we return and let execution to be run in a separate thread. When the execution is finished, we can replace the placeholder with the real data. This is OK since execution thread will hold the device lock which prevent others to access datas.
This is also what we do in LTC to overlap the tracing and execution.

## Motivation

When I profile the training using the existing `aot_xla_trace_once` bridge, I found there is a significant between graph executions. It seems like aot_autograd needs to perform some computation between forward and backward graphs. Those does not need to be blocked on device execution since it is unlikely to require the actual value of the tensor. This should hopefully make our bridge runs a bit faster.

## Pitch

in `_run_cached_graph`, make device execution happens in a separate thread. We also need to know the output shape of the graph given a hash in order to create the data placeholder.

@alanwaketan @shunting314 @will-cromar

PR #4425, opened 2023-01-07
Make dyanmo execute async

JackCaoG:
This is to implement https://github.com/pytorch/xla/issues/4402,

I verified `test/dynamo/test_dynamo.py` worked. Next I will try to rerun the inference and training benchmark to verified this does not regress inference and hopefully make training faster.

Update:
for training

command I used
```
XLA_IR_DEBUG=0 XLA_HLO_DEBUG=0 USE_FAKE_TENSOR=0 python benchmarks/dynamo/torchbench.py --randomize-input --performance --trace-on-xla --training --backend=aot_torchxla_trace_once --only $model -n 10
```
on TPUv4 nightly with/without this change

| model | with this change | without this change | speed up | 
| --- | ----------- | ----------- |----------- |
| resnet50 | 0.758x | 0.724x | 1.047
| resnet18 | 0.665x | 0.653x | 1.018
| BERT_pytorch | 1.441x | 1.400x | 1.029
| resnext50_32x4d | 0.870x | 0.849x | 1.025
| alexnet | 0.632x | 0.662x | 0.955
| mobilenet_v2 | 0.549x | 0.539x | 1.019
| mnasnet1_0 | 0.698x | 0.669x | 1.043
| vgg16  | 0.712x | 0.721x | 0.988
| average |           |             | 1.0155

At least for the single step training, overlapping the execution and training does not help the speed too much.

FYI @wconstab @shunting314 @alanwaketan @wonjoolee95

### DIFF
--- a/torch_xla/csrc/xla_graph_executor.cpp
+++ b/torch_xla/csrc/xla_graph_executor.cpp
@@ -543,16 +543,39 @@ void XLAGraphExecutor::TensorCollectionBarrier(SyncTensorCollection* coll) {
 
 std::vector<torch::lazy::BackendDataPtr>
 XLAGraphExecutor::ExecuteComputationWithBarrier(
-    torch::lazy::ComputationPtr computation,
+    torch::lazy::hash_t hash,
     c10::ArrayRef<torch::lazy::BackendDataPtr> arguments,
     const torch::lazy::BackendDevice& device) {
+  MaybeDumpGraph("dynamo", hash);
+  auto cachedComputation =
+      XLAGraphExecutor::Get()->GetComputationCache()->Get(hash);
+  // TODO implement a fallback mechanism, or make sure those entries
+  // never get kicked out
+  XLA_CHECK(cachedComputation)
+      << "Failed to get computation by hash " << torch::lazy::HashToString(hash)
+      << ". Maybe the entry get "
+         "kicked out of the LRU cache";
   std::vector<torch::lazy::ExceptionCleanup> unlocker;
   unlocker = DeviceLockerArena::Get()->LockDevices({device});
+
   // Create DataPlaceHolder that will get filled in async executions.
-  // vector<xla::Shape>* output_shapes =
-  // DeviceContextArena::Get()->GetOutputShapesByHash();
-  return torch::lazy::getBackend()->ExecuteComputation(computation, arguments,
-                                                       device);
+  std::vector<xla::Shape>* output_shapes =
+      DeviceContextArena::Get()->GetOutputShapesByHash(hash);
+  std::vector<torch::lazy::BackendDataPtr> placeholders;
+  placeholders.reserve(output_shapes->size());
+  for (const xla::Shape& shape : *output_shapes) {
+    torch::lazy::BackendDataPtr handle =
+        WrapXlaData(xla::ComputationClient::Get()->CreateDataPlaceholder(
+            device.toString(), std::move(shape)));
+    placeholders.push_back(handle);
+  }
+
+  // std::shared_ptr<XLAGraphExecutor::Async> async = std::make_shared<Async>(
+  //   coll, std::move(parameters_data), std::move(tensors_data),
+  //   std::move(cached_computation));
+
+  return torch::lazy::getBackend()->ExecuteComputation(
+      cachedComputation->computation, arguments, device);
 }
 
 std::vector<at::Tensor> XLAGraphExecutor::GetTensorsOpByOp(
--- a/torch_xla/csrc/xla_graph_executor.cpp
+++ b/torch_xla/csrc/xla_graph_executor.cpp
@@ -140,9 +140,9 @@ void XLAGraphExecutor::DeviceContextArena::SaveGraphAsString(
 }
 
 void XLAGraphExecutor::DeviceContextArena::SaveOutputShapes(
-    torch::lazy::hash_t hash, std::vector<xla::Shape> outptu_shapes) {
+    torch::lazy::hash_t hash, std::vector<xla::Shape> output_shapes) {
   if (hash_to_output_shape_map.find(hash) == hash_to_output_shape_map.end()) {
-    hash_to_output_shape_map[hash] = std::move(outptu_shapes);
+    hash_to_output_shape_map[hash] = std::move(output_shapes);
   }
 }
 
@@ -386,13 +386,13 @@ torch::lazy::hash_t XLAGraphExecutor::GetGraphHash(
   SyncTensorCollection coll = CollectSyncTensors(tensors, config);
   absl::Span<const size_t> indices = coll.indices;
   std::vector<torch::lazy::Value> ir_values;
-  std::vector<xla::Shape> outptu_shapes;
+  std::vector<xla::Shape> output_shapes;
   ir_values.reserve(indices.size());
-  outptu_shapes.reserve(indices.size());
+  output_shapes.reserve(indices.size());
   for (auto index : indices) {
     XLATensorPtr tensor = tensors[index];
     ir_values.push_back(tensor->CurrentIrValue());
-    outptu_shapes.push_back(MakeShapeWithDeviceLayout(
+    output_shapes.push_back(MakeShapeWithDeviceLayout(
         tensor->shape(),
         static_cast<XlaDeviceType>(tensor->GetDevice().type())));
   }
@@ -400,7 +400,7 @@ torch::lazy::hash_t XLAGraphExecutor::GetGraphHash(
   torch::lazy::hash_t res_hash = torch::lazy::HashCombine(
       coll.hash, torch::lazy::Hash(po_data.parameter_sequence));
   DeviceContextArena::Get()->SaveOutputShapes(res_hash,
-                                              std::move(outptu_shapes));
+                                              std::move(output_shapes));
   DeviceContextArena::Get()->SaveGraphAsString(res_hash, tensors,
                                                &coll.indices);
   return res_hash;
--- a/torch_xla/csrc/xla_graph_executor.cpp
+++ b/torch_xla/csrc/xla_graph_executor.cpp
@@ -139,6 +139,13 @@ void XLAGraphExecutor::DeviceContextArena::SaveGraphAsString(
   }
 }
 
+void XLAGraphExecutor::DeviceContextArena::SaveOutputShapes(
+    torch::lazy::hash_t hash, std::vector<xla::Shape> outptu_shapes) {
+  if (hash_to_output_shape_map.find(hash) == hash_to_output_shape_map.end()) {
+    hash_to_output_shape_map[hash] = std::move(outptu_shapes);
+  }
+}
+
 std::string XLAGraphExecutor::DeviceContextArena::GetGraphByHash(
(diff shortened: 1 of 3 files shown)


JackCaoG:
OK With training I am seeing some slight improvement, with master 
```
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -334,7 +334,12 @@ PjRtComputationClient::ExecuteComputation(
 
   std::vector<DataPtr> datas;
   datas.reserve(results.size());
+  bool waited = false;
   for (auto& result : results) {
+    if (!waited) {
+      auto status = result->GetReadyFuture().Await();
+      waited = true;
+    }
     std::unique_ptr<xla::PjRtBuffer> buffer = std::move(result);
 
     std::shared_ptr<PjRtData> data = std::make_shared<PjRtData>(
```
(diff is needed to make PJRT benchmark accurate)

resnet50 training output
```
cpu  train resnet50                           0.709x SAME
```

with this change
```
cpu  train resnet50                           0.754x SAME
```

so roughly a 6% improvement. I will test it on more models and post the update here.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -370,13 +386,21 @@ torch::lazy::hash_t XLAGraphExecutor::GetGraphHash(
   SyncTensorCollection coll = CollectSyncTensors(tensors, config);
   absl::Span<const size_t> indices = coll.indices;
   std::vector<torch::lazy::Value> ir_values;
+  std::vector<xla::Shape> outptu_shapes;
```
alanwaketan, line 389: `outptu_shapes ` => `output_shapes`

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -519,13 +543,58 @@ void XLAGraphExecutor::TensorCollectionBarrier(SyncTensorCollection* coll) {
 
 std::vector<torch::lazy::BackendDataPtr>
 XLAGraphExecutor::ExecuteComputationWithBarrier(
-    torch::lazy::ComputationPtr computation,
-    c10::ArrayRef<torch::lazy::BackendDataPtr> arguments,
+    torch::lazy::hash_t hash,
+    std::vector<torch::lazy::BackendDataPtr> arguments,
```
alanwaketan, line 547: Should this be `const std::vector<torch::lazy::BackendDataPtr>&`? Otherwise, a copy will be introduced.

torch_xla/csrc/xla_graph_executor.h
```diff
@@ -161,8 +161,8 @@ class XLAGraphExecutor : public torch::lazy::LazyGraphExecutor {
   ComputationCache* GetComputationCache();
 
   std::vector<torch::lazy::BackendDataPtr> ExecuteComputationWithBarrier(
-      torch::lazy::ComputationPtr computation,
-      c10::ArrayRef<torch::lazy::BackendDataPtr> arguments,
+      torch::lazy::hash_t hash,
```
alanwaketan, line 164: Should this be const std::vector<torch::lazy::BackendDataPtr>&?

torch_xla/csrc/xla_graph_executor.h
```diff
@@ -209,12 +209,21 @@ class XLAGraphExecutor : public torch::lazy::LazyGraphExecutor {
         const std::vector<size_t>* indices,
         DebugUtil::GraphFormat format = DebugUtil::GetDefaultGraphFormat());
 
+    void SaveOutputShapes(torch::lazy::hash_t hash,
+                          std::vector<xla::Shape> outptu_shapes);
```
alanwaketan, line 213: `outptu_shapes` => `output_shapes`, and `const std::vector<xla::Shape>&` perhaps?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -139,6 +139,13 @@ void XLAGraphExecutor::DeviceContextArena::SaveGraphAsString(
   }
 }
 
+void XLAGraphExecutor::DeviceContextArena::SaveOutputShapes(
+    torch::lazy::hash_t hash, std::vector<xla::Shape> outptu_shapes) {
```
alanwaketan, line 143: outptu_shapes => output_shapes, and const std::vector<xla::Shape>& perhaps?

JackCaoG: will fix

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: will fix

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: I actually do this intentionally, if you check the code below, `arguments` is `std::move` to the `async` object because we want to maintain a copy for the execution thread.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -519,13 +543,58 @@ void XLAGraphExecutor::TensorCollectionBarrier(SyncTensorCollection* coll) {
...
+         "kicked out of the LRU cache";
+
+  // Create DataPlaceHolder that will get filled in async executions.
+  std::vector<xla::Shape>* output_shapes =
+      DeviceContextArena::Get()->GetOutputShapesByHash(hash);
+  std::vector<torch::lazy::BackendDataPtr> placeholders;
+  placeholders.reserve(output_shapes->size());
+  for (const xla::Shape& shape : *output_shapes) {
+    torch::lazy::BackendDataPtr handle =
+        WrapXlaData(xla::ComputationClient::Get()->CreateDataPlaceholder(
```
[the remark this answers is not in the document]
JackCaoG: `xla::shape` contains the both `dimension` and `type`, it looks like `s32[100, 100]` for example.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -519,13 +543,58 @@ void XLAGraphExecutor::TensorCollectionBarrier(SyncTensorCollection* coll) {
...
+               << torch::lazy::HashToString(hash) << " on device "
+               << async->device << " done!";
+    // Updating placeholder with actual output handle.
+    for (size_t i = 0; i < results.size(); ++i) {
+      XLA_CHECK(async->tensors_data[i] != nullptr);
+      async->tensors_data[i]->Assign(*results[i]);
+    }
+  };
+
+  xla::env::ScheduleIoClosure(async->mwait.Completer(std::move(syncfn)));
```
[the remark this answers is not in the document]
JackCaoG: sure, so if you look at above code
```
coll.unlocker = DeviceLockerArena::Get()->LockDevices({device});
std::shared_ptr<XLAGraphExecutor::Async> async = std::make_shared<Async>(
      &coll, std::move(arguments), placeholders, std::move(cachedComputation));
```

we grab the device lock and pass it to the `syncfn`. Every operation that requires reads real data from the device requires a device lock. Async execution thread will hold the device lock until it finished computation, and release the lock when stack object is destroyed. If user wants to access tensor's data like `t1.cpu()`, it will try to grab the lock which will force them to wait for async execution to finish.

torch_xla/csrc/xla_graph_executor.h
JackCaoG: we actually wants to make a copy and pass it to sync fn(separate thread), we can pass a const ref but then we need to make a copy inside the `ExecuteComputationWithBarrier`.

torch_xla/csrc/xla_graph_executor.h
JackCaoG: `outptu_shapes` will be saved in a global map and outlive the stack object from the caller, I think it has to be a copy.

JackCaoG: it is also `std::move` so we are not making an extra copy.

JackCaoG:
command I used
```
XLA_IR_DEBUG=0 XLA_HLO_DEBUG=0 USE_FAKE_TENSOR=0 python benchmarks/dynamo/torchbench.py --randomize-input --performance --trace-on-xla --training --backend=aot_torchxla_trace_once --only $model -n 10
```
on TPUv4 nightly with/without this change

| model | with this change | without this change | speed up | 
| --- | ----------- | ----------- |----------- |
| resnet50 | 0.758x | 0.724x | 1.047
| resnet18 | 0.665x | 0.653x | 1.018
| BERT_pytorch | 1.441x | 1.400x | 1.029
| resnext50_32x4d | 0.870x | 0.849x | 1.025
| alexnet | 0.632x | 0.662x | 0.955
| mobilenet_v2 | 0.549x | 0.539x | 1.019
| mnasnet1_0 | 0.698x | 0.669x | 1.043
| vgg16  | 0.712x | 0.721x | 0.988
| average |           |             | 1.0155

At least for the single step training, overlapping the execution and training does not help the speed too much.

JackCaoG:
I will clean up this pr and consider making this execution async configurable.

torch_xla/csrc/xla_graph_executor.cpp
alanwaketan: Sounds good.

torch_xla/csrc/xla_graph_executor.h
alanwaketan: Sounds good.

torch_xla/csrc/xla_graph_executor.h
alanwaketan: I guess my hobby has always been explicitly making the copy in the code instead of relying on value passing. That's why I always ask. Otherwise, it's too much thinking on deciding whether a parameter should be passed by value or reference. Just easier to determine if it's const reference or r reference.

alanwaketan (approved):
LGTM.

torch_xla/csrc/xla_graph_executor.h
JackCaoG: sg, will update.

torch_xla/csrc/xla_graph_executor.cpp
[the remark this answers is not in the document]
shunting314: Thanks for the explanation!

One more thing confirm, so the lock is not a tensor level lock but a device level lock?  i.e., for each device, there is a single such lock?

shunting314 (approved):

torch_xla/csrc/xla_graph_executor.cpp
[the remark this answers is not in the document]
JackCaoG: yes, it is a device level lock, hence the name `LockDevices`

torch_xla/csrc/xla_graph_executor.h
JackCaoG: ah, actually in this case we want to keep it as `std::vector<xla::Shape>`. This way caller can use `std::move` to avoid the extra copy we need to do if we only pass in a reference.

alanwaketan: Haha, it's tricky. I'm actually fine with either ways. Just trying to share some of my thoughts here.

PR #4429, opened 2023-01-07
 Udpate TF pin to enable `libtpu-nightly==0.1.dev.20221223`

yeounoh:
This is to enable libtpu-nightly==0.1.dev.20221223 -- it was previously reverted because of GCB build issues.

### DIFF
--- a/scripts/build_torch_wheels.sh
+++ b/scripts/build_torch_wheels.sh
@@ -182,7 +182,7 @@ function install_and_setup_conda() {
   fi
   ENVNAME="pytorch"
   if conda env list | awk '{print $1}' | grep "^$ENVNAME$"; then
-    conda remove --name "$ENVNAME" --all
+    conda remove -y --name "$ENVNAME" --all
   fi
   if [ -z "$PYTHON_VERSION" ]; then
     PYTHON_VERSION=$DEFAULT_PYTHON_VERSION
--- a/scripts/build_torch_wheels.sh
+++ b/scripts/build_torch_wheels.sh
@@ -252,6 +252,12 @@ function build_and_install_torch() {
 
   python setup.py bdist_wheel
   pip install dist/*.whl
+
+  # collect_wheels expects this
+  if [ ! -d "/pytorch/dist" ]; then
+    mkdir -p /pytorch/dist
+    cp dist/*.whl /pytorch/dist/
+  fi
 }
 
 function build_and_install_torch_xla() {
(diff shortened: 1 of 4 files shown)


scripts/build_torch_wheels.sh
```diff
@@ -293,6 +311,12 @@ function install_torchaudio_from_source() {
   python setup.py bdist_wheel
   pip install dist/*.whl
   popd
+
+  # collect_wheels expects this
+  if [ ! -d "/pytorch/audio/dist" ]; then
+    mkdir /pytorch/audio/dist
+    cp audio/dist/*.whl /pytorch/audio/dist/
```
JackCaoG, line 318: I thought we already have script to do that.

JackCaoG: check https://github.com/pytorch/xla/blob/master/docker/common.sh#L49

yeounoh:
Verified on 
- locally (TPUVM - v4): don't see any obvious regression
```
| Training Device=xla:0/2 Epoch=1 Step=2340 Loss=0.00135 Rate=1767.83 GlobalRate=1039.97 Time=23:50:23
| Training Device=xla:0/1 Epoch=1 Step=2340 Loss=0.00135 Rate=1767.58 GlobalRate=1059.54 Time=23:50:23
| Training Device=xla:0/0 Epoch=1 Step=2340 Loss=0.00135 Rate=1767.48 GlobalRate=1064.44 Time=23:50:23
| Training Device=xla:0/3 Epoch=1 Step=2340 Loss=0.00135 Rate=1767.71 GlobalRate=1040.13 Time=23:50:23
```

will-cromar (approved):
Thanks!

scripts/build_torch_wheels.sh
yeounoh: Yea, that actually assumes that the wheel is already placed under `/pytorch/...` https://github.com/pytorch/xla/blob/570c6d44871175785793ab79601b7cb9e3f16146/docker/common.sh#L27.

JackCaoG: hmm, why wouldn't that be the case for this pr? Wheel generally be placed under `dist/`

yeounoh: It's under /pytorch/dist, if pytorch is installed at the root /. It would fail if it's installed and build from else where, since the build script uses relative paths!

JackCaoG: ok, sg

yeounoh:
The GCB wheel build also succeeded.

yeounoh:
Rebased

JackCaoG (approved):

PR #4433, opened 2023-01-09
Remove a dynamic shape related comment that doesn't hold anymore.

vanbasten23:


test/cpp/run_tests.sh
```diff
@@ -43,7 +43,6 @@ done
 shift $(($OPTIND - 1))
 
 if [[ "$TPUVM_MODE" != "1" ]]; then
-  # Dynamic shape is not supported on the tpuvm.
   export XLA_EXPERIMENTAL
 fi
```
JackCaoG, line 47: we can remove the whole block, this block will reset `XLA_EXPERIMENTAL` when running `run_test.sh` on TPUVM

JackCaoG: before removing it, first try to run some dynamic shapes python tests on TPUVM and confirm test will pass.

vanbasten23: Actually I'm not sure what this blocks does:

Seems this file is only used by circle CI but CI only run test on CPU and GPU instead of TPU. Do you know why we are checking if we are using TPU ("$TPUVM_MODE" != "1")? Or if other places also use this file?

JackCaoG: it is also used by xlml test, check python-op-test in our xlml tests

vanbasten23: I see what you meant. I've removed the block.
I'm developing on my TPU VM and the DS tests pass when the backend is TPU so I think we're good.

JackCaoG (approved):

will-cromar (approved):

vanbasten23:
Thanks for the review!

PR #4436, opened 2023-01-10
Fix TF patch command in experimental build

will-cromar:
```
$ find -wholename 'tf_patches/*.diff'
$ find tf_patches -name *.diff
tf_patches/stream_executor.diff
tf_patches/xplane.diff
tf_patches/thread_local_random.diff
tf_patches/f16_abi_clang.diff
```


steventk-g (approved):
LGTM! Thanks Will!

PR #4438, opened 2023-01-11
Lower masked_scatter to fix masked_scatter op test in xla/test/test_ops.py

wonjoolee95:
Lower masked_scatter to fix masked_scatter op test in xla/test/test_ops.py #4438

This PR removes the old in-place version of the `masked_scatter_` op with the out-of-place version `masked_scatter`. It also updates the tensor broadcasting logic to fix the `masked_scatter` python op tests which is explained in a comment below. 

---

Earlier investigations (just for context):

Initially, the `masked_scatter` op test failed with:
```
RuntimeError: The operator aten::as_strided appears to be a view operator, but it has no implementation for the backend "xla:0". View operators don't support falling back to run on the CPU, since the tensor's storage cannot be shared across devices.
```

Tried to fix it by just lowering `masked_scatter` with a manual functionalization pass, but it failed due to segmentation fault. 

Then finally tried to manually lower `masked_scatter` by reusing the existing lowered in-place op `maksed_scatter_`. Now the segmentation fault is gone, but failing due to tensor comparison:
```

Mismatched elements: 13 / 25 (52.0%)
Greatest absolute difference: 14.423441886901855 at index (4, 3) (up to 0.001 allowed)
Greatest relative difference: 13.10500368692721 at index (1, 4) (up to 0.001 allowed)
```

### DIFF
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -1545,11 +1545,20 @@ void masked_fill_(XLATensorPtr& input, const XLATensorPtr& mask,
       value));
 }
 
-void masked_scatter_(XLATensorPtr& input, const XLATensorPtr& mask,
-                     const XLATensorPtr& source) {
+XLATensorPtr masked_scatter(XLATensorPtr& input, const XLATensorPtr& mask,
+                            const XLATensorPtr& source) {
   torch::lazy::ScopePusher ir_scope(at::aten::masked_scatter.toQualString());
-  input->SetIrValue(torch::lazy::MakeNode<MaskedScatter>(
-      input->GetIrValue(), MaybeExpand(mask->GetIrValue(), input->shape()),
+  auto input_value = input->GetIrValue();
+  // This ensures that input tensor is at least the same shape as mask tensor.
+  // Note that we can't use the existing MaybeExpand function since
+  // input tensor may sometimes be bigger than the mask tensor, and MaybeExpand
+  // requires the first parameter to always be less or equal to the second
+  // parameter.
+  if (input->shape().get().dimensions() < mask->shape().get().dimensions()) {
+    input_value = MaybeExpand(input->GetIrValue(), mask->shape());
+  }
+  return input->CreateFrom(torch::lazy::MakeNode<MaskedScatter>(
+      input_value, MaybeExpand(mask->GetIrValue(), GetXlaShape(input_value)),
       source->GetIrValue()));
 }
 
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -5443,8 +5443,7 @@ TEST_F(AtenXlaTensorTest, TestMaskedScatter) {
       // calls.
       ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
     }
-    ExpectCounterChanged("xla::masked_scatter_",
-                         cpp_test::GetIgnoredCounters());
+    ExpectCounterChanged("xla::masked_scatter", cpp_test::GetIgnoredCounters());
     ResetCounters();
   });
 }
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -117,7 +117,7 @@ allowed_opinfo = set(
             AllowedOpInfoEntry('lt'),
             AllowedOpInfoEntry('lu_unpack'),
             AllowedOpInfoEntry('masked_fill'),
-            AllowedOpInfoEntry('masked_scatter')
+            AllowedOpInfoEntry('masked_scatter'),
             AllowedOpInfoEntry('masked_select'),
             AllowedOpInfoEntry('matrix_exp'),
             AllowedOpInfoEntry('max', 'binary'),
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -117,6 +117,7 @@ allowed_opinfo = set(
             AllowedOpInfoEntry('lt'),
             AllowedOpInfoEntry('lu_unpack'),
             AllowedOpInfoEntry('masked_fill'),
+            AllowedOpInfoEntry('masked_scatter')
             AllowedOpInfoEntry('masked_select'),
             AllowedOpInfoEntry('matrix_exp'),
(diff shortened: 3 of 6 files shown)


wonjoolee95:
With the latest commit now the previous `AssertionError: Tensor-likes are not close!` is gone. And the op test progresses further. However, now is failing with another assertion at `BuildExpand` (at https://github.com/pytorch/xla/blob/master/torch_xla/csrc/data_ops.cpp#L135):
```
XLA_CHECK_LE(input_sizes.size(), output_sizes.size());
```

alanwaketan:
Looks like the CIs were passing. Maybe we can do a rebase and merge this fix?

alanwaketan:
> Looks like the CIs were passing. Maybe we can do a rebase and merge this fix?

Never mind. The CI is green because of keep going...

wonjoolee95:
All the tests are finally succeeding. The problem was that now functionalization passes in `self` tensors that may be smaller in dimension that the `mask` tensor for the `masked_scatter` op. If this is the case, we need to broadcast the `self` tensor to the same size with the `mask` tensor. However, since this never happened when functionalization was disabled, such broadcasting logic didn't exist. Now we added the broadcasting logic, the tests are passing.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1545,11 +1545,20 @@ void masked_fill_(XLATensorPtr& input, const XLATensorPtr& mask,
...
 
-void masked_scatter_(XLATensorPtr& input, const XLATensorPtr& mask,
-                     const XLATensorPtr& source) {
+XLATensorPtr masked_scatter(XLATensorPtr& input, const XLATensorPtr& mask,
+                            const XLATensorPtr& source) {
   torch::lazy::ScopePusher ir_scope(at::aten::masked_scatter.toQualString());
-  input->SetIrValue(torch::lazy::MakeNode<MaskedScatter>(
-      input->GetIrValue(), MaybeExpand(mask->GetIrValue(), input->shape()),
+  auto input_value = input->GetIrValue();
+  // This ensures that input tensor is at least the same shape as mask tensor.
```
alanwaketan, line 1552: I don't think we should expand the input to match the mask's shape.  See: https://pytorch.org/docs/stable/generated/torch.Tensor.masked_scatter_.html#torch.Tensor.masked_scatter_.

alanwaketan: Wonder if you have dumped the parameters to have a reduced test case and then compare the behavior with eager?

alanwaketan: If not, I guess you could utilize this: https://github.com/pytorch/xla/blob/master/test/test_ops.py#L415?

wonjoolee95: I also thought this was the case initially looking at https://github.com/pytorch/xla/blob/master/test/test_ops.py#L415, and it really just seems like it's a lack of documentation. I should've described more in detail in this PR, I'll put down some more details. So the exact set of tensors that fail are:
```
WONJOO: self=
 4
 3
 4
-1
-6
[ XLALongType{5} ]
WONJOO: mask=
 0  0  0  1  1
 1  1  1  1  1
 1  0  1  1  0
 0  1  0  0  0
 0  1  0  0  0
[ XLABoolType{5,5} ]
WONJOO: source=
-8 -4  3 -4 -6
-1 -7 -5  9 -3
-3  5  5  5  7
 8  2  8  1  5
-4 -4 -6  1 -3
```
When I mimic this (same shape of self, mask, source tensors) in PyTorch only (without XLA), the result is:
```
>>> import torch
>>> self = torch.tensor([1, 1, 1, 1, 1])
>>> mask = torch.tensor([[0, 0, 0, 1, 1], [1, 1, 1, 1, 1], [1, 0, 1, 1, 0],[0, 1, 0, 0, 0],[0, 1, 0, 0, 0]])
>>> source = torch.tensor([[0, 0, 0, 0, 0], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0],[0, 0, 0, 0, 0],[0, 0, 0, 0, 0]])
>>> self.masked_scatter(mask, source)
tensor([[1, 1, 1, 0, 0],
        [0, 0, 0, 0, 0],
        [0, 1, 0, 0, 1],
        [1, 0, 1, 1, 1],
        [1, 0, 1, 1, 1]])
```
As we can see, the `self` tensor is expanded to the shape of the `mask` tensor.

Previously, such example case will fail with the error at https://github.com/pytorch/xla/pull/4438#issuecomment-1380899277 in XLA. With this PR, the XLA behavior is matched with the above PyTorch behavior:
```
>>> import torch; import torch_xla
>>> self = torch.tensor([1, 1, 1, 1, 1], device="xla:0")
>>> mask = torch.tensor([[0, 0, 0, 1, 1], [1, 1, 1, 1, 1], [1, 0, 1, 1, 0],[0, 1, 0, 0, 0],[0, 1, 0, 0, 0]], device="xla:0")
>>> source = torch.tensor([[0, 0, 0, 0, 0], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0],[0, 0, 0, 0, 0],[0, 0, 0, 0, 0]], device="xla:0")
>>> self.masked_scatter(mask, source)
tensor([[1, 1, 1, 0, 0],
        [0, 0, 0, 0, 0],
        [0, 1, 0, 0, 1],
        [1, 0, 1, 1, 1],
        [1, 0, 1, 1, 1]], device='xla:0')

```

So what it is, really, seems to be a lack of better documentation at https://github.com/pytorch/xla/blob/master/test/test_ops.py#L415. Let me know if this sounds okay, @alanwaketan.

alanwaketan: Sounds good. Can you file a GH issue on pytorch?

alanwaketan (approved):
LGTM. Thanks, Wonjoo. The investigation is intriguing.

torch_xla/csrc/tensor_methods.cpp
wonjoolee95: Created an issue for PyTorch documentation at https://github.com/pytorch/pytorch/issues/94353.

PR #4447, opened 2023-01-12
Code-gen addcdiv again

### LINKED ISSUE #4213: Seeing a roughly 25% slow down with torch-xla=1.13 because of the addcdiv_ and addcmul_ ops

alanwaketan:
Summary:
This pull request redoes the addcdiv and addcmul code-gen, and adds a test case to verify that if we reuse the DataCache for scalars.

This needs pytorch/pytorch#92066 to function.

Test Plan:
PJRT_DEVICE=CPU python test/test_operations.py -v -k test_cached_addcdiv

Fixes #4213.


test/test_operations.py
```diff
@@ -1893,6 +1893,21 @@ def test_sigmoid_bounds(self):
...
+  def test_cached_addcdiv(self):
+    xla_device = xm.xla_device()
+    met.clear_all()
+
+    t1 = torch.randn(1, 3).to(xla_device)
+    t2 = torch.randn(1, 3).to(xla_device)
+    t3 = torch.randn(1, 3).to(xla_device)
+    t1.addcdiv_(t2, t3, value=0.1)
+    # The following two scalars shouldn't trigger TransferToServerTime.
+    t1.addcdiv_(t2, t3, value=0.1)
```
JackCaoG, line 1905: Let's may this test a bit more explicit by checking the `met.metric_data("TransferToServerTime")[0]` before second `addcdiv_` and after second `addcdiv_` to verify that second `addcdiv_` does not introduce new transfer.

alanwaketan: Okay, let me do that.

JackCaoG (approved):
Thanks!

alanwaketan:
Thanks Jack for approving.

wonjoolee95 (approved):

PR #4451, opened 2023-01-13
Split virtual device and basic sharding tests

steventk-g:
When we run virtual device tests with basic sharding tests in the same process, we get crashes. When we run these tests in their own processes, they all pass. 

Since we currently only support configuring virtual device at the host process level (i.e. as an environment variable), we should have no loss of real world coverage separating these tests.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -158,7 +158,8 @@ function run_op_tests {
   run_pjrt python3 "$CDIR/pjrt/test_experimental_tpu.py"
   run_pjrt python3 "$CDIR/pjrt/test_ddp.py"
   run_pjrt python3 "$CDIR/pjrt/test_mesh_service.py"
-  run_pjrt python3 "$CDIR/test_xla_sharding.py"
+  run_pjrt python3 "$CDIR/spmd/test_xla_sharding.py"
+  run_pjrt python3 "$CDIR/spmd/test_xla_virtual_device.py"
   run_test python3 "$CDIR/test_operations_hlo.py" "$@" --verbosity=$VERBOSITY
   run_test python3 "$CDIR/test_input_output_aliases.py"
 }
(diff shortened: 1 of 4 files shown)


test/test_xla_virtual_device.py
```diff
@@ -0,0 +1,90 @@
+import os
```
JackCaoG, line 1: I think we need to update run_test.sh. Can we also move all SPMD test into a separate dir like `test/spmd/`

torch_xla/test/test_xla_sharding_base.py
```diff
@@ -0,0 +1,41 @@
...
+from torch import nn
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.experimental.xla_sharding as xs
+from torch_xla.experimental.pjrt import using_pjrt
+
+
+@unittest.skipIf(not using_pjrt() or xm.get_xla_supported_devices("GPU"),
+                 f"Requires PJRT_DEVICE set to `TPU` or `CPU`.")
+class XlaShardingTest(unittest.TestCase):
```
JackCaoG, line 14: I think it is better to put this file under `/test/spmd` instead of `torch_xla/test`

steventk-g: As far as I can tell, we need to put it under `torch_xla` so that it can be imported by tests (like `test_util.py` in `torch_xla/test`. I could create `torch_xla/test/spmd` to make it more explicit.

JackCaoG: I can you put import from python file at the same path, check https://github.com/pytorch/xla/blob/master/test/args_parse.py

steventk-g: Done

JackCaoG (approved):

yeounoh (commented):
Post-approval. LGTM.

PR #4461, opened 2023-01-14
Adds Keep Going to CPP tests

alanwaketan:
Summary:
CPP tests was recently re-enabled but without the new keep going feature. Let's add it.

Test Plan:
CI.


alanwaketan:
@wonjoolee95 Now that we have keep going in three different scripts. Maybe it's good to think about how to keep it in one place.

PR #4464, opened 2023-01-17
Enable lowering for upsample_bilinear2d with scale factor

### LINKED ISSUE #2703: lowering upsample_bilinear2d
## 🚀 Feature
A request for a lowering of upsample_bilinear2d_backward
reference https://github.com/pytorch/xla/issues/2702

(and these 2 as well)
Counter: aten::_local_scalar_dense
Counter: aten::isnan

## Motivation
stylegan makes use of nn.Upsample for its generator, which is causing a call to `grad` to be very slow.  please see 
https://colab.research.google.com/drive/1MEyQ2KMDn1IjxJ2FLHcEgmySBlWNuR5q?usp=sharing
 
## Pitch

I noticed this because my calls using perceptual path length were slow, ie 9 seconds per step. (when doing 100k steps for a GAN, it can be a big slow down)  you can see the example in the colab above. lowering it would make it faster and thus stylegan would be feasible on pytorch tpu

## Alternatives

I don't know of any

lsy323:
This fixes #2703

Enable lowering for upsample_bilinear2d with scale factor

Added a new unit tests of upsample_bilinear2d with scale factor != 1. The test will fail with the current implementation, since it will fallback to Aten with scale factor != 1. The test passed after the lowering for upsample_bilinear2d with scale factor is supported.

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -2911,7 +2911,7 @@ at::Tensor XLANativeFunctions::upsample_bilinear2d(
       torch::lazy::ToVector<int64_t>(output_size);
   if ((scales_h && *scales_h != 1.0) || (scales_w && *scales_w != 1.0)) {
     scaled_output_size = GetOutputSizeWithScale(input_dims, scales_h, scales_w,
-                                            scaled_output_size);
+                                                scaled_output_size);
   }
   return bridge::AtenFromXlaTensor(tensor_methods::upsample_bilinear2d(
       self_tensor, scaled_output_size, align_corners));
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -2911,6 +2911,11 @@ at::Tensor XLANativeFunctions::upsample_bilinear2d(
   if ((scales_h && *scales_h != 1.0) || (scales_w && *scales_w != 1.0)) {
     scaled_output_size = GetOutputSizeWithScale(input_dims, scales_h, scales_w,
                                                 scaled_output_size);
+    if (!output_size.empty()) {
+      XLA_CHECK(scaled_output_size.at(0) == output_size.at(0) &&
+                scaled_output_size.at(1) == output_size.at(1))
+          << "Inferred output size and output_size from upstream are different";
+    }
   }
   return bridge::AtenFromXlaTensor(tensor_methods::upsample_bilinear2d(
       self_tensor, scaled_output_size, align_corners));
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2905,16 +2905,16 @@ at::Tensor XLANativeFunctions::upsample_bilinear2d(
     c10::optional<double> scales_h, c10::optional<double> scales_w) {
   TORCH_LAZY_FN_COUNTER("xla::");
   XLATensorPtr self_tensor = bridge::GetXlaTensor(self);
```
JackCaoG, line 2907: I think we need to make sure only one of (`scales_h ` + `scales_w` ) and `output_size` is specified, otherwise I don't know which one we should believe. Ideally upsteram already does that check but we can make it more explict.

lsy323: After investigating using GDB, I found the `output_size` will always be filled by upstream (at: https://github.com/pytorch/pytorch/blob/88366a907549abdd7e2c402a961b60c2be910824/aten/src/ATen/native/UpSampleBilinear2d.cpp#L166)

With the current upstream implementation, we can rely on the `output_size` inferred from scales_h/w by upstream. However, I think we can keep the scale factor shape inference to be future-proof.

In the scale factor shape inference block, I think we can do a shape validation if `output_size` is not empty, to make sure the output shape and the inferred shape are the same.

JackCaoG (approved):

JackCaoG:
For a sanity check, @lsy323 can you check if you use python api `upsample_bilinear2d ` with scale factor, will it fall back to CPU(you can check that by looking at coutner before after the op)?

lsy323:
> For a sanity check, @lsy323 can you check if you use python api `upsample_bilinear2d ` with scale factor, will it fall back to CPU(you can check that by looking at coutner before after the op)?

Will do, thanks!

lsy323:
> For a sanity check, @lsy323 can you check if you use python api `upsample_bilinear2d ` with scale factor, will it fall back to CPU(you can check that by looking at coutner before after the op)?

Tested with the following Python script, `upsample_bilinear2d` with scale factor will be lowered instead of falling back to CPU

```
import torch
import unittest


class UpsampleBilinear2DTest(unittest.TestCase):

  def test_upsample_bilinear2d(self):
    xla_device = xm.xla_device()
    t1 = torch.rand(2, 2, 5, 5, device=xla_device)
    upsample_bilinear2d = torch.nn.UpsamplingBilinear2d(
        scale_factor=(8.0 / 5.0, 8.0 / 5.0))
    t2 = upsample_bilinear2d(t1)
    xm.mark_step()
    assert (met.counter_value("xla::upsample_bilinear2d") == 1)


if __name__ == '__main__':
  test = unittest.main()
  sys.exit(0 if test.result.wasSuccessful() else 1)
```

PR #4469, opened 2023-01-18
[Functionalization] Fix cpp tests

wonjoolee95:
Summary:
Lowers ops `copy`, `copy_`, `_prelu_kernel`, and `diag_embed` to fix cpp tests that were failing with functionalization. 

Note that we lower both `copy` and `copy_` ops. When we only lower `copy` op, the in-place version `copy_` op gets auto-generated. However, upstream PyTorch had an assertion that this auto-generated was wrong. To by pass, we're manually lowering `copy_` as well.

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -2305,29 +2305,6 @@ at::Tensor XLANativeFunctions::pow(const at::Scalar& self,
       tensor_methods::pow(self, bridge::GetXlaTensor(exponent)));
 }
 
-// at::Tensor XLANativeFunctions::prelu(const at::Tensor& self,
-//                                      const at::Tensor& weight) {
-//   TORCH_LAZY_FN_COUNTER("xla::");
-//   // If multiple weights, check channel size == number of weights.
-//   int64_t weight_num = weight.numel();
-//   if (weight.numel() > 1) {
-//     int64_t input_dim = self.dim();
-//     XLA_CHECK_GT(input_dim, 0) << "Input tensor dimension cannot be 0";
-
-//     int64_t channel_size = input_dim > 1 ? self.size(1) : 1;
-//     XLA_CHECK_EQ(channel_size, weight_num)
-//         << "Mismatch of parameter numbers and input channel size. Found "
-//            "parameter numbers = "
-//         << weight_num << " and channel size = " << channel_size;
-//   }
-
-//   XLATensorPtr self_tensor = bridge::GetXlaTensor(self);
-//   XLATensorPtr weight_tensor = bridge::GetXlaTensor(weight);
-
-//   return bridge::AtenFromXlaTensor(
-//       tensor_methods::prelu(self_tensor, weight_tensor));
-// }
-
 at::Tensor XLANativeFunctions::_prelu_kernel(const at::Tensor& self,
                                              const at::Tensor& weight) {
   TORCH_LAZY_FN_COUNTER("xla::");
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -2295,7 +2295,29 @@ at::Tensor XLANativeFunctions::pow(const at::Scalar& self,
 at::Tensor XLANativeFunctions::prelu(const at::Tensor& self,
                                      const at::Tensor& weight) {
   TORCH_LAZY_FN_COUNTER("xla::");
+  // If multiple weights, check channel size == number of weights.
+  int64_t weight_num = weight.numel();
+  if (weight.numel() > 1) {
+    int64_t input_dim = self.dim();
+    XLA_CHECK_GT(input_dim, 0) << "Input tensor dimension cannot be 0";
 
+    int64_t channel_size = input_dim > 1 ? self.size(1) : 1;
+    XLA_CHECK_EQ(channel_size, weight_num)
+        << "Mismatch of parameter numbers and input channel size. Found "
+           "parameter numbers = "
+        << weight_num << " and channel size = " << channel_size;
+  }
+
+  XLATensorPtr self_tensor = bridge::GetXlaTensor(self);
+  XLATensorPtr weight_tensor = bridge::GetXlaTensor(weight);
+
+  return bridge::AtenFromXlaTensor(
+      tensor_methods::prelu(self_tensor, weight_tensor));
+}
+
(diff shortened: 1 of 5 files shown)


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2295,7 +2323,29 @@ at::Tensor XLANativeFunctions::pow(const at::Scalar& self,
 at::Tensor XLANativeFunctions::prelu(const at::Tensor& self,
```
alanwaketan, line 2323: Why don't we just replace prelu with _prelu_kernel?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1107,6 +1107,34 @@ at::Tensor XLANativeFunctions::diag(const at::Tensor& self, int64_t diagonal) {
       tensor_methods::diag(bridge::GetXlaTensor(self), diagonal));
 }
 
+at::Tensor XLANativeFunctions::diag_embed(const at::Tensor& self,
```
alanwaketan, line 1110: It looks like the TS backend implemented it via passing it to functionalize again, wdyt?: https://github.com/pytorch/pytorch/blob/c29cfe4d02d9d3479961fc1423f87d1fb29aa959/torch/csrc/lazy/ts_backend/ts_native_functions.cpp#L525

torch_xla/csrc/elementwise.cpp
```diff
@@ -227,13 +227,8 @@ xla::XlaOp BuildPrelu(xla::XlaOp input, xla::XlaOp weight) {
...
   const xla::Shape& weight_shape = XlaHelpers::ShapeOfXlaOp(weight);
 
-  int64_t weight_num = xla::ShapeUtil::ElementsIn(weight_shape);
-  int64_t broadcast_dim = weight_num == 1 ? 0 : 1;
-
   xla::XlaOp zero = xla::Zero(input.builder(), input_shape.element_type());
-  xla::XlaOp broadcasted_weight =
-      xla::BroadcastInDim(weight, input_shape.dimensions(), {broadcast_dim});
-  xla::XlaOp product = xla::Mul(input, broadcasted_weight);
+  xla::XlaOp product = xla::Mul(input, weight);
```
alanwaketan, line 231: Can you explain a little bit on this change?

torch_xla/csrc/aten_xla_type.cpp
wonjoolee95: We can definitely do that. Good point, updated.

torch_xla/csrc/aten_xla_type.cpp
wonjoolee95: I gave it a try, seems like the tests are still passing. Good catch, updated.

torch_xla/csrc/elementwise.cpp
wonjoolee95: Now, the `at::Tensor& weight` that is passed to `XLANativeFunctions::_prelu_kernel` already broadcasted. So updated our lowering to remove the unnecessary broadcasting logic. 

Also updated in the PR summary above to include this reasoning.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1052,6 +1052,19 @@ XLANativeFunctions::convolution_backward_overrideable(
                      : at::Tensor());
 }
 
+at::Tensor XLANativeFunctions::copy(const at::Tensor& self,
+                                    const at::Tensor& src, bool non_blocking) {
+  TORCH_LAZY_FN_COUNTER("xla::");
+  return _copy_from(src, self, non_blocking);
```
alanwaketan, line 1058: I was doing the same thing on the other pull request: https://github.com/pytorch/xla/pull/4477. And then it fails some of the CI. I guess this won't be the right approach. Can you maybe exclude this one and then land the remaining changes?

torch_xla/csrc/elementwise.cpp
alanwaketan: Sounds good to me.

alanwaketan (approved):
Please double check the copy lowering. Otherwise, LGTM.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1052,6 +1052,19 @@ XLANativeFunctions::convolution_backward_overrideable(
...
+}
+
+at::Tensor& XLANativeFunctions::copy_(at::Tensor& self, const at::Tensor& src,
```
alanwaketan, line 1061: Wait a minute. Yea, I think this additional copy_ lowering should be able to fix the issue I'm facing on the other pull request.

I hit an internal assertion on the other pull request that basically suggest the auto-generated copy_ lowering is wrong.

If the CIs are all green, I will let you handle the copy/copy_ lowering in your PR here. And then I will do a rebase after your merge.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1052,6 +1052,19 @@ XLANativeFunctions::convolution_backward_overrideable(
...
+                                      bool non_blocking) {
+  TORCH_LAZY_FN_COUNTER("xla::");
+  at::Tensor result = _copy_from(self, src, non_blocking);
+  return result;
```
alanwaketan, line 1065: Would it be better to?
```
_copy_from(self, src, non_blocking);
return self;
```

wonjoolee95: Okay, I noticed that `test_copy_noncontig_xla` test was failing with:
```
```
Looking back at this, it should've been `_copy_from(src, self, non_blocking)`, not `_copy_from(self, src, non_blocking)`. After changing this, the `test_copy_noncontig_xla` test is now passing locally. I'll wait for the CI to verify. Now this op looks like:
```
at::Tensor& XLANativeFunctions::copy_(at::Tensor& self, const at::Tensor& src,
                                      bool non_blocking) {
  _copy_from(src, self, non_blocking);
  return self;
}
```

alanwaketan: Thanks for the update. Can you leave a comment on the commit log to suggest why both copy and copy_ are lowered? It could be simple as we don't want to use the auto-generated copy_ lowering.

wonjoolee95: Sounds good, updated.

wonjoolee95:
CI is all green, will merge this.

PR #4472, opened 2023-01-18
Create Python unit test for dynamic ops.

vanbasten23:


vanbasten23:
cc @miladm @JackCaoG

out.txt
```diff
@@ -0,0 +1,217 @@
++ OPTS=()
```
vanbasten23, line 1: please disregard this file. I'll remove it in the next iteration.

PR #4476, opened 2023-01-19
refactor: move dynamo/TorchXLA bridge to pytorch/xla repo

shunting314:
Check the pytorch side PR for details: https://github.com/pytorch/pytorch/pull/92601 . 

The pytorch one needs to wait for this one to be merged first.

### DIFF
--- a/test/dynamo/test_bridge.py
+++ b/test/dynamo/test_bridge.py
@@ -1,0 +1,213 @@
+import copy
+
+import torch
+
+import torch._dynamo.test_case
+import torch._dynamo.testing
+from functorch.compile import aot_module_simplified, make_boxed_compiler
+from torch._dynamo import disable
+
+import torch_xla.core.dynamo_bridge as bridge
+import torch_xla.core.xla_model as xm
+import torch_xla.debug.metrics as metrics
+from torch import fx, nn
+
+
+class BasicModule(nn.Module):
+
+  def __init__(self):
+    super(BasicModule, self).__init__()
+
+  def forward(self, x, y):
+    return x + y
+
+  def get_random_inputs(self):
+    return (torch.randn(10), torch.randn(10))
+
+
+class MatmulModule(nn.Module):
+
+  def __init__(self):
+    super(MatmulModule, self).__init__()
+
+  def forward(self, x, y):
+    return x @ y
+
+  def get_random_inputs(self):
+    return (torch.randn(5, 100), torch.randn(100, 5))
+
+
+class LinearModule(nn.Module):
+
+  def __init__(self):
+    super().__init__()
+    self.linear = nn.Linear(10, 5)
+
+  def forward(self, x):
+    return self.linear(x)
+
+  def get_random_inputs(self):
+    return (torch.randn(2, 10),)
+
+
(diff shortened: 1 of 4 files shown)


shunting314:
@JackCaoG  should I be worried about the CI break?
1. for the lint error, it looks like xla/ repo wants 2 spaces indent while pytorch/ repo usually uses 4. I can run some quick 'sed' command to format the code. Or let me know if there is already some script in the repo that can do the job.
2. the build error seems to be caused by some unrelated network error.

JackCaoG:
I just restarted the CI, let's see how it goes. I think you also want to update the test under `test/dynamo` since the bridge is under a new path.

shunting314:
Looks like the exsting tests under test/dynamo does not refer to the bridge directly. They specify the backend by string name which is not affected by the refactoring.

wconstab:
should I review the code in this PR as if it is just a code move from pytorch/pytorch or is there also some significant change to the code?

shunting314:
@wconstab  there is no significant change. Just some code movements and some necessary renames.

JackCaoG:
Thanks Shutning, I will try to take a look tmr.

test/dynamo/test_bridge.py
```diff
@@ -0,0 +1,204 @@
+import copy
```
JackCaoG, line 1: I assume this test is moved from upstram `test_torchxla_integration.py`. Let's replace 
https://github.com/pytorch/xla/blob/aff3513234fe344bbf159af5e376e5a45792f5dc/test/run_tests.sh#L135

with the new path then.

test/dynamo/test_num_output.py
```diff
@@ -0,0 +1,105 @@
+import unittest
```
JackCaoG, line 1: ditto, let's add this test to the `run_test.sh`

torch_xla/csrc/tensor.cpp
```diff
@@ -702,12 +702,12 @@ c10::SymNode XLASymNodeImpl::neg() {
                    << " has not been implemented.";
 }
 
-c10::SymNode XLASymNodeImpl::min(const c10::SymNode& other) {
+c10::SymNode XLASymNodeImpl::sym_min(const c10::SymNode& other) {
```
JackCaoG, line 705: if you rebase, these file should be updated on master already

torch_xla/dynamo_bridge.py
```diff
@@ -0,0 +1,317 @@
+import dataclasses
```
JackCaoG, line 1: let's move this file into `torch_xla/core` maybe? we generally don't have python file under `torch_xla` dir directly.

JackCaoG (commented):
Mostly LGTM, thanks Shunting!

JackCaoG (approved):
Let's fix the linter then we can merge this one. Thanks @shunting314 !

JackCaoG:
You can follow https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md#before-submiting-a-pull-request

JackCaoG:
failure is unrealted, merging

PR #4477, opened 2023-01-19
[Functionalization] Fix torch.distributed tests

alanwaketan:
Summary:
This pull request tries to fix the following two tests:
1. test_allgather: allgather uses copy_ internally to update the input and copy_ being decomposed into copy which we need to lower.
2. test_send: send uses t.data to update t’s data which is not supported in functionalization. See https://github.com/pytorch/pytorch/issues/91093. After switching to copy_, it looks like the problem is fixed.

In addition, it fixes two cpp tests as they also require copy.

Test Plan:
CI.

### DIFF
--- a/xla_native_functions.yaml
+++ b/xla_native_functions.yaml
@@ -107,8 +107,6 @@ supported:
   - _adaptive_avg_pool3d_backward
   - _amp_foreach_non_finite_check_and_unscale_
   - _amp_update_scale_
-  - copy
-  - copy_
   - _copy_from
   - _copy_from_and_resize
   - _index_put_impl_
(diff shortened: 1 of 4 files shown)


xla_native_functions.yaml
```diff
@@ -107,6 +107,8 @@ supported:
   - _adaptive_avg_pool3d_backward
   - _amp_foreach_non_finite_check_and_unscale_
   - _amp_update_scale_
+  - copy
```
wonjoolee95, line 110: nit: I believe these two should be already there after the rebase?

JackCaoG (approved):

wonjoolee95 (approved):
LGTM, thanks!

xla_native_functions.yaml
alanwaketan: Let me remove.

PR #4478, opened 2023-01-19
Basic test for `PjrtComputationClient`.

stgpetrovic:
For now, the test does an end-to-end exercise of the client: it transfers tensors to device, compiles a computation (R2 add),
 and transfers the result back to host, ensuring correctness.

This works on CPU for now and is not yet executed during releases.

### DIFF
--- a/third_party/xla_client/pjrt_computation_client_test.cc
+++ b/third_party/xla_client/pjrt_computation_client_test.cc
@@ -61,7 +61,8 @@ TEST(PjRtComputationClientTest, Init) {
       xla::LiteralUtil::CreateR2<float>({{5.0f, 6.0f}, {7.0f, 8.0f}});
 
   // Compile the graph.
-  auto cptr = client->Compile(std::move(instances));
+  std::vector<ComputationClient::ComputationPtr> computations =
+      client->Compile(std::move(instances));
 
   // Copy inputs to device.
   ComputationClient::ExecuteComputationOptions options{};
@@ -70,9 +71,9 @@ TEST(PjRtComputationClientTest, Init) {
       TensorSourceFromLiteral(device, literal_y)};
 
   // Execute the graph.
-  auto results = client->ExecuteComputation(
-      *cptr[0], client->TransferToServer(absl::MakeConstSpan(args)), device,
-      options);
+  std::vector<ComputationClient::DataPtr> results = client->ExecuteComputation(
+      *computations[0], client->TransferToServer(absl::MakeConstSpan(args)),
+      device, options);
 
   // Copy the output from device back to host and assert correctness..
   ASSERT_EQ(results.size(), 1);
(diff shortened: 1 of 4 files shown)


third_party/xla_client/pjrt_computation_client_test.cc
```diff
@@ -0,0 +1,86 @@
...
+      &shape));
+
+  // Prepare inputs.
+  xla::Literal literal_x =
+      xla::LiteralUtil::CreateR2<float>({{1.0f, 2.0f}, {3.0f, 4.0f}});
+  xla::Literal literal_y =
+      xla::LiteralUtil::CreateR2<float>({{5.0f, 6.0f}, {7.0f, 8.0f}});
+
+  // Compile the graph.
+  auto cptr = client->Compile(std::move(instances));
```
will-cromar, line 64: nit: can you change this to `std::vector<ComputationPtr> computations`? The interfaces of `PjRtClient` and `ComputationClient` are very similar, so I always get the exact return types mixed up. Same for `results` and `result_literals`

Confession: I just had to look up the return type here because I forgot that `Compile` returned a vector

third_party/xla_client/pjrt_computation_client_test.cc
```diff
@@ -0,0 +1,86 @@
...
+    std::memcpy(dest_buffer, literal.data<float>().data(),
+                dest_buffer_size * sizeof(literal.data<float>().data()));
+  };
+  return ComputationClient::TensorSource(literal.shape(), device,
+                                         std::move(populate_fn));
+}
+
+TEST(PjRtComputationClientTest, Init) {
+  // Get a CPU client.
+  tsl::setenv("PJRT_DEVICE", "CPU", true);
```
will-cromar, line 45: Can you change `overwrite` to `false`? Let's make CPU a default, but I do want to be able to run this test on TPU/GPU (e.g. `PJRT_DEVICE=TPU`)

will-cromar (commented):
This is really helpful. Thanks!

JackCaoG (commented):
This is great, thanks Goran!

will-cromar (approved):
LGTM. Thanks! Feel free to merge once we fix the CI.

third_party/xla_client/pjrt_computation_client_test.cc
```diff
@@ -0,0 +1,87 @@
...
+                         void* dest_buffer, size_t dest_buffer_size) {
+    std::memcpy(dest_buffer, literal.data<float>().data(),
+                dest_buffer_size * sizeof(literal.data<float>().data()));
+  };
+  return ComputationClient::TensorSource(literal.shape(), device,
+                                         std::move(populate_fn));
+}
+
+TEST(PjRtComputationClientTest, Init) {
+  // Get a CPU client.
```
vanbasten23, line 44: Great comments. Very helpful! Thanks.

vanbasten23:
Thanks for adding the test. Can you also add instruction on how to run the test? Also, will you add the test to the CI?

JackCaoG (approved):
Thanks!

PR #4485, opened 2023-01-20
[FSDP] Replace all .data usages with modern approaches

alanwaketan:
Summary:
The .data API is deprecated by PyTorch. Let's replace all .data usages with modern approaches.

1. we replace `.data =` with `.set_`. We also lower `.set_` within this PR.
2. we then remove other `.data` given there is no needs to detach.

There is one caveat as we now requires torch.no_grad() to run inferences. Otherwise, it will introduce weird error. Need to follow up later.

Test Plan:
1. PJRT_DEVICE=TPU python test/test_train_mp_mnist_fsdp_with_ckpt.py --batch_size 16 --drop_last --num_epochs 2  --use_nested_fsdp
Expected Max Accuracy: 98.88%+

3. python test/test_operations.py -v -k test_set


alanwaketan:
@JackCaoG @ronghanghu Does FSDP have any unit tests or any faster way to validate my change? Running one epoch for every increment change is not fun...

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -658,7 +658,7 @@ def _shard_parameters_(self, params_to_shard) -> None:
       self.sharded_params.append(p_shard)
       # Free the full parameter storage (here we free its `.data`) but keep the tensor itself
       # for auto-grad tracing (like `torch.autograd.Variable` before the tensor-variable merge).
-      p.data = p.data.new_zeros(1)
+      p.copy_(p.data.new_zeros(1))
```
alanwaketan, line 661: Is there a way for us to validate the data is still freed as expected?

JackCaoG: there is a counter corresponding to release of buffer.. let me look.

alanwaketan:
Also, not sure if there is any sort of correctness test without comparing the learning curves?

ronghanghu:
> @JackCaoG @ronghanghu Does FSDP have any unit tests or any faster way to validate my change? Running one epoch for every increment change is not fun...

I think the MNIST example is faster -- one can do 
```bash
PJRT_DEVICE=TPU python3 test/test_train_mp_mnist_fsdp_with_ckpt.py \
  --batch_size 16 --drop_last --num_epochs 2 \
  --use_nested_fsdp
```

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1365,8 +1365,9 @@ def _rebuild_full_params(self,
         if apply_opt_barrier:
           self.optimization_barrier_op([p_padded])
         if self._shard_param_on_dim_0:
-          p.data = p_padded[:p_shard._orig_size[0]]
+          p.copy_(p_padded[:p_shard._orig_size[0]])
         else:
+          # RuntimeError: /workspaces/work/pytorch/xla/torch_xla/csrc/data_ops.cpp:135 : Check failed: input_sizes.size() <= output_sizes.size() (4 vs. 1)
```
alanwaketan, line 1370: This is the error when I was trying to do the replacement.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1398,6 +1399,7 @@ def _free_full_params(self,
     for p in full_params:
       if p._has_full_param:
         # free the original full parameter
+        # RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [XLAFloatType [1000, 2048]] is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```
alanwaketan, line 1402: This is the error when I was trying to do the replacement.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1433,9 +1435,11 @@ def _apply_opt_barrier_to_params_and_tensors(
                                  dependency_tensors)
 
     for p, p_data in zip(p_list, p_data_list):
+      # RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [XLAFloatType [1000, 2048]] is at version 3; expected version 1 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```
alanwaketan, line 1438: This is the error when I was trying to do the replacement.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
JackCaoG: for xrt https://github.com/pytorch/xla/blob/master/third_party/xla_client/xrt_computation_client.cc#L1285 should be called, I think we have similar function for PJRT.

alanwaketan: Let me double check. Thanks, Jack.

alanwaketan:
> > @JackCaoG @ronghanghu Does FSDP have any unit tests or any faster way to validate my change? Running one epoch for every increment change is not fun...
> 
> I think the MNIST example is faster -- one can do
> 
> ```shell
> PJRT_DEVICE=TPU python3 test/test_train_mp_mnist_fsdp_with_ckpt.py \
>   --batch_size 16 --drop_last --num_epochs 2 \
>   --use_nested_fsdp
> ```
> 
> It should get around `Max Accuracy: 98.88%`

Thanks, Ronghang!!!

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
ronghanghu: The tricky part was that we hoped to free a tensor `t`'s storage (and reconstruct it later via `t.data = xxx`) without bumping its version counter `t._version`.

I think every time we call `.data` it creates a new Tensor that shares storage but not the version counter. But `.copy_` changes the version counter

alanwaketan: I also tried to set it to None but PyTorch as a type check... The new value must be a Tensor.

alanwaketan:
I tried https://github.com/pytorch/pytorch/pull/92924/files and it looks like it fixes our problems about autograd complaining the tensor versions and still produced:
```
PJRT_DEVICE=TPU python3 test/test_train_mp_mnist_fsdp_with_ckpt.py --batch_size 16 --drop_last --num_epochs 2  --use_nested_fsdp
Max Accuracy: 98.94%
```

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
alanwaketan: @ronghanghu Any thoughts about this last error?

alanwaketan:
Once we fix all the error, then we need to double check:
1) speed with real data and fake data,
2) memory usage with real data and fake data.

alanwaketan:
Verified that with this change, the https://github.com/pytorch/xla/pull/4158/files#diff-d7a18ad7c8a5902f1c95d906f4d6f4a86c3578c816ae9f9c2c050cfab1029e03R139 can work again on the functionalization branch.

Now we just need to make sure that the memory are freed properly after the `.data =` usages are removed.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
alanwaketan: Hit a crash while trying to replace all `.copy_` with `.set_` and then discovered that we need to lower `- set_.source_Tensor`. Trying to lower that op and see if it works.

alanwaketan: After commenting out the eval part, things can continue without any problems...

ronghanghu: Hmm, this is typically because some hooks didn't fire -- let me take a closer look at it

alanwaketan: Thanks, Ronghang. Yea, I'm not sure why would eval overwrites hooks.

ronghanghu: Hi @alanwaketan, I wonder could you verify whether the issue still occurs if we put the test loop in this line (https://github.com/pytorch/xla/blob/3f091bdd43944ee23a6acf17ef592f73adb06673/test/test_train_mp_imagenet_fsdp.py#L356) under `torch.no_grad`? (This shouldn't be a requirement for the test loop, but could help debug what's wrong.) If my guess is correct, then the error above should disappear when testing under `torch.no_grad`.

My guess is that during evaluation, the `_pre_backward_hook` somehow fired unexpectedly, causing the model to go into `TrainingState.BACKWARD_PRE` state. This hook needs to be registered during the forward pass unless it's under `torch.no_grad` (since we don't know if this is in an evaluation loop and there is a chance that the user would later call `loss.backward` even when the model is in eval mode). But registering this hook alone doesn't cause it to be tiggered, and the previous (expected) behavior is that this look should only fire when the user actually call something like `loss.backward()` -- but somehow this behavior is broken.

alanwaketan: @ronghanghu You are amazing... It solves the problem...
```
PJRT_DEVICE=TPU python3 test/test_train_mp_mnist_fsdp_with_ckpt.py --batch_size 16 --drop_last --num_epochs 2 --use_nested_fsdp
...
Epoch 2 test end 20:16:26, Accuracy=98.94
checkpoint saved to /tmp/mnist-fsdp/final_ckpt_rank-00000000-of-00000004.pth
checkpoint saved to /tmp/mnist-fsdp/final_ckpt_rank-00000002-of-00000004.pth
checkpoint saved to /tmp/mnist-fsdp/final_ckpt_rank-00000001-of-00000004.pth
checkpoint saved to /tmp/mnist-fsdp/final_ckpt_rank-00000003-of-00000004.pth
found 4 checkpoint files in /tmp/mnist-fsdp/final_ckpt_rank-*-of-*.pth
saved consolidated model to /tmp/mnist-fsdp/final_ckpt_consolidated.pth
Checkpoint consolidated, Accuracy=98.97 (note: it can be slightly different from the final training accuracy due to non-sync BatchNorm2d in the model)
Max Accuracy: 98.94%
```

ronghanghu: I see, thanks for verifying it quickly! Then it's weird why `_pre_backward_hook` would fire during a test loop when we don't call `.backward()` -- the expected behavior is that this hook wouldn't fire in such cases

ronghanghu: @alanwaketan I see, thanks for verifying it quickly! Then it's weird why _pre_backward_hook would now fire during a test loop when we don't call .backward() -- the expected behavior is that this hook wouldn't fire in such cases.

alanwaketan: On the other hand, I have observed that the number of `_post_backward_hook` being fired is incorrect for epoch 2. Maybe, without torch.no_grad, it actually modifies some other aspects of the model.

alanwaketan:
Just added `test_data_handle_release` which verify that `.set` can release the underneath XLATensor (BackendDataHandle) as `.data=`.

alanwaketan:
For performance comparison using rates generated by `PJRT_DEVICE=TPU python test/test_train_mp_mnist_fsdp_with_ckpt.py --batch_size 16 --drop_last --num_epochs 2 --use_nested_fsdp`:
```
Master:
Type      Mean    Median    90th %    Std Dev    CV
------  ------  --------  --------  ---------  ----
Rate    203.10    203.49    205.56       2.83  0.01

This branch:
Type      Mean    Median    90th %    Std Dev    CV
------  ------  --------  --------  ---------  ----
Rate    192.51    193.18    194.43       2.72  0.01
```

So we can see that after this change the performance is regressed around 5%.

P.S. Testbed: V4-8, PJRT, Debug build.

alanwaketan:
Fixed the performance regression:
```
Original FSDP:
Type      Mean    Median    90th %    Std Dev    CV
------  ------  --------  --------  ---------  ----
Rate    203.10    203.49    205.56       2.83  0.01

FSDP with vanilla set:
Type      Mean    Median    90th %    Std Dev    CV
------  ------  --------  --------  ---------  ----
Rate    192.51    193.18    194.43       2.72  0.01

FSDP with set dummy:
Type      Mean    Median    90th %    Std Dev    CV
------  ------  --------  --------  ---------  ----
Rate    218.24    219.15    220.75       3.31  0.02
```

Actually we are seeing a performance gain with the latest change. That bothers me. Will point out the code change in the PR.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1398,11 +1394,12 @@ def _free_full_params(self,
...
       if p._has_full_param:
         # free the original full parameter
-        p.data = self._dummy_data_placeholder
+        with torch.autograd._unsafe_preserve_version_counter(p):
+          p.set_(self._dummy_data_placeholder)
         p._has_full_param = False
 
     if apply_opt_barrier:
-      self._apply_opt_barrier_to_params_and_tensors(full_params, sharded_params,
+      self._apply_opt_barrier_to_params_and_tensors([p for p in full_params if p._has_full_param], sharded_params,
```
alanwaketan, line 1402: @ronghanghu Wonder why we want to `_apply_opt_barrier_to_params_and_tensors`  to p even if it doesn't hold the full param?

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1341,7 +1336,7 @@ def _rebuild_full_params(self,
       dependency_tensors = []
 
     if apply_opt_barrier:
-      self._apply_opt_barrier_to_params_and_tensors(self.full_params,
+      self._apply_opt_barrier_to_params_and_tensors([p for p in self.full_params if p._has_full_param],
```
alanwaketan, line 1339: @ronghanghu Wonder why we want to `_apply_opt_barrier_to_params_and_tensors`  to p even if it doesn't hold the full param?

ronghanghu: @alanwaketan Hmm, it probably isn't necessary to run `_apply_opt_barrier_to_params_and_tensors` on those tensors without full parameters. Back in April 2022, I had some struggle getting the optimization barrier to work. I ended up adding this to a lot of places to finally get it working, but with hindsight it's probably not needed in several places. (The main purpose of this optimization barrier is to prevent the XLA compiler from de-duplicating the two `_rebuild_full_params` in forward and backward passes and holding the full parameters after forward -- which saves collective ops but causes OOM.)

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
ronghanghu: @alanwaketan Yeah, you're right! It's probably not needed here on those `p` without full param. (I probably introduced several unnecessary optimization barriers in the implementation earlier.)

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
ronghanghu: @alanwaketan When looking at https://github.com/pytorch/xla/pull/4670, I suspect the issue there might be similar to what we had here -- somehow the `_pre_backward_hook` fired unexpectedly when we only do forward pass without a backward pass (e.g. when a submodule didn't participate in backward, or when we are running eval).

alanwaketan:
@JackCaoG @ronghanghu I think the PR is ready for reviews.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
alanwaketan: Thanks for the heads up. I added a TODO.

ronghanghu:
Thanks @alanwaketan! I'll take another look by tomorrow.

test/test_train_mp_imagenet_fsdp.py
```diff
@@ -353,7 +353,10 @@ def test_loop_fn(loader, epoch):
     run_eval = ((not FLAGS.test_only_at_end and
                  epoch % FLAGS.eval_interval == 0) or epoch == FLAGS.num_epochs)
     if run_eval:
-      accuracy = test_loop_fn(test_device_loader, epoch)
+      # TODO(alanwaketan): Investigate why we inference would impact
+      # the next epoch's training.
```
JackCaoG, line 357: what's the issue here?

JackCaoG: nvm, saw the comment, I think you have a typo here(remove `we`)

JackCaoG (approved):
Thanks!

ronghanghu (approved):

alanwaketan:
Thanks @ronghanghu and @JackCaoG for approving the pull request. Rerunning the CPU test as the failure doesn't seem to be related. Will land after that.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
ronghanghu: Minor update: I suspect this change in `_pre_backward_hook` might be related to the recent upstream change in PyTorch: [Hooks registered on tensor to always run, even if they are the inputs to .grad() (#85849)](https://github.com/pytorch/pytorch/pull/85849), as mentioned in the [PT 2.0 release note](https://gist.github.com/drisspg/09547b9ab16c5e236c496f120547c330#hooks-registered-on-tensor-to-always-run-even-if-they-are-the-inputs-to-grad-85849)

PR #4487, opened 2023-01-21
Fix HLO dump file names for PJRT

will-cromar:
`SaveTensorsGraphInfo` relied on a global save file path and ordinal (as an env var) and wrote all HLO dumps to the same file. Use thread local ordinal and save file path to support PJRT.

Added a PJRT multi-CPU test and tested manually with XRT.

### DIFF
--- a/test/pjrt/test_experimental_pjrt_multi_cpu.py
+++ b/test/pjrt/test_experimental_pjrt_multi_cpu.py
@@ -108,6 +108,23 @@ class TestExperimentalPjrtMultiCpu(parameterized.TestCase):
     indices = sorted(queue.get(block=False) for _ in range(queue.qsize()))
     self.assertListEqual(indices, list(range(4)))
 
+  @staticmethod
+  def _hlo_dump(tmpdir: str):
+    os.environ['XLA_SAVE_TENSORS_FMT'] = 'hlo'
+    os.environ['XLA_SAVE_TENSORS_FILE'] = os.path.join(tmpdir, 'save.hlo')
+
+    x = torch.randn((3, 3), device=xm.xla_device())
+    xm.mark_step()
+    x.cpu()
+
+  def test_hlo_dump(self):
+    tmpdir = self.create_tempdir().full_path
+    pjrt._run_multiprocess(self._hlo_dump, tmpdir)
+
+    files = os.listdir(tmpdir)
+    for i in range(4):
+      self.assertIn(f'save.hlo.{i}', files)
+
 
 if __name__ == '__main__':
   absltest.main()
--- a/torch_xla/csrc/debug_util.cpp
+++ b/torch_xla/csrc/debug_util.cpp
@@ -116,8 +116,8 @@ void DebugUtil::SaveTensorsGraphInfo(const char* name,
                                      absl::Span<const XLATensorPtr> tensors,
                                      const std::vector<size_t>* indices,
                                      GraphFormat format) {
-  static const std::string save_file =
-      xla::sys_util::GetEnvOrdinalPath("XLA_SAVE_TENSORS_FILE", "");
+  thread_local const std::string save_file = xla::sys_util::GetEnvOrdinalPath(
+      "XLA_SAVE_TENSORS_FILE", "", GetCurrentDevice().ordinal());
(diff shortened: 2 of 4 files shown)


test/pjrt/test_experimental_pjrt_multi_cpu.py
```diff
@@ -108,6 +108,23 @@ def test_spawn(self, spawn):
...
+    xm.mark_step()
+    x.cpu()
+
+  def test_hlo_dump(self):
+    tmpdir = self.create_tempdir().full_path
+    pjrt._run_multiprocess(self._hlo_dump, tmpdir)
+
+    files = os.listdir(tmpdir)
+    for i in range(4):
+      self.assertIn(f'save.hlo.{i}', files)
```
vanbasten23, line 126: do you need to clean up the hlo dump at the end of the test?

test/pjrt/test_experimental_pjrt_multi_cpu.py
```diff
@@ -108,6 +108,23 @@ def test_spawn(self, spawn):
     indices = sorted(queue.get(block=False) for _ in range(queue.qsize()))
     self.assertListEqual(indices, list(range(4)))
 
+  @staticmethod
+  def _hlo_dump(tmpdir: str):
+    os.environ['XLA_SAVE_TENSORS_FMT'] = 'hlo'
+    os.environ['XLA_SAVE_TENSORS_FILE'] = os.path.join(tmpdir, 'save.hlo')
```
JackCaoG, line 114: hmm, I actually don't know if `os.environ` is local to this function. Will `os.environ` got reset upon return of this function?

torch_xla/csrc/debug_util.cpp
```diff
@@ -116,8 +116,8 @@ void DebugUtil::SaveTensorsGraphInfo(const char* name,
                                      absl::Span<const XLATensorPtr> tensors,
                                      const std::vector<size_t>* indices,
                                      GraphFormat format) {
-  static const std::string save_file =
-      xla::sys_util::GetEnvOrdinalPath("XLA_SAVE_TENSORS_FILE", "");
+  thread_local const std::string save_file = xla::sys_util::GetEnvOrdinalPath(
```
JackCaoG, line 119: haha I need to google first to confirm that `thread_local` is implicitly static.

test/pjrt/test_experimental_pjrt_multi_cpu.py
will-cromar: It's not local, but we only change `os.environ` in the sub process, so it does get reset between tests

torch_xla/csrc/debug_util.cpp
will-cromar: Same! https://stackoverflow.com/a/22794640

test/pjrt/test_experimental_pjrt_multi_cpu.py
will-cromar: No. We got the temp dir from the test class, which cleans up temp files at the end

test/pjrt/test_experimental_pjrt_multi_cpu.py
```diff
@@ -108,6 +108,23 @@ def test_spawn(self, spawn):
...
+
+    x = torch.randn((3, 3), device=xm.xla_device())
+    xm.mark_step()
+    x.cpu()
+
+  def test_hlo_dump(self):
```
JackCaoG, line 120: can we add the same test under maybe `test_operations`? I want to make sure this change does not break non-pjrt work loads. If it is too complicated we can do it in a separate pr.

will-cromar: This is actually already tested on XRT under this test case: https://github.com/pytorch/xla/blob/master/test/dynamo/test_dynamo_graph_dump.py

I know that because it caught a bug in the name of my files :smile:

JackCaoG: lol OK, I forgot I added that test.

JackCaoG (approved):

PR #4488, opened 2023-01-21
[NOT FOR REVIEW] Support dynamism in empty_symint

vanbasten23:


torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1247,6 +1250,29 @@ XLATensorPtr full(absl::Span<const int64_t> size, const at::Scalar& fill_value,
...
+    }
+    return true;
+  })) << "Dimensions cannot be negative numbers";
+
+  SymIntElements size_elements = SymIntElements(sym_size);
+  // /workspaces/work/pytorch/xla/torch_xla/csrc/layout_manager.h:20:12: note: candidate function not viable: no known conversion from 'std::vector<bool>' to 'absl::Span<const bool>' for 2nd argument
+  xla::Shape shape =
+      MakeArrayShapeFromDimensions(size_elements.GetUpperBounds(),
+                                   //absl::Span<const bool>(size_elements.GetDynamicDims().data()),
+                                   size_elements.GetDynamicDims(), // absl::Span<const bool>(size_elements.GetDynamicDims().data()) doesn't work either.
```
vanbasten23, line 1268: Just want to check the build failure here after my investigation. The build now fails with `note: candidate function not viable: no known conversion from 'std::vector<bool>' to 'absl::Span<const bool>' for 2nd argument` due to the 2nd argument of `MakeArrayShapeFromDimensions`.

I wonder if there is some code sample that can convert a 'std::vector<bool>' to 'absl::Span<const bool>' or if there are other way to fix it

alanwaketan: Hmm, std::vector should be able to be implicitly converted to a absl::Span. Let me double check. Will get back to you.

alanwaketan: Try the following:
```
diff --git a/torch_xla/csrc/tensor_methods.cpp b/torch_xla/csrc/tensor_methods.cpp
index 32c5e92c..e2694ef2 100644
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -1261,11 +1261,13 @@ XLATensorPtr empty(at::SymIntArrayRef sym_size, const at::Scalar& fill_value,
   })) << "Dimensions cannot be negative numbers";
 
   SymIntElements size_elements = SymIntElements(sym_size);
+  auto dynamic_dims_tmp = size_elements.GetDynamicDims();
+  absl::InlinedVector<bool, xla::InlineRank()> dynamic_dims(dynamic_dims_tmp.begin(), dynamic_dims_tmp.end());
   // /workspaces/work/pytorch/xla/torch_xla/csrc/layout_manager.h:20:12: note: candidate function not viable: no known conversion from 'std::vector<bool>'
 to 'absl::Span<const bool>' for 2nd argument
   xla::Shape shape =
       MakeArrayShapeFromDimensions(size_elements.GetUpperBounds(),
                                    //absl::Span<const bool>(size_elements.GetDynamicDims().data()),
-                                   size_elements.GetDynamicDims(), // absl::Span<const bool>(size_elements.GetDynamicDims().data()) doesn't work eithe
r.
+                                   dynamic_dims, // absl::Span<const bool>(size_elements.GetDynamicDims().data()) doesn't work either.
                                    MakeXlaPrimitiveType(scalar_type, &device),
                                    static_cast<XlaDeviceType>(device.type()));
   return XLATensor::Create(
```

It looks like vector<bool> is one of the special kind of vector: https://stackoverflow.com/questions/71161221/std-vector-data-return-void. Maybe we have a mismatch between our C++ version and absl version, and therefore the implicit convection doesn't work. After trying a few hacks, the above is the one actually works. BTW, please put a not to suggest we could potentially remove the hack in the future.

vanbasten23: Thanks for the suggestion! It compiles but it fails when I run it: `dynamic_dims.size()` returns 42453634 which doesn't make sense. I think maybe I should try replace vector<bool> with absl::InlinedVector<bool, xla::InlineRank()>.

alanwaketan: Another way to try is to init dynamic_dims with a for loop instead of using the iterator based ctor.

PR #4489, opened 2023-01-21
Switch to pytorch sccache

huydhn:
Recently, there are some flaky sccache start-up failures on PyTorch when building XLA, for example:

* https://hud.pytorch.org/pytorch/pytorch/commit/c5cb46ecdbc0c21697d99e240116bc3411674744

The full list can be found [here](https://hud.pytorch.org/failure/sccache%3A%20error%3A%20Timed%20out%20waiting%20for%20server%20startup).  The error, strangely, comes only from XLA.

It turns out that XLA `pytorch/xla_base:v0.6` uses upstream sccache from `https://github.com/mozilla/sccache.git` while the rest of PyTorch CI uses a custom fork from `https://github.com/pytorch/sccache.git` as defined in https://github.com/pytorch/pytorch/blob/master/.circleci/docker/common/install_cache.sh#L12

IMO, it's easier to stick to `https://github.com/pytorch/sccache.git` as the rest of the CI till we have the capacity to do the switch back to upstream sccache.

AFAIK, `https://github.com/pytorch/sccache.git` has some fixes to work with nvcc.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,8 +23,8 @@ launch_docker_and_build: &launch_docker_and_build
     # To allow the upstream workflow to access PyTorch/XLA build images, we
     # need to have them in the ECR. This is not expensive, and only pushes it
     # if image layers are not present in the repo.
-    docker tag ${GCR_DOCKER_IMAGE} ${ECR_DOCKER_IMAGE_BASE}:v0.6 >/dev/null
-    docker push ${ECR_DOCKER_IMAGE_BASE}:v0.6 >/dev/null
+    docker tag ${GCR_DOCKER_IMAGE} ${ECR_DOCKER_IMAGE_BASE}:v0.7 >/dev/null
+    docker push ${ECR_DOCKER_IMAGE_BASE}:v0.7 >/dev/null
     sudo pkill -SIGHUP dockerd
     # To enable remote caching, use SCCACHE_BUCKET=${SCCACHE_BUCKET}
     # To disable remote caching, use SCCACHE_BUCKET=${LOCAL_BUCKET}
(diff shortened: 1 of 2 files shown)


JackCaoG:
@yeounoh Can you take this one?

huydhn:
> @yeounoh Can you take this one?

Thank @JackCaoG.  The change looks simple enough and build seems to pass.  Let me know if this works as expected

yeounoh:
@huydhn will verify, thank you!

.circleci/config.yml
```diff
@@ -23,8 +23,8 @@ launch_docker_and_build: &launch_docker_and_build
     # To allow the upstream workflow to access PyTorch/XLA build images, we
     # need to have them in the ECR. This is not expensive, and only pushes it
     # if image layers are not present in the repo.
-    docker tag ${GCR_DOCKER_IMAGE} ${ECR_DOCKER_IMAGE_BASE}:v0.6 >/dev/null
-    docker push ${ECR_DOCKER_IMAGE_BASE}:v0.6 >/dev/null
+    docker tag ${GCR_DOCKER_IMAGE} ${ECR_DOCKER_IMAGE_BASE}:v0.7 >/dev/null
```
yeounoh, line 26: Could we use `v0.8`? `v0.7` was already taken, so I pushed under `v0.8` while testing/verifying your image in our CI.

yeounoh:
Oh, I think we also need to rebase before merging 🙏

.circleci/config.yml
yeounoh: Actually, I can merge and bump up the version separately.

huydhn:
@yeounoh Thank you for doing the rebase for me.  I'll update XLA docker image on PyTorch CI to v0.8 accordingly

PR #4490, opened 2023-01-23
Forward fix for 81763

lezcano:
See https://github.com/pytorch/pytorch/pull/81763#issuecomment-1399187907 and https://github.com/pytorch/pytorch/pull/81763#issuecomment-1399318349

Now FrobeniusNorm is just left for backwards compatibility, and it dispatches to `at::norm`, which dispatches to `xla::norm`. As such, we don't need the `sqrt` counters as they will not be updated. We remove these checks as those from `TestNorm*` do not have them either.

It's not very clear to me how to handle this merge, as it's a forward fix. I guess we should simply merge it and then revert https://github.com/pytorch/pytorch/pull/92634 in master? cc @huydhn @kit1980


kit1980:
Before reverting https://github.com/pytorch/pytorch/pull/92634 we'll need also to update xla hash on pytorch/pytorch.

kit1980 (approved):

PR #4491, opened 2023-01-23
ResNet optimized

chandrasekhard2:
In progress.

### DIFF
--- a/test/pjrt/test_train_pjrt_imagenet.py
+++ b/test/pjrt/test_train_pjrt_imagenet.py
@@ -1,4 +1,5 @@
 import args_parse
+import time
 
 SUPPORTED_MODELS = [
     'alexnet', 'densenet121', 'densenet161', 'densenet169', 'densenet201',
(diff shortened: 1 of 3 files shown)


test/pjrt/test_train_pjrt_imagenet.py
```diff
@@ -1,4 +1,5 @@
 import args_parse
+import time
```
JackCaoG, line 2: this test is not really being used anymore. We have been using the `test/test_train_mp_imagenet.py` directly since PJRT supports `xmp.spawn` now.

JackCaoG:
@chandrasekhard2 @miladm If we want to talk about this speed improvement on PyTorch 2.0 we want to merge this change before the branch cut(EOW)

test/lars.py
```diff
@@ -0,0 +1,175 @@
...
+            for p in group['params']:
+                if p.grad is None:
+                    continue
+                decayed_grad = p.grad
+                scaled_lr = lr
+                '''
+                if lars:
+                    w_norm = torch.norm(p)
+                    g_norm = torch.norm(p.grad)
+                    trust_ratio = torch.where(
```
chandrasekhard2, line 101: Training hangs at this step. If commented it runs as expected.

PR #4494, opened 2023-01-24
[Functionalization] Fix SPMD tests

alanwaketan:
Summary:
This pull request tries to fix three SPMD tests that fails with Functionalization.
1. test_model_weight_metrics: Before Functionalization, model.weight is an at::tensor stored in the XLATensor, but it's an IRValue now. Therefore, the test no longer holds true.
2. test_optimizer_step_with_sharding/test_inplace_add_with_sharding: we propagate the sharding spec via _propagate_xla_data which  is an operator that help us during in-place op functionalized pass.

Test Plan:
CI.

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -110,7 +110,7 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     xm.mark_step()  # mark_step should preserve the sharding
     self.assertEqual(sharding_spec, torch_xla._XLAC._get_xla_sharding_spec(xt))
 
-  @unittest.skip("fails with functionalization")
+  # @unittest.skip("fails with functionalization")
   def test_optimizer_step_with_sharding(self):
     model = self.SimpleLinear().to(xm.xla_device())
     xs.mark_sharding(model.fc1.weight, self._get_mesh((1, self.n_devices)),
@@ -132,7 +132,7 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertEqual(sharding_spec,
                      torch_xla._XLAC._get_xla_sharding_spec(model.fc1.weight))
 
-  @unittest.skip("fails with functionalization")
+  # @unittest.skip("fails with functionalization")
   def test_inplace_add_with_sharding(self):
     xt = torch.ones(2, 2).to(xm.xla_device())
     xs.mark_sharding(xt, self._get_mesh((1, self.n_devices)), (0, 1))
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -110,7 +110,6 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     xm.mark_step()  # mark_step should preserve the sharding
     self.assertEqual(sharding_spec, torch_xla._XLAC._get_xla_sharding_spec(xt))
 
-  # @unittest.skip("fails with functionalization")
   def test_optimizer_step_with_sharding(self):
     model = self.SimpleLinear().to(xm.xla_device())
     xs.mark_sharding(model.fc1.weight, self._get_mesh((1, self.n_devices)),
@@ -132,7 +131,6 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertEqual(sharding_spec,
                      torch_xla._XLAC._get_xla_sharding_spec(model.fc1.weight))
 
-  # @unittest.skip("fails with functionalization")
   def test_inplace_add_with_sharding(self):
     xt = torch.ones(2, 2).to(xm.xla_device())
     xs.mark_sharding(xt, self._get_mesh((1, self.n_devices)), (0, 1))
--- a/test/spmd/test_xla_virtual_device.py
+++ b/test/spmd/test_xla_virtual_device.py
@@ -42,15 +42,13 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
     self.assertIn("VirtualDeviceUsage", met.counter_names())
     self.assertNotEqual(met.counter_value("VirtualDeviceUsage"), 0)
 
-  @unittest.skip("fails with functionalization")
   def test_model_weight_metrics(self):
     met.clear_counters()
     partition_spec = (0, 1)
     model = nn.Linear(128, 64).to(xm.xla_device())
     xs.mark_sharding(model.weight, self._get_mesh((1, self.n_devices)),
                      partition_spec)
-    self.assertIn("VirtualDeviceUsage", met.counter_names())
-    self.assertNotEqual(met.counter_value("VirtualDeviceUsage"), 0)
+    self.assertNotIn("VirtualDeviceUsage", met.counter_names())
 
   def test_no_sharding(self):
     t1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -2359,10 +2359,17 @@ void XLANativeFunctions::_propagate_xla_data(const at::Tensor& input,
   TORCH_LAZY_FN_COUNTER("xla::");
   // This op is only called when functionalize pass is transforming an in-place
   // op. Therefore, we can populate some meta data to maintain any optimization
-  // for in-place ops we have in hands. 1) Aid XLA's InputOutputAlias.
+  // for in-place ops we have in hands.
+
+  // 1) Aid XLA's InputOutputAlias.
   auto input_tensor = bridge::GetXlaTensor(input);
   auto output_tensor = bridge::GetXlaTensor(output);
   output_tensor->data()->alias_id = input_tensor->GetUniqueId();
+
+  // 2) Aid SPMD.
+  if (input_tensor->sharding_spec()) {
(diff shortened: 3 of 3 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1501,9 +1501,12 @@ void InitXlaModuleBindings(py::module m) {
...
-                xtensor->CurrentDataHandle()->HasValue())
-          << "Cannot shard tensor. Data not present on any device.";
+      {
+        XLAGraphExecutor::Get()->DeviceBarrier(xtensor->GetDevice());
+        XLA_CHECK(xtensor->GetXlaData() != nullptr &&
+                  xtensor->CurrentDataHandle()->HasValue())
+            << "Cannot shard tensor. Data not present on any device.";
+      }
       std::vector<XLATensorPtr> xla_tensors{xtensor};
       cpu_tensor = XLAGraphExecutor::Get()->GetTensors(&xla_tensors)[0];
```
alanwaketan, line 1511: Can we replace L1501 to 1511 with:
```
cpu_tensor = xtensor.ToTensor(true);
```

I tried but the figure out that it could break test_outbound_data_metrics. outbound_with_virtual_device will be equal to outbound_without_virtual_device.

steventk-g: I'm not sure if `ToTensor` does what we need here.

alanwaketan: So what exactly do you need?

test/spmd/test_xla_virtual_device.py
```diff
@@ -42,16 +42,6 @@ def test_metrics_recorded(self):
     self.assertIn("VirtualDeviceUsage", met.counter_names())
     self.assertNotEqual(met.counter_value("VirtualDeviceUsage"), 0)
 
-  @unittest.skip("fails with functionalization")
```
steventk-g, line 45: Rather than removing the test, should we add a metric for the other codepath (that handles IR) and check that it's used?

test/spmd/test_xla_sharding.py
```diff
@@ -126,17 +125,14 @@ def test_optimizer_step_with_sharding(self):
       loss.backward()
       optimizer.step()
       xm.mark_step()
+      # After functionalization, in-place ops like tensor.add_ are no longer availble. Therfore, we need
+      # a new way to carry the sharding spec between iterations. This is a tentative fix.
+      xs.mark_sharding(model.fc1.weight, self._get_mesh((1, self.n_devices)),
```
JackCaoG, line 130: the purpose of this test is to make sure in a regular pytorch training case, user only need to call  `mark_sharding` once on weight once. This is currently broken. Since we don't want to ask user to add `mark_sharding` on every step, we can just leave this test disabled.

test/spmd/test_xla_sharding.py
```diff
@@ -126,17 +125,14 @@ def test_optimizer_step_with_sharding(self):
...
     self.assertEqual(sharding_spec,
                      torch_xla._XLAC._get_xla_sharding_spec(model.fc1.weight))
 
-  @unittest.skip("fails with functionalization")
-  def test_inplace_add_with_sharding(self):
-    xt = torch.ones(2, 2).to(xm.xla_device())
-    xs.mark_sharding(xt, self._get_mesh((1, self.n_devices)), (0, 1))
-    sharding_spec = torch_xla._XLAC._get_xla_sharding_spec(xt)
-    xt.add_(1)  # inplace update should preserve the sharding
-    self.assertEqual(sharding_spec, torch_xla._XLAC._get_xla_sharding_spec(xt))
```
JackCaoG, line 138: ditto, we should just leave test disabled? We don't want inplace operation to remove the sharding spec?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1501,9 +1501,12 @@ void InitXlaModuleBindings(py::module m) {
       // If the at::Tensor data is not present, we need to re-download the
       // tensor from the physical device to CPU. In that case, the value
       // must be present on the backend device.
-      XLA_CHECK(xtensor->GetXlaData() != nullptr &&
-                xtensor->CurrentDataHandle()->HasValue())
-          << "Cannot shard tensor. Data not present on any device.";
+      {
+        XLAGraphExecutor::Get()->DeviceBarrier(xtensor->GetDevice());
```
JackCaoG, line 1505: why do we need to add a barrier here?

yeounoh: Let's have a offline discussion, I would like to understand as well

torch_xla/csrc/init_python_bindings.cpp
yeounoh: I think we need to sync first why we are changing the API impl. This doesn't seem like enabling the tensor sharding attachment in the higher stack?

test/spmd/test_xla_sharding.py
alanwaketan: Got you. I'm fine of skipping the tests as it is. This is an attempt trying to provide a way to workaround the test.

torch_xla/csrc/init_python_bindings.cpp
alanwaketan: If we don't use the barrier, GetXlaData will error out if there is ongoing operation on the device.

test/spmd/test_xla_sharding.py
alanwaketan: We need to reach an agreement on the design and then how we can decouple the landing of the functionalization and development of the new SPMD design.

yeounoh:
"Currently SPMD relies on tensor.add_ to populate sharding spec for weights across different iterations."
>> No, the sharding annotation is attached to the XLATensor backing the torch.Tensor. It's copied to IR during tracing, but XLATensor holds the sharidng info ultimately. 
"However, Functionalization removes mutations and therefore this method is not working anymore. "
>> If I understood correctly (haven't tried/verified myself, but based on the info you shared), the problem we are seeing is that, the sharding annotation attached to the XLATensor holding the annotation is no longer associated with the torch.Tensor it seems, after the functionalization pass. The way functionalization removes the mutation is by creating intermediary tensnor to chain the compuation. 
"the SPMD team is trying to figure out a new place to store the sharding spec. Potentially, the DTensor or XLAShardedTensor such that the python object is the ultimate source of truth and can be kept intact regardless any dispatcher magic. We shouldn't need propagation in this case."
>> This would require the underlying PyTorch/XLA implementation to access the sharding info in Python layer, after the initial XLATensor construction. This is blocked on its feasibility check. Another alternative is to investigate a custom sharding propagation in the PyTorch/XLA layer, given that PyTorch/XLA layer receives all the IRs after the functionalization & at the same time, allow IR (with propagated sharding) to be the source of truth sometimes (that is, the new intermediary XLATensor created by functinalization may not have the sharding, so we copy to it).

cc @alanwaketan @JackCaoG @steventk-g

alanwaketan:
> Quoting from the PR desc.
> 
> > > "Currently SPMD relies on tensor.add_ to populate sharding spec for weights across different iterations."
> 
> No, the sharding annotation is attached to the XLATensor backing the torch.Tensor. It's copied to IR during tracing, but XLATensor holds the sharidng info ultimately.
> 

Correct, I can rephrase it.

> > > "However, Functionalization removes mutations and therefore this method is not working anymore. "
> 
> If I understood correctly (haven't tried/verified myself, but based on the info you shared), the problem we are seeing is that, the sharding annotation attached to the XLATensor holding the annotation is no longer associated with the torch.Tensor it seems, after the functionalization pass. The way functionalization removes the mutation is by creating intermediary tensnor to chain the compuation.

Correct.

> 
> > > "the SPMD team is trying to figure out a new place to store the sharding spec. Potentially, the DTensor or XLAShardedTensor such that the python object is the ultimate source of truth and can be kept intact regardless any dispatcher magic. We shouldn't need propagation in this case."
> 
> This would require the underlying PyTorch/XLA implementation to access the sharding info in Python layer, after the initial XLATensor construction. This is blocked on its feasibility check. Another alternative is to investigate a custom sharding propagation in the PyTorch/XLA layer, given that PyTorch/XLA layer receives all the IRs after the functionalization & at the same time, allow IR (with propagated sharding) to be the source of truth sometimes (that is, the new intermediary XLATensor created by functinalization may not have the sharding, so we copy to it).
> 
> cc @alanwaketan @JackCaoG @steventk-g @miladm

I guess it will be hard for us to know how to propagate the information in our layer properly because:
1) we want to propagate it if we know the created tensor is a temporarily tensor that is going to replace the input.
2) we don't want to propagate it for other cases to violate the compiler optimization.

That's why I think it's better to store it in the python layer which sits above the dispatcher and won't be affected by any dispatcher magic. This time it's functionalization, but who knows what program transfer comes next?

torch_xla/csrc/init_python_bindings.cpp
steventk-g: Following up: `ToTensor` makes sense to me. Thanks Jiewen!

alanwaketan: Okay, I will leave that part to you to fix in the master, and I will remove the DeviceBarrier here.

steventk-g (approved):
LGTM! Thanks!

yeounoh (approved):
LGTM

JackCaoG (approved):

alanwaketan:
Thanks Steven, Yeounoh, and Jack.

PR #4496, opened 2023-01-24
[Functionalization] Fix more cpp tests that require new lowerings

wonjoolee95:
Fix rest of the cpp tests that requires new lowerings. These include:
- SLogDet
- Inverse
- Pinverse

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -564,6 +564,41 @@ at::Tensor& XLANativeFunctions::_index_put_impl_(
                                                    accumulate);
 }
 
+std::tuple<at::Tensor, at::Tensor, at::Tensor, at::Tensor>
+XLANativeFunctions::_linalg_slogdet(const at::Tensor& self) {
+  TORCH_LAZY_FN_COUNTER("xla::");
+  XLATensorPtr self_tensor = bridge::GetXlaTensor(self);
+  auto outputs = tensor_methods::slogdet(self_tensor);
+  return std::make_tuple(bridge::AtenFromXlaTensor(std::get<0>(outputs)),
+                         bridge::AtenFromXlaTensor(std::get<1>(outputs)),
+                         bridge::AtenFromXlaTensor(XLATensorPtr()),
+                         bridge::AtenFromXlaTensor(XLATensorPtr()));
+}
+
+std::tuple<at::Tensor, at::Tensor, at::Tensor> XLANativeFunctions::_linalg_svd(
+    const at::Tensor& self, bool full_matrices, bool compute_uv,
+    c10::optional<c10::string_view> /* driver */) {
+  TORCH_LAZY_FN_COUNTER("xla::");
+  // As per https://pytorch.org/docs/stable/generated/torch.svd.html,
+  // The second boolean argument is exactly opposite between
+  // torch::svd and torch::_linalg_svd, hence the negation of full_matrices.
+  auto results = tensor_methods::svd(bridge::GetXlaTensor(self), !full_matrices,
+                                     compute_uv);
+  auto u = std::get<0>(results);
+  auto s = std::get<1>(results);
+  auto v = std::get<2>(results);
+  if (!compute_uv) {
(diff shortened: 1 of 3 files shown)


wonjoolee95:
The failures seem real, the `test_ops.py` for the updated ops are failing. Working on a fix.

wonjoolee95:
Okay, the `test_ops.py` still seem to give some trouble. For example the `AllowedOpInfoEntry('linalg.svd')` fails with error:

> ```2023-01-25 05:56:04.965700: F 2566733 /workspace/pytorch/xla/third_party/tensorflow/bazel-tensorflow/tensorflow/compiler/xla/xla_client/debug_macros.h:27] Non-OK-status: status.status() status: INVALID_ARGUMENT: Slice dim size 1 greater than dynamic slice dimension: 0.```

It looks the exact root cause of this failure is at https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ops/svd.cpp#L19. In the `ShapeOfXlaOp` call, this error is raised. The input that causes this error seems to be:
```
xla_tensor=1
full_matrices=0
compute_uv=1
```

wonjoolee95:
The newly failing `test/test_ops.py` ops include:
- `AllowedOpInfoEntry('geqrf')`
- `AllowedOpInfoEntry('linalg.det')`
- `AllowedOpInfoEntry('linalg.eig')`
- `AllowedOpInfoEntry('linalg.inv')`
- `AllowedOpInfoEntry('linalg.matrix_rank')`
- `AllowedOpInfoEntry('linalg.svd')`
- `AllowedOpInfoEntry('linalg.svdvals')`

I've commented out just to let the CI run for now.

Many of them (including `geqrf` op which seemingly doesn't appear related) fail at `SVD::Lower` like:
```
	xla::Shape const* ConsumeValue<xla::Shape const*>(tsl::StatusOr<xla::Shape const*>&&)
	
	torch_xla::SVD::Lower(torch_xla::LoweringContext*) const
```
This is the same error described in https://github.com/pytorch/xla/pull/4496#issuecomment-1403202882.

wonjoolee95:
The `_linalg_svd` *should* be fixed by https://github.com/pytorch/xla/pull/4537. I'll rebase once that PR merges.

wonjoolee95:
https://github.com/pytorch/xla/pull/4537 is merged with master, so some of the errors that we were seeing *should* be gone. While I test this locally, re-running CI to also verify.

wonjoolee95:
Okay, locally at least `test/test_ops.py` is all passing. I'll let CI verify the rest.

wonjoolee95:
CPU CI passing, will still wait for the GPU CI but this should be finally ready for review.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1544,6 +1555,20 @@ at::Tensor XLANativeFunctions::lift_fresh(const at::Tensor& tensor) {
   return at::functionalization::impl::to_functional_tensor(tensor);
 }
 
+std::tuple<at::Tensor, at::Tensor> XLANativeFunctions::linalg_inv_ex(
+    const at::Tensor& self, bool check_errors) {
+  TORCH_LAZY_FN_COUNTER("xla::");
```
alanwaketan, line 1560: It looks like we didn't do anything with `check_errors`. An easy fix for now is to fallback to cpu if check_errors is provided. wdyt?
https://pytorch.org/docs/stable/generated/torch.linalg.inv_ex.html

alanwaketan (commented):
Generally LGTM. Left one comment.

torch_xla/csrc/aten_xla_type.cpp
wonjoolee95: Good point, makes sense. According to https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/native_functions.yaml#L13160, the default value for `check_errors` is `false`, so I've updated to fallback to cpu when this is set to `true`. Also added a comment explaining this in the code.

alanwaketan (approved):
LGTM.

PR #4499, opened 2023-01-24
Try bootstrapping tpu env from env vars

steventk-g:
### Description
In [GKE](https://cloud.google.com/tpu/docs/kubernetes-engine-setup) and other restricted environments, we won't have access to the GCE Metadata returned from `_get_metadata()` in [`tpu.py`](https://github.com/pytorch/xla/blob/master/torch_xla/experimental/tpu.py), and we'll need another way to configure topologies for TPUs in such environments. This PR adds support for configuring topologies based on environment variables, proposed by the libTPU team to standardize configuration on their end.

### PJRT Experimental TPU Test Results
```
(pytorch) root@t1v-n-804806aa-w-0:/pytorch/xla# PJRT_DEVICE=TPU python test/pjrt/test_experimental_tpu.py
Running tests under Python 3.8.8: /root/anaconda3/envs/pytorch/bin/python
[ RUN      ] TestExperimentalTpu.test_configure_tpu_topology_v3-8_process_0
/root/anaconda3/envs/pytorch/lib/python3.8/unittest/case.py:1215: DeprecationWarning: assertDictContainsSubset is deprecated
  warnings.warn('assertDictContainsSubset is deprecated',
[       OK ] TestExperimentalTpu.test_configure_tpu_topology_v3-8_process_0
[ RUN      ] TestExperimentalTpu.test_configure_tpu_topology_v4-16_worker_1_process_0
[       OK ] TestExperimentalTpu.test_configure_tpu_topology_v4-16_worker_1_process_0
[ RUN      ] TestExperimentalTpu.test_configure_tpu_topology_v4-8_process_0
[       OK ] TestExperimentalTpu.test_configure_tpu_topology_v4-8_process_0
[ RUN      ] TestExperimentalTpu.test_configure_tpu_topology_v4-8_process_3
[       OK ] TestExperimentalTpu.test_configure_tpu_topology_v4-8_process_3
[ RUN      ] TestExperimentalTpu.test_get_worker_ips_four_hosts
[       OK ] TestExperimentalTpu.test_get_worker_ips_four_hosts
[ RUN      ] TestExperimentalTpu.test_get_worker_ips_one_host
[       OK ] TestExperimentalTpu.test_get_worker_ips_one_host
[ RUN      ] TestExperimentalTpu.test_num_local_processes_default_one_host
[       OK ] TestExperimentalTpu.test_num_local_processes_default_one_host
[ RUN      ] TestExperimentalTpu.test_num_local_processes_multi_process_one_host
[       OK ] TestExperimentalTpu.test_num_local_processes_multi_process_one_host
[ RUN      ] TestExperimentalTpu.test_num_local_processes_multi_process_v4-16
[       OK ] TestExperimentalTpu.test_num_local_processes_multi_process_v4-16
[ RUN      ] TestExperimentalTpu.test_num_local_processes_multi_process_v4-32
[       OK ] TestExperimentalTpu.test_num_local_processes_multi_process_v4-32
[ RUN      ] TestExperimentalTpu.test_num_local_processes_one_process_one_host
[       OK ] TestExperimentalTpu.test_num_local_processes_one_process_one_host
[ RUN      ] TestExperimentalTpu.test_process_bounds_size_default_one_host
[       OK ] TestExperimentalTpu.test_process_bounds_size_default_one_host
[ RUN      ] TestExperimentalTpu.test_process_bounds_size_multi_process_one_host
[       OK ] TestExperimentalTpu.test_process_bounds_size_multi_process_one_host
[ RUN      ] TestExperimentalTpu.test_process_bounds_size_multi_process_v4-16
[       OK ] TestExperimentalTpu.test_process_bounds_size_multi_process_v4-16
[ RUN      ] TestExperimentalTpu.test_process_bounds_size_multi_process_v4-32
[       OK ] TestExperimentalTpu.test_process_bounds_size_multi_process_v4-32
[ RUN      ] TestExperimentalTpu.test_process_bounds_size_one_process_one_host
[       OK ] TestExperimentalTpu.test_process_bounds_size_one_process_one_host
[ RUN      ] TestExperimentalTpu.test_task_id0 (None, None)
[       OK ] TestExperimentalTpu.test_task_id0 (None, None)
[ RUN      ] TestExperimentalTpu.test_task_id1 ('0', 0)
[       OK ] TestExperimentalTpu.test_task_id1 ('0', 0)
[ RUN      ] TestExperimentalTpu.test_task_id2 ('1', 1)
[       OK ] TestExperimentalTpu.test_task_id2 ('1', 1)
[ RUN      ] TestExperimentalTpu.test_task_id3 ('15', 15)
[       OK ] TestExperimentalTpu.test_task_id3 ('15', 15)
[ RUN      ] TestExperimentalTpu.test_tpu_env_from_env_vars_all-vars-set
[       OK ] TestExperimentalTpu.test_tpu_env_from_env_vars_all-vars-set
[ RUN      ] TestExperimentalTpu.test_tpu_env_from_env_vars_defaults-only
[       OK ] TestExperimentalTpu.test_tpu_env_from_env_vars_defaults-only
[ RUN      ] TestExperimentalTpu.test_tpu_env_from_gce_metadata
[       OK ] TestExperimentalTpu.test_tpu_env_from_gce_metadata
Ran 23 tests in 0.012s

OK
```


JackCaoG:
@steventk-g can you add the description that what this pr is for?

torch_xla/experimental/tpu.py
```diff
@@ -94,14 +109,15 @@ def version() -> int:
...
-  match = re.match(r'^v(\d)-(\d+)$', env['ACCELERATOR_TYPE'])
+  match = re.match(r'^v(\d)-(\d+)$', env[xenv.ACCELERATOR_TYPE])
   return int(match.groups()[0])
 
 
 def get_worker_ips() -> List[str]:
   """Returns ordered list of TPU worker IPs from TPU metadata."""
-  metadata = _get_metadata('worker-network-endpoints')
-
+  metadata = xu.getenv_as(xenv.TPU_WORKER_HOSTNAMES, str, '')
```
thebinaryone1, line 118: TPU_WORKER_HOSTNAMES is of the format: host-name-1,host-name-2,..........,host-name-n

Change the code to the following perhaps?
  if metadata is '':
    metadata = _get_metadata('worker-network-endpoints')
    # Workers have format 'hostname:uid:ip,hostname:uid:ip,...'
    workers = metadata.split(',')
    ips = [worker.split(':')[2] for worker in workers]
  else:
    ips = metadata.split(',')
  return ips if len(ips) > 1 else ['localhost']

steventk-g: Done

test/pjrt/test_experimental_tpu.py
```diff
@@ -46,14 +46,17 @@ def test_task_id(self, task_id, expected):
 
     self.assertEqual(i, expected)
 
-  def test_tpu_env(self):
+  def test_tpu_env_from_gce_metadata(self):
+    if xenv.TPU_ACCELERATOR_TYPE in os.environ:
```
will-cromar, line 50: Since we only need to manipulate the env vars in python and don't need to pass them down into C++, its better to `mock.patch` `os.environ` to make cleanup easier and prevent changes to `os.environ` from leaking between test cases. Example:

https://github.com/pytorch/xla/blob/05195c7ee63b75b395f794de659c7016f8c55335/test/pjrt/test_experimental_tpu.py#L43-L44

torch_xla/experimental/tpu.py
```diff
@@ -135,8 +152,8 @@ def configure_topology(local_rank: int,
   """
   tpu_env = get_tpu_env()
 
-  accelerator_type = tpu_env['ACCELERATOR_TYPE']
-  if tpu_env['ACCELERATOR_TYPE'].startswith('v4'):
+  accelerator_type = tpu_env[xenv.ACCELERATOR_TYPE]
+  if tpu_env[xenv.ACCELERATOR_TYPE].startswith('v4'):
```
will-cromar, line 156: Let's change this to `tpu.version() == 4`

torch_xla/experimental/tpu.py
```diff
@@ -81,11 +81,26 @@ def task_id() -> Optional[int]:
   return xu.getenv_as(xenv.CLOUD_TPU_TASK_ID, int)
 
 
+def build_tpu_env_from_vars() -> Dict[str, str]:
+  metadata = defaultdict(str)
```
will-cromar, line 85: Why use `defaultdict`?

torch_xla/experimental/tpu.py
```diff
@@ -81,11 +81,26 @@ def task_id() -> Optional[int]:
...
+  return metadata
+
+
 def get_tpu_env() -> Dict[str, str]:
   """Fetches and parses `tpu-env` metadata field."""
-  metadata = _get_metadata('tpu-env')
-
-  return yaml.load(metadata, yaml.Loader)
+  metadata = build_tpu_env_from_vars()
+  if metadata[xenv.ACCELERATOR_TYPE] is None:
```
will-cromar, line 100: I think the logic here could be easier to follow, since this calls the env vars "metadata" and then conditionally ignores it and returns the actual metadata. Also, the first definition of `metadata` is a `dict` and the second is a `str`.

Let's work out the condition is for using the environment variables compared to metadata and put that up front, rather than having the logic broken up. e.g. something of the form

```
if using_env_vars():
  return build_tpu_env_from_vars()

# otherwise fetch GCE metadata
```

Also see my main comment for a design question.

torch_xla/experimental/tpu.py
```diff
@@ -94,18 +109,20 @@ def version() -> int:
...
 
 def get_worker_ips() -> List[str]:
   """Returns ordered list of TPU worker IPs from TPU metadata."""
-  metadata = _get_metadata('worker-network-endpoints')
-
-  # Workers have format 'hostname:uid:ip,hostname:uid:ip,...'
-  workers = metadata.split(',')
-  ips = [worker.split(':')[2] for worker in workers]
-
+  metadata = xu.getenv_as(xenv.TPU_WORKER_HOSTNAMES, str, '')
```
will-cromar, line 118: nit: call this something more specific like `hostnames_env` (feel free to come up with a better name). I don't want readers to confuse this with GCE metadata, and we want to make it clearer to the reader this has a different format than the other usage of `metadata` down below.

torch_xla/experimental/tpu.py
```diff
@@ -94,18 +109,20 @@ def version() -> int:
...
+  if metadata is '':
```
will-cromar, line 119: nit: `if not metadata:`

torch_xla/experimental/tpu.py
```diff
@@ -94,18 +109,20 @@ def version() -> int:
...
+    metadata = _get_metadata('worker-network-endpoints')
+    # Workers have format 'hostname:uid:ip,hostname:uid:ip,...'
+    workers = metadata.split(',')
+    ips = [worker.split(':')[2] for worker in workers]
+  else:
+    ips = metadata.split(',')
```
will-cromar, line 125: super-nit: this is technically not a list of IPs anymore, since we may pass hostnames in as `TPU_WORKER_HOSTNAMES`. I suggest `hostnames` to be more general

will-cromar (commented):
Thanks! This is very useful!

I have a design question: how do we want to handle a partially available set of env vars? I see three options:

1. Use env vars if one specific env var is set (current implementation)
2. Use env vars if _any_ is set
3. Use env vars only is _all_ are set
4. If there's an incomplete set, fetch the GCE metadata and override with the available env vars.

I think there is at least one case where we can set `TPU_ACCELERATOR_TYPE`, but want to use the metadata for everything else. But we could still keep (1) if we look for e.g. `TPU_HOST_BOUNDS`. 

I'm leaning in favor of (4), since it would be easier to add new settings in the future as long as they're available in the GCE metadata. We should print a warning in that case if there are incomplete env vars, though, since not every environment will actually have metadata available.

torch_xla/experimental/tpu.py
steventk-g: The idea was to prevent crashes when consumers of `tpu_env` query using a key that would be in GCE metadata but is missing from the environment variables

vanbasten23:
re comment#1, do we run the tpu ci in the GKE? Also, can you add a link to the `tpu.py` you mentioned?

test/pjrt/test_experimental_tpu.py
steventk-g: Done, thanks for the tip!

torch_xla/experimental/tpu.py
steventk-g: Done

torch_xla/experimental/tpu.py
steventk-g: Replaced with `dict()`

torch_xla/experimental/tpu.py
steventk-g: Done

torch_xla/experimental/tpu.py
steventk-g: Done

torch_xla/experimental/tpu.py
steventk-g: Done

torch_xla/experimental/tpu.py
steventk-g: Done

steventk-g:
> re comment#1, do we run the tpu ci in the GKE? Also, can you add a link to the `tpu.py` you mentioned?

We run the TPU CI in a hacky version of GKE (using self registered TPUs instead of letting GKE manage resources) which will hopefully be replaced with full scale TPUVM<>GKE integration sometime this year. As far as fully testing this change on GKE+PT/XLA, I think there's still more work to be done before that's fully available. This PR is one step in that direction.

I've added some useful hyperlinks in the description for GKE and the tpu.py file.

will-cromar (approved):
Thanks! I like using `TPU_SKIP_MDS_QUERY` as the way to decide whether or not to use env var initialization.

torch_xla/experimental/tpu.py
```diff
@@ -81,10 +80,28 @@ def task_id() -> Optional[int]:
   return xu.getenv_as(xenv.CLOUD_TPU_TASK_ID, int)
 
 
+def using_env_vars() -> bool:
```
will-cromar, line 83: nit: prefix this function with an underscore because it shouldn't be used outside of this module

steventk-g: Done

torch_xla/experimental/tpu.py
```diff
@@ -81,10 +80,28 @@ def task_id() -> Optional[int]:
...
+  return xu.getenv_as(xenv.TPU_SKIP_MDS_QUERY, str, False)
+
+
+def build_tpu_env_from_vars() -> Dict[str, str]:
```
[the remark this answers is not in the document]
steventk-g: Done

PR #4503, opened 2023-01-24
[PJRT] Don't use futures on `ExecuteSharded` output buffers

will-cromar:
Setting callbacks on each output buffer (see #4400) led to a significant performance regression on both TPU v3 and TPU v4. Remove those callbacks entirely.

Instead, only use the `returned_future` of `ExecuteSharded` when not using PJRT C API. This means the `ExecuteTime` metric will be broken with the PJRT C API until our next pin update, when we can remove this condition.

Tested with both `PJRT_DEVICE=TPU` and `PJRT_DEVICE=TPU_C_API`. Both get performance similar to before #4400.


JackCaoG (approved):
Thanks!

PR #4506, opened 2023-01-25
[Functionalization] Mark test_logcumsumexp NTBF

alanwaketan:
Summary:
test_logcumsumexp fails because the shape check on logcumsumexp.out is not executed after functionalized decomposition. As disccused in pytorch/pytorch#92912, this will be hard to fix. Since we never lower logcumsumexp, our users are highly likely not going to use it.

Test Plan:
CI.


JackCaoG (approved):

alanwaketan:
Thanks Jack for approving.

PR #4508, opened 2023-01-25
Transfer shards from server using replication

yeounoh:
Allow shards transfer to host, so we can do
```
xt1 = torch.ones(16, 16).to(xm.xla_device())
xs.mark_sharding(xt1, self._get_mesh((1, self.n_devices)), (0, 1))
t1 = xt1.cpu()
```
Currently, `xt1.cpu()` would crash. This is not exposing the shards, which will be addressed in the future.

### DIFF
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -5,6 +5,7 @@
 #include "absl/strings/ascii.h"
 #include "absl/types/span.h"
 #include "pjrt_computation_client.h"
+#include "tensorflow/compiler/xla/client/xla_builder.h"
 #include "tensorflow/compiler/xla/client/xla_computation.h"
 #include "tensorflow/compiler/xla/layout_util.h"
 #include "tensorflow/compiler/xla/literal.h"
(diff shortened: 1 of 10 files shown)


yeounoh:
The failure is due to a newly disabled test, 
```
[ FAILED  ] 1 test, listed below:
[  FAILED  ] AtenXlaTensorTest.TestNonzero
```
rebasing now.

JackCaoG:
@yeounoh can you add a comment in the description regarding what problem are you trying to solve?

JackCaoG:
Thanks, I will take a look today

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -193,6 +195,45 @@ ComputationClient::DataPtr PjRtComputationClient::CopyToDevice(
...
+ComputationClient::DataPtr PjRtComputationClient::ReplicateShardedData(
+    const ComputationClient::DataPtr& handle) {
+  PjRtShardedData& sharded_data = dynamic_cast<PjRtShardedData&>(*handle);
+  xla::XlaBuilder b("ReplicateShardedData");
+  xla::Shape shape = sharded_data.shape();
+  b.SetSharding(sharded_data.GetSharding());
+  auto x = xla::Parameter(&b, 0, shape, "p0");
+  b.SetSharding(xla::HloSharding::Replicate().ToProto());
+  auto y = xla::Div(x, ConstantR0<float>(&b, 2));
+  auto z = xla::Add(y, y);
```
JackCaoG, line 207: can we add some comments here?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -193,6 +195,45 @@ ComputationClient::DataPtr PjRtComputationClient::CopyToDevice(
                                     std::move(status_or.value()));
 }
 
+ComputationClient::DataPtr PjRtComputationClient::ReplicateShardedData(
```
JackCaoG, line 198: If I understand correctly,  this functions takes a sharded data and then replicate it on all device and then return? This is useful when you want to all_gather a sharded data and return to the user?

JackCaoG (commented):
Mostly lgtm, minor comments

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Ack, will do.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Correct, basically we perform a simple/light identity calculation, so that the compiler replicates the output (reassembled input).

JackCaoG (approved):
Thanks!

PR #4518, opened 2023-01-26
[Functionalization] Re-enable some of the dynamic shape test

alanwaketan:
Summary:
The dynamic shape tests are broken because of upstream changes that decompose dynamic ops into different ways. Adds new lowering to fix that.

Corresponding upstream changes for SymBool: https://github.com/pytorch/pytorch/pull/92149.

Test Plan:
CI

### DIFF
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -607,6 +607,11 @@ bool XLATensor::ShouldSyncIrNode() {
   return this->data()->ir_value->op() != xla_device_data;
 }
 
+bool XLASymNodeImpl::is_bool() {
+  // TODO: handle not is int
+  return true;
+}
+
 bool XLASymNodeImpl::is_int() {
   // TODO: handle not is int
   return true;
@@ -663,8 +668,10 @@ c10::SymNode XLASymNodeImpl::eq(const c10::SymNode& other) {
 }
 
 c10::SymNode XLASymNodeImpl::ne(const c10::SymNode& other) {
-  XLA_CHECK(false) << "XLASymNodeImpl::" << __FUNCTION__
-                   << " has not been implemented.";
+  TORCH_LAZY_FN_COUNTER("xla::size_");
+  auto p_other = dynamic_cast<XLASymNodeImpl*>(other.get());
+  auto n_ne = torch::lazy::MakeNode<SizeNe>(node(), p_other->node());
+  return c10::make_intrusive<XLASymNodeImpl>(n_ne);
 }
 
 c10::SymNode XLASymNodeImpl::gt(const c10::SymNode& other) {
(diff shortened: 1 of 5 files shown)


torch_xla/csrc/tensor.cpp
```diff
@@ -607,6 +607,11 @@ bool XLATensor::ShouldSyncIrNode() {
   return this->data()->ir_value->op() != xla_device_data;
 }
 
+bool XLASymNodeImpl::is_bool() {
```
alanwaketan, line 610: I actually don't know what's the right thing to do so. Just try a few different values and this one works.

torch_xla/csrc/tensor.cpp
```diff
@@ -746,6 +753,10 @@ double XLASymNodeImpl::guard_float(const char* file, int64_t line) {
                    << " has not been implemented.";
 }
 
+bool XLASymNodeImpl::guard_bool(const char* file, int64_t line) {
```
alanwaketan, line 756: Not sure what are we supposed to deal with the parameters and the word "guard".

alanwaketan (commented):
Upstream has some new changes regarding to the bool things. I just poke around and make things work. @JackCaoG @vanbasten23 @miladm Please advise what's the right approach to deal with it.

torch_xla/csrc/tensor.cpp
JackCaoG: @ezyang Can you shed some light on the expected behavior on this function?(along with `is_int` and `guard_bool`)

ezyang: So the intended semantics is, when we do an equality operation, the resulting SymNode should be treated as a bool. You'll have to add some sort of internal tagging to XLA to do this, I think.

alanwaketan: Thanks, Ed. Implemented corresponding changes.

wonjoolee95 (approved):

torch_xla/csrc/tensor.cpp
```diff
@@ -607,6 +607,17 @@ bool XLATensor::ShouldSyncIrNode() {
   return this->data()->ir_value->op() != xla_device_data;
 }
 
+bool XLASymNodeImpl::is_bool() {
+  auto op = node()->op().op;
+  // Reference:
+  // https://github.com/pytorch/pytorch/blob/master/torch/fx/experimental/symbolic_shapes.py#L403
+  if (op == at::aten::eq || op == at::aten::ne || op == at::aten::ge ||
+      op == at::aten::lt) {
```
ezyang, line 615: if you support boolean negation, that'll be good too. This is a weird way to do it but absent proper type propagation it's probably ok.

JackCaoG (approved):

torch_xla/csrc/tensor.cpp
alanwaketan: Yea, otherwise I will need to type check the object which is probably weird too.

alanwaketan:
Thanks Wonjoo, Jack and Ed for reviewing the patch.

PR #4504, opened 2023-01-25
[PJRT] Release the GIL during `TransferFromServer`

will-cromar:
Because execution in PJRT is asynchronous, if we call `TransferFromServer` on a buffer produced by a computation that is in progress, `TransferToServer` will block until the underlying buffer is ready. However, in this case, the thread calling `TransferFromServer` is also holding the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock). That means that in a multithreaded context, we can end up with a deadlock where one thread is holding the GIL indefinitely while it waits for another replica to participate in a computation, and the thread that must run that computation is blocked because the first thread is holding the GIL.

This started happening consistently with PJRT on TPU v3 some time in mid-December, triggered by an unrelated commit. I confirmed the GIL deadlock by inspecting hanging threads with `gdb`:

<details>
<summary>`gdb` output</summary>

```
Thread 456 (Thread 0x7f3aa37fe700 (LWP 856191)):
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00007f405a13278e in absl::lts_20220623::synchronization_internal::Waiter::Wait(absl::lts_20220623::synchronization_internal::KernelTimeout) () from /pytorch/xla/torch_xla/lib/libxla_computation_client.so
#2  0x00007f405a132652 in AbslInternalPerThreadSemWait_lts_20220623 () from /pytorch/xla/torch_xla/lib/libxla_computation_client.so
#3  0x00007f405a133a43 in absl::lts_20220623::Mutex::Block(absl::lts_20220623::base_internal::PerThreadSynch*) () from /pytorch/xla/torch_xla/lib/libxla_computation_client.so
#4  0x00007f404d4d7c9c in absl::lts_20220623::Mutex::LockSlowWithDeadline(absl::lts_20220623::MuHowS const*, absl::lts_20220623::Condition const*, absl::lts_20220623::synchronization_internal::KernelTimeout, int) [clone .cold.39] () from /pytorch/xla/torch_xla/lib/libxla_computation_client.so
#5  0x00007f404d4d7cb0 in absl::lts_20220623::Mutex::LockSlow(absl::lts_20220623::MuHowS const*, absl::lts_20220623::Condition const*, int) () from /pytorch/xla/torch_xla/lib/libxla_computation_client.so
#6  0x00007f405a1351d3 in absl::lts_20220623::Notification::WaitForNotification() const () from /pytorch/xla/torch_xla/lib/libxla_computation_client.so
[...]

PR #4520, opened 2023-01-26
[PJRT] Experimental support for `torch.distributed` and DDP on TPU v2/v3

will-cromar:
- Use experimental [`ThreadLocalWorld`](https://github.com/pytorch/pytorch/blob/801d831d7a83c3da41b5d457616cef08769d7281/torch/testing/_internal/distributed/multi_threaded_pg.py#L301-L335) to enable multithreading
- Implement a new `torch.distributed` `init_method` that uses PJRT runtime parameters and supports multithreading
- `torch.distributed` "rank" will become the same as our "ordinal", meaning we have one fewer set of indices to track
- Converge the PJRT ImageNet and MNIST tests with the original ones
- Remove experimental `pjrt.DistributedDataParallel` now that the upstream version works on v3

Performance comparison using ResNet50 with fake data on TPU v3:

Runtime | DDP | Throughput (ex/sec/replica)
-- | -- | --
XRT | No | 418.54
XRT | Yes | 395.97
PJRT | No | ~640
PJRT | Yes | ~565

Example usage:

```python
import torch
import torch_xla.experimental.pjrt_backend

def _all_gather(index: int):
  dist.init_process_group('xla', init_method='pjrt://')
  t = torch.tensor([index], dtype=torch.int32, device=xm.xla_device())
  output = [torch.zeros_like(t) for _ in range(dist.get_world_size())]
  dist.all_gather(output, t)
  xm.mark_step()
  print(output)

if __name__ == '__main__':
  xmp.spawn(self._all_gather)
```

Needs rebasing after #4504 merges

Follow-up:
- Update automated tests
- Update PJRT and DDP documentation before release


JackCaoG:
@will-cromar is this one ready for review?

will-cromar:
I'll take another pass tomorrow to polish and add some comments, but this should be ready for review.

test/pjrt/test_ddp.py
```diff
@@ -18,39 +17,29 @@
...
+    pjrt.spawn_threads(self._ddp_init)
 
-  @parameterized.named_parameters(*ddp_parameters)
-  def test_ddp_correctness_large_net(self, ddp: type):
+  @parameterized.named_parameters(('small_net', False), ('large_net', True))
+  def test_ddp_correctness(self, use_large_net: bool):
     pjrt._run_multiprocess(
-        util.ddp_correctness, ddp=ddp, use_large_net=True, debug=FLAGS.debug)
+        util.ddp_correctness,
+        init_method='pjrt://',
```
alanwaketan, line 40: Wonder if we want to parameterized the init_method with env as well?

test/distributed_util.py
```diff
@@ -128,10 +129,11 @@ def ddp_correctness(ddp: type = torch.nn.parallel.DistributedDataParallel,
   offset = rank * local_batch_size
   for step in range(steps):
     # To make torch.randn produce same results across devices.
-    torch.manual_seed(2022 + step)
+    rng = torch.Generator().manual_seed(2022 + step)
```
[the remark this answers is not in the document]
will-cromar: The other option would be to wrap these `randn` calls in a lock and give them a common global seed, but explicitly creating a new generator with the same seed seems clearer to me. I would have done the same for module initialization, but that case doesn't support a custom RNG.

test/pjrt/test_ddp.py
will-cromar: Good idea. Added another test that skips for TPU <= v3, since `env://` doesn't work nicely with multithreading.

alanwaketan (approved):
LGTM.

test/distributed_util.py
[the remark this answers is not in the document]
alanwaketan: One last comment. Maybe add a comment to the test case to suggest the reasoning behind?

test/test_train_mp_imagenet.py
```diff
@@ -31,7 +31,7 @@
     '--ddp': {
         'action': 'store_true',
     },
-    '--ddp_pjrt': {
+    '--pjrt_distributed': {
```
JackCaoG, line 34: can we add comment here to explain how is `pjrt_distributed` different than `ddp`?

JackCaoG (approved):
Thanks!

Issue #4475, opened 2023-01-19
TPU dynamo speed up on Training analysis


JackCaoG:
# Context:
I am running inference benchmark using dynamo bridge + torch_bench on TPU v4 single device. This thread is more to update the current info and some todos. 

title |  value
-- | --
PyTorch Version | Nightly at 01/2023
PyTorch/XLA Version | Nightly at 01/2023
TorchBench + TorchText + TorchAudio Version | Nightly at 01/2023
Accelerator | v4-8
Runtime | PJRT
Average Speedup | 0.8315

</div></b>

# Result
![image](https://user-images.githubusercontent.com/59073027/213330339-d958b18b-6caf-437f-8c6a-166b0a0ff32c.png)

# Sample Profile
gs://tpu-pytorch/tmp/dynamo_profile/dynamo_tracing_train/try1/
![image](https://user-images.githubusercontent.com/59073027/213330512-823e0bdb-3c6d-4e78-801d-7abe1f326476.png)

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-16eb41f2-7fff-2e72-9510-db9c75e0b903"><div dir="ltr" style="margin-left:0pt;" align="center">

  | Device Executation Time(ms)
-- | --
Lazy | 17.85
Dynamo Input Sync | 0.083
Dynamo Forward | 9.6
Dynamo Cleanup 1 | 0.6
Dynamo Backward | 13.1
Dynamo Cleanup 2 | 0.289
Dynamo Combined | 23.672

  | Tracing TIme(ms) | Total Device Execution Time(ms) | Wall time(ms)
-- | -- | -- | --
Lazy | 13.67 | 17.85 | 39
Dynamo | N/A | 23.672 (0.75x) | 52.6 (0.74x)

JackCaoG:
The Lazy Tensor causes the first Device execution. The lazy tracing takes 13.67ms and the device execution takes 17.85 ms, LTC’s walltime is ~39ms. This is expected since each `step_marker` and `ExecuteGraph` has some overhead.

If we look at the Dynamo graph, there are 5 graphs being executed. Only the forward and backward takes more than 1 ms, but there is some very noticeable gap(9.91 ms between forward and cleanup1 and 5.137 between cleanup 1 and backward) between each device execution. The key to improve the dynamo training speed would be to understand and minimize the gap between each device executions.

Going forward I think there are three areas we should investigate
[...]

PR #4523, opened 2023-01-27
Optimize Dynamo Training Execute

JackCaoG:
This will help https://github.com/pytorch/xla/issues/4475

Currently each dynamo training step(without optimizer) will result in 5 graph execution

They are
1. `sync_multi` to sync input to fwd graph
2. Execute the fwd graph
3. `sync_multi` to sync input to bwd graph
4. Execute bwd graph
5. `mark_step` at the end of `optimized_mod`

My finding is that
1. The first graph executation is uecessary. It is triggered by `GetXlaData` in 
https://github.com/pytorch/xla/blob/661ab6d27bdf94483c141a6c218e8be9712b5528/torch_xla/csrc/init_python_bindings.cpp#L1700-L1702
However if I check the HLO, it is actually a `DeviceData` IR(not sure why). We don't need to call `applyPendingGraph` to trigger execeution to retrive the XLAData. Hence my fix in `GetXlaData`

2. The graph 3 and graph 5 are both triggered by `aot_autograd`. Instead of use `xla_sync_multi` to only sync the partial pending tensors and use `mark_step` to sync the rest, we can just do one `mark_step` in the beginning to combine 2 graphs. One thing to note is that we need to make first `mark_step` to be blocking.

With above 2 fixes, we can cut graph from 5 -> 3. 

With the latest change, I am seeing a much more promising result
<google-sheets-html-origin><style type="text/css"></style>

model | Old Speed Up | new Speed up
-- | -- | --
resnet50 | 0.758 | 0.937
resnet18 | 0.66 | 1.003
BERT_pytorch | 1.441 | 1.869
resnext50_32x4d | 0.87 | 1.139
alexnet | 0.632 | 0.802
mobilenet_v2 | 0.549 | 0.672
mnasnet1_0 | 0.698 | 0.967
vgg16 | 0.712 | 0.742
timm_vision_transformer | 1.275 | 1.69
squeezenet1_1 | 0.72 | 0.958
Avg | 0.8315 | 1.0779

### DIFF
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -280,7 +280,10 @@ def extract_compiled_graph(xla_model: torch.fx.GraphModule, xla_args):
   torch_xla._XLAC._clear_pending_irs(str(xm.xla_device()))
 
   def optimized_mod(*args):
-    torch_xla._XLAC._xla_sync_multi(args, [])
+    # mark_step needs to be blocking since we want to access args's XLADatas
+    # and they can't be placeholder.
+    if any(torch_xla._XLAC._check_tensor_need_materialization(args)):
+      xm.mark_step(wait=True)
     enter_ts = time.time()
     if len(args_and_out) == 0:
       return ()
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -315,7 +315,10 @@ def extract_compiled_graph(xla_model: torch.fx.GraphModule, xla_args):
     if debug:
       print(f"optimized_mod takes {time.time() - enter_ts} seconds overall")
 
-    xm.mark_step()
+    if any(
+        torch_xla._XLAC._check_device_tensor_need_materialization(
+            str(xm.xla_device()))):
+      xm.mark_step()
     none_remover.add_nones(result)
     return result
 
--- a/test/dynamo/test_num_output.py
+++ b/test/dynamo/test_num_output.py
@@ -104,3 +104,8 @@ class TestNumOutput(unittest.TestCase):
   def test_direct_return_with_duplicated_inplace_update(self):
     self.do_test(
         DirectReturnWithDuplicatedInplaceUpdateModule, expected_num_output=3)
+
+
+if __name__ == '__main__':
+  test = unittest.main()
+  sys.exit(0 if test.result.wasSuccessful() else 1)
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -941,7 +941,7 @@ def _run_step_closures():
   return devctx
 
 
-def mark_step():
+def mark_step(wait=False):
   if xu.getenv_as('XLA_EMIT_STEPLOG', bool, False):
     print(
         'torch_xla.core.xla_model::mark_step\n',
@@ -950,7 +950,7 @@ def mark_step():
         flush=True)
   torch_xla._XLAC._xla_step_marker(
       torch_xla._XLAC._xla_get_default_device(), [],
-      wait=xu.getenv_as('XLA_SYNC_WAIT', bool, False))
+      wait=xu.getenv_as('XLA_SYNC_WAIT', bool, wait))
   # Only emit metrics from the first local device index, to avoid emitting the
   # same values from different threads.
   if is_master_ordinal():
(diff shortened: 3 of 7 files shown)


JackCaoG:
I run the benchmark again using this pr and sadly find out the improvement is not as big as I would hope
image

The mark_step between forward and backward are very expensive, despite the actual device execution is very tiny. I will look into 
1. If there is a way to make this mark_step faster
2. If there is a way to make this mark_step async

For the second part I actually had some idea. I might merge this pr as it is for now.

JackCaoG:
<google-sheets-html-origin><style type="text/css"></style>

model | Old Speed Up | New Speed Up |  
-- | -- | -- | --
resnet50 | 0.758 | 0.731 |  
resnet18 | 0.66 | 0.693 |  
BERT_pytorch | 1.441 | 1.459 |  
resnext50_32x4d | 0.87 | 0.845 |  
alexnet | 0.632 | 0.747 |  
mobilenet_v2 | 0.549 | 0.546 |  
mnasnet1_0 | 0.698 | 0.674 |  
vgg16 | 0.712 | 0.731 |  
timm_vision_transformer | 1.275 | 1.335 |  
squeezenet1_1 | 0.72 | 0.767 |  
Avg | 0.8315 | 0.8528 | 2.56% better

torch_xla/core/dynamo_bridge.py
```diff
@@ -280,7 +280,10 @@ def extract_compiled_graph(xla_model: torch.fx.GraphModule, xla_args):
   torch_xla._XLAC._clear_pending_irs(str(xm.xla_device()))
 
   def optimized_mod(*args):
-    torch_xla._XLAC._xla_sync_multi(args, [])
+    # mark_step needs to be blocking since we want to access args's XLADatas
+    # and they can't be placeholder.
+    if any(torch_xla._XLAC._check_tensor_need_materialization(args)):
+      xm.mark_step(wait=True)
```
shunting314, line 286: In general, does a synchronous `mark_step` can be slower than `_xla_sync_multi`? Since the latter does not wait for the execution.

JackCaoG: `_xla_sync_multi` is also synchronous, but it might be executing a smaller graph.

JackCaoG:
Oh interesting, if I remove the `mark_step` at the end of the `optimized_mod`, resnet50 speed up jump from 
`0.731` -> `0.928`

I think what happened is that even `mark_step` at the end might be a no-op, it will still try to acquire the device lock. However device lock is hold by the async thread of `torch_xla._XLAC._run_cached_graph` which blocks the `mark_step`.

I think we should not unconditionally perform the `mark_step` at the end

torch_xla/core/dynamo_bridge.py
shunting314: Ah, I didn't realize that the default behavior of 'mark_step (not wait for compute)' and '_xla_sync_multi (wait for compute)' are different. Thanks for clarification!

JackCaoG:
> Oh interesting, if I remove the `mark_step` at the end of the `optimized_mod`, resnet50 speed up jump from `0.731` -> `0.928`
> 
> I think what happened is that even `mark_step` at the end might be a no-op, it will still try to acquire the device lock. However device lock is hold by the async thread of `torch_xla._XLAC._run_cached_graph` which blocks the `mark_step`.
> 
> I think we should not unconditionally perform the `mark_step` at the end

This also explains why my previous pr to make execution async for dynam doesn't help, it is because `mark_step` will block on that async execution anyway.

alanwaketan (approved):
LGTM.

JackCaoG:
With the latest change, I am seeing a much more promising result
<google-sheets-html-origin><style type="text/css"></style>

model | Old Speed Up | new Speed up
-- | -- | --
resnet50 | 0.758 | 0.937
resnet18 | 0.66 | 1.003
BERT_pytorch | 1.441 | 1.869
resnext50_32x4d | 0.87 | 1.139
alexnet | 0.632 | 0.802
mobilenet_v2 | 0.549 | 0.672
mnasnet1_0 | 0.698 | 0.967
vgg16 | 0.712 | 0.742
timm_vision_transformer | 1.275 | 1.69
squeezenet1_1 | 0.72 | 0.958
Avg | 0.8315 | 1.0779

I think there still are some unit test failing, need to investigate a bit

JackCaoG:
Ok I think I fixed all of the failing tests, let's wait for CI to confirm

PR #4527, opened 2023-01-27
Move XLA testing to Python-3.8

malfet:
As 3.7 is approaching EOL and getting deprecated in pytorch/pytorch

Also, remove misleading `BASE_IMAGE_TAG` which is defined, but not used anywhere according to
https://cs.github.com/?scopeName=All+repos&scope=&q=BASE_IMAGE_TAG+repo%3Apytorch%2Fxla


JackCaoG (approved):
Thanks!

PR #4528, opened 2023-01-27
Bazel

stgpetrovic:
Bazel integration for PyTorch/XLA's `libxla_computation_client.so`.

This is a large change but should have no effect on the build.

Work done:

- bazel WORKSPACE created
- tensorflow moved to a external dependency, patched and pinned via WORKSPACE
- bazelrc ported to ensure equivalent build
- BUILD file reworked to include files form external tesorflow dependency
- includes reworked to match new file locations
- includes in torch_xla/crsc rewored too
- setup modified to find files in new locations and build new targets
- integration for bazel and clangd for development support
- integration for bazel for coverage for calculating and rendering coverage
- sample tests written to demonstrate coverage
- fixes to code to remove warnings (and warning suppressions)

This enables for:

- simpler programming since the code is not copied in tensorflow tree
- visual studio code references
- visual studio autocompletion
- visual studio refactoring and smart navigation
- faster rebuild times
- deterministic builds with simple `bazel build`, no shell or the like

Future work:

- migrate `torch_xla` to bazel too to get all perks named above for the whole codebase
- support for remote build execution so that build times can be much faster
- untangling certain parts of code, splitting into smaller compilation units, for incremental build speed
- and more


WORKSPACE
```diff
@@ -0,0 +1,54 @@
...
+# To update TensorFlow to a new revision,
+# a) update URL and strip_prefix to the new git commit hash
+# b) get the sha256 hash of the commit by running:
+#    curl -L https://github.com/tensorflow/tensorflow/archive/<git hash>.tar.gz | sha256sum
+#    and update the sha256 with the result.
+http_archive(
+    name = "org_tensorflow",
+    patch_args = ["-p1"],
+    patch_tool = "patch",
+    patches = [
```
will-cromar, line 12: We don't need to apply patches manually anymore, right? Can you remove references to `apply_patches.sh`? That should also fix the CI script:

https://github.com/pytorch/xla/blob/ce48e42abf486afc82940f24db762e15712e3459/.circleci/common.sh#L25-L28

WORKSPACE
```diff
@@ -0,0 +1,54 @@
...
+#    bazel --override_repository=org_tensorflow=/path/to/tensorflow
+#    or
+# b) by commenting out the http_archive above and uncommenting the following:
+# local_repository(
+#    name = "org_tensorflow",
+#    path = "/path/to/tensorflow",
+# )
+
+# Initialize TensorFlow's external dependencies.
+load("@org_tensorflow//tensorflow:workspace3.bzl", "tf_workspace3")
```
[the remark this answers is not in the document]
JackCaoG: lol what are these `tf_workspace0` -> `tf_workspace3` for?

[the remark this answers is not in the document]
will-cromar: See the TF repo: https://github.com/tensorflow/tensorflow/blob/947920176ca4b464f4c979c6d244aa2ab3a30a4d/WORKSPACE#L3-L8

We have to set up our workspace the same way as TF, but we can't actually just "import" their WORKSPACE file :cry:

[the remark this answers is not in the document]
stgpetrovic: This will dissapear soon, bazel 6.0.0. introduced modules: https://bazel.build/build/bzlmod. We wait till all tf's deps migrate and tf will follow and then we can follow too. For now, repo rule init is not transitive :S

build_torch_xla_libs.sh
```diff
@@ -34,70 +34,41 @@ if [[ "$XLA_BAZEL_VERBOSE" == "1" ]]; then
...
-  OPTS+=("--define=build_with_acl=true")
+  OPTS+=(--config=acl)
 fi
 
 if [ "$CMD" == "clean" ]; then
-  pushd $THIRD_PARTY_DIR/tensorflow
   bazel clean
-  popd
-else
-  # Overlay llvm-raw secondary cache. The remote cache should be updated
```
JackCaoG, line 81: @yeounoh regarding removing this secondary cahce, I think we added to make upstream CI more stable.

test/cpp/cpp_test_util.h
```diff
@@ -9,7 +9,7 @@
 #include <unordered_set>
 
 #include "absl/types/span.h"
-#include "tensorflow/compiler/xla/xla_client/computation_client.h"
+#include "third_party/xla_client/computation_client.h"
```
JackCaoG, line 12: since we don't need to move `xla_client` into tf to build anymore, maybe we can move the whole `xla_client` under the `torch_xla/csrc/`? We can do that in the next pr.

JackCaoG: oh nvm, I think xla_client has its own build file and uses bazel. It is more than just copying file over.

build_torch_xla_libs.sh
```diff
@@ -34,70 +34,41 @@ if [[ "$XLA_BAZEL_VERBOSE" == "1" ]]; then
...
-  unset CC
-  unset CXX
-  BUILD_STRATEGY="local"
-fi
+BUILD_STRATEGY="sandboxed --sandbox_base=${SANDBOX_BASE}"
 
-TPUVM_FLAG=
 if [[ "$TPUVM_MODE" == "1" ]]; then
-  TPUVM_FLAG="--define=with_tpu_support=true"
+  OPTS+=(--config=tpu)
```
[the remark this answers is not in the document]
stgpetrovic: (in .bazelrc)

test/cpp/cpp_test_util.h
stgpetrovic: This is the first (easier) step; migrating the libxla_c_c.so to bazel, since it's already in bazel. When we migrate the plugin to bazel then we can merge the two builds, that will give us more cache, faster build times, faster incremental builds, etc.

We can consider doing like jax does in the future, and splitting into jax/jaxlib equivalents where one is python+thin layer of cc binidngs, and the other is the cc library, mostly for ease of pushes where you'd change one more often than the other. We can discuss that after the migration and see what makes most sense.

WORKSPACE
will-cromar: Actually we might rely on `apply_patches.sh` to update to the torch pin in PRs, so it's better to add a way to skip the TF patches than remove this script right now.

miladm:
CC @ManfeiBai

.bazelrc
```diff
@@ -0,0 +1,176 @@
...
+build:linux --define=INCLUDEDIR=$(PREFIX)/include
+build:linux --define=PROTOBUF_INCLUDE_PATH=$(PREFIX)/include
+
+# On linux, we dynamically link small amount of kernels
+build:linux --config=dynamic_kernels
+
+# For projects which use TensorFlow as part of a Bazel build process, putting
+# nothing in a bazelrc will default to a monolithic build. The following line
+# opts in to modular op registration support by default.
+build --define framework_shared_object=true
```
miladm, line 121: General comment: in case we introduce / modify environment variables, we should keep [configuration.yaml](https://github.com/pytorch/xla/blob/master/configuration.yaml) up to date.

third_party/xla_client/pjrt_computation_client_test.cc
```diff
@@ -1,4 +1,6 @@
-#include "tensorflow/compiler/xla/xla_client/pjrt_computation_client.h"
+#include "third_party/xla_client/pjrt_computation_client.h"
+
+#include <gtest/gtest.h>
```
miladm, line 3: We have `gtest` pin dependency in PyTorch/XLA. Does it make sense to transition this dependency to bazel also?

.bazelrc
stgpetrovic: Agreed.

I would not introduce any env variables, I would remove all of them eventually. Configuration should happen through ansible/bazel in the future. We should only keep the ones that are required to configure downstream stuff. wdyt?

third_party/xla_client/pjrt_computation_client_test.cc
stgpetrovic: Yes, but since that touches both `torch_xla` and the `libxla_cc_.so` I'd migrate that with the last part when we remove cmake. wdyt?

JackCaoG:
@stgpetrovic The branch cut for the pytorch 2.0 is scheduled to be this Friday, I am a bit hesitant to merge this one right before the branch cut. Do you have any opinion?

stgpetrovic:
I think that we should first cut and then we have a large merge window for this large change; something's bound to explode here.

build_torch_xla_libs.sh
stgpetrovic: Not sure what to do here. We can patch the TF repo with this change.

stgpetrovic: I think that [https://github.com/tensorflow/tensorflow/blob/master/third_party/llvm/workspace.bzl#L15-L16](llvm) tensorflow external repository already names both of these URLs so bazel will fetch from both and there's no need to do this anymore.

Not 100% sure but it looks like it.

wdyt?

stgpetrovic: I've patched in TF with this 2 `sed` equivalents so we have it as a patch, not as a in-place replacement.

third_party/xla_client/pjrt_computation_client_test.cc
stgpetrovic: Once we migrate the tests to bazel, we should just depend on the pinned gtest that comes in as an external repository from tensorflow and not have cmake do it.

PR #4537, opened 2023-01-31
Lower _linalg_svd

wonjoolee95:
Lower _linalg_svd

This change lowers _linalg_svd by mostly reusing the existing svd lowering. However, there are a couple of notable differences between these two ops:
1. torch.svd()'s `some` is the opposite of torch.linalg.svd()'s `full_matrices`. Note that default value for both is True, so the default behavior is effectively the opposite.
2. torch.svd() returns V, whereas torch.linalg.svd() returns Vh (V transposed).
3. If `compute_uv` is False, torch.svd() returns zero-filled tensors for U and Vh, whereas torch.linalg.svd() returns empty tensors.

---
As a result of this change, we also disable the following ops in xla/test/test_ops.py:
- `geqrf`
- `linalg.det`
- `linalg.inv_ex`
- `linalg.matrix_rank`
- `linalg.svd`
- `linalg.svdvals`

These ops all call into the newly lowered `_linalg_svd` and cause failure: `Slice dim size 1 greater than dynamic slice dimension: 0`. This failure seems to be also happening in our master branch for the already lowered op `svd`. In xla/test/test_ops.py, ops such as `svd`, `pinverse`, or `nuclear_norm` that call `svd` are failing due to the same error and are disabled. However, to ensure correctness for the newly lowered `_linalg_svd`, this PR adds unit tests and does not cause any regressing cpp test failures.


wonjoolee95:
CI failing due to seemingly unrelated error: `ERROR: test_view_copy_out_xla (__main__.TestViewOpsXLA)`. Let me try to rebase.

wonjoolee95:
Update: this test seems like it's failing on head.

JackCaoG (approved):
Thanks! Feel free to merge after test is green

JackCaoG:
crashed with frame info
```
        xla::Shape const* ConsumeValue<xla::Shape const*>(tsl::StatusOr<xla::Shape const*>&&)
	torch_xla::SVD::Lower(torch_xla::LoweringContext*) const
	torch_xla::LoweringContext::LowerNode(torch::lazy::Node const*)
```

seems like some shape in SVD is invalid..

wonjoolee95:
Seems like I missed an op in the `test/test_ops.py` that fails with the same error message in the PR summary. Let me re-run the entire op locally and confirm. If it's failing due to the same SVD shape, I'll comment that one out as well.

wonjoolee95:
Added some more ops failing due to the same existing error with `svd`, including some more `linalg` methods along with `lu` and `qr` that XLA currently doesn't support. 

Now locally, the tests are passing but I'll let the CI verify.

test/test_ops.py
```diff
@@ -288,14 +280,20 @@ def __new__(cls, name, variant_test_name=""):
...
             # AllowedOpInfoEntry('linalg.eigvalsh'),
+            # AllowedOpInfoEntry('linalg.inv'),  # Slice dim size 1 greater than dynamic slice dimension: 0
             # AllowedOpInfoEntry('linalg.inv_ex'),  # Slice dim size 1 greater than dynamic slice dimension: 0
+            # AllowedOpInfoEntry('linalg.slogdet'),  # Slice dim size 1 greater than dynamic slice dimension: 0
+            # AllowedOpInfoEntry('linalg.qr'),  # Slice dim size 1 greater than dynamic slice dimension: 0
             # AllowedOpInfoEntry('linalg.lstsq'),
             # AllowedOpInfoEntry('linalg.norm'),
             # AllowedOpInfoEntry('linalg.matrix_norm'),
             # AllowedOpInfoEntry('linalg.matrix_rank'),  # Slice dim size 1 greater than dynamic slice dimension: 0
+            # AllowedOpInfoEntry('linalg.matrix_power'),
```
JackCaoG, line 291: can we add comments here?

wonjoolee95: Updated

JackCaoG (approved):

PR #4538, opened 2023-01-31
[OpenXLA]Migrate Stream_Executor to OpenXLA

ManfeiBai:


ManfeiBai:
need to check with files under `third_party/xla_client`

ManfeiBai:
need to check file of tf_patchs: `tf_patches/stream_executor.diff`

ManfeiBai:
failed at patches's update

Issue #4541, opened 2023-01-31
Race condition on GPU device


ymwangg:
## 🐛 Bug

We found there's likely a race condition on XLA:GPU device while debugging ViT model. I suspect it's between GPU main stream execution and BFC allocator asynchronous deallocation.

Previously we reported a race condition between nccl async communication stream and BFC allocator asynchronous deallocation https://github.com/tensorflow/tensorflow/pull/58022, but now it seems the main stream also has the same issue.

## To Reproduce

Patch the `test/test_train_mp_imagenet.py` file to support ViT model.
```
diff --git a/test/test_train_mp_imagenet.py b/test/test_train_mp_imagenet.py
index 7a7a1300..9cb75252 100644
--- a/test/test_train_mp_imagenet.py
+++ b/test/test_train_mp_imagenet.py
@@ -4,7 +4,7 @@ SUPPORTED_MODELS = [
     'alexnet', 'densenet121', 'densenet161', 'densenet169', 'densenet201',
     'inception_v3', 'resnet101', 'resnet152', 'resnet18', 'resnet34',
     'resnet50', 'squeezenet1_0', 'squeezenet1_1', 'vgg11', 'vgg11_bn', 'vgg13',
-    'vgg13_bn', 'vgg16', 'vgg16_bn', 'vgg19', 'vgg19_bn'
+    'vgg13_bn', 'vgg16', 'vgg16_bn', 'vgg19', 'vgg19_bn', 'vit_b_16',
 ]
 
 MODEL_OPTS = {
```

Run training with [imagenet](https://www.kaggle.com/competitions/imagenet-object-localization-challenge/data) dataset on a single gpu:
```
 GPU_NUM_DEVICES=1 python test_train_mp_imagenet.py --model vit_b_16 --datadir /dataset/ --batch_size 128
```
Typically there will be NaN loss shown within the first 100 step.

The following methods can fix the NaN loss issue:
1. Set `CUDA_LAUNCH_BLOCKING=1` (a good indication that there's race condition).
2. Set `AllowsAsynchronousDeallocation()=false` for BFC allocator.

## Environment

 - Reproducible on XLA backend [GPU]: tested on Nvidia-V100 and T4.
 - torch_xla version: docker image `gcr.io/tpu-pytorch/xla:nightly_3.7_cuda_11.2`, image_id = `e1d95d077920`.

## Additional context
[...]

PR #4542, opened 2023-01-31
Fix race condition on GPU

ymwangg:
This is a temporary fix for https://github.com/pytorch/xla/issues/4541 until it can be fixed on tensorflow side.

Btw: this PR should also fix the flaky ddp test on GPU.

### DIFF
--- a/tf_patches/gpu_race_condition.diff
+++ b/tf_patches/gpu_race_condition.diff
@@ -1,0 +1,14 @@
+diff --git a/tensorflow/compiler/xla/service/gpu/gpu_executable.cc b/tensorflow/compiler/xla/service/gpu/gpu_executable.cc
+index ed55b32c47a..260fddb3f5c 100644
+--- a/tensorflow/compiler/xla/service/gpu/gpu_executable.cc
++++ b/tensorflow/compiler/xla/service/gpu/gpu_executable.cc
+@@ -489,8 +489,7 @@ StatusOr<ExecutionOutput> GpuExecutable::ExecuteAsyncOnStreamImpl(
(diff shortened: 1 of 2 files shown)


tf_patches/gpu_race_condition.diff
```diff
@@ -0,0 +1,14 @@
...
+index ed55b32c47a..260fddb3f5c 100644
+--- a/tensorflow/compiler/xla/service/gpu/gpu_executable.cc
++++ b/tensorflow/compiler/xla/service/gpu/gpu_executable.cc
+@@ -489,8 +489,7 @@ StatusOr<ExecutionOutput> GpuExecutable::ExecuteAsyncOnStreamImpl(
+       "GpuExecutable::ExecuteAsyncOnStreamImpl(", module_name_, ")"));
+   se::DeviceMemoryAllocator* const memory_allocator = run_options->allocator();
+   // Force synchronous execution if the allocator requires it.
+-  const bool block_host_until_done =
+-      !memory_allocator->AllowsAsynchronousDeallocation();
++  const bool block_host_until_done = true;
```
JackCaoG, line 11: Will this has speed implication?

ymwangg: Appears not. For all the models I have tested: resnet50, ViT, bert, the speed remains the same with this patch.

JackCaoG (approved):

JackCaoG:
Thanks @ymwangg !

JackCaoG:
FYI @alanwaketan since we can enable the DDP test for GPU now.

PR #4543, opened 2023-01-31
[OpenXLA]Migrate all codes to use XLA from OpenXLA

ManfeiBai:


JackCaoG:
I think this change will have some conflict with @stgpetrovic 's bazel pr, maybe we should try to land that first?

JackCaoG:
https://github.com/pytorch/xla/pull/4528

PR #4549, opened 2023-02-01
Fix bug in ShardTensor to generate correct shards for each device

jonb377:
During multihost testing, it was found that the shard tiling was not being applied consistently, resulting in repeated results in the resulting tensor.

This change ensures that the tiles are generated based on the row-major ordering of the tile index. It also generalizes the logic to allow for higher than 5-dimensional sharding.

This issue was originally missed in unit tests because all devices were on a single axis. This change updates `test_mark_sharding_4d` to shard along two axes to check the logic here.

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -58,11 +58,14 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     expected = t + t
 
     xt = t.to(xm.xla_device())
-    xs.mark_sharding(xt, self._get_mesh((1, 1, 1, self.n_devices)),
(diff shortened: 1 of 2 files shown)


test/spmd/test_xla_sharding.py
```diff
@@ -58,11 +58,14 @@ def test_mark_sharding_4d(self):
     expected = t + t
 
     xt = t.to(xm.xla_device())
-    xs.mark_sharding(xt, self._get_mesh((1, 1, 1, self.n_devices)),
+    # Shard along two axes if four or more devices are available
```
yeounoh, line 61: +1 great!

yeounoh (approved):
LGTM, feel free to merge once all the tests pass.

PR #4524, opened 2023-01-27
Make empty_symint support dynamism

vanbasten23:

PR #4550, opened 2023-02-01
Make empty_symint support dynamism.

vanbasten23:
To address comment https://github.com/pytorch/xla/pull/4524#issuecomment-1412542586.

### DIFF
--- a/test/cpp/test_aten_xla_tensor.cpp
+++ b/test/cpp/test_aten_xla_tensor.cpp
@@ -5173,7 +5173,12 @@ TEST_F(AtenXlaTensorTest, TestOneIndexTransfer) {
   }
 }
 
-// Temporarily disable test. See  https://github.com/pytorch/xla/issues/4501
+// Temporarily disable test. Original issue
+// https://github.com/pytorch/xla/issues/4501 has been resolved. The next error
+// is https://gist.github.com/vanbasten23/b3a79e0cc7f17edc0018eb83cdd5d738 (see
+// https://github.com/pytorch/xla/issues/4432 for more info). The next error
+// happens on TPU but not on CPU.
+//
 /*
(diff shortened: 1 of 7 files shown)


test/cpp/test_aten_xla_tensor.cpp
```diff
@@ -5173,7 +5173,12 @@ TEST_F(AtenXlaTensorTest, TestOneIndexTransfer) {
   }
 }
 
-// Temporarily disable test. See  https://github.com/pytorch/xla/issues/4501
+// Temporarily disable test. Original issue
```
alanwaketan, line 5176: I guess this comment is not relevant to the master?

test/test_dynamic_shapes.py
```diff
@@ -4,6 +4,7 @@
 import torch, torch_xla
 import torch_xla.core.xla_model as xm
 
+pd = torch._C._EnablePythonDispatcher()
```
alanwaketan, line 7: Do we want to enable python dispatcher for master?

test/cpp/test_aten_xla_tensor.cpp
vanbasten23: This test is also disabled in the master branch https://github.com/pytorch/xla/blob/889a10390e388e29d804e23c99b566238c884740/test/cpp/test_aten_xla_tensor.cpp#L5176.

The comment also applies as both branches are impacted by the `nonzero` regression.

test/test_dynamic_shapes.py
vanbasten23: Please correct me if I'm wrong. To my understanding, for dynamic shape feature to run, we enable python dispatch by default.

alanwaketan (approved):
LGTM.

PR #4551, opened 2023-02-01
Implement missing XLASymNodeImpl::Sub

vanbasten23:
This is to unblock the failing dynamic shape tests when functionalization is enabled.

### DIFF
--- a/torch_xla/csrc/ops/dynamic_ir.cpp
+++ b/torch_xla/csrc/ops/dynamic_ir.cpp
@@ -96,6 +96,35 @@ XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
   return ReturnOp((input1 + input2), loctx);
 }
 
+SizeSub::SizeSub(torch::lazy::Value a, torch::lazy::Value b)
+    : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString("aten::sub")}, // TODO: should it be something like aten::size_sub? Try it out.
+              {a, b},
+              xla::ShapeUtil::MakeShape(
+                  GetShapeDimensionType(/*device*/nullptr), {}),
+              1){
+  const torch::lazy::DimensionNode* dim_node_0 = DimCast(operand(0));
+  const torch::lazy::DimensionNode* dim_node_1 = DimCast(operand(1));
+  XLA_CHECK(dim_node_0);
+  XLA_CHECK(dim_node_1);
+  upper_bound_ = dim_node_0->getStaticValue() - dim_node_1->getStaticValue();
+};
+
+int64_t SizeSub::getDynamicValue() const {
+  const torch::lazy::DimensionNode* dim_node_0 = DimCast(operand(0));
+  const torch::lazy::DimensionNode* dim_node_1 = DimCast(operand(1));
+  XLA_CHECK(dim_node_0);
+  XLA_CHECK(dim_node_1);
+  return dim_node_0->getDynamicValue() - dim_node_1->getDynamicValue(); 
+}
+
+std::string SizeSub::ToString() const { return "aten::sub_size"; }
+
+XlaOpVector SizeSub::Lower(LoweringContext* loctx) const {
+  xla::XlaOp input0 = loctx->GetOutputOp(operand(0));
+  xla::XlaOp input1 = loctx->GetOutputOp(operand(1));
+  return ReturnOp((input0 - input1), loctx);
+}
+
 SizeEq::SizeEq(torch::lazy::Value a, torch::lazy::Value b)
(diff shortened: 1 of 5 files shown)


torch_xla/csrc/tensor.cpp
```diff
@@ -639,8 +639,9 @@ c10::SymNode XLASymNodeImpl::add(const c10::SymNode& other) {
 }
 
 c10::SymNode XLASymNodeImpl::sub(const c10::SymNode& other) {
-  XLA_CHECK(false) << "XLASymNodeImpl::" << __FUNCTION__
```
alanwaketan, line 642: Wonder if we should have something like:
```
TORCH_LAZY_FN_COUNTER("xla::size_");
```

alanwaketan (approved):
LGTM.

torch_xla/csrc/tensor.cpp
vanbasten23: I feel it's ok not to increase the counter here because it's upstream that redirect the "-" operator to the `sub` method. Also, without the change here, the test will fail with error `XLASymNodeImpl::sub not implemented`

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -96,6 +96,37 @@ XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
   return ReturnOp((input1 + input2), loctx);
 }
 
+SizeSub::SizeSub(torch::lazy::Value a, torch::lazy::Value b)
+    : XlaNode(torch::lazy::OpKind{c10::Symbol::fromQualString(
+                  "aten::sub")},  // TODO: should it be something like
```
vanbasten23, line 101: I feel the OpKind here can be more specific such as "aten::size_sub", as opposed to the general ones such as "aten::sub". wdyt? @miladm @JackCaoG @alanwaketan

torch_xla/csrc/tensor.cpp
alanwaketan: The TORCH_LAZY_FN_COUNTER is more or less a way to suggest that our version of the op is called instead of the fallback one. The MARCO actually will include the method name in the counter: xla::size_sub.

torch_xla/csrc/ops/dynamic_ir.cpp
alanwaketan: c10::Symbol is something defined in upstream. If you can find the corresponding enum in upstream, I guess it's better. Yea, we don't store "aten::sub" as a string. It's an enum.

vanbasten23: I guess the enums are defined in pytorch/aten/src/ATen/core/interned_strings.h? `size_sub` is not one of the enums but I could still compile locally and run the test.

torch_xla/csrc/tensor.cpp
vanbasten23: done.

JackCaoG (approved):

torch_xla/csrc/ops/dynamic_ir.cpp
```diff
@@ -96,6 +96,36 @@ XlaOpVector SizeAdd::Lower(LoweringContext* loctx) const {
   return ReturnOp((input1 + input2), loctx);
 }
 
+SizeSub::SizeSub(torch::lazy::Value a, torch::lazy::Value b)
+    : XlaNode(
+          torch::lazy::OpKind{c10::Symbol::fromQualString("aten::size_sub")},
```
JackCaoG, line 101: Well, I am ok with either way as long as it runs, but if you want to update this for SizeSub, you need to update this for all nodes including `SizeAdd` above.

alanwaketan: From the code, it looks like c10::Symbol would just assign any unseen string an incremental ID. I guess the problem I'm having is that there is no straightforward way you can do a comparison later on. For any aten ops, you can simply do:
```
op == at::aten::sub
```

For any custom op, I guess you have to do:
```
op == c10::Symbol::fromQualString("aten::size_sub")
```

Unless you could define an enum like structure in our end. So what benefits do you think we get from aten::size_sub?

vanbasten23: It seems semantically cleaner to me: `at::aten::sub` feels like we are doing subtraction between 2 tensors, but here it's between 2 shape dimension (eg tensor.shape[0]). wdyt?

alanwaketan: From the code perspective, yes. But I wonder if this helps anything practical for instance, LazyIR dump or HLO dump.

PR #4555, opened 2023-02-02
Add move hashing to constructor

steventk-g:


torch_xla/csrc/tensor.cpp
```diff
@@ -242,15 +242,15 @@ void XLATensor::SetShardingSpec(const ShardingSpec& sharding) {
         << ", must be cleared before applying a new one, "
         << sharding.sharding.DebugString();
   }
-  dynamic_cast<XlaNode*>(GetIrValue().node.get())
-      ->SetSharding(sharding_spec()->sharding);
+  CreateShardedIrValue(sharding_spec());
```
JackCaoG, line 245: Nit, `CreateXXX` sounds like we will return something, but this method actually inplace update the `data()`. Let's maybe call it `SetShardedIrValue` instead?

torch_xla/csrc/tensor.cpp
```diff
@@ -242,15 +242,15 @@ void XLATensor::SetShardingSpec(const ShardingSpec& sharding) {
...
 }
+
 void XLATensor::ClearShardingSpec() {
   data()->sharding = nullptr;
   torch::lazy::Value ir_value = CurrentIrValue();
   if (ir_value) {
     // This should be a no-op if there is no sharding.
-    dynamic_cast<XlaNode*>(ir_value.node.get())->ClearSharding();
+    CreateUnshardedIrValue();
```
JackCaoG, line 253: ditto

torch_xla/csrc/tensor.cpp
```diff
@@ -284,6 +284,33 @@ void XLATensor::SetXlaData(torch::lazy::BackendDataPtr handle, bool sync) {
   }
 }
 
+void XLATensor::CreateShardedIrValue(
+    const ShardingSpecPtr sharding_spec) const {
+  torch::lazy::Value old_value = GetIrValue();
+  XLA_CHECK(old_value && old_value.node != nullptr)
+      << "Cannot create a sharded IR value if an IR value does not already "
+         "exist";
+  XLA_CHECK(old_value.node->op() == xla_device_data)
```
JackCaoG, line 293: maybe we can relax this check now?

torch_xla/csrc/tensor.cpp
```diff
@@ -284,6 +284,33 @@ void XLATensor::SetXlaData(torch::lazy::BackendDataPtr handle, bool sync) {
...
+      << "Can only set sharding for device data";
+  torch::lazy::Value new_ir = dynamic_cast<DeviceData*>(old_value.node.get())
+                                  ->CloneWithSharding(sharding_spec->sharding);
+  data()->ir_value = std::move(new_ir);
+  data()->generation += 1;
```
JackCaoG, line 298: generation is only used by `view`, I don't think we need to update it here?

torch_xla/csrc/tensor.cpp
```diff
@@ -284,6 +284,33 @@ void XLATensor::SetXlaData(torch::lazy::BackendDataPtr handle, bool sync) {
...
+
+void XLATensor::CreateUnshardedIrValue() {
+  torch::lazy::Value old_value = CurrentIrValue();
+  XLA_CHECK(old_value && old_value.node != nullptr)
+      << "Cannot create a unsharded IR value if an IR value does not already "
+         "exist";
+  XLA_CHECK(old_value.node->op() == xla_device_data)
+      << "Can only clear sharding for device data";
+  torch::lazy::Value new_ir =
+      dynamic_cast<DeviceData*>(old_value.node.get())->Clone();
```
JackCaoG, line 309: Seems like we need to implement a method in each IR Node to return a unsharded IR hmm,..

torch_xla/csrc/tensor.cpp
```diff
@@ -284,6 +284,33 @@ void XLATensor::SetXlaData(torch::lazy::BackendDataPtr handle, bool sync) {
...
+  data()->ir_value = std::move(new_ir);
+  data()->generation += 1;
```
JackCaoG, line 311: ditto

JackCaoG (commented):
Mostly lGTM, thanks!

torch_xla/csrc/ops/device_data.cpp
```diff
@@ -13,16 +13,34 @@ DeviceData::DeviceData(std::shared_ptr<torch::lazy::BackendData> data)
...
   ss << XlaNode::ToString() << ", device=" << data_->device();
   return ss.str();
 }
 
+torch::lazy::NodePtr DeviceData::Clone() const {
+  return Clone(operands_as_oplist());
+}
+
 torch::lazy::NodePtr DeviceData::Clone(torch::lazy::OpList operands) const {
   return torch::lazy::MakeNode<DeviceData>(data_);
```
alanwaketan, line 35: operands are not used at all in this method...

torch_xla/csrc/ops/device_data.h
```diff
@@ -10,10 +10,18 @@ class DeviceData : public XlaNode {
...
 
+  DeviceData(std::shared_ptr<torch::lazy::BackendData> data,
+             torch::lazy::OpList ops, xla::Shape xla_shape,
+             xla::OpSharding sharding);
+
   std::string ToString() const override;
 
+  torch::lazy::NodePtr Clone() const;
+
   torch::lazy::NodePtr Clone(torch::lazy::OpList operands) const override;
```
alanwaketan, line 21: Hmmm, this Clone method is only used in https://github.com/pytorch/xla/blob/master/torch_xla/csrc/ir_util.h#L15, and those two Util::Clone are not used anywhere...

I guess we can actually remove it. But let's do it in a separate patch. Could you leave a TODO in XlaNode::Clone.

steventk-g: Added a TODO https://github.com/pytorch/xla/issues/4568

torch_xla/csrc/tensor.cpp
steventk-g: Done

torch_xla/csrc/tensor.cpp
steventk-g: Done

torch_xla/csrc/tensor.cpp
steventk-g: Done

torch_xla/csrc/tensor.cpp
steventk-g: Done

torch_xla/csrc/tensor.cpp
steventk-g: Done

torch_xla/csrc/ops/device_data.cpp
```diff
@@ -13,16 +13,34 @@ DeviceData::DeviceData(std::shared_ptr<torch::lazy::BackendData> data)
...
+      data_(std::move(data)) {}
+
 std::string DeviceData::ToString() const {
   std::stringstream ss;
   ss << XlaNode::ToString() << ", device=" << data_->device();
   return ss.str();
 }
 
+torch::lazy::NodePtr DeviceData::Clone() const {
+  return Clone(operands_as_oplist());
```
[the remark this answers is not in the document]
steventk-g: Done

torch_xla/csrc/tensor.cpp
steventk-g: We only need to add it for the node types which will be cloned at runtime, but I'm not sure if that would be everything. We can always leave it with the base implementation (throwing a not implemented error) otherwise.

yeounoh:
Let's mark the PR (DO NOT MERGE YET) and/or add a comment about the ongoing investigation.

torch_xla/csrc/ir.cpp
```diff
@@ -48,15 +48,30 @@ XlaNode::XlaNode(torch::lazy::OpKind op, torch::lazy::OpList operands,
     : torch::lazy::Node(op, operands, std::move(shapes), num_outputs),
       xla_shape_(std::move(xla_shape)),
       node_hash_(torch::lazy::HashCombine(op.hash(), hash_seed)),
-      dag_hash_(GetOperandHashes(operands, node_hash_)) {}
+      dag_hash_(GetOperandHashes(operands, node_hash_)),
+      oplist_(std::optional<torch::lazy::OpList>(operands)) {}
```
yeounoh, line 52: Is it possible to construct OpList from `operands_` in `lazy::Node`? We are passing `operands` as input to the parent constructor and also to `oplist_`. Hopefully, that will free us from adding this additional constructor argument and possibly be safer, as we don't have to maintain two copies of `operands_` in two different formats. `operands_` is protected, so should be accessible.

torch_xla/csrc/ir.cpp
```diff
@@ -101,11 +121,14 @@ XlaNode::XlaNode(torch::lazy::OpKind op, torch::lazy::Shape shape,
     : torch::lazy::Node(op, shape, num_outputs),
       xla_shape_(std::move(xla_shape)),
       node_hash_(GetOpHash(op, xla_shape_, hash_seed)),
-      dag_hash_(node_hash_) {}
+      dag_hash_(node_hash_),
+      oplist_(std::nullopt) {}
```
yeounoh, line 125: Along the same line with my previous [comment](https://github.com/pytorch/xla/pull/4555/files#r1114930890), we are introducing a gap here between `operands_`, `operands_as_outputs_` and `oplist_` (empty). And we probably want to avoid this additional null initialization if possible.

torch_xla/csrc/ir.h
```diff
@@ -151,11 +151,12 @@ class XlaNode : public torch::lazy::Node {
   xla::Shape xla_shape_;
   torch::lazy::hash_t node_hash_ = 0;
   torch::lazy::hash_t dag_hash_;
-  torch::lazy::hash_t sharding_hash_ = 0;
 
   // Experimental sharding annotation attached to the IR node.
   // TODO(yeounoh): make sure that view update doesn't reset this.
   std::shared_ptr<xla::OpSharding> output_sharding_ = nullptr;
+
+  std::optional<torch::lazy::OpList> oplist_;
```
yeounoh, line 159: I guess a major issue we need to work around would be keeping the cross references between ops up-to-date, given that the graph is a tree. It might be doable with additional bookkeeping and traversals, but not sure how expensive that would be.

torch_xla/csrc/ir.cpp
steventk-g: I followed your suggestion here to move the logic into XlaNode. Should I go back to using lazy::Node like I was originally? https://github.com/pytorch/pytorch/pull/92579/files#r1105075942

torch_xla/csrc/ops/device_data.cpp
```diff
@@ -13,16 +13,34 @@ DeviceData::DeviceData(std::shared_ptr<torch::lazy::BackendData> data)
               /*hash_seed=*/(uint32_t)101),
       data_(std::move(data)) {}
 
+DeviceData::DeviceData(std::shared_ptr<torch::lazy::BackendData> data,
+                       torch::lazy::OpList ops, xla::Shape xla_shape,
+                       xla::OpSharding sharding)
+    : XlaNode(xla_device_data, ops, {data->shape()}, xla_shape, sharding,
```
yeounoh, line 19: Do we need `xla_shape` as input? Could we do this like in the other constructor `UnwrapXlaData(data)->shape()`?

torch_xla/csrc/ir.h
steventk-g: I think the best way to do it will be storing ops as smart pointers, either shared or weak, so that we can access them within an oplist by index while also being able to replace the node with a new node.

torch_xla/csrc/ir.cpp
steventk-g: I was planning to close the upstream PR and localize these changes in XlaNode. Do you think we should move it back to lazy::Node like what I had before?

torch_xla/csrc/ops/device_data.cpp
```diff
@@ -13,16 +13,34 @@ DeviceData::DeviceData(std::shared_ptr<torch::lazy::BackendData> data)
...
+              /*hash_seed=*/(uint32_t)101),
+      data_(std::move(data)) {}
+
 std::string DeviceData::ToString() const {
   std::stringstream ss;
   ss << XlaNode::ToString() << ", device=" << data_->device();
   return ss.str();
 }
 
+torch::lazy::NodePtr DeviceData::Clone() const { return Clone({}); }
```
yeounoh, line 30: Is this calling?
```
// TODO(https://github.com/pytorch/xla/issues/4567) Remove this clone method
  virtual torch::lazy::NodePtr Clone(torch::lazy::OpList operands) const;
  ```
  Let's add some comments why we are passing an empty `oplist` argument here.
  
  Will this be removed/updated when we address https://github.com/pytorch/xla/issues/4567?

Comment (yeounoh) [reply torch_xla/csrc/ops/device_data.cpp:30]: I see that you have a similar comment below, 
```
// TODO(steventk) Ideally, we are either passing the actual oplist, or we
  // don't pass ops at all and use another XlaNode constructor
  ```

steventk-g: I'll add the comment to where we make the call too

torch_xla/csrc/ops/device_data.cpp
steventk-g: I can experiment with it. Any reason to exclude xla_shape?

torch_xla/csrc/ops/device_data.h
```diff
@@ -10,10 +10,19 @@ class DeviceData : public XlaNode {
...
+             torch::lazy::OpList ops, xla::Shape xla_shape,
+             xla::OpSharding sharding);
+
   std::string ToString() const override;
 
+  torch::lazy::NodePtr Clone() const override;
+
   torch::lazy::NodePtr Clone(torch::lazy::OpList operands) const override;
 
+  torch::lazy::NodePtr CloneWithSharding(
```
yeounoh, line 23: What happens to the other node types? We can attach sharding to any.

steventk-g: Unless they have this method implemented it will crash at runtime. So far, only generic and device data are effected. If we need to support sharding for any node, we will need to add these functions to each node class.

torch_xla/csrc/ops/generic.cpp
```diff
@@ -37,11 +38,48 @@ Generic::Generic(torch::lazy::OpKind op, xla::Shape shape, LowerFn lower_fn,
...
 torch::lazy::NodePtr Generic::Clone(torch::lazy::OpList operands) const {
   return torch::lazy::MakeNode<Generic>(op(), operands, xla_shape(), lower_fn_,
                                         num_outputs(), hash_seed_);
 }
 
+torch::lazy::NodePtr Generic::CloneWithSharding(
+    xla::OpSharding sharding) const {
+  // TODO(steventk) Right now, we drop the operands on clone, because the oplist
+  // memory becomes unsafe when we clone the other nodes and they go out of
+  // scope. Instead of initializing the oplist below to nullopt, we want to use
```
yeounoh, line 70: Thiking out loud, constructing the `oplist` from the parent's `operands_`, if possible, could simplify things. As long as the operands are alive, they shoulnd't go out of scope. For the recreated node, we should propagate that change (old->new), though.

torch_xla/csrc/tensor.cpp
yeounoh: it's little verbose, but I actually preferred `dynamic_cast<XlaNode*>(ir_value.node.get())->XYZ` to show that the tensor method `ClearShardingSpec` calls the underlying node's method.

torch_xla/csrc/tensor.cpp
```diff
@@ -284,6 +284,26 @@ void XLATensor::SetXlaData(torch::lazy::BackendDataPtr handle, bool sync) {
   }
 }
 
+void XLATensor::CreateShardedIrValue(const ShardingSpecPtr sharding_spec) {
+  torch::lazy::Value old_value = GetIrValue();
```
yeounoh, line 288: Use `CurrentIrValue()` ?

torch_xla/csrc/tensor.cpp
```diff
@@ -284,6 +284,26 @@ void XLATensor::SetXlaData(torch::lazy::BackendDataPtr handle, bool sync) {
...
+  XLA_CHECK(old_value && old_value.node != nullptr)
+      << "Cannot create a sharded IR value if an IR value does not already "
+         "exist";
+  torch::lazy::Value new_ir = dynamic_cast<XlaNode*>(old_value.node.get())
+                                  ->CloneWithSharding(sharding_spec->sharding);
+  data()->ir_value = std::move(new_ir);
```
yeounoh, line 294: some node's operands could be pointing to the old_ir.

torch_xla/csrc/tensor.cpp
steventk-g: I was getting crashes with Current, it looks like ir_value is not always defined when we call this method.

torch_xla/csrc/tensor.cpp
steventk-g: Yeah I think this is the root cause of the problem I was seeing before. If I can keep the operands up to date, the clone should work smoothly.

torch_xla/csrc/ops/generic.cpp
steventk-g: > constructing the oplist from the parent's operands_

Ah, like "clone" the operand in the oplist instead of the parent. That makes sense to me, but I'm not sure how to get a reference to the parent

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -512,8 +512,7 @@ XLAGraphExecutor::SyncTensorCollection XLAGraphExecutor::CollectSyncTensors(
           // sharding takes the precedence as the source of the truth.
           XLATensor::ShardingSpecPtr sharding = tensors[i]->sharding_spec();
           if (sharding) {
-            dynamic_cast<XlaNode*>(ir_value.node.get())
-                ->SetSharding(sharding->sharding);
+            tensors[i]->CreateShardedIrValue(sharding);
```
yeounoh, line 515: ditto, https://github.com/pytorch/xla/pull/4555/files#r1114969286

torch_xla/csrc/ir.cpp
yeounoh: No circular dependency here. Instead of doing it in `lazy::Node`, can we just take `operands_` from the parent (`lazy::Node`) and construct OpList in `XlaNode`? 

Also,

yeounoh: Synced offline with @alanwaketan , `node(node)` from `torch::lazy::Value` ctor would hold the reference to `Node`.

torch_xla/csrc/ops/device_data.cpp
yeounoh: Generally, if it's not needed or redundant, then I would not add more arguments to the constructor.

torch_xla/csrc/ops/device_data.h
yeounoh: I see.. I think we have data & linear model sharding test cases. Could you try with conv layer as well, to see if that would also require touching other nodes? I hope we can cover most of common cases generic & data.

alanwaketan:
@steventk-g Can you remind me what are the problems we are trying to solve? And if there is a test case to reproduce the problem?

steventk-g:
> @steventk-g Can you remind me what are the problems we are trying to solve? And if there is a test case to reproduce the problem?

@alanwaketan The original runtime bug is solved and tested in https://github.com/pytorch/xla/pull/4287, this PR is intended to clean up the IR hash logic so that the hash for a specific node isn't mutable.

alanwaketan:
> > @steventk-g Can you remind me what are the problems we are trying to solve? And if there is a test case to reproduce the problem?
> 
> @alanwaketan The original runtime bug is solved and tested in #4287, this PR is intended to clean up the IR hash logic so that the hash for a specific node isn't mutable.

You mentioned some bugs are exposed if we are not taking the new approach. That's what I'm curious about. Thanks for the briefing.

PR #4557, opened 2023-02-03
Modify tpu version regex to match new names

steventk-g:
New names will be used for ACCELERATOR_TYPE field, so we need to relax our regex matching to match those names correctly and extract the version


ManfeiBai (approved):
excellent！

ManfeiBai (approved):

ManfeiBai (approved):

will-cromar (approved):
Thanks!

chandrasekhard2:
Thanks @steventk-g @ManfeiBai @will-cromar

PR #4560, opened 2023-02-03
Skip git pull if we're in the xla repo

steventk-g:
Today, the wheel build script is designed to run on our docker images. However, this script has been duplicating the pytorch and xla repos due to some old logic, which has result in a large image. This has impacts for image pull and container start times, and can introduce issues with running containers on TPUVMs with limited storage capacity.

Specifically, this is affecting our CI, causing timeouts before the kubernetes pod running our container can start. This change is intended to cut down the image size to improve container start times and restore TPU CI to green.

### DIFF
--- a/scripts/build_torch_wheels.sh
+++ b/scripts/build_torch_wheels.sh
@@ -86,14 +86,17 @@ function maybe_install_sources {
     popd
   fi
 
-  # Check if we have cloned pytorch and cd into the pytorch dir. Within the pytorch
-  # dir there is a subdir `torch`.
-  if [ ! -d "torch" ]; then
+  # Check if we have cloned pytorch or xla and cd into the pytorch dir. 
+  # Within the pytorch dir there is a subdir `torch`, and within xla 
+  # is `torch_xla`.
+  if [ ! -d "torch" && ! -d "torch_xla" ]; then
     sudo apt-get install -y git
     git clone --recursive https://github.com/pytorch/pytorch.git
     cd pytorch
     git clone --recursive https://github.com/pytorch/xla.git
     export RELEASE_VERSION="nightly"
+  elif [ -d "torch_xla" ]; then
+    cd ..
   fi
 }
 


scripts/build_torch_wheels.sh
```diff
@@ -86,14 +86,17 @@ function maybe_install_sources {
...
+  # Check if we have cloned pytorch or xla and cd into the pytorch dir. 
+  # Within the pytorch dir there is a subdir `torch`, and within xla 
+  # is `torch_xla`.
+  if [ ! -d "torch" && ! -d "torch_xla" ]; then
     sudo apt-get install -y git
     git clone --recursive https://github.com/pytorch/pytorch.git
     cd pytorch
     git clone --recursive https://github.com/pytorch/xla.git
     export RELEASE_VERSION="nightly"
+  elif [ -d "torch_xla" ]; then
```
vanbasten23, line 98: do you mind adding a comment about under what circumstance there will be a `torch_xla` dir?

ManfeiBai (approved):

scripts/build_torch_wheels.sh
```diff
@@ -86,14 +86,17 @@ function maybe_install_sources {
     popd
   fi
 
-  # Check if we have cloned pytorch and cd into the pytorch dir. Within the pytorch
-  # dir there is a subdir `torch`.
-  if [ ! -d "torch" ]; then
+  # Check if we have cloned pytorch or xla and cd into the pytorch dir.
```
yeounoh, line 89: Shouldn't the comment be: We are checking "if we ... and cd into the `pytorch` or `pytorch/xla`  dir." 

Also, it may not apply to our cloudbuild triggers, but what happens if we have `torch`, but not `xla` which contains `torch_xla`? The first `pytorch` dir contains `xla` with some starter scripts -- without actual code. 

I don't want to over-convolute the case, so as long as our trigger builds run, I am happy.

scripts/build_torch_wheels.sh
yeounoh: if you git clone PyTorch/XLA repo, then the project should contain `torch_xla` dir. So the short answer, if you have cloned the repo. Just like checking for `torch` dir, I think this `torch_xla` check can be implied IMO.

vanbasten23: yeah I know, I guess what I want to know is under what occasion we have done a `git clone pytorch/xla repo`.

scripts/build_torch_wheels.sh
steventk-g: For the comment, we actually want to end up in the `pytorch` directory after this block. Previously, we clone `pytorch`, `cd` in, then clone `xla`, but stay in `pytorch`. Now, if we are in `xla`, we can `cd ..` to `pytorch`. This way, the change doesn't break any other logic that relies on us being in `pytorch` after this block gets executed.

As for the edge cases, this can definitely still be broken if we run the wheel build script under different circumstances. Hopefully this will be relatively stable. 

The build succeeded, I'm confirming that the duplicate repos aren't present on the new image.

ManfeiBai:
`ci/circleci: pytorch_xla_run_build` 's test failed might disappear after rebase with https://github.com/pytorch/xla/pull/4561

JackCaoG (approved):

JackCaoG:
@steventk-g can you rebase this branch?

yeounoh (approved):
LGTM

PR #4561, opened 2023-02-03
Remove unused docker tagging in .circleci/config.yml

yeounoh:
The docker tag is not used. We use a version tag instead. This fixes the issue because extracting the tag was failing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,9 +83,6 @@ calculate_docker_image: &calculate_docker_image
   command: |
     # Just a quick smoke test to see if we can actually extract the tag
     git clone --quiet https://github.com/pytorch/pytorch.git "/tmp/pytorch"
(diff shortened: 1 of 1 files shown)


vanbasten23:
can you explain how this pr would fix the error 
```
HEAD:.circleci/docker
fatal: path '.circleci/docker' does not exist in 'HEAD'
```
?

steventk-g (approved):

.circleci/config.yml
```diff
@@ -83,9 +83,6 @@ calculate_docker_image: &calculate_docker_image
   command: |
     # Just a quick smoke test to see if we can actually extract the tag
```
steventk-g, line 84: NB can we remove this comment as well?

ManfeiBai (approved):

JackCaoG (approved):

PR #4570, opened 2023-02-04
Actual build and install sccache from source

huydhn:
This is a miss from my previous PR https://github.com/pytorch/xla/pull/4489.  I didn't realize that the existing command `cargo install sccache` would download, build, and install sccache (v0.3.3) from upstream repo like conda. That command never even uses the local checkout from git.  So in fact, https://github.com/pytorch/xla/pull/4489 is a no-op.

The correct command to use is `cargo install --path .` to build and install from local source.

### DIFF
--- a/.circleci/docker/Dockerfile
+++ b/.circleci/docker/Dockerfile
@@ -90,7 +90,7 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 RUN . $CARGO_HOME/env && \
     git clone --recursive https://github.com/pytorch/sccache.git && \
     cd sccache && \
-    cargo install sccache && \
+    cargo install --path . && \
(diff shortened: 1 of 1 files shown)


huydhn:
@yeounoh Could you help take a look a this follow-up fix?  I didn't up the docker image version, but let me know if I need to do so.

huydhn:
FYI, I did a test on PyTorch https://github.com/pytorch/pytorch/actions/runs/4089793155/jobs/7053582177 to ensure that XLA build and test jobs looks correct now. The test manually installs and uses sccache version from https://github.com/pytorch/sccache

yeounoh:
Thanks @huydhn I will handle the image version update (I want to do it with CUDA update).

yeounoh (approved):
LGTM, I will merge this along with the CUDA version update so we only need to update the base image once.

PR #4572, opened 2023-02-06
[FSDP] Use state_dict hook to save shard metadata for consolidating checkpoint

YangFei1990:
Currently the way to consolidate full checkpoint for FSDP works as saving a `shard_metadate` object which is expecting to create a one-to-one sharding metadata information mapping to the model `state_dict` parameters. And those sharding metadata information will be used to restore the full parameters. However, the current implementation will fail to create such one-to-one mapping when there are shared modules. Consider the following example:
```
import torch
import torch.nn as nn

from torch_xla.distributed.fsdp import (
    XlaFullyShardedDataParallel as FSDP,
    consolidate_sharded_model_checkpoints,
)

class TieWeightEmbedding(nn.Embedding):
    """
        An embedding layer whose weight can also be used in linear transformation.
    """

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # A bias term for the linear transformation.
        self.bias = nn.Parameter(self.weight.new_zeros(self.weight.size(0)))

    def forward(self, indices_or_hidden_states,
                use_as_linear_transformation=False):
        if use_as_linear_transformation:
            return torch.nn.functional.linear(indices_or_hidden_states,
                                              self.weight, self.bias)

        return super().forward(indices_or_hidden_states)

class LMHead(torch.nn.Module):
    def __init__(self, m):
        super().__init__()
        self.lm_head = m
    
    def forward(self, input):
        return self.lm_head(input, use_as_linear_transformation=True)


class TestModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.embedding = FSDP(TieWeightEmbedding(10, 3))
        self.linear = nn.Linear(3, 3)
        self.lm_head = LMHead(self.embedding)
    
    def forward(self, input):
        x = self.embedding(input)
        x = self.linear(x)
        x = self.lm_head(x)
        return x

def run():
    device = xm.xla_device()
    print(f"get_ordinal {xm.get_ordinal()}")
    model = TestModule()
    model = FSDP(model)
    model.to(device)
    dummy_data = torch.LongTensor([[1,2,4,5],[4,3,2,9]]).to(device)
    output = model(dummy_data)
    xm.mark_step()
    # Save the final model checkpoint
    rank = xm.get_ordinal()
    ckpt_prefix = "fsdp"
    ckpt_path = f'{ckpt_prefix}_rank-{rank:08d}-of-{world_size:08d}.pth'
    ckpt = {
        'model': model.state_dict(),
        'shard_metadata': model.get_shard_metadata(),
    }
    xm.save(ckpt, ckpt_path, master_only=False)
    print(f'checkpoint saved to {ckpt_path}\n', end='')

    # Consolidate the sharded model checkpoints and test its accuracy
    if xm.is_master_ordinal(local=False):
      consolidate_sharded_model_checkpoints(
          ckpt_prefix=ckpt_prefix, ckpt_suffix="_rank-*-of-*.pth")
    xm.rendezvous('ckpt_consolidation')

    

def _mp_fn(index):
    run()

if __name__ == "__main__":
    xmp.spawn(_mp_fn, args=(), nprocs=8)
``` 
The same `TieWeightEmbedding` module is used for both `embedding` and `lm_head`, which is a very common use case for LLM. However the sharding metadata for parameter `lm_head.lm_head` will be missing since it is a shared module, but `lm_head.lm_head.weight` will exist in the model state_dict.

To solve such issue, this PR use `module._register_state_dict_hook` to register the `get_shard_metadata` hook, so that the shard metadata will be saved together with the parameter. This implementation enforces the one-to-one mapping relationship between `shard_metadata` and model's state_dict parameters, as well as simplifies the user experience as they do not need to save an extra `shard_metadata` object.


JackCaoG:
@ronghanghu Do you have cycle to take a look? I will also try to take a look.

ronghanghu:
Thanks @YangFei1990 and @JackCaoG! I'll take a look some time today

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -483,6 +487,9 @@ def __init__(
       gc.collect()
       xm.mark_step()
 
+    if save_shard_metadata:
+      self._register_state_dict_hook(self.get_shard_metadata)
```
ronghanghu, line 491: If I understand correctly, this `_register_state_dict_hook` will cause `self.get_shard_metadata` to be called on **every** FSDP module (not just the root FSDP module)? I believe we would only want to call `self.get_shard_metadata` on the root instance

The implementation in this PR seems to lead to some problems in nested FSDP as well as multiple root FSDP instances. Consider the following model hierarchy
```
class A(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.linear_fsdp1 = FSDP(torch.nn.Linear(10, 3))
        self.linear_regular = torch.nn.Linear(10, 4)

class B(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.a = FSDP(A())
        self.linear_fsdp2 = FSDP(torch.nn.Linear(10, 5))

b = B()
```
We will only want to `self.get_shard_metadata` on `b.a` and `b.linear_fsdp2` since they are the root FSDP instances (but not `b.a.linear_fsdp1` since it's not a root FSDP). But when calling `b.state_dict()`, it cannot guarantee which FSDP's metadata will be saved to `destination["shard_metadata"]`.

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1471,48 +1478,58 @@ def get_original_names_and_sharded_parameters(self):
...
+          "shard_info": shard_info,
+          "flatten_info": flatten_info,
+          "buffer_info": buffer_info,
+          "world_size": m.world_size,
+          "rank": m.rank,
+      }
+    else:
+      destination["shard_metadata"]["shard_info"].update(shard_info)
+      destination["shard_metadata"]["flatten_info"].update(flatten_info)
+      destination["shard_metadata"]["buffer_info"].update(buffer_info)
```
ronghanghu, line 1532: And would this cause a name conflict if there is a parameter or a buffer under the same named `"shard_metadata"`?

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1471,48 +1478,58 @@ def get_original_names_and_sharded_parameters(self):
...
-          flatten_info[param_name] = m.metadata(i)
-
-    for name, buf in self.named_buffers():
+    # remove "_fpw_module." from module names since it is also removed in
+    # XlaFullyShardedDataParallel's state_dict()
+    module_name = prefix.replace("_fpw_module.", "")
+    # remove the trailing "." since state_dict will add it
+    module_name = module_name.rstrip(".")
+
+    if isinstance(m, XlaFullyShardedDataParallel):
```
ronghanghu, line 1501: I'm still trying to understand this. If `get_shard_metadata` is registered to be a state dict hook of the the `XlaFullyShardedDataParallel` class, then isn't `m` guaranteed to be an instance of `XlaFullyShardedDataParallel` (as how it's called in `nn.Module` in https://github.com/pytorch/pytorch/blob/v1.13.1/torch/nn/modules/module.py#L1449)?

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1471,48 +1478,58 @@ def get_original_names_and_sharded_parameters(self):
...
+      shard_info[module_name] = sharded_param_info
+
+    if isinstance(m, XlaFlattenParamsWrapper):
+      for i in range(len(m.flat_params)):
+        param_name = f"flat_param_{i}"
+        if module_name != "":
+          param_name = module_name + "." + param_name
+        flatten_info[param_name] = m.metadata(i)
+
+    for name, buf in m.named_buffers(recurse=False):
```
ronghanghu, line 1517: If we have `recurse=False`, would we miss buffers such as `model.module1.module2.buffer` when we only wrap the base `model` with the FSDP class?

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1471,48 +1478,58 @@ def get_original_names_and_sharded_parameters(self):
...
     """
     Get the shard metadata to consolidate the sharded model checkpoints.
-    The output from this method should be saved in a checkpoint file and
-    used in ``consolidate_sharded_model_checkpoints``.
+    The output from this method will be used in ``consolidate_sharded_model_checkpoints``.
     """
     shard_info = {}
     flatten_info = {}
     buffer_info = {}
-    for module_name, m in self.named_modules():  # includes self
```
ronghanghu, line 1483: I wonder would the `lm_head.lm_head` example in this PR be fixed if we change this line to 
```
for module_name, m in self.named_modules(remove_duplicate=False): # includes self
```

ronghanghu:
Hi @YangFei1990, thanks for the PR, I left a few comments above. 

Overall I feel I didn't fully understand the change, or why it is needed to move the `get_shard_metadata` to state dict hook. I'm also a bit skeptical of putting it into `state_dict` since it would risk overriding a param/buffer under the same name "shard_metadata".

Regarding the `lm_head.lm_head` example in this PR, could it be fixed if we change this line (https://github.com/pytorch/xla/blob/38417751e660714a8f815a24b679940c25245ad6/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py#L1483) to 
```
for module_name, m in self.named_modules(remove_duplicate=False): # includes self
```
to count for shared modules via `remove_duplicate=False`?

YangFei1990:
Thanks for all the comments @ronghanghu ! As mentioned in the description, the motivation for this PR is to solve the module reuse issue. I was under the impression that the parameter in the `shard_metadata` should have one-to-one mapping to the parameter in the state_dict, and thus came up with the idea to use state_dict hook. However I did miss the part for buffer and `XlaFlattenParamsWrapper`. The solution with `remove_duplicate=False` looks good to me, I have modified the PR to use it.

ronghanghu (approved):
Thanks @YangFei1990, looks good to me!

(And I think this needs to be split into two lines now to make the linter happy.)

JackCaoG:
@YangFei1990 can you rebase? If CI pass I can merge this one.

PR #4573, opened 2023-02-06
Support multihost SPMD execution

jonb377:
The only main change to support multihost execution is to restrict the generated shards in ShardTensor to those which belong to addressable devices.

### DIFF
--- a/torch_xla/csrc/tensor_util.cpp
+++ b/torch_xla/csrc/tensor_util.cpp
@@ -931,30 +931,26 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
     std::vector<xla::ComputationClient::DataPtr> new_handles;          // out
     if (shardings[i] != nullptr) {
       xla::OpSharding sharding = shardings[i]->sharding;
-      // TODO(yeounoh) PJRT runs a process per host for SPMD and without cross
-      // host communications. This means that we may need to manually shard
-      // across global devices for multi-host training.
+      // GetLocalDevices returns the list of local devices specified by their
+      // global ordinals (e.g. ["TPU:4", "TPU:5", "TPU:6", "TPU:7"]).
       std::vector<std::string> local_devices =
-          xla::ComputationClient::Get()->GetAllDevices();
+          xla::ComputationClient::Get()->GetLocalDevices();
       // Shards the input tensors with padding, to split evenly.
       // The execution requires consistent shard sizes, and the zero-padded
       // values should be ignored.
-      std::vector<at::Tensor> shards = ShardingUtil::ShardTensor(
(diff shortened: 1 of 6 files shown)


torch_xla/experimental/xla_sharding.py
```diff
@@ -88,7 +89,7 @@ def mark_sharding(t: Union[torch.Tensor, XLAShardedTensor], mesh: Mesh,
     Examples
     —------------------------------
     mesh_shape = (4, 2)
-    num_devices = len(xm.get_xla_supported_devices())
+    num_devices = pjrt.global_device_count()
```
yeounoh, line 92: Great :)

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -175,26 +175,41 @@ xla::HloModuleProto ShardingUtil::SpmdPartitioningPass(
   return module.get()->ToProto();
 }
 
+// Builds a map from the device's global ordinal to its index in the `devices`
+// array.
+static std::unordered_map<int, int> build_index_map(
+    const std::vector<std::string>& devices) {
+  std::unordered_map<int, int> device_index;
+  for (int i = 0; i < devices.size(); ++i) {
+    int global_ordinal = ParseDeviceString(devices[i]).ordinal();
```
yeounoh, line 184: The first global device gets the local index 0, so the order of the input devices list is important. Is this a correct understanding? Can we add some comments on this?

yeounoh (changes requested):
LGTM, a minor comment.

torch_xla/csrc/xla_sharding_util.cpp
jonb377: The first device in the list gets local index 0, but the order of the global ordinals within `devices` doesn't matter. I'll add some more documentation around this.

torch_xla/csrc/tensor_util.cpp
```diff
@@ -931,30 +931,24 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
     std::vector<xla::ComputationClient::DataPtr> new_handles;          // out
     if (shardings[i] != nullptr) {
       xla::OpSharding sharding = shardings[i]->sharding;
-      // TODO(yeounoh) PJRT runs a process per host for SPMD and without cross
-      // host communications. This means that we may need to manually shard
-      // across global devices for multi-host training.
       std::vector<std::string> local_devices =
```
yeounoh, line 934: Does `GetLocalDevices()` return local devices with global ordinals? If so, let's leave a comment.

yeounoh (approved):
LGTM, have 2 comments --nit.

PR #4577, opened 2023-02-06
Fix ScatterReduce

alanwaketan:
Summary:
ScatterReduce::reduce_ uses a unsafe type c10::string_view to store the string. Replace it with std::string and re-enable all previous failing tests.

Test Plan:
CI.

### DIFF
--- a/torch_xla/csrc/ops/scatter_reduce.h
+++ b/torch_xla/csrc/ops/scatter_reduce.h
@@ -19,7 +19,7 @@ class ScatterReduce : public XlaNode {
   int64_t dim() const { return dim_; };
 
  private:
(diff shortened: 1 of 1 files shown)


wonjoolee95 (approved):

vanbasten23 (approved):

alanwaketan:
Thanks Wonjoo and Xiongfei.

alanwaketan:
> Thanks! Can you explain why c10::string_view is unsafe and why there is a `reduce_=�`? Lots of places in pytorch or pt/xla uses c10::string_view

It's basically a const reference. Since the data is already gone, it just reads from random data in the memory that doesn't comply with any human readable encoding.

PR #4582, opened 2023-02-07
Fix test_forward_pass_dynamic_input_compile_once

vanbasten23:
The current `test_forward_pass_dynamic_input_compile_once` is easy to break: the number of compilation varies when the functionalization is enabled or not. This PR is intended to make it more stable.

### DIFF
--- a/test/test_dynamic_shape_models.py
+++ b/test/test_dynamic_shape_models.py
@@ -135,9 +135,9 @@ class TestDynamicShapeModels(unittest.TestCase):
       for j in range(num_features):
         x_test[i][j] = 1
         num_non_zero_added += 1
-        if num_non_zero_added == num_non_zero_added:
+        if num_non_zero_added == num_non_zeros:
           break
-      if num_non_zero_added == num_non_zero_added:
+      if num_non_zero_added == num_non_zeros:
         break
 
     num_non_zero_added = 0
@@ -145,7 +145,7 @@ class TestDynamicShapeModels(unittest.TestCase):
     for i in range(num_test_samples * 2):
       y_test[i] = 1
       num_non_zero_added += 1
-      if num_non_zero_added == num_non_zero_added:
+      if num_non_zero_added == num_non_zeros:
         break
 
     x_test_xla = x_test.to(device)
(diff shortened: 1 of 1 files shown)


test/test_dynamic_shape_models.py
```diff
@@ -117,11 +124,29 @@ def test_backward_pass_with_dynamic_input(self):
...
+                               num_features,
+                               device,
+                               num_non_zeros=1):
+    x_test = torch.zeros(num_test_samples, num_features)
+    num_non_zero_added = 0
+    for i in range(num_test_samples):
+      for j in range(num_features):
+        x_test[i][j] = 1
+        num_non_zero_added += 1
+        if num_non_zero_added == num_non_zero_added:
```
alanwaketan, line 138: What does this condition do? Is it supposed to be always True?

test/test_dynamic_shape_models.py
```diff
@@ -117,11 +124,29 @@ def test_backward_pass_with_dynamic_input(self):
...
+          break
+      if num_non_zero_added == num_non_zero_added:
```
alanwaketan, line 140: Ditto.

test/test_dynamic_shape_models.py
```diff
@@ -117,11 +124,29 @@ def test_backward_pass_with_dynamic_input(self):
...
+        break
+
+    num_non_zero_added = 0
+    y_test = torch.zeros(num_test_samples * 2)
+    for i in range(num_test_samples * 2):
+      y_test[i] = 1
+      num_non_zero_added += 1
+      if num_non_zero_added == num_non_zero_added:
```
alanwaketan, line 148: Ditto.

test/test_dynamic_shape_models.py
vanbasten23: oops, fixed.

test/test_dynamic_shape_models.py
vanbasten23: fixed

test/test_dynamic_shape_models.py
vanbasten23: fixed

alanwaketan (approved):
LGTM.

PR #4583, opened 2023-02-07
[PJRT] Slice literals to return correct dynamic shape

will-cromar:
PJRT allocates a literal according to the maximum bounded size of a buffer: https://github.com/tensorflow/tensorflow/blob/315e5e897f20d111afd1d65fbeaefeca233c9672/tensorflow/compiler/xla/pjrt/pjrt_client.h#L825-L826

This leads to dynamically shaped tensors being padded with uninitialized data up to the dynamic size. For example:

```
PJRT_DEVICE=CPU XLA_EXPERIMENTAL=nonzero python3
>>> import torch
>>> x = torch.randn((2, 2), device=xm.xla_device())
>>> i = torch.tensor([[False, True], [False, False]], device=xm.xla_device())
>>> x
tensor([[ 1.3344,  1.0886],
        [-1.3762,  1.0890]], device='xla:0')
>>> x[i]
tensor([1.0886e+00, 7.9192e+08, 9.2642e+19, 1.0769e-19], device='xla:0')
```

Unfortunately, the underlying runtime implementations copy the entire buffer up to the bounded size into whichever `xla::Literal` we pass into `ToLiteralSync` ([TFRT:CPU, for example](https://github.com/tensorflow/tensorflow/blob/ef82912d9d452dd24ade8407e24649b9736cba2b/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc#L1085)), which prevents us from just passing in a correctly-sized literal. Instead, we must let PJRT copy the data into a larger literal, and copy the slice of the data we need into a smaller literal.

This change if effectively a no-op for any buffer that already has a static shape. The PJRT C API does not support `logical_on_device_shape` yet, so the behavior is effectively unchanged. I confirmed this change does not affect the performance of our ResNet50 benchmark on TPU v4.

Fixes the remaining C++ tests on CPU.

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -158,8 +158,10 @@ function run_torch_xla_tests() {
     pushd test/cpp
     echo "Running C++ Tests on XRT"
     ./run_tests.sh
-    echo "Running C++ Tests on PJRT"
-    PJRT_DEVICE=CPU ./run_tests.sh
+    if [ -z "$GPU_NUM_DEVICES" ];  then
+      echo "Running C++ Tests on PJRT"
+      PJRT_DEVICE=CPU ./run_tests.sh
+    fi
 
     if ! [ -x "$(command -v nvidia-smi)"  ]
     then
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -156,8 +156,10 @@ function run_torch_xla_tests() {
     fi
 
     pushd test/cpp
-    echo "Running C++ Tests"
+    echo "Running C++ Tests on XRT"
     ./run_tests.sh
+    echo "Running C++ Tests on PJRT"
+    PJRT_DEVICE=CPU ./run_tests.sh
 
     if ! [ -x "$(command -v nvidia-smi)"  ]
     then
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -158,7 +158,8 @@ function run_torch_xla_tests() {
     pushd test/cpp
     echo "Running C++ Tests on XRT"
     ./run_tests.sh
-    if [ -z "$GPU_NUM_DEVICES" ];  then
+    # TODO(wcromar): Enable PJRT C++ tests on GPU
+    if [ -z "$GPU_NUM_DEVICES" ]; then
       echo "Running C++ Tests on PJRT"
       PJRT_DEVICE=CPU ./run_tests.sh
     fi
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -264,12 +264,11 @@ std::vector<xla::Literal> PjRtComputationClient::TransferFromServer(
 
     // TODO(wcromar): Only use logical_on_device_shape when PJRT C API supports
     // it.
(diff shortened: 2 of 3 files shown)


will-cromar:
Adding PJRT C++ tests to the CI in this PR now that they're passing. This won't affect CI latency because the GPU tests are so much longer.

We can start phasing out this and other tests for XRT:CPU in future PRs.

.circleci/common.sh
```diff
@@ -156,8 +156,13 @@ function run_torch_xla_tests() {
     fi
 
     pushd test/cpp
-    echo "Running C++ Tests"
+    echo "Running C++ Tests on XRT"
     ./run_tests.sh
+    # TODO(wcromar): Enable PJRT C++ tests on GPU
+    if [ -z "$GPU_NUM_DEVICES" ]; then
+      echo "Running C++ Tests on PJRT"
+      PJRT_DEVICE=CPU ./run_tests.sh
```
JackCaoG, line 164: wouldn't this double our CI test runtime?

JackCaoG:
Thanks Will, will PJRT eventually fixed this upstream so we don't need to do this hack?

.circleci/common.sh
will-cromar: I'm skipping it on GPU, which is the much longer test suite, so the total time for CI results will not increase. This is also only adding the C++ tests for PJRT, which account for a small portion of the total runtime of the CPU tests. IME the python tests take much longer because there are more of them.

will-cromar:
Yes, part of the work on PJRT needs to happen in libtpu, so it's being tracked internally.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -199,49 +200,49 @@ ComputationClient::DataPtr PjRtComputationClient::ReplicateShardedData(
...
-            absl::StrSplit(shard->device(), ':');
-        XLA_CHECK_EQ(device_spec.size(), 2)
-            << "Invalid device specification: " << shard->device();
-        int device_i = std::stoi(device_spec[1]);
-        arguments_by_device[device_i][0] = shard;
-      }
-      xla::ComputationClient::ExecuteReplicatedOptions execute_options;
-      return ExecuteReplicated(*computations.front(), arguments_by_device,
-                               GetLocalDevices(), execute_options)[0][0];
+    TF_VLOG(1) << "ReplicateShardedData (handle=" << handle->GetOpaqueHandle()
```
JackCaoG, line 203: I didn't find anything changed in this part of the code, did I miss anything?

JackCaoG (approved):
Mostly lGTM, thanks!

.circleci/common.sh
JackCaoG: oh ok I missed that this only runs CPP test.

third_party/xla_client/pjrt_computation_client.cc
will-cromar: `clang-format` reduced the indent here from 4 to 2 spaces to match the rest of the file. Not sure why that didn't get caught in the CI originally

PR #4588, opened 2023-02-08
Split `xla_client` shared library into build targets.

stgpetrovic:
Split the `//third_party/xla_client:libxla_computation_client.so` monolith into small C++ libraries.
Add simple unit tests for `sys_util` and `util` to demonstrate how `bazel coverage` works.

This requires removing the `computation_client` dependency from `metrics_analyzer`; the metrics should be injected from the outside, not fetched from global state (singleton) into the analyzer.

This improves the recompilation time and linking time, especially incrementally.


.bazelrc
```diff
@@ -78,7 +78,7 @@ common --experimental_repo_remote_exec
 try-import %workspace%/.bazelrc.user
 
 # Compile database generation config.
-build:compdb --features=-layering_check
+build:compdb --features=-layering_check --java_runtime_version=remotejdk_11
```
will-cromar, line 81: What's the purpose of this change?

.bazelrc
```diff
@@ -118,7 +118,7 @@ build:linux --config=dynamic_kernels
 # For projects which use TensorFlow as part of a Bazel build process, putting
 # nothing in a bazelrc will default to a monolithic build. The following line
 # opts in to modular op registration support by default.
-build --define framework_shared_object=true
+build --define framework_shared_object=false
```
will-cromar, line 121: Can you update the comment above too since you're changing this value?

will-cromar (commented):
Thanks!

.bazelrc
stgpetrovic: Compiling tests requires JRE/JDK. I'm not sure why but I think that it stems from coverage collection for lcov being in java. Because `compdb` (compilation database) indexes all targets it fails without having the java runtime defined. Same is true for coverage. Essentially, `bazel test` fails without it.

.bazelrc
stgpetrovic: Done.

PR #4599, opened 2023-02-09
[PJRT] Use PJRT by default if XRT is not configured

will-cromar:
If you don't set XRT configuration today, we print a warning that there is no XLA configuration. Instead, let's set the default device to PJRT:CPU or PJRT:TPU, depending on what is available.

Existing workloads that already set up XRT configuration will be unaffected by this change. However, we'll still print a warning for now to let the user know that they're using PJRT instead of XRT.

Example output:

```
# With libtpu installed:
$ python -c "import torch_xla.core.xla_model as xm; print(xm.get_xla_supported_devices('TPU'))"
WARNING:torch_xla:XRT configuration not detected. Defaulting to preview PJRT runtime. To silence this warning and continue using PJRT, explicitly set PJRT_DEVICE to a supported device. To use XRT, set any of the following environment variables: ['XRT_TPU_CONFIG', 'XRT_DEVICE_MAP', 'XRT_WORKERS']
WARNING:torch_xla:For more information about the status of PJRT, see https://github.com/pytorch/xla/blob/master/docs/pjrt.md
WARNING:torch_xla:libtpu.so and TPU device found. Setting PJRT_DEVICE=TPU.
['xla:0', 'xla:1', 'xla:2', 'xla:3']
$ PJRT_DEVICE=TPU python -c "import torch_xla.core.xla_model as xm; print(xm.get_xla_supported_devices('TPU'))"
['xla:0', 'xla:1', 'xla:2', 'xla:3']

# With libtpu uninstalled
$ python -c "import torch_xla.core.xla_model as xm; print(xm.get_xla_supported_devices('CPU'))"
WARNING:torch_xla:XRT configuration not detected. Defaulting to preview PJRT runtime. To silence this warning and continue using PJRT, explicitly set PJRT_DEVICE to a supported device. To use XRT, set any of the following environment variables: ['XRT_TPU_CONFIG', 'XRT_DEVICE_MAP', 'XRT_WORKERS']
WARNING:torch_xla:For more information about the status of PJRT, see https://github.com/pytorch/xla/blob/master/docs/pjrt.md
WARNING:torch_xla:Defaulting to PJRT_DEVICE=CPU
['xla:0']
```

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -11,6 +11,7 @@ XRT_RUN_SERVER_PROCESS = 'torch_xla.core._xrt_run_server'
 XRT_SERVER_REGEX = '^python3 -m {} [0-9]+$'.format(XRT_RUN_SERVER_PROCESS)
 XRT_CONFIG_ENV_VARS = ['XRT_TPU_CONFIG', 'XRT_DEVICE_MAP', 'XRT_WORKERS']
 
+
 def server_is_alive():
   # pgrep returns 0 when at least one running process matches the requested name.
   # Otherwise, the exit code is 1. If pgrep is not availiable in the system, it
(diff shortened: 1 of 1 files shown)


torch_xla/__init__.py
```diff
@@ -10,6 +9,7 @@
 
 XRT_RUN_SERVER_PROCESS = 'torch_xla.core._xrt_run_server'
 XRT_SERVER_REGEX = '^python3 -m {} [0-9]+$'.format(XRT_RUN_SERVER_PROCESS)
+XRT_CONFIG_ENV_VARS = ['XRT_TPU_CONFIG', 'XRT_DEVICE_MAP', 'XRT_WORKERS']
```
JackCaoG, line 12: currently if user sets `GPU_NUM_DEVICES` it will default to xrt:gpu, user have to overwrite that with additional `PJRT_DEVICE=GPU`. I guess we intentionally don't change that behavior there?

JackCaoG (approved):

torch_xla/__init__.py
will-cromar: Right. I left GPU out intentionally because we don't have the same easy way of knowing that the user intended XRT, and there are still some feature gaps on PJRT GPU.

PR #4607, opened 2023-02-10
Update LOCAL_RANK to XLA_LOCAL_RANK

steventk-g:


JackCaoG:
Can we confirm is this will solve the HF issue?

AlexWertheim:
> Can we confirm is this will solve the HF issue?

I can help test this, since the local changes are easy enough to copy over.

test/allreduce_torchrun.py
```diff
@@ -45,7 +46,7 @@ def _mp_fn_xla_backend():
   print(
       'master_port:{}, master_addr:{}, rank:{}, local_rank:{}, size:{}'.format(
           os.environ['MASTER_PORT'], os.environ['MASTER_ADDR'],
-          os.environ['RANK'], os.environ['LOCAL_RANK'],
+          os.environ['RANK'], os.environ[xenv.XLA_LOCAL_RANK],
```
will-cromar, line 49: nit: can we make this `PJRT_LOCAL_RANK` if we only use this in PJRT?

torch_xla/distributed/xrt_init.py
```diff
@@ -178,7 +178,7 @@ def init_xrt_context(master_addr=None, master_port=None, store=None):
 
   # Call this in the actual test case, to work with torch/xla workers
   rank = int(os.environ['RANK'])
-  local_rank = int(os.environ['LOCAL_RANK'])
+  local_rank = int(os.environ[xenv.XLA_LOCAL_RANK])
```
will-cromar, line 181: Will this break `torchrun`? Code that depends on torchrun will look for `LOCAL_RANK`

test/allreduce_torchrun.py
vanbasten23, line 49: nit: how do you feel about `PTXLA_LOCAL_RANK` just to differentiate ourselves from tensorflow xla?

JackCaoG: XLA should be fine, many of our env var started with XLA.

torch_xla/distributed/xrt_init.py
steventk-g: Good point, updated the torchrun tests and variables.

PR #4609, opened 2023-02-10
[Functionalization] Properly skip test_empty_strided

alanwaketan:
Summary:
Functionalization breaks torch.autograd.grad for us. The reason to skip the test is due to the followings:
1. We don't have other test cases or use cases that suggest we officially support torch.autograd.grad.
2. This test case suggested that it's testing test_empty_strided and not torch.autograd.grad.
3. We have cpp test: AtenXlaTensorTest.TestEmptyStrided to test empty_strided and it passes.

Therefore, we believe temporarily breaking torch.autograd.grad should be okay and we should skip the test. pytorch/pytorch#91199 is tracking the progress of the fix.

Test Plan:
CI.

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -599,7 +599,7 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
     xla_result = xla_base[:, torch.empty(0, 6, dtype=torch.int64)]
     self.assertEqual(result, xla_result)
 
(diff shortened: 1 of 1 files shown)


wonjoolee95 (approved):

alanwaketan:
Thanks, Wonjoo.

JackCaoG (approved):

PR #4611, opened 2023-02-10
 POC of dynamicShape+functionalization integration

vanbasten23:


test/test_dynamic_shapes.py
```diff
@@ -175,13 +191,173 @@ def test_expand_symint_correctness(self):
...
+
+    self.assertEqual(len(t2_unsqueeze.size()), 3)
+    self.assertIsInstance(t2_unsqueeze.shape[0], int)
+    self.assertIsInstance(t2_unsqueeze.shape[1], torch.SymInt)
+    self.assertIsInstance(t2_unsqueeze.shape[2], int)
+    self.assertEqual(str(t2_unsqueeze.shape[0]), '1')
+    self.assertEqual(str(t2_unsqueeze.shape[1]), '<=12')
+    self.assertEqual(str(t2_unsqueeze.shape[2]), '2')
+    self.assertEqual(t2_unsqueeze.shape[0], 1)
+    self.assertEqual(t2_unsqueeze.shape[1], 7)
```
JackCaoG, line 216: general comment, this line will trigger a device execution. We should have a test doing something similar to 
```
self.assertEqual(t2_unsqueeze.shape[1], 7)
# check execution counter incremented by 1
xm.mark_step()
# check execution counter incremented by 1
self.assertEqual(t2_unsqueeze.shape[1], 7)
# check execution counter does not increase
```
I am actually not sure if `t2_unsqueeze.shape` will get materalized upon a `mark_step`, that;s one thing we want to get some clarity. I am guessing it won't because this dynamic IR is not being assigned to any tensor.. now I am not sure what's going to happen in above case, which is why we need a test.

JackCaoG (approved):
Fell free to merge once ci is green

test/test_dynamic_shapes.py
vanbasten23: Thanks for the suggestion. I wonder by checking the execution counter, what aspect of the code are we testing?

JackCaoG: we check that accessing the real dynamic size will trigger a device computation. I was hoping to test that after `mark_step` the `tensor.size` will get also materalize, but I am not sure if that;s the case.

vanbasten23: Thanks. let me add the test in a followup pr.

PR #4612, opened 2023-02-11
Skip execution for extract_compiled_graph

JackCaoG:
Currently we use `xla_sync_multi` to force a execution which will warm up the cache for future dynamo runs, this is not ideal because
1. execution can be time consuming
2. `xla_sync_multi` is considered as a `mark_step` like sync and will trigger buffer aliasing. However we actually saved some XLA data when we cache the dynamo graph. Those buffer might be aliased during the `xla_sync_multi` and cause random crash when we actually execute the dynamo graph

confirmed that this change does not regress speed on TPU when running torchbench.

FYI @wconstab @will-cromar

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -58,13 +58,16 @@ class DynamoInferenceBasicTest(unittest.TestCase):
     resnet18.eval()
     xla_resnet18 = torchvision.models.resnet18().to(device)
     xla_resnet18.eval()
+    # materalize the fake data for test purpose
+    xm.mark_step()
+    xm.wait_device_ops()
+    met.clear_all()
     for data, _ in loader:
       output = self.run_model_with_dynamo(xla_resnet18, data)
       torch.allclose(resnet18(data.cpu()), output.cpu())
-    # One graph for initial input data materialization. Another grpah for the
-    # real model code.
-    self.assertEqual(met.metric_data('CompileTime')[0], 2)
-    self.assertEqual(met.metric_data('ExecuteTime')[0], sample_count + 2)
+    # We only expect one graph for the resnet18 inference.
+    self.assertEqual(met.metric_data('CompileTime')[0], 1)
+    self.assertEqual(met.metric_data('ExecuteTime')[0], sample_count)
     self.assertEqual(
         met.metric_data('RunCachedGraphInputData')[0], sample_count)
     self.assertEqual(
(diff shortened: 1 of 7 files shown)


vanbasten23:
Where can I read and learn the `buffer aliasing`?

JackCaoG:
Ah, one thing I realized is we actually need to materializing some of the tensor, at least those one we save for future execution. Let me tweak it a bit.

JackCaoG:
> Where can I read and learn the `buffer aliasing`?

If you search Aliasing in XLA you should see some very very brief doc.. I mostly just read codes.

JackCaoG:
@shunting314 @alanwaketan This pr is ready for review.

shunting314:
>xla_sync_multi is considered as a mark_step like sync and will trigger buffer aliasing. However we actually saved some XLA data when we cache the dynamo graph. Those buffer might be aliased during the xla_sync_multi and cause random crash when we actually execute the dynamo graph

Is it easy to have a unit test to show the buffer aliasing related issue?

JackCaoG:
yea, actually `PJRT=CPU` + the current resnet18 test crashes(currently ci is using XRT, we are migrating test to PJRT runtime then I find this issue). Do you want to add a smaller unit test?

shunting314:
TBH, I don't quite understand what the 'buffer aliasing' issue fixed by this PR is. Or more specifically, how does it cause crashes. A simple unit test should help.

But it's not blocking, I think we can merge this PR first as well.

shunting314 (approved):

alanwaketan (approved):
LGTM.

JackCaoG:
@shunting314 I should be more specified. The buffer aliasing is only enabled during a `mark_step` or equivant(`sync_multi` ) The code is in 
https://github.com/pytorch/xla/blob/574cfdab9bff5c60a1cfb0259d243c3b236a3c1e/torch_xla/csrc/xla_graph_executor.cpp#L1104-L1105

In this pr I set the `config.force_ltc_data` to False during the `warm_up_cache` hence it will skip the aliasing. Regarding what Aliasing does, we have a simple test in 

https://github.com/pytorch/xla/blob/574cfdab9bff5c60a1cfb0259d243c3b236a3c1e/test/test_input_output_aliases.py#L26-L34

You see that we perform two inplace operation on `t1` and `t2`. We will have a hlo similar to 
```
hlo(input1, input2) -> (output1, output2, output3)
```
`input` in this case is `t1` and `t2`, `output` is the `t3` and new `t1` and `t2`. What aliasing does is instead of allocating new buffer for the `output1` and `output2`, it reuses the input buffer since pytorch/xla can tell compiler this is a inplace operation and we don't need `t1` and `t2`'s original buffer after this computation.

The bug we had prior this pr is we actually saved some input for each dynamo graph. However when we do `xla_sync_multi` to warm up the cache, we incorrectly trigger the aliasing. This means the buffer we saved was donated to the output of the `xla_sync_multi` which will crash when we tried to use saved input for the real dynamo run.

PR #4613, opened 2023-02-13
Remove reference to `torch._six` compat layer

XuehaiPan:
Remove reference to `torch._six`, required by pytorch/pytorch#94709.

- pytorch/pytorch#94709

### DIFF
--- a/torch_xla/_patched_functions.py
+++ b/torch_xla/_patched_functions.py
@@ -1,7 +1,7 @@
 import inspect
 import torch
 import torch.nn as nn
-from torch._six import inf
+from torch import inf
 from typing import Iterable, Union, Optional
 
 _tensor_or_tensors = Union[torch.Tensor, Iterable[torch.Tensor]]
--- a/scripts/bench_tensor_io.py
+++ b/scripts/bench_tensor_io.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 
-from __future__ import print_function
-
 import argparse
 import threading
 import torch
--- a/scripts/cond_patch.py
+++ b/scripts/cond_patch.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 
-from __future__ import print_function
-
 import argparse
 import glob
 import os
--- a/scripts/debug_run.py
+++ b/scripts/debug_run.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 
-from __future__ import print_function
-
 import argparse
 import copy
 import getpass
--- a/scripts/dump_stacks.py
+++ b/scripts/dump_stacks.py
@@ -6,8 +6,6 @@
 #  echo 0 > /proc/sys/kernel/yama/ptrace_scope
 #
 
-from __future__ import print_function
-
 import argparse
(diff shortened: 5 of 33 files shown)


torch_xla/_patched_functions.py
```diff
@@ -1,7 +1,7 @@
 import inspect
 import torch
 import torch.nn as nn
-from torch._six import inf
+from torch import inf
```
XuehaiPan, line 4: We can directly import `inf` from `torch` now.

JackCaoG (approved):
Thanks!

JackCaoG:
@wonjoolee95 Can you take of the landing of this one? I think this branch needs to be recreated from the master instead of a fork so we can pin it.

JackCaoG:
@XuehaiPan can this pr merge without the upstream pr? If that's the case(and it won't break us), we can just merge this one instead doing the 2 way pinging which will takes longer to merge.

wonjoolee95:
@XuehaiPan, left a comment in your PyTorch PR regarding the XLA CI pin update, let me know if you have any questions.

Skylion007:
> @XuehaiPan can this pr merge without the upstream pr? If that's the case(and it won't break us), we can just merge this one instead doing the 2 way pinging which will takes longer to merge.

Yeah, it can. The PR is just removing unnecessary code.

wonjoolee95:
Ah I see, then please disregard my comment on the XLA pin. Then we should be good to merge this as its own.

JackCaoG:
@Skylion007 If that's the case @wonjoolee95 can you merge this one directly after all test passed(and pin removed). We can then pin upstream to pytorch/xla master and merge upstream.

wonjoolee95:
@JackCaoG, this PR already doesn't have a torch_pin, so we should be good to merge this? But for some reason, the CPU/GPU tests succeeded but the build got stuck.

JackCaoG:
I think it is because that build was retrigger but stuck.. Let me retrigger CI.

torch_xla/_patched_functions.py
HungNgoCT: Hi
How can we use "from torch._six import string_classes" in torch 2.0?
Thanks in advance if can help.

XuehaiPan: > How can we use `from torch._six import string_classes` in torch 2.0?

Hi @HungNgoCT, you can use `str` directly. Refer to [PyTorch 2.0 Release Note](https://github.com/pytorch/pytorch/releases/tag/v2.0.0):

> ## Releng
> 
> Deprecated private API `torch._six` (https://github.com/pytorch/pytorch/pull/94709)
> Removed the Python 2 and 3 compatibility library `six` and `future` and `torch._six`.
> 
> 2.0
> 
> ```python
> # from torch._six import string_classes
> str
> # from torch._six import int_classes
> int
> # from torch._six import inf, nan
> from torch import inf, nan
> # torch._six.string_classes
> str
> ```

HungNgoCT: Thank you @XuehaiPan . Work now.

PR #4615, opened 2023-02-13
Update TF pin to 2/13

will-cromar:


JackCaoG:
OK This is one of those annoying error that only show up when we build with `XLA_CUDA`. What I usually do is
1. Pull CI docker in TPUVM and build with `XLA_CUDA=1` 
2. find the offending CL and see if I can rebase to a commit before that
3. if not, maybe revert the pr in a patch and fix it internally

This is what I do and why we ended up with the temporary `[ffp_gpu.diff]`

will-cromar:
From the error here:

```
tensorflow/compiler/xla/stream_executor/cuda/cuda_dnn.cc:3498:43: error: ‘CUDNN_TENSOR_REORDERING_INT8x32’ was not declared in this scope
                                         ? CUDNN_TENSOR_REORDERING_INT8x32
                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I'm going to take an educated guess that this is the problem commit: https://github.com/tensorflow/tensorflow/commit/3e2405541f8be5db4bb1594fd3520a369bcfb211

I'll try patching that out while I set up a GPU build environment.

tf_patches/cudnn_int8x32.diff
```diff
@@ -0,0 +1,563 @@
+diff --git a/tensorflow/compiler/xla/debug_options_flags.cc b/tensorflow/compiler/xla/debug_options_flags.cc
```
JackCaoG, line 1: Can we add a TO_BE_REMOVED comment? I am concerned about the increasing number of diffs we carry.. These kind of cuda error usually will be fixed by someone after a couple days, it is fine for us to carry the patch for a short period of time.

will-cromar: Yeah, I'll add in a comment before I submit. Holding off on pushing another commit until the build passes or fails

JackCaoG (approved):

JackCaoG:
CI error is unrelated merging for now.

PR #4617, opened 2023-02-13
[deprecated][internal test][OpenXLA] Migrate xla_client to pull xla from OpenXLA

ManfeiBai:


ManfeiBai:
update code's reference first;

has not updated the build and env set up, will open and let trigger test once the build and env set up has also been updated

miladm (commented):
I see you've removed tf pin in favor of openxla pin. Given we have other dependencies to tf (e.g. XRT), I expect comilation failures that rise from the absence of target source code in tf. I wonder if we should address that first.

ManfeiBai:
> Details

Thanks, Milad, have bring TF back in the newest commit

ManfeiBai:
tried to import openxla as a submodule in https://github.com/pytorch/xla/pull/4625

ManfeiBai:
tracked in a newer PR: https://github.com/pytorch/xla/pull/4878

PR #4619, opened 2023-02-14
Fix HLO dumping

JackCaoG:
After a LTC change, we delay the lock grabbing logic which improved the performance, but it also make HLO dumping triggered via `XLA_SAVE_TENSORS_FMT='hlo' XLA_SAVE_TENSORS_FILE` happen before the last execution finished. It is fine for IR, but for HLO this might trigger an access to placeholder IR. 

verfiied that with this change 
```
XLA_SAVE_TENSORS_FMT='hlo' XLA_SAVE_TENSORS_FILE='tmp/save.hlo' python test/dynamo/test_dynamo.py DynamoTrainingBasicTest.test_resnet18
```

passed

### DIFF
--- a/torch_xla/csrc/debug_util.cpp
+++ b/torch_xla/csrc/debug_util.cpp
@@ -16,6 +16,7 @@
 #include "torch_xla/csrc/ir.h"
 #include "torch_xla/csrc/ir_dump_util.h"
 #include "torch_xla/csrc/ir_util.h"
+#include "torch_xla/csrc/xla_graph_executor.h"
 
 namespace torch_xla {
 namespace {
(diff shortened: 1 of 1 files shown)


will-cromar (approved):

torch_xla/csrc/debug_util.cpp
```diff
@@ -120,6 +121,13 @@ void DebugUtil::SaveTensorsGraphInfo(const char* name,
       "XLA_SAVE_TENSORS_FILE", "", GetCurrentDevice().ordinal());
   if (!save_file.empty()) {
     static std::mutex lock;
+    if (format == DebugUtil::GraphFormat::kHlo && indices->size() > 0) {
+      // Dumping the HLO might access the placeholder data created during
+      // previous execution. We need to wait for last execution to finish before
+      // proceeding.
```
vanbasten23, line 127: I wonder what consequence we get if we don't wait, for example what error would you get when you run the python script

PR #4626, opened 2023-02-15
Make WaitDeviceOps block until device execution finishes

JackCaoG:
In PyTorch/XLA we have a device lock, current thread/process will need device lock if
1. want to transfer data from device
2. execute a program

In the execution, we let the async thread to hold the device lock(until `ExecuteComputation` call finished) while the main process can start tracing for the next iteration. Next process will be blocked when it tries to prepare the input for the current execution because at that point we need last execution to be finished otherwise we might see input to current program as a placeholder.

PJRT go a step further, it let `ExecuteComputation` return before the actual device execution finished. This can further increase the overlap between 2 steps. However this also make `ExecuteTime` metrics inaccurate and `WaitDeviceOps`(which suppose to wait for device exectuion to finish but it actually only waits for the device lock to be release) unreliable. 

For the `ExecuteTime`, we fixed it in https://github.com/pytorch/xla/commit/afe7bb6b6562edcd368055899e17c691824575ac . In this pr I did a similar trick by introducing a new lock that can correclt reflect the `DeviceExectuion` finishes and make `WaitDeviceOps` depend on that.

In a tweaked test, I was able to see
```
2023-02-15 01:49:55.201317: I  401088 /pytorch/xla/torch_xla/csrc/xla_graph_executor.cpp:616] Executing Dynamo IR graph hash 8abb69318a12f48b7f71aab2a071be6f on device CPU:0 ...


wait_device_ops



....



2023-02-15 01:49:55.404479: I  400859 tensorflow/compiler/xla/xla_client/pjrt_computation_client.cc:610] Waiting for device execution forCPU:0 to finish

2023-02-15 01:49:55.404541: I  400859 tensorflow/compiler/xla/xla_client/pjrt_computation_client.cc:613] Waiting for device execution forCPU:0 to finish.. Done

wait_device_ops done
```

TODO:
make sure this does not regress our speed on TPU.

FYI @stgpetrovic @alanwaketan

### DIFF
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -28,42 +28,6 @@ namespace xla {
 
 namespace {
 
-class PJRTLocker {
- public:
-  void Lock() {
-    std::unique_lock<std::mutex> lock(mutex_);
-    cv_.wait(lock, [this] { return !locked_; });
-    locked_ = true;
-  }
-
-  void Unlock() {
-    std::lock_guard<std::mutex> lock(mutex_);
-    locked_ = false;
-    cv_.notify_all();
-  }
-
-  void Barrier() {
-    std::unique_lock<std::mutex> lock(mutex_);
-    cv_.wait(lock, [this] { return !locked_; });
-    cv_.notify_all();
-  }
-
- private:
-  std::mutex mutex_;
-  std::condition_variable cv_;
-  bool locked_ = false;
-  std::exception_ptr exptr_;
-};
-
-std::shared_ptr<PJRTLocker> GetLockerForDevice(std::string device) {
-  static std::unordered_map<std::string, std::shared_ptr<PJRTLocker>>
-      device_lockers;
-  if (device_lockers.find(device) == device_lockers.end()) {
-    device_lockers[device] = std::make_shared<PJRTLocker>();
-  }
-  return device_lockers[device];
-}
-
 std::string PjRtDeviceToString(PjRtDevice* const device) {
   std::string platform =
       absl::AsciiStrToUpper(device->client()->platform_name());
@@ -147,6 +111,7 @@ PjRtComputationClient::PjRtComputationClient() {
   for (auto* device : client_->devices()) {
     std::string device_str = PjRtDeviceToString(device);
     string_to_device_.emplace(device_str, device);
+    device_locks_.emplace(device_str, std::make_unique<std::shared_mutex>());
   }
 }
 
@@ -455,21 +420,15 @@ PjRtComputationClient::ExecuteComputation(
                            returned_future)
           .value();
 
-  std::shared_ptr<PJRTLocker> locker = GetLockerForDevice(device);
-  // This call should not be blocking because there is at most one device
-  // execution happens at the same time. This call is after current thread's
-  // `ExecuteSharded` means execution from last async thread must have finished
-  // and released the locker.
-  locker->Lock();
-
+  // Grab the shared lock and block the `WaitDeviceOps` untill buffer is ready.
+  auto lock = lock_device_shared(device);
   // Signal that `ExecuteSharded` has completed for the ExecuteTime metric.
   // Copies the `timed` shared pointer into the lambda.
-  returned_future->OnReady([timed, locker](Status unused) mutable {
-    timed.reset();
-    // Release the device lock so `WaitDeviceExections` can return.
-    locker->Unlock();
-    TF_VLOG(3) << "returned_future->OnReady finished";
-  });
+  returned_future->OnReady(
+      [timed, lock = std::move(lock)](Status unused) mutable {
+        timed.reset();
+        TF_VLOG(3) << "ExecuteComputation returned_future->OnReady finished";
+      });
 
   std::vector<DataPtr> datas;
   datas.reserve(results.size());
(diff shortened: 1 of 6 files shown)


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -415,9 +455,21 @@ PjRtComputationClient::ExecuteComputation(
                            returned_future)
           .value();
 
+  std::shared_ptr<PJRTLocker> locker = GetLockerForDevice(device);
```
stgpetrovic, line 458: Would it be better to have the PjrtComputationClient to have a field with the map of shared pointers to the lockers, rather than a function-level static variable? Then this would be a member function call, and the storage would be the same as the classes storage stategy?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -547,6 +599,28 @@ xla::PjRtDevice* PjRtComputationClient::StringToPjRtDevice(
   return pjrt_device;
 }
 
+void PjRtComputationClient::WaitDeviceExections(
```
stgpetrovic, line 602: `Executions` typo, same in header.

third_party/xla_client/pjrt_computation_client.h
```diff
@@ -75,6 +75,8 @@ class PjRtComputationClient : public ComputationClient {
 
   void PrepareToExit() override { return; };
 
+  void WaitDeviceExections(const std::vector<std::string>& devices) override;
```
stgpetrovic, line 78: Maybe document that if `devices` is empty, it waits for all devices.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -547,6 +599,28 @@ xla::PjRtDevice* PjRtComputationClient::StringToPjRtDevice(
...
+    const std::vector<std::string>& devices) {
+  std::set<std::string> wait_devices;
```
stgpetrovic, line 604: Do we need an ordered set here, a rb tree?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -28,6 +28,42 @@ namespace xla {
 
 namespace {
 
+class PJRTLocker {
```
will-cromar, line 31: nit: We should follow `PjRt` from TensorFlow

third_party/xla_client/pjrt_computation_client.h
will-cromar: +1 or make this an optional, where a null optional means "wait for all devices"

third_party/xla_client/pjrt_computation_client.cc
will-cromar: Can we even just have a map of device strings to [`shared_mutex`](https://en.cppreference.com/w/cpp/thread/shared_mutex)? Each execution can take a [`shared_lock`](https://en.cppreference.com/w/cpp/thread/shared_lock) and hold it until the callback finishes, and then you can implement `PJRTLocker::Barrier` by blocking on a `unique_lock` on the same `shared_mutex`

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -28,6 +28,42 @@ namespace xla {
...
+
+ private:
+  std::mutex mutex_;
+  std::condition_variable cv_;
+  bool locked_ = false;
+  std::exception_ptr exptr_;
+};
+
+std::shared_ptr<PJRTLocker> GetLockerForDevice(std::string device) {
+  static std::unordered_map<std::string, std::shared_ptr<PJRTLocker>>
```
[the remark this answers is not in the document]
JackCaoG: lol I thought I did, let me add that.

third_party/xla_client/pjrt_computation_client.cc
JackCaoG: yea unordered_set works better.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -28,6 +28,42 @@ namespace xla {
...
+  }
+
+ private:
+  std::mutex mutex_;
+  std::condition_variable cv_;
+  bool locked_ = false;
+  std::exception_ptr exptr_;
+};
+
+std::shared_ptr<PJRTLocker> GetLockerForDevice(std::string device) {
```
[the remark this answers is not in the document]
JackCaoG: good catch, will do

third_party/xla_client/pjrt_computation_client.h
JackCaoG: I will leave a comment, this is align with current `waitdeviceops` behaviors, check https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_graph_executor.cpp#L373 and https://github.com/pytorch/xla/blob/master/torch_xla/core/xla_model.py#L965

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -415,9 +420,15 @@ PjRtComputationClient::ExecuteComputation(
                            returned_future)
           .value();
 
+  // Grab the shared lock and block the `WaitDeviceOps` untill buffer is ready.
```
will-cromar, line 423: typo: `until`

third_party/xla_client/pjrt_computation_client.cc
will-cromar: What is the set for? Would it be better to just directly iterate through `devices` if not empty and `GetLocalDevices` otherwise?

will-cromar (approved):
LGTM. Thanks!

third_party/xla_client/pjrt_computation_client.cc
JackCaoG: it is in case people passed in duplicated device names, I should make it a unordered_set.

JackCaoG:
I was able to confirm that on PJRT + v4-8 there is no speed regression.

JackCaoG:
The test is too unstable, it passed on my machine but not the CI. I will remove the check of `assertNotIn` to make it more stable.

JackCaoG:
@Liyang90 FYI, without this pr, the benchmark for dynamo can be a bit misleading when using PJRT. I will try to backport this to the r2.0 branch and we can rerun the benchmark after release is more stable.

PR #4629, opened 2023-02-15
Support dynamism on view_copy_symint

vanbasten23:
This PR intend to support dynamism on view_copy_symint.
To make it work, this PR also implements:
- XLASymNodeImpl:mod
- SizeMod node.

With the change in this PR, the test `test_backward_pass_with_dynamic_input` passes.

### DIFF
--- a/test/pjrt/test_mesh_service.py
+++ b/test/pjrt/test_mesh_service.py
@@ -12,7 +12,6 @@ class PjRtMeshServiceTest(parameterized.TestCase):
     payload = b'message %d' % xm.get_ordinal()
     return xm.rendezvous("test rendezvous", payload)
 
-  @unittest.skip("fails with functionalization")
   def test_rendezvous_static_size(self):
     results = pjrt._run_multiprocess(self._rendezvous_static_size)
 
@@ -24,7 +23,6 @@ class PjRtMeshServiceTest(parameterized.TestCase):
     payload = b'message' * xm.get_ordinal()
     return xm.rendezvous("test rendezvous", payload)
 
-  @unittest.skip("fails with functionalization")
   def test_rendezvous_dynamic_size(self):
     results = pjrt._run_multiprocess(self._rendezvous_dynamic_size)
 
@@ -36,14 +34,12 @@ class PjRtMeshServiceTest(parameterized.TestCase):
     replicas = list(range(pjrt.global_device_count()))
     return xm.rendezvous("test rendezvous", b'message', replicas)
 
-  @unittest.skip("fails with functionalization")
   def test_rendezvous_replica_groups(self):
     results = pjrt._run_multiprocess(self._rendezvous_replica_groups)
 
     expected = [b'message'] * len(results)
     self.assertDictEqual(results, {r: expected for r in results})
 
-  @unittest.skip("fails with functionalization")
   def test_rendezvous_empty_payload(self):
     test_fn = functools.partial(xm.rendezvous, 'test rendezvous', b'')
     results = pjrt._run_multiprocess(test_fn)
@@ -61,7 +57,6 @@ class PjRtMeshServiceTest(parameterized.TestCase):
   def _mesh_reduce():
     return xm.mesh_reduce('test mesh reduce', xm.get_ordinal(), sum)
 
-  @unittest.skip("fails with functionalization")
   def test_mesh_reduce(self):
     results = pjrt._run_multiprocess(self._mesh_reduce)
     values = list(results.values())
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -253,6 +253,32 @@ class TestDynamicShapes(test_utils.XlaTestCase):
     t2 = t1.fill_(10)
     self.assertIsInstance(t2.shape[0], torch.SymInt)
 
+  # xw32 TODO: implement this.
+  def test_sizeMod(self):
+    met.clear_all()
+
+    size1 = 5
+    size2 = 2
+    t1 = torch.zeros([size1, size2], device=dev)
+    t1[3][0] = 1
+    # t2 has size [<=10, 2] with real size [1, 2]
+    t2 = torch.nonzero(t1)
+    # Create a SizeMod IR node.
+    # t2.shape[1] generates a SizeConstant node.
+    dyn_size = t2.shape[0] % t2.shape[1]
+    self.assertGreater(met.counter_value("xla::size_mod"), 0)
+    # Exercises SizeMod::getDynamicValue.
+    dynamic_size = int(dyn_size)
+    self.assertEqual(dynamic_size, 1)
+    self.assertEqual(str(dyn_size), '<=0')
+
+    # t3 has size [<=10, 2] with real size [1, 2]
+    t3 = torch.nonzero(t1)
+    dyn_size = t2.shape[0] % t3.shape[0]
+    dynamic_size = int(dyn_size)
+    self.assertEqual(dynamic_size, 0)
+    self.assertEqual(str(dyn_size), '<=0')
+
   def test_sizeGe(self):
     met.clear_all()
 
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -252,32 +252,13 @@ class TestDynamicShapes(test_utils.XlaTestCase):
     self.assertEqual(t8.cpu(), t8_aten.cpu())
 
   def test_view_copy_symint_with_dyn_input_static_input_shape(self):
-    # If the input tensor and shape are “statically” incompatible, a compilation error is raised.
-    t1 = torch.tensor([1, 0, 3, 5, 0, 6], device=dev)
-    # t2.shape=torch.Size([<=6, 1]) with real size [4, 1]
-    # t2 = [[0], [2], [3], [5]]
+    # If the input tensor is dynamic and input shape is static,
+    # it should fail because we will not likely have this case
+    # in reality.
+    t1 = torch.tensor([1, 1, 3, 5, 1, 6], device=dev)
+    # t2.shape=torch.Size([<=6, 1]) with real size [6, 1]
     t2 = torch.nonzero(t1)
-    # static 6 is incompatible with 7
-    self.assertRaises(RuntimeError, lambda: t2.view(7))
-
-    # If their “dynamic” values are incompatible, a RuntimeError is raised.
-    # statically compatible.
-    # It will fail in pytorch layer.
-    self.assertRaises(RuntimeError, lambda: t2.view(6))
-
-    # verify if dynamism is propagated correctly.
-    t5 = torch.tensor([1, 1, 3, 5, 1, 6], device=dev)
-    t6 = torch.nonzero(t5)
-    t7 = t6.view(6)
-    self.assertIsInstance(t7.shape[0], torch.SymInt)
-    self.assertEqual(str(t7.shape[0]), '<=6')
-    self.assertEqual(t7.shape[0], 6)
-
-    # verify correctness.
-    t5_aten = torch.tensor([1, 1, 3, 5, 1, 6])
-    t6_aten = torch.nonzero(t5_aten)
-    t7_aten = t6_aten.view(6)
-    self.assertEqual(t7.cpu(), t7_aten.cpu())
+    self.assertRaises(RuntimeError, lambda: t2.view(2, 3))
 
   def test_view_copy_symint_with_dyn_input_dyn_input_shape(self):
     # If the input tensor and shape are “statically” incompatible, a compilation error is raised.
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -228,11 +228,13 @@ class TestDynamicShapes(test_utils.XlaTestCase):
     print('xw32 running view(4)')
     # doing t3.view(5) would fail in python level with error
     # https://paste.googleplex.com/6642485810954240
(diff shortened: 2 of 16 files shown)


test/test_dynamic_shapes.py
```diff
@@ -222,6 +222,162 @@ def test_unsqueeze_copy_dynamism(self):
...
+    t4 = torch.nonzero(t3)
+    self.assertRaises(RuntimeError, lambda: t2.view(t4.shape[0]))
+
+    # If their “dynamic” values are incompatible, a RuntimeError is raised.
+    t5 = torch.tensor([1, 2, 3, 4, 5, 6, 0], device=dev)
+    # t6.shape=torch.Size([<=7, 1]) with real size [6, 1]
+    t6 = torch.nonzero(t5)
+    # statically compatible but dynamically incompatible.
+    # It will fail in pytorch layer.
+    self.assertRaises(RuntimeError, lambda: t6.view(t4.shape[0]))
```
vanbasten23, line 300: comment to me: it fails in pytorch layer: https://paste.googleplex.com/5104119952965632

test/test_dynamic_shapes.py
```diff
@@ -222,6 +222,162 @@ def test_unsqueeze_copy_dynamism(self):
...
+    # If the input tensor and shape are “statically” incompatible, a compilation error is raised.
+    t1 = torch.tensor([1, 0, 3, 5, 0, 6], device=dev)
+    # t2.shape=torch.Size([<=6, 1]) with real size [4, 1]
+    # t2 = [[0], [2], [3], [5]]
+    t2 = torch.nonzero(t1)
+    # static 6 is incompatible with 7
+    self.assertRaises(RuntimeError, lambda: t2.view(7))
+
+    # If their “dynamic” values are incompatible, a RuntimeError is raised.
+    t4 = torch.randint(10, (2, 3), device=dev)
```
vanbasten23, line 264: delete unused t4

JackCaoG:
Thanks, I will try to take a look today

test/test_dynamic_shapes.py
```diff
@@ -222,6 +222,161 @@ def test_unsqueeze_copy_dynamism(self):
...
+    t8_aten = torch.nonzero(t7_aten)
+    # t8_aten.size=[5, 1]
+    t3_aten = torch.tensor([1, 0, 3, 5, 0, 6, 7])
+    t4_aten = torch.nonzero(t3_aten)
+    # t4_aten.size=[5, 1]
+    t9_aten = t8_aten.view(t4_aten.shape[0])
+    self.assertEqual(t9.cpu(), t9_aten.cpu())
+
+  def test_view_copy_symint_negative_input_shape(self):
+    t5 = torch.tensor([1, 1, 3, 5, 1, 6], device=dev)
```
JackCaoG, line 321: it is a bit weird that tensor started with t5 lol, can we update that to t1?

test/test_dynamic_shapes.py
```diff
@@ -222,6 +222,161 @@ def test_unsqueeze_copy_dynamism(self):
...
+    self.assertEqual(t7.shape[1], 3)
+
+    # verify correctness.
+    t5_aten = torch.tensor([1, 1, 3, 5, 1, 6])
+    t6_aten = torch.nonzero(t5_aten)
+    t7_aten = t6_aten.view(2, -1)
+    self.assertEqual(t7.cpu(), t7_aten.cpu())
+
+  def test_GetDynamicReshape(self):
+    t5 = torch.tensor([1, 2, 3, 4, 5, 6, 7, 8], device=dev)
```
JackCaoG, line 340: ditto

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -3161,11 +3161,17 @@ std::tuple<at::Tensor, at::Tensor> XLANativeFunctions::var_mean(
...
-      bridge::GetXlaTensor(self), XlaHelpers::I64List(size)));
+  c10::optional<at::IntArrayRef> int_sizes =
+      c10::asIntArrayRefSlowOpt(sym_size);
+  bool all_dims_static = int_sizes.has_value();
+  if (all_dims_static) {
+    return bridge::AtenFromXlaTensor(tensor_methods::view(
+        bridge::GetXlaTensor(self), XlaHelpers::I64List(int_sizes.value())));
+  }
+
+  return bridge::AtenFromXlaTensor(
```
JackCaoG, line 3173: nit, can we put this in a else branch so it is more clear?

torch_xla/csrc/helpers.cpp
```diff
@@ -254,27 +254,23 @@ xla::XlaOp XlaHelpers::ReshapeToRank(xla::XlaOp input, int64_t expected_rank,
...
-    for (int64_t i = 0; i <= input_dynamic_dimension; ++i) {
-      size_at_dyndim *= input_shape.dimensions(i);
+    int64_t size_prod_until_dyndim = 1;
+    for (int64_t i = 0; i <= input_dyndim_idx; ++i) {
+      size_prod_until_dyndim *= input_shape.dimensions(i);
     }
     int64_t dynamic_dimension = -1;
     int64_t out_size = 1;
     for (int64_t i = 0; i < output_sizes.size(); ++i) {
-      XLA_CHECK_LE(out_size, size_at_dyndim / input_shape.dimensions(
```
JackCaoG, line 272: why is this check removed?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2639,6 +2641,21 @@ XLATensorPtr view(const XLATensorPtr& input,
   return input->CreateViewTensor(std::move(view_info));
```
JackCaoG, line 2641: @wonjoolee95 @alanwaketan I vaguely remember we reuse a lot of old View code even during functionization pass, do we have a plan to remove them completely? For example ideally we shouldn't call `CreateViewTensor` and drop `viewinfo` class.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2639,6 +2641,21 @@ XLATensorPtr view(const XLATensorPtr& input,
...
+  xla::util::MaybeRef<xla::Shape> input_shape = input->shape();
+  SymIntElements size_elements(sym_size);
+  std::vector<int64_t> complete_dimensions = GetCompleteShape(
+      size_elements.GetUpperBounds(), input_shape.get().dimensions());
+  xla::Shape result_shape = xla::ShapeUtil::MakeShape(
+      input_shape.get().element_type(), complete_dimensions,
+      size_elements.GetDynamicDims());
+  ViewInfo view_info(ViewInfo::Type::kReshape, std::move(result_shape),
+                     input_shape);
+  return input->CreateViewTensor(std::move(view_info));
```
JackCaoG, line 2656: hmm.. so reshape handles dynamism out of the box? I guess that make sense.

test/test_dynamic_shapes.py
vanbasten23: lol, done

test/test_dynamic_shapes.py
vanbasten23: done

torch_xla/csrc/aten_xla_type.cpp
vanbasten23: done

torch_xla/csrc/helpers.cpp
vanbasten23: It's the trickest part of the PR. Let's use the test `test_view_copy_symint_negative_input_shape` for example. In the test, we do `t3=t2.view(2,3)` where t2.shape=torch.Size([<=6, 1]) with real size [6, 1]. To my understanding, the point of this function `XlaHelpers::GetDynamicReshapeInfo` is to find out which output tensor dimension is dynamic. Theoretically, we should be able to reshape a [<=6,1] tensor to a [2, <=3] tensor but in reality this `XLA_CHECK_LE` would fail for no good reason: when i=1, out_size=2, size_at_dyndim/input_shape.dimensions(input_dynamic_dimension)=6/6=1

Addditionally, after I comment out this check, the CI is green so it seems to me we're not gaining much from the check. And the original PR https://github.com/pytorch/xla/pull/1496 doesn't have tests for this check..

torch_xla/csrc/tensor_methods.cpp
miladm: Even so, I'd wonder if we need to verify this with a unit test to confirm. Ideally, all ops we touch should have a 'simple' unit test to be whitelisted for dynamism - we can even have a whitelisting `yaml` file that tracks support level. For ops we do proper dynamism sypport, we do more extensive testing. With this approach, future regressions will less likely become a surprise to us. Thoughts?

torch_xla/csrc/tensor_methods.h
```diff
@@ -883,6 +883,8 @@ std::tuple<XLATensorPtr, XLATensorPtr> var_mean(const XLATensorPtr& input,
 // Like reshape, but it returns a view into the original tensor.
 XLATensorPtr view(const XLATensorPtr& input,
```
miladm, line 884: Do we keep both `view` and `view_symint` going forward?

torch_xla/csrc/tensor_methods.cpp
vanbasten23: +1 on adding unit tests. For this PR, I have verified this path with 4 unit tests in pytorch/xla/test/test_dynamic_shapes.py in this PR. Let me know if I have missed any scenarios. Also, for all ops that support dynamism, I have added enough unit tests (the remaining untested code are the Size nodes which I will add tests this week.)

torch_xla/csrc/tensor_methods.h
vanbasten23: Yeah.

For view op, I've modified my code to only use `view_symint`.

 However, there're 3 other references in tensor_ops.cpp that uses this function which belongs to other ops. This PR is only intended to fix the `view` ops so I'll leave it there.

torch_xla/csrc/helpers.cpp
vanbasten23: Now that the `view` op doesn't depends on this function, I'd still think we can remove this check, for the above reason.

vanbasten23:
fyi for reviewers, the build error is 
```
cat: xla/third_party/tensorflow/third_party/llvm/workspace.bzl: No such file or directory
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 
mirror.tensorflow.org/github.com/llvm/ is not available.
If the issue persists, plesase report at github.com/pytorch/xla
```
I probably need to do a rebase.

torch_xla/csrc/helpers.cpp
JackCaoG: is this true through? I don't think we can always reshape [<=6,1] tensor to a [2, <=3], what if it is a `[3,1]`, there is no way for us to reshape it [2, x].

Reshaping dynamic tensor is tricky, I wonder if XLA has any guideline on this.

vanbasten23: ok, I added the check back, since this PR doesn't necessarily need to use this function. I can deal with it later if it becomes a blocker.

Though I'd like to keep the renaming change as it's really confusing to read this part of code.

I also did some [experiment](https://github.com/pytorch/xla/issues/4686). For `I don't think we can always reshape [<=6,1] tensor to a [2, <=3]`, see case 6. It would succeed without the check https://github.com/pytorch/xla/blob/58894f0e0cc89c2a6d899a22e5619cac00cc8914/torch_xla/csrc/helpers.cpp#L272-L275. Wrt `what if it is a [3,1], there is no way for us to reshape it [2, x]`, see case 7. So I could be wrong but it looks to me the intention of the original author is: if we try to reshape/view a [<=6, 1] tensor to a shape [2, 3], the expected resulting shape would be a dynamic one [2, <=3]. But it makes lots of thing more complicated. So I agree with you that we can make it a error case when the input is dynamic and input_shape is static.

test/pjrt/test_mesh_service.py
```diff
@@ -12,7 +12,6 @@ def _rendezvous_static_size():
     payload = b'message %d' % xm.get_ordinal()
     return xm.rendezvous("test rendezvous", payload)
 
-  @unittest.skip("fails with functionalization")
```
vanbasten23, line 15: to reviewers: the change in this file is irrelevant to this PR. It's to make the CI green. We need this change probably due to I made a mistake when rebasing origin/funtionalization_dynamic_shape onto origin/functionalization. The change made here is consistent to what origin/functionalization has.

JackCaoG (approved):

torch_xla/csrc/tensor_methods.cpp
alanwaketan: Yea, there is a task to replace those hacks with more formal lowering: https://docs.google.com/document/d/1d1u8oLrL7iZz55NbCfQSYzwLP4M60kb7igvdBVE_X7s/edit?resourcekey=0-W3QvBNHNPUeX84-TkRJjlA#bookmark=id.tux3bndmql98

PR #4630, opened 2023-02-15
[PJRT] Defer default device selection until we need to know the runtime

will-cromar:
#4599 breaks `xla_dist` by selecting PJRT as the default runtime upon importing PT/XLA, because `xla_dist` is the only situation where you don't need to set an XRT env var manually ahead of time. For example:

```
# python -m torch_xla.distributed.xla_dist --help
WARNING:torch_xla:XRT configuration not detected. Defaulting to preview PJRT runtime. To silence this warning and continue using PJRT, explicitly set PJRT_DEVICE to a supported device. To use XRT, set any of the following environment variables: ['XRT_TPU_CONFIG', 'XRT_DEVICE_MAP', 'XRT_WORKERS', 'GPU_NUM_DEVICES']
```

Instead, wait until we need to know the runtime to select a default device. This behavior can be disabled with `PJRT_SELECT_DEFAULT_DEVICE=0`.

Also warn the user when calling `xla_dist` with a `PJRT_DEVICE` selected.

<details>
<summary>Example output</summary>

Default with `libtpu` installed:

```
# python
Python 3.8.13 (default, Jun 23 2022, 11:46:15) 
[GCC 8.3.0] on linux
>>> xm.xla_device()
2023-02-15 22:17:31.669229: W tensorflow/compiler/xla/xla_client/computation_client.cc:373] XRT configuration not detected. Defaulting to preview PJRT runtime. To silence this warning and continue using PJRT, explicitly set PJRT_DEVICE to a supported device or configure XRT.
2023-02-15 22:17:31.669284: W tensorflow/compiler/xla/xla_client/computation_client.cc:377] For more information about the status of PJRT, see https://github.com/pytorch/xla/blob/master/docs/pjrt.md
2023-02-15 22:17:31.669302: W tensorflow/compiler/xla/xla_client/computation_client.cc:382] libtpu.so and TPU device found. Setting PJRT_DEVICE=TPU.
device(type='xla', index=0)
```

Default with no `libtpu`:

```
# python
Python 3.8.13 (default, Jun 23 2022, 11:46:15) 
[GCC 8.3.0] on linux
>>> xm.xla_device()
2023-02-15 22:18:42.051619: W tensorflow/compiler/xla/xla_client/computation_client.cc:373] XRT configuration not detected. Defaulting to preview PJRT runtime. To silence this warning and continue using PJRT, explicitly set PJRT_DEVICE to a supported device or configure XRT.
2023-02-15 22:18:42.051699: W tensorflow/compiler/xla/xla_client/computation_client.cc:377] For more information about the status of PJRT, see https://github.com/pytorch/xla/blob/master/docs/pjrt.md
2023-02-15 22:18:42.051712: W tensorflow/compiler/xla/xla_client/computation_client.cc:386] Defaulting to PJRT_DEVICE=CPU
device(type='xla', index=0)
>>> 
```

`PJRT_DEVICE=TPU`:

```
# PJRT_DEVICE=TPU python
Python 3.8.13 (default, Jun 23 2022, 11:46:15) 
[GCC 8.3.0] on linux
>>> xm.xla_device()
device(type='xla', index=0)
>>>
```

`xla_dist`:

```
# python -m torch_xla.distributed.xla_dist --help
usage: xla_dist.py [-h] --tpu TPU [--vm VM] [--docker-container DOCKER_CONTAINER]
                   [--docker-image DOCKER_IMAGE] [--docker-run-flag DOCKER_RUN_FLAG]
                   [--conda-env CONDA_ENV] [--env ENV] [--restart-tpuvm-pod-server]
                   [--tpuvm-server-port TPUVM_SERVER_PORT]
                   positional [positional ...]
```

`xla_dist` with PjRt:

```
# PJRT_DEVICE=CPU python -m torch_xla.distributed.xla_dist --help
WARNING:root:PJRT runtime detected. `xla_dist` is NOT compatible with PJRT, and you may run into unexpected errors. Unset $PJRT_DEVICE to silence this warning.
usage: xla_dist.py [-h] --tpu TPU [--vm VM] [--docker-container DOCKER_CONTAINER]
                   [--docker-image DOCKER_IMAGE] [--docker-run-flag DOCKER_RUN_FLAG]
                   [--conda-env CONDA_ENV] [--env ENV] [--restart-tpuvm-pod-server]
                   [--tpuvm-server-port TPUVM_SERVER_PORT]
                   positional [positional ...]
```
</details>

TODO:

- [x] Check `xla_dist` on a pod
- [x] Add unit tests


will-cromar:
Confirmed that moving the default device selection does fix `xla_dist`. However, this change won't work well as written because of a long-unfixed issue in Python where `os.environ` ignores updates not made directly to `os.environ`. See this thread https://stackoverflow.com/a/33642551

Basically, there's no way for us to update the environment from within C++ and have that update reflected in Python, unless we replace `os.environ` references, which is just asking for trouble. Even if we fix this now, someone in the future will forget they can't rely on `os.environ`. 

One option to drop the `PJRT_DEVICE` env var except for initialization and store the state somewhere else, which would require cleaning up any references to it in our code. That becomes tricky because we rely on the `PJRT_DEVICE` setting to get propagated from a parent process to its children, which wouldn't be true for e.g. a global variable.

The other option is to set the variable in Python. I'm reasonably sure that the first call that the first call where we have to know whether or not we're using PJRT will go through `pjrt.using_pjrt` (e.g. `xm.xla_device`, `xmp.spawn`), so we could set it exclusively there, and never try to infer the default device when initializing the client. If there's an edge case I'm not thinking of here where the user might initialize the client before going through `pjrt.using_pjrt`, they'll get the current complaint about missing XLA configuration. If we try to infer the device when initializing the client, then we won't be able to trivially communicate that back to `os.environ`, and any code that relies on `PJRT_DEVICE` will break.

test/pjrt/test_experimental_pjrt.py
```diff
@@ -52,6 +50,49 @@ def test_xla_device_error(self):
...
+      'XRT_TPU_CONFIG': 'localservice;0;localhost:51011'
+  }, False), ('pjrt_tpu_precedence', {
+      'PJRT_DEVICE': 'TPU',
+      'XRT_TPU_CONFIG': 'localservice;0;localhost:51011'
+  }, True), ('xrt_gpu', {
+      'GPU_NUM_DEVICES': '4'
+  }, False), ('pjrt_gpu', {
+      'PJRT_DEVICE': 'GPU',
+      'GPU_NUM_DEVICES': '4'
+  }, True))
```
JackCaoG, line 78: should we add tests for xla_dist case, for example `XRT_LOCAL_WORKER` in https://github.com/pytorch/xla/blob/master/third_party/xla_client/computation_client.cc#L196. It is a bit tricky since even if we set all env vars, it will fail because there is no XRT server to distribue the TPU topology in this case.. 

Maybe we should just have a test with mesh server setup and make sure it doesn't got overwrite to PJRT_DEVICE. I think we can configure that mesh timeout to some shorter time, defualt is 300 s.

JackCaoG: or we can just set any `XRT_` and expect the test to fail instead of getting a PJRT device.

JackCaoG (approved):
mostly lgtm, we need to backport this to 2.0 release branch.

test/pjrt/test_experimental_pjrt.py
will-cromar: Good idea. I don't know why I forgot to add a unit test for the bug I'm fixing.

Added an an `xla_dist` test case that just sets `XRT_LOCAL_WORKER`. Since we don't actually initialize the client in this unit test, it works without having to set up a mesh server.

will-cromar:
All tests passed on 62c5ca611d417819885c8e54585ab51f72cb594c. Merging now since the last commit is just formatting with yapf.

PR #4636, opened 2023-02-16
Bazel

stgpetrovic:
Migration of libxla_computation_client to bazel.

* Replace tensorflow with a bazel external repository

* Basic migration to bazel for xla_client.

* Revert to blob

* Add vscode config.

* Update newlines

* Merge with pjrt client test build changes.

* Migrate tests to new build

* Format test and plugin

* Order imports

* Conditionally apply tf patches; apply pt patches always.

* Format python

* configure formatters

* Mirror TF pin update an fixes in bazel.

* Support local and sandboxed build based on flags

* Add cloud cache URLs for llvm.

* Merge with upstream

* Update TF pin

* Fix patching regression


JackCaoG:
@ymwangg @cicirori @aws-rhsoln FYI, the most notable change is you won't see `third_party/tensorflow/` anymore but it is part of our dependency.

PR #4641, opened 2023-02-16
Re-enable sandbox builds in Docker

will-cromar:
Will kick off TPU CI on this branch to confirm the fix

### DIFF
--- a/docker/experimental/Dockerfile
+++ b/docker/experimental/Dockerfile
@@ -74,7 +74,7 @@ COPY third_party/ third_party/
 COPY build_torch_xla_libs.sh .
 
 # TODO: Remove this when it's not required anymore
-ENV XLA_SANDBOX_BUILD=0
+ENV XLA_SANDBOX_BUILD=1
 ENV XLA_SANDBOX_BASE "/dev/shm"
 
 ARG tpuvm
--- a/docker/experimental/Dockerfile
+++ b/docker/experimental/Dockerfile
@@ -77,18 +77,16 @@ COPY build_torch_xla_libs.sh .
 ENV XLA_SANDBOX_BUILD=1
 ENV XLA_SANDBOX_BASE "/dev/shm"
 
-ARG tpuvm
-ARG cuda
-ARG tf_cuda_compute_capabilities
-ARG bazel_jobs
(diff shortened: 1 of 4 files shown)


will-cromar:
The TPU CI and my local build started failing after #4636. Tested these changes locally.

If others are seeing the same issue, we can change the default in `build_torch_xla_libs.sh`

cc @stgpetrovic

docker/experimental/Dockerfile
```diff
@@ -74,7 +74,7 @@ COPY third_party/ third_party/
 COPY build_torch_xla_libs.sh .
 
 # TODO: Remove this when it's not required anymore
-ENV XLA_SANDBOX_BUILD=0
+ENV XLA_SANDBOX_BUILD=1
```
JackCaoG, line 77: Do we have plan to remove sandbox in the future?

JackCaoG (approved):

docker/experimental/Dockerfile
will-cromar: I'll defer to @stgpetrovic on that. My understanding is that we need sandboxing to take advantage of remote build execution.

will-cromar:
I'm still getting this error: `ERROR: Config value 'tpu' is not defined in any .rc file`. Any ideas?

stgpetrovic:
The docker file is not copying .bazel{rc,version} nor WORKSPACE so it has
no config.

see the fix in the bazel branch i pushed, i had it running the cloudbuild
test manually before repushing to see if there are additional issues.

cheers,

will-cromar:
Now I'm getting this error after the bazel build starts: `LLVM ERROR: IO failure on output stream: No space left on device`

stgpetrovic:
This is /dev/shm, we can remove that part so it builds slowly on disk, or
make the image have larger shm, ive seen that sonewhere in cloudbuild
configuring the shm size. You can revert and I can debug it all until it
passes manually.

will-cromar:
I'll try moving the TPU CI build to the same pool that we use for the nightly builds so that it will have the same shm size and bazel jobs.

steventk-g (approved):
Thanks Will!

docker/experimental/Dockerfile
stgpetrovic: The sandbox is required by some bug; bazel complains about some llvm mlir deps, that's why the sandbox was introduced before this migration. Once we're running remote, there's no need for sandbox,

stgpetrovic: OK this makes sense: this build was using 8 jobs and was running locally before. We can either leave it at that, or make the RAM a bit bigger and make it much faster.

For now I'd like to fix this so I'll make it run as it did before; with the new .bazelrc we default to shm/sandboxed.

will-cromar:
See #4649

PR #4644, opened 2023-02-16
Add dynamic support for topk

ymwangg:
This PR adds dynamic shape support for topk.


test/test_dynamic_shapes.py
```diff
@@ -178,6 +178,37 @@ def test_expand_symint_correctness(self):
...
+    self.assertIsInstance(values.shape[0], torch.SymInt)
+    self.assertIsInstance(indices.shape[0], torch.SymInt)
+    self.assertEqual(str(t2.shape[0]), '<=10')
+    self.assertEqual(str(values.shape[0]), '<=10')
+    self.assertEqual(str(indices.shape[0]), '<=10')
+    self.assertEqual(t3.shape[1], 2)
+    self.assertEqual(values.shape[1], 2)
+    self.assertEqual(indices.shape[1], 2)
+
+    def test_fn(*tensors):
```
JackCaoG, line 199: this and above test are independent I think? Maybe split them into a `test_topk_symint_trival` and `test_topk_symint `

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2841,13 +2841,22 @@ at::Tensor XLANativeFunctions::threshold_backward(const at::Tensor& grad_output,
...
 }
 
-std::tuple<at::Tensor, at::Tensor> XLANativeFunctions::topk(
-    const at::Tensor& self, int64_t k, int64_t dim, bool largest, bool sorted) {
-  TORCH_LAZY_FN_COUNTER("xla::");
-  auto results = tensor_methods::topk(bridge::GetXlaTensor(self), k, dim,
-                                      largest, sorted, /*stable=*/false);
-  return std::make_tuple(bridge::AtenFromXlaTensor(std::get<0>(results)),
-                         bridge::AtenFromXlaTensor(std::get<1>(results)));
+std::tuple<at::Tensor, at::Tensor> XLANativeFunctions::topk_symint(
```
JackCaoG, line 2844: hmm, will static topk got dispatched to this function too? Seems like it since our topk test doesn't break. I guess upstream just check if downstream implemented `topk` or `topk_symint` ?

torch_xla/csrc/ops/topk.cpp
```diff
@@ -18,6 +18,23 @@ xla::Shape NodeOutputShape(const torch::lazy::Value& input, int64_t k,
   return InferOutputShape({GetXlaShape(input)}, lower_for_shape_fn);
 }
 
+xla::Shape NodeOutputShapeSymInt(const torch::lazy::Value& input,
+                                 int64_t k_upper_bound, int64_t dim,
```
JackCaoG, line 22: Can we create a topk_symint.h/cpp instead? it is easier to have a op->file name mapping

torch_xla/csrc/ops/topk.cpp
```diff
@@ -18,6 +18,23 @@ xla::Shape NodeOutputShape(const torch::lazy::Value& input, int64_t k,
...
+                                 bool largest, bool sorted, bool stable) {
+  xla::Shape input_shape = GetXlaShape(input);
+  std::vector<int64_t> dimensions(input_shape.dimensions().begin(),
+                                  input_shape.dimensions().end());
+  XLA_CHECK_LT(dim, input_shape.rank());
+  dimensions[dim] = k_upper_bound;
+  xla::Shape values_shape =
+      xla::ShapeUtil::MakeShape(input_shape.element_type(), dimensions);
+  xla::Shape indices_shape =
+      xla::ShapeUtil::MakeShape(xla::PrimitiveType::S64, dimensions);
```
JackCaoG, line 32: index shape is actully s32, can you use https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor_util.cpp#L1254 ? context is `SetDimensionShape` and `getDimensionShpe` currently only handle s32, this is where we have slight difference with upstream.

torch_xla/csrc/aten_xla_type.cpp
JackCaoG: nvm, now I remember we have a `symint` field in the yaml file.

test/test_dynamic_shapes.py
```diff
@@ -178,6 +178,37 @@ def test_expand_symint_correctness(self):
     self.assertEqual(t3.shape[0], 2)
     self.assertEqual(expand_out_aten.cpu(), expand_out_xla.cpu())
 
+  def test_topk_symint(self):
+    dev = xm.xla_device()
```
vanbasten23, line 182: nit: there is already a `dev = xm.xla_device()` defined at the top of the file

test/test_dynamic_shapes.py
```diff
@@ -178,6 +178,37 @@ def test_expand_symint_correctness(self):
...
+      torch.manual_seed(0)
+      x = torch.rand(10, 10).to(tensors[0].device)
+      results = []
+      for tensor in tensors:
+        k = torch.nonzero(tensor).shape[0]
+        for dim in range(2):
+          results += list(torch.topk(x, k=k, dim=dim))
+      return results
+
+    self.runAtenTest([torch.randint(0, 2, size=(10,)) for _ in range(5)],
```
vanbasten23, line 209: this is to test the correctness of the `topk`?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2506,6 +2506,20 @@ std::tuple<XLATensorPtr, XLATensorPtr> topk(const XLATensorPtr& input,
       input->CreateFrom(torch::lazy::Value(node, 1), at::ScalarType::Long));
 }
 
+std::tuple<XLATensorPtr, XLATensorPtr> topk_symint(const XLATensorPtr& input,
```
vanbasten23, line 2509: quick question, now that we have a symint version of `topk`, do we still need the original `topk`?

test/test_dynamic_shapes.py
```diff
@@ -178,6 +178,37 @@ def test_expand_symint_correctness(self):
...
     self.assertEqual(expand_out_aten.cpu(), expand_out_xla.cpu())
 
+  def test_topk_symint(self):
+    dev = xm.xla_device()
+    t1 = torch.zeros([5, 2], device=dev)
+    t1[3][0] = 1
+    t1[3][1] = 1
+    t2 = torch.nonzero(t1)
+    t3 = torch.zeros([10, 2], device=dev)
+    values, indices = torch.topk(t3, t2.shape[0], dim=0)
```
vanbasten23, line 188: I wonder if your change can handle the case where t3 is dynamic, eg t3=torch.nonzero(t1).

torch_xla/csrc/aten_xla_type.cpp
ymwangg: That's right. I basically followed the expand_symint example.

torch_xla/csrc/ops/topk.cpp
ymwangg: Actually the static version also returns s64 type, because of a s32 to s64 cast here: https://github.com/pytorch/xla/blob/1bbe4da79753bf0a4bd94ff106cf1267f841f302/torch_xla/csrc/xla_lower_util.cpp#L380-L383
Do I also need to disable this cast?

test/test_dynamic_shapes.py
ymwangg: That's right. The `runAtenTest` function will run and compare the results with native pytorch.

torch_xla/csrc/tensor_methods.cpp
ymwangg: I basically followed the `expand_symint` design, which keeps the static expand implementation. I guess the original `topk` implementation is still needed because the codegen for handling SymInt is slightly different (e.g. SetDimensionSize).

test/test_dynamic_shapes.py
ymwangg: I think this feature is already supported by xla through dynamic padder (static topk also works with dynamic t3). I'll add a test case for it later.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2506,6 +2507,20 @@ std::tuple<XLATensorPtr, XLATensorPtr> topk(const XLATensorPtr& input,
...
+                                                   bool largest, bool sorted,
+                                                   bool stable) {
+  SymIntElements k_symint = SymIntElements(k);
+  torch::lazy::NodePtr node = torch::lazy::MakeNode<TopKSymInt>(
+      input->GetIrValue(), k_symint,
+      torch::lazy::GetCanonicalDimensionIndex(dim, input->shape().get().rank()),
+      largest, sorted, stable);
+  return std::make_tuple(
+      input->CreateFrom(torch::lazy::Value(node, 0)),
+      input->CreateFrom(torch::lazy::Value(node, 1), at::ScalarType::Long));
```
ymwangg, line 2521: Some pytorch tests require the static topk returned indices type to be int64, otherwise it throws an error in pytorch [here](https://github.com/pytorch/pytorch/blob/db6cfff827158c3f22a6707aed35d2013ec2d994/aten/src/ATen/native/TensorAdvancedIndexing.cpp#L1960). I saw `nonzero` returns int32 indices. Why not just do a `int64 (pytorch) -> int32 (xla) -> int64 (pytorch)` conversion for handling indices?

torch_xla/csrc/tensor_methods.cpp
JackCaoG: yea, let's keep the static version their.

torch_xla/csrc/tensor_methods.cpp
JackCaoG: so what happened here is that underlying XLA type will still be `xla::s32`, however we mark the logical element type will be `at::ScalarType::Long`. This will create a problem since underlying s32 can not represent all numbers for long. We run into issues when we try to cast this kind of tensor(long logical type, real s32 type) to other dtypes, because `Cast` op doesn't know how to handle this kind of cases.

The right thing to do is probably do a manual s64->s32->s64 cast instead of playing the trick of `logical_element_type` here.

ymwangg: Thanks for the clarification! Do we want to add something like `res = torch::lazy::MakeNode<Cast>(res, at::ScalarType::Long)` here or use
`xla::ConvertElementType(indices, xla::PrimitiveType::S64, /*device=*/nullptr))` in `CreateTopK` like before?

JackCaoG: Let's handle it inside `CreateTopK` so it is cleaner.

vanbasten23 (approved):

ymwangg:
This PR requires removing torch_pin. Let me ping someone for review.

JackCaoG:
@ymwangg can upstream pr merge without this one? If so we can merge the upstream one first, otherwise we would have to do the two way pining.

ymwangg:
@JackCaoG Yes, the upstream PR does not depend on anything on xla.

JackCaoG:
Sorry I have overlooked this pr. @vanbasten23 can you take over this pr and rebase? I think we should merge this pr since it is already here.

PR #4646, opened 2023-02-17
[PJRT] Change the name of `LOCAL_RANK` to fix conflicts

will-cromar:
Our use of `LOCAL_RANK` conflicts with HuggingFace, which uses `LOCAL_RANK` to decide whether to use `torch.distributed` ([example](https://github.com/huggingface/transformers/blob/d3b1adf59fff726f1c8f324728e562237f080ce6/src/transformers/training_args.py#L1469)). This causes two problems:

1. We don't want to initialize `torch.distributed` in this case.
2. `LOCAL_RANK` is subtly different in PJRT, since it's the rank of a process and not necessarily a replica. But, we need to know the local process rank somehow to compute the correct local ordinal.

Make the PJRT local process rank env var different/more verbose to remove this conflict.

I'm avoiding changing other instances of `LOCAL_RANK` that relate to `torchrun`, but we don't yet support PJRT with `torchrun`.


will-cromar:
Validated manually by @AlexWertheim. I'll hold off on merging this until I can run the TPU CI.

torch_xla/experimental/pjrt.py
```diff
@@ -211,11 +211,9 @@ def _merge_replica_results(
 
 
 @requires_pjrt
-def _run_thread_per_device(local_rank: int,
-                           local_world_size: int,
-                           fn: Callable[[], R],
-                           initializer_fn: Callable[[int, int], None],
-                           master_port: int = 12355) -> Dict[int, R]:
```
JackCaoG, line 218: so we don't need master port anymore?

torch_xla/core/xla_env_vars.py
```diff
@@ -28,6 +28,8 @@
 TPU_MESH_CTLER_PORT = 'TPU_MESH_CONTROLLER_PORT'
 PJRT_DEVICE = 'PJRT_DEVICE'
 PJRT_SELECT_DEFAULT_DEVICE = 'PJRT_SELECT_DEFAULT_DEVICE'
+PJRT_LOCAL_PROCESS_RANK = 'PJRT_LOCAL_PROCESS_RANK'
```
steventk-g, line 31: Do we need to worry about the usage of LOCAL_RANK in our cpp code?
https://github.com/pytorch/xla/blob/master/third_party/xla_client/env_vars.cc#L27

torch_xla/experimental/pjrt.py
will-cromar: I also did a clean up of unused variables in the functions that I touched. Sorry, should have included that in the PR description.

steventk-g (approved):
Approved with a question about usage of local rank in CPP

torch_xla/core/xla_env_vars.py
will-cromar: Good catch! Yeah, I need to handle that for GPU intialization.

JackCaoG (approved):

AlexWertheim (approved):

PR #4649, opened 2023-02-17
Disable docker sandbox

stgpetrovic:
Revert to local builds on cloud vs sandboxed (shm size issues).

There are a couple of issues there that intermingle.

Bazel has a [bug](https://github.com/bazelbuild/bazel/issues/15359) where the cppmap files don't get included as deps when building with clang when not in sandbox, from an external repository (local repository is OK). Using a cache might alleviate this.
The suggested solution is building in the sandbox.

For some reason, `docker build --shm-size=16g` does not prduce a 16g `/dev/shm` (it is actually 64M which is the default). There are differences in docker build/run but during build it should have 16g but it does not (as checked with `RUN df -h` showing 64M).

So the solution for this issue is to build sandboxed (without cache, unlike the other build that has cache and does not suffer from this; we could replicate that in tpu tests too), and in docker, but to run with `RUN --mount=type=tmpfs,target=/dev/shm,rw cmd` which then mounts `tmfps` as `/dev/shm` and builds properly. The build needs about `4G` of temp files so `16G` is an overkill there.

This makes the build work (avoid bazel bug by sandboxing), and makes it fast (avoid sandbox base being on HDD, and avoid docker `shm-size` issue by mounting the `tmpfs`).

We can rework this as needed in the future when we streamline all different builds into a single config.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -10,7 +10,8 @@ build --announce_rc
 # TODO(goranpetrovic): figure out visibility of tensorflow libraries.
 build --nocheck_visibility
 
-#build --define open_source_build=true
+# We can set this to `standalone` after https://github.com/bazelbuild/bazel/issues/15359 is resolved.
+build --spawn_strategy=sandboxed
 
 build --enable_platform_specific_config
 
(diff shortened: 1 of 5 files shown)


will-cromar (approved):
Thanks for fixing this!

yeounoh (approved):
Thank you!

PR #4650, opened 2023-02-17
Bazel documentation

stgpetrovic:
Document basics for bazel usage.


JackCaoG:
Can we also update https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md to match the new bazel?

docs/bazel.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Code coverage
+
+When running tests, it can be useful to calculate code coverage.
+
+```bash
+bazel coverage //third_party/xla_client/...
+```
+
+Coverage can be visualized using `lcov` as described in [Bazel's documentation](https://bazel.build/configure/coverage), or in your editor of choice with lcov plugins, e.g. [Coverage Gutters](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters) for VSCode.
```
steventk-g, line 41: What are the prereqs for doing this on our source code? Since we already use gtest, would lcov through bazel be seamless?

stgpetrovic: We can already do it for the 3rdparty/xla_client and soon when i migrate the torch_cla/csrc on that as well.I plan to migrate the python tests to bazel as well when that will give way too. Bazel produces lcov, and it can be consed by genhtml or whatever else.

docs/bazel.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+Tests can also be executed remotely in the same fashion as builds.
+
+## Code coverage
+
+When running tests, it can be useful to calculate code coverage.
+
+```bash
+bazel coverage //third_party/xla_client/...
```
vanbasten23, line 38: I used to use blaze. How much difference are there between bazel and blaze besides the fact that bazel is the open source version of blaze?

docs/bazel.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+Coverage can be visualized using `lcov` as described in [Bazel's documentation](https://bazel.build/configure/coverage), or in your editor of choice with lcov plugins, e.g. [Coverage Gutters](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters) for VSCode.
+
+
+## Language Server
+
+Bazel can power a language server like [clangd](https://clangd.llvm.org/) that brings code references, autocompletion and semantic understanding of the underlying code to your editor of choice. For VSCode,
```
vanbasten23, line 46: wow, it's awesome!

docs/bazel.md
```diff
@@ -0,0 +1,58 @@
...
+one can use [Bazel Stack](https://github.com/stackb/bazel-stack-vscode-cc) that can be combined with
+[clangd](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd) functionality to bring powerful features to assist code editing.
+
+## Building PyTorch/XLA
+
+As always, PyTorch/XLA is built using Python `distutils`:
+
+```bash
+BAZEL_REMOTE=1 TPUVM_MODE=1 python setup.py bdist_wheel
```
vanbasten23, line 55: currently many pt/xla devs uses this doc https://docs.google.com/document/d/1X2Z0VmKjVV6vlUTRmu-soWW8RcDOyshxZSdSUZLkc-U/edit?resourcekey=0-yjGpRCaBlMBIEW0WFoa0qw#bookmark=id.xaa5do61ce0c, i.e. `TPUVM_MODE=1 python setup.py develop --user`.  Suggest to also add `BAZEL_REMOTE=1 TPUVM_MODE=1 python setup.py develop --user` to your doc.

will-cromar (commented):
Can you also include some information about how to set up a local repository for TF (and eventually `torch`) and apply patches? Following the comment in the WORKSPACE file, I checked out a local copy and added `--override_repository=org_tensorflow=$(pwd)/third_party/tensorflow`, but I get this error:

```
ERROR: Analysis of target '//third_party/xla_client:libxla_computation_client.so' failed; build aborted: no such package 'tensorflow': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /workspaces/pjrt/pytorch/xla/tensorflow
```

I probably set something up wrong, and unfortunately Bazel's error message isn't really helpful here.

docs/bazel.md
stgpetrovic: THere should be no differences. Bazel has some quirks (not monorepo) so that's a bit different, and some coverage html support is not there, but when it works it should be equivalent.

PR #4656, opened 2023-02-18
Bazel rbe

stgpetrovic:
Test remote build execution.


.bazelrc
```diff
@@ -75,6 +76,93 @@ build:tpu --define=with_tpu_support=true
...
+# Enforces consistent action environment variables. This is important to
+# address rebuild sensitivity on changes to the environment variables and
+# helps cache.
+build:rbe --incompatible_strict_action_env
+
+
+# Setup ResultStore
+common:rbe --bes_backend=buildeventservice.googleapis.com
+common:rbe --bes_instance_name=tpu-pytorch
+common:rbe --remote_instance_name="projects/tpu-pytorch/instances/default_instance"
```
will-cromar, line 89: Would this still work outside of our team for people without access to `tpu-pytorch`? Is there a way to make the project name a parameter that others can change?

will-cromar (commented):
Is there any other project-level setup that is needed to use RBE?

.bazelrc
stgpetrovic: Others can override this parameter on the command line and it will be taken.

I'll create a new role for members of the project and add devs in so they can just use default credentials, and that will work out of the box; same for our cloud builds. For circleci, we use the cloud token, and it authenticates like that.

For anyone else, specifying remote builds will fail by default with authentication errors unless they point to a different remote instance where they have access.

PR #4659, opened 2023-02-19
Split `libxla_computation_client.so` into libraries.

stgpetrovic:
To do this split, reworked the metrics analysis.
Currently, metrics analysis would get a global singleton of the computation client to get the metrics. This change switches to injection, so the python bindings init can use the singleton to pass the metrics down to the analysis, removing the dependency from the analysis to the whole client.

Add some tests.


will-cromar (approved):
Thanks!

PR #4661, opened 2023-02-21
Lower aten::_unique2

ymwangg:
This PR adds `aten::_unique2`. Similar to `aten::nonzero`, this op is currently marked as experimental.

### DIFF
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -192,7 +192,8 @@ class TestDynamicShapes(test_utils.XlaTestCase):
 
     self.runAtenTest([torch.randint(4, 10, size=(10,)) for _ in range(2)] +
                      [torch.randint(4, 10, size=(10, 10)) for _ in range(2)] +
-                     [torch.rand(10, 10) for _ in range(2)], test_fn)
+                     [torch.rand(10, 10) for _ in range(2)] + [torch.ones(1)],
+                     test_fn)
 
 
 if __name__ == '__main__':
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -178,6 +178,22 @@ class TestDynamicShapes(test_utils.XlaTestCase):
     self.assertEqual(t3.shape[0], 2)
     self.assertEqual(expand_out_aten.cpu(), expand_out_xla.cpu())
 
+  def test_unique_correctness(self):
+
+    def test_fn(*tensors):
+      results = []
+      for t in tensors:
+        results += [torch.unique(t, sorted=True)]
+        results += list(torch.unique(t, sorted=True, return_inverse=True))
+        results += list(
+            torch.unique(
+                t, sorted=True, return_inverse=True, return_counts=True))
+      return results
+
+    self.runAtenTest([torch.randint(4, 10, size=(10,)) for _ in range(2)] +
+                     [torch.randint(4, 10, size=(10, 10)) for _ in range(2)] +
+                     [torch.rand(10, 10) for _ in range(2)], test_fn)
+
 
 if __name__ == '__main__':
   assert os.environ['XLA_EXPERIMENTAL'] != ''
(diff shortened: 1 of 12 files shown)


test/test_dynamic_shapes.py
```diff
@@ -178,6 +178,23 @@ def test_expand_symint_correctness(self):
...
+      results = []
+      for t in tensors:
+        results += [torch.unique(t, sorted=True)]
+        results += list(torch.unique(t, sorted=True, return_inverse=True))
+        results += list(
+            torch.unique(
+                t, sorted=True, return_inverse=True, return_counts=True))
+      return results
+
+    self.runAtenTest([torch.randint(4, 10, size=(10,)) for _ in range(2)] +
```
JackCaoG, line 193: can we also check result of `unique` is dynamic?

JackCaoG (approved):
Mostly LGTM, thanks @ymwangg !

xla_native_functions.yaml
```diff
@@ -321,6 +321,7 @@ supported:
   - triangular_solve
   - unbind.int
   - uniform_
+  - _unique2
```
vanbasten23, line 324: I wonder how `_unique2` differs from `unique`

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -3190,4 +3190,25 @@ at::Tensor XLANativeFunctions::_cdist_forward(
       bridge::GetXlaTensor(x1), bridge::GetXlaTensor(x2), p));
 }
 
+std::tuple<at::Tensor, at::Tensor, at::Tensor> XLANativeFunctions::_unique2(
```
vanbasten23, line 3193: FYI: I don't think this PR will be impacted but just in case you need to develop dynamic models, you'd need to develop on `functionization_dynamic_shape` branch which has functionalization feature enabled.

xla_native_functions.yaml
ymwangg: The `_unique` [function](https://github.com/pytorch/pytorch/blob/e22d791287d1c00cd11b9229ad15d030224ab4b1/aten/src/ATen/native/native_functions.yaml#L5842) does not support `return_counts`. It looks like it's not accessible to normal user because `torch.unique` is dispatched to either `unique_dim` or `_unique2` [here](https://github.com/pytorch/pytorch/blob/1e2d82b8e436fc5c72071c3b78bc9d13f4a02268/torch/functional.py#L722-L724).

torch_xla/csrc/ops/unique2.h
```diff
@@ -0,0 +1,16 @@
...
+
+namespace torch_xla {
+
+class Unique2 : public XlaNode {
+ public:
+  Unique2(const torch::lazy::Value& input);
+
+  torch::lazy::NodePtr Clone(torch::lazy::OpList operands) const override;
+
+  XlaOpVector Lower(LoweringContext* loctx) const override;
```
vanbasten23, line 13: nit: mind adding a ToString()?

torch_xla/csrc/ops/unique2.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "torch_xla/csrc/tensor_util.h"
+#include "torch_xla/csrc/xla_lower_util.h"
+
+namespace torch_xla {
+namespace {
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input) {
+  xla::Shape input_shape = GetXlaShape(input);
+  int64_t num_elements = xla::ShapeUtil::ElementsIn(input_shape);
+  xla::PrimitiveType indices_type = GetShapeDimensionType(/*device=*/nullptr);
```
ymwangg, line 13: Same question here, do we want the indices type to be S32 or S64? Though returning S32 does not break any tests.

JackCaoG: I would leave most of the op in S32 to avoid extra cost lol. I think eventually we want to make `setDimensionSize` and `getDimensionSize` handle S64, which I don't think is super complicated, but someone need to work on it...

vanbasten23 (approved):

torch_xla/csrc/ops/unique2.h
ymwangg: It appears ToString() override is only needed when the op has input types other than lazy::Value. So I guess the default should be fine.

PR #4664, opened 2023-02-21
[FSDP] Add meta device and torchdistx deferred_init feature in FSDP

zachzzc:
This PR brings the deferred initialization from torch FSDP into torch-xla FSDP

To use the FSDP with auto_wrap, we will likely to encounter a OOM issue when instantiate the gigantic model before wrapping it with FSDP.

```
model = gigantic_model() ----> OOM here
model = XlaFullyShardedDataParallel(
      model,
      auto_wrap_policy=always_wrap
)
```

To address this issue, we need a mechanism to delay the initialization to the FSDP stage, where we should initialize the model layer by layer (module by module) and partition the parameters immediately to reduce the peak memory. To achieve this, we need two things: 1) During the instantiation we should not allocate the memory immediately. 2) Re-initialize the sub module correctly.

There are two ways to delay the memory allocation. The first one is the `meta` device. The actual memory allocation will not happen to the tensors with `meta` device, and only in the FSDP stage we move the tensors from meta device to xla device, allocate the memory and re-do the initialization. 

```
module = MyModule(device="meta")
def my_init_fn(module):
   # responsible for initializing a module, such as with reset_parameters
   ...
fsdp_model = FSDP(module, param_init_fn=my_init_fn, auto_wrap_policy=size_based_auto_wrap_policy)
```
However, this method has a duplicated-initalization problem that is also raised in PyTorch-FSDP https://github.com/pytorch/pytorch/issues/90465. This problem will cause the initialization result mismatches.

The second approach is to use torchdistX package. Once the module is wrapped by `deferred_init`, all the tensors inside will become fake tensors. And It provides extra ability to record the operations applied to the fake tensors. When we want to materialize the module, it will allocate the memory and replay all the operations.

```
# With torchdistX
module = deferred_init.deferred_init(MyModule)
# Will initialize via deferred_init.materialize_module().
fsdp_model = FSDP(module, auto_wrap_policy=size_based_auto_wrap_policy)
```

TorchdistX part is tested locally with source-build torchdistx. The disadvantage of this approach is that the official support of torchdistX is till PT1.12. I tried the other versions. It can work with PT-XLA 1.13 out-of-the-box and need minor changes to work with latest PT. I don't know why they stop maintaining this package. Feel free to comment on it if anyone has insights.

@JackCaoG @ronghanghu

### DIFF
--- a/test/test_torch_distributed_fsdp_meta.py
+++ b/test/test_torch_distributed_fsdp_meta.py
@@ -8,7 +8,7 @@ import torch.distributed as dist
 import torch_xla.distributed.xla_multiprocessing as xmp
 import torch_xla.core.xla_model as xm
 
-import traceback
+import sys
 
 _TORCHDISTX_AVAIL = True
 try:
@@ -141,14 +141,20 @@ class TestFSDPWithMetaDevice():
 
 
 def _mp_fn(index):
-  dist.init_process_group(
-      'xla', world_size=xm.xrt_world_size(), rank=xm.get_ordinal())
-  test = TestFSDPWithMetaDevice()
-  test.test_simple_model_with_meta_device_reset_params()
-  test.test_simple_model_with_meta_default_reset_params()
-  if _TORCHDISTX_AVAIL:
-    test.test_simple_model_with_torchdistX_init_fn()
-    test.test_simple_model_with_default_torchdistX()
+  device = xm.xla_device()
+  if xm.xla_device_hw(device) in ('TPU', 'GPU'):
+    dist.init_process_group(
+        'xla', world_size=xm.xrt_world_size(), rank=xm.get_ordinal())
+    test = TestFSDPWithMetaDevice()
+    test.test_simple_model_with_meta_device_reset_params()
+    test.test_simple_model_with_meta_default_reset_params()
+    if _TORCHDISTX_AVAIL:
+      test.test_simple_model_with_torchdistX_init_fn()
+      test.test_simple_model_with_default_torchdistX()
+  else:
+    print(
+        'Default device {} is not a TPU or GPU device'.format(device),
+        file=sys.stderr)
 
 
 if __name__ == '__main__':
--- a/test/test_torch_distributed_fsdp_meta.py
+++ b/test/test_torch_distributed_fsdp_meta.py
@@ -1,0 +1,155 @@
+import torch
+import torch.nn as nn
+from torch_xla.distributed.fsdp import XlaFullyShardedDataParallel
+from torch_xla.distributed.fsdp.wrap import (
+    always_wrap_policy as always_wrap,)
+
+import torch.distributed as dist
+import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla.core.xla_model as xm
+
+import traceback
+
+_TORCHDISTX_AVAIL = True
+try:
+  from torchdistx import deferred_init
+except ImportError:
+  _TORCHDISTX_AVAIL = False
+
+
+def _reset_params_if_meta(is_meta, model):
+  # For torchdistX init, we don't need to call reset_params, as
+  # deferred_init(model).materialize() is equivalent to model().
+  if is_meta:
+    model.reset_parameters()
+
+
+class MyLinear(nn.Linear):
+  """
+    Linear layer with deterministic reset_parameters for testing.
+    """
+
+  def __init__(self, *args, **kwargs):
+    super().__init__(*args, **kwargs)
+
+  def reset_parameters(self, *args, **kwargs):
+    with torch.no_grad():
+      self.weight.fill_(1)
+
+
+class MyModel(nn.Module):
+
+  def __init__(self, device):
+    super().__init__()
+    self.lin1 = MyLinear(2, 2, bias=False, device=device)
+    self.lin2 = MyLinear(2, 2, bias=False, device=device)
+
+  def forward(self, x):
+    return self.lin2(self.lin1(x))
+
+  def reset_parameters(self, *args, **kwargs):
+    for m in [self.lin1, self.lin2]:
+      if not isinstance(m, XlaFullyShardedDataParallel):
+        m.reset_parameters()
+
+
+def _init_with_reset_params(module):
+  """
+    to_empty + reset_parameters() init function example for modules
+    initailized with device="meta"
+    """
+  is_meta = any(t.is_meta for t in module.parameters())
+  if is_meta:
+    module.to_empty(device=xm.xla_device())
+  with torch.no_grad():
+    module.reset_parameters()
+
+
+def _init_with_torchdistX(module):
+  """
+    torchdistX-based deferred module initialization function example
+    using ``materialize_module``.
+    """
+  assert _TORCHDISTX_AVAIL
+
+  def check_fn(k):
+    return not isinstance(k, XlaFullyShardedDataParallel)
+
+  deferred_init.materialize_module(module, check_fn=check_fn)
+
+
+class TestFSDPWithMetaDevice():
+
+  def _compare_fsdp(self, fsdp1, fsdp2):
+    for p1, p2 in zip(fsdp1.parameters(), fsdp2.parameters()):
+      assert (torch.allclose(p1.cpu(), p2.cpu()))
+
+  def _test_simple_model_with_meta_device(self, meta_module_fn, init_fn=None):
+    # Create model on meta device and wrap with FSDP.
+    model = meta_module_fn()
+    inp = torch.randn(10, 2, device=xm.xla_device())
+
+    fsdp_meta = XlaFullyShardedDataParallel(
+        model,
+        auto_wrap_policy=always_wrap,
+        param_init_fn=init_fn,
+    )
+    meta_opt = torch.optim.SGD(fsdp_meta.parameters(), lr=1e-3)
+    fsdp_meta(inp).sum().backward()
+    meta_opt.step()
+    xm.mark_step()
+
+    regular = MyModel(device=xm.xla_device())
+    fsdp_regular = XlaFullyShardedDataParallel(
+        regular, auto_wrap_policy=always_wrap)
+    regular_opt = torch.optim.SGD(fsdp_regular.parameters(), lr=1e-3)
+    fsdp_regular(inp).sum().backward()
+    regular_opt.step()
+    xm.mark_step()
+
+    self._compare_fsdp(fsdp_meta, fsdp_regular)
+
+  def test_simple_model_with_meta_device_reset_params(self):
+
+    def meta_module_fn():
+      return MyModel(device="meta")
+
+    self._test_simple_model_with_meta_device(meta_module_fn,
+                                             _init_with_reset_params)
+
+  def test_simple_model_with_meta_default_reset_params(self):
+
+    def meta_module_fn():
+      return MyModel(device="meta")
+
+    self._test_simple_model_with_meta_device(meta_module_fn)
+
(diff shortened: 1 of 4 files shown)


JackCaoG:
CI failure seems unrelated, can you rebase?

JackCaoG:
@AlexWertheim FYI

JackCaoG:
I saw that last commit to [torchdistX](https://github.com/pytorch/torchdistx) is last Sept, @wconstab Do you know if that branch is still being developed/maintianed?

@zachzzc Do you mind explaining how will our dependency to torchdistX looks like?

zachzzc:
> I saw that last commit to [torchdistX](https://github.com/pytorch/torchdistx) is last Sept, @wconstab Do you know if that branch is still being developed/maintianed?
> 
> @zachzzc Do you mind explaining how will our dependency to torchdistX looks like?

The main branch of torchdistX doesn't compile with the latest PyTorch. It works after I fixed the compilation locally. Ideally we should fix torchdistX and just use the `pip install torchdistX` to leverage it.

zachzzc:
> @zachzzc so this pr by itself is not enough, user has to manully build the torchdistX and install it? This doesn't sounds like a very smooth user journey, which is my concern.

Totally agree. However, this is currently the only way to get address the large model initialization problem and also is used in torch-fsdp. I will say this is a short-term solution, and long-term we need the PyTorch Core to adopt the fake tensors mechanism and update the logic here.

ronghanghu:
@zachzzc thanks for the PR! Just to clarify, for the OOM here, is it OOM on the host side (CPU memory), or OOM on the XLA device side?

After https://github.com/pytorch/xla/pull/3992, we can now initialize the full model on the host side, and only send sharded parameters to the XLA device
```
module = MyModule(device="cpu")
def my_init_fn(module):
   # responsible for initializing a module, such as with reset_parameters
   ...
fsdp_model = FSDP(module, param_init_fn=my_init_fn, auto_wrap_policy=size_based_auto_wrap_policy)
```

> The first one is the meta device. The actual memory allocation will not happen to the tensors with meta device, and only in the FSDP stage we move the tensors from meta device to xla device, allocate the memory and re-do the initialization.

Personally I'm slightly more in favor of this approach :) I guess this depends on how much consistency do we need between different ways of model initialization/wrapping. (In my experience on TPU v3 with PJRT devices, we had a lot of randomness e.g. due to Python thread concurrency in PJRT runtime, so it was really had for me to get consistent random initialization.)

zachzzc:
> 

It is about host CPU OOM. 

I don't have an evidence to prove the effect of the initialization variance yet. In general I think it is good to leave both interfaces here so user can choose.

zachzzc:
@JackCaoG @ronghanghu  any other suggestions about this PR? I think we should leave both `meta` device option and torchdistx option to the users and also align with torch FSDP. If users are okay with the randomness in the initialization, they can use `meta` approach. Otherwise, users can manually install torchdistx.

JackCaoG:
I am hesitant to take any dependencies on anything that is experimental since it is really hard to deprecate things once it is out and it leaves a bunch of maintaince burden on our end. There is also no sign that https://github.com/pytorch/torchdistx is being maintained so I am more worried. If Ronghang's method already work, I will prefer not to merge this pr.

I am fine with meta arpproach through(assuming we can fix the bug..), that's something in the pytorch master at least.

ronghanghu:
I had the same feeling as Jack, that torchdistx could be a troublesome dependency. It could be hard to make it compatible with future PyTorch as the development on torchdistx has stopped.

YangFei1990:
I'm inclined to add support for this feature. The reasons are:
- IMO this is not a dependency. Users can still use torch_xla without torchdistx. But this offers users extra capability to use torchdistx.
- This aligns with the native Pytorch FSDP implementation: https://github.com/pytorch/pytorch/blob/master/torch/distributed/fsdp/_init_utils.py#L408

I understand that we have concerns that torchdistx seems not under active development, but since upstream PT also do the same way, once there is a native solution (which is likely the fake tensor mentioned by @zachzzc), we can easily switch to it together with upstream PT.

amithrm:
Hi All, do we know the reason why the Fake tensor mechanism is deprecated? Any strong technical drawbacks for example. 
If not, we would need this support immediately, otherwise it will be a blocker to launch large models. 

Also, what is the main difference between using the meta device "vs" fake tensor approach? The fake tensor approach using deferred_init() also uses the meta device underneath correct.
So, it look like we need both: Fake tensors for recording operations + Meta device?

JackCaoG:
Hmm, ok if upstream pytorch FSDP takes the same dependency I guess I am ok with it. Please add a comment in the code saying this is experimental and might be remove in the future releases. @alanwaketan I will let you and ronghang take another pass in the pr. 

@ronghanghu I think AWS trainium has 32 devices per host, so they will start 32 process, each trying to init the whole parameter hence it is a much bigger issue than TPU.

torch_xla/distributed/fsdp/_init_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+def _get_orig_params(
+    module: nn.Module,
+    ignored_params: Set[nn.Parameter],
+) -> Iterator[nn.Parameter]:
+  """
+    Returns an iterator over the original parameters in ``module``, ignoring
+    the parameters in ``ignored_params``, any ``FlatParameter`` s (which may be
+    present due to nested FSDP wrapping), and any original parameters already
+    flattened (only relevant when ``use_orig_params=True``).
```
alanwaketan, line 94: What's `use_orig_params`? Is this method even needed for XLA?

zachzzc: Good catch. This is not used by torch-xla. Will remove

test/test_torch_distributed_fsdp_meta.py
```diff
@@ -0,0 +1,161 @@
...
+
+import torch.distributed as dist
+import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla.core.xla_model as xm
+
+import sys
+
+_TORCHDISTX_AVAIL = True
+try:
+  from torchdistx import deferred_init
```
alanwaketan, line 15: Do we need to install torchdistx in our CI for this coverage?

JackCaoG:
@YangFei1990 Do you mind open an issue on pytorch and  cc `ezyang`  and `awgu` to discuss whether deferred_init will be merged to the master?

torch_xla/distributed/fsdp/_init_utils.py
```diff
@@ -0,0 +1,102 @@
...
+try:
+  from torchdistx import deferred_init, fake  # type: ignore[import]
+except ImportError:
+  _TORCHDISTX_AVAIL = False
+
+
+def _materialize_module(
+    module: nn.Module,
+    param_init_fn: Optional[Callable[[nn.Module], None]],
+    ignored_params: Set[nn.Parameter],
```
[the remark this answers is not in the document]
zachzzc: yes I leave here in case we need it in the future. The auto_wrap also has ignored_params interface but not sued, so theoretically we can leverage them together

test/test_torch_distributed_fsdp_meta.py
zachzzc: No, I see upstream FSDP also has similar issue that it is not tested in the CI for the coverage. With latest torch we need to build torchdistx from source + some changes to match pytorch, so it is hard to maintain the C if we add it.

alanwaketan: Sad. But I guess as long as you guys are testing it and we brand it as experimental feature. Then it should be fine.

torch_xla/distributed/fsdp/_init_utils.py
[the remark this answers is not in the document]
alanwaketan: Maybe add a TODO?

alanwaketan (approved):
LGTM. Thanks, Zack.

YangFei1990:
> @YangFei1990 Do you mind open an issue on pytorch and cc `ezyang` and `awgu` to discuss whether deferred_init will be merged to the master?

Sorry for the delay. I posted an issue here: https://github.com/pytorch/pytorch/issues/96161

zachzzc:
@ronghanghu @JackCaoG  could you help take a final look and merge this PR?

ronghanghu (approved):

JackCaoG:
@zachzzc Do you have the `TorchdistX` python 3.8 whl that works with the head? I think if we are going to let user use it, we should provide a working whl.

JackCaoG:
@zachzzc or if you can share the diff I can try to build it on my end.

zachzzc:
@JackCaoG  Hi Jack. I upload the fixes to here https://github.com/zachzzc/torchdistx. You can try pull it and build
I can also send the wheels, but not sure what's a good way to share it. Let me know you want it this way.

JackCaoG:
@zachzzc Thanks, let me give it a try.

JackCaoG:
@zachzzc I run into build issue
```
(pytorch) root@81b6081019e5:/pytorch/torchdistx# cmake -DTORCHDIST_INSTALL_STANDALONE=ON -B build
-- The CXX compiler identification is Clang 8.0.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/clang++-8 - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Python3: /root/anaconda3/envs/pytorch/bin/python3.8 (found version "3.8.8") found components: Interpreter Development.Module 
CMake Error at cmake/Helpers.cmake:365 (add_subdirectory):
  The source directory

    /pytorch/torchdistx/third-party/pybind11

  does not contain a CMakeLists.txt file.
Call Stack (most recent call first):
  CMakeLists.txt:72 (torchdist_add_third_party)


-- MKL_ARCH: None, set to ` intel64` by default
-- MKL_ROOT /root/anaconda3/envs/pytorch
-- MKL_LINK: None, set to ` dynamic` by default
-- MKL_INTERFACE_FULL: None, set to ` intel_ilp64` by default
-- MKL_THREADING: None, set to ` intel_thread` by default
-- MKL_MPI: None, set to ` intelmpi` by default
CMake Warning at /root/anaconda3/envs/pytorch/lib/python3.8/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:22 (message):
  static library kineto_LIBRARY-NOTFOUND not found.
Call Stack (most recent call first):
  /root/anaconda3/envs/pytorch/lib/python3.8/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:127 (append_torchlib_if_found)
  CMakeLists.txt:87 (find_package)


-- Found Torch: /root/anaconda3/envs/pytorch/lib/python3.8/site-packages/torch/lib/libtorch.so (Required is at least version "1.11") 
-- Configuring incomplete, errors occurred!
See also "/pytorch/torchdistx/build/CMakeFiles/CMakeOutput.log".
```

seems to do with `kineto_LIBRARY-NOTFOUND`.

JackCaoG:
lol, we usually upload the wheel to gcs but I guess it is hard to share stuff since we use two different cloud. Maybe we can do it old school that you email me the jackcao@google.com, or you can upload it to a pytorch/xla branch then I can just copy it there.

JackCaoG:
ok I know what the issue is lol I forgot to do the submodule update

JackCaoG:
@zachzzc I am able to get it to work, thanks a lot for your help!

zachzzc:
@JackCaoG NP. Yeah I also saw the `kineto_LIBRARY-NOTFOUND` and it looks like just warnining

PR #4665, opened 2023-02-21
[WIP] Add C++ tests to TPU CI

will-cromar:


setup.py
```diff
@@ -374,9 +374,9 @@ def make_relative_rpath(path):
         ],
     },
     data_files=[
-        'scripts/fixup_binary.py',
+        ('scripts', ['scripts/fixup_binary.py']),
```
vanbasten23, line 377: i wonder what the change here does

will-cromar: I'm fixing a warning from using a blank install directory: https://docs.python.org/3/distutils/setupscript.html#installing-additional-files

will-cromar:
It looks like the use of env vars within the test script is breaking because we pass the job spec through `envsubst` first:

```
Step #2 - "create-job": + TPU_LIBRARY_PATH=/lib/libtpu.so
Step #2 - "create-job": + /test_ptxla
```

It's basically replacing the prefixes for each path with an empty string because those variables aren't set in the GCB process.

will-cromar:
I'm still running into issues trying to get the tests to actually run:

`Step #2 - "create-job": /usr/local/test_ptxla: error while loading shared libraries: libc10.so: cannot open shared object file: No such file or directory`

PR #4669, opened 2023-02-22
[Internal Test]Enable HLO->MHLO->HLO Round trip

lsy323:
Use this draft PR to keep track of the work on HLO->StableHLO->HLO PoC.
The goal is to build PT/XLA with the util function defined in "tensorflow/compiler/xla/translate/hlo_to_mhlo/hlo_to_mlir_hlo.h" and "tensorflow/compiler/xla/translate/mhlo_to_hlo/mlir_hlo_to_hlo.h".


lsy323:
cc @miladm @vanbasten23 @ManfeiBai

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1138,6 +1140,10 @@ XLAGraphExecutor::CompilationResult XLAGraphExecutor::Compile(
   }
 
   xla::XlaComputation computation = ConsumeValue(lowering_ctx.BuildXla());
+  mlir::MLIRContext context;
+  mlir::OwningModuleRef module = mlir::ModuleOp::create(mlir::UnknownLoc::get(&context));
+  HloModuleProto proto(computation.proto());
+  TF_RETURN_IF_ERROR(ConvertHloToMlirHlo(module.get(), &proto, /*import_all_computations=*/false));
```
JackCaoG, line 1146: hmm this is not roundtrip right? You just convert HLO to MHLO

lsy323: Not round trip. Just want to solve compiling issue first. Open the draft PR to keep track of progress and allow more people with the work

lsy323:
Currently still hit a compiling error:
```
/workspaces/work/pytorch/xla/bazel-xla/external/llvm-project/llvm/include/llvm/ADT/STLExtras.h:26:10: fatal error: 'llvm/Config/abi-breaking.h' file not found
#include "llvm/Config/abi-breaking.h"
```
"llvm/Config/abi-breaking.h" doesn't exist in llvm-project repo, only "llvm/Config/abi-breaking.h.cmake" exists, it seems that an additional step is needed to generate headers in llvm-project

torch_xla/csrc/xla_graph_executor.cpp
miladm: +1

We also want to understand the latency of these translations to StableHLO ABD back from StableHLO

miladm: I expect the two extra latencies end up being asymmetric - but tolerable

third_party/xla_client/mhlo_helper.cc
```diff
@@ -0,0 +1,19 @@
+#include "third_party/xla_client/mhlo_helper.h"
+#include "tensorflow/compiler/xla/translate/hlo_to_mhlo/hlo_to_mlir_hlo.h"
+#include "tensorflow/compiler/xla/translate/mhlo_to_hlo/mlir_hlo_to_hlo.h"
+
+namespace xla {
+
+void hlo_mhlo_hlo_roundtrip_helper(HloModuleProto* proto) {
+  mlir::MLIRContext context;
+  mlir::ModuleOp mlir_module = mlir::ModuleOp::create(mlir::UnknownLoc::get(&context));
+  auto status = ConvertHloToMlirHlo(mlir_module, proto, /*import_all_computations=*/false);
```
GleasonK, line 10:
```
  if (!status.ok()) {
    // Do some error handling / logging 
  }
```

Here and below, will help track where issues are happening.

lsy323: Thanks for the comment. Yes, will add more logging when debugging the tests failure.

lsy323:
HLO->MHLO round trip dependency issue is resolved. Here are the details:

The most straightforward way is to add MLIR dependency to `libptxla.so`. However, PyTorch/XLA uses setuptools to build csrcs while TF:MLIR uses Bazel. So it's hard to introduce TF:MLIR dependency to `libptxla.so` directly.

Then the plan is to include MLIR headers in PyTorch/XLA, add add MLIR dependency in `libxla_computation_client.so`. Then `libptxla.so` links with `libxla_computation_client.so`. This can pass compilation but hit runtime errors on `undefined symbol llvm::DisableABIBreakingChecks`. After investigation, it caused by the LLVM headers included by `libptxla.so` are not properly processed.

To resolve the header dependency issue above,  LLVM header dependency needs to be removed in `libptxla.so`. Wrappers functions for HLO->MHLO conversion API is added to `libxla_computation_client.so`, and all of the TF:MLIR, LLVM dependency will be handled by bazel when building `libxla_computation_client.so`.

third_party/xla_client/mhlo_helper.cc
```diff
@@ -1,18 +1,33 @@
...
   mlir::MLIRContext context;
   mlir::ModuleOp mlir_module = mlir::ModuleOp::create(mlir::UnknownLoc::get(&context));
   auto status = ConvertHloToMlirHlo(mlir_module, proto, /*import_all_computations=*/false);
+  if (!status.ok()) {
+    std::cout << "hlo2mhlo not ok" << std::endl;
+  }
+  if (!mlir::verify(mlir_module).succeeded()) {
+    std::cout << "verify not ok" << std::endl;
+    TF_ASSIGN_OR_RETURN(auto hlo_module, xla::util::CreateModuleFromProto(*proto));
+    std::cout << hlo_module->ToString() << std::endl;
```
GleasonK, line 21: (I think) you should also be able to dump the invalid MLIR using `mlir_module.dump()`. Would probably be helpful as well.

Also should we return in this `if`? Log invalid case and return the original HLO before swapping. That would allow us to get all the failing test HLOs if a crash or anything is blocking them from running.

GleasonK: Oh nice! Both failing on dynamic reshape?

lsy323: Yes, I think they are the same case with minor difference. The error messages are the same

PR #4675, opened 2023-02-22
Torch pin for release branch should be under root

JackCaoG:


JackCaoG:
This is also what we do in https://github.com/pytorch/xla/blob/r1.13/.torch_pin (it was later changed to a tag after 1.13 branch is released). I think some of our build script requires release branch to have pin under root.

steventk-g (approved):

PR #4678, opened 2023-02-22
Implement XLASymNodeImpl.has_hint and re-enable failing DS tests.

vanbasten23:


torch_xla/csrc/tensor.cpp
```diff
@@ -796,6 +842,8 @@ bool XLASymNodeImpl::bool_() {
   return dn->getDynamicValue() != 0;
 }
 
+bool XLASymNodeImpl::has_hint() { return true; }
```
vanbasten23, line 845: hi @ezyang , I'm not sure if it's the correct implementation for XLA. I was reading the description of the pr https://github.com/pytorch/pytorch/pull/94431 and want to clarify with you about the meaning of `has_hint`. From the description, it sound "a SymInt has_hint" is equivalent to "a SymInt is backed". What does it mean by "a SymInt is backed"?

ezyang: Unbacked SymInt is the result of a data dependent output like nonzero; we don't know what the value is because it's data dependent. Because XLA uses max bound, I believe you just always return true here.

ezyang (approved):

alanwaketan (approved):

torch_xla/csrc/tensor.cpp
vanbasten23: Thanks Ed!

PR #4684, opened 2023-02-23
[PJRT] Use the new local rank env var to calculate local_ordinal

will-cromar:
Fix some assertions that erroneously let the unit tests pass, and enable more unit tests on the TPU CI to catch bugs like this in the future.


Liyang90 (approved):
LGTM

PR #4697, opened 2023-02-27
Fix dynamo test no op

JackCaoG:
Currently all of the `torch.allclose` in the `test_dynamo.py` is no op since `torch.allclose` returns a bool instead of fail when it failed to match two inputs. Correct the test, this pr need to be backported to 2.0 branch.

In the long term, we should not compare the dynamo result vs cpu result, since with optimization, the result very likely be off. A better approach is to compare the result of dynamo vs result with lazy.

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -31,16 +31,16 @@ class DynamoInferenceBasicTest(unittest.TestCase):
     res_cpu = self.fn_simple(x, y)
     res_xla_dynamo = self.fn_simple_dynamo(xla_x, xla_y)
     self.assertIn('xla::add', met.counter_names())
-    torch.allclose(res_cpu, res_xla_dynamo.cpu())
+    assert torch.allclose(res_cpu, res_xla_dynamo.cpu())
     # verifiy that tracing is skipped in following runs
     met.clear_counters()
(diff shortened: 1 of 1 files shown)


test/dynamo/test_dynamo.py
```diff
@@ -31,16 +31,16 @@ def test_simple_model(self):
     res_cpu = self.fn_simple(x, y)
     res_xla_dynamo = self.fn_simple_dynamo(xla_x, xla_y)
     self.assertIn('xla::add', met.counter_names())
-    torch.allclose(res_cpu, res_xla_dynamo.cpu())
+    assert torch.allclose(res_cpu, res_xla_dynamo.cpu())
```
alanwaketan, line 34: Will it be better to use `self.assert`?

alanwaketan (approved):
LGTM.

test/dynamo/test_dynamo.py
JackCaoG: Let me do that in a follow up pr lol, don't want to rerun ci since I want to backport this to 2.0 ASAP.

steventk-g:
Do we need to patch this into `r2.0`?

JackCaoG:
@steventk-g yes, let me open the pr

PR #4698, opened 2023-02-27
Add test for dynamo and sgd optimizer

JackCaoG:
A lot of issues I run into 2 weeks ago seems to be fixed on pytorch nightly, now I can run optimizer and see it being captured as part of the graph and give correct update to input. Adding a test to make sure this does not regress.

FYI @wconstab


JackCaoG:
Given that I run into 2 weeks ago's pytorch head, I don't think it is a good idea for me to port this pr into 2.0 release branch. I will only talk about fwd and bwd in the upcoming blog post.

alanwaketan (approved):
LGTM.

PR #4703, opened 2023-02-28
Add RoIAlign tests

vanbasten23:
Add a RoIAlign test for dynamic shape inspired by ezyang's post https://dev-discuss.pytorch.org/t/a-pure-python-implementation-of-roi-align-that-looks-just-like-its-cuda-kernel/1266.

The test doesn't crash but generate different result on xla devices compared with running on CPU.

Will try to merge the pr first so other changes won't make the RoIAlign test crash. Will try to figure out what cause it generate different results next.

Note: per ROIAlign [cuda implementation](pytorch/vision/blob/9b7c7d395d44832104b583e82dfc5c5272535ebb/torchvision/csrc/ops/cuda/roi_align_kernel.cu#L14-L143 22), it doesn't seem to involve dynamism. The [IR graph](https://gist.github.com/vanbasten23/7b43a87c6b9f6e415fa9bffcb6e9f5f9) demonstrates it.

What's strange is that the RoIAlign test succeeds on TPU v3 but not on TPU v4.

cc @miladm @JackCaoG


vanbasten23:
For the forward pass, after I enabled logging (which can be found [here](https://gist.github.com/vanbasten23/992ab2f90bc9b5262f68d1bfbdfe0caf)), it appears the program hangs at 
```
I0228 22:24:55.053719  273468 2a886c8_compiler_base.cc:4179] XLA::TPU running hlo passes for 460,600 instructions, modules: SyncTensorsGraph.460601
I0228 22:24:55.061100  273468 2a886c8_compiler_base.cc:4230] Initial HLO module: SyncTensorsGraph.460601 instructions: 460,600 fingerprint: 
I0228 22:25:00.518923  273468 2a886c8_compiler_base.cc:4279] HLO optimizing module: SyncTensorsGraph.460601 instructions: 531,832
I0228 22:25:00.519016  273468 2a886c8_compiler_base.cc:4294] XLA::TPU HLO optimization
 
I0228 22:27:54.404236  273468 2a886c8_compiler_base.cc:3767] XLA::TPU HLO PostOptimizationPipeline
E0228 22:30:34.215286  274750 slow_operation_alarm.cc:65] 
********************************
[Run backend for module group] Very slow compile?  If you want to file a bug, run with envvar XLA_FLAGS=--xla_dump_to=/tmp/foo and attach the results.
********************************
```
So I followed the prompt and reran the program with `XLA_FLAGS=--xla_dump_to=/tmp/foo` and got the xla dump

JackCaoG:
Seems like it is just a super slow compilation, you can dump the HLO and open a bug to XLA tema, Through I think we won't use the native roialign anyway so I am not sure how relevant is this bug. Can you try our custom roialign?

vanbasten23:
Thanks for the reply. By `native roialign`, you meant the torchvision's version of RoIAlign? Where can I find our custom roialign?

vanbasten23:
Updates:

1. For RoIAlign forward test, I was able to make it run to completion when DS is enabled. See [commit](https://github.com/pytorch/xla/pull/4703/commits/b04dcd94c625292ebc8ab74889d3d51325a05f1e). So I got all xla op involved and summarized in this [issue](https://github.com/pytorch/xla/issues/4707). Current issue is the RoIAlign forward result on TPU is different from the result on CPU so the test would fail.

2. The slow compilation ticket I filed b/271173239 got resolved. It looks there's a bug in XLA:TPU. In the next tf pin update, we will get their fix.

I'm also trying to answer the following questions:

1. Where does the dynamism come to the picture in the RoIAlign forward test?
I got the IR [dump](https://gist.github.com/vanbasten23/e0abc0e941ce0340d4d19226920bd2c2). I then modified the input data by reducing one data from `rois = torch.tensor([[0, 0, 0, 9, 9], [0, 0, 5, 4, 9], [0, 5, 5, 9, 9], [1, 0, 0, 9, 9],...)` to `rois = torch.tensor([[0, 0, 0, 9, 9], [0, 0, 5, 4, 9], [0, 5, 5, 9, 9],...)`. The IR dump becomes [this](https://gist.github.com/vanbasten23/8d757cf7b0c0399d9df9312a28a4ba46). So I don't think dynamism is even involved in the ROIAlign forward pass.

1. Where does the code run after this line https://github.com/pytorch/vision/blob/e3da44bb02cbcbb4e7269606905590b754b712e8/torchvision/csrc/ops/roi_align.cpp#L25?
It's /workspaces/work/pytorch/vision/torchvision/csrc/ops/cpu/roi_align_kernel.cpp, line=25.

vanbasten23:
I experimented Ed's RoiAlign implementation in this PR pytorch/xla/test/ds/roiAlign.py. When running, it failed at line
```
pooled_features = roi_align(features, rois, spatial_scale, output_size[0], output_size[1], -1, False)
```
with error:
```
terminate called after throwing an instance of 'c10::Error'
  what():  The operator aten::as_strided appears to be a view operator, but it has no implementation for the backend "xla:0". View operators don't support falling back to run on the CPU, since the tensor's storage cannot be shared across devices.
```
 ([full error](https://gist.github.com/vanbasten23/7e7e719bef42bc4b4d1f70fcbea727c1))

If I use the torchvision's implementation (the same file pytorch/xla/test/ds/roiAlign.py at the end), it fails with [error](https://gist.github.com/vanbasten23/26eb6887a5cd78e99cb2153475285116):
```
NotImplementedError: Could not run 'torchvision::roi_align' with arguments from the 'CPU' backend. This could be because the operator doesn't exist for this backend, or was omitted during the selective/custom build process (if using custom build).
```

vanbasten23:
After rebase, the cpp RoIAlign implementation runs without crashing.

vanbasten23:
TPU CI is failing with torchvision not found ([error](https://gist.github.com/vanbasten23/c450e30ffbba59df18593f1bcb03406f)). I've added torchvision installation in the tpu ci image.

test/ds/test_dynamic_shape_models.py
```diff
@@ -155,6 +155,35 @@ def create_dynamic_test_data(self,
...
+                        dtype=torch.float,
+                        device=xla_dev)
+
+    # Set output size and spatial scale
+    output_size = (7, 7)
+    spatial_scale = 1.0 / 16.0
+
+    from torchvision.ops import roi_align as roi_align_torchvision
+    print(
+        roi_align_torchvision(features, rois, output_size, spatial_scale).sum())
```
JackCaoG, line 179: why print here?

test/ds/test_dynamic_shape_models.py
```diff
@@ -155,6 +155,35 @@ def create_dynamic_test_data(self,
     y_test_nonzero_dev = torch.nonzero(y_test_xla.int()).float().squeeze()
     return x_test_nonzero_dev, y_test_nonzero_dev
 
+  def test_roialign(self):
```
JackCaoG, line 158: what's this test trying to test? running `roi_align_torchvision` will not crash?

vanbasten23: yes, that's the first step. By running this in our CI, we can make sure all other changes (pytorch, xla, runtime) won't break it (it won't crash.)

Correctness and number of compilation test will come next.

test/ds/test_dynamic_shape_models.py
vanbasten23: It'll execute the HLO graph and show the result. Later when we check the correctness, we'll compare it with the aten result.

infra/ansible/roles/build_srcs/tasks/main.yaml
```diff
@@ -17,6 +17,24 @@
     name: "{{ pytorch_wheels.files | map(attribute='path') }}"
     state: "forcereinstall"
 
+- name: Build torchvision
```
will-cromar, line 20: Do you need to build torchvision from source? Or can you use the existing nightly CPU build of torchvision?

e.g. https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/250015f2394ae38e4f3ec4fa150ba33e72f237a7/tests/pytorch/nightly/common.libsonnet#L103-L104

vanbasten23: Right, building torchvision from source may not be a good idea. Removed the change.

vanbasten23:
As discussed in the meeting, the roiAlign implementation doesn't involve dynamism so I'll close this pr now.

PR #4709, opened 2023-03-01
[PJRT] Remove start method from `_run_singleprocess`

will-cromar:
Since `_run_singleprocess` doesn't spawn anything, it doesn't have a `start_method`

Enable more unit tests that would have caught this bug in CI.


JackCaoG (approved):
Thanks!

alanwaketan:
Late to the party but thanks Will!

PR #4710, opened 2023-03-01
Lower UpsampleBilinear/Nearest2DBackward with scale factor on TPU

lsy323:
Currently, the following cases upsample_nearest2d_backward won't be lowered:

1. Device is not TPU (see comment at: https://github.com/pytorch/xla/blob/5b5dd27f20ec80706e2d88f59b1826a5924c66f2/torch_xla/csrc/aten_xla_type.cpp#L2990)
2. The upsampling has scaler not equal to 1

In this PR
- The 2nd case above will be lowered on TPU.
- Add tests

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -2951,16 +2951,26 @@ at::Tensor XLANativeFunctions::upsample_bilinear2d_backward(
   // our XLA lowering.
   XlaDeviceType hw_type =
       static_cast<XlaDeviceType>(grad_output_tensor->GetDevice().type());
-  if (hw_type != XlaDeviceType::TPU || (scales_h && *scales_h != 1.0) ||
-      (scales_w && *scales_w != 1.0)) {
+  if (hw_type != XlaDeviceType::TPU) {
     return at::native::call_fallback_fn<
         &xla_cpu_fallback,
         ATEN_OP(upsample_bilinear2d_backward)>::call(grad_output, output_size,
                                                      input_size, align_corners,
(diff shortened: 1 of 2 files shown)


vanbasten23:
Is there an associated gh issue?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2951,16 +2951,26 @@ at::Tensor XLANativeFunctions::upsample_bilinear2d_backward(
...
                                                      input_size, align_corners,
                                                      scales_h, scales_w);
   }
+  std::vector<int64_t> scaled_output_size =
+      torch::lazy::ToVector<int64_t>(output_size);
+  if ((scales_h && *scales_h != 1.0) || (scales_w && *scales_w != 1.0)) {
+    scaled_output_size = GetOutputSizeWithScale(input_size, scales_h, scales_w,
+                                                scaled_output_size);
+    if (!output_size.empty()) {
+      XLA_CHECK(scaled_output_size.at(0) == output_size.at(0) &&
```
vanbasten23, line 2967: nit: do you need to test the error case?

JackCaoG (approved):

torch_xla/csrc/aten_xla_type.cpp
lsy323: The torch.nn Python API requires either output shape or scale factor is specified, so there is already a guard for the inconsistency between the output shape and scale factor. This is more like a 2nd guard.
ref: https://pytorch.org/docs/stable/generated/torch.nn.UpsamplingNearest2d.html

PR #4717, opened 2023-03-03
Move dependencies on Tensorflow under third_party/xla_client and torch_xla/csrc to use TSL where possible

ddunl:
The goal of this PR is to make the move from TF to OpenXLA easier. Now the only remaining dependencies appear to be XRT or necessary because XRT.


ddunl:
I see that I missed a file with clang-format, I'll rectify that tomorrow morning.

JackCaoG (approved):
This is great! Thanks @ddunl !

alanwaketan (approved):
LGTM. Thanks, David.

PR #4718, opened 2023-03-03
Enable PjRt multi-GPU tests in CI

jonb377:
GPU test failures seem to be due to shm size. Using this PR to identify what changes need to happen in CI for tests to pass.


will-cromar (approved):
Thanks!

PR #4720, opened 2023-03-03
[PJRT] Detect number of local TPU chips

will-cromar:
Remove hardcoded `local_chips = 4`. Instead, check `/dev/` to see how many TPU devices there actually are to decide how many processes need to be created on the current host.

Tweak `tpu.process_bounds_size` to return `None` when no process bounds are set yet.

### DIFF
--- a/torch_xla/experimental/tpu.py
+++ b/torch_xla/experimental/tpu.py
@@ -72,7 +72,7 @@ def _get_metadata(key: str) -> str:
 
 
 def process_bounds_size() -> Optional[int]:
-  """Returns number of processes across all TPU hosts."""
+  """Returns number of processes across all TPU hosts, or None if unknown."""
   process_bounds = xu.getenv_as(xenv.TPU_PROCESS_BOUNDS, str)
   return MeshShape.from_string(process_bounds).size if process_bounds else None
 
--- a/torch_xla/experimental/tpu.py
+++ b/torch_xla/experimental/tpu.py
@@ -1,4 +1,5 @@
 import functools
+import glob
 import operator
 import os
 import re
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/tpu.py
```diff
@@ -70,17 +71,24 @@ def _get_metadata(key: str) -> str:
...
   process_bounds = xu.getenv_as(xenv.TPU_PROCESS_BOUNDS, str)
-  return MeshShape.from_string(
-      process_bounds).size if process_bounds else default
+  return MeshShape.from_string(process_bounds).size if process_bounds else None
 
 
-def num_local_processes(local_chips: int = 4) -> int:
+def num_available_chips() -> int:
+  """Returns the number of local chips in /dev/"""
+  return len(glob.glob('/dev/accel?'))
```
JackCaoG, line 82: what's gonna happen if it is not on a TPU device?

will-cromar: `tpu.num_available_chips` will return `0` in that case

alanwaketan: Seems a bit hacky? Is there an API to allow you do that? The `/dev/accel?` could change or could be non-TPU.

JackCaoG (approved):

ManfeiBai (approved):

torch_xla/experimental/tpu.py
will-cromar: You can query PJRT for the number of devices, but we actually have to run this check before we initialize PJRT, since it tells us how many PJRT clients to create.

I agree this is a little hacky, but it helps us handle the case where you pass through one TPU device to a container out of multiple, compared to creating a mapping between TPU types and local chips. FWIW, `/dev/accel*` is hardcoded  within `libtpu` and the TPU driver as well, so I don't think that convention is likely to change without us knowing.

torch_xla/experimental/tpu.py
```diff
@@ -70,17 +71,24 @@ def _get_metadata(key: str) -> str:
...
+  """Returns number of processes across all TPU hosts, or None if unknown."""
   process_bounds = xu.getenv_as(xenv.TPU_PROCESS_BOUNDS, str)
-  return MeshShape.from_string(
-      process_bounds).size if process_bounds else default
+  return MeshShape.from_string(process_bounds).size if process_bounds else None
 
 
-def num_local_processes(local_chips: int = 4) -> int:
+def num_available_chips() -> int:
+  """Returns the number of local chips in /dev/"""
```
vanbasten23, line 81: qq: does it apply to v3/v2? Asking it because iirc there are 4 chips on a v3-8 donut but we can see /dev/accel0 up to /dev/accel7.

will-cromar: Actually, each `/dev/accel*` represents one chip, not one core, so only 4 devices will show up in `/dev/`. For example, on a TPU v2:

```
$ ls /dev/accel*
/dev/accel0  /dev/accel1  /dev/accel2  /dev/accel3
```

PR #4696, opened 2023-02-27
Preserve parameter sharding & propagate to data nodes

yeounoh:
This addresses the model parameter sharding bug, where the sharded model parameter inputs are replicated after execution. In addition to preserving the sharding annotation, we also need to shard the input data nodes, as they are replicated after execution/update.

- Introduce `ShardingUtil::ShardInputDataNodes` for propagating sharding to input data nodes
- Introduce `ShardingUtil::TensorShardingStore` to caching per tensor sharding annotation information
- Add `XLAShardingTest.TensorShardingStore` test
- Add `XLAShardingTest.ShardInputDataNodes` test
- Extend `BasicShardingTest.test_optimizer_step_with_sharding` test with `ShardInputDataNodes` met counter check.

This is tested with & without functionalization pass. I am also landing my  code to propagate sharding annotation between input and output nodes, that can be removed after we land the functinolization code.

PR #4721, opened 2023-03-03
[SPMD] Preserve parameter sharding with output data sharding

yeounoh:
This addresses the same problem as in #4696 with an alternative solution. We shard the replicated output while handling the computation results. This avoids post traversal pass to replace original data node with a sharded one, thus more efficient. Key changes include:
- Introduce `ShardingUtil::OutputHandler`
- Add `XLAShardingTest.OutputHandler` test for unit testing, `test_optimizer_step_with_sharding` checks the validity of the change with a simple e2e example already.
- Add `std::optional<xla::Shape>` to `ShardingSpec`
- Add `std::optional<xla::OpSharding>`  to `PjRtShardedData`
- Pass an additional `std::vector<XLATensor::ShardingSpecPtr>` param to `XLAGraphExecutor::ScheduleSyncTensorsGraph`, since the async function now calls `ShardingUtil::OutputHandler`
- Introduce & call `XLAGraphExecutor::CollectShardingSpecs` before calling `ScheduleSyncTensorsGraph`
- Introduce `WrapDataShards` and `GetDataSharding` APIs in `ComputationClient`.

### DIFF
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -160,9 +160,9 @@ TEST_F(XLAShardingTest, EqualShardingSpecs) {
       xla::HloSharding::Tile({{{0, 1}, {2, 3}, {4, 5}, {6, 7}}}).ToProto());
   XLATensor::ShardingSpec replicated(xla::HloSharding::Replicate().ToProto());
   EXPECT_TRUE(ShardingUtil::EqualShardingSpecs(tiled_2d, tiled_2d));
-  EXPECT_TRUE(!ShardingUtil::EqualShardingSpecs(tiled_2d, tiled_3d));
+  EXPECT_FALSE(ShardingUtil::EqualShardingSpecs(tiled_2d, tiled_3d));
   EXPECT_TRUE(ShardingUtil::EqualShardingSpecs(replicated, replicated));
-  EXPECT_TRUE(!ShardingUtil::EqualShardingSpecs(tiled_2d, replicated));
+  EXPECT_FALSE(ShardingUtil::EqualShardingSpecs(tiled_2d, replicated));
 }
 
 TEST_F(XLAShardingTest, CreateTensorsData) {
@@ -275,8 +275,8 @@ TEST_F(XLAShardingTest, OutputHandler) {
   auto shards =
       xla::ComputationClient::Get()->GetDataShards(sharded_outputs[0]);
   EXPECT_EQ(shards.size(), devices.size());
-  EXPECT_TRUE(
-      !xla::Shape::Equal().IgnoreLayout()(shards[0]->shape(), tensor_shape));
+  EXPECT_FALSE(
+      xla::Shape::Equal().IgnoreLayout()(shards[0]->shape(), tensor_shape));
 
   // Wrap sharded data into a PjRtShardedData with `devices.size()` shards.
   std::vector<xla::ComputationClient::DataPtr> wrapped_outputs =
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -239,5 +239,55 @@ TEST_F(XLAShardingTest, InputHandler) {
                                  at::kFloat));
 }
 
+TEST_F(XLAShardingTest, OutputHandler) {
+  if (xla::sys_util::GetEnvString(xla::env::kEnvPjRtDevice, "") == "") {
+    GTEST_SKIP() << "`PJRT_DEVICE` is not set.";
+  }
+
+  std::vector<std::string> devices =
+      xla::ComputationClient::Get()->GetLocalDevices();
+
+  // Prepare an input vecotr `outputs` with 2 arguments per device.
+  std::vector<std::vector<xla::ComputationClient::DataPtr>> outputs;
+  outputs.reserve(devices.size());
+  at::Tensor tensor = at::ones({8}, at::TensorOptions(at::kFloat));
+  for (auto device : devices) {
+    outputs.push_back(
+        UnwrapXlaData(CreateTensorsData({tensor, tensor}, {device, device})));
+  }
+
+  xla::Shape tensor_shape =
+      CreateComputationShapeFromTensor(tensor, GetDefaultDevice());
+  auto sharding_spec = std::make_shared<XLATensor::ShardingSpec>(
+      xla::HloSharding::Tile1D(
+          CreateComputationShapeFromTensor(tensor, GetDefaultDevice()),
+          devices.size())
+          .ToProto(),
+      tensor_shape);
+  std::vector<XLATensor::ShardingSpecPtr> sharding_specs{sharding_spec,
+                                                         sharding_spec};
+
+  // Shard a PjRtData into a PjRtShardedData.
+  std::vector<xla::ComputationClient::DataPtr> sharded_outputs =
+      ShardingUtil::OutputHandler(outputs, sharding_specs,
+                                  /*replicated_output=*/true);
+  EXPECT_EQ(sharded_outputs.size(), 2);
+  auto shards =
+      xla::ComputationClient::Get()->GetDataShards(sharded_outputs[0]);
+  EXPECT_EQ(shards.size(), devices.size());
+  EXPECT_TRUE(
+      !xla::Shape::Equal().IgnoreLayout()(shards[0]->shape(), tensor_shape));
+
+  // Wrap sharded data into a PjRtShardedData with `devices.size()` shards.
+  std::vector<xla::ComputationClient::DataPtr> wrapped_outputs =
+      ShardingUtil::OutputHandler(outputs, sharding_specs,
+                                  /*replicated_output=*/false);
+  EXPECT_EQ(wrapped_outputs.size(), 2);
+  shards = xla::ComputationClient::Get()->GetDataShards(wrapped_outputs[0]);
+  EXPECT_EQ(shards.size(), devices.size());
+  EXPECT_TRUE(
+      xla::Shape::Equal().IgnoreLayout()(shards[0]->shape(), tensor_shape));
+}
+
 }  // namespace cpp_test
 }  // namespace torch_xla
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -240,8 +240,12 @@ TEST_F(XLAShardingTest, InputHandler) {
 }
 
 TEST_F(XLAShardingTest, OutputHandler) {
-  if (xla::sys_util::GetEnvString(xla::env::kEnvPjRtDevice, "") == "") {
-    GTEST_SKIP() << "`PJRT_DEVICE` is not set.";
+  if ((xla::sys_util::GetEnvString(xla::env::kEnvPjRtDevice, "") == "") ||
+      (xla::ComputationClient::Get()->GetLocalDevices().size() < 2)) {
+    GTEST_SKIP()
+        << "`PJRT_DEVICE` is not set, with more than 2 local devices, ("
+        << xla::ComputationClient::Get()->GetLocalDevices().size()
+        << " local devices detected).";
   }
 
   std::vector<std::string> devices =
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -354,6 +354,9 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
     if (instance.is_sharded) {
       // TODO(yeounoh) multi-host, multi-slice configurations
       compile_options.executable_build_options.set_use_spmd_partitioning(true);
+      // TODO(yeounoh) this is set to false by default, but explicitly set here
+      // to expose the knob for future reference. We can override the compiler's
+      // default behavior to further optimize parameter sharding in the future.
       compile_options.executable_build_options
           .set_allow_spmd_sharding_propagation_to_output({false});
       compile_options.executable_build_options.set_num_partitions(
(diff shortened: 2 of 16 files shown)


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -332,6 +354,8 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
     if (instance.is_sharded) {
       // TODO(yeounoh) multi-host, multi-slice configurations
       compile_options.executable_build_options.set_use_spmd_partitioning(true);
+      compile_options.executable_build_options
+          .set_allow_spmd_sharding_propagation_to_output({false});
```
JackCaoG, line 358: can you add a comment here to explain why it is needed?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -507,7 +533,8 @@ PjRtComputationClient::ExecuteReplicated(
     data_handles.push_back(datas);
   }
 
-  TF_VLOG(1) << "Returning " << data_handles.size() << " sets of results";
+  TF_VLOG(1) << "Returning " << data_handles.size() << " sets of results "
```
JackCaoG, line 536: This will be quite verbose, no? Maybe make it a 2 or 3? IMO the 1 should only contains the critial messages like "I init", "I execute". Details debug information should be in higher levels.

third_party/xla_client/pjrt_computation_client.h
```diff
@@ -193,10 +202,10 @@ class PjRtComputationClient : public ComputationClient {
       return true;
     }
 
-    xla::OpSharding GetSharding() { return sharding; }
+    std::optional<xla::OpSharding> GetSharding() { return sharding; }
 
     std::vector<std::shared_ptr<PjRtData>> shards;
-    xla::OpSharding sharding;
+    std::optional<xla::OpSharding> sharding;
```
JackCaoG, line 208: hmm, why do we make it optional?

torch_xla/csrc/tensor.h
```diff
@@ -259,8 +259,11 @@ class XLATensor : public torch::lazy::LazyTensor {
   // HloSharding for replication, manual and tile shardings.
   struct ShardingSpec {
     ShardingSpec(const xla::OpSharding& sharding) : sharding(sharding) {}
+    ShardingSpec(const xla::OpSharding& sharding, const xla::Shape& shape)
+        : sharding(sharding), shape(shape) {}
 
     xla::OpSharding sharding;
+    std::optional<xla::Shape> shape;
```
JackCaoG, line 266: what;s this shape for?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1519,12 +1519,15 @@ void InitXlaModuleBindings(py::module m) {
                                  const py::list& tile_assignment,
                                  bool replicated = false, bool manual = false) {
     TORCH_LAZY_COUNTER("XlaMarkSharding", 1);
+    XLATensorPtr xtensor = bridge::GetXlaTensor(input);
+    XLAGraphExecutor::Get()->UnregisterTensor(xtensor->data().get());
```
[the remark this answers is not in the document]
JackCaoG: I think this unregister approach will conflict with the Clone pr @steventk-g is working on.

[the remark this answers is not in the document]
JackCaoG: So what will happen if we don't register, will compiler returns us 2 duplicate output?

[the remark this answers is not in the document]
yeounoh: The node hashing work is to avoid hash collision at the node level; this helps us from confusing two computation graphs, sharded and unsharded, with the same hash.

This is to sanitize sync tensors and the parameters lists. The registration happens at the device level, and we register one tensor to one XLATensor. If we don't register.. then I think we are still associating the unsharded XLATensor to the input tensor, which is not correct. And we can't or should be careful trying to associate two XLATensors to a tensor... let's talk more, I feel like there is much cost to our goal.

torch_xla/csrc/tensor.h
yeounoh: Original shape, let me comment.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Ack.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Ack.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Actually, this will be not that lengthy/verbose, and it's on par with unsharded computation logging: https://github.com/pytorch/xla/blob/a6d186f7c6f27b94d6357b4d7ef50683b10612d7/third_party/xla_client/pjrt_computation_client.cc#L446

third_party/xla_client/pjrt_computation_client.h
yeounoh: Unwrapped the optionality

torch_xla/csrc/init_python_bindings.cpp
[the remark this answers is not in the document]
yeounoh: Synced offline, we will register, but not unregister to rely on the reference counter.

torch_xla/csrc/tensor.h
yeounoh: Done

test/cpp/test_xla_sharding.cpp
```diff
@@ -239,5 +239,55 @@ TEST_F(XLAShardingTest, InputHandler) {
...
+
+  // Shard a PjRtData into a PjRtShardedData.
+  std::vector<xla::ComputationClient::DataPtr> sharded_outputs =
+      ShardingUtil::OutputHandler(outputs, sharding_specs,
+                                  /*replicated_output=*/true);
+  EXPECT_EQ(sharded_outputs.size(), 2);
+  auto shards =
+      xla::ComputationClient::Get()->GetDataShards(sharded_outputs[0]);
+  EXPECT_EQ(shards.size(), devices.size());
+  EXPECT_TRUE(
```
JackCaoG, line 278: nit is there a EXPECT_FALSE ? `EXPECT_TRUE (!)` seems a bit awkward

third_party/xla_client/pjrt_computation_client.cc
JackCaoG: Did I miss the comment somewhere?

third_party/xla_client/xrt_computation_client.h
```diff
@@ -248,10 +248,18 @@ class XrtComputationClient : public ComputationClient {
...
-  std::vector<DataPtr> GetDataShards(DataPtr data) override {
+  std::vector<DataPtr> GetDataShards(DataPtr data) override { return {data}; }
+
+  DataPtr WrapDataShards(const std::vector<DataPtr>& shards, std::string device,
+                         xla::Shape shape, xla::OpSharding sharding) override {
     XLA_ERROR() << __FUNCTION__ << " not implemented";
   }
 
+  std::optional<xla::OpSharding> GetDataSharding(DataPtr handle) override {
+    // Returns an empty sharding result, ad XRT does not support sharding.
```
JackCaoG, line 259: is `ad XRT does not` a typo?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -227,6 +228,47 @@ ShardingUtil::InputHandler(
...
+  std::vector<xla::ComputationClient::DataPtr> outputs;
+  outputs.reserve(sharding_specs.size());
+  for (int i = 0; i < sharding_specs.size(); ++i) {
+    XLATensor::ShardingSpecPtr sharding = sharding_specs[i];
+    if (sharding &&
+        (sharding->sharding.type() != xla::OpSharding::REPLICATED)) {
+      if (replicated_output) {
+        // Reshards replicated output if `sharding` is present.
+        std::vector<at::Tensor> tensors =
+            XlaDataToTensors({WrapXlaData(sharded_results[0][i])}, at::kFloat);
```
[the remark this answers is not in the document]
JackCaoG: I felt like it should be actual result type, instead of kFloat.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -898,32 +927,39 @@ XLAGraphExecutor::ScheduleSyncTensorsGraph(
...
                 UnwrapXlaData(async->parameters_data), devices);
         xla::ComputationClient::ExecuteReplicatedOptions execute_options;
         TF_VLOG(3) << "Executing IR graph hash "
                    << torch::lazy::HashToString(hash)
                    << " on devices: " << absl::StrJoin(devices, ",");
-        results = WrapXlaData(xla::ComputationClient::Get()->ExecuteReplicated(
-            *async->cached_computation->computation->client_computation(),
-            device_arguments, devices,
-            execute_options)[0]);  // TODO(yeounoh) assumes replicated outputs
+        // OutputHandler creates sharded data for for sharded
```
JackCaoG, line 952: typo `for for`

torch_xla/csrc/xla_sharding_util.h
```diff
@@ -46,7 +46,23 @@ class ShardingUtil {
...
+  // leaves the results in replicated states, which is aligned with the default
+  // exepctation `replicated_output=true`. However, if we override the
+  // compiler's default behavior and allow the execution to return sharded
+  // results, then we should set `replicated_output=false` and wrap sharded
+  // arguments into `PjRtShardedData`. This returns a vector of size that is
+  // equal to the number of arguments.
+  static std::vector<xla::ComputationClient::DataPtr> OutputHandler(
+      std::vector<std::vector<xla::ComputationClient::DataPtr>> sharded_results,
+      std::vector<XLATensor::ShardingSpecPtr> sharding_specs,
+      bool replicated_output = true);
```
JackCaoG, line 63: are we ever going to need `replicated_output=false`?

yeounoh: Yes, when we change this,
```
// TODO(yeounoh) this is set to false by default, but explicitly set here
// to expose the knob for future reference. We can override the compiler's
// default behavior to further optimize parameter sharding in the future.
compile_options.executable_build_options
          .set_allow_spmd_sharding_propagation_to_output({false});
```

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: Thanks!

third_party/xla_client/xrt_computation_client.h
yeounoh: Done

torch_xla/csrc/xla_sharding_util.cpp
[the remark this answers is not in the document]
yeounoh: Ack, good catch!

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Done

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -831,6 +848,18 @@ std::vector<torch::lazy::BackendDataPtr> XLAGraphExecutor::SetTensorData(
       tensor->data()->view = nullptr;
       tensor->data()->tensor_data = c10::nullopt;
     }
+    // Create sharded data placeholder, this will be used to
+    // hold the corresponding computation results.
+    if (tensor->sharding_spec()) {
```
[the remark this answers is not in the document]
yeounoh: Sounds good, done.

test/cpp/test_xla_sharding.cpp
yeounoh: Sounds good, done.

JackCaoG:
`OutputHandler` seems to crash

yeounoh:
> `OutputHandler` seems to crash

Yea, we need at least 2 devices to create Hlo sharding. 
```
2023-03-07 21:01:25.297118: F external/org_tensorflow/tensorflow/compiler/xla/hlo/ir/hlo_sharding.cc:54] Check failed: num_tiles > 1 (1 vs. 1)
```
Added the safeguard.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -856,6 +873,18 @@ void XLAGraphExecutor::ExtractIRAndPrepareXlaData_(
...
+      auto sharding = tensor->sharding_spec();
+      if (!sharding->shape.has_value()) {
+        sharding->shape = tensor->shape();
+      }
+      handle = WrapXlaData(xla::ComputationClient::Get()->WrapDataShards(
+          {UnwrapXlaData(handle)}, GetVirtualDevice().toString(),
+          sharding->shape.value(), sharding->sharding));
+      tensor->data()->handle = handle;
```
JackCaoG, line 886: do we need to do it here? I thought we just need to push it to `tensor_data_vec` and it will get assigned later.

JackCaoG: I think ti is being down in 
https://github.com/pytorch/xla/blob/50141f2e19cedab7121d34f0921694402d686d9d/torch_xla/csrc/xla_graph_executor.cpp#L824-L827

so you can skip it here.

yeounoh: Good catch, done.

JackCaoG (approved):
Thanks!

jonb377 (approved):
LGTM, thanks!

PR #4723, opened 2023-03-06
[PJRT] Replace XRT CI tests with PJRT

will-cromar:


jonb377 (approved):
LGTM, thanks Will!

test/run_tests.sh
```diff
@@ -94,30 +95,42 @@ function run_xla_backend_mp {
...
     echo "the tests need atleast two XLA workers to validate"
   fi
 }
 
+function run_xrt_tests {
+  echo "Running XRT tests"
+  run_opbyop python3 "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
+  run_async_scalar python3 "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
+  run_xrt python3 "$CDIR/test_torch_distributed_xla_backend.py"
+  run_xrt python3 "$CDIR/test_mp_rendezvous.py"
```
JackCaoG, line 130: do we know why these two test failed on PJRT?

JackCaoG (approved):
Thanks @will-cromar ! This is exciting!

test/run_tests.sh
will-cromar: Rendezvous uses replica groups, which we don't support with the new rendezvous implementation.

The distributed backend test relies on some implementation details of XRT that don't apply to PJRT. I can create a PJRT version, since `torch.distributed` largely works.

PR #4724, opened 2023-03-07
Catch info from metadata

ManfeiBai:
Update code to catch info from metadata

### DIFF
--- a/torch_xla/experimental/tpu.py
+++ b/torch_xla/experimental/tpu.py
@@ -1,10 +1,8 @@
 import functools
-import glob
 import operator
 import os
 import re
 from typing import Dict, NamedTuple, Optional, List, Tuple
-from typing_extensions import TypedDict
 import requests
 import yaml
 
@@ -31,14 +29,8 @@ _ACCELERATOR_TYPE_TO_HOST_BOUNDS = {
     'v3-1024': '8,16,1',
     'v3-2048': '16,16,1',
     # Get v4 host bounds from TPU metadata
+    # Get v5 host bounds from TPU metadata
 }
-
-
-class TpuEnv(TypedDict):
-  accelerator_type: str
-  tpu_process_bounds: str
-  tpu_chips_per_process_bound: str
-  worker_id: int
 
 
 class MeshShape(NamedTuple):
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/tpu.py
```diff
@@ -172,7 +172,7 @@ def configure_topology(local_rank: int,
   tpu_env = get_tpu_env()
 
   accelerator_type = tpu_env[xenv.ACCELERATOR_TYPE]
-  if version() == 4:
+  if tpu_env['ACCELERATOR_TYPE'].startswith('v4') or tpu_env['ACCELERATOR_TYPE'].startswith('v5'):
```
JackCaoG, line 175: @will-cromar what's the right place to add a test for this one? I felt like we had similar tests somwehre.

will-cromar: The relevant unit test is here: https://github.com/pytorch/xla/blob/18674daf0c44bef0162cbeddaaef7acc206ecea5/test/pjrt/test_experimental_tpu.py#L90

Can we just do `tpu.version() >= 4` here?

ManfeiBai: tried version() >=4, failed with https://paste.googleplex.com/5365404993060864

torch_xla/experimental/tpu.py
```diff
@@ -127,7 +127,7 @@ def version() -> int:
   except requests.HTTPError as e:
     raise EnvironmentError('Failed to get TPU metadata') from e
 
-  match = re.match(r'^v(\d)([A-Za-z]?){7}-(\d+)$', env[xenv.ACCELERATOR_TYPE])
```
suexu1025, line 130: need to be `([A-Za-z]?){7}-(\d+)$` according to previous change # 96fd3ce

suexu1025 (changes requested):
one question regarding #ACCELERATOR_TYPE match

torch_xla/experimental/tpu.py
ManfeiBai: r'^v(\d)([A-Za-z]?){7}-(\d+)$' could match v0abcdefg;

and we also have v0abcd;

r'^v(\d).*' match v0abcd and v0abcdefg

ManfeiBai: maybe we could try r'^v(\d).*-(\d+)$'

yeounoh (approved):
LGTM

torch_xla/experimental/tpu.py
```diff
@@ -127,8 +127,7 @@ def version() -> int:
...
-  return int(match.groups()[0])
+  return int((env[xenv.ACCELERATOR_TYPE])[1])
```
will-cromar, line 130: IMO the regex here is clearer. Do you have a reason to change it? The existing one works in all of the test cases:

```
>>> re.match(r'^v(\d)([A-Za-z]?){7}-(\d+)$', 'v5abcdefg-4').groups()[0]
'5'
>>> re.match(r'^v(\d)([A-Za-z]?){7}-(\d+)$', 'v5abcd-4').groups()[0]
'5'
>>> re.match(r'^v(\d)([A-Za-z]?){7}-(\d+)$', 'v5-4').groups()[0]
'5'
```

will-cromar (changes requested):

torch_xla/experimental/tpu.py
ManfeiBai: updated to `r'^v(\d).*' ` to match `v0abcdefg-higklmnop-0`

torch_xla/experimental/tpu.py
ManfeiBai: Thanks, updated

ManfeiBai: Thanks, Will @will-cromar , after the modification based on your advice, current code has passed the CI tests too, do we want to mark this request change as finished and submit this PR?

will-cromar (approved):
Thanks!

PR #4689, opened 2023-02-23
Fix starting profiler server in resnet training test script

lsy323:
`test/test_train_mp_imagenet.py` fails with `--profile`. This patch fixes the profiler failure on commit `8949dd124d6eb7dfe`. However, both `test_train_mp_imagenet.py` and `test_profile_mp_mnist.py` hangs after rebase.

This diff makes the profiler server starting in `test/test_train_mp_imagenet.py` to be consistent with `test/test_profile_mp_mnist.py`, resolving the `--profile` failure.

PR #4726, opened 2023-03-07
fix starting profiler server (#4689)

JackCaoG:
Backporting this this fix to 2.0


JackCaoG:
Force merge now to unblock rc1 build

Issue #4713, opened 2023-03-02
Support XLA_GPU and CUDA cowork by default


kevint324:
## 🚀 Feature

The feature request is pretty much the same as  this issue [Running regular PyTorch with PT-XLA CUDA Container](https://github.com/pytorch/xla/issues/3052 ).

## Motivation

It is a little bit counter intuititve that XLA CUDA and CUDA cannot coexist.
Usually XLA_GPU users and CUDA users are highly overlapped.
And a very common scenario is to compare performance between the native CUDA mode and XLA mode.
Due to current design limitation we cannot do it within the same docker.

Support XLA_GPU/CUDA cowork will greatly improve PyTorch/XLA's usability and user experience.

## Pitch

1. Turn on CUDA support in PyTorch/XLA docker by default
2. Add certain tests to ensure XLA_GPU and CUDA could work together.

## Additional context

Same confusion raised in this [post](https://discuss.pytorch.org/t/why-use-cuda-must-be-0-when-xla-cuda-1/172692)

> We disabled the pytorch wheel with `USE_CUDA=0` since there was a old bug that we can’t have both pytorch and pytorch/xla both trying to grab the CUDA device. That issue might or might not still be the case.

Same  confusion 2

https://github.com/pytorch/xla/issues/3452

JackCaoG:
Enabling the CUDA build for pytorch should be easy. For moving tensor between CUDA and XLA:GPU I believe @cicirori has some ideas in https://github.com/pytorch/xla/issues/4692

qipengh:
@JackCaoG I also do some benchmark test of performance with XLA_GPU and Torch_CUDA， like #4699 。I see torch_xla does some update code of PJRT after release 1.13, so I want to try new features with r2.0 or master 。And I install torch_xla and torch within GPU docker of CUDA 11.7 as follow: 
```python
$ pip install https://download.pytorch.org/whl/nightly/torch/torch-2.0.0.dev20230226+cu117-cp38-cp38-linux_x86_64.whl
$ pip install https://storage.googleapis.com/tpu-pytorch/wheels/cuda/117/torch_xla-nightly+20230228-cp38-cp38-linux_x86_64.whl
```
But I got error with `_XLAC.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZNK5torch8autograd4Node4nameB5cxx11Ev`

```bash
[...]

PR #4728, opened 2023-03-07
Make pytorch cuda build configureable

JackCaoG:
This is to check if we can build both pytorch and xla with CUDA enabled. If this work, it will solve https://github.com/pytorch/xla/issues/4713

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,6 +8,7 @@ ARG python_version="3.7"
 ARG release_version="nightly"
 ARG xla_branch=""
 ARG example_branch="master"
+ARG pytorch_cuda="0"
 ARG cuda="0"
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -21,7 +22,7 @@ RUN apt-get install -y git sudo python-pip python3-pip
 RUN git clone https://github.com/pytorch/pytorch
 
 # Disable CUDA for PyTorch
```
vanbasten23, line 24: nit: update the comment?

JackCaoG:
seems like build timeout, I retrigger the build

JackCaoG:
Ok, I think what happened is building pytorch with cuda takes a long time, so it exceed the build timeout.. let me try to bump that as well.

PR #4729, opened 2023-03-07
Disable ABI for release

JackCaoG:
ABI should be disabled for release build, check https://github.com/pytorch/xla/blob/r1.13/docker/Dockerfile#L11


JackCaoG:
Release wheel from upstream will have ABI off, hence we also need to do that.

will-cromar (approved):

JackCaoG:
Seems like build started without issue(if ABI build failed, it will fail at the beginning of the pytorch build), I will merge this pr and retrigger build for 2.0.

For future reference, we need https://github.com/pytorch/xla/blob/r2.0/torch_patches/81058.diff to get `ABI=0` to build

PR #4730, opened 2023-03-07
Support tf tf repo macros in external repos.

stgpetrovic:
Support tf tf repo macros in external repos.

Using //tensorflow in a macro makes it unusable in repote repos; using clean_dep which is basically str(label($)) makes it work from wherever it's included.

Copy the `bazel` directory into the docker image so the build process works within it.


vanbasten23:
In the past when we debugged tensorflow related issues, we used to directly modify files in pytorch/xla/third_party/tensorflow. Given we have removed `pytorch/xla/third_party` folder, how can we achieve the same? In the team chat, Will suggested we can add tf patches as workaround. I'm wondering what the long term solution is with bazel.

third_party/xla_client/BUILD
```diff
@@ -11,6 +10,10 @@ load(
     "@org_tensorflow//tensorflow/tsl/platform/default:cuda_build_defs.bzl",
     "if_cuda_is_configured",
 )
+load(
+    "//bazel:tensorflow.bzl",
+    "if_with_tpu_support",
```
vanbasten23, line 15: I wonder how or where the flag `if_true`/`if_false` is passed into this macro.

stgpetrovic: At the call site, see below at [https://github.com/pytorch/xla/blob/0961d43f46a84c53e8a45bc9a583063a40188367/third_party/xla_client/BUILD#L138](usage).

JackCaoG:
I triggered the TPU CI at this pr, let's see if that is going to pass.

third_party/xla_client/BUILD
vanbasten23: Thanks!

JackCaoG:
TPU CI seems to work as well.

PR #4733, opened 2023-03-08
Make tpu ci run test for dynamic shape.

vanbasten23:
Make TPU CI run tests for dynamic shape with `XLA_EXPERIMENTAL="nonzero:masked_select"` and `PJRT_DEVICE=TPU_LEGACY`. 

Since we start running dynamic shape related test on pjrt stream executor instead of xrt stream executor, I don't think the current GPU test would cover the scenario.

### DIFF
--- a/test/tpu/xla_test_job.yaml
+++ b/test/tpu/xla_test_job.yaml
@@ -51,3 +51,21 @@ spec:
     env:
     - name: PJRT_DEVICE
       value: TPU
+  - name: xla-test-tpu-legacy
+    securityContext:
+      privileged: true
+    image: gcr.io/$PROJECT_ID/pytorch-xla-test:$BUILD_ID
+    command:
+    - bash
+    - -c
+    - |
+      python3 pytorch/xla/test/test_dynamic_shape_models.py -v
+      python3 pytorch/xla/test/test_dynamic_shapes.py -v
+    volumeMounts:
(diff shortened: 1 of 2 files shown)


test/tpu/xla_test_job.yaml
```diff
@@ -38,3 +38,21 @@ spec:
     env:
     - name: PJRT_DEVICE
       value: TPU
+  - name: xla-test-tpu-legacy
```
vanbasten23, line 41: hi @will-cromar , do you know any documentation showing me how to write the yaml file?

will-cromar: See the Kubernetes documentation on Pods: https://kubernetes.io/docs/concepts/workloads/pods/

This will very likely fail, since it will run two containers concurrently that are going to try to open the TPU at the same time. Instead, add these scripts to the `xla-test` container and override the env vars as needed like `PJRT_DEVICE=TPU_LEGACY XLA_EXPERIMENTAL=... python3 /path/to/script`. Eventually we'll be able to remove both env vars and this should work by default

vanbasten23: Thanks. Fixed.

will-cromar (approved):
Thanks!

vanbasten23:
> Thanks!

Thanks for the review!

PR #4736, opened 2023-03-08
Add device check in aten::_copy_from

ymwangg:
This is to make the error more informative instead of core dump if a user accidentally copies data from cuda device to xla device.


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -460,6 +460,9 @@ at::Tensor XLANativeFunctions::_copy_from(const at::Tensor& self,
                                           const at::Tensor& dst,
                                           bool non_blocking) {
   TORCH_LAZY_FN_COUNTER("xla::");
+  XLA_CHECK(self.device().type() == c10::kCPU ||
```
JackCaoG, line 463: hmm, so I guess this is useful when someone tries to copy a cuda tensor to xla device?

ymwangg: yes, currently torch_xla will die with core dump due to illegal memory access if a user tries to do so.

JackCaoG (approved):

PR #4743, opened 2023-03-09
[PJRT] Update `torch.distributed` unit test to use PJRT

will-cromar:
Also generally clean up test, including:

- Use `mock.patch` instead of overriding object attributes directly to improve test isolation
- Test user-facing API (`dist.*`)
- Use parameterized test where applicable
- Use unit test assertions where applicable
- Reset IRs instead of running unsupported ops and then catching the exception
- Put common cleanup in `tearDown`

Add explanation for XRT tests remaining in CI. Remove rendezvous test since it's covered for PJRT in `test_mesh_service.py`

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -123,11 +123,10 @@ function run_torchrun {
 }
 
 function run_xrt_tests {
+  # For features not supported in PJRT
   echo "Running XRT tests"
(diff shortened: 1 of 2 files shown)


alanwaketan (approved):
The mock techniques are pretty cool! Thanks for improving the tests.

PR #4747, opened 2023-03-09
Set XLA_DISABLE_FUNCTIONALIZATION to true in xla-dynamo bridge

wonjoolee95:
This is an attempt to remove functionalization code path in dynamo. This does this by introducing a new env var `DISABLE_FUNCTIONALIZATION` that is set/unset in the dynamo bridge. 

On a good note, I can see the `DynamoTrainingBasicTest` that was failing before is now succeeding:
```
(base) jenkins@26d7adccbc26:/workspace/pytorch/xla$ python test/dynamo/test_dynamo.py -k DynamoTrainingBasicTest.test_resnet18
.
----------------------------------------------------------------------
Ran 1 test in 32.409s

OK
(base) jenkins@26d7adccbc26:/workspace/pytorch/xla$ 
```
However, I'm now seeing the Optimizer dynamo test fail to a new reason:
```
ERROR: test_resnet18 (__main__.DynamoTrainingOptimizerTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/dynamo/test_dynamo.py", line 252, in test_resnet18
    xla_output = dynamo_train_model(xla_resnet18, data, target, xla_optimizer)
  File "/opt/conda/lib/python3.8/site-packages/torch/_dynamo/eval_frame.py", line 231, in _fn
    return fn(*args, **kwargs)
  File "test/dynamo/test_dynamo.py", line 199, in train_model
    optimizer.step()
  File "/opt/conda/lib/python3.8/site-packages/torch/optim/optimizer.py", line 33, in _use_grad
    ret = func(self, *args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/torch/optim/sgd.py", line 76, in step
    sgd(params_with_grad,
  File "/opt/conda/lib/python3.8/site-packages/torch/optim/sgd.py", line 187, in sgd
    def sgd(params: List[Tensor],
  File "/opt/conda/lib/python3.8/site-packages/torch/_dynamo/eval_frame.py", line 231, in _fn
    return fn(*args, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/torch/_functorch/aot_autograd.py", line 2847, in forward
    return compiled_fn(full_args)
  File "/opt/conda/lib/python3.8/site-packages/torch/_functorch/aot_autograd.py", line 1222, in g
    return f(*args)
  File "/opt/conda/lib/python3.8/site-packages/torch/_functorch/aot_autograd.py", line 1952, in runtime_wrapper
[...]

PR #4751, opened 2023-03-10
Introduce a new flag to disable functionalization tensor wrapping

wonjoolee95:
Partially picks up https://github.com/pytorch/xla/pull/4747/

---

Sometimes, we may want to disable functionalization tensor wrapping. For example, in dynamo-xla bridge the fx graph we're passed is already functionalized by PyTorch and does not need additional functionalization from us. This PR introduces a flag to disable functionalization. Note that this flag is not yet used anywhere in this PR.

Test: CI

### DIFF
--- a/torch_xla/csrc/torch_util.cpp
+++ b/torch_xla/csrc/torch_util.cpp
@@ -46,6 +46,15 @@ at::Tensor UnwrapNumber(const at::Tensor& tensor, at::ScalarType dtype) {
                                                            : tensor;
 }
 
+at::Tensor MaybeWrapTensorToFunctional(const at::Tensor& tensor) {
+  bool disable_functionalization =
+      xla::sys_util::GetEnvBool("DISABLE_FUNCTIONALIZATION", false);
+  if (disable_functionalization) {
+    return tensor;
+  }
+  return at::functionalization::impl::to_functional_tensor(tensor);
+}
+
 }  // namespace torch_xla
 
 namespace torch {
--- a/torch_xla/csrc/torch_util.cpp
+++ b/torch_xla/csrc/torch_util.cpp
@@ -48,7 +48,7 @@ at::Tensor UnwrapNumber(const at::Tensor& tensor, at::ScalarType dtype) {
 
 at::Tensor MaybeWrapTensorToFunctional(const at::Tensor& tensor) {
   bool disable_functionalization =
-      xla::sys_util::GetEnvBool("DISABLE_FUNCTIONALIZATION", false);
+      xla::sys_util::GetEnvBool("XLA_DISABLE_FUNCTIONALIZATION", false);
   if (disable_functionalization) {
     return tensor;
   }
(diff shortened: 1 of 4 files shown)


torch_xla/csrc/torch_util.cpp
```diff
@@ -46,6 +46,15 @@ at::Tensor UnwrapNumber(const at::Tensor& tensor, at::ScalarType dtype) {
                                                            : tensor;
 }
 
+at::Tensor MaybeWrapTensorToFunctional(const at::Tensor& tensor) {
+  bool disable_functionalization =
+      xla::sys_util::GetEnvBool("DISABLE_FUNCTIONALIZATION", false);
```
JackCaoG, line 51: maybe prefix it with `XLA` to match our other env vars?

wonjoolee95: Updated

wonjoolee95:
@alanwaketan, this should be a ready for review. Thanks!

torch_xla/csrc/torch_util.h
```diff
@@ -59,6 +60,10 @@ at::Scalar MakeFloatScalar(T value) {
 // Unwraps tensor to target dtype if it's a wrapped number.
 at::Tensor UnwrapNumber(const at::Tensor& tensor, at::ScalarType dtype);
 
+// Wraps tensor to functional tensor if XLA_DISABLE_FUNCTIONALIZATION is false
+// or not set.
+at::Tensor MaybeWrapTensorToFunctional(const at::Tensor& tensor);
```
alanwaketan, line 65: Can you add a comment:
```
For unwrapping, `torch::lazy::maybe_unwrap_functional()` will only unwrap tensors that are functional. So, nothing needs to be done there.
```

alanwaketan (approved):
LGTM. Thanks Wonjoo. Please follow up with another change that we can actually enable this flag.

I will suggest that in that change, we can run test_operations.py with the flag on. That should be enough for the coverage.

torch_xla/csrc/torch_util.h
wonjoolee95: Added, thanks for the suggestion -- very helpful comment.

wonjoolee95:
> LGTM. Thanks Wonjoo. Please follow up with another change that we can actually enable this flag.
> 
> I will suggest that in that change, we can run test_operations.py with the flag on. That should be enough for the coverage.

Thanks for the approval. Sounds good, will do. I'll merge this PR now and follow-up with a new PR to do that.

alanwaketan:
> > LGTM. Thanks Wonjoo. Please follow up with another change that we can actually enable this flag.
> > I will suggest that in that change, we can run test_operations.py with the flag on. That should be enough for the coverage.
> 
> Thanks for the approval. Sounds good, will do. I'll merge this PR now and follow-up with a new PR to do that.

This is probably unrelated to the PR. Another nice follow up to have is to leave comments on the view infra to suggest that it's kept for a/b testing functionalization only. New feature development shouldn't bother considering it.

wonjoolee95:
Merging now as the CIs are green. Will follow-up with the comments mentioned earlier in the thread.

PR #4752, opened 2023-03-10
Jack cao g/r20 backportdoc

JackCaoG:
Back porting a bunch of documentation change to 2.0

### DIFF
--- a/docs/dynamo.md
+++ b/docs/dynamo.md
@@ -1,0 +1,90 @@
+## TorchDynamo(torch.compile) integration in PyTorch XLA
+
+Torchdynamo is a Python-level JIT compiler designed to make unmodified PyTorch programs faster. It provides a clean API for compiler backends to hook in and its biggest feature is to dynamically modify Python bytecode right before it is executed. In the pytorch/xla 2.0 release, PyTorch/XLA provided an experimental backend for the TorchDynamo for both inference and training. 
+
+The way that XLA bridge works is that Dynamo will provide a TorchFX graph when it recognizes a model pattern and PyTorch/XLA will use existing Lazy Tensor technology to compile the FX graph and return the compiled function. 
+
+### Inference
+Here is a small code example of running resnet18 with `torch.compile`
+
+```python
+import torch
+imprt torchvision
+import torch_xla.core.xla_model as xm
+
+def eval_model(loader):
+  device = xm.xla_device()
+  xla_resnet18 = torchvision.models.resnet18().to(device)
+  xla_resnet18.eval()
+  dynamo_resnet18 = torch.compile(
+      xla_resnet18, backend='torchxla_trace_once')
+  for data, _ in loader:
+    output = dynamo_resnet18(data)
+```
+> **NOTE:** inference backend name `torchxla_trace_once` is subject to change.
+
+With the `torch.compile` you will see that PyTorch/XLA only traces the resent18 model once during the init time and executes the compiled binary everytime `dynamo_resnet18` is invoked, instead of tracing the model every time. Note that currently Dynamo does not support fallback so if there is an op that can not be traced by XLA, it will error out. We will fix this issue in the upcoming 2.1 release. Here is a inference speed analysis to compare Dynamo and Lazy using torch bench on Cloud TPU v4-8
+
+| model | Speed up |
+| --- | ----------- |
+resnet18 | 1.768
+resnet50 | 1.61
+resnext50_32x4d	| 1.328
+alexnet | 1.261
+mobilenet_v2 | 2.017
+mnasnet1_0 | 1.686
+vgg16 | 1.155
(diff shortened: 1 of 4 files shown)


docs/dynamo.md
```diff
@@ -0,0 +1,90 @@
+## TorchDynamo(torch.compile) integration in PyTorch XLA
+
+Torchdynamo is a Python-level JIT compiler designed to make unmodified PyTorch programs faster. It provides a clean API for compiler backends to hook in and its biggest feature is to dynamically modify Python bytecode right before it is executed. In the pytorch/xla 2.0 release, PyTorch/XLA provided an experimental backend for the TorchDynamo for both inference and training.
```
miladm, line 3: Can we add this reference to pytorch dynamo - for additional : https://pytorch.org/tutorials/intermediate/dynamo_tutorial.html?

docs/dynamo.md
```diff
@@ -0,0 +1,90 @@
+## TorchDynamo(torch.compile) integration in PyTorch XLA
+
+Torchdynamo is a Python-level JIT compiler designed to make unmodified PyTorch programs faster. It provides a clean API for compiler backends to hook in and its biggest feature is to dynamically modify Python bytecode right before it is executed. In the pytorch/xla 2.0 release, PyTorch/XLA provided an experimental backend for the TorchDynamo for both inference and training. 
+
+The way that XLA bridge works is that Dynamo will provide a TorchFX graph when it recognizes a model pattern and PyTorch/XLA will use existing Lazy Tensor technology to compile the FX graph and return the compiled function.
```
miladm, line 5: can you add a few words on how a training cycle may consist of >1 graph - a contrast to the lazy execution flow.

for inference: we can make dynamo guarantee 1 graph to be generated if proper flag is set.

for training: today, we get at least 3 graphs.

docs/dynamo.md
```diff
@@ -0,0 +1,90 @@
...
+  xla_resnet18 = torchvision.models.resnet18().to(device)
+  xla_resnet18.eval()
+  dynamo_resnet18 = torch.compile(
+      xla_resnet18, backend='torchxla_trace_once')
+  for data, _ in loader:
+    output = dynamo_resnet18(data)
+```
+> **NOTE:** inference backend name `torchxla_trace_once` is subject to change.
+
+With the `torch.compile` you will see that PyTorch/XLA only traces the resent18 model once during the init time and executes the compiled binary everytime `dynamo_resnet18` is invoked, instead of tracing the model every time. Note that currently Dynamo does not support fallback so if there is an op that can not be traced by XLA, it will error out. We will fix this issue in the upcoming 2.1 release. Here is a inference speed analysis to compare Dynamo and Lazy using torch bench on Cloud TPU v4-8
```
miladm, line 26: do you have a few words on how the lack of fallback support impacts modeling - something along the lines of what class of (example of) models should work easily and what models should fail.

docs/dynamo.md
```diff
@@ -0,0 +1,90 @@
...
+
+We are currently working on the optimizer support and that will be availiable on nightly soon but won't be in the 2.0 release.
+
+### Take away
+TorchDynamo provides a really promising way for the compiler backend to hide the complexity from the user and easily retrieve the modeling code in a graph format. Compared with PyTorch/XLA’s traditional Lazy Tensor way of extracting the graph, TorchDynamo can skip the graph tracing for every iteration hence provide a much better inference response time. However TorchDynamo does not trace the communication ops(like `all_reduce` and `all_gather`) yet and it provides separate graphs for the forward and the backward which hurts xla performance. These feature gaps compared to Lazy Tensor makes it less efficient in real world training use cases, especially the tracing cost can be overlapped with the execution in training. The PyTorch/XLA team will keep investing in TorchDynamo and work with upstream to mature the training story.
```
miladm, line 90: suggesting we explicitly call out this as a "distributed support feature gap"

miladm: the paragraph would be more actionable if it outlines what type of models (examples) should / shouldn't work because of this gap. otherwose, I worry it unnecessarily discourages the user.

plus we should clarify this across training vs. inference scenarios.

it's like a 2x2 matrix in my mind if that makes sense.

miladm (commented):
Thanks @JackCaoG - I added a few thoughts.

JackCaoG:
@miladm This is a cherry-pick to 2.0, I will make corresponding changes to master and then do any backport sometime today

docs/dynamo.md
JackCaoG: I do mentioned that below 
```
With the `torch.compile` you will see that PyTorch/XLA only traces the resent18 model once during the init time and executes the compiled binary everytime `dynamo_resnet18` is invoked
```
and
```
We expect to extract and execute 3 graphs per training step instead of one training step if you use the Lazy tensor.
```

PR #4753, opened 2023-03-10
[Internal Test] HLO<->StableHLO roundtrip

lsy323:


third_party/xla_client/mhlo_helper.cc
```diff
@@ -0,0 +1,135 @@
...
+#include "tensorflow/compiler/xla/translate/mhlo_to_hlo/mlir_hlo_to_hlo.h"
+#include "tensorflow/compiler/xla/mlir_hlo/mhlo/transforms/passes.h"
+#include "mlir/IR/Verifier.h"  // from @llvm-project
+#include "mlir/Pass/PassManager.h" // from @llvm-project
+
+#include <iostream>
+
+namespace xla {
+
+void hlo_mhlo_hlo_roundtrip_helper(HloModuleProto* proto) {
```
vanbasten23, line 13: A function with the same name `hlo_mhlo_hlo_roundtrip_helper` seems to be defined in third_party/xla_client/computation_client.cc. I wonder where the function in this file is used.

PR #4755, opened 2023-03-10
Bazel torchxla

stgpetrovic:
Migrate `torch_xla` fully to bazel.

PyTorch/XLA can now be built fully with bazel with remote caching. PyTorch has to be build before; all dependencies are satisfied from the built pytorch repository that's the parent directory of the xla checkout.
 
- [x] PyTorch is a bazel remote repository
- [x] torch_xla/csrc sources migrated to bazel
- [x] PyTorch dev headers offered as a library 
- [x] PyTorch runtime headers offered as a library
- [x] PyTorch shared objects (.so) offered as a library
- [x] Tests migrated to bazel
- [x] Bazel remote caching enabled


yeounoh:
Exciting!

.bazelrc
```diff
@@ -84,14 +104,13 @@ try-import %workspace%/.bazelrc.user
...
 
 # Compiling tests requires Java.
-common --java_runtime_version=remotejdk_11
+build --java_runtime_version=remotejdk_11
 
 # Coverage requires Java and GCC.
 coverage --config=coverage
 coverage --build_tests_only
 build:coverage --copt=-DNDEBUG
 build:coverage --combined_report=lcov
```
steventk-g, line 113: How can we generate and upload this lcov coverage report? My understanding is that we need to build and link with special args to generate coverage, and that the bazel output is a `.dat` file which we can export to html (https://bazel.build/configure/coverage)

stgpetrovic: Yeah, we can use `genthml` from the `lcov` package to generate the HTML version from the `.dat`.

torch_xla/csrc/token_handler.h
```diff
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef XLA_TORCH_XLA_CSRC_TOKEN_HANDLER_H_
+#define XLA_TORCH_XLA_CSRC_TOKEN_HANDLER_H_
```
vanbasten23, line 2: I wonder why we need this `#define XLA_TORCH_XLA_CSRC_XLA_GRAPH_EXECUTOR_H_` macro.

.bazelrc
yeounoh: We would upload the original `lcov` format to the metrics gcs bucket, as `xxx_coverage.out`. The extension matters less as long as the file content is in `lcov` format. Can we comment about what tests are being included for coverage, and maybe a TODO if there is anything left out?

torch_xla/csrc/tensor_methods.h
```diff
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef XLA_TORCH_XLA_CSRC_TENSOR_METHODS_H_
+#define XLA_TORCH_XLA_CSRC_TENSOR_METHODS_H_
```
vanbasten23, line 2: I may miss something here. I thought the c++ macro takes the form `#define identifier replacement`. So what's the replacement for the macro identifier `XLA_TORCH_XLA_CSRC_TENSOR_METHODS_H_`?

.circleci/docker/Dockerfile
```diff
@@ -34,11 +34,6 @@ ENV CUDA_PATH /usr/local/cuda
 ENV CC "${cc}"
 ENV CXX "${cxx}"
 
-# Whether to build torch and torch_xla libraries with CXX ABI
-ENV _GLIBCXX_USE_CXX11_ABI "${cxx_abi}"
```
yeounoh, line 38: Dockerfile change needs to be coordinated with the upstream, when are we planning to merge this PR? Can we aim for early next week?

.vscode/settings.json
```diff
@@ -13,8 +15,8 @@
         "./bazel-out/_coverage/_coverage_report.dat"
     ],
     "lcov.path": [
-        "./.bazel-out/_coverage/_coverage_report.dat"
+        "./bazel-out/_coverage/_coverage_report.dat"
```
yeounoh, line 18: Do you have a sample .dat file for us to look at?

setup.py
```diff
@@ -16,29 +16,36 @@
...
+#     file containing the auth tokens for remote cache/build. implies remote cache.
+#
+#   BAZEL_REMOTE_CACHE=""
+#     whether to use remote cache for builds
+#
+#   TPUVM_MODE=0
+#     whether to build for TPU
+#
+#   SILO_NAME=""
+#     name of the remote build cache silo
```
vanbasten23, line 41: can you explain more about what's "remote build cache silo"?

CODEGEN_MIGRATION_GUIDE.md
```diff
@@ -30,14 +30,14 @@ All file mentioned below lives under the `xla/torch_xla/csrc` folder, with the e
   - Contains all the op XLA supported today. Most of the ops are under the supported category, the goal of this document is to move most of the ops to the full_codegen category.
 - xla/scripts/gen_lazy_tensor.py
   - Provides necessary XLA versions of the codegen Codegen class and calls the upstream codegen API.
-- xla/torch_xla/csrc/generated/XLANativeFunctions.cpp
-  - Result of the full_codegen column of the xla/xla_native_functions.yaml. The op function defined here will implement the op declared in the XLANativeFunctions.h. Each op will take at::tensor and return another at::tensor wrapped around a XLATensor.
-- xla/torch_xla/csrc/generated/LazyIr.h
-  - Result of the full_codegen column of the xla/xla_native_functions.yaml.  Defines the IR that is used to construct the full_codegen ops.
+- xla/torch_xla/csrc/XLANativeFunctions.cpp
+  - Result of the full_codegen column of the xla/codegen/xla_native_functions.yaml. The op function defined here will implement the op declared in the XLANativeFunctions.h. Each op will take at::tensor and return another at::tensor wrapped around a XLATensor.
```
yeounoh, line 34: should we add xla/codegen to .gitignore?

setup.py
```diff
@@ -16,29 +16,36 @@
...
+#     whether to use ACL
+#
 #   BUNDLE_LIBTPU=0
 #     include libtpu in final wheel
 #
+#   GCLOUD_SERVICE_KEY_FILE=''
+#     file containing the auth tokens for remote cache/build. implies remote cache.
+#
+#   BAZEL_REMOTE_CACHE=""
+#     whether to use remote cache for builds
```
vanbasten23, line 35: is there a guidance about on what occasion we should or should not use remote cache?

setup.py
```diff
@@ -16,29 +16,36 @@
...
-#
-#   XLA_BAZEL_VERBOSE=0
+#   BAZEL_VERBOSE=0
 #     turn on verbose messages during the bazel build of the xla/xrt client
 #
 #   XLA_CUDA=0
 #     build the xla/xrt client with CUDA enabled
 #
+#   XLA_CPU_USE_ACL=0
+#     whether to use ACL
```
vanbasten23, line 26: can you explain more about what this access control list does for xla_cpu?

setup.py
```diff
@@ -258,86 +187,96 @@ def run(self):
...
+    # Debug build.
+    if DEBUG:
+      bazel_argv.append('--compilation_mode=dbg')
+
+    # Remote cache authentication.
+
+    if _check_env_flag('BAZEL_REMOTE_CACHE'):
+      bazel_argv.append('--config=remote_cache')
+
+    if GCLOUD_KEY_FILE:
```
vanbasten23, line 248: can you explain to me why we need gcloud during build?

docker/cloudbuild.yaml
```diff
@@ -47,7 +46,6 @@ substitutions:
     _CUDA_COMPUTE: '3.7,7.0,7.5,8.0'
     _GITHUB_XLA_BRANCH: ''
     _GITHUB_EXAMPLE_BRANCH: 'master'
-    _BUILD_CPP_TESTS: '1'
```
yeounoh, line 50: Do we now have another way to turn on/off the CPP test build? If so, we need to expose it to GCB, since in some environments the test build fails... having said that, have we tested the new bazel build with Python3.9 (both with and without CPP tests)?

vanbasten23, line 50: So after we switch to bazel, do we build cpp test by default?

torch_xla/csrc/convolution.h
```diff
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef XLA_TORCH_XLA_CSRC_CONVOLUTION_H_
```
alanwaketan, line 1: Why not using `#pragma once` anymore?

setup.py
```diff
@@ -16,29 +16,36 @@
...
 #     change the package name to something other than 'torch_xla'
 #
-#   COMPILE_PARALLEL=1
-#     enable parallel compile
-#
-#   BUILD_CPP_TESTS=1
-#     build the C++ tests
-#
-#   XLA_DEBUG=0
-#     build the xla/xrt client in debug mode
```
yeounoh, line 26: We need a way to control the debug mode. Is this always on?

codegen/xla_native_functions.yaml
alanwaketan, line 1: Why moving this file to code-gen?

setup.py
```diff
@@ -117,34 +124,6 @@ def create_version_files(base_dir, version, xla_git_sha, torch_git_sha):
...
-  if child.returncode != 0:
-    print(
-        'Failed to generate lazy files: {}'.format(generate_lazy_cmd),
-        file=sys.stderr)
-    sys.exit(1)
-
-
-def build_extra_libraries(base_dir, build_mode=None):
-  build_libs_cmd = [os.path.join(base_dir, 'build_torch_xla_libs.sh')]
-  cxx_abi = getattr(torch._C, '_GLIBCXX_USE_CXX11_ABI', None)
```
yeounoh, line 136: `cxx_abi` is also a param we wanted to control for debugging + and for release. We also had an incident where `torch._C._GLIBCXX_USE_CXX11_ABI` returns an incorrect flag (always false).

yeounoh (commented):
What a great work, thank you Goran. I left some questions/comments.

docker/Dockerfile
```diff
@@ -14,7 +14,6 @@ ARG cxx_abi="1"
 ARG tpuvm=""
 ARG bazel_jobs=""
 ARG git_clone="true"
-ARG build_cpp_tests="0"
 
 ENV BUILD_CPP_TESTS "${build_cpp_tests}"
```
wonjoolee95, line 18: Seems like we deleted all the cpp tests related flag, is this `BUILD_CPP_TESTS` used anywhere now?

.circleci/docker/Dockerfile
stgpetrovic: Sounds good, we can do it after the pin update whenever it makes most sense for the team.

.vscode/settings.json
stgpetrovic: This is the same file you'd expect from `blaze coverage` in lcov format. https://bazel.build/configure/coverage

codegen/xla_native_functions.yaml
stgpetrovic: Before, a script (gen_lazy_tensor) generated files and wrote them in the repository, and they were in gitignore. This is not ideal because it adds state to the repository and its build process. Now, we use codegen/ directory to host the python script that generates code, and we have a `genrule` that generates the files on the fly during the build process. These files don't exist in the repository (they are in bazel-bin), and there is an explicit dependency to the genrule's output in the targets including the generated files. Having everything needed for this process in codegen/ made sense to me, but it can also be under root if there is a reason for it.

docker/cloudbuild.yaml
stgpetrovic: Please see the comment in docker/Dockerfile.

We don't need to prebuild tests; they will be built during the test execution phase; most of the stuff will be cached because the largest chunk of the code is xla.

I used 3.8, 3.9, 3.10, 3.11 python versions over the time.

scripts/build_torch_wheels.sh
```diff
@@ -5,7 +5,6 @@ set -x  # Display commands being run.
 
 PYTHON_VERSION=$1
 RELEASE_VERSION=$2  # rX.Y or nightly
-BUILD_CPP_TESTS="${3:-0}"
```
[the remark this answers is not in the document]
stgpetrovic: Please see comment in docker/Dockerfile.

What's the need to control this?

.gitignore
```diff
@@ -10,9 +10,6 @@ torch_xla/csrc/version.cpp
 *.pyc
 *.so
 
-# Directory autogenerated by full_codegen
-torch_xla/csrc/generated/
```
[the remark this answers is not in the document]
stgpetrovic: No, this is now handled with a `genrule`, there is no state of the repository outside of bazel (it's hermetic and deterministic).

CODEGEN_MIGRATION_GUIDE.md
```diff
@@ -197,7 +197,7 @@ to
 
 ## Run the test and verify the result
 Run the C++ op test or a simple test that only involves the generated ops. To run the C++ test:
-1. Build the xla through `python setup.py install` (note: don't use the `BUILD_CPP_TESTS=0` flag since this will skip building the C++ tests)
```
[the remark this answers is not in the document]
stgpetrovic: You can run it manually; or run first the wheel and then run the tests, it will reuse 99% of the build cache.

setup.py
stgpetrovic: There was a difference before for DEBUG and XLA_DEBUG. Now only DEBUG is relevant since it's all a single binary, and that will be applied. Is that enough?

.bazelrc
stgpetrovic: Currently all cpp tests are included. They are run with various settings currently (XRT, PJRT), so I suppose I should investigate how to collect coverage for separate invocations.
I'll probably have to merge them, I can follow the current process and try to replicate it.

setup.py
stgpetrovic: We're using gcloud remote build cache on gcloud and for that we need to authenticate with it.

There are 3 scenarions

 - gcloud/workstation will use default authentication
 - circleci will use the service account key
 - cloudbuild will use the cloudbuild service account authentication

https://developers.google.com/remote-build-execution/docs/remote-cache-overview

setup.py
stgpetrovic: I have no idea what this is :)

This is the current behaviour and it comes from here:
https://github.com/pytorch/xla/blob/HEAD/build_torch_xla_libs.sh#L50

I attempted to replicate it and not change semantics.
Probably someone from the team can answer this.

setup.py
stgpetrovic: I'll add documentation to this PR, but generally we should be using it always.

We want to minimize cache misses so maybe each separate environment can use it's own
`SILO_NAME`, but CI and cloudbuild have their custom ones set, others can use `dev`
by default and we can see how that goes.

GCloud authentication is required for this feature.

setup.py
stgpetrovic: Builds and tests that use the remote cache are siloed within the cache with a cache silo key. Each uniquely-configured machine running Bazel should have a unique cache silo key.

torch_xla/csrc/convolution.h
stgpetrovic: We had header guards in one part of the code and pragma in the other.

Since google code style recommends header guards (https://google.github.io/styleguide/cppguide.html#The__define_Guard), and discourages the `#pragma once` even on windows (https://google.github.io/styleguide/cppguide.html#Windows_Code), I went for the header guards. They are standardized and recommended.

I have no opinion really outisde of these guidances.

setup.py
stgpetrovic: Cool. Added a `CXX_ABI` flag that can be set from the outside . If empty, it will default to `torch._C`.

docker/cloudbuild.yaml
stgpetrovic: See above.

.bazelrc
stgpetrovic: I see that you only have python coverage for now, right?

vanbasten23:
One general question about dev workflow with bazel, let's say I modify a file locally pytorch/xla/torch_xla/csrc/helpers.cpp and I want to run a test. Is it the workflow like: blaze build //pytorch/xla/torch_xla/csrc:tensor and then run test like `blaze test //...`?

.circleci/common.sh
```diff
@@ -14,6 +14,8 @@ fi
 # 2. CONDA_PREFIX (if it exists)
 # 3. The conda install directory (if it exists)
 export CMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH:-${CONDA_PREFIX:-"$(dirname $(which conda))/../"}}
+export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$(python3-config --prefix)/lib"
```
will-cromar, line 17: How hard would it be to convert this constellation of testing bash scripts to bazel [`test_suite`s](https://bazel.build/reference/be/general#test_suite)? Could bazel handle setting env vars, flags, etc? Would we need custom logic to generate targets for different variants (e.g. TPU, single-GPU, multi-GPU, etc)?

This PR is already quite large, so no need to make that change now

scripts/build_torch_wheels.sh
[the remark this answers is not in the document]
will-cromar: We used to include a pre-built binary that contains the C++ tests in the wheel, which extends the build time and inflates the wheel size, so we added this flag to disable it. My understanding is that the C++ test binary will not be included in the wheel at all after this PR. Is that right?

will-cromar (approved):
Amazing work. Thank you so much!

test/cpp/run_tests.sh
```diff
@@ -39,35 +37,38 @@ do
       BUILD_ONLY=1
       ;;
     F)
-      FILTER="--gtest_filter=$OPTARG"
+      FILTER="--test_filter=$OPTARG"
```
alanwaketan, line 40: Does `--gtest_filter` still work or not?

torch_xla/csrc/BUILD
```diff
@@ -0,0 +1,289 @@
...
+        "no-remote-exec",
+    ],
+    tools = [
+        "//codegen:fix_includes",
+        "//codegen:lazy_tensor_generator",
+    ],
+)
+
+ptxla_cc_library(
+    name = "tensor",
```
alanwaketan, line 32: May I ask what's the rules to define each module?

alanwaketan (commented):
This is great. Finally went through it! Leave more comments.

scripts/build_torch_wheels.sh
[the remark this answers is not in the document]
stgpetrovic: Ah that makes sense. I hope those days are gone.

We can bring it back if we need to, or more like work until that requirement is gone.

codegen/BUILD
```diff
@@ -0,0 +1,20 @@
+exports_files(["xla_native_functions.yaml"])
+
+# Requires `torchgen` locally, via pip `torch` package.
```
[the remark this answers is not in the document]
stgpetrovic: This python script imports `torchgen` which is then used to generate the files.

This requires that `torch` package is installed, usually by `pip install torch`,
or by installing from a built wheel `pip install dist/*whl`.

This is uncommon for bazel; usually everything is hermetic and provided by the repository itself,
so it's worth a comment.

Running `bazel build` in the XLA repository will fail if `torch` package is not installed within the machine's python distribution.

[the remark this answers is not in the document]
alanwaketan: Ah, can you rephrase it to `Requires `torchgen` locally, via pip `torch` package or local builds.`?

codegen/lazy_tensor_generator.py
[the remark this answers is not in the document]
stgpetrovic: Good point, removing.

docs/bazel.md
```diff
@@ -0,0 +1,166 @@
...
+needs to have remote build permissions in GCP. In the CI, the service account key is used for authentication and is passed to
+bazel using `--config=remote_cache --google_credentials=path/to/service.key`. On [Cloud Build](https://cloud.google.com/build), `docker build --network=cloudbuild` is used to pass the authentication from the service account running the cloud build down into the docker image doing the compilation: [Application Default Credentials](https://cloud.google.com/docs/authentication/provide-credentials-adc) does the work there and authenticates as the service account. All accounts, both
+user and service accounts, need to have remote cache read/write permissions.
+
+Remote cache uses cache silos. Each unique machine and build should specify a unique silo key to benefit from consistent caching. The silo key can be passed using a flag: `-remote_default_exec_properties=cache-silo-key=SOME_SILO_KEY'`.
+
+Running the build with remote cache:
+
+```bash
+BAZEL_REMOTE_CACHE=1 SILO_NAME="cache-silo-YOUR-USER" TPUVM_MODE=1 python setup.py bdist_wheel
```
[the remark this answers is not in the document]
stgpetrovic: 1. Yes
2. any user like `alanwaketan`. The idea is that one machine always uses the same key, to maximize cache hits, minimize misses.

Will add some explanation.

docs/bazel.md
```diff
@@ -0,0 +1,166 @@
...
+authentication with gcloud. See `.bazelrc` for the configuration.
+
+Using bazel makes it easy to express complex dependencies and there is a lot of gain from having a single build graph
+with everything expressed in the same way. Therefore, there is no need to build the XLA libraries separately from the
+rest of the pluing as used to be the case, building the whole repository, or the plugin shared object that links everythin
+else in, is enough.
+
+## How to build the Torch/XLA plugin
+
+The normal build can be achieved by the invoking the standard `python setup.py bdist_wheel`, but C++ bindings can be built simply with:
```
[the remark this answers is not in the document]
stgpetrovic: Yes, the change is in replacing the torch CppExtension (that does `g++` and `ld`) with bazel to produce the `.so`, python build works the same as before.

docs/bazel.md
```diff
@@ -0,0 +1,166 @@
...
+```bash
+bazel build --override_repository=org_tensorflow=/path/to/exported/tf_repo //...
+```
+
+```bash
+bazel build --override_repository=torch=/path/to/exported/and/built/torch_repo //...
+```
+
+Please make sure that the overridden repositories are at the appropriate revisions and in case of `torch`, that it
+has been built with `USE_CUDA=0 python setup.py bdist_wheel` to make sure that all expected build objects are present;
```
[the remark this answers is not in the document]
stgpetrovic: Please see previous comment.

setup.py
stgpetrovic: When running from scripts (ci/test), it is autoconfigured. The default is to use a common silo, so maybe a bit poorer cache performance but still great. I've seen docker digest being used as the key in some places.

test/cpp/run_tests.sh
stgpetrovic: The filter works here when passed with `-F`. The correct flag is `--test_filter` in bazel, it passes it down to the xUnit (e.g. `googletest`) correctly. Tests can be manually filtered in bazel simply as `bazel test --test_filter=TensorTest.* //`. You could also pass a `test_arg` to it, circumventing bazel and just interacting with the C++ binary that is the test, like `bazel test --test_arg=--gtest_filter=TensorTest.* //`. I prefer the first mechanism but both will work.

docs/bazel.md
```diff
@@ -0,0 +1,166 @@
...
+Bazel can power a language server like [clangd](https://clangd.llvm.org/) that brings code references, autocompletion and semantic understanding of the underlying code to your editor of choice. For VSCode,
+one can use [Bazel Stack](https://github.com/stackb/bazel-stack-vscode-cc) that can be combined with
+[clangd](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd) functionality to bring powerful features to assist code editing.
+
+## Building PyTorch/XLA
+
+As always, PyTorch/XLA is built using Python `distutils`:
+
+```bash
+BAZEL_REMOTE_CACHE=1 TPUVM_MODE=1 python setup.py bdist_wheel
```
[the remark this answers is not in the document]
alanwaketan: Got it. So SILO_NAME would improve the performance of BAZEL_REMOTE_CACHE but it's not required.

torch_xla/csrc/BUILD
stgpetrovic: Ideally 1 C++ file/header should be 1 library. But this code is written in such a way that everything includes with everything else so that cannot be done as-is. I'll spend some time unwrapping the code in the future to reduce this target. The benefits would be faster incremental compilation, better caching and faster linking time.

.vscode/settings.json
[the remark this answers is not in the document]
stgpetrovic: https://code.visualstudio.com/docs/getstarted/settings

These should be somewhat universal; if something is a user-setting here, it should be moved out.

torch_xla/csrc/BUILD
alanwaketan: Yea, it's very unintuitive. Something like grouping with directory makes more sense. haha. Anyway, I guess we can have this cleanup as a follow up.

codegen/fix_includes.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# torchgen insists on including the tensor header as a system <tensor.h> header
+sed -i 's#<torch_xla/csrc/tensor.h>#"torch_xla/csrc/tensor.h"#' $@
+
+# remove the runfiles-prefix used in codegen for pytorch
+sed -i 's#bazel-out/k8-opt/bin/codegen/lazy_tensor_generator.runfiles/torch/##' $@
```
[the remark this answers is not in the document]
stgpetrovic: Because of the way `torchgen` works, it will generate includes relative to the runfiles directory,
so we update that here. It's a bit ugly.

docs/bazel.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+The first time the code is compiled using a new cached key will be slow because it will compile everything from scratch, but incremental compilations will be very fast. On updating the TensorFlow pin, it will once again be a bit slower the first time per key, and then until the next update quite fast again.
+
+## Running tests
+
+Bazel is a test plafrom too, making it easy to run tests:
+
+```bash
+bazel test //...
```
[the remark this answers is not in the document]
stgpetrovic: The old way still works; I've replaced the cmake build + exec with `bazel test`.

You can run any target you like, e.g. `bazel test //test/cpp:main`.

Python is not yet migrated to bazel.

.vscode/settings.json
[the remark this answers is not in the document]
alanwaketan: So basically, it should be automatically merged if I open the xla directory as the workspace?

torch_xla/csrc/BUILD
stgpetrovic: Grouping by directory makes for big slow (re)compilation and linking, with remote caching small nimble objects are the best. I'll do an unspagetti-run some time in the future.

docs/bazel.md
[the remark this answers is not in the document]
alanwaketan: I think updating the comment with the above example and suggesting it only works for C++ would be good.

codegen/fix_includes.sh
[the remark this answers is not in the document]
alanwaketan: Maybe we can make the comment more expressive as your explanation above?

torch_xla/csrc/BUILD
alanwaketan: I guess I would prefer readability over performance, but that's my own opinion and I haven't used bazel before. Anyway as long as we can provide a general rule and put it as a comment at the top of this file, I'm fine with it.

.vscode/settings.json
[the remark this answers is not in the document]
stgpetrovic: Yes, that's the idea. This will also come with a separate repo that contains the vscode devcontainer setup that loads the dev image that contains all prerequisites to build and all vscode plugins too, so 0-setup time.

codegen/fix_includes.sh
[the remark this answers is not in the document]
stgpetrovic: Added a comment. We could try to get support for a parameter in `torchgen` to support passing in the relative include root.

.bazelrc
yeounoh: Yes, the pipeline is working.

BUILD
```diff
@@ -0,0 +1,57 @@
...
+        "//third_party/xla_client:xla_util",
+        "//torch_xla/csrc:computation",
+        "//torch_xla/csrc:device",
+        "//torch_xla/csrc:init_python_bindings",
+        "//torch_xla/csrc:tensor",
+        "//torch_xla/csrc:version",
+        "@com_google_absl//absl/container:flat_hash_map",
+        "@com_google_absl//absl/strings",
+        "@com_google_absl//absl/types:variant",
+        "@org_tensorflow//tensorflow/compiler/xla/python/profiler/internal:traceme_wrapper",
```
ManfeiBai, line 40: are this `"@org_tensorflow//tensorflow/compiler/xla/python/profiler/internal:traceme_wrapper"` is different from the following `"@org_tensorflow//tensorflow/python/profiler/internal:profiler_pywrap_impl",`?

do we have the difference in their usage?

BUILD
```diff
@@ -0,0 +1,57 @@
...
+        "@org_tensorflow//tensorflow/compiler/xla/service:hlo_parser",
+        "@org_tensorflow//tensorflow/compiler/xla/service:hlo_pass_pipeline",
+        "@org_tensorflow//tensorflow/compiler/xla/service:hlo_verifier",
+        "@org_tensorflow//tensorflow/compiler/xla/service:sharding_propagation",
+        "@org_tensorflow//tensorflow/compiler/xla/service/spmd:spmd_partitioner",
+        "@org_tensorflow//tensorflow/core",
+        "@org_tensorflow//tensorflow/core:protos_all_cc",
+        "@org_tensorflow//tensorflow/core/platform:env",
+        "@org_tensorflow//tensorflow/core/profiler/lib:traceme",
```
ManfeiBai, line 49: are these `tensorflow/core` all needed?

we might want to delete `tensorflow/core`'s useage when migrate to pull XLA from OpenXLA, so we want to check whether all the deps could be changed to OpenXLA or not: https://github.com/openxla/xla

torch_xla/csrc/BUILD
```diff
@@ -0,0 +1,289 @@
...
+        "@org_tensorflow//tensorflow/compiler/xla/client/lib:logdet",
+        "@org_tensorflow//tensorflow/compiler/xla/client/lib:math",
+        "@org_tensorflow//tensorflow/compiler/xla/client/lib:slicing",
+        "@org_tensorflow//tensorflow/compiler/xla/stream_executor:dnn",
+        "@org_tensorflow//tensorflow/core:lib",
+        "@org_tensorflow//tensorflow/core/framework:tensor_shape",
+        "@org_tensorflow//tensorflow/core/kernels:conv_ops",
+        "@org_tensorflow//tensorflow/core/lib/core:errors",
+        "@org_tensorflow//tensorflow/core/lib/gtl:inlined_vector",
+        "@org_tensorflow//tensorflow/core/profiler/lib:traceme",
```
ManfeiBai, line 151: same question, are these `tensorflow/core`'s dep are needed? are they used directly or generally?

we would migrate to pull XLA from OpenXLA, so would delete TensorFlow, and will delete `tensorflow/core` deps too

third_party/xla_client/BUILD
```diff
@@ -360,7 +359,6 @@ cc_library(
         "@org_tensorflow//tensorflow/compiler/xla:types",
         "@org_tensorflow//tensorflow/core:lib",
```
ManfeiBai, line 360: same question, do we need this `tensorflow/core` deps directly or generally?

we would migrate to pull XLA from OpenXLA, and would delete tensorflow too, so would delete `tensorflow/core` deps too

third_party/xla_client/BUILD
```diff
@@ -485,8 +479,8 @@ tf_cc_shared_object(
         "@org_tensorflow//tensorflow/compiler/xla/client/lib:svd",
         "@org_tensorflow//tensorflow/compiler/xla/rpc:grpc_stub",
         "@org_tensorflow//tensorflow/core:lib",
```
ManfeiBai, line 481: same question, do we need this `tensorflow/core` deps directly or generally?

we would migrate to pull XLA from OpenXLA, and would delete tensorflow too, so would delete `tensorflow/core` deps too

torch_xla/csrc/BUILD
```diff
@@ -0,0 +1,289 @@
...
+    hdrs = ["layout_manager.h"],
+    deps = [
+        ":device",
+        "//third_party/xla_client:debug_macros",
+        "//third_party/xla_client:sys_util",
+        "//third_party/xla_client:tf_logging",
+        "//third_party/xla_client:util",
+        "@com_google_absl//absl/strings",
+        "@org_tensorflow//tensorflow/compiler/xla:shape_util",
+        "@org_tensorflow//tensorflow/core/framework:tensor_shape",
```
ManfeiBai, line 193: same question, do we need this `tensorflow/core` deps directly or generally?

we would migrate to pull XLA from OpenXLA, and would delete tensorflow too, so would delete `tensorflow/core` deps too

torch_xla/csrc/BUILD
```diff
@@ -0,0 +1,289 @@
...
+)
+
+ptxla_cc_library(
+    name = "function_call_tracker",
+    srcs = ["function_call_tracker.cpp"],
+    hdrs = ["function_call_tracker.h"],
+    deps = [
+        "//third_party/xla_client:sys_util",
+        "@com_google_absl//absl/strings",
+        "@org_tensorflow//tensorflow/core/platform:stacktrace",
```
ManfeiBai, line 224: same question, do we need this `tensorflow/core` deps directly or generally?

we would migrate to pull XLA from OpenXLA, and would delete tensorflow too, so would delete `tensorflow/core` deps too

ManfeiBai (approved):
LGTM

left some comments of my questions about `tensorflow/core` deps

BUILD
stgpetrovic: I don't really know, I just copied whatever was used before.

BUILD
stgpetrovic: They are needed because the repository still depends on non-xla tf, right?

torch_xla/csrc/BUILD
stgpetrovic: Yeah, once we remove the logical dependency to tensorflow, then these will not be needed any more.

third_party/xla_client/BUILD
stgpetrovic: same as above.

BUILD
will-cromar: `@org_tensorflow//tensorflow/python/profiler/internal:profiler_pywrap_impl` sounds like the python bindings that JAX would use, but I don't know for sure. Let's keep this PR as minimal as possible (ie still huge) and change as few dependencies as possible.

.circleci/docker/Dockerfile
JackCaoG: Is this still a valid concern?

setup.py
JackCaoG: Hey @stgpetrovic in your experince how reliable is the TF's debug build? In my experience it almost never worked hence I usually have `DEBUG=1` and `XLA_DEBUG=0` so at least I get the stack trace info for torch_xla.

setup.py
JackCaoG: it is from AWS CPU team, they want to use pytorch/xla for ARM CPU I think.

JackCaoG (approved):
This is super exciting! Thanks @stgpetrovic !

setup.py
stgpetrovic: Hm, it fails to build in debug mode currently. We need to fix that. Let me look into it. We can't have the codebase in a state where `#ifdef NDEBUG` causes issues.

stgpetrovic: There are some more issues with `dbg` builds :S

Not sure what to do here. I'll keep digging.

JackCaoG: Thanks Goran, that has been my experience :D. I fix one debug build error and then there is next, I felt like debug build is either not used or used differently internally so people break it pretty often without noticing.

setup.py
```diff
@@ -258,92 +190,97 @@ def run(self):
...
+
+    # Build configuration.
+    if _check_env_flag('BAZEL_VERBOSE'):
+      bazel_argv.append('-s')
+    if _check_env_flag('XLA_CUDA'):
+      bazel_argv.append('--config=cuda')
+    if _check_env_flag('XLA_CPU_USE_ACL'):
+      bazel_argv.append('--config=acl')
+
+    if IS_WINDOWS:
```
will-cromar, line 270: Do we need this case? We don't support Windows.

setup.py
stgpetrovic: I managed to get it to work and I think it consistently will.

Basically, we can control per-file `copts` with bazel, so I told it to compile all non-pytorch/xla files with `-g0 -DNDEBUG`, which makes our whole repo (including the stuff in xla_client) have debug symbols, and the external stuff (tf, llvm, mlir) not. We can in the future maybe play with this to enable more debug symbols form TF in TF/OpenXLA core, but not the underlying deps that usually are the root cause of issues when compiling in debug mode.

This is a nice compromse IMHO.

PR #4763, opened 2023-03-13
Fix XLASymNodeImpl::has_hint

vanbasten23:
This PR modifies `XLASymNodeImpl::has_hint` implementation and also implements missing XLASymNodeImpl methods: `is_non_overlapping_and_dense`, `wrap_bool`, and `sym_or`.

Currently, torch_xla's [implementation](https://github.com/pytorch/xla/blob/5da150236974aa892124891f9729350df3a9f99f/torch_xla/csrc/ops/dynamic_ir.h#L39-L54) is we track 2 fields: upper_bound and real_size separately. When we trace a program
```
# assuming x has shape torch.Size([10])
y = torch.nonzero(x) 
```
we only know and propagate `y`'s upper_bound. (`print(y)` outputs `torch.Size([<=10, 1])`.) To obtain its real value, either:
- PyTorch/XLA has to compile and execute via a `mark_step` call explicitly
- the code encounter a condition or assertion (such as `assert(y.shape[0], 2)` for example.) which torch_xla implicitly trigger a compilation and execution.

Test plan: `is_non_overlapping_and_dense`, `wrap_bool`, and `sym_or` are covered by the test ` TestDynamicShapeModels.test_backward_pass_with_dynamic_input_multibatch_correctness` and `TestDynamicShapes.test_t_copy`.

### DIFF
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -849,9 +849,7 @@ bool XLASymNodeImpl::bool_() {
 // "a SymInt has_hint" is equivalent to "a SymInt is backed". Unbacked SymInt is
 // the result of a data dependent output like nonzero; we don't know what the
 // value is because it's data dependent.
-bool XLASymNodeImpl::has_hint() {
-  return false;
-}
+bool XLASymNodeImpl::has_hint() { return false; }
 
 std::string XLASymNodeImpl::str() {
   return "<=" + std::to_string(DimCast(node().get())->getStaticValue());
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -750,8 +750,8 @@ c10::SymNode XLASymNodeImpl::sym_max(const c10::SymNode& other) {
 }
 
 c10::SymNode XLASymNodeImpl::sym_or(const c10::SymNode& other) {
-  XLA_CHECK(false) << "XLASymNodeImpl::" << __FUNCTION__
-                   << " has not been implemented.";
+  auto error_node = torch::lazy::MakeNode<SizeError>();
+  return c10::make_intrusive<XLASymNodeImpl>(error_node, PyType::BOOL);
 }
 
 c10::SymNode XLASymNodeImpl::sym_and(const c10::SymNode& other) {
@@ -789,10 +789,12 @@ c10::SymNode XLASymNodeImpl::is_channels_last_strides_3d(
   XLA_CHECK(false) << "XLASymNodeImpl::" << __FUNCTION__
                    << " has not been implemented.";
 }
+
+// xw32: what's the reason that this function should never be called in practice?
 c10::SymNode XLASymNodeImpl::is_non_overlapping_and_dense(
     at::ArrayRef<c10::SymNode> sizes, at::ArrayRef<c10::SymNode> strides) {
-  XLA_CHECK(false) << "XLASymNodeImpl::" << __FUNCTION__
-                   << " has not been implemented.";
+  auto error_node = torch::lazy::MakeNode<SizeError>();
+  return c10::make_intrusive<XLASymNodeImpl>(error_node, PyType::BOOL);
 }
 
 c10::SymNode XLASymNodeImpl::clone() {
@@ -813,6 +815,11 @@ c10::SymNode XLASymNodeImpl::wrap_int(int64_t num) {
 c10::SymNode XLASymNodeImpl::wrap_float(double num) {
   XLA_CHECK(false) << "XLASymNodeImpl::" << __FUNCTION__
                    << " has not been implemented.";
+}
+
+c10::SymNode XLASymNodeImpl::wrap_bool(bool num) {
+  auto cnst = torch::lazy::MakeNode<SizeConstant>(num);
+  return c10::make_intrusive<XLASymNodeImpl>(cnst, PyType::BOOL);
 }
 
 int64_t XLASymNodeImpl::guard_int(const char* file, int64_t line) {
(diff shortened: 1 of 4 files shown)


ezyang (changes requested):
Maybe this accidentally works but this doesn't look right to me. An unbacked SymInt is still an int!

torch_xla/csrc/tensor.cpp
```diff
@@ -840,7 +840,15 @@ bool XLASymNodeImpl::bool_() {
   return dn->getDynamicValue() != 0;
 }
 
-bool XLASymNodeImpl::has_hint() { return true; }
+// "a SymInt has_hint" is equivalent to "a SymInt is backed". Unbacked SymInt is
```
alanwaketan, line 843: From Ed: If the ops are hinted, then the outputs are hinted. Starting point: 1) Inputs are hinted. 2) Data dependent ops are all false. (nonzero is special, 1st dim is unbacked, 2nd dim is backed.)

vanbasten23:
> Maybe this accidentally works but this doesn't look right to me. An unbacked SymInt is still an int!

Sorry for the late reply as I was OOO last week. I see your point. You are saying an unbacked SymInt is still an int type. Let me try something else.

torch_xla/csrc/tensor.cpp
vanbasten23: Hi @ezyang , I'm not sure if I follow your suggestion. But before delving into it, it seems torch_xla only creates an `XLASymNodeImpl` when the dimension is dynamic ([code](https://github.com/pytorch/xla/blob/aa532ff18da0c909efcd100de817e1703228fa40/torch_xla/csrc/tensor_impl.cpp#L191-L196)). IOW, as long as we have an `XLASynNodeImpl`, the dimension it represent is dynamic. So would it make sense to you that our `has_hint()` always return false?

ezyang: Uhh, so I guess my question is, does XLA support compilation with varying batch size without recompiling the graph every time? If the answer is no, then yes I agree with your assessment: you can have has_hint always return false. But this is unexpected, and certainly not how dynamo dynamic shapes works. If you are going to do this, it is very very important that dimensions are rarely dynamic (only after, e.g., nonzero call).

vanbasten23: > does XLA support compilation with varying batch size without recompiling the graph every time?

The answer is no. As I talked with Milad, for the 2 types of dynamism: the one introduced by dynamic ops (such as nonzero) and the one introduced by dynamic input (such as the varying batch size that you mentioned), current torch_xla dynamic shape design only intends to address the former. For latter, we may address it in the future. So I'll try to make `has_hint()` always return false for now.

> If you are going to do this, it is very very important that dimensions are rarely dynamic (only after, e.g., nonzero call).

Thanks for the headsup. Right now, we only have 2 ops `nonzero` and `unique` that introduce dynamic dimensions.

ezyang: In PyTorch proper, we return an opaque IsNonOverlappingAndDense(...) size expression and hope that we don't have to ever actually evaluate it.

vanbasten23: By making `has_hint()` return false, I discovered that I need to implement the missing methods on XLASymNodeImpl:  IsNonOverlappingAndDense(), wrap_bool(), sym_or(), which may some effort.

So my first question is: is making `has_hint()` return false the correct implementation to you (considering you said `But this is unexpected, and certainly not how dynamo dynamic shapes works.`)? I asked because I don't want to take a seemingly short path and have to revisit this part later. If it's not the correct implementation, can you shed some light on the correct implementation?

ezyang: Yes, I expected that you would have to do this. I guess we should have a call to talk through what is going on with these at a high level.

vanbasten23: Thanks Ed. Really appreciate it. Maybe we can chat on Tuesday's meeting between PyTorch/XLA and PyTorch? Meanwhile, I can prepare my question in the [doc](https://docs.google.com/document/d/1KQlTojpwDQgauL0RKu1zrq9M2Cf3E7PUkliycV2Clsk/edit).

torch_xla/csrc/tensor.cpp
```diff
@@ -789,10 +789,12 @@ c10::SymNode XLASymNodeImpl::is_channels_last_strides_3d(
   XLA_CHECK(false) << "XLASymNodeImpl::" << __FUNCTION__
                    << " has not been implemented.";
 }
+
+// xw32: what's the reason that this function should never be called in practice?
 c10::SymNode XLASymNodeImpl::is_non_overlapping_and_dense(
```
vanbasten23, line 794: hi @ezyang , what's the reason that both `is_non_overlapping_and_dense` and `sym_or` won't be used in practice respectively so that we can just return an `ErrorNode`?

torch_xla/csrc/tensor.cpp
```diff
@@ -840,7 +847,21 @@ bool XLASymNodeImpl::bool_() {
...
+// What if we have
+/*
+t0=torch.zeros(3)
+t2=torch.nonzero(t1)
+sz=t2.shape[0]+t0.shape[0]
+*/
+// then sz is always an unbacked SymInt.
+bool XLASymNodeImpl::has_hint() {
+  std::cout << "xw32, file=" << __FILE__ << ", line=" << __LINE__ << "function=" << __FUNCTION__ << ": " << std::endl;
+  return false;
```
vanbasten23, line 863: hi @ezyang , just want to confirm if my understanding is correct. The reason why we can always return false for `XLASymNodeImpl::has_hint` is that torch_xla only create a `SymNodeImpl` for `nonzero` output, such as [here](https://github.com/pytorch/xla/blob/8e1d3ae98f644f5c05645342d8408d0553fa62ba/torch_xla/csrc/tensor_impl.cpp#L191-L195). Moreover, during propagation such as in 
```
t0=torch.zeros(3)
t2=torch.nonzero(t1)
sz=t2.shape[0]+t0.shape[0]
```
`sz` will remain representing a unbacked SymInt. So it's ok to always return false. Does it sound good to you?

torch_xla/csrc/tensor.cpp
ezyang: it's because I happen to know that these are only used to compute contiguity fields on tensors like "is non overlapping and dense" and if they are never fetched it is fine for them to error only if poked

torch_xla/csrc/tensor.cpp
ezyang: yes

ezyang (approved):

alanwaketan (approved):
LGTM.

vanbasten23:
Thanks for the review.

PR #4764, opened 2023-03-13
Return instead of exit when failing install_deps_pytorch_xla

huydhn:
After adding retry for `install_deps_pytorch_xla` in https://github.com/pytorch/pytorch/pull/96352, I hope to see this step retrying when there are network flakiness downloading those dependencies.  However, a network flaky issue today wasn't retried correctly as I expect https://github.com/pytorch/pytorch/actions/runs/4408570131/jobs/7724097579

It turns out that the exit statement in `install_deps_pytorch_xla` is too unforgiving and exit the shell script right way without giving retry the opportunity to perform.  As a function, it should return the failure exit code instead so that it can be handled by the callers accordingly


huydhn:
AFAICT, the failed GPU test look unrelated to this change.

huydhn:
@yeounoh What do you think about this change?

huydhn:
Any updates on this PR?  Pinging as our oncall starts to notice this failure on PyTorch, cc @ZainRizvi @clee2000

yeounoh (approved):
LGTM, please rebase before merging! Thank you @huydhn

PR #4768, opened 2023-03-13
[OpenXLA]Replace xrt_world_size with rt_world_size

ManfeiBai:
To deprecate XRT, replace cofig `XRT_SHARD_WORLD_SIZE` with `PJRT_SHARD_WORLD_SIZE`, and replace `xrt_world_size` with `rt_world_size`


ManfeiBai:
[test result]: All checks have passed

PR #4784, opened 2023-03-15
[Functionalization] Lower diagonal_scatter properly

alanwaketan:
Summary:
This change replaces the old implementation with DiagonalViewUpdate.

Test Plan:
PJRT_DEVICE=CPU python ../test/test_view_ops.py -v -k TestViewOpsXLA.test_diagonal

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1118,8 +1118,8 @@ at::Tensor XLANativeFunctions::diagonal_scatter(const at::Tensor& base,
   auto mutated_view_ = bridge::GetXlaTensor(mutated_view);
   return bridge::AtenFromXlaTensor(
       base_->CreateFrom(torch::lazy::MakeNode<DiagonalViewUpdate>(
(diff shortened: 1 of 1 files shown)


wonjoolee95 (approved):
LGTM! The CPU CI looks stuck for some reason, may want to re-trigger it.

On an unrelated note, I took a look at lowering `select_scatter` properly earlier today. I was hoping that I would be able to re-use an existing lowered op like this, but I don't think there is a suitable existing lowering that can do `select_scatter`. This would require me to lower the op from scratch manually, is this the correct understanding? Just wanted to double check with you.

alanwaketan:
> LGTM! The CPU CI looks stuck for some reason, may want to re-trigger it.
> 
> On an unrelated note, I took a look at lowering `select_scatter` properly earlier today. I was hoping that I would be able to re-use an existing lowered op like this, but I don't think there is a suitable existing lowering that can do `select_scatter`. This would require me to lower the op from scratch manually, is this the correct understanding? Just wanted to double check with you.

Check this out: https://github.com/pytorch/xla/blob/master/torch_xla/csrc/view.cpp#L81. I guess you can just re-apply the logic to your select_scatter.

PR #4787, opened 2023-03-15
Enable ubuntu:focal for Colab

yeounoh:
Colab has moved to `ubuntu:focal` as base image, and this one liner seems to enable the build.

Tested locally (docker) and on the cloud (GCB).


steventk-g (approved):
LGTM, thanks Yeounoh!

yeounoh:
> Do we need to cherry pick this into 2.0?

Probably not now, if we are going to cut 2.1 release branch for colab support.

PR #4788, opened 2023-03-15
[SPMD] Support partial replication.

yeounoh:
Enable partial replication where the input tensor is sharded across groups of devices (tiled) and replicated within each group (a.k.a. device subgroup). This is needed before we allow mapping lower rank tensor to a 3D mesh. cc @khatwanimohit @alanwaketan @jonb377

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -141,33 +141,24 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertTrue(torch.allclose(expected, actual))
 
   def test_mark_sharding_partial(self):
-    t1 = torch.randn(4, 4, 4, 4, device='cpu')
-    t2 = torch.randn(4, 4, 4, 4, device='cpu')
-    expected = t1 @ t2
+    device = xm.xla_device()
+    xt1 = torch.randn(4, 4).to(xm.xla_device())
+    xt2 = torch.randn(4, 4).to(xm.xla_device())
+    expected = (xt1 @ xt2).cpu()
 
-    xt1 = t1.to(xm.xla_device())
-    xt2 = t2.to(xm.xla_device())
     # Shard along two axes if four or more devices are available
     z_dim = 2 if self.n_devices >= 4 else 1
-    mesh = self._get_mesh((1, 1, z_dim, self.n_devices // z_dim))
-    xs.mark_sharding(xt1, mesh, (0, 1, 2, None))
-    xs.mark_sharding(xt2, mesh, (0, 1, None, 3))
+    mesh = self._get_mesh((z_dim, self.n_devices // z_dim))
+    xs.mark_sharding(xt1, mesh, (0,  None))
 
     # partial replication requires >1 devices; otherwise, it's replicated.
     if self.n_devices > 1:
       # xt1 is sharded `z_dim`-way, replicated `n_devices/z_dim`-way.
       self.assertTrue('last_tile_dim_replicate' in
                       torch_xla._XLAC._get_xla_sharding_spec(xt1))
-      self.assertTrue('[1,1,%d,1,%d]' %
+      self.assertTrue('[%d,1,%d]' %
                       (z_dim, self.n_devices //
                        z_dim) in torch_xla._XLAC._get_xla_sharding_spec(xt1))
-      # xt2 is sharded `n_devices/z_dim`-way, replicated `z_dim`-way.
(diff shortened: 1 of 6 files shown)


yeounoh:
@khatwanimohit let's sync offline sometime next week, so we can discuss the 3D sharding use case.

test/spmd/test_xla_sharding.py
```diff
@@ -50,9 +50,11 @@ def test_xla_shards(self):
...
       expected = torch.arange(start, end, dtype=torch.float32)
       self.assertTrue(torch.allclose(shard.data, expected))
-      self.assertIsInstance(shard.indices, list)
-      self.assertEqual(len(shard.indices), len(t.shape))
-      self.assertEqual(shard.indices[0], slice(start, end, 1))
+      if isinstance(shard.indices, list):
+        self.assertEqual(len(shard.indices), len(t.shape))
+        self.assertEqual(shard.indices[0], slice(start, end, 1))
+      else:
+        self.assertIsInstance(shard.indices, type(Ellipsis))
```
jonb377, line 57: Ah sharding across one device is `MAXIMAL` which becomes `REPLICATED`, so this happens on CPU 👍

test/spmd/test_xla_sharding.py
```diff
@@ -131,15 +137,61 @@ def test_mark_sharding_4d(self):
...
+    if self.n_devices > 1:
+      annotation = '{devices=[1,1,%d,%d]%s}' % (
+          z_dim, self.n_devices // z_dim, ','.join(
+              [str(i) for i in range(self.n_devices)]))
+      self.assertEqual(annotation, torch_xla._XLAC._get_xla_sharding_spec(xt))
 
     actual = (xt + xt).cpu()
     self.assertTrue(torch.allclose(expected, actual))
 
+  def test_mark_sharding_partial(self):
```
jonb377, line 150: Can we also test local_shards under partial replication?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -332,6 +405,13 @@ ShardingUtil::GetShardIndicesForDevices(
       int offset = i;
       std::vector<at::indexing::TensorIndex> indices;
       for (int j = tile_shape.size() - 1; j >= 0; j--) {
+        if (sharding.replicate_on_last_tile_dim() &&
+            j == tile_shape.size() - 1) {
+          // the last tile assignment dimension is replicated, which implies
+          // that the consecutive `tile_shape[j]` devices hold the replicated.
+          offset /= tile_shape[j];
+          continue;
+        }
```
jonb377, line 414: High-level question - how does `tile_shape` change when the tensor is partially replicated?

Right now, this logic makes sense to me if partial replication adds a dimension to track replica group and `tile_shape[j] == 1` on the replicated dimensions, is that a correct understanding?

yeounoh: That's correct.

test/spmd/test_xla_sharding.py
yeounoh: +1 yes, we could with the new shards access API now.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -36,7 +43,7 @@ std::vector<int64_t> TileAssignmentDimensions(
     XLA_CHECK(r.size() > 0)
         << "Invalid argument: empty list is not a valid element type.";
     dims.push_back(r.size());
-    auto type = r[0].attr("__class__").attr("__name__").cast<std::string>();
+    std::string type = GetPyType(r[0]);
```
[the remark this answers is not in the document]
yeounoh: I like it, let me take a look.

jonb377 (approved):
LGTM, thanks Yeounoh! 🚀

yeounoh:
Thanks for the comments, I will have a follow-up PR, since they are minor (+ i dont' have access to tpuvm at the moment... to test)

torch_xla/experimental/xla_sharding.py
```diff
@@ -71,6 +72,59 @@ def get_logical_mesh(self):
...
+  elif any(d is None for d in partition_spec):
+    sharding_type = ShardingType.PARTIAL
+  return sharding_type
+
+
+def _get_tile_assignment(mesh: Mesh) -> List[int]:
+  return mesh.get_logical_mesh().tolist()
+
+
+def _get_group_assignment(
```
alanwaketan, line 101: Can you explain a bit on what this helper does?

PR #4812, opened 2023-03-23
Add C++ test coverage

steventk-g:
This is a temporary method to gather CPP coverage while we wait for Bazel to land for Pytorch/XLA. It modifies the build clang process on CircleCI to build the executable with symbols to generate coverage, then gathers the generated output into an `.info` file for upload to an internal GCP bucket.

Generates CPP coverage files in CircleCI.

Passing CI: https://app.circleci.com/pipelines/github/pytorch/xla/18216/workflows/7156393f-9c9b-4516-9a1a-3864830eeefa/jobs/44264


test/cpp/run_tests.sh
```diff
@@ -6,6 +6,7 @@ BUILDTYPE="Release"
 VERB=
 FILTER=
 BUILD_ONLY=0
+USE_COVERAGE=1
```
yeounoh, line 9: Can we set it to 0 as default, and pass `USE_COVERAGE=1` when calling `run_tests.sh` else where? Like, https://github.com/pytorch/xla/blob/1217d33ba9253755b0320fbc71def094d7db5fc6/.circleci/config.yml#L191, guess this may be already doing that?

yeounoh: The above goes through. https://github.com/pytorch/xla/blob/1217d33ba9253755b0320fbc71def094d7db5fc6/.circleci/test.sh#L28,

yeounoh (changes requested):
Mostly LGTM, left a comment -- thanks @steventk-g

test/cpp/run_tests.sh
steventk-g: Good catch! Removed

stgpetrovic (approved):

yeounoh (approved):
LGTM

PR #4813, opened 2023-03-24
[FSDP] Don't use detach in _rebuild_full_params

alanwaketan:
Summary:
Somehow the `detach_copy` implementation has some issues with FSDP that it messed up with FSDP's memory management mechanism. Let's not use it.

Before:
image

After:
image

Test Plan:
CI.


ronghanghu (approved):
Thanks, looks good to me! (It's a bit unexpected to me that `detach` would make a difference, though)

alanwaketan:
> Thanks, looks good to me! (It's a bit unexpected to me that `detach` would make a difference, though)

Me too. That's why I tagged Brian on #4813 to see what's the correct way for detach. Thanks, Ronghang.

PR #4814, opened 2023-03-24
[Functionalization] Slightly improve detach_copy

alanwaketan:
Summary:
Somehow the current `detach_copy` has increased the memory usage of GPT-2 with FSDP a lot, see #4813. We may not implement it correctly. This fix won't fix the memory overhead as well.

Test Plan:
CI.

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1103,10 +1103,7 @@ at::Tensor XLANativeFunctions::cumsum(const at::Tensor& self, int64_t dim,
 // TODO(alanwaketan): Let's rewrite a without reusing other native functions.
 at::Tensor XLANativeFunctions::detach_copy(const at::Tensor& self) {
   TORCH_LAZY_FN_COUNTER("xla::");
-  auto new_tensor =
-      empty_symint(self.sym_sizes(), at::typeMetaToScalarType(self.dtype()),
-                   c10::nullopt, self.device(), c10::nullopt, c10::nullopt);
-  return set_(new_tensor, self);
+  return bridge::AtenFromXlaTensor(bridge::GetXlaTensor(self));
 }
 
 at::Tensor XLANativeFunctions::diag(const at::Tensor& self, int64_t diagonal) {


alanwaketan:
@bdhirsh I wonder how should I interpret detach_copy?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1092,10 +1092,7 @@ at::Tensor XLANativeFunctions::cumsum(const at::Tensor& self, int64_t dim,
 // TODO(alanwaketan): Let's rewrite a without reusing other native functions.
 at::Tensor XLANativeFunctions::detach_copy(const at::Tensor& self) {
   TORCH_LAZY_FN_COUNTER("xla::");
-  auto new_tensor =
-      empty_symint(self.sym_sizes(), at::typeMetaToScalarType(self.dtype()),
-                   c10::nullopt, self.device(), c10::nullopt, c10::nullopt);
-  return set_(new_tensor, self);
+  return bridge::AtenFromXlaTensor(bridge::GetXlaTensor(self));
```
alanwaketan, line 1095: Same as view_copy.

alanwaketan: Replace this with view_copy actually increase the memory footprint. So, the current implementation is actually better.

JackCaoG:
@alanwaketan is this still relevant?

alanwaketan:
Yea, let me rebase...

alanwaketan:
@wonjoolee95 Can I get a review?

torch_xla/csrc/aten_xla_type.cpp
JackCaoG: hmm, is this real a copy? Aren't we just return the same XLATensor back to the caller?

wonjoolee95:
> @bdhirsh I wonder how should I interpret detach_copy?

I wonder if we know the answer to this now. Is it okay for our `detach_copy` to just return the `XLATensor` directly without the actual copy as Jack mentioned above?

torch_xla/csrc/aten_xla_type.cpp
JackCaoG: for example if someone call `mark_sharding` on this XLATensor, I think it will affect both new and old tensors. I don't think that's the right behavior.

alanwaketan: Well, detach is never a copy. So I think it's okay: https://pytorch.org/docs/stable/generated/torch.Tensor.detach.html

Just like view_copy, we also don't copy anything. Instead, we make a view node. Therefore, I think here we just need to respect the original behavior  and no needs to deal with "copying".

wonjoolee95 (approved):

alanwaketan:
Thanks, Wonjoo.

PR #4822, opened 2023-03-27
Use ToXlaTensorArena for sharded data loader

jonb377:
Relying on `.to` in the data loader caused segfaults for sharded inputs. This change relies on the original `ToXlaTensorArena` implementation even for sharded data, applying sharding after the initial transfer.

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -1081,22 +1081,15 @@ def _maybe_convert_to_cpu(data, convert=True):
 
 
 def send_cpu_data_to_device(data, device, input_sharding=None):
-  if input_sharding:
-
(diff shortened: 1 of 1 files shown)


steventk-g:
Is there any test we can add that shows failing before this change / passing after?

yeounoh (approved):
LGTM, thanks Jon!

alanwaketan (approved):
LGTM. Is it trivial to provide a test case?

steventk-g (approved):
LGTM, added a comment about testing

yeounoh:
It's not trivial to create a test, esp. if it's a unit test. This is an implementation change to resolve a perf bug. Functionality is the same before and after the change, and the original issue (segfault) arise with a large batch size.

Issue #4742, opened 2023-03-09
[RFC] Supporting CPU fallback for the dynamo-xla bridge


wonjoolee95:
## 🚀 Feature

We propose PyTorch/XLA to leverage `CapabilityBasedPartitioner` from `torch.fx.passes.infra.partitioner` to support CPU fallback for the dynamo-xla bridge. The `CapabilityBasedPartitioner` a tool in torch.fx to partition the FX graph into "backend" parts, given a list of supported operators. The partitioner tries to form the largest subgraphs that only contain the supported ops.

## Motivation

In the current dynamo-xla integration, the dynamo bridge passes down a fx module and expects the backend to return a single function/hash representing the whole model. However, in the cases of that fx model containing the ops that PyTorch/XLA does not support, PyTorch/XLA will fallback at unsupported ops and generate more than one graph. In such cases, the current `torch_xla_bridge` will crash. 

This issue was originally tracked in an upstream issue [feature request](https://github.com/pytorch/torchdynamo/issues/1823). 

## Pitch

In order to support CPU fallback for the dynamo-xla bridge, we propose PyTorch/XLA to leverage `CapabilityBasedPartitioner` from `torch.fx.passes.infra.partitioner`. The `CapabilityBasedPartitioner` a tool in torch.fx to partition the FX graph into "backend" parts given a list of supported operators. This partitioner tries to forms the largest subgraphs that only contain the supported ops. 

Currently, other backends such as `nvFuser` and `onnxruntimer` leverage `CapabilityBasedPartitioner` to achieve a similar use case. 

### CapabilityBasedPartitioner

Example usage of `CapabilityBasedPartitioner` is shown below. First, we need to initialize `CapabilityBasedPartitioner` with the FX graph module and a list of supports ops. Then we can call the partitioner's `partition_and_fuse()` method to produce the `fused_graph_module`.
```
partitioner = CapabilityBasedPartitioner(graph_module, self.supported_ops)
fused_graph_module = partitioner.partition_and_fuse()
```
Now, we need to override fused_module's `__call__()` function with our custom backend function. This looks like: 
```
for node in fused_graph_module.graph.nodes:
    # Identify fused submodule
    if "fused_" in node.name:
        fused_module = getattr(fused_graph_module, node.name)
        fused_module._wrapped_call = my_backend_function
```

[...]

PR #4825, opened 2023-03-28
[WIP] Support CPU fallback for unsupported ops in dynamo

wonjoolee95:
Fixes https://github.com/pytorch/xla/issues/4742

---

[WIP] Support CPU fallback for unsupported ops in dynamo

TODO
- [ ] Verify correctness and add unit tests
- [ ] Support shape/input based fallback
- [ ] Add more comments

### DIFF
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -1,4 +1,5 @@
 import dataclasses
+import operator
 
 import functools
 import itertools
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -7,6 +7,7 @@ import time
 from typing import Any, Dict, List
 
 import torch
+from torch.fx.passes.infra.partitioner import CapabilityBasedPartitioner
 import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.debug.metrics as metrics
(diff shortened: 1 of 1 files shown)


wonjoolee95:
@seanlatias, a quick verification that we can start with would be check if dynamo works with some of the unsupported ops. For example, something like:
```
import torch
import torch_xla

  return torch.addmm(M, mat1, mat2, beta=beta)


M = torch.randn(2, 3, device=xm.xla_device())
mat1 = torch.randn(2, 3, device=xm.xla_device())
mat2 = torch.randn(3, 3, device=xm.xla_device())

res = fn_fallback(M, mat1, mat2, 0.5)
```

One more thing to verify would be to check if the `is_node_supported` function working as intended. We need to double check if the lines:
```
    return node.op == "call_function" and (getattr(node.target, "xla",
                                                   None) is not None or
                                           node.target == operator.getitem)
```
are enough to check if the node is supported by xla.

seanlatias:
@wonjoolee95 is there a known list of unsupported ops? The example you show doesn't work. I think it is because of the limitation you mentioned before: the current implementation could not check the operator-operand combination.

wonjoolee95:
Ah that's right, good catch. That's still a todo for this PR. So all ops that are supported by PyTorch/XLA are listed at https://github.com/pytorch/xla/blob/master/xla_native_functions.yaml. And all PyTorch ops are listed at https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/native_functions.yaml. So any op that's not in the PyTorch/XLA yaml file is not supported.

seanlatias:
With the following code, I identified two issues.
```python
import torch
import torch_xla
    
def fn_fallback(M):
    return torch.cummin(M, 1)                      
                                           
M = torch.randn(5, 10, device=xm.xla_device())
res = fn_fallback(M)                                       
```

1. We need to enable the `allows_single_node_partition`. Otherwise there would be no partition here.
2. The `partitioner.fuse_partitions()` seems not working properly. The fused graph is simply wrong. I'll see what's the problem.
3. Shouldn't we move the fallback ops to CPU? Or we just compile the ops that can be offloaded of XLA. In that case we should fuse the graph at a later stage. If we just partition and fuse, we will result in the same graph even if the `fuse_partitions()` function works properly.

seanlatias:
I think I find a workaround. Will push the code soon.

wonjoolee95:
Thanks for the investigation, @seanlatias. I guess another simple test case would have the `torchxla_trace_once` wrapped under a function with a supported op *and* an unsupported op. 

Also, regarding the `allows_single_node_partition`, the `CapacityBasedPartitioner` seems to accept a parameter that controls this -- https://github.com/pytorch/pytorch/blob/master/torch/fx/passes/infra/partitioner.py#L39. And it is `False` by default, so we many manually want to turn that flag to `True`.

torch_xla/core/dynamo_bridge.py
```diff
@@ -217,6 +241,9 @@ def extract_compiled_graph(xla_model: torch.fx.GraphModule, xla_args):
       tensor_id: i for i, tensor_id in enumerate(args_tensor_ids)
   }
 
+  # partition the graph for unsupported ops
+  xla_model = maybe_partition_graph(xla_model)
```
JackCaoG, line 245: now `xla_model` becomes a `partitioned_graph`, and what happen next? In below code we will trace through this model `xla_out = xla_model(*xla_args)`, are we going to get a single graph and a single hash?

PR #4855, opened 2023-04-05
Add flakiness to kokoro

steventk-g:
Adds necessary sponge tags to use kokoro for flakiness and greenness reporting.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -61,8 +61,6 @@ export_to_sponge_config() {
 
 flakiness()
 {
-  export_to_sponge_config "ng3_sponge_url" \
-    "https://fusion2.corp.google.com/invocations/${KOKORO_BUILD_ID}"
   export_to_sponge_config "ng3_commit" "$(git rev-parse HEAD)"
   export_to_sponge_config "ng3_cl_target_branch" "master"
   export_to_sponge_config "ng3_project_id" "pytorchxla"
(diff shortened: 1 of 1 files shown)


.kokoro/build.sh
```diff
@@ -44,6 +44,34 @@ pip install --user https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch
...
+export_to_sponge_config() {
+  local key="${1}"
+  local value="${2}"
+  printf "${key},${value}\n" >> "${CUSTOM_SPONGE_CONFIG}"
+}
+
+flakiness()
+{
+  export_to_sponge_config "ng3_sponge_url" \
+    "https://fusion2.corp.google.com/invocations/${KOKORO_BUILD_ID}"
```
miladm, line 65: How can we skip adding internal URLs to OSS code?

miladm (changes requested):

.kokoro/build.sh
steventk-g: I'll test to see if things work without this field. It doesn't seem like this is strictly required based on the documentation.

steventk-g: Looks like the field with a link is unnecessary. Removed.

vanbasten23:
I wonder how the metrics generated by Kokoro different from the one generated from bazel. Or is it bazel won't generate the flakiness/greeness metrics?

vanbasten23:
btw, can you point me to the kokoro documentation?

miladm (approved):

PR #4856, opened 2023-04-06
[WIP] Support non-fp32 weight all-gather in XLA FSDP inference pass

lsy323:
The goal of this PR is to support int8 weight all-gather in XLA FSDP inference pass. The motivation of adding this feature is to support shard int8 quantized weights in weight-only quantization workflow.

### Changes made to XLA FSDP impelementation:
Currently the change is not ready for review, it's for experimental purpose, so it's hacky now. In the current implementation, there is an assumption that the computation dtype and the weight type are the same. However, in weight-only quantized workflow, this assumption will break. An additional flag (quantized_weigths, we can have a better name for this) is introduce to specify the weight dtype and compute dtype are different.

### Testing Plan:
Comparing the numerical results of a quantized MNIST model (Linear layers' weights are quantized to int8) and a FSDP wrapped one. The numerical results are the same.

The HLO from the test is also attached in the PR for reference purpose, will remove upon completion
In the dump HLO, the all-reduce (Since all-gather in XLA is implemented by padding + all-reduce) now has int8 operand
Quantized:
```
%all-reduce.14 = (s8[320,50]{0,1}, s8[]) all-reduce(s8[320,50]{1,0} %get-tuple-element.8, s8[] %get-tuple-element.9), replica_groups={}, constrain_layout=true, to_apply=%AddComputation.10
```
bf16:
```
%all-reduce.141 = (bf16[52,320]{1,0}, bf16[]) all-reduce(bf16[52,320]{1,0} %get-tuple-element.135, bf16[] %get-tuple-element.136), replica_groups={}, constrain_layout=true, to_apply=%AddComputation.137
```

Example cmd:
```
PJRT_DEVICE=TPU \
    python3 test/fsdp_mnist_quant_test.py --auto_wrap_policy type_based \
    --shard_param_on_dim_0 --compute_dtype bfloat16 --quantized --compare_cpu
```

cc @alanwaketan


lsy323:
cc @ronghanghu @AlexWertheim @alanwaketan @miladm

lsy323:
As @alanwaketan pointed out, using non-float32 weights could cause training converge much slower, which is one of the reasons for having the float32 guard on the weights in the current code. Since quantization is only for inference purpose, I'll make changes to make the dtype guard to be relaxed only for inference pass.

lsy323:
> I suggest we find a way to not include the hlo files directly in the source code. one way is to take relevant snippets of the files and paste here. otherwise, we can point a link to the files stored elsewhere. wdyt?

Yes, that makes sense. I think including the command to run the tests and snippets is less distractive. Just removed the dump in the PR

PR #4857, opened 2023-04-06
[PJRT] Improve `HasValue()` error message

will-cromar:
The current error looks something like this:

```
RuntimeError: ./third_party/xla_client/pjrt_computation_client.h:167 : Check failed: HasValue() 
```

This can either mean the buffer is null _or_ a non-null buffer has been deleted by the runtime. But, we don't actually tell the user or developer which one of these is true, which makes debugging harder than it has to be.

cc @gkroiz


alanwaketan (approved):
This change is awesome. I have a hard time to reason about this error all the time...

PR #4861, opened 2023-04-06
Spmd dynamo materialization

steventk-g:


yeounoh:
Is this still needed?

steventk-g:
Not required - users can work around the tensor materialization issue by setting up the data loader to use XLA tensors.

PR #4862, opened 2023-04-07
[WIP] Integrate Dynamo + SPMD for Inference

steventk-g:
Includes @yeounoh's changes to use SPMD with `torch.compile`


test/spmd/test_dynamo_spmd_inference_latency.py
```diff
@@ -0,0 +1,114 @@
+"""
```
yeounoh, line 1: Let's remove this script before merging

test/spmd/test_inference_spmd_dynamo_imagenet.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def inference_imagenet():
+  print('==> Preparing data..')
+  img_dim = get_model_property('img_dim')
+  device = xm.xla_device()
+  if FLAGS.fake_data:
+    assert FLAGS.test_set_batch_size == 1
+    test_loader = xu.SampleGenerator(
+        data=(torch.zeros(FLAGS.test_set_batch_size, 3, img_dim, img_dim).to(device),
```
yeounoh, line 96: Maybe we comment on that we needed `to(device)` here for spmd?

yeounoh (changes requested):
Mostly LGTM, added a few minor comments. Also we need to address linter checks.

test/spmd/test_inference_spmd_dynamo_imagenet.py
```diff
@@ -0,0 +1,238 @@
...
+  model = torchvision.models.resnet50().to(
+      device)  # get_model_property('model_fn')().to(device)
+
+  input_mesh = None
+  if FLAGS.sharding:
+    num_devices = pjrt.global_device_count()
+    device_ids = np.arange(num_devices)
+    # Model sharding
+    if 'conv' in FLAGS.sharding:
+      # Shard the model's convlution layers along two dimensions
```
yeounoh, line 136: typo "convolution"

test/spmd/test_inference_spmd_dynamo_imagenet.py
```diff
@@ -0,0 +1,238 @@
...
+        test_dataset,
+        batch_size=FLAGS.test_set_batch_size,
+        sampler=None,
+        drop_last=FLAGS.drop_last,
+        shuffle=False)
+
+  torch.manual_seed(42)
+
+  model = torchvision.models.resnet50().to(
+      device)  # get_model_property('model_fn')().to(device)
```
yeounoh, line 128: Need to clean this up, either use `get_model_property('model_fn')().to(device)` or stick to `resnet50()`, in which case we don't need some of the model props code above.

test/dynamo/test_bridge.py
```diff
@@ -9,6 +9,7 @@
 
 import torch_xla.core.dynamo_bridge as bridge
 import torch_xla.core.xla_model as xm
+import torch_xla.experimental.xla_sharding as xs
```
yeounoh, line 12: this is needed?

test/dynamo/test_bridge.py
```diff
@@ -108,6 +109,7 @@ def test_wrapper(self):
     optimized_mod = bridge.extract_compiled_graph(
         fx.symbolic_trace(xla_module), inputs)
 
+    print('\nCompiled graph extraction done!\n')
```
yeounoh, line 112: this is a debugging probe?

test/tpu/xla_test_job.yaml
```diff
@@ -44,7 +44,8 @@ spec:
       python3 pytorch/xla/test/pjrt/test_experimental_pjrt_tpu.py
       python3 pytorch/xla/test/spmd/test_xla_sharding.py
       python3 pytorch/xla/test/spmd/test_xla_virtual_device.py
-      python3 pytorch/xla/test/spmd/test_train_spmd_linear_model.py
+      XLA_USE_SPMD=1 python3 pytorch/xla/test/spmd/test_train_spmd_linear_model.py --sharding batch
+      XLA_USE_SPMD=1 python3 pytorch/xla/test/spmd/test_inference_spmd_dynamo_imagenet.py  --fake_data --use_dynamo --sharding batch
```
yeounoh, line 48: awesome, thank you!

torch_xla/core/dynamo_bridge.py
```diff
@@ -190,6 +190,10 @@ def add_nones(self, value_list):
 
 
 def is_xla_tensor(tensor: torch.Tensor) -> bool:
+  # TODO(yeounoh) check if tensor sharding annotation can be accessed here
```
yeounoh, line 193: I think we can remove my comment here now.

yeounoh (changes requested):
Added more comments

test/spmd/test_inference_spmd_dynamo_imagenet.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+      test_loader = pl.MpDeviceLoader(
+          test_loader,
+          device,
+          input_sharding=xs.ShardingSpec(input_mesh, (0, 1, 2, 3)))
+
+  model.eval()
+  xm.mark_step()
+  xm.wait_device_ops()
+  met.clear_all()
```
JackCaoG, line 191: why are these needed? I had these in my dynamo test because test data is fake so I need a `mark_step` to materalize the fake input, if that;'s not the case here we don't need to do a mark_step here.

test/spmd/test_inference_spmd_dynamo_imagenet.py
```diff
@@ -0,0 +1,238 @@
...
+    writer = test_utils.get_summary_writer(FLAGS.logdir)
+
+  def inference_loop_fn(loader):
+    start_0 = time.time()
+    for step, (data, _) in enumerate(loader):
+      if step == 1:
+        start_1 = time.time()
+      output = model(data)
+    end = time.time()
+    return end - start_0, end - start_1
```
JackCaoG, line 208: If this test is inference only, we should surface that in the test name.

test/spmd/test_inference_spmd_dynamo_imagenet.py
```diff
@@ -0,0 +1,238 @@
+import args_parse
```
JackCaoG, line 1: We also need to basic test similar to https://github.com/pytorch/xla/blob/master/test/dynamo/test_dynamo.py#L45 to check the correctness of the dynamo + SPMD. In this case I guess you should expect `dynamo + SPMD` yield the same result as `SPMD` along to test this integration. Test can be either under `SPMD` dir or `dynamo` dir. Currently we don't even know if SPMD is emitting the correct result when used with dynamo(I suppose we did some e2e test).

test/tpu/xla_test_job.yaml
JackCaoG: How long will this test take? It is inference so I guess it won't be too bad but I am trying to prevent us from adding too many burden to the already very long CI.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -613,17 +621,41 @@ XLAGraphExecutor::ExecuteComputationWithBarrier(
   std::shared_ptr<XLAGraphExecutor::Async> async = std::make_shared<Async>(
       &coll, std::move(arguments), placeholders, std::move(cachedComputation));
 
-  auto syncfn = [async, hash]() {
+  auto syncfn = [async, hash, sharding_specs]() {
```
JackCaoG, line 624: why `sharding_specs` needs to be passed from original function? It is init as default value right? We might as well just init it within this function.

JackCaoG: ok, I guess the point here is we need to supply a proper sharding spec for the output, but it is currently not supported yet. As a follow up, what's the right way to determine the sharding spec for output?

JackCaoG (commented):
mostly LGTM.

FYI @wonjoolee95 in this pr we break a really important assumption of the dynamo which is "same fx graph will result in same IR/HLO graph". In the SPMD case, even if the fx graph is the same, we can annotate the input XLATensor and make it generate different IR/HLO graph.

There is one thing that is still unclear to me and I hope @yeounoh and maybe @alanwaketan can give me some answer. AFAIK, during the inital phase of the dynamo `extract_compile_graph`, dynamo will pass us the fx graph along with the fake XLATensor with same shape with the real model. Does dyanmo somehow correctly move the sharding annotation from the real XLATensor to the Fake XLATensor? Otherwise I would expect we hit a cache miss when we execute the real SPMD computation.

test/spmd/test_inference_spmd_dynamo_imagenet.py
steventk-g: Done

test/dynamo/test_bridge.py
steventk-g: Removed

test/dynamo/test_bridge.py
steventk-g: Removed

test/dynamo/test_bridge.py
```diff
@@ -117,6 +119,8 @@ def test_wrapper(self):
       # make sure above lazy computation is executed.
       xm.mark_step()
 
+      print(f'\nexpected: {expected}\n')
```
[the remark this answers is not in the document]
steventk-g: Removed

test/spmd/test_inference_spmd_dynamo_imagenet.py
steventk-g: Where should we add a reference to inference? The main function in this file is `inference_imagenet` and the file name is `test/spmd/test_inference_spmd_dynamo_imagenet.py`

test/spmd/test_inference_spmd_dynamo_imagenet.py
steventk-g: We need to materialize the input here as well. @yeounoh may be able to add more context

JackCaoG: ok, maybe add a comment above to mentioned that this `mark_step` is to materialize the fake input then.

test/spmd/test_inference_spmd_dynamo_imagenet.py
JackCaoG: oh I miss the inference in the name, my bad.

JackCaoG:
@steventk-g let me know when you want me to take another look

test/spmd/test_inference_spmd_dynamo_imagenet.py
steventk-g: Tried adding a test, currently crashing on `res_xla_spmd_dynamo = self.fn_simple_dynamo(sharded_x, sharded_y)`

test/spmd/test_inference_spmd_dynamo_imagenet.py
steventk-g: Done

torch_xla/csrc/xla_graph_executor.cpp
steventk-g: Good question, I'm not sure the right way to determine sharding spec for output. @yeounoh might be able to answer or provide more context

JackCaoG: That's fine, we will leave that as a follow up.

test/spmd/test_inference_spmd_dynamo_imagenet.py
JackCaoG: let me know if you need help.

JackCaoG:
I am taking over this commit, since it is on @steventk-g 's fork and it already has conflcits with the master, I will just open a new pr and patch from master.

alanwaketan:
> I am taking over this commit, since it is on @steventk-g 's fork and it already has conflcits with the master, I will just open a new pr and patch from master.

Let me know when it's ready. I can review it.

JackCaoG:
Close this one in favor of https://github.com/pytorch/xla/pull/5002

PR #4867, opened 2023-04-10
[SPMD] Enable readonly shards on host with DTensor API

steventk-g:


torch_xla/experimental/xla_distributed_tensor.py
```diff
@@ -0,0 +1,108 @@
...
+class _ToTorchTensor(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, input: 'XLAShard'):
+        ctx.dtensor_spec = input._spec
+        return input._local_tensor.detach()
+
+    @staticmethod
+    def backward(ctx, grad_output: torch.Tensor):
+        dtensor_spec = ctx.dtensor_spec
+        dtensor_meta = dtensor_spec.tensor_meta
```
wanchaol, line 64: do you need autograd enabled in the `to_local()` API? If not it doesn't have to be an autograd function

PR #4868, opened 2023-04-11
Enable dynamism on xla::add

vanbasten23:
- created another SymIntElements ctor
- add some change
- add more logging.

### DIFF
--- a/torch_xla/csrc/helpers.cpp
+++ b/torch_xla/csrc/helpers.cpp
@@ -481,11 +481,63 @@ xla::Shape XlaHelpers::GetPromotedShape(const xla::Shape& shape1,
 
 xla::Shape XlaHelpers::GetPromotedBinaryOpShape(const xla::Shape& shape1,
                                                 const xla::Shape& shape2) {
-  return xla::ShapeUtil::MakeShape(
-      PromoteType(shape1.element_type(), shape2.element_type()),
-      torch::lazy::GetPromotedShape(
-          xla::util::ToVector<int64_t>(shape1.dimensions()),
-          xla::util::ToVector<int64_t>(shape2.dimensions())));
+  if (!shape1.is_dynamic() && !shape2.is_dynamic()) {
+    return xla::ShapeUtil::MakeShape(
+        PromoteType(shape1.element_type(), shape2.element_type()),
+        torch::lazy::GetPromotedShape(
+            xla::util::ToVector<int64_t>(shape1.dimensions()),
+            xla::util::ToVector<int64_t>(shape2.dimensions())));
+  }
+  return GetPromotedDynamicShape(shape1, shape2);
+}
+
+xla::Shape XlaHelpers::GetPromotedDynamicShape(const xla::Shape& shape1,
+                                               const xla::Shape& shape2) {
+  std::vector<int64_t> upper_bounds1 = xla::util::ToVector<int64_t>(shape1.dimensions());
+  std::vector<int64_t> upper_bounds2_cp = xla::util::ToVector<int64_t>(shape1.dimensions());
+
+  std::vector<int64_t> upper_bounds2 = xla::util::ToVector<int64_t>(shape2.dimensions());
+  absl::Span<const bool> dyn_dims1 = shape1.dynamic_dimensions();
+  absl::Span<const bool> dyn_dims2 = shape2.dynamic_dimensions();
+  std::vector<int64_t> upper_bounds;
+  std::vector<bool> dyn_dims;
+
+  if (upper_bounds1.size() > upper_bounds2.size()) {
+    upper_bounds.insert(
+      upper_bounds.end(),
+      upper_bounds1.begin(),
+      upper_bounds1.begin() + (upper_bounds1.size() - upper_bounds2.size()));
+    dyn_dims.insert(
+      dyn_dims.end(),
+      dyn_dims1.begin(),
+      dyn_dims1.begin() + (dyn_dims1.size() - dyn_dims2.size()));
+  } else {
+    upper_bounds.insert(
+      upper_bounds.end(),
+      upper_bounds2.begin(),
+      upper_bounds2.begin() + (upper_bounds2.size() - upper_bounds1.size()));
+    dyn_dims.insert(
+      dyn_dims.end(),
+      dyn_dims2.begin(),
+      dyn_dims2.begin() + (dyn_dims2.size() - dyn_dims1.size()));
+  }
+  size_t min_size = std::min(upper_bounds1.size(), upper_bounds2.size());
+  for (const auto i : c10::irange(min_size)) {
+    int64_t ubound1 = upper_bounds1[upper_bounds1.size() - min_size + i];
+    int64_t ubound2 = upper_bounds2[upper_bounds2.size() - min_size + i];
+    upper_bounds.push_back(std::max<int64_t>(ubound1, ubound2));
+    
+    bool ddim1 = dyn_dims1[dyn_dims1.size() - min_size + i];
+    bool ddim2 = dyn_dims2[dyn_dims2.size() - min_size + i];
+    dyn_dims.push_back(ddim1||ddim2);
+  }
+  const xla::Shape& promoted_shape = xla::ShapeUtil::MakeShape(
+    PromoteType(shape1.element_type(), shape2.element_type()),
+    upper_bounds, dyn_dims);
+  
+  std::cout << "xw32, file=" << __FILE__ << ", line=" << __LINE__ << "function=" << __FUNCTION__ << ": promoted_shape=" << promoted_shape << std::endl;
+  
+  return promoted_shape;
 }
 
 std::pair<xla::XlaOp, xla::XlaOp> XlaHelpers::PromoteShapes(xla::XlaOp op1,
--- a/torch_xla/csrc/helpers.cpp
+++ b/torch_xla/csrc/helpers.cpp
@@ -523,11 +523,12 @@ xla::Shape XlaHelpers::GetPromotedDynamicShape(const xla::Shape& shape1,
   for (const auto i : c10::irange(min_size)) {
     int64_t ubound1 = upper_bounds1[upper_bounds1.size() - min_size + i];
     int64_t ubound2 = upper_bounds2[upper_bounds2.size() - min_size + i];
-    int64_t ubound = std::max<int64_t>(ubound1, ubound2);
-    upper_bounds.push_back(ubound);
-    
     bool ddim1 = dyn_dims1[dyn_dims1.size() - min_size + i];
     bool ddim2 = dyn_dims2[dyn_dims2.size() - min_size + i];
+    XLA_CHECK((!ddim1&&!ddim2)||(ddim1&&!ddim2&&ubound2==1)||(ddim2||!ddim1||ubound1==1)) << "At dimension " << i << ", operand1 has dimension size " << ubound1 << " isDynamic=" << ddim1 << " vs operand2 has dimension size " << ubound2 << " isDynamic=" << ddim2;
+
+    int64_t ubound = std::max<int64_t>(ubound1, ubound2);
+    upper_bounds.push_back(ubound);
     bool ddim = ddim1||ddim2;
     dyn_dims.push_back(ddim);
     std::cout << "xw32, file=" << __FILE__ << ", line=" << __LINE__ << "function=" << __FUNCTION__ << ": ubound=" << ubound << ", ddim=" << ddim << std::endl;
--- a/torch_xla/csrc/helpers.cpp
+++ b/torch_xla/csrc/helpers.cpp
@@ -494,8 +494,6 @@ xla::Shape XlaHelpers::GetPromotedBinaryOpShape(const xla::Shape& shape1,
 xla::Shape XlaHelpers::GetPromotedDynamicShape(const xla::Shape& shape1,
                                                const xla::Shape& shape2) {
   std::vector<int64_t> upper_bounds1 = xla::util::ToVector<int64_t>(shape1.dimensions());
-  std::vector<int64_t> upper_bounds2_cp = xla::util::ToVector<int64_t>(shape1.dimensions());
-
   std::vector<int64_t> upper_bounds2 = xla::util::ToVector<int64_t>(shape2.dimensions());
   absl::Span<const bool> dyn_dims1 = shape1.dynamic_dimensions();
   absl::Span<const bool> dyn_dims2 = shape2.dynamic_dimensions();
@@ -525,11 +523,14 @@ xla::Shape XlaHelpers::GetPromotedDynamicShape(const xla::Shape& shape1,
   for (const auto i : c10::irange(min_size)) {
     int64_t ubound1 = upper_bounds1[upper_bounds1.size() - min_size + i];
     int64_t ubound2 = upper_bounds2[upper_bounds2.size() - min_size + i];
-    upper_bounds.push_back(std::max<int64_t>(ubound1, ubound2));
+    int64_t ubound = std::max<int64_t>(ubound1, ubound2);
+    upper_bounds.push_back(ubound);
     
     bool ddim1 = dyn_dims1[dyn_dims1.size() - min_size + i];
     bool ddim2 = dyn_dims2[dyn_dims2.size() - min_size + i];
-    dyn_dims.push_back(ddim1||ddim2);
+    bool ddim = ddim1||ddim2;
+    dyn_dims.push_back(ddim);
+    std::cout << "xw32, file=" << __FILE__ << ", line=" << __LINE__ << "function=" << __FUNCTION__ << ": ubound=" << ubound << ", ddim=" << ddim << std::endl;
   }
   const xla::Shape& promoted_shape = xla::ShapeUtil::MakeShape(
     PromoteType(shape1.element_type(), shape2.element_type()),
--- a/torch_xla/csrc/helpers.cpp
+++ b/torch_xla/csrc/helpers.cpp
@@ -502,7 +502,8 @@ xla::Shape XlaHelpers::GetPromotedDynamicShape(const xla::Shape& shape1,
   std::vector<int64_t> upper_bounds;
   std::vector<bool> dyn_dims;
 
-  // See https://pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics
+  // See
+  // https://pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics
   if (upper_bounds1.size() > upper_bounds2.size()) {
     upper_bounds.insert(
         upper_bounds.end(), upper_bounds1.begin(),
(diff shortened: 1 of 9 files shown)


JackCaoG:
@vanbasten23 based on our conversation with Blake last time, I think XLA will execute whatever you passed to them and not throw an error. For us I can only see two ways
1. materialize the size (not desirable)
2. do a check based on the bound. For example if we try to add a `1, <=3` and `1, <=6`, since `<=3` and `<=6` might not be compatible, we will just throw an error.

I don't think we should do 1, but I am not sure if we should do 2 because that will introduce a lot of check in our code base. What does pytorch upstream do? Did they just disable all size check if input is dynamic?

vanbasten23:
hi @ezyang , would you mind pointing to me how PyTorch handles `torch.add` under the context of dynamic shape? Specifically, in torch_xla, to check if 2 operands of `torch.add` are broadcastable (according to the [rule](https://pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics)) we do something like [this](https://github.com/pytorch/pytorch/blob/0eb59ad0937e07ce5cdf0993e87efa797675ed4f/torch/csrc/lazy/core/helpers.cpp#L107-L113). But doing so would trigger an compilation and execution (consequently a graph break) in order to get the real size and that doesn't sound ideal. I assume upstream PyTorch has similar problems. I wonder how PyTorch deal with the issue and if there is a way to avoid the graph break.

I was searching in /workspaces/work/pytorch/aten/src/ATen/native but didn't find it. Sorry I'm not very familiar with pytorch code base.

ezyang:
Ah, this is what unsound 0/1 specialization on unbacked symints is for. We assume that the result of, e.g., nonzero, NEVER compares equal to 0 or 1 (even though there's no reason it couldn't compare equal.) Thus the only nontrivial test is x == y, which will likely be true without requiring a break.

vanbasten23:
> Thus the only nontrivial test is x == y, which will likely be true without requiring a break.

Thanks for the response. Do you mean we can assume x == y hence skip the check? I tried to do that but XLA would give incorrect results, for example in this PR I have this test:
```
  def test_add(self):
    ...
    # t2.shape=torch.Size([<=6, 2]) with real size [4, 2]
    # t2=[[0,0], [0,2], [0,3], [0,5]]
    # t4.shape=torch.Size([<=6, 2]) with real size [2, 2]
    # t4=[[0,0], [0,1]]
    t5 = torch.add(t2, t4)
    print(t5)
```
`print(t5)` prints [[0, 0], [0, 3]] but I expect doing `torch.add(t2, t4)` should throw a runtime error. So it seems we have to check x == y somehow.

ezyang:
Sorry I wasn't clear. It's not that you can skip the check, but that I expect that x and y would be the same symbol and so you'd be able to pass the check without forcing running the net.

vanbasten23:
hi @ezyang , I'll post my questions here as there was not enough time to cover them in our meeting.

As follow-up questions to how pytorch handles torch.add broadcasting ((rule symInt1==1 || symInt2==1 || symInt1==symInt2)) with dynamic tensors,
1. Regarding your [comment](https://github.com/pytorch/xla/pull/4868#issuecomment-1528868580) "but that I expect that x and y would be the same symbol", for torch.add, when we check if the 2 operands are broadcastable (specifically for the check, symInt1 == symInt2), why can you “expect that x and y would be the same symbol”? Shouldn’t the 2 operands be independent?
2. For the check `symInt1==1 || symInt2==1`, you mentioned unbacked symints never compare equal to 0/1. Then for example:
```
tensor1.size: 2  1(<=6)
tensor2.size: 2  2
torch.add(tensor1, tensor2)
```
Pytorch eager mode would determine t1 and t2 are broadcastable. How does pytorch do for this case with dynamic shape? Would pytorch assume tensor1.size()[1] is not 1 and skip the broadcasting check?

ezyang:
> Pytorch eager mode would determine t1 and t2 are broadcastable. How does pytorch do for this case with dynamic shape? Would pytorch assume tensor1.size()[1] is not 1 and skip the broadcasting check?

We don't. 👹  We intentionally error anyway, even though eager would have "accidentally" allowed the broadcast.

> why can you “expect that x and y would be the same symbol”? Shouldn’t the 2 operands be independent?

If these are truly unbacked symbols, there must be *some* reason to believe they are equal a priori. If you add x.nonzero() + y.nonzero(), there is no reason to believe these tensors actually have the same size, and we'll just always error (even though, if the stars aligned, they could have the same result.) If you know, out-of-band, that two independent operands must be equal, then you can use `constrain_unify` to declare "these must be equal!" But by the nature of thing, this cannot be a synchronous check; at best, it can be checked asynchronously only when we actually have the values.

torch_xla/csrc/helpers.cpp
```diff
@@ -481,11 +481,62 @@ xla::Shape XlaHelpers::GetPromotedShape(const xla::Shape& shape1,
...
+      xla::util::ToVector<int64_t>(shape1.dimensions());
+  std::vector<int64_t> upper_bounds2 =
+      xla::util::ToVector<int64_t>(shape2.dimensions());
+  absl::Span<const bool> dyn_dims1 = shape1.dynamic_dimensions();
+  absl::Span<const bool> dyn_dims2 = shape2.dynamic_dimensions();
+  std::vector<int64_t> upper_bounds;
+  std::vector<bool> dyn_dims;
+
+  if (upper_bounds1.size() > upper_bounds2.size()) {
+    upper_bounds.insert(
```
JackCaoG, line 506: some comment here would be nice. also what happen if `upper_bounds1.size() == upper_bounds2.size()`?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -652,8 +652,14 @@ XLATensorPtr abs(const XLATensorPtr& input) {
 XLATensorPtr add(const XLATensorPtr& input, const XLATensorPtr& other,
                  const at::Scalar& alpha,
                  c10::optional<at::ScalarType> logical_element_type) {
+  SymIntElements sym_int_elements(other->GetIrValue());
```
JackCaoG, line 655: should we separate the static and dynamic case here?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -282,13 +291,6 @@ torch::lazy::Value XLAGraphExecutor::GetIrValueForScalar(
   return GetIrValueForScalar(value, type, shape.dimensions(), device);
 }
 
-torch::lazy::Value XLAGraphExecutor::GetIrValueForScalar(
```
JackCaoG, line 285: any reason to move this function up?

vanbasten23: reverted the change.

torch_xla/csrc/tensor_methods.cpp
vanbasten23: Since SymInt can be either static or dynamic, so I feel it's ok not separate the 2 cases. If CI is green, then this change should be good for the static case. wdyt?

JackCaoG: this will result in static code path to run unnecessary dynamic related code and make tracing slower. I'd prefer to separate the cases.

torch_xla/csrc/helpers.cpp
vanbasten23: If `upper_bounds1.size() == upper_bounds2.size()`, then line 512-516 should be no op. The python tests in this PR test this case.

Also added a comment for the broadcast rule.

torch_xla/csrc/tensor_methods.cpp
vanbasten23: done.

Btw, how can I see the tracing time? Is it a metric in `met.metrics_report()`?

vanbasten23:
The current issue is that the test `ERROR: test_index_types (__main__.TestAtenXlaTensor)` fails with error:
```
  File "/tmp/pytorch/xla/test/test_operations.py", line 1637, in test_index_types
    self.runAtenTest([
  File "/tmp/pytorch/xla/test/test_utils.py", line 391, in runAtenTest
    xla_results = xu.as_list(fn(*xla_tensors))
  File "/tmp/pytorch/xla/test/test_operations.py", line 1635, in test_fn
    return [x[idx] for idx in indices]
  File "/tmp/pytorch/xla/test/test_operations.py", line 1635, in <listcomp>
    return [x[idx] for idx in indices]
RuntimeError: torch_xla/csrc/helpers.cpp:527 : Check failed: (!ddim1 && !ddim2) || (ddim1 && !ddim2 && ubound2 == 1) || (ddim2 && !ddim1 && ubound1 == 1) 
*** Begin stack trace ***
	torch_xla::XlaHelpers::GetPromotedDynamicShape(xla::Shape const&, xla::Shape const&)
	torch_xla::XlaHelpers::GetPromotedBinaryOpShape(xla::Shape const&, xla::Shape const&)
	torch_xla::operator+(torch::lazy::Value const&, torch::lazy::Value const&)
	torch_xla::IndexByTensors(c10::intrusive_ptr<torch_xla::XLATensor, c10::detail::intrusive_target_default_null_type<torch_xla::XLATensor> > const&, absl::lts_20230125::Span<c10::intrusive_ptr<torch_xla::XLATensor, c10::detail::intrusive_target_default_null_type<torch_xla::XLATensor> > const>, long)
	torch_xla::tensor_methods::index(c10::intrusive_ptr<torch_xla::XLATensor, c10::detail::intrusive_target_default_null_type<torch_xla::XLATensor> > const&, absl::lts_20230125::Span<c10::intrusive_ptr<torch_xla::XLATensor, c10::detail::intrusive_target_default_null_type<torch_xla::XLATensor> > const>, long)
	torch_xla::XLANativeFunctions::index(at::Tensor const&, c10::List<c10::optional<at::Tensor> > const&)
	at::_ops::index_Tensor::redispatch(c10::DispatchKeySet, at::Tensor const&, c10::List<c10::optional<at::Tensor> > const&)
	at::_ops::index_Tensor::call(at::Tensor const&, c10::List<c10::optional<at::Tensor> > const&)
	torch::autograd::THPVariable_getitem(_object*, _object*)
...
At dimension 0, operand1 has dimension size 10 isDynamic=1 vs operand2 has dimension size 10 isDynamic=0
```

I spent some time investigating it. A small repro is:
```
    idx_tensor = torch.randint(0, 1, size=(10,), dtype=torch.bool)
    xla_tensor = idx_tensor.to(device).detach().requires_grad_(idx_tensor.requires_grad)
    x = torch.arange(10).to(device)
    res = x[xla_tensor]
```
The reason why it failed is that `tensor_methods::index` internally calls `nonzero` ([here](https://github.com/pytorch/xla/blob/30b2aba0ec2d9f7c664025f805bf4677c0f8a13e/torch_xla/csrc/ops/index_ops.cpp#L65)), then when we sum over 2 torch::lazy::Value [here](https://github.com/pytorch/xla/blob/30b2aba0ec2d9f7c664025f805bf4677c0f8a13e/torch_xla/csrc/ops/index_ops.cpp#L152) it failed. If failed even if I change [these lines](https://github.com/pytorch/xla/blob/30b2aba0ec2d9f7c664025f805bf4677c0f8a13e/torch_xla/csrc/ops/index_ops.cpp#L151-L155) to:
```
    if (!dim_index->shape().get().is_dynamic()) {
      wrapped_dim_index = XLATensor::Create(
        dim_index->GetIrValue() +
            XLAGraphExecutor::Get()->GetIrValueForScalar(
                dim_size, dim_index->shape(), base->GetDevice()),
        base->GetDevice());
    } else {
      SymIntElements sym_int_elements(dim_index->GetIrValue());
      xla::PrimitiveType primitive_type = dim_index->shape().get().element_type();
      wrapped_dim_index = XLATensor::Create(
        dim_index->GetIrValue() +
            XLAGraphExecutor::Get()->GetIrValueForScalar(
                dim_size, sym_int_elements, primitive_type, base->GetDevice()),
        base->GetDevice());
    }
```
With the new change, the error basically remain the same except for the bold font part.:
```
  File "pytorch/xla/test/test_operations.py", line 441, in test_index_types
    res = x[xla_tensor]
RuntimeError: torch_xla/csrc/helpers.cpp:528 : Check failed: (!ddim1 && !ddim2) || (ddim1 && !ddim2 && ubound2 == 1) || (ddim2 && !ddim1 && ubound1 == 1) 
*** Begin stack trace ***
        torch_xla::XlaHelpers::GetPromotedDynamicShape(xla::Shape const&, xla::Shape const&)
        torch_xla::XlaHelpers::GetPromotedBinaryOpShape(xla::Shape const&, xla::Shape const&)
At dimension 0, operand1 has dimension size 10 isDynamic=1 vs operand2 has dimension size 10 **isDynamic=1**
```


Note to myself:
```
>>> import torch
>>> t=torch.tensor([1, 0])
>>> t=torch.tensor([[0, 0]])
>>> torch.select(t, 1, 0)
tensor([0])
>>> torch.select(t, 0, 0)
tensor([0, 0])
>>> t1=torch.tensor([0, 0, 1, 1])
>>> t2=torch.nonzero(t1)
>>> t2
tensor([[2],
        [3]])

>>> x = torch.arange(10)
>>> idx_long=torch.randint(0, 1, size=(10,), dtype=torch.long)
>>> x[idx_long]
tensor([0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
>>> idx_bool=torch.randint(0, 1, size=(10,), dtype=torch.bool)
>>> x[idx_bool]
tensor([], dtype=torch.int64)

>>> import torch_xla, torch_xla.core.xla_model as xm
>>> device = xm.xla_device()
>>> idx_bool_xla=torch.randint(0, 1, size=(10,), dtype=torch.bool, device=device)
>>> x_xla = torch.arange(10).to(device)
>>> x_xla[idx_bool_xla]
tensor([], device='xla:0', dtype=torch.int64)

>>> idx_bool_xla=torch.randint(0, 1, size=(10,), dtype=torch.bool, device=device)
>>> x_xla = torch.arange(10).to(device)
>>> x_xla[idx_bool_xla]
tensor([], device='xla:0', dtype=torch.int64)
>>> x = torch.arange(10)
>>> idx_bool=torch.randint(0, 1, size=(9,), dtype=torch.bool)
>>> x[idx_bool]
IndexError: The shape of the mask [9] at index 0 does not match the shape of the indexed tensor [10] at index 0
```

vanbasten23:
hi @JackCaoG @miladm , I'm currently blocked by the [issue](https://github.com/pytorch/xla/pull/4868#issuecomment-1552120188) above. Do you know for a `xla::Shape shape`, how to get its real dimension size (in terms of int)?

JackCaoG:
I think I know what's going on, correct me if I am wrong

you are doing something similar to 
```
// t1 shape is [<=10]
t2 = t1 + 1
```
and in the code we expand the scalar to t1's shape. However because some implementation detail, we think both value has dynamic shapes hence we can't add them. If you commented out the check, would result be correct?

vanbasten23:
> and in the code we expand the scalar to t1's shape. However because some implementation detail, we think both value has dynamic shapes hence we can't add them. If you commented out the check, would result be correct?

What you described looks correct to me. What is happening is "XLANativeFunction::index" calls `nonzero` which introduces dynamism, then it calls TensorMethods::index which calls function [WrapIndicesOnce](https://github.com/pytorch/xla/blob/3150573fcafa06b6f5e4321964f9b437f224d37a/torch_xla/csrc/ops/index_ops.cpp#L152-L154) where we add 2 dynamic IrValues. So indeed: t2=t1+1. And by removing the check:
```
 XLA_CHECK((!ddim1 && !ddim2) || (ddim1 && !ddim2 && ubound2 == 1) ||
               (ddim2 && !ddim1 && ubound1 == 1))
         << "At dimension " << i << ", operand1 has dimension size " << ubound1
         << " isDynamic=" << ddim1 << " vs operand2 has dimension size "
         << ubound2 << " isDynamic=" << ddim2;
```
the test passes.

But should we remove the check? Note that the real value of t1.shape[0] may not be 1. Removing the check will return incorrect result as in the [gh issue](https://github.com/pytorch/xla/issues/4870).

vanbasten23:
As discussed, I'll only compare upper bound for dynamic dimensions for now.

vanbasten23:
Note to myself: a simple test to repro the "::index" failure:
```
  def test_index_types(self):

    def test_fn(*indices):
      x = torch.arange(10).to(indices[0].device)
      return [x[idx] for idx in indices]

    # self.runAtenTest([
    #     torch.randint(0, 1, size=(10,), dtype=dtype)
    #     for dtype in (torch.long, torch.int32, torch.bool)

    device = xm.xla_device()
    idx_tensor = torch.randint(0, 1, size=(10,), dtype=torch.bool)
    xla_tensor = idx_tensor.to(device).detach().requires_grad_(idx_tensor.requires_grad)
    x = torch.arange(10).to(device)
    res = x[xla_tensor]
    print('res=', res)

root@t1v-n-15d4adcb-w-0:/workspaces/work# XLA_EXPERIMENTAL="nonzero:masked_select"   python3 pytorch/xla/test/test_operations.py TestAtenXlaTensor.test_index_types 2>&1 | tee out.txt
```

torch_xla/csrc/helpers.cpp
```diff
@@ -481,11 +481,65 @@ xla::Shape XlaHelpers::GetPromotedShape(const xla::Shape& shape1,
...
+    dyn_dims.insert(dyn_dims.end(), dyn_dims2.begin(),
+                    dyn_dims2.begin() + (dyn_dims2.size() - dyn_dims1.size()));
+  }
+  size_t min_size = std::min(upper_bounds1.size(), upper_bounds2.size());
+  for (const auto i : c10::irange(min_size)) {
+    int64_t ubound1 = upper_bounds1[upper_bounds1.size() - min_size + i];
+    int64_t ubound2 = upper_bounds2[upper_bounds2.size() - min_size + i];
+    bool ddim1 = dyn_dims1[dyn_dims1.size() - min_size + i];
+    bool ddim2 = dyn_dims2[dyn_dims2.size() - min_size + i];
+    XLA_CHECK((!ddim1 && !ddim2) || (ddim1 && !ddim2 && ubound2 == 1) ||
```
JackCaoG, line 526: can you add a comment to explain this check here?

torch_xla/csrc/helpers.cpp
```diff
@@ -481,11 +481,65 @@ xla::Shape XlaHelpers::GetPromotedShape(const xla::Shape& shape1,
...
+        upper_bounds.end(), upper_bounds2.begin(),
+        upper_bounds2.begin() + (upper_bounds2.size() - upper_bounds1.size()));
+    dyn_dims.insert(dyn_dims.end(), dyn_dims2.begin(),
+                    dyn_dims2.begin() + (dyn_dims2.size() - dyn_dims1.size()));
+  }
+  size_t min_size = std::min(upper_bounds1.size(), upper_bounds2.size());
+  for (const auto i : c10::irange(min_size)) {
+    int64_t ubound1 = upper_bounds1[upper_bounds1.size() - min_size + i];
+    int64_t ubound2 = upper_bounds2[upper_bounds2.size() - min_size + i];
+    bool ddim1 = dyn_dims1[dyn_dims1.size() - min_size + i];
```
JackCaoG, line 524: nit, maybe `dim1_is_dynamic` instead of `ddim1 `? Took me a while to figure out what it meant.

JackCaoG (commented):
Mostly lgtm, minor nits

torch_xla/csrc/helpers.cpp
vanbasten23: done

torch_xla/csrc/helpers.cpp
vanbasten23: done

JackCaoG (approved):
Thanks @vanbasten23 !

PR #4869, opened 2023-04-11
Add support to build PyTorch/XLA with PyTorch installed from pip.

kevint324:
Add a new env var BUILD_CPP_TESTS_USE_SITE_PYTORCH to fix the build of cpp test with PyTorch installed from pip.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,11 +74,24 @@ To build from source:
   sudo apt-get update
   ```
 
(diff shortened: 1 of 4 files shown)


kevint324:
I'm not sure about why building pytorch is necessary before build torch/xla. Maybe some historical reason?

I've tested this fix on Ubuntu 20 with gcc 9.4 and clang-8.

This should make developers' life a little bit easier by avoid building PyTorch and mess with the CXX11_ABI stuff.

kevint324:
trigger CircleCI

kevint324:
close

PR #4872, opened 2023-04-12
Apply input sharding to all compatible input tensors

jonb377:
The existing implementation expects a single (input, label) pair, but in practice we see multiple such pairs or, in the case of GPT-2, input_ids and attention_mask. This change will support these additional use cases.

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -10,6 +10,7 @@ import torch
 import torch.nn.functional as F
 import torch_xla
 from torch_xla.experimental import pjrt
+import torch_xla.experimental.xla_sharding as xs
 import torch_xla.core.xla_env_vars as xenv
 import torch_xla.debug.metrics_saver as ms
 import torch_xla.utils.utils as xu
@@ -1080,15 +1081,17 @@ def _maybe_convert_to_cpu(data, convert=True):
   return ToXlaTensorArena(convert_fn, select_fn).transform(data)
 
 
-def send_cpu_data_to_device(data, device, input_sharding=None):
+def send_cpu_data_to_device(data,
+                            device,
+                            input_sharding: xs.ShardingSpec = None):
 
   def convert_fn(tensors):
     devices = [str(device)] * len(tensors)
     xtensors = torch_xla._XLAC._xla_tensors_from_aten(tensors, devices)
     if input_sharding:
-      assert(len(xtensors) == 2), \
-        f"Expected input data and label pair, but got {xtensors}."
-      input_sharding.apply(xtensors[0])
+      for xtensor in xtensors:
+        if input_sharding.can_apply(xtensor):
+          input_sharding.apply(xtensor)
     return xtensors
 
   def select_fn(v):
--- a/torch_xla/experimental/xla_sharding.py
+++ b/torch_xla/experimental/xla_sharding.py
@@ -148,6 +148,12 @@ class ShardingSpec:
   mesh: Mesh
   partition_spec: Tuple[Union[int, None]]
 
+  def can_apply(self, t: torch.Tensor) -> bool:
+    """
+    Test whether the ShardingSpec is compatible with the given torch.Tensor.
+    """
+    return len(self.partition_spec) == len(t.shape)
+
   def apply(self, t: torch.Tensor):
     # TODO(yeounoh) use virtual device interface when available.
     assert (t.device == xm.xla_device())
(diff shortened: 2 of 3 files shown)


steventk-g:
Could you add a unit test for the sharding selector logic?

steventk-g:
What is a real world example of using the sharding selector for sharding generic input formats? When would a user want to apply sharding based on the index?

jonb377:
> Could you add a unit test for the sharding selector logic?

Sure! I couldn't find a good home for a unit test, do you know where it should go?

> What is a real world example of using the sharding selector for sharding generic input formats? When would a user want to apply sharding based on the index?

It's an edge case, but if the input/label pairs have the same shape, a selector would be necessary. Since this edge case is the only reason for having the selector, I can just remove that logic if that's preferable.

steventk-g:
> Sure! I couldn't find a good home for a unit test, do you know where it should go?
 
Thanks! I think this file makes sense https://github.com/pytorch/xla/blob/master/test/spmd/test_xla_sharding.py#L16 

> It's an edge case, but if the input/label pairs have the same shape, a selector would be necessary. Since this edge case is the only reason for having the selector, I can just remove that logic if that's preferable.

Ah, that makes sense. I see why we'd want it as an optional param here. I wonder how common this edge case is, or if there's any way to conceal this complexity from the user. Could we inject the selector if we see that input/label pairs have the same shape?

jonb377:
Synced offline - there's not a real benefit to having the selector, so we'll apply sharding to all compatible tensors. I'll update the PR shortly.

torch_xla/core/xla_model.py
```diff
@@ -1080,15 +1081,17 @@ def select_fn(v):
...
+                            input_sharding: xs.ShardingSpec = None):
 
   def convert_fn(tensors):
     devices = [str(device)] * len(tensors)
     xtensors = torch_xla._XLAC._xla_tensors_from_aten(tensors, devices)
     if input_sharding:
-      assert(len(xtensors) == 2), \
-        f"Expected input data and label pair, but got {xtensors}."
-      input_sharding.apply(xtensors[0])
+      for xtensor in xtensors:
```
JackCaoG, line 1092: can we add a unit test for this? or a e2e case that uses paraallel loader with SPMD.

jonb377: We have an e2e case using parallel loader in [test_train_spmd_imagenet.py](https://github.com/pytorch/xla/blob/master/test/spmd/test_train_spmd_imagenet.py#L255-L258)

torch_xla/experimental/xla_sharding.py
```diff
@@ -148,6 +148,12 @@ class ShardingSpec:
   mesh: Mesh
   partition_spec: Tuple[Union[int, None]]
 
+  def can_apply(self, t: torch.Tensor) -> bool:
+    """
+    Test whether the ShardingSpec is compatible with the given torch.Tensor.
+    """
+    return len(self.mesh.mesh_shape) == len(t.shape)
```
yeounoh, line 155: @jonb377 this check is not correct, mesh rank is not constraint by the input rank (e.g., if mesh rank is 1, and we have a 2D input tensor, it will be linearly sharded in one direction). Use partition_spec instead, 
```
len(t.shape) >= len(partition_spec)
```

We will have more checks inside `mark_sharding` but those should not be used for compatibility, and should crash if not met.

yeounoh (changes requested):
Mostly LGTM, have one comment.

torch_xla/experimental/xla_sharding.py
jonb377: Done, thanks Yeounoh!

yeounoh (approved):
Thanks, LGTM

torch_xla/core/xla_model.py
jonb377: Yeah, this is the case with real imagenet data in the above script. We have an XL ML test which runs this configuration, and it's currently failing due to multiple (input, label) pairs since the existing implementation only accepts a single pair.

JackCaoG (approved):

alanwaketan:
This breaks mnist:
```
root@t1v-n-307ffe96-w-0:/workspaces/work/pytorch/xla# PJRT_DEVICE=TPU python test/test_train_mp_mnist.py --num_epochs=1
  File "test/test_train_mp_mnist.py", line 2, in <module>
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 14, in <module>
  File "/workspaces/work/pytorch/xla/torch_xla/core/xla_model.py", line 11, in <module>
    import torch_xla.experimental.xla_sharding as xs
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/xla_sharding.py", line 10, in <module>
    from torch_xla.experimental.pjrt import requires_pjrt
ImportError: cannot import name 'requires_pjrt' from partially initialized module 'torch_xla.experimental.pjrt' (most likely due to a circular import) (/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py)
```

PR #4880, opened 2023-04-14
Add documentation for Terraform setup

mateuszlewko:


docker/experimental/Terraform.md
```diff
@@ -0,0 +1,53 @@
...
+If not, install Terraform from the [official source](https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli).
+
+## First time initialization
+
+1. Run `gcloud auth application-default login` on your local workstation.
+2. Go to the directory of the desired Terraform setup, for example 
+   [`tpu-pytorch-releases`](./tpu-pytorch-releases).
+3. Run `terraform init`.
+
+## Enforce entire Terraform setup
```
will-cromar, line 17: I'm still new to Terraform, so I want to understand better what a safe workflow looks like. We only want people to actually run `apply` after the changes are merged to `master`, right? Otherwise two people working on Terraform may overwrite each others' changes. Should the team even be running `terraform apply` themselves, or is that something we should build into our CI/CD pipeline?

will-cromar (approved):
This is the best documentation we've ever had for managing releases and triggers. Thank you!

docker/experimental/Terraform.md
mateuszlewko: > Should the team even be running terraform apply themselves, or is that something we should build into our CI/CD pipeline?

Yeah we should, we could have a cloud build trigger that runs terraform apply.

PR #4882, opened 2023-04-14
Update for SymInt API changes

ezyang:


torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1242,8 +1242,10 @@ XLATensorPtr full_symint(at::SymIntArrayRef sym_size,
                          const torch::lazy::BackendDevice& device,
                          at::ScalarType scalar_type) {
   XLA_CHECK(std::all_of(sym_size.begin(), sym_size.end(), [](at::SymInt dim) {
-    if (!dim.is_symbolic()) {
-      return dim >= 0;
+    // TODO: It should be OK to perform this test on symbolic ints too, not
```
vanbasten23, line 1245: One reason I can think of is that if we do something like `symint>0` then PyTorch/XLA will trigger a compilation and execution which in turn cause a graph break.

vanbasten23:
Thanks Ed for making the change.

I was looking at the upstream pr https://github.com/pytorch/pytorch/pull/98968 but I didn't see the API change somehow. But from this PR, it seems:
1. Upstream replaced `SymInt.is_symbolic()` with `SymInt.maybe_as_int()`.
2. Upstream removed `SymInt.toSymNodeImpl()` and we should use `SymInt.toSymNodeImplUnowned()`.

Is the above summary accurate?

torch_xla/csrc/torch_util.cpp
```diff
@@ -9,22 +9,22 @@
 namespace torch_xla {
 
 void SymIntElements::AddSymIntNodeElements(c10::SymInt& size) {
-  if (size.is_symbolic()) {
+  if (auto s = size.maybe_as_int()) {
```
vanbasten23, line 12: nit: do you need `.has_value()`?

ezyang:
@vanbasten23 Oops I pinned the wrong PR; it's actually https://github.com/pytorch/pytorch/pull/99157

torch_xla/csrc/torch_util.cpp
ezyang: optional when used in a bool-y context will be true if it is full (even if its contents is zero) and false if empty. This is important since I can't do binding AND also test `has_value` for the conditional lol.

torch_xla/csrc/tensor_methods.cpp
ezyang: But even for a dynamically sized tensor, the sizes should be guaranteed to be >= 0.

vanbasten23 (approved):

PR #4899, opened 2023-04-18
Expose map_location in `torch_xla.utils.serialization.load`

carmocca:


JackCaoG:
Thanks for the contribution, what's the motivation behind this pr? We usually recommend user to load and save tensors on CPU device.

carmocca:
And I agree with the recommendation! In Lightning we expose the `map_location` argument for those who might need it, and if this `save` function is used, then the argument becomes dead because we cannot plug it in.

Since this was an easy patch, I went ahead and added the necessary plumbing

carmocca:
I added the recommendation to the docstring

JackCaoG:
@Liyang90 Can you take a look since it is lighting related?

Liyang90 (commented):
LGTM

JackCaoG:
Can you fix the linter? according to yapf
```
Installing collected packages: yapf
Successfully installed yapf-0.30.0
diff --git a/torch_xla/utils/serialization.py b/torch_xla/utils/serialization.py
index 6e2ec733..74dddac3 100644
--- a/torch_xla/utils/serialization.py
+++ b/torch_xla/utils/serialization.py
@@ -96,9 +96,7 @@ def load(path, map_location=None):
       rewritten_tensors.append(
           torch.load(
               _get_tensor_file(tensor_folder, t.tid),
-              map_location=map_location
-          )
-      )
+              map_location=map_location))
     return rewritten_tensors
 
   def select_fn(v):
```

torch_xla/utils/serialization.py
```diff
@@ -77,11 +77,13 @@ def save(data, path, master_only=True, global_master=False):
   xm.rendezvous('torch_xla.utils.serialization.save')
 
 
-def load(path):
+def load(path, map_location=None):
```
vanbasten23, line 80: per your comment, how do you feel about defaulting the `map_location` to CPU?

carmocca: It would be a breaking change and it would be different to the default behavior of `torch.load`. I wouldn't suggest it

PR #4901, opened 2023-04-19
[Draft] Move partition outside of extract_compile_graph for dynamo CPU fallback

seanlatias:
In this PR, I modified the cpu fallback logic, where I move the capability partitioner outside of `extract_compile_graph`. I rename the original function into `extract_internal`, which will be called by the fused graphs (i.e., the graphs that can be lowered to XLA). In this code, I hardcoded the OP check function to ensure my test works. This is the test that I use:
```python
import torch
import torch_xla

    A = torch.cummin(M, 1)
    B = torch.add(A[0], mat1)
    return torch.add(B, mat2)
    
M = torch.randn(5, 10, device=xm.xla_device())
mat1 = torch.randn(5, 10, device=xm.xla_device())
mat2 = torch.randn(5, 10, device=xm.xla_device())  
      
res = fn_fallback(M, mat1, mat2, 0.5)
        
print(res)
```
The `torch.cummin` function cannot be lowered to XLA. With my code, the above code can work successfully without having any issue.

----
Graph after partitioning
```
def forward(self, L_M_ : torch.Tensor, L_mat1_ : torch.Tensor, L_mat2_ : torch.Tensor):
    l_m_ = L_M_
    l_mat1_ = L_mat1_
    l_mat2_ = L_mat2_
    cummin = torch.cummin(l_m_, 1);  l_m_ = None
    getitem = cummin[0];  cummin = None
    fused_0 = self.fused_0(getitem, l_mat1_, l_mat2_);  getitem = l_mat1_ = l_mat2_ = None
    return (fused_0,)
```
You can see that the two `torch.add` functions are fused.

cc @alanwaketan

### DIFF
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -205,21 +205,6 @@ class XlaOperatorSupport(torch.fx.passes.operator_support.OperatorSupport):
 def is_xla_tensor(tensor: torch.Tensor) -> bool:
   return tensor.device.type == "xla"
 
-"""
-def maybe_partition_graph(xla_model: torch.fx.GraphModule):
-  supported_ops = XlaOperatorSupport()
-  partitioner = CapabilityBasedPartitioner(xla_model, supported_ops)
-  partitions = partitioner.propose_partitions()
-  partitioned_graph = partitioner.fuse_partitions(partitions)
-
-  for node in partitioned_graph.graph.nodes:
-    if node.op == "call_module" and "fused_" in node.name:
-      fused_module = getattr(partitioned_graph, node.name)
-      fused_module._wrapped_call = extract_compiled_graph
-
-  return partitioned_graph
-"""
-
 def collect_inputs(model, *args):
   model.xla_args = args
   delattr(model, '_wrapped_call')
@@ -250,14 +235,10 @@ def extract_internal(xla_model: torch.fx.GraphModule):
       tensor_id: i for i, tensor_id in enumerate(args_tensor_ids)
   }
 
-  # partition the graph for unsupported ops
-  #xla_model = maybe_partition_graph(xla_model)
-
   # get_fallback_ops below uses counters to detect torch_xla fallbacks.
   # Clear the counters here so we ignore pre-existing fallbacks and
   # only detect fallbacks happening when running the xla_model below.
   metrics.clear_counters()
-  #delattr(xla_model, '_wrapped_call')
   xla_out = xla_model(*xla_args)
 
   fallback_ops = get_fallback_ops()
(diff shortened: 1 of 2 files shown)


seanlatias:
Another thing to notice is that with my implementation, the fused graph will be compiled each time because we use `wrapped_call` to trigger the compilation. Not sure if there is a way to check if the graph is already compiled and in the cache. If so, we can do the check first and compile the graph if necessary.

seanlatias:
Another alternative of implementation might be using similar appraoches as the nvfuse backend. Bascially we replace the fused graph with a compiled graph. The issue with this approach is that we need to get the inputs of each fused graph for compilation, which requires more work I believe.

torch_xla/core/dynamo_bridge.py
```diff
@@ -345,4 +349,18 @@ def optimized_mod(*args):
...
+def extract_compile_graph(xla_model, xla_args):
+  supported_ops = XlaOperatorSupport()
+  partitioner = CapabilityBasedPartitioner(xla_model, supported_ops)
+  partitions = partitioner.propose_partitions()
+  partitioned_graph = partitioner.fuse_partitions(partitions)
+
+  for node in partitioned_graph.graph.nodes:
+    if node.op == "call_module" and "fused_" in node.name:
+      fused_module = getattr(partitioned_graph, node.name)
+      fused_module._wrapped_call = extract_internal
```
JackCaoG, line 363: oh OK, I think this logic make more sense to me. We return multiple fused module and each of them will call `extract_internal` (maybe rename this function to  `extract_and_partition_graph ` and leave origional one `extract_compile_graph`). @wonjoolee95 I also left a comment in you origional pr, maybe you can also give this approach a shot since core logic is intact.

seanlatias: Renaming will require corresponding changes in dynamo's backend file. I will note this and make the changes once everything else is fixed.

seanlatias:
@JackCaoG I updated my method based on what nvfuser is doing (i.e., we replace the fused graph with our compiled fused graph). In this case, we will only need to compile once. However, I faced the following error when running my code
```
    if any(torch_xla._XLAC._check_tensor_need_materialization(args)):
RuntimeError: /home/ubuntu/pytorch/xla/torch_xla/csrc/init_python_bindings.cpp:824 : Check failed: false 
*** Begin stack trace ***     
...
*** End stack trace ***
_check_tensor_need_materialization currently does not handle XLATensor without XLAData and IR
```
Do you have any clue?

seanlatias:
@JackCaoG  If I remove the `_check_tensor_need_materialization` and simply call the `mark_step(wait=True)`, my code works properly. May I ask what's the reason that we add that check in the original dynamo_bridge?

wonjoolee95:
Thanks for the investigation, Sean. IIRC, the `_check_tensor_need_materialization` is used determine if an arg is updated in-place by the model. In fact, that exact error seems to be thrown in [our init_python_bindings.cpp code here](https://github.com/pytorch/xla/blob/master/torch_xla/csrc/init_python_bindings.cpp#L790), so we can probably just update this `else` statement of the code. 

cc @JackCaoG to confirm

seanlatias:
@wonjoolee95 I pushed a version where my test case should work. Now we just need to fix the part where we check whether an aten op can be supported or not.

seanlatias:
Moreover, this approach can also solve the unsupported operator+operand issue.

PR #4848, opened 2023-04-04
[WIP] Multinomial op lowering

cowanmeg:

PR #4904, opened 2023-04-19
Faster multinomial implementation

cowanmeg:
New implementation removes the while loop. 
Following up from: https://github.com/pytorch/xla/pull/4848

### DIFF
--- a/torch_xla/csrc/xla_lower_util.cpp
+++ b/torch_xla/csrc/xla_lower_util.cpp
@@ -1157,8 +1157,17 @@ xla::XlaOp BuildMultinomial(xla::XlaOp input, int64_t num_samples,
   // Map samples to categories
   std::vector<int64_t> broadcast_size = XlaHelpers::SizesOfXlaOp(input);
   broadcast_size.push_back(num_samples);
-  cumprob = xla::BroadcastInDim(cumprob, broadcast_size, {dim});
-  rng = xla::BroadcastInDim(rng, broadcast_size, {input_shape.rank()});
+  std::vector<int64_t> cumprob_broadcast_dim;
+  std::vector<int64_t> rng_broadcast_dim;
+  if (input_shape.rank() == 1) {
+    cumprob_broadcast_dim = {0};
+    rng_broadcast_dim = {1};
+  } else {
+    cumprob_broadcast_dim = {0, 1};
+    rng_broadcast_dim = {0, 2};
+  }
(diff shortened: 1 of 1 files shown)


wonjoolee95 (approved):
This is really cool. LGTM pending CI completion.

wonjoolee95:
One thing -- can we try to enable the `multinomial` test we disabled in the previous PR, @cowanmeg?

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -1154,48 +1154,18 @@ xla::XlaOp BuildMultinomial(xla::XlaOp input, int64_t num_samples,
...
+  cumprob = xla::BroadcastInDim(cumprob, broadcast_size, {dim});
+  rng = xla::BroadcastInDim(rng, broadcast_size, {input_shape.rank()});
 
-  std::vector<xla::XlaOp> results = ConsumeValue(xla::ForEachIndex(
-      sizes[dim], output_type, body_fn, {output, max_index, cumprob, rng},
-      "MultinomialMapping", output.builder()));
-  return results[0];
+  // Build comparison mask and sum along K dimension
+  auto output_type = xla::PrimitiveType::S64;
+  auto mask = BuildComparisonOp(at::aten::gt, rng, cumprob);
```
wonjoolee95, line 1165: This line seems to be responsible for the CI failure, related to shape differences between `rng` vs `cumprob`

cowanmeg: Yeah, the BroadcastInDim calls didn't have proper arguments for 2D inputs.

cowanmeg:
> One thing -- can we try to enable the `multinomial` test we disabled in the previous PR, @cowanmeg?

The failing test is: `test_multinomial_rng_state_advance`
Unfortunately, it still generates an error from long constant folding, so I'll keep it disabled.
I verified that the test passes when run isolated. 

```
 Constant folding an instruction is taking > 1s:

  reduce-window.70 (displaying the full instruction incurs a runtime overhead. Raise your logging level to 4 or above).

This isn't necessarily a bug; constant-folding is inherently a trade-off between compilation time and speed at runtime.  XLA has some guards that attempt to keep constant folding from taking too long, but fundamentally you'll always be able to come up with an input program that takes a long time.
```

wonjoolee95 (approved):
LGTM, thanks!

miladm:
Thanks @cowanmeg. Wondering if verified the perf latency gain of this op with/without the fix in python unit test script?

PR #4905, opened 2023-04-19
[Distributed] Remove sea-of-devices mode

alanwaketan:
Summary:
This change removes sea-of-devices mode which is no longer valid and simply the code in all_reduce a lot. This is the 1st step to adopt pytorch/pytorch#93173.

Test Plan:
PJRT_DEVICE=TPU python test/test_mp_sync_batch_norm.py
and CI...

### DIFF
--- a/test/test_torch_distributed_xla_backend.py
+++ b/test/test_torch_distributed_xla_backend.py
@@ -26,21 +26,6 @@ def new_group_barrier_disabled():
 
 
 @contextlib.contextmanager
-def always_intercore_reduce():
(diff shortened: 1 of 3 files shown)


JackCaoG (approved):
LGTM, not super confident about the host_ordinal part but CI should verify.

alanwaketan:
> LGTM, not super confident about the host_ordinal part but CI should verify.

Thanks, Jack. That part of code looks suspicious. But let's see how CI sees it.

PR #4912, opened 2023-04-20
[Distributed] Move the cached all reduce token to C++

alanwaketan:
Summary:
For all the cc ops, we use a token to introduce control dependencies among them such that they will be executed in order. This token is cached in the Python layer and this pull request moves it to C++ given the upcoming https://github.com/pytorch/pytorch/issues/93173 won't carry the token from Python to C++.

Test Plan:
CI.

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -452,8 +452,9 @@ def all_reduce(reduce_type, inputs, scale=1.0, groups=None, pin_layout=True):
     torch_xla._XLAC._set_all_reduce_token(result[1])
     results = [result[0]]
   else:
-    torch_xla._XLAC._set_all_reduce_token(torch_xla._XLAC._xla_all_reduce_inplace(
-        reduce_type, inputs, token, scale, groups, pin_layout))
+    torch_xla._XLAC._set_all_reduce_token(
+        torch_xla._XLAC._xla_all_reduce_inplace(reduce_type, inputs, token,
+                                                scale, groups, pin_layout))
     results = inputs
 
   return results[0] if isinstance(inputs, torch.Tensor) else results
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -272,7 +272,7 @@ def set_replication(device, devices):
   else:
     torch_xla._XLAC._xla_set_replication_devices([])
     devctx.device_index = 0
-  devctx.all_reduce_token = None
+  torch_xla._XLAC._set_all_reduce_token(None)
   torch_xla._XLAC._xla_set_default_device(device)
 
 
@@ -449,11 +449,11 @@ def all_reduce(reduce_type, inputs, scale=1.0, groups=None, pin_layout=True):
   if isinstance(inputs, torch.Tensor):
     result = torch_xla._XLAC._xla_all_reduce(reduce_type, inputs, token, scale,
                                              groups, pin_layout)
-    devctx.all_reduce_token = result[1]
+    torch_xla._XLAC._set_all_reduce_token(result[1])
     results = [result[0]]
   else:
-    devctx.all_reduce_token = torch_xla._XLAC._xla_all_reduce_inplace(
-        reduce_type, inputs, token, scale, groups, pin_layout)
+    torch_xla._XLAC._set_all_reduce_token(torch_xla._XLAC._xla_all_reduce_inplace(
+        reduce_type, inputs, token, scale, groups, pin_layout))
     results = inputs
 
   return results[0] if isinstance(inputs, torch.Tensor) else results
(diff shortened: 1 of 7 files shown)


alanwaketan:
Okay, GPU CI is happy without test_zero1.py. Let's skip that and I will follow up next week.

torch_xla/core/xla_model.py
```diff
@@ -548,12 +546,12 @@ def all_gather(value, dim=0, groups=None, output=None, pin_layout=True):
...
                                                     shard_count, groups or [],
                                                     pin_layout)
-    devctx.all_reduce_token = new_token
+    torch_xla._XLAC._set_all_reduce_token(new_token)
     return output
 
   result = torch_xla._XLAC._xla_all_gather(value, token, dim, shard_count,
                                            groups or [], pin_layout)
-  devctx.all_reduce_token = result[1]
+  torch_xla._XLAC._set_all_reduce_token(result[1])
```
JackCaoG, line 554: hmm, I guess in the long term if we don't even set the token in python, there is no need to return the token to the python layer? We can look into this later, maybe it is still cleaner this way.

torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -11,10 +11,18 @@
...
 
 namespace torch_xla {
 namespace {
 
+struct AllReduceToken {
+  std::shared_ptr<torch::lazy::Value> token;
+  std::mutex lock;
+};
+
+AllReduceToken g_all_reduce_token;
```
JackCaoG, line 24: Shouldn't we have one token per device? Under the PJRT for v3 cases, each process will have 2 thread and 1 device per thread. In that case those two thread should not share the same token.

torch_xla/core/xla_model.py
alanwaketan: Yea, I'm just too lazy to make the changes to all cc ops to remove the token in the python layer.

torch_xla/csrc/cross_replica_reduces.cpp
alanwaketan: Good point. Completely forgot the V3 case...

torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -259,4 +280,18 @@ ReduceScatterResult BuildReduceScatter(
...
+    const torch::lazy::BackendDevice& device) {
+  if (!g_all_reduce_token.token) {
+    std::lock_guard<std::mutex> lock(g_all_reduce_token.lock);
+    g_all_reduce_token.token = CreateToken(device);
+  }
+  return *g_all_reduce_token.token;
+}
+
+void SetAllReduceToken(const std::shared_ptr<torch::lazy::Value>& token) {
+  std::lock_guard<std::mutex> lock(g_all_reduce_token.lock);
```
JackCaoG, line 293: Hmm, why do we need lock here? I guess `mark_step` will set this token to None, but there shouldn't be any async process trying to access this token IR right? Are you just trying to be extra safe?

alanwaketan: Yea, just to be safe. I can remove it if you think it's not necessary...

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1616,6 +1601,15 @@ void InitXlaModuleBindings(py::module m) {
         [](at::Tensor& self, const at::Tensor& source) -> at::Tensor& {
           return XLANativeFunctions::set_(self, source);
         });
+  m.def("_get_all_reduce_token",
```
JackCaoG, line 1604: I think it is better to call it `_get_cc_token` or `_get_xla_token`, although it is currently only for `all_reduce`. We can also do this after we convert second cc op to use cpp token

alanwaketan: I'm just following the traditional in the python layer where it's named as all_reduce_token. haha.

JackCaoG (commented):
mostly lgtm, @pratnali @amithrm FYI we are moving the token to cpp so it can be traced by the dynamo.

torch_xla/csrc/cross_replica_reduces.cpp
JackCaoG: Don't want to introduce any noise here, but we can consider removing it later.

alanwaketan: Already removed.

torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -16,12 +16,9 @@
 namespace torch_xla {
 namespace {
 
-struct AllReduceToken {
-  std::shared_ptr<torch::lazy::Value> token;
-  std::mutex lock;
-};
-
-AllReduceToken g_all_reduce_token;
+// For V3, we have 4 processes and each process has 2 threads to manage the 8 cores.
```
JackCaoG: Nit `For PJRT + TPU v3`

JackCaoG (approved):

alanwaketan:

pratnali:
Thanks for the feedback everyone.

PR #4913, opened 2023-04-20
[Distributed] Hide token from Python for _xla_all_reduce path

alanwaketan:
Summary:
Now that we have the token cached in the C++ layer, and we really don't need to expose it to Python anymore. This is a first effort to to route one of the all_reduce to use the C++ cached token and remove the token from the pipes.  This is one of the efforts to integrate https://github.com/pytorch/pytorch/issues/93173.

Test Plan:

### DIFF
--- a/torch_xla/csrc/cross_replica_reduces.cpp
+++ b/torch_xla/csrc/cross_replica_reduces.cpp
@@ -293,8 +293,8 @@ void SetAllReduceToken(const torch::lazy::BackendDevice& device,
   g_all_reduce_tokens[device.ordinal()] = token;
 }
 
-void ResetAllReduceToken() {
-  g_token.reset();
+void SetAllReduceToken(const std::shared_ptr<torch::lazy::Value>& token) {
+  g_token = token;
 }
 
 }  // namespace torch_xla
--- a/torch_xla/csrc/cross_replica_reduces.cpp
+++ b/torch_xla/csrc/cross_replica_reduces.cpp
@@ -293,8 +293,4 @@ void SetAllReduceToken(const torch::lazy::BackendDevice& device,
   g_all_reduce_tokens[device.ordinal()] = token;
 }
 
-void SetAllReduceToken(const std::shared_ptr<torch::lazy::Value>& token) {
-  g_token = token;
-}
-
 }  // namespace torch_xla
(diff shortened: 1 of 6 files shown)


torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -11,10 +11,18 @@
...
 #include "torch_xla/csrc/token_handler.h"
+#include "torch_xla/csrc/xla_graph_executor.h"
 
 namespace torch_xla {
 namespace {
 
+struct AllReduceToken {
+  std::shared_ptr<torch::lazy::Value> token;
+  std::mutex lock;
+};
```
JackCaoG, line 22: lol I guess you need to manage the merge conflict with the other pr?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -205,18 +205,14 @@ std::shared_ptr<torch::lazy::Value> AllReduceInPlace(
...
-  std::tie(result, new_token) = tensor_methods::all_reduce(
-      bridge::GetXlaTensor(input), *token, GetReduceType(reduce_type), scale,
-      replica_groups, pin_layout);
-  return std::pair<at::Tensor, std::shared_ptr<torch::lazy::Value>>(
-      bridge::AtenFromXlaTensor(std::move(result)),
-      std::make_shared<torch::lazy::Value>(new_token));
+at::Tensor AllReduce(const std::string& reduce_type, const at::Tensor& input,
+                     double scale,
+                     const std::vector<std::vector<int64_t>>& replica_groups,
+                     bool pin_layout) {
```
JackCaoG, line 211: One thing I am worried about here is that now all_reduce is using a different token than other cc ops.

This means all of the `all_reduce` has a forced dependency and all other cc ops has a force dependency. All_reduce and other cc ops's order will not be enforced. I am OK with this being a short term thing but we do need to fix this before the release.

torch_xla/csrc/cross_replica_reduces.cpp
alanwaketan: Yea, no worries.

torch_xla/csrc/init_python_bindings.cpp
alanwaketan: Why? They are the same token from the C++ cache.

JackCaoG: oh actually nvm, all ops uses c++ token now, my bad.

JackCaoG (approved):

alanwaketan:

PR #4915, opened 2023-04-20
[Distributed] Adopts traceable all_reduce

alanwaketan:
Summary:
This pull request adopts the traceable all_reduce as mentioned in https://github.com/pytorch/pytorch/issues/93173.
1. It registers an implementation to c10d_functional.all_reduce via TORCH_LIBRARY_IMPL interface.
2. It then hooks xm.all_reduce to use the torch.ops.c10d_functional.all_reduce op, which will route to the above implementation.
3. Currently it only supports a very basic usage that assumes scale == 1.0 and groups == [] and pin_layout.

Test Plan:

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -450,10 +450,11 @@ def all_reduce(reduce_type, inputs, scale=1.0, groups=None, pin_layout=True):
     result = None
     if scale == 1.0 and groups == [] and pin_layout:
       # TODO(alanwaketan): Support groups.
-      result = torch.ops.c10d_functional.all_reduce(inputs, reduce_type, "", [], 0)
+      result = torch.ops.c10d_functional.all_reduce(inputs, reduce_type, "", [],
+                                                    0)
     else:
-      result = torch_xla._XLAC._xla_all_reduce(reduce_type, inputs, scale, groups,
-                                             pin_layout)
+      result = torch_xla._XLAC._xla_all_reduce(reduce_type, inputs, scale,
+                                               groups, pin_layout)
     results = [result]
   else:
     token, devctx = _get_all_reduce_token()
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -450,6 +450,7 @@ def all_reduce(reduce_type, inputs, scale=1.0, groups=None, pin_layout=True):
     result = None
     if scale == 1.0 and groups == [] and pin_layout:
       # TODO(alanwaketan): Support groups.
+      # Only c10d_functional version cc ops are traceable by Dynamo.
       result = torch.ops.c10d_functional.all_reduce(inputs, reduce_type, "", [],
                                                     0)
     else:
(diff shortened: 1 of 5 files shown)


torch_xla/core/xla_model.py
```diff
@@ -446,8 +447,14 @@ def all_reduce(reduce_type, inputs, scale=1.0, groups=None, pin_layout=True):
   """
   groups = groups or []
   if isinstance(inputs, torch.Tensor):
-    result = torch_xla._XLAC._xla_all_reduce(reduce_type, inputs, scale, groups,
-                                             pin_layout)
+    result = None
+    if scale == 1.0 and groups == [] and pin_layout:
+      # TODO(alanwaketan): Support groups.
+      result = torch.ops.c10d_functional.all_reduce(inputs, reduce_type, "", [],
```
wconstab, line 453: ranks=[] and world_size=0?  seems like you are not quite using the API as intended, just curious why

torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -103,6 +105,31 @@ std::shared_ptr<torch::lazy::Value> CreateToken(
...
+                      c10::string_view /*tag*/, at::IntArrayRef /*ranks*/,
+                      int64_t /*group_size*/) {
+  TORCH_LAZY_FN_COUNTER("xla::");
+  auto self_tensor = bridge::GetXlaTensor(self);
+  // TODO(alanwaketan): Use ranks and group_size to generate groups. Currently
+  // we just suse {} as a workaround. Scale is always 1.0 here, and we always
+  // pin layout.
+  auto result = tensor_methods::all_reduce(self_tensor, GetReduceType(reduceOp),
+                                           /*scale*/ 1.0,
+                                           /*groups*/ {}, /*pin_layout*/ true);
```
wconstab, line 125: do you need some way to translate between our `tag rank stride` format and xla's groups format?

torch_xla/core/xla_model.py
alanwaketan: Oh, the intention of this PR is to integrate the case where only one group in the whole world. Will follow up with the correct implementation of adopting the ranks and world_size later. In PyTorch/XLA, our runtime manages those information for us and not the users. Therefore, we usually don't need to pass down any ranks/world_size information down except for grouping.

torch_xla/csrc/cross_replica_reduces.cpp
alanwaketan: Yea, will follow up with that later.

torch_xla/core/xla_model.py
JackCaoG: Can we add a comment here explain that only c10d op is traceable for the dynamo?

JackCaoG (approved):

torch_xla/core/xla_model.py
alanwaketan: Sure, will do.

alanwaketan:

PR #4916, opened 2023-04-20
Migrate ci-tpu-test GCB trigger to use new Terraform setup.

mateuszlewko:
Use Ansible, Terraform and the Dockerfile in infra/ansible to build images used for e2e tests.
With this configuration we can get rid of files in docker/experimental as they were only used for e2e tests.


mateuszlewko:
Tests in progress: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/361a9ac8-3328-46ca-8248-5b1a849e14b6?jsmode=o&mods=allow_workbench_image_override&project=tpu-pytorch

will-cromar:
LGTM in principle. However, your test run failed

will-cromar:
We may need to extend the timeout when waiting for the pod. From the GCB logs:

```
Step #4 - "run_e2e_tests": ++ kubectl wait --for condition=ready --timeout=10m pod/xla-test-job-ngvp5 -o name
Step #4 - "run_e2e_tests": error: timed out waiting for the condition on pods/xla-test-job-ngvp5
Step #4 - "run_e2e_tests": + pod_name=
Finished Step #4 - "run_e2e_tests"
```

The pod did schedule and fail anyway:

```
$ kubectl logs xla-test-job-ngvp5
+ python3 pytorch/xla/test/test_operations.py -v
python3: can't open file 'pytorch/xla/test/test_operations.py': [Errno 2] No such file or directory
```

mateuszlewko:
Ah, I think it's because I'm not preserving xla sources in the release dockerfile (/infra/ansible/Dockerfile).

I'll move the PR to draft while I work on solution.

mateuszlewko:
Tests succeeded for the most recent version: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/7821f7f3-5daa-4073-8605-674cf40e11ea?project=tpu-pytorch. 
I'll rebase

infra/ansible/e2e_tests.Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+FROM python:${python_version}-${debian_version}
+
+WORKDIR /ansible
+RUN pip install ansible
+COPY . /ansible
+
+ARG ansible_vars
+
+# Build and install PyTorch and PyTorch/XLA wheels.
+RUN ansible-playbook -vvv playbook.yaml -e "stage=build" -e "${ansible_vars}"
```
will-cromar, line 13: It looks like the test image here includes all of the build artifacts, which blows the size up to `7.1 GB` compared to `1.4 GB` in the current test image. Is it possible to use the image build we use for nightly releases? The current `nightly_3.8_tpuvm` is only 3 GB (we can optimize the size further in a follow-up).

Large images increase the run time of the CI and contribute to flakiness as the disk on the test host machines fill up. Auto-repair is unfortunately not available for our GKE setup right now.

will-cromar:
Can you do a test run after rebasing on the bazel PR if you haven't already?

mateuszlewko:
> Can you do a test run after rebasing on the bazel PR if you haven't already?

The last run started on Started on Apr 26, 2023, 7:21:55 AM, so I think it included bazel PR

infra/ansible/e2e_tests.Dockerfile
mateuszlewko: I'll try to optimise it. Running a new build & test: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/133b06ba-ed73-48a7-9fdb-6393565aa99f?project=tpu-pytorch

will-cromar:
> > Can you do a test run after rebasing on the bazel PR if you haven't already?
> 
> The last run started on Started on Apr 26, 2023, 7:21:55 AM, so I think it included bazel PR.

Do the runs copy the source again from master, or do they use the commit they were triggered with? I believe you started that run before you rebased past the bazel build

mateuszlewko:
> > > Can you do a test run after rebasing on the bazel PR if you haven't already?
> > 
> > 
> > The last run started on Started on Apr 26, 2023, 7:21:55 AM, so I think it included bazel PR.
> 
> Do the runs copy the source again from master, or do they use the commit they were triggered with? I believe you started that run before you rebased past the bazel build

They actually fetch sources from a specific git revision (HEAD) in this case, which is done by ansible setup, see: https://github.com/pytorch/xla/blob/master/infra/ansible/roles/fetch_srcs/tasks/main.yaml#L17.

mateuszlewko:
I'm running the trigger again to check the image size: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/7717a315-2460-4404-abdc-549381d8b891?project=tpu-pytorch

will-cromar:
I see. The default should be to use the current working copy of the code. Our old build process used to work similarly, and debugging the build locally was difficult because you had to push your changes to github and then point the build at it.

In this specific case, there's an edge case where if you submit two PRs very close together, then the run against the first PR may pick up the changes from the second PR.

If you're using the ansible config at all, then that means that you've checked out pytorch/xla, and we should normally use that. I guess the exception would be if you wanted to build an old branch that predates ansible, but that case won't be relevant for much longer.

mateuszlewko:
> In this specific case, there's an edge case where if you submit two PRs very close together, then the run against the first PR may pick up the changes from the second PR.

I see, good point. We can fetch xla sources at $COMMIT_SHA, which is resolved by cloudbuild to git sha of the commit at which the build was triggered.

will-cromar:
`$COMMIT_SHA` makes sense here, and the updated image is only `2 GB`. I'll follow up and see if there's an easy way we can just copy the working directory into the Docker build instead of fetching XLA again. It's handy to be able to run the build locally with your working copy for debugging.

will-cromar (approved):
Thanks!

PR #4917, opened 2023-04-20
Skip test/test_zero1.py for TPU

alanwaketan:
Summary:
The test crashes on my TPU V4-8 run. The test is introduced in #4648.

Test Plan:
N.A.


wonjoolee95 (approved):
Seems like this test didn't run in our`tpu-ci` trigger, let me try to find out why. But LGTM, thanks!

JackCaoG:
Seems like reduce_scatter is not configure correctly on TPU, let's skip it now but maybe we should look into it later.

JackCaoG (approved):

alanwaketan:
Thanks Wonjoo and Jack.

PR #4921, opened 2023-04-21
Use dynamic shape support in TFRT TPU and CPU

will-cromar:
Removing a workaround for a bug in PJRT where it would crash or return incorrect results for dynamic buffers. The underlying issue is already fixed for TPU in our current TF pin, and the fix is present for CPU in https://github.com/tensorflow/tensorflow/commit/7f045484290f57ec3dc91877af77de39275a4f78.

For some reason, this workaround does not work for TFRT TPU specifically, but it does work for SE TPU, TFRT CPU, and SE GPU. Temporarily patching the CPU fix so I can remove the workaround altogether. After this CL, _all C++ tests pass on TFRT TPU_. 

```
[==========] 692 tests from 12 test suites ran. (10908461 ms total)
[  PASSED  ] 675 tests.
[  SKIPPED ] 17 tests, listed below:
```

### DIFF
--- a/tf_patches/cpu_dynamic_shape.diff
+++ b/tf_patches/cpu_dynamic_shape.diff
@@ -1,0 +1,61 @@
+# Required for dynamic shape on TFRT:CPU. Remove when we sync past https://github.com/tensorflow/tensorflow/commit/7f045484290f57ec3dc91877af77de39275a4f78
+diff --git a/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc b/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc
+index f23bbc13e8c..25f68ffab74 100644
+--- a/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc
++++ b/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc
+@@ -1075,18 +1075,25 @@ PjRtFuture<Status> TfrtCpuBuffer::ToLiteral(MutableLiteralBase* literal) {
+
+   bool should_sync_copy = device_buffer_wait_avs.empty() &&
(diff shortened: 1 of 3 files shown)


tf_patches/cpu_dynamic_shape.diff
```diff
@@ -0,0 +1,61 @@
+# Required for dynamic shape on TFRT:CPU. Remove when we sync past https://github.com/tensorflow/tensorflow/commit/7f045484290f57ec3dc91877af77de39275a4f78
+diff --git a/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc b/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc
+index f23bbc13e8c..25f68ffab74 100644
+--- a/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc
++++ b/tensorflow/compiler/xla/pjrt/tfrt_cpu_pjrt_client.cc
+@@ -1075,18 +1075,25 @@ PjRtFuture<Status> TfrtCpuBuffer::ToLiteral(MutableLiteralBase* literal) {
+
+   bool should_sync_copy = device_buffer_wait_avs.empty() &&
+                           literal->size_bytes() < kSmallDataTransferByteSize;
++  StatusOr<Shape> device_shape = logical_on_device_shape();
```
JackCaoG, line 10: Do we plan to merge this fix on upstream PJRT directly?

will-cromar: This is a copy of the linked commit from TensorFlow. I have it patched here temporarily until we update again.

JackCaoG: SG, please communicate with whoever is doing next pin update so they can delete this patch.

JackCaoG (approved):

PR #4922, opened 2023-04-21
[Distributed] Make xm.all_gather a single graph in Dynamo

alanwaketan:
Summary:
This pull request makes xm.all_gather, the _all_gather_using_all_reduce path, a single graph in Dynamo. To do that, it:
1. removes a hardware type check, specialize CPU doesn't seem to be worth it.
2. caches ordinal and xrt_world_size.

Test Plan:
PJRT_DEVICE=TPU python test/test_mp_all_gather.py

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -97,7 +97,7 @@ def xrt_world_size(defval=1):
   if pjrt.using_pjrt():
     g_xrt_world_size = pjrt.world_size()
   else:
-    g_xrt_world_size = xu.getenv_as(xenv.XRT_WORLD_SIZE, int, defval=defval)
+    g_xrt_world_size = xu.getenv_as(xenv.WORLD_SIZE, int, defval=defval)
   return g_xrt_world_size
 
 g_ordinal = None
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -78,7 +78,7 @@ def get_xla_supported_devices(devkind=None, max_devices=None):
     if kind_devices:
       return kind_devices[:max_devices] if max_devices else kind_devices
 
-
+g_xrt_world_size = None
 def xrt_world_size(defval=1):
   """Retrieves the number of devices which is taking part of the replication.
 
@@ -90,12 +90,17 @@ def xrt_world_size(defval=1):
   Returns:
     The number of devices which is taking part of the replication.
   """
+  global g_xrt_world_size
+  if g_xrt_world_size is not None:
+    return g_xrt_world_size
+
   if pjrt.using_pjrt():
-    return pjrt.world_size()
+    g_xrt_world_size = pjrt.world_size()
+  else:
+    g_xrt_world_size = xu.getenv_as(xenv.XRT_WORLD_SIZE, int, defval=defval)
+  return g_xrt_world_size
 
-  return xu.getenv_as(xenv.WORLD_SIZE, int, defval=defval)
-
-
+g_ordinal = None
 def get_ordinal(defval=0):
   """Retrieves the replication ordinal of the current thread.
 
@@ -109,10 +114,15 @@ def get_ordinal(defval=0):
   Returns:
     The replication ordinal of the current thread.
   """
-  if pjrt.using_pjrt():
-    return pjrt.global_ordinal()
+  global g_ordinal
+  if g_ordinal is not None:
+    return g_ordinal
 
-  return xu.getenv_as(xenv.ORDINAL, int, defval=defval)
+  if pjrt.using_pjrt():
+    g_ordinal = pjrt.global_ordinal()
+  else:
+    g_ordinal = xu.getenv_as(xenv.ORDINAL, int, defval=defval)
+  return g_ordinal
 
 
 def get_local_ordinal(defval=0):
@@ -533,8 +543,7 @@ def all_gather(value, dim=0, groups=None, output=None, pin_layout=True):
     A tensor which has, in the ``dim`` dimension, all the values from the
     participating replicas.
   """
-  if pin_layout and xla_device_hw(
-      value.device) in ('TPU', 'GPU', 'XPU') and output == None:
+  if pin_layout and output == None:
     # There is not an easy way to pin the all_gather layout on TPU and GPU, use
     # all_reduce based all_gather for this purpose.
     return _all_gather_using_all_reduce(
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -135,6 +135,7 @@ def get_ordinal(defval=0):
   Returns:
     The replication ordinal of the current thread.
   """
+  global _ORDINAL
   if _ORDINAL is not None:
     return _ORDINAL
 
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -30,6 +30,7 @@ _DEVICE_CONTEXTS_LOCK = threading.Lock()
 _WORLD_SIZE = None
 _ORDINAL = None
 
+
 def _init_world_size_ordinal():
   global _WORLD_SIZE, _ORDINAL
 
@@ -37,12 +38,13 @@ def _init_world_size_ordinal():
     return
 
   # We don't support V3-8. See Note [V3-8 Threading]
-  if tpu.version() <= 3:
+  if tpu.version() < 3:
     return
 
   if _WORLD_SIZE is None:
     _WORLD_SIZE = xrt_world_size()
     _ORDINAL = get_ordinal()
+
 
 class DeviceContext(object):
 
@@ -113,6 +115,7 @@ def xrt_world_size(defval=1):
 
   if pjrt.using_pjrt():
     return pjrt.world_size()
+
   return xu.getenv_as(xenv.WORLD_SIZE, int, defval=defval)
 
 
@@ -134,6 +137,7 @@ def get_ordinal(defval=0):
 
   if pjrt.using_pjrt():
     return pjrt.global_ordinal()
+
   return xu.getenv_as(xenv.ORDINAL, int, defval=defval)
 
 
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -9,6 +9,7 @@ import torch.distributed._functional_collectives
 import torch.nn.functional as F
 import torch_xla
 from torch_xla.experimental import pjrt
+from torch_xla.experimental import tpu
 import torch_xla.core.xla_env_vars as xenv
 import torch_xla.debug.metrics_saver as ms
 import torch_xla.utils.utils as xu
@@ -26,6 +27,22 @@ REDUCE_MAX = 'max'
 _DEVICE_CONTEXTS = dict()
 _DEVICE_CONTEXTS_LOCK = threading.Lock()
 
+_WORLD_SIZE = None
+_ORDINAL = None
+
+def _init_world_size_ordinal():
+  global _WORLD_SIZE, _ORDINAL
+
+  if not pjrt.using_pjrt():
+    return
+
+  # We don't support V3-8. See Note [V3-8 Threading]
+  if tpu.version() <= 3:
+    return
+
+  if _WORLD_SIZE is None:
+    _WORLD_SIZE = xrt_world_size()
+    _ORDINAL = get_ordinal()
 
 class DeviceContext(object):
 
@@ -79,9 +96,6 @@ def get_xla_supported_devices(devkind=None, max_devices=None):
       return kind_devices[:max_devices] if max_devices else kind_devices
 
 
-g_xrt_world_size = None
-
-
 def xrt_world_size(defval=1):
   """Retrieves the number of devices which is taking part of the replication.
 
@@ -93,20 +107,13 @@ def xrt_world_size(defval=1):
   Returns:
     The number of devices which is taking part of the replication.
   """
-  global g_xrt_world_size
-  if g_xrt_world_size is not None:
-    return g_xrt_world_size
+  global _WORLD_SIZE
+  if _WORLD_SIZE is not None:
+    return _WORLD_SIZE
 
   if pjrt.using_pjrt():
-    g_xrt_world_size = pjrt.world_size()
-  else:
-    g_xrt_world_size = xu.getenv_as(xenv.WORLD_SIZE, int, defval=defval)
-  return g_xrt_world_size
-
-
-# See Note [V3-8 Threading]
-g_ordinals = {}
-g_thread_id = None
+    return pjrt.world_size()
+  return xu.getenv_as(xenv.WORLD_SIZE, int, defval=defval)
 
 
 def get_ordinal(defval=0):
@@ -122,17 +129,12 @@ def get_ordinal(defval=0):
   Returns:
     The replication ordinal of the current thread.
   """
-  global g_ordinals
-  global g_thread_id
-  if g_thread_id is not None:
-    return g_ordinals[g_thread_id]
-  g_thread_id = threading.get_native_id()
+  if _ORDINAL is not None:
+    return _ORDINAL
 
   if pjrt.using_pjrt():
-    g_ordinals[g_thread_id] = pjrt.global_ordinal()
-  else:
-    g_ordinals[g_thread_id] = xu.getenv_as(xenv.ORDINAL, int, defval=defval)
-  return g_ordinals[g_thread_id]
+    return pjrt.global_ordinal()
+  return xu.getenv_as(xenv.ORDINAL, int, defval=defval)
 
 
 def get_local_ordinal(defval=0):
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -100,7 +100,9 @@ def xrt_world_size(defval=1):
     g_xrt_world_size = xu.getenv_as(xenv.WORLD_SIZE, int, defval=defval)
   return g_xrt_world_size
 
-g_ordinal = None
+# See Note [V3-8 Threading]
+g_ordinal = {}
+g_thread_id = None
 def get_ordinal(defval=0):
   """Retrieves the replication ordinal of the current thread.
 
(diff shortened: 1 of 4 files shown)


torch_xla/core/xla_model.py
```diff
@@ -78,7 +78,7 @@ def get_xla_supported_devices(devkind=None, max_devices=None):
     if kind_devices:
       return kind_devices[:max_devices] if max_devices else kind_devices
 
-
+g_xrt_world_size = None
 def xrt_world_size(defval=1):
```
alanwaketan, line 82: @wconstab This is the python function that I want to use in 'allow_in_graph'.

torch_xla/core/xla_model.py
```diff
@@ -90,12 +90,17 @@ def xrt_world_size(defval=1):
...
+    return g_xrt_world_size
 
+  if pjrt.using_pjrt():
+    g_xrt_world_size = pjrt.world_size()
+  else:
+    g_xrt_world_size = xu.getenv_as(xenv.XRT_WORLD_SIZE, int, defval=defval)
+  return g_xrt_world_size
 
+g_ordinal = None
 def get_ordinal(defval=0):
```
alanwaketan, line 104: @wconstab This is the python function that I want to use in 'allow_in_graph'.

torch_xla/core/xla_model.py
wconstab: hmm, if you are going to manually cache the value of this anyway, then i think just using allow_in_graph without the caching is the same thing.

the issue with allow_in_graph is if you expect the value to be updated on later iterations, allow_in_graph will prevent that from working.  But if you expect the value to be a constant for the whole execution, then allow_in_graph will capture the value during compile and reuse it later (e.g. cache it)

alanwaketan: Here is how I use allow_in_graph:
```
ptxla@t1v-n-307ffe96-w-0:/workspaces/work/pytorch/xla$ git diff
diff --git a/torch_xla/core/xla_model.py b/torch_xla/core/xla_model.py
index 6ff4a5a5..a07ff472 100755
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -6,6 +6,7 @@ import time
 from typing import List, Optional
 import torch
 import torch.distributed._functional_collectives
+from torch._dynamo import allow_in_graph
 import torch.nn.functional as F
 import torch_xla
@@ -1088,3 +1089,6 @@ def optimization_barrier_(tensors):
     tensors (List[torch.Tensor]): List of `torch.Tensor` to add barrier to.
   """
   torch_xla._XLAC._xla_optimization_barrier_(tensors)
+
+
+allow_in_graph(xrt_world_size)
```

And here is the error:
```
root@t1v-n-307ffe96-w-0:/workspaces/work/pytorch/xla# PJRT_DEVICE=TPU python test/test_mp_all_gather.py 
concurrent.futures.process._RemoteTraceback: 
"""
  File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 198, in _process_chunk
    return [fn(*args) for args in chunk]
  File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 198, in <listcomp>
    return [fn(*args) for args in chunk]
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 92, in wrapper
    return fn(*args, **kwargs)
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 245, in _run_thread_per_device
    replica_results = list(
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 619, in result_iterator
    yield fs.pop().result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 238, in _thread_fn
    return fn()
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 341, in __call__
    self.fn(global_ordinal(), *self.args, **self.kwargs)
  File "/workspaces/work/pytorch/xla/test/test_mp_all_gather.py", line 32, in _mp_fn
    result = compiled_all_gather(ordinal_tensor, dim=0)
  File "/workspaces/work/pytorch/torch/_dynamo/eval_frame.py", line 252, in _fn
    return fn(*args, **kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/eval_frame.py", line 405, in catch_errors
    return callback(frame, cache_size, hooks, frame_state)
  File "/workspaces/work/pytorch/torch/_dynamo/convert_frame.py", line 122, in _fn
    return fn(*args, **kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/convert_frame.py", line 331, in _convert_frame_assert
    return _compile(
  File "/workspaces/work/pytorch/torch/_dynamo/utils.py", line 169, in time_wrapper
    r = func(*args, **kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/convert_frame.py", line 401, in _compile
    out_code = transform_code_object(code, transform)
  File "/workspaces/work/pytorch/torch/_dynamo/bytecode_transformation.py", line 1000, in transform_code_object
    transformations(instructions, code_options)
  File "/workspaces/work/pytorch/torch/_dynamo/convert_frame.py", line 386, in transform
    tracer.run()
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 1972, in run
    super().run()
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 670, in run
    and self.step()
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 630, in step
    getattr(self, inst.opname)(inst)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 355, in wrapper
    return inner_fn(self, inst)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 1138, in CALL_FUNCTION_KW
    self.call_function(fn, args, kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 521, in call_function
    self.push(fn.call_function(self, args, kwargs))
  File "/workspaces/work/pytorch/torch/_dynamo/variables/functions.py", line 269, in call_function
    return super().call_function(tx, args, kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/variables/functions.py", line 102, in call_function
    return tx.inline_user_function_return(
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 557, in inline_user_function_return
    result = InliningInstructionTranslator.inline_call(self, fn, args, kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 2077, in inline_call
    return cls.inline_call_(parent, func, args, kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 2155, in inline_call_
    tracer.run()
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 670, in run
    and self.step()
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 630, in step
    getattr(self, inst.opname)(inst)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 355, in wrapper
    return inner_fn(self, inst)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 1138, in CALL_FUNCTION_KW
    self.call_function(fn, args, kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 521, in call_function
    self.push(fn.call_function(self, args, kwargs))
  File "/workspaces/work/pytorch/torch/_dynamo/variables/functions.py", line 269, in call_function
    return super().call_function(tx, args, kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/variables/functions.py", line 102, in call_function
    return tx.inline_user_function_return(
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 557, in inline_user_function_return
    result = InliningInstructionTranslator.inline_call(self, fn, args, kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 2077, in inline_call
    return cls.inline_call_(parent, func, args, kwargs)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 2155, in inline_call_
    tracer.run()
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 670, in run
    and self.step()
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 630, in step
    getattr(self, inst.opname)(inst)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 355, in wrapper
    return inner_fn(self, inst)
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 1086, in CALL_FUNCTION
    self.call_function(fn, args, {})
  File "/workspaces/work/pytorch/torch/_dynamo/symbolic_convert.py", line 521, in call_function
    self.push(fn.call_function(self, args, kwargs))
  File "/workspaces/work/pytorch/torch/_dynamo/variables/torch.py", line 603, in call_function
    tensor_variable = wrap_fx_proxy(
  File "/workspaces/work/pytorch/torch/_dynamo/variables/builder.py", line 923, in wrap_fx_proxy
    return wrap_fx_proxy_cls(
  File "/workspaces/work/pytorch/torch/_dynamo/variables/builder.py", line 1098, in wrap_fx_proxy_cls
    unimplemented(
  File "/workspaces/work/pytorch/torch/_dynamo/exc.py", line 107, in unimplemented
    raise Unsupported(msg)
torch._dynamo.exc.Unsupported: torch.* op returned non-Tensor int call_function <function xrt_world_size at 0x7fb184e94ca0>

from user code:
   File "/workspaces/work/pytorch/xla/torch_xla/core/xla_model.py", line 550, in all_gather
    return _all_gather_using_all_reduce(
  File "/workspaces/work/pytorch/xla/torch_xla/core/xla_model.py", line 511, in _all_gather_using_all_reduce
    left, right = ordinal, xrt_world_size() - 1 - ordinal

Set torch._dynamo.config.verbose=True or TORCHDYNAMO_VERBOSE=1 for more information


You can suppress this exception and fall back to eager by setting:
    torch._dynamo.config.suppress_errors = True

"""

The above exception was the direct cause of the following exception:

  File "test/test_mp_all_gather.py", line 66, in <module>
    xmp.spawn(_mp_fn, args=())
  File "/workspaces/work/pytorch/xla/torch_xla/distributed/xla_multiprocessing.py", line 367, in spawn
    return pjrt.spawn(fn, nprocs, start_method, args)
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 365, in spawn
    _run_multiprocess(spawn_fn, start_method=start_method)
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 92, in wrapper
    return fn(*args, **kwargs)
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 322, in _run_multiprocess
    replica_results = list(
  File "/workspaces/work/pytorch/xla/torch_xla/experimental/pjrt.py", line 323, in <genexpr>
  File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 484, in _chain_from_iterable_of_lists
    for element in iterable:
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 619, in result_iterator
    yield fs.pop().result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
torch._dynamo.exc.Unsupported: torch.* op returned non-Tensor int call_function <function xrt_world_size at 0x7fb184e94ca0>

from user code:
   File "/workspaces/work/pytorch/xla/torch_xla/core/xla_model.py", line 550, in all_gather
    return _all_gather_using_all_reduce(
  File "/workspaces/work/pytorch/xla/torch_xla/core/xla_model.py", line 511, in _all_gather_using_all_reduce
    left, right = ordinal, xrt_world_size() - 1 - ordinal

Set torch._dynamo.config.verbose=True or TORCHDYNAMO_VERBOSE=1 for more information


You can suppress this exception and fall back to eager by setting:
    torch._dynamo.config.suppress_errors = True

root@t1v-n-307ffe96-w-0:/workspaces/work/pytorch/xla# 
```

torch_xla/core/xla_model.py
```diff
@@ -90,12 +90,17 @@ def xrt_world_size(defval=1):
...
-  if pjrt.using_pjrt():
-    return pjrt.world_size()
-
-  return xu.getenv_as(xenv.WORLD_SIZE, int, defval=defval)
+  global g_xrt_world_size
+  if g_xrt_world_size is not None:
+    return g_xrt_world_size
 
+  if pjrt.using_pjrt():
+    g_xrt_world_size = pjrt.world_size()
```
JackCaoG, line 98: @will-cromar to double check, I believe `world_size` will never changed after the first init, but correct me if I am wrong.

will-cromar: That's correct. This call will force the PJRT client to initialize, but the devices and world size will not change after that.

torch_xla/core/xla_model.py
```diff
@@ -109,10 +114,15 @@ def get_ordinal(defval=0):
   Returns:
     The replication ordinal of the current thread.
   """
-  if pjrt.using_pjrt():
-    return pjrt.global_ordinal()
+  global g_ordinal
```
JackCaoG, line 117: I think this will break PJRT + v3 cases, the implementation we had checks the devices in
```
  m.def("_xla_get_default_device_ordinal", []() {
    std::string device_str = GetCurrentThreadDevice();
    torch::lazy::BackendDevice device =
        bridge::AtenDeviceToXlaDevice(device_str);
    return device.ordinal();
  });
```

alanwaketan: Hmm, this is confusing. That call is in the C++ layer. Then allow_in_graph won't work here.

But we can work around by caching a map...

JackCaoG: I am not sure actually, effectively this function won't return constant in the v3 cases because there are two devices per process. This is a bit tricky.

torch_xla/core/xla_model.py
alanwaketan: @will-cromar Do you know if there is a place in pjrt where we can call xrt_world_size and get_ordinal to cache the values?

torch_xla/core/xla_model.py
JackCaoG: I wonder if we can bypass the v3 cases for now, what's going to happen if you add a condition here to skip this cahce value of we are on v3 + PJRT?

alanwaketan: It will introduce graph breaks in Dynamo.

will-cromar: Should this use thread local storage instead?

alanwaketan: That's cool. Was not aware python has this feature. Let me work on it.

alanwaketan: Dynamo doesn't seem to compile in the same thread as the user code. threading.local doesn't work here.

torch_xla/core/xla_model.py
```diff
@@ -533,8 +553,7 @@ def all_gather(value, dim=0, groups=None, output=None, pin_layout=True):
     A tensor which has, in the ``dim`` dimension, all the values from the
     participating replicas.
   """
-  if pin_layout and xla_device_hw(
-      value.device) in ('TPU', 'GPU', 'XPU') and output == None:
```
JackCaoG, line 537: I think we had it because CPU was not supported at some point. Do you need to remove it because it will break dynamo?

alanwaketan: Yea.

torch_xla/core/xla_model.py
```diff
@@ -109,10 +122,17 @@ def get_ordinal(defval=0):
   Returns:
     The replication ordinal of the current thread.
   """
-  if pjrt.using_pjrt():
-    return pjrt.global_ordinal()
+  global g_ordinals
+  global g_thread_id
```
JackCaoG, line 126: I am not an expert on python threading, but wouldn't you want `g_thread_id` to be thread local? I think `global` means it is shared across devices? If my understanding is correct, we also need to make `g_ordinals` to be thread safe since it might be modified by multiple threads? @will-cromar to correct me if I am wrong.

alanwaketan: You are right, my bad. I might have found a place where we could safely cache these values.

JackCaoG: ok but then you mentioned dynamo compiles in a different threads... Would it work if we cache in the `_xla_get_default_device_ordinal` instead?

alanwaketan: Well, the thing is to prevent dynamo graph breaks. So we need to cache it in the topest layer. Otherwise, other things could introduce graph breaks as well.

alanwaketan: Actually I'm confused. If the threading is implemented in the python layer for V3-8, how do other global variables work?

JackCaoG: I think they are being shared across threads? Just like two threads will share the same pytorch random seed. A lot of variables is per device(like old device token) so they are OK.

alanwaketan: This threading support in V3 is very confusing. We might better to not support Dynamo in V3?

JackCaoG: yea, I think we don't need to support v3 in this dynamo case now, at least not on PJRT. If there is way to workaround it and not v3 for normal case we should do that.

torch_xla/core/xla_model.py
```diff
@@ -26,6 +27,22 @@
 _DEVICE_CONTEXTS = dict()
 _DEVICE_CONTEXTS_LOCK = threading.Lock()
 
+_WORLD_SIZE = None
```
JackCaoG, line 30: Please add a comment here to state that we cached these value for dynamo, or it might get removed one day when people don't know why we did it.

torch_xla/core/xla_model.py
```diff
@@ -26,6 +27,22 @@
...
+_ORDINAL = None
+
+def _init_world_size_ordinal():
+  global _WORLD_SIZE, _ORDINAL
+
+  if not pjrt.using_pjrt():
+    return
+
+  # We don't support V3-8. See Note [V3-8 Threading]
+  if tpu.version() <= 3:
```
JackCaoG: What would this return if we are on a GPU or CPU device.

torch_xla/core/xla_model.py
alanwaketan: At least they delete the test case as well. Otherwise, there won't be silent failures...

But agree, let me add a comment.

torch_xla/core/xla_model.py
alanwaketan: Added another guard.

JackCaoG (approved):
Thanks!

alanwaketan:
Thanks Jack for approving.

PR #4924, opened 2023-04-21
Fix hf transformer test to load correct dataset.

vanbasten23:


will-cromar (approved):
Good catch! Thanks!

test/pjrt/test_train_hf_transformer.py
```diff
@@ -54,7 +54,7 @@ def finetune(rank, train_dataset, test_dataset, tokenizer, flags):
       drop_last=True,
       generator=rng)
   test_loader = torch.utils.data.DataLoader(
-      train_dataset,
+      test_dataset,
```
JackCaoG, line 57: lol, nice catch!

PR #4927, opened 2023-04-21
[PJRT] Add `tpu.num_available_devices` to get number of XLA TPU devices

will-cromar:
This can be used to get the expected number of total TPU devices without initializing the client like we were able to do with `TPU_NUM_DEVICES` in XRT, without the user having to manually set it.


Liyang90 (commented):
LGTM

JackCaoG (approved):

PR #3792, opened 2022-07-28
Support dim=None for std

wonjoolee95:
Fixes https://github.com/pytorch/pytorch/pull/81845

PR #4935, opened 2023-04-24
[Dynamo] Refine CPU fallback for TD+XLA

seanlatias:
In this PR, we refined the CPU fallback mechanism originally implemented by @wonjoolee95 and made the following changes.

1. Fix the checking for XLA support.
  We use `torch.fx.Interpreter` to execute each node of the input module and use `torch_xla.debug.metrics` to check whether that node goes through CPU fallback or not. If it does, it means it is not supported by XLA.

2. Fix the fallback meachnism.
    2.1. Partition the graph according to our results from (1), which results in a new graph with subgraphs containing ops supported by XLA only.
    2.2. Compile each subgraph into a compiled function call.
    2.3. Replace the subgraph with the compiled function call.

With the above changes, we can even cover cases where a combination of operator and operand is not supported.

cc: @JackCaoG @alanwaketan

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -176,7 +176,7 @@ function run_xla_op_tests {
   run_test "$CDIR/dynamo/test_dynamo_integrations_util.py"
   run_test "$CDIR/dynamo/test_dynamo.py"
   run_test "$CDIR/dynamo/test_bridge.py"
-  run_test "$CDIR/dynamo/test_dynamo_fallback.py"
+  run_test "$CDIR/dynamo/test_fallback.py"
   run_test "$CDIR/dynamo/test_num_output.py"
   run_save_tensor_file "$CDIR/dynamo/test_dynamo_graph_dump.py"
   run_downcast_bf16 "$CDIR/test_data_type.py"


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -786,10 +786,7 @@ std::vector<bool> check_materialization_helper(
         need_materialization.push_back(true);
       }
     } else {
-      // TODO: maybe also handle it is a XLATensor with tensor_data case
-      XLA_CHECK(false)
-          << "_check_tensor_need_materialization "
-             "currently does not handle XLATensor without XLAData and IR";
+      need_materialization.push_back(true);
```
JackCaoG, line 789: I think the right thing to do is to check `xtensor->CurrentTensorData()`, if it is not nullptr then 
```
need_materialization.push_back(false);
```
since we don't need to execute a computation with a XLATensor with `tensor_data`(tensor data is a cpu at::Tensor that we know the value). We should leave the last `else` branch to throw an error to catch any future unhanlded case.

test/dynamo/test_dynamo_fallback.py
```diff
@@ -0,0 +1,43 @@
...
+import torch._dynamo.test_case as dt
+
+
+class XlaDynamoFallbackTest(dt.TestCase):
+
+  def test_operator_fallback(self):
+
+    @dynamo.optimize("torchxla_trace_once")
+    def fn_fallback(M, mat1, mat2):
+      A = torch.cummin(M, 1)
```
JackCaoG, line 14: is cumin suppose to fallback?

seanlatias: Yes, I have checked that cummin is not supported. I think one problem of this kind of testing is that once we support the unsupported ops, we might need to update the test. I'm thinking maybe I can create a custom op that will never be supported to avoid this problem.

test/dynamo/test_dynamo_fallback.py
```diff
@@ -0,0 +1,43 @@
...
+      B = torch.add(A[0], mat1)
+      return torch.add(B, mat2)
+
+    M = torch.randn(5, 10, device=xm.xla_device())
+    mat1 = torch.randn(5, 10, device=xm.xla_device())
+    mat2 = torch.randn(5, 10, device=xm.xla_device())
+
+    res = fn_fallback(M, mat1, mat2)
```
JackCaoG, line 22: I think we should
1. Check that `res` match the eager result.
2. Check the counter (try `met.short_metrics_report()` and you should see the fallback counter), I actually don't know what to expect but if there is a fallback I assume the `ExecuteTime` will be 2 instead of 1. You also need to run the `fn_fallback` once to let it compile and then do a clear counter than rerun to see `ExecuteTime` to be 2 instead of 1.

For more counter related stuff, check https://github.com/pytorch/xla/blob/master/torch_xla/debug/metrics.py

test/dynamo/test_dynamo_fallback.py
```diff
@@ -0,0 +1,43 @@
...
+      # xla currently only support alpha and beta == 1
+      A = torch.addmm(M, mat1, mat2, beta=beta)
+      B = torch.add(A, mat1)
+      return torch.add(B, mat1)
+
+    M = torch.randn(2, 3, device=xm.xla_device())
+    mat1 = torch.randn(2, 3, device=xm.xla_device())
+    mat2 = torch.randn(3, 3, device=xm.xla_device())
+
+    res = fn_fallback(M, mat1, mat2, 0.5)
```
JackCaoG, line 37: ditto

test/dynamo/test_dynamo_fallback.py
seanlatias: Will do.

test/dynamo/test_dynamo_fallback.py
wonjoolee95: For the scope of this PR, I think you can just leave a TODO comment. We can track that kind of specialized unit test in a separate issue, if that sounds okay.

test/dynamo/test_dynamo_fallback.py
wonjoolee95: If we don't see the mentioned PyTorch's counter, we can also create our own metric when we collect the fallback nodes -- could be useful for debugging/testing.

torch_xla/core/dynamo_bridge.py
```diff
@@ -319,3 +311,67 @@ def optimized_mod(*args):
...
+
+  for node in partitioned_graph.graph.nodes:
+    if node.op == "call_module" and "fused_" in node.name:
+      fused_module = getattr(partitioned_graph, node.name)
+      partitioned_graph.delete_submodule(node.target)
+      with partitioned_graph.graph.inserting_after(node):
+        new_node = partitioned_graph.graph.call_function(
+            extract_internal(fused_module), node.args, None)
+        node.replace_all_uses_with(new_node)
+      partitioned_graph.graph.erase_node(node)
```
wonjoolee95, line 373: Wondering why we need this `partitioned_graph.graph.erase_node(node)`? Can we do `graph.eliminate_dead_code()` like https://github.com/pytorch/pytorch/blob/0d66db1b2a9470a50d930308dbffda017500b80b/torch/_prims/nvfuser_executor.py#L465 or is this something different?

wonjoolee95 (commented):
Thanks, @seanlatias! Seems like some existing dynamo tests are failing, mostly due to metrics. The errors should be able to be reproduced locally:
```
python test/dynamo/test_dynamo.py DynamoInferenceBasicTest.test_resnet18
```

seanlatias:
> Thanks, @seanlatias! Seems like some existing dynamo tests are failing, mostly due to metrics. The errors should be able to be reproduced locally:
> 
> ```
> python test/dynamo/test_dynamo.py DynamoInferenceBasicTest.test_resnet18
> ```

Do you know why sometimes `mark_step` increases the compilation count?

JackCaoG:
`mark_step` will cut the current graph and compile/execute it. It is expected that `mark_step` will increase both `Compile` and `Execute` count. After the initial dynamo init, running the dynamo program should not trigger `mark_step`. You can check https://github.com/pytorch/xla/blob/master/test/dynamo/test_dynamo.py#L71 for a more detail example.

seanlatias:
Ok, because I see in the current dynamo bridge implementation, there are two places that increase the `Compile` count. One is the `mark_step` at the entry point (https://github.com/pytorch/xla/blob/8db3f891507424c91238cf5919c498e256398326/torch_xla/core/dynamo_bridge.py#L207) and the other one is when caching the graph (https://github.com/pytorch/xla/blob/8db3f891507424c91238cf5919c498e256398326/torch_xla/core/dynamo_bridge.py#L279). However, in the restnet19 test, it expects the `Compile` count to be just one (https://github.com/pytorch/xla/blob/8db3f891507424c91238cf5919c498e256398326/test/dynamo/test_dynamo.py#L82). And I think this is one of the errors I'm getting.

wonjoolee95:
The metric comparison failure seems like it has to do with how the `metrics.clear_counters()` was moved. Previously, we do a `metrics.clear_counters()` after `mark_step()` at https://github.com/pytorch/xla/blob/master/torch_xla/core/dynamo_bridge.py#L223. With this PR, we moved this `metrics.clear_counters()` to the `run_node()` function at https://github.com/pytorch/xla/pull/4935/files#diff-158d06d13623de8a2b4c4ee54902f148a8e0abda80870e5f25ee0bef3cc369b0R323. 

Let me also try to build this branch locally to confirm.

seanlatias:
Thanks Wonjoo. But the `clear_counters()` does not affect the `Compile` metric. Please let me know if you also see similar issues.

torch_xla/csrc/init_python_bindings.cpp
seanlatias: Done.

seanlatias:
I also found an interesting issue (coming from the optimizer). If we have an input graph like this
```
a = compute(x)
b = compute(y)
c = compute(z)
return (a, b, c)
```
Suppose all three compute can be offloaded to XLA, then after partitioner pass, we would get something like this
```
a = subgraph(x)
b = subgraph(y)
c = subgraph(z)
return (a, b, c)
```
Hower, ideally we want
```
abc = subgraph(x, y, z)
a = abc.getitem(0)
b = abc.getitem(1)
c = abc.getitem(2)
return (a, b, c)
```
I'm trying to see if there's a way we can optimize it.

seanlatias:
For the optimizer test, the number is way off because of an issue with the partitioner. I have created a PR (https://github.com/pytorch/pytorch/pull/100195) to solve it.

wonjoolee95:
> After second thought, I think it makes sense for my approach to introduce extra `Compile` and `Execute` because I indeed execute the graph one more time to collect fallback info. So if this approach makes sense to you guys, the assertions related to metric checking in other dynamo-related tests will need to be changed. I will first push a code with only metric checking failures and we can discuss from there.

Thanks for the investigation, Sean. Nice, that makes a lot sense to me. Let's update the `Compile` and `Execute` metrics in the tests. I also wonder what kind of implications this would have regarding performance since we're now adding another layer of execution on each initial trace, but we can look into that later. 

Also if you could add this finding to comments in our dynamo tests explaining this change in compile/execute time, that'd be great.

wonjoolee95:
> For the optimizer test, the number is way off because of an issue with the partitioner. I have created a PR ([pytorch/pytorch#100195](https://github.com/pytorch/pytorch/pull/100195)) to solve it.

Thanks for opening the PR, I've just triggered the CI run on it. LGTM but let's see if the reviewer has any feedback. Also, to make this PR build/test based of that PyTorch PR, we can add a PyTorch pin to this PR. If you add a file `.torch_pin` under `torch_patches/` with the PyTorch PR number, the CI on this PR will build PyTorch based off that PR. Example PR of adding such pin: https://github.com/pytorch/xla/pull/3792/commits/40f41fb98b0f2386d287eeac0bae86e873d4a9d8.

seanlatias:
> > For the optimizer test, the number is way off because of an issue with the partitioner. I have created a PR ([pytorch/pytorch#100195](https://github.com/pytorch/pytorch/pull/100195)) to solve it.
> 
> Thanks for opening the PR, I've just triggered the CI run on it. LGTM but let's see if the reviewer has any feedback. Also, to make this PR build/test based of that PyTorch PR, we can add a PyTorch pin to this PR. If you add a file `.torch_pin` under `torch_patches/` with the PyTorch PR number, the CI on this PR will build PyTorch based off that PR. Example PR of adding such pin: [40f41fb](https://github.com/pytorch/xla/commit/40f41fb98b0f2386d287eeac0bae86e873d4a9d8).

Thanks Wonjoo. Do you know how I can request review in PyTorch or who I should tag?

JackCaoG:
@seanlatias I pinged Sherlock offline, we can help you find a reviewer for that pr.

wonjoolee95:
Had a quick sync with @seanlatias offline. Oddly, it seems like some fallback works on master branch if we just remove the assertion in our dynamo bridge at https://github.com/pytorch/xla/blob/master/torch_xla/core/dynamo_bridge.py#L226-L230. 

Using a similar example that was in our original RFCs (https://github.com/pytorch/xla/issues/4742 and https://github.com/pytorch/pytorch/issues/93601), the code below runs without any errors on master:
```
import torch
import torch_xla

  ret = torch.addmm(M, mat1, mat2, beta=beta)
  ret2 = ret * 2;
  return ret2;

device = xm.xla_device()
M = torch.randn(2, 3, device=device)
mat1 = torch.randn(2, 3, device=device)
mat2 = torch.randn(3, 3, device=device)

res1 = fn_fallback(M, mat1, mat2, 0.5)
res2 = fn_fallback(M, mat1, mat2, 0.5)
print('res1:', res1)
print('res2:', res2)
print('----------')
print(met.metrics_report())
```
And with the metrics report I can see that the `aten` op did fallback to CPU:
```
Counter: aten::addmm
  Value: 1
```
Previously, this would fail with an error like
```
torch._dynamo.exc.BackendCompilerFailed: torchxla_trace_once raised AssertionError: compiler_fn did not return callable
```
cc @JackCaoG, am I missing something here? It appears the fallback that wasn't working before is now working on master. 

Thanks @seanlatias for the investigation, let me know if I'm missing any details here.

seanlatias:
Just to provide more details. I have also tried using the same fix (i.e., simply remove the check and the assertion) on Dynamo test suite. So far 3 huggingface models and 4 timm models that used to fail because of CPU fallback now work correctly (by using `--accuracy` flag). I also dump the generated HLO codes and they also look good to me. Just want to make sure it is safe if we remove this check because I see that the check was there from the very beginning when Dynamo+XLA integration is introduced (https://github.com/pytorch/pytorch/pull/87741).

test/dynamo/test_fallback.py
```diff
@@ -0,0 +1,78 @@
...
+    M = torch.randn(5, 10)
+    mat1 = torch.randn(5, 10)
+    mat2 = torch.randn(5, 10)
+    xla_M = M.to(xm.xla_device())
+    xla_mat1 = mat1.to(xm.xla_device())
+    xla_mat2 = mat2.to(xm.xla_device())
+
+    cpu_res = fn_fallback(M, mat1, mat2)
+    xla_res = dynamo_fn(xla_M, xla_mat1, xla_mat2)
+
```
JackCaoG, line 27: can you use something similar to https://github.com/pytorch/xla/blob/master/torch_xla/core/dynamo_bridge.py#L57 to check there is no fallback ops in this case?

If this works, the fallback should happen on the pytorch cpu side and does not trigger a fallback counter on pytorch/xla end.

We should also check that `xla::cummin` is not in the counter, in case we lower this op in the future.

test/dynamo/test_fallback.py
```diff
@@ -0,0 +1,78 @@
...
+    mat1 = torch.randn(2, 3)
+    mat2 = torch.randn(3, 3)
+    xla_M = M.to(xm.xla_device())
+    xla_mat1 = mat1.to(xm.xla_device())
+    xla_mat2 = mat2.to(xm.xla_device())
+
+    cpu_res = fn_fallback(M, mat1, mat2, 0.5)
+    xla_res = dynamo_fn(M, mat1, mat2, 0.5)
+
+    self.assertTrue(torch.allclose(cpu_res, xla_res.cpu()))
```
JackCaoG, line 49: ditto, we should add the counter check. Did we also handle the opernad shape invalid case in this pr? I thought we just handle ops that we don't lower today.

test/dynamo/test_fallback.py
```diff
@@ -0,0 +1,78 @@
...
```
JackCaoG, line 72: ditto

JackCaoG: for this one we should also check the metrics for `ExecuteTime`(you need to clear metrics first). We are expecting to see 2 executions since fallback happens in the middle.

torch_xla/core/dynamo_bridge.py
```diff
@@ -316,6 +309,83 @@ def optimized_mod(*args):
       print(f"optimized_mod takes {time.time() - enter_ts} seconds overall")
 
     none_remover.add_nones(result)
-    return result
+    if len(result) == 1:
+      return result[0]
```
JackCaoG, line 313: why is this needed?

test/dynamo/test_fallback.py
wonjoolee95: This PR now actually handles invalid shape/operand case because of how the updated `FallBackNodeCollector` works. The `FallBackNodeCollector` now executes each node of the input module so it can figure out which ops fall back (regardless of if it falls back because it's just not implemented in XLA or because specific shapes/operands are not supported).

wonjoolee95:
@seanlatias, I was hoping to push directly to PR/branch but seems like I can't since it's a local forked branch. Are you able to push directly to https://github.com/pytorch/xla/pull/5000? Just rebased this PR with latest and fixed some conflicts.

seanlatias:
> @seanlatias, I was hoping to push directly to PR/branch but seems like I can't since it's a local forked branch. Are you able to push directly to #5000? Just rebased this PR with latest and fixed some conflicts.

@wonjoolee95 yeah I can do that. But before doing so, I'm wondering if this approach is still valid? Because we probably only need to remove the fallback checking and things will just work.

seanlatias:
@wonjoolee95 maybe let me do this. We can create an env var that determines whether we should use explicit CPU fallback (my method), or go through the original flow with checking removed. The default one will be the original flow with checking removed. How does that sound to you? If you agree, I can add the logics and push it to your branch.

seanlatias:
@wonjoolee95 Sounds good. Will do.

PR #4936, opened 2023-04-24
Fix test_mp_save

JackCaoG:
In https://github.com/pytorch/xla/commit/4f7f3dcd4b82aaa9acfcb63057cebd80e989ebf7 we removed the `rendezvous` in `xm.save`, hence we need to manually do it in the test to guarantee the correctness.

### DIFF
--- a/test/test_mp_save.py
+++ b/test/test_mp_save.py
@@ -40,7 +40,7 @@ def _mp_fn(index, temp_file):
   dd = _create_state_dict(device)
   xm.save(dd, temp_file)
   # User needs to manually rendezvous since only master process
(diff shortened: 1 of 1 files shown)


test/test_mp_save.py
```diff
@@ -39,6 +39,10 @@ def _mp_fn(index, temp_file):
   device = xm.xla_device()
   dd = _create_state_dict(device)
   xm.save(dd, temp_file)
+  # User needs to manually rendezvous since only master process
+  # will perfomrn the save and other processes needs to wait.
```
will-cromar, line 43: nit: "perform"

will-cromar (approved):

PR #4941, opened 2023-04-25
Add a test to verify XLATensorImpl memoize the SizeNode it creates.

vanbasten23:
It turned out that PyTorch/XLA already supports it. This is because of `XLATensorImpl.generation_` at https://github.com/pytorch/xla/blob/74eff29120682d4764bf52419b9e65e5b6b16a4c/torch_xla/csrc/tensor_impl.cpp#L162


JackCaoG (approved):

vanbasten23:
Thanks for the review!

PR #4948, opened 2023-04-26
Pass `cxx_abi` Docker arg to setup.py

will-cromar:
See https://github.com/pytorch/xla/blob/2870ac2fd0d646a6ffc3a4182563267b4a4c534d/setup.py#L208-L211

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -29,8 +29,9 @@ RUN git clone https://github.com/pytorch/pytorch
 ENV USE_CUDA "0"
 ENV USE_MPI "0"
 
-# Value for -D_GLIBCXX_USE_CXX11_ABI flag; if empty, it is infered from `torch._C`.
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -29,6 +29,9 @@ RUN git clone https://github.com/pytorch/pytorch
 ENV USE_CUDA "0"
 ENV USE_MPI "0"
 
+# Value for -D_GLIBCXX_USE_CXX11_ABI flag; if empty, it is infered from `torch._C`.
+ENV CXX_ABI "${cxx_abi}"
```
JackCaoG, line 33: lol, didn't saw it is being deleted in last pr

will-cromar: I kicked off a test build with your trigger. Should we wait for it to finish before merging?

PR #4952, opened 2023-04-27
Correct coverage path

stgpetrovic:
Coverage bazel-out is i the repository root, and the script does `pushd test/cpp` so the file is not relatively there.


will-cromar (approved):
Good catch!

PR #4960, opened 2023-04-28
Parallel tests

stgpetrovic:
Make tests run in parallel for CPU; makes them much faster.


JackCaoG (approved):

JackCaoG:
@stgpetrovic This is breaking on master, I will revert it for now.

PR #4965, opened 2023-05-02
[PJRT] Cache world size and ordinal only when addressable_device_count == 1

will-cromar:
We currently check TPU version instead, which gives incorrect results when debugging multithreaded workloads on TPU v4 (e.g. `TPU_PROCESS_BOUNDS=1,1,1 TPU_CHIPS_PER_PROCESS_BOUNDS=2,2,1 TPU_VISIBLE_CHIPS=0,1,2,3`)

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -37,11 +37,8 @@ _ORDINAL = None
 def _init_world_size_ordinal():
   global _WORLD_SIZE, _ORDINAL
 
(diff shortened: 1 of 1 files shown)


alanwaketan (approved):
LGTM.

will-cromar:
This is only relevant for testing multithreading on TPU v4 to simulate TPU v2/v3. Real use cases would see a performance hit from using multithreading.

JackCaoG (approved):

PR #4969, opened 2023-05-02
Add instructions for creating non-master branch build trigger

mateuszlewko:
Also updated docs to reflect that Terraform setup is now automatically provisioned.


JackCaoG (approved):
The doc make sense to me, but I am sure when I actully tried it I will have more problems. Thanks @mateuszlewko !

mateuszlewko:
No worries, we can iterate on the docs and the setup until it's simple enough.

PR #4971, opened 2023-05-02
Wrap implicitly-replicated data in PjRtShardedData

jonb377:
This change wraps replicated tensors in `PjRtShardedData`, when the `XLA_USE_SPMD` environment variable is set. Doing so eliminates the overhead from implicitly replicating tensors each step, since the replicated handles are tracked within the `PjRtShardedData` wrapper. On Hugginface GPT-2, this decreases TPU idle time between steps from ~450ms to ~8ms

This takes effect in three places:
- Unsharded `CreateTensorsData`
- Sharded `CreateTensorsData` when sharding is null
- `TensorToXlaData` when the virtual device optimization is bypassed (as is the case for scalars/seeds)

Additionally, this change adds a static variable to test whether the virtual device is active to avoid querying the environment variable each time, and it adds the requirement that the environment variable is set to run in SPMD mode.

### DIFF
--- a/test/spmd/test_xla_virtual_device.py
+++ b/test/spmd/test_xla_virtual_device.py
@@ -72,17 +72,10 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
     xs.mark_sharding(xt1, self._get_mesh((1, self.n_devices)), partition_spec)
     outbound_with_virtual_device = met.metric_data("OutboundData")[1]
 
-    os.environ["XLA_USE_SPMD"] = "0"
-
-    met.clear_all()
-    xt2 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
-                       dtype=torch.float,
-                       device=xm.xla_device())
-    xs.mark_sharding(xt2, self._get_mesh((1, self.n_devices)), partition_spec)
-    outbound_without_virtual_device = met.metric_data("OutboundData")[1]
-
+    # Without virtual device optimization, we expect the data to be transferred to
+    # device at least twice, so assert that the actual transfer amount is less.
     self.assertLess(outbound_with_virtual_device,
-                    outbound_without_virtual_device)
+                    2 * xt1.nelement() * xt1.element_size())
 
 
 if __name__ == '__main__':
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -228,6 +228,11 @@ ComputationClient::DataPtr PjRtComputationClient::TransferShardsToServer(
     xla::Shape shape, xla::OpSharding sharding) {
   TF_VLOG(1) << "TransferShardsToServer with " << tensor_shards.size()
              << " shards.";
+  // TODO(jonbolin): Consider using CopyToDevice when sharding is REPLICATED.
+  // We are opting out of CopyToDevice for now due to the synchronization
+  // issues observed in ShardingUtil::InputHandler, but because CopyToDevice
+  // directly copies buffers between devices using ICI, it can be much faster
+  // than transferring from the host to each device.
   auto data_shards = TransferToServer(tensor_shards);
   std::vector<std::shared_ptr<PjRtData>> pjrt_data_shards;
   for (auto& shard : data_shards) {
@@ -265,6 +270,10 @@ ComputationClient::DataPtr PjRtComputationClient::ReplicateShardedData(
           dynamic_cast<PjRtShardedData*>(handle.get())) {
     TF_VLOG(1) << "ReplicateShardedData (handle=" << handle->GetOpaqueHandle()
                << ", shape=" << handle->shape() << ")";
+    if (sharded_data->GetSharding().type() == xla::OpSharding::REPLICATED) {
+      // Data is replicated, return the first shard
+      return sharded_data->shards[0];
+    }
     xla::XlaBuilder b("ReplicateShardedData");
     xla::Shape shape = sharded_data->shape();
     b.SetSharding(sharded_data->GetSharding());
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -43,6 +43,7 @@
 #include "torch_xla/csrc/tensor_util.h"
 #include "torch_xla/csrc/torch_util.h"
 #include "torch_xla/csrc/xla_graph_executor.h"
+#include "torch_xla/csrc/xla_sharding_util.h"
 
 // [Implementation Guidelines]
 // - If you want to call a at::func which doesn't have a kernel registered
@@ -474,7 +475,7 @@ at::Tensor XLANativeFunctions::_copy_from(const at::Tensor& self,
   if (!self_tensor) {
     static bool sync_update =
         xla::sys_util::GetEnvBool("XLA_TENSOR_UPDATE_SYNC", true) &&
-        !xla::sys_util::GetEnvBool("XLA_USE_SPMD", false);
+        !ShardingUtil::UseVirtualDevice();
     XLA_CHECK(dst_tensor);
     dst_tensor->UpdateFromTensor(self, /*sync=*/sync_update);
   } else if (!dst_tensor) {
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1474,6 +1474,8 @@ void InitXlaModuleBindings(py::module m) {
                                  const py::list& tile_assignment,
                                  bool replicated = false, bool manual = false) {
     TORCH_LAZY_COUNTER("XlaMarkSharding", 1);
+    XLA_CHECK(ShardingUtil::UseVirtualDevice())
+        << "Please set `XLA_USE_SPMD=1`";
     XLATensorPtr xtensor = bridge::GetXlaTensor(input);
     xla::OpSharding sharding =
         ShardingUtil::CreateOpSharding(tile_assignment, replicated, manual);
@@ -1483,8 +1485,7 @@ void InitXlaModuleBindings(py::module m) {
                       static_cast<XlaDeviceType>(xtensor->GetDevice().type())));
 
     at::Tensor cpu_tensor;
-    if (xla::sys_util::GetEnvBool("XLA_USE_SPMD", false) &&
-        xtensor->CurrentTensorData().has_value()) {
+    if (xtensor->CurrentTensorData().has_value()) {
       TORCH_LAZY_COUNTER("VirtualDeviceUsage", 1);
       // When virtual device is enabled for SPMD, we defer the initial data
       // transfer to the device and retain the original data on the host, until
--- a/torch_xla/csrc/tensor_util.cpp
+++ b/torch_xla/csrc/tensor_util.cpp
@@ -35,41 +35,6 @@ struct DataAsync {
   std::vector<xla::util::ExceptionCleanup> handle_unlockers;
 };
 
-void TransferToServerAsync(std::shared_ptr<DataAsync> async,
-                           const std::vector<std::string>& devices) {
-  TORCH_LAZY_TIMED("TransferToServerAsync");
-
-  std::vector<xla::ComputationClient::DataPtr> async_xla_datas =
-      xla::ComputationClient::Get()->CreateAsyncDatas(async->source_tensors);
-  async->handle_unlockers =
-      xla::ComputationClient::Get()->LockAsyncDatas(async_xla_datas);
-  async->async_datas = WrapXlaData(async_xla_datas);
-  auto mwait = std::make_shared<xla::util::MultiWait>(/*num_wait=*/1);
-  auto update_data = [async, async_xla_datas]() {
-    try {
-      xla::ComputationClient::Get()->TransferToServer(async->source_tensors,
-                                                      async_xla_datas);
-    } catch (...) {
-      // There are two paths of discovery of an exception happening on an
-      // asynchronous task. One happens if the creator of the asynchronous task
-      // explicitly waits for completion, in which case the exception will be
-      // thrown from the Wait() API. Re-throwing the exception below makes sure
-      // this will be captured by the completer function created below, and
-      // surfaced by the Wait() API. But we also need to surface the exception
-      // even in case the caller does not wait, and that is accomplished by
-      // setting the unlockers status. In that case the exception will be
-      // surfaced when the user tries to acquire the device locks the next time.
-      std::exception_ptr exptr = std::current_exception();
-      for (auto& unlocker : async->handle_unlockers) {
-        unlocker.SetStatus(exptr);
-      }
-      throw;
-    }
-  };
-  xla::env::ScheduleIoClosure(
-      xla::util::MultiWait::Completer(mwait, std::move(update_data)));
-}
-
 bool ShouldUseBF16() {
   bool use_bf16 = xla::sys_util::GetEnvBool("XLA_USE_BF16", false);
   if (use_bf16) {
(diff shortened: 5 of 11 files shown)


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -152,6 +152,12 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::GetDataShards(
       shards.push_back(std::make_shared<PjRtData>(
           shard->device(), shard->shape(), shard->buffer));
     }
+  } else if (PjRtReplicatedData* replicated_data =
```
jonb377, line 155: This is piggybacking on `GetDataShards`, so it might be better to move this to a new ComputationClient function. I'm happy to make that change

JackCaoG:
Thanks, I will try to take a look later today.

will-cromar (approved):

third_party/xla_client/pjrt_computation_client.h
```diff
@@ -174,6 +179,41 @@ class PjRtComputationClient : public ComputationClient {
     std::shared_ptr<PjRtBuffer> buffer;
   };
 
+  struct PjRtReplicatedData : public Data {
```
alanwaketan, line 182: I'm confused. Why would this wrapper help given it's just barely a vector of PjRtData?

jonb377: In the existing implementation, we only track the DataPtr for a single device's copy of the replicated data across steps, so after each step we discard the DataPtr for the replicated data on all other devices. This makes us `CopyToDevice` to all other devices every step. Additionally, CopyToDevice will block if the devices are busy (i.e. executing the computation), which causes the additional overhead.

PjRtReplicatedData just provides a Data class that tracks all of the replicated handles, but the benefit is that we can track all devices' DataPtr across steps. This lets us only CopyToDevice once in the `TransferReplicatedDataToServer` call and update the replicated data values by `Assign` after the computation when the output is replicated.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Looking at this dynamic cast & check, I am thinking maybe we should use the PjRtShardedData to hold the replicated data to track. PjRtShardedData can have a flag `replicated` or we can supply OpSharding::Replicated to it (it already holds the sharding spec) for lookup. This way, we don't need to extra auxiliary APIs, like `WrapReplicatedData`, and just re-use the existing similar APIs.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -223,6 +241,24 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
   return datas;
 }
 
+ComputationClient::DataPtr PjRtComputationClient::TransferReplicatedDataToServer(
```
yeounoh, line 244: `TransferShardsToServer` takes in `OpSharding`. Can this be done using `TransferShardsToServer` if `OpSharding` sharding type is replicated?

yeounoh (changes requested):
Left some comments, let's investigate if we can make it more lean.

third_party/xla_client/pjrt_computation_client.cc
jonb377: Sounds good, I'll try reusing PjRtShardedData with `OpSharding::Replicated`

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1136,7 +1142,9 @@ XLAGraphExecutor::CompilationResult XLAGraphExecutor::Compile(
     lowering_ctx.AddResult(root);
   }
   // Annotate HLO sharding selectively in the compuation.
-  bool is_sharded = ShardingUtil::SetHloSharding(&lowering_ctx);
+  // TODO(jonbolin): Always treat the computation as sharded when using SPMD.
+  bool is_sharded = ShardingUtil::SetHloSharding(&lowering_ctx) ||
+                    xla::sys_util::GetEnvBool("XLA_USE_SPMD", false);
```
yeounoh, line 1147: For our offline discussion, let's not rely on `XLA_USE_SPMD` -- it's meant to be replaced & for virtual device optimization. For implicit replication, we shouldn't rely on this flag -- instead let's see if we can have `InputHandler` decide.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -882,6 +882,12 @@ void XLAGraphExecutor::ExtractIRAndPrepareXlaData_(
       handle = WrapXlaData(xla::ComputationClient::Get()->WrapDataShards(
           {UnwrapXlaData(handle)}, GetVirtualDevice().toString(),
           sharding->shape.value(), sharding->sharding));
+    } else if (xla::sys_util::GetEnvBool("XLA_USE_SPMD", false)) {
+      // Create a placeholder PjRtReplicatedData when using SPMD
```
JackCaoG, line 886: hmm, so the assumption is that if we are under `SPMD` mode, all of the tensor that does not have the sharding annotation must be replicated. Is that the right reading?

JackCaoG: Can you remind me how virtual device is impacted.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -262,6 +262,16 @@ std::vector<xla::ComputationClient::DataPtr> ShardingUtil::OutputHandler(
             shards, GetVirtualDevice().toString(), sharding->shape.value(),
             sharding->sharding));
       }
+    } else if (xla::sys_util::GetEnvBool("XLA_USE_SPMD", false)) {
```
JackCaoG, line 265: If we are going to relie on a env var, make it a staic variable and store it in a helper function. Search for `XLA_USE_BF16` for example. We should not keep quering for the env var.

torch_xla/csrc/xla_graph_executor.cpp
jonb377: That's the correct reading: under SPMD mode, every tensor that's not explicitly sharded should will be replicated. cc: @yeounoh in case I'm overlooking something.

Virtual device will defer data transfer until CreateTensorsData, which in the `mark_sharding` path will be called with a ShardingSpec to distribute the tensor across local devices. If we don't call `mark_sharding`, then the data should instead be replicated. Currently, this happens in InputHandler through `CopyToDevice`, but this change moves the replication step into CreateTensorsData.

torch_xla/csrc/xla_graph_executor.cpp
jonb377: Following up on this - we aren't able to use InputHandler to track the handles across steps, since we only have access to the DataPtrs at that stage. We decided to rely on XLA_USE_SPMD in CreateTensorsData for now.

test/spmd/test_xla_virtual_device.py
```diff
@@ -72,17 +72,9 @@ def test_outbound_data_metrics(self):
...
-                       dtype=torch.float,
-                       device=xm.xla_device())
-    xs.mark_sharding(xt2, self._get_mesh((1, self.n_devices)), partition_spec)
-    outbound_without_virtual_device = met.metric_data("OutboundData")[1]
-
-    self.assertLess(outbound_with_virtual_device,
-                    outbound_without_virtual_device)
+    # Without virtual device optimization, we expect the data to be transferred to
+    # device at least twice, so assert that the actual transfer amount is less.
+    self.assertLess(outbound_with_virtual_device, 2 * xt1.nelement() * xt1.element_size())
```
jonb377, line 77: We can't override XLA_USE_SPMD's value at runtime since we're tracking as a static variable, so I've modified this test to assert we only transfer once. cc @steventk-g

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -228,6 +228,7 @@ ComputationClient::DataPtr PjRtComputationClient::TransferShardsToServer(
     xla::Shape shape, xla::OpSharding sharding) {
   TF_VLOG(1) << "TransferShardsToServer with " << tensor_shards.size()
              << " shards.";
+  // TODO(jonbolin): Consider using CopyToDevice when sharding is REPLICATED
```
yeounoh, line 231: This happens even after the InputHandler, rather, don't we want to move this to the earlier steps?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -474,7 +475,7 @@ at::Tensor XLANativeFunctions::_copy_from(const at::Tensor& self,
   if (!self_tensor) {
     static bool sync_update =
         xla::sys_util::GetEnvBool("XLA_TENSOR_UPDATE_SYNC", true) &&
-        !xla::sys_util::GetEnvBool("XLA_USE_SPMD", false);
+        !ShardingUtil::UseVirtualDevice();
```
yeounoh, line 478: Good :)

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1474,6 +1474,9 @@ void InitXlaModuleBindings(py::module m) {
                                  const py::list& tile_assignment,
                                  bool replicated = false, bool manual = false) {
     TORCH_LAZY_COUNTER("XlaMarkSharding", 1);
+    XLA_CHECK(ShardingUtil::UseVirtualDevice())
```
yeounoh, line 1477: Let's write "Please set `XLA_USE_SPMD=1`"

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -872,10 +872,16 @@ void XLAGraphExecutor::ExtractIRAndPrepareXlaData_(
     torch::lazy::BackendDataPtr handle =
         WrapXlaData(xla::ComputationClient::Get()->CreateDataPlaceholder(
             tensor_device.toString(), std::move(shape)));
-    // Create sharded data placeholder, this will be used to
-    // hold the corresponding computation results.
-    if (tensor->sharding_spec()) {
+    if (ShardingUtil::UseVirtualDevice()) {
```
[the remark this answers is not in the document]
yeounoh: Do you need to attach the implicit sharding annotation (replicated) here?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -203,25 +216,11 @@ ShardingUtil::InputHandler(
...
-      int source_device_i = device_index[global_ordinal];
-      arguments_by_device[source_device_i][argument_i] = shards[0];
-      for (int64_t device_i = 0; device_i < devices.size(); ++device_i) {
-        if (device_i != source_device_i) {
-          arguments_by_device[device_i][argument_i] =
-              xla::ComputationClient::Get()->CopyToDevice(shards[0],
-                                                          devices[device_i]);
-        }
-      }
+    // Input is sharded across addressable devices
```
yeounoh, line 219: Also note the implicit replication case, where the shards are identical.

torch_xla/csrc/xla_sharding_util.h
```diff
@@ -14,6 +14,11 @@ namespace torch_xla {
 
 class ShardingUtil {
  public:
+  // Test whether the XLA_USE_SPMD environment variable is set to enable the
+  // virtual device optimization.
+  // TODO(jonbolin): Setting this is required for SPMD execution.
```
yeounoh, line 19: we can remove TODO if we already have added a check in mark_sharding?

torch_xla/experimental/xla_sharding.py
```diff
@@ -157,5 +157,4 @@ def can_apply(self, t: torch.Tensor) -> bool:
   def apply(self, t: torch.Tensor):
     # TODO(yeounoh) use virtual device interface when available.
     assert (t.device == xm.xla_device())
-    assert (xu.check_env_flag('XLA_USE_SPMD'))
```
yeounoh, line 160: Good :)

torch_xla/csrc/tensor_util.cpp
```diff
@@ -893,6 +926,18 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
     // populate_fn to be called.
     populate_mwait->Wait();
     return async->async_datas;
+  } else if (ShardingUtil::UseVirtualDevice()) {
```
yeounoh, line 929: Let's make move this `if` block to the top, before we check for `if(transfer_async)`. This can be an independent terminal condition, no `if-else`.

torch_xla/csrc/xla_graph_executor.cpp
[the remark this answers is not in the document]
yeounoh: Ok, this is to safe-guard explicit replication.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -872,10 +872,16 @@ void XLAGraphExecutor::ExtractIRAndPrepareXlaData_(
...
+      // Create sharded data placeholder, this will be used to
+      // hold the corresponding computation results.
       auto sharding = tensor->sharding_spec();
+      if (!sharding) {
+        // Implicitly replicate tensors without a user-provided sharding
+        // annotation.
+        sharding = std::make_shared<XLATensor::ShardingSpec>(
```
yeounoh, line 882: For sharding spec, let's also put the tensor shape to initialize.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Let's also explain why we are opting out of copytodevice for now, but also note that copytodevice without blocking is faster than async data transfer from the host.

yeounoh (approved):
LGTM

torch_xla/csrc/xla_graph_executor.cpp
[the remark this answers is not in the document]
jonb377: This ensures implicit replication follows the same codepath as sharded tensors/explicit replication - in this case it's a necessary check. I'll verify all other uses are necessary.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1483,7 +1485,7 @@ void InitXlaModuleBindings(py::module m) {
                       static_cast<XlaDeviceType>(xtensor->GetDevice().type())));
 
     at::Tensor cpu_tensor;
-    if (xla::sys_util::GetEnvBool("XLA_USE_SPMD", false) &&
+    if (ShardingUtil::UseVirtualDevice() &&
```
alanwaketan, line 1488: Do we still need this given the XLA_CHECK above?

torch_xla/csrc/tensor_util.cpp
```diff
@@ -35,6 +35,34 @@ struct DataAsync {
   std::vector<xla::util::ExceptionCleanup> handle_unlockers;
 };
 
+void PopulateTensorBuffer(const at::Tensor& tensor,
```
alanwaketan, line 38: What's this?

torch_xla/csrc/init_python_bindings.cpp
jonb377: Ah good catch! This can be removed

torch_xla/csrc/tensor_util.cpp
alanwaketan: Okay, why not move PopulateTensorBuffer directly here?

jonb377: There are a few related functions to PopulateTensorBuffer that it makes sense to stay near. What do you think about just moving the TransferToServer wrappers down in the file instead? I don't think there's any reason that `TransferToServerAsync` has to be at the top.

jonb377: Updated the PR with the changes discussed

torch_xla/csrc/tensor_util.cpp
```diff
@@ -685,6 +708,11 @@ torch::lazy::BackendDataPtr TensorToXlaData(
     // called.
     populate_mwait->Wait();
     return async->async_datas.front();
+  } else if (ShardingUtil::UseVirtualDevice()) {
+    std::vector<std::string> local_devices =
+        xla::ComputationClient::Get()->GetLocalDevices();
+    return WrapXlaData(TransferToServerReplicated(tensor, device.toString(),
```
alanwaketan, line 714: It looks like we unconditionally replicate all the data? I thought tiled data shouldn't be replicated?

torch_xla/csrc/tensor_util.cpp
```diff
@@ -869,7 +897,19 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
...
-  if (transfer_async) {
+  if (ShardingUtil::UseVirtualDevice()) {
+    // If running in SPMD mode, replicate the data across addressable devices
+    std::vector<std::string> local_devices =
+        xla::ComputationClient::Get()->GetLocalDevices();
+    std::vector<xla::ComputationClient::DataPtr> handles;
+    for (size_t i = 0; i < tensors.size(); ++i) {
+      auto device = ParseDeviceString(devices[i]);
+      auto shape = CreateComputationShapeFromTensor(tensors[i], &device);
+      handles.push_back(TransferToServerReplicated(tensors[i], devices[i],
```
alanwaketan, line 908: It looks like we unconditionally replicate all the data? I thought tiled data shouldn't be replicated?

alanwaketan (commented):
I don't want to block this PR. Feel free to land it and then answer my questions later.

torch_xla/csrc/tensor_util.cpp
jonb377: In this codepath, the tensor is unsharded, and data is being immediately created on a physical device (i.e. bypassing virtual device). This data should be replicated to all devices. This is the path taken by `IrValueFromScalar`

jonb377: FYI I just moved this condition above the `transfer_async && tensor.dim() == 0 && tensor.numel() == 1` condition - PjRt always transfers async, and we should give the replicated transfer precedence

torch_xla/csrc/tensor_util.cpp
```diff
@@ -869,7 +899,19 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
...
+                                                   shape, local_devices));
+    }
+    return WrapXlaData(handles);
+  } else if (transfer_async) {
```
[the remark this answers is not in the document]
jonb377: Hmm this `else if` just covers the original async transfer case, this should still apply for some non-SPMD cases.

torch_xla/csrc/tensor_util.cpp
jonb377: Sounds good, I'll add a comment. The `else if` should still be necessary for some non-SPMD scenarios, I'm just making it so that the SPMD case takes precedence here.

alanwaketan: Yea, it's more like a NIT. To me, it feels like the logic is:
1. SPMD
2. Non SPMD:
2.1. transfer_async
2.2. transfer_sync

But then your coding is:
1. SPMD
2. transfer_async
3. transfer_sync

torch_xla/csrc/tensor_util.cpp
alanwaketan: It looks like there are a lot of places calling `TensorToXlaData`. You mentioned you patched this code path because of a case where the tensor created bypasses the virtual device and then you check UseVirtualDevice before executing the code?

jonb377: Right, this case handles the scenario where `TensorToXlaData` is called for a tensor whose device isn't `DeviceType::SPMD`. Tensors created through [UpdateFromTensor](https://github.com/pytorch/xla/blob/72d2874d83188661325886b32977782478abb47d/torch_xla/csrc/tensor.cpp#L499) will be placed on the virtual device and their data not moved off of the CPU until the `CreateTensorsData` call, and `TensorToXlaData` just returns a placeholder for them here. (cc @steventk-g)

If the virtual device is active and the tensor isn't following the `UpdateFromTensor` path, we want to implicitly replicate it here. In GPT-2 I saw this for scalars in `IrValueFromScalar`, but there may be other scenarios.

Let me know if this explanation makes sense or not, I'd be happy to adjust the approach.

torch_xla/csrc/tensor_util.cpp
jonb377: Ah gotcha! That makes sense. So something like

```
if (SPMD) {
...
return;
}

if (transfer_async) {
...
} else {
...
}
```

I'll update this part to make the logic more clear.

JackCaoG: ehh, I will get rid of `transfer_async` once we get rid of XRT, that was a hacked I did which make codes really messy.

torch_xla/csrc/tensor_util.cpp
```diff
@@ -664,7 +687,14 @@ torch::lazy::BackendDataPtr TensorToXlaData(
 
   static const bool transfer_async =
       xla::sys_util::GetEnvBool("XLA_TRANSFER_SCALAR_ASYNC", false);
-  if (transfer_async && tensor.dim() == 0 && tensor.numel() == 1) {
+  if (ShardingUtil::UseVirtualDevice()) {
```
alanwaketan, line 690: Then should we move this above `if (device.type() == (int8_t)XlaDeviceType::SPMD)`:
```
if (ShardingUtil::UseVirtualDevice()) {
if (device.type() == (int8_t)XlaDeviceType::SPMD) {
...
}
// The tensor is bypassing the virtual device, so it should be replicated
// to all devices.
...
}
```

torch_xla/csrc/tensor_util.cpp
alanwaketan: Got it. Commented on the new code.

torch_xla/csrc/tensor_util.cpp
jonb377: Sure, I'll do that now

alanwaketan (approved):
LGTM. Thanks, Jon.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -48,6 +48,19 @@ std::vector<int64_t> TileAssignmentDimensions(
 
 }  // namespace
 
+bool ShouldUseVirtualDevice() {
+  bool use_virtual_device = xla::sys_util::GetEnvBool("XLA_USE_SPMD", false);
```
JackCaoG, line 52: nit, `static bool`. User should not be able to change env var after process is started

JackCaoG: oh nvm, saw the below function.

JackCaoG (approved):

PR #4980, opened 2023-05-04
Fix TransferToServer Timing

chandrasekhard2:


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -183,9 +183,10 @@ std::optional<xla::OpSharding> PjRtComputationClient::GetDataSharding(
 
 std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
     absl::Span<const TensorSource> tensors) {
-  metrics::TimedSection timed(TransferToServerMetric());
-  tsl::profiler::TraceMe activity("PjRtComputationClient::TransferToServer",
-                                  tsl::profiler::TraceMeLevel::kInfo);
+  auto timed = std::make_shared<metrics::TimedSection>(TransferToServer());
```
JackCaoG, line 186: this is not enough, you need to set the `returned_future` and use the `onReady`, check https://github.com/pytorch/xla/commit/afe7bb6b6562edcd368055899e17c691824575ac. You also need a test.

chandrasekhard2: I'm not an expert but  thought Future is when we're waiting for the result from another process. All we're doing here is explicitly transferring objects one by one. So I thought we don't need those additional future related code.

chandrasekhard2: ignore my previous comment. I added the necessary future logic.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -183,12 +183,15 @@ std::optional<xla::OpSharding> PjRtComputationClient::GetDataSharding(
 
 std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
     absl::Span<const TensorSource> tensors) {
-  metrics::TimedSection timed(TransferToServerMetric());
-  tsl::profiler::TraceMe activity("PjRtComputationClient::TransferToServer",
-                                  tsl::profiler::TraceMeLevel::kInfo);
+  auto timed = std::make_shared<metrics::TimedSection>(TransferToServerMetric());
+  tensorflow::profiler::TraceMe activity(
```
will-cromar, line 187: This should stay `tsl::profiler`, right?

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -209,14 +212,20 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
                 literal_pointer->shape().dimensions(), byte_strides,
                 xla::PjRtClient::HostBufferSemantics::
                     kImmutableUntilTransferCompletes,
-                [literal{std::move(literal)}]() { /* frees literal */ },
+                [literal{std::move(literal)}, &completed_transfers, &returned_future, total_tensors = tensors.size()]() {
```
will-cromar, line 215: Can we make this more simple by copying the `shared_ptr` to the `TimedSection` into each instance of this lambda? Then when the ref count hits 0 (ie all transfer futures have finished), the `shared_ptr` will destruct.

third_party/xla_client/pjrt_computation_client.cc
chandrasekhard2: In ExecuteComputation, I see that its tensorflow::profiler. Whats the difference b/w tensorflow::profiler and tsl::profiler?

third_party/xla_client/pjrt_computation_client.cc
chandrasekhard2: Done!

will-cromar:
Thank, this looks right to me. Can you add a test to confirm?

chandrasekhard2:
> 

Added

PR #4990, opened 2023-05-08
[PJRT] Reduce the number of CPU transfers in rendezvous

will-cromar:
- If the payloads are all empty, just return empty data immediately rather than transferring padding
- For non-empty data, transfer the sizes to CPU once rather than once per replica


JackCaoG (approved):
Thanks！

PR #4991, opened 2023-05-08
Set some helpful defaults in TPU devcontainer

will-cromar:
- Use the [`privileged` option](https://containers.dev/implementors/json_reference/#general-properties)
- Add increased `--shm-size`
- Enable Bazel remote caching with a unique silo by default
- Add the Python VSCode extension

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,6 +2,9 @@
   "name": "pytorch_xla",
   "image": "us-central1-docker.pkg.dev/tpu-pytorch/docker/development:tpu",
   "privileged": true,
+  "runArgs": [
+    "--shm-size=16G"
+  ],
   "containerEnv": {
     "TPUVM_MODE": "1",
     "BAZEL_REMOTE_CACHE": "1",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -8,7 +8,7 @@
   "containerEnv": {
     "TPUVM_MODE": "1",
     "BAZEL_REMOTE_CACHE": "1",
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,13 +1,18 @@
...
   "image": "us-central1-docker.pkg.dev/tpu-pytorch/docker/development:tpu",
+  "privileged": true,
   "runArgs": [
-    "--privileged"
+    "--shm-size=16G"
   ],
+  "containerEnv": {
+    "TPUVM_MODE": "1",
+    "BAZEL_REMOTE_CACHE": "1",
+    "SILO_NAME": "cache-silo-${localEnv:USER}"
```
JackCaoG, line 11: nit, can we do `cache-silo-${localEnv:USER}-tpuvm`? so it doesn't accidentally conflict with cache key with other setups.

mateuszlewko (approved):

.devcontainer/devcontainer.json
stgpetrovic: This is good to have; I had this set up in my user.bazelrc, but this makes more sense since it can take `$USER` as var.

stgpetrovic (approved):

.devcontainer/devcontainer.json
mateuszlewko: TPUVM_MODE: 1, shouldn't be needed, because in the tpu dev image, files .bashrc, and .zshrc, should already contain that env var, see: https://github.com/pytorch/xla/pull/4997/files#diff-4525e4b0185b6f8448b67e991c9f73de2a3ef3ea9b4f745c38172167d7794f58L14.

PR #4992, opened 2023-05-08
Add support for in place operation for dynamo bridge

JackCaoG:
Fix the issue where in place update will error out after pytorch started to pass real input to the bridge. Without this change, all in place update to input tensors will crash the dynamo bridge.

FYI @shunting314


torch_xla/core/dynamo_bridge.py
```diff
@@ -205,6 +205,20 @@ def extract_compiled_graph(xla_model: torch.fx.GraphModule, xla_args):
...
+  # Clone the input tensors which can be used to restore the origional value of the xla_args
+  # if model applied inplace operations to the input.
+  # Note that we currently can't use `cloned_xla_args` to perform the tracing since cloned
+  # Tensor and base tensor will share the same XLAData and XLAData only store the tensor ID
+  # of the first XLATensor, which is the base tensor. When we try to map the XLAData back to
+  # TensorID in `_get_tensors_xla_device_data_node` to create the mapping, the wrong Tensor ID
+  # will be returned.
+  # TODO(JackCaoG): fix the cloned tensor can't be used to warm up the cache.
+  cloned_xla_args = [
+      torch.clone(xla_arg) if isinstance(xla_arg, torch.Tensor) else xla_arg
```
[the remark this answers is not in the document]
JackCaoG: good catch.. didn't really think of that use case. I will adjust the code

JackCaoG:
ok one thing I realized is I only handle the `xla_args` being inplace update case, but I didn't really handle the `model.parameter` being inplace updated case. Looking into that now..

JackCaoG:
> ok one thing I realized is I only handle the `xla_args` being inplace update case, but I didn't really handle the `model.parameter` being inplace updated case. Looking into that now..

Ah with a simple example I can see that is not likely to happen
```
  def model_inplace_update_paramter(self, model, input):
    # RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.
    for i in model.parameters():
      i += 1
    return model(input)

  def test_inplace_update_model_parameter_correctness(self):
    device = xm.xla_device()
    xla_linear = nn.Linear(8, 16).to(device=device)
    xla_batch = torch.rand(4, 8).to(device=device)
    import pdb
    pdb.set_trace()
    return self.model_inplace_update_paramter(xla_linear, xla_batch)
    dynamo_linear = torch.compile(xla_linear, backend='torchxla_trace_once')
```

JackCaoG:
Ok, I tried another test case for the tuple and list and failed with
```
  def inplace_update_container(self, container):
    container[0] += 1
    return container

  def test_inplace_update_container_correctness(self):
    dynamo_inplace_container = torch.compile(
        self.inplace_update_container,
        backend="torchxla_trace_once",
        fullgraph=True)
    t1 = torch.tensor([0, 1, 2], device=xm.xla_device())
    t2 = torch.tensor([0, -1, -2], device=xm.xla_device())
    tuple_t = (t1, t2)
    for i in range(10):
      dynamo_inplace_container(tuple_t)
    self.assertTrue(
        torch.all(torch.eq(tuple_t[0].cpu(), torch.tensor([10, 11, 12]))))
```

```
  File "/usr/local/lib/python3.8/site-packages/torch/_dynamo/variables/base.py", line 253, in call_method
    raise unimplemented(f"call_method {self} {name} {args} {kwargs}")
  File "/usr/local/lib/python3.8/site-packages/torch/_dynamo/exc.py", line 133, in unimplemented
    raise Unsupported(msg)
torch._dynamo.exc.Unsupported: call_method TupleVariable() __setitem__ [ConstantVariable(int), TensorVariable()] {}

from user code:
   File "test_dynamo.py", line 28, in inplace_update_container
    container[0] += 1

Set torch._dynamo.config.verbose=True or TORCHDYNAMO_VERBOSE=1 for more information


You can suppress this exception and fall back to eager by setting:
    import torch._dynamo
    torch._dynamo.config.suppress_errors = True
```

seems like it is something dynamo does not support and we can ignore it for now.

alanwaketan (approved):
LGTM.

test/dynamo/test_dynamo.py
```diff
@@ -18,6 +18,21 @@
 import test_utils
 
 
+class DynamoInPlaceTest(unittest.TestCase):
+
+  def inplace_update(self, a):
+    a += 1
+    return a
+
+  def test_inplace_update_correctness(self):
```
wonjoolee95, line 27: nit: should we be adding compile/execute metrics checks here?

wonjoolee95 (approved):
Thanks! A small nit comment, LGTM. 

cc @seanlatias you'll probably need to rebase your PR one more time to fix some conflicts with this PR merges.

test/dynamo/test_dynamo.py
JackCaoG: good catch, let me add that in a follow up pr.

PR #4995, opened 2023-05-09
Apply sharding when creating tensors data in data loader

jonb377:
Currently, the ParallelLoader path calls `CreateTensorsData` directly to create the XLA tensors without delaying data transfer. When the input is sharded, this incurs overhead of `TransferToServer(full batch)->TransferFromServer(full batch)->TransferShardsToServer(sharded batch)`, when only `TransferShardsToServer(sharded batch)` is required. In general, this overhead is avoided through the virtual device optimization.

In the case of data loading, we do not want to delay transfer at all since the sharding spec is known at the time of XLA tensor creation.

This change enables converting the Python type `xs.ShardingSpec` into the C++ `xla::OpSharding` for use in the sharded `CreateTensorsData` to enable creating sharded data immediately in the ParallelLoader.


test/spmd/test_xla_sharding.py
```diff
@@ -164,6 +164,22 @@ def test_transfer_sharded_data_to_host(self):
...
     self.assertTrue(torch.allclose(t1, torch.ones(16, 16)))
 
+  def test_send_cpu_data_to_device_with_sharding(self):
+    tensor = torch.arange(16).reshape(4, 4)
+    mesh = self._get_mesh((1, self.n_devices))
+    sharding_spec = xs.ShardingSpec(mesh, (0, 1))
+    self.assertTrue(sharding_spec.can_apply(tensor))
+    xtensors = xm.send_cpu_data_to_device([tensor],
+                                          xm.xla_device(),
+                                          input_sharding=sharding_spec)
```
JackCaoG, line 174: can we add a counter check for `TransferToServer` here?

test/spmd/test_xla_sharding.py
```diff
@@ -164,6 +164,22 @@ def test_transfer_sharded_data_to_host(self):
     t1 = xt1.cpu()
     self.assertTrue(torch.allclose(t1, torch.ones(16, 16)))
 
+  def test_send_cpu_data_to_device_with_sharding(self):
+    tensor = torch.arange(16).reshape(4, 4)
+    mesh = self._get_mesh((1, self.n_devices))
+    sharding_spec = xs.ShardingSpec(mesh, (0, 1))
+    self.assertTrue(sharding_spec.can_apply(tensor))
+    xtensors = xm.send_cpu_data_to_device([tensor],
+                                          xm.xla_device(),
```
JackCaoG, line 173: so... the device here will be a virtual device? because `XLA_USE_SPMD=1`?

JackCaoG (commented):
mostly lgtm

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -408,15 +408,64 @@ std::ptrdiff_t GetTensorId(const at::Tensor& tensor) {
...
+// along with the tensors to create sharded XLA tensors. A value of None in
+// the input list corresponds to no sharding on the tensor, which in SPMD mode
+// results in implicit replication of the tensor across devices.
+//
+// If the input `py_sharding_specs` is std::nullopt, then the return value is
+// also std::nullopt.
+std::optional<std::vector<XLATensor::ShardingSpecPtr>> ShardingSpecFromPyHandle(
+    const std::optional<py::list>& py_sharding_specs) {
+  static const py::object pyShardingSpec =
+      py::module_::import("torch_xla.experimental.xla_sharding")
```
alanwaketan, line 422: Can we define a proper PYBIND11_MODULE for sharding spec? Then I guess you don't need this helper.

alanwaketan: Okay, I guess we have two different ShardingSpec type in python and c++ such that you need a conversion. What's the differences? Can we merge them?

jonb377: I don't think we'll be able to merge them, the ShardingSpec in python provides a lightweight wrapper around the Mesh and PartitionSpec while the xla::OpSharding is the proto passed to the XLA compiler.

I like your idea though, I can define a pybind class here wrapping the `xla::OpSharding` to support the conversion in the python layer. I'll try that out, thanks Jiewen!

test/spmd/test_xla_sharding.py
yeounoh: This would be `xla` device in Python, and all `xla` devices will be understood as a virtual `SPMD:0` device, until the actual materialization. Unless @steventk-g already worked on the virtual device Python interface, we should take this change separately, to expose a Python level virtual device.

torch_xla/csrc/init_python_bindings.cpp
yeounoh: +1 @alanwaketan for defining a proper module in Pybind11 for "Python" `ShardingSpec`, I think what you are describing @jonb377 could make things cleaner.

Also, just to add some more context -- the python `ShardingSpec` is to represent sharding info for the data loader and the `mark_sharding` API, while the C++ `ShardingSpec` is a wrapper around `xla::OpSharding` proto for the compiler to work with. 

Some naming conflicts there, but the user should only care about the python one, and how they want to think about it wrt the `mark_sharding` API. The proto should be transparent.

alanwaketan: Thanks for the clarification. I guess maybe we can rename one of them later as a follow up to reduce confusions.

torch_xla/experimental/xla_sharding.py
```diff
@@ -148,6 +148,27 @@ class ShardingSpec:
   mesh: Mesh
   partition_spec: Tuple[Union[int, None]]
 
+  # Derived fields
+  _tile_assignment: List[int] = field(init=False)
+  _replicated: bool = field(init=False)
+  _partial: bool = field(init=False)
+
+  def __post_init__(self):
```
alanwaketan, line 156: Just for my own education, what's the benefit of using __post_init__ v.s. __init__?

alanwaketan (commented):
Mostly LGTM. Looking forward to the official patch.

torch_xla/experimental/xla_sharding.py
jonb377: `dataclass` generates the `__init__` method, which will call into `__post_init__` after initialization.

https://docs.python.org/3/library/dataclasses.html#post-init-processing

alanwaketan: Then I guess ShardingSpec is no longer a data class then. Anyway, I'm fine with it as it is.

alanwaketan (approved):
LGTM.

yeounoh (approved):
LGTM

JackCaoG:
This test failed the TPU CI 
```
Step #4 - "run_e2e_tests": + python3 /src/pytorch/xla/test/spmd/test_xla_sharding.py
Step #4 - "run_e2e_tests": .........F...
Step #4 - "run_e2e_tests": ======================================================================
Step #4 - "run_e2e_tests": FAIL: test_send_cpu_data_to_device_with_sharding (__main__.BasicShardingTest)
Step #4 - "run_e2e_tests": ----------------------------------------------------------------------
Step #4 - "run_e2e_tests": Traceback (most recent call last):
Step #4 - "run_e2e_tests":   File "/src/pytorch/xla/test/spmd/test_xla_sharding.py", line 188, in test_send_cpu_data_to_device_with_sharding
Step #4 - "run_e2e_tests":     self.assertEqual(outbound, tensor.element_size() * tensor.nelement())
Step #4 - "run_e2e_tests": AssertionError: 128.0 != 64
Step #4 - "run_e2e_tests": 
Step #4 - "run_e2e_tests": ----------------------------------------------------------------------
Step #4 - "run_e2e_tests": Ran 13 tests in 0.408s
Step #4 - "run_e2e_tests": 
Step #4 - "run_e2e_tests": FAILED (failures=1)
Step #4 - "run_e2e_tests": ++ kubectl get pod/xla-test-job-2xs9w -o 'jsonpath={.status.containerStatuses[?(@.name=="xla-test")].state.terminated.exitCode}'
Step #4 - "run_e2e_tests": + exit 1
Finished Step #4 - "run_e2e_tests"
ERROR
ERROR: build step 4 "google/cloud-sdk" failed: step exited with non-zero status: 1
```
I am going to revert it

PR #4998, opened 2023-05-11
Make it an error if calling sizes() on a dynamic tensor.

vanbasten23:


vanbasten23:
The current fix
```
at::IntArrayRef XLATensorImpl::sizes_custom() const {
  // XLA_CHECK(!has_symbolic_sizes_strides_)
  xla::Shape xla_shape = tensor_->shape().get();
  XLA_CHECK(!xla_shape.is_dynamic())
      << "Cannot call sizes_custom() on an XLA tensor with symbolic "
         "sizes/strides";
```
is wrong. Doing so would fail at `nonzero`:
```
*** Begin stack trace ***
        torch_xla::XLATensorImpl::sizes_custom() const
      at::functionalization::FunctionalStorageImpl::FunctionalStorageImpl(at::Tensor const&)
        at::FunctionalTensorWrapper::FunctionalTensorWrapper(at::Tensor const&)
        at::functionalization::impl::to_functional_tensor(at::Tensor const&)
        at::_ops::nonzero::redispatch(c10::DispatchKeySet, at::Tensor const&)
        at::_ops::nonzero::call(at::Tensor const&)
```
So I think the correct fix is that we have to use `has_symbolic_sizes_strides_` in tensorImpl instead of xla::Shape. But right now, `has_symbolic_sizes_strides_` is not set so need to figure out where it should be set.

vanbasten23:
hi @ezyang , in this PR I'm trying to make it an error when calling tensor.sizes() on a tensor with dynamic dimensions. Specifically, torch_xla has not set `has_symbolic_sizes_strides_ ` on XLATensorImpl before so this PR tries to do so. Can I check with you about these questions?

1. With my change, `nonzero` op fails in functionalization layer even before we do `dynamic_tensor.sizes()` with [error](https://gist.github.com/vanbasten23/059a0bd84eecd879259a2f45a77cbfe1). I could reproduce the error when running our test `XLA_EXPERIMENTAL="nonzero:masked_select"   python3 pytorch/xla/test/ds/test_dynamic_shapes.py TestDynamicShapes.test_simple_expand`. Upon taking a close look, I found https://github.com/pytorch/pytorch/blob/e248016472fd3b5c8aed270704b0fd21fac20842/aten/src/ATen/FunctionalStorageImpl.cpp#L83 evaluate to false.
- Do you know how to fix it? I tried to add the c10::DispatchKey::Python to https://github.com/pytorch/xla/blob/2855373cd016449f810989b0f9a6dd83b758cf8b/torch_xla/csrc/tensor_impl.cpp#L60 but it didn't help.
- As you know, torch_xla's tensors don't have storage. What should we expect to get from this call `value.storage().sym_nbytes()` from https://github.com/pytorch/pytorch/blob/e248016472fd3b5c8aed270704b0fd21fac20842/aten/src/ATen/FunctionalStorageImpl.cpp#L84?

Thanks.

ezyang:
I wouldn't add Python key to XLA tensor. You may need to modify PyTorch core to properly fix this. I think what you want to do, is make the "recompute from scratch code" (that XLA was previously exercising) work with SymInts. That's mostly calling sym_storage_offset instead of storage_offset, and likewise for anything else that needs it.

vanbasten23:
> is make the "recompute from scratch code" (that XLA was previously exercising) work with SymInts.

Thanks for the response @ezyang . Can you be more specific about where the `recompute from scratch code (that XLA was previously exercising)` is in pytorch's codebase?

ezyang:
this https://github.com/pytorch/pytorch/blob/e248016472fd3b5c8aed270704b0fd21fac20842/aten/src/ATen/FunctionalStorageImpl.cpp#L87-L88

vanbasten23:
The unit test `test_index_types` failed because `XLANativeFunctions::index` uses `nonzero` which introduces dynamism. It failed because in https://github.com/pytorch/xla/blob/7a1e9c74ddf2b229f839d4ad5ec334b13beeb17b/torch_xla/csrc/tensor_util.h#L139-L165, we use `sizes()` on tensors with dynamic dimensions. I guess this PR catches this bug.

Uploaded a commit to fix https://github.com/pytorch/xla/blob/7a1e9c74ddf2b229f839d4ad5ec334b13beeb17b/torch_xla/csrc/tensor_util.h#L139-L165.

vanbasten23:
The last unit test is fixed but found more unit test failures:
```
ERROR: test_fill_ (__main__.TestDynamicShapes)
  File "/tmp/pytorch/xla/test/ds/test_dynamic_shapes.py", line 399, in test_fill_
    t2.fill_(1)
RuntimeError: torch_xla/csrc/aten_xla_type.cpp:1244: SymIntArrayRef expected to contain only concrete integers

FAIL: test_SizeEq_should_not_compile_for_identical_symints (__main__.TestDynamicShapes)
  File "/tmp/pytorch/xla/test/ds/test_dynamic_shapes.py", line 496, in test_SizeEq_should_not_compile_for_identical_symints
    self.assertIsNone(met.metric_data('CompileTime'))
AssertionError: (1, 15493685.0, ((1684807076.5097666, 15493685.0),)) is not None
```

Will fix it today.

vanbasten23:
For the failing test:
```
ERROR: test_fill_ (__main__.TestDynamicShapes)
  File "/tmp/pytorch/xla/test/ds/test_dynamic_shapes.py", line 399, in test_fill_
    t2.fill_(1)
RuntimeError: torch_xla/csrc/aten_xla_type.cpp:1244: SymIntArrayRef expected to contain only concrete integers
```
I would need to symintify `XLANativeFunctions::empty_strided_symint`([here](https://github.com/pytorch/xla/blob/3150573fcafa06b6f5e4321964f9b437f224d37a/torch_xla/csrc/aten_xla_type.cpp#L1239)), `XLANativeFunctions::as_strided_copy`, and `tensor_methods::as_strided`. 

Since torch_xla has enabled functionalization, do we still need the `XLANativeFunctions::empty_strided_symint`?

vanbasten23:
There are 2 remaining test failures:
1) `pytorch/xla/test/test_operations.py TestAtenXlaTensor.test_index_types`: when I run the test it crashes (with `PJRT_DEVICE=TPU`): `Non-OK-status: pjrt_data.buffer->ToLiteralSync(&literal) status: INTERNAL: Error converting to literal: Can't slice buffer of length 2048 with start_offset=1024 length=0`[full output](https://gist.github.com/vanbasten23/0f72cb1a7ed06bf38997234913e05604). But when I use PJRT_DEVICE=TPU_LEGACY, the test passes.
2) `pytorch/xla/test/ds/test_dynamic_shapes.py TestDynamicShapes.test_fill_`: it seems in order to fix it, I need to fix function `XLANativeFunctions::empty_strided_symint`: create a new `XLANativeFunctions::as_strided_copy_symint`, a new `tensor_methods::as_strided_symint`, and a new IR node `AsStrided_Symint`, assuming we can merge https://github.com/pytorch/xla/pull/5084.

hi @will-cromar , regarding the above test failure (1), do you have some input on how to debug it?

vanbasten23:
It seems `XLANativeFunctions::empty_strided_symint` today outputs incorrect stride info.

I added a test:
```
 def test_empty_strided(self):
    a = torch.empty_strided((2, 3), (1, 2), device=dev)
    print('a.stride()=', a.stride())
    print('a.size()=', a.size())
    xm.mark_step()
```
When running `# PJRT_DEVICE=TPU_LEGACY  XLA_EXPERIMENTAL="nonzero:masked_select"   python3 pytorch/xla/test/ds/test_dynamic_shapes.py TestDynamicShapes.test_empty_strided`, I got output:
```
a.stride()= (3, 1)
a.size()= torch.Size([2, 3])
```
However, the correct output for ` a.stride()` should be `(1, 2)` per https://pytorch.org/docs/stable/generated/torch.empty_strided.html.

vanbasten23:
CPU CI is failing. To repo, do `ptxla@t1v-n-0d95297d-w-0:/workspaces/work$ PJRT_DEVICE=CPU CPU_NUM_DEVICES=1  python3 pytorch/xla/test/test_ops.py -v -k TestOpInfoXLA.test_reference_eager_frexp_xla_float32`

vanbasten23:
It turns out that I need to directly return `empty_symint` in `XLANativeFunctions::empty_strided_symint`. This is because otherwise `TestOpInfoXLA.test_reference_eager_frexp_xla_float32` would failed with [error](https://gist.github.com/vanbasten23/e1a36c493ddc54aa15366e8cf58a61c6). The root cause for that is the `stride` passed in `XLANativeFunctions::empty_strided_symint` is [0, 3, 1] ([prints](https://gist.github.com/vanbasten23/e1a36c493ddc54aa15366e8cf58a61c6)): strides should be a positive number. Since pt/xla shouldn't care about stride so I feel we could ignore it even for static size/stride case.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1241,21 +1241,12 @@ at::Tensor XLANativeFunctions::empty_strided_symint(
...
-  } else {
-    return torch_xla::XLANativeFunctions::as_strided_copy(t, size.value(),
-                                                          stride.value(),
-                                                          /*storage_offset=*/0);
-  }
+  // As XLATensor doesn't have a storage, it should not care about the memory
+  // format or how to jump to the next element (strides). So the term stride
+  // does not mean much to us. The size of the tensor has been set by the
+  // above `empty_symint` so we feel it is ok to return here.
+  return empty_symint(sym_size, dtype, layout, device, pin_memory,
```
JackCaoG, line 1248: @alanwaketan to confirm

torch_xla/csrc/ops/index_ops.cpp
```diff
@@ -260,7 +260,9 @@ CanonicalIndexInfo GetCanonicalIndexInfo(
   CheckIndexTensorTypes(orig_indices);
   // First expand ByteTensor (boolean masks) into 1 or more LongTensors, then
   // broadcast all index tensors together.
-  auto indices = xla_expand_outplace(ExpandByteTensors(base, orig_indices));
+  std::vector<at::Tensor> expand_byte_tensors =
+      ExpandByteTensors(base, orig_indices);
+  std::vector<at::Tensor> indices = xla_expand_outplace(expand_byte_tensors);
```
JackCaoG, line 265: what's this change for?

torch_xla/csrc/tensor_util.h
```diff
@@ -146,8 +146,50 @@ inline bool tensor_has_dym_dim(at::Tensor t) {
   return !c10::asIntArrayRefSlowOpt(sym_sizes).has_value();
 }
 
+inline bool tensors_has_dym_dim(at::TensorList tensors) {
+  for (const at::Tensor t : tensors) {
```
JackCaoG, line 150: maybe `const at::Tensor& t` to avoid the copy here?

torch_xla/csrc/tensor_util.h
```diff
@@ -146,8 +146,50 @@ inline bool tensor_has_dym_dim(at::Tensor t) {
...
+    if (tensor_has_dym_dim(t)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+inline std::vector<at::Tensor> xla_expand_outplace_symint(
```
JackCaoG, line 158: I think it is better to call this function `xla_expand_outplace_symint_helper`  and you create another `xla_expand_outplace_helper` which handle the static case. In `xla_expand_outplace` it will have a if check to dispatch to one of the helper.

torch_xla/csrc/ops/index_ops.cpp
vanbasten23: It's just a refactor, breaking one line into two.

test/ds/test_dynamic_shapes.py
```diff
@@ -598,7 +598,10 @@ def test_SizeEq_should_not_compile_for_identical_symints(self):
     dyn_size = t2.shape[0]
     self.assertEqual(dyn_size, dyn_size)
     # Without the code change, met.metric_data('CompileTime')[0] returns 1.
-    self.assertIsNone(met.metric_data('CompileTime'))
+    # self.assertIsNone(met.metric_data('CompileTime'))
```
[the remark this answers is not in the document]
vanbasten23: This is to show what to test after the below `todo` is done. So do you still prefer me to remove this?

torch_xla/csrc/aten_xla_type.cpp
vanbasten23: @alanwaketan has already approved the unconditional case in https://github.com/pytorch/xla/pull/5239..

torch_xla/csrc/tensor_util.h
vanbasten23: Thanks. Fixed.

torch_xla/csrc/tensor_util.h
vanbasten23: fixed

JackCaoG (approved):

vanbasten23:
Thanks for the review!

PR #5002, opened 2023-05-12
Enable SPMD + dynamo for inference

JackCaoG:
This work was done by @yeounoh and I am trying to land this pr in his behalf. The last attempt was made for @steventk-g in https://github.com/pytorch/xla/pull/4862.

Currently test failed with an `Check failed: handle->HasValue()`, so still WIP.

### DIFF
--- a/test/spmd/test_dynamo_spmd.py
+++ b/test/spmd/test_dynamo_spmd.py
@@ -46,9 +46,9 @@ class DynamoSpmdInferenceTest(test_xla_sharding_base.XlaShardingTest):
     xla_res = linear(xla_x)
     xm.mark_step()
 
-    # TODO: this currently failed with `Check failed: handle->HasValue() `
     dynamo_linear = torch.compile(linear, backend="torchxla_trace_once")
     dynamo_res = dynamo_linear(xla_x)
+    torch.allclose(xla_res.cpu(), dynamo_res.cpu())
 
 
 if __name__ == '__main__':
--- a/test/spmd/test_dynamo_spmd.py
+++ b/test/spmd/test_dynamo_spmd.py
@@ -49,6 +49,8 @@ class DynamoSpmdInferenceTest(test_xla_sharding_base.XlaShardingTest):
     dynamo_linear = torch.compile(linear, backend="torchxla_trace_once")
     dynamo_res = dynamo_linear(xla_x)
     torch.allclose(xla_res.cpu(), dynamo_res.cpu())
+    # TODO(JackCaoG): add counter checks after ExecuteReplicated also creates
+    # a ExecuteMetric.
 
 
 if __name__ == '__main__':
--- a/test/spmd/test_dynamo_spmd.py
+++ b/test/spmd/test_dynamo_spmd.py
@@ -1,0 +1,56 @@
+import os
+import sys
+
+import torch
+import torch.nn as nn
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.experimental.xla_sharding as xs
+import torch_xla.debug.metrics as met
+import unittest
+
+import test_xla_sharding_base
+
+
+class SimpleLinear(nn.Module):
+
+  def __init__(self):
+    super(SimpleLinear, self).__init__()
+    self.fc1 = nn.Linear(128, 128)
+    self.relu = nn.ReLU()
+    self.fc2 = nn.Linear(128, 1)
+    # Add an additional 1x1 layer at the end to ensure the final layer
+    # is not sharded.
+    self.fc3 = nn.Linear(1, 1)
+
+  def forward(self, x):
+    y = self.relu(self.fc1(x))
+    z = self.fc2(y)
+    return self.fc3(z)
+
+
+class DynamoSpmdInferenceTest(test_xla_sharding_base.XlaShardingTest):
+
+  @classmethod
+  def setUpClass(cls):
+    os.environ["XLA_USE_SPMD"] = "1"
+    super().setUpClass()
+
+  def test_dynamo_spmd_basic(self):
+    device = xm.xla_device()
+    linear = SimpleLinear().to(device)
+    linear.eval()
+    xla_x = torch.randn(1, 128, device=device)
+    xs.mark_sharding(linear.fc2.weight, self._get_mesh((1, self.n_devices)),
+                     (1, 0))
+    xla_res = linear(xla_x)
+    xm.mark_step()
+
+    # TODO: this currently failed with `Check failed: handle->HasValue() `
+    dynamo_linear = torch.compile(linear, backend="torchxla_trace_once")
+    dynamo_res = dynamo_linear(xla_x)
+
+
+if __name__ == '__main__':
+  test = unittest.main()
+  sys.exit(0 if test.result.wasSuccessful() else 1)
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1728,7 +1728,10 @@ void InitXlaModuleBindings(py::module m) {
             -> std::vector<at::Tensor> {
           XLA_CHECK(hash_str.size() == sizeof(torch::lazy::hash_t));
           torch::lazy::hash_t hash = *(torch::lazy::hash_t*)(hash_str.c_str());
-          torch::lazy::BackendDevice device = torch_xla::GetCurrentDevice();
+          // Device will be Virtual device if SPMD is enabled.
+          torch::lazy::BackendDevice device =
+              ShardingUtil::UseVirtualDevice() ? ParseDeviceString("SPMD:0")
+                                               : torch_xla::GetCurrentDevice();
           auto results = XLAGraphExecutor::Get()->ExecuteComputationWithBarrier(
               hash, graph_inputs, device);
           std::vector<at::Tensor> retlist;
--- a/torch_xla/csrc/xla_graph_executor.cpp
+++ b/torch_xla/csrc/xla_graph_executor.cpp
@@ -605,9 +605,6 @@ XLAGraphExecutor::ExecuteComputationWithBarrier(
     placeholders.push_back(handle);
   }
 
-  // TODO(yeounoh) supply proper sharding specs for sharded results.
-  std::vector<XLATensor::ShardingSpecPtr> sharding_specs(placeholders.size());
-
   SyncTensorCollection coll;
   coll.device = device;
   coll.unlocker = DeviceLockerArena::Get()->LockDevices({device});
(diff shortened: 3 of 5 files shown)


JackCaoG:
OK There are 2 issues
1. Dynamo async function currently silently fail if there is an exception happened. we need to add a rethrow logic
2. dynamo's `PjRtComputationClient::PjRtData::Assign(const Data& data)` failed with bad_cast error

```
void PjRtComputationClient::PjRtData::Assign(const Data& data) {
  TF_VLOG(3) << "enter assign\n";
  const PjRtData& pjrt_data = dynamic_cast<const PjRtData&>(data);
  if (&pjrt_data != this) {
    buffer = pjrt_data.buffer;
  }
  TF_VLOG(3) << "left assign\n";
}
```

```
2023-05-13 00:26:38.471263: I third_party/xla_client/pjrt_computation_client.cc:135] enter assign

E
RuntimeError: std::bad_cast
```

JackCaoG:
Ah Ok.. I think I know what's the problem, the result of the dynamo graph is a `PjRtShardedData`, and we tried to cast it to `PjRtData`. This might has to do with @jonb377 's recent pr that make most things implicitly replicated. This should be a easy fix, I can work on it next week. 

FYI @yeounoh

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1728,7 +1728,10 @@ void InitXlaModuleBindings(py::module m) {
             -> std::vector<at::Tensor> {
           XLA_CHECK(hash_str.size() == sizeof(torch::lazy::hash_t));
           torch::lazy::hash_t hash = *(torch::lazy::hash_t*)(hash_str.c_str());
-          torch::lazy::BackendDevice device = torch_xla::GetCurrentDevice();
+          // Device will be Virtual device if SPMD is enabled.
+          torch::lazy::BackendDevice device =
+              ShardingUtil::UseVirtualDevice() ? ParseDeviceString("SPMD:0")
+                                               : torch_xla::GetCurrentDevice();
```
JackCaoG, line 1734: @yeounoh I am not sure if we should just update `GetCurrentDevice`, any thought? We need to sit down and think about how to surface this virtual device to user soon..

test/spmd/test_dynamo_spmd.py
```diff
@@ -0,0 +1,56 @@
...
+    xs.mark_sharding(linear.fc2.weight, self._get_mesh((1, self.n_devices)),
+                     (1, 0))
+    xla_res = linear(xla_x)
+    xm.mark_step()
+
+    dynamo_linear = torch.compile(linear, backend="torchxla_trace_once")
+    dynamo_res = dynamo_linear(xla_x)
+    torch.allclose(xla_res.cpu(), dynamo_res.cpu())
+
+
```
JackCaoG, line 53: I will try to add more test cases tmr.

JackCaoG:
I think this one is ready for review, I will add more test cases(input data sharding, which I am not sure if it works or not) and features in the next pr.

test/spmd/test_dynamo_spmd.py
```diff
@@ -0,0 +1,56 @@
+import os
```
JackCaoG, line 1: oh I forgot to add this test to the test script... it passes on my local TPUVM

torch_xla/csrc/init_python_bindings.cpp
alanwaketan: I voted for GetCurrentDevice as there might be other scenario where the caller will also need to distinguish SPMD:0 with XLA:0.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -590,9 +593,21 @@ XLAGraphExecutor::ExecuteComputationWithBarrier(
     torch::lazy::BackendDataPtr handle =
         WrapXlaData(xla::ComputationClient::Get()->CreateDataPlaceholder(
             device.toString(), std::move(shape)));
+    // if SPMD is enabled, we assume all output will be replicated
+    if (ShardingUtil::UseVirtualDevice()) {
```
alanwaketan, line 597: Why we now start adding this for the dynamo path? We don't need this for the LTC path?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -590,9 +593,21 @@ XLAGraphExecutor::ExecuteComputationWithBarrier(
...
+              xla::HloSharding::Replicate().ToProto(), shape);
+      handle = WrapXlaData(xla::ComputationClient::Get()->WrapDataShards(
+          {UnwrapXlaData(handle)}, GetVirtualDevice().toString(),
+          sharding->shape.value(), sharding->sharding));
+    }
     placeholders.push_back(handle);
   }
 
+  // TODO(yeounoh) supply proper sharding specs for sharded results.
+  std::vector<XLATensor::ShardingSpecPtr> sharding_specs(placeholders.size());
```
alanwaketan, line 609: Should we merge this with L640?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -608,6 +623,9 @@ XLAGraphExecutor::ExecuteComputationWithBarrier(
       if (auto xla_tensor_ptr = bridge::TryGetXlaTensor(ivalue.toTensor())) {
         dataptr = xla_tensor_ptr->GetXlaData();
       } else {
+        XLA_CHECK(device.type() != (int8_t)XlaDeviceType::SPMD)
```
alanwaketan, line 626: What's this XLA_CHECK for?

torch_xla/csrc/xla_graph_executor.cpp
alanwaketan: Looks like this patch is dynamo exclusive... Should we hint this somewhere?

alanwaketan (approved):
Generally, LGTM.

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: the lazy code path already have this logic, in fact I copt this logic from lazy code path lol

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: I think there is a plan to create this `sharding_specs` properly in the future(after we merge the ouput sharding change), hence left it as a standalont variable for now.

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: hmm, not sure, I copy this from @yeounoh 's diff. @yeounoh any idea?

torch_xla/csrc/xla_graph_executor.cpp
alanwaketan: Then maybe we should move it closer to where it is first used?

torch_xla/csrc/xla_graph_executor.cpp
alanwaketan: I smile an opportunity to merge two code paths more. But let's do it in a follow up.

torch_xla/csrc/init_python_bindings.cpp
JackCaoG: `GetCurrentDevice` is being used over 30 places in our code base now, mostly during tracing and caller trying to figure out the hw type. I think it should be fine as long as `SPMD:0` can be resolved into correct hardware type. I would leave that in a separate pr since it touches too many codes and might introduce noise.

test/spmd/test_dynamo_spmd.py
```diff
@@ -0,0 +1,58 @@
...
+class SimpleLinear(nn.Module):
+
+  def __init__(self):
+    super(SimpleLinear, self).__init__()
+    self.fc1 = nn.Linear(128, 128)
+    self.relu = nn.ReLU()
+    self.fc2 = nn.Linear(128, 1)
+    # Add an additional 1x1 layer at the end to ensure the final layer
+    # is not sharded.
+    self.fc3 = nn.Linear(1, 1)
```
jonb377, line 24: Is this due to the lack of output sharding propagation?

jonb377 (approved):
LGTM, thanks Jack

test/spmd/test_dynamo_spmd.py
JackCaoG: yea, in this pr I tried to keep it that output is replicated. We can expand this after output sharding pr is ready.

alanwaketan (approved):

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Got it.

yeounoh:
> I think this one is ready for review, I will add more test cases(input data sharding, which I am not sure if it works or not) and features in the next pr.

Input sharding should (used to) work if the sharded input is used for the torch compilation. Let me know. Will take a pass on the chages now as well, thanks.

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: It's not needed, but more for a sanity check I probably added to ensure that this doesn't happen. Basically, we want to make sure that the SPMD device type is always on the backend (device data).

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -590,6 +593,15 @@ XLAGraphExecutor::ExecuteComputationWithBarrier(
     torch::lazy::BackendDataPtr handle =
         WrapXlaData(xla::ComputationClient::Get()->CreateDataPlaceholder(
```
yeounoh, line 594: If it's SPMD virtual device, then we should always use PjRtShardedData handle.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -590,6 +593,15 @@ XLAGraphExecutor::ExecuteComputationWithBarrier(
...
             device.toString(), std::move(shape)));
+    // if SPMD is enabled, we assume all output will be replicated
+    if (ShardingUtil::UseVirtualDevice()) {
+      XLATensor::ShardingSpecPtr sharding =
+          std::make_shared<XLATensor::ShardingSpec>(
+              xla::HloSharding::Replicate().ToProto(), shape);
+      handle = WrapXlaData(xla::ComputationClient::Get()->WrapDataShards(
```
yeounoh, line 601: It seems that you are already creating a sharded data handle in case of virtual device, maybe we can update the comment, "if SPMD is enabled, we assume all output will be replicated" to "If SPMD is enabled, we assume all output will be sharded or replicated and wrapped inside PjRtShardedData handle."; this change should work either way.

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: hmm, is the logic below to call `WrapDataShards` not enough? This code path is shared between spmd and non-spmd code path.

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: sg, let me do that.

PR #5008, opened 2023-05-12
[OpenXLA] Migrate Profile to tf/compiler/xla before migrate to use openxla

ManfeiBai:


ManfeiBai:
current migration to tf/compiler/xla is blocked by this func: `xla_bridge.get_backend()`;

do we have a similar func like `xla_bridge.get_backend()` to find and get the backend/device and make sure libtpu initialization? @JackCaoG

ManfeiBai:
will try `torch_xla._found_libtpu`

miladm (commented):
Thanks Manfei. This PR is great. Let's make sure some folks who often use profiling have visibility to this change and can confirm this change will NOT impact them negatively. 

@Liyang90 @JackCaoG @yeounoh @AlexWertheim please comment here.

will-cromar (commented):
I'm all for bringing over profiler improvements from JAX, but can you keep the API consistent with this change? I see there are several functions that change completely with this change, e.g. `trace` to `start_trace`.

I recommend trying to keep this PR as small as possible to get OpenXLA working. We can iteratively make improvements to the profiler module in a compatible way in future PRs.

miladm (commented):
@ManfeiBai are we going to merge this PR? My understand is we no longer plan to support this API. Is so, should we close the PR? Please share your thoughts.

PR #5011, opened 2023-05-15
Move PJRT Python APIs out of `torch_xla.experimental.*`

will-cromar:
Reorganize the experimental PJRT Python APIs.

- Add new `_internal` module for APIs that are well-tested, but likely to change. I moved device-specific logic here, since I expect to rework it in the near future. All of these functions are mainly used for framework development. In general, users shouldn't have to call them directly.
- Most functionality that interacts directly with the runtime is moved into the new `torch.runtime` module.
- Added `deprecation` module to register deprecated aliases for all public functions that are moving out into other parts of of `torch_xla`.
- Print a warning once per deprecated function. For example:

```
>>> from torch_xla.experimental import tpu
>>> tpu.num_available_chips()
WARNING:root:torch_xla.experimental.tpu.num_available_chips is deprecated. Use torch_xla._internal.tpu.num_available_chips instead.
4
>>> tpu.num_available_chips()
4
>>> tpu.version()
WARNING:root:torch_xla.experimental.tpu.version is deprecated. Use torch_xla._internal.tpu.version instead.
4
```

- Update references across the repository.

Summary of new modules:

- `torch_xla.runtime`
- `torch_xla._internal.tpu`
- `torch_xla._internal.gpu`
- `torch_xla._internal.pjrt`


test/pjrt/test_experimental_pjrt_gpu.py
```diff
@@ -11,14 +11,15 @@
 import torch_xla.core.xla_env_vars as xenv
```
JackCaoG, line 11: should we renamed these `experimental` files?

torch_xla/runtime.py
```diff
@@ -0,0 +1,236 @@
...
+  if torch_xla._found_libtpu and os.path.exists('/dev/accel0'):
+    logging.warning('libtpu.so and TPU device found. Setting PJRT_DEVICE=TPU.')
+    os.environ[xenv.PJRT_DEVICE] = 'TPU'
+  else:
+    logging.warning('Defaulting to PJRT_DEVICE=CPU')
+    os.environ[xenv.PJRT_DEVICE] = 'CPU'
+  # TODO(wcromar): Detect GPU device too
+
+
+def device_type() -> Optional[str]:
```
JackCaoG, line 54: I think we have similar functions in `torch_xla.core.xla_model`, do we want to do some clean up?

JackCaoG: I want to take this chance to do some clean up. I am always confuse what function to call for `local ordinal`, `gloabal ordinal`, `worla_size` etc and what do they really mean in a pod context. If we can restructure those api a bit and maybe have those apis in this runtime instead that would be nice...(random idea, might need more thinking)

JackCaoG: `xla_model` becomes a kitch sink and we put random things in it, if we can move all runtime related bits in this file it is actually nicer..

will-cromar: Discussed this offline. We'll start to move APIs that interact directly with the runtime to a new module, and leave any modeling-related APIs in xla_model. I moved the PJRT version of rendezvous back to xla_model, and the old `rendezvous` is will be an alias of that implementation when PJRT is enabled.

test/pjrt/test_experimental_pjrt_gpu.py
will-cromar: Yes, good catch.

torch_xla/runtime.py
```diff
@@ -0,0 +1,188 @@
...
+def _maybe_select_default_device():
+  # Skip if runtime is already configured
+  if xu.getenv_as(
+      xenv.PJRT_SELECT_DEFAULT_DEVICE, str, '1'
+  ) == '0' or xenv.PJRT_DEVICE in os.environ or xenv.GPU_NUM_DEVICES in os.environ or any(
+      env.startswith('XRT_') for env in os.environ):
+    return
+
+  logging.warning(
+      'XRT configuration not detected. Defaulting to preview PJRT '
```
JackCaoG, line 37: do we want to change preview to stable here?

JackCaoG (approved):
LGTM, but prefer to merge on Monday

PR #5016, opened 2023-05-16
Expose local shards as XLAShard

jonb377:
This change exposes local shards as XLAShard containing the shard data and indices into the global tensor. The exposed shards will contain padding, while the indices will be into the unpadded global tensor. Because of this, an additional field `unpadded_data` is introduced to remove padding from the shards.

To support the indices not containing padding, a change is made to `ShardingUtil::GetShardIndicesForDevices`. Previously, this method returned the shard indices as though the global tensor were padded, and the index slicing operations ignored indices which were out-of-bound. This change ensures that the end index is bounded to the tensor dimension sizes, and slices with `start` beyond `stop` do not select any indices (e.g. `tensor[4:3] = tensor([])`)

### DIFF
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -55,17 +55,17 @@ TEST_F(XLAShardingTest, GetShardIndicesForDevices) {
   });
   auto sharding = xla::HloSharding::Tile(mesh).ToProto();
   auto shard_shape = ShardingUtil::GetShardShape(tensor, sharding);
-  auto shard_indices =
-      ShardingUtil::GetShardIndicesForDevices(shard_shape, sharding, devices);
+  auto shard_indices = ShardingUtil::GetShardIndicesForDevices(
+      shard_shape, tensor.sizes().vec(), sharding, devices);
   EXPECT_EQ(shard_indices.size(), devices.size());
   /* Tiled indices should be:
                  dim=0 dim=1
        device=0  [0:4,  0:4]
-       device=1  [0:4,  4:8]
+       device=1  [0:4,  4:7]
        device=2  [4:8,  0:4]
-       device=3  [4:8,  4:8] */
+       device=3  [4:8,  4:7] */
   std::vector<std::vector<int>> slice_starts = {{0, 0}, {0, 4}, {4, 0}, {4, 4}};
-  std::vector<std::vector<int>> slice_ends = {{4, 4}, {4, 8}, {8, 4}, {8, 8}};
+  std::vector<std::vector<int>> slice_ends = {{4, 4}, {4, 7}, {8, 4}, {8, 7}};
   for (int device = 0; device < shard_indices.size(); ++device) {
     EXPECT_EQ(shard_indices[device].size(), tensor.sizes().size());
     for (int dim = 0; dim < shard_indices[device].size(); ++dim) {
@@ -79,8 +79,8 @@ TEST_F(XLAShardingTest, GetShardIndicesForDevices) {
 
   sharding = xla::HloSharding::Replicate().ToProto();
   shard_shape = ShardingUtil::GetShardShape(tensor, sharding);
-  shard_indices =
-      ShardingUtil::GetShardIndicesForDevices(shard_shape, sharding, devices);
+  shard_indices = ShardingUtil::GetShardIndicesForDevices(
+      shard_shape, tensor.sizes().vec(), sharding, devices);
   EXPECT_EQ(shard_indices.size(), devices.size());
   for (int i = 0; i < devices.size(); ++i) {
     EXPECT_EQ(shard_indices[i].size(), 1);
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1,11 +1,13 @@
 import copy
 
 import unittest
+import math
 import numpy as np
 import os
 
 import torch
 from torch import nn
+import torch.nn.functional as F
 import torch.optim as optim
 import torch_xla
 import torch_xla.core.xla_model as xm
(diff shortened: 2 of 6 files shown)


test/spmd/test_xla_sharding.py
```diff
@@ -32,6 +34,64 @@ def test_xla_sharded_tensor(self):
...
+          torch.allclose(shard.cpu().unpadded_data, t[shard.indices]))
+
+  def test_replicated_xla_shards(self):
+    num_element = self.n_devices
+    mesh = self._get_mesh((self.n_devices,))
+    t = torch.arange(num_element, dtype=torch.float32)
+    xt = xs.mark_sharding(t.to(xm.xla_device()), mesh, (None,))
+    shards = xt.local_shards
+    self.assertEqual(len(shards), self.n_devices)
+    for i, shard in enumerate(shards):
```
yeounoh, line 88: Can we also do a quick test to show that the replicated data is not padded?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1537,6 +1537,57 @@ void InitXlaModuleBindings(py::module m) {
     }
     return std::string();
   });
+  m.def("_get_local_shards",
```
yeounoh, line 1540: Add a comment saying that the returned shards are padded, while the indices are not, excluding the pads.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1537,6 +1537,57 @@ void InitXlaModuleBindings(py::module m) {
...
+        [](const at::Tensor& input) -> std::vector<at::Tensor> {
+          XLATensorPtr xtensor = bridge::GetXlaTensor(input);
+          XLA_CHECK(xtensor->GetXlaData() != nullptr)
```
yeounoh, line 1543: This may be redundant, but shall we also check for `XLA_USE_SPMD=1`?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -292,7 +292,8 @@ std::vector<int64_t> ShardingUtil::GetShardShape(
 
 std::vector<std::vector<at::indexing::TensorIndex>>
 ShardingUtil::GetShardIndicesForDevices(
-    const std::vector<int64_t>& shard_shape, const xla::OpSharding sharding,
+    const std::vector<int64_t>& shard_shape,
+    const std::vector<int64_t>& tensor_shape, const xla::OpSharding sharding,
     const std::vector<std::string>& devices) {
   std::vector<std::vector<at::indexing::TensorIndex>> shard_indices(
```
yeounoh, line 298: It's device by tensor dim, is it? Can we also comment on this - a few words?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -292,7 +292,8 @@ std::vector<int64_t> ShardingUtil::GetShardShape(
 
 std::vector<std::vector<at::indexing::TensorIndex>>
 ShardingUtil::GetShardIndicesForDevices(
```
yeounoh, line 294: Would this work for local devices only?

torch_xla/experimental/xla_sharded_tensor.py
```diff
@@ -3,15 +3,37 @@
...
 @dataclass
 class XLAShard:
+  # A snapshot of the shard data from the time of XLAShard creation.
   data: torch.Tensor
-  rank: int
+
+  # The indices of the shard into the global tensor. If the tensor is replicated
+  # across local devices, the value of `indices` is Ellipsis. Otherwise, it is a
+  # list of the index slices across each dimension.
+  # The indices do not reflect padding, since the padding does not exist on the
```
yeounoh, line 19: Good :)

yeounoh (approved):
LGTM, left some minor comments.

test/cpp/test_xla_sharding.cpp
```diff
@@ -55,17 +55,17 @@ TEST_F(XLAShardingTest, GetShardIndicesForDevices) {
...
+      shard_shape, tensor.sizes().vec(), sharding, devices);
   EXPECT_EQ(shard_indices.size(), devices.size());
   /* Tiled indices should be:
                  dim=0 dim=1
        device=0  [0:4,  0:4]
-       device=1  [0:4,  4:8]
+       device=1  [0:4,  4:7]
        device=2  [4:8,  0:4]
-       device=3  [4:8,  4:8] */
+       device=3  [4:8,  4:7] */
```
alanwaketan, line 66: Shouldn't all 8 be replaced by 7?

test/spmd/test_xla_sharding.py
```diff
@@ -32,6 +34,64 @@ def test_xla_sharded_tensor(self):
...
+      self.assertEqual(shard.data.shape, (num_element,))
+      start, end = i * num_element, (i + 1) * num_element
```
alanwaketan, line 90: I don't think we need `start, end` here?

alanwaketan (approved):
LGTM.

test/cpp/test_xla_sharding.cpp
jonb377: Only in the second dimension - the underlying tensor's shape is `(8, 7)`. The first dimension wasn't padded, so it remains unchanged.

test/spmd/test_xla_sharding.py
jonb377: Sure, I'll add an assertion that `data` equals `unpadded_data`

test/spmd/test_xla_sharding.py
jonb377: Ah thanks! I'll clean that up.

torch_xla/csrc/init_python_bindings.cpp
jonb377: Good point, I'll add that check. I'll also enforce that `sharding_spec()` is non-null, since otherwise we'll return a single shard wrapping the tensor's data.

torch_xla/csrc/xla_sharding_util.cpp
jonb377: This will work for any set of devices, including nonlocal.

torch_xla/csrc/xla_sharding_util.cpp
jonb377: That's correct, I'll add a comment

test/cpp/test_xla_sharding.cpp
alanwaketan: Missed the hidden lines...

jonb377:
I'll merge after TPU CI finishes

PR #5018, opened 2023-05-16
[SPMD] Output param sharding

yeounoh:
This is a follow-up work of #4721 , to shard the computation outputs. This will help us avoid resharding the replicated outputs for more efficient stepping. Most notably, 
- we introduce `ShardingUtil::PrepareOutputShardingPropagation` to extract `SharidngSpecPtr` from sync tensors and prepare `PjRtShardedData` placeholders. 
- remove the logic to prepare PjRtShardedData placeholders in `ExtractIRAndPrepareXlaData_` 
- remove `CollectSharidngSpecs` also in favor of `PrepareOutputShardingPropagation`.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -188,7 +188,8 @@ function run_xla_op_tests {
   run_test "$CDIR/pjrt/test_mesh_service.py"
   run_test "$CDIR/spmd/test_xla_sharding.py"
   run_test "$CDIR/spmd/test_xla_virtual_device.py"
-  run_test "$CDIR/spmd/test_dynamo_spmd.py"
+  # TODO(yeounoh) SPMD output sharding is blocking Dynamo integration.
+  #run_test "$CDIR/spmd/test_dynamo_spmd.py"
   run_test "$CDIR/test_operations_hlo.py" "$@" --verbosity=$VERBOSITY
   run_test "$CDIR/test_input_output_aliases.py"
   run_test "$CDIR/test_torch_distributed_xla_backend.py"
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -365,7 +365,7 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
       // to expose the knob for future reference. We can override the compiler's
       // default behavior to further optimize parameter sharding in the future.
       compile_options.executable_build_options
-          .set_allow_spmd_sharding_propagation_to_output({false});
+          .set_allow_spmd_sharding_propagation_to_output({true});
       compile_options.executable_build_options.set_num_partitions(
           client_->device_count());
       compile_options.executable_build_options.set_num_replicas(1);
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -274,6 +274,7 @@ ComputationClient::DataPtr PjRtComputationClient::ReplicateShardedData(
     const ComputationClient::DataPtr& handle) {
   if (PjRtShardedData* sharded_data =
           dynamic_cast<PjRtShardedData*>(handle.get())) {
+    XLA_COUNTER("ReplicateShardedData", 1);
     TF_VLOG(1) << "ReplicateShardedData (handle=" << handle->GetOpaqueHandle()
                << ", shape=" << handle->shape() << ")";
     if (sharded_data->GetSharding().type() == xla::OpSharding::REPLICATED) {
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -231,8 +231,9 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::TransferToServer(
 ComputationClient::DataPtr PjRtComputationClient::TransferShardsToServer(
     absl::Span<const TensorSource> tensor_shards, std::string device,
     xla::Shape shape, xla::OpSharding sharding) {
-  TF_VLOG(1) << "TransferShardsToServer with " << tensor_shards.size()
-             << " shards.";
+  tsl::profiler::TraceMe activity(
+      "PjRtComputationClient::TransferShardsToServer",
+      tsl::profiler::TraceMeLevel::kInfo);
   // TODO(jonbolin): Consider using CopyToDevice when sharding is REPLICATED.
   // We are opting out of CopyToDevice for now due to the synchronization
   // issues observed in ShardingUtil::InputHandler, but because CopyToDevice
@@ -361,9 +362,9 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
     if (instance.is_sharded) {
       // TODO(yeounoh) multi-host, multi-slice configurations
       compile_options.executable_build_options.set_use_spmd_partitioning(true);
-      // TODO(yeounoh) this is set to false by default, but explicitly set here
-      // to expose the knob for future reference. We can override the compiler's
-      // default behavior to further optimize parameter sharding in the future.
+      // We can override the compiler's default behavior to replicate the
+      // outputs. Setting this to true would wrapping the sharded outputs in
+      // PjRtShardedData.
       compile_options.executable_build_options
           .set_allow_spmd_sharding_propagation_to_output({true});
       compile_options.executable_build_options.set_num_partitions(
--- a/torch_xla/csrc/ops/device_data.h
+++ b/torch_xla/csrc/ops/device_data.h
@@ -23,8 +23,11 @@ class DeviceData : public XlaNode {
   }
 
   // With SPMD sharding propagation, we need to update the unpartitioned
-  // backend data with a partitioned one in the node operands.
+  // backend data with a partitioned one in the node operands. Note that
+  // this is permitted only if the node holds a placeholder.
   void Assign(std::shared_ptr<torch::lazy::BackendData> data) {
+    // TODO(yeounoh) check if the existing data is a placeholder after we
+    // address the issue where some of the sync tensors spill with device node.
     XLA_CHECK(data->shape() == data_->shape())
         << "Shape mismatch: expected (" << data_->shape().to_string()
         << "), actual (" << data->shape().to_string() << ")";
--- a/torch_xla/csrc/ops/device_data.h
+++ b/torch_xla/csrc/ops/device_data.h
@@ -22,6 +22,13 @@ class DeviceData : public XlaNode {
     return data_;
   }
 
+  // With SPMD sharding propagation, we need to update the unpartitioned
+  // backend data with a partitioned one in the node operands. The node and its
+  // device data operands are fixed, but the underlying device data address need
+  // to be updated. An alternative to `Assign` would be make the operands of
+  // nodes mutable and modify with new device data nodes.
+  void Assign(std::shared_ptr<torch::lazy::BackendData> data) { data_ = data; }
+
   static DeviceData* Cast(const torch::lazy::Node* node);
 
  private:
--- a/torch_xla/csrc/ops/device_data.h
+++ b/torch_xla/csrc/ops/device_data.h
@@ -23,11 +23,13 @@ class DeviceData : public XlaNode {
   }
 
   // With SPMD sharding propagation, we need to update the unpartitioned
-  // backend data with a partitioned one in the node operands. The node and its
-  // device data operands are fixed, but the underlying device data address need
-  // to be updated. An alternative to `Assign` would be make the operands of
-  // nodes mutable and modify with new device data nodes.
-  void Assign(std::shared_ptr<torch::lazy::BackendData> data) { data_ = data; }
+  // backend data with a partitioned one in the node operands.
+  void Assign(std::shared_ptr<torch::lazy::BackendData> data) {
+    XLA_CHECK(data->shape() == data_->shape())
+        << "Shape mismatch: expected (" << data_->shape().to_string()
+        << "), actual (" << data->shape().to_string() << ")";
+    data_ = data;
+  }
 
   static DeviceData* Cast(const torch::lazy::Node* node);
 
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -197,6 +197,9 @@ torch::lazy::BackendDataPtr XLATensor::GetXlaData() {
   if (up_to_date) {
     torch::lazy::BackendDataPtr handle = CurrentDataHandle();
     if (handle != nullptr) {
+      if (CurrentIrValue())
+        std::cout << "- Current IR: " << CurrentIrValue()->ToString()
+                  << std::endl;
       XLA_CHECK(handle->HasValue())
           << "Trying to access XLA data while an async operation is in flight: "
           << handle->shape();
@@ -451,6 +454,7 @@ at::Tensor XLATensor::ToTensor(bool detached) {
   c10::optional<at::Tensor> tensor_data = CurrentTensorData();
   if (!tensor_data) {
     XLAGraphExecutor::Get()->DeviceBarrier(GetDevice());
+
     // The GetXlaData() call will trigger an ApplyPendingGraph() if an IR
     // XlaNode is available on the tensor.
     std::vector<at::Tensor> tensors = XlaDataToTensors({GetXlaData()}, dtype());
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -451,7 +451,6 @@ at::Tensor XLATensor::ToTensor(bool detached) {
   c10::optional<at::Tensor> tensor_data = CurrentTensorData();
   if (!tensor_data) {
     XLAGraphExecutor::Get()->DeviceBarrier(GetDevice());
-
     // The GetXlaData() call will trigger an ApplyPendingGraph() if an IR
     // XlaNode is available on the tensor.
     std::vector<at::Tensor> tensors = XlaDataToTensors({GetXlaData()}, dtype());
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -197,9 +197,6 @@ torch::lazy::BackendDataPtr XLATensor::GetXlaData() {
   if (up_to_date) {
     torch::lazy::BackendDataPtr handle = CurrentDataHandle();
     if (handle != nullptr) {
-      if (CurrentIrValue())
-        std::cout << "- Current IR: " << CurrentIrValue()->ToString()
-                  << std::endl;
       XLA_CHECK(handle->HasValue())
           << "Trying to access XLA data while an async operation is in flight: "
           << handle->shape();
--- a/torch_xla/csrc/xla_graph_executor.cpp
+++ b/torch_xla/csrc/xla_graph_executor.cpp
@@ -495,14 +495,14 @@ XLAGraphExecutor::SyncTensorCollection XLAGraphExecutor::CollectSyncTensors(
   if (!unique_device) {
     return coll;
   }
-
+  std::cout << "*** CollectSyncTensors ... " << std::endl;
   std::vector<at::Tensor> at_tensors;
   std::vector<std::string> devices;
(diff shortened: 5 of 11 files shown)


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -351,7 +352,7 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
       // to expose the knob for future reference. We can override the compiler's
```
JackCaoG, line 352: maybe you should update the comment here as well.

torch_xla/csrc/ops/device_data.h
```diff
@@ -22,6 +22,13 @@ class DeviceData : public XlaNode {
     return data_;
   }
 
+  // With SPMD sharding propagation, we need to update the unpartitioned
+  // backend data with a partitioned one in the node operands. The node and its
+  // device data operands are fixed, but the underlying device data address need
+  // to be updated. An alternative to `Assign` would be make the operands of
+  // nodes mutable and modify with new device data nodes.
+  void Assign(std::shared_ptr<torch::lazy::BackendData> data) { data_ = data; }
```
JackCaoG, line 30: we should at very least check if shape matches before modifying the underlying data

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1028,7 +1008,12 @@ XLAGraphExecutor::ScheduleSyncTensorsGraph(
     const std::vector<torch::lazy::BackendDataPtr>& tensor_data_vec) {
   auto tensors_data =
       SetTensorData(tensors, coll->config, coll->indices, tensor_data_vec);
-  auto sharding_specs = CollectShardingSpecs(tensors, coll->indices);
+  std::vector<XLATensor::ShardingSpecPtr> sharding_specs(coll->indices.size(),
+                                                         nullptr);
+  // Extract sharding specs for the results and prepares the data placeholders.
+  ShardingUtil::PrepareOutputShardingPropagation(
+      tensors, coll->indices, cached_computation->computation, &tensors_data,
+      &sharding_specs);
```
JackCaoG, line 1016: shouldn't we check whether graph is sharded before calling these functions?

yeounoh: This call is a no-op if the graph is not sharded, as this is used in the SyncTensors code path.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -382,4 +383,52 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
...
+      (*sharding_specs)[i] = std::make_shared<XLATensor::ShardingSpec>(
+          tuple_shardings[i],
+          MakeShapeWithDeviceLayout(
+              xtensor->shape().get(),
+              static_cast<XlaDeviceType>(xtensor->GetDevice().type())));
+      xtensor->SetShardingSpec(*(*sharding_specs)[i]);
+
+      // Create sharded data placeholder, this will be used to
+      // hold the corresponding computation results.
+      auto sharded_data_placeholder =
```
jonb377, line 413: In `xla_graph_executor.cpp`, the sharded data preparation for implicitly replicated data is removed. Should we still prepare a sharded placeholder regardless of `tuple_shardings[i].type()`?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -382,4 +383,52 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
...
+      xtensor->data()->handle = (*data_placeholders)[i];
+      if (auto ir_value = xtensor->CurrentIrValue()) {
+        if (DeviceData* device_data_node =
+                DeviceData::Cast(ir_value.node.get())) {
+          device_data_node->Assign(xtensor->data()->handle);
+        }
+      }
+    } else {
+      // Clear sharding if the output parameter is no longer sharded.
+      xtensor->ClearShardingSpec();
```
JackCaoG, line 430: in what case output tensor will not be sharded? I thought under SPMD it is either sharded or replicated.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -542,25 +543,25 @@ XLAGraphExecutor::SyncTensorCollection XLAGraphExecutor::CollectSyncTensors(
...
   coll.hash = torch::lazy::MHash(
       coll.hash,
       xla::ComputationClient::Get()->GetResourceDomain(coll.device.toString()));
   if (!at_tensors.empty()) {
     TORCH_LAZY_COUNTER("SyncTensorsToData", at_tensors.size());
     // Create data handles with shardings. If a tensor has a
     // sharding annotation, then a BackendDataPtr with PjRtShardedData is
-    // returned; if there is no sharding annotation, then a BackendDataPtr with
-    // PjRtData is returned.
+    // returned; if there is no sharding annotation, then a BackendDataPtr
```
JackCaoG, line 555: any chance we can undo these comment changes? They make reviewing a bit difficult

torch_xla/csrc/tensor.cpp
```diff
@@ -451,6 +451,7 @@ at::Tensor XLATensor::ToTensor(bool detached) {
   c10::optional<at::Tensor> tensor_data = CurrentTensorData();
   if (!tensor_data) {
     XLAGraphExecutor::Get()->DeviceBarrier(GetDevice());
+
```
JackCaoG, line 454: maybe undo this

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: In summary, we shouldn't prepare the sharded data placeholder in the graph executor, but do so using this `ShardingUtil` API. This happens based on `tuple_shardings[i].type`, not "regardless of" it. Um, let me know if you still have questions.

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: right, I would still prefer put this under a if condition, because
1. it is easy to read, I don't need to go into `PrepareOutputShardingPropagation` and figured that it is a no-op if it is not sharded
2. we don't need to allocate `sharding_specs` for non-SPMD
3. if one day we update `PrepareOutputShardingPropagation` and forgot about the caller always call(even for non-spmd) we might introduce weird issue.

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: Correct. When it's replicated, it's no longer sharded. If user decides to unshard or the compiler decides to remove the propagation, then we that needs to be reflected on the XLATensor's as source of the truth. This is a safeguard for the edge use cases.

JackCaoG: oh ok, when you say it is not `sharded`, I thought it means it does not has sharding, hence it won't even be replicated. I guess comment here can be more specified about the replication.

torch_xla/csrc/xla_sharding_util.cpp
jonb377: Just for my understanding - in the implicit replication case, the data is wrapped in `PjRtShardedData` and `xtensor->sharding_spec()` remains nullptr. In the `OutputHandler`, we're wrapping the resulting shards in `PjRtShardedData`. If we don't override the placeholder with `PjRtShardedData` here, we'll be trying to `Assign` a `PjRtShardedData` to a `PjRtData` placeholder.

If this is working, it seems like the `tuple_shardings[i].type()` is non-false even for the implicitly replicated case, unless I'm missing something?

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: I see, good point. I think a better place would be to initialize and call `PrepareOutputShardingPropagation` if (async->is_sharded...)` just before the execution. Let me make some changes.

JackCaoG: make sure do that before the `barrier` call in https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_graph_executor.cpp#L746, otherwise this will be blocked by the execution previous execution and can not be overlapped (which will increase the overhead of tracing in general).

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -382,4 +383,55 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
   return shards;
 }
 
+void ShardingUtil::PrepareOutputShardingPropagation(
+    std::vector<XLATensorPtr>* tensors, absl::Span<const size_t> indices,
+    ComputationPtr computation,
+    std::vector<torch::lazy::BackendDataPtr>* data_placeholders,
```
alanwaketan, line 389: why pointer instead of reference?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -382,4 +383,55 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
...
+  if (tuple_shardings.empty()) {
+    return;
+  }
+  XLA_CHECK(indices.size() == tuple_shardings.size())
+      << "Expected " << indices.size() << " sharding annotations, but got "
+      << tuple_shardings.size();
+
+  for (int i = 0; i < indices.size(); ++i) {
+    auto xtensor = (*tensors)[indices[i]];
+    std::cout << "*** " << xtensor->CurrentIrValue()->ToString() << " sharded? "
```
alanwaketan, line 403: Log?

alanwaketan (commented):
Can you briefly describe the code being removed?

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: No particular or a strong reason, just wanted to be consistent with other inputs `std::vector<XLATensorPtr>* tensors` .

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: The PR's in draft -- I am testing a few extra bits with probing. Will ping you when it's ready :) thanks.

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: After merging with the recent implicit replication changes, all the outputs in case of sharded computation should be wrapped inside PjRtShardedData. This is a bug, will update the PR.

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: Great, I am also adding the implicitly replicated PjRtShardedData here after rebasing.

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: yea, also sourcing the inputs to the function seems untidy.. going with your original suggestion:
```
// Extract sharding specs for the results and prepare the sharded data
  // placeholders if the computation is sharded.
  if (cached_computation->is_sharded) {
    ShardingUtil::PrepareOutputShardingPropagation(
        tensors, coll->indices, cached_computation->computation, &tensors_data,
        &sharding_specs);
  }
```
we still need to initialize the sharding_specs with null, in case of a unsharded compuation.

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: Ok removed the `std::cout`, I will mark this PR as ready-for-review tmr morning.

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: Removed unnecessary linting changes.

third_party/xla_client/pjrt_computation_client.cc
yeounoh: > Can you briefly describe the code being removed?

Updated the PR description, thanks :)

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -382,4 +383,60 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
...
+  if (tuple_shardings.empty()) {
+    return;
+  }
+  XLA_CHECK(indices.size() == tuple_shardings.size())
+      << "Expected " << indices.size() << " sharding annotations, but got "
+      << tuple_shardings.size();
+
+  for (int i = 0; i < indices.size(); ++i) {
+    auto xtensor = (*tensors)[indices[i]];
+    if (tuple_shardings[i].type()) {
```
jonb377, line 403: Could you add a comment explaining what this is testing?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1028,7 +1000,17 @@ XLAGraphExecutor::ScheduleSyncTensorsGraph(
     const std::vector<torch::lazy::BackendDataPtr>& tensor_data_vec) {
   auto tensors_data =
       SetTensorData(tensors, coll->config, coll->indices, tensor_data_vec);
```
jonb377, line 1002: It seems like we have to override some of the work done in `SetTensorData` in `PrepareOutputShardingPropagation`, is there a reason we can't call `PrepareOutputShardingPropagation` first?

jonb377 (approved):
LGTM, thanks Yeounoh! This is great

torch_xla/csrc/xla_sharding_util.cpp
alanwaketan: We can clean up this later. It seems weird to use raw pointers...

alanwaketan (approved):
LGTM.

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: Good question, and the answer is no (though we can make changes to both `SetTensorData` and `PrepareOutputShardingPropagation` to make it work in the reverse order, more work) sine we still want to hit `handle == nullptr` case there. I think since output sharding prop is conditional, I would prefer keeping the non SPMD code path intact if possible.

yeounoh:
The CI is failing, which seems real. I will address & add more unit test cases:
```
ERROR: test_mark_sharding_2d (__main__.BasicShardingTest)
  File "/tmp/pytorch/xla/test/spmd/test_xla_sharding.py", line 60, in test_mark_sharding_2d
    actual = (xt1 + xt2).cpu()
RuntimeError: std::bad_cast

ERROR: test_mark_sharding_4d (__main__.BasicShardingTest)
  File "/tmp/pytorch/xla/test/spmd/test_xla_sharding.py", line 78, in test_mark_sharding_4d
    actual = (xt + xt).cpu()
RuntimeError: std::bad_cast

ERROR: test_optimizer_step_with_sharding (__main__.BasicShardingTest)
  File "/tmp/pytorch/xla/test/spmd/test_xla_sharding.py", line 138, in test_optimizer_step_with_sharding
    xm.mark_step()
  File "/opt/conda/lib/python3.8/site-packages/torch_xla-2.1.0-py3.8-linux-x86_64.egg/torch_xla/core/xla_model.py", line 832, in mark_step
    torch_xla._XLAC._xla_step_marker(
RuntimeError: std::bad_cast
```

torch_xla/csrc/xla_sharding_util.cpp
yeounoh:
```
// Output parameter sharding annotations, defaults to REPLICATED(0) if unset.
auto tuple_shardings = ...
...
if (tuple_shardings[i].type()) {
      // Tensor sharding annotation type is non-zero (sharded).
```

torch_xla/csrc/ops/device_data.h
```diff
@@ -22,6 +22,15 @@ class DeviceData : public XlaNode {
     return data_;
   }
 
+  // With SPMD sharding propagation, we need to update the unpartitioned
+  // backend data with a partitioned one in the node operands.
+  void Assign(std::shared_ptr<torch::lazy::BackendData> data) {
+    XLA_CHECK(data->shape() == data_->shape())
```
JackCaoG, line 28: nit, let's also checkout about `!data_->hasValue()`, we should only call Assign on placeholders.

JackCaoG: I take this as a short term solution. In the long term we should figure out why DeviceData is being generated by on a placeholder during a `SyncTensorGraphsInternal` call.

yeounoh: Ah, for the current use case it's true -- do we want to restrict the API like that? Alternatively, we can pass the ownership (move) if the data is not a placeholder if someone tries to use the API. Hmm I am split, and see that the restriction would be safer semantically. WDYT?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -382,4 +383,64 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
...
+            (*sharding_specs)[i]->shape.value(),
+            (*sharding_specs)[i]->sharding));
+
+    // Register the sharded data placeholder to the tensor and its node.
+    (*data_placeholders)[i] = sharded_data_placeholder;
+    xtensor->data()->handle = (*data_placeholders)[i];
+    if (auto ir_value = xtensor->CurrentIrValue()) {
+      if (DeviceData* device_data_node =
+              DeviceData::Cast(ir_value.node.get())) {
+        device_data_node->Assign(xtensor->data()->handle);
```
JackCaoG, line 441: Do we still expect this part of the code being executed? There shouldn't be any IR being generated when there is a placeholder in place. I think we should error out here instead of silently patching the node.

torch_xla/csrc/ops/device_data.h
yeounoh: Will restrict it so that this is only used for placholders until we find more use cases (justification) cc @JackCaoG

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: This is needed since we don't reset IR when we set placeholders -- https://github.com/pytorch/xla/blob/dc057dcda60355ea38d5686e5baafe677b838564/torch_xla/csrc/xla_graph_executor.cpp#L916

JackCaoG: I am still a bit confuse in 
https://github.com/pytorch/xla/blob/dc057dcda60355ea38d5686e5baafe677b838564/torch_xla/csrc/xla_graph_executor.cpp#L965-L967
we reset IRs for all of the tensors that will get a new `XlaData` after the execution and in https://github.com/pytorch/xla/blob/dc057dcda60355ea38d5686e5baafe677b838564/torch_xla/csrc/xla_graph_executor.cpp#L915-L923 we assign the placeholder to those XlaTensor(which doesn't have any IR). 

In logic here we are replacing the `unsharded placeholder` with the `sharded_data_placeholder`. I am not sure in what case there will be a `DeviceData` Node that is pointed to the `unsharded placeholder`

alanwaketan:
cc @khatwanimohit

torch_xla/csrc/xla_sharding_util.cpp
JackCaoG: discussed offline with yeounoh, we will land this pr as it is and I will pick this us and fiugre out what is going on so we don't need to block the benchmark.

test/run_tests.sh
```diff
@@ -188,7 +188,8 @@ function run_xla_op_tests {
   run_test "$CDIR/pjrt/test_mesh_service.py"
   run_test "$CDIR/spmd/test_xla_sharding.py"
   run_test "$CDIR/spmd/test_xla_virtual_device.py"
-  run_test "$CDIR/spmd/test_dynamo_spmd.py"
+  # TODO(yeounoh) SPMD output sharding is blocking Dynamo integration.
```
yeounoh, line 191: @JackCaoG we need to revisit SPMD+Dynamo integration after output sharding. Let's take it outside the current scope, as it requires some more work & design.

yeounoh:
GPT-2 benchmark and basic sharding tests are passing. Noticed that this is breaking the current SPMD+Dynamo integration, I will take that outside the current scope and handle separately. We need to enable Dynamo with output sharding anyway. cc @JackCaoG

PR #5023, opened 2023-05-17
Tf pin update

cowanmeg:
Update libtpu version and tf pin

TF patches:
- Removed: gpu_hlo_schedule.diff, small_constants_optimizer.diff
- Added: absl_statusor.diff (reverts: 0946f40bab84fb3f23648eb2544c7a46fd9e0700) and cuda_graph.diff (reverts: 8aa20d47ea5c863b9edeb072489ffb7c38b6e8eb)

XLA updates:
 - tensorflow::MakeXlaBackpropFilterConvOp no longer requires PrecisionConfig as an argument


JackCaoG:
@cowanmeg Did you have a chance to run this with new libtpu and make sure the speed does not regress on a simple resnet50 example?

vanbasten23:
Can you also run the TPU CI?

tf_patches/cuda_graph.diff
```diff
@@ -0,0 +1,194 @@
+Revert 8aa20d47ea5c863b9edeb072489ffb7c38b6e8eb
```
JackCaoG, line 1: curious, why do we need to revert this cl?

JackCaoG: ok Let's follow up in this change later. If I understand correctly this is a compiler issue and it will get fixed if we bump our compiler version? This file will be needed even if get rid of the XRT, so we need to figure out a way before this diff becomes even larger. @stgpetrovic in case you have some insights.

cowanmeg: Yes, both reverts can be removed once we bump our compiler version.

vanbasten23 (approved):

JackCaoG (approved):
Thanks!

PR #5032, opened 2023-05-18
Place XLAShard data on CPU on creation

jonb377:
As a follow-up from #5016, we should place the shards on CPU by default. This makes it clear to the user that the shards will not remain up-to-date with the underlying XLAShardedTensor, and it eases interoperability of the shards' data with other tensors.

Expanding on the second point, the shards are backed by PjRtData when returned from `_get_local_shards`. In SPMD mode, we expect all computation inputs to have PjRtShardedData handles, which makes the on-device shards incompatible with SPMD execution. In order to use the shards in an on-device computation, they will now need to be transferred back to the device, which implies implicit replication.

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -45,14 +45,15 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertEqual(len(shards), self.n_devices)
     shard_len = math.ceil(num_element / self.n_devices)
     for i, shard in enumerate(shards):
+      self.assertEqual(shard.data.device, torch.device('cpu'))
       self.assertEqual(shard.data.shape, (shard_len,))
       start, end = (i, i + 1) * shard_len
       expected = torch.arange(start, end, dtype=torch.float32)
-      self.assertTrue(torch.allclose(shard.data.cpu(), expected))
+      self.assertTrue(torch.allclose(shard.data, expected))
       self.assertIsInstance(shard.indices, list)
       self.assertEqual(len(shard.indices), len(t.shape))
       self.assertEqual(shard.indices[0], slice(start, end, 1))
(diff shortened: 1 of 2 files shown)


test/spmd/test_xla_sharding.py
```diff
@@ -48,11 +48,11 @@ def test_xla_shards(self):
       self.assertEqual(shard.data.shape, (shard_len,))
       start, end = (i, i + 1) * shard_len
       expected = torch.arange(start, end, dtype=torch.float32)
-      self.assertTrue(torch.allclose(shard.data.cpu(), expected))
+      self.assertTrue(torch.allclose(shard.data, expected))
```
JackCaoG, line 51: can we add a assert above to check that `shard.data` is on cpu device. Just to make check more explict.

jonb377: Good point, added

alanwaketan (approved):
LGTM.

JackCaoG (approved):

torch_xla/experimental/xla_sharded_tensor.py
```diff
@@ -97,7 +94,7 @@ def __new__(cls, elem: torch.Tensor, *args, **kwargs):
 
   # Shards on the devices are materialized/available after the lazy
   # execution of the SPMDPartitioned HLO graph. Each XLAShard points
-  # to torch.Tensor (xla::device_data). The shards represent a snapshot
+  # to torch.Tensor (xla::device_data). The shards represent a snapshot on CPU
```
yeounoh, line 97: nit. Could we update the comment here, "execution of the partitioned HLO graph. Each XLAShard points to torch.Tensor. The shards represent a snapshot on CPU, detached from the global tensor." ?

yeounoh (approved):
Left a minor comment, LGTM.

PR #5064, opened 2023-05-24
Add dev container config for external contributors

mateuszlewko:
1. Now when you invoke "Reopen in Container" there will by 2 options: `tpu-internal` and `tpu-contributor`. The second one doesn't use bazel cache.
2. Both dev containers now use the public docker development image.

### DIFF
--- a/.devcontainer/tpu-contributor/devcontainer.json
+++ b/.devcontainer/tpu-contributor/devcontainer.json
@@ -1,16 +1,12 @@
 {
-  "name": "pytorch_xla",
-  "image": "us-central1-docker.pkg.dev/tpu-pytorch/docker/development:tpu",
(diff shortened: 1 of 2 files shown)


will-cromar (approved):
Thanks! Does VSCode give us any tools to merge the common configuration? The only difference here seems to be the `containerEnv`

will-cromar (approved):
Thanks! Does VSCode give us any tools to merge the common configuration? The only difference here seems to be the `containerEnv`

mateuszlewko:
Yeah, env is the only difference. Based on https://containers.dev/implementors/json_reference/ I don't see any option to inherit or override devcontainer configs. We'd have to use some other tool (jsonnet?) for deduplicating json.

PR #5065, opened 2023-05-24
Breaking depenency cycles.

stgpetrovic:
The plugin C++ code is very tangled, every header including almost every other. This is an attempt in simplifying the build graph and consequently speeding up the incremental build.


bazel/rules_def.bzl
```diff
@@ -28,6 +28,7 @@ def ptxla_cc_test(
             "-isystemexternal/torch",  # Required for system includes.
             "-fexceptions",  # Required for testing crashes.
         ],
+        hdrs_check = "strict",
```
will-cromar, line 31: Is this an option defined in tf_cc_test? I don't see it documented in Bazel: https://bazel.build/reference/be/c-cpp

torch_xla/csrc/xla_data.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "third_party/xla_client/computation_client.h"
+
+namespace torch_xla {
+
+// To be noted, the most appropriate way to adopt BackendData
+// should actually be letting ComputationClient::Data inherit it.
+// Since ComputationClient is within TensorFlow, and TF cannot
+// depend on PyTorch. Therefore, we have this intermediate wrapper.
+class XLAData : public torch::lazy::BackendData {
```
will-cromar, line 16: `Since ComputationClient is within TensorFlow, and TF cannot depend on PyTorch.` isn't true anymore. @JackCaoG can we flatten the dependency chain here?

We can do this in a follow-up PR if so.

bazel/rules_def.bzl
will-cromar: From that document: `All header files that are used in the build must be declared in the hdrs or srcs of cc_* rules. This is enforced.`

Does that option have any effect in open-source bazel?

https://bazel.build/reference/be/c-cpp#hdrs

torch_xla/csrc/xla_data.h
JackCaoG: yea, if computation client can import pytorch file, we should get rid of all of these `Wrap` and `Unwrap` thing.

will-cromar: I can take this action item since I'm already untangling the computation_client build target.

bazel/rules_def.bzl
stgpetrovic: This should be the default, but it does not work with gcc I think (there's a gazillion transitive imports in this project and it does not fail the build). I'll remove it though, we should move to clang when upstream issues are fixed with the cppmap declarative outputs.

torch_xla/csrc/unwrap_data.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <memory>
+#include <vector>
+
+#include "absl/types/span.h"
+#include "third_party/xla_client/computation_client.h"
+#include "torch_xla/csrc/xla_data.h"
+
+namespace torch_xla {
+
+xla::ComputationClient::DataPtr UnwrapXlaData(
```
JackCaoG, line 16: Let's also get rid of it in later pr, if xla_client can inherit from pytorch Data class directl.y

JackCaoG (approved):
Thanks @stgpetrovic, this is amazing!

PR #5068, opened 2023-05-24
Don't let autograd initialize the ComputationClient.

will-cromar:
Print a warning at a verbose log level when the client is null.


JackCaoG (approved):

carmocca:
Is something blocking this PR?

will-cromar:
My bad. I'll remove the conflict and merge when tests pass.

PR #5082, opened 2023-05-25
Python test bazel

stgpetrovic:
Migrate some python tests to bazel.

### DIFF
--- a/bazel/dependencies.bzl
+++ b/bazel/dependencies.bzl
@@ -1,1 +1,1 @@
-PYTORCH_LOCAL_DIR = "../"
+PYTORCH_LOCAL_DIR = "../pytorch/"
--- a/bazel/rules_def.bzl
+++ b/bazel/rules_def.bzl
@@ -84,13 +84,13 @@ def union(a, b):
 def ptxla_py_test(
         name = "name",
         modes = ["default"],
-        accelerators = ["cpu"],
+        accelerators = ["cpu", "gpu", "tpu"],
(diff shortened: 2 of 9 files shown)


vanbasten23:
Can you help me understand the benefit of moving python test to bazel?

Also, would the old way (python3 pytorch/xla/test/test_foo.py) still work after the pr?

stgpetrovic:
Taking a look at the GPU tests.

The old way would still work. I think that we should consolidate everything under bazel to get consistent caching, parallelism, coverage, symbol cross references etc.

You can get parallelism speedups on CPU tests that are the first to run on each PR: you can see here it took 39min to run CPU tests (including the C++ tests and the unmigrated torch-op tests that are still the bulk of python tests), which is already 6min less than `master` branch has it. And when we migrate the torch op tests, the critical path will go down by a lot.

I suppose you could paralellize the test runs in the bash script too, but that's not easy to maintain (in bazel with `--config=tpu` it will run with a single test executor worker making it serial).

Also, having the configuration be declarative (`modes = ["default", "dynamic"]`) as a property of the test, rather than having 2 invocations of bash functions is more manageable and when tests fail, the name of the test contains both the accelerator (tpu,cpu,gpu) and the configuration it was running in (downcast_bf16, use_bf16), it's a bit easier to debug.

I'm doing this just to see what the potential speedups are but I'm happy to leave the branch here and you can evaluate later if you'd like something like this or not and then merge it.

PR #5078, opened 2023-05-25
Move the Lock acquireing to the end of ExecuteComputation and Execute…

JackCaoG:
@jonb377  I move the lock acquiring to a async function, hopefully this can make the gap go away..

PR #5085, opened 2023-05-26
Revert "Move the Lock acquireing to the end of ExecuteComputation and Execute…"

JackCaoG:
Reverts pytorch/xla#5078 while I am investigating the failure in TPU CI


JackCaoG:
nvm, I think I know how to fix the issue, will send a fix shortly.

PR #5086, opened 2023-05-26
pass the PJRTFuture to lambda directly

JackCaoG:
There seems to be an issue of passing `std::optional` to the lambda function. I was able to see that `returned_future ->isValid` become false when passing optional. Passing the real PJRTFuture with `std::move` solved the issue.

### DIFF
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -482,7 +482,7 @@ PjRtComputationClient::ExecuteComputation(
   CreateDataHandlesCounter()->AddValue(datas.size());
 
   auto mwait = std::make_shared<util::MultiWait>(1);
-  auto lockfn = [&, this, device, returned_future = std::move(returned_future),
+  auto lockfn = [&, this, device, returned_future = std::move(*returned_future),
                  timed]() mutable {
     TF_VLOG(5) << "ExecuteComputation acquiring PJRT device lock for "
                << device;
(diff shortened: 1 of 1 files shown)


JackCaoG:
I will let TPU CI to verify the result, test passed on my TPUVM after this change.

jonb377 (approved):

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -482,7 +482,8 @@ PjRtComputationClient::ExecuteComputation(
   CreateDataHandlesCounter()->AddValue(datas.size());
 
   auto mwait = std::make_shared<util::MultiWait>(1);
-  auto lockfn = [&, this, device, returned_future, timed]() mutable {
+  auto lockfn = [&, this, device, returned_future = std::move(*returned_future),
```
yeounoh, line 485: It would be good to assert that `returned_future.has_value()` before invoking this lambda function. Also do we need `mutable` here? Does it crash without?

yeounoh (approved):
Left some minor comments, approve to unblock the CI.

yeounoh (approved):
Left some minor comments, approve to unblock the CI.

third_party/xla_client/pjrt_computation_client.cc
JackCaoG: vaugley remember we need because of `returned_future, I can give it another test next week.

JackCaoG: hmm actually this is not going to be a silent failure, that's why we see the TPU CI failing.. Let me add the check anyway.

PR #5091, opened 2023-05-26
[SPMD] API for getting attributes for global devices

khatwanimohit:


torch_xla/experimental/pjrt.py
```diff
@@ -188,6 +188,9 @@ def process_count() -> int:
 def device_attributes(device: str) -> Dict[str, object]:
   return torch_xla._XLAC._xla_get_device_attributes(device)
 
+@requires_pjrt
+def global_device_attributes() -> List[Dict[str, object]]:
```
jonb377, line 193: Could you add a test for this in `test/pjrt/test_experimental_pjrt.py`? Just something simple and backend-agnostic, we need it to pass CI for TPU and CPU

will-cromar (commented):
+1 to adding an integration test. Otherwise LGTM

Does it make sense to add a pybind function to get the attributes for one device? ie a function that takes a device ordinal and returns the attributes for that ordinal? Or do you always want to get _all_ device attributes?

khatwanimohit:
> Does it make sense to add a pybind function to get the attributes for one device? ie a function that takes a device ordinal and returns the attributes for that ordinal? Or do you always want to get _all_ device attributes?
Then only use case for getting these attributes are for all global devices to create an efficient mesh.

will-cromar (approved):
Thanks!

alanwaketan:
@khatwanimohit Thanks, Mohit. Can you cc me on PRs related to multislice such that I can better keep track of the progress?

PR #5100, opened 2023-05-30
Add isvalid check for return futures

JackCaoG:
@yeounoh we still need mutable because we tried to reset the `timed` which is init on outside scope.


JackCaoG:
@yeounoh TPU CI also passed, can you take a look?

PR #5105, opened 2023-06-01
Revert "Disable cmake macros"

malfet:
Reverts pytorch/xla#5021 as it does not solve the underlying problem (of headers not being installed in `torch/include`)

As one can see in https://github.com/pytorch/pytorch/pull/102446 XLA can be built without this change


stgpetrovic (approved):
Thanks!

PR #5107, opened 2023-06-01
[CI] Pass CXX_ABI to cpp tests

malfet:
This should allow bazel to reuse majority of the build artifacts between
XLAC.so and cpp tests builds. Otherwise, different `cxx_opts` options
forces build system to start everything anew.


stgpetrovic (approved):
Nice. Thanks!

JackCaoG (approved):
Thanks! I will merge once CI is green

PR #5108, opened 2023-06-01
[test][OpenXLA] Find failed reason for bazel build failed

ManfeiBai:


ManfeiBai:
this PR is old and will continue work in https://github.com/pytorch/xla/pull/5202

PR #5109, opened 2023-06-01
Add StableHLO Dump Python API

lsy323:
In this PR, new Python APIs and debugging flag to dump StableHLO are added. The StableHLO is generated from HLO -> StableHLO conversion instead of emitted from PyTorch/XLA natively.
 
### StableHLO dump Python API
`xm.get_stablehlo(tensors : List[torch.tensor] = []) -> string`

Returns the StableHLO in string format. If `tensors` is not empty, dump the graph containing the tensors in the parameters. If `tensors` is empty, dump the whole graph.

In this API, no XLA compilation and graph execution is done. So a new workflow with early return is added to `XlaGraphExecutor` to support this use case.

### Enable StableHLO dump via Debugging flag
Currently, environmental variable `XLA_SAVE_TENSORS_FILE` and `XLA_SAVE_TENSORS_FMT=hlo` are set to dump HLO for debugging before the XLA compilation. Add support to use `XLA_SAVE_TENSORS_FMT=stablehlo` to dump StableHLO in the existing debugging workflow.

### Implementation
1. Add helper function (`hlo_to_stablehlo_str`) to do HLO->StableHLO conversion and dump StableHLO in string format.
2. Add a new parameter to extend `DumpUtil::ToHlo` for StableHLO dumping.
5. Add a new workflow in `XlaGraphExecutor` to return early (Before XLA compilation and execution) and dump StableHLO

### Test
An example script to use the StableHLO API is added to `test/test_stablehlo_dump.py`

### DIFF
--- a/test/test_stablehlo_dump.py
+++ b/test/test_stablehlo_dump.py
@@ -1,27 +1,34 @@
 import torch_xla
 import torch_xla.core.xla_model as xm
 import torch
-'''
-The following MLIR module should be dump in this test
-module @IrToHlo.9 attributes {mhlo.cross_program_prefetches = [], mhlo.dynamic_parameter_bindings = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
-  func.func @main(%arg0: tensor<1xi64>, %arg1: tensor<1xi64>) -> tuple<tensor<1xi64>, tensor<1xi64>, tensor<1xi64>> {
-    %0 = stablehlo.constant dense<1> : tensor<i64>
-    %1 = stablehlo.constant dense<1> : tensor<1xi64>
-    %2 = stablehlo.multiply %arg1, %1 : tensor<1xi64>
-    %3 = stablehlo.add %arg0, %2 : tensor<1xi64>
-    %4 = stablehlo.tuple %arg0, %arg1, %3 {xla_shape = "(s64[1]{0}, s64[1]{0}, s64[1]{0})"} : tuple<tensor<1xi64>, tensor<1xi64>, tensor<1xi64>>
-    return %4 : tuple<tensor<1xi64>, tensor<1xi64>, tensor<1xi64>>
-  }
-}
-'''
+import torchvision
+import unittest
 
-x = torch.tensor([3], device=xm.xla_device())
-y = torch.tensor([3], device=xm.xla_device())
-z = x + y
 
-# Example usage of dumping StableHLO given output tensors
-stablehlo = xm.get_stablehlo([z])
-# print(stablehlo)
-# Example usage of dump StableHLO of the entire graph
-stablehlo = xm.get_stablehlo()
-# print(stablehlo)
+class StableHloDumpTest(unittest.TestCase):
+
+  def test_simple(self):
+    device = xm.xla_device()
+    x = torch.tensor([3], device=device)
+    y = torch.tensor([3], device=device)
+    z = x + y
+    # Example usage of dumping StableHLO given output tensors
+    stablehlo = xm.get_stablehlo([z])
+    self.assertEqual(stablehlo.count("stablehlo.multiply"), 1)
+    self.assertEqual(stablehlo.count("stablehlo.add"), 1)
+
+  def test_resnet18(self):
+    device = xm.xla_device()
+    xla_resnet18 = torchvision.models.resnet18()
+    xla_resnet18.eval()
+    xla_resnet18 = xla_resnet18.to(device)
+    data = torch.randn(4, 3, 224, 224, device=device)
+    output = xla_resnet18(data)
+    stablehlo = xm.get_stablehlo()
+    self.assertEqual(stablehlo.count("convolution"), 20)
+    print(stablehlo)
+
+
+if __name__ == '__main__':
+  test = unittest.main()
+  sys.exit(0 if test.result.wasSuccessful() else 1)
--- a/test/test_stablehlo_dump.py
+++ b/test/test_stablehlo_dump.py
@@ -1,7 +1,6 @@
 import torch_xla
 import torch_xla.core.xla_model as xm
 import torch
-
 '''
 The following MLIR module should be dump in this test
 module @IrToHlo.9 attributes {mhlo.cross_program_prefetches = [], mhlo.dynamic_parameter_bindings = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
--- a/test/test_stablehlo_dump.py
+++ b/test/test_stablehlo_dump.py
@@ -1,0 +1,28 @@
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch
+
+'''
+The following MLIR module should be dump in this test
+module @IrToHlo.9 attributes {mhlo.cross_program_prefetches = [], mhlo.dynamic_parameter_bindings = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
+  func.func @main(%arg0: tensor<1xi64>, %arg1: tensor<1xi64>) -> tuple<tensor<1xi64>, tensor<1xi64>, tensor<1xi64>> {
+    %0 = stablehlo.constant dense<1> : tensor<i64>
+    %1 = stablehlo.constant dense<1> : tensor<1xi64>
+    %2 = stablehlo.multiply %arg1, %1 : tensor<1xi64>
+    %3 = stablehlo.add %arg0, %2 : tensor<1xi64>
+    %4 = stablehlo.tuple %arg0, %arg1, %3 {xla_shape = "(s64[1]{0}, s64[1]{0}, s64[1]{0})"} : tuple<tensor<1xi64>, tensor<1xi64>, tensor<1xi64>>
+    return %4 : tuple<tensor<1xi64>, tensor<1xi64>, tensor<1xi64>>
+  }
+}
+'''
+
+x = torch.tensor([3], device=xm.xla_device())
+y = torch.tensor([3], device=xm.xla_device())
+z = x + y
+
+# Example usage of dumping StableHLO given output tensors
+stablehlo = xm.get_xla_tensors_stablehlostep([z])
+print(stablehlo)
+# Example usage of dump StableHLO of the entire graph
+stablehlo = xm.xla_get_stablehlo()
+print(stablehlo)
--- a/test/test_stablehlo_dump.py
+++ b/test/test_stablehlo_dump.py
@@ -26,7 +26,6 @@ class StableHloDumpTest(unittest.TestCase):
     output = xla_resnet18(data)
     stablehlo = xm.get_stablehlo()
     self.assertEqual(stablehlo.count("convolution"), 20)
-    print(stablehlo)
 
 
 if __name__ == '__main__':
--- a/test/test_stablehlo_dump.py
+++ b/test/test_stablehlo_dump.py
@@ -25,7 +25,9 @@ class StableHloDumpTest(unittest.TestCase):
     data = torch.randn(4, 3, 224, 224, device=device)
     output = xla_resnet18(data)
     stablehlo = xm.get_stablehlo([output])
-    self.assertEqual(stablehlo.count("convolution"), 20)
+    self.assertEqual(stablehlo.count("stablehlo.convolution"), 20)
+    stablehlo = xm.get_stablehlo()
+    self.assertEqual(stablehlo.count("stablehlo.convolution"), 20)
 
 
 if __name__ == '__main__':
--- a/test/test_stablehlo_dump.py
+++ b/test/test_stablehlo_dump.py
@@ -24,7 +24,7 @@ class StableHloDumpTest(unittest.TestCase):
     xla_resnet18 = xla_resnet18.to(device)
     data = torch.randn(4, 3, 224, 224, device=device)
     output = xla_resnet18(data)
-    stablehlo = xm.get_stablehlo()
+    stablehlo = xm.get_stablehlo([output])
     self.assertEqual(stablehlo.count("convolution"), 20)
 
 
--- a/test/test_stablehlo_dump.py
+++ b/test/test_stablehlo_dump.py
@@ -20,8 +20,8 @@ y = torch.tensor([3], device=xm.xla_device())
 z = x + y
 
 # Example usage of dumping StableHLO given output tensors
-stablehlo = xm.get_xla_tensors_stablehlostep([z])
-print(stablehlo)
+stablehlo = xm.xla_get_stablehlo([z])
+# print(stablehlo)
 # Example usage of dump StableHLO of the entire graph
 stablehlo = xm.xla_get_stablehlo()
-print(stablehlo)
+# print(stablehlo)
--- a/test/test_stablehlo_dump.py
+++ b/test/test_stablehlo_dump.py
@@ -20,8 +20,8 @@ y = torch.tensor([3], device=xm.xla_device())
 z = x + y
 
 # Example usage of dumping StableHLO given output tensors
-stablehlo = xm.xla_get_stablehlo([z])
+stablehlo = xm.get_stablehlo([z])
 # print(stablehlo)
 # Example usage of dump StableHLO of the entire graph
-stablehlo = xm.xla_get_stablehlo()
+stablehlo = xm.get_stablehlo()
 # print(stablehlo)
--- a/third_party/xla_client/BUILD
+++ b/third_party/xla_client/BUILD
@@ -347,6 +347,19 @@ cc_library(
 )
 
 cc_library(
+    name = "stablehlo_helper",
+    srcs = ["stablehlo_helper.cc"],
+    hdrs = ["stablehlo_helper.h"],
+    deps = [
+        ":types",
+        ":xla_util",
+        "@org_tensorflow//tensorflow/compiler/xla/translate/hlo_to_mhlo:hlo_to_mlir_hlo",
+        "@org_tensorflow//tensorflow/compiler/xla/translate/mhlo_to_hlo:mlir_hlo_to_hlo",
+        "@org_tensorflow//tensorflow/compiler/xla/mlir_hlo:all_passes",
+    ],
+)
+
+cc_library(
     name = "sys_util",
     srcs = ["sys_util.cc"],
     hdrs = ["sys_util.h"],
--- a/third_party/xla_client/stablehlo_helper.cc
+++ b/third_party/xla_client/stablehlo_helper.cc
@@ -25,39 +25,47 @@ static std::string getMlirModuleStr(mlir::ModuleOp& mlir_module) {
 }
 
 static absl::Status hloToMhloHelper(const HloModuleProto* proto,
-                            mlir::ModuleOp* mlir_module) {
+                                    mlir::ModuleOp* mlir_module) {
   auto status = ConvertHloToMlirHlo(*mlir_module, proto,
                                     /*import_all_computations=*/false);
   if (!status.ok()) {
     return status;
   }
   if (!mlir::verify(*mlir_module).succeeded()) {
-    return absl::Status(absl::StatusCode::kInternal, "MHLO Module from HLO -> MHLO conversion is not legal.");
+    return absl::Status(
+        absl::StatusCode::kInternal,
+        "MHLO Module from HLO -> MHLO conversion is not legal.");
   }
   return absl::OkStatus();
 }
 
 static absl::Status mhloToStablehloHelper(mlir::ModuleOp* mlir_module,
-                                  mlir::MLIRContext* context) {
+                                          mlir::MLIRContext* context) {
   mlir::PassManager pm(context);
   pm.addPass(mlir::mhlo::createHloLegalizeToStablehloPass());
   if (!mlir::succeeded(pm.run(*mlir_module))) {
-    return absl::Status(absl::StatusCode::kInternal, "StableHLO Module from MHLO -> StableHLO conversion is not leagal.");
+    return absl::Status(
+        absl::StatusCode::kInternal,
+        "StableHLO Module from MHLO -> StableHLO conversion is not leagal.");
   }
-  return absl::OkStatus();;
+  return absl::OkStatus();
+  ;
 }
 
 std::string hloToStablehloStr(const HloModuleProto* proto) {
   mlir::MLIRContext context;
   mlir::ModuleOp mlir_module =
       mlir::ModuleOp::create(mlir::UnknownLoc::get(&context));
-  static const std::string err_msg = "Please open a github issue to PyTorch/XLA.\nOriginal HLO dump:\n";
+  static const std::string err_msg =
+      "Please open a github issue to PyTorch/XLA.\nOriginal HLO dump:\n";
   auto status = hloToMhloHelper(proto, &mlir_module);
-  XLA_CHECK(status.ok()) <<
-    "HLO -> MHLO conversion failed.\n" << status.message() << err_msg << getHloModuleStr(proto);
+  XLA_CHECK(status.ok()) << "HLO -> MHLO conversion failed.\n"
+                         << status.message() << err_msg
+                         << getHloModuleStr(proto);
   status = mhloToStablehloHelper(&mlir_module, &context);
-  XLA_CHECK(status.ok()) <<
-    "MHLO -> StableHLO conversion failed.\n" << status.message() << err_msg << getHloModuleStr(proto);
+  XLA_CHECK(status.ok()) << "MHLO -> StableHLO conversion failed.\n"
+                         << status.message() << err_msg
+                         << getHloModuleStr(proto);
   return getMlirModuleStr(mlir_module);
 }
 
--- a/third_party/xla_client/stablehlo_helper.cc
+++ b/third_party/xla_client/stablehlo_helper.cc
@@ -1,0 +1,44 @@
+#include "third_party/xla_client/stablehlo_helper.h"
+
+#include <iostream>
+
+#include "mlir/IR/Verifier.h"       // from @llvm-project
+#include "mlir/Pass/PassManager.h"  // from @llvm-project
+#include "tensorflow/compiler/xla/mlir_hlo/mhlo/transforms/passes.h"
+#include "tensorflow/compiler/xla/translate/hlo_to_mhlo/hlo_to_mlir_hlo.h"
+#include "tensorflow/compiler/xla/translate/mhlo_to_hlo/mlir_hlo_to_hlo.h"
+#include "third_party/xla_client/debug_macros.h"
+#include "third_party/xla_client/xla_util.h"
+
+namespace xla {
+
+static void hlo_mhlo_helper(const HloModuleProto* proto,
+                            mlir::ModuleOp* mlir_module) {
+  auto status = ConvertHloToMlirHlo(*mlir_module, proto,
+                                    /*import_all_computations*/ false);
+  XLA_CHECK(status.ok()) << "error in HLO -> MHLO conversion.";
+  XLA_CHECK(mlir::verify(*mlir_module).succeeded())
+      << "mhlo from hlo2mhlo verify not ok.";
+}
+
+static void mhlo_stablehlo_helper(mlir::ModuleOp* mlir_module,
+                                  mlir::MLIRContext* context) {
+  mlir::PassManager pm(context);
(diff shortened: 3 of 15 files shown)


third_party/xla_client/BUILD
```diff
@@ -56,102 +56,163 @@ cc_test(
 )
 
 cc_library(
-    name = "computation_client",
```
JackCaoG: can you rebase? I think these are from different prs.

test/test_stablehlo_dump.py
```diff
@@ -0,0 +1,27 @@
...
+x = torch.tensor([3], device=xm.xla_device())
+y = torch.tensor([3], device=xm.xla_device())
+z = x + y
+
+# Example usage of dumping StableHLO given output tensors
+stablehlo = xm.get_xla_tensors_stablehlostep([z])
+print(stablehlo)
+# Example usage of dump StableHLO of the entire graph
+stablehlo = xm.xla_get_stablehlo()
+print(stablehlo)
```
JackCaoG, line 27: remove the print maybe?

test/test_stablehlo_dump.py
```diff
@@ -0,0 +1,27 @@
...
+
+x = torch.tensor([3], device=xm.xla_device())
+y = torch.tensor([3], device=xm.xla_device())
+z = x + y
+
+# Example usage of dumping StableHLO given output tensors
+stablehlo = xm.get_xla_tensors_stablehlostep([z])
+print(stablehlo)
+# Example usage of dump StableHLO of the entire graph
+stablehlo = xm.xla_get_stablehlo()
```
JackCaoG, line 26: Maybe we don't need 2 api, we just need one api that takes optional arguments. If no argument tensor is passed, we dump the whole graph.

third_party/xla_client/stablehlo_helper.cc
```diff
@@ -0,0 +1,44 @@
...
+#include "mlir/Pass/PassManager.h"  // from @llvm-project
+#include "tensorflow/compiler/xla/mlir_hlo/mhlo/transforms/passes.h"
+#include "tensorflow/compiler/xla/translate/hlo_to_mhlo/hlo_to_mlir_hlo.h"
+#include "tensorflow/compiler/xla/translate/mhlo_to_hlo/mlir_hlo_to_hlo.h"
+#include "third_party/xla_client/debug_macros.h"
+#include "third_party/xla_client/xla_util.h"
+
+namespace xla {
+
+static void hlo_mhlo_helper(const HloModuleProto* proto,
```
JackCaoG, line 15: nit, maybe `hlo_to_mhlo_helper`

third_party/xla_client/stablehlo_helper.cc
```diff
@@ -0,0 +1,44 @@
...
+                            mlir::ModuleOp* mlir_module) {
+  auto status = ConvertHloToMlirHlo(*mlir_module, proto,
+                                    /*import_all_computations*/ false);
```
JackCaoG, line 18: nit `/*import_all_computations=*/` to match with rest of the code base.

third_party/xla_client/stablehlo_helper.cc
```diff
@@ -0,0 +1,44 @@
...
+  XLA_CHECK(status.ok()) << "HLO -> MHLO conversion failed.";
+  XLA_CHECK(mlir::verify(*mlir_module).succeeded())
+      << "MHLO from HLO->MHLO is not leagal.";
```
JackCaoG, line 21: maybe add a line `Please open a github issue to PyTorch/XLA`, we should also dump the HLO graph that is illegal.

third_party/xla_client/stablehlo_helper.cc
```diff
@@ -0,0 +1,44 @@
...
+}
+
+static void mhlo_stablehlo_helper(mlir::ModuleOp* mlir_module,
```
JackCaoG, line 24: nit, `mhlo_to_stablehlo_helper `

torch_xla/core/xla_model.py
```diff
@@ -840,6 +840,15 @@ def mark_step(wait=False):
   torch_xla._XLAC._set_all_reduce_token(devctx.device, None)
 
 
+def get_xla_tensors_stablehlostep(tensors):
```
JackCaoG, line 843: we need doc strings to describe the behavior and input types, return types.

test/test_stablehlo_dump.py
```diff
@@ -0,0 +1,27 @@
...
+    %3 = stablehlo.add %arg0, %2 : tensor<1xi64>
+    %4 = stablehlo.tuple %arg0, %arg1, %3 {xla_shape = "(s64[1]{0}, s64[1]{0}, s64[1]{0})"} : tuple<tensor<1xi64>, tensor<1xi64>, tensor<1xi64>>
+    return %4 : tuple<tensor<1xi64>, tensor<1xi64>, tensor<1xi64>>
+  }
+}
+'''
+
+x = torch.tensor([3], device=xm.xla_device())
+y = torch.tensor([3], device=xm.xla_device())
+z = x + y
```
JackCaoG, line 20: can we have a different test that dumps resnet18 StableHLO for example?

test/test_stablehlo_dump.py
lsy323: Yes, that's a good idea. Let's keep the `xla_get_stablehlo`

test/test_stablehlo_dump.py
lsy323: Yes, will remove it

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -345,6 +345,16 @@ void StepMarker(const std::string& device_str,
   }
 }
 
+std::string SyncLiveTensorGetHlo(const std::string& device_str,
```
JackCaoG, line 348: maybe `GetLiveTensorsStableHLO` and move it near to https://github.com/pytorch/xla/blob/master/torch_xla/csrc/init_python_bindings.cpp#LL364C13-L364C33

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1259,9 +1274,18 @@ void InitXlaModuleBindings(py::module m) {
         [](const std::string& device, const std::vector<std::string>& devices,
            bool wait) {
           NoGilSection nogil;
-          StepMarker(device, devices, wait);
+          // StepMarker(device, devices, wait);
+          std::cout << "in step marker";
+          std::cout << SyncLiveTensorGetHlo(device, devices);
```
JackCaoG, line 1279: remove debugging comments.

torch_xla/core/xla_model.py
lsy323: Will add doc string for the updated API

torch_xla/csrc/debug_util.cpp
```diff
@@ -52,6 +54,34 @@ DebugUtil::GraphFormat DebugUtil::GetDefaultGraphFormat() {
   return format;
 }
 
+std::string DebugUtil::GetTensorsGraphHlo(
+    absl::Span<const XLATensorPtr> tensors, const std::vector<size_t>* indices,
+    bool dump_stablehlo) {
+  std::vector<torch::lazy::Value> root_values;
+  xla::util::Unique<torch::lazy::BackendDevice> unique_device;
+  if (indices != nullptr) {
```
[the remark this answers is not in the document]
lsy323: That's a good idea, but an other local `vector<size_t>` is needed. If indices is nullptr, we need to fill the local vector with `iota` and have `indices` pointing to the local vector.

Actually this snippet is duplicated from line 92. Maybe create a helper function to get `root_nodes`, `root_hashes`,  `root_values` and `unique_device ` from tensor list and tensor indices

third_party/xla_client/stablehlo_helper.cc
lsy323: Will update

third_party/xla_client/stablehlo_helper.cc
lsy323: Yes, that' s a good idea, let me add it

third_party/xla_client/stablehlo_helper.cc
lsy323: Will update to match the code base style.

third_party/xla_client/stablehlo_helper.cc
lsy323: Will update

test/test_stablehlo_dump.py
miladm: good ideas here - at this stage, based on our discussions today, we need to test:

1- dumping stablehlo string (with optional tensor input)
2- converting to stablehlo and doing the actual lowering

I don't think this test addresses the second bullet

third_party/xla_client/stablehlo_helper.cc
miladm: another nit: shouldn't our cpp code be camelcase?

test/test_stablehlo_dump.py
lsy323: This PR focuses on StableHLO dump API - see description.

For 2, did you mean compiling & compilation with StableHLO?

third_party/xla_client/stablehlo_helper.cc
lsy323: Yes, good catch, just realized that and will update.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -345,6 +345,16 @@ void StepMarker(const std::string& device_str,
   }
 }
 
+std::string SyncLiveTensorGetHlo(const std::string& device_str,
+                                 const std::vector<std::string>& devices) {
+  tsl::profiler::TraceMe activity("SyncLiveTensorGetHlo",
```
miladm, line 350: also rename here.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -358,6 +359,23 @@ void XLAGraphExecutor::SyncTensorsGraph(std::vector<XLATensorPtr>* tensors,
...
+  tsl::profiler::TraceMe activity("SyncTensorsGraphDumpHlo",
+                                  tsl::profiler::TraceMeLevel::kInfo);
+  SyncTensorsConfig config;
+  config.sync_ltc_data = sync_ltc_data;
+  SyncTensorCollection coll = CollectSyncTensors(*tensors, config);
+  if (coll.indices.empty()) {
+    /* Enure previous execution is complete before exiting this
+     * function */
+    TensorCollectionBarrier(&coll);
+    return "";
```
miladm, line 374: doesn't make sense to return something more meaningful than empty string?

test/test_stablehlo_dump.py
miladm: on a small model, we should do a diff comparison of the hlo and stablehlo graphs to confirm they emit the same set of operations.

test/test_stablehlo_dump.py
burmako, line 26: What do you think about calling this new API: `xm.get_stablehlo` (rather than `xm.xla_get_stablehlo`)? StableHLO is not part of XLA - it's a framework- and compiler-agnostic opset.

third_party/xla_client/stablehlo_helper.h
```diff
@@ -0,0 +1,12 @@
+#ifndef XLA_MHLO_HELPER_H_
+#define XLA_MHLO_HELPER_H_
```
burmako, line 2: `XLA_CLIENT_STABLEHLO_HELPER_H_`?

torch_xla/core/xla_model.py
burmako: Perhaps these doc strings could also describe anticipated use cases?

test/test_stablehlo_dump.py
GleasonK, line 27: Should this do some sort of comparison to see if this test breaks?

If something strange happened that this API failed conversion and fell back to dumping an empty `module {}`, this would not be caught by this test.

Given that textual assembly is not stable, perhaps use some substring comparisons (`stablehlo.add`, `stablehlo.multiply`, any attributes of interest).

third_party/xla_client/stablehlo_helper.cc
GleasonK: Does the status message appear in logs anywhere as it is? Other wise consider including the `status.message()` as well.

third_party/xla_client/stablehlo_helper.cc
```diff
@@ -0,0 +1,44 @@
...
+
+std::string hlo_to_stablehlo_str(const HloModuleProto* proto) {
+  mlir::MLIRContext context;
+  mlir::ModuleOp mlir_module =
+      mlir::ModuleOp::create(mlir::UnknownLoc::get(&context));
+  hlo_mhlo_helper(proto, &mlir_module);
+  mhlo_stablehlo_helper(&mlir_module, &context);
+  std::string txt_mlir_module;
+  llvm::raw_string_ostream os{txt_mlir_module};
+  mlir_module.print(os);
```
GleasonK, line 40: LGTM. Future enhancement ideas:

1. If the goal is for this to be used anywhere in production (i.e. dump a thing we can compile later) consider adding an option to dump as bytecode or StableHLO portabl artifact.
2. If the HLO has any interesting debug information, it will be thrown away here since prettyprint (the default) does not include file/line/col info. If preserving is desired, can add [enableDebugInfo](https://github.com/openxla/xla/blob/98f817b8d377f86e897cfd50b708412a2233ded1/xla/service/dump.cc#L620), but might not be necessary since I'm not sure if pytorch/xla includes any debug info currently (could be an interesting experiment).

test/test_stablehlo_dump.py
GleasonK: Similar feedback for the internal `_xla_get_stablehlo` method naming.

GleasonK: Possibly related (maybe more related to 2), is there any way to run a test suite and either:
1. Dump StableHLO for each test
2. Convert to StableHLO and compile

An environment variable or otherwise? This will be useful as we add support for the remaining pieces needed by PyTorch/XLA in StableHLO, to make sure we have everything covered.

GleasonK (commented):
Overall LGTM! Left some minor feedback. Will be good to get this infrastructure in place. Evolutions of this API to cater to more specific user needs will be much easier with these methods in `stablehlo_helper` submitted, i.e. getting these dependencies and conversions in place is likely the hard part.

test/test_stablehlo_dump.py
lsy323: That's a good idea, will add the string comparisons in the test.

torch_xla/core/xla_model.py
lsy323: Good idea, anticipated use cases are added.

test/test_stablehlo_dump.py
lsy323: That's a good idea👍 

1. The environment variables to enable StableHLO dump is added in the PR. Details in the "Enable StableHLO dump via Debugging flag" section in the description. However, we don't have a workflow to validate the dump from the tests. Maybe adding a separate Python script that compares the dump with the pre-generated dump is a way to go.

2. This is not covered by this PR. Will include this workflow in the 2nd PR that changes the compilation workflow. This PR focuses on the StableHLO dump.

lsy323: This makes sense to me, will rename the API accordingly.

third_party/xla_client/stablehlo_helper.cc
lsy323: Yeah, that's a good idea to provide the error message to the user.

third_party/xla_client/stablehlo_helper.h
lsy323: Good catch!

torch_xla/csrc/xla_graph_executor.cpp
lsy323: My understanding is that when the tensors to be collected is empty, that means there are async execution is happening. For this case, the current synced graph should be empty, this aligns with the current handling logic in the `mark_step` workflow. So in this case we just return empty string to users. cc @JackCaoG for confirmation on my understanding and ideas on this logic.

lsy323:
Copying @GleasonK's idea on further enhancement of the dump API here, so they won't lost as the code review progresses. Thanks @GleasonK for the enhancement ideas 👍 

Future enhancement ideas:

- If the goal is for this to be used anywhere in production (i.e. dump a thing we can compile later) consider adding an option to dump as bytecode or StableHLO portabl artifact.
- If the HLO has any interesting debug information, it will be thrown away here since prettyprint (the default) does not include file/line/col info. If preserving is desired, can add [enableDebugInfo](https://github.com/openxla/xla/blob/98f817b8d377f86e897cfd50b708412a2233ded1/xla/service/dump.cc#L620), but might not be necessary since I'm not sure if pytorch/xla includes any debug info currently (could be an interesting experiment).

lsy323:
- Rename APIs according to review comments
- Add ResNet18 test cases
- Enhance error handling and error message
- Improve doc string

test/test_stablehlo_dump.py
```diff
@@ -0,0 +1,33 @@
...
+    self.assertEqual(stablehlo.count("stablehlo.add"), 1)
+
+  def test_resnet18(self):
+    device = xm.xla_device()
+    xla_resnet18 = torchvision.models.resnet18()
+    xla_resnet18.eval()
+    xla_resnet18 = xla_resnet18.to(device)
+    data = torch.randn(4, 3, 224, 224, device=device)
+    output = xla_resnet18(data)
+    stablehlo = xm.get_stablehlo([output])
```
JackCaoG, line 27: can we have a test for `xm.get_stablehlo` without inputs? You might need to do a `mark_step` during init to clear IRs from other tests.

torch_xla/csrc/debug_util.cpp
```diff
@@ -106,6 +136,10 @@ std::string DebugUtil::GetTensorsGraphInfo(
   } else if (format == GraphFormat::kHlo) {
     graph_str = DumpUtil::ToHlo(
         root_values, unique_device ? *unique_device : GetCurrentDevice());
+  } else if (format == GraphFormat::kStableHlo) {
+    graph_str = DumpUtil::ToHlo(
+        root_values, unique_device ? *unique_device : GetCurrentDevice(),
+        /* to_stablehlo */ true);
```
JackCaoG, line 142: nit /*to_stablehlo=*/

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -917,6 +928,10 @@ void InitXlaModuleBindings(py::module m) {
         [](const std::vector<at::Tensor>& tensors) -> std::string {
           return GetTensorsHloGraph(tensors);
         });
+  m.def("_get_xla_tensors_stablehlo",
+        [](const std::vector<at::Tensor>& tensors) -> std::string {
+          return GetTensorsHloGraph(tensors, true);
```
JackCaoG, line 933: nit `/*get_stable_hlo=*/true`

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -358,6 +359,23 @@ void XLAGraphExecutor::SyncTensorsGraph(std::vector<XLATensorPtr>* tensors,
   }
 }
 
+std::string XLAGraphExecutor::SyncTensorsGraphDumpHlo(
+    std::vector<XLATensorPtr>* tensors, absl::Span<const std::string> devices,
+    bool sync_ltc_data) {
+  tsl::profiler::TraceMe activity("SyncTensorsGraphDumpHlo",
+                                  tsl::profiler::TraceMeLevel::kInfo);
+  SyncTensorsConfig config;
+  config.sync_ltc_data = sync_ltc_data;
```
JackCaoG, line 368: you don't really need to pass `sync_ltc_data`, since you are not updating the IR.

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: hmm, shouldn't it be
```
  if (!coll.indices.empty()) {
    /* Enure previous execution is complete before exiting this
     * function */
    TensorCollectionBarrier(&coll);
  } else {
    return "";
  }
```
if indicies is empty, you don't need to dump anything so there is no point of waiting for the lock.

torch_xla/csrc/ir_dump_util.h
```diff
@@ -25,7 +25,8 @@ class DumpUtil {
       absl::Span<const torch::lazy::Node* const> roots);
 
   static std::string ToHlo(c10::ArrayRef<torch::lazy::Value> values,
-                           const torch::lazy::BackendDevice& device);
+                           const torch::lazy::BackendDevice& device,
+                           bool to_stable_hlo = false);
```
burmako, line 29: Nit: the canonical spelling for StableHLO in code is "stablehlo" or "Stablehlo" (without underscores or capitalization in the middle of the word).

burmako (approved):
I don't have much experience with PyTorch/XLA internals, so I didn't review that part of the PR, but the overall API, documentation and the code that uses MHLO/StableHLO infrastructure LGTM.

lsy323:
- Minor refactor
- Enhance Test Coverage

torch_xla/csrc/xla_graph_executor.cpp
lsy323: I was referring the logic at https://github.com/pytorch/xla/blob/83c487111978ac221df53840f9a7e91cd94335eb/torch_xla/csrc/xla_graph_executor.cpp#L1301. Is there anything wrong with the logic there?

third_party/xla_client/stablehlo_helper.cc
```diff
@@ -0,0 +1,72 @@
...
+static std::string getMlirModuleStr(mlir::ModuleOp& mlir_module) {
+  std::string txt_mlir_module;
+  llvm::raw_string_ostream os{txt_mlir_module};
+  mlir_module.print(os);
+  return txt_mlir_module;
+}
+
+static absl::Status hloToMhloHelper(const HloModuleProto* proto,
+                                    mlir::ModuleOp* mlir_module) {
+  auto status = ConvertHloToMlirHlo(*mlir_module, proto,
```
vanbasten23, line 29: I wonder where `ConvertHloToMlirHlo ` is defined.

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: hmm yea.. I think `coll.indices.empty()` means there is nothing for us to sync. I need another pair of eyes, @alanwaketan wdyt? I think this is a bug.

JackCaoG: actually you can delete logic here, if you are dumping HLO(and stableHLO too), you will run
https://github.com/pytorch/xla/blob/83c487111978ac221df53840f9a7e91cd94335eb/torch_xla/csrc/debug_util.cpp#L122-L130

`WaitDeviceOps` will call `barrier` I think, you don't need to do barrier here.

third_party/xla_client/stablehlo_helper.h
```diff
@@ -0,0 +1,12 @@
+#ifndef STABLEHLO_HELPER_H_
```
alanwaketan, line 1: Why do we add this to the xla_client instead of torch_xla?

torch_xla/csrc/xla_graph_executor.cpp
lsy323: @alanwaketan This is a workflow similar to `XLAGraphExecutor::SyncTensorsGraph`, but returns early without going through XLA compilation and graph execution. So I decided to put it in `XLAGraphExecutor`

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -391,6 +392,16 @@ std::string GetLiveTensorsReport(size_t nodes_threshold,
   return ss.str();
 }
 
+static std::string GetLiveTensorsStableHLO(
+    const std::string& device_str, const std::vector<std::string>& devices) {
+  tsl::profiler::TraceMe activity("GetLiveTensorsStableHLO",
+                                  tsl::profiler::TraceMeLevel::kInfo);
+  torch::lazy::BackendDevice device = GetDeviceOrCurrent(device_str);
+  auto tensors = XLAGraphExecutor::Get()->GetLiveTensors(&device);
+  return XLAGraphExecutor::Get()->SyncTensorsGraphDumpHlo(
```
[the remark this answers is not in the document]
lsy323: I want to make sure my understanding on "Live Tensors" and "Sync" is correct. Please correct if I'm wrong

"Live Tensors" refers to all tensors that are already on device, including both materialized and unmaterialized tensors.

"Sync" refers to materialize the IR Nodes that are not on the device yet. It involves XLA compilation and graph execution.

This API is built for the use case that user would like to get the StableHLO of the whole graph. It's like a `mark_step()` but doesn't compile and execute the graph.

Since `SyncTensorsGraphDumpHlo` introduces a new workflow that returns early before XLA compilation and graph execution. This expands the definition of "Sync" above from "materialize IR Nodes" to "Lower the IR Nodes"

third_party/xla_client/stablehlo_helper.h
lsy323: I found currently all TF dependencies used in PyTorch/XLA are wrapped in `libxla_computation_client.so`, and `_XLAC.so` links with `libxla_computation_client.so` for TF dependencies. The MLIR/StableHLO APIs can be considered TF dependencies now, so I put them in xla_client

torch_xla/csrc/xla_graph_executor.cpp
lsy323: @alanwaketan I see your point, it should be cleaner to call `XLAGraphExecutro::CollectSyncTensors` outside and then pass the tensors to dump util functions for lowering and StableHLO dump.

lsy323:
- Use 1 pybind API
- Remove syncing previous execution logic

torch_xla/csrc/init_python_bindings.cpp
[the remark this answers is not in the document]
alanwaketan: "Live Tensors" means all xla tensor that are currently live in the Python space. It doesn't matter whether they are device data or IR or aten tensors. I think that's all you need and you don't need to sync them. Can you try just passing the live tensors to the hlodump and see if that gives you what you wants?

JackCaoG:
ehh this `error: command '/usr/local/bin/bazel' failed with exit code 37` is annoying. @lsy323 can you rebase and verify that build is passing locally.

lsy323:
> 

Hi @JackCaoG, I rebased to the HEAD yesterday and it still fails. The build passed locally.

third_party/xla_client/stablehlo_helper.h
```diff
@@ -0,0 +1,12 @@
...
+#define STABLEHLO_HELPER_H_
```
vanbasten23, line 2: Curious, do you know why we need to define this macro `STABLEHLO_HELPER_H_ ` and how we use it?

torch_xla/csrc/xla_graph_executor.h
```diff
@@ -121,6 +122,12 @@ class XLAGraphExecutor : public torch::lazy::LazyGraphExecutor {
                         absl::Span<const std::string> devices, bool wait,
                         bool sync_ltc_data, bool warm_up_cache_only = false);
 
+  // An early-return version of SyncTensorsGraph, the function returns after the
+  // tensors are synced. The graph won't be compiled and executed. This workflow
+  // is created for HLO/StableHLO dump.
+  std::string SyncTensorsGraphDumpHlo(std::vector<XLATensorPtr>* tensors,
```
alanwaketan, line 128: I guess now you can just remove this.

third_party/xla_client/stablehlo_helper.h
lsy323: Similar to `#pragma once`, please see https://en.wikipedia.org/wiki/Include_guard

torch_xla/csrc/xla_graph_executor.h
lsy323: Yes, thanks for catching it!

JackCaoG:
Mirror of this pr https://github.com/pytorch/xla/pull/5150 has merged, I am going to close this one.

PR #5047, opened 2023-05-22
Add load_local_shards to XLAShardedTensor

jonb377:
This change enables loading a list of local `XLAShard` on CPU into a sharded tensor's device data. To do this, the function `ShardingUtil::CreateShardedData` is added which transfers a list of local CPU shards to the devices and wraps the resulting handles in PjRtShardedData.

A few notes:
- Only non-replicated sharding is allowed to ensure the underlying shards remain consistent with their sharding (e.g. to prevent passing different tensors as shards to a replicated tensor).
- `PopulateTensorBuffer` is exposed from `tensor_util` for use in `ShardingUtil::CreateShardedData`.
- Each `XLAShard` will now also track which device it originated from.

PR #5110, opened 2023-06-01
[SPMD] index local shards correctly in replication

yeounoh:
Replication across multi-host devices was buggy due to incorrect index mapping. Tested locally with 2x v4-8 slice.

### DIFF
--- a/third_party/xla_client/pjrt_computation_client.cc
+++ b/third_party/xla_client/pjrt_computation_client.cc
@@ -53,6 +53,21 @@ MaybeInitializeDistributedRuntimeClient(int local_rank,
   return std::move(client);
 }
 
+// Builds a map from the device's global ordinal to its index in the `devices`
+// array.
+static std::unordered_map<int, int> build_index_map(
+    const std::vector<std::string>& devices) {
+  std::unordered_map<int, int> device_index;
+  for (int i = 0; i < devices.size(); ++i) {
+    std::vector<std::string> device_spec = absl::StrSplit(devices[i], ':');
+    XLA_CHECK_EQ(device_spec.size(), 2)
+        << "Invalid device specification: " << devices[i];
+    int global_ordinal = std::stoi(device_spec[1]);
+    device_index[global_ordinal] = i;
+  }
+  return device_index;
+}
+
 }  // namespace
 
 std::string PjRtComputationClient::PjRtDeviceToString(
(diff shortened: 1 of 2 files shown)


third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -315,17 +315,25 @@ ComputationClient::DataPtr PjRtComputationClient::ReplicateShardedData(
     XLA_CHECK_EQ(shards.size(), GetLocalDevices().size());
     std::vector<std::vector<ComputationClient::DataPtr>> arguments_by_device(
         GetLocalDevices().size(), std::vector<ComputationClient::DataPtr>(1));
-    for (auto shard : shards) {
+    for (int i = 0; i < shards.size(); ++i) {
```
jonb377, line 318: Can we reuse ShardingUtil::InputHandler here?

yeounoh: +1 let's do that

yeounoh: ah, forgot about the circular dependencies issue. We should try avoid using the xla sharding util in the runtime client.

third_party/xla_client/pjrt_computation_client.cc
```diff
@@ -315,17 +315,25 @@ ComputationClient::DataPtr PjRtComputationClient::ReplicateShardedData(
...
+      auto shard = shards[i];
       std::vector<std::string> device_spec =
           absl::StrSplit(shard->device(), ':');
       XLA_CHECK_EQ(device_spec.size(), 2)
           << "Invalid device specification: " << shard->device();
-      int device_i = std::stoi(device_spec[1]);
-      arguments_by_device[device_i][0] = shard;
+      TF_VLOG(3) << shard->device() << " is mapped to local device index " << i;
+      arguments_by_device[i][0] = shard;
```
jonb377, line 325: Using `i` directly here requires that the order of the shards within PjRtShardedData corresponds to the order of devices returned from `GetLocalDevices` - is that a safe assumption? In the `InputHandler`, we explicitly [remap](https://github.com/pytorch/xla/blob/09d6775d4c42ce1d0582ef5e797e8bd741166ffb/torch_xla/csrc/xla_sharding_util.cpp#L226) the shard index to the device index.

In my `load_local_shards` [PR](https://github.com/pytorch/xla/pull/5047/files#diff-12cff199d1701b121499c962ef4dabbe2bba7b134cae3f6b54d8082a52a1c85aR1659), the order within PjRtShardedData will depend on the order specified by the user, but other than that I don't think we ever change the order - we just don't have any checks that the order matches.

jonb377 (approved):
LGTM, thanks Yeounoh!

third_party/xla_client/pjrt_computation_client.cc
yeounoh: Yea, we should try to explicitly remap.

PR #5114, opened 2023-06-02
Add option to remove XRT from the build

will-cromar:
Set `PJRT_ONLY=1` in `setup.py` (or `--config=pjrt_only` in bazel) to exclude XRT from the build process.

- Add `--config=pjrt_only` to bazel build, which excludes XRT from the final binary. This lowers the binary size significantly.
- Attempt to clean up dependency structure by hand. We'll need better tooling to do a complete cleanup.
- After hitting repeated linker and undefined symbol errors related to the profiler, consolidate all direct dependencies on the profiler under the `:profiler` target. Backport https://github.com/tensorflow/tensorflow/commit/0a2945f73f3839d7a19cb95263fd9f42a7fee6e2 (moves trace to TSL) to make this easier. We can remove the patch in the next pin update. XRT is obscuring these implicit dependencies (mainly proto implementations) by including so much of TF.


setup.py
```diff
@@ -40,6 +40,9 @@
 #   TPUVM_MODE=0
 #     whether to build for TPU
 #
+#   PJRT_ONLY=1
+#     whether to build for TPU
```
JackCaoG, line 44: typo, PJRT instead of TPU

tf_patches/profiler_trace.diff
```diff
@@ -0,0 +1,611 @@
+diff --git a/tensorflow/python/profiler/internal/BUILD b/tensorflow/python/profiler/internal/BUILD
```
JackCaoG, line 1: lol what's this diff about?

torch_xla/csrc/BUILD
```diff
@@ -275,6 +275,7 @@ ptxla_cc_library(
         "@com_google_absl//absl/container:flat_hash_map",
         "@com_google_absl//absl/strings",
         "@com_google_absl//absl/types:variant",
+        "@org_tensorflow//tensorflow/tsl/profiler/lib:traceme",
```
JackCaoG, line 278: nice.

tf_patches/profiler_trace.diff
will-cromar: Added comment to this file and PR description.

JackCaoG:
seems to have some issue with profiler
```

2023-06-05T17:39:52.9263584Z ======================================================================
2023-06-05T17:39:52.9264028Z ERROR: test_trace_and_metrics (__main__.ProfilerTest)
2023-06-05T17:39:52.9264614Z ----------------------------------------------------------------------
2023-06-05T17:39:52.9265737Z Traceback (most recent call last):
2023-06-05T17:39:52.9266289Z   File "/tmp/pytorch/xla/test/test_profiler.py", line 77, in test_trace_and_metrics
2023-06-05T17:39:52.9266676Z     xp.trace(
2023-06-05T17:39:52.9267486Z   File "/opt/conda/lib/python3.8/site-packages/torch_xla-2.1.0-py3.8-linux-x86_64.egg/torch_xla/debug/profiler.py", line 82, in trace
2023-06-05T17:39:52.9268193Z     torch_xla._XLAC.profiler.trace(
2023-06-05T17:39:52.9268722Z RuntimeError: No trace event was collected because there were no responses from clients or the responses did not have trace data.
```

will-cromar:
The latest CPU CI run just got stuck on `test_mp_decorator` until it timed out:

```
2023-06-06T19:35:01.1032801Z + python3 /tmp/pytorch/xla/test/test_operations.py --verbosity=2
2023-06-06T19:35:02.8676315Z /home/jenkins/.local/lib/python3.8/site-packages/torchvision/io/image.py:13: UserWarning: Failed to load image Python extension: ''If you don't plan on using image functionality from `torchvision.io`, you can ignore this warning. Otherwise, there might be something wrong with your environment. Did you have `libjpeg` or `libpng` installed before building `torchvision` from source?
2023-06-06T19:35:02.8677189Z   warn(
2023-06-06T19:35:02.8678084Z /opt/conda/lib/python3.8/site-packages/torch/__init__.py:615: UserWarning: torch.set_default_tensor_type() is deprecated as of PyTorch 2.1, please use torch.set_default_dtype() and torch.set_default_device() as alternatives. (Triggered internally at /tmp/pytorch/torch/csrc/tensor/python_tensor.cpp:453.)
2023-06-06T19:35:02.8678774Z   _C._set_default_tensor_type(t)
2023-06-06T19:35:02.8683339Z test_mp_decorator (__main__.MpDecoratorTest) ... ERROR
2023-06-06T20:28:49.3865509Z ##[error]The operation was canceled.
```

Locally (and even in previous CI runs), I can't reproduce the issue

will-cromar:
The TPU CI test hits a SIGSEGV within one of the all reduce tests:

```
Step #4 - "run_e2e_tests": [ RUN      ] TestTpuCollectiveOps.test_all_gather_pinned
Step #4 - "run_e2e_tests": https://symbolize.stripped_domain/r/?trace=7f2d297b663b,7f2e65fa272f,7f2d298c05f5,7f2d296aab3b,7f2d2962794b,7f2d29628f24,7f2d29628fd5,7f2e58222027,1a,3a6c616e6f697462&map=7300d17483f73bf84c4699d2a61d8df18d5c2d8c:7f2d234ff000-7f2d3517c560 
Step #4 - "run_e2e_tests": *** SIGSEGV (@(nil)), see gl__________57#s15 received by PID 8353 (TID 10249) on cpu 84; stack trace: ***
Step #4 - "run_e2e_tests": PC: @     0x7f2d297b663b  (unknown)  std::vector<>::_M_realloc_insert<>()
Step #4 - "run_e2e_tests":     @     0x7f2d229426ba       1152  (unknown)
Step #4 - "run_e2e_tests":     @     0x7f2e65fa2730       3408  (unknown)
Step #4 - "run_e2e_tests":     @     0x7f2d298c05f6        272  torch_xla::AllReduce::AllReduce()
```

This is a bad sign, but I don't think any of the changes I made since the last time this passed are material here. None of the other collective tests failed.

I'll wait and see how the GPU test turns out. If it actually runs, I'll roll back the changes that enable `PJRT_ONLY=1` in the CI and move that change to a follow-up PR.

will-cromar:
On closer inspection, both runs are hanging on the XRT-only tests:

```
Running in OpByOp mode: /tmp/pytorch/xla/test/test_operations.py --verbosity=2
+ XLA_GET_TENSORS_OPBYOP=1
+ XLA_SYNC_TENSORS_OPBYOP=1
+ run_xrt /tmp/pytorch/xla/test/test_operations.py --verbosity=2
++ command -v nvidia-smi
+ '[' -x /usr/bin/nvidia-smi ']'
+ '[' 1 '!=' 0 ']'
+ GPU_NUM_DEVICES=2
+ run_coverage /tmp/pytorch/xla/test/test_operations.py --verbosity=2
+ '[' 0 '!=' 0 ']'
+ python3 /tmp/pytorch/xla/test/test_operations.py --verbosity=2
/home/jenkins/.local/lib/python3.8/site-packages/torchvision/io/image.py:13: UserWarning: Failed to load image Python extension: ''If you don't plan on using image functionality from `torchvision.io`, you can ignore this warning. Otherwise, there might be something wrong with your environment. Did you have `libjpeg` or `libpng` installed before building `torchvision` from source?
  warn(
/opt/conda/lib/python3.8/site-packages/torch/__init__.py:615: UserWarning: torch.set_default_tensor_type() is deprecated as of PyTorch 2.1, please use torch.set_default_dtype() and torch.set_default_device() as alternatives. (Triggered internally at /tmp/pytorch/torch/csrc/tensor/python_tensor.cpp:453.)
  _C._set_default_tensor_type(t)
test_mp_decorator (__main__.MpDecoratorTest) ... ERROR
```

I'm going to call this PR a success since the other tests passed. I'll revert my change to the CI and build config. Follow-up

1. Decide whether to completely disable XRT in the CI
2. Complain loudly instead of hanging in op-by-op execution

JackCaoG:
Let's keep at least one XRT test until we delete it from our code base.

setup.py
```diff
@@ -40,6 +40,9 @@
 #   TPUVM_MODE=0
 #     whether to build for TPU
 #
+#   PJRT_ONLY=1
```
JackCaoG, line 43: are we going to default PJRT_ONLY=1 on head? I am afraid we will miss the XRT build failures/bugs after TF updates. I felt like setting this flag almost is equivalent to delete all XRT from our code base.

BUILD
```diff
@@ -17,37 +17,7 @@ tf_cc_shared_object(
     ],
     visibility = ["//visibility:public"],
     deps = [
-        "//third_party/xla_client:runtime",
```
JackCaoG, line 20: curious why can we delete these xla_client dependency, are we getting it from somewhere else or these are not needed for the `XLAC.so`?

setup.py
will-cromar: Nope, this is a typo that I reverted by accident.

BUILD
will-cromar: We needed them before migrating `csrc` to bazel, since code in there depended on `libcomputation_client.so`. Now that everything is in bazel, bazel willnclude all of the transitive deps of `init_python_bindings` on its own.

will-cromar:
> Let's keep at least one XRT test until we delete it from our code base.

We want to run all of our PJRT tests after building with `PJRT_ONLY=1` to ensure we're not taking any new implicit dependencies on TF through XRT. But, of course, we can't run the XRT tests without XRT.

I think we have to add another build/test config to GHA:

1. An extra suite of tests with `PJRT_ONLY=1`
2. An extra suite of tests with XRT

What if we added some CPU-only tests with XRT as a sanity-check?

JackCaoG (approved):
This is exciting! Thanks @will-cromar

JackCaoG:
yea cpu only XRT test sgtm, we only need to test a subset of tests, maybe `test_operations` with normal and opbyop.

will-cromar:
Passing TPU CI run from last commit: https://github.com/pytorch/xla/runs/14055560855

PR #5117, opened 2023-06-05
[OpenXLA] Delete unused `tensor_shape.h` dep

ManfeiBai:
Before migrate PyTorch/XLA to pull XLA from OpenXLA, PyTorch/XLA want to delete some unused TensorFlow deps for clean


ManfeiBai:
@ddunl, please help to review this PR, thanks!

ddunl:
I'm suspicious that this PR is correct - I suspect this passes only because we get this header transitively via `conv_op_helpers`

ManfeiBai:
> I'm suspicious that this PR is correct - I suspect this passes only because we get this header transitively via `conv_op_helpers`

will continue this discuss in task `conv_op_helpers`

PR #5120, opened 2023-06-06
Restrict the mark_sharding api to devicedata

JackCaoG:
This pr restrict the `mark_sharding` to device data. 

Through on a second thought this will disable the usage like
```
t1 = torch.randn(100, device='xla:0')
t1.mark_sharding()...
```

since randn will produce a pending computation.


JackCaoG:
I will drop this pr due to the randn issue for now,

PR #5128, opened 2023-06-07
Add utility functions for distributed checkpointing

jonb377:
This change adds a few utility functions to support distributed checkpointing. The following changes are included:
- Add `sharding_type` to `XLAShardedTensor` to get the ShardingType
- Add `wrap_if_sharded` to convert `torch.Tensor` into `XLAShardedTensor` if the underlying data is sharded.
- Remove `devices` parameter from `_get_local_shard_indices` and instead always return the shard indices in the order of the shards
- Clamp the `start` bound of the index slices to the tensor's size.
- Add a setter to the `unpadded_data` property of `XLAShard`

### DIFF
--- a/torch_xla/csrc/xla_sharding_util.cpp
+++ b/torch_xla/csrc/xla_sharding_util.cpp
@@ -169,6 +169,26 @@ bool ShardingUtil::SetHloSharding(LoweringContext* lowering_ctx) {
   return is_sharded;
 }
 
+ShardingUtil::ShardingType ShardingUtil::GetShardingType(
+    xla::OpSharding& sharding) {
+  switch (sharding.type()) {
+    case xla::OpSharding::REPLICATED:
+      return ShardingType::REPLICATED;
+    case xla::OpSharding::MAXIMAL:
+      return ShardingType::MAXIMAL;
+    case xla::OpSharding::TUPLE:
+      return ShardingType::TUPLE;
+    case xla::OpSharding::OTHER:
+      // OTHER sharding can indicate either PARTIAL or TILED sharding.
+      return sharding.replicate_on_last_tile_dim() ? ShardingType::PARTIAL
+                                                   : ShardingType::TILED;
+    case xla::OpSharding::MANUAL:
+      return ShardingType::MANUAL;
+    default:
+      TF_LOG(ERROR) << "Unsupported sharding type: " << sharding.type();
+  }
+}
+
 bool ShardingUtil::EqualShardingSpecs(const XLATensor::ShardingSpec& a,
                                       const XLATensor::ShardingSpec& b) {
   return xla::protobuf_util::ProtobufEquals(a.sharding, b.sharding);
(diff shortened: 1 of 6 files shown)


jonb377:
cc @yashs97

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -413,9 +433,9 @@ ShardingUtil::GetShardIndicesForDevices(
           continue;
         }
         int64_t n_j = offset % tile_shape[j];
-        int start = n_j * shard_shape[j];
-        // Clamp the end of the slice to the tensor shape to accurately reflect
+        // Clamp the slice bounds to the tensor shape to accurately reflect
         // the shard size without padding.
+        int start = std::min(n_j * shard_shape[j], tensor_shape[j]);
```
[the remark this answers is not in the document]
jonb377: I would call this a latent bug, but it wasn't breaking anything because torch indexing handles negative-length indices as though they were empty. It just breaks the expectation that `stop - start` reflects the size of the unpadded shard, which we rely on in distributed checkpointing.

You're right - these index slices will end up empty, but this is the desired outcome when the shard consists entirely of padding.

alanwaketan (approved):
LGTM.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -169,6 +169,26 @@ bool ShardingUtil::SetHloSharding(LoweringContext* lowering_ctx) {
...
+  switch (sharding.type()) {
+    case xla::OpSharding::REPLICATED:
+      return ShardingType::REPLICATED;
+    case xla::OpSharding::MAXIMAL:
+      return ShardingType::MAXIMAL;
+    case xla::OpSharding::TUPLE:
+      return ShardingType::TUPLE;
+    case xla::OpSharding::OTHER:
+      // OTHER sharding can indicate either PARTIAL or TILED sharding.
+      return sharding.replicate_on_last_tile_dim() ? ShardingType::PARTIAL
```
alanwaketan, line 183: This seems pretty hacky. But I guess we don't have other ways round?

jonb377: My understanding is that we distinguish partial replication as a different sharding type whereas XLA treats partial and tiled as the same type `OTHER`. @yeounoh could you confirm?

JackCaoG:
@jonb377 @alanwaketan is this pr ready to merge?

jonb377:
Yes, I'll merge after TPU CI finishes

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: Yes, this is actually the correct way, the compiler treats TILED and PARTIAL the same as the OTHER type. The differences between the two would be how the tile shards are assigned to difference devices.

PR #5129, opened 2023-06-07
Update Python device API for SPMD

JackCaoG:
The purpose of this pr is to clean up the xla_model api call behavior under the spmd context. IMO we should always return `xla:0` to the user and use that to represent the global device. The other api like `global_device_count` and `ordinal` will also behave as there is only one device. The idea here is that all of the existing api under `xla_model` is about the `xla` device. Under SPMD context there is one XLA device, which is a virtual device.

We provided a set of separate api that has `runtime` in their name for user to query the real runtime device information. for example, user should use 
```
xr. global_runtime_device_count()
```

instead of 

```
xr. global_device_count()  --> always return 1 when `XLA_USE_SPMD=1`
```
or
```
len(xm.get_xla_supported_devices()) --> always return 1 when `XLA_USE_SPMD=1`
```

This pr will require some model code change, hence I want to push this before next release.

Next step
1. add similar test for apis under `runtime.py`
2. add a warning message when user using `global_device_count`(and other similar apis) under SPMD context
3. clean up the `SPMD` device in C++ land

### DIFF
--- a/test/pjrt/test_runtime_tpu.py
+++ b/test/pjrt/test_runtime_tpu.py
@@ -188,22 +188,22 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
           {i: torch.device(f'xla:{i}') for i in range(self.num_devices)})
 
   @staticmethod
-  def _physical_device_attributes():
-    return xr.physical_device_attributes(str(xm.xla_device()))
+  def _runtime_device_attributes():
+    return xr.runtime_device_attributes(str(xm.xla_device()))
 
-  def test_physical_device_attributes(self):
-    result = pjrt.run_multiprocess(self._physical_device_attributes)
+  def test_runtime_device_attributes(self):
+    result = pjrt.run_multiprocess(self._runtime_device_attributes)
     for device in result.values():
       self.assertCountEqual(['coords', 'core_on_chip'], list(device.keys()))
       self.assertIsInstance(device['coords'], list)
       self.assertIsInstance(device['core_on_chip'], int)
 
   @staticmethod
-  def _global_physical_device_attributes():
-    return xr.global_physical_device_attributes()
+  def _global_runtime_device_attributes():
+    return xr.global_runtime_device_attributes()
 
-  def test_global_physical_device_attributes(self):
-    results = pjrt.run_multiprocess(self._global_physical_device_attributes)
+  def test_global_runtime_device_attributes(self):
+    results = pjrt.run_multiprocess(self._global_runtime_device_attributes)
     for result in results.values():
       for device in result:
         self.assertCountEqual(['coords', 'core_on_chip'], list(device.keys()))
--- a/test/pjrt/test_runtime_tpu.py
+++ b/test/pjrt/test_runtime_tpu.py
@@ -188,22 +188,22 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
           {i: torch.device(f'xla:{i}') for i in range(self.num_devices)})
 
   @staticmethod
-  def _device_attributes():
-    return xr.device_attributes(str(xm.xla_device()))
+  def _physical_device_attributes():
+    return xr.physical_device_attributes(str(xm.xla_device()))
 
-  def test_device_attributes(self):
-    result = pjrt.run_multiprocess(self._device_attributes)
+  def test_physical_device_attributes(self):
+    result = pjrt.run_multiprocess(self._physical_device_attributes)
     for device in result.values():
       self.assertCountEqual(['coords', 'core_on_chip'], list(device.keys()))
       self.assertIsInstance(device['coords'], list)
       self.assertIsInstance(device['core_on_chip'], int)
 
   @staticmethod
-  def _global_device_attributes():
-    return xr.global_device_attributes()
+  def _global_physical_device_attributes():
+    return xr.global_physical_device_attributes()
 
-  def test_global_device_attributes(self):
-    results = pjrt.run_multiprocess(self._global_device_attributes)
+  def test_global_physical_device_attributes(self):
+    results = pjrt.run_multiprocess(self._global_physical_device_attributes)
     for result in results.values():
       for device in result:
         self.assertCountEqual(['coords', 'core_on_chip'], list(device.keys()))
--- a/test/spmd/test_spmd_xla_model_api.py
+++ b/test/spmd/test_spmd_xla_model_api.py
@@ -1,0 +1,59 @@
+import unittest
+import os
+import sys
+
+import torch
+import torch_xla
+import torch_xla.core.xla_model as xm
+import test_xla_sharding_base
+
+
+class BasicXMAPITest(test_xla_sharding_base.XlaShardingTest):
+
+  @classmethod
+  def setUpClass(cls):
+    os.environ["XLA_USE_SPMD"] = "1"
+    super().setUpClass()
+
+  def test_get_xla_supported_devices(self):
+    device_type = os.environ['PJRT_DEVICE']
+    devices = xm.get_xla_supported_devices(device_type)
+    self.assertEqual(len(devices), 1)
+
+  def test_world_size(self):
+    self.assertEqual(xm.xrt_world_size(), 1)
+
+  def test_get_ordinal(self):
+    self.assertEqual(xm.get_ordinal(), 0)
+
+  def test_get_local_ordinal(self):
+    self.assertEqual(xm.get_local_ordinal(), 0)
+
+  def test_is_master_ordinal(self):
+    self.assertTrue(xm.is_master_ordinal())
+
+  def test_xla_device(self):
+    device = xm.xla_device()
+    self.assertEqual(device, torch.device('xla:0'))
+
+  def test_xla_real_devices(self):
+    device = xm.xla_device()
+    device_type = os.environ['PJRT_DEVICE']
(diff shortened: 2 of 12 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -967,12 +967,37 @@ void InitXlaModuleBindings(py::module m) {
...
+    }
+  });
+  m.def("_xla_num_devices", []() -> int64_t {
+    if (ShardingUtil::UseVirtualDevice()) {
+      return 1;
+    } else {
+      return xla::GetComputationClient()->GetNumDevices();
+    }
+  });
+  m.def("_xla_get_all_devices", []() {
```
jonb377, line 989: This would cause `global_device_count()` to return 1, but we rely on that API to construct the mesh

JackCaoG: yea, I am thinking about has 2 concepts, `num_devices` -> number of xla devices, in SPMD case it will be 1. and another concept is `num_real_devices` which will return the actual TPU devices. I am still debating whether it make things more clear or more confusing..

vanbasten23, line 989: In the other thread you mentioned: `num_devices -> number of xla device` and `another concept is num_real_devices which will return the actual TPU devices`. Just want to check if my understanding is correct:

- `num_devices` refers to the logical device. So in a v4-8 non-spmd case, the number should be 4.
- `num_real_devices` refers to the physical device. In a v4-8 non-spmd case, the number should be 8?

JackCaoG: `num_devices` should return 1 and `num_real_devices` should reutrn 4. Since there is 1 logic(virtual) device, and 4 real device. We don't need to worry about core vs chip.

torch_xla/experimental/pjrt.py
```diff
@@ -7,9 +7,9 @@
 
 aliases = [
     runtime.addressable_device_count,
-    runtime.device_attributes,
```
will-cromar, line 10: You'll have to rename these two functions like `rendezvous` below, otherwise the API will just be `pjrt.physical_device_attributes`. e.g. `device_attributes = deprecated(this_module, runtime.physical_device_attributes)`

will-cromar (commented):
LGTM in principle

I'm on the fence about the naming convention here. So far, I've been thinking about the distinction between a TPU chip (an actual PCI device) and a TPU core (a "device" within that chip) by calling the TPU core a "logical" device and the TPU chip a "physical" device. 

We should settle on terms we can use by convention for each of these sets:

1. The set of all runtime devices (e.g. TPU cores)
2. The set of all PCI devices (e.g. TPU chip)
3. The set of all PyTorch devices (either SPMD or runtime devices)

To me, "physical" makes the most sense to me as (2), since these are the actual devices you see if you run `lspci`.

alanwaketan:
> LGTM in principle
> 
> I'm on the fence about the naming convention here. So far, I've been thinking about the distinction between a TPU chip (an actual PCI device) and a TPU core (a "device" within that chip) by calling the TPU core a "logical" device and the TPU chip a "physical" device.
> 
> We should settle on terms we can use by convention for each of these sets:
> 
> 1. The set of all runtime devices (e.g. TPU cores)
> 2. The set of all PCI devices (e.g. TPU chip)
> 3. The set of all PyTorch devices (either the SPMD virtual device or runtime devices)
> 
> To me, "physical" makes the most sense to me as (2), since these are the actual devices you see if you run `lspci`.

I'm thinking this is very architecture dependent?
For V4, each device is actually corresponding to a chip.
For V3, it's the story you present.

alanwaketan (approved):
LGTM. Pending Will's comment.

will-cromar:
Right, (1) and (2) are going to be the same for TPU v4 because the driver represents both cores as one large device. The same is true for GPU.

For TPU v2/v3 and CPU, (1) will be larger than (2) because the runtime will represent the cores as separate devices within the same process.

JackCaoG:
I am flexibile with the naming convention here, I forgot about the chip vs core for v3... Did we ever expose the chip vs core to user through any api? For example is there any api that returns `number of chip` which will will return `4` for 'v3-8'?

I think it would be best if user don't need to know about the core vs chips since it is really confusing.

in my mind, under SPMD, we have 1 logic device. In normal data parallel, v3-8 has 8 logical device and v4-8 has 4 logic device.

SPMD 
Accelator | logical | physcial |
|---|---|---|
v3-8 | 1 | 8
v4-8 | 1 | 4

non-SPMD 
Accelator | logical | physcial |
|---|---|---|
v3-8 | 8 | 8
v4-8 | 4 | 4

will-cromar:
Yeah, there are APIs in `tpu.py` the differentiate between chips and cores: https://github.com/pytorch/xla/blob/56ae70732b8a101e433feccdc5bb3d8999e7c86a/torch_xla/_internal/tpu.py#L92-L119

Concretely, the number of chips is the expected number of local processes when using multiprocessing. I agree that for normal use cases, we should bury this distinction as much as we can. A clear naming convention is important for tools building on us (e.g. Lightning and Accelerate) and for ourselves once we forget this conversation.

I'm fine with "logical" devices being the of `torch.device`s we expose. What do you think of these?

1. "physical" devices = PCI devices
2. "logical" devices =  `torch.device`s
3. "runtime" devices = `PjRtDevice`s

JackCaoG:
yea, I am ok with the name `runtime` devices, I can rename my api to `global_runtime_device_count `

JackCaoG:
I am half way editing this pr but now I have a different thought. In https://github.com/pytorch/xla/blob/master/torch_xla/experimental/xla_sharding.py we used term `physcial` a lot. I also think `physical` is an easier to understand term, since most people probably don't know what runtime we are using and how `runtime` device is different than `logical` device. If we are only going to use the PCI device concept internally(user don't access it), can we just call it `pci_device` and leave the `physical_device` to be the number of device that user can create mesh on?

JackCaoG:
@will-cromar gentle ping on your opinion toward using `physical` to represent `runtime` device.

JackCaoG:
chatted with Will offline, I think I am just going to use the term defined in https://github.com/pytorch/xla/pull/5129#issuecomment-1610279532 and update all `physcial` in this pr to be `runtime` to be accurate.

test/pjrt/test_runtime_tpu.py
```diff
@@ -188,22 +188,22 @@ def test_spawn_threads(self):
           {i: torch.device(f'xla:{i}') for i in range(self.num_devices)})
 
   @staticmethod
-  def _device_attributes():
-    return xr.device_attributes(str(xm.xla_device()))
+  def _physical_device_attributes():
```
jonb377, line 191: Should these be `runtime` instead of `physical` in this file?

JackCaoG: good catch

torch_xla/experimental/pjrt.py
JackCaoG: done

test/pjrt/test_runtime_tpu.py
JackCaoG: fixed.

JackCaoG:
@will-cromar @jonb377 Can I get a review for this one?

jonb377 (approved):
Overall LGTM, just a couple of questions to help my understanding

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -836,12 +836,47 @@ void InitXlaModuleBindings(py::module m) {
...
         [](const at::Tensor& tensor) { return GetTensorId(tensor); });
-  m.def("_xla_get_devices",
+  m.def("_xla_get_devices", []() {
+    if (UseVirtualDevice()) {
+      // Under SPMD context, there is only one virtual devices from user
+      // perspective.
+      std::vector<std::string> all_devices =
+          runtime::GetComputationClient()->GetAllDevices();
+      all_devices.resize(1);
+      return all_devices;
```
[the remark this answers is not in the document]
JackCaoG: I think `TPU:0`, this is also how we `get_hw_type` works, by looking at the prefix of the device.

test/spmd/test_spmd_xla_model_api.py
```diff
@@ -0,0 +1,59 @@
...
+
+  @classmethod
+  def setUpClass(cls):
+    os.environ["XLA_USE_SPMD"] = "1"
+    super().setUpClass()
+
+  def test_get_xla_supported_devices(self):
+    device_type = os.environ['PJRT_DEVICE']
+    devices = xm.get_xla_supported_devices(device_type)
+    self.assertEqual(len(devices), 1)
```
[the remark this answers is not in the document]
JackCaoG: good question, let me check.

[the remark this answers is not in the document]
JackCaoG: it is `xla:0`, I think we never return `tpu:0` to user, through we do store it in the `_DEVICE`.

JackCaoG:
@amithrm @Liyang90 @jonb377 @alanwaketan @baoleai I am going to merge this pr now, the most noticeable change to users are you should not use `xr. global_device_count ` under `SPMD` since it will return 1(since there is only one xla virtual device). You should use `xr. global_runtime_device_count ` which will tell you the actual number of devices that runtime sees.

PR #5130, opened 2023-06-07
Implement SPMDLoadPlanner to enable distributed checkpoint loading

jonb377:
This implements the [LoadPlanner interface](https://github.com/pytorch/pytorch/blob/2800a04a17d727dc12353dbea4f4007682194822/torch/distributed/checkpoint/planner.py#L245) from torch.distributed.checkpoint. This implementation only directly handles sharded tensors and relies on the default planner's logic for everything else.

A high-level overview of each of the Planner interface methods:
- `set_up_planner`: Called with the state_dict to be restored and metadata from the checkpoint. Our implementation will split the state_dict into a sharded and unsharded portion so that we can defer to the default planner logic for the unsharded part.
- `create_local_plan`: ReadItems are generated for every item in the state_dict. The default planner is used for the unsharded objects, and we generate a ReadItem for each shard of each XLAShardedTensor with non-REPLICATED sharding type.
- `create_global_plan`: The coordinator process makes any global decisions for the restoration. There is no custom logic here.
- `finish_plan`: The process can adjust its plan after global coordination. Again, no custom logic here.
- `load_bytes`: This is how non-tensor data is restored. We defer to the default planner's logic.
- `resolve_tensor`: This function returns a tensor to store the read result of the associated ReadItem. If the ReadItem doesn't correspond to a sharded tensor, we defer to the default planner logic. Otherwise, we return the unpadded data of the local shard associated with the ReadItem.
- `commit_tensor`: This is called after the data has been loaded into the tensor. This is a no-op in the default planner, but for sharded tensors we track each shard that has been committed. Once all shards are committed for a tensor, they are loaded into the XLAShardedTensor.

This PR depends on https://github.com/pytorch/xla/pull/5128 for some utility functions.

### DIFF
--- a/test/spmd/test_xla_distributed_checkpoint.py
+++ b/test/spmd/test_xla_distributed_checkpoint.py
@@ -1,0 +1,165 @@
+import os
+import tempfile
+import unittest
+import test_xla_sharding_base
+
+import torch
+import torch.distributed.checkpoint as dist_cp
+import torch_xla.core.xla_model as xm
+import torch_xla.runtime as xr
+import torch_xla.experimental.xla_sharding as xs
+
+from torch.distributed.checkpoint.default_planner import (
+    create_default_local_save_plan,
+    create_default_global_save_plan,
+)
+from torch_xla.experimental.distributed_checkpoint import SPMDLoadPlanner
+
+
+class DistributedCheckpointTestBase(test_xla_sharding_base.XlaShardingTest):
+
+  @classmethod
+  def setUpClass(cls):
+    os.environ['XLA_USE_SPMD'] = '1'
+    super().setUpClass()
+
+  def _get_sharded_model(self, mesh_shape=None):
+    # Return a sharded SimpleLinear model with fc1.weight sharded and
+    # fc2.weight explicitly replicated
+    mesh_shape = mesh_shape or (1, self.n_devices)
+    model = self.SimpleLinear().to(xm.xla_device())
+    mesh = self._get_mesh(mesh_shape)
+    xs.mark_sharding(model.fc1.weight, mesh, (0, 1))
+    xs.mark_sharding(model.fc2.weight, mesh, (None, None))
+    return model
+
+  def _get_default_local_metadata(self):
+    # Create a default Metadata instance for a SimpleLinear model by using
+    # the default local save plan on the model's CPU state_dict
+    cpu_state_dict = self.SimpleLinear().state_dict()
+    local_plan = create_default_local_save_plan(cpu_state_dict, True)
+    _, md = create_default_global_save_plan([local_plan])
+    return md
+
+  def _same_shard_data(self, shards, others) -> bool:
+    for a, b in zip(shards, others):
+      if not torch.allclose(a.data, b.data):
+        return False
+    return True
+
+
+class ReshardingTest(DistributedCheckpointTestBase):
+
+  def _save_and_restore(self,
+                        model_in,
+                        model_out,
+                        save_planner=None,
+                        load_planner=None):
+    """
+    Checkpoint model_in using the provided save_planner and load into model_out
+    using the provided load_planner, and assert model_out equals model_in after
+    the load. If either planner is not specified, the DefaultPlanner is used.
+    """
+    tmpdir = tempfile.mkdtemp()
+
+    # Save an unsharded model using the provided save planner
+    dist_cp.save_state_dict(
+        state_dict=model_in.state_dict(),
+        storage_writer=dist_cp.FileSystemWriter(tmpdir),
+        planner=save_planner,
+        no_dist=True,  # Single-host checkpoint doesn't require a process group
+    )
+
+    # Load the checkpoint using the provided load planner
+    for p1, p2 in zip(model_in.parameters(), model_out.parameters()):
+      self.assertFalse(torch.allclose(p1, p2))
+    dist_cp.load_state_dict(
+        state_dict=model_out.state_dict(),
+        storage_reader=dist_cp.FileSystemReader(tmpdir),
+        planner=load_planner,
+        no_dist=True,  # Single-host checkpoint doesn't require a process group
+    )
+    for p1, p2 in zip(model_in.parameters(), model_out.parameters()):
+      self.assertTrue(torch.allclose(p1, p2))
+
+  def test_unsharded_to_sharded(self):
+    # Save an unsharded model using the DefaultSavePlanner and load into a
+    # sharded model using the SPMDLoadPlanner
+    model = self.SimpleLinear().to(xm.xla_device())
+    sharded_model = self._get_sharded_model()
+    self._save_and_restore(model, sharded_model, load_planner=SPMDLoadPlanner())
+
+
+class SPMDLoadPlannerTest(DistributedCheckpointTestBase):
+
+  def _get_load_planner(self, model):
+    # Create an SPMDLoadPlanner for the given model.
+    md = self._get_default_local_metadata()
+    planner = SPMDLoadPlanner()
+    planner.set_up_planner(model.state_dict(), md, True)
+    return planner
+
+  def test_state_dict_separation(self):
+    model = self._get_sharded_model()
+    planner = self._get_load_planner(model)
+    if self.n_devices > 1:
+      # The state_dict should be flattened and separated
+      self.assertCountEqual(planner.sharded_state_dict, ['fc1.weight'])
+      # `fc2.weight` should be in the unsharded_state_dict despite having
+      # an explicit mark_sharding call because it is replicated.
+      self.assertCountEqual(planner.unsharded_state_dict,
+                            ['fc1.bias', 'fc2.weight', 'fc2.bias'])
+    else:
+      # With a single device, no tensors are sharded.
+      self.assertCountEqual(
+          planner.unsharded_state_dict,
+          ['fc1.weight', 'fc1.bias', 'fc2.weight', 'fc2.bias'])
+
+  def test_local_load_plan(self):
+    model = self._get_sharded_model()
+    planner = self._get_load_planner(model)
+    plan = planner.create_local_plan()
+    parameter_count = len(list(model.parameters()))
+    if self.n_devices > 1:
+      # When the model is sharded across devices, fc1.weight will result in
+      # self.n_devices ReadItems while all other tensors result in a single
+      # ReadItem.
+      sharded_read_items = [
+          ri for ri in plan.items if ri.dest_index.fqn == 'fc1.weight'
+      ]
+      self.assertEqual(self.n_devices, len(sharded_read_items))
+      # Every other parameter should have a single ReadItem
+      unsharded_read_items = set(plan.items) - set(sharded_read_items)
+      self.assertEqual(parameter_count - 1, len(unsharded_read_items))
+    else:
+      # If unsharded, there should be a single ReadItem per model parameter
+      self.assertEqual(parameter_count, len(plan.items))
+
+  @unittest.skipIf(xr.global_device_count() == 1,
+                   "Multiple devices required to shard tensors")
+  def test_resolve_and_commit_sharded_tensor(self):
+    model = self._get_sharded_model()
+    planner = self._get_load_planner(model)
+    plan = planner.create_local_plan()
+
+    xtensor = xs.wrap_if_sharded(model.fc1.weight)
+    old_shards = xtensor.local_shards
+    sharded_read_items = [
+        ri for ri in plan.items if ri.dest_index.fqn == 'fc1.weight'
+    ]
+    self.assertEqual(self.n_devices, len(sharded_read_items))
+    for read_item in sharded_read_items:
+      # Before all ReadItems have been processed, the tensor shards should not
+      # be updated
+      self.assertTrue(self._same_shard_data(xtensor.local_shards, old_shards))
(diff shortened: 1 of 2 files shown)


jonb377:
cc @yashs97

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -0,0 +1,145 @@
...
+    for read_item in sharded_read_items:
+      # Before all ReadItems have been processed, the tensor shards should not be updated
+      self.assertTrue(self._same_shard_data(xtensor.local_shards, old_shards))
+      tensor = planner.resolve_tensor(read_item)
+      planner.commit_tensor(read_item, -tensor)
+    # After all ReadItems are processed, the local_shards should reflect the new values
+    self.assertFalse(self._same_shard_data(xtensor.local_shards, old_shards))
+
+  def test_load_state_dict(self):
+    dist.init_process_group(
```
alanwaketan, line 115: Why do we need the process_group?

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -0,0 +1,145 @@
...
+        backend='gloo',
+        rank=0,
+        world_size=1,
+        store=dist.FileStore("/tmp/filestore", 1))
+    tmpdir = tempfile.mkdtemp()
+
+    # Save an unsharded model using the default planner in dist_cp.save_state_dict
+    model = self.SimpleLinear().to(xm.xla_device())
```
alanwaketan, line 123: Can you illustrate more on how a saved unsharded model can be loaded into a sharded model?

alanwaketan (commented):
Thanks @jonb377. This is great. Left a few comments.

test/spmd/test_xla_distributed_checkpoint.py
jonb377: `torch.distributed.checkpoint` requires a process group to coordinate merging the local plans into a global plan. Even though this is just a single instance, we still need to initialize the PG to use the `save_state_dict` and `load_state_dict` APIs.

test/spmd/test_xla_distributed_checkpoint.py
jonb377: This is a feature from the upstream - resharding is handled transparently for us when we use the [create_read_items_for_chunk_list](https://github.com/pytorch/pytorch/blob/47cfcf566ab76573452787335f10c9ca185752dc/torch/distributed/checkpoint/planner_helpers.py#L141) API to create our ReadItems. Even sharded checkpoints should be able to be restored into different device meshes or an unsharded models.

torch_xla/experimental/checkpoint.py
```diff
@@ -0,0 +1,205 @@
...
+
+  def create_global_plan(
+      self, all_plans: List[SavePlan]) -> Tuple[List[SavePlan], Metadata]:
+    raise NotImplemented
+
+  def finish_plan(self, new_plan: SavePlan) -> SavePlan:
+    raise NotImplemented
+
+
+class XLALoadPlanner(LoadPlanner):
```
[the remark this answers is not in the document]
jonb377: Good point! I'll move the declarations into `__init__`. `set_up_planner` will still handle initializing most of the fields, but this will definitely help with readability.

torch_xla/experimental/checkpoint.py
```diff
@@ -0,0 +1,205 @@
...
+      is_coordinator: bool,
+  ) -> None:
+    self.metadata = metadata
+    self.is_coordinator = is_coordinator
+    self.original_state_dict = state_dict
+
+    # Flatten the state_dict to allow separating sharded XLA tensors from
+    # types that can be handled by the default planner, and ensure all sharded
+    # tensors are wrapped in XLAShardedTensor
+    state_dict, self.mappings = flatten_state_dict(state_dict)
```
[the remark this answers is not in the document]
jonb377: Hmm I actually don't know why it's optional in the DefaultLoadPlanner, I'll follow up on that. The flattened state_dict is easier for us to work with since we need to split the input into sharded/unsharded parts, which is difficult if the state_dict is nested.

One downside of flattening is that we aren't operating directly on the input state_dict, so non-tensor items need to be mapped back to the original state_dict.

torch_xla/experimental/checkpoint.py
```diff
@@ -0,0 +1,205 @@
...
+      sizes=torch.Size(ind.stop - ind.start for ind in index))
+
+
+def _create_xla_read_items(sharded_state_dict: STATE_DICT_TYPE,
+                           metadata: Metadata) -> List[ReadItem]:
+  """
+  Iterate through the state_dict and return ReadItems for all local shards.
+  """
+  items = []
+  for fqn, t in sharded_state_dict.items():
```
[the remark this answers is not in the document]
jonb377: "Fully qualified name", e.g. `fc1.weight` - these are the keys after flattening the state_dict.

test/spmd/test_xla_distributed_checkpoint.py
alanwaketan: That makes sense now.

test/spmd/test_xla_distributed_checkpoint.py
alanwaketan: Hmmm, is this because we are saving the unsharded CPU tensor? If it's required all the time, I feel like it's not a good UX.

torch_xla/experimental/checkpoint.py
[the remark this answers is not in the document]
alanwaketan: But that down side is handled by the default planner helpers, right?

test/spmd/test_xla_distributed_checkpoint.py
jonb377: A PG is always required, but I'll need to check if we can use the XLA backend - my first attempt to replace `gloo` with an `xla` group errored out when creating the global plan.

If we build a higher-level interface, this could be hidden this from the user.

torch_xla/experimental/checkpoint.py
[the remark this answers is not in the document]
jonb377: That's correct.

test/spmd/test_xla_distributed_checkpoint.py
alanwaketan: Most of our own distributed API does't require PG. So, maybe we should try keep the consistency. Do you know why they require PG?

jonb377: It's used to centrally coordinate the global plans, e.g. all worker plans are sent to the coordinator using collectives in the process group.

Actually, I see we have the option to run without a PG by setting `use_dist=False` in the `(save|load)_state_dict` functions, but we lose out on the global coordination. This shouldn't be an issue for the LoadPlanner, and I suspect we can make the SavePlanner work without global planning as well - we'll just need to track replica ranks and dedupe tensors locally.

Thanks for bringing this up Jiewen, I'll experiment with it and update the PR.

jonb377: Synced offline - when taking a checkpoint, we would lose the consistency provided by the process group (the coordinator waits until all workers are finished before writing metadata, which depends on a PG).

I'll use `no_dist=True` in the tests to simplify the test code, but we will still expect users to have a CPU process group when taking a checkpoint.

torch_xla/experimental/checkpoint.py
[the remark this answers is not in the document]
kumpera: It's optional on DefaultLoadPlaner mostly to help with testing. Its default value is True, so most users will always have it enabled.

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -0,0 +1,165 @@
...
+        state_dict=model_in.state_dict(),
+        storage_writer=dist_cp.FileSystemWriter(tmpdir),
+        planner=save_planner,
+        no_dist=True,  # Single-host checkpoint doesn't require a process group
+    )
+
+    # Load the checkpoint using the provided load planner
+    for p1, p2 in zip(model_in.parameters(), model_out.parameters()):
+      self.assertFalse(torch.allclose(p1, p2))
+    dist_cp.load_state_dict(
```
alanwaketan, line 76: It looks like the LoadPlanner should be able to handle the resharding of the consolidated checkpoints. I wonder how does it handle the case where the save checkpoint has a different sharding spec than the loading model?

alanwaketan: Seems like it's handled by create_read_items_for_chunk_list?

alanwaketan (approved):
Generally speaking, it LGTM! Thanks for getting this complicated work being done so quick. I left a bunch of comments just for my education to learn more in-depth on how this planner is supposed to be used by load_state_dict API.

Some of the missing pieces that I imagine the load_state_dict will do to fill some of the gaps of planners are:
1. Loading the actual metadata and storage.
2. Loading the storage and slice it into the proper size of tensors for the ReadItems.

It will be great if you can elaborate more on the E2E flow on how the high level APIs are interacting with our low level implementation. I also commented this on your design doc.

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -0,0 +1,165 @@
...
+          planner.unsharded_state_dict,
+          ['fc1.weight', 'fc1.bias', 'fc2.weight', 'fc2.bias'])
+
+  def test_local_load_plan(self):
+    model = self._get_sharded_model()
+    planner = self._get_load_planner(model)
+    plan = planner.create_local_plan()
+    parameter_count = len(list(model.parameters()))
+    if self.n_devices > 1:
+      # When the model is sharded across devices, fc1.weight will result in
```
alanwaketan, line 124: So you suggest the number of ReadItem will match the number of XLAShard for a particular XLAShardedTensor. Does this hold true for the following resharding example?
1. stored shards <= loading shard, i.e., the models is sharded in 2 ways but need to loaded into 4 ways.
2. stored shards > loading shard, i.e., the models is sharded in 4 ways but need to loaded into 2 ways.

I can see that holds true for case 1 but it's hard for me to imagine for case 2. Can you elaborate? Commented in your design as well.

jonb377: That's true - I'll clarify the comment that this comes from using the unsharded checkpoint metadata used when creating the Planner in `_get_load_planner`.

torch_xla/experimental/distributed_checkpoint.py
```diff
@@ -0,0 +1,241 @@
...
+        ind.start for ind in xla_shard.indices
+    ), "ReadItem does not correspond to the correct shard"
+    # Return padded data since the tensor will be narrowed to match
+    # the ReadItem in `transform_tensor`.
+    return xla_shard.data
+
+  def transform_tensor(self, read_item: ReadItem, tensor: torch.Tensor):
+    offsets = read_item.dest_offsets
+    index = read_item.dest_index
+    if index.fqn in self.sharded_state_dict:
```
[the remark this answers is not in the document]
jonb377: I'll add a comment - the storage layer expects that the tensor returned from `resolve_tensor` matches the shape of the ReadItem's `lengths` field, so the tensor must be narrowed here.

torch_xla/experimental/distributed_checkpoint.py
```diff
@@ -0,0 +1,241 @@
...
+  return items
+
+
+def _is_sharded_tensor(x: Any) -> bool:
+  """Return true if the tensor's data is sharded across multiple devices"""
+  return isinstance(
+      x, XLAShardedTensor) and x.sharding_type != ShardingType.REPLICATED
+
+
+def _unwrap_sharded_tensor(x: Any) -> Any:
```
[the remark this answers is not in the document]
jonb377: Correct, that's a good point - the name is confusing. This is used to ensure the default planner is operating on `torch.Tensor` instead of `XLAShardedTensor`. Maybe `_unwrap_xla_sharded_tensor`?

test/spmd/test_xla_distributed_checkpoint.py
alanwaketan: Then is there a simple formula to calculate the number of ReadItems? Or I have to refer to the create_read_items_for_chunk_list algorithm?

test/spmd/test_xla_distributed_checkpoint.py
jonb377: If we are loading into a coarser mesh, e.g. loading a checkpoint taken on a (4, 4) mesh into a model sharded across (2, 2), each local shard in the (2, 2) mesh will require multiple ReadItems. On our side, we will still generate a single ChunkStorageMetadata for each shard, and the utility function `create_read_items_for_chunk_list` translates these into the ReadItems necessary by finding the overlap of `ChunkStorageMetadata` with the chunks in the storage metadata.

The algorithm for `create_read_items_for_chunk_list` just iterates across all chunks in the storage metadata, identifies overlap with the `ChunkStorageMetadata` describing our XLAShard, and generates a ReadItem if there is overlap.

alanwaketan: High level speaking, the number of ReadItems will then be determined by the max(number of shards in loading model, number of shards in saving model)?

alanwaketan: Perfect. Thanks, Jon!

PR #5135, opened 2023-06-07
Disable XRT in default CI jobs

will-cromar:
- Add CPU-only XRT tests as a sanity check.
- Rename `PJRT_ONLY` to `DISABLE_XRT`

### DIFF
--- a/.circleci/test_xrt.sh
+++ b/.circleci/test_xrt.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -ex
+
+source ./xla_env
+source .circleci/common.sh
+
+PYTORCH_DIR=/tmp/pytorch
+XLA_DIR=$PYTORCH_DIR/xla
+USE_COVERAGE="${USE_COVERAGE:-0}"
+
+# Needs to be kept in sync with .jenkins/pytorch/common_utils.sh in pytorch/pytorch.
+TORCHVISION_COMMIT="$(cat $PYTORCH_DIR/.github/ci_commit_pins/vision.txt)"
+
+function pip_install() {
+  # retry 3 times
+  # old versions of pip don't have the "--progress-bar" flag
+  pip install --progress-bar off "$@" || pip install --progress-bar off "$@" || pip install --progress-bar off "$@" ||\
+  pip install "$@" || pip install "$@" || pip install "$@"
+}
+
+function install_torchvision() {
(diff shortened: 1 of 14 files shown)


infra/ansible/playbook.yaml
```diff
@@ -21,6 +21,8 @@
           pattern: ^(aarch64|amd64)$
         - name: accelerator
           pattern: ^(tpu|cuda)$
+        - name: disable_xrt
+          pattern: ^(0|1)$
```
mateuszlewko, line 25: This shouldn't be here. Only variables that don't have default values and need to be passed on every invocation of `ansible-playbook` should be here. 

The build error is there, because task "validate required variables" runs before "Include vars from config files" (which loads vars from vars.yaml).

The variable can still be overwritten

will-cromar: Thanks. I removed the validation here since this flag is temporary and the var gets picked up as expected: https://github.com/pytorch/xla/runs/14112366156

JackCaoG:
> This test is hanging forever, but only on GPU:
> 
> ```
> 2023-06-08T05:09:03.6953379Z + python test/test_autocast.py
> 2023-06-08T05:09:06.1696515Z test_autocast_banned (__main__.TestAutocast) ... ERROR
> 2023-06-08T06:47:50.8195553Z ##[error]The operation was canceled.
> ```
> 
> I think it's trying to run with XRT and getting stuck, since the PJRT device is set in `test/run_tests.sh`. @JackCaoG do you know why this test is not included in `test/run_tests.sh`?
> 
> https://github.com/pytorch/xla/blob/e8dc573769934e4685b964edae75e7ebfb2eeff8/.circleci/common.sh#L133-L135

This test used to only work for GPU hence I left it in circleCI script, it is a AMP test.

.circleci/test_xrt.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -ex
```
JackCaoG, line 3: one day we should get rid of this `.circleci` dir and move it to `.github` or something.

JackCaoG: oh we already have the `.github/workflows`, maybe we can move things over in a separate pr.

will-cromar: Yeah, let's move this in a follow-up PR

JackCaoG (approved):
Thanks @will-cromar !

PR #5136, opened 2023-06-07
Ignore changes in torch_xla/csrc/generated.

vanbasten23:
Without the change, doing `git status` would output:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        torch_xla/csrc/generated/
```
But I don't think doing `git status` should show the changes in the generated folder.

The added line was removed in https://github.com/pytorch/xla/pull/4755. But it seems someone else asked if we should remove that line and @stgpetrovic replied "No, this is now handled with a genrule, there is no state of the repository outside of bazel (it's hermetic and deterministic).".


stgpetrovic (commented):
This dir should not exist/be created by setup. If it does we should look into why. It is safe to ignore it. Genrules generate their output under bazel-out/ as a build artifact.

will-cromar:
Do these files get generated again if you delete them?

vanbasten23:
> Do these files get generated again if you delete them?

ah good callout. They don't. Thanks Will. I'll close the PR.

Issue #5031, opened 2023-05-18
Lower count_nonzero


A9isha:
## 🚀 Feature
Request to lower count_nonzero, https://pytorch.org/docs/stable/generated/torch.count_nonzero.html
## Motivation

This op can be used for fixing autoregressive decoding in TPUs for models like GPT-2 which uses positional embedding

E.g.:

JackCaoG:
@vanbasten23 do you want to pick this up when you have some cycle? This is actually not a dynamic op, I think all we need to do is to first convert the element to contains only `0` and `1` using some mask then reduce in the corresponding dimension to get number of non_zero elements.

vanbasten23:
@JackCaoG  Sorry I didn't see your message earlier. Let me book some time for it this week.

JackCaoG:
The pr is merged, this should be reflected on our nightly tmr and next release.

vanbasten23:
Thanks @JackCaoG for the review.

@A9isha , please give it a try when you have time. I'll close the issue for now. In case you find any issue, feel free to reopen it.

PR #5137, opened 2023-06-07
Lower count_nonzero

vanbasten23:
This pr is to fix issue https://github.com/pytorch/xla/issues/5031 by lowering count_nonzero.


JackCaoG:
can you resolve the merge conflict?

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -2196,6 +2196,60 @@ TEST_F(AtenXlaTensorTest, TestOneIndexTransfer) {
...
+  a[0][2] = 1.0;
+  a[2][2] = 1.0;
+  torch::Tensor b = torch::count_nonzero(a);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = torch::count_nonzero(xla_a);
+    AllClose(b, torch::_cast_Long(xla_b));
+
+    ExpectCounterChanged("xla::count_nonzero", cpp_test::GetIgnoredCounters());
+    ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
```
JackCaoG, line 2211: nit, move this outside the `ForEachDevice`, I don't think you need to call `ReetCounters`?

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -2196,6 +2196,60 @@ TEST_F(AtenXlaTensorTest, TestOneIndexTransfer) {
...
+  a[0][2] = 1.0;
+  a[2][2] = 1.0;
+  c10::optional<long int> dim = 0;
+  torch::Tensor b = torch::count_nonzero(a, dim);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = torch::count_nonzero(xla_a, dim);
+    AllClose(b, torch::_cast_Long(xla_b));
+
+    ExpectCounterChanged("xla::count_nonzero", cpp_test::GetIgnoredCounters());
```
JackCaoG, line 2228: ditto

torch_xla/csrc/ops/count_nonzero.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "torch_xla/csrc/xla_lower_util.h"
+
+namespace torch_xla {
+namespace {
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input,
+                           std::vector<int64_t> dims) {
+  xla::Shape input_shape = GetXlaShape(input);
+  std::vector<int64_t> dimensions;
+  if (dims.empty()) {
```
JackCaoG, line 13: you can do 
`if (!dims.empty())` for line 17-> 22, and remove line 14. since return logic is the same.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -757,6 +757,26 @@ xla::XlaOp BuildLinspace(const torch::lazy::BackendDevice& device,
...
+  static const xla::PrimitiveType kConditionType = xla::PrimitiveType::S32;
+  xla::XlaOp ne_int = xla::ConvertElementType(ne, kConditionType);
+  xla::XlaOp zeros = xla::ZerosLike(ne_int);
+  xla::XlaOp compared =
+      xla::ConvertElementType(xla::Gt(ne_int, zeros), kConditionType);
+  if (dims.empty()) {
+    return xla::ReduceAll(
+        compared, xla::Zero(ne.builder(), kConditionType),
+        xla::CreateScalarAddComputation(kConditionType, ne.builder()));
+  }
```
JackCaoG, line 774: else here maybe? I think in our repo we tend to have `if else` pair to make logic more clear.

test/cpp/test_aten_xla_tensor_1.cpp
vanbasten23: done

test/cpp/test_aten_xla_tensor_1.cpp
vanbasten23: done

torch_xla/csrc/ops/count_nonzero.cpp
vanbasten23: done

torch_xla/csrc/xla_lower_util.cpp
vanbasten23: done

torch_xla/csrc/ops/count_nonzero.cpp
```diff
@@ -0,0 +1,50 @@
...
+    return xla::ShapeUtil::MakeShape(input_shape.element_type(), dimensions);
+  }
+
+  std::unordered_set<int64_t> dims_set(dims.begin(), dims.end());
+  for (int64_t i = 0; i < input_shape.rank(); i++) {
+    if (dims_set.find(i) != dims_set.end()) {
+      dimensions.push_back(i);
```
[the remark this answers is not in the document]
vanbasten23: good catch. Pytorch throws an error. So I'll follow pytorch's behavior.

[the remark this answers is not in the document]
JackCaoG: if pytorch already have that check, we don't need to check here. The assumption is such behavior will be caught by upstream. We can safely assume `dim` passed to us is valid.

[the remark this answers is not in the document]
vanbasten23: It isn't caught by pytorch. Instead, it's caught by xla:
```
XLA builder error: INVALID_ARGUMENT
```
So i would be nice if we can fail it earlier?

torch_xla/csrc/ops/count_nonzero.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input,
+                           std::vector<int64_t> dims) {
+  xla::Shape input_shape = GetXlaShape(input);
+  std::vector<int64_t> dimensions;
+  if (!dims.empty()) {
+    std::unordered_set<int64_t> dims_set(dims.begin(), dims.end());
+    for (int64_t i = 0; i < input_shape.rank(); i++) {
+      if (dims_set.find(i) != dims_set.end()) {
+        dimensions.push_back(i);
```
JackCaoG, line 17: nit, dims here should already be unique given the check in `XLANativeFunctions::count_nonzero`. If all you want is to sort it why not just `std::sort (dims.begin(), dims.end())`

are you also trying to handle the case that `dims` contains `value >= input_shape.rank()` ?  Is that a valid case for this op?

vanbasten23: right, it was meant to remove duplicate dimensions. After I added the check, this is not needed anymore.

JackCaoG (approved):
Thanks @vanbasten23 !

PR #5138, opened 2023-06-07
[test] Update tf_cc_test from OpenXLA

ManfeiBai:


ManfeiBai:
this build will fail because we miss some func definition from tf when we using cc_test

stgpetrovic (commented):
Hm, where do we get the //tensorflow include here? We have only //bazel:rules_def.bzl, but I don't see a tensorflow dir in this patch.

Copying the missing functions seems like the way to go tho.

ManfeiBai:
> Hm, where do we get the //tensorflow include here? We have only //bazel:rules_def.bzl, but I don't see a tensorflow dir in this patch.
> 
> Copying the missing functions seems like the way to go tho.

Thanks for pointing that out, @stgpetrovic, we could use tsl and absl too, so I change to use `cc_test` from tsl

ManfeiBai:
will close this PR and track in https://github.com/pytorch/xla/pull/5165/

PR #5147, opened 2023-06-08
[SPMD] Hybrid Device mesh creation

khatwanimohit:


khatwanimohit:
cc @alanwaketan

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,129 @@ def get_logical_mesh(self):
...
+    granule_mesh = np.arange(len(granules)).reshape(dcn_mesh_shape)
+    blocks = np.vectorize(
+        lambda i: per_granule_meshes[i], otypes=[object])(
+            granule_mesh)
+    device_mesh = np.block(blocks.tolist())
+    return device_mesh
+
+  def get_logical_mesh(self):
+    if np.prod(self.dcn_mesh_shape) > 1:  # multislice
+      return self._create_hybrid_device_mesh(self.ici_mesh_shape,
```
jonb377, line 190: Do we need to regenerate this on each `mark_sharding` call, or can we put this call in the constructor and pass the result to `Mesh::__init__`?

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,129 @@ def get_logical_mesh(self):
...
+        transpose.append(int(y))
+    return physical_mesh.transpose(transpose).reshape(mesh_shape), assignment
+
+  def _create_device_mesh(self, mesh_shape: Sequence[int],
+                          devices: Sequence[Any]) -> np.ndarray:
+    if np.prod(mesh_shape) != len(devices):
+      raise ValueError(
+          f'Number of devices {len(devices)} must equal the product '
+          f'of mesh_shape {mesh_shape}')
+    physical_mesh = self._get_physical_tpu_mesh(devices)
```
jonb377, line 161: Let's add an assertion on the hardware type here - this logic is TPU-specific, and we'll need to handle other hardware types separately. e.g. `assert xm.xla_device_hw() == 'TPU'`

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,129 @@ def get_logical_mesh(self):
...
+    device_mesh, assignment = self._create_device_mesh_for_nd_torus(
+        physical_mesh, mesh_shape)
+    return device_mesh
+
+  def _create_hybrid_device_mesh(self, mesh_shape: Sequence[int],
+                                 dcn_mesh_shape: Sequence[int],
+                                 devices: Sequence[Any]) -> np.ndarray:
```
jonb377, line 168: It looks like all of the logic here is independent of the user-specified devices, since `device_attributes` is for all global devices. Do we need to support user-specified devices for hybrid mesh, or can we assume it will always be across all devices?

khatwanimohit: Removed the dependency on devices. Thanks

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,128 @@ def get_logical_mesh(self):
...
+    assert len(ici_mesh_shape) == len(dcn_mesh_shape)
+    super().__init__(device_ids, mesh_shape, axis_names)
+    self.device_attributes = pjrt.global_device_attributes()
+    if 'slice_index' in self.device_attributes[0] and np.prod(
+        dcn_mesh_shape) == 1:
+      raise ValueError('Provide dcn_mesh_shape to create a mesh for multislice')
+    self.ici_mesh_shape = ici_mesh_shape
+    self.dcn_mesh_shape = dcn_mesh_shape
+    if np.prod(dcn_mesh_shape) > 1:  # multislice
+      self.mesh = self._create_hybrid_device_mesh(self.ici_mesh_shape,
```
jonb377, line 95: Is it a correct understanding that the resulting `device_mesh.shape` should equal `self.mesh_shape`? I'm wondering if we can just use these functions to rearrange the list of device_ids and use the result to construct the `Mesh` superclass, e.g. `device_ids = self._create_hybrid_device_mesh(...).flatten()` and `super().__init__(device_ids, mesh_shape, axis_names)`

This is just to let us reuse more superclass logic, what you have now works too.

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,128 @@ def get_logical_mesh(self):
     return self.device_ids.reshape(self.mesh_shape)
 
 
+class HybridMesh(Mesh):
+  ici_mesh_shape: Tuple[int, ...]
+  dcn_mesh_shape: Tuple[int, ...]
+
+  def __init__(self,
+               device_ids: Union[np.ndarray, List],
```
jonb377, line 79: Like we discussed offline, let's look into removing `device_ids` as a parameter here and defaulting to all global devices - it looks like right now we're silently ignoring the user's input in the hybrid_mesh case

cc @alanwaketan

jonb377 (commented):
Looking great Mohit! Could we also add some basic unit tests in https://github.com/pytorch/xla/blob/master/test/spmd/test_xla_sharding.py?

torch_xla/experimental/xla_sharding.py
alanwaketan: May I ask why, it looks like Mesh takes the device_ids as well.

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,128 @@ def get_logical_mesh(self):
...
+               ici_mesh_shape: Tuple[int, ...],
+               dcn_mesh_shape: Tuple[int, ...] = None,
+               axis_names: Tuple[str, ...] = None):
+    if dcn_mesh_shape == None:
+      dcn_mesh_shape = tuple([1] * len(ici_mesh_shape))
+    mesh_shape = tuple([x * y for x, y in zip(ici_mesh_shape, dcn_mesh_shape)])
+    assert len(ici_mesh_shape) == len(dcn_mesh_shape)
```
alanwaketan, line 86: Shouldn't we assert before before the previous line?

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,128 @@ def get_logical_mesh(self):
...
+    super().__init__(device_ids, mesh_shape, axis_names)
+    self.device_attributes = pjrt.global_device_attributes()
+    if 'slice_index' in self.device_attributes[0] and np.prod(
+        dcn_mesh_shape) == 1:
+      raise ValueError('Provide dcn_mesh_shape to create a mesh for multislice')
+    self.ici_mesh_shape = ici_mesh_shape
+    self.dcn_mesh_shape = dcn_mesh_shape
+    if np.prod(dcn_mesh_shape) > 1:  # multislice
```
alanwaketan, line 94: What if we don't have 'slice_index' in this case? I guess you need to expand the condition on the Line 89.

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,128 @@ def get_logical_mesh(self):
...
+    if np.prod(mesh_shape) != len(devices):
+      raise ValueError(
+          f'Number of devices {len(devices)} must equal the product '
+          f'of mesh_shape {mesh_shape}')
+    physical_mesh = self._get_physical_tpu_mesh(devices)
+    device_mesh, assignment = self._create_device_mesh_for_nd_torus(
+        physical_mesh, mesh_shape)
+    return device_mesh
+
+  def _create_hybrid_device_mesh(self, mesh_shape: Sequence[int],
```
alanwaketan, line 171: mesh_shape => ici_mesh_shape?

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,128 @@ def get_logical_mesh(self):
...
+                                 dcn_mesh_shape: Sequence[int]) -> np.ndarray:
+    granule_dict = defaultdict(list)
+    slice_index_attr = [d['slice_index'] for d in self.device_attributes]
```
alanwaketan, line 174: Do you use this variable at all?

alanwaketan (commented):
It will help me easier to read your code if you can briefly describe what you did in the PR especially on describing some of the key complex functionalities in the code, like _create_hybrid_device_mesh.

Or you can leave a comment on the code.

torch_xla/experimental/xla_sharding.py
alanwaketan: @jonb377 and I discussed offline. I got what Jon means.

torch_xla/experimental/xla_sharding.py
khatwanimohit: changed to ici_mesh_shape

torch_xla/experimental/xla_sharding.py
khatwanimohit: added an extra check

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,134 @@ def get_logical_mesh(self):
...
+    super().__init__(device_ids, mesh_shape, axis_names)
+
+  def _get_physical_tpu_mesh(self, devices: Sequence[Any]) -> np.ndarray:
+    r"""Rearrange TPU devices in a slice into a physical mesh."""
+    assert xm.xla_device_hw(xm.xla_device()) == 'TPU'
+    device_coords = [self.device_attributes[d]['coords'] for d in devices]
+    dims = tuple(d + 1 for d in max(device_coords))
+    out = np.empty(dims, dtype=object)
+    for coords, d in zip(device_coords, devices):
+      out[coords[0], coords[1], coords[2]] = d
```
alanwaketan, line 112: Do you know why coords[0], coords[1], coords[2] has index hard-coded?

alanwaketan: It implies the topology is always 3D?

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,134 @@ def get_logical_mesh(self):
...
+                                             self.dcn_mesh_shape)
+    else:
+      mesh = self._create_device_mesh(self.ici_mesh_shape)
+    device_ids = mesh.flatten()
+    super().__init__(device_ids, mesh_shape, axis_names)
+
+  def _get_physical_tpu_mesh(self, devices: Sequence[Any]) -> np.ndarray:
+    r"""Rearrange TPU devices in a slice into a physical mesh."""
+    assert xm.xla_device_hw(xm.xla_device()) == 'TPU'
+    device_coords = [self.device_attributes[d]['coords'] for d in devices]
```
alanwaketan, line 108: So, each coord is a list of the 3D coord of the device? Can you comment that in the code?

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,134 @@ def get_logical_mesh(self):
...
+
+  def _create_device_mesh_for_nd_torus(
+      self, physical_mesh: np.ndarray,
+      mesh_shape: Sequence[int]) -> Tuple[np.ndarray, List[Tuple[int, ...]]]:
+    # Remaining physical axes to be assigned to logical axes.
+    assignable_physical_mesh = list(physical_mesh.shape)
+    # Map each logical axis to a subset of physical axes.
+    assignment: List[Tuple[int, ...]] = [() for _ in mesh_shape]
+    # Assign logical axes from highest network intensity to lowest.
+    # `mesh_shape` is assumed to ordered by lowest network intensity first, so
```
alanwaketan, line 123: I think we should have this information in the HybridMesh doc string. Can you create one?

torch_xla/experimental/xla_sharding.py
```diff
@@ -70,6 +71,134 @@ def get_logical_mesh(self):
...
+    r"""Rearrange TPU devices in a slice into a physical mesh."""
+    assert xm.xla_device_hw(xm.xla_device()) == 'TPU'
+    device_coords = [self.device_attributes[d]['coords'] for d in devices]
+    dims = tuple(d + 1 for d in max(device_coords))
+    out = np.empty(dims, dtype=object)
+    for coords, d in zip(device_coords, devices):
+      out[coords[0], coords[1], coords[2]] = d
+    return out
+
+  def _create_device_mesh_for_nd_torus(
```
alanwaketan, line 115: Can you explain how this function optimize the performance according to the TPU physical topology? What's the algorithm? Is it the inner ring has the highest performance, so we should assign the back of the mesh_shape to it?

test/spmd/test_xla_sharding.py
```diff
@@ -452,6 +452,16 @@ def test_no_sharding(self):
     t3_expected = [9.0, 9.0, 9.0, 9.0, 9.0, 9.0, 9.0, 9.0]
     self.assertEqual(t3.tolist()[0], t3_expected)
 
+  @unittest.skipUnless(
+      xm.get_xla_supported_devices("TPU"),
+      f"Requires PJRT_DEVICE set to `TPU`.")
+  def test_hybrid_mesh(self):
```
alanwaketan, line 458: We probably can add test cases for MultiSlice but only enables them for the XL-ML tests?

khatwanimohit: Is there a way to add XL ML tests here. I was planning to add them in the other repo.

alanwaketan: It depends on you. You can have a stond-alone test in other repo. Or you can have a test here which is skipped for single slice. Then you run the test in other repo through a config.

jonb377: @khatwanimohit let's add a unit test here, you can mock out `pjrt.global_device_attributes` to spoof the device attributes. It will be good to have some explicit unit tests to verify the logic, and we can have an E2E test in XL ML tests.

test/spmd/test_xla_sharding.py
```diff
@@ -452,6 +454,62 @@ def test_no_sharding(self):
...
+        'coords': [0, 1, 0],
+        'slice_index': 1,
+        'core_on_chip': 0
+    }, {
+        'core_on_chip': 0,
+        'coords': [1, 1, 0],
+        'slice_index': 1
+    }]
+    hybrid_mesh = xs.HybridMesh(
+        ici_mesh_shape=(1, self.n_devices), dcn_mesh_shape=(num_slices, 1))
```
jonb377, line 508: @khatwanimohit Let's remove the `skipIf` on this test and just hard-code `ici_mesh_shape=(1, 4)`, since we're mocking out the global devices this test should work on all hardware types

khatwanimohit: Right, I'll remove it

jonb377: Hmm good point, maybe we can mock out `xm.xla_device_hw` to `TPU`?

khatwanimohit: Yes, added the xla_device_hw mock

test/spmd/test_xla_sharding.py
```diff
@@ -452,6 +454,61 @@ def test_no_sharding(self):
...
+        'core_on_chip': 0
+    }, {
+        'core_on_chip': 0,
+        'coords': [1, 1, 0],
+        'slice_index': 1
+    }]
+    hybrid_mesh = xs.HybridMesh(
+        ici_mesh_shape=(1, 4), dcn_mesh_shape=(num_slices, 1))
+    print(hybrid_mesh.get_logical_mesh())
+    self.assertEqual(hybrid_mesh.get_logical_mesh().tolist(),
```
alanwaketan, line 509: Does this result respect the _create_device_mesh_for_nd_torus algorithm?

khatwanimohit: Yes, I have confirmed this with the jax's mesh

torch_xla/experimental/xla_sharding.py
alanwaketan: Speaking with Mohit offline. The rule is that the TPU topology is always 3D. And the inner 2D tensors have a faster ICI than the ones connect across them. Therefore, we should group the most speed demanding rank, i.e., highest rank of the mesh, to the inner 2D tensors.

alanwaketan: Now that I read more into the code. This algorithm seems quite restrict:
1. It only works with mapping a 2D or 3D logical mesh into the 3D physical mesh.
2. Then for 3D mesh, I think the logical mesh needs to be a transpose of the physical mesh.
3. Then for 2D mesh, it's just trying to map a combination of the axes into each of the dimension of the logical mesh.

After these simple rules, it then makes sure that devices that are physically close to each other are assigned close to each other in the logical mesh as well. For example, assuming the logical mesh is 2D, the devices that are in mesh[0] are always be a 2D slice of the 3D physical mesh.

If my understanding is correct, @khatwanimohit can you polish my comments and make it into the comment of this helper?

test/spmd/test_xla_sharding.py
alanwaketan: Can you make the `ici_mesh_shap=(2, 2)`? I think that can better show how the algorithm works?

alanwaketan (changes requested):
I just noticed that most of the helpers @khatwanimohit you introduced are inspired by https://github.com/google/jax/blob/bfe8acb31e04a540daad3f568239ec0e5c3f0d0f/jax/experimental/mesh_utils.py. And in fact, all those helpers have a very nice docstring to explain what the helpers are doing.

I recommend next time if you are going to import some JAX utils to PyTorch/XLA, you'd better:
1. List the source on each utils you imported.
2. Import their docstring as well. Those are really critical for the readability of the code.

Also, have you checked the licenses to make sure that you can copy code from JAX into PyTorch/XLA? If not, I can do the research for you.

test/spmd/test_xla_sharding.py
khatwanimohit: Changed ici_mesh_shape

torch_xla/experimental/xla_sharding.py
alanwaketan: You can add:
```
This is imported from JAX: https://github.com/google/jax/blob/main/jax/experimental/mesh_utils.py#L64.
```

torch_xla/experimental/xla_sharding.py
```diff
@@ -64,12 +65,203 @@ def size(self):
...
+        0]:  # multislice
+      mesh = self._create_hybrid_device_mesh(self.ici_mesh_shape,
+                                             self.dcn_mesh_shape)
+    else:
+      mesh = self._create_device_mesh(self.ici_mesh_shape)
+    device_ids = mesh.flatten()
+    super().__init__(device_ids, mesh_shape, axis_names)
+
+  def _get_physical_tpu_mesh(self, devices: Sequence[Any]) -> np.ndarray:
+    r"""Rearrange TPU devices in a slice into a physical mesh."""
```
alanwaketan, line 127: Can you add:
1.
```
This is imported from JAX: https://github.com/google/jax/blob/main/jax/experimental/mesh_utils.py#L172
```
2. The following description of the function:
```
  r"""Rearrange TPU devices in a slice into a physical mesh.

  Args:
    devices: A list of device logical ordinals in a TPU slice.

  Returns:
    A np.ndarray of device logical ordinals with shape [global_x, global_y, global_z]. On
      v2 and v3, global_z is instead cores_per_chip (i.e., 2).
  """
```

torch_xla/experimental/xla_sharding.py
```diff
@@ -64,12 +65,203 @@ def size(self):
...
+    if np.prod(mesh_shape) != len(devices):
+      raise ValueError(
+          f'Number of devices {len(devices)} must equal the product '
+          f'of mesh_shape {mesh_shape}')
+    physical_mesh = self._get_physical_tpu_mesh(devices)
+    device_mesh, assignment = self._create_device_mesh_for_nd_torus(
+        physical_mesh, mesh_shape)
+    return device_mesh
+
+  def _create_hybrid_device_mesh(self, ici_mesh_shape: Sequence[int],
```
alanwaketan, line 239: Can you add:
1.
```
This is imported from JAX: https://github.com/google/jax/blob/main/jax/experimental/mesh_utils.py#L288.
```
2. And the follow function description:
```
"""Creates a device mesh for hybrid (e.g., ICI and DCN) parallelism.

  Args:
    ici_mesh_shape: shape of the logical mesh for the faster/inner network, ordered
      by increasing network intensity, e.g. [replica, data, mdl] where mdl has
      the most network communication requirements.
    dcn_mesh_shape: shape of the logical mesh for the slower/outer network,
      in the same order as mesh_shape.

  Returns:
    A np.ndarray of device logical ordinal with ici_mesh_shape * dcn_mesh_shape as its shape
    that can be fed into HybridMesh for hybrid parallelism.
  """
```

alanwaketan (commented):
Mostly looking good to me. Thanks, @khatwanimohit.

Please address the comments on readability.

torch_xla/experimental/xla_sharding.py
```diff
@@ -64,12 +65,240 @@ def size(self):
...
+          )
+    # Flatten the assignment
+    transpose: List[int] = []
+    for x in assignment:
+      for y in x:
+        transpose.append(int(y))
+    return physical_mesh.transpose(transpose).reshape(mesh_shape), assignment
+
+  # This is imported from JAX: https://github.com/google/jax/blob/main/jax/experimental/mesh_utils.py#L231
+  def _create_device_mesh(self,
```
alanwaketan, line 236: I didn't mention this one given your logic is quite different. I suggest you can undo it.

khatwanimohit: Fixed the comment

alanwaketan (approved):
LGTM. Thanks, Mohit.

alanwaketan:
Let's have a follow up to disable the test for TPU.

You can do that by following: https://github.com/pytorch/xla/blob/master/test/test_zero1.py#L13

PR #5151, opened 2023-06-08
[test] OpenXLA migration branch test directly

ManfeiBai:


ManfeiBai:
try to update openxla to may09, and failed, I think tf's version and openxla's version match has a gap in time, so our current tf pin is at 0509, but the matched openxla's date might not be 0509, so will skip it now and try a older openxla version now

ManfeiBai:
this PR is old and will continue work in https://github.com/pytorch/xla/pull/5202

PR #5152, opened 2023-06-08
[OpenXLA] Delete unused `topology.proto` after delete XRT

ManfeiBai:
Before PyTorch migrate to pull XLA from OpenXLA, delete some unused TensorFlow deps after deprecated XRT

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -50,6 +50,7 @@ http_archive(
         "//tf_patches:thread_local_random.diff",
         "//tf_patches:topk_rewriter.diff",
         "//tf_patches:triton_filesystem.diff",
+        "//tf_patches:xla_bzl.diff",
         "//tf_patches:xplane.diff",
     ],
     strip_prefix = "tensorflow-d577af9cac504776a2d0ddbb0a445ba311aa1fea",
--- a/torch_xla/csrc/runtime/mesh_service.proto
+++ b/torch_xla/csrc/runtime/mesh_service.proto
@@ -1,7 +1,5 @@
 syntax = "proto2";
 
-import "tensorflow/core/protobuf/tpu/topology.proto";
-
 package torch_xla.runtime.service.grpc;
 
 message Device {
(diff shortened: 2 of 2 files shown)


ManfeiBai:
Hi, @will-cromar , only delete `topology.proto` will cause CPU test failed when bazel build `xrt_computation_client`,

what would be your suggestion to skip build `xrt_computation_client` in `run_xla_backend_mp /tmp/pytorch/xla/test/test_fsdp_auto_wrap.py`:
https://github.com/pytorch/xla/actions/runs/5248464052/jobs/9480595454?pr=5152

miladm:
Do we have an existing plan to update our CI to run XRT and non-XRT tests with the appropriate build? Let me know if we should scope this work.

CC @JackCaoG @will-cromar

JackCaoG:
Ok.. I think I know what's going on. so
1. We already separate the XRT and PJRT build, which is why you see that `Build & Test / Build XLA / build` passed
2. All of the python test in `Build & Test / CPU tests / test ` passed, they are purely PJRT
3. The reason you see it failed at the end is that when we are building the C++ test for PJRT, it pull in XRT dependency which failed because of lacking of topo. I think it is because our C++ test still uses some tf macro in the build file.

overall I won't worry too much about this failure because
1. it is still a XRT thing that failed
2. all python test passed on PJRT

ManfeiBai:
Thanks, Jack, that's what we met now;

should we leave this CPU/GPU test failure here and keep this PR open until we removed XRT dependency from C++ test?

or we merge this PR directly without passing CPU/GPU test, but not sure whether we could merge with `some checks were not successful`

JackCaoG:
you can't merge this pr until XRT is removed anyway. Now we know that PJRT does not depend on `topology.proto`, that's all we care. You can leave this pr as it is now.

ManfeiBai:
Thanks, Jack, will leave this PR here for now until we need it

ManfeiBai:
merged with the newest code

ManfeiBai:
since https://github.com/pytorch/xla/pull/5202 has been merged without `topology.proto`, will close this PR

PR #5160, opened 2023-06-12
Move `third_party/xla_client` to `torch_xla/csrc/runtime`

will-cromar:
`xla_client` used to be a set of patches to TensorFlow. Thanks to @stgpetrovic's work moving us to Bazel, we can now move all of our code out of the upstream `xla::` namespace and out of `third_party`. This will make it easier to identify if a symbol is from upstream XLA or PyTorch/XLA.


JackCaoG (approved):
This is great! Thanks @will-cromar

ManfeiBai (approved):
LGTM

PR #5166, opened 2023-06-13
Refine the comment for TensorCollectionBarrier in SyncTensorsGraphInt…

JackCaoG:
…ernal

The logic was introduced in https://github.com/pytorch/xla/pull/3457/files#diff-17ef64dc5ef931bed0cb0d971cf537a610e90a9b4351658d6587ff59448f0f87R1694, took me a while to remember and reason about why this `TensorCollectionBarrier` is actually needed. updated the comment so we don't need to go through this mental exercise in 6 months.


alanwaketan (approved):
LGTM.

PR #5170, opened 2023-06-13
Implement SPMDSavePlanner to take distributed checkpoints

jonb377:
This implements the [SavePlanner interface](https://github.com/pytorch/pytorch/blob/2800a04a17d727dc12353dbea4f4007682194822/torch/distributed/checkpoint/planner.py#L92) from torch.distributed.checkpoint. This implementation only directly handles sharded tensors and relies on the default planner's logic for everything else.

A high-level overview of each of the SavePlanner interface methods:

`set_up_planner`: Called with the state_dict to be checkpointed. Our implementation will split the state_dict into a sharded and unsharded portion so that we can defer to the default planner logic for the unsharded part.
`create_local_plan`: WriteItems are generated for every item in the state_dict. The default planner is used for the unsharded objects, and we generate a WriteItem for each shard of each XLAShardedTensor with non-REPLICATED sharding type.
`create_global_plan`: The coordinator process makes any global decisions for the restoration. There is no custom logic here.
`finish_plan`: The process can adjust its plan after global coordination. Again, no custom logic here.
`resolve_data`: Return the data to be written for a given WriteItem. We return the local shard for sharded tensors or relevant portions

This change also enables distributed checkpointing tests in CI.

### DIFF
--- a/torch_xla/experimental/distributed_checkpoint.py
+++ b/torch_xla/experimental/distributed_checkpoint.py
@@ -32,7 +32,6 @@ from torch.distributed.checkpoint.metadata import (
     STATE_DICT_TYPE,
 )
 from torch.distributed.checkpoint.utils import find_state_dict_object
-from torch.distributed._shard._utils import narrow_tensor_by_index
 from torch.utils._pytree import tree_map
 from torch_xla.experimental.xla_sharding import (XLAShardedTensor, XLAShard,
                                                  ShardingType)
@@ -41,6 +40,7 @@ from torch_xla.experimental._distributed_checkpoint_helpers import (
     flatten_state_dict,
     dedup_tensors,
     set_element,
+    narrow_tensor_by_index,
 )
 from typing import Any, Dict, List, Tuple, Union
 
--- a/torch_xla/experimental/distributed_checkpoint.py
+++ b/torch_xla/experimental/distributed_checkpoint.py
@@ -31,17 +31,17 @@ from torch.distributed.checkpoint.metadata import (
     Metadata,
     STATE_DICT_TYPE,
 )
-from torch.distributed.checkpoint._nested_dict import (
-    FLATTEN_MAPPING,
-    flatten_state_dict,
-)
 from torch.distributed.checkpoint.utils import find_state_dict_object
-from torch.distributed.checkpoint._dedup_tensors import dedup_tensors
-from torch.distributed.checkpoint._traverse import set_element
 from torch.distributed._shard._utils import narrow_tensor_by_index
 from torch.utils._pytree import tree_map
 from torch_xla.experimental.xla_sharding import (XLAShardedTensor, XLAShard,
                                                  ShardingType)
+from torch_xla.experimental._distributed_checkpoint_helpers import (
+    FLATTEN_MAPPING,
+    flatten_state_dict,
+    dedup_tensors,
+    set_element,
+)
 from typing import Any, Dict, List, Tuple, Union
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


torch_xla/experimental/distributed_checkpoint.py
```diff
@@ -214,6 +304,52 @@ def commit_tensor(self, read_item: ReadItem, tensor: torch.Tensor) -> None:
       self.sharded_state_dict[fqn].load_local_shards_(local_shards)
 
 
+def _create_write_item_from_indices(fqn: str, shard_index: int,
```
alanwaketan, line 307: Just noticed that you created all the helpers as the private functions of the module instead of private methods within each class. Just curious on why? You don't need to change this.

torch_xla/experimental/distributed_checkpoint.py
```diff
@@ -45,26 +53,108 @@
...
                    write_item: WriteItem) -> Union[torch.Tensor, io.BytesIO]:
-    raise NotImplemented
+    obj = self.lookup_object(write_item.index)
+    return self.transform_object(write_item, obj)
+
+  def lookup_object(self, index: MetadataIndex) -> Any:
+    if index.fqn in self.unsharded_state_dict:
+      return self.unsharded_state_dict[index.fqn]
+
+    if index.fqn not in self._local_shards:
```
alanwaketan, line 137: It seems to me that the logic of having `_local_shards` is not necessary. You are basically just using it to keep track of duplicated writes. Will that happen? For LoadPlanner, it is necessary because you only want to load the shards to XLAShardedTensor when all the shards are in the host memory and then you need to count it. I'm not sure why the SavePlanner needs this logic. Am I missing anything?

alanwaketan (commented):
Generally LGTM. Thanks for shaping the distributed checkpointing story so quick.

torch_xla/experimental/distributed_checkpoint.py
jonb377: Ah no real reason, I was just following the pattern from the upstream implementation - helpers to generate Read/WriteItems are private in the module for the DefaultPlanners.

torch_xla/experimental/distributed_checkpoint.py
jonb377: The motivation is that each call to `XLAShardedTensor::local_shards` will move all of the shards from device to CPU, and since we only need one shard at a time it's more efficient to only transfer once. I'll expand the comment in `__init__` to explain the need here.

torch_xla/experimental/distributed_checkpoint.py
alanwaketan: I still won't call myself a Python expert and that's why I ask questions on python programming patterns when they contradict to my C++ instinct.

torch_xla/experimental/distributed_checkpoint.py
alanwaketan: That makes sense to me now. Thanks for the explanation!

alanwaketan (approved):
LGTM.

jonb377:
@alanwaketan FYI I removed the distributed checkpointing test from CI, some of the APIs we're depending on aren't stable and we may need to adjust the imports once the upstream exposes them. Once they're stable, I'll add back to CI.

alanwaketan:
> @alanwaketan FYI I removed the distributed checkpointing test from CI, some of the APIs we're depending on aren't stable and we may need to adjust the imports once the upstream exposes them. Once they're stable, I'll add back to CI.

Can you be more specific? And have talked @kumpera?

alanwaketan:
> @alanwaketan Across both planners, the APIs we're taking a dependency on are:
> 
> * `torch.distributed.checkpoint._nested_dict`
> * `torch.distributed.checkpoint._dedup_tensors`
> * `torch.distributed.checkpoint._traverse`
> 
> I've spoken with @kempura, and he's looking into which he can make stable. They're pretty small helper functions for the most part, so we can reimplement here if they can't be made stable in the upstream.

Thanks for the updates. Can we dup the code now? And leave a GH issue and TODO in the code to follow up? I'm hesitate to call the MVP feature complete without a test.

jonb377:
@alanwaketan I've pulled all of the unstable dependencies into `_distributed_checkpoint_helpers.py`. I'll merge this PR once the TPU CI finishes.

alanwaketan:
Thanks, @jonb377!

PR #5174, opened 2023-06-13
[test] Replace `core:lib` with `"@org_tensorflow//tensorflow/tsl/plat…

ManfeiBai:
…form:hash",`

PR #5176, opened 2023-06-13
[OpenXLA] Delete `core:lib` in `torch_xla/csrc/BUILD`

ManfeiBai:
Before migrate to pull XLA from OpenXLA, PyTorch/XLA delete/replace deps from TensorFlow to OpenXLA

### DIFF
--- a/torch_xla/csrc/BUILD
+++ b/torch_xla/csrc/BUILD
@@ -146,7 +146,6 @@ ptxla_cc_library(
         "@org_tensorflow//tensorflow/compiler/xla/client/lib:sorting",
         "@org_tensorflow//tensorflow/compiler/xla/client/lib:svd",
         "@org_tensorflow//tensorflow/compiler/xla/stream_executor:dnn",
-        "@org_tensorflow//tensorflow/core:lib",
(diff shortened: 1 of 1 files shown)


torch_xla/csrc/BUILD
```diff
@@ -146,7 +146,6 @@ ptxla_cc_library(
         "@org_tensorflow//tensorflow/compiler/xla/client/lib:sorting",
         "@org_tensorflow//tensorflow/compiler/xla/client/lib:svd",
         "@org_tensorflow//tensorflow/compiler/xla/stream_executor:dnn",
-        "@org_tensorflow//tensorflow/core:lib",
```
JackCaoG, line 149: lol I am surprised that this is just a dead dependency..

JackCaoG (approved):

torch_xla/csrc/BUILD
ManfeiBai: me too!

and according to https://github.com/pytorch/xla/pull/5174/files, `torch_xla/csrc/runtime:util` is the only way to include all the deps from `core:lib`

PR #5177, opened 2023-06-13
Make sure dynamo give us different graph when input shape changes

JackCaoG:
Add a test to make sure when DS is enabled on dynamo it does not break the assumption that XLA has on dynamo bridge.

FYI @ezyang

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -75,14 +75,14 @@ class DynamoInferenceBasicTest(unittest.TestCase):
     xla_y = torch.randn(5, 5).to(device)
     xla_z = torch.randn(10, 10).to(device)
     self.fn_simple_dynamo(xla_x, xla_x)
-    compiel_count = met.metric_data('CompileTime')[0]
+    compile_count = met.metric_data('CompileTime')[0]
     # Execute with input with same shape should not trigger additional compilation
     self.fn_simple_dynamo(xla_y, xla_y)
-    self.assertEqual(met.metric_data('CompileTime')[0], compiel_count)
+    self.assertEqual(met.metric_data('CompileTime')[0], compile_count)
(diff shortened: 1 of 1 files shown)


ezyang (approved):

test/dynamo/test_dynamo.py
```diff
@@ -68,6 +68,25 @@ def test_simple_model(self):
...
     self.assertTrue(torch.allclose(res_cpu_3, res_xla_dynamo_3.cpu()))
 
+  def test_simple_model_with_different_input_shape(self):
+    met.clear_counters()
+    device = xm.xla_device()
+    xla_x = torch.randn(5, 5).to(device)
+    xla_y = torch.randn(5, 5).to(device)
+    xla_z = torch.randn(10, 10).to(device)
+    self.fn_simple_dynamo(xla_x, xla_x)
+    compiel_count = met.metric_data('CompileTime')[0]
```
wonjoolee95, line 78: nit: `compile` spelling 😄

wonjoolee95 (approved):

test/dynamo/test_dynamo.py
JackCaoG: nice catch!

alanwaketan (approved):
LGTM.

alanwaketan:
Not sure how this is related to DS though...

JackCaoG:
Ed was talking about enabling the DS on dynamo directly. I am worried once that happened, dynamo will start thinking about
```
resnet([18,3,512])
```
and
```
resnet([8,3,512])
```
to be the same program. Pretty much I am not sure if dynamo will guard on input shape anymore, but we need it.

alanwaketan:
If Dynamo won't guard on input shapes after DS is enabled, it will blow out my mind. Nevertheless, I appreciate the additional guards.

PR #5186, opened 2023-06-14
Add mask rcnn model

vanbasten23:
referenced:
- https://github.com/Liyang90/benchmark/tree/main/torchbenchmark/models/vision_maskrcnn

### DIFF
--- a/test/ds/coco_utils.py
+++ b/test/ds/coco_utils.py
@@ -1,0 +1,77 @@
+import torch
+from pycocotools import mask as coco_mask
+from torchvision.transforms import functional as F
+
+def convert_coco_poly_to_mask(segmentations, height, width):
+    masks = []
+    for polygons in segmentations:
+        rles = coco_mask.frPyObjects(polygons, height, width)
+        mask = coco_mask.decode(rles)
+        if len(mask.shape) < 3:
+            mask = mask[..., None]
+        mask = torch.as_tensor(mask, dtype=torch.uint8)
+        mask = mask.any(dim=2)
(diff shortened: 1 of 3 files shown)


vanbasten23:
When I run `root@t1v-n-2e71f932-w-0:/workspaces/work# python pytorch/xla/test/ds/test_mask_rcnn_models.py TestDynamicShapeMaskRCNN.test_maskrcnn_resnet50_fpn_val`, the model crashed with error:
```
RuntimeError: torch_xla/csrc/aten_xla_type.cpp:1399 : Check failed: bridge::IsXlaTensor(self) && indices_on_cpu_or_xla 
```
([full error](https://gist.github.com/vanbasten23/ecc3b0c7ed9a7fbfee8109d0f961cdee)).

Based on the stacktrace, I stuck some prints. I found out:
- Right before https://github.com/pytorch/pytorch/blob/8ce4fee68d77c6c820837d105ae36ed34bbc8167/torch/_decomp/decompositions.py#L2832, print `print('xw32 decompositions.py upsample_bilinear2d line 2790 input.device=', input.device, ', x_floor_view.device=', x_floor_view.device, ', y_floor.device=', y_floor.device)` output: `xw32 decompositions.py upsample_bilinear2d line 2790 input.device= xla:0 , x_floor_view.device= xla:0 , y_floor.device= xla:0`.

That means `the tensor` in `XLANativeFunctions::index(const at::Tensor& self, ...` used to be on xla device, but later when we dispatch the call to torch_xla, the device info is lost.

PR #5187, opened 2023-06-14
Update package version in ansible build

will-cromar:


JackCaoG (approved):
Thanks!

mateuszlewko:
FYI, this is the default value for package version, but all build triggers override it. For nighty build triggers you can change it here: https://github.com/pytorch/xla/blob/master/infra/tpu-pytorch-releases/artifacts.auto.tfvars#L1
For releases here: https://github.com/pytorch/xla/blob/master/infra/tpu-pytorch-releases/artifacts.auto.tfvars#L24

will-cromar:
Thanks for the heads up. Sent https://github.com/pytorch/xla/pull/5192

Is there a good way for us to pull this version from PyTorch's [version.txt](https://github.com/pytorch/pytorch/blob/main/version.txt) instead of having it in multiple places?

mateuszlewko:
It's possible to set an Ansible variable based on file contents, but are
you sure we want that?
I see they set something like "2.1.0a0". We later set it as an env variable
during build here:
https://github.com/pytorch/xla/blob/master/infra/ansible/config/env.yaml#LL30C5-L30C47
.

If it's okay, I can set it up (soonish).

PR #5191, opened 2023-06-15
Fix import in SPMD resnet

jonb377:
The SPMD resnet training test needs to import the new runtime package.

Also, I removed an outdated flag to enable virtual device.

### DIFF
--- a/test/spmd/test_train_spmd_imagenet.py
+++ b/test/spmd/test_train_spmd_imagenet.py
@@ -32,9 +32,6 @@ MODEL_OPTS = {
         'nargs': '+',
         'default': [],
     },
-    '--use_virtual_device': {
-        'action': 'store_true',
-    },
     '--use_gradient_checkpointing': {
         'action': 'store_true',
     }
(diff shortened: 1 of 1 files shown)


test/spmd/test_train_spmd_imagenet.py
```diff
@@ -67,6 +64,7 @@
 from torch_xla.distributed.fsdp.wrap import (recursive_wrap,
                                              transformer_auto_wrap_policy)
 from torch_xla.distributed.fsdp.utils import checkpoint_module
+from torch_xla import runtime as xr
```
JackCaoG, line 67: hmm, why do we need to import if there is no code using it below?

jonb377: We use it for building the device mesh: `    num_devices = xr.global_device_count()`. When migrating `pjrt` to `runtime`, the references were updated but the import was missed

JackCaoG (approved):
Thanks!

JackCaoG:
I guess we can skip the CI, I don;t think this test is part of the CI(maybe it should be part of the TPU CI? running for couple steps for example).

PR #5193, opened 2023-06-15
Enable dynamism for torch.sub

vanbasten23:


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2969,6 +2971,13 @@ at::Tensor XLANativeFunctions::sub(const at::Tensor& self,
                                    const at::Tensor& other,
                                    const at::Scalar& alpha) {
   TORCH_LAZY_FN_COUNTER("xla::");
+  // Currently, we disallow the case when both operands contain dynamic
+  // dimensions. This is consistent with PyTorch's behavior.
+  XLA_CHECK(!(tensor_has_dym_dim(self) && tensor_has_dym_dim(other)))
+      << "Both operands of torch.add cannot have dynamic dimensions at the "
```
JackCaoG, line 2977: typo here, `torch.sub`

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2473,8 +2473,22 @@ std::tuple<XLATensorPtr, XLATensorPtr> std_mean(const XLATensorPtr& input,
...
+    constant = XLAGraphExecutor::Get()->GetIrValueForScalar(
+        alpha, other->shape(), logical_element_type, input->GetDevice());
+  } else {
+    SymIntElements sym_int_elements(other->GetIrValue());
+    xla::PrimitiveType primitive_type =
+        logical_element_type
+            ? MakeXlaPrimitiveType(*logical_element_type, &(input->GetDevice()))
+            : other->shape().get().element_type();
+    constant = XLAGraphExecutor::Get()->GetIrValueForScalar(
+        alpha, sym_int_elements, primitive_type, input->GetDevice());
```
JackCaoG, line 2489: curious why function signature here different from the static version of `GetIrValueForScalar`. is it because `sym_int_elements` does not contain the `primitive_type`? If every place we call this new function(there might be a lot) we have to explicitly run 
```
    xla::PrimitiveType primitive_type =
        logical_element_type
            ? MakeXlaPrimitiveType(*logical_element_type, &(input->GetDevice()))
            : other->shape().get().element_type();
```
it is a bit annoying, I'd rather encapsulate that logic in the dynamic version of the `GetIrValueForScalar` so it is more concise.

vanbasten23: good point. Fixed it.

JackCaoG (approved):
Thanks!

PR #5197, opened 2023-06-16
[OpenXLA] Replace `core:lib` in `util` from TensorFlow to tensorflow/compiler/xla

ManfeiBai:
Before migrate to pull XLA from OpenXLA, PyTorch/XLA delete/replace deps from TensorFlow to OpenXLA/TSL


.bazelrc
```diff
@@ -167,7 +167,7 @@ build:linux --config=dynamic_kernels
 # the monolitih build because otherwise there are missing dependencies and
 # linking fails.
 build --define framework_shared_object=false
-build --define tsl_protobuf_header_only=true
+build --define tsl_protobuf_header_only=false
```
miladm, line 170: @ManfeiBai do we know why this change resolves the recent failures?

JackCaoG (approved):

.bazelrc
ManfeiBai: the recent failures is due to missing proto after delete `core:lib`, and `--define tsl_protobuf_header_only=false` will enable PyTorch/XLA build with these protos again

PR #5203, opened 2023-06-16
Enable source line info in StableHLO

lsy323:
Enable source line info in StableHLO


torch_xla/csrc/runtime/stablehlo_helper.cc
```diff
@@ -22,7 +22,9 @@ static std::string getHloModuleStr(const xla::HloModuleProto* proto) {
 static std::string getMlirModuleStr(mlir::ModuleOp& mlir_module) {
   std::string txt_mlir_module;
   llvm::raw_string_ostream os{txt_mlir_module};
-  mlir_module.print(os);
+  // Enable Debug Info to include source line info in the StableHLO dump.
+  mlir_module.print(
```
JackCaoG, line 26: I guess if we don;t enable `XLA_HLO_DEBUG=1` this will be a no-op? I am trying to think whether we need another flag like `XLA_SHLO_DEBUG` but that seems to be an overkill.

lsy323: Yes, that makes sense.

I think we should turn on the debug info in MLIR Module dump when `XLA_HLO_DEBUG=1`. I don't think we need another flag, since there isn't a use case that we want HLO to have meta data but do not want those in StableHLO : )

JackCaoG:
Can we add a test case for this?

lsy323:
> 

Sure, let me add one.

lsy323:
It's pretty weird that the source line info isn't showing in the test script. However, it's showing in my local resnet18 test script. Need to take a closer look.

test/run_tests.sh
```diff
@@ -184,6 +184,7 @@ function run_xla_op_tests {
   run_use_bf16 "$CDIR/test_data_type.py"
   run_xla_ir_debug "$CDIR/test_env_var_mapper.py"
   run_xla_hlo_debug "$CDIR/test_env_var_mapper.py"
+  run_xla_hlo_debug "$CDIR/test_stablehlo_dump.py"
```
JackCaoG, line 187: let's create a dir called `stablehlo` or `inference` and put tests under that dir, there are too many tests under `test/` dir right now

test/test_stablehlo_dump.py
```diff
@@ -14,7 +14,6 @@ def test_simple(self):
     z = x + y
     # Example usage of dumping StableHLO given output tensors
     stablehlo = xm.get_stablehlo([z])
-    self.assertEqual(stablehlo.count("stablehlo.multiply"), 1)
```
JackCaoG, line 17: why is this removed?

torch_xla/core/xla_model.py
```diff
@@ -852,6 +852,8 @@ def get_stablehlo(tensors=None):
   For inference graph, it is recommended to pass the model outputs to `tensors`.
   For training graph, it is not straightforward to identify the "outputs". Using empty `tensors` is recommended.
 
+  To enable source line info in StableHLO, please set env var XLA_HLO_DEBUG=1.
```
JackCaoG, line 855: can we add this function to https://github.com/pytorch/xla/blob/master/docs/source/index.rst#xla_model as well?

test/run_tests.sh
lsy323: Sure, that's a good idea, let me create a new folder, maybe just use `stablehlo`

test/test_stablehlo_dump.py
lsy323: Typo, the graph shouldn't have multiply op.

torch_xla/core/xla_model.py
lsy323: Yes, let me add it.

JackCaoG (approved):
Thanks! @lsy323

PR #5204, opened 2023-06-17
Add tesor debugging util API

JackCaoG:
I think having an internal API to dump XLATensor info is useful in debugging, it can dump something similar to 
```
XLATensor {
TensorID: 5
Device: CPU:0
XLA Shape: f32[5]
IR: [] aten::add, xla_shape=f32[5]{0}
XLAData: None
Tensor on host: None
}
```

I didn't include view since I assume view will not be there after we clean up the functionization pass.


JackCaoG:
I can also add a test on Monday, context is I am debugging a SPMD issue and I can't figure out why XLAData is a placeholder when it shouldn't, so I created this API

JackCaoG:
@wonjoolee95 @alanwaketan This should be ready to review now.

alanwaketan (approved):
LGTM.

wonjoolee95 (approved):

PR #5202, opened 2023-06-16
[OpenXLA] Migrate to pull XLA from OpenXLA

ManfeiBai:
**[Description]**
PyTorch/XLA migrate to pull XLA from OpenXLA by replacing TensorFlow with OpenXLA after deprecating XRT usage, and replace TensorFlow-pin with OpenXLA-pin to May09

**[What has changed]**
1. change from `@org_tensorflow//tensorflow/compiler/xla/...` to `@xla//xla/...`
2. change from `@org_tensorflow//tensorflow/tsl/...` to `@tsl//tsl/...`
3. change from `#include "tensorflow/compiler/xla/...` to `#include "xla/...`
4. change from `#include "tensorflow/tsl/...` to `#include "tsl/...`
5. change from pull XLA from [TensorFlow](https://github.com/tensorflow/tensorflow) with `tf_patches` to pull XLA from [OpenXLA](https://github.com/openxla/xla) with `openxla_patches`

**[Start to do OpenXLA-pin]**
After this PR, we will update OpenXLA-pin instead of TensorFlow-pin. We used to update `WORKSPACE` and `tf_patches` to update used TensorFlow's version, aka TensorFlow-pin update. After this PR, we will update OpenXLA-pin by updating `WORKSPACE` and `openxla_patches` to update used OpenXLA's version.

Thanks, @miladm @JackCaoG @will-cromar @puneith @ddunl @stgpetrovic @wonjoolee95 @lsy323 @aspirido
According to https://github.com/pytorch/xla/pull/5202#discussion_r1237741669, we should give AWS folks a heads up when we merge this pr.

PR #5209, opened 2023-06-20
[test][OpenXLA]Newbranchopxlatest

ManfeiBai:


ManfeiBai:
since all tests passed, will close and continue track in https://github.com/pytorch/xla/pull/5202

PR #5214, opened 2023-06-20
Add function to generate stablehlo based callable from pytorch model

qihqi:
Added function
`torch_xla.experimental.stablehlo_saved_model.export_pytorch_model`. This function will take a pytorch Module and convert it into stablehlo bytecode.

### DIFF
--- a/torch_xla/experimental/stablehlo_saved_model.py
+++ b/torch_xla/experimental/stablehlo_saved_model.py
@@ -1,0 +1,119 @@
+import copy
+import shutil
+import os
+import re
+import numpy as np
+import torch
+from torch import nn
+import torch_xla
+from torch_xla.core import xla_model as xm
+import tensorflow as tf
+import torchvision
+import torch._dynamo as torchdynamo
+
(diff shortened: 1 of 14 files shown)


torch_xla/experimental/stablehlo_saved_model.py
```diff
@@ -0,0 +1,126 @@
...
+  pos_to_orig_pos = {}
+  pos_to_param = {}
+  for hlo_input_pos, tensor_id in enumerate(graph_input_tensor_ids):
+    if tensor_id in input_ids:  # this is input
+      pos_to_orig_pos[hlo_input_pos] = input_ids[tensor_id]
+    else:
+      pos_to_param[hlo_input_pos] = graph_input_xla_values[
+          hlo_input_pos].detach().cpu().numpy()
+
+  output_numpy = output.cpu().detach().numpy()
```
JackCaoG, line 116: calling `output.cpu()` will trigger an execution(since we moved the output to cpu). Can we use `output.size` and `output.dtype` directly?

PR #5216, opened 2023-06-21
Add function to generate stablehlo based callable from pytorch model

qihqi:
Added function
`torch_xla.experimental.stablehlo_saved_model.export_pytorch_model`. This function will take a pytorch Module and convert it into stablehlo bytecode.

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -838,7 +838,7 @@ def mark_step(wait=False):
   torch_xla._XLAC._set_all_reduce_token(devctx.device, None)
 
 
-def get_stablehlo(tensors=None):
+def get_stablehlo(tensors=None) -> str:
   """Get StableHLO for the computation graph in string format.
 
   If `tensors` is not empty, the graph with `tensors` as outputs will be dump.
@@ -860,7 +860,31 @@ def get_stablehlo(tensors=None):
   if tensors is None:
     tensors = []
   return torch_xla._XLAC._get_stablehlo(
-      tensors, torch_xla._XLAC._xla_get_default_device(), [])
+      tensors, torch_xla._XLAC._xla_get_default_device(), [],
+      False).decode('utf-8')
+
+
+def get_stablehlo_bytecode(tensors=None) -> bytes:
+  """Get StableHLO for the computation graph in bytecode format.
+
+  If `tensors` is not empty, the graph with `tensors` as outputs will be dump.
+  If `tensors` is empty, the whole computation graph will be dump.
+  TODO(lsy323): When `tensors` is empty, the some intermediate tensors will also be
+  dump as outputs. Need further investigation.
+
+  For inference graph, it is recommended to pass the model outputs to `tensors`.
+  For training graph, it is not straightforward to identify the "outputs". Using empty `tensors` is recommended.
+
+  Args:
+    tensors (list[torch.Tensor], optional): The tensors contained in the StableHLO graph.
+
+  Returns:
+    StableHLO Module in bytecode format.
+  """
+  if tensors is None:
+    tensors = []
+  return torch_xla._XLAC._get_stablehlo(
+      tensors, torch_xla._XLAC._xla_get_default_device(), [], True)
 
 
 def wait_device_ops(devices=[]):
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -852,7 +852,7 @@ def get_stablehlo(tensors=None) -> str:
   To enable source line info in StableHLO, please set env var XLA_HLO_DEBUG=1.
 
   Args:
-    tensors (list[torch.Tensor], optional): The tensors contained in the StableHLO graph.
+    tensors (list[torch.Tensor], optional): Tensors that represent the output/root of the StableHLO graph.
 
   Returns:
     StableHLO Module in string format.
@@ -876,7 +876,7 @@ def get_stablehlo_bytecode(tensors=None) -> bytes:
   For training graph, it is not straightforward to identify the "outputs". Using empty `tensors` is recommended.
 
   Args:
-    tensors (list[torch.Tensor], optional): The tensors contained in the StableHLO graph.
+    tensors (list[torch.Tensor], optional): Tensors that represent the output/root of the StableHLO graph.
 
   Returns:
     StableHLO Module in bytecode format.
--- a/torch_xla/csrc/runtime/BUILD
+++ b/torch_xla/csrc/runtime/BUILD
@@ -351,6 +351,7 @@ cc_library(
         "@org_tensorflow//tensorflow/compiler/xla/translate/hlo_to_mhlo:hlo_to_mlir_hlo",
         "@org_tensorflow//tensorflow/compiler/xla/translate/mhlo_to_hlo:mlir_hlo_to_hlo",
         "@org_tensorflow//tensorflow/compiler/xla/mlir_hlo:all_passes",
+        "@stablehlo//:stablehlo_serialization",
     ],
 )
 
--- a/torch_xla/experimental/stablehlo_saved_model.py
+++ b/torch_xla/experimental/stablehlo_saved_model.py
@@ -1,0 +1,119 @@
+import copy
+import shutil
+import os
+import re
+import numpy as np
+import torch
+from torch import nn
+import torch_xla
+from torch_xla.core import xla_model as xm
+import tensorflow as tf
+import torchvision
+import torch._dynamo as torchdynamo
+
+from tensorflow.compiler.tf2xla.python import xla as tfxla
+
+from typing import Tuple, Type, Callable
+
+import sys
+
+
+class SHLOModel:
+
+  def __init__(self, model_bytecode, pos_to_orig_pos, pos_to_param,
+               output_shape, output_dtype):
+    self.model_bytecode = model_bytecode
+    self.pos_to_orig_pos = pos_to_orig_pos
+    self.pos_to_param = pos_to_param
+    self._total_number = len(pos_to_orig_pos) + len(pos_to_param)
+
+    self.tout = (output_dtype,)
+    self.sout = (tuple(output_shape),)
+
+  def __call__(self, *args):
+
+    call_args = []
+    for i in range(self._total_number):
+      if i in self.pos_to_orig_pos:
+        call_args.append(args[self.pos_to_orig_pos[i]])
+      else:
+        call_args.append(self.pos_to_param[i])
+
+    return tfxla.call_module(
+        tuple(call_args),
+        version=5,
+        Tout=self.tout,  # dtype information
+        Sout=self.sout,  # Shape information
+        function_list=[],
+        platforms=('CPU',),
+        module=self.model_bytecode,
+    )[0]
+
+
+def export_torch_model(model: torch.nn.Module,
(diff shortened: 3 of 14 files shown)


test/test_stablehlo_dump.py
```diff
@@ -24,11 +26,40 @@ def test_resnet18(self):
...
     data = torch.randn(4, 3, 224, 224, device=device)
     output = xla_resnet18(data)
-    stablehlo = xm.get_stablehlo([output])
+    stablehlo = xm.get_stablehlo([output]).decode('utf-8')
     self.assertEqual(stablehlo.count("stablehlo.convolution"), 20)
-    stablehlo = xm.get_stablehlo()
+    stablehlo = xm.get_stablehlo().decode('utf-8')
     self.assertEqual(stablehlo.count("stablehlo.convolution"), 20)
 
+  def test_resnet18_inference(self):
```
JackCaoG, line 34: let's keep this test as a `stablehlo_dump` test, and put inference/saved_model spcified bits to a different tests. I was talking to @lsy323 about creating a new dir called `stablehlo` or `inference` under `/test/`

JackCaoG:
Thanks, I will take a look tmr.

torch_xla/experimental/stablehlo_saved_model.py
```diff
@@ -0,0 +1,110 @@
...
+import numpy as np
+import torch
+from torch import nn
+import torch_xla
+from torch_xla.core import xla_model as xm
+import tensorflow as tf
+import torchvision
+import torch._dynamo as torchdynamo
+
+from tensorflow.compiler.tf2xla.python import xla as tfxla  # type: ignore[import]
```
lsy323, line 14: Why is `# type: ignore[import]` needed here?

lsy323 (approved):
Thanks for adding StableHLO Bytecode serialization 👍 Overall LGTM

lsy323:
I saw StableHLO test is failing in CI. It's because `get_stablehlo` now returns `bytes`, probably call [bytes.decode](https://docs.python.org/3.10/library/stdtypes.html?highlight=decode#bytes.decode) to decode the bytes in the [test](https://github.com/pytorch/xla/blob/b3f4c7bc1e9f812069770c83e38e5bd2dc20aef6/test/stablehlo/test_stablehlo_dump.py#L16)?

JackCaoG:
I am not an python API expert, but would it be OK to let `get_stablehlo` to return `string` for readable and `py::byte` for `bytecode`? or this is impossible due to the best python practice or limitation of the pybind. The alternative is to have 2 different api.

Asking user to call `bytes.decode` when they want the string format is a bit annoying.

torch_xla/experimental/stablehlo_saved_model.py
```diff
@@ -0,0 +1,110 @@
...
+    on the sample input, or inputs sufficiently similar* to sample input,  
+    it is will to return same result as the original model.
+
+    * sufficiently similar input because this function will use tracing to extract the model operations
+      so it might specialize on the shapes of the sample input.
+    
+    For now, model has to only take Tensors as input and has to return 1 tensor as output.
+
+    """
+  xm.mark_step()
```
JackCaoG, line 76: nit: add a comment here to explain why a mark_step is needed(something along the line of "Materialize the computation before this call to make sure that StableHLO captured below only reflect the model passed in")

torch_xla/experimental/stablehlo_saved_model.py
```diff
@@ -0,0 +1,110 @@
...
+  sample_outputs = model(*sample_inputs)
+
+  device = xm.xla_device()
+  model.to(device=device)
+  sample_inputs_lazy = tuple(map(lambda x: x.to(device=device), sample_inputs))
+  input_ids = {
+      torch_xla._XLAC._xla_get_tensor_id(tensor): i
+      for i, tensor in enumerate(sample_inputs_lazy)
+  }
+  output = model(*sample_inputs_lazy)
```
JackCaoG, line 90: You want to call `clear_pending_irs` similar to https://github.com/pytorch/xla/blob/b3f4c7bc1e9f812069770c83e38e5bd2dc20aef6/torch_xla/core/dynamo_bridge.py#L299-L305

otherwise you will leave a pending graph on device which might mess up the future execution and wasting compute.

torch_xla/experimental/stablehlo_saved_model.py
```diff
@@ -0,0 +1,110 @@
...
+
+class SHLOModel:
+
+  def __init__(self, model_bytecode, pos_to_orig_pos, pos_to_param,
+               sample_output):
+    self.model_bytecode = model_bytecode
+    self.pos_to_orig_pos = pos_to_orig_pos
+    self.pos_to_param = pos_to_param
+    self._total_number = len(pos_to_orig_pos) + len(pos_to_param)
+    output_tf = tf.convert_to_tensor(sample_output)
```
JackCaoG, line 29: curious why sample_output is needed. if they only need to know the shape, we can just create a fake tensor using `output.size` and `output.dtype`. Running the model on CPU once just to get the output is a bit wasteful

JackCaoG (commented):
mostly LGTM, minor nits

torch_xla/experimental/stablehlo_saved_model.py
qihqi: done

torch_xla/experimental/stablehlo_saved_model.py
qihqi: done

qihqi:
> I am not an python API expert, but would it be OK to let `get_stablehlo` to return `string` for readable and `py::byte` for `bytecode`? or this is impossible due to the best python practice or limitation of the pybind. The alternative is to have 2 different api.
> 
> Asking user to call `bytes.decode` when they want the string format is a bit annoying.

Oki I made 2 distinct APIs instead. Usually a function's return type should be stable (although serialization related stuff sometimes gets an exception: i.e. `json.loads`). PTAL

torch_xla/experimental/stablehlo_saved_model.py
qihqi: removed

torch_xla/experimental/stablehlo_saved_model.py
qihqi: Sure. Although the user probably won't know output shape / dtype without first running it.

test/inference/test_stablehlo.py
```diff
@@ -0,0 +1,49 @@
+import torch_xla
```
JackCaoG, line 1: you would need to add this test to `test/run_test.sh` otherwise it won't get run

torch_xla/core/xla_model.py
```diff
@@ -860,7 +860,31 @@ def get_stablehlo(tensors=None):
...
+  If `tensors` is not empty, the graph with `tensors` as outputs will be dump.
+  If `tensors` is empty, the whole computation graph will be dump.
+  TODO(lsy323): When `tensors` is empty, the some intermediate tensors will also be
+  dump as outputs. Need further investigation.
+
+  For inference graph, it is recommended to pass the model outputs to `tensors`.
+  For training graph, it is not straightforward to identify the "outputs". Using empty `tensors` is recommended.
+
+  Args:
+    tensors (list[torch.Tensor], optional): The tensors contained in the StableHLO graph.
```
JackCaoG, line 879: nit, can we update this along with comment above to 
```
tensors (list[torch.Tensor], optional): Tensors that represent the output/root of the StableHLO graph.
```

torch_xla/core/xla_model.py
```diff
@@ -860,7 +860,31 @@ def get_stablehlo(tensors=None):
   if tensors is None:
     tensors = []
   return torch_xla._XLAC._get_stablehlo(
-      tensors, torch_xla._XLAC._xla_get_default_device(), [])
+      tensors, torch_xla._XLAC._xla_get_default_device(), [],
+      False).decode('utf-8')
+
+
+def get_stablehlo_bytecode(tensors=None) -> bytes:
```
JackCaoG, line 867: Thanks! can we add the new api to https://github.com/pytorch/xla/blob/master/docs/source/index.rst#xla_model ?

torch_xla/csrc/runtime/BUILD
```diff
@@ -364,6 +364,7 @@ cc_library(
         "@org_tensorflow//tensorflow/compiler/xla/translate/hlo_to_mhlo:hlo_to_mlir_hlo",
         "@org_tensorflow//tensorflow/compiler/xla/translate/mhlo_to_hlo:mlir_hlo_to_hlo",
         "@org_tensorflow//tensorflow/compiler/xla/mlir_hlo:all_passes",
+        "@stablehlo//:stablehlo_serialization",
```
JackCaoG, line 367: @ManfeiBai FYI since this dependency is new.

torch_xla/experimental/stablehlo_saved_model.py
JackCaoG: I think you can also get `output_shape` and `output_dtype` by doing `output.dtype` and `otuput.size()` here

qihqi: would that materialize `output? and thus lose the LazyIR stored in it?

JackCaoG: nope, shape and dtype are generated during tracing, it won't trigger execution.

torch_xla/csrc/runtime/BUILD
ManfeiBai: if [[stablehlo imported in TF](https://github.com/tensorflow/tensorflow/blob/1ec9f2e7163aff9c959ed0f478497ff475a99c16/third_party/stablehlo/workspace.bzl)] is the total same as [stablehlo imported in OpenXLA](https://github.com/openxla/xla/tree/main/third_party/stablehlo), this new dep `"@stablehlo//:stablehlo_serialization",` could be used after we migrate to pull XLA from OpenXLA, @lsy323 

my thought is this would be very like `tsl` because `tsl` is a third party included by OpenXLA too

cc @JackCaoG

lsy323: @ManfeiBai Good catch, you are right. It shouldn't be a problem when migrating to OpenXLA

ManfeiBai: testing PR: https://github.com/pytorch/xla/pull/5237

ManfeiBai: since build successed on tpu: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/46f7838f-af3f-4681-8f2e-04c5b46ea6e9?project=tpu-pytorch&e=13802955&jsmode=o&mods=allow_workbench_image_override

we could build with this new dep `@stablehlo//:...` after migrate to pull XLA from OpenXLA

ManfeiBai (approved):
LGTM

torch_xla/experimental/stablehlo_saved_model.py
JackCaoG: Took a another look at this pr, once we fixed this comment, this pr lgtm

qihqi: done

JackCaoG (approved):
Thanks @qihqi !

PR #5227, opened 2023-06-21
[test][OpenXLA]Newbranchopxlaraiseerrordelete rt shard ordinal

ManfeiBai:


ManfeiBai:
since pass tests in https://github.com/pytorch/xla/pull/5229, will apply this change

PR #5229, opened 2023-06-21
[test][OpenXLA] Newbranchopxlametrics

ManfeiBai:


ManfeiBai:
since passed tests on TPU, will apply this change because the other tests didn't start without any known reason

PR #5230, opened 2023-06-22
Flatten `Computation` classes

will-cromar:
- Remove the wrapper `torch_xla::Computation` class now that we can mix PyTorch and OpenXLA in the same build.
- Remove unused fields and make them explicitly unimplemented.
- Always set name and hash, even if name is unused.
- Update comments from `computation.h` (please make sure my understanding is correct).

To be honest, I still don't think the difference between the three uses of Computation is clear, so I'll likely follow up with more refactoring to make the differences explicit.


JackCaoG:
Let me take a look now..

torch_xla/csrc/runtime/computation_client.h
```diff
@@ -64,31 +70,67 @@ class ComputationClient {
...
+    // 3. To represent a computation that is already compiled. In this case
+    //    name_ and hash_ are not required. Computation will be a wrapper around
+    //    an executable, PjRtComputationClient::PjRtComputation in our case. It
+    //    is not ideal to use same class for 3 different purposes but this is
+    //    the path took by upstream ltc.
     Computation(xla::XlaComputation computation,
                 std::vector<std::string> devices)
-        : computation_(std::move(computation)), devices_(std::move(devices)) {
-      program_shape_ = ConsumeValue(computation_.GetProgramShape());
-    }
```
JackCaoG, line 85: any reason `program_shape_` is only being inited on one of the constructors?

torch_xla/csrc/runtime/computation_client.h
```diff
@@ -107,13 +149,41 @@ class ComputationClient {
...
+    int parameters_size() const override {
+      return program_shape().parameters_size();
+    }
+
+    const std::vector<torch::lazy::Shape>& parameter_shapes() const override {
+      XLA_ERROR() << "Unimplemented";
+    }
+
+    const std::vector<std::string>& parameter_names() const override {
+      return program_shape().parameter_names();
```
JackCaoG, line 163: seems like we can't alwasy reply on that program_shape exists

torch_xla/csrc/runtime/computation_client.h
```diff
@@ -107,13 +149,41 @@ class ComputationClient {
...
     const xla::ProgramShape& program_shape() const { return program_shape_; }
 
+    const torch::lazy::hash_t& hash() const { return hash_; }
+
+    int parameters_size() const override {
+      return program_shape().parameters_size();
+    }
+
+    const std::vector<torch::lazy::Shape>& parameter_shapes() const override {
+      XLA_ERROR() << "Unimplemented";
```
JackCaoG, line 159: are these `XLA_ERROR` because they are not currently being used?

JackCaoG (commented):
mostly lgtm

torch_xla/csrc/runtime/computation_client.h
will-cromar: All of the other constructors delegate to the one that sets `program_shape_` and `hash_`.

https://secure.wikimedia.org/wikipedia/en/wiki/C++11#Object_construction_improvement

torch_xla/csrc/runtime/computation_client.h
will-cromar: Correct. There was not even any logic before to convert the XLA parameter shapes to lazy shapes in the previous implementation. We just returned a default-initialized object, which seems more dangerous to me.

Upstream defines this as a pure virtual function, but I could not find any usage.

torch_xla/csrc/runtime/computation_client.h
will-cromar: All of the constructors initialize program shape, as long as its available from the `XlaComputation`. See my other comment.

JackCaoG (approved):

alanwaketan (approved):
This makes our code a lot cleaner!

PR #5235, opened 2023-06-22
Revert absl::Status patch

will-cromar:
:crossed_fingers:


.github/workflows/_build.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Download docker image from GCR
         shell: bash
-        run: docker pull "${GCR_DOCKER_IMAGE}"
+        run: docker pull us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:tpu
```
JackCaoG, line 66: we should use cuda one right? since CI is also running gpu test

.github/workflows/_build.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Download docker image from GCR
         shell: bash
-        run: docker pull "${GCR_DOCKER_IMAGE}"
+        run: docker pull us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.8_cuda_11.8
```
JackCaoG, line 66: @lsy323 FYI, since you were planning do similar thing

will-cromar:
I was hoping I could easily drop in our dev image, but we probably need to do a larger rework of our CIs.

PR #5242, opened 2023-06-23
[test][JUST FOR TEST][OpenXLA] rebase

ManfeiBai:


ManfeiBai:
since all tests passed, will update #5202 according to this PR

PR #5243, opened 2023-06-23
[test][openxla][JUST FOR TEST] test and fix cycle

ManfeiBai:


ManfeiBai:
test without rebase has failed locally, will not try this change and patch

PR #5245, opened 2023-06-23
[test][openxla][JUST FOR TEST] no cycle without rebase

ManfeiBai:


ManfeiBai:
test failed locally, will stop this test and patch

PR #5256, opened 2023-06-27
Jack cao g/fix spmd buff is null

JackCaoG:
Currently resnet50 batch sharding will fail with  `buffer is null`. The root cause is that virtual device will try to delay the transfer of the `at::Tensor` to device until we know the sharding annotation(when mark_step is called, all tensor without sharding will be considered as replicated).

However for a case like
```
xt2 = xt1 / 0.5
```
we will transfer scalar `0.5` to the device since it is part of the computation. If virtual device kicks in, we will create a data placeholder for `0.5`. The problem is there is no `XLATensor` corresponding to the `scalar 0.5`. During the `mark_step`, we will just ignore this place holder which result in this `buffer_is_null` error.

I think the long term fix is to only use `virtual_device` for value that actually has a `XLATensor`, I need to change the function signature to make caller explicitly about the `at::tensor` it is trying to send is a actual pytorch tensor or just a temp scalar.

### DIFF
--- a/test/spmd/test_xla_virtual_device.py
+++ b/test/spmd/test_xla_virtual_device.py
@@ -77,6 +77,16 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
     self.assertLess(outbound_with_virtual_device,
                     2 * xt1.nelement() * xt1.element_size())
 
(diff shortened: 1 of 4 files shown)


alanwaketan (approved):
LGTM.

PR #5259, opened 2023-06-27
Disable Bazel remote cache for forked PR

lsy323:
### Background
Since forked PRs doesn't have access to Github Repository secrets, the gcloud credentials passed by secrets cannot be picked up by forked PRs, causing the CI to fail due to empty gcloud credential files.

Thanks to @stgpetrovic pointing out, whenever the self-hosted workers are used in CI, the workers will have access to the remote cache thus the credential won't be needed at that time.

cc @mateuszlewko who is working on using self-hosted workers in TPU CI.

### Temporary Solution
Since gcloud credential is only required when Bazel remote cache is enabled, the size of the credential file is checked in `setup.py`, if the credential is empty (Actually the credential file is 1 byte), we assume it's triggered from a forked PR and skip the remote cache.

### Next Step
After the self-hosted workers are setup for CPU/GPU CI, we can remove the gcloud credential in `setup.py`.

### DIFF
--- a/scripts/run_bazel_coverage.sh
+++ b/scripts/run_bazel_coverage.sh
@@ -2,11 +2,20 @@
 export XRT_DEVICE_MAP="CPU:0;/job:localservice/replica:0/task:0/device:XLA_CPU:0"
 export XRT_WORKERS="localservice:0;grpc://localhost:40934"
 export XLA_EXPERIMENTAL="nonzero:masked_select"
-bazel coverage --config=remote_cache --remote_default_exec_properties=cache-silo-key=cache-silo-coverage //...
+
+if [ ! -z "$GCLOUD_SERVICE_KEY_FILE" ]; then
+  file_size=$(stat -c%s "$GCLOUD_SERVICE_KEY_FILE")
+  if [ "$file_size" -le 1 ]; then
+    BAZEL_REMOTE_CACHE_CONFIG=""
+  fi
+fi
+BAZEL_REMOTE_CACHE_CONFIG="--config=remote_cache --remote_default_exec_properties=cache-silo-key=cache-silo-coverage"
+
+bazel coverage $BAZEL_REMOTE_CACHE_CONFIG //...
 cp "$(bazel info output_path)/_coverage/_coverage_report.dat" /tmp/cov_xrt.dat
 
 export PJRT_DEVICE="CPU"
-bazel coverage --config=remote_cache --remote_default_exec_properties=cache-silo-key=cache-silo-coverage //test/...
+bazel coverage $BAZEL_REMOTE_CACHE_CONFIG //test/...
 cp "$(bazel info output_path)/_coverage/_coverage_report.dat" /tmp/cov_pjrt.dat
 
 # requires `apt-get install lcov`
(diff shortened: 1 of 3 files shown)


lsy323:
cc @JackCaoG @will-cromar

JackCaoG:
Thanks, let me know when this pr is ready for review.

lsy323:
Hi @stgpetrovic, before self-hosted workers are set up, I think we should disable remote cache for PRs from forks. The size of the credential file from forks is 1 byte, based on this fact, I made the changes in the related files. Would you please take a pass? Feel free to take it over if you know a more elegant way to do it. Thanks

lsy323:
I tried to update `setup.py`, the build passed, but test failed since remote cache config is also in `test.sh` and test coverage files. Add the check there as well.

scripts/run_bazel_coverage.sh
```diff
@@ -2,11 +2,20 @@
...
 export XLA_EXPERIMENTAL="nonzero:masked_select"
-bazel coverage --config=remote_cache --remote_default_exec_properties=cache-silo-key=cache-silo-coverage //...
+
+if [ ! -z "$GCLOUD_SERVICE_KEY_FILE" ]; then
+  file_size=$(stat -c%s "$GCLOUD_SERVICE_KEY_FILE")
+  if [ "$file_size" -le 1 ]; then
+    BAZEL_REMOTE_CACHE_CONFIG=""
+  fi
+fi
+BAZEL_REMOTE_CACHE_CONFIG="--config=remote_cache --remote_default_exec_properties=cache-silo-key=cache-silo-coverage"
```
stgpetrovic, line 12: Should this be in some "else" branch?

stgpetrovic (changes requested):

scripts/run_bazel_coverage.sh
lsy323: Thanks, fixed.

stgpetrovic (approved):

setup.py
```diff
@@ -251,16 +251,23 @@ def bazel_build(self, ext):
...
-    if GCLOUD_KEY_FILE:
-      bazel_argv.append('--google_credentials=%s' % GCLOUD_KEY_FILE)
-      if not _check_env_flag('BAZEL_REMOTE_CACHE'):
+    gclout_key_file_size = os.path.getsize(GCLOUD_KEY_FILE)
+    # Temporary workaround to allow PRs from forked repo to run CI. See details at (#5259).
+    # TODO: Remove the check once self-hosted GHA workers are avaialble to CPU/GPU CI.
+    print("check GCLOUD_KEY_FILE: {}".format(GCLOUD_KEY_FILE))
+    print("check if file exists: {}".format(os.path.exists(GCLOUD_KEY_FILE)))
+    print("check size: {}".format(os.path.getsize(GCLOUD_KEY_FILE)))
+    if gclout_key_file_size > 1:
```
will-cromar, line 260: Won't this prevent us from using the remove cache if we don't have a `GCLOUD_KEY_FILE` set? I've been running the remote cache with my GCE machine's default credentials, not a key file.

stgpetrovic: Yes, there thw adc is implicit. The criterion could be iff the key file.env var is set to an existing file that has size 1 or less, turn it off.

lsy323:
Update the logic in `setup.py`, so the workflow where credential file isn't explicitly set in `GCLOUD_KEY_FILE` can still use remote cache.

will-cromar (approved):
Thanks!

JackCaoG (approved):

PR #5263, opened 2023-06-27
[TEST] Make sure bazel remote cache is available after #5259

lsy323:


lsy323:
With #5259, remote cache for branch PRs is still enabled.

PR #5266, opened 2023-06-28
Update Sharded graph HLO dumping

JackCaoG:
Currently if we try to dump the HLO graph for a sharded computation, the logic in `DumpUtil::ToHlo` will return the post-optimization HLO, which contains a bunch of `fused_compuation.clone` and it is pretty unreadable. I think we want the HLO that's closer to the original IR for the debugging purpose, after this change, we will see something like
```
HloModule IrToHlo.8, entry_computation_layout={(f32[], f32[1,8]{1,0})->(f32[1,8]{1,0})}

ENTRY %IrToHlo.8 (p0.2: f32[], p1.4: f32[1,8]) -> (f32[1,8]) {
  %p1.4 = f32[1,8]{1,0} parameter(1), sharding={devices=[1,4]0,1,2,3}
  %p0.2 = f32[] parameter(0), sharding={replicated}
  %constant.1 = f32[] constant(1)
  %multiply.3 = f32[] multiply(f32[] %p0.2, f32[] %constant.1)
  %broadcast.5 = f32[1,8]{1,0} broadcast(f32[] %multiply.3), dimensions={}
  %add.6 = f32[1,8]{1,0} add(f32[1,8]{1,0} %p1.4, f32[1,8]{1,0} %broadcast.5)
  ROOT %tuple.7 = (f32[1,8]{1,0}) tuple(f32[1,8]{1,0} %add.6)
}
```

which gives us enough information about the input sharding which is useful.

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -454,6 +454,19 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     t3_expected = [9.0, 9.0, 9.0, 9.0, 9.0, 9.0, 9.0, 9.0]
     self.assertEqual(t3.tolist()[0], t3_expected)
 
+  def test_xla_sharded_hlo_dump(self):
+    partition_spec = (0, 1)
+    xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                       dtype=torch.float,
+                       device=xm.xla_device())
+    xst1 = xs.mark_sharding(xt1, self._get_mesh((1, self.n_devices)),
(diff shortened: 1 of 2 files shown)


alanwaketan:
I wonder what's the original intention to add this special case?

JackCaoG:
actual @yeounoh already implemented `_xla_partitioning_pass` in 
https://github.com/pytorch/xla/blob/e7ad09ed8ae670ed2a0d5a4c225fe0a262174c8f/torch_xla/csrc/init_python_bindings.cpp#L1497-L1505

for that purpose, I think we should leave `XLA_SAVE_TENSORS_FILE` intact.

PR #4693, opened 2023-02-24
Add feature to increase the number of host to device transfer threads

chandrasekhard2:
This config improves the ResNet training performance on TPU v4 chips. 

```
Epoch 3 train begin 19:02:08
| Training Device=xla:0/1 Epoch=3 Step=5100 Loss=3.82812 Rate=2014.43 GlobalRate=1220.24 Time=19:02:14
| Training Device=xla:0/3 Epoch=3 Step=5100 Loss=3.39062 Rate=2014.43 GlobalRate=1223.24 Time=19:02:14
| Training Device=xla:0/0 Epoch=3 Step=5100 Loss=3.64062 Rate=2014.44 GlobalRate=1221.29 Time=19:02:14
| Training Device=xla:0/2 Epoch=3 Step=5100 Loss=3.34375 Rate=2014.61 GlobalRate=1221.54 Time=19:02:14
| Training Device=xla:0/0 Epoch=3 Step=5400 Loss=3.51562 Rate=2221.05 GlobalRate=1254.91 Time=19:02:30
| Training Device=xla:0/3 Epoch=3 Step=5400 Loss=3.40625 Rate=2221.03 GlobalRate=1256.85 Time=19:02:30
| Training Device=xla:0/1 Epoch=3 Step=5400 Loss=3.54688 Rate=2221.02 GlobalRate=1253.85 Time=19:02:30
| Training Device=xla:0/2 Epoch=3 Step=5400 Loss=3.43750 Rate=2221.59 GlobalRate=1255.17 Time=19:02:30
| Training Device=xla:0/1 Epoch=3 Step=5700 Loss=2.89062 Rate=2267.96 GlobalRate=1284.59 Time=19:02:47
| Training Device=xla:0/3 Epoch=3 Step=5700 Loss=3.50000 Rate=2267.92 GlobalRate=1287.56 Time=19:02:47
| Training Device=xla:0/0 Epoch=3 Step=5700 Loss=3.42188 Rate=2267.86 GlobalRate=1285.63 Time=19:02:47
| Training Device=xla:0/2 Epoch=3 Step=5700 Loss=3.45312 Rate=2268.25 GlobalRate=1285.89 Time=19:02:47
| Training Device=xla:0/0 Epoch=3 Step=6000 Loss=3.03125 Rate=2323.80 GlobalRate=1315.59 Time=19:03:03
[...]

PR #5267, opened 2023-06-28
Fix race condition when use multi threads to transfer data in parallel Loader

aws-tianquaw:
This pull request includes a fix to the issue brought by [the previous pull request](https://github.com/pytorch/xla/pull/4693) that added feature to increase the number of host to device transfer threads. When using > 1 workers to transfer data, a possible race condition could happen and caused some data to be lost:

Currently, each thread will call [queue.close_write()](https://github.com/pytorch/xla/blob/master/torch_xla/distributed/parallel_loader.py#L174) when there is no more data in the loader queue. But if one thread called `queue.close_write()` while others still need to to put data to the queue, these batches of data will be lost. In this case. the `next_item()` call in [this line](https://github.com/pytorch/xla/blob/master/torch_xla/distributed/parallel_loader.py#L46) will return `None` while it should return valid data.

Ideally, we should only call `queue.close_write()` after all threads have completed writing data to the queue to avoid this race condition.


JackCaoG:
@Liyang90 Do you have cycle to take a look at this one?

chandrasekhard2 (approved):

JackCaoG:
@aws-tianquaw DO you mind rebasing this pr? Then ci should start passing

PR #5273, opened 2023-06-30
Always do build_ext in python setup.py develop

qihqi:
Bazel should figure out that _XLAC.so is current
or not, and trigger rebuild if any cpp files changed.


stgpetrovic (approved):

JackCaoG (approved):

will-cromar (approved):
Thanks!

PR #5277, opened 2023-06-30
Fix the error when export_torch_model is given a non-tensor

qihqi:
However the generated StableHLO graph still hardcodes the non-tensor value. this is not correct, will fix later.


lsy323 (approved):
LGTM

PR #5281, opened 2023-07-05
Fix the error where mark_step does not materalize tensors on SPMD:0

JackCaoG:
This is a ugly fix. The problem is that currently `mark_step` only happens on one device(if unspecified, the default device which is `TPU:0`, 'CPU:0' or 'GPU:0'). Under SPMD, we have a weird pattern that some tensors on `TPU:0` and some tensors on `SPMD:0`. 

I am planing to rewrite the virtual device logic and make all tensor son `spmd:0` in XLATensor level if SPMD is enabled in C++ land.

### DIFF
--- a/test/spmd/test_xla_virtual_device.py
+++ b/test/spmd/test_xla_virtual_device.py
@@ -79,6 +79,9 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
 
   def test_non_tensor_scalar(self):
     sharding_spec = xs.ShardingSpec(self._get_mesh((1, self.n_devices)), (0, 1))
+    # TODO(JackCaoG)currently, execution will only happen if there is at least one
+    # tensor on non-spmd:0 device.
+    t1 = torch.randn(3, 3, device=xm.xla_device())
     # tensor will have device as `SPMD:0` in c++
     xt1 = xm.send_cpu_data_to_device([torch.randn(3, 3)],
                                      xm.xla_device(),
@@ -86,6 +89,7 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
     # we will transfer 0.5 as a device_data to the 'SPMD:0' device, need to make sure
     # that virtual device can handle this case.
     xt2 = xt1 / 0.5
+    xm.mark_step(wait=True)
     torch.allclose(xt2.cpu(), xt1.cpu() / 0.5)
 
   def test_mark_step_on_virtual_device(self):
@@ -99,7 +103,7 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
                                      xm.xla_device(),
                                      input_sharding=sharding_spec)[0]
     xt2 = xt1 / 0.5
-    xm.mark_step()
+    xm.mark_step(wait=True)
     # after mark_step, xt2 should be materalized
     self.assertNotIn('aten::div',
                      torch_xla._XLAC._get_xla_tensor_debug_info(xt2))
--- a/test/spmd/test_xla_virtual_device.py
+++ b/test/spmd/test_xla_virtual_device.py
@@ -88,6 +88,22 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
     xt2 = xt1 / 0.5
     torch.allclose(xt2.cpu(), xt1.cpu() / 0.5)
 
+  def test_mark_step_on_virtual_device(self):
+    xm.mark_step()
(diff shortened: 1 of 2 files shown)


test/spmd/test_xla_virtual_device.py
```diff
@@ -88,6 +88,22 @@ def test_non_tensor_scalar(self):
     xt2 = xt1 / 0.5
     torch.allclose(xt2.cpu(), xt1.cpu() / 0.5)
 
+  def test_mark_step_on_virtual_device(self):
+    xm.mark_step()
+    sharding_spec = xs.ShardingSpec(self._get_mesh((1, self.n_devices)), (0, 1))
+    # TODO(JackCaoG)currently, execution will only happen if there is at least on
```
alanwaketan, line 94: `at least one`?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -502,7 +505,10 @@ XLAGraphExecutor::SyncTensorCollection XLAGraphExecutor::CollectSyncTensors(
                                   tsl::profiler::TraceMeLevel::kInfo);
   runtime::util::Unique<torch::lazy::BackendDevice> unique_device;
   for (size_t i = 0; i < tensors.size(); ++i) {
-    unique_device.set(tensors[i]->GetDevice());
+    // TODO(JackCaoG): all tensors should be on spmd:0 in SPMD mode.
+    if (tensors[i]->GetDevice().toString() != "SPMD:0") {
```
alanwaketan, line 509: So currently we allow SPMD:0 and TPU:0 to coexist? I wonder what's the deal with the code later on that deals with unique_device?

JackCaoG: I will move all tensors on `SPMD:0` under spmd context. Planning on work on that later this week. Currently the code that go through
```
tensor.to(xla_device)
```
will be on `TPU:0` and tensors go through
```
# data loader
`xm.send_cpu_data_to_device`
```
will be on `SPMD:0`, this is really confusing and make logic nasty.

alanwaketan: Right, but we have code in this method later on use unique_device? And it looks like it's important: `coll.device = *unique_device;` How does it work today?

And then if SPMD:0 and TPU:0 really coexist in this method, shouldn't `unique_device.set` crash?

JackCaoG: oh, so here is where things become messy. What happened is that we will pass `TPU:0` as the device to `Compile` and `ScheduleSyncTensorgraph` on xla_graph_executor. `ScheduleSyncTensorsGraph` will ignore the device and check if graph is sharded and execute on all device if that's the case

https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_graph_executor.cpp#L999-L1000

It is a very broken design now. Different function is checking different thing. I want to unify all those later.

alanwaketan: Err... whatever, I will approve that and then wait for your redesign.

alanwaketan (approved):
LGTM.

JackCaoG:
Sigh, there is another issue. On dynamo we always just use `SPMD:0` as the device for all parameters...which is the right thing to do... Let me try to fix dynamo failures.

test/spmd/test_xla_virtual_device.py
```diff
@@ -88,6 +88,22 @@ def test_non_tensor_scalar(self):
     xt2 = xt1 / 0.5
     torch.allclose(xt2.cpu(), xt1.cpu() / 0.5)
 
+  def test_mark_step_on_virtual_device(self):
+    xm.mark_step()
+    sharding_spec = xs.ShardingSpec(self._get_mesh((1, self.n_devices)), (0, 1))
+    # TODO(JackCaoG)currently, execution will only happen if there is at least one
+    # tensor on non-spmd:0 device.
+    t1 = torch.randn(3, 3, device=xm.xla_device())
```
alanwaketan, line 96: What's this t1 about?

JackCaoG: so.. her eis another weird bug.. if both `xt1` and `xt2` on `spmd:0`, the `mark_step` will see that there is no tensor on `TPU:0` device and it will skip the execution.. I think it is easier for me to rewrite whole virtual device instead of fixing these bugs to suit both `spmd:0` and `TPU:0` exist at the same time.

alanwaketan: Good luck... I wonder how was I able to do GPT-2 experiments all the time...

JackCaoG: I am surprised that code runs at all lol.

PR #5284, opened 2023-07-06
[SPMD][Virtual Device]All tensors should be in SPMD:0 C++ device

JackCaoG:
Context:
On current head, depending on how you created sharded tensor, the XLATensor(in C++)'s device might be either `SPMD:0` or `TPU:0`. This creates a bunch of issues and make codes really hard to understand. In this par I want to make all XLATensor device to be SPMD:0.

Other things I changed
1. for distributed checkpointing, each sharded used to be on devices like `xla:0`, `xla:1` ... I changed them to the actual hardware device like `tpu:0`, `tpu:1` since I only want to expose `xla:0` as virtual device to the user. This is fine on TPU but if we want to take a checkpoint on TPU and load it to GPU it will cause an error. I think we can just mark it as a todo now. @jonb377 FYI
2. Currently in some edge cases virtual device will create data placeholder for things when sharding is unkown. I think we should never do that since it is very problematic. IMO the only places that virtual device needs to optimize is
```
device = xm.xla_device()
fc = nn.Linear(FLAGS.input_dim, FLAGS.input_dim // 2)
fc.to(device)                                       ----> virtual device should delay the transfer
xs.mark_sharding(fc.parameters[0], sharding_spec)         ----> transfer the parameter[0] to device sharded

fc(input)                                                ----> all other paramter should be replicated
xm.mark_step()                                        ----> no device upload should happen during mark_step
```

TODO:
There are a bunch of places we use `UseVirtualDevice` to check things, we should check `XLATensor->Device()` whenever possible instead.

### DIFF
--- a/test/spmd/test_xla_virtual_device.py
+++ b/test/spmd/test_xla_virtual_device.py
@@ -103,7 +103,7 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
                      torch_xla._XLAC._get_xla_tensor_debug_info(xt2))
 
   def test_virtual_device_no_upload(self):
-    met.clear_counters()
+    met.clear_all()
     device = xm.xla_device()
     t1 = torch.randn(5, 5).to(device)
     t1_debug_info = torch_xla._XLAC._get_xla_tensor_debug_info(t1)
@@ -116,7 +116,7 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
     self.assertIn("XLAData: None", t1_debug_info)
 
   def test_virtual_device_upload_after_mark_sharding(self):
-    met.clear_counters()
+    met.clear_all()
     partition_spec = (0, 1)
     device = xm.xla_device()
     t1 = torch.randn(8, 8).to(device)
@@ -131,15 +131,13 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
     self.assertIn("TransferToServerTime", met.metric_names())
 
   def test_virtual_device_upload_after_tracing(self):
-    met.clear_counters()
+    met.clear_all()
     device = xm.xla_device()
     t1 = torch.randn(8, 8).to(device)
     t1_debug_info = torch_xla._XLAC._get_xla_tensor_debug_info(t1)
     self.assertIn("Tensor on host: with size [8, 8]", t1_debug_info)
     t2 = t1 + t1
     t1_debug_info_new = torch_xla._XLAC._get_xla_tensor_debug_info(t1)
-    import pdb
-    pdb.set_trace()
     # tensor should be uploaded to device after being used as input to other op.
     self.assertIn("Tensor on host: None", t1_debug_info_new)
     self.assertIn("xla::device_data", t1_debug_info_new)
--- a/test/spmd/test_xla_virtual_device.py
+++ b/test/spmd/test_xla_virtual_device.py
@@ -79,9 +79,6 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
 
   def test_non_tensor_scalar(self):
     sharding_spec = xs.ShardingSpec(self._get_mesh((1, self.n_devices)), (0, 1))
-    # TODO(JackCaoG)currently, execution will only happen if there is at least one
-    # tensor on non-spmd:0 device.
-    t1 = torch.randn(3, 3, device=xm.xla_device())
     # tensor will have device as `SPMD:0` in c++
     xt1 = xm.send_cpu_data_to_device([torch.randn(3, 3)],
                                      xm.xla_device(),
@@ -95,9 +92,6 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
   def test_mark_step_on_virtual_device(self):
     xm.mark_step()
     sharding_spec = xs.ShardingSpec(self._get_mesh((1, self.n_devices)), (0, 1))
-    # TODO(JackCaoG)currently, execution will only happen if there is at least one
-    # tensor on non-spmd:0 device.
-    t1 = torch.randn(3, 3, device=xm.xla_device())
     # tensor will have device as `SPMD:0` in c++
     xt1 = xm.send_cpu_data_to_device([torch.randn(3, 3)],
                                      xm.xla_device(),
--- a/test/spmd/test_xla_virtual_device.py
+++ b/test/spmd/test_xla_virtual_device.py
@@ -102,6 +102,65 @@ class VirtualDeviceTest(test_xla_sharding_base.XlaShardingTest):
     self.assertNotIn('aten::div',
                      torch_xla._XLAC._get_xla_tensor_debug_info(xt2))
 
+  def test_virtual_device_no_upload(self):
+    met.clear_counters()
+    device = xm.xla_device()
+    t1 = torch.randn(5, 5).to(device)
+    t1_debug_info = torch_xla._XLAC._get_xla_tensor_debug_info(t1)
+    # t1's upload to device should be deferred
+    self.assertIn("Tensor on host: with size [5, 5]", t1_debug_info)
+    self.assertNotIn("TransferToServerTime", met.metric_names())
+    # t1 should be on SPMD device under spmd context
+    self.assertIn("Device: SPMD:0", t1_debug_info)
+    self.assertIn("IR: None", t1_debug_info)
+    self.assertIn("XLAData: None", t1_debug_info)
+
+  def test_virtual_device_upload_after_mark_sharding(self):
+    met.clear_counters()
+    partition_spec = (0, 1)
+    device = xm.xla_device()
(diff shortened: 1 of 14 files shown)


JackCaoG:
All of the spmd unit test passed locally, I am going to try model test then adding more unit test to make sure all tensors will be on SPMD:0 C++ device.

JackCaoG:
local resnet test also run with reasonable speed on v4-8 with fake data 
```
| Training Device=xla:0/0 Epoch=1 Step=180 Loss=0.00367 Rate=7287.42 GlobalRate=3370.65 Time=00:41:50
| Training Device=xla:0/0 Epoch=1 Step=200 Loss=0.00337 Rate=7289.18 GlobalRate=3561.16 Time=00:41:55
| Training Device=xla:0/0 Epoch=1 Step=220 Loss=0.00310 Rate=7289.86 GlobalRate=3734.01 Time=00:42:01
| Training Device=xla:0/0 Epoch=1 Step=240 Loss=0.00286 Rate=7289.67 GlobalRate=3891.53 Time=00:42:07
```

I will work on the additional unit tests.

torch_xla/csrc/tensor_util.cpp
```diff
@@ -106,13 +106,18 @@ bool Use32BitLong() {
   return use_32bit_long;
 }
 
+bool IsTpuDevice(XlaDeviceType hw_type) {
+  // TODO(JackCaoG): We can't assume all SPMD device are TPU deivce.
+  return hw_type == XlaDeviceType::TPU && hw_type == XlaDeviceType::SPMD;
```
jonb377, line 111: Wouldn't this always return false?

torch_xla/experimental/xla_sharded_tensor.py
```diff
@@ -109,8 +109,7 @@ def __new__(cls, elem: torch.Tensor, *args, **kwargs):
   # which results from the sharding.
   @property
   def local_shards(self) -> List[XLAShard]:
-    shards = torch_xla._XLAC._get_local_shards(self.global_tensor)
-    devices = [str(shard.device) for shard in shards]
+    shards, devices = torch_xla._XLAC._get_local_shards(self.global_tensor)
     indices = torch_xla._XLAC._get_local_shard_indices(self.global_tensor)
     return [
         XLAShard(s.cpu(), i, d) for s, i, d in zip(shards, indices, devices)
```
jonb377, line 115: If `shards` are on CPU when returned from `_get_local_shards`, we can remove the `.cpu()` call on `s`

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1489,8 +1498,8 @@ void InitXlaModuleBindings(py::module m) {
           << "Input shard shape must include padding: " << shard.sizes()
           << " vs " << shard_shape;
     }
-    auto xla_devices = GetXlaDevices(devices);
-    auto xla_data = ShardingUtil::CreateShardedData(shards, xla_devices,
+    // auto xla_devices = GetXlaDevices(devices);
```
jonb377, line 1501: Nit: delete this

torch_xla/csrc/tensor_util.cpp
JackCaoG: good catch, it should a `||`

torch_xla/experimental/xla_sharded_tensor.py
JackCaoG: I actually don't remember, I can check.

torch_xla/csrc/tensor_util.cpp
JackCaoG: OK we actually also run SPMD on CPU.. I need to use a better way to tell if underlying type is TPU

test/spmd/test_xla_virtual_device.py
```diff
@@ -108,6 +102,65 @@ def test_mark_step_on_virtual_device(self):
...
+
+  def test_virtual_device_upload_after_tracing(self):
+    met.clear_counters()
+    device = xm.xla_device()
+    t1 = torch.randn(8, 8).to(device)
+    t1_debug_info = torch_xla._XLAC._get_xla_tensor_debug_info(t1)
+    self.assertIn("Tensor on host: with size [8, 8]", t1_debug_info)
+    t2 = t1 + t1
+    t1_debug_info_new = torch_xla._XLAC._get_xla_tensor_debug_info(t1)
+    import pdb
```
alanwaketan: What's this about?

torch_xla/csrc/tensor.cpp
```diff
@@ -338,6 +338,7 @@ torch::lazy::Value XLATensor::GetIrValue() const {
   c10::optional<at::Tensor> tensor_data = CurrentTensorData();
   XLA_CHECK(tensor_data);
   AssignIrValue(GetIrValueForTensor(*tensor_data, GetDevice()));
+  data()->tensor_data = c10::nullopt;
```
alanwaketan, line 341: What's this about?

test/spmd/test_xla_virtual_device.py
JackCaoG: updated

torch_xla/experimental/xla_sharded_tensor.py
JackCaoG: it is on CPU already, so I removed the `.cpu` below

JackCaoG:
@alanwaketan @jonb377 Can you guys take another pass at this pr?

torch_xla/experimental/xla_sharded_tensor.py
```diff
@@ -109,12 +109,9 @@ def __new__(cls, elem: torch.Tensor, *args, **kwargs):
...
   @property
   def local_shards(self) -> List[XLAShard]:
-    shards = torch_xla._XLAC._get_local_shards(self.global_tensor)
-    devices = [str(shard.device) for shard in shards]
+    shards, devices = torch_xla._XLAC._get_local_shards(self.global_tensor)
     indices = torch_xla._XLAC._get_local_shard_indices(self.global_tensor)
-    return [
-        XLAShard(s.cpu(), i, d) for s, i, d in zip(shards, indices, devices)
-    ]
+    return [XLAShard(s, i, d) for s, i, d in zip(shards, indices, devices)]
```
alanwaketan, line 114: It looks like `XLAShard.shard_device` will be used by `torch_xla._XLAC._load_local_shards`. Then within, it calls `GetXlaDevices` which uses `bridge::AtenDeviceToXlaDevice(c10::Device(device_str))` to create the backend data. I wonder how this works now after you return tpu:0 instead of xla:0.

alanwaketan (commented):
Left one comment, otherwise LGTM.

torch_xla/experimental/xla_sharded_tensor.py
JackCaoG: we used to convert `tpu:0` -> `xla:0` and return `xla:0` as `XLAShards`'s device. In the load part we call `GetXlaDevices` to convert `xla:0` back to `tpu:0`.

After this change we directly save `tpu:0` as the `XLAShard`'s device. The downside is if we saved a shard on TPU, we can't load it back to GPU since now device is hard code as TPU. This is fixable through, I just don't want to expose `xla:1` -> `xla:3` to user since there should only one `xla:0` which is a virtual device.

alanwaketan: By reading the code, I believe `device_str` in `c10::Device(device_str)` is `tpu:0`? Then how can `c10::Device` parse it?

jonb377: Just a note - we don't serialize the device string (e.g. `tpu:0`) when checkpointing, the device string is only tracked so we can map the CPU shard back to the shard on device. This change doesn't break anything for checkpointing.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1489,8 +1498,7 @@ void InitXlaModuleBindings(py::module m) {
           << "Input shard shape must include padding: " << shard.sizes()
           << " vs " << shard_shape;
     }
-    auto xla_devices = GetXlaDevices(devices);
-    auto xla_data = ShardingUtil::CreateShardedData(shards, xla_devices,
+    auto xla_data = ShardingUtil::CreateShardedData(shards, devices,
```
JackCaoG, line 1501: @alanwaketan this is where device `tpu:0` being used. it used to be `devices` is `xla:0` and we used `GetXlaDevices` to convert it to `tpu:0`. Now we passed `tpu:0` directly to `CreateShardedData`. Now we just don't call `GetXlaDevices` at all.

jonb377 (approved):

torch_xla/csrc/init_python_bindings.cpp
alanwaketan: Okay, then I think it's safe. Thanks, @JackCaoG!

alanwaketan (approved):

PR #5285, opened 2023-07-06
Disable torch._dynamo.config.automatic_dynamic_shapes

wonjoolee95:
With https://github.com/pytorch/pytorch/pull/103623 that enables dynamic shape on PyTorch Dynamo by default, our test DynamoInferenceBasicTest.test_simple_model_with_different_input_shape was broken. 

More context: https://github.com/pytorch/pytorch/pull/103623#issuecomment-1622571952

While we investigate for a long term fix, manually setting this flag (https://github.com/pytorch/pytorch/pull/103623/files#diff-5f3e7e4866f93af05691e2382d2258cad959091492f900583a49e55266f5e897L67) to False.

With this, the test is succeeding again:
```
wonjoo@t1v-n-9eb87547-w-0:~/pytorch/xla$ python3 test/dynamo/test_dynamo.py DynamoInferenceBasicTest.test_simple_model_with_different_input_shape
.
Ran 1 test in 3.191s

OK

```

cc @JackCaoG, let me know if you think this is okay.

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -68,7 +68,6 @@ class DynamoInferenceBasicTest(unittest.TestCase):
     res_cpu_3 = self.fn_simple(x + y, y * 3)
     self.assertTrue(torch.allclose(res_cpu_3, res_xla_dynamo_3.cpu()))
 
-  @unittest.skip("currently failing at head")
   def test_simple_model_with_different_input_shape(self):
     met.clear_counters()
     device = xm.xla_device()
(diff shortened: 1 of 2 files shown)


JackCaoG (approved):

wonjoolee95:
Hmm, looking back at this, ideally I want to set this flag ***only*** when XLA Dynamo is being used. But this PR sets this to false for entire XLA. 

I tried adding this line in our dynamo_bridge.py, but it doesn't seem to properly set this flag to false when placed there.

wonjoolee95:
I guess that doesn't really makes sense since XLA Dynamo uses LTC code path too. So if we disable only once XLA Dynamo is ran, all subsequent executions with be ran with this flag set to False anyways.

I'll go ahead and merge this once CI is green.

PR #5288, opened 2023-07-07
Add `_sharded_cpu_state_dict` for distributed checkpointing

yashs97:
Add `_sharded_cpu_state_dict` for distributed checkpointing


torch_xla/experimental/_distributed_checkpoint_helpers.py
```diff
@@ -186,4 +190,22 @@ def narrow_tensor_by_index(tensor: torch.Tensor, offsets: Sequence[int],
...
+
+
+def _sharded_cpu_state_dict(state_dict: STATE_DICT_TYPE) -> STATE_DICT_TYPE:
+  """
+  Converts a state_dict on XLA device to a sharded state_dict on CPU.
+  """
+
+  def move_state_dict_to_cpu(v):
+    v = xs.wrap_if_sharded(v)
+    return _CpuShards(shards=v.local_shards, global_shape=v.global_tensor.shape)
```
jonb377, line 209: We should only wrap tensors wit non-REPLICATED sharding, similar to https://github.com/pytorch/xla/blob/42d92709b58ad86a12ba94b834399ead1a6392f4/torch_xla/experimental/distributed_checkpoint.py#L378-L381. You can move that method into this file to avoid circular dependencies.

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -244,6 +247,18 @@ def test_resolve_shard_data(self):
...
+class DistributedCheckpointHelpersTest(DistributedCheckpointTestBase):
+
+  def test_sharded_cpu_state_dict(self):
+    model = self.SimpleLinear().to(xm.xla_device())
+    state_dict = model.state_dict()
+    sharded_cpu_state_dict = _sharded_cpu_state_dict(state_dict)
+    self.assertCountEqual(sharded_cpu_state_dict,
+                          ['fc1.weight', 'fc1.bias', 'fc2.weight', 'fc2.bias'])
+    self.assertTrue(
+        isinstance(sharded_cpu_state_dict['fc1.weight'], _CpuShards))
```
jonb377, line 259: Let's also add an assertion that the other tensors are not wrapped in `_CpuShards` but are on the `CPU` device.

```
for name, param in sharded_cpu_state_dict.items():
    if name == 'fc1.weight':
       # Assert isinstance _CpuShards
    else:
      # Assert isinstance torch.Tensor and device is CPU
```

torch_xla/experimental/_distributed_checkpoint_helpers.py
```diff
@@ -212,7 +218,9 @@ def _sharded_cpu_state_dict(state_dict: STATE_DICT_TYPE) -> STATE_DICT_TYPE:
   """
 
   def move_state_dict_to_cpu(v):
-    v = xs.wrap_if_sharded(v) if _is_sharded_tensor(v) else v
+    if not _is_sharded_tensor(v):
```
jonb377, line 221: We need to `wrap_if_sharded` before calling this function

torch_xla/experimental/_distributed_checkpoint_helpers.py
```diff
@@ -186,4 +191,36 @@ def narrow_tensor_by_index(tensor: torch.Tensor, offsets: Sequence[int],
...
+
+def _sharded_cpu_state_dict(state_dict: STATE_DICT_TYPE) -> STATE_DICT_TYPE:
+  """
+  Converts a state_dict on XLA device to a sharded state_dict on CPU.
+  """
+
+  def move_state_dict_to_cpu(v):
+    v = xs.wrap_if_sharded(v)
+    if not _is_sharded_tensor(v):
+      return _unwrap_xla_sharded_tensor(v).cpu()
```
jonb377, line 223: We should generalize this some - the state dict may contain non-Tensor fields like scalars, so we should only call `.cpu` if the result of `_unwrap` is a tensor.

jonb377 (approved):
LGTM, thanks Yash!

alanwaketan:
@yashs97 @jonb377 Can you share me some pointers about what does this enable? I thought the MVP is sort of functional.

jonb377:
@alanwaketan This is a preliminary change to eventually support async checkpointing. It isn't currently in use.

alanwaketan:
Thanks @jonb377 for the context.

PR #5299, opened 2023-07-12
[SPMD] Add FSDP sharding for test_train_spmd_linear_model.py

alanwaketan:
Summary:
This diff adds FSDP sharding for test_train_spmd_linear_model.py.

Test Plan:
PJRT_DEVICE=TPU XLA_USE_SPMD=1 python test/spmd/test_train_spmd_linear_model.py --sharding fsdp

### DIFF
--- a/test/spmd/test_train_spmd_linear_model.py
+++ b/test/spmd/test_train_spmd_linear_model.py
@@ -15,7 +15,7 @@ from torch import nn
 
 MODEL_OPTS = {
     '--sharding': {
-        'choices': ['batch', 'megatron-lm'],
+        'choices': ['batch', 'megatron-lm', 'fsdp'],
         'nargs': '+',
         'default': [],
     },
@@ -58,7 +58,6 @@ device = xm.xla_device()
 
 def train():
   print('===> Preparing data..')
-  num_epochs = 18
   lr = 0.1
   train_loader = xu.SampleGenerator(
       data=(torch.zeros(FLAGS.batch_size, FLAGS.input_dim),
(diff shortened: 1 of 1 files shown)


test/spmd/test_train_spmd_linear_model.py
```diff
@@ -78,6 +77,14 @@ def train():
     train_loader = pl.MpDeviceLoader(
         train_loader, device, input_sharding=xs.ShardingSpec(mesh, (0, 1)))
 
+  if 'fsdp' in FLAGS.sharding:
+    train_loader = pl.MpDeviceLoader(
+        train_loader, device, input_sharding=xs.ShardingSpec(mesh, (0, 1)))
+    print('Sharding model weights')
+    # Shard the weights according to their 0th dim
+    xs.mark_sharding(model.fc1.weight, mesh, (0, 1))
+    xs.mark_sharding(model.fc2.weight, mesh, (0, 1))
```
JackCaoG, line 86: hmm we are using `mark_sharding` as an inplace op, if that's how we are going to let user uses it, we should make it `mark_sharding_` instead.

JackCaoG (approved):

test/spmd/test_train_spmd_linear_model.py
alanwaketan: Good catch. I feel like we probably shouldn't make it inplace op. Let me make a follow up to make it more clear.

alanwaketan:
Thanks Jack for approving.

PR #5300, opened 2023-07-12
[SPMD] Avoid recompilations in xs.mark_sharding()

alanwaketan:
Summary:
This pull requests fixes the recompilation issue in xs.mark_sharding(). xtensor->GetXlaData() will compile the program if xtensor is an IR in order to get the BackendData. I believe this is not intended given the error message below suggests only data type xtensors are supported.

Test Plan:
PJRT_DEVICE=TPU XLA_USE_SPMD=1 python test/spmd/test_xla_sharding.py

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -488,6 +488,7 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     # scalar 5 should be replicated
     self.assertIn('%p0.2 = f32[] parameter(0), sharding={replicated}', hlo)
 
+  @unittest.skip("TODO(alanwaketan): Implement IR sharding to re-enable this.")
   def test_2d_tensor_3d_mesh(self):
     ct1 = torch.randn(16, 16, device='cpu')
     ct2 = torch.randn(16, 16, device='cpu')
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -223,13 +223,15 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
 
   def test_mark_sharding_partial(self):
     device = xm.xla_device()
-    t1 = torch.randn(4, 4)
-    t2 = torch.randn(4, 4)
+    t1 = torch.randn(4, 4).to(device)
+    t2 = torch.randn(4, 4).to(device)
     # Somehow the eager cpu result is different from the xla result.
-    expected = (t1.to(device) @ t2.to(device)).cpu()
+    expected = t1 @ t2
+    # To re-materialize t1 and t2.
+    xm.mark_step()
+    xm.wait_device_ops()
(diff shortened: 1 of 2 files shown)


test/spmd/test_xla_sharding.py
```diff
@@ -223,10 +223,13 @@ def test_mark_sharding_4d(self):
 
   def test_mark_sharding_partial(self):
     device = xm.xla_device()
-    t1 = torch.randn(4, 4).to(xm.xla_device())
-    t2 = torch.randn(4, 4).to(xm.xla_device())
-    expected = (t1 @ t2).cpu()
+    t1 = torch.randn(4, 4)
+    t2 = torch.randn(4, 4)
+    # Somehow the eager cpu result is different from the xla result.
+    expected = (t1.to(device) @ t2.to(device)).cpu()
```
JackCaoG, line 229: why do we need to make changes here?

alanwaketan: Because the xtensors used in expected should be copies of the ones used in actual spmd. The previous tests doesn't take care of those and t1 after expected calculation becomes IR.

JackCaoG: ok, can we do 
```
    t1 = torch.randn(4, 4).to(xm.xla_device())
    t2 = torch.randn(4, 4).to(xm.xla_device())
    expected = (t1 @ t2).cpu()
    xm.mark_step()
```
instead? This way we don't need to move same tensor to xla device multiple times. You can leave a comment explaining why mark_step is needed.

alanwaketan: Let me try. That should work too.

alanwaketan: Fixed.

alanwaketan:
@jonb377 @JackCaoG This is ready for another round of reviews.

test/spmd/test_xla_sharding.py
```diff
@@ -223,9 +223,14 @@ def test_mark_sharding_4d(self):
 
   def test_mark_sharding_partial(self):
     device = xm.xla_device()
-    t1 = torch.randn(4, 4).to(xm.xla_device())
-    t2 = torch.randn(4, 4).to(xm.xla_device())
-    expected = (t1 @ t2).cpu()
+    t1 = torch.randn(4, 4).to(device)
+    t2 = torch.randn(4, 4).to(device)
+    # Somehow the eager cpu result is different from the xla result.
```
JackCaoG, line 228: hmm... I expect them to be off by a bit, due to optimization. They should not be too different through.

JackCaoG (approved):

test/spmd/test_xla_sharding.py
alanwaketan: Hmm... Actually quite off. Probably worth looking next.

alanwaketan:
Thanks Jack for approving.

PR #5305, opened 2023-07-13
Supoort unordered sharding spec correctly

JackCaoG:
We need to respect the sharding spec and transpose tile_assigniment accordingly.

TODO:
support unordered partial replication. In partial replication `tile_assigniment` is ignored in favor of `group_assignment` and `replication_groups`, need to figure out how to update that.

### DIFF
--- a/torch_xla/experimental/xla_sharding.py
+++ b/torch_xla/experimental/xla_sharding.py
@@ -319,8 +319,28 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
   return sharding_type
 
 
-def _get_tile_assignment(mesh: Mesh) -> List[int]:
-  return mesh.get_logical_mesh().tolist()
+def _get_tile_assignment(mesh: Mesh,
+                         partition_spec: Tuple[Union[int, None]]) -> List[int]:
+  # Use Torch.tensor here to make use of the torch.transpose_
+  mesh_list_tensor = torch.tensor(mesh.get_logical_mesh().tolist())
+  # This is 2d tensor 3d mesh case, tile_assigniment will be ignore in favor of
+  # group_assignment and replication_groups.
+  if (mesh_list_tensor.dim() != len(partition_spec)):
+    return mesh_list_tensor.tolist()
+  dim_to_current_dim = {}
+  for i in range(len(partition_spec)):
+    dim_to_current_dim[i] = i
+  for i in range(len(partition_spec)):
+    # we need transpose the mesh to match the partition_spec
+    if partition_spec[i] != None and partition_spec[i] != dim_to_current_dim[i]:
+      # need to find where the target dimension to transpose to
+      target_dim = dim_to_current_dim[partition_spec[i]]
+      mesh_list_tensor.transpose_(i, target_dim)
+      # now update the mapping by swapping the current and target dim_to_current_dim
+      # since transposed happened
+      dim_to_current_dim[partition_spec[i]] = dim_to_current_dim[i]
+      dim_to_current_dim[i] = target_dim
+  return mesh_list_tensor.tolist()
 
 
 def _get_group_assignment(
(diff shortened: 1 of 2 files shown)


JackCaoG:
This should help https://github.com/pytorch/xla/issues/5264

torch_xla/experimental/xla_sharding.py
```diff
@@ -319,8 +319,28 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
 
 
-def _get_tile_assignment(mesh: Mesh) -> List[int]:
-  return mesh.get_logical_mesh().tolist()
+def _get_tile_assignment(mesh: Mesh,
+                         partition_spec: Tuple[Union[int, None]]) -> List[int]:
+  # Use Torch.tensor here to make use of the torch.transpose_
+  mesh_list_tensor = torch.tensor(mesh.get_logical_mesh().tolist())
+  # This is 2d tensor 3d mesh case, tile_assigniment will be ignore in favor of
+  # group_assignment and replication_groups.
```
jonb377, line 327: Does `group_assignment` already handle the reordering?

jonb377 (approved):
LGTM, thanks Jack!

torch_xla/experimental/xla_sharding.py
JackCaoG: nope lol I plan to support that in next pr.

torch_xla/experimental/xla_sharding.py
```diff
@@ -319,8 +319,28 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
+    return mesh_list_tensor.tolist()
+  dim_to_current_dim = {}
+  for i in range(len(partition_spec)):
+    dim_to_current_dim[i] = i
+  for i in range(len(partition_spec)):
+    # we need transpose the mesh to match the partition_spec
+    if partition_spec[i] != None and partition_spec[i] != dim_to_current_dim[i]:
+      # need to find where the target dimension to transpose to
+      target_dim = dim_to_current_dim[partition_spec[i]]
+      mesh_list_tensor.transpose_(i, target_dim)
```
alanwaketan, line 338: I believe you can use permute instead of looping transpose.
https://pytorch.org/docs/stable/generated/torch.permute.html

alanwaketan (approved):
Generally LGTM besides one comment.

torch_xla/experimental/xla_sharding.py
JackCaoG: oh nice! let me try that.

torch_xla/experimental/xla_sharding.py
```diff
@@ -319,8 +319,24 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
+                         partition_spec: Tuple[Union[int, None]]) -> List[int]:
+  # Use Torch.tensor here to make use of the torch.transpose_
+  mesh_list_tensor = torch.tensor(mesh.get_logical_mesh().tolist())
+  # This is partial sharding case, tile_assigniment will be ignore in favor of
+  # group_assignment and replication_groups.
+  if (mesh_list_tensor.dim() != len(partition_spec)):
+    return mesh_list_tensor.tolist()
+  partition_spec_list = list(partition_spec)
+  for i in range(len(partition_spec_list)):
+    if partition_spec_list[i] == None:
```
alanwaketan, line 332: Not quite understanding this part and then it doesn't seem like you have a test case for it as well.

JackCaoG: For partial partial and replicated, the sharding spec will contains Nones. For example `[0, 1, None]`, `[None, None]`, we can't call permute in a array with `None`, so we have to replace them with numbers.

That being said, I am still working on partial partion on next pr, this check here just to make sure fully replicated and partial partion does not crash.

JackCaoG: without this, existing test for fully replicated will error out. since we can't do `tensor.permute([2, 1, None])`

alanwaketan: Thanks for the context. LGTM.

PR #5308, opened 2023-07-14
Add tf independent serialization

qihqi:
Create a serialization format for StableHLO graphs and weights without tf.saved_model

Need to not use tensorflow because tensorflow is no longer dependency of pytorch/xla.
Information saved are enough to reconstruct the tf.saved_model for serving.
Information stored:
* metadata on which tensor maps which input position
* StableHLO version number
* metadata on which tensor corresponds to user input or parameter
* metadata on shape and dtype of each tensor.
Tensors themselves are saved as numpy arrays using `np.save`.

### DIFF
--- a/torch_xla/experimental/stablehlo_saved_model.py
+++ b/torch_xla/experimental/stablehlo_saved_model.py
@@ -1,4 +1,5 @@
 import copy
+from dataclasses import dataclass
 import shutil
 import os
 import re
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/stablehlo_saved_model.py
```diff
@@ -137,3 +143,257 @@ def export_torch_model(model: torch.nn.Module,
...
+  def constant(cls, position):
+    return cls(type_=VariableType.CONSTANT, position=position)
+
+
+@dataclass
+class StableHLOFunctionMeta:
+  # name of the callable.
+  name: str
+  # version.
+  stablehlo_version: int
```
lsy323, line 184: Not sure if we should use a str? Seems StableHLO version is major.minor.patch. https://github.com/openxla/stablehlo/blob/bf4ff668528f789bd12480de2dfbe68cab73dcc0/stablehlo/dialect/Version.h#L34

lsy323 (approved):
LGTM, one comment for stablehlo version

PR #5315, opened 2023-07-17
Use `_sharded_cpu_state_dict` functionality to Write Items for SPMD Save Planner

yashs97:
Use `_sharded_cpu_state_dict` functionality to Write Items for SPMD Save Planner


jonb377 (changes requested):

torch_xla/experimental/distributed_checkpoint.py
```diff
@@ -217,8 +214,8 @@ def set_up_planner(
     state_dict, self.mappings = flatten_state_dict(state_dict)
     state_dict = tree_map(xs.wrap_if_sharded, state_dict)
 
-    # Select only XLAShardedTensors which are not replicated, since the
-    # default planner can handle everything else.
+    # Select only XLAShardedTensors which are not replicated or _CpuShards,
```
jonb377, line 217: nit: revert this comment as well

jonb377 (commented):
Looks good Yash! Could we also add a unit test for saving a state_dict with `_CpuShards`?

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -208,6 +208,25 @@ def test_state_dict_separation(self):
           planner.unsharded_state_dict,
           ['fc1.weight', 'fc1.bias', 'fc2.weight', 'fc2.bias'])
 
+  def test_save_state_dict_with_cpu_shards(self):
+    model = self._get_sharded_model()
+    planner = self._get_save_planner(model)
```
jonb377, line 213: We want to create the SavePlanner with the result of `_sharded_cpu_state_dict` instead of the model's state_dict, so we shouldn't use this helper function. Something like this would work:

```
    sharded_state_dict = _sharded_cpu_state_dict(model.state_dict())
    planner = SPMDSavePlanner()
    planner.set_up_planner(sharded_state_dict, True)
````

Also, let's remove the `_is_sharded_tensor` check on line 214, since if n_devices > 1, fc1.weight should be _CpuShards in the state_dict.

File: torch_xla/experimental/distributed_checkpoint.py [hunk 3]
```diff
@@ -88,12 +83,14 @@ def set_up_planner(self, state_dict: STATE_DICT_TYPE,
     # types that can be handled by the default planner, and ensure all sharded
     # tensors are wrapped in XLAShardedTensor
     state_dict, self.mappings = flatten_state_dict(state_dict)
-    state_dict = tree_map(xs.wrap_if_sharded, state_dict)
+    state_dict = _sharded_cpu_state_dict(state_dict)
```
jonb377, line 86: We don't want to call _sharded_cpu_state_dict within the planner, this should still just tree_map `wrap_if_sharded`

yashs97: this is what I had earlier but the tests failed so I changed it

test/spmd/test_xla_distributed_checkpoint.py
yashs97: this is what I had earlier but the tests failed so I changed it

torch_xla/experimental/xla_sharding.py
```diff
@@ -461,7 +461,8 @@ def wrap_if_sharded(x: Any) -> Any:
   Otherwise, returns the input.
   """
   if (isinstance(x, torch.Tensor) and not isinstance(x, XLAShardedTensor) and
-      torch_xla._XLAC._get_xla_sharding_type(x) is not None):
+      torch_xla._XLAC._get_xla_sharding_type(x) is not None and
+      x.device.type == 'xla'):
```
jonb377, line 465: We need to do the `device.type` check first, `_get_xla_sharding_type` asserts that the device is an XLA device

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -208,6 +208,26 @@ def test_state_dict_separation(self):
           planner.unsharded_state_dict,
           ['fc1.weight', 'fc1.bias', 'fc2.weight', 'fc2.bias'])
 
+  def test_save_state_dict_with_cpu_shards(self):
```
jonb377, line 211: Let's add two more tests - one to verify the number of WriteItem generated, and one to verify the resolve_data call returns the local shards.

jonb377 (commented):
Looking great! Last note: we need to handle `lookup_object` for `_CpuShards`. The relevant section is line 135 in your PR

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -98,6 +106,11 @@ def test_sharded_to_unsharded(self):
     model = self.SimpleLinear().to(xm.xla_device())
     sharded_model = self._get_sharded_model()
     self._save_and_restore(sharded_model, model, save_planner=SPMDSavePlanner())
+    self._save_and_restore(
```
jonb377, line 109: We should re-initialize the models before running the test. If we can't parametrize the test, let's make this a separate test

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -66,18 +67,22 @@ def _save_and_restore(self,
     tmpdir = tempfile.mkdtemp()
 
     # Save an unsharded model using the provided save planner
+    model_in_state_dict = model_in.state_dict(
+    ) if not is_sharded_cpu_state_dict else _sharded_cpu_state_dict(
+        model_in.state_dict())
```
jonb377, line 72: This line is a little hard to follow due to the formatting, could we break it into multiple lines:

```
model_in_state_dict = model_in.state_dict()
if is_sharded_cpu_state_dict:
  model_in_state_dict = _sharded_cpu_state_dict(model_in_state_dict)
```

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -95,9 +100,15 @@ def test_unsharded_to_sharded(self):
   # TODO(jonbolin): Enable tests for resharding into coarser meshes
   @unittest.skip("View assignment with virtual device is not yet supported")
   def test_sharded_to_unsharded(self):
-    model = self.SimpleLinear().to(xm.xla_device())
-    sharded_model = self._get_sharded_model()
-    self._save_and_restore(sharded_model, model, save_planner=SPMDSavePlanner())
+    for i in [True, False]:
```
jonb377, line 103: nit: `i` isn't very descriptive, can we use e.g. `chkpt_on_cpu`?

jonb377 (approved):
Thanks Yash! A couple of small formatting comments, but overall LGTM

PR #5316, opened 2023-07-17
Support unordered sharding spec for partial replication

JackCaoG:


torch_xla/experimental/xla_sharding.py
```diff
@@ -339,26 +335,28 @@ def _get_tile_assignment(mesh: Mesh,
   return mesh_list_tensor.permute(partition_spec_list).tolist()
 
 
-def _get_group_assignment(
-    sharding_type: ShardingType, mesh: Mesh,
-    partition_spec: Tuple[Union[int, None]]) -> Tuple[List, List]:
+def _get_group_assignment(sharding_type: ShardingType, mesh: Mesh,
```
jonb377, line 338: It looks like the `mesh` parameter isn't used anymore

jonb377 (approved):
Thanks Jack! LGTM

torch_xla/experimental/xla_sharding.py
JackCaoG: nice catch, I should remove it.

JackCaoG: @jonb377 I will fix this in a follow up pr.

PR #5320, opened 2023-07-19
Enable Some input output aliasing under SPMD

JackCaoG:
I was able to see buffer being aliased when running conv sharding for `Resnet50`. 

Note that this pr mostly enables aliasing for parameters that we explictly call `mark_sharding` and applied in place operation in optimizer(in place operation will inherit the sharding in XLATensor level). For those tensors that does not have explicit output sharding we currently skipped aliasing since compiler might assign different sharing.


alanwaketan (approved):
LGTM. Thanks, Jack.

PR #5331, opened 2023-07-21
[SPMD] Multi-host batch sharded data loading

khatwanimohit:
This PR is changing the way we do multi-host data loading. Here are some changes user script will need. 
1. Add a Sampler 
```
sampler = torch.utils.data.DistributedSampler(dataset, num_replicas = xr.process_count(), rank = xr.process_index())
```
2. Using sampler in `torch.utils.data.DataLoader`
```
loader = torch.utils.data.DataLoader(
        dataset,
        batch_size=FLAGS.batch_size // xr.process_count(),
        sampler=sampler,
        drop_last=FLAGS.drop_last,
        shuffle=False)
 ```
3. Change sharding_spec in `MpDeviceLoader`
```
loader = pl.MpDeviceLoader(
          loader,
          device,
          input_sharding=xs.ShardingSpec(input_mesh, (0, 1, 2, 3), minibatch = True))
 ```
`minibatch` flag denotes that input is already sharded along the batch axes. Multi-host dataloading currently only supports batch dimension sharding.

### DIFF
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -50,8 +50,7 @@ TEST_F(XLAShardingTest, GetShardShape) {
   // For tiled sharding, each dimension should be halved
   EXPECT_EQ(shard_shape, std::vector<int64_t>({4, 4}));
 
-  sharding = xla::HloSharding::Replicate().ToProto();
-  sharding_spec->sharding = sharding;
+  sharding_spec->sharding = xla::HloSharding::Replicate().ToProto();
   shard_shape = ShardingUtil::GetShardShape(sharding_spec);
   // For replicated sharding, each dimension should be preserved
   EXPECT_EQ(shard_shape, std::vector<int64_t>({8, 7}));
@@ -61,7 +60,8 @@ TEST_F(XLAShardingTest, GetShardIndicesForDevices) {
   std::vector<std::string> devices = {"TPU:0", "TPU:1", "TPU:2", "TPU:3"};
 
   auto tensor = at::ones({8, 7}, at::TensorOptions(at::kFloat));
-  xla::Shape tensor_shape = CreateComputationShapeFromTensor(tensor, nullptr);
+  xla::Shape tensor_shape =
+      CreateComputationShapeFromTensor(tensor, GetDefaultDevice());
   xla::Array2D<int64_t> mesh({
       {0, 1},
       {2, 3},
@@ -91,7 +91,6 @@ TEST_F(XLAShardingTest, GetShardIndicesForDevices) {
       EXPECT_EQ(slice.step(), 1);
     }
   }
-
   sharding = xla::HloSharding::Replicate().ToProto();
   sharding_spec->sharding = sharding;
   shard_shape = ShardingUtil::GetShardShape(sharding_spec);
@@ -110,7 +109,8 @@ TEST_F(XLAShardingTest, ShardTensor) {
 
   // 1D tiled
   at::Tensor tensor = at::ones({8}, at::TensorOptions(at::kFloat));
-  xla::Shape tensor_shape = CreateComputationShapeFromTensor(tensor, nullptr);
+  xla::Shape tensor_shape =
+      CreateComputationShapeFromTensor(tensor, GetDefaultDevice());
   xla::OpSharding sharding =
       xla::HloSharding::Tile1D(
           CreateComputationShapeFromTensor(tensor, GetDefaultDevice()),
@@ -127,7 +127,7 @@ TEST_F(XLAShardingTest, ShardTensor) {
   // 2D tiled, The first dim is halved and the last replicated. The last shard
   // size should be smaller in dim=1 because it's not evenly divisible.
   tensor = at::ones({8, 7, 4}, at::TensorOptions(at::kFloat));
-  tensor_shape = CreateComputationShapeFromTensor(tensor, nullptr);
+  tensor_shape = CreateComputationShapeFromTensor(tensor, GetDefaultDevice());
   xla::Array2D<int64_t> mesh({
       {0, 1, 2, 3},
       {4, 5, 6, 7},
@@ -144,8 +144,7 @@ TEST_F(XLAShardingTest, ShardTensor) {
   // 3D tiled, the first dim is replicated and the last halved. The last shard
   // size should be smaller in dim=1 because it's not evenly divisible.
   xla::Array3D<int64_t> cube({{{0, 1}, {2, 3}, {4, 5}, {6, 7}}});
-  sharding = xla::HloSharding::Tile(cube).ToProto();
-  sharding_spec->sharding = sharding;
+  sharding_spec->sharding = xla::HloSharding::Tile(cube).ToProto();
   shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
                                      /*padded=*/false);
   EXPECT_EQ(shards.size(), 8);
@@ -153,8 +152,7 @@ TEST_F(XLAShardingTest, ShardTensor) {
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 1, 2}));
 
   // Replicated, all shards should be identical.
-  sharding = xla::HloSharding::Replicate().ToProto();
-  sharding_spec->sharding = sharding;
+  sharding_spec->sharding = xla::HloSharding::Replicate().ToProto();
   shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
                                      /*padded=*/false);
   EXPECT_EQ(shards.size(), 8);
@@ -165,7 +163,7 @@ TEST_F(XLAShardingTest, ShardTensor) {
   // last shard size should be smaller in dim=2 because it's not evenly
   // divisible.
   tensor = at::ones({1, 8, 7, 4}, at::TensorOptions(at::kFloat));
-  tensor_shape = CreateComputationShapeFromTensor(tensor, nullptr);
+  tensor_shape = CreateComputationShapeFromTensor(tensor, GetDefaultDevice());
   xla::Array4D<int64_t> tesseract({{{{0, 1}, {2, 3}, {4, 5}, {6, 7}}}});
   sharding = xla::HloSharding::Tile(tesseract).ToProto();
   sharding_spec =
@@ -187,7 +185,7 @@ TEST_F(XLAShardingTest, ShardTensor) {
   // last shard size should be smaller in dim=2 because it's not evenly
   // divisible.
   tensor = at::ones({10, 1, 8, 7, 4}, at::TensorOptions(at::kFloat));
-  tensor_shape = CreateComputationShapeFromTensor(tensor, nullptr);
+  tensor_shape = CreateComputationShapeFromTensor(tensor, GetDefaultDevice());
   xla::Array<int64_t> hypercube(std::vector<int64_t>{1, 1, 2, 2, 2});
   hypercube.FillIota(0);
   sharding = xla::HloSharding::Tile(hypercube).ToProto();
@@ -212,7 +210,8 @@ TEST_F(XLAShardingTest, ShardTensorMultiHost) {
 
   // 2D tiled, The first dim is halved and the last replicated.
   at::Tensor tensor = at::ones({8, 7, 4}, at::TensorOptions(at::kFloat));
-  xla::Shape tensor_shape = CreateComputationShapeFromTensor(tensor, nullptr);
+  xla::Shape tensor_shape =
+      CreateComputationShapeFromTensor(tensor, GetDefaultDevice());
   xla::Array2D<int64_t> mesh({
       {4, 5, 0, 1},
       {6, 7, 2, 3},
@@ -234,8 +233,7 @@ TEST_F(XLAShardingTest, ShardTensorMultiHost) {
       {0, 1, 4, 5},
       {2, 3, 6, 7},
   });
-  sharding = xla::HloSharding::Tile(mesh).ToProto();
-  sharding_spec->sharding = sharding;
+  sharding_spec->sharding = xla::HloSharding::Tile(mesh).ToProto();
   shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
                                      /*padded=*/false);
   EXPECT_EQ(shards.size(), 4);
@@ -248,25 +246,25 @@ TEST_F(XLAShardingTest, ShardTensorMiniBatch) {
   at::Tensor minibatch_tensor =
       at::ones({8, 7, 4}, at::TensorOptions(at::kFloat));
   xla::Shape tensor_shape =
-      CreateComputationShapeFromTensor(minibatch_tensor, nullptr);
+      CreateComputationShapeFromTensor(minibatch_tensor, GetDefaultDevice());
   tensor_shape.set_dimensions(
       0, minibatch_tensor.sizes()[0] * 2);  // Assuming 2 hosts
-  xla::Array2D<int64_t> mesh({
-      {0},
-      {1},
-      {2},
-      {3},
-      {4},
-      {5},
-      {6},
-      {7},
+  xla::Array3D<int64_t> mesh({
+      {{0}},
+      {{1}},
+      {{2}},
+      {{3}},
+      {{4}},
+      {{5}},
+      {{6}},
+      {{7}},
   });
 
   auto sharding = xla::HloSharding::Tile(mesh).ToProto();
   auto sharding_spec = std::make_shared<XLATensor::ShardingSpec>(
       sharding, tensor_shape, /*minibatch=*/true);
   auto shards = ShardingUtil::ShardTensor(minibatch_tensor, sharding_spec,
-                                          devices, /*padded=*/false);
+                                          devices, /*padded=*/true);
   EXPECT_EQ(shards.size(), 4);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({2, 7, 4}));
   EXPECT_EQ(shards[3].sizes(), c10::ArrayRef<long>({2, 7, 4}));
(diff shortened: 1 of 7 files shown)


torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -457,9 +480,12 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
     XLA_CHECK(sharding.tile_shape().dimensions_size() <= 2);
     XLA_CHECK(tensor.sizes().size() >= sharding.tile_shape().dimensions_size());
 
-    auto shard_shape = GetShardShape(tensor, sharding);
+    auto shard_shape = GetShardShape(tensor.sizes().vec(), sharding);
```
jonb377, line 483: We should use the global shape to determine the shard shape here

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -790,12 +790,13 @@ void InitXlaModuleBindings(py::module m) {
...
                        const py::list& group_assignment,
-                       const py::list& replication_groups, int sharding_type) {
+                       const py::list& replication_groups, int sharding_type,
+                       bool minibatch) {
         return std::make_shared<XLATensor::ShardingSpec>(
             ShardingUtil::CreateOpSharding(
                 tile_assignment, group_assignment, replication_groups,
                 ShardingUtil::ShardingType(sharding_type)),
-            CreateComputationShapeFromTensor(tensor, nullptr));
+            CreateComputationShapeFromTensor(tensor, nullptr), minibatch);
```
jonb377, line 799: We need to attach the global shape to the ShardingSpec here. `CreateComputationShapeFromTensor(tensor, nullptr)` will return the shape of the minibatch.

I think we get the correct xla::Shape by using `CreateComputationShapeFromTensor` on the minibatch to get the local shape and use `set_dimensions` to set dim 0 to `tensor.sizes[0] * global_device_count / local_device_count`.

We'll also need to enforce that all devices in the tiling assignment are on the batch axis.

torch_xla/csrc/tensor_util.cpp
```diff
@@ -937,8 +937,10 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
       std::vector<std::string> local_devices =
           runtime::GetComputationClient()->GetLocalDevices();
       xla::OpSharding sharding;
+      bool minibatch;
```
jonb377, line 940: We should initialize, `bool minibatch = false;`

torch_xla/csrc/tensor_util.cpp
```diff
@@ -947,8 +949,13 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
       // Shards the input tensors with padding, to split evenly.
       // The execution requires consistent shard sizes, and the zero-padded
       // values should be ignored.
-      std::vector<at::Tensor> local_shards = ShardingUtil::ShardTensor(
-          tensors[i], sharding, local_devices, /*padded=*/true);
+      std::vector<at::Tensor> local_shards =
+          ShardingUtil::ShardTensor(tensors[i], sharding, local_devices,
+                                    /*padded=*/true, /*minibatch=*/minibatch);
+      if (minibatch) {  // change global shape as tensor is already sharded
```
jonb377, line 955: We can read the global shape from the ShardingSpec.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -401,42 +402,64 @@ ShardingUtil::GetShardIndicesForDevices(
     std::fill_n(shard_indices.begin(), shard_indices.size(), indices);
   } else if (sharding.type() == xla::OpSharding::OTHER) {
     auto device_index = build_index_map(devices);
-    for (size_t i = 0; i < sharding.tile_assignment_devices().size(); i++) {
-      int64_t core = sharding.tile_assignment_devices()[i];
-      if (device_index.find(core) == device_index.end()) {
-        // Skip any shards whose device is not part of the `devices` list.
-        continue;
+    if (minibatch) {
```
jonb377, line 405: We can simplify the logic some here:

```
  for (int i = 0; i < devices.size(); i++) {
    auto indices = std::vector<at::indexing::TensorIndex>(tensor_shape.size(), at::indexing::Slice());
    indices[0] = at::indexing::Slice(i * shard_shape[0], (i + 1) * shard_shape[0]);
    shard_indices.push_back(slices);
  }
```

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -385,7 +386,7 @@ std::vector<std::vector<at::indexing::TensorIndex>>
 ShardingUtil::GetShardIndicesForDevices(
```
jonb377, line 386: I'm thinking we should leave this function untouched and add a new helper function to get indices for minibatch sharding. Wdyt?

torch_xla/csrc/xla_sharding_util.h
```diff
@@ -87,8 +87,8 @@ class ShardingUtil {
   // Returns the shape of the resulting shards of `tensor` after applying
   // `sharding`. This assumes the shards will be padded to ensure they all
   // have the same shape.
-  static std::vector<int64_t> GetShardShape(const at::Tensor& tensor,
-                                            const xla::OpSharding sharding);
+  static std::vector<int64_t> GetShardShape(
+      const std::vector<int64_t>& tensor_shape, const xla::OpSharding sharding);
```
jonb377, line 91: Let's rename this parameter `global_shape`

alanwaketan:
Do we have a design doc for this? It seems quite a large change in the user experience land. Trying to understand more about this before reviewing the actual code.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -457,9 +482,18 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
     XLA_CHECK(sharding.tile_shape().dimensions_size() <= 2);
     XLA_CHECK(tensor.sizes().size() >= sharding.tile_shape().dimensions_size());
 
-    auto shard_shape = GetShardShape(tensor, sharding);
-    auto shard_indices = GetShardIndicesForDevices(
-        shard_shape, tensor.sizes().vec(), sharding, devices);
+    auto shard_shape = GetShardShape(tensor.sizes().vec(), sharding);
+    if (minibatch) {
+      shard_shape[0] = tensor.sizes().vec()[0] / devices.size();
```
jonb377, line 487: Can we call `GetShardShape` with the global shape instead of modifying the shard shape here? I think you'd just need to change the `sharding` parameter from `xla::OpSharding` to `ShardingSpecPtr`, then the global shape can come from the ShardingSpec,

We could then also get rid of the `minibatch` parameter

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -790,12 +790,26 @@ void InitXlaModuleBindings(py::module m) {
...
                        const py::list& group_assignment,
-                       const py::list& replication_groups, int sharding_type) {
+                       const py::list& replication_groups, int sharding_type,
+                       bool minibatch) {
+        xla::Shape tensor_shape =
+            CreateComputationShapeFromTensor(tensor, nullptr);
+        int num_local_devices =
+            runtime::GetComputationClient()->GetLocalDevices().size();
+        int num_global_devices =
+            runtime::GetComputationClient()->GetAllDevices().size();
```
jonb377, line 800: These can go in the `if (minibatch)` block

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -790,12 +790,27 @@ void InitXlaModuleBindings(py::module m) {
       m, "XlaShardingSpec")
       .def(py::init([](at::Tensor tensor, const py::list& tile_assignment,
                        const py::list& group_assignment,
-                       const py::list& replication_groups, int sharding_type) {
+                       const py::list& replication_groups, int sharding_type,
+                       bool minibatch) {
+        xla::Shape tensor_shape =
```
jonb377, line 795: Can we rename this `global_shape` to be more explicit?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1497,14 +1512,15 @@ void InitXlaModuleBindings(py::module m) {
...
         << "Replicated tensor should not be loaded from _load_local_shards - "
            "use copy_";
-    auto shard_shape = ShardingUtil::GetShardShape(tensor, sharding);
+    auto shard_shape = ShardingUtil::GetShardShape(sharding_spec);
     for (auto shard : shards) {
       XLA_CHECK(shard.sizes() == shard_shape)
-          << "Input shard shape must include padding: " << shard.sizes()
-          << " vs " << shard_shape;
+          << "Input shard shape must include padding: " << shard.sizes();
+      // << " vs " << shard_shape;
```
jonb377, line 1523: This should be okay to add back in since `GetShardShape` returns a vector

torch_xla/csrc/tensor_util.cpp
```diff
@@ -947,8 +949,13 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
       // Shards the input tensors with padding, to split evenly.
       // The execution requires consistent shard sizes, and the zero-padded
       // values should be ignored.
-      std::vector<at::Tensor> local_shards = ShardingUtil::ShardTensor(
-          tensors[i], sharding, local_devices, /*padded=*/true);
+      std::vector<at::Tensor> local_shards =
+          ShardingUtil::ShardTensor(tensors[i], shardings[i], local_devices,
```
jonb377, line 953: `shardings[i]` could be nullptr, in which case we want to pass in a REPLICATED sharding spec to ShardTensor. Or, we can handle a null input in this if block: https://github.com/pytorch/xla/blob/8ada333234e376ed217b1937a0ce2da86b832d70/torch_xla/csrc/xla_sharding_util.cpp#L477

torch_xla/csrc/tensor_util.cpp
```diff
@@ -947,8 +949,13 @@ std::vector<torch::lazy::BackendDataPtr> CreateTensorsData(
...
+                                    /*padded=*/true);
+      if (minibatch) {  // change global shape as tensor is already sharded
+                        // accross batch dimesion.
+        shape = shardings[i]->shape;
+      }
       new_handles.push_back(ShardingUtil::CreateShardedData(
           local_shards, local_devices, shape, sharding));
```
jonb377, line 960: This is turning into a larger refactor, but can we also just pass the `ShardingSpecPtr` into `CreateShardedData`?

Then we won't need L955-958 and L939-948

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -326,13 +327,13 @@ std::vector<runtime::ComputationClient::DataPtr> ShardingUtil::OutputHandler(
     XLATensor::ShardingSpecPtr sharding = sharding_specs[i];
     if (replicated_output && sharding &&
         (sharding->sharding.type() != xla::OpSharding::REPLICATED)) {
-      XLA_CHECK(sharding->shape.has_value())
-          << "Sharding or Wrapping data shards in OutputHandler requires "
-             "unpartitioned tensor shape.";
+      // XLA_CHECK(sharding->shape.has_value())
+      //     << "Sharding or Wrapping data shards in OutputHandler requires "
+      //        "unpartitioned tensor shape.";
```
jonb377, line 332: Should this be uncommented?

khatwanimohit: Since we have made sharding->shape no Optional now. I'm not sure what are we checking here ?

jonb377: Oh gotcha, I didn't read the whole `XLA_CHECK`, you can just delete this

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -790,12 +790,27 @@ void InitXlaModuleBindings(py::module m) {
...
+                       const py::list& replication_groups, int sharding_type,
+                       bool minibatch) {
+        xla::Shape tensor_shape =
+            CreateComputationShapeFromTensor(tensor, nullptr);
+        if (minibatch) {
+          int num_local_devices =
+              runtime::GetComputationClient()->GetLocalDevices().size();
+          int num_global_devices =
+              runtime::GetComputationClient()->GetAllDevices().size();
+          XLA_CHECK(tile_assignment.size() == num_global_devices)
```
alanwaketan, line 802: Naive question: how does this guarantee that only the batch dimension is sharded?

khatwanimohit: In order to shard the input only along the batch axis. User would create a mesh with all the devices in the first dimension ( which will create a tile assignment like an array of all devices)

torch_xla/csrc/tensor_util.cpp
alanwaketan: I'm also curious on the deletion of the null check.

torch_xla/csrc/tensor.h
```diff
@@ -253,13 +253,17 @@ class XLATensor : public torch::lazy::LazyTensor {
...
     ShardingSpec(const xla::OpSharding& sharding, const xla::Shape& shape)
         : sharding(sharding), shape(shape) {}
+    ShardingSpec(const xla::OpSharding& sharding, const xla::Shape& shape,
+                 const bool& minibatch)
+        : sharding(sharding), shape(shape), minibatch(minibatch) {}
 
     xla::OpSharding sharding;
     // Optional source tensor shape unpartitioned.
-    std::optional<xla::Shape> shape;
+    xla::Shape shape;
```
alanwaketan, line 264: Why this is not optional anymore? Then I guess you need to delete the above comment as well.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -352,35 +350,62 @@ std::vector<runtime::ComputationClient::DataPtr> ShardingUtil::OutputHandler(
...
+    const std::vector<std::string>& devices) {
+  std::vector<std::vector<at::indexing::TensorIndex>> shard_indices(
+      devices.size());
+  if (sharding.type() == xla::OpSharding::OTHER) {
+    for (int i = 0; i < devices.size(); i++) {
+      std::vector<at::indexing::TensorIndex> indices;
+      for (int j = 0; j < tensor_shape.size(); j++) {
+        indices.push_back(at::indexing::Slice(0, tensor_shape[j]));
+      }
+      indices[0] =
```
alanwaketan, line 399: Why overwrites the 0th index?

alanwaketan: We only care about index 0 because that's the batch dim and then the remaining are just copys.

torch_xla/csrc/tensor.h
alanwaketan: Now that global tensor's shape doesn't reflect the truth.

torch_xla/csrc/tensor_util.cpp
alanwaketan: This is handled within ShardingUtil::ShardTensor line 485..

alanwaketan (approved):
LGTM. Please address @jonb377 comments as well.

torch_xla/csrc/xla_sharding_util.h
```diff
@@ -99,6 +99,14 @@ class ShardingUtil {
                             const xla::OpSharding sharding,
                             const std::vector<std::string>& devices);
 
+  // Returns the indices for the shards. Supports `OTHER` sharding types and
+  // called when input is sharded along the batch axis.
+  static std::vector<std::vector<at::indexing::TensorIndex>>
+  GetShardIndicesForMinibatchTensor(const std::vector<int64_t>& shard_shape,
```
alanwaketan, line 105: shard_shape is the local shard shape for the host. tensor_shape is the global tensor shape. devices is the tpu devices for the local host.

test/cpp/test_xla_sharding.cpp
```diff
@@ -208,23 +233,58 @@ TEST_F(XLAShardingTest, ShardTensorMultiHost) {
...
   EXPECT_EQ(shards.size(), 4);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({4, 2, 4}));
   EXPECT_EQ(shards[3].sizes(), c10::ArrayRef<long>({4, 1, 4}));
 }
 
+TEST_F(XLAShardingTest, ShardTensorMiniBatch) {
+  std::vector<std::string> devices = {"TPU:4", "TPU:5", "TPU:6", "TPU:7"};
+  at::Tensor minibatch_tensor =
+      at::ones({8, 7, 4}, at::TensorOptions(at::kFloat));
+  xla::Shape tensor_shape =
```
jonb377, line 248: nit: `global_shape` instead of `tensor_shape`

jonb377 (approved):
LGTM, great stuff Mohit! 🚀

test/cpp/test_xla_sharding.cpp
```diff
@@ -208,23 +233,58 @@ TEST_F(XLAShardingTest, ShardTensorMultiHost) {
...
-      ShardingUtil::ShardTensor(tensor, sharding, devices, /*padded=*/false);
+  sharding_spec->sharding = xla::HloSharding::Tile(mesh).ToProto();
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/false);
   EXPECT_EQ(shards.size(), 4);
   EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({4, 2, 4}));
   EXPECT_EQ(shards[3].sizes(), c10::ArrayRef<long>({4, 1, 4}));
 }
 
+TEST_F(XLAShardingTest, ShardTensorMiniBatch) {
```
miladm, line 244: @khatwanimohit did we tests `minibatch=True` on SPMD 2D sharding?

cc @tengyifei

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -365,35 +363,61 @@ std::vector<runtime::ComputationClient::DataPtr> ShardingUtil::OutputHandler(
...
+
     return shard_shape;
   } else {
     TF_LOG(ERROR) << "Unsupported OpSharding type " << sharding.type();
   }
 }
 
+std::vector<std::vector<at::indexing::TensorIndex>>
+ShardingUtil::GetShardIndicesForMinibatchTensor(
+    const std::vector<int64_t>& shard_shape,
```
miladm, line 401: @khatwanimohit wonder if we can revisit the `minibatch` logic for the 2D model sharding scheme.

More specifically, I am curious to learn if the logic around `shard_shape` properly handles 2D sharding.

cc @tengyifei

miladm: looks like 2D + minibatch is disallowed; @khatwanimohit to help confirm.

https://github.com/pytorch/xla/blob/2f5dc1f10e41afbf1c73497ea426a79728b2a1bd/torch_xla/csrc/runtime/tf_logging.h

miladm: cc @jonb377

PR #5332, opened 2023-07-21
Disable cxx_abi when building PyTorch/XLA in r2.0.

vanbasten23:
Following Mateusz's [suggestion](https://github.com/pytorch/xla/issues/5325#issuecomment-1643808739) to disable cxx_abi for pytorch/xla branch r2.0.

PR #5106, opened 2023-06-01
Checkout Ansible setup at origin's master, not local, untracked master.

mateuszlewko:
This will ensure that older versioned builds get the newest Ansible setup version.

PR #5104, opened 2023-05-31
Backport infra/ansible to r2.0

wonjoolee95:
Backport infra/ansible to r2.0

PR #5340, opened 2023-07-25
Backport infra/ansible to r1.13

vanbasten23:
This PR is similar to what we did to r2.0 in https://github.com/pytorch/xla/pull/5104 for the similar reason.

The reason why we need to do this is: I need to revert this PR https://github.com/pytorch/xla/pull/5106 so that terraform/ansible can pick up my [change](https://github.com/pytorch/xla/pull/5332/files) on branch r2.0. Without doing this, the 1.13 build will fail because it cannot find ansible on r1.13 branch.

### DIFF
--- a/infra/ansible/.ansible-lint
+++ b/infra/ansible/.ansible-lint
@@ -1,0 +1,6 @@
+---
+# .ansible-lint
+
+profile: moderate
+skip_list:
+  - schema[tasks]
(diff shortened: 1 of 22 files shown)


vanbasten23:
Pending work: update tag to the latest commit.

wonjoolee95 (approved):

vanbasten23:
it looks like we don't want to touch the r1.13 branch and don't need more wheels hence closing it.

PR #5342, opened 2023-07-25
Make all-reduce a no-op when world size is 1

will-cromar:
All reduce was a no-op if `xm.set_replication` was not called before https://github.com/pytorch/xla/commit/397ea4d571c7b6363779a5b960afe3e71132e0b7. This lets you easily debug scripts with one process, even if there are multiple runtime devices present. If there are multiple runtime devices, you must run the collective on all of them.

Example command that hangs before this PR (since MP collectives don't actually work on CPU):

```
PJRT_DEVICE=CPU CPU_NUM_DEVICES=2 python pytorch/xla/test/test_train_mp_mnist.py --fake_data --num_cores=1
```

More importantly, this allows single-process debugging (ie `xmp.spawn(..., nprocs=1)`) on TPU v2/v3, where there are always at least two runtime devices.

It looks like this logic only applied to `all_reduce` in the last release. I can add other no-ops in a follow-up if we believe that is the correct behavior for all of them.

### DIFF
--- a/test/pjrt/test_runtime_multi_cpu.py
+++ b/test/pjrt/test_runtime_multi_cpu.py
@@ -4,6 +4,7 @@ import queue
 from absl.testing import absltest, parameterized
 
 import torch
+import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.core.xla_env_vars as xenv
 import torch_xla.distributed.xla_multiprocessing as xmp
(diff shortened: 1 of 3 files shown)


alanwaketan:
That's interesting. I don't know what my PR has modified to lead to this regression, but thanks for fixing it.

alanwaketan:
Now I got it... Yea, I missed the clone part...

alanwaketan (approved):
LGTM.

JackCaoG (approved):

ManfeiBai (approved):

Issue #5325, opened 2023-07-19
Can PyTorch/XLA wheel for release branch build with cxx_abi disabled?


vanbasten23:
Hi @mateuszlewko ,

I wonder if the new wheel build process (with ansible) can disable cxx_abi when it builds a torch_xla wheel for a release branch (such as r2.0). We recently build a torch_xla wheel (on pt/xla branch r2.0, cuda 11.8, python=3.10). From the [log](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/e115b688-f467-45e4-b19c-a627de605a74?e=13803378&jsmode=o&mods=allow_workbench_image_override&project=tpu-pytorch-releases), it seems it still enables cxx_abi (I see `-D_GLIBCXX_USE_CXX11_ABI=1` in the log above, which make me think it enables cxx_abi. Please correct me if I'm wrong.). Building an official torch_xla with cxx_abi enabled causes torch_xla wheel to be incompatible with torch's wheel.

What we used to do in the release branch, is to first apply a torch patch (as in this [pr](https://github.com/pytorch/xla/commit/dffb9157f16d2fe36d8b56a811840e2f4cb9a566)), then disable the cxx_abi (as in this [pr](https://github.com/pytorch/xla/pull/4134)). So my question is
1. With ansible, does the wheel have cxx abi enabled?
2. If so for above question, with ansible, is it possible to apply the torch patch file first and then set the flag to be false?

Thanks.

cc: @JackCaoG @miladm

mateuszlewko:
Hey,

First some background information.

*Building process*

1. All build related tasks are present in this role:
https://github.com/pytorch/xla/blob/master/infra/ansible/roles/build_srcs/tasks/main.yaml.
They should be self-descriptive, but please reach out if something is
unclear.
2. For non-nightly releases, you should look at Ansible setup version at a
given tag, branch or commit, so in this case branch r2.0:
infra/ansible/roles/build_srcs/tasks/main.yaml#L32-L36
<https://github.com/pytorch/xla/blob/r2.0/infra/ansible/roles/build_srcs/tasks/main.yaml#L32-L36>
-
this is also the step that builds torch_xla.
3. Each ansible.builtin.command task in Ansible has a separate shell
environment, i.e. previous tasks are not polluting env vars of other tasks.
4. Having said that, most tasks are loading env_vars
[...]

PR #5344, opened 2023-07-25
Check out the release branch instead of origin/master

vanbasten23:
See https://github.com/pytorch/xla/issues/5325 for details. In summary, I disabled cxx_abi in r2.0 branch but once we ran the trigger it didn't pick up my change. Looking at the [log](https://github.com/pytorch/xla/issues/5325#issuecomment-1648791918), it first check out the correct commit (on r2.0 branch) and then switched back to origin/master (because of this [PR](https://github.com/pytorch/xla/pull/5106).). To unblock the dlvm p1 ticket (need py 3.10, cuda 11.8, r2.0 torch_xla wheel), I need to make the change.

### DIFF
--- a/infra/tpu-pytorch-releases/artifacts_builds.tf
+++ b/infra/tpu-pytorch-releases/artifacts_builds.tf
@@ -176,7 +176,6 @@ module "versioned_builds" {
 
   # Use Ansible setup from master branch for versioned release, because source
   # code at older version doesn't contain Ansible setup.
-  ansible_git_rev = "origin/master"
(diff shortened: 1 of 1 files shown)


ManfeiBai (approved):
LGTM

will-cromar (approved):
Thanks!

vanbasten23:
Thanks for the review!

vanbasten23:
Also checked with the team that we don't need to build wheels for r1.13 or r1.12 so we don't need to backport ansible to those 2 release branches like what we did in https://github.com/pytorch/xla/pull/5104.

PR #5348, opened 2023-07-26
Don't rewrite index hints in global save planning

jonb377:
For XLAShardedTensor, we expect the index hints to remain unchanged after global planning. The default save plan rewrites index hints to support ShardedTensor, which breaks multihost checkpointing. This change uses https://github.com/pytorch/pytorch/pull/105861 to avoid modifying index hints during global planning.

This will also enable distributed checkpointing tests in TPU CI, enables resharding tests, and fixes some bugs from `_CpuShard` support that were missed due to the lacking CI coverage.

The new test `test_multihost_checkpoint` will be used in a nightly test running on multiple hosts.

### DIFF
--- a/torch_xla/experimental/distributed_checkpoint.py
+++ b/torch_xla/experimental/distributed_checkpoint.py
@@ -92,7 +92,9 @@ class SPMDSavePlanner(SavePlanner):
         for k, v in state_dict.items()
         if _is_sharded_tensor(v) or isinstance(v, _CpuShards)
     }
-    unsharded = dict(state_dict.items() - self.sharded_state_dict.items())
+    unsharded = {
+        k: v for k, v in state_dict.items() if k not in self.sharded_state_dict
+    }
     self.unsharded_state_dict = tree_map(_unwrap_xla_sharded_tensor, unsharded)
 
   def create_local_plan(self) -> SavePlan:
@@ -112,7 +114,8 @@ class SPMDSavePlanner(SavePlanner):
     # Deduplicate write items across plans
     all_plans = dedup_tensors(all_plans)
 
-    global_plan, metadata = create_default_global_save_plan(all_plans)
+    global_plan, metadata = create_default_global_save_plan(
+        all_plans, rewrite_index_hints=False)
 
     # Combine mappings from all plans
     planner_data_dict = [p.planner_data for p in global_plan]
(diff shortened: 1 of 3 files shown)


torch_xla/experimental/distributed_checkpoint.py
```diff
@@ -92,7 +92,9 @@ def set_up_planner(self, state_dict: STATE_DICT_TYPE,
         for k, v in state_dict.items()
         if _is_sharded_tensor(v) or isinstance(v, _CpuShards)
     }
-    unsharded = dict(state_dict.items() - self.sharded_state_dict.items())
+    unsharded = {
```
alanwaketan, line 95: Curious on why you made this change?

torch_xla/experimental/distributed_checkpoint.py
```diff
@@ -340,13 +345,12 @@ def _create_write_items_for_xla_sharded_tensor(
 def _create_write_items_for_cpu_shards(
     fqn: str, cpu_shards: _CpuShards) -> List[WriteItem]:
   items = []
-  for xla_shard in cpu_shards.shards:
+  for shard_ind, xla_shard in enumerate(cpu_shards.shards):
```
alanwaketan, line 348: Is this the bug you are fixing?

alanwaketan (approved):
LGTM.

torch_xla/experimental/distributed_checkpoint.py
jonb377: This is to support `_CpuShard`. The old approach tries to hash the values of the dict as well, since `.items()` is `(key, value)` pairs. When we're checkpointing on CPU, the `_CpuShards` contains `List[XLAShard]` which isn't hashable.

torch_xla/experimental/distributed_checkpoint.py
jonb377: Yeah, what we had originally was incorrect. It looks like the CPU tests didn't actually hit this codepath, and we weren't running on TPU CI.

torch_xla/experimental/distributed_checkpoint.py
alanwaketan: Hmm, but  unsharded  is still a KV pair?

jonb377: The issue is with `state_dict.items() - self.sharded_state_dict.items()` - the `-` operator between two `dict_items` is a set difference, which will hash the entire `(k, v)` tuple. Using dict comprehension, only `k` is hashed.

alanwaketan: That's interesting. Thanks for the python lecture!

PR #5362, opened 2023-07-27
[Dynamo|TPU] Skip`DynamoTrainingBasicTest.test_resnet18` on TPU

ManfeiBai:
Before bring `test_dynamo.py` to TPU CI in https://github.com/pytorch/xla/pull/5369, set skip `DynamoTrainingBasicTest.test_resnet18` on TPU

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -277,9 +277,11 @@ class DynamoTrainingBasicTest(unittest.TestCase):
       cpu_data.requires_grad = True
       cpu_target = target.detach().cpu()
       cpu_output = self.train_model(resnet18, cpu_data, cpu_target)
-      self.assertTrue(
-          torch.allclose(
-              xla_output.cpu(), cpu_output.cpu(), rtol=1e-05, atol=1e-05))
+      # Disable the accuracy check here due to training is experimental.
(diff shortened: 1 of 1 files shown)


wonjoolee95:
Curious on which hardware this is failing -- I assume this is failing on TPU due to precision? Or is it failing on GPU as well?

ManfeiBai:
> Curious on which hardware this is failing -- I assume this is failing on TPU due to precision? Or is it failing on GPU as well?

didn't test on GPU, this failure appear on TPU, maybe we should make a patch for TPU only to fix this?

wonjoolee95:
The CI is green on head, which means it should succeed on CPU and GPU.

We should be able to easily skip it only on TPU with a `@skipif`, something like https://github.com/pytorch/xla/blob/master/test/test_operations.py#L59.

ManfeiBai:
> The CI is green on head, which means it should succeed on CPU and GPU.
> 
> We should be able to easily skip it only on TPU with a `@skipif`, something like https://github.com/pytorch/xla/blob/master/test/test_operations.py#L59.

thanks, @wonjoolee95, cool, really good point, has applied `@skipOnTpu`

wonjoolee95 (approved):

PR #5392, opened 2023-08-01
Add manual seed in multihost checkpoint test

jonb377:
In a multihost environment, we need all hosts to agree on the initial state of the tensors. This change sets the seed and moves parameters to CPU before comparison.


JackCaoG (approved):

alanwaketan (approved):
LGTM.

PR #5401, opened 2023-08-02
Update CI image with Ubuntu 20.04 base image

lsy323:
Try run CI with pre-built ubuntu 20.04 base image. See if Ubuntu based GPU image also hit the GPU test hanging issue in #5290

Thanks @yeounoh for providing the image for testing.


lsy323:
The image doesn't work with the current CI script out-of-box. I tried to build PyTorch and PyTorch/XLA in the ubu20 container locally. It can compile with some tweaks, will add the tweaks to the CI script and have an other CI run.

PR #5403, opened 2023-08-02
Fix transferToServer time

chandrasekhard2:
Fix Transfer To Server Time in profiler

### DIFF
--- a/test/test_metrics.py
+++ b/test/test_metrics.py
@@ -185,7 +185,14 @@ class MetricsTest(unittest.TestCase):
     # of `ExecuteComputation`, but the actual async time.
     self.assertGreater(execute_time_ns, .5 * wall_time_ns)
 
+  def test_transfer_to_server_time(self):
+    device = xm.xla_device()
+    tensor = torch.ones(128, 3, 224, 224)
+    begin = time.perf_counter_ns()
+    tensor = xm.send_cpu_data_to_device(tensor, device)
+    wall_time_ns = time.perf_counter_ns() - begin
+    transfer_time_ns = met.metric_data('TransferToServerTime')[1]
+    self.assertIn("TransferToServerTime", met.metric_names())
+    self.assertGreater(transfer_time_ns, wall_time_ns)
 
 if __name__ == '__main__':
-  test = unittest.main()
-  sys.exit(0 if test.result.wasSuccessful() else 1)


test/test_metrics.py
```diff
@@ -185,6 +185,16 @@ def test_execute_time_metric(self):
...
 
+  def test_transfer_to_server_time(self):
+    device = xm.xla_device()
+    tensor = torch.ones(128, 3, 224, 224)
+    begin = time.perf_counter_ns()
+    tensor = xm.send_cpu_data_to_device(tensor, device)
+    wall_time_ns = time.perf_counter_ns() - begin
+    transfer_time_ns = met.metric_data('TransferToServerTime')[1]
+    self.assertIn("TransferToServerTime", met.metric_names())
+    self.assertGreater(transfer_time_ns, wall_time_ns)
```
will-cromar, line 196: Can you add a comment explaining the test condition here?

Do you need to add a mark_step after you record the wall_time_ns to make sure the transfer is complete before you check metrics?

test/test_metrics.py
```diff
@@ -185,6 +185,16 @@ def test_execute_time_metric(self):
     # of `ExecuteComputation`, but the actual async time.
     self.assertGreater(execute_time_ns, .5 * wall_time_ns)
 
+  def test_transfer_to_server_time(self):
```
will-cromar, line 188: Does this test case work on GPU/TPU? I'm asking because we added a skip for the execute time metric test when using CPU.

chandrasekhard2: @JackCaoG Confirmed that it works across GPU&TPU

test/test_metrics.py
chandrasekhard2: I'm trying to send 128 batch images and checking if the TransferToServer Time is more than the wall time. Based on the logic here - torch_xla/csrc/runtime/pjrt_computation_client.cc  line 262, I believe the transfer to server time is more than the wall time , but it isn't the case for some reason. I see the output of the test as

Even with mark_step(), the transfer to server time is less than the wall time.

```
AssertionError: 143920.0[TransferToServer Time] not greater than 33337409[Wall time] 
```

Am I doing the right test here?

will-cromar: This is probably just a matter of the overhead being much larger than the transfer time in practice. For the test case above, we had to choose an arbitrary factor of `.5` experimentally. Before the change, execute time was normally much less than half of the wall time, and after, it was normally much more. It's hacky, but I don't know how to make the test assertion more directly.

I suggest you try timing `TransferToServer` before and after your change, and then pick a factor that would make the test fail before your change and pass after.

PR #5404, opened 2023-08-03
Fix ReplicateShardedData for int type

JackCaoG:
We used to perform `res = (res/2) + (res/2)` as identify function to get the unsharded result. However the `res/2` part will make result always even for the int type. This pr performs a `res + 0` and force the compiler not to overwrite the output sharding instead.

### DIFF
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -31,6 +31,8 @@
 #include "xla/shape.h"
 #include "xla/stream_executor/tpu/tpu_initializer_helper.h"
 
+using xla::internal::XlaBuilderFriend;
+
 namespace torch_xla {
 namespace runtime {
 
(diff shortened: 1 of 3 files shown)


torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -324,21 +326,22 @@ ComputationClient::DataPtr PjRtComputationClient::ReplicateShardedData(
...
-    b.SetSharding(xla::HloSharding::Replicate().ToProto());
-    xla::XlaOp scalar_two_op =
-        xla::ConvertElementType(xla::ConstantR0(&b, 2), shape.element_type());
-    auto y = xla::Div(x, scalar_two_op);
-    auto z = xla::Add(y, y);
+    xla::XlaOp x = xla::Parameter(&builder, 0, shape, "p0");
+    builder.SetSharding(xla::HloSharding::Replicate().ToProto());
+    xla::XlaOp scalar_zero_op = xla::ConvertElementType(
+        xla::ConstantR0(&builder, 0), shape.element_type());
+    xla::XlaOp y = xla::Add(x, scalar_zero_op);
```
jonb377, line 339: Is it possible to have a totally no-op computation and remove the `Add`? Since we're setting `allow_spmd_sharding_propagation_to_output=false`, the result will always be replicated regardless of the input sharding

JackCaoG: let me try

JackCaoG: commented out the `xla::add` gives incorrect result. I think there are ways to make it work but since it is not a priority I will leave that to a separate pr.

jonb377: Got it, thanks for trying.

jonb377 (commented):
LGTM

JackCaoG:
> LGTM

lol you did not hit the approve button.

jonb377 (approved):
LGTM (second attempt)

PR #5409, opened 2023-08-04
Partially replicate lower-rank tensors

jonb377:
The current support for sharding a rank n tensor across a rank (n+1) mesh requires initially replicating tensors before reshaping and sharding. This change will instead use the last dimension to determine replication groups and use partial replication to shard the tensor.

PR #5411, opened 2023-08-05
SPMD cross slice-replication using partial_replication sharding

yeounoh:
This reverts some changes made in #5305 , #5409 to make partial replication support more general and not restrictive to certain conditions. 

This addresses unordered partial replication, and includes a test case with 3D unordered partial replication, modified the case with 2D tensor over 3D mesh sharding.


test/spmd/test_xla_sharding.py
```diff
@@ -318,78 +317,21 @@ def test_mark_sharding_not_ordered_partial_3d(self):
     # Shard along two axes if four or more devices are available
     z_dim = 2 if self.n_devices >= 4 else 1
     mesh = self._get_mesh((z_dim, 1, self.n_devices // z_dim))
-
-    # Expect local shard size to be [8, 16 / z_dim, 32]
-    xt1 = xs.mark_sharding(t1, mesh, (1, 0, None))
-
-    for local_shard in xt1.local_shards:
-      self.assertEqual(local_shard.data.size()[0], 8)
```
JackCaoG, line 326: any reason we removed the local shard size check here?

test/spmd/test_xla_sharding.py
```diff
@@ -638,14 +580,12 @@ def test_2d_tensor_3d_mesh(self):
 
     t1 = ct1.to(xm.xla_device())
     t2 = ct2.to(xm.xla_device())
-    mesh = self._get_mesh((1, self.n_devices, 1))
-    t1 = xs.mark_sharding(t1, mesh, partition_spec=(1, 2))
+    mesh = self._get_mesh((2, self.n_devices // 2, 1))
```
JackCaoG, line 583: wouldn't cpu test just fail at this case? since there is only 1 device.

test/spmd/test_xla_sharding.py
yeounoh: Thanks, I am going to work with @khatwanimohit @jonb377 to add/refactor the tests.

test/spmd/test_xla_sharding.py
yeounoh: ditto

torch_xla/experimental/xla_sharding.py
```diff
@@ -412,35 +402,24 @@ def mark_sharding(t: Union[torch.Tensor, XLAShardedTensor], mesh: Mesh,
...
-  tensor_expand = 0
-  if tensor_expand < len(mesh.get_logical_mesh().shape) - len(partition_spec):
-    tensor_expand = len(mesh.get_logical_mesh().shape) - len(partition_spec)
-    partition_spec = (None,) * tensor_expand + partition_spec
-    shape = (1,) * tensor_expand + (*original_shape,)
-    t = t.expand(shape)
-
   tile_assignment = _get_tile_assignment(mesh, partition_spec)
-  sharding_type = _get_sharding_type(partition_spec, num_devices)
+  mesh_size = 1 if type(mesh.mesh_shape) == int else len(mesh.mesh_shape)
```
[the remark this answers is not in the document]
yeounoh: Good question, single element int tuple could be passed as int.

torch_xla/experimental/xla_sharding.py
```diff
@@ -319,45 +319,35 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
+    partition_spec: Tuple[Union[int, None]]) -> Tuple[List, List]:
   group_assignment = list()
   replication_groups = list()
-  mesh_shape_list = list(torch.tensor(tile_assignment).size())
   if sharding_type is ShardingType.PARTIAL:
     # Shard across groups and replicate within subgroups; replicated dims
     # will be used to group replication devices.
     tile_dims = [d for d in partition_spec if d is not None]
-    replicated_dims = set(range(len(mesh_shape_list))) - set(tile_dims)
+    replicated_dims = set(range(len(mesh.mesh_shape))) - set(tile_dims)
```
JackCaoG, line 340: this doesn't seems to be used anywhere.

yeounoh: Legacy code. Thanks.

torch_xla/experimental/xla_sharding.py
```diff
@@ -319,45 +319,35 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
       group_list = _group_list
     replication_groups = [group.flatten().tolist() for group in group_list]
 
-    group_tile_shape = mesh_shape_list
-    for d in replicated_dims:
-      group_tile_shape[d] = 1
+    group_tile_shape = [mesh.mesh_shape[d] if d is not None else 1 for d in partition_spec]
     group_assignment = np.arange(len(replication_groups)).reshape(
         tuple(group_tile_shape)).tolist()
   return group_assignment, replication_groups
```
JackCaoG, line 353: can you add a comment here or in the function declaration regarding how to interpret `group_assignment` and `replication group`?

torch_xla/experimental/xla_sharding.py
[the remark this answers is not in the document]
jonb377: Single-element tuples should be declared as e.g. `(1,)` instead of `(1)`. I'm concerned about having to special-case for this, do we get much benefit from supporting both paradigms?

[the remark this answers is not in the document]
jonb377: We would need to handle it in other places too, like https://github.com/pytorch/xla/pull/5411/files#diff-66621cfc79f7e139b195eaee8de1df71b4870d57294fd8e416a2e7a597bea0b0R340

test/spmd/test_xla_sharding.py
yeounoh: @JackCaoG actually, just realized that this was replaced by the assertion in L334

torch_xla/experimental/xla_sharding.py
yeounoh: Done

torch_xla/experimental/xla_sharding.py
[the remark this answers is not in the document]
yeounoh: The input is type checkes `:Tuple` and should be passed as `(1,)`. I am also surprised to see that it's sometimes interpreted as int. Let me check.

test/spmd/test_xla_sharding.py
```diff
@@ -638,15 +580,20 @@ def test_2d_tensor_3d_mesh(self):
 
     t1 = ct1.to(xm.xla_device())
     t2 = ct2.to(xm.xla_device())
-    mesh = self._get_mesh((1, self.n_devices, 1))
-    t1 = xs.mark_sharding(t1, mesh, partition_spec=(1, 2))
+
```
yeounoh, line 583: cc @khatwanimohit @jonb377 I updated this unit test, could you confirm that this would be relevant for the multi-slice test cases, also note that in case of 1-device, it falls back to replicated.

khatwanimohit: Yes, this is how we shard model parameters in shardings.py

torch_xla/experimental/xla_sharding.py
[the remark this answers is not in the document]
yeounoh: Ok, found the bug in `test_mark_sharding_ir_with_multiple_output`, whcih caused the 
```
TypeError: object of type 'int' has no len()
```
error.

JackCaoG (approved):

test/spmd/test_xla_sharding.py
```diff
@@ -755,7 +701,7 @@ def test_mark_sharding_ir_with_multiple_output(self):
     # max return 2 tensors `value` and `indices`. They are the output
     # of the same IR Node `MaxInDim`
     (xt_val, xt_index) = torch.max(xt1, 1)
-    xst_val = xs.mark_sharding(xt_val, self._get_mesh((self.n_devices)),
```
yeounoh, line 758: cc @jonb377 this was the culprit, thanks for the comment :)

torch_xla/experimental/xla_sharding.py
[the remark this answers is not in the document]
yeounoh: https://github.com/pytorch/xla/pull/5411/files#r1286401442

jonb377 (approved):
LGTM, thanks Yeounoh!

alanwaketan (approved):
LGTM.

khatwanimohit (approved):
LGTM

Issue #5410, opened 2023-08-04
Torchbench dynamo training failure with openxla dynamo backend


JackCaoG:
## 🐛 Bug

with command
```
python benchmarks/dynamo/torchbench.py --randomize-input --performance --training --trace-on-xla --backend=openxla --only resnet50
```
after setting up the torchbench, we can see a error
```
RuntimeError: ./torch_xla/csrc/runtime/pjrt_computation_client.h:166 : Check failed: HasValue() 
*** Begin stack trace ***
	tsl::CurrentStackTrace[abi:cxx11]()
	torch_xla::runtime::PjRtComputationClient::PjRtData::GetOpaqueHandle()
	torch::lazy::LazyGraphExecutor::RunPostOrder(std::vector<torch::lazy::Value, std::allocator<torch::lazy::Value> > const&, torch::lazy::LazyGraphExecutor::SyncTensorCollection*)
	torch_xla::XLAGraphExecutor::RunPostOrder(std::vector<torch::lazy::Value, std::allocator<torch::lazy::Value> > const&, torch::lazy::LazyGraphExecutor::SyncTensorCollection*)
	torch_xla::XLAGraphExecutor::GetGraphHash(std::vector<c10::intrusive_ptr<torch_xla::XLATensor, c10::detail::intrusive_target_default_null_type<torch_xla::XLATensor> >, std::allocator<c10::intrusive_ptr<torch_xla::XLATensor, c10::detail::intrusive_target_default_null_type<torch_xla::XLATensor> > > > const&)
```

Note that we need to patch the benchmark script similar to https://github.com/pytorch/xla/issues/4174#issuecomment-1666268488

@shunting314 @jansel @wconstab @wonjoolee95 @alanwaketan

JackCaoG:
with some debugging message I can see that it is a tensor with shape `f32[32,1000]` being null.
```
buffer with shape f32[32,1000] on device TPU:0 is null
```

and I can see that we clear the value of this tensor
```
clear pending ir for tensor with size f32[32,1000]
```
because we think it is a garbage value created during cache warm up which is incorrect. Should be able to target this tensor pretty easily since there is only one tensor with shape `f32[32,1000]` and it is being inplcaed updated.

JackCaoG:
I think one of the cause of this issue is a python syntax mistake, we have logic to inplace update the argument with its clone if there is a inpalce update happened during tracing. our logic looks like
```
  if args_need_update_bool:
[...]

Issue #4174, opened 2022-11-09
Run backward on CPU before importing torch_xla cause future backward on XLA crash


shunting314:
## 🐛 Bug

I'm integrating torchdynamo with torchxla for training. One weird crash I encountered is:
```
RuntimeError: 0 <= device.index() && device.index() < static_cast<c10::DeviceIndex>(device_ready_queues_.size()) INTERNAL ASSERT FAILED at "/pytorch/torch/csrc/autograd/engine.cpp":1342, please report a bug to PyTorch.
```
when doing backward on an XLA model.

It turns out that this only happens if dynamo runs a backward pass on CPU before importing torch_xla.

This is definitely not an issue for dynamo. But I'm not sure if it should be an issue for pytorch or torchxla. Assigning it to torchxla for now.

cc @JackCaoG @wconstab  @jansel 

## To Reproduce

```
from torch import nn
import os

os.environ["GPU_NUM_DEVICES"] = "1"

class LinearModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(2, 3)

    def forward(self, inp):
        return self.linear(inp)

    def get_example_inputs(self):
        return (torch.rand(10, 2),)

model = LinearModel()
inputs = model.get_example_inputs()
# import torch_xla # uncomment this line to workaround
model(*inputs).sum().backward()

import torch_xla.core.xla_model as xm
xla_device = xm.xla_device()
model = model.to(device=xla_device)
inputs = map(lambda x: x.to(device=xla_device), inputs)
model(*inputs).sum().backward()
print("bye")
```

The above script will crash with error:
```
Traceback (most recent call last):
  File "myscripts/xla_model_backward_crash.py", line 27, in <module>
    model(*inputs).sum().backward()
  File "/pytorch/torch/_tensor.py", line 451, in backward
    self, gradient, retain_graph, create_graph, inputs=inputs
  File "/pytorch/torch/autograd/__init__.py", line 199, in backward
    allow_unreachable=True, accumulate_grad=True)  # Calls into the C++ engine to run the backward pass
RuntimeError: 0 <= device.index() && device.index() < static_cast<c10::DeviceIndex>(device_ready_queues_.size()) INTERNAL ASSERT FAILED at "/pytorch/torch/csrc/autograd/engine.cpp":1342, please report a bug to PyTorch.
```

Uncomment the line can work around the issue:
```
# import torch_xla # uncomment this line to workaround
```

## Environment

[...]

PR #5414, opened 2023-08-07
Fix the incorect clone arg condition in dynamo bridge

JackCaoG:
This should fix https://github.com/pytorch/xla/issues/5410. For more detail please take a look at https://github.com/pytorch/xla/issues/5410#issuecomment-1666333291

After this pr and a patch on the pytorch side (for more detail check https://github.com/pytorch/xla/issues/4174)

```
diff --git a/benchmarks/dynamo/common.py b/benchmarks/dynamo/common.py
index a354299d3ab..7134e1b238e 100644
--- a/benchmarks/dynamo/common.py
+++ b/benchmarks/dynamo/common.py
@@ -37,7 +37,7 @@ import torch.fx._pytree as fx_pytree
 from scipy.stats import gmean, ttest_ind
 from torch._dynamo.profiler import fx_insert_profiling, Profiler
 from torch._dynamo.testing import dummy_fx_compile, format_speedup, same
-from torch._dynamo.utils import clone_inputs, graph_break_reasons
+from torch._dynamo.utils import clone_inputs#, graph_break_reasons
 from torch._functorch.aot_autograd import set_model_name
 from torch._inductor import config as inductor_config
 from torch._inductor.utils import fresh_inductor_cache
@@ -55,6 +55,7 @@ except ImportError:
 
 try:
+    device = xm.xla_device()
 except ImportError:
     # ignore the error if torch_xla is not installed
     pass
@@ -666,6 +667,7 @@ def speedup_experiment(args, model_iter_fn, model, example_inputs, **kwargs):
     timings = np.zeros((args.repeat, 2), np.float64)
     # if we randomize the input, we should also check the result is correct
     should_check_result = should_randomize_input = args.randomize_input
+    should_check_result = False
 
     import contextlib
 
@@ -1074,20 +1076,21 @@ def baselines(models, model_iter_fn, example_inputs, args):
     return result
```

I am able to run the dynamo training again with 
```
:/src/pytorch# python benchmarks/dynamo/torchbench.py --randomize-input --performance --training --trace-on-xla --backend=openxla --only ${model}
```
and produce the new result in https://github.com/pytorch/xla/issues/4475#issuecomment-1668418731

TODO
1. currently if I don't pass the `--only` flag and try to run more model, the script will error out and seems to suggest there are multiple process trying to grab the device(I was using TPU to benchmark this)
2. Find a way to wrok around the pytorch patch or just merge it(should_check_result is due to cpu and TPU precision issue, it is a known limitation)
3. Above command is more to compare the `Dynamo` vs `Lazy`, we need a more proper way to compare different compilers(like XLA:GPU and inductor). The current known issue is when using XLA:GPU we always compile PyTorch with `USE_CUDA=0` and XLA with `XLA_CUDA=1` to avoid both pytorch and xla trying to grab the GPU at the same time. This should not be needed in theory, need to test that.
4. Setup a nightly torchbench run for both TPU and GPU for inference and training.

FYI @wconstab @jansel


shunting314 (approved):

wonjoolee95 (approved):
Thanks!

PR #5420, opened 2023-08-08
Update test_runtime_tpu.py

ManfeiBai:


ManfeiBai:
continued in the follow up PR

PR #5421, opened 2023-08-08
[PJRT|TPU] Update `test_xla_devices_single_process_all_chips` for expected device number

ManfeiBai:
For `test_xla_devices_single_process_all_chips`, edit expected device number to match expected TPU configs

### DIFF
--- a/test/pjrt/test_runtime_tpu.py
+++ b/test/pjrt/test_runtime_tpu.py
@@ -92,11 +92,10 @@ class TestExperimentalPjrtTpu(parameterized.TestCase):
     devices_per_process = pjrt.run_multiprocess(xm.xla_device)
     self.assertDictEqual(devices_per_process, expected)
 
+  @absltest.skipIf(tpu.num_available_devices() > 4, "Not implemented")
   def test_xla_devices_single_process_all_chips(self):
-    cores_per_process = tpu.num_available_devices()
     expected = _ordinal_to_device(
-        processes=1,
-        cores_per_process=cores_per_process if cores_per_process <= 4 else 4)
+        processes=1, cores_per_process=tpu.num_available_devices())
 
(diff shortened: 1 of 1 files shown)


test/pjrt/test_runtime_tpu.py
```diff
@@ -93,8 +93,10 @@ def test_xla_devices_multiprocess(self):
     self.assertDictEqual(devices_per_process, expected)
 
   def test_xla_devices_single_process_all_chips(self):
```
will-cromar, line 95: Since `TPU_VISIBLE_CHIPS` is hardcoded with 4 devices in this test case, I think it makes sense to skip it if case if `tpu.num_available_chips` is not `4`.

ManfeiBai: sounds good to me, will do

test/pjrt/test_runtime_tpu.py
```diff
@@ -92,6 +92,7 @@ def test_xla_devices_multiprocess(self):
     devices_per_process = pjrt.run_multiprocess(xm.xla_device)
     self.assertDictEqual(devices_per_process, expected)
 
+  @absltest.skipIf(tpu.num_available_devices() > 4, "Not implemented")
```
will-cromar, line 95: Please make this `tpu.num_available_chips`. This test case is still valid when there are multiple devices per chip. Also, it's safer to do `tpu.num_available_chips() != 4` here. I don't know the behavior of `TPU_VISIBLE_CHIPS=0,1,2,3` when there are fewer than 4 chips actually available.

ManfeiBai: Thanks, Will, updated

will-cromar (approved):

Issue #5419, opened 2023-08-08
PyTorch/XLA wheel for release branch build failed to find cudnn8.7.0


ManfeiBai:
Hi @mateuszlewko and @will-cromar,

I wonder how build process find specific cudnn version like `libcudnn8=8.7.0.84-1+cuda11.8`, do we installed it already and saved in a bucket or we could only access specific version with limitation?

now we want to build release branch(2.0) with cudnn8.7.0.84-1+cudnn11.8 like https://github.com/pytorch/xla/blob/r2.0/infra/ansible/config/cuda_deps.yaml, but failed with `msg: no available installation candidate for libcudnn8=8.7.0.84-1+cuda11.8`, we confirmed `libcudnn8=8.7.0.84-1+cuda11.8` according to [blog](https://stackoverflow.com/questions/76465821/how-to-modify-this-command-to-install-specific-cudnn-and-torch-version-apt-inst) due to no access to nvidia yet.

So my question is how to enable building system find and install this specific verison of cudnn(`libcudnn8=8.7.0.84-1+cuda11.8`)?

Thanks.

cc: @JackCaoG @miladm @will-cromar

mateuszlewko:
Hey Manfei,

It seems that previous APT repo for CUDA didn't contain such package version. I slightly updated the config and the CUDA apt repo. I created https://github.com/pytorch/xla/pull/5425. Note that PR will only affect nightly builds. To change r2.0, you should push similar change to the r2.0 branch.

Thanks,
Mateusz

ManfeiBai:
Thanks, will make a same PR for r2.0 too

vanbasten23:
It seems we still have such issue:
```
2023-09-18 22:05:13.946634: E external/xla/xla/stream_executor/cuda/cuda_dnn.cc:435] Loaded runtime CuDNN library: 8.7.0 but source was compiled with: 8.8.0.  CuDNN library needs to have matching major version and equal or higher minor version. If using a binary install, upgrade your CuDNN library.  If building from sources, make sure the library loaded at runtime is compatible with the version specified during compile configuration.
```
in our nightly docker image.

To repro:
```
(base) xiowei@xiowei-gpu:~$ sudo docker pull us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.8_cuda_11.8

nightly_3.8_cuda_11.8: Pulling from tpu-pytorch-releases/docker/xla
[...]

PR #5425, opened 2023-08-09
Add repo for libcudnn8=8.7.0.84 and CUDA 11.8

mateuszlewko:
1. Use debian11 repo for CUDA deps (seems to work anyway on debian 10).
2. Remove variables from nvidia repo key. The key is fixed (doesn't depend on repo, contrary to the docs). The latest version can be taken directly from nvidia's docs (added link).

This change will only affect nightly builds for CUDA 11.8.

Related: https://github.com/pytorch/xla/issues/5419


stgpetrovic (approved):

ManfeiBai (approved):
lgtm, verified on r2.0

Issue #5046, opened 2023-05-22
Load PjRT Plugin API for Custom PjRT Backend


aws-kingrj:
## ❓ Questions and Help

We are currently trying to use the load_pjrt_plugin() API for loading custom C-API pjrt plugins but the function is not exposed in torch_xla. I can see it defined here in the xla.cc file of the tensorflow xla commit used for torch_xla 2.0: https://github.com/tensorflow/tensorflow/blob/f7759359f8420d3ca7b9fd19493f2a01bd47b4ef/tensorflow/compiler/xla/python/xla.cc#L325 as a part of the xla_extension.so file as a pybind11 module but that xla_extensions.so file isn't exposed in torch_xla. 

Is there another way we can used this API in python to load our C-API pjrt plugin with the current torch_xla 2.0 or is there some other API exposed for us to load pjrt plugins?

will-cromar:
We don't support the xla_extensions pybind module in PyTorch/XLA. You have to load the module in C++ during the `ComputationClient` creation. For now, you can add a branch like this one here: https://github.com/pytorch/xla/blob/bc28ca581535b2c61af8187cc28743580e76b881/third_party/xla_client/pjrt_computation_client.cc#L109-L113

You can use the XPU PR as a reference to add a new plugin: https://github.com/pytorch/xla/pull/4891

I'm planning on updating this logic to allow dynamic plugin loading and remove all of those device type enums/lists, but that may have to wait until after we remove XRT.

amithrm:
Thanks @will-cromar ! We will try doing this and update you

aws-kingrj:
Neuron PR for this https://github.com/pytorch/xla/pull/5428/

aws-kingrj:
Neuron PR for import hook to be added: https://github.com/pytorch/xla/pull/5429

ysiraichi:
I believe we are already able to load plugins dynamically, so I'm closing this. Feel free to re-open this issue if needed.

PR #5428, opened 2023-08-09
Enable PJRT C API Client and other changes for Neuron

aws-kingrj:
- Adds ability to discover Neuron device in PJRT C API Client
- Adds support for pjrt spawn with neuron
- Adds master port support for pjrt backend with torchrun and neuron
- Adds typecasting defaults for NEURON device type
- Tested on AWS Neuron internal testing 
- https://github.com/pytorch/xla/issues/5046

### DIFF
--- a/torch_xla/_internal/pjrt.py
+++ b/torch_xla/_internal/pjrt.py
@@ -53,7 +53,7 @@ def _run_thread_per_device(
     Dict of the form {thread_rank: return_value}, where return_value is the
     result of calling `fn`.
   """
-  if runtime.device_type() == "NEURON":
+  if runtime.device_type() == 'NEURON':
     neuron.intialize_env(local_rank)
 
   initializer_fn(local_rank, local_world_size)
--- a/torch_xla/_internal/pjrt.py
+++ b/torch_xla/_internal/pjrt.py
@@ -109,8 +109,7 @@ def initialize_multiprocess(local_rank: int, local_world_size: int):
 
   if runtime.device_type() == 'TPU':
     tpu.configure_topology(local_rank, local_world_size)
-  
-  if runtime.device_type() == 'NEURON':
+  elif runtime.device_type() == 'NEURON':
     neuron.initialize_env(local_rank)
 
   devices = xm.get_xla_supported_devices()
--- a/torch_xla/_internal/pjrt.py
+++ b/torch_xla/_internal/pjrt.py
@@ -11,7 +11,7 @@ import torch_xla
 import torch_xla.core.xla_env_vars as xenv
 import torch_xla.core.xla_model as xm
 import torch_xla.distributed.xla_backend
-from torch_xla._internal import tpu, gpu
+from torch_xla._internal import tpu, gpu, neuron
 from torch_xla import runtime
 
 R = TypeVar('R')
@@ -53,6 +53,9 @@ def _run_thread_per_device(
     Dict of the form {thread_rank: return_value}, where return_value is the
     result of calling `fn`.
   """
+  if runtime.device_type() == "NEURON":
+    neuron.intialize_env(local_rank)
+
   initializer_fn(local_rank, local_world_size)
 
   devices = xm.get_xla_supported_devices()
@@ -137,6 +140,8 @@ def run_multiprocess(fn: Callable[..., R],
   elif runtime.device_type() == 'GPU':
     num_processes = gpu.num_local_processes()
     gpu.initialize_distributed_runtime(num_processes)
+  elif runtime.device_type() == 'NEURON':
+    num_processes = neuron.num_local_devices()
   else:
     num_processes = 1
 
--- a/torch_xla/_internal/pjrt.py
+++ b/torch_xla/_internal/pjrt.py
@@ -53,9 +53,6 @@ def _run_thread_per_device(
     Dict of the form {thread_rank: return_value}, where return_value is the
     result of calling `fn`.
   """
-  if runtime.device_type() == 'NEURON':
-    neuron.intialize_env(local_rank)
-
   initializer_fn(local_rank, local_world_size)
 
   devices = xm.get_xla_supported_devices()
@@ -113,6 +110,9 @@ def _initialize_multiprocess(local_rank: int, local_world_size: int):
 
   if runtime.device_type() == 'TPU':
     tpu.configure_topology(local_rank, local_world_size)
+  
+  if runtime.device_type() == 'NEURON':
+    neuron.initialize_env(local_rank)
 
 
 @runtime.requires_pjrt
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -71,7 +71,7 @@ def is_xla_tensor(tensor):
 
 
 def parse_xla_device(device):
-  m = re.match(r'(CPU|TPU|GPU|XPU):(\d+)$', device)
+  m = re.match(r'(CPU|TPU|GPU|XPU|NEURON):(\d+)$', device)
   if m:
     return (m.group(1), int(m.group(2)))
 
@@ -80,8 +80,8 @@ def get_xla_supported_devices(devkind=None, max_devices=None):
   """Returns a list of supported devices of a given kind.
 
   Args:
-    devkind (string..., optional): If specified, one of `TPU`, `GPU`, `XPU` or `CPU`
-      (the 'GPU' XLA device is currently not implemented).
+    devkind (string..., optional): If specified, one of `TPU`, `GPU`, `XPU`, 
+      `NEURON` or `CPU` (the 'GPU' XLA device is currently not implemented).
     max_devices (int, optional): The maximum number of devices to be returned of
       that kind.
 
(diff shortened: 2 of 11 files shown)


torch_xla/core/xla_model.py
```diff
@@ -216,8 +217,8 @@ def xla_device_hw(device):
       real device.
 
   Returns:
-    A string representation of the hardware type (`CPU`, `TPU`, `XPU`, `GPU`) of the
-    given device.
+    A string representation of the hardware type (`CPU`, `TPU`, `XPU`, `NEURON`, `GPU`)
```
will-cromar, line 220: Yikes, we really have to get rid of these device type lists.

torch_xla/distributed/xla_backend.py
```diff
@@ -35,6 +35,10 @@ def __init__(self, prefix_store, rank, size, timeout):
     self.prefix_store = prefix_store  # reserved for future use.
     self.timeout = timeout
     self._mesh = []
+    if os.getenv('PJRT_DEVICE',
+                 None) == 'NEURON' and os.getenv('XLA_USE_SPMD') != '1':
+      dev = xm.xla_device()
+      xm.set_replication(dev, [dev])
```
will-cromar, line 41: Do you normally call `dist.init_process_group` without `xmp.spawn`? That should already have set the replication. If it's not handling the SPMD case correctly, then I can follow up with a fix.

torch_xla/experimental/pjrt_backend.py
```diff
@@ -19,8 +20,17 @@ def _pjrt_rendezvous_handler(url: str,
   if not master_ip:
     master_ip = tpu.discover_master_worker_ip() if xr.device_type(
     ) == 'TPU' else 'localhost'
-
-  master_port = xu.getenv_as('MASTER_PORT', int, 12355)
+  # Workaround for Neuron to prevent socket in use error with torchrun
+  # MASTER_PORT is set by torchrun and will clash with the port here
```
will-cromar, line 24: I'm not sure I understand the bug here. Which two processes are trying to use this port? Does `torchrun` itself listen on `MASTER_PORT` or start a process that does? This implementation is nearly identical to the built-in `env://` `init_method`, except for some magic around discovering the world size and rank from the runtime. We also handle the multithreaded edge case for TPU v2 and v3, which AFAIK is not relevant to Neuron

torch_xla/experimental/pjrt_backend.py
```diff
@@ -19,8 +20,17 @@ def _pjrt_rendezvous_handler(url: str,
...
+  if dist.is_torchelastic_launched() and xr.device_type() == 'NEURON':
+    master_port = xu.getenv_as('NEURON_PJRT_MASTER_PORT', int, 12355)
+    # In the case of torchrun pjrt.world_size() is not yet initialized
```
will-cromar, line 27: Why doesn't `xr.world_size` work here? The runtime will always be initialized before this function returns, at least by `xr.global_ordinal`.

will-cromar (commented):
This is exciting!

torch_xla/distributed/xla_backend.py
aws-kingrj: Yeah we usually call it in torchrun cases as well

torch_xla/experimental/pjrt_backend.py
aws-kingrj: Yeah torchrun has a default MASTER_PORT so in the case here it will try to listen on the same MASTER_PORT and get a socket in use error

torch_xla/experimental/pjrt_backend.py
aws-kingrj: I'm not fully sure, but any time I printed the world size (on PT 2.0) it would always be 1

torch_xla/distributed/xla_backend.py
aws-kingrj: Previously we had an xrt context function that would call the setting replication in this spot, we removed all of that since there's no xrt but we still needed the replication to be set

torch_xla/experimental/pjrt_backend.py
aws-kingrj: I believe it has to do with the fact that the replication is not set here before we call world_size() since we set the replication in xla_backend.py

torch_xla/experimental/pjrt_backend.py
will-cromar: Got it. I didn't realize `torchrun` itself actually listens on the MASTER_PORT -- I thought it just told the master which port to listen on. This is a bug on our side then.

I found the implementation of the `env://` `init_method`, which [sets `is_master=False` when using torchrun](https://github.com/pytorch/pytorch/blob/a14d99bb6c16233f9c305c089d1bdcdd5b8ea5ac/torch/distributed/rendezvous.py#L166-L174) and wraps `TCPStore` in a `PrefixStore`. I can copy that logic in a follow-up, or you can add it here.

torch_xla/distributed/xla_backend.py
will-cromar: `spawn` will have already called `set_replication` with all available devices in the current thread (usually just the one returned by `xm.xla_device`) by the time you reach this line: https://github.com/pytorch/xla/blob/f06c31f5685fd537e9f220f3fdae8116d0bd0c32/torch_xla/_internal/pjrt.py#L58-L59

torch_xla/experimental/pjrt_backend.py
will-cromar: Actually, this should be `world_size`, because the `torch.distributed` world size should be 1 when doing single-process debugging (either not calling `xmp.spawn` or passing `nprocs=1`).

Getting a world size of 1 sounds suspicious to me. `xmp.spawn` will have already `set_replication` for you before you get to `init_process_group`, so [this case](https://github.com/pytorch/xla/blob/f06c31f5685fd537e9f220f3fdae8116d0bd0c32/torch_xla/runtime.py#L127-L128) will be false. Can you make sure your implementation of `PjRtClient::devices` is returning the correct set of devices, i.e. _all_ devices across all hosts/processes? `global_device_count` returns the [length](https://github.com/pytorch/xla/blob/f06c31f5685fd537e9f220f3fdae8116d0bd0c32/torch_xla/runtime.py#L121-L121) of [this list](https://github.com/pytorch/xla/blob/f06c31f5685fd537e9f220f3fdae8116d0bd0c32/torch_xla/csrc/runtime/pjrt_computation_client.cc#L708-L710)

torch_xla/_internal/pjrt.py
```diff
@@ -53,6 +53,9 @@ def _run_thread_per_device(
     Dict of the form {thread_rank: return_value}, where return_value is the
     result of calling `fn`.
   """
+  if runtime.device_type() == 'NEURON':
+    neuron.intialize_env(local_rank)
```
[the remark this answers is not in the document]
aws-kingrj: Will do, testing if it works internally now

torch_xla/distributed/xla_backend.py
aws-kingrj: yeah this is mean to be for torchrun only - i can add a if torch.distributed.is_torchelastic_launched() here? Also, the only reason i added this only call if not GSPMD is because we don't have it enabled yet for neuron

torch_xla/experimental/pjrt_backend.py
aws-kingrj: This is specifically for torchrun where it's needed

torch_xla/experimental/pjrt_backend.py
aws-kingrj: Yeah could you copy that

will-cromar: Sounds good. I'll follow up with a PR that disables `is_master` and uses torchrun's MASTER_PORT instead.

torch_xla/distributed/xla_backend.py
will-cromar: Got it. I just forgot that `xmp.spawn` doesn't behave correctly with torchrun. This is fine for now. I'm 90% sure you can remove `os.getenv('XLA_USE_SPMD') != '1'` with no issue here, even when you do get to enabling it.

Guarding this behind `dist.is_torchelastic_lauched` makes sense to me.

The logic still doesn't seem quite right to me, since it would make sense to initialize everything before you get to the PJRT rendezvous so that you can trust the `xm.world_size` there. I'll think through the problem some more and copy you on any changes I make to this logic.

will-cromar (commented):
LGTM. Thanks!

FYI, we're in the early planning for a more plugin-like interface for PJRT C API implementations, so that device vendors don't have to go add their device type in several places in this repository. Ideally users would just install plugin packages like `torch-xla-tpu`, `torch-xla-neuron`, etc as needed. We'll keep you in the loop when we have something more concrete.

will-cromar:
It doesn't look like the GPU test failure is related. Can you fix the linter complaints and we can run the CI again?

torch_xla/distributed/xla_backend.py
will-cromar: #5438 sets up multiprocessing in `init_process_group` when launched with torchrun and uses `torchrun`'s store instead of creating a new one in process 0. You should be able to remove this condition and the one in https://github.com/pytorch/xla/pull/5428#discussion_r1290526436.

aws-kingrj:
Since you added [#5438 ](https://github.com/pytorch/xla/pull/5438/), I just removed the set replication in the xla_backend.py. Since you now have it setting in the initialize_multiprocess in pjrt_rendezvous_handler.py

aws-kingrj:
Where are the linter issues, or now we need to rerun the tests since I have committed again?

will-cromar:
I just approved another CI run.

will-cromar (approved):

PR #5429, opened 2023-08-09
Neuron import hook

aws-kingrj:
- Neuron hook needed for initialization functions from neuron libraries
- Tested internally with AWS Neuron

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -68,10 +68,6 @@ def _aws_ec2_inf_trn_init():
   else:
     xla.init()
 
-# Setup Neuron library for AWS EC2 inf/trn instances.
-# It needs to be called before _XLAC is loaded so that _XLAC can pick up environment variables
-_aws_ec2_inf_trn_init()
-
 
 def _setup_tpu_vm_library_path() -> bool:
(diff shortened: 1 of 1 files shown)


torch_xla/__init__.py
```diff
@@ -60,6 +60,19 @@ def _summarize_fn_tracker():
...
+def _aws_ec2_inf_trn_init():
+  try:
+    from torch_neuronx import xla
+  except ImportError:
+    return
+  else:
+    xla.init()
+
+# Setup Neuron library for AWS EC2 inf/trn instances.
+# It needs to be called before _XLAC is loaded so that _XLAC can pick up environment variables
```
will-cromar, line 72: What does `_XLAC` execute during loading? We used to to have static initializers that would immediately create the device client, but I thought I removed those

awshaichen: Our real goal is to "automatically" load the Neuron PJRT C-API plugin, currently achieved by setting the `NEURON_LIBRARY_PATH` environment variable prior to `_XLAC` loading. See https://github.com/pytorch/xla/pull/5428 for more context.

We are open to any suggestion/advice on the proper way of PJRT C-API plugin loading.

PR #5431, opened 2023-08-09
[author: jluntamazon] Adding more explicit HLO lowering control by exposing LoweringContext…

aws-kingrj:
… (and utilities) to python for Neuron

- Currently needed for PyTorch Inference with Neuron
- Has been tested in Neuron version of torch-xla for many releases

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2081,11 +2081,9 @@ class TestLoweringContext(test_utils.XlaTestCase):
   def test_api(self):
     device = xm.xla_device()
     example = torch.tensor([1.0, 2.0, 3.0, 4.0], device=device)
+    xm.mark_step()
 
-    def network(x):
-      return x + 2.0
-
-    result = network(example)
+    result = example + 2
 
     ctx = torch_xla._XLAC.lowering.LoweringContext()
     ctx.build([result])
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2076,6 +2076,8 @@ class RegisterXLAKeyTest(test_utils.XlaTestCase):
     self.assertEqual(met.counter_value("RegisterXLAFunctions"), 1)
 
 
+# Only fails in CI https://github.com/pytorch/xla/pull/5431
+@unittest.skip
 class TestLoweringContext(test_utils.XlaTestCase):
 
   def test_api(self):
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2080,21 +2080,20 @@ class TestLoweringContext(test_utils.XlaTestCase):
 
   def test_api(self):
     device = xm.xla_device()
-    example = torch.tensor([1.0, 2.0, 3.0, 4.0], device=device)
+    a = torch.rand(10, device=device)
+    b = torch.rand(10, device=device)
     xm.mark_step()
 
-    result = example + 2
+    result = a + b
 
     ctx = torch_xla._XLAC.lowering.LoweringContext()
     ctx.build([result])
     hlo = ctx.hlo()
     hlo_text = ctx.hlo_text()
     self.assertTrue('opcode: "parameter"' in hlo_text)
+    self.assertTrue('opcode: "add"' in hlo_text)
     mapping = ctx.parameter_id_tensor_mapping()
     self.assertEqual(len(mapping), 2)
-    input_parameter_id = ctx.tensor_parameter_id(example)
-    # id 0 is the constant
-    self.assertEqual(input_parameter_id, 1)
 
 
 class TestGeneric(test_utils.XlaTestCase):
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2076,6 +2076,30 @@ class RegisterXLAKeyTest(test_utils.XlaTestCase):
     self.assertEqual(met.counter_value("RegisterXLAFunctions"), 1)
 
 
+class TestLoweringContext(test_utils.XlaTestCase):
+
+  def test_api(self):
+    device = xm.xla_device()
+    example = torch.tensor([1.0, 2.0, 3.0, 4.0], device=device)
+
+    def network(x):
+        return x + 2.0
+
+    result = network(example)
+
+    ctx = torch_xla._XLAC.lowering.LoweringContext()
+    ctx.build([result])
+    hlo = ctx.hlo()
+    hlo_text = ctx.hlo_text()
+    self.assertTrue('opcode: "parameter"' in hlo_text)
+    self.assertTrue('opcode: "broadcast"' in hlo_text)
+    self.assertTrue('opcode: "add"' in hlo_text)
+    mapping = ctx.parameter_id_tensor_mapping()
+    self.assertEqual(len(mapping), 2)
+    input_parameter_id = ctx.tensor_parameter_id(example)
+    # id 0 is the constant
+    self.assertEqual(input_parameter_id, 1)
+
 class TestGeneric(test_utils.XlaTestCase):
 
   def test_zeros_like_patch(self):
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2092,7 +2092,6 @@ class TestLoweringContext(test_utils.XlaTestCase):
     hlo = ctx.hlo()
     hlo_text = ctx.hlo_text()
     self.assertTrue('opcode: "parameter"' in hlo_text)
-    self.assertTrue('opcode: "add"' in hlo_text)
     mapping = ctx.parameter_id_tensor_mapping()
     self.assertEqual(len(mapping), 2)
     input_parameter_id = ctx.tensor_parameter_id(example)
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2083,7 +2083,7 @@ class TestLoweringContext(test_utils.XlaTestCase):
     example = torch.tensor([1.0, 2.0, 3.0, 4.0], device=device)
 
     def network(x):
-        return x + 2.0
+      return x + 2.0
 
     result = network(example)
 
@@ -2099,6 +2099,7 @@ class TestLoweringContext(test_utils.XlaTestCase):
     input_parameter_id = ctx.tensor_parameter_id(example)
     # id 0 is the constant
     self.assertEqual(input_parameter_id, 1)
+
 
 class TestGeneric(test_utils.XlaTestCase):
 
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2092,7 +2092,6 @@ class TestLoweringContext(test_utils.XlaTestCase):
     hlo = ctx.hlo()
     hlo_text = ctx.hlo_text()
     self.assertTrue('opcode: "parameter"' in hlo_text)
-    self.assertTrue('opcode: "broadcast"' in hlo_text)
     self.assertTrue('opcode: "add"' in hlo_text)
     mapping = ctx.parameter_id_tensor_mapping()
     self.assertEqual(len(mapping), 2)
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -742,6 +744,155 @@ void BuildProfilerSubmodule(py::module* m) {
...
+class PyLoweringContext {
+ public:
+  PyLoweringContext() : PyLoweringContext(GetCurrentDevice()) {}
+
+  PyLoweringContext(torch::lazy::BackendDevice device)
+      : lowering_ctx("PyLoweringContext", device) {}
+
+  // Builds a HLO graph given a set of output tensors. It is assumed that the
+  // tensors passed to this function are the result of the forward pass of a
+  // given Module/function.
```
JackCaoG, line 756: this is not necessary true right? builder just takes a list a tensor and treat them as root then do post-order-traversal.  You can pretty much pass any tensor to it.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -742,6 +744,155 @@ void BuildProfilerSubmodule(py::module* m) {
...
+  void Build(std::vector<at::Tensor> tensors) {
+    // Get the backing XLA tensors from the output torch tensor handles
+    std::vector<XLATensorPtr> xtensors = GetXlaTensors(tensors, /*want_all=*/true);
+
+    // Get the lazy IR value from the output XLA tensors
+    std::vector<torch::lazy::Value> ir_values;
+    for (auto& xtensor : xtensors) {
+      torch::lazy::Value value = xtensor->CurrentIrValue();
+      ir_values.push_back(value);
+    }
```
JackCaoG, line 766: what if `CurrentIrValue` return a defualt value? this call here assumes all xtensor will have a IR value which is not necessary true.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -742,6 +744,155 @@ void BuildProfilerSubmodule(py::module* m) {
...
+  }
+
+  // Get the parameter identifier of a given tensor. If the tensor is not a
+  // parameter this will always return -1. This is useful in conjunction with
+  // GetParameterIdTensorMapping to identify which values can be baked into
+  // the graph and which values must remain parameters.
+  int64_t GetTensorParameterId(at::Tensor tensor) {
+    // Convert tensor into the backing lazy node
+    XLATensorPtr xtensor = bridge::GetXlaTensor(tensor);
+    torch::lazy::Value value = xtensor->CurrentIrValue();
```
JackCaoG, line 811: again this function assumes tensor has a IR value.

JackCaoG:
@jluntamazon

torch_xla/csrc/init_python_bindings.cpp
jluntamazon: Yes, this comment is outdated, we can fix this.

torch_xla/csrc/init_python_bindings.cpp
jluntamazon: Under what conditions does an xtensor not have an IR value? 

I am aware that this could cause problems if you had both CPU/XLA allocated tensors, but is there another condition where an IR value would not exist?

I think the correct behavior would be that if you were passed a tensor that did not have an IR value then we should exit with an exception (unless this is otherwise recoverable?)

JackCaoG: XLATensor can be in many format, it can has a `XLAData` which means it is a data on device, it can be a `IR Value` which means it is an computation to be computed. It can also be `at::Tensor` means it is still a tensor on host and we haven't upload the tensor to device yet.

If you want to force the IR value(XLAData and `at::tensor` can become IR, which is a DeviceData IR), you want to use `GetIrValue` instead of `CurrentIRValue`

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -743,6 +745,154 @@ void BuildProfilerSubmodule(py::module* m) {
...
+   *         return x + 2.0
+   *
+   *     result = network(example)
+   *
+   *     ctx = torch_xla._XLAC.lowering.LoweringContext()
+   *     ctx.build([result])
+   *     hlo = ctx.hlo()
+   *     hlo_text = ctx.hlo_text()
+   *     mapping = ctx.parameter_id_tensor_mapping()
+   *     input_parameter_id = ctx.tensor_parameter_id(example)
```
JackCaoG, line 877: Can you add this part as test to somewhere near https://github.com/pytorch/xla/blob/master/test/test_operations.py#L2078 as a new test? I am worried about adding new APIs without test. For new apis we introduced we tried to make sure they actually have tests.

seanlatias: Sure. I have added the test. Please take a look.

JackCaoG (approved):

JackCaoG:
we can merge it after CI is green. @wonjoolee95 can you cherry0pick this into 2.2 once merged?

test/test_operations.py
```diff
@@ -2076,6 +2076,29 @@ def test_multi_init_xla_backend(self):
...
+    example = torch.tensor([1.0, 2.0, 3.0, 4.0], device=device)
+
+    def network(x):
+      return x + 2.0
+
+    result = network(example)
+
+    ctx = torch_xla._XLAC.lowering.LoweringContext()
+    ctx.build([result])
+    hlo = ctx.hlo()
```
vanbasten23, line 2092: is this line needed?

seanlatias: It's just there to test whether the API would fail or not. The returned results is a binary and thus cannot be checked.

vanbasten23: I see.  is it to test `ctx.hlo()` doesn't throw an error

seanlatias:
Hmmm... The test results in CI are different from what I observe locally. Need to figure out why.

seanlatias:
@JackCaoG @wonjoolee95 is there a way I can directly test in the CI environment?

JackCaoG:
can you try to run the whole test('test_operations.py') instead of single test? If I have to guess it is other test running before this(which is in the same process) affect the result.

seanlatias:
Yeah, I tried that locally but it still passed.

JackCaoG:
There are logs in https://github.com/pytorch/xla/actions/runs/6174499874/job/16761837847?pr=5431 under `Download and Run docker xxx` show you how to download the docker of the CI and we have logs that how test is being run. You can pretty much replicate that process on your local machine.

seanlatias:
Cool, thanks.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -761,7 +761,7 @@ class PyLoweringContext {
     // Get the lazy IR value from the output XLA tensors
     std::vector<torch::lazy::Value> ir_values;
     for (auto& xtensor : xtensors) {
-      torch::lazy::Value value = xtensor->CurrentIrValue();
+      torch::lazy::Value value = xtensor->GetIrValue();
```
JackCaoG, line 764: oh haha, I should catch this when reviewing

seanlatias:
I don't have the access to the docker image. Now the error messages are all because of `Check failed: HasValue() `. Do you have ideas what is usually the cuase for this?

JackCaoG:
The failure is 
```
buffer with shape f32[10] on device CPU:0 is deleted
```
and log is
```
	torch_xla::runtime::PjRtComputationClient::PjRtData::GetOpaqueHandle()
	torch_xla::LoweringContext::GetParameter(std::shared_ptr<torch::lazy::BackendData> const&)
	torch_xla::DeviceData::Lower(torch_xla::LoweringContext*) const
```

it is suggesting that it is trying to get the parameter but the parameter buffer is already deleted. This is a bit weird, because the test code is
```
    a = torch.rand(10, device=device)
    b = torch.rand(10, device=device)
    xm.mark_step()

    result = a + b
```

The only thing I can guess is aliasing somehow messed up. can you try run with `XLA_ENABLE_PARAM_ALIASING=0`?

seanlatias:
Jsut to double check, you mean setting the envvar when running this unit test?

JackCaoG:
yea, were you able to repo locally?

seanlatias:
No, no luck. I can still run the test sucessfully locally with and without the envvar.

JackCaoG:
```
+ echo 'Running in DynamicShape mode: /tmp/pytorch/xla/test/test_operations.py' --verbosity=2
Running in DynamicShape mode: /tmp/pytorch/xla/test/test_operations.py --verbosity=2
+ XLA_EXPERIMENTAL=nonzero:masked_select:masked_scatter
+ run_test /tmp/pytorch/xla/test/test_operations.py --verbosity=2
+ echo 'Running in PjRt runtime: /tmp/pytorch/xla/test/test_operations.py' --verbosity=2
Running in PjRt runtime: /tmp/pytorch/xla/test/test_operations.py --verbosity=2
++ command -v nvidia-smi
+ '[' -x '' ']'
+ PJRT_DEVICE=CPU
+ CPU_NUM_DEVICES=1
+ run_coverage /tmp/pytorch/xla/test/test_operations.py --verbosity=2
+ '[' 0 '!=' 0 ']'
+ python3 /tmp/pytorch/xla/test/test_operations.py --verbosity=2
```

seems like it is dynamic shape failing. can you run with 
```
XLA_EXPERIMENTAL=nonzero:masked_select:masked_scatter
PJRT_DEVICE=CPU
CPU_NUM_DEVICES=1
python3 /tmp/pytorch/xla/test/test_operations.py --verbosity=2
```

JackCaoG:
@wonjoolee95 Do you have time to take a look at this one?

wonjoolee95:
Let me try to build this and PyTorch master locally and see if I can reproduce it, it should be quick (like 15 minutes or so). If not, we can just disable the test in the CI for now.

seanlatias:
Thanks @wonjoolee95

wonjoolee95:
I can also see that running both the entire `test_operations.py` and this specific test succeed in my CPU VM:
```
# Running this one test only
(base) wonjoo@wonjoo-cpu:~/pytorch/xla$ python test/test_operations.py TestLoweringContext.test_api
Ran 1 test in 0.127s

OK

# Running entire `test_operations.py`
(base) wonjoo@wonjoo-cpu:~/pytorch/xla$ python test/test_operations.py 
/home/wonjoo/pytorch/xla/test/test_operations.py:906: UserWarning: The .grad attribute of a Tensor that is not a leaf Tensor is being accessed. Its .grad attribute won't be populated during autograd.backward(). If you indeed want the .grad field to be populated for a non-leaf Tensor, use .retain_grad() on the non-leaf Tensor. If you access the non-leaf Tensor by mistake, make sure you access the leaf Tensor instead. See github.com/pytorch/pytorch/pull/30531 for more informations. (Triggered internally at /home/wonjoo/pytorch/build/aten/src/ATen/core/TensorBody.h:489.)
  self.assertIsNone(a.grad)
Ran 158 tests in 37.020s

OK (skipped=2)
```
Since we can both verify that this passes successfully in our dev env and cannot reproduce, let's just skip the test to keep the CI happy. @seanlatias, we can skip this with `@unittest.skip`, and let's leave a comment saying that this failure only happens in CI and leave a link to this PR for context.

seanlatias:
@wonjoolee95 not sure what happens to the GPU test.

wonjoolee95:
Can you rebase with master one more time so it re-triggers the CI?

seanlatias:
It's still the same. It seems the CI cancels itself automatically when it reaches 4 hr. I also see similar results in other PRs.

JackCaoG:
it is a known issue, because forked pr can't use the remote cache. We can just merge.

seanlatias:
Thanks @JackCaoG @wonjoolee95

PR #5433, opened 2023-08-10
[SPMD] Make IR sharding custom sharding op

alanwaketan:
Summary:
This pull request changes the syntax of IR sharding by making it a new node instead of just attaching the sharding spec to the tensor. On the same time, we will still attach a sharding spec to the newly created XLATensor which will hold the new IR node.

This new IR node will be a CustomSharding node and in hlo:
```
%annotate = f32[6,3] custom-call(%copy), custom_call_target="Sharding", sharding={devices=[2,1]0,1}
```

Test Plan:
PJRT_DEVICE=TPU XLA_USE_SPMD=1 python test/spmd/test_xla_sharding.py -v -k test_mark_sharding_ir 
PJRT_DEVICE=TPU XLA_USE_SPMD=1 python test/spmd/test_xla_sharding.py -v -k test_inplace_add_with_sharding


alanwaketan:
Somehow XLAShardedTensor will be propogated through PyTorch ops as well, but then the propagated one doesn't have the global tensor attribute. So basically breaking the whole system. Disable it for now.

alanwaketan:
Somehow the backward is gone. Probably we need to make it an in-place op instead of returning a new tensor.

JackCaoG:
a bunch of CPU test is failing.. looking, I think we break some previous assumption of the `mark_sharding` is a inplace operation to the IR.

torch_xla/experimental/xla_sharding.py
```diff
@@ -444,14 +446,9 @@ def mark_sharding(
...
-  if isinstance(t, XLAShardedTensor):
-    torch_xla._XLAC._xla_mark_sharding(t.global_tensor, tile_assignment,
-                                       group_assignment, replication_groups,
-                                       int(sharding_type))
-    return t
-  torch_xla._XLAC._xla_mark_sharding(t, tile_assignment, group_assignment,
+  t = torch_xla._XLAC._xla_mark_sharding(t, tile_assignment, group_assignment,
                                      replication_groups, int(sharding_type))
-  return XLAShardedTensor(t)
+  return t
```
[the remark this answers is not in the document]
JackCaoG: I think this is why unit test is failing.

.vscode/settings.json
```diff
@@ -18,5 +18,82 @@
         "./bazel-out/_coverage/_coverage_report.dat"
     ],
     "python.formatting.provider": "yapf",
-    "editor.formatOnSave": true
-}
+    "editor.formatOnSave": true,
+    "files.associations": {
```
JackCaoG, line 22: lol let's clean this part later. If we want to change it it should be a separate pr.

test/spmd/test_xla_sharding.py
```diff
@@ -695,6 +691,7 @@ def test_sharded_tensor_aliasing(self):
     xm.mark_step()
     self.assertEqual(met.metric_data("InputOutputAliasCount")[0], 1)
 
+  @unittest.skip("Irrelevant, maybe the whole PR can be rollback.")
```
JackCaoG, line 694: I think I know what's the issue, `setShardingSpec` has a bug where it assign the sharding spec to the whole IR, instead the one output of the IR.

JackCaoG: not sure why but this test actually passed on my v4.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2400,7 +2399,8 @@ void XLANativeFunctions::_propagate_xla_data(const at::Tensor& input,
 
   // 2) Aid SPMD.
   if (input_tensor->sharding_spec()) {
-    output_tensor->SetShardingSpec(*(input_tensor->sharding_spec()));
+    tensor_methods::custom_sharding(output_tensor,
```
JackCaoG, line 2402: nit, this method should be called `custom_sharding_` if it is an inplace op.

alanwaketan: Yea, will fix that. Originally that was not the plan haha.

alanwaketan: Fixed.

JackCaoG:
@alanwaketan let me know when you want me to review this pr

alanwaketan:
> @alanwaketan let me know when you want me to review this pr

I think it's ready for review now.

alanwaketan:
Lol, if I'm not targeting the master, it won't run any tests. Any suggestions on how can I fix this?

JackCaoG:
oh.... haha I have no ideas. You can at least run the TPU CI.

alanwaketan:
The CIs are all green.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1398,7 +1398,7 @@ void InitXlaModuleBindings(py::module m) {
     if (xtensor->CurrentIrValue()) {
       device_data_node = DeviceData::Cast(xtensor->CurrentIrValue().node.get());
       if (!device_data_node) {
-        xtensor->SetShardingSpec(*new_sharding_spec);
+        tensor_methods::custom_sharding_(xtensor, new_sharding_spec);
```
JackCaoG, line 1401: I vaguely remember that we discussed about  for IR we want to do custom sharding and for input/output we want to directly injected the sharding spec to the HLO. Is that still the case?

alanwaketan: Yea, output means the ROOT node which should not be affected by mark_sharding.

JackCaoG: hmm then how does output get the sharding? Through in-place update and sharding spec being inherited to the output tensor? I remember there are cases where we want to manually mark sharding on output.

alanwaketan: There is a logic to extract the sharding spec from the output and then assign it to the output tensors: https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_sharding_util.cpp#L629

This logic works no matter whether the ROOT node has sharding spec or not, i.e., whether or not the output sharding propagation is enabled.

alanwaketan:
Can I get a review on this pull request?

JackCaoG:
Let me take a look, busy debugging something else lol

torch_xla/csrc/init_python_bindings.cpp
JackCaoG: so the idea here is that only `mark_sharding` and functionization should use `custom_sharding`? I saw there are a couple other places like in `clone` still use `setShardingSpec`.

For data through data loader I think it is attach sharding to HLO directly
For model weights, we will use `mark_sharding` so `custom_sharding`
For output it depends on whether `mark_sharding` is being called on the output explictly
For Clone it will be directly attach sharding to the IR cloned.

Is this the right understanding?

alanwaketan: The custom_sharding op should only be applied to IR node which means not inputs or outputs.

Data from data loader and weights are inputs so they will use setShardingSpec directly.
Output with output sharding propagation, we don't mark anything. Output means ROOT node. Without propagation, we will do setShardingSpec directly.

For Clone, I have no ideas... Can you point me to the code?

JackCaoG: [I am glad we get to meet a couple times during TGIT, that was fun. ](https://github.com/pytorch/xla/blob/4f378e694a68192fd634a070a3485c73afab1c71/torch_xla/csrc/tensor_methods.cpp#L906-L912)

alanwaketan: Got it, I think it should be custom-call. In fact, all SetShardingSpec except inputs and outputs should be replaced by custom-calls. But I want to do it as a follow up after we were able to reproduce the LLaMA2 results on the RC wheels.

alanwaketan: File #5495.

JackCaoG (approved):

alanwaketan:
Thanks, Jack!

torch_xla/csrc/init_python_bindings.cpp
yeounoh: For reference, clone, copy -- and even functionalization related path should be annotated with custom call.

yeounoh (commented):
LGTM (after-the-fact), thank you @alanwaketan !

PR #5435, opened 2023-08-10
Remove TPU loading hack

will-cromar:
A transitive dependency XRT added a static initializer that opened the TPU. Without XRT, we don't have to worry about it.

This commit should _not_ be cherry-picked to the XRT branch.


will-cromar:
See #3616 for original context.

JackCaoG (approved):

PR #5436, opened 2023-08-11
[WIP] Patch nn.Linear

alanwaketan:
Summary:
This pull request patch nn.Linear to avoid PyTorch's default flattening logic of the input tensor.

Test Plan:
PJRT_DEVICE=TPU TPU_LIBRARY_PATH=../../libtpu.so python test/test_operations.py -v -k test_patched_linear


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1595,6 +1605,13 @@ void InitXlaModuleBindings(py::module m) {
     MapXlaEnvVarsToLazy();
     InitXlaBackend();
   });
+  m.def("_xla_linear",
+        [](const at::Tensor& input, const at::Tensor& weight, const c10::optional<at::Tensor>& bias) -> at::Tensor {
```
jonb377, line 1609: `bias` isn't used, but it isn't an issue for our current tests

alanwaketan:
Let's close this as replacing matmul in the nn.Linear with einsum is much more appropriate than this approach.

PR #5437, opened 2023-08-11
Add warning to Linear patching when Linear subclasses are used

carmocca:
The current Linear patching code assumes that the Linear layer will not be subclassed.

LoRA layers are generally implemented by subclassing Linear: https://github.com/microsoft/LoRA/blob/main/loralib/layers.py#L90

This means that the patching logic is completely and silently skipped for them. XLA should try to support them, in the meantime, this PR adds a warning to give some visibility into the issue for users who could experience issues related to this.

cc @gkroiz

### DIFF
--- a/torch_xla/distributed/fsdp/utils.py
+++ b/torch_xla/distributed/fsdp/utils.py
@@ -144,7 +144,7 @@ def apply_xla_patch_to_nn_linear(module):
       return
     if getattr(forward_method, "__func__", None) != torch.nn.Linear.forward:
       m_cls = m.__class__
-      if issubclass(m_cls, torch.nn.Linear):
+      if m_cls is not torch.nn.Linear and issubclass(m_cls, torch.nn.Linear):
         logging.warning(
           f"`{m_cls.__module__}.{m_cls.__name__}` is a subclass of `torch.nn.Linear`. PyTorch XLA needs to monkeypatch"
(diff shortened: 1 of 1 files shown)


gkroiz (commented):
~~Should this warning be generalized for all of Lightning when someone uses xla? To my understanding all `Linear()` Layers will have to go through this patch.~~

torch_xla/distributed/fsdp/utils.py
```diff
@@ -142,6 +143,13 @@ def _try_patching_forward_method(m, forward_method_name="forward"):
     if forward_method is None:
       return
     if getattr(forward_method, "__func__", None) != torch.nn.Linear.forward:
+      m_cls = m.__class__
+      if m_cls is not torch.nn.Linear and issubclass(m_cls, torch.nn.Linear):
+        logging.warning(
```
gkroiz, line 148: Should this warning be generalized for all of Lightning when someone uses xla? To my understanding all Linear() Layers will have to go through this patch.

gkroiz, line 148: Should this warning be generalized for all of Lightning when someone uses xla? To my understanding all Linear() Layers will have to go through this patch.

gkroiz:
cc @JackCaoG

JackCaoG (approved):
lgtm, @Liyang90 do you have any concern?

JackCaoG:
@carmocca can you fix the linter?

PR #5439, opened 2023-08-11
Bugfix + add more test for llama

qihqi:
Bugfix details:
1. When the graph have mutations the exported graph will have additional inputs. For now we are dropping them.
2. We should trace with args instead of final_args.

### DIFF
--- a/test/stablehlo/llama_model2.py
+++ b/test/stablehlo/llama_model2.py
@@ -1,28 +1,14 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
-# This software may be used and distributed according to the terms of the Llama 2 Community License Agreement.
+# this version contains modification to make it easier to trace
 
 import math
 from dataclasses import dataclass
-from typing import Any, Optional, Tuple
+from typing import Any, Optional, Tuple, List
 
 import torch
 import torch.nn.functional as F
 from torch import nn
 
-
-@dataclass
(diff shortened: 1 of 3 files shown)


test/stablehlo/llama_model.py
```diff
@@ -259,11 +259,11 @@ def __init__(self, params: ModelArgs):
   def forward(self, tokens: torch.Tensor, start_pos: int):
     _bsz, seqlen = tokens.shape
     h = self.tok_embeddings(tokens)
-    # freqs_cis = self.freqs_cis[start_pos : start_pos + seqlen]
+    freqs_cis = self.freqs_cis[start_pos:start_pos + seqlen]
     # freqs_cis = self.freqs_cis[:seqlen] #(self.freqs_cis, 0, 0, seqlen)
-    freqs_cis = torch.index_select(
```
lsy323, line 264: nit: In the XLA execution workflow, we need to use `index_select` to avoid a cpu fallback. Otherwise the indexing operation will fallback to cpu. Call this out just in case we want the model code to be executed.

lsy323 (approved):
LGTM, minor comment on llama1 change.

test/stablehlo/llama_model.py
qihqi: yeah the llama_model2 version have the one using index_select

PR #5441, opened 2023-08-12
Update fix_includes.sh

aazzolini:
Without this patch I cannot get torch_xla to build outside of the docker. This should fix it.


alanwaketan (approved):
LGTM. Let's wait until all CI are green.

JackCaoG (commented):
not sure what these are @stgpetrovic any idea :D?

alanwaketan:
> not sure what these are @stgpetrovic any idea :D?

I believe torchgen will hardcore some of the includes in code-gen files. This is a way to hack that infra...

JackCaoG:
I think gpu build just timeput

PR #5445, opened 2023-08-14
Update gcc to 10.

qihqi:
Also use unversioned clang-format to be more flexible in updating distro version.

### DIFF
--- a/infra/ansible/config/apt.yaml
+++ b/infra/ansible/config/apt.yaml
@@ -13,7 +13,7 @@ apt:
       - rename
       - vim
       - wget
-      - clang-format-7
+      - clang-format
       - lcov
       - less
 
--- a/infra/ansible/config/apt.yaml
+++ b/infra/ansible/config/apt.yaml
@@ -4,6 +4,8 @@ apt:
     build_common:
       - ccache
       - curl
+      - gcc-10
+      - g++-10
       - git
       - gnupg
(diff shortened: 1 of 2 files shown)


infra/ansible/e2e_tests.Dockerfile
```diff
@@ -1,8 +1,12 @@
 ARG python_version=3.8
-ARG debian_version=buster
+ARG debian_version=bullseye
 
 FROM python:${python_version}-${debian_version} AS build
 
+# install gcc10
+RUN apt-get update
+RUN apt-get install -y gcc-10
```
will-cromar, line 8: Should gcc-10 be installed by ansible instead? This image installs all of the packages from the `build` stage.

infra/ansible/config/apt.yaml
```diff
@@ -13,7 +13,7 @@ apt:
       - rename
       - vim
       - wget
-      - clang-format-7
+      - clang-format
       - lcov
```
JackCaoG, line 17: can we add gcc-10 here directly?

will-cromar: gcc-10 is already included in the build stage for `aarch64`. Maybe just move it to `build_common`?

qihqi: done.

JackCaoG (approved):

PR #5458, opened 2023-08-16
Update insutrction in case gcc version on the build env is too old

JackCaoG:


CONTRIBUTING.md
```diff
@@ -19,6 +19,16 @@ We recommend you to use our prebuilt Docker image to start your development work
   ```
   All of the code below will be assumed to be run within the docker.
 
+* Update gcc version if it is < gcc 10 (only necessary with an old docker image)
+
+  ```shell
+  gcc --version
+  # if it is < 10
+  echo 'deb http://deb.debian.org/debian bullseye main' >> /etc/apt/sources.list
```
will-cromar, line 27: I'm not sure adding this to the docs is a good idea. There's a good chance IMO that doing this breaks the container anyway, since you're just adding the next releases repository for `main` and not updating anything else. Also not sure if this even overrides this line that's already in there: `deb http://deb.debian.org/debian buster main`

I think it's enough to say we need gcc-10, and if you're on an old version of the image, pull a new one or attempt an in-place upgrade according to Debian's docs: https://wiki.debian.org/DebianUpgrade

PR #5462, opened 2023-08-17
wrap dynamo inference execution with no grad

JackCaoG:
For more detail please take a look at https://github.com/pytorch/xla/issues/5430#issuecomment-1681367696, we want to run inference workload using `no_grad` to prevent aot_autograd to perform some training specified operation.


JackCaoG:
actually nvm, https://github.com/pytorch/xla/issues/5430#issuecomment-1683060034 this is still slower with aot backend.

JackCaoG:
I rootcuased that issue, I am hesitating landing this one. I felt like it is safer to use `openxla_eval` as the inference backend until we fix that `as_strided` issue..

PR #5467, opened 2023-08-18
Fix data type in Pow with Scalar base and Tensor exponent

### LINKED ISSUE #5118: `torch.pow` gives incorrect results compared to pure python and `torch.float_power`
## 🐛 Bug

When doing the following computation: `0.995**100` we see that the following operators produce these results:
```
torch.pow - tensor(0, device='xla:1')
toch.float_power - tensor(0.6058, device='xla:1')
** - 0.6057704364907279
```

The following HLOs are produced for each the first two operators
```
HloModule SyncTensorsGraph.5, entry_computation_layout={(s64[],s64[])->(s64[])}

ENTRY %SyncTensorsGraph.5 (p0: s64[], p1: s64[]) -> (s64[]) {
  %p1 = s64[] parameter(1), frontend_attributes={neff_input_name="input1"}
  %p0 = s64[] parameter(0), frontend_attributes={neff_input_name="input0"}
  %power = s64[] power(s64[] %p1, s64[] %p0)
  ROOT %tuple = (s64[]) tuple(s64[] %power), frontend_attributes={neff_output_names="output0"}
}
```

```
HloModule SyncTensorsGraph.6, entry_computation_layout={(s64[],f32[])->(f32[])}

ENTRY %SyncTensorsGraph.6 (p0: s64[], p1: f32[]) -> (f32[]) {
  %p1 = f32[] parameter(1), frontend_attributes={neff_input_name="input1"}
  %p0 = s64[] parameter(0), frontend_attributes={neff_input_name="input0"}
  %convert = f32[] convert(s64[] %p0)
  %power = f32[] power(f32[] %p1, f32[] %convert)
  ROOT %tuple = (f32[]) tuple(f32[] %power), frontend_attributes={neff_output_names="output0"}
}
```

## To Reproduce

Execute the following script:
```
import torch

beta = 0.995
def func1(step):
    step = torch.tensor(step, device=xm.xla_device())
    factor = torch.pow(beta, step)
    print(factor)

def func2(step):
    step = torch.tensor(step, device=xm.xla_device())
    factor = torch.float_power(beta, step)
    factor = factor.float()
    print(factor)

def func3(step):
    factor = (beta ** step)
    print(factor)

func1(100)
func2(100)
func3(100)
```

## Expected behavior

`torch.pow` should produce the same results as CPU python and `torch.float_power`

## Environment

 - Reproducible on XLA backend [CPU/TPU]: Neuron
 - torch_xla version: `1.13.1+torchneuron6`

lsy323:
This fixes #5118. When pow base is a floating point scalar, exponent is an int tensor, the result is wrong.

The root cause is the following
- when the scalar is converted to Lazy IR from `at::Scalar`, it's using the shape info of exponent tensor. This makes the scalar to have integer dtype when exponent is an integer tensor.
- Another problem is that the result tensor is created by `exponent->CreateFrom`, which uses exponent primitive type by default, leading to the result tensor has the same dtype as exponent. However, we should use the promoted dtype for the result tensor.

The following changes are made to resolve the above 2 problems:

- Use Scalar dtype when Lazy IR Node is created.
- Use promoted dtype for the result tensor

Issue with the fix:

I found when a Python constant is captured into `at::Scalar`, fp32/64 are not distinguished any more ([code ref](https://github.com/pytorch/pytorch/blob/4ee622476702fba3480572cbc712b552fd878605/c10/core/Scalar.h#L258C16-L270)), the dtype check util function in PyTorch/XLA [here] is doing the same(https://github.com/pytorch/xla/blob/028df4da388468fa9a41b1f98ea08bfce13b4c63/torch_xla/csrc/torch_util.cpp#L49). This causes the result to be in either double dtype or integer dtype. However the default floating point dtype in Python is float32, and the PyTorch eager gives float32 result. This discrepancy needs to be investigated. I have to cast the xla tensor back to fp32 in the cpp tests now

To generate the same dtype as PyTorch Eager, we need to distinguish float32/64 when the Python constant is captured into an `aten::Scalar`.

### DIFF
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -3679,7 +3679,7 @@ TEST_F(AtenXlaTensorTest, TestPowScalarTensor) {
   torch::Tensor result = torch::pow(base, exponent);
   ForEachDevice([&](const torch::Device& device) {
     torch::Tensor xla_exponent = CopyToDevice(exponent, device);
-    torch::Tensor xla_result = torch::pow(base, xla_exponent);
+    torch::Tensor xla_result = torch::pow(base, xla_exponent).to(torch::kFloat);
     AllClose(result, xla_result, /*rtol=*/1e-3, /*atol=*/1e-5);
   });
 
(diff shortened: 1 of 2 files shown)


wonjoolee95 (approved):
LGTM. Pending CI to verify rest of the unit tests.

JackCaoG (approved):

PR #5468, opened 2023-08-18
[original author: mrnikwaws] fixing incorrect stride information on xla tensors

aws-kingrj:
- fixing incorrect xla tensor stride information
- breaks log softmax lowering
- tested on AWS Neuron internal testing
- original author mrnikwaws

### DIFF
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -3679,7 +3679,7 @@ TEST_F(AtenXlaTensorTest, TestPowScalarTensor) {
   torch::Tensor result = torch::pow(base, exponent);
   ForEachDevice([&](const torch::Device& device) {
     torch::Tensor xla_exponent = CopyToDevice(exponent, device);
-    torch::Tensor xla_result = torch::pow(base, xla_exponent);
+    torch::Tensor xla_result = torch::pow(base, xla_exponent).to(torch::kFloat);
     AllClose(result, xla_result, /*rtol=*/1e-3, /*atol=*/1e-5);
   });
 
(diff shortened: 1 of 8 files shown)


JackCaoG:
Ok I think I know what's going on. The pr from the fork can not use remote cache so build just timeout. Let me see if I can bump the timeout.

JackCaoG:
pr in https://github.com/pytorch/xla/pull/5470. Let me also grant you write access so you can create branch on pytorch/xla directly. This way all of the CI will use the bazel remote cache and runs much faster.

JackCaoG:
@aws-kingrj if you rebase the build should start run pass timeout.

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -161,6 +161,11 @@ at::IntArrayRef XLATensorImpl::strides_custom() const {
   return strides_default();
 }
```
JackCaoG, line 163: You mentioned this `breaks log softmax lowering`. Can you add a test that will fail without this change? The change in general make sense but without a test it is hard to make sure this is correct nor can we prevent this from regressing again.

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -3700,6 +3700,19 @@ TEST_F(AtenXlaTensorTest, TestPowIntExponent) {
   ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestPowFloatScalarBaseIntExponent) {
```
JackCaoG, line 3703: I think the commit is a bit messed up, this seems to be from other pr

PR #5469, opened 2023-08-18
[Dynamo][SPMD] Fix the input sharding for dynamo

JackCaoG:
without this change the test failed with
```
but got buffer with incompatible size 1024 (s64[]{:T(128)})
```

the real issue is that after `mark_step` there is a dangling `DeviceData` IR which blocks the correct `DeviceData` IR being created using the real `PjrtData` object. This will creates all sorts of issue including the `TensorID` in the `DeviceData`'s info being missing.

I still need to figure out why the `DeviceData` Ir still exist in the middle of the `mark_step`, it should get deleted during the `CollectSyncTensor` step(given this is considered as the output of the current execution).

### DIFF
--- a/torch_xla/csrc/xla_sharding_util.cpp
+++ b/torch_xla/csrc/xla_sharding_util.cpp
@@ -692,11 +692,9 @@ void ShardingUtil::PrepareOutputShardingPropagation(
     // Register the sharded data placeholder to the tensor and its node.
     (*data_placeholders)[i] = sharded_data_placeholder;
     xtensor->data()->handle = (*data_placeholders)[i];
-    if (auto ir_value = xtensor->CurrentIrValue()) {
-      if (DeviceData* device_data_node =
-              DeviceData::Cast(ir_value.node.get())) {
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -692,11 +692,9 @@ void ShardingUtil::PrepareOutputShardingPropagation(
     // Register the sharded data placeholder to the tensor and its node.
     (*data_placeholders)[i] = sharded_data_placeholder;
     xtensor->data()->handle = (*data_placeholders)[i];
```
yeounoh, line 694: So we were setting the device data and the device data IR's handle, instead of resetting the IR.

yeounoh (approved):
LGTM, thanks

torch_xla/csrc/xla_sharding_util.cpp
JackCaoG: yea, we should just delete that device data. So it can be created again correctly later when used.

PR #5472, opened 2023-08-21
Add topological sorting to dynamo partitions

wonjoolee95:
As a follow-up to https://github.com/pytorch/xla/pull/5309, this PR sets `allows_single_node_partition=True` and calls `torch.fx.passes.utils.fuser_utils.topo_sort` to ensure the topological ordering of the partitioned nodes. 

With the topological ordering added, the test `test_simple_model_with_in_place_ops` succeeds consistently (previously, we were seeing occasional failure due to the unsorted partitioned nodes when `allows_single_node_partition=True`). 

Test plan:
- `python test/dynamo/test_dynamo.py` all passing locally
- CI

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -101,8 +101,17 @@ class DynamoInferenceBasicTest(unittest.TestCase):
         super().__init__()
         self.self_tensor = torch.zeros((5, 3), device=device)
 
-      def forward(self, index, copy_tensor, input_tensor):
+      def copy_(self, index, copy_tensor):
         self.self_tensor.index_copy_(0, index, copy_tensor)
+
+      def add_(self, index, other_tensor):
+        self.self_tensor.add_(other_tensor)
+
+      def abs_(self, index, other_tensor):
+        self.self_tensor.abs_()
+
+      def forward(self, index, copy_tensor, input_tensor, op_name):
+        getattr(self, op_name)(index, copy_tensor)
         output = input_tensor + self.self_tensor
         return output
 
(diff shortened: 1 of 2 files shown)


test/dynamo/test_dynamo.py
```diff
@@ -245,22 +245,22 @@ def fn_fallback(t):
     cpu_res = fn_fallback(t)
     xla_dynamo_res = dynamo_fn(t_xla)
     self.assertTrue(torch.allclose(cpu_res, xla_dynamo_res.cpu()))
-    self.assertEqual(met.metric_data('CompileTime')[0], 4)
-    self.assertEqual(met.metric_data('ExecuteTime')[0], 8)
+    self.assertEqual(met.metric_data('CompileTime')[0], 3)
```
wonjoolee95, line 248: The `CompileTime` and `ExecuteTime` only differs for this `test_fallback_multiple_submodules` unit test. With `allows_single_node_partition=True`, the partition is different and sees slightly different `CompileTime` and `ExecuteTime` metrics.

test/dynamo/test_dynamo.py
```diff
@@ -101,33 +101,43 @@ def __init__(self, device=None):
...
-    self.assertTrue(torch.allclose(res_cpu, res_xla_dynamo.cpu()))
+    input_tensor = torch.ones(3)
+    copy_tensor = torch.rand(5, 3)
+    index = torch.tensor([0, 4, 2, 1, 3])
+    xla_input_tensor = input_tensor.to(device)
+    xla_copy_tensor = copy_tensor.to(device)
+    xla_index = index.to(device)
+
+    in_place_ops = ['copy_', 'add_', 'abs_']
+    for in_place_op in in_place_ops:
```
wonjoolee95, line 135: Updated the test to call multiple in-place ops in iteration, which should trigger separate `extract_compiled_graph` calls.

JackCaoG:
Thanks, I will take a look today.

PR #5473, opened 2023-08-21
[NOT MERGE] Test on adding dynamic shape and tags

qihqi:


torch_xla/csrc/lowering_context.cpp
```diff
@@ -94,13 +98,21 @@ LoweringContext::LoweringContext(
...
-    const std::shared_ptr<torch::lazy::BackendData>& data) {
+    const std::shared_ptr<torch::lazy::BackendData>& data, 
+    const std::vector<uint32_t>& dynamic_dims
+    ) {
   torch::lazy::BackendData::Handle handle = data->GetHandle();
   auto it = parameters_map_.find(handle);
   if (it == parameters_map_.end()) {
+    xla::Shape shape = UnwrapXlaData(data)->shape();
+    for (const int dim : dynamic_dims) {
+      shape.set_dynamic_dimension(dim, true);
```
sdasgup3, line 109: I think we should also add
```
// import xla::Shape.h to inlcude the following defintion.
static constexpr int64_t kUnboundedSize = std::numeric_limits<int64_t>::min();

shape.set_dimensions(dim, kUnboundedSize);
```

PR #5474, opened 2023-08-21
Introduce `torch_xla.runtime.use_spmd()`

yeounoh:
This is the first step to deprecate `XLA_USE_SPMD` flag, to enable SPMD. While the flag is being deprecated, we will allow both usages, where user can either set `XLA_USE_SPMD=1` flag or call `torch_xla.runtime.use_spmd()` in their script/program.

The recommended way is to use `use_spmd()` API, as follows:
```python
xr.use_spmd()
assert xr.is_spmd() == True
```
and it raises a warning, if user set `XLA_USE_SPMD` before the invocation.

### DIFF
--- a/test/spmd/test_dynamo_spmd.py
+++ b/test/spmd/test_dynamo_spmd.py
@@ -34,7 +34,7 @@ class DynamoSpmdInferenceTest(test_xla_sharding_base.XlaShardingTest):
 
   @classmethod
   def setUpClass(cls):
-    xr.set_use_spmd()
+    xr.use_spmd()
     super().setUpClass()
 
   def test_dynamo_spmd_basic(self):
@@ -72,19 +72,6 @@ class DynamoSpmdInferenceTest(test_xla_sharding_base.XlaShardingTest):
     self.assertEqual(met.counter_value('UncachedOutputSharding'), 1)
     dynamo_res = dynamo_linear(xla_y)
     self.assertEqual(met.counter_value('UncachedOutputSharding'), 1)
-
-  def test_dynamo_sharded_input(self):
-    device = xm.xla_device()
-    linear = SimpleLinear().to(device)
-    linear.eval()
-    xla_x = torch.randn(8, 128, device=device)
-    xs.mark_sharding(xla_x, self._get_mesh((1, self.n_devices)), (1, 0))
-    xla_res = linear(xla_x)
-    xm.mark_step()
-
-    dynamo_linear = torch.compile(linear, backend="openxla")
-    dynamo_res = dynamo_linear(xla_x)
-    torch.allclose(xla_res.cpu(), dynamo_res.cpu())
 
 
 if __name__ == '__main__':
--- a/test/spmd/test_dynamo_spmd.py
+++ b/test/spmd/test_dynamo_spmd.py
@@ -4,6 +4,7 @@ import sys
 import torch
 import torch.nn as nn
 import torch_xla
+import torch_xla.runtime as xr
 import torch_xla.core.xla_model as xm
 import torch_xla.experimental.xla_sharding as xs
 import torch_xla.debug.metrics as met
@@ -33,7 +34,7 @@ class DynamoSpmdInferenceTest(test_xla_sharding_base.XlaShardingTest):
 
   @classmethod
   def setUpClass(cls):
-    os.environ["XLA_USE_SPMD"] = "1"
+    xr.set_use_spmd()
     super().setUpClass()
 
   def test_dynamo_spmd_basic(self):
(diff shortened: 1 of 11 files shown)


yeounoh:
Still in draft, DO_NOT_MERGE_YET

yeounoh:
Let's start iterating this PR. Removing DO_NOT_MERGE_YET after running some unit tests with the new API. I will also update the SPMD user guide and the dev blog post for it.

test/spmd/test_spmd_graph_dump.py
```diff
@@ -18,7 +18,7 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
 
   @classmethod
   def setUpClass(cls):
-    os.environ["XLA_USE_SPMD"] = "1"
+    xr.set_use_spmd()
```
alanwaketan, line 21: How about just `xr.use_spmd()`?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -852,6 +852,7 @@ void InitXlaModuleBindings(py::module m) {
         [](const at::Tensor& tensor) { return GetTensorViewAliasId(tensor); });
   m.def("_xla_get_tensor_id",
         [](const at::Tensor& tensor) { return GetTensorId(tensor); });
+  m.def("_xla_get_lock_spmd_config", []() { return lock_spmd_config; });
```
alanwaketan, line 855: This syntax seems wrong?

test/spmd/test_spmd_graph_dump.py
yeounoh: lol, sounds good -- cc @JackCaoG wdyt?

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Hm,.. can't see what you are seeing. Could you help me find the bug here, and I will test it again?

test/spmd/test_spmd_graph_dump.py
JackCaoG: yea, sgtm

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Agreed, let me make that change.

test/spmd/test_spmd_graph_dump.py
JackCaoG: I guess we never meant to provide a option to do `xr.use_spmd(False)` right? It should be a one time thing

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Done, thanks @alanwaketan for the suggestion :)

test/spmd/test_spmd_graph_dump.py
yeounoh: @JackCaoG true, right now (the first step) it's a drop-in replacement to the flag.

test/spmd/test_dynamo_spmd.py
```diff
@@ -33,7 +34,7 @@ class DynamoSpmdInferenceTest(test_xla_sharding_base.XlaShardingTest):
 
   @classmethod
   def setUpClass(cls):
-    os.environ["XLA_USE_SPMD"] = "1"
+    xr.use_spmd()
```
JackCaoG, line 37: One thought, we need api to query whether spmd is being used on python level. I am thinking something along the line of `xr.spmd_enabled`. given that would it be better to name this api to `xr.enable_spmd`? @alanwaketan @yeounoh

yeounoh: I see, I can expose an API to check if it's SPMD. As SPMD is an execution "mode", I personally prefer simple API names like "is_spmd" & "use_spmd", but open to the alternatives, "spmd_enabled", "enable_spmd". @JackCaoG do you think it's better with "enable"?

yeounoh: All tests are passing, I will wait for @JackCaoG @alanwaketan 's feedback on the naming.

torch_xla/csrc/device.cpp
```diff
@@ -27,6 +27,10 @@ std::string XlaDeviceTypeToString(XlaDeviceType hw_type) {
   XLA_ERROR() << "Invalid device type";
 }
 
+// This is set when any device is initialized, so to prevent using non-virtual
+// device and virtual device together.
+static bool lock_spmd_config = false;
```
JackCaoG, line 32: nit, I felt like you should call this variabloe something along the line of `spmd_config_is_locked`. `lock_spmd_config` sounds like an action not a state.

JackCaoG (approved):
LGTM

alanwaketan (approved):
LGTM.

jonb377 (approved):

torch_xla/csrc/device.cpp
yeounoh: +1 will change as i rebase with the master

PR #5477, opened 2023-08-22
Support input sharding changed after first dynamo tracing

JackCaoG:
This is mostly to handle the case whether input is being in place updated by the SPMD after first execution. If the input sharding changed, we need to retrace the graph through dynamo.

I try to time how expensive this return sharding as string is, it took ~ 0.28ms for 6 runs and each runs has 7 tensors, one of them has sharding spec. I can rerun the experiment on llama.


torch_xla/core/dynamo_bridge.py
```diff
@@ -307,8 +309,32 @@ def extract_internal(xla_model: torch.fx.GraphModule):
...
+    nonlocal none_remover
+    nonlocal graph_input_matcher
+    nonlocal dumb_return_handler
+    nonlocal xla_args_need_update
+    print(xla_args_sharding_spec)
+    current_arg_sharding_spec = torch_xla._XLAC._get_xla_sharding_specs(args)
+    if current_arg_sharding_spec != xla_args_sharding_spec:
+      (xla_args_sharding_spec, args_and_out, graph_hash,
+      arg_index_to_need_update_index, none_remover, graph_input_matcher,
+      dumb_return_handler, xla_args_need_update) = extract_graph_helper(xla_model)
```
JackCaoG, line 336: now I think about it, I am not sure if `args` that has the new sharding info will be updated in the `xla_model.xla_args`, need to verify that tmr.

JackCaoG: this is fixed, we need to update `xla_model.xla_arg`

JackCaoG:
oh.. on llama something really scary happened, the step time went from 7 s to 22.8 s. Let me look into it..

JackCaoG:
regression should be fixed by https://github.com/pytorch/xla/pull/5489
TODO
1. fix cpu test
2. make threadshold configurable

JackCaoG:
This pr is ready for review.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1471,6 +1471,27 @@ void InitXlaModuleBindings(py::module m) {
     XLATensorPtr xtensor = bridge::GetXlaTensor(input);
     return GetXLAShardingSpec(xtensor);
   });
+  m.def("_get_xla_sharding_specs",
+        [](const std::vector<at::Tensor>& tensors) -> std::vector<std::string> {
+          tsl::profiler::TraceMe activity("_get_xla_sharding_specs",
+                                          tsl::profiler::TraceMeLevel::kInfo);
+          TORCH_LAZY_TIMED("_get_xla_sharding_specs");
+          std::vector<std::string> sharding_specs;
+          sharding_specs.reserve(sharding_specs.size());
```
wonjoolee95, line 1480: nit: should this be `tensors.size()` inside the `reserve()`?

torch_xla/core/dynamo_bridge.py
```diff
@@ -308,12 +315,53 @@ def extract_internal(xla_model: torch.fx.GraphModule):
...
+          arg_index_to_need_update_index, none_remover, graph_input_matcher,
+          dumb_return_handler, xla_args_need_update)
+
+
+def extract_internal(xla_model: torch.fx.GraphModule):
+  (xla_args_sharding_spec, args_and_out, graph_hash,
+   arg_index_to_need_update_index, none_remover, graph_input_matcher,
+   dumb_return_handler, xla_args_need_update) = extract_graph_helper(xla_model)
+  skip_checking_input_sharding_threashold = xu.getenv_as(
+      'XLA_DYNAMO_INPUT_SHARDING_CHECK_THREASHOLD', int, 5)
```
wonjoolee95, line 328: also nit: "threashold" -> "threshold" 😃

torch_xla/core/dynamo_bridge.py
```diff
@@ -308,12 +315,53 @@ def extract_internal(xla_model: torch.fx.GraphModule):
...
+        if torch_xla._XLAC._get_xla_sharding_specs(
+            args) != xla_args_sharding_spec:
+          # update the xla_args with the input with new sharding and retrace
+          xla_model.xla_args = args
+          (xla_args_sharding_spec, args_and_ou_copy, graph_hash,
+           arg_index_to_need_update_index, none_remover, graph_input_matcher,
+           dumb_return_handler,
+           xla_args_need_update) = extract_graph_helper(xla_model)
+
+        skip_checking_input_sharding_threashold -= 1
```
wonjoolee95, line 363: Not very urgent to this PR -- but if the sharding specs are different and we do the retracing, should we still subtract the `skip_checking_input_sharding_threashold` here? Since sharding specs can change any number of times, if `5` or user-provided `skip_checking_input_sharding_threashold` is lower than the actual number of change, dynamo won't recognize the sharding change anymore.

wonjoolee95 (approved):
LGTM overall, had some small comments not urgent to this PR.

torch_xla/core/dynamo_bridge.py
JackCaoG: yea I guess we should reset the `threashold` if it changed. good point

torch_xla/csrc/init_python_bindings.cpp
JackCaoG: nice catch!

torch_xla/core/dynamo_bridge.py
JackCaoG: done

JackCaoG:
I will merge once CI is green

test/spmd/test_dynamo_spmd.py
```diff
@@ -85,6 +85,92 @@ def test_dynamo_spmd_output_sharding_cache(self):
...
+    dynamo_linear = torch.compile(linear, backend="openxla")
+    dynamo_res = dynamo_linear(xla_x)
+    self.assertIn('CompileTime', met.metric_names())
+    self.assertEqual(met.metric_data('CompileTime')[0], 1)
+
+    # Shard the original input
+    xs.mark_sharding(xla_x, self._get_mesh((1, self.n_devices)), (1, 0))
+    dynamo_res_sharded = dynamo_linear(xla_x)
+    torch.allclose(dynamo_res.cpu(), dynamo_res_sharded.cpu())
+    # one graph is being generated by .cpu call above
```
alanwaketan, line 119: Should .cpu() trigger another compilation? Maybe a separate topic but I'm curious. I assume it shouldn't?

torch_xla/core/dynamo_bridge.py
```diff
@@ -308,12 +315,55 @@ def extract_internal(xla_model: torch.fx.GraphModule):
...
+          arg_index_to_need_update_index, none_remover, graph_input_matcher,
+          dumb_return_handler, xla_args_need_update)
+
+
+def extract_internal(xla_model: torch.fx.GraphModule):
+  (xla_args_sharding_spec, args_and_out, graph_hash,
+   arg_index_to_need_update_index, none_remover, graph_input_matcher,
+   dumb_return_handler, xla_args_need_update) = extract_graph_helper(xla_model)
+  skip_checking_input_sharding_threashold = xu.getenv_as(
+      'XLA_DYNAMO_INPUT_SHARDING_CHECK_THRESHOLD', int, 5)
```
alanwaketan, line 328: Why would we want this flag? And how to use this flag properly?

JackCaoG: getting the shardingspec for all input tensor is not free, it has some speed implication for the inference(~ 5% - 10%) on 7B. The idea of this flag is that if input sharding is the same for `XLA_DYNAMO_INPUT_SHARDING_CHECK_THRESHOLD` times, we will assume everything is OK and stop checking the input shardings. It is more or less a hack.

test/spmd/test_dynamo_spmd.py
JackCaoG: I am guessing the graph is something similar to 
```
%1 = device_data()
```
didn't check the actual HLO

torch_xla/core/dynamo_bridge.py
JackCaoG: The core idea is that user should not change the input sharding. We are mainly solving the problem for compiler overwrite the sharding. Compiler usually overwrites sharding for the 1st run and then sharding will be the same.

alanwaketan: Then I guess this flag should be off by default? We should ensure program correctness first than provide some risky hacks for users to tune the performance?

alanwaketan: > The core idea is that user should not change the input sharding. We are mainly solving the problem for compiler overwrite the sharding. Compiler usually overwrites sharding for the 1st run and then sharding will be the same.

Okay, it makes sense then.

alanwaketan (approved):

PR #5478, opened 2023-08-22
[SPMD] Fix HybridMesh ordering

alanwaketan:
Summary:
In xs.HybridMesh, it assumes the xr.global_runtime_device_attributes() will return the attributes according to the PyTorch/XLA's logical global ordinals. However, it turns out not to be the case.

To fix this, we pass the logical global ordinal as one of the attributes and xs.HybridMesh will sort the attributes according to this new attribute before using the array.

Test Plan:
PJRT_DEVICE=TPU USE_XLA_SPMD=1 python test/spmd/test_xla_sharding.py -v -k test_hybrid_mesh

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -618,40 +618,48 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     device_attributes_mock.return_value = [{
         'coords': [0, 0, 0],
         'core_on_chip': 0,
-        'slice_index': 0
+        'slice_index': 0,
+        'name': 'TPU:2'
     }, {
         'core_on_chip': 0,
         'coords': [1, 0, 0],
-        'slice_index': 0
+        'slice_index': 0,
+        'name': 'TPU:1'
     }, {
         'slice_index': 0,
         'core_on_chip': 0,
-        'coords': [0, 1, 0]
+        'coords': [0, 1, 0],
+        'name': 'TPU:0'
     }, {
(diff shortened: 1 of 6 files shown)


alanwaketan:
> There are some linter changes that are not relevant to the function of this PR.

Realized I have used a wrong version of yapf...

test/spmd/test_xla_sharding.py
```diff
@@ -618,40 +618,48 @@ def test_hybrid_mesh(self, xla_device_mock, device_attributes_mock):
...
         'coords': [1, 1, 0],
-        'slice_index': 1
+        'slice_index': 1,
+        'name': 'TPU:5'
     }]
     hybrid_mesh = xs.HybridMesh(
         ici_mesh_shape=(2, 2), dcn_mesh_shape=(num_slices, 1))
     self.assertEqual(hybrid_mesh.get_logical_mesh().tolist(),
-                     [[0, 1], [2, 3], [4, 5], [6, 7]])
+                     [[2, 1], [0, 3], [4, 7], [6, 5]])
```
jonb377, line 662: Is this the correct ordering for any specific hardware, or just the correct ordering given the mocked return value?

jonb377 (approved):
Thanks Jiewen!

test/spmd/test_xla_sharding.py
alanwaketan: This is the ordering from my funny mocks. Doesn't represent any real world use cases...

alanwaketan:
Thanks @jonb377!

PR #5484, opened 2023-08-23
Fix FSDP for Models with Frozen Weights

Liyang90:
Fixing the issue where XLAFSDP is not freeing the full params of the frozen parameters (`requires_grad==False`) after a backward pass.

### DIFF
--- a/test/test_torch_distributed_fsdp_frozen_weight.py
+++ b/test/test_torch_distributed_fsdp_frozen_weight.py
@@ -1,4 +1,5 @@
 import os
+
 os.environ["XLA_DISABLE_FUNCTIONALIZATION"] = "1"
 
 import sys
@@ -10,26 +11,26 @@ from torch_xla.distributed.fsdp import XlaFullyShardedDataParallel as FSDP
 
 
 def _mp_fn(index):
-    dev = xm.xla_device()
-    if xm.xla_device_hw(dev) not in ('TPU', 'GPU'):
-      print(
+  dev = xm.xla_device()
+  if xm.xla_device_hw(dev) not in ('TPU', 'GPU'):
+    print(
         'Default device {} is not a TPU or GPU device'.format(dev),
         file=sys.stderr)
-      return
+    return
 
-    model = nn.Linear(1024, 1024)
-    model.weight.requires_grad = False  # the weight param is frozen
+  model = nn.Linear(1024, 1024)
+  model.weight.requires_grad = False  # the weight param is frozen
 
-    model = FSDP(model)  # wrapping the linear module with FSDP
+  model = FSDP(model)  # wrapping the linear module with FSDP
 
-    input = torch.rand((2, 1024), device=xm.xla_device())
+  input = torch.rand((2, 1024), device=xm.xla_device())
 
-    output = model(input)
-    loss = torch.sum(output)
-    loss.backward()
-    assert not any(p._has_full_param for p in model.full_params), \
-      'Expecting all the full params to be freed at this moment.'
+  output = model(input)
+  loss = torch.sum(output)
+  loss.backward()
+  assert not any(p._has_full_param for p in model.full_params), \
+    'Expecting all the full params to be freed at this moment.'
 
 
 if __name__ == "__main__":
-    xmp.spawn(_mp_fn, args=())
+  xmp.spawn(_mp_fn, args=())
--- a/test/test_torch_distributed_fsdp_frozen_weight.py
+++ b/test/test_torch_distributed_fsdp_frozen_weight.py
@@ -1,7 +1,3 @@
-import os
-
-os.environ["XLA_DISABLE_FUNCTIONALIZATION"] = "1"
-
 import sys
 import torch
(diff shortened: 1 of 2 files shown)


JackCaoG:
any chance we can add a small test to make sure this does not regress?

gkroiz:
Just want to clarify here, the user should freeze parameters before sharding? What difference does it make if the parameters are frozen after sharding? If there is a distinction here, we should add to a brief message regarding how to correctly freeze parameters in the readme.

Liyang90:
> Just want to clarify here, the user should freeze parameters before sharding? 

Right 

> What difference does it make if the parameters are frozen after sharding? If there is a distinction here, we should add to a brief message regarding how to correctly freeze parameters in the readme.
The line of code here:
 https://github.com/pytorch/xla/blob/8a98a88724248fc3ce96a486d57f3dc81a2d260e/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py#L697
and 
https://github.com/pytorch/xla/blob/8a98a88724248fc3ce96a486d57f3dc81a2d260e/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py#L707

indicates the original `requires_grad` status is copied over to the sharded copy of the parameter and the full param place holder tensor as well. So it needs to be correctly set by then for the sharded parameter to have the correct value.

Setting `requires_grad` after wrapping my not have the behavior as expected.

The ultimate cause of the issue is the `post_backward_hook` does not fire on params that do not `requires_grad`, so setting `requires_grad=False` after wrapping seems to mask the issue because FSDP still think these params `requires_grad`.

carmocca:
@Liyang90 To be super clear, do you mean that until this PR lands, you recommend that `requires_grad=False` is done after wrapping?

Liyang90:
> @Liyang90 To be super clear, do you mean that until this PR lands, you recommend that `requires_grad=False` is done **after** wrapping?
> 
> And that it's done **before** after it lands?

No, I do not recommend setting `requires_grad=False` after the wrapping. That may lead to correctness error.

carmocca:
I see, but empirically, the memory usage is decreased by doing it after.

May I ask if this PR will land before the 2.1 release is published? This seems to be a critical bug in the finetuning case. Thank you

Liyang90:
> May I ask if this PR will land before the 2.1 release is published? This seems to be a critical bug in the finetuning case. Thank you

I agree. @alanwaketan

Liyang90:
> I see, but empirically, the memory usage is decreased by doing it after.

That might be because the FSDP wrapper still treats all the weights as `requires_grad==True`

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -1299,13 +1299,19 @@ def _wait_for_post_backward(self) -> None:
     # A backward pass is done, clean up below.
     def _finalize_parameters(fsdp_module: XlaFullyShardedDataParallel) -> None:
       """Helper used below on all fsdp modules."""
+      frozen_params = []
```
alanwaketan, line 1302: Can we identify which _free_full_params calls doesn't free the full params instead of adding a new pass at the end?

We might have bigger problem there.

alanwaketan: Will it be better to do it here?
https://github.com/pytorch/xla/blob/master/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py#L1327

alanwaketan: After looking into the code more, I think this place is fine.

test/test_torch_distributed_fsdp_frozen_weight.py
```diff
@@ -0,0 +1,36 @@
+import os
+
+os.environ["XLA_DISABLE_FUNCTIONALIZATION"] = "1"
```
alanwaketan, line 3: May I ask why? Does functionalization prevent this fix from working?

alanwaketan:
Also, can you make a nicer title for the PR before landing?

test/test_torch_distributed_fsdp_frozen_weight.py
Liyang90: This can be removed probably now. I previously check the `.data` shape of the full param for the test, and that would give me error because https://github.com/pytorch/xla/blob/5dc7e9058806d6f86c4ad7364743a2308ae7b623/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py#L1430C41-L1430C41 

Now I just test `_has_full_param`, so that's okay.

alanwaketan: .data is deprecated. You shouldn't use that anymore.

alanwaketan (approved):
LGTM.

wonjoolee95:
@Liyang90, is this ready to be merged? Asking since we want to cherry-pick into the r2.1 branch as well. Thanks!

PR #5485, opened 2023-08-23
[SPMD] Fix XLA_DUMP_POST_OPTIMIZATIONS test

alanwaketan:
Summary:
XLA_DUMP_POST_OPTIMIZATIONS was set as static which means that the value will be fixed during the whole test run for a particular test suite.

Therefore, let's make a separate file.

Test Plan:
PJRT_DEVICE=TPU USE_XLA_SPMD=1 python test/spmd/test_xla_sharding.py
PJRT_DEVICE=TPU USE_XLA_SPMD=1 python test/spmd/test_xla_sharding_hlo.py

### DIFF
--- a/torch_xla/csrc/ir_dump_util.cpp
+++ b/torch_xla/csrc/ir_dump_util.cpp
@@ -266,7 +266,7 @@ std::string DumpUtil::ToHlo(c10::ArrayRef<torch::lazy::Value> values,
   auto is_sharded = ShardingUtil::SetHloSharding(&lowering_ctx);
   xla::XlaComputation computation = ConsumeValue(lowering_ctx.BuildXla());
 
-  static bool dump_post_optimizations =
+  bool dump_post_optimizations =
       runtime::sys_util::GetEnvBool("XLA_DUMP_POST_OPTIMIZATIONS", false);
   if (dump_post_optimizations) {
     xla::Shape shape = MakeShapeWithDeviceLayout(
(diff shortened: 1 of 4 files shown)


torch_xla/csrc/ir_dump_util.cpp
```diff
@@ -266,7 +266,7 @@ std::string DumpUtil::ToHlo(c10::ArrayRef<torch::lazy::Value> values,
   auto is_sharded = ShardingUtil::SetHloSharding(&lowering_ctx);
   xla::XlaComputation computation = ConsumeValue(lowering_ctx.BuildXla());
 
-  static bool dump_post_optimizations =
+  bool dump_post_optimizations =
```
JackCaoG, line 269: I think this is intentional, all of the `XLA_xxxx` debug flag is static by design. What we usually do is control the env var in https://github.com/pytorch/xla/blob/master/test/run_tests.sh#L101

alanwaketan: Yea, but then I have to separate the test out. Otherwise, other tests will require it to be off.

JackCaoG: right.. I would recommend doing that. I am hesitating to make this env var different than every other debug flag which might cause confusion to user/developer.

alanwaketan: Sure.

alanwaketan: Done.

JackCaoG (approved):

alanwaketan:
Thanks, Jack.

PR #5486, opened 2023-08-23
[original author: mrnikwaws] fixing incorrect stride information on xla tensors 

aws-kingrj:
- fixing incorrect xla tensor stride information
- breaks log softmax lowering
- tested on AWS Neuron internal testing
- original author mrnikwaws
- old PR #5468

### DIFF
--- a/torch_xla/csrc/tensor_impl.cpp
+++ b/torch_xla/csrc/tensor_impl.cpp
@@ -228,8 +228,8 @@ void XLATensorImpl::SetupSymSizeProperties() {
   sym_sizes_ = sym_sizes;
 
   c10::SymInt prod{1};
- 
-  while (index > 0 ) {
+
+  while (index > 0) {
     --index;
     sym_strides[index] = prod;
     prod *= sym_sizes[index];
--- a/torch_xla/csrc/tensor_impl.cpp
+++ b/torch_xla/csrc/tensor_impl.cpp
@@ -229,12 +229,12 @@ void XLATensorImpl::SetupSymSizeProperties() {
 
   c10::SymInt prod{1};
  
-  while( index > 0 ) {
-    --index; 
+  while (index > 0 ) {
+    --index;
     sym_strides[index] = prod;
     prod *= sym_sizes[index];
   }
- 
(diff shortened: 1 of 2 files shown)


aws-kingrj:
Sorry, had to create a new PR because of the rebase conflicts

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -161,6 +161,11 @@ at::IntArrayRef XLATensorImpl::strides_custom() const {
   return strides_default();
 }
 
+c10::SymIntArrayRef XLATensorImpl::sym_strides_custom() const {
```
JackCaoG, line 164: can we add a test? It is a bit hard for me to tell what this pr is fixing and how it affect user.

mrnikwaws: Let me check in with Ryan and create a test.  

This issue occurs when pytorch checks the strides of a tensor have the same rank as the shape of a tensor.  By default XLA returns strides of one.  This will cause log_softmax to fail lowering, so using this lowering should form a simple test.

mrnikwaws:
```
(test_env) ubuntu@ip-172-31-63-138:~/waldronn/asr$ python
Python 3.8.10 (default, May 26 2023, 14:05:08) 
[GCC 9.4.0] on linux
>>> import torch
>>> import torch_neuronx
>>> example_input = torch.rand(1, 80, 643, dtype=torch.float)
>>> print(example_input.stride())
(51440, 643, 1)
>>> example_input = example_input.to('xla')
>>> print(example_input.stride())
(1,)
```
With the change the strides will match

JackCaoG: great, can we make this a test case? You can add it somewhere in https://github.com/pytorch/xla/blob/master/test/test_operations.py#L908 `test_operation` and create a new test for it.

aws-kingrj:
This has already been fixed in 2.1, after doing testing with log softmax lowering and getting the correct stride information

PR #5488, opened 2023-08-23
Support tuples in partition spec

jonb377:
In general, a single mesh should be used across all shardings in an SPMD program. This change simplifies the reuse of a mesh by allowing multiple logical mesh axes to be specified for a single tensor axis in the partition spec.

For example, with a (2, 16, 16) mesh with logical axes (replica, data, model), the same mesh can be used for parameter and activation sharding as follows:

```
mesh = xs.Mesh(range(512), (2, 16, 16), ('replica', 'data', 'model'))

xs.mark_sharding(input, mesh, (('replica', 'data'), None))
xs.mark_sharding(parameter, mesh, ('data', 'model'))
xs.mark_sharding(activation, mesh, (('replica', 'data'), 'model')
```

To summarize the changes:
- We will always permute the mesh in `_get_tile_assignment`, even in the partial replication case
- `_get_group_assignment` now operates over the permuted tile assignment instead of the unmodified logical mesh.
- `_get_group_assignment` no longer takes the partition spec as input; instead, the group tile shape is derived based on the replicated tensor dimensions and the tensor rank, both of which are derived from the partition spec.

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -337,6 +337,16 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     actual = (xt1 + t2).cpu()
     self.assertTrue(torch.allclose(expected, actual))
 
+  @unittest.skipUnless(xr.global_runtime_device_count() > 1,
+                       "Multiple devices required for tupled partition spec")
+  def test_tupled_partition_spec(self):
+    mesh = self._get_mesh((2, self.n_devices // 2))
+    t = torch.randn(16).to(xm.xla_device())
+    xs.mark_sharding(t, mesh, ((0, 1),))
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(t), "{devices=[%d]%s}" %
+        (self.n_devices, ','.join(str(x) for x in range(self.n_devices))))
+
   def test_partial_replication_addmm(self):
     device = xm.xla_device()
     z_dim = 2 if self.n_devices >= 4 else 1
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1,5 +1,6 @@
 import copy
 
+from itertools import chain
 import unittest
 from unittest.mock import patch
 import math
@@ -346,6 +347,40 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertEqual(
         torch_xla._XLAC._get_xla_sharding_spec(t), "{devices=[%d]%s}" %
         (self.n_devices, ','.join(str(x) for x in range(self.n_devices))))
+  
+  @unittest.skipUnless(xr.global_runtime_device_count() >= 4,
+                       "Multiple devices required for tupled partition spec")
+  def test_named_partial_tupled_partition_spec(self):
+    mesh = xs.Mesh(range(self.n_devices), (1, 2, self.n_devices // 2), ('r', 'b', 'm'))
+    # Shard the first dimension on `r` and `b`, replicate the second dimension
+    t = torch.randn(16, 16).to(xm.xla_device())
+    xs.mark_sharding(t, mesh, (('r', 'b'), None))
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(t), "{devices=[2,1,%d]%s last_tile_dim_replicate}" %
+        (self.n_devices // 2, ','.join(str(x) for x in range(self.n_devices))))
+
+    # Replicate the first dimension, shard the second on `b` and `m`
+    u = torch.randn(16, 16).to(xm.xla_device())
+    xs.mark_sharding(u, mesh, (None, ('b', 'm')))
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(u), "{devices=[1,%d]%s}" %
+        (self.n_devices, ','.join(str(x) for x in range(self.n_devices))))
+
+    # Replicate the first dimension, shard the second on `r` and `m`
+    v = torch.randn(16, 16).to(xm.xla_device())
+    xs.mark_sharding(v, mesh, (None, ('r', 'm')))
+    device_order = chain(range(0, self.n_devices, 2), range(1, self.n_devices, 2))
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(v), "{devices=[1,%d,2]%s last_tile_dim_replicate}" %
+        (self.n_devices // 2, ','.join(str(x) for x in device_order)))
+
+    # Replicate the first dimension, shard the second on `m` and `b`
+    v = torch.randn(16, 16).to(xm.xla_device())
+    xs.mark_sharding(v, mesh, (None, ('m', 'b')))
+    device_order = chain(range(0, self.n_devices, 2), range(1, self.n_devices, 2))
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(v), "{devices=[1,%d]%s}" %
+        (self.n_devices, ','.join(str(x) for x in device_order)))
 
   def test_partial_replication_addmm(self):
     device = xm.xla_device()
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -388,6 +388,18 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
         (self.n_devices, ','.join(str(x) for x in device_order)))
 
   @unittest.skipUnless(xr.global_runtime_device_count() > 1,
+                       'Multiple devices required for tupled partition spec')
+  def test_multiple_tuples_in_spec(self):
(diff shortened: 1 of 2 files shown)


test/spmd/test_xla_sharding.py
```diff
@@ -336,6 +336,16 @@ def test_mark_sharding_partial_unordered(self):
     actual = (xt1 + t2).cpu()
     self.assertTrue(torch.allclose(expected, actual))
 
+  @unittest.skipUnless(xr.global_runtime_device_count() > 1,
+                       "Multiple devices required for tupled partition spec")
+  def test_tupled_partition_spec(self):
```
jonb377, line 341: I'm adding more tests, just wanted to get the PR out for review earlier.

jonb377:
> what happened to the `replica` mesh dimension here?

When a mesh dimension is not specified, the tensor will be replicated along it. For model parameters, we want to replicate along the mesh's replica axis, so this example demonstrates that we can achieve that by just omitting the `replica` axis from the partition spec.

torch_xla/experimental/xla_sharding.py
```diff
@@ -492,11 +533,12 @@ class ShardingSpec:
   @xr.requires_pjrt
   def __post_init__(self):
     partition_spec, mesh = self.partition_spec, self.mesh
-    self._tile_assignment = _get_tile_assignment(mesh, partition_spec)
+    tile_assignment, partition_spec = _get_tile_assignment(mesh, partition_spec)
```
JackCaoG, line 536: if this method also returns partition_spec, you should call it something like `_get_tile_assignment_and_update_partition_spec`

jonb377: Good point, maybe I can do a bit more refactoring and not return the partition spec here. It feels a bit clunky to be rewriting the partition spec anyway.

JackCaoG (approved):
I have to admit I don't fully understand the pr, the test case make sense. I am going to approve to unblock.

torch_xla/experimental/xla_sharding.py
```diff
@@ -331,13 +331,44 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
+  The resulting partition spec and tiling assignment should be used to generate
+  the group assignment and replica groups in the case of partial replication,
+  and the tiling assignment should be used to create the final sharding spec.
+  """
+  # Flatten the partition spec and ensure that it is fully specified over the
+  # mesh for permutation.
+  tiled_dims = [x for x in partition_spec if x is not None]
+  permutation = np.hstack(tiled_dims).tolist() if tiled_dims else []
+  missing_axes = sorted(set(range(len(mesh.shape()))) - set(permutation))
+  tile_assignment = mesh.get_logical_mesh().transpose(permutation +
```
alanwaketan, line 356: I was shocked by the transpose grammar and then realized it's np.array. I'm curious what's the advantage to use np.array over cpu torch.tensor. This is just for discussion. There is no needs to make any changes.

torch_xla/experimental/xla_sharding.py
```diff
@@ -331,13 +331,44 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
+                                                      missing_axes)
+
+  # For any tuples in the partition_spec, the grouped axes will be adjacent
+  # after the permutation. Combine these dimensions into a single axis.
+  for i, spec in reversed(list(enumerate(tiled_dims))):
```
alanwaketan, line 361: Why reversed(list())? Is enumerate enough?

torch_xla/experimental/xla_sharding.py
```diff
@@ -331,13 +331,44 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
+    if isinstance(spec, tuple):
+      shape = tile_assignment.shape
+      tile_assignment = tile_assignment.reshape(shape[:i] + (-1,) +
```
alanwaketan, line 364: This `-1` is the magic to infer the product of the tuple. lol.

torch_xla/experimental/xla_sharding.py
```diff
@@ -355,16 +386,17 @@ def _get_group_assignment(
     # will be used to group replication devices.
     tile_dims = [d for d in partition_spec if d is not None]
 
-    group_list = [np.array(mesh.get_logical_mesh().tolist())]
+    logical_mesh = tile_assignment
```
alanwaketan, line 389: Not sure what this is about? Maybe we can stick on the same term either logical_mesh or tile_assignment.

torch_xla/experimental/xla_sharding.py
alanwaketan: You mean in-placed updating the partition spec?

alanwaketan (approved):
LGTM. Left some comments for discussion.

Great work. @jonb377

test/spmd/test_xla_sharding.py
```diff
@@ -336,6 +337,65 @@ def test_mark_sharding_partial_unordered(self):
...
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(v), "{devices=[1,%d]%s}" %
+        (self.n_devices, ','.join(str(x) for x in device_order)))
+
+  @unittest.skipUnless(xr.global_runtime_device_count() > 1,
+                       'At least 2 devices needed for 2D mesh')
+  def test_3d_tensor_2d_mesh(self):
+    mesh = self._get_mesh((2, self.n_devices // 2))
+    t = torch.randn(16, 16, 16).to(xm.xla_device())
+    xs.mark_sharding(t, mesh, (None, 0, 1))
```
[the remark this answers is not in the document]
jonb377: I noticed that we were missing coverage for this case, and my initial approach broke it.

torch_xla/experimental/xla_sharding.py
jonb377: I've cleaned this up in the latest revision

torch_xla/experimental/xla_sharding.py
jonb377: Just generally having to modify it. The partition spec has served its purpose in `_get_tile_assignment`, we should be able to get the group assignment without using the partition spec.

torch_xla/experimental/xla_sharding.py
jonb377: Good catch! You're right - we should be collapsing the axes in a forward iteration, not reverse.

My tests didn't catch this because there's only one tuple per spec, I'll add some more coverage.

test/spmd/test_xla_sharding.py
```diff
@@ -336,6 +337,65 @@ def test_mark_sharding_partial_unordered(self):
...
+    v = torch.randn(16, 16).to(xm.xla_device())
+    xs.mark_sharding(v, mesh, (None, ('r', 'm')))
+    device_order = chain(
+        range(0, self.n_devices, 2), range(1, self.n_devices, 2))
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(v),
+        "{devices=[1,%d,2]%s last_tile_dim_replicate}" %
+        (self.n_devices // 2, ','.join(str(x) for x in device_order)))
+
+    # Replicate the first dimension, shard the second on `m` and `b`
```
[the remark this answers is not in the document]
jonb377: I'm not sure on the approach taken by JAX here... I'll add a comment for now, but we should try to match JAX's behavior on this.

test/spmd/test_xla_sharding.py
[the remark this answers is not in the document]
alanwaketan: Good catch.

test/spmd/test_xla_sharding.py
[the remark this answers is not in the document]
jonb377: Confirmed the order matters in JAX as well:

```
>>> z = jax.device_put(x, NamedSharding(mesh, P(('a', 'b'))))
>>> jax.debug.visualize_array_sharding(z)
┌───────────────────────┐
│         TPU 0         │
├───────────────────────┤
│         TPU 1         │
├───────────────────────┤
│         TPU 2         │
├───────────────────────┤
│         TPU 3         │
└───────────────────────┘
>>> z = jax.device_put(x, NamedSharding(mesh, P(('b', 'a'))))
>>> jax.debug.visualize_array_sharding(z)
┌───────────────────────┐
│         TPU 0         │
├───────────────────────┤
│         TPU 2         │
├───────────────────────┤
│         TPU 1         │
├───────────────────────┤
│         TPU 3         │
└───────────────────────┘
```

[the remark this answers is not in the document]
alanwaketan: Awesome!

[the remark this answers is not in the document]
JackCaoG: we should have this `visualize_array_sharding` haha, it looks so nice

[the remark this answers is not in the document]
yeounoh: Lol very good.

torch_xla/experimental/xla_sharding.py
yeounoh: +1

PR #5491, opened 2023-08-24
[SPMD] Patch nn.Linear

alanwaketan:
Summary:
This pull request introduces a patched version of `torch.nn.functional.linear` that uses einsum instead of torch.matmul which will flatten the tensors to 2D and collide the sharded dimensions. The torch.matmul default behavior makes it very hard for XLA compiler to propagate the sharding annotation.

Test Plan:
PJRT_DEVICE=CPU python test/test_operations.py -v -k test_patched_linear

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -44,7 +44,7 @@ import torch_xla.utils.serialization as xser
 import torch_xla.core.xla_model as xm
 import torch_xla.core.functions as xf
 import torch_xla.debug.profiler as xp
-# import torchvision
+import torchvision
 import unittest
 import test_utils
 
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1714,6 +1714,34 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
 
     self.assertTrue(torch.allclose(linear.bias.grad.cpu(), linear_cpu.bias.grad))
 
+  def test_patched_linear_2D_bias(self):
+    linear_cpu = nn.Linear(2, 4)
+    input_cpu = torch.randn(4, 2, requires_grad=True)
+    input_cpu.retain_grad()
+    output_cpu = linear_cpu(input_cpu)
+
+    # It looks like nn.Module.to is in-place.
+    linear = copy.deepcopy(linear_cpu).to('xla')
+    apply_xla_patch_to_nn_linear(linear, xs.xla_patched_nn_linear_forward)
+    input = copy.deepcopy(input_cpu).to('xla')
+    input.retain_grad()
+    output = linear(input)
+
+    # Make sure the forward result is correct.
+    self.assertTrue(torch.allclose(output.cpu(), output_cpu))
+
+    # Now work on the backward.
+    linear_cpu.weight.retain_grad()
+    loss_cpu = output_cpu.sum()
+    loss_cpu.backward()
+
+    loss =output.sum()
+    loss.backward()
+
+    self.assertTrue(torch.allclose(linear.weight.grad.cpu(), linear_cpu.weight.grad))
+    self.assertTrue(torch.allclose(input.grad.cpu(), input_cpu.grad))
+    self.assertTrue(torch.allclose(linear.bias.grad.cpu(), linear_cpu.bias.grad))
+
 
 class MNISTComparator(nn.Module):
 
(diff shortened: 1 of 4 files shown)


JackCaoG:
I thought we decided to let XLA compiler handle the reshape?

torch_xla/distributed/fsdp/utils.py
```diff
@@ -122,7 +122,9 @@ def _xla_patched_nn_linear_forward(m, input):
   return XLAPatchedLinear.apply(input, m.weight, m.bias)
 
 
-def apply_xla_patch_to_nn_linear(module):
+def apply_xla_patch_to_nn_linear(module,
```
JackCaoG, line 125: will this function always be called? or only when fsdp being used.

alanwaketan: It's called in the FSDP wrapper. For GSPMD, we don't have a wrapper so we have to call it explicitly.

test/test_operations.py
```diff
@@ -1656,6 +1658,124 @@ def test_conv2d_backward(self):
...
 
+  def test_patched_linear_3D(self):
+    linear_cpu = nn.Linear(2, 4, bias=False)
+    input_cpu = torch.randn(4, 3, 2, requires_grad=True)
+    input_cpu.retain_grad()
+    output_cpu = linear_cpu(input_cpu)
+
+    # It looks like nn.Module.to is in-place.
+    linear = copy.deepcopy(linear_cpu).to('xla')
+    apply_xla_patch_to_nn_linear(linear, xs.xla_patched_nn_linear_forward)
```
JackCaoG, line 1669: I assume if you call it once, the whole file will have the linear patched, no?

JackCaoG: nvm, I see what's happening here.

torch_xla/experimental/xla_sharding.py
```diff
@@ -521,3 +521,42 @@ def apply(self, t: torch.Tensor):
...
+  dimensions. The torch.matmul default behavior makes it very hard for XLA compiler
+  to propagate the sharding annotation.
+
+  TODO (alanwaketan): Let's patch it on the dispatcher level.
+  """
+
+  @staticmethod
+  def forward(ctx, input, weight, bias=None):
+    # bias is an optional argument
+    ctx.save_for_backward(input, weight, bias)
```
JackCaoG, line 539: hmm.. are we doing explicit gradient checkpointing here?

JackCaoG: not family with how nn moudle is being written but I mainly ask because the `torch.no_grad` below.

JackCaoG (approved):
lgtm, thanks

torch_xla/experimental/xla_sharding.py
alanwaketan: No, it's just a convention in autograd function to save the activations. So not the gradient checkpointing or re-materialization.

alanwaketan:
Thanks, Jack!

torch_xla/distributed/fsdp/utils.py
yeounoh: @alanwaketan did it, using einsum instead of matmul, improve the benchmark training result? I recall that it didn't actually have an impact (or lowered MFU in @jonb377 's result?)? If so, we should make a comment explaining that this may reduce the performance, too. And +1 to not enabling by default...

yeounoh:
> > I thought we decided to let XLA compiler handle the reshape?
> 
> For XLA compiler, I guess that's a workaround. Ultimately, I think it's right for us to handle this ourselves.

If this change is not what we used... not sure if we should merge & enable for FSDP. Iirc, didn't @AlexWertheim say that using einsum resulted in lower MFU in FSDP? @jonb377 had a similar result for GSPMD lately, no? Maybe we leave a comment so people can use this with caution on demand?

torch_xla/distributed/fsdp/utils.py
jonb377: The lower MFU was from user-mode FSDP, I don't have a head-to-head benchmark on SPMD. @alanwaketan had some data to suggest it improves performance for SPMD.

alanwaketan: This is the solution in our end to stop the reshape on nn.Linear's matmul.

alanwaketan:
> > > I thought we decided to let XLA compiler handle the reshape?
> > 
> > 
> > For XLA compiler, I guess that's a workaround. Ultimately, I think it's right for us to handle this ourselves.
> 
> If this change is not what we used... not sure if we should merge & enable for FSDP. Iirc, didn't @AlexWertheim say that using einsum resulted in lower MFU in FSDP? @jonb377 had a similar result for GSPMD lately, no? Maybe we leave a comment so people can use this with caution on demand?

I used this change for all the latest results. I don't think it introduces any performance regression on my side.

PR #5493, opened 2023-08-24
Refactor stablehlo API and put them in official location.

qihqi:
Changes include:

* make end point in torch_xla/__init__.py for exposed APIs `torch_xla.save_as_stablehlo` and `torch_xla.save_torch_model_as_stablehlo`. 
* All tf related integration to its own file.
* Remove `args` as argument (because it will spear inside of ExportedProgram) but allow user to override it (which we use for now.


lsy323 (approved):
LGTM

PR #5494, opened 2023-08-24
Always use ExecuteReplicated with SPMD

jonb377:
SPMD computations should always follow the ExecuteReplicated codepath, even when there are no sharding annotations present in the HLO.

### DIFF
--- a/torch_xla/csrc/xla_graph_executor.cpp
+++ b/torch_xla/csrc/xla_graph_executor.cpp
@@ -1329,8 +1329,10 @@ XLAGraphExecutor::CompilationResult XLAGraphExecutor::Compile(
         torch::lazy::Output(ir_value.node.get(), ir_value.index));
     lowering_ctx.AddResult(root);
   }
+  // Always execute sharded when running in SPMD mode
+  bool is_sharded = (coll.device == GetVirtualDevice());
   // Annotate HLO sharding selectively in the compuation.
-  bool is_sharded = ShardingUtil::SetHloSharding(&lowering_ctx);
+  ShardingUtil::SetHloSharding(&lowering_ctx);
 
   std::vector<std::pair<int64_t, int64_t>> input_output_alias_pair;
   // TODO(yeounoh) aliasing is disabled for partitioned computation,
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1329,8 +1329,10 @@ XLAGraphExecutor::CompilationResult XLAGraphExecutor::Compile(
         torch::lazy::Output(ir_value.node.get(), ir_value.index));
     lowering_ctx.AddResult(root);
   }
+  // Always execute sharded when running in SPMD mode
+  bool is_sharded = (coll.device == GetVirtualDevice());
   // Annotate HLO sharding selectively in the compuation.
-  bool is_sharded = ShardingUtil::SetHloSharding(&lowering_ctx);
+  ShardingUtil::SetHloSharding(&lowering_ctx);
```
alanwaketan, line 1335: Should we remove is_sharded in SetHloSharding?

alanwaketan: It looks like it's being used by other places. nvm.

alanwaketan (approved):
LGTM. Let's trigger the TPU CI to make sure there is no regression.

Not sure if we can add any test cases?

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG, line 1335: can we add a test for non-sharding execution with spmd mode?

jonb377: It looks like we already have some coverage here: https://github.com/pytorch/xla/blob/4f378e694a68192fd634a070a3485c73afab1c71/test/spmd/test_xla_virtual_device.py#L55

I'll sync up with @alanwaketan on the issue he encountered and use that as a test case

jonb377: Added a test case which induces the error without this patch

alanwaketan: Sorry for missing your messages, but the `Unknown device: SPMD:0` is the right error.

JackCaoG (approved):

jonb377:
TPU CI passed, submitting this change.

PR #5498, opened 2023-08-25
[SPMD] Use actual device instead of SPMD:0 as the BackendDevice

### LINKED ISSUE #5497: [SPMD] AMP autocast failed when set XLA_USE_SPMD=1
## 🐛 Bug

AMP autocast failed when set XLA_USE_SPMD=1

## To Reproduce

Let's use the following example

```
import os
import torch

from torch_xla.amp import autocast


device = xm.xla_device()
t1 = torch.ones([2,3], device=device, dtype=torch.float32)
t2 = torch.ones([3,2], device=device, dtype=torch.float32)

with autocast(device, dtype=torch.bfloat16):
  t3 = torch.matmul(t1, t2)

print(t3.dtype)
xm.mark_step()
```
, the expected result is `torch.bfloat16`, but when setting `os.environ["XLA_USE_SPMD"] = "1"`, the actual output
is `torch.float32`.

I found that the reason is that when XLA_USE_SPMD is set to 1, the devices in the C++ are all SPMD:0, and PyTorch autocast does not recognize the SPMD:0 device. cc @JackCaoG 

## Environment

 - Reproducible on XLA backend [GPU/TPU]: GPU
 - torch_xla version: master

baoleai:
fix #5497, we should use actual device type instead of SPMD:0 as BackendDevice to make it recognizable to PyTorch.

### DIFF
--- a/torch_xla/csrc/aten_xla_bridge.cpp
+++ b/torch_xla/csrc/aten_xla_bridge.cpp
@@ -39,15 +39,10 @@ class AtenXlaDeviceMapper {
 
  private:
   AtenXlaDeviceMapper() {
-    if (UseVirtualDevice()) {
-      devices_.emplace_back(ParseDeviceString("SPMD:0"));
-      devices_ordinals_[devices_.back()] = 0;
-    } else {
-      for (auto& device_str :
-           torch_xla::runtime::GetComputationClient()->GetLocalDevices()) {
-        devices_.emplace_back(ParseDeviceString(device_str));
-        devices_ordinals_[devices_.back()] = devices_.size() - 1;
-      }
+    for (auto& device_str :
+         torch_xla::runtime::GetComputationClient()->GetLocalDevices()) {
+      devices_.emplace_back(ParseDeviceString(device_str));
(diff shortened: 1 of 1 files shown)


torch_xla/csrc/aten_xla_bridge.cpp
```diff
@@ -39,15 +39,10 @@ class AtenXlaDeviceMapper {
...
-    } else {
-      for (auto& device_str :
-           torch_xla::runtime::GetComputationClient()->GetLocalDevices()) {
-        devices_.emplace_back(ParseDeviceString(device_str));
-        devices_ordinals_[devices_.back()] = devices_.size() - 1;
-      }
+    for (auto& device_str :
+         torch_xla::runtime::GetComputationClient()->GetLocalDevices()) {
+      devices_.emplace_back(ParseDeviceString(device_str));
+      devices_ordinals_[devices_.back()] = devices_.size() - 1;
```
JackCaoG, line 45: this will break whole SPMD, I think we need to fix it in a different way...

baoleai: Yes, found UT break.

PR #5513, opened 2023-08-28
mark_sharding over a replicated tensor is allowed.

yeounoh:
Replicated sharding should be treated as unsharded in SPMD mode. Thus, allowing `mark_sharding` to overwrite it when a tensor is already marked replicated.

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1432,12 +1432,13 @@ void InitXlaModuleBindings(py::module m) {
       cpu_tensor = xtensor->CurrentTensorData().value();
     } else {
       // A new input tensor is not expected to be sharded. But sometimes,
-      // the same input is used sharding annotation, in which case we can
-      // skip if it's the same sharding; however, if it's the same input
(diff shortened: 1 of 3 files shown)


alanwaketan (approved):
LGTM.

jonb377 (approved):
Thanks Yeounoh!

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1437,7 +1437,8 @@ void InitXlaModuleBindings(py::module m) {
       // with a different sharding then we block & ask the user to clear
       // the existing sharding first.
       auto current_sharding_spec = xtensor->sharding_spec();
-      if (current_sharding_spec) {
+      if (current_sharding_spec && (current_sharding_spec->sharding.type() !=
```
jonb377, line 1440: More of a general comment, but what do you think about adding a TraceMe for the profiler after this check (at L1448)? That way it's easy to tell if a TransferFromServer call is due to resharding.

yeounoh: Good idea :)

yeounoh:
Locally tested & passed.

PR #5514, opened 2023-08-28
Add stablehlo.md

lsy323:


lsy323:
cc @burmako @GleasonK

lsy323:
Had offline discussion with @qihqi, we will expose API in https://github.com/pytorch/xla/blob/master/torch_xla/stablehlo.py to users, @qihqi will prepare the updated StableHLO README.

PR #5520, opened 2023-08-29
Enable megacore_dense by default

alanwaketan:
Summary:
This change enables megacore_dense by default to allow asynchorous cc ops especailly for GSPMD.

Test Plan:
CI

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -40,6 +40,7 @@ def _setup_default_env():
   _set_missing_env('GRPC_VERBOSITY', 'ERROR')
   _set_missing_env('ALLOW_MULTIPLE_LIBTPU_LOAD', '1')
   _set_missing_env('TPU_ML_PLATFORM', 'PyTorch/XLA')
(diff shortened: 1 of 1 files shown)


torch_xla/__init__.py
```diff
@@ -40,6 +40,7 @@ def _setup_default_env():
   _set_missing_env('GRPC_VERBOSITY', 'ERROR')
   _set_missing_env('ALLOW_MULTIPLE_LIBTPU_LOAD', '1')
   _set_missing_env('TPU_ML_PLATFORM', 'PyTorch/XLA')
+  _set_missing_env('TPU_MEGACORE', 'megacore_dense')
```
jonb377, line 43: Is this a no-op for non-v4 TPUs?

jonb377 (approved):
Thanks!

torch_xla/__init__.py
alanwaketan: I guess so? TPU CI will tell us.

alanwaketan:
Thanks, Jon. Merged since CIs are green.

PR #5522, opened 2023-08-29
Remove TF flags from .bazelrc

apivovarov:
branch 2.1.0 and master do not use/build tensorflow anymore.
But we still have TF related flags/options in `.bazelrc`

I removed some TF related flags/options (from `.bazelrc`) which are not used anymore.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -11,13 +11,6 @@ build --enable_platform_specific_config
 
 build --experimental_cc_shared_library
 
-# Disable enabled-by-default TensorFlow features that we don't care about.
-build --define=no_aws_support=true
-build --define=no_hdfs_support=true
-build --define=no_hdfs_support=true
-build --define=no_kafka_support=true
-build --define=no_ignite_support=true
-
 build --define=grpc_no_ares=true
 
 build -c opt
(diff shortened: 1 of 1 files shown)


JackCaoG:
@will-cromar @stgpetrovic Can you take a look?

.bazelrc
```diff
@@ -44,9 +37,6 @@ build:avx_linux --host_copt=-mavx
 build:native_arch_posix --copt=-march=native
 build:native_arch_posix --host_copt=-march=native
 
-build:mkl_open_source_only --define=tensorflow_mkldnn_contraction_kernel=1
-
-build:cuda --repo_env TF_NEED_CUDA=1
```
will-cromar, line 49: Some of these options are still in use in openxla/xla, e.g. https://github.com/openxla/xla/blob/ab3e2de65f3c077e4ea7b4d474dd2453e54e9c8e/.bazelrc#L228

Feel free to remove any options here that don't exist in openxla. However, I think we should leave in anything they still use to be safe. Their build is very complicated and nobody really understands it, so I'm inclined to play it safe. @stgpetrovic what do you think?

PR #5526, opened 2023-08-30
Replace pytree_to_str with treespec_dumps

apivovarov:
`pytree_to_str` is deprecated. [Internally](https://github.com/pytorch/pytorch/blob/main/torch/utils/_pytree.py#L532) it simply calls `treespec_dumps`. 

We should use `treespec_dumps` directly instead of using `pytree_to_str`.

This PR fixes the following warning:
```
>>> stablehlo_program = exported_program_to_stablehlo(exported)
/usr/local/lib/python3.8/site-packages/torch/utils/_pytree.py:533: UserWarning: pytree_to_str is deprecated. Please use treespec_dumps
  warnings.warn("pytree_to_str is deprecated. Please use treespec_dumps")
```

### DIFF
--- a/torch_xla/stablehlo.py
+++ b/torch_xla/stablehlo.py
@@ -325,8 +325,8 @@ def _exported_program_to_stablehlo_bundle(exported_model,
       input_signature=input_signatures,
       output_signature=output_signature,
       input_locations=input_locations,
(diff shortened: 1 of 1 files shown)


qihqi (approved):
thanks!

PR #5557, opened 2023-09-12
Update PJRT documentation for the 2.1 release

will-cromar:


README.md
```diff
@@ -64,12 +68,13 @@ To update your existing training loop, make the following changes:
...
 
 
-```
+```diff
  import torch.distributed as dist
 -import torch.multiprocessing as mp
 +import torch_xla.core.xla_model as xm
 +import torch_xla.distributed.parallel_loader as pl
 +import torch_xla.distributed.xla_multiprocessing as xmp
++import torch_xla.experimental.pjrt_backend
```
JackCaoG, line 77: why do we need this import?

will-cromar: This registers the `pjrt://` init method. I missed it in the original draft.

JackCaoG: can we make it an implicit important somehow? This seems to be easy to forget.

will-cromar: Do you mean to register the rendezvous handler implicitly in `torch_xla.__init__`? I guess that's probably safe enough as long as I separate out the [hack for multithreading](https://github.com/pytorch/xla/blob/a1b927964543db0a23510bd3284ee9725800246d/torch_xla/experimental/pjrt_backend.py#L50-L55). I will send a PR.

will-cromar: Actually, I think it makes more sense to include this in `xla_backend`.

will-cromar: Done.

JackCaoG (approved):

PR #5560, opened 2023-09-13
Change `pjrt://` init method to `xla://`

will-cromar:
- Rename `pjrt://` init method to `xla://`
- Register `xla://` init method with `xla` backend to save the user an import statement
- Experimental `pjrt_backend` package will remain and register the old `pjrt://` name for compatibility
  - Experimental package will also keep hack for TPU v2/v3. This will not move into stable API


alanwaketan (approved):
LGTM.

JackCaoG (approved):

PR #5563, opened 2023-09-13
Support kwargs in export and converting to stablehlo

qihqi:
Torch side is adding support for kwargs in export's pytree in_spec: 
https://github.com/pytorch/pytorch/pull/109160

This change will make it so save_as_stablehlo works with and without the above change.


angelayi (approved):
Thanks Han for the quick fix!! 😊

lsy323 (approved):
LGTM

PR #5440, opened 2023-08-11
Use setBufferDonor api instead of SetUpAlias

JackCaoG:
XLA pin update needs to go in first.

PR #5567, opened 2023-09-14
Support for mulit-hop input output aliases.

baoleai:
For in-place operations, XLA utilizes InputOutputAlias to reuse buffers for inputs and outputs. However, the current implementation only allows reusing buffers for a single in-place operation. If a tensor undergoes multiple in-place operations, only the inputs and outputs of the last operation can reuse buffers, while previous in-place operations cannot. The goal of this commit is to allow multiple in-place operations to reuse a single buffer, thereby reducing memory usage.

cc @JackCaoG @bdhirsh

### DIFF
--- a/test/test_input_output_aliases.py
+++ b/test/test_input_output_aliases.py
@@ -28,6 +28,7 @@ class InputOutputAliasesTest(unittest.TestCase):
     # check in place op aliasing.
     t3 = t1 + t2
     t1 *= 2.0
+    t1 *= 2.0
     t2 += 2.0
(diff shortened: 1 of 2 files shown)


JackCaoG:
XLA had a new API that allow us to just annoate buffer that we don;t need and compiler will pick the aliasing for us. I had a pr open in https://github.com/pytorch/xla/pull/5440. Maybe we should just do that.

baoleai:
Looking forward to the merge of new features.

PR #5568, opened 2023-09-14
fix GPU AMP under SPMD

### LINKED ISSUE #5497: [SPMD] AMP autocast failed when set XLA_USE_SPMD=1

baoleai:
fix #5497 
cc @JackCaoG @yeounoh

### DIFF
--- a/torch_xla/csrc/aten_xla_bridge.cpp
+++ b/torch_xla/csrc/aten_xla_bridge.cpp
@@ -299,6 +299,13 @@ std::vector<torch::lazy::BackendDevice> GetBackendDevices() {
   return AtenXlaDeviceMapper::Get()->GetAllDevices();
 }
 
+XlaDeviceType GetRunTimeDeviceType() {
+    std::string device_str =
+        torch_xla::runtime::GetComputationClient()->GetAllDevices()[0];
+    torch::lazy::BackendDevice device = ParseDeviceString(device_str);
+    return static_cast<XlaDeviceType>(device.type());
+}
+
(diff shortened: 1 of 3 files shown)


yeounoh (changes requested):
This refactors the device type checking code -- which doesn't resolve the real issue cc @JackCaoG . Having said that, AMP should have worked as-is for SPMD, but not with SPMD (if we properly call autocast with xla device)

Hi @baoleai could you verify that you've compiled/installed `torch` with `USE_CUDA=0` and `torch_xla` with `XLA_CUDA=1`, and re-run your test code.

I will raise a PR today to enable `XLA:GPU` for autocast. I am blocking this PR, feel free to close.

PR #5527, opened 2023-08-31
[Pin Update] Version 20230826

alanwaketan:
Summary:
This change bumps libTPU.so to 20230826 and update the corresponding openxla/xla pin.

Test Plan:
CI.

PR #5570, opened 2023-09-14
Enable xla:gpu autocast for bfloat16 if not restricted

yeounoh:
`torch_xla.amp.autocast` was not using `XLA:GPU` but `torch.cuda` (GPU) backend. We should use `XLA` devices for XLA tensors, and for running `torch_xla.amp.autocast`.

Clarification 1:
> AMP executes on XLA:GPU, unless it's bfloat16 in which case, it has to fall back to cuda and requires USE_CUDA=1. We should address it.

This pytorch/pytorch#109302 to update the dispatcher comment is not needed. This also addresses #5497.

### DIFF
--- a/torch_xla/amp/autocast_mode.py
+++ b/torch_xla/amp/autocast_mode.py
@@ -6,29 +6,25 @@ import warnings
 
 class autocast(torch.amp.autocast_mode.autocast):
   r"""
-    See :class:`torch.autocast`.
-    ``torch_xla.amp.autocast(device, **kwargs)`` is equivalent to 
-    ``torch.autocast("xla", **kwargs)`` for TPUs
-    ``torch.autocast("cuda", **kwargs)`` for GPUs.
-    """
+  `torch.autocast` for XLA backend devices. See :class:`torch.autocast`.
+  ``torch_xla.amp.autocast(device, **kwargs)`` is equivalent to
+  ``torch.autocast("xla", **kwargs)`` for XLA:TPU and XLA:GPU backends.
+  """
 
   def __init__(self,
                device,
                enabled: bool = True,
                dtype: torch.dtype = None,
                cache_enabled: bool = True):
-    if xm.xla_device_hw(device) == 'GPU':
-      if dtype is None:
-        dtype = torch.float16
-      super().__init__(
-          "cuda", enabled=enabled, dtype=dtype, cache_enabled=cache_enabled)
-    elif xm.xla_device_hw(device) == 'TPU':
+    assert 'xla' in device.__str__(), "torch_xla.autocast is available for XLA:TPU, XLA:GPU"
+    xla_device = xm.xla_device_hw(device)
+    if xla_device in ['TPU', 'GPU']:
       if dtype is None:
         dtype = torch.bfloat16
       if dtype != torch.bfloat16:
-        error_message = "In TPU autocast, but the target dtype is not supported. Disabling autocast.\n"
+        error_message = "In torch_xla autocast, but the target dtype is not supported. Disabling autocast.\n"
         error_message += (
-            "TPU Autocast only supports dtype of torch.bfloat16 currently.")
+            "torch_xla Autocast only supports dtype of torch.bfloat16 currently.")
         warnings.warn(error_message)
         enabled = False
       super().__init__(
--- a/torch_xla/amp/autocast_mode.py
+++ b/torch_xla/amp/autocast_mode.py
@@ -16,7 +16,8 @@ class autocast(torch.amp.autocast_mode.autocast):
                enabled: bool = True,
                dtype: torch.dtype = None,
                cache_enabled: bool = True):
-    assert 'xla' in device.__str__(), "torch_xla.autocast is available for XLA:TPU, XLA:GPU"
+    assert 'xla' in device.__str__(
+    ), "torch_xla.autocast is available for XLA:TPU, XLA:GPU"
     xla_device = xm.xla_device_hw(device)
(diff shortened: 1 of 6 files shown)


JackCaoG:
there were some histories behind this. We used upstream autocase for cuda because we want to reuse the GPU AMP rules. Are you saying currently AMP does not execute on XLA:GPU when enabled?

JackCaoG:
we also had test for it https://github.com/pytorch/xla/blob/master/test/test_autocast.py

JackCaoG:
I am pretty sure, at least in 2.0 that our autocast actually worked with XLA:GPU and executed on XLA:GPU

yeounoh:
I will investigate why our tests didn't catch this issue and modify or add proper tests, before merging.

yeounoh:
> I am pretty sure, at least in 2.0 that our autocast actually worked with XLA:GPU and executed on XLA:GPU

Yea, it did. I think this issue was surfaced becasue of `bfloat16`, and we were hard-coding the dtype `float16` for GPU back then.

yeounoh:
> there were some histories behind this. We used upstream autocase for cuda because we want to reuse the GPU AMP rules. Are you saying currently AMP does not execute on XLA:GPU when enabled?

Got it, AMP executes on XLA:GPU, unless it's `bfloat16` in which case, it has to fall back to `cuda` and requires `USE_CUDA=1`. We should address it.

baoleai:
> Got it, AMP executes on XLA:GPU, unless it's `bfloat16` in which case, it has to fall back to `cuda` and requires `USE_CUDA=1`. We should address it.

How is this issue resolved?

yeounoh:
> > Got it, AMP executes on XLA:GPU, unless it's `bfloat16` in which case, it has to fall back to `cuda` and requires `USE_CUDA=1`. We should address it.
> 
> How is this issue resolved?

Hi @baoleai it was resolved, but making changes to our approach. Will let you know when the PR is ready.

yeounoh:
GPU test CI run timed out at the end.... increasing the timeout. I was able to pass all the GPU tests locally, on my GPU/TPU boxes. Will wait for the new CI run.

ManfeiBai:
> GPU test CI run timed out at the end.... increasing the timeout. I was able to pass all the GPU tests locally, on my GPU/TPU boxes. Will wait for the new CI run.

Hi, does the GPU CI test timed out at a test? didn't found the time out error in the last two commits' log

Pin update PR met similar issue and fixed by increased the timeout, do we want to try this too? https://github.com/pytorch/xla/pull/5527/commits/55a08239b0cbdd5daebd15efca0f978a9369a560

torch_xla/runtime.py
```diff
@@ -84,6 +84,16 @@ def wrapper(*args, **kwargs):
   return wrapper
 
 
+def is_bf16_supported():
+  """Returns whether torch.bfloat16 is supported on this environment.
+  """
+  try:
+    torch.tensor([1.], dtype=torch.bfloat16, device=xm.xla_device())
```
JackCaoG, line 91: lol this is smart

JackCaoG (approved):

yeounoh:
> > GPU test CI run timed out at the end.... increasing the timeout. I was able to pass all the GPU tests locally, on my GPU/TPU boxes. Will wait for the new CI run.
> 
> 
> 
> Hi, does the GPU CI test timed out at a test? didn't found the time out error in the last two commits' log
> 
> 
> 
> Pin update PR met similar issue and fixed by increased the timeout, do we want to try this too? https://github.com/pytorch/xla/pull/5527/commits/55a08239b0cbdd5daebd15efca0f978a9369a560

I increased timeout and was able to pass as well. I am going to address the gpu test time sometime soon.

torch_xla/amp/autocast_mode.py
```diff
@@ -1,48 +1,85 @@
...
+    if self._xla_device == 'GPU':
+      backend = 'cuda'
       if dtype is None:
         dtype = torch.float16
+      elif dtype == torch.bfloat16:
+        if xr.is_bf16_supported() and not torch.cuda.is_available():
+          # XLA:GPU with bfloat16 should run on `xla` backend
+          # unless torch.autocast is compiled with cuda.
+          backend = 'xla'
+        else:
```
baoleai, line 36: Why does not support bfloat16 now? Previously torch_xla was able to support bfloat16 with `torch.cuda.is_available = True`.  Do you mean we have to compile PyTorch with USE_CUDA=0 after this commit? And in most scenarios, our PyTorch needs to support both XLA and GPUs simultaneously, i.e. with USE_CUDA=1.

yeounoh: If torch is already compiled with `USE_CUDA=1` and `torch.cuda.is_available == True` then it would run with `cuda` backend and supports `bfloat16` as the cuda backend allows (note that for some platforms this wouldn't be the case). 

For the context, XLA:GPU, note that we guide users to compile `torch` with `USE_CUDA=0` to run ops with XLA, not torch cuda backend -- [gpu guide](https://github.com/pytorch/xla/blob/master/docs/gpu.md#develop-pytorchxla-on-a-gpu-instance-build-pytorchxla-from-source-with-gpu-support).

We've casted bfloat16 to float16/3 on XLA:GPU, but this should resolve it https://github.com/pytorch/xla/pull/5598

baoleai: > then it would run with cuda backend and supports bfloat16 as the cuda backend allows (note that for some platforms this wouldn't be the case).

But here the type is forced to torch.float16, even I set dtype=torch.bfloat16.

PR #5581, opened 2023-09-14
Improve bernoulli rng-bit-generation memory footprint

yeounoh:
Bernoulli rng-bit-generation is using f32 -> u32 precision, which is not needed when the range is small. We should be able to opt in for lower precision to save memory, when it's applicable. We don't compute the dynamic range on the fly to avoid/minimize computation.


torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -476,7 +476,8 @@ xla::XlaOp BuildBernoulli(xla::XlaOp probability, xla::XlaOp seed,
       xla::Zero(probability.builder(), probability_shape.element_type());
   xla::XlaOp one =
       xla::One(probability.builder(), probability_shape.element_type());
-  xla::XlaOp noise = RngUniform(seed, probability_shape, zero, one);
+  xla::XlaOp noise =
+      RngUniform(seed, probability_shape, zero, one, /*downcast=*/true);
```
JackCaoG, line 480: is there a reason we only downcast for `Bernoulli`. Would it be better if we always downcast?

yeounoh: Yea, I think so. For instance we could do for multinomial, since its also using it to create a random mask, and I don't think the extra precision matters. I guess we could try to extend the application when it becomes an issue or surface in the actual use case like Bernoulli, and it's easier to verify and test its benefit :)

JackCaoG (approved):

PR #5552, opened 2023-09-09
Move where clear pending IR is called to avoid crash

JackCaoG:
Without this patch
```
python benchmarks/dynamo/torchbench.py --randomize-input --performance --training --trace-on-xla --backend=openxla --only hf_Bert
```
will crash with
```
RuntimeError: ./torch_xla/csrc/runtime/pjrt_computation_client.h:168 : Check failed: HasValue() 
```
We expect `FallBackNodeCollector ` to introduce in place operation to `xla_args` so we want to replaced it with the cloned arg and clearing the pending IR. What I found is that `CapabilityBasedPartitioner` (or something close to that region of the code) will also introduce IRs. On top of that after the `Partitioner`, the `xla_args` passed to the `extract_internal` is not the same `xla_arg` passed to `extract_compiled_graph`. If we called `clearPendingIr` after `Partitioner` we might remove the pending IR of the copied `xla_arg` and we have no way to restore those values then dynamo will crash.

I chose to move the `clearpendingIr` to a earlier region of the code and use `mark_step` to turn them into device data. I have some concern of the correctness of this approach but I need more time to debug why `Partitioner` will introduce IRs to begin with.

PR #5582, opened 2023-09-15
Cherry-pick Move where clear pending IR is called to avoid crash (#5552)

JackCaoG:
* Move where clear pending IR is called to avoid crash

* fix CI

* fix CI and add some debugging messages


wonjoolee95 (approved):
LGTM.

PR #5593, opened 2023-09-15
Cache OpSharding by mesh and partition_spec

jonb377:
For large meshes, generating the OpSharding proto is expensive to generate and can increase tracing time. The value is easily cached for reuse across steps during tracing. 

This change also optimizes `UpdateShardingHash` to better handle a large number of devices and caches the return value for `xr.global_runtime_device_count`, both of which become more significant overheads with a greater number of devices.

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -847,6 +847,25 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertNotEqual(torch_xla._XLAC._get_xla_sharding_spec(xla_x), '')
     self.assertTrue(torch.allclose(xla_x.cpu(), x))
 
+  def test_op_sharding_cache(self):
+    met.clear_all()
+    mesh = self._get_mesh((1, self.n_devices))
+
+    t = torch.randn(1, self.n_devices).to(xm.xla_device())
+    xs.mark_sharding(t, mesh, (0, 1))
+    self.assertIn("CreateOpSharding", met.counter_names())
+    self.assertEqual(met.counter_value("CreateOpSharding"), 1)
+
+    # Sharding with the same partition spec should not result in another call
+    u = torch.randn(1, self.n_devices).to(xm.xla_device())
+    xs.mark_sharding(u, mesh, (0, 1))
+    self.assertEqual(met.counter_value("CreateOpSharding"), 1)
+
+    # Changing the partition spec will result in another CreateOpSharding
+    v = torch.randn(1, self.n_devices).to(xm.xla_device())
+    xs.mark_sharding(v, mesh, (0, None))
+    self.assertEqual(met.counter_value("CreateOpSharding"), 2)
+
 
 if __name__ == '__main__':
   test = unittest.main()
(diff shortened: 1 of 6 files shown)


torch_xla/experimental/xla_sharding.py
```diff
@@ -396,7 +404,12 @@ def _get_group_assignment(sharding_type: ShardingType,
     ]
     group_assignment = np.arange(len(replication_groups)).reshape(
         tuple(group_tile_shape)).tolist()
-  return group_assignment, replication_groups
+    _get_group_assignment.cache[cache_key] = totuple(
+        [group_assignment, replication_groups])
+  return tolist(_get_group_assignment.cache[cache_key])
```
jonb377, line 409: @yeounoh It might be cleaner to modify `_xla_mark_sharding`'s signature to accept `py::tuple` instead of converting back to list here, wdyt?

alanwaketan (approved):
LGTM.

alanwaketan:
Maybe we can have some sort of tests. One easy way to query the size of the cache?

jonb377:
> Maybe we can have some sort of tests. One easy way to query the size of the cache?

Good idea, I'll add some tests around this.

torch_xla/csrc/ir.cpp
```diff
@@ -199,10 +199,12 @@ void XlaNode::UpdateShardingHash() {
       sharding_hash_ = torch::lazy::HashCombine(
           sharding_hash_, (uint32_t)tile_assignment_dimension);
     }
-    for (const auto& tile_assignment_device :
```
alanwaketan, line 202: I guess this for loop is expansive as well.

torch_xla/experimental/xla_sharding.py
```diff
@@ -79,6 +81,27 @@ def get_axis_name_idx(self, name: str) -> int:
       return None
     return self.axis_names.index(name)
 
+  def get_op_sharding(self,
```
alanwaketan, line 84: Why don't you use `@functools.lru_cache()` like `global_runtime_device_count`

test/spmd/test_xla_sharding.py
```diff
@@ -847,6 +847,25 @@ def test_shard_device_data_ir_after_mark_step(self):
     self.assertNotEqual(torch_xla._XLAC._get_xla_sharding_spec(xla_x), '')
     self.assertTrue(torch.allclose(xla_x.cpu(), x))
 
+  def test_op_sharding_cache(self):
+    met.clear_all()
+    mesh = self._get_mesh((1, self.n_devices))
+
+    t = torch.randn(1, self.n_devices).to(xm.xla_device())
+    xs.mark_sharding(t, mesh, (0, 1))
+    self.assertIn("CreateOpSharding", met.counter_names())
```
alanwaketan, line 856: It's smart to use a counter here.

alanwaketan (approved):
LGTM.

torch_xla/experimental/xla_sharding.py
jonb377: I didn't think about `functools.lru_cache` until handling `global_runtime_device_count`, but I guess I reimplemented it for `get_op_sharding` lol! Let's use it here too.

jonb377: I set `maxsize=None` so that nothing gets evicted.

yeounoh (commented):
Post merge review, very good -- thanks you @jonb377 :)

PR #5594, opened 2023-09-15
Don't trigger CI build on release tag push

will-cromar:
The second run will fail because the first one already ran, e.g. `tag invalid: The image tag 'latest-1e6c11d1ff124ca79fa57e8de42333beb7895c8e' already exists in the 'pytorch/xla_base' repository and cannot be overwritten because the repository is immutable.`


ManfeiBai (approved):
Thanks! LGTM

PR #5606, opened 2023-09-18
Use XLA's new API to setup aliasing

JackCaoG:
XLA provided this new `AddBufferDonor` API to replace the old `SetUpAlias` one. The advantage of this new API is that it doesn't require the caller to explicitly match the computation input and output, which is hard to do in SPMD context since we don't know how compiler wants to shard the output. The `AddBufferDonor` promise that as long as user can tell which input buffer is not required after the execution, compiler can reuse this buffer for output.

TODO
1. add test
2. verify on SPMD

### DIFF
--- a/test/spmd/test_xla_sharding_hlo.py
+++ b/test/spmd/test_xla_sharding_hlo.py
@@ -10,6 +10,7 @@ import torch_xla
 import torch_xla.runtime as xr
 import torch_xla.core.xla_model as xm
 import torch_xla.experimental.xla_sharding as xs
+import torch_xla.debug.metrics as met
 
 import test_xla_sharding_base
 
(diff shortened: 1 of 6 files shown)


JackCaoG:
Ok I was able to resue the existing test, with the new API I saw
```
HloModule SyncTensorsGraph.4, is_scheduled=true, input_output_alias={ {0}: (0, {}, may-alias), {1}: (1, {}, may-alias) }, entry_computation_layout={(f32[4,2,2]{0,2,1:T(2,128)}, f32[4,2,2]{0,2,1:T(2,128)})->(f32[4,2,2]{0,2,1:T(2,128)}, f32[4,2,2]{0,2,1:T(2,128)})}

ENTRY SyncTensorsGraph.4 {
  p1.2 = f32[4,2,2]{0,2,1:T(2,128)} parameter(1)
  p0.1 = f32[4,2,2]{0,2,1:T(2,128)} parameter(0)
  ROOT tuple = (f32[4,2,2]{0,2,1:T(2,128)}, f32[4,2,2]{0,2,1:T(2,128)}) tuple(p0.1, p1.2)
}
```

i can confirm that `input_output_alias` works. Now I am checking the SPMD case.

JackCaoG:
hmm on small examples everything seems to work. However when I tried to run resnet with SPMD I don't see aliasing..

JackCaoG:
chatted with XLA folks internally, there is a bug that X64Rewritter will drop the buffer donor list. The fix is WIP so I will hold on this pr for now.

baoleai:
Let's take the xla/test//test_input_output_aliases.py as example:
After we add another in-place Op see https://github.com/baoleai/xla/blob/8dfa8d625353d555c0cb6d2a1f7dc59aa479e2db/test/test_input_output_aliases.py#L31 , the `InputOutputAliasCount` (https://github.com/baoleai/xla/blob/8dfa8d625353d555c0cb6d2a1f7dc59aa479e2db/test/test_input_output_aliases.py#L35) should not be changed ( it's still 4) if all in-place Ops can reuse buffer.

JackCaoG:
Relevant part of this part has landed.

Issue #5611, opened 2023-09-18
Support of op lowering for `hstack` and `vstack`.


zpcore:

zpcore:
Noticed that XLA op `stack` is not covered in the unit test. Will increase the test coverage also.

PR #5614, opened 2023-09-18
Support of vstack and hstack op lowering.

zpcore:
Solves #5611 

1. Add the op lowering support for `hstack` and `vstack`;
2. Add the code unit test for `stack`, `hstack` and `vstack`.


zpcore:
Noticed issues with the code logic.

PR #5617, opened 2023-09-19
Add cuda 12.1 support by adding cuda 12.1 trigger

vanbasten23:
Per [PyTorch's release page](https://github.com/pytorch/pytorch/blob/main/RELEASE.md#release-compatibility-matrix), PyTorch has added experimental support for `CUDA 12.1, CUDNN 8.9.2.26`. We should do the same.

### DIFF
--- a/infra/ansible/config/cuda_deps.yaml
+++ b/infra/ansible/config/cuda_deps.yaml
@@ -3,11 +3,13 @@
 cuda_deps:
   # List all libcudnn8 versions with `apt list -a libcudnn8`
   libcudnn:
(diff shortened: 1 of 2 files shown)


will-cromar (approved):
Thanks! Should we clean up old triggers to keep the number of builds down?

vanbasten23:
> Thanks! Should we clean up old triggers to keep the number of builds down?

Yes. We should do that. I plan to verify the new cuda 12.1 build. If everything looks fine, we can remove the old cuda 12.0 trigger since PyTorch doesn't support cuda 12.0.

vanbasten23:
Thanks for the review!

PR #5620, opened 2023-09-19
Add op lowering for hstack and vstack.

zpcore:
Solves #5611

### DIFF
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -1470,6 +1470,34 @@ TEST_F(AtenXlaTensorTest, TestStack) {
   }
 }
 
+TEST_F(AtenXlaTensorTest, TestHstack) {
+  torch::Tensor a = torch::rand({2, 4, 3}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor b = torch::rand({2, 4, 3}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor c = torch::rand({2, 4, 3}, torch::TensorOptions(torch::kFloat));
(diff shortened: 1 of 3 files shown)


zpcore:
Originally I tried with codegen, but it looks like `vstack` and `hstack` are not supported since I saw no file updates after I added those ops under full_codegen section in xla_native_functions.yaml.

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -1470,6 +1470,34 @@ TEST_F(AtenXlaTensorTest, TestStack) {
   }
 }
 
+TEST_F(AtenXlaTensorTest, TestHstack) {
```
lsy323, line 1473: Can we also check the counter in the test, so we can make sure the op is lowered instead falling back to cpu

lsy323:
> Originally I tried with codegen, but it looks like `vstack` and `hstack` are not supported since I saw no file updates after I added those ops under full_codegen section in xla_native_functions.yaml.

Weird, I saw hstack/vstack is in `native_functions.yaml` at https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L5517

test/cpp/test_aten_xla_tensor_1.cpp
zpcore: Good point. I was thinking the XLA op will automatically throws out the error if 1) the op lowering is not implemented and also 2) the input tensor is copied to the device. Looks like this is not the case. Thanks

lsy323:
hstack/vstack is calling `cat` before dispatching happens. So lowering `cat` is enough for hstack/vstack

PR #5619, opened 2023-09-19
Fix log spam when libtpu is loaded

will-cromar:
Swap absl logging for TSL logging, since we don't initialize absl.

Currently, every process prints this warning:

```
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1692408336.588809 3320398 tpu_initializer_framework_helper.cc:78] Libtpu path is: /usr/local/lib/python3.8/site-packages/libtpu/libtpu.so
I0000 00:00:1692408374.448069 3320398 pjrt_c_api_client.cc:110] PjRtCApiClient created.
```

With this PR, those logs are visible only if you enabled verbose logging:

```
# python rand.py 
tensor([[-0.5182, -0.7280, -0.1179],
        [-0.7625,  1.7006, -0.2758],
        [-1.1760, -1.6168, -0.9752]], device='xla:0')
# TF_CPP_MIN_LOG_LEVEL=0 python rand.py 
2023-09-19 19:04:55.129028: I external/xla/xla/pjrt/pjrt_api.cc:98] GetPjrtApi was found for tpu at /usr/local/lib/python3.8/site-packages/libtpu/libtpu.so
2023-09-19 19:04:55.129101: I external/xla/xla/pjrt/pjrt_api.cc:67] PJRT_Api is set for device type tpu
2023-09-19 19:04:55.129109: I external/xla/xla/pjrt/pjrt_api.cc:72] PJRT plugin for tpu has PJRT API version 0.23. The framework PJRT API version is 0.23.
2023-09-19 19:04:58.376339: I external/xla/xla/pjrt/pjrt_c_api_client.cc:116] PjRtCApiClient created.
tensor([[-0.5182, -0.7280, -0.1179],
        [-0.7625,  1.7006, -0.2758],
        [-1.1760, -1.6168, -0.9752]], device='xla:0')
```

I will propose this fix in upstream OpenXLA, then we can remove this patch after the next pin update.

PR #5621, opened 2023-09-19
Fix log spam when libtpu is loaded (#5619)

ManfeiBai:
Cherry-pick #5619

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -42,6 +42,11 @@ http_archive(
         "//openxla_patches:f16_abi_clang.diff",
         "//openxla_patches:gpu_race_condition.diff",
         "//openxla_patches:constexpr_return.diff",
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -42,6 +42,11 @@ http_archive(
         "//openxla_patches:f16_abi_clang.diff",
         "//openxla_patches:gpu_race_condition.diff",
         "//openxla_patches:constexpr_return.diff",
+<<<<<<< HEAD
```
will-cromar, line 45: Please fix this conflict

will-cromar (approved):
Thanks!

WORKSPACE
ManfeiBai: Thanks, updated

PR #5623, opened 2023-09-20
Support persistent computation caching to disk

jonb377:
Allow saving computations to disk for reuse across executions. Only supports SPMD-mode for now. To use, configure the environment variable `XLA_PERSISTENT_CACHE_PATH` to the correct path for the persistent cache to live.

The persistent cache does *not* use LRU to evict compilation results from disk.

This change depends on https://github.com/pytorch/xla/pull/5633 to stabilize the hash for `Constant` ops.

### DIFF
--- a/torch_xla/csrc/runtime/cache.h
+++ b/torch_xla/csrc/runtime/cache.h
@@ -128,11 +128,12 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
   using TypePtr = std::shared_ptr<T>;
 
   explicit PersistentCache(
-      int kMaxCacheSize, std::string cache_dir,
+      int kMaxCacheSize, std::string cache_dir, bool readonly,
       std::function<void(TypePtr&, std::ostream&)> serialize,
       std::function<TypePtr(std::istream&)> deserialize)
       : memcache_(kMaxCacheSize),
         cache_dir_(cache_dir),
+        readonly_(readonly),
         serialize_(serialize),
         deserialize_(deserialize) {
     std::filesystem::create_directories(cache_dir);
@@ -140,7 +141,7 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
 
   TypePtr Add(K key, TypePtr obj) override {
     std::string path = GetPath(key);
-    if (!Exists(path)) {
+    if (!Exists(path) && !readonly_) {
       std::ofstream out(path, std::ios::binary);
       serialize_(obj, out);
     }
@@ -175,7 +176,7 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
 
   bool Erase(const K& key) override {
     memcache_.Erase(key);
-    return std::remove(GetPath(key).c_str());
+    return !readonly_ && std::remove(GetPath(key).c_str());
   }
 
  private:
@@ -194,6 +195,7 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
   std::function<void(TypePtr&, std::ostream&)> serialize_;
   std::function<TypePtr(std::istream&)> deserialize_;
   std::string cache_dir_;
+  bool readonly_;
 };
 
 }  // namespace util
--- a/torch_xla/csrc/runtime/cache.h
+++ b/torch_xla/csrc/runtime/cache.h
@@ -2,6 +2,7 @@
 #define XLA_CLIENT_CACHE_H_
 
 #include <sys/stat.h>
+#include <torch/csrc/lazy/core/metrics.h>
 
 #include <filesystem>
 #include <fstream>
--- a/torch_xla/csrc/runtime/cache.h
+++ b/torch_xla/csrc/runtime/cache.h
@@ -1,10 +1,15 @@
 #ifndef XLA_CLIENT_CACHE_H_
 #define XLA_CLIENT_CACHE_H_
 
+#include <sys/stat.h>
+
+#include <filesystem>
+#include <fstream>
 #include <functional>
 #include <list>
 #include <memory>
 #include <mutex>
+#include <sstream>
 #include <unordered_map>
 #include <utility>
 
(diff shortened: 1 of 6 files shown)


torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -529,6 +529,45 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
...
+bool PjRtComputationClient::SerializeComputation(ComputationPtr computation,
+                                                 std::ostream& out) {
+  XLA_CHECK(computation->devices().size() == 1 &&
+            computation->devices()[0] == "SPMD:0")
+      << "Persistent compilation cache only supports SPMD";
+  const PjRtComputation& pjrt_computation =
+      dynamic_cast<const PjRtComputation&>(*computation);
+
+  auto executable = pjrt_computation.executable->SerializeExecutable();
+  if (!executable.ok()) return false;
```
will-cromar, line 541: Is it useful to log here if serialization fails?

Same for other cases where you check `ok`

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -468,10 +468,37 @@ void XLAGraphExecutor::MaybeDumpGraph(std::string name,
...
 }
 
-XLAGraphExecutor::ComputationCache* XLAGraphExecutor::GetComputationCache() {
+XLAGraphExecutor::ComputationCache* createComputationCache() {
   static const size_t kMaxCacheSize =
       runtime::sys_util::GetEnvInt("XLA_COMPILATION_CACHE_SIZE", 1024);
-  static ComputationCache* cache = new ComputationCache(kMaxCacheSize);
+  static std::string persistentCacheDir = runtime::sys_util::GetEnvString(
+      "XLA_PERSISTENT_COMPILATION_CACHE_PATH", "");
+  if (persistentCacheDir != "") {
```
will-cromar, line 476: nit: I suggest [`std::string::empty`](https://cplusplus.com/reference/string/string/empty/)

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -468,10 +468,37 @@ void XLAGraphExecutor::MaybeDumpGraph(std::string name,
...
+      auto client_computation =
+          runtime::GetComputationClient()->DeserializeComputation(istream);
+      if (!client_computation) return nullptr;
+      auto computation =
+          std::make_shared<torch_xla::Computation>(client_computation);
+      return std::make_shared<XLAGraphExecutor::CachedComputation>(
+          computation, /*is_sharded=*/true);
+    };
+    /* In a distributed SPMD context, only one worker should write. */
+    bool readonly = runtime::GetComputationClient()->GetProcessIndex() != 0;
```
will-cromar, line 492: I'm not sure this is right in general. It's easy for an ordinal to sneak into the graph as a constant with multiprocessing, particularly when constructing a mask for e.g. pinned all_gather. I believe each replica will need its own cache.

What if you always write to `$XLA_PERSISTENT_COMPILATION_CACHE_PATH/$ORDINAL`? This should have no impact on the SPMD case. You'll just write to `$XLA_PERSISTENT_COMPILATION_CACHE_PATH/0`.

will-cromar (commented):
Wow! Thank you so much for taking a shot at this.

Is it possible to add any sort of test? The coverage in `torch_xla::runtime` is a nightmare, so I'm not going to make you get 100% coverage, but I'd feel a lot more comfortable with a basic test case.

torch_xla/csrc/xla_graph_executor.cpp
jonb377: Oh nice! TIL a new C++ API

torch_xla/csrc/xla_graph_executor.cpp
jonb377: This PR only supports SPMD, so we'll need to follow up with MP. This is an optimization for a shared mount, since thousands of workers writing to the cache at once could cause some issues.

In MP, would we need a different cache for each worker? Any difference in the graph should be reflected in the hash, right?

torch_xla/csrc/runtime/pjrt_computation_client.cc
jonb377: Oh good point, I didn't immediately remember how to use VLOG so I took the coward's way out (no logging) 😂

torch_xla/csrc/runtime/cache.h
```diff
@@ -106,6 +122,83 @@ class Cache {
...
+    ss << cache_dir_ << "/" << key << ".bin";
+    return ss.str();
+  }
+
+  bool Exists(std::string path) {
+    struct stat buffer;
+    return stat(path.c_str(), &buffer) == 0;
+  }
+
+  Cache<K, T, H, E> memcache_;
```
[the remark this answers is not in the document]
jonb377: Haha good point, how about `memory_cache_`? I just want the name to reflect that this cache layer isn't persistent.

torch_xla/csrc/xla_graph_executor.cpp
jonb377: I'll make `readonly` controlled by a flag. It won't always be a shared mount even for SPMD.

alanwaketan:
Is this ready for review? Or still a draft?

jonb377:
> Is this ready for review? Or still a draft?

@alanwaketan I don't plan to merge until the hash takes the compilation environment (e.g. XLA flags, openxla version, hardware type) into account, so there's still some changes to come. But the functionality in the PR is pretty stable if you want to take a pass.

No rush though, I'll mark the PR ready when everything is in place.

torch_xla/csrc/xla_graph_executor.cpp
jonb377: I updated to read `readonly` from an environment variable.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -468,10 +468,35 @@ void XLAGraphExecutor::MaybeDumpGraph(std::string name,
   }
 }
 
-XLAGraphExecutor::ComputationCache* XLAGraphExecutor::GetComputationCache() {
+XLAGraphExecutor::ComputationCache* createComputationCache() {
```
yeounoh, line 471: nit. put this in a unnamed namesmape at the top, if not exposed via the 
 header file.

yeounoh: Also, should we call it `CreateComputationCache`?

torch_xla/csrc/runtime/cache.h
```diff
@@ -106,6 +122,83 @@ class Cache {
...
+      return nullptr;
+    }
+    std::ifstream in(path, std::ios::binary);
+    TypePtr val = deserialize_(in);
+    if (!val) {
+      TORCH_LAZY_COUNTER("PersistentCacheDeserializeFailure", 1);
+      return nullptr;
+    }
+    TORCH_LAZY_COUNTER("PersistentCacheHit", 1);
+    // Make sure the memory_cache_ tracks the value to prevent multiple loads
```
yeounoh, line 169: +1

jonb377:
I will split this into smaller PRs

PR #5627, opened 2023-09-20
Update the amp.md according to the latest changes

yeounoh:
Update the amp.md according to the latest changes, to prevent user from crashing for a certain use case (using `torch.cuda.autocast` directly with XLA:GPU).


JackCaoG (approved):

yeounoh:
Merging as it only updates the doc.

PR #5628, opened 2023-09-20
Create cuda 12.1 dev image trigger.

vanbasten23:
Need to test multihost local change on a newer cuda version than 11.8, since we support cuda 12.1 now.

Without this trigger, it's very tricky/non-trivial to set up all dependencies and environment variable correctly in order to build torch_xla.


JackCaoG (approved):

vanbasten23:
Thanks for the review!

PR #5630, opened 2023-09-20
Add API to assemble CPU shards to a sharded tensor

jonb377:
Currently, to convert CPU shards to an XLAShardedTensor, the API `XLAShardedTensor::load_local_shards_` must be used. This loads CPU shards in-place to an existing sharded tensor. A more convenient method for use cases outside of distributed checkpointing is to directly assemble the shards into a global tensor on device.

This PR adds a private _XLAC API `_get_global_tensor_from_cpu_shards` to allow directly creating a sharded tensor from a list of CPU shards and an OpSharding generated by `Mesh::get_op_sharding`.

I currently plan to use this new API in a few places:
 - tpu.py::discover_master_worker_ip_address: place each host's IP into a global tensor and pull out the zeroth entry for worker 0. The existing IP discovery API currently doesn't work with SPMD, but it's needed for distributed checkpointing.
 - Distributed data loading: A per-device, sharding-aware dataloader can use this API to efficiently create device data from the loaded CPU shards.

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -881,6 +881,12 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     with self.assertRaises(RuntimeError):
       XLAShardedTensor.from_cpu_shards(shards[:-1], op_sharding)
 
+    # Test replicated sharding. The result should equal a shard.
+    op_sharding = mesh.get_op_sharding((None,))
+    shards = [torch.LongTensor([0])] * self.n_devices
+    global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding)
+    self.assertTrue(torch.allclose(global_tensor.cpu(), shards[0]))
+
     # Test an invalid global shape - too many values.
     with self.assertRaises(RuntimeError):
       bad_size = torch.Size((2 * self.n_devices,))
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -866,6 +866,65 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     xs.mark_sharding(v, mesh, (0, None))
     self.assertEqual(met.counter_value("CreateOpSharding"), 2)
 
+  def test_from_cpu_shards(self):
+    # Create an OpSharding with all devices on a single axis
+    mesh = self._get_mesh((self.n_devices,))
+    op_sharding = mesh.get_op_sharding((0,))
+
+    # Infer the global shape from the OpSharding
+    shards = [torch.LongTensor([i]) for i in range(self.n_devices)]
+    global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding)
+    self.assertTrue(
+        torch.allclose(global_tensor.cpu(), torch.arange(self.n_devices)))
+
+    # Test incorrect number of shards
+    with self.assertRaises(RuntimeError):
+      XLAShardedTensor.from_cpu_shards(shards[:-1], op_sharding)
+
+    # Test an invalid global shape - too many values.
+    with self.assertRaises(RuntimeError):
+      bad_size = torch.Size((2 * self.n_devices,))
+      XLAShardedTensor.from_cpu_shards(shards, op_sharding, bad_size)
+
+    if self.n_devices > 1:
+      # Test an invalid global shape - incorrect rank
+      with self.assertRaises(RuntimeError):
+        bad_size = torch.Size((self.n_devices // 2, 2))
+        XLAShardedTensor.from_cpu_shards(shards, op_sharding, bad_size)
+
+      # Test a valid global shape - restrict the number of meaningful values
+      # to 1, treating the rest as padding.
+      global_shape = torch.Size((1,))
+      global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding,
+                                                       global_shape)
+      self.assertTrue(torch.allclose(global_tensor.cpu(), torch.arange(1)))
+
+      # Use a 2d mesh
+      mesh_2d = self._get_mesh((self.n_devices // 2, 2))
+      shards = [torch.LongTensor([[i]]) for i in range(self.n_devices)]
+      op_sharding = mesh_2d.get_op_sharding((0, 1))
+      global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding)
+      expected = torch.arange(self.n_devices).reshape(2, self.n_devices // 2)
+      self.assertTrue(torch.allclose(global_tensor.cpu(), expected))
+
+      # Use partial replication. With eight devices, the mesh and shards are:
+      #       mesh_2d            shards
+      #  | TPU:0 | TPU:1 |     | 0 | 1 |
+      #  | TPU:2 | TPU:3 |     | 0 | 1 |
+      #  | TPU:4 | TPU:5 |     | 0 | 1 |
+      #  | TPU:6 | TPU:7 |     | 0 | 1 |
+      #
+      # Partial replication along the 0th axis represents a global tensor
+      # of torch.Tensor([[0, 1]]).
+      shards = [torch.LongTensor([[i]]) for i in range(2)] * (
+          self.n_devices // 2)
+      op_sharding = mesh_2d.get_op_sharding((None, 1))
+      global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding)
+      expected = torch.arange(self.n_devices)
+      self.assertTrue(
+          torch.allclose(global_tensor.cpu(),
+                         torch.arange(2).reshape(1, 2)))
+
 
 if __name__ == '__main__':
   test = unittest.main()
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -866,70 +866,138 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     xs.mark_sharding(v, mesh, (0, None))
     self.assertEqual(met.counter_value("CreateOpSharding"), 2)
 
-  def test_from_cpu_shards(self):
+  def test_from_cpu_shards_replicated(self):
+    from_cpu_shards = torch_xla._XLAC._global_tensor_from_cpu_shards
+
     # Create an OpSharding with all devices on a single axis
     mesh = self._get_mesh((self.n_devices,))
-    op_sharding = mesh.get_op_sharding((0,))
+    partition_spec = (None,)
+    op_sharding = mesh.get_op_sharding(partition_spec)
+    shards = [torch.arange(4)] * self.n_devices
 
-    # Infer the global shape from the OpSharding
+    # No shape should result in the shape of a single shard.
+    global_tensor = from_cpu_shards(shards, op_sharding)
+    self.assertTrue(torch.allclose(global_tensor.cpu(), shards[0]))
+
+    # Specify a valid shape for the global tensor
+    global_tensor = from_cpu_shards(shards, op_sharding, shards[0].shape)
+    self.assertTrue(torch.allclose(global_tensor.cpu(), shards[0]))
+
+    # All invalid shapes should raise
+    with self.assertRaises(RuntimeError):
+      from_cpu_shards(shards, op_sharding, torch.Size((5,)))
+    with self.assertRaises(RuntimeError):
+      from_cpu_shards(shards, op_sharding, torch.Size((3,)))
+    with self.assertRaises(RuntimeError):
+      from_cpu_shards(shards, op_sharding, torch.Size((2, 2)))
+
+  def test_from_cpu_shards_tiled(self):
+    from_cpu_shards = torch_xla._XLAC._global_tensor_from_cpu_shards
+
+    # Create an OpSharding with all devices on a single axis
+    mesh = self._get_mesh((self.n_devices,))
+    partition_spec = (0,)
+    op_sharding = mesh.get_op_sharding(partition_spec)
     shards = [torch.LongTensor([i]) for i in range(self.n_devices)]
-    global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding)
+
+    global_tensor = from_cpu_shards(shards, op_sharding)
     self.assertTrue(
         torch.allclose(global_tensor.cpu(), torch.arange(self.n_devices)))
 
     # Test incorrect number of shards
     with self.assertRaises(RuntimeError):
-      XLAShardedTensor.from_cpu_shards(shards[:-1], op_sharding)
-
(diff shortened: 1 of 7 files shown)


torch_xla/runtime.py
```diff
@@ -131,6 +131,12 @@ def global_device_count() -> int:
   return len(torch_xla._XLAC._xla_get_all_devices())
 
 
+@requires_pjrt
+def local_runtime_devices() -> List[str]:
```
jonb377, line 135: @JackCaoG What do you think about this API? I wanted to include it so there's some way to discover the local device ID ordering.

test/spmd/test_xla_sharding.py
```diff
@@ -866,6 +866,71 @@ def test_op_sharding_cache(self):
     xs.mark_sharding(v, mesh, (0, None))
     self.assertEqual(met.counter_value("CreateOpSharding"), 2)
 
+  def test_from_cpu_shards(self):
+    # Create an OpSharding with all devices on a single axis
+    mesh = self._get_mesh((self.n_devices,))
+    op_sharding = mesh.get_op_sharding((0,))
```
yeounoh, line 872: Can we break this down to 
```
partition_spec = (0,)
op_sharding = mesh.get_op_sharding(partition_spec)
```
for readability?

test/spmd/test_xla_sharding.py
```diff
@@ -866,6 +866,71 @@ def test_op_sharding_cache(self):
...
+    shards = [torch.LongTensor([i]) for i in range(self.n_devices)]
+    global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding)
+    self.assertTrue(
+        torch.allclose(global_tensor.cpu(), torch.arange(self.n_devices)))
+
+    # Test incorrect number of shards
+    with self.assertRaises(RuntimeError):
+      XLAShardedTensor.from_cpu_shards(shards[:-1], op_sharding)
+
+    # Test replicated sharding. The result should equal a shard.
```
yeounoh, line 884: nit, "equal to a shard." Handling replication transparently to the frontend is great. Thanks!

test/spmd/test_xla_sharding.py
```diff
@@ -866,6 +866,71 @@ def test_op_sharding_cache(self):
...
+
+    # Test an invalid global shape - too many values.
+    with self.assertRaises(RuntimeError):
+      bad_size = torch.Size((2 * self.n_devices,))
+      XLAShardedTensor.from_cpu_shards(shards, op_sharding, bad_size)
+
+    if self.n_devices > 1:
+      # Test an invalid global shape - incorrect rank
+      with self.assertRaises(RuntimeError):
+        bad_size = torch.Size((self.n_devices // 2, 2))
```
yeounoh, line 898: Does it work with 1 device case?

test/spmd/test_xla_sharding.py
```diff
@@ -866,6 +866,71 @@ def test_op_sharding_cache(self):
...
+    self.assertTrue(
+        torch.allclose(global_tensor.cpu(), torch.arange(self.n_devices)))
+
+    # Test incorrect number of shards
+    with self.assertRaises(RuntimeError):
+      XLAShardedTensor.from_cpu_shards(shards[:-1], op_sharding)
+
+    # Test replicated sharding. The result should equal a shard.
+    op_sharding = mesh.get_op_sharding((None,))
+    shards = [torch.LongTensor([0])] * self.n_devices
```
yeounoh, line 886: I think we should emphasize/comment that `from_cpu_shards` won't move the shards, but copy and the shards are detached.

test/spmd/test_xla_sharding.py
```diff
@@ -866,6 +866,71 @@ def test_op_sharding_cache(self):
...
+      XLAShardedTensor.from_cpu_shards(shards, op_sharding, bad_size)
+
+    if self.n_devices > 1:
+      # Test an invalid global shape - incorrect rank
+      with self.assertRaises(RuntimeError):
+        bad_size = torch.Size((self.n_devices // 2, 2))
+        XLAShardedTensor.from_cpu_shards(shards, op_sharding, bad_size)
+
+      # Test a valid global shape - restrict the number of meaningful values
+      # to 1, treating the rest as padding.
```
yeounoh, line 902: Is it always the case, or due to the replicated op sharding.

test/spmd/test_xla_sharding.py
```diff
@@ -866,6 +866,71 @@ def test_op_sharding_cache(self):
...
+    with self.assertRaises(RuntimeError):
+      XLAShardedTensor.from_cpu_shards(shards[:-1], op_sharding)
+
+    # Test replicated sharding. The result should equal a shard.
+    op_sharding = mesh.get_op_sharding((None,))
+    shards = [torch.LongTensor([0])] * self.n_devices
+    global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding)
+    self.assertTrue(torch.allclose(global_tensor.cpu(), shards[0]))
+
+    # Test an invalid global shape - too many values.
```
yeounoh, line 890: What happens if the global shape is smaller, do we treat the rest as padding or error?

test/spmd/test_xla_sharding.py
```diff
@@ -866,6 +866,71 @@ def test_op_sharding_cache(self):
...
+
+      # Test a valid global shape - restrict the number of meaningful values
+      # to 1, treating the rest as padding.
+      global_shape = torch.Size((1,))
+      global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding,
+                                                       global_shape)
+      self.assertTrue(torch.allclose(global_tensor.cpu(), torch.arange(1)))
+
+      # Use a 2d mesh
+      mesh_2d = self._get_mesh((self.n_devices // 2, 2))
```
yeounoh, line 909: Would this work with 1 device case?

test/spmd/test_xla_sharding.py
```diff
@@ -866,6 +866,71 @@ def test_op_sharding_cache(self):
...
+      #  | TPU:4 | TPU:5 |     | 0 | 1 |
+      #  | TPU:6 | TPU:7 |     | 0 | 1 |
+      #
+      # Partial replication along the 0th axis represents a global tensor
+      # of torch.Tensor([[0, 1]]).
+      shards = [torch.LongTensor([[i]]) for i in range(2)] * (
+          self.n_devices // 2)
+      op_sharding = mesh_2d.get_op_sharding((None, 1))
+      global_tensor = XLAShardedTensor.from_cpu_shards(shards, op_sharding)
+      expected = torch.arange(self.n_devices)
```
yeounoh, line 929: this is not used, but `torch.arange(2).reshape(1, 2)`?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1660,6 +1660,66 @@ void InitXlaModuleBindings(py::module m) {
...
+            << "Please enable SPMD via `torch_xla.runtime.use_spmd()`";
+        auto local_devices = runtime::GetComputationClient()->GetLocalDevices();
+        XLA_CHECK(local_devices.size() == shards.size())
+            << "Must specify a shard for each local device";
+
+        if (!global_shape.has_value()) {
+          // Set a default value for the global shape based on the sharding
+          // type.
+          if (sharding.type() == xla::OpSharding::OTHER) {
+            // Infer the global shape to be the shard shape scaled by the tiling
```
yeounoh, line 1681: This may not hold true, in case of uneven tiling. Let's make a note on this.

yeounoh: or, actually for the right most shard in each dim, we can add the sizes, as the padding is always on the last dims.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1660,6 +1660,66 @@ void InitXlaModuleBindings(py::module m) {
...
+            // dimensionality.
+            auto tile_shape = sharding.tile_assignment_dimensions();
+            global_shape = std::vector<long int>();
+            for (int dim = 0; dim < shards[0].sizes().size(); ++dim) {
+              auto global_dim = tile_shape[dim] * shards[0].sizes()[dim];
+              global_shape->push_back(global_dim);
+            }
+          } else {
+            // If the sharding type is not tiled, the tensor shards are
```
yeounoh, line 1690: This may not hold true in the future, can you check for the replicated type, and have a catch-all block for raising error?

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Ok we can't do that, since `"Input shard shape must include padding: " << shard.sizes()"`

torch_xla/experimental/xla_sharding.py
```diff
@@ -87,6 +87,14 @@ def get_op_sharding(self,
     Return the OpSharding for the given partition spec. This is an expensive
     operation as the mesh grows, so the value is cached for reuse.
     """
+    partition_spec = _translate_named_partition_spec(self, partition_spec)
```
yeounoh, line 90: Thanks for refactoring 👍

torch_xla/runtime.py
yeounoh: I think this could be handy.

yeounoh (commented):
Let some comments.

test/spmd/test_xla_sharding.py
jonb377: Good call, I can make this a single-device test.

test/spmd/test_xla_sharding.py
jonb377: That's a good point. Do you think adding that to the docstring is sufficient?

test/spmd/test_xla_sharding.py
jonb377: This is always the case, I'll add another test for tiled sharding.

test/spmd/test_xla_sharding.py
jonb377: Thanks! I'll update the assertion to use `expected`.

torch_xla/csrc/init_python_bindings.cpp
jonb377: Let me know if I should revisit this. I decided on this approach because padding can cross over multiple devices, e.g. sharding a tensor with shape `(1, 2)` on the mesh `(1, 4)` will have shards with shape `(1, 1)` with no real data on the last two devices.

test/spmd/test_xla_sharding.py
jonb377: Oh I see what you mean, by making the mesh `(1, self.n_devices)`? I wanted this test to have multiple tile dims to cover the inferred global shape. I can make these tests run single-device as well.

test/spmd/test_xla_sharding.py
jonb377: It will infer padding up to a point, but if it's too small it will also fail the assertion. I'll add another test case for that.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1660,6 +1660,66 @@ void InitXlaModuleBindings(py::module m) {
...
           return std::nullopt;
         });
+  // Reassemble the CPU shards into a global tensor. A new sharded tensor is
+  // created from the local shards with the provided sharding annotation
+  // attached. The order of the shards should coincide with the order of
+  // devices returned by `torch_xla.runtime.local_runtime_devices()`.
+  m.def(
+      "_global_tensor_from_cpu_shards",
+      [](const std::vector<at::Tensor>& shards, const xla::OpSharding& sharding,
+         std::optional<std::vector<long int>>& global_shape) -> at::Tensor {
```
alanwaketan, line 1670: Can we use int64_t here?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1660,6 +1660,66 @@ void InitXlaModuleBindings(py::module m) {
...
+              global_shape->push_back(global_dim);
+            }
+          } else {
+            // If the sharding type is not tiled, the tensor shards are
+            // replicated.
+            global_shape = shards[0].sizes().vec();
+          }
+        }
+
+        xla::Shape tensor_shape =
```
alanwaketan, line 1696: Is this supposed to be global shape? Then why use shards[0]?

torch_xla/experimental/xla_sharded_tensor.py
```diff
@@ -102,6 +102,19 @@ def __new__(cls, elem: torch.Tensor, *args, **kwargs):
     r.global_tensor = elem.detach() if r.requires_grad else elem
     return r
 
+  @staticmethod
+  def from_cpu_shards(shards: List[torch.Tensor],
+                      sharding: torch_xla._XLAC.OpSharding,
+                      global_shape: torch.Size = None):
```
alanwaketan, line 108: What's the benefit of providing a global_shape?

torch_xla/csrc/init_python_bindings.cpp
jonb377: This is a bit of a shortcut to use `CreateComputationShapeFromTensor`, the shape's dimensions are updated in the next few lines. I'll actually just switch to the lower-level API `MakeArrayShapeFromDimensions` instead, that will be easier to follow.

torch_xla/csrc/init_python_bindings.cpp
jonb377: Good point, I'll change this.

torch_xla/experimental/xla_sharded_tensor.py
jonb377: If the shards are padded, the global_shape can be used to ignore padding.

torch_xla/runtime.py
jonb377: I'm removing this, we'll keep the new `_global_tensor_from_cpu_shards` private.

torch_xla/csrc/init_python_bindings.cpp
alanwaketan: ping @yeounoh

yeounoh (approved):
LGTM

torch_xla/csrc/init_python_bindings.cpp
yeounoh: I think for the most part, user should specify the global shape. I am thinking that we should actually make it explicit, not optional... but if not, this way of handling, inferring the default shape is good.

jonb377: That's a good point... I chose to let it be inferred for convenience, since for e.g. distributed data loading, the shards are provided on CPU with the correct padded local shape, and deriving the global shape could be difficult for more sophisticated shardings.

I'll go ahead and land with it optional for now, since we're keeping the API private. Thanks Yeounoh!

jonb377:
Thanks @yeounoh and @alanwaketan for the review. I'll merge after a TPU CI run.

alanwaketan (approved):
LGTM.

PR #5633, opened 2023-09-20
Stabilize Constant hash across processes

jonb377:
The hash for `Constant` ops currently depends on the [AbslHash implementation](https://github.com/openxla/xla/blob/46baaafb19d6819d01b3f91be78b5a1e8cc9e14f/xla/literal.h#L323-L341) for XLA literals. AbslHash intentionally randomizes the hash seed on every process, so this breaks the persistent compilation caching in https://github.com/pytorch/xla/pull/5623.

This change ports the AbslHash logic to use torch::lazy::hash to ensure the hash is consistent across executions.

### DIFF
--- a/torch_xla/csrc/ops/constant.cpp
+++ b/torch_xla/csrc/ops/constant.cpp
@@ -4,12 +4,16 @@
 #include <sstream>
 
 #include "torch_xla/csrc/lowering_context.h"
+#include "torch_xla/csrc/runtime/xla_util.h"
+#include "torch_xla/csrc/torch_util.h"
+#include "xla/layout_util.h"
+#include "xla/shape_util.h"
 
 namespace torch_xla {
 
 Constant::Constant(xla::Literal value)
     : XlaNode(torch::lazy::OpKind(at::prim::Constant), value.shape(),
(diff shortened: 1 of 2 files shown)


JackCaoG (approved):
Thanks! I am trying to see what's a good way to test this.. One idea is to use API in https://github.com/pytorch/xla/blob/e14bdc80224a162b98ce5fa50f5ab08251dec37d/torch_xla/csrc/init_python_bindings.cpp#L1956C10-L1965 to dump the API of a `arange`, but then I can't think of an easy way to run the test twice and then collect the hash..


I guess it is OK but when we implement the persisent cache we should has more rigid unit test to make sure hash changed/not changed under different cases.

jonb377:
> Thanks! I am trying to see what's a good way to test this.. One idea is to use API in https://github.com/pytorch/xla/blob/e14bdc80224a162b98ce5fa50f5ab08251dec37d/torch_xla/csrc/init_python_bindings.cpp#L1956C10-L1965 to dump the API of a `arange`, but then I can't think of an easy way to run the test twice and then collect the hash..
> 
> I guess it is OK but when we implement the persisent cache we should has more rigid unit test to make sure hash changed/not changed under different cases.

I was thinking about the same, I can use `Popen` to run in multiple processes within a single Python script.

PR #5636, opened 2023-09-21
Fix bug for low-precision ranges

jonb377:
When creating a range on the XLA device using low-precision floating types like bf16, the process will hang. For example, this will hang:

```
$ python -c 'import torch, torch_xla; print(torch.arange(255, 260, dtype=torch.bfloat16, device="xla"))'
# Hangs
```

Meanwhile, upstream torch executes successfully:

```
$ python -c 'import torch, torch_xla; print(torch.arange(255, 260, dtype=torch.bfloat16))'
tensor([255., 256., 256., 258., 260.], dtype=torch.bfloat16)
```

The issue is that when generating the range, we accumulate in the low-precision data type. In the case of bfloat16, numbers like 257 and 259 don't have a representation in the number format, so we enter an infinite loop adding 1 to 256. We should accumulate in a higher-precision format before downcasting to the actual datatype.

With this change, our result matches that of the upstream:

```
$ python -c 'import torch, torch_xla; print(torch.arange(255, 260, dtype=torch.bfloat16, device="xla"))'
tensor([255., 256., 256., 258., 260.], device='xla:0', dtype=torch.bfloat16)
```

### DIFF
--- a/torch_xla/csrc/helpers.h
+++ b/torch_xla/csrc/helpers.h
@@ -334,10 +334,13 @@ class XlaHelpers {
   static xla::XlaOp PromotedLogicalUnaryOp(
       xla::XlaOp op, const std::function<xla::XlaOp(xla::XlaOp)>& unary_op);
 
-  template <typename T>
-  static xla::Literal Range(T start, T end, T step) {
+  // T is the returned type, A is the type used for accumulation. In general,
+  // A should have higher-or-equal-precision to T.
+  template <typename T, typename A = T>
+  static xla::Literal Range(A start, A end, A step) {
+    std::vector<A> accumulated = runtime::util::Range<A>(start, end, step);
     return xla::LiteralUtil::CreateR1<T>(
-        runtime::util::Range<T>(start, end, step));
+        std::vector<T>(accumulated.begin(), accumulated.end()));
   }
 
   static xla::PrecisionConfig::Precision mat_mul_precision() {
--- a/torch_xla/csrc/ops/ops.cpp
+++ b/torch_xla/csrc/ops/ops.cpp
@@ -431,20 +431,16 @@ torch::lazy::NodePtr ARange(const at::Scalar& start, const at::Scalar& end,
   xla::Literal values;
   switch (type) {
     case xla::PrimitiveType::BF16:
-      values = XlaHelpers::Range<tsl::bfloat16>(
-          static_cast<tsl::bfloat16>(start.toFloat()),
-          static_cast<tsl::bfloat16>(end.toFloat()),
-          static_cast<tsl::bfloat16>(step.toFloat()));
+      values = XlaHelpers::Range<tsl::bfloat16, double>(
+          start.toDouble(), end.toDouble(), step.toDouble());
       break;
     case xla::PrimitiveType::F16:
-      values =
-          XlaHelpers::Range<xla::half>(static_cast<xla::half>(start.toHalf()),
-                                       static_cast<xla::half>(end.toHalf()),
-                                       static_cast<xla::half>(step.toHalf()));
+      values = XlaHelpers::Range<xla::half, double>(
+          start.toDouble(), end.toDouble(), step.toDouble());
       break;
     case xla::PrimitiveType::F32:
-      values = XlaHelpers::Range<float>(start.toFloat(), end.toFloat(),
-                                        step.toFloat());
+      values = XlaHelpers::Range<float, double>(
+          start.toDouble(), end.toDouble(), step.toDouble());
       break;
     case xla::PrimitiveType::F64:
       values = XlaHelpers::Range<double>(start.toDouble(), end.toDouble(),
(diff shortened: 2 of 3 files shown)


torch_xla/csrc/helpers.h
```diff
@@ -322,10 +322,13 @@ class XlaHelpers {
   static xla::XlaOp PromotedLogicalUnaryOp(
       xla::XlaOp op, const std::function<xla::XlaOp(xla::XlaOp)>& unary_op);
 
-  template <typename T>
-  static xla::Literal Range(T start, T end, T step) {
+  // T is the returned type, A is the type used for accumulation. In general,
+  // A should have higher-or-equal-precision to T.
+  template <typename T, typename A = T>
+  static xla::Literal Range(A start, A end, A step) {
+    std::vector<A> accumulated = runtime::util::Range<A>(start, end, step);
```
will-cromar, line 329: Should `runtime::util::Range` just be defined in terms of double, or just written more safely? e.g. creating an iota and multiplying it by `step` should prevent accumulating imprecision and hangs.

This is an easy mistake to make, so I'd rather make `Range` less dangerous if we can.

jonb377:
This also impacts integral types when the values would overflow, e.g.:

```
$ python -c 'import torch, torch_xla; print(torch.arange(255, 260, dtype=torch.int8, device="xla"))'
```

torch_xla/csrc/helpers.h
jonb377: Using double for everything would have some corner cases with integral types I think, but I agree a better fix would be in `Range`. I'll see if I can find how the upstream is handling this.

yeounoh: We could add one or two safeguards in the runtime::range impl, where we check if the step is >= 0, and/or cur and prev are not the same in each iteration. If we would support negative step, then we would need to update this path, too:
```
} else {
    for (; start > end; start += step) {
      result.push_back(start);
    }
  }
 ```
To be clear, the range here is inclusive? I thought the general range functions are exclusive (of end).

Comment (jonb377) [reply torch_xla/csrc/helpers.h:329]: It's exclusive, the 260 in the output comes from 259's bfloat16 representation:

```
>>> torch.Tensor([259]).to(dtype=torch.bfloat16)
tensor([260.], dtype=torch.bfloat16)
```

Will pointed me to this xla::Array method we might be able to use instead of having our own Range implementation, I'm going to give it a shot: https://github.com/openxla/xla/blob/0756a113f9e65ac46805080e3171b030e6a9b4c6/xla/array.h#L244-L249

File: torch_xla/csrc/helpers.h [hunk 2]
```diff
@@ -324,8 +324,14 @@ class XlaHelpers {
 
   template <typename T>
   static xla::Literal Range(T start, T end, T step) {
+    int64_t range_size = (int64_t)std::ceil((end - start) / step);
+    xla::Array<T> range({range_size});
+    range.FillWithMultiples(step);
+    range.Each([start](absl::Span<const int64_t> index, T* value) {
```
will-cromar, line 330: micro-nit: can you capture `start` by reference?

From our style guide: [`Use default capture by reference ([&]) only when the lifetime of the lambda is obviously shorter than any potential captures.`](https://google.github.io/styleguide/cppguide.html#Lambda_expressions)

torch_xla/csrc/helpers.h
```diff
@@ -324,8 +324,14 @@ class XlaHelpers {
...
+      *value += start;
+    });
     return xla::LiteralUtil::CreateR1<T>(
-        runtime::util::Range<T>(start, end, step));
+        absl::MakeConstSpan(range.data(), range.dim(0)));
```
will-cromar, line 334: Can you create from the array directly? https://github.com/openxla/xla/blob/fbcb441c8b5a327c7092bc21d31a3e598f11dd0c/xla/literal_util.h#L132-L133

will-cromar:
Can you also delete `runtime::util::Range`? This is the only usage I could find.

torch_xla/csrc/helpers.h
```diff
@@ -324,8 +324,14 @@ class XlaHelpers {
 
   template <typename T>
   static xla::Literal Range(T start, T end, T step) {
+    int64_t range_size = (int64_t)std::ceil((end - start) / step);
```
will-cromar, line 327: nit: `static_cast<int64_t>`

https://google.github.io/styleguide/cppguide.html#Casting

torch_xla/csrc/helpers.h
jonb377: Nice find! I missed that API.

jonb377:
The current approach of filling with multiples doesn't always agree with the CPU result:

```
root@t1v-n-f494979e-w-0:/root# python -c 'import torch, torch_xla; print(torch.arange(0, 1024, 257, dtype=torch.bfloat16))'
tensor([  0., 256., 512., 772.], dtype=torch.bfloat16)
root@t1v-n-f494979e-w-0:/root# python -c 'import torch, torch_xla; print(torch.arange(0, 1024, 257, dtype=torch.bfloat16, device="xla"))'
tensor([  0., 256., 512., 768.], device='xla:0', dtype=torch.bfloat16)
```

The original downcasting approach does agree.

jonb377:
Making this a draft PR, we need to find a compatible solution.

will-cromar (approved):

PR #5643, opened 2023-09-25
lower native_dropout

### LINKED ISSUE #5588: OP lowering for op `native_dropout`
Try manual lowing the op for `native_dropout` including adding the unit test. 

As @JackCaoG has mentioned, `native_dropout` regress a lot model training on torchbench. We plan to add this into the 2.1 release.

zpcore:
Fix #5588.

Added the manual lowering for native_dropout op. We need to add the test to verify whether the logic is correct or not.

### DIFF
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -1400,6 +1400,29 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
   ExpectCounterChanged("xla::bernoulli", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestNativeDropout) {
+  torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
+  float allowance = 0.04;
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    for (float probability : {0.1, 0.5}) {
+      auto [xla_b_val, xla_b_mask] =
+          torch::native_dropout(xla_a, probability, /*train=*/true);
+      double prob = static_cast<double>(
+                        xla_b_val.cpu().eq(0.0f).sum().item().toDouble()) /
+                    a.numel();
+      EXPECT_GT(prob, probability - allowance);
+      EXPECT_LT(prob, probability + allowance);
+      EXPECT_EQ(xla_b_val.scalar_type(), torch::kFloat);
+      EXPECT_EQ(xla_b_mask.scalar_type(), torch::kBool);
+    }
+  });
+
+  ExpectCounterNotChanged("aten::(?!_local_scalar_dense).*",
+                          cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::native_dropout", cpp_test::GetIgnoredCounters());
+}
+
 TEST_F(AtenXlaTensorTest, TestNativeDropoutNotTrain) {
   torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
   ForEachDevice([&](const torch::Device& device) {
@@ -1416,19 +1439,15 @@ TEST_F(AtenXlaTensorTest, TestNativeDropoutNotTrain) {
   ExpectCounterChanged("xla::native_dropout", cpp_test::GetIgnoredCounters());
 }
 
-TEST_F(AtenXlaTensorTest, TestNativeDropout) {
+TEST_F(AtenXlaTensorTest, TestNativeDropoutMask) {
   torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
   ForEachDevice([&](const torch::Device& device) {
     torch::Tensor xla_a = CopyToDevice(a, device);
     auto [xla_b_val, xla_b_mask] =
-        torch::native_dropout(xla_a, 0.1, /*train=*/true);
-    double prob =
-        static_cast<double>(xla_b_val.cpu().ne(0.0f).sum().item().toDouble()) /
-        a.numel();
-    EXPECT_GT(prob, 0.06);
-    EXPECT_LT(prob, 0.14);
-    EXPECT_EQ(xla_b_val.scalar_type(), torch::kFloat);
-    EXPECT_EQ(xla_b_mask.scalar_type(), torch::kBool);
+        torch::native_dropout(xla_a, 0.5, /*train=*/true);
+    auto count1 = xla_b_val.cpu().eq(0.0f).sum().item().toInt();
+    auto count2 = xla_b_mask.cpu().eq(0.0f).sum().item().toInt();
+    EXPECT_EQ(count1, count2);
   });
 
   ExpectCounterNotChanged("aten::(?!_local_scalar_dense).*",
@@ -1436,15 +1455,17 @@ TEST_F(AtenXlaTensorTest, TestNativeDropout) {
   ExpectCounterChanged("xla::native_dropout", cpp_test::GetIgnoredCounters());
 }
 
-TEST_F(AtenXlaTensorTest, TestNativeDropoutMask) {
+TEST_F(AtenXlaTensorTest, TestNativeDropoutZeroProbability) {
   torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
   ForEachDevice([&](const torch::Device& device) {
     torch::Tensor xla_a = CopyToDevice(a, device);
     auto [xla_b_val, xla_b_mask] =
-        torch::native_dropout(xla_a, 0.5, /*train=*/true);
+        torch::native_dropout(xla_a, 0, /*train=*/true);
     auto count1 = xla_b_val.cpu().ne(0.0f).sum().item().toInt();
     auto count2 = xla_b_mask.cpu().ne(0.0f).sum().item().toInt();
+    auto count3 = xla_a.cpu().ne(0.0f).sum().item().toInt();
     EXPECT_EQ(count1, count2);
+    EXPECT_EQ(count2, count3);
   });
 
   ExpectCounterNotChanged("aten::(?!_local_scalar_dense).*",
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -1406,7 +1406,7 @@ TEST_F(AtenXlaTensorTest, TestNativeDropoutNotTrain) {
     torch::Tensor xla_a = CopyToDevice(a, device);
     auto [xla_b_val, xla_b_mask] =
         torch::native_dropout(xla_a, 0.5, /*train=*/false);
-    AllClose(xla_b_val, xla_a);
+    AllEqual(xla_b_val, xla_a);
     EXPECT_EQ(xla_b_val.scalar_type(), torch::kFloat);
     EXPECT_EQ(xla_b_mask.scalar_type(), torch::kBool);
   });
@@ -1429,7 +1429,22 @@ TEST_F(AtenXlaTensorTest, TestNativeDropout) {
     EXPECT_LT(prob, 0.14);
     EXPECT_EQ(xla_b_val.scalar_type(), torch::kFloat);
     EXPECT_EQ(xla_b_mask.scalar_type(), torch::kBool);
-    // AllClose(xla_b_val*xla_b_mask, xla_b_val);
+  });
+
+  ExpectCounterNotChanged("aten::(?!_local_scalar_dense).*",
+                          cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::native_dropout", cpp_test::GetIgnoredCounters());
+}
+
+TEST_F(AtenXlaTensorTest, TestNativeDropoutMask) {
+  torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    auto [xla_b_val, xla_b_mask] =
+        torch::native_dropout(xla_a, 0.5, /*train=*/true);
+    auto count1 = xla_b_val.cpu().ne(0.0f).sum().item().toInt();
+    auto count2 = xla_b_mask.cpu().ne(0.0f).sum().item().toInt();
+    EXPECT_EQ(count1, count2);
   });
 
   ExpectCounterNotChanged("aten::(?!_local_scalar_dense).*",
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -1400,6 +1400,24 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
   ExpectCounterChanged("xla::bernoulli", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestNativeDropout) {
+  torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    auto [xla_b_val, xla_b_mask] = torch::native_dropout(xla_a, 0.1, /*train=*/true);
+    double prob =
+        static_cast<double>(xla_b_val.cpu().ne(0.0f).sum().item().toDouble()) /
+        a.numel();
+    EXPECT_GT(prob, 0.86);
+    EXPECT_LT(prob, 0.94);
+  });
+
+  ExpectCounterNotChanged("aten::(?!_local_scalar_dense).*",
+                          cpp_test::GetIgnoredCounters());
+  // dropout is composed of many arithmetic ops.
+  ExpectCounterChanged("xla::bernoulli", cpp_test::GetIgnoredCounters());
+}
+
 TEST_F(AtenXlaTensorTest, TestRandperm) {
   int n = 5;
   torch::Tensor shuffle = torch::randperm(
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -1400,22 +1400,41 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
   ExpectCounterChanged("xla::bernoulli", cpp_test::GetIgnoredCounters());
 }
 
-TEST_F(AtenXlaTensorTest, TestNativeDropout) {
+TEST_F(AtenXlaTensorTest, TestNativeDropoutNotTrain) {
   torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
   ForEachDevice([&](const torch::Device& device) {
     torch::Tensor xla_a = CopyToDevice(a, device);
-    auto [xla_b_val, xla_b_mask] = torch::native_dropout(xla_a, 0.1, /*train=*/true);
-    double prob =
-        static_cast<double>(xla_b_val.cpu().ne(0.0f).sum().item().toDouble()) /
-        a.numel();
-    EXPECT_GT(prob, 0.86);
-    EXPECT_LT(prob, 0.94);
+    auto [xla_b_val, xla_b_mask] =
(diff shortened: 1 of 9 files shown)


torch_xla/csrc/ops/native_dropout.cpp
```diff
@@ -0,0 +1,36 @@
...
+NativeDropout::NativeDropout(const torch::lazy::Value& input, float p,
+                             c10::optional<bool> train,
+                             const torch::lazy::Value& seed)
+    : XlaNode(torch::lazy::OpKind(at::aten::native_dropout), {input, seed},
+              [&]() { return NodeOutputShape(input); }, 2),
+      p_(p),
+      train_(train) {}
+
+torch::lazy::NodePtr NativeDropout::Clone(torch::lazy::OpList operands) const {
+  return torch::lazy::MakeNode<NativeDropout>(operands.at(0), operands.at(1),
```
lsy323, line 26: `operands` should only contain Lazy IR Node. So here it should be:
```
return torch::lazy::MakeNode<NativeDropout>(operands.at(0), p_, train_);
```

torch_xla/csrc/ops/native_dropout.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "torch_xla/csrc/helpers.h"
+#include "torch_xla/csrc/lowering_context.h"
+#include "torch_xla/csrc/xla_lower_util.h"
+
+namespace torch_xla {
+namespace {
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input) {
+  xla::Shape input_shape = GetXlaShape(input);
+  return xla::ShapeUtil::MakeTupleShape({input_shape, input_shape});
```
lsy323, line 12: `xla::Shape` contains dtype info. In PyTorch, there is Bool Tensor in PyTorch, but there is no `bool` in `xla::PrimitiveType`. So let's use `uint8` instead.

```
shape.set_element_type(xla::PrimitiveType::U8);
```

lsy323: In the test, we may check the output tensor type, note that we may hit dtype mismatch `bool` vs `uint8`. If that happens, let's just assert `uint8` for the mask output and add a comment in the test code.

torch_xla/csrc/ops/native_dropout.cpp
zpcore: Should we also consider seed (with type torch::lazy::Value) as a lazy IR node?

torch_xla/csrc/ops/native_dropout.cpp
JackCaoG: I think bool in XLA is called `xla::PRED`

torch_xla/csrc/ops/native_dropout.cpp
zpcore: I just checked example multinomial.cpp, the seed is also considered as the lazy IR node. Also fixed p_ and train_ to be val. Thanks!

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -1400,6 +1400,24 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
...
+        static_cast<double>(xla_b_val.cpu().ne(0.0f).sum().item().toDouble()) /
+        a.numel();
+    EXPECT_GT(prob, 0.86);
+    EXPECT_LT(prob, 0.94);
+  });
+
+  ExpectCounterNotChanged("aten::(?!_local_scalar_dense).*",
+                          cpp_test::GetIgnoredCounters());
+  // dropout is composed of many arithmetic ops.
+  ExpectCounterChanged("xla::bernoulli", cpp_test::GetIgnoredCounters());
```
JackCaoG, line 1418: you want to check the counter for `native_dropout` instead of `bernoulli`. It should just be a `native_dropout` otherwise it is not testing your lowering.

torch_xla/csrc/ops/native_dropout.cpp
```diff
@@ -0,0 +1,36 @@
...
+NativeDropout::NativeDropout(const torch::lazy::Value& input, float p,
+                             c10::optional<bool> train,
+                             const torch::lazy::Value& seed)
+    : XlaNode(torch::lazy::OpKind(at::aten::native_dropout), {input, seed},
+              [&]() { return NodeOutputShape(input); }, 2),
+      p_(p),
+      train_(train) {}
+
+torch::lazy::NodePtr NativeDropout::Clone(torch::lazy::OpList operands) const {
+  return torch::lazy::MakeNode<NativeDropout>(operands.at(0), p_, train_, operands.at(3));
```
lsy323, line 27: Size of operands should be 2, only Lazy IR nodes in the `operands`

torch_xla/csrc/ops/native_dropout.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+
+XlaOpVector NativeDropout::Lower(LoweringContext* loctx) const {
+  xla::XlaOp input = loctx->GetOutputOp(operand(0));
+  xla::XlaOp seed = loctx->GetOutputOp(operand(3));
```
lsy323, line 32: same here, `operand(1)` for seed

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -1400,6 +1400,24 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
   ExpectCounterChanged("xla::bernoulli", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestNativeDropout) {
```
lsy323, line 1403: I think we need at least 3 test cases to cover the lowering logic

- prob > 0, train = true
- prob = 0, train = true
- train = false

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -1400,6 +1400,24 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
...
+  torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    auto [xla_b_val, xla_b_mask] = torch::native_dropout(xla_a, 0.1, /*train=*/true);
```
lsy323, line 1407: `p` is the probability that the elements will be zeroed out. So I think we should check a range around `p` instead of `1 - p`

torch_xla/csrc/ops/native_dropout.cpp
zpcore: Thanks for pointing out! With this fix, the number of output mismatch issue got resolved.

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -1400,6 +1400,24 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
...
+    double prob =
+        static_cast<double>(xla_b_val.cpu().ne(0.0f).sum().item().toDouble()) /
+        a.numel();
+    EXPECT_GT(prob, 0.86);
+    EXPECT_LT(prob, 0.94);
```
lsy323, line 1412: We also need to check the correctness of mask, we can check `prob_activation == prob_mask`

Also we need to check the output dtype of activation and mask
```
EXPECT_EQ(xla_b_val.scalar_type(), torch::kFloat);
EXPECT_EQ(xla_b_mask.scalar_type(), torch::kBool);
```

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1886,6 +1887,15 @@ std::tuple<XLATensorPtr, XLATensorPtr, XLATensorPtr> native_batch_norm_backward(
                          std::move(grad_bias));
 }
 
+std::tuple<XLATensorPtr, XLATensorPtr> native_dropout(
+    const XLATensorPtr& input, double p, c10::optional<bool> train) {
+  torch::lazy::NodePtr node = torch::lazy::MakeNode<NativeDropout>(
+      input->GetIrValue(), p, train,
+      XLAGraphExecutor::Get()->GetRngSeed(input->GetDevice()));
+  return std::make_tuple(input->CreateFrom(torch::lazy::Value(node, 0)),
+                         input->CreateFrom(torch::lazy::Value(node, 1)));
```
lsy323, line 1896: Since the result torch tensor will be built from Lazy IR node, we need to set the correct type here. we can pass an additional arg for dtype at https://github.com/pytorch/xla/blob/a0a3aba9e5263e5aafd6e7895f36e2703ed3c4cd/torch_xla/csrc/tensor.h#L167

We should pass `at::ScalarType::Bool` for mask

test/cpp/test_aten_xla_tensor_1.cpp
zpcore: Cool! updated with 
```
EXPECT_GT(prob, 0.06);
EXPECT_LT(prob, 0.14);
```
And now the condition passed.

test/cpp/test_aten_xla_tensor_1.cpp
zpcore: Cool! updated with 
```
EXPECT_GT(prob, 0.06);
EXPECT_LT(prob, 0.14);
```
And now the condition passed.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -504,6 +504,24 @@ xla::XlaOp BuildDropout(xla::XlaOp input, float probability, xla::XlaOp seed) {
...
+std::vector<xla::XlaOp> BuildNativeDropout(xla::XlaOp input, float probability,
+                                           c10::optional<bool> train,
+                                           xla::XlaOp seed) {
+  const xla::Shape& shape = ShapeHelper::ShapeOfXlaOp(input);
+  if (!train.has_value() || *train) {
+    xla::XlaOp prob =
+        XlaHelpers::ScalarBroadcast<float>(probability, shape, input.builder());
+    xla::XlaOp mask = BuildBernoulli(prob, seed, shape.element_type());
+    if (probability > 0.0f) {
+      mask = mask / prob;
```
lsy323, line 516: We should scale the activation by `1/(1-p)`. 

Let's do
```
xla::XlaOp one = xla::One(input.builder(), shape.element_type());
mask = mask / (one - prob);
```

torch_xla/csrc/ops/native_dropout.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "torch_xla/csrc/lowering_context.h"
+#include "torch_xla/csrc/xla_lower_util.h"
+
+namespace torch_xla {
+namespace {
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input) {
+  xla::Shape input_shape = GetXlaShape(input);
+  input_shape.set_element_type(xla::PrimitiveType::U8);
+  return xla::ShapeUtil::MakeTupleShape({input_shape, input_shape});
```
lsy323, line 13: You are setting both activation and mask to be bool here. You can do:
```
xla::Shape mask_shape = GetXlaShape(input);
mask_shape.set_element_type(xla::PrimitiveType::PRED);
```

torch_xla/csrc/ops/native_dropout.cpp
zpcore: Thanks, we have converted the mask tensor into `xla::PRED`.

torch_xla/csrc/xla_lower_util.cpp
zpcore: You are right, that matches the implementation: https://github.com/pytorch/pytorch/blob/f87863335c8e8e5a6f51c6b35194f7b68d44be17/aten/src/ATen/native/Dropout.cpp#L116C1-L118. Fixed!

torch_xla/csrc/ops/native_dropout.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "torch_xla/csrc/helpers.h"
+#include "torch_xla/csrc/lowering_context.h"
+#include "torch_xla/csrc/xla_lower_util.h"
+
+namespace torch_xla {
+namespace {
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input) {
+  xla::Shape input_shape = GetXlaShape(input);
+  input_shape.set_element_type(xla::PrimitiveType::U8);
```
lsy323, line 12: We should remove this line, this forces the output activation to be U8

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -504,6 +504,26 @@ xla::XlaOp BuildDropout(xla::XlaOp input, float probability, xla::XlaOp seed) {
...
+std::vector<xla::XlaOp> BuildNativeDropout(xla::XlaOp input, xla::XlaOp seed,
+                                           float probability,
+                                           c10::optional<bool> train) {
+  const xla::Shape& shape = ShapeHelper::ShapeOfXlaOp(input);
+  if (!train.has_value() || *train) {
+    xla::XlaOp prob =
+        XlaHelpers::ScalarBroadcast<float>(probability, shape, input.builder());
+    xla::XlaOp one = xla::One(input.builder(), shape.element_type());
+    xla::XlaOp mask = BuildBernoulli(prob, seed, shape.element_type());
+    mask = mask / (one - prob);
```
lsy323, line 516: this line should be in the `if (probability > 0.0f) {` block?

torch_xla/csrc/ops/native_dropout.cpp
```diff
@@ -0,0 +1,39 @@
...
+  return xla::ShapeUtil::MakeTupleShape({input_shape, mask_shape});
+}
+
+}  // namespace
+
+NativeDropout::NativeDropout(const torch::lazy::Value& input,
+                             const torch::lazy::Value& seed, float p,
+                             c10::optional<bool> train)
+    : XlaNode(torch::lazy::OpKind(at::aten::native_dropout), {input, seed},
+              [&]() { return NodeOutputShape(input); }, 2),
```
lsy323, line 24: The hash is missing in the XlaNode constructor. You need it to distinguish different nativeDropout nodes. 

Just do
```
XlaNode(..., torch::lazy::MHash(p, train))
```

lsy323: I guess that's what's happening, if we don't have a hash for the nativedropout node, first a `train=false` node is created, then when the other test creates a node `train=true`, it will consider the `train=false` node to be the same node, and reused it.

zpcore: Thanks, I thought the default hash function will work... Could you give more details about why the default hash causes the weird status issue?

lsy323: The constructor has a default value for the hash value. see https://github.com/pytorch/xla/blob/00272c34aad2aaf61b37909ce11453c8be58aad9/torch_xla/csrc/ir.h

zpcore: Thanks for the explanation!

torch_xla/csrc/xla_lower_util.cpp
zpcore: Nice catch!

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -504,6 +504,26 @@ xla::XlaOp BuildDropout(xla::XlaOp input, float probability, xla::XlaOp seed) {
...
+                                           c10::optional<bool> train) {
+  const xla::Shape& shape = ShapeHelper::ShapeOfXlaOp(input);
+  if (!train.has_value() || *train) {
+    xla::XlaOp prob =
+        XlaHelpers::ScalarBroadcast<float>(probability, shape, input.builder());
+    xla::XlaOp one = xla::One(input.builder(), shape.element_type());
+    xla::XlaOp mask = BuildBernoulli(prob, seed, shape.element_type());
+    if (probability > 0.0f) {
+      mask = mask / (one - prob);
+      mask = mask / prob;
```
lsy323, line 518: We need to remove this line right?

zpcore: Removed the typo line.

test/cpp/test_aten_xla_tensor_1.cpp
lsy323: @zpcore I think prob = 0 train = true test is missing. Can you please add this case? This is a special handling in the lowering logic

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -1400,6 +1400,58 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
...
+}
+
+TEST_F(AtenXlaTensorTest, TestNativeDropout) {
+  torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    auto [xla_b_val, xla_b_mask] =
+        torch::native_dropout(xla_a, 0.1, /*train=*/true);
+    double prob =
+        static_cast<double>(xla_b_val.cpu().ne(0.0f).sum().item().toDouble()) /
```
lsy323, line 1426: Sorry for the confusion, I think we should check `xla_b_val.cpu().eq(0.0f)` here. `p` is the probability that zero out.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -504,6 +504,25 @@ xla::XlaOp BuildDropout(xla::XlaOp input, float probability, xla::XlaOp seed) {
   return input * mask;
 }
 
+std::vector<xla::XlaOp> BuildNativeDropout(xla::XlaOp input, xla::XlaOp seed,
+                                           float probability,
+                                           c10::optional<bool> train) {
+  const xla::Shape& shape = ShapeHelper::ShapeOfXlaOp(input);
+  if (!train.has_value() || *train) {
+    xla::XlaOp prob =
+        XlaHelpers::ScalarBroadcast<float>(probability, shape, input.builder());
```
lsy323, line 513: here I think it should be `1 - probability`. Because in the `BuildBernoulli` call. It will return a mask that `prob` of the elements are 1, the rest are 0.

torch_xla/csrc/ops/native_dropout.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "torch_xla/csrc/helpers.h"
+#include "torch_xla/csrc/lowering_context.h"
+#include "torch_xla/csrc/xla_lower_util.h"
+
+namespace torch_xla {
+namespace {
+
+xla::Shape NodeOutputShape(const torch::lazy::Value& input) {
+  xla::Shape input_shape = GetXlaShape(input);
```
lsy323, line 11: I cannot repro the CI failure on native_dropout_backward. But it seems type of mask output mismatches with the type specified here.

Shall we just do the following here.
```
return xla::ShapeUtil::MakeTupleShape({input_shape, input_shape});
```
I had a try locally, the torch tensor would still have boolean type, seems the dtype would be casted when the tensor is built from lazy ir.

zpcore: Yes, the conversion here doesn't affect type.

lsy323 (approved):

lsy323:
@zpcore Thanks for working on this, good work!

zpcore:
Thanks @JackCaoG and @lsy323 for the feedback, it's a great learning experience!

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -1400,6 +1400,79 @@ TEST_F(AtenXlaTensorTest, TestDropoutInPlace) {
   ExpectCounterChanged("xla::bernoulli", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestNativeDropout) {
+  torch::Tensor a = torch::rand({17, 21}, torch::TensorOptions(torch::kFloat));
+  float allowance = 0.04;
```
vanbasten23, line 1405: Very cool! 

quick question, how is value 0.04 determined?

zpcore: Thanks, this value was referred from the existing dropout test code [here](https://github.com/pytorch/xla/blob/498be74cb14d01ec2a535d212615907f68e41a1b/test/cpp/test_aten_xla_tensor_1.cpp#L1367C1-L1383C2). They didn't mention the value 0.04 directly, but based on the [0.86, 0.94] range, the value `0.04` is the allowance. To be honest, I don't know why they choose this number.

PR #5644, opened 2023-09-25
Dynamic PJRT plugin registration API

will-cromar:
First pass at implementing a common API for device plugins. The eventual goal is to remove any cases where we have to hard-code the device type in our build, allowing truly dynamic plugins through the PJRT plugin API.

- Add `DevicePlugin` API, including sample implementation for TPU
- Enable dynamic plugins with `PJRT_DYNAMIC_PLUGINS=1` or `plugins.use_dynamic_plugins`
- Add registration mechanism for PJRT plugins. If you register a valid plugin with `plugin.register_plugin` and enable plugins, then you can use the same device by name by setting `PJRT_DEVICE`. (see integration test in this PR for an example)
  - Completely new devices will probably not work yet. We still rely too much on parsing hard-coded strings in the internals.
- Default behavior doesn't change at all for now

Future work:

- Remove or update `XlaDeviceType` so plugins don't have to register their device strings in this repository
- Move GPU client into PJRT Plugin
- Automatically register TPU and GPU plugins and remove hard-coded PJRT client initialization.


will-cromar:
Heads up @jzhoulon @aws-kingrj, I'm working on a new way for external packages to register PJRT plugins with `torch_xla`. No action is required from you right now. I'll keep Neuron and XPU working within this repository while we develop the idea.

When this API is finalized, we can move plugin registration (something like `TpuPlugin` in this PR) into your respective packages.

will-cromar:
Leaving this as draft for now until I rebase after #5677, but this PR is largely ready for comments.

test/pjrt/test_dynamic_plugin_tpu.py
```diff
@@ -0,0 +1,34 @@
...
+  def setUpClass(cls):
+    xr.set_device_type('TPU')
+
+  @staticmethod
+  def _assert_tpus_exist(index=0):
+    del index
+    assert len(xm.get_xla_supported_devices('TPU')) > 0
+
+  def test_dynamic_plugin_api(self):
+    with concurrent.futures.ProcessPoolExecutor(max_workers=1) as executor:
```
vanbasten23, line 26: I guess the difference b/w test_dynamic_plugin_api and test_spawn is that the former test single processing and the latter test the multi-processing?

torch_xla/experimental/plugins.py
```diff
@@ -0,0 +1,67 @@
...
+  return xu.getenv_as(xenv.PJRT_DYNAMIC_PLUGINS, bool, False)
+
+
+def default() -> DevicePlugin:
+  return _plugin_registry[xr.device_type()]
+
+
+def register_plugin(name: str, device_plugin: DevicePlugin):
+  _plugin_registry[name.upper()] = device_plugin
+  torch_xla._XLAC._register_pjrt_plugin(name, device_plugin.library_path())
```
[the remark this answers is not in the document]
will-cromar: None right now. GPU support is statically linked in. When that moves to a plugin (say `libsegpu.so`), it will be the path to that binary.

test/pjrt/test_dynamic_plugin_tpu.py
will-cromar: Yeah, I wrote `test_dynamic_plugin_api` before the other. I'll change the name to something like `test_single_process` to be more clear

test/pjrt/test_dynamic_plugin_tpu.py
```diff
@@ -0,0 +1,34 @@
...
+
+class TestDynamicTpuPlugin(absltest.TestCase):
+
+  @classmethod
+  def setUpClass(cls):
+    xr.set_device_type('TPU')
+
+  @staticmethod
+  def _assert_tpus_exist(index=0):
+    del index
```
JackCaoG, line 22: what's this for?

torch_xla/_internal/pjrt.py
```diff
@@ -96,7 +97,9 @@ def _run_singleprocess(fn: Callable[..., R], *args, **kwargs) -> Dict[int, R]:
   """
   os.environ.setdefault(xenv.PJRT_LOCAL_PROCESS_COUNT, '1')
 
-  if runtime.device_type() == 'TPU':
+  if plugins.using_dynamic_plugins():
+    plugins.default().configure_single_process()
```
JackCaoG, line 101: should we throw a warning or something when pople configure `PJRT_DEVICE` while also register the plugin in the code?

torch_xla/csrc/runtime/computation_client.cc
```diff
@@ -53,6 +53,21 @@ int64_t ComputationClient::GetDeviceOrdinal(const std::string& device) {
   return std::stoi(device.substr(pos + 1));
 }
 
+std::unordered_map<std::string, std::string> pjrt_plugins_;
```
JackCaoG, line 56: I thought user can only register one plug in? What;s the use case of registering multiple?

test/pjrt/test_dynamic_plugin_tpu.py
will-cromar: `index` is required for `spawn`, but we don't need it. I just explicitly delete it to mark it unused

torch_xla/_internal/pjrt.py
will-cromar: You still select the device type with `PJRT_DEVICE`. Plugins will just let you register new device types when we clean up all of the hardcoded strings.

torch_xla/csrc/runtime/computation_client.cc
will-cromar: We'll register TPU and GPU as default options, and then other packages will add plugins on top of those. JAX is also using [Python entry points](https://packaging.python.org/en/latest/specifications/entry-points/) to register available plugins automatically, which we may also want to do.

test/pjrt/test_dynamic_plugin_tpu.py
```diff
@@ -0,0 +1,34 @@
+import concurrent.futures
+
+from absl.testing import absltest
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.xla_multiprocessing as xmp
+from torch_xla.experimental import plugins
+import torch_xla.runtime as xr
+from torch_xla._internal import tpu
+
+plugins.register_plugin('TPU', tpu.TpuPlugin())
```
JackCaoG, line 10: are you going to put this in our init file eventually?

JackCaoG (approved):

test/pjrt/test_dynamic_plugin_tpu.py
will-cromar: Yeah. I'm avoiding any changes to the default behavior while this is WIP.

vanbasten23 (approved):
LGTM! Thanks.

PR #5653, opened 2023-09-27
Make `ComputationClient::Data` inherit from `torch::lazy::BackendData`

will-cromar:
- Remove `XLAData` wrapper
- Make our `Data` inherit directly from `BackendData`
  - Remove some virtual functions that duplicate parent
- Move current device state to `aten_xla_bridge` since there were already duplicate functions there
  - Breaks circular dependency (see comment below)
- Remove wrap/unwrap functions for single data pointers
- Leave wrap/unwrap functions for vectors, since we can't cleanly upcast or downcast them inline


will-cromar:
XLAData papers over a circular dependency:

1. `Data` should inherit from `BackendData`
2. Data needs to parse the device string into a `BackendDevice` for `BackendData`'s constructor
3. "Parsing" the empty string is [how we select the default device](https://github.com/pytorch/xla/blob/ea5073b4c445f14b1a6ecdfa86101b0d7da68a0d/torch_xla/csrc/device.cpp#L90-L94). [This requires the runtime](https://github.com/pytorch/xla/blob/ea5073b4c445f14b1a6ecdfa86101b0d7da68a0d/torch_xla/csrc/device.cpp#L44-L58).
4. The runtime includes Data
5. See (2)

Short term, we can separate the stateful functions in `device.cpp` (mainly default device selection) from the device type [parsing](https://github.com/pytorch/xla/blob/ea5073b4c445f14b1a6ecdfa86101b0d7da68a0d/torch_xla/csrc/device.cpp#L43) and [un-parsing](https://github.com/pytorch/xla/blob/ea5073b4c445f14b1a6ecdfa86101b0d7da68a0d/torch_xla/csrc/device.cpp#L38). Long-term, I'd rather just do away with [representing devices as strings at all](https://wiki.c2.com/?StringlyTyped).

JackCaoG:
yea I think we have to represent it as a string because the XRT being in a separate build target, now since device class can be shared we can just that instead.

JackCaoG:
is this ready for review?

will-cromar:
Not yet. I'm still doing cleanup. I'll move it out of draft when it's ready

torch_xla/csrc/aten_xla_bridge.cpp
```diff
@@ -337,23 +339,43 @@ std::string ToXlaString(const c10::Device& device) {
   return absl::StrCat("xla:", device.index());
 }
 
+const torch::lazy::BackendDevice* GetDefaultDevice() {
+  std::string default_device_spec =
```
JackCaoG, line 343: nit add `static`. `UseVirtualDevice` should not changed after default device is inited.

torch_xla/csrc/runtime/runtime.h
```diff
@@ -11,10 +14,6 @@ ComputationClient* GetComputationClient();
 
 ComputationClient* GetComputationClientIfInitialized();
 
-// Run the XRT local service, this will block the caller unitl the server
-// being stopped.
-void RunLocalService(uint64_t service_port);
```
JackCaoG, line 16: hahaha

torch_xla/csrc/unwrap_data.cpp
```diff
@@ -9,14 +8,13 @@
 
 #include "absl/types/span.h"
 #include "torch_xla/csrc/runtime/computation_client.h"
-#include "torch_xla/csrc/xla_data.h"
 
 namespace torch_xla {
 
 runtime::ComputationClient::DataPtr UnwrapXlaData(
```
JackCaoG, line 14: I actually would rename this function to something like `DynamicCastToXlaDatas` and other one to `DynamicCastToLazyDatas`, since it is not really a wrapping. We can do that in a follow up pr through.

JackCaoG (approved):
Mostly LGTM, thanks!

torch_xla/csrc/runtime/runtime.h
will-cromar: Oops! These changes are leftover from an earlier draft.

torch_xla/csrc/aten_xla_bridge.cpp
will-cromar: Done

torch_xla/csrc/unwrap_data.cpp
will-cromar: That name makes more sense. I wonder if we can/should add a general utility to dynamic-cast a vector of shared pointers. I'll see if there's any other usage.

alanwaketan (commented):
Sorry for late review, but it LGTM.

PR #5654, opened 2023-09-27
Expose replica_id in XLAShard

jonb377:
To optimize distributed checkpoint writing, we should distribute the writes for replicated shards across the hosts tracking the shard. This PR exposes the replica ID of each shard, which can be used to select a host responsible for writing when preparing the global plan. The value of `replica_id` depends on the type of sharding:
- `REPLICATED`: Since every host has a copy of the shard and there is no tiling, the replica ID is equal to the global device ordinal.
- `TILED`: When fully tiled, each shard lives only on a single device, so the replica id is always 0.
- `PARTIAL`: Partial replication is the main use case for `replica_id`. With this sharding type, the replica id is taken from the last tile dimension, which indicates the replication group assignment.

Selecting the shard's `replica_id` depends on the tiling assignment for partial replication, which is a small extension to the existing logic for `GetShardIndicesForDevices`. I opted to refactor this method to also return the shard's rank instead of declaring a new method. 

Aside from the refactor for the return type, the actual logic for choosing shard's replica_id is added to `ShardingUtil:: GetShardReplicaAndIndicesForDevices` in `torch_xla/csrc/xla_sharding_util.cpp`, and the change to add `replica_id` to `XLAShard` is in `xla_sharded_tensor.py`

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1694,13 +1694,14 @@ void InitXlaModuleBindings(py::module m) {
           }
           return std::make_tuple(shards, str_devices);
         });
-  // For each local shard, returns the replica_id and the indices into the
-  // global tensor as either a Python list of slices for each dimension or a
-  // Python Ellipsis object indicating that the tensor is replicated. These
-  // indices will not reflect any padding that has been applied to the shards.
-  // The order of the returned indices matches the order of the shards returned
-  // from
-  // `_get_local_shards`.
+  // For each local shard, returns the tuple:
+  //        (replica_id: int, indices: Union[List[Slice], Ellipsis]),
+  // where `replica_id` is the replica the shard belongs to and `indices` index
+  // into the global tensor. The value of `indices` is either a Python list of
+  // slices for each dimension or an Ellipsis object indicating that the tensor
+  // is replicated. These indices will not reflect any padding that has been
+  // applied to the shards. The order of the returned indices matches the order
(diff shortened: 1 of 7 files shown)


yeounoh:
@jonb377 I have a suggestion: instead of overloading the term `rank` which is more generic for host processes in PyTorch, let's name it as `replica_id` which is essentially what you want and precise, imo.

yeounoh (changes requested):
Will sync offline about https://github.com/pytorch/xla/pull/5654#issuecomment-1743564329

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1694,13 +1694,15 @@ void InitXlaModuleBindings(py::module m) {
...
         });
-  // Returns the indices of the shards into the global tensor as either
-  // a Python list of slices for each dimension or a Python Ellipsis object
-  // indicating that the tensor is replicated. These indices will not reflect
-  // any padding that has been applied to the shards. The order of the returned
-  // indices matches the order of the shards returned from `_get_local_shards`.
-  m.def("_get_local_shard_indices",
-        [](const at::Tensor& input) -> std::vector<py::object> {
+  // For each local shard, returns the replica_id and the indices into the
+  // global tensor as either a Python list of slices for each dimension or a
```
yeounoh, line 1698: a list of (rank_id, Union[slice, elipsis]) tuples, or something like that.

yeounoh (approved):
LGTM

jonb377:
Thanks @yeounoh!

PR #5656, opened 2023-09-29
Move distributed checkpointing to a subdirectory of experimental

jonb377:
The number of checkpointing-related files is growing in the experimental folder, so this change moves them to a dedicated subdirectory.


alanwaketan (approved):
LGTM as long as the helpers are not exposed to the users.

jonb377:
> LGTM as long as the helpers are not exposed to the users.

@alanwaketan Is keeping the helpers file private sufficient? Like `_helpers.py`

jonb377:
Rebased, I'll merge after TPU CI passes.

Issue #5558, opened 2023-09-13
[PJRT] Support GPU Multiple Hosts


wbmc:
## 🚀 Feature

Support GPU Multiple hosts in PJRT Runtime

## Motivation

requirements of multiple hosts multiple GPUs

## Alternatives

Introduce TorchElastic style to describes the information of Distributed Runtime 

## Additional context

```[tasklist]
### Tasks
- [ ] https://github.com/pytorch/xla/pull/5559
```

vanbasten23:
Since you've create the PR, how did you test it? Specifically, what command did you run?

wbmc:
> Since you've create the PR, how did you test it? Specifically, what command did you run?
I test on 2 gpu hosts. It needs to change the `test_gpu_multi_hosts.py` a few codes.


## The first host
The first host runs the command:
`CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 PJRT_DEVICE=GPU GPU_NUM_DEVICES=8 MASTER_PORT=30286 python test_gpu_multi_hosts.py`
```

world_size = 16
master_proc = Process(target=master_fn, args=(world_size,))
master_proc.start()
os.environ['GPU_NUM_DEVICES'] = str(world_size)
os.environ['PJRT_WORLD_SIZE'] = str(world_size)
configs = (
      ((0, 0, 2, world_size), {
          'cuda_visible_devices': '0,1'
      }),
      ((1, 0, 2, world_size), {
          'cuda_visible_devices': '0,1'
      }),
      ((0, 1, 2, world_size), {
          'cuda_visible_devices': '2,3'
      }),
      ((1, 1, 2, world_size), {
          'cuda_visible_devices': '2,3'
      }),
      ((0, 2, 2, world_size), {
          'cuda_visible_devices': '4,5'
      }),
      ((1, 2, 2, world_size), {
          'cuda_visible_devices': '4,5'
      }),
      ((0, 3, 2, world_size), {
          'cuda_visible_devices': '6,7'
      }),
      ((1, 3, 2, world_size), {
          'cuda_visible_devices': '6,7'
      }),
}
procs = []
for config in configs:
  procs.append(Process(target=worker_fn, args=config[0], kwargs=config[1]))
  procs[-1].start()
for p in procs:
  p.join()
master_proc.kill()
```

## The second host
The second host changes the world size to `16` and runs the command:
[...]

PR #5657, opened 2023-09-29
Add multi-host GPU support

vanbasten23:
Collaborating with @wbmc 

To start the multi-host (or multi-node) training, do:
```
PJRT_DEVICE=GPU torchrun \
--nproc_per_node=${NUMBER_LOCAL_GPU_DEVICES} \
--nnodes=${NUMBER_GPU_VM} \
--node_rank=${CURRENT_HOST_RANK} \
--rdzv_endpoint=<internal_ip_address> multinode_training_script.py
```
on each host.

The documentation will be updated in a follow-up PR.


JackCaoG:
@wbmc I have granted you the write access

wbmc:
> @wbmc I have granted you the write access

OK, Thanks!

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1844,7 +1844,7 @@ void InitXlaModuleBindings(py::module m) {
   m.def("_xla_get_distributed_runtime_service",
         [](int num_nodes) -> std::unique_ptr<xla::DistributedRuntimeService> {
           std::string dist_service_addr =
-              runtime::sys_util::GetEnvString("PJRT_DIST_SERVICE_ADDR", "");
+              runtime::sys_util::GetEnvString("MASTER_ADDR", "") + ":8547";
```
wbmc, line 1847: The port number `8547` seems to be a hardcoded value, which could potentially lead to conflicts with other processes in a cloud environment. Is it possible to allow users to specify a different port number? Otherwise, they would need to recompile the torch_xla program.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -120,10 +120,22 @@ PjRtComputationClient::PjRtComputationClient() {
...
+    //   auto local_process_count = sys_util::GetEnvInt(env::kEnvPjRtLocalProcessCount, 1);
+    //   options.num_nodes = local_process_count;
+    //   dist_runtime_service_ = xla::GetDistributedRuntimeService(dist_service, options).value();
+    // }
+    
+    // std::string dist_service_addr =
+    //     sys_util::GetEnvString(env::kEnvPjrtDistServiceAddr, "");
     std::string dist_service_addr =
-        sys_util::GetEnvString(env::kEnvPjrtDistServiceAddr, "");
+        sys_util::GetEnvString("MASTER_ADDR", "")+":8547";
```
wbmc, line 135: The port number `8547` seems to be a hardcoded value, which could potentially lead to conflicts with other processes in a cloud environment. Is it possible to allow users to specify a different port number? Otherwise, they would need to recompile the torch_xla program.

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: Yea, that makes sense. It should be possible.

vanbasten23: I added a flag COORDINATOR_PORT to set an alternative port to be used. In practice, I imagine the default port should be available most of the time. In case it isn't, users can unblock themselves that way.

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: ditto

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1843,8 +1843,9 @@ void InitXlaModuleBindings(py::module m) {
                                   py::call_guard<py::gil_scoped_release>());
   m.def("_xla_get_distributed_runtime_service",
         [](int num_nodes) -> std::unique_ptr<xla::DistributedRuntimeService> {
+          std::string port = runtime::sys_util::GetEnvString("COORDINATOR_PORT", "8547");
           std::string dist_service_addr =
-              runtime::sys_util::GetEnvString("PJRT_DIST_SERVICE_ADDR", "");
+              runtime::sys_util::GetEnvString("MASTER_ADDR", "127.0.0.1") + ":" + port;
```
miladm, line 1848: For better readability, can we introduce a variable to describe this IP address is the default parameter? e.g/ `LOCAL_HOST_IP_DEFAULT`

vanbasten23: sure

test/test_train_mp_imagenet_torchrun.py
```diff
@@ -0,0 +1,385 @@
+from absl import logging
```
vanbasten23, line 1: to reviewers: this is a torchrun version of test/test_train_mp_imagenet.py. 
Maybe I should merge it with test/test_train_mp_imagenet.py..

test/pjrt/test_torchrun.py
```diff
@@ -28,6 +28,41 @@ def test_all_gather(self):
 
     expected = torch.arange(0, expected_world_size, step=1, dtype=torch.float32)
     torch.testing.assert_close(result.cpu(), expected)
+    dist.destroy_process_group()
+
+  def test_all_reduce(self):
+    # The test is inspired by https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_reduce
+    dist.init_process_group('xla', init_method='xla://')
```
jonb377, line 35: Can we init/destroy the process group in setUpTest/tearDownTest to cover the early exit case?

test/pjrt/test_torchrun.py
```diff
@@ -28,6 +28,41 @@ def test_all_gather(self):
...
+
+    dist_world_size = xu.getenv_as('WORLD_SIZE', int)
+    devices_per_thread = xr.addressable_device_count()
+    world_size = dist_world_size * devices_per_thread
+    tensors = [torch.arange(2, dtype=torch.int64) + 1 + 2 * r for r in range(world_size)]
```
jonb377, line 40: Could you add a comment on what value the tensor should have? It's a bit hard to follow at the moment.

test/test_train_mp_imagenet_torchrun.py
jonb377: Are there any differences?

torch_xla/_internal/gpu.py
```diff
@@ -26,13 +28,12 @@ def initialize_distributed_runtime(global_world_size: int) -> None:
...
   """
   if global_world_size > 1:
-    # TODO(jonbolin): For multi-host, this needs to be consistent across hosts
-    os.environ.setdefault(xenv.PJRT_DIST_SERVICE_ADDR, '127.0.0.1:8547')
     global distributed_service
     if distributed_service is None:
       num_nodes = global_world_size
       distributed_service = torch_xla._XLAC._xla_get_distributed_runtime_service(
           num_nodes)
+      atexit.register(shutdown_distributed_runtime)
```
jonb377, line 36: Neat! So this will shutdown the distributed runtime when the rank-0 process exits?

I'm curious if this process somehow finishes first, will the shutdown block until all clients are destroyed?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1843,8 +1843,9 @@ void InitXlaModuleBindings(py::module m) {
                                   py::call_guard<py::gil_scoped_release>());
   m.def("_xla_get_distributed_runtime_service",
         [](int num_nodes) -> std::unique_ptr<xla::DistributedRuntimeService> {
+          std::string port = runtime::sys_util::GetEnvString("COORDINATOR_PORT", "8547");
           std::string dist_service_addr =
-              runtime::sys_util::GetEnvString("PJRT_DIST_SERVICE_ADDR", "");
```
jonb377, line 1847: How about we make the address and port parameters to `_xla_get_distributed_runtime_service`? I'll need this shortly for autocheckpointing, so happy to pick up the refactor after you land this PR.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -120,10 +135,10 @@ PjRtComputationClient::PjRtComputationClient() {
     TF_VLOG(1) << "Initializing PjRt GPU client...";
     bool async = sys_util::GetEnvBool(env::kEnvPjrtAsyncGpuClient, true);
     int local_rank = sys_util::GetEnvInt(env::kEnvPjRtLocalRank, 0);
-    std::string dist_service_addr =
-        sys_util::GetEnvString(env::kEnvPjrtDistServiceAddr, "");
+    
+    int global_rank = getGlobalRank(local_rank);
```
jonb377, line 139: How about `sys_util::GetEnvInt("RANK", local_rank);`, then we don't need the helper function

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -120,10 +135,10 @@ PjRtComputationClient::PjRtComputationClient() {
...
     auto distributed_client =
-        MaybeInitializeDistributedRuntimeClient(local_rank, dist_service_addr);
+        MaybeInitializeDistributedRuntimeClient(global_rank);
     auto allowed_devices =
         std::make_optional<std::set<int>>(std::set{local_rank});
```
jonb377, line 143: We need to generalize this to support `CUDA_VISIBLE_DEVICES` and single-process-multi-device

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1843,8 +1843,9 @@ void InitXlaModuleBindings(py::module m) {
                                   py::call_guard<py::gil_scoped_release>());
   m.def("_xla_get_distributed_runtime_service",
         [](int num_nodes) -> std::unique_ptr<xla::DistributedRuntimeService> {
+          std::string port = runtime::sys_util::GetEnvString("COORDINATOR_PORT", "8547");
```
jonb377, line 1846: Is `COORDINATOR_PORT` documented anywhere?

torch_xla/_internal/gpu.py
```diff
@@ -14,7 +15,8 @@ def num_local_processes() -> int:
   """
   assert xenv.GPU_NUM_DEVICES in os.environ, \
       "Must set `GPU_NUM_DEVICES` environment variable to use the PjRt GPU client"
-  return int(os.environ[xenv.GPU_NUM_DEVICES])
+  os.environ[xenv.WORLD_SIZE] = os.environ[xenv.GPU_NUM_DEVICES]
```
jonb377, line 18: Is this right? We'll clobber the torchrun-set world size. Also wondering if we need to keep `GPU_NUM_DEVICES` in the first place

torch_xla/_internal/pjrt.py
```diff
@@ -111,6 +118,11 @@ def initialize_multiprocess(local_rank: int, local_world_size: int):
     tpu.configure_topology(local_rank, local_world_size)
   elif runtime.device_type() == 'NEURON':
     neuron.initialize_env(local_rank)
+  elif runtime.device_type() == 'GPU':
+    global_world_size = xu.getenv_as('WORLD_SIZE', int)
+    global_rank = get_global_rank(local_rank)
```
jonb377, line 123: Also for consideration: `xu.getenv_as(xenv.RANK, int, local_rank)`

jonb377 (commented):
Looking great Xiongfei!

test/test_train_mp_imagenet_torchrun.py
vanbasten23: Only minor changes. In test_train_mp_imagenet.py, we have
```
if __name__ == '__main__':
  xmp.spawn(_mp_fn, args=(FLAGS,), nprocs=FLAGS.num_cores)
```
This file stops calling `spawn` and changes to:
```
if __name__ == '__main__':
  if not dist.is_torchelastic_launched():
    logging.error('Test must be launched with torchrun!')
    exit(1)
  _mp_fn(FLAGS)
```

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: Yeah, we initially plan to incorporate CUDA_VISIBLE_DEVICES into this PR but we encountered some errors such as https://github.com/pytorch/xla/issues/5558#issuecomment-1756482280 and https://github.com/pytorch/xla/issues/5558#issuecomment-1751574292. We still plan to do it but probably in a follow-up pr

torch_xla/_internal/gpu.py
vanbasten23: This change is mainly to make single-host case work with `spawn` (e.g. the tests in pytorch/xla/test/pjrt/test_runtime_gpu.py in which we use `spawn`). To provide a similar UX as PyTorch, we should still support `spawn` for the single host case (fwiw, PyTorch supports it as https://screenshot.googleplex.com/7nKD68dXNUUskF7). But I like your idea of replacing "GPU_NUM_DEVICES" with "WORLD_SIZE".

Also, torchrun doesn't invoke `spawn` and in turn this function so it wouldn't clobber the torchrun-set world size.

So how about I replace `GPU_NUM_DEVICES ` with `WORLD_SIZE `?

test/test_train_mp_imagenet_torchrun.py
will-cromar: Can you merge these two cases? ie if the program has been launched with torchrun, call `_mp_fn`, otherwise call `xmp.spawn`

torch_xla/_internal/gpu.py
will-cromar: `WORLD_SIZE` isn't quite right here, since this function returns the expected number of _local_ processes. In `torchrun`, that's set as `LOCAL_WORLD_SIZE`. In `xmp.spawn`, we had to set it as something different like `PJRT_LOCAL_WORLD_SIZE` because `LOCAL_WORLD_SIZE` caused some issue with a third party package.

So maybe `os.environ.get('LOCAL_WORLD_SIZE') or os.environ.get('PJRT_LOCAL_WORLD_SIZE')`? It's clunky, but it covers both cases.

To use a subset of local GPUs with `xmp.spawn`, a user could set then `LOCAL_WORLD_SIZE` themselves instead of `GPU_NUM_DEVICES`.

torch_xla/_internal/gpu.py
will-cromar: TIL this module exists. Cool!

doc for other reviewers who are curious https://docs.python.org/3/library/atexit.html

torch_xla/core/xla_env_vars.py
```diff
@@ -26,3 +26,5 @@
 PJRT_GPU_ASYNC_CLIENT = 'PJRT_GPU_ASYNC_CLIENT'
 PJRT_DIST_SERVICE_ADDR = 'PJRT_DIST_SERVICE_ADDR'
 LOCAL_RANK = 'LOCAL_RANK'
+RANK = 'RANK'
```
will-cromar, line 29: nit: are these constants provided by upstream `torch.distributed` somewhere?

torch_xla/csrc/init_python_bindings.cpp
will-cromar: I think this name needs to be more verbose, unfortunately. There are too many coordinators in our system already: TPU slice discovery, multislice discovery, torch.distributed process groups, and `xla::DistributedRuntimeService`. Maybe `XLA_COORDINATOR_PORT` and `XLA_COORDINATOR_ADDR`?

torch_xla/_internal/pjrt.py
```diff
@@ -101,6 +102,12 @@ def _run_singleprocess(fn: Callable[..., R], *args, **kwargs) -> Dict[int, R]:
 
   return fn(*args, **kwargs)
 
+def get_global_rank(local_rank: int):
```
will-cromar, line 105: This overlaps confusingly with the `global_ordinal` and the `process_index` in `torch_xla.runtime`, and this usage of the `RANK` env var or the local rank strikes me as a special case. `RANK` is not always set.

AFAICT, this index is used in one place: to determine whether to start the distributed runtime. How about something like `is_distributed_master(local_rank: int)` or `should_start_xla_coordinator(local_rank:int)` so we don't have another another index-getter?

cc @jonb377 for input since you're going to need to use this for checkpointing

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -39,17 +39,22 @@ static std::string spmd_device_str = "SPMD:0";
...
-    client = xla::GetDistributedRuntimeClient(dist_service_addr, options);
-    XLA_CHECK(client->Connect().ok())
-        << "Failed to initialize distributed runtime client";
+  int global_world_size = sys_util::GetEnvInt("WORLD_SIZE", 1);
+  if (global_world_size < 2) {
+    return std::move(client);
   }
+  std::string master_addr = sys_util::GetEnvString("MASTER_ADDR", "127.0.0.1");
+  std::string port = runtime::sys_util::GetEnvString("COORDINATOR_PORT", "8547");
+  std::string dist_service_addr = master_addr+":"+port ;
```
will-cromar, line 50: nit: can you use `absl::StrJoin` instead? https://abseil.io/docs/cpp/guides/strings#abslstrjoin-for-joining-elements-within-a-string

torch_xla/csrc/runtime/pjrt_computation_client.cc
will-cromar: It's time for me to bring up everyone's favorite topic: TPU v2/v3 multithreading :confetti_ball: 

Normally we could ignore this case entirely since we only needed the distributed runtime for GPUs. But, @jonb377 will be implementing it for TPUs as well, so we have to be more careful about the difference with process rank (shared between threads) and ordinal (unique to each thread) in this context.

Short term, can you make the naming more specific? ie `local_process_rank`, `global_process_rank`, and `getProcessRank` to emphasize these are process-scoped indices?

will-cromar (commented):
Great work! I'm really excited to see your progress here.

While this PR is in review, can you start updating the documentation for GPUs in this repository? We should have some documentation covering how/why the GPU runtime works and make it clear that we expect people to use `torchrun` for multihost use cases.

torch_xla/_internal/pjrt.py
vanbasten23: SG

torch_xla/_internal/pjrt.py
vanbasten23: This is handy. Thanks!
I replaced it with a function `should_initialize_dist_runtime` per a previous comment.

torch_xla/core/xla_env_vars.py
vanbasten23: I did a search but couldn't find it. A lot of places in upstream uses the string literals.

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: > Is COORDINATOR_PORT documented anywhere?

COORDINATOR_PORT is used to overwrite the default port for pjrt's dist runtime service. I added it in response to another comment. Before it, we initialize the dist runtime service at the port 8547 by default. But when this port is not available (the possibility may be low I assume), users wouldn't be able to start the service and would have to recompile torch_xla. In that case, users can overwrite the default port and unblock themselves.

> Is there ever a case where the torch.distributed master will be different than the distributed runtime master?

So far, torch.distributed master and pjrt's dist runtime master can share the same address and use different ports without any issues. So I'll rename COORDINATOR_PORT to XLA_COORDINATOR_PORT.

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: Thanks!

test/pjrt/test_torchrun.py
vanbasten23: done

test/pjrt/test_torchrun.py
vanbasten23: done

test/test_train_mp_imagenet_torchrun.py
vanbasten23: done

torch_xla/_internal/pjrt.py
vanbasten23: done

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: done

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: done

torch_xla/_internal/gpu.py
vanbasten23: Agreed that `WORLD_SIZE` is confusing. OTOH, `WORLD_SIZE` is what PyTorch single-node training uses: https://screenshot.googleplex.com/BH27HAYTpbNU4KA, if we want to stay closer to PyTorch.

I think clarity is more important here. How do you think we replace `GPU_NUM_DEVICES` with `PJRT_LOCAL_WORLD_SIZE` (or something like `PT_XLA_LOCAL_WORLD_SIZE` if we don't want to leak the underlying implementation detail), so the way we run single-host-multi-GPU `GPU_NUM_DEVICES=4 python3 xla/test/test_train_mp_imagenet.py ...` turns to `PJRT_LOCAL_WORLD_SIZE=4 python3 xla/test/test_train_mp_imagenet.p ..`?  @will-cromar @jonb377

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: per our discussion offline, let's push it down to the cpp layer

torch_xla/_internal/gpu.py
will-cromar: I think it's okay to rely on the `LOCAL_WORLD_SIZE` env var here. We know torchrun will set it for sure, and the manual single-host-multi-GPU case can become `PJRT_DEVICE=GPU LOCAL_WORLD_SIZE=4 python script.py`*. We can get rid of `GPU_NUM_DEVICES`.

\* As a follow up, I would like to implement more automatic configuration like we have with TPUs so users don't have to set anything in the default case

We definitely don't want to override `torchrun`'s settings here.

I forgot that `PJRT_LOCAL_WORLD_SIZE` is set after this function is called (and probably just set to the output of this function). So we can ignore that variable here.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -143,26 +151,27 @@ PjRtComputationClient::PjRtComputationClient() {
...
+    int global_world_size = sys_util::GetEnvInt("WORLD_SIZE", 1);
+    TF_VLOG(3) << "Getting StreamExecutorGpuClient for node_id="
+               << global_process_rank << ", num_nodes=" << global_world_size;
+    client_ = std::move(xla::GetStreamExecutorGpuClient(
+                            /*asynchronous=*/async, xla::GpuAllocatorConfig{},
+                            /*node_id=*/global_process_rank,
+                            /*num_nodes=*/
+                            global_world_size,
+                            /*allowed_devices=*/allowed_devices,
+                            /*platform_name*/ "gpu",
```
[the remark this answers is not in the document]
vanbasten23: done

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1922,22 +1922,26 @@ void InitXlaModuleBindings(py::module m) {
...
-                  .value());
-        });
+  m.def(
+      "_xla_get_distributed_runtime_service",
+      [](int num_nodes) -> std::unique_ptr<xla::DistributedRuntimeService> {
+        std::string port =
+            runtime::sys_util::GetEnvString("XLA_COORDINATOR_PORT", "8547");
+        std::string LOCAL_HOST_IP_ADDR = "127.0.0.1";
+        std::string dist_service_addr =
+            runtime::sys_util::GetEnvString("MASTER_ADDR", LOCAL_HOST_IP_ADDR) +
```
will-cromar, line 1932: nit: does this have to be an IP? Otherwise, you can just hardcode `"localhost"` as the default

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1922,22 +1922,26 @@ void InitXlaModuleBindings(py::module m) {
...
+            ":" + port;
+        XLA_CHECK(!dist_service_addr.empty())
+            << "Must set PJRT_DIST_SERVICE_ADDR environment variable to use "
```
will-cromar, line 1935: nit: update the error message to `MASTER_ADDR` too

torch_xla/_internal/pjrt.py
vanbasten23: Yeah, it's gone in the latest iteration.

torch_xla/_internal/gpu.py
vanbasten23: Yeah, I used `LOCAL_WORLD_SIZE ` here.

I'll replace `GPU_NUM_DEVICES` with `LOCAL_WORLD_SIZE` in a follow-up PR.

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: yeah, "localhost" is better!

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: done

torch_xla/csrc/init_python_bindings.cpp
jonb377: Do you plan to take this in a follow-up PR? If so, I can drive it to unblock autocheckpointing, the GPU-related changes to the distributed runtime seem complete (feel free to let me know otherwise).

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1922,22 +1922,22 @@ void InitXlaModuleBindings(py::module m) {
...
-              xla::GetDistributedRuntimeService(dist_service_addr, options)
-                  .value());
-        });
+  m.def(
+      "_xla_get_distributed_runtime_service",
+      [](int num_nodes) -> std::unique_ptr<xla::DistributedRuntimeService> {
+        std::string master_addr =
+            runtime::sys_util::GetEnvString("MASTER_ADDR", "localhost");
+        std::string port =
+            runtime::sys_util::GetEnvString("XLA_COORDINATOR_PORT", "8547");
```
jonb377, line 1931: I'm concerned about hard-coding the default value in two places, can we make this a const somewhere? Or if you want, this can be addressed in the follow-up to move everything into C++.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -144,26 +151,28 @@ PjRtComputationClient::PjRtComputationClient() {
...
-                      /*num_nodes=*/
-                      sys_util::GetEnvInt(env::kEnvPjRtLocalProcessCount, 1),
-                      /*allowed_devices=*/allowed_devices,
-                      /*platform_name*/ "gpu",
-                      /*should_stage_host_to_device_transfers*/ true,
-                      /*kv_get*/ kv_get,
-                      /*kv_put*/ kv_put)
-                      .value());
+    int global_world_size = sys_util::GetEnvInt(
+        "WORLD_SIZE", sys_util::GetEnvInt("LOCAL_WORLD_SIZE", 1));
```
jonb377, line 155: nit: Same here, splitting into two lines would be a bit cleaner.

jonb377 (approved):
LGTM, thanks Xiongfei!

torch_xla/_internal/pjrt.py
```diff
@@ -111,6 +120,12 @@ def initialize_multiprocess(local_rank: int, local_world_size: int):
     tpu.configure_topology(local_rank, local_world_size)
   elif runtime.device_type() == 'NEURON':
     neuron.initialize_env(local_rank)
+  elif runtime.device_type() in ('GPU', 'ROCM', 'CUDA'):
+    global_world_size = xu.getenv_as(
+        xenv.WORLD_SIZE, int, xu.getenv_as(xenv.LOCAL_WORLD_SIZE, int, 1))
```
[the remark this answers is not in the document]
vanbasten23: will resolve in a follow-up PR.

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: Yes, it's in https://github.com/pytorch/xla/pull/5714.

torch_xla/_internal/pjrt.py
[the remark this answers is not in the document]
vanbasten23: it's in https://github.com/pytorch/xla/pull/5714

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: Fair. Addressed it in https://github.com/pytorch/xla/pull/5714

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: done in https://github.com/pytorch/xla/pull/5714

PR #5662, opened 2023-09-29
Print dynamo fallback ops for pt_xla_debug flag

JackCaoG:
`PT_XLA_DEBUG` should capture the dynamo fallback message as well.


alanwaketan (commented):
Sorry for late review, but it looks good to me.

PR #5669, opened 2023-10-04
Update `setup.py` and README for r2.1 release.

will-cromar:
- Backport build metadata and README changes from release branch
- Attempt to clean up various build lists
- Bury old versions in `<details>` to make the lists less overwhelming


README.md
```diff
@@ -204,53 +217,58 @@ pip3 install torch_xla[tpuvm]
...
 
-| Version | GPU CUDA 11.2 + Python 3.7 Docker |
+| Version | GPU CUDA 11.2 + Docker |
 | --- | ----------- |
-1.13 | `gcr.io/tpu-pytorch/xla:r1.13_3.7_cuda_11.2` |
-1.12 | `gcr.io/tpu-pytorch/xla:r1.12_3.7_cuda_11.2` |
+| 1.13 | `gcr.io/tpu-pytorch/xla:r1.13_3.7_cuda_11.2` |
+| 1.12 | `gcr.io/tpu-pytorch/xla:r1.12_3.7_cuda_11.2` |
 
+</details>
```
JackCaoG, line 271: There is a place in README mentions `pip install torch_xla[tpuvm]`, can you add notes saying that only works for release prior to 2.0? for nightly do we use `[tpuvm]` or `[tpu]`?

will-cromar: I left `tpuvm` on master (see comment in `setup.py`) so I don't break people's scripts. We should remove it eventually, so I'd rather leave this as an undocumented "feature".

The instruction in this doc already says "before PyTorch/XLA 2.0"

JackCaoG (approved):

will-cromar:
Since I touched `setup.py`, I'll wait until the build passes before I merge.

ManfeiBai (approved):
LGTM

README.md
```diff
@@ -130,26 +132,37 @@ Our comprehensive user guides are available at:
...
+
+GPU, XRT (legacy runtime), and nightly builds are available in our public GCS
+bucket.
 
 | Version | Cloud TPU VMs Wheel |
 | --- | ----------- |
-| 2.0 (Python 3.8) | `https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch_xla-2.0-cp38-cp38-linux_x86_64.whl` |
-| nightly >= 2023/04/25 (Python 3.8) | `https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-nightly-cp38-cp38-linux_x86_64.whl` |
-| nightly >= 2023/04/25 (Python 3.10) | `https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-nightly-cp310-cp310-linux_x86_64.whl` |
+| 2.1 (CUDA 12.0 + Python 3.8) | `https://storage.googleapis.com/pytorch-xla-releases/wheels/cuda/12.0/torch_xla-2.1.0-cp38-cp38-manylinux_2_28_x86_64.whl` |
```
vanbasten23, line 150: pytorch uses cuda 12.1 instead of 12.0: `pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/test/cu121`. Do we want to do the same?

setup.py
```diff
@@ -334,12 +356,12 @@ def run(self):
     },
     extras_require={
         # On Cloud TPU VM install with:
-        # $ sudo pip3 install torch_xla[tpuvm] -f https://storage.googleapis.com/tpu-pytorch/wheels/tpuvm/torch_xla-1.11-cp38-cp38-linux_x86_64.whl
+        # pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
```
vanbasten23, line 359: What's the difference between `pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html` and `pip install torch_xla[tpuvm]`? IOW, under what circumstance should I use the former versus the latter?

README.md
will-cromar: Oh interesting. I just went off of what we had built. Do you know if our 12.0 wheel is compatible with their 12.1 wheel? Or do we need to add a new build config for the 2.0 release?

setup.py
will-cromar: The technical difference is that `tpuvm` refers to a private package repository directly. That's why you don't have to pass an `--extra-index`. PyPI disallows this, so with release wheels starting with 2.1, you have to use `tpu`, which specifies a libtpu version but not the location. The location comes from the index url.

I left `tpuvm` on master/nightly for now since there's no rush to remove it, but I think we might want to anyway to reduce confusion.

If you're not sure, just use the `tpu` extra dependency since we will have to support that going forward.

vanbasten23: Got it. Thanks!

README.md
vanbasten23: > Do you know if our 12.0 wheel is compatible with their 12.1 wheel?

If we do `pip3 install torch torchvision torchaudio
`, it'll install cuda 12.1 by default per https://screenshot.googleplex.com/BDFoq3C7TXiWeXn. I can do some testing.

But before that, we do need to add 2 new build config for the 2.1 release, one for cuda12.1+py3.8 and another one for cuda12.1+py3.10. Once we get the wheel and validate that it is compatible with pytorch's, we can update our GPU doc. Wdyt?

will-cromar: Sounds good to me. Thanks for catching the issue.

vanbasten23: hi @ManfeiBai , it seems we need cuda 12.1 wheels instead of 12.0 because by default PyTorch uses cuda 12.1. Can you please generate a py=3.8 cuda=12.1 wheel and a py=3.10 cuda=12.1 wheel for torch_xla release 2.1?

ManfeiBai: PR for CUDA12.1 wheel: https://github.com/pytorch/xla/pull/5683

PR #5671, opened 2023-10-04
Workaround the crash in TF (#4224)

JackCaoG:
this will resolve the pod crash during init for xrt, verified locally that patch can be applied.


will-cromar (approved):

JackCaoG:
I am going to wait for the build to finish before merging

JackCaoG:
GPU test failure is not related to this pr. merging.

PR #5673, opened 2023-10-04
change `master` to `main`

ManfeiBai:
after PyTorch/XLA 2.1 release, plan to change default branch name from `master` to `main`
[x] change `/master/` to `/main/`
[x] change `master` to `main`
[] chaneg `https://pytorch.org/xla/master/` to `https://pytorch.org/xla/main/`


will-cromar:
@mateuszlewko can correct me if I'm wrong, but you actually want to merge this into master. The existing build triggers will be watching master until this PR goes through

ManfeiBai:
> @mateuszlewko can correct me if I'm wrong, but you actually want to merge this into master. The existing Terraform deployment triggers will be watching master until this PR goes through

if we change default branch from master to main, we will use `main` as the default branch, and yes, @mateuszlewko will know the detail about how the Terraform deployment triggers will watching, will wait for @mateuszlewko's reply too

mateuszlewko:
> @mateuszlewko can correct me if I'm wrong, but you actually want to merge this into master. The existing Terraform deployment triggers will be watching master until this PR goes through

Yes, that's correct.

The triggers in infra/tpu-pytorch-releases/infra_triggers.tf and  infra/tpu-pytorch/infra_triggers.tf are watching the `master` branch.
So if we only merge this change to the main branch, nothing will happen.
If we first merge this into `master` branch, the existing trigger in GCP that watches the `master`branch will update itself (and other dev / nightly triggers) to watch the `main` branch.  Then (5 minutes later should be enough), merge the same changes into the `main` branch. The infra trigger will watch the `main` branch changes and will update itself, as well as other trigger, but the change should be a no-op.

ManfeiBai:
Thanks for you clarify, Mateusz!

If you gonna start moving master to main, do you want to continue in this PR or create a new PR? I have added you permission to modify https://github.com/ManfeiBai/PyTorchXLA, and please ping me directly if we failed to push code to this PR too

mateuszlewko (approved):
LGTM for the infra changes, but please also get an approval from one other person.

PR #5674, opened 2023-10-04
Add fsdp tests back to GPU CI

vanbasten23:
fsdp test was broken in one of the tf pin updates. 
This pr tests if it's possible to reenable it in the GPU CI and add it back if so.


alanwaketan (approved):
LGTM. Do we have a bug to track the memory regression?

JackCaoG (approved):

vanbasten23:
> LGTM. Do we have a bug to track the memory regression?

Just cc'ed you in the bug. Thanks Jiewen!

vanbasten23:
Thanks for the review!

PR #5677, opened 2023-10-05
IFRT prototype

will-cromar:
- Implement new `ComputationClient` with IFRT, which currently just wraps PJRT.
  -  Enable IFRT wrapper with `XLA_USE_IFRT=1`
- Move PJRT client initialization into `initialize_pjrt.cc/h` since IFRT wraps the same `PjRtClient`.
- Minor fixes to existing `pjrt_computation_client`: 
  - Make `spmd_device_str` `const`
  - Change the owned `PjRtClient` to a `unique_ptr`. Only SE:TPU required us to use `shared_ptr`, which is now removed.

There's still some opportunities to refactor common functionality up to `ComputationClient`, but I'm trying to minimize changes to the high-level API for this PR.

IFRT is still highly experimental. Use at your own risk. See [my comments](https://github.com/pytorch/xla/pull/5677#issuecomment-1834935876) below for caveats and limitations.


torch_xla/csrc/runtime/ifrt_computation_client.h
```diff
@@ -0,0 +1,249 @@
...
+  std::string PjRtDeviceToString(xla::PjRtDevice* const device) const;
+  std::vector<std::string> PjRtDevicesToString(
+      absl::Span<xla::PjRtDevice* const> devices) const;
+
+  struct IfrtData : public Data {
+    IfrtData(std::string device, xla::Shape device_shape)
+        : Data(std::move(device), std::move(device_shape)) {}
+
+    IfrtData(std::string device, xla::Shape device_shape,
+             tsl::RCReference<xla::ifrt::Array> buffer,
```
yeounoh, line 127: Some pointer to verify -- We are wrapping `xla::ifrt::pjrt_client`, and that probably means that you want to work with the same `PjRtBuffer` instead of the `ifrt::Array` [link](http://google3/third_party/tensorflow/compiler/xla/python/pjrt_ifrt/pjrt_client.h;l=43;rcl=558837006). 

I would think we should re-use `PjRtData` and `PjRtShardedData` as-is, and later add a new data type `IfRtArray`, when it's supported in `ifrt::pjrt_client` ?

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -313,6 +314,10 @@ ShardingUtil::InputHandler(
     std::vector<std::string> devices) {
   tsl::profiler::TraceMe activity("InputHandler",
                                   tsl::profiler::TraceMeLevel::kInfo);
+  // TODO: use vector<DataPtr> directly since DataPtr is already sharded
```
yeounoh, line 317: Wait, I think that it would be nicer to have all the changes in the runtime wrapper, and PyTorch/XLA still calls the same API, work with the same data prepresentation for SPMD with the same sementic. I.e., have the same-level of abstraction in our wrapper client, so the underlying details can be hidden (DataPtr goes to a single device buffer) or exposed (ArrayPtr goes to block across devices), instead of asking to assume DataPtr is now sharded or not shsarded. We really shoulnd't reimplement PyTorch/XLA for this, IFRT was designed to support that.

torch_xla/csrc/runtime/ifrt_computation_client.h
yeounoh: Actually, I see array is already being supported in the client there.

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: PjRtShardedData abstraction (or call it IfRtShardedData) stores shards as Array, which used to be a vector<PjRtData>. But the abstraction and its level-of-abstraction stays the same, and our implementation still oeprates the same.

will-cromar: In this case, we should change the signature of `ExecuteReplicated` to accept one vector of sharded tensors. Any assembling/disassembling can happen in the computation client.

I'll separate any ComputationClient API changes into a new PR to ensure they work with PJRT.

will-cromar:
At this point, I am able to run our SPMD ResNet50 example successfully, but it's extremely slow. Any case where I resharded an array, I chose "copy" semantics for safety. I'll have to take another pass and carefully think through the ownership of the underlying data, since the excessive copies are likely contributing to poor performance.

Sharded execution does not appear to be implemented at all within the IFRT/PJRT wrapper, so I marked that method unimplemented for now. Likewise, I `xla::ifrt::HloSharding` (a wrapper over `xla::HloSharding`) is not implemented in the wrapper, so we'll have to carry the `xla::OpSharding` in our own object for now.

Dynamic shape is currently unsupported by `xla::ifrt::Shape`. 

Until we have feature parity, I will keep IFRT as a separate `ComputationClient` implementation so we don't break PJRT.

will-cromar:
Performance on LLama 7B is not bad! It's somewhere between PJRT now and PJRT before I started working on some optimizations this month:

```
Totally decoded 1007 tokens in 7.03615 seconds
```

JackCaoG:
If you don't intend to merge this for 2.2 release, I will hold on the review until the branch cut.

will-cromar:
Merging after the cut sounds good to me. This won't be useful in the 2.2 release.

JackCaoG:
I will take a look today

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+#include "xla/shape.h"
+
+using xla::internal::XlaBuilderFriend;
+
+namespace torch_xla {
+namespace runtime {
+
+namespace {
+
+static const std::string spmd_device_str = "SPMD:0";
```
JackCaoG, line 41: lol we should make this a attribute or enum or something in computation client later.

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+
+// Initializes a distributed runtime client if dist_service_addr is specified
```
JackCaoG, line 43: is this for GPU?

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+    client = xla::GetDistributedRuntimeClient(dist_service_addr, options);
+    XLA_CHECK(client->Connect().ok())
+        << "Failed to initialize distributed runtime client";
+  }
+  return std::move(client);
+}
+
+// Builds a map from the device's global ordinal to its index in the `devices`
+// array.
+std::unordered_map<int, int> build_index_map(
```
JackCaoG, line 61: for these utils, can we share between pjrt and ifrt? so they are actually different?

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+        << "Invalid device specification: " << devices[i];
+    int global_ordinal = std::stoi(device_spec[1]);
+    device_index[global_ordinal] = i;
+  }
+  return device_index;
+}
+
+}  // namespace
+
+std::string IfrtComputationClient::PjRtDeviceToString(
```
JackCaoG, line 76: nit Ifrt

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+    xla::PjRtDevice* const device) const {
+  std::string platform =
+      absl::AsciiStrToUpper(device->client()->platform_name());
+  int ordinal = global_ordinals_.at(device->id());
+  std::string str = absl::StrFormat("%s:%d", platform, ordinal);
+  return str;
+}
+
+std::vector<std::string> IfrtComputationClient::PjRtDevicesToString(
```
JackCaoG, line 85: ditto

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+
+  for (auto* device : devices) {
+    strs.push_back(PjRtDeviceToString(device));
+  }
+
+  return strs;
+}
+
+IfrtComputationClient::IfrtComputationClient() {
+  std::string device_type = sys_util::GetEnvString(env::kEnvPjRtDevice, "");
```
JackCaoG, line 98: let's do a grep of `pjrt` in this file and replace them with `ifrt`. Through I am curious did you intend to query the EnvPjrtDevice here?

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+  tsl::profiler::TraceMe activity("IfrtComputationClient::GetDataShard",
+                                  tsl::profiler::TraceMeLevel::kInfo);
+  return GetDataShards(data)[index];
+}
+
+ComputationClient::DataPtr IfrtComputationClient::WrapDataShards(
+    const std::vector<DataPtr>& shards, std::string device, xla::Shape shape,
+    xla::OpSharding sharding) {
+  // TODO: implement CreateDataPlaceholder for sharded data
+  if (shards.size() == 0) {
```
JackCaoG, line 201: is there a legit use case of `shards.zie() == 0`?

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+  return std::make_shared<IfrtData>(device, shape, sharded_array, sharding);
+}
+
+std::optional<xla::OpSharding> IfrtComputationClient::GetDataSharding(
+    DataPtr handle) {
+  auto ifrt_data = std::dynamic_pointer_cast<IfrtData>(handle);
+  return ifrt_data->sharding_;
+}
+
+std::vector<ComputationClient::DataPtr> IfrtComputationClient::TransferToServer(
```
JackCaoG, line 242: lol we should really rename this to TransferToDevice.

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+    absl::Span<const std::shared_ptr<const TensorSource>> tensors) {
+  metrics::TimedSection timed(TransferToServerMetric());
```
JackCaoG, line 244: is transfer async? If so this metric won't capture the real device transfer time right?

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,643 @@
...
+  OutboundDataMetric()->AddSample(total_size);
+  CreateDataHandlesCounter()->AddValue(datas.size());
+
+  return datas;
+}
+
+ComputationClient::DataPtr IfrtComputationClient::TransferShardsToServer(
+    absl::Span<const std::shared_ptr<const TensorSource>> tensor_shards,
+    std::string device, xla::Shape shape, xla::OpSharding sharding) {
+  tsl::profiler::TraceMe activity(
```
JackCaoG, line 282: ditto, is transfer async?

torch_xla/csrc/runtime/ifrt_computation_client.h
```diff
@@ -0,0 +1,210 @@
...
+                         xla::Shape shape, xla::OpSharding sharding) override;
+
+  std::optional<xla::OpSharding> GetDataSharding(DataPtr handle) override;
+
+  std::vector<DataPtr> TransferToServer(
+      absl::Span<const std::shared_ptr<const TensorSource>> tensors) override;
+
+  // Use XLA replication to re-assemble the sharded data.
+  // DataPtr ReplicateShardedData(const DataPtr& handle);
+
```
JackCaoG, line 48: remove empty line maybe

torch_xla/csrc/runtime/runtime.cc
```diff
@@ -19,10 +20,14 @@ ComputationClient* CreateClient() {
     tsl::testing::InstallStacktraceHandler();
   }
 
+  static bool use_ifrt = sys_util::GetEnvBool("XLA_USE_IFRT", false);
```
JackCaoG, line 23: why not just `IFRT_DEVICE`? Do you intend to hide this under the PJRT env var?

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: This was intentional. I wanted to minimize changes to common code, and ideally we would be able to remove the PJRT computation client at some point. Would you rather I try to factor out all common functionality in this PR?

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: We should get rid of device strings from our code base entirely

torch_xla/csrc/runtime/runtime.cc
will-cromar: We still use `PJRT_DEVICE` for the "IFRT wraps PJRT" case, since we're initializing exactly the same client. This is just a switch to enable or disable the IFRT wrapper.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: Agreed. I'll follow up after this PR.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: This is how sharded data placeholders get created right now:

https://github.com/pytorch/xla/blob/a2f80e466aedd7f1457bda75e9beea9c76509f1b/torch_xla/csrc/xla_sharding_util.cpp#L590-L593

I'd rather update `CreateDataPlaceholder` to take a sharding and make it more explicit

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: Oops! Done. Also updated a bunch of aliases to use the IFRT name.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: Yeah, see https://github.com/pytorch/xla/pull/5677#discussion_r1425871830

torch_xla/csrc/runtime/ifrt_computation_client.cc
jonb377: @will-cromar This can be removed - the logic is encapsulated in XlaCoordinator now.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: This one is synchronous

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: Fixed

torch_xla/csrc/runtime/ifrt_computation_client.h
will-cromar: Done.

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,678 @@
...
+    std::vector<std::string> device_spec = absl::StrSplit(devices[i], ':');
+    XLA_CHECK_EQ(device_spec.size(), 2)
+        << "Invalid device specification: " << devices[i];
+    int global_ordinal = std::stoi(device_spec[1]);
+    device_index[global_ordinal] = i;
+  }
+  return device_index;
+}
+
+torch::lazy::hash_t hash_comp_env(
```
jonb377, line 57: Should we make this shared between the two ComputationClients? If we need to update the env hashing logic, it will need to apply in both places.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: Oops. This one should have been removed during rebasing.

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,678 @@
...
+    ComputationClient::DataPtr data) {
+  tsl::profiler::TraceMe activity("IfrtComputationClient::GetDataShards",
+                                  tsl::profiler::TraceMeLevel::kInfo);
+  std::vector<ComputationClient::DataPtr> shards;
+  if (data->HasSharding()) {
+    auto ifrt_data = std::dynamic_pointer_cast<IfrtData>(data);
+    std::vector<tsl::RCReference<xla::ifrt::Array>> arrays =
+        ifrt_data->buffer
+            ->DisassembleIntoSingleDeviceArrays(
+                xla::ifrt::ArrayCopySemantics::kAlwaysCopy)
```
jonb377, line 195: Why do we need to copy here?

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,678 @@
...
+  //   xla::ifrt::MemoryKind(),
+  //   xla::HloSharding::FromProto(sharding).value()
+  // );
+  tsl::RCReference<xla::ifrt::Array> sharded_array =
+      client_
+          ->AssembleArrayFromSingleDeviceArrays(
+              ifrt_shape, std::move(ifrt_sharding), absl::MakeSpan(arrays),
+              xla::ifrt::ArrayCopySemantics::kAlwaysCopy)
+          .value();
+  return std::make_shared<IfrtData>(device, shape, sharded_array, sharding);
```
jonb377, line 251: Is `device` always expected to be `SPMD:0`?

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,678 @@
...
+
+std::vector<ComputationClient::DataPtr> IfrtComputationClient::TransferToServer(
+    absl::Span<const std::shared_ptr<const TensorSource>> tensors) {
+  auto timed = std::make_shared<metrics::TimedSection>(TransferToServerMetric());
+  tsl::profiler::TraceMe activity("IfrtComputationClient::TransferToServer",
+                                  tsl::profiler::TraceMeLevel::kInfo);
+  std::vector<ComputationClient::DataPtr> datas;
+  datas.reserve(tensors.size());
+  int64_t total_size = 0;
+  for (auto& tensor : tensors) {
```
jonb377, line 269: No need to address in this PR, just curious: would this benefit from distributing the transfers over the thread pool?

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: Moved this constant to `protected` section of `ComputationClient` so both implementations can use it.

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,678 @@
...
+          .value();
+  return std::make_shared<IfrtData>(device, shape, sharded_array, sharding);
+}
+
+ComputationClient::DataPtr IfrtComputationClient::CopyToDevice(
+    ComputationClient::DataPtr data, std::string dst) {
+  XLA_ERROR() << __FUNCTION__ << " not implemented";
+}
+
+tsl::RCReference<xla::ifrt::Array> IfrtComputationClient::ReplicateShardedData(
```
jonb377, line 344: Do we still need this API in Ifrt or could we `CopyToHostBuffer` directly on an Array backed by sharded data?

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -0,0 +1,678 @@
...
+
+    // TODO: handle dynamic shapes
+    auto& literal = literals.emplace_back(
+        xla::ShapeUtil::DeviceShapeToHostShape(ifrt_data->shape()));
+    std::vector<int64_t> byte_strides(literal.shape().dimensions_size());
+    XLA_CHECK_OK(xla::ShapeUtil::ByteStrides(literal.shape(),
+                                             absl::MakeSpan(byte_strides)));
+    XLA_CHECK_OK(
+        replicated_array
+            ->CopyToHostBuffer(literal.untyped_data(), byte_strides,
```
jonb377, line 422: Follow-up: we'll need to parallelize this if possible for async checkpointing performance.

torch_xla/csrc/runtime/ifrt_computation_client.h
```diff
@@ -0,0 +1,222 @@
...
+  std::vector<std::string> IfrtDevicesToString(
+      absl::Span<xla::ifrt::Device* const> devices) const;
+
+  struct IfrtData : public Data {
+    IfrtData(std::string device, xla::Shape device_shape)
+        : Data(std::move(device), std::move(device_shape)) {}
+
+    IfrtData(std::string device, xla::Shape device_shape,
+             tsl::RCReference<xla::ifrt::Array> buffer,
+             std::optional<xla::OpSharding> sharding = std::nullopt)
```
jonb377, line 143: I wonder if IfrtData needs to track the sharding, or can we just pull it from the underlying Array? Same for device_shape.

I assume there may be some overhead reading these off of the array, so it may be better to track these separately.

jonb377 (approved):
Nice one Will! This is awesome.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: It doesn't quite work since the client types are actually different.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: I'm going for safety over performance with this PR. We'll have to go back over this with a careful eye for Array ownership to get the best performance. PJRT let us ignore ownership because aliasing was functionally impossible, but it's very easy to alias underlying data in IFRT.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: I think so. We can realistically remove the `device` argument from `ComputationClient::WrapDataShards` if we are always placing data on the `SPMD` "device"

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: Almost certainly. I'll work on performance some more in follow up PRs.

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: > could we CopyToHostBuffer directly on an Array backed by sharded data?

IIUC this will not work with Arrays sharded over non-addressable devices

torch_xla/csrc/runtime/ifrt_computation_client.cc
will-cromar: Agreed

torch_xla/csrc/runtime/ifrt_computation_client.h
will-cromar: > I wonder if IfrtData needs to track the sharding, or can we just pull it from the underlying Array? Same for device_shape.

So, the IFRT array does carry its own sharding, but it's in IFRT's format. There is an IFRT wrapper over `xla::HloSharding` (aka the non-proto version of `xla::OpSharding`), but I was running into crashes while trying to use it on our input data. Also, IFRT does not guarantee that the arrays returned after execution specifically have the `HloSharding` type.

This is a place where the IFRT abstraction doesn't really work for us, since we rely on the `xla::OpSharding`/`xla::HloSharding` for building HLOs. To use `xla::ifrt::Array`'s sharding full-time, we'd have to update our code to operate on `xla::ifrt::Sharding`. For now, I'm recycling most of our existing logic that we used for `PjRtShardedData`'s sharding.

torch_xla/csrc/runtime/ifrt_computation_client.cc
jonb377: Oh that's interesting - thanks for the clarification, I had assumed Arrays represented the global array.

will-cromar: Arrays can represent _either_ global arrays  or shards of a global array. You can even have both a global and local array pointing to the same underlying data.

JackCaoG (approved):

PR #5688, opened 2023-10-09
Fix `masked_fill` broadcasting.

ysiraichi:
This PR fixes tensor broadcasting for `masked_fill(input, mask, scalar)`. Specifically, when they have the same rank.

**Problem:** when both `input` and `mask` were same rank tensors, say `[2, 2]` and `[2, 1]` respectively, we always expanded the `input`.

**Solution:** inlined part of the code of `XlaHelpers::PromoteShapes` (without the type comparison).

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -309,6 +309,19 @@ class TestSelect(test_utils.XlaTestCase):
     tx = t.select(1, 12)
     self.assertEqual(tx, sx.data.cpu())
 
+  def test_masked_fill_scalar(self):
+
+    def fn(tensor):
+      # Build a mask from the first line of tensor.
+      # Also, make it have the same rank as the original tensor.
+      mask = tensor[0].ge(0.5).unsqueeze(dim=0)
+      # Call masked_fill.
+      return tensor.masked_fill(mask, 10)
+
+    x = _gen_tensor(2, 2, device=xm.xla_device())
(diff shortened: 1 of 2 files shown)


JackCaoG:
@wonjoolee95 can you take a look?

wonjoolee95 (approved):
Thanks for the catch, @ysiraichi!

test/test_operations.py
```diff
@@ -309,6 +309,19 @@ def test_get_xla_tensor(self):
     tx = t.select(1, 12)
     self.assertEqual(tx, sx.data.cpu())
 
+  def test_masked_fill_scalar(self):
+
+    def fn(tensor):
+      # Build a mask from the first line of tensor.
+      # Also, make it have the same rank as the original tensor.
+      mask = tensor[0].ge(0.5).unsqueeze(dim=0)
```
vanbasten23, line 317: I can see the problem (the test would crash before the change) but confused about the PR description, it says the problem happens when we have 2 tensors with the same rank. But here, it seems mask's shape is `torch.Size([1, 2, 2])` and tensor's shape is torch.Size([2, 2]).

ysiraichi: `tensor` is indeed [2, 2]. However, `tensor[0]` is [2]. Which makes mask `[1, 2]` after the `unsqueeze` call.

PR #5689, opened 2023-10-09
[WIP] Filter tensor arguments from traced model.

ysiraichi:
This PR filters tensor arguments, when collecting tensor information, from the list of arguments that would be given to the model.

**Problem:** dynamo bridge assumed all arguments were tensors. 
**Solution:** filter tensor arguments so that we correctly collect tensor information.

### DIFF
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -230,8 +230,12 @@ def extract_graph_helper(xla_model: torch.fx.GraphModule):
       for xla_arg in xla_args
   ]
 
+  xla_tensor_args = [
+      (i, xla_arg) for i, xla_arg in enumerate(xla_args) if isinstance(xla_arg, torch.Tensor)
+  ]
+
   args_tensor_ids = [
-      torch_xla._XLAC._xla_get_tensor_id(xla_arg) for xla_arg in xla_args
+      (index, torch_xla._XLAC._xla_get_tensor_id(xla_arg)) for index, xla_arg in xla_tensor_args
   ]
 
   if dynamo_debug:
@@ -239,7 +243,7 @@ def extract_graph_helper(xla_model: torch.fx.GraphModule):
     print(f"args_tensor_ids {args_tensor_ids}")
 
   tensor_id_to_arg_idx = {
-      tensor_id: i for i, tensor_id in enumerate(args_tensor_ids)
+      tensor_id: index for index, tensor_id in args_tensor_ids
   }
 
   if xr.is_spmd():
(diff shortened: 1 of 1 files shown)


JackCaoG:
Thanks, can you add a test cast that would fail without this change?

test/dynamo/test_bridge.py
```diff
@@ -207,6 +207,24 @@ class TorchXLAReuseGraphTest(torch._dynamo.test_case.TestCase):
...
+    class Emb(torch.nn.Embedding):
+
+      def __init__(self):
+        super().__init__(num_embeddings=10, embedding_dim=10, padding_idx=0)
+
+    device = xm.xla_device()
+    module = Emb()
+    module.to(device)
+
+    @torch.compile(backend="openxla")
```
ysiraichi, line 221: @JackCaoG replaced `openxla_eval` by `openxla`.

torch_xla/core/dynamo_bridge.py
```diff
@@ -230,16 +230,19 @@ def extract_graph_helper(xla_model: torch.fx.GraphModule):
       for xla_arg in xla_args
   ]
 
-  args_tensor_ids = [
-      torch_xla._XLAC._xla_get_tensor_id(xla_arg) for xla_arg in xla_args
-  ]
+  xla_tensor_args = [(i, xla_arg)
```
JackCaoG, line 233: naming is a bit misleading here, it is more of `arg_idx_and_xla_tensor_args`... open to other suggestions but I think we wan to make it clear that it is a list of `[idx, xla_arg]` tuple

torch_xla/core/dynamo_bridge.py
```diff
@@ -230,16 +230,19 @@ def extract_graph_helper(xla_model: torch.fx.GraphModule):
...
+                     for i, xla_arg in enumerate(xla_args)
+                     if isinstance(xla_arg, torch.Tensor)]
+
+  args_tensor_ids = [(index, torch_xla._XLAC._xla_get_tensor_id(xla_arg))
```
JackCaoG, line 237: ditto, maybe `arg_idx_and_tensor_ids`

torch_xla/core/dynamo_bridge.py
ysiraichi: What about: `index_and_xla_tensor_args`?

JackCaoG: sgtm

wonjoolee95 (approved):
Thanks for the PR! LGTM pending the CI tests.

test/dynamo/test_bridge.py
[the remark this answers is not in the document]
JackCaoG: I think it is aot-autograd handles the argument. What's the non-tensor input passed to `openxla_eval`?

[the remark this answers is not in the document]
ysiraichi: Yep, exactly. One of the inputs passed was a custom class that inherited from `nn.Embedding`.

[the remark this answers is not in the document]
JackCaoG: actually @bdhirsh any thought?

JackCaoG:
I think it is safer to have this change in, I am going to merge it.

PR #5690, opened 2023-10-09
Disable xla backend for SPMD

jonb377:
We should not allow using the xla backend with SPMD mode execution. Conceptually, SPMD has a single replica, and users should not need to directly manage collectives on the devices since the compiler will determine where the collectives should be.


alanwaketan (approved):
LGTM. Do you want to add a test case?

yeounoh (approved):
LGTM,, thx

PR #5691, opened 2023-10-09
Update artifacts.auto.tfvars for wheel with libtpu for py38 of releas…

ManfeiBai:
…e 2.1

### DIFF
--- a/infra/tpu-pytorch-releases/artifacts.auto.tfvars
+++ b/infra/tpu-pytorch-releases/artifacts.auto.tfvars
@@ -59,6 +59,13 @@ versioned_builds = [
   {
     git_tag         = "v2.1.0"
     pytorch_git_rev = "v2.1.0"
+    package_version = "2.1.0+libtpu"
+    accelerator     = "tpu"
+    bundle_libtpu   = "1"
+  },
+  {
+    git_tag         = "v2.1.0"
+    pytorch_git_rev = "v2.1.0"
     package_version = "2.1.0"
     accelerator     = "tpu"
     python_version  = "3.10"


JackCaoG:
what's the purpose of adding this trigger?

ManfeiBai:
> what's the purpose of adding this trigger?

this trigger is spirited by https://github.com/pytorch/xla/pull/5682, it looks like we suggest user to use torch_xla wheel with libtpu inside, and we have torch_xla+libtpu for py310 already, so add this trigger for py38 for torch_xla+libtpu

infra/tpu-pytorch-releases/artifacts.auto.tfvars
```diff
@@ -56,6 +56,13 @@ versioned_builds = [
     accelerator     = "tpu"
     bundle_libtpu   = "0"
   },
+  {
+    git_tag         = "v2.1.0"
+    pytorch_git_rev = "v2.1.0"
+    package_version = "2.1.0+libtpu"
+    accelerator     = "tpu"
+    bundle_libtpu   = "1"
```
mateuszlewko, line 64: Consider specifying python version explicitly.

will-cromar:
Do we need this build? The cases that I know of where the bundled libtpu is helpful:

1. Kaggle (which has a fixed python version)
2. Docker image (which provides its own python interpreter, so we don't need to publish multiple versions)

Otherwise, IMO it's better to install `torch_xla` from PyPI with the `tpu` optional dependency

JackCaoG:
Do we still need this?

PR #5693, opened 2023-10-10
Support synchronous saving and loading in CheckpointManager

jonb377:
This PR adds the initial functionality for CheckpointManager to manage synchronously taking checkpoints, restoring checkpoints, and managing how many checkpoints it tracks.

### DIFF
--- a/test/spmd/test_xla_distributed_checkpoint.py
+++ b/test/spmd/test_xla_distributed_checkpoint.py
@@ -26,9 +26,9 @@ def run_with_tmpdir(f):
 
   @functools.wraps(f)
   def run(*args, **kwargs):
-    assert 'tmpdir' not in kwargs
     with tempfile.TemporaryDirectory() as tmpdir:
-      f(*args, **kwargs, tmpdir=tmpdir)
+      kwargs.setdefault('tmpdir', tmpdir)
+      f(*args, **kwargs)
 
   return run
 
@@ -338,7 +338,7 @@ class CheckpointManagerTest(DistributedCheckpointTestBase):
     super().setUp()
     # Initialize the a minimal process group
     dist.init_process_group(
-        init_method='tcp://127.1:8932', world_size=1, rank=0)
+        backend='gloo', init_method='tcp://127.1:8932', world_size=1, rank=0)
 
   def tearDown(self):
     super().tearDown()
--- a/test/spmd/test_xla_distributed_checkpoint.py
+++ b/test/spmd/test_xla_distributed_checkpoint.py
@@ -1,3 +1,4 @@
+import functools
 import os
 import sys
 import tempfile
@@ -15,9 +16,21 @@ from torch.distributed.checkpoint.default_planner import (
     create_default_local_save_plan,
     create_default_global_save_plan,
 )
-from torch_xla.experimental.distributed_checkpoint import SPMDLoadPlanner, SPMDSavePlanner
+from torch_xla.experimental.distributed_checkpoint import SPMDLoadPlanner, SPMDSavePlanner, CheckpointManager
 from torch_xla.experimental.distributed_checkpoint._helpers import (
     _sharded_cpu_state_dict, _CpuShards, _is_sharded_tensor)
+
+
+# Wrapper to manage a temporary directory for the wrapped test
+def run_with_tmpdir(f):
+
+  @functools.wraps(f)
+  def run(*args, **kwargs):
+    assert 'tmpdir' not in kwargs
+    with tempfile.TemporaryDirectory() as tmpdir:
+      f(*args, **kwargs, tmpdir=tmpdir)
+
+  return run
 
 
 class DistributedCheckpointTestBase(test_xla_sharding_base.XlaShardingTest):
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -1,9 +1,17 @@
...
 import torch_xla.experimental.distributed_checkpoint as xc
 
-from typing import List, Optional
+from fsspec.core import url_to_fs
+from os.path import basename
+from typing import List, Optional, Union
 from torch.distributed.checkpoint.metadata import STATE_DICT_TYPE
 
+# TODO(jonbolin): Import path will change
+from torch.distributed.checkpoint._fsspec_filesystem import FsspecReader, FsspecWriter
```
jonb377, line 13: The import path will change when the API becomes public in the upstream. @alanwaketan @yeounoh do you have any thoughts on how to handle this?

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -319,6 +332,78 @@ def test_sharded_cpu_state_dict(self):
         self.assertTrue(param.device == torch.device("cpu"))
 
 
+class CheckpointManagerTest(DistributedCheckpointTestBase):
+
+  def setUp(self):
+    super().setUp()
+    # Initialize the a minimal process group
+    dist.init_process_group(
```
alanwaketan, line 340: Should we specify a backend?

torch_xla/experimental/distributed_checkpoint/manager.py
alanwaketan: It's okay. The upstream test will break our CI in the upstream, and then we can have a companion change to fix it.

torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -77,14 +96,35 @@ def __init__(self,
...
+    self.base_path = path
+    self.save_period = save_period
+    self.max_to_keep = max_to_keep
+
+  def _get_path(self, step: int) -> str:
+    return os.path.join(self.base_path, str(step))
+
+  def _release_oldest_checkpoints(self):
+    if self.max_to_keep > 0:
+      tracked_steps = sorted(self.all_steps(), reverse=True)
```
alanwaketan, line 113: Hmm, in python, we can't just have a queue structure?

torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -139,10 +188,18 @@ def restore(self, step: int, state_dict: STATE_DICT_TYPE) -> None:
...
+        storage_reader=FsspecReader(path),
+        planner=xc.SPMDLoadPlanner(),
+    )
 
   def all_steps(self) -> List[int]:
     """
     List all steps tracked by the CheckpointManager.
     """
-    raise NotImplementedError
+    fs, raw_path = url_to_fs(self.base_path)
```
alanwaketan, line 202: What's this? Why don't we just keep a queue as a private member?

test/spmd/test_xla_distributed_checkpoint.py
jonb377: Good point, I can make it explicit

torch_xla/experimental/distributed_checkpoint/manager.py
jonb377: The `all_steps` method returns the tracked steps from the fsspec backend, so I'm not directly tracking steps in the CheckpointManager. `tracked_steps` will hold all steps which exist on the persistent storage in decreasing order, so the earliest steps are at the end of the list to enable O(1) popping.

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -15,11 +16,23 @@
...
 from torch_xla.experimental.distributed_checkpoint._helpers import (
     _sharded_cpu_state_dict, _CpuShards, _is_sharded_tensor)
 
 
+# Wrapper to manage a temporary directory for the wrapped test
+def run_with_tmpdir(f):
+
+  @functools.wraps(f)
+  def run(*args, **kwargs):
+    assert 'tmpdir' not in kwargs
```
yeounoh, line 29: nit. if `tmpdir` already exists, then can we use it instead, instead of blocking?

torch_xla/experimental/distributed_checkpoint/manager.py
jonb377: `url_to_fs` takes in a path like `gs://my-bucket/path/to/chkpt` and returns an fsspec FileSystem object and a raw path without the storage prefix, e.g. `my-bucket/path/to/chkpt`. The FileSystem object is what's used to interface with the storage backend.

I'd prefer to read from the source-of-truth each time in case there are any changes on the backend, but we could make this a one-time operation and track a queue within the manager. Let me know what you prefer.

alanwaketan: I see. It depends on how hot this path will be? If it's a hot path, it'd better cached in the manager. Otherwise, it seems okay.

jonb377: It will be called once per checkpoint, but we may as well cache it.

torch_xla/experimental/distributed_checkpoint/manager.py
alanwaketan: It seems this will be a hot path? Basically, each time a ckpt is saved. You have O(N) to grep the all_steps and O(nlogn) to sort it. Then another O(N) to pop the items. It seems only the last O(N) is necessary.

torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -77,14 +97,48 @@ def __init__(self,
...
+    assert save_period > 0, "save_period must be positive"
+    assert async_queue_size > 0, "async_queue_size must be positive"
+    assert max_to_keep >= 0, "max_to_keep must be non-negative"
+
+    self.base_path = path
+    self.save_period = save_period
+    self.max_to_keep = max_to_keep
+
+    # Cache tracked steps in a heap for efficient clearing.
+    self._tracked_steps = self._load_tracked_steps()
```
alanwaketan, line 110: Why a heap? Can the .save be called out of order?

jonb377: It's sort of an edge case, but if the user restores from an earlier checkpoint and changes the save_period, the steps will be out of order. Using a heap ensures that we always delete the lowest step number.

alanwaketan: heap is still o(logn). If sort it after loading, then the later deletion from queue is just o(1)

jonb377: I'll update to delete the oldest by timestamp instead of lowest step number.

torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -53,10 +79,21 @@ class CheckpointManager:
   https://github.com/google/orbax/blob/efc079c4e5b437782a80138913d322cb3ed365c7/checkpoint/orbax/checkpoint/checkpoint_manager.py
   """
 
+  # The base path to write checkpoints to. Each checkpoint taken by the manager
+  # will be written into a subdirectory of this path, identified by the
+  # checkpoint's step.
+  base_path: Union[str, os.PathLike]
+
+  # The period to take checkpoints, in steps.
+  save_period: int
```
yeounoh, line 88: super nit. period -> interval. Feel free to keep as-is.

yeounoh (approved):
LGTM

alanwaketan (approved):
LGTM.

jonb377:
Thanks @yeounoh and @alanwaketan for the review! I'll merge after TPU CI.

PR #5696, opened 2023-10-10
Conditionally set default TPU settings in `__init__.py`

will-cromar:
Replace `_set_missing_env` with standard library call.

Only set `TPU_MEGACORE` on TPU v4. Otherwise, use the default value.


JackCaoG (approved):

torch_xla/_internal/tpu.py
```diff
@@ -266,6 +266,9 @@ def configure_topology(local_rank: int,
   os.environ.setdefault(xenv.TPU_VISIBLE_CHIPS, str(local_rank))
   os.environ.setdefault(xenv.TPU_PROCESS_PORT, str(ports[local_rank]))
 
+  if version() == 4:
```
alanwaketan, line 269: Is it guaranteed that this will be called all the time and will be before any modeling code involved tpu?

will-cromar: Actually, this will only get called for sure when using `torchrun` or `xmp.spawn`.. Good catch.

It makes the most logical sense to me to set these up during client initialization, but that happens in C++ and all of the TPU utilities are in Python. I'll make this setting conditional in `__init__.py` for now.

alanwaketan (approved):
LGTM. Thanks, Will.

PR #5697, opened 2023-10-10
Support async checkpointing through CheckpointManager

jonb377:
Support asynchronous checkpointing through the CheckpointManager interface. This will move the state_dict to CPU before starting the checkpoint, which unblocks the calling thread.

This PR depends on #5693 for synchronous checkpointing functionality.

### DIFF
--- a/test/spmd/test_xla_distributed_checkpoint.py
+++ b/test/spmd/test_xla_distributed_checkpoint.py
@@ -412,6 +412,44 @@ class CheckpointManagerTest(DistributedCheckpointTestBase):
     self.assertTrue(chkpt_mgr.save(20, state_dict))
     self.assertEqual(set(chkpt_mgr.all_steps()), {20, 10})
 
+  @run_with_tmpdir
+  def test_manager_async_checkpoint(self, tmpdir):
+    chkpt_mgr = CheckpointManager(tmpdir, save_period=10)
+    state_dict = self._get_sharded_model().state_dict()
+
+    self.assertEqual(chkpt_mgr.all_steps(), [])
+
+    # Steps not divisible by 10 should not be saved
+    for step in range(1, 10):
+      self.assertFalse(chkpt_mgr.save_async(step, state_dict))
+      self.assertEqual(chkpt_mgr.all_steps(), [])
+
+    # Steps divisible by 10 should be saved
+    saved = set()
+    for step in range(0, 100, 10):
+      self.assertTrue(chkpt_mgr.save_async(step, state_dict))
+      saved.add(step)
+
+    # Delete the checkpoint manager to block this thread until all pending
+    # async checkpoints are complete.
+    del chkpt_mgr
+
+    # The manager should track all steps which were asynchronously saved.
+    chkpt_mgr = CheckpointManager(tmpdir, save_period=10)
+    self.assertEqual(set(chkpt_mgr.all_steps()), saved)
+
+    # Load a checkpoint into a new state_dict
+    new_state_dict = self._get_sharded_model().state_dict()
+    self.assertFalse(
+        any(
+            torch.allclose(v, new_state_dict[k])
+            for k, v in state_dict.items()))
+    chkpt_mgr.restore(0, new_state_dict)
+    self.assertTrue(
+        all(
+            torch.allclose(v, new_state_dict[k])
+            for k, v in state_dict.items()))
+
 
 if __name__ == '__main__':
   test = unittest.main()
--- a/test/spmd/test_xla_distributed_checkpoint.py
+++ b/test/spmd/test_xla_distributed_checkpoint.py
@@ -4,6 +4,7 @@ import sys
 import tempfile
 import unittest
 import test_xla_sharding_base
+import threading
 
 import torch
 import torch.distributed as dist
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -105,6 +109,27 @@ def __init__(self,
     self.save_period = save_period
     self.max_to_keep = max_to_keep
 
+    self._async_queue = queue.Queue(maxsize=async_queue_size)
+    self._chkpt_thread = threading.Thread(target=self._async_worker, daemon=True)
+    self._chkpt_thread.start()
+
+    # Create a CPU process group to coordinate the checkpoint.
+    self.pg = dist.new_group(backend='gloo')
```
alanwaketan, line 117: I thought you suggested that the user should do this?

torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -105,6 +109,27 @@ def __init__(self,
...
+
+  def __del__(self):
+    # Ensure pending checkpoints are finished
+    self._async_queue.join()
```
alanwaketan, line 121: Interesting ideas!

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -403,6 +403,44 @@ def test_manager_max_to_keep(self, tmpdir):
...
+
+    # Steps divisible by 10 should be saved
+    saved = set()
+    for step in range(0, 100, 10):
+      self.assertTrue(chkpt_mgr.save_async(step, state_dict))
+      saved.add(step)
+
+    # Delete the checkpoint manager to block this thread until all pending
+    # async checkpoints are complete.
+    del chkpt_mgr
```
alanwaketan, line 426: Not sure if this can test the saving is async but not sync?

torch_xla/experimental/distributed_checkpoint/manager.py
yeounoh: A few questions:
- does this backend always work? `gloo` can be used not only for CPU, but also GPU. If you don't specify the members, does it only contain the CPU processes?
- how doe this differ from the default gloo pg?

jonb377: The user will need to initialize the default process group, but since the default group is thread-local, we need to create a handle on a subgroup for use in the background thread.

To @yeounoh:
- This will inherit the ranks from the default group. With SPMD on TPUs, this has to be a CPU group, but with GPU it could be on-device. Probably the right approach is to not specify `gloo` here and just inherit the default group's backend.
- The resulting group should be identical to the default group.

test/spmd/test_xla_distributed_checkpoint.py
jonb377: I'm worried about introducing flakiness to the tests... Do you have any thoughts on how to assert it's an async operation? Maybe I can add some counters.

torch_xla/experimental/distributed_checkpoint/manager.py
alanwaketan: Now I missed the [event loop](https://docs.python.org/3/library/asyncio-eventloop.html) in any other modern programming languages...

test/spmd/test_xla_distributed_checkpoint.py
alanwaketan: We can mock it in a way that the exec function in the manager never return until a signal sent later in your test after a check.

torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -175,7 +201,12 @@ def save_async(self,
     Returns:
       True if a checkpoint was taken and False otherwise.
     """
-    raise NotImplementedError
+    if self.should_save(step) or force:
+      # Move the state_dict to CPU
+      cpu_state_dict = _sharded_cpu_state_dict(state_dict)
+      self._async_queue.put((step, cpu_state_dict))
```
yeounoh, line 207: is this blocking when the size exceeds or at the capacity?

torch_xla/experimental/distributed_checkpoint/manager.py
jonb377: Ah unfortunately it seems I can't count on __del__ to be called... I'll have to revisit this.

torch_xla/experimental/distributed_checkpoint/manager.py
jonb377: Calls to `put` will block when the queue is at capacity

test/spmd/test_xla_distributed_checkpoint.py
yeounoh: Yea, this is concise and nice, though. it deduces it from the number of saved chkpts and threading.Thread(target=self._async_worker, daemon=True) did the job with the ending `_async_queue.join()` call.

@jonb377 it looks like you are spawning a single-thread per manager, how do you control the number of threads?

yeounoh (approved):
LGTM, left some questions.

torch_xla/experimental/distributed_checkpoint/manager.py
yeounoh: Then, you may also want to avoid spawning `daemon` threads.

test/spmd/test_xla_distributed_checkpoint.py
jonb377: That's a great callout, it should be addressed in the next patch. I'm going to allow the threads to exit when the manager is deleted.

torch_xla/experimental/distributed_checkpoint/manager.py
jonb377: I have a better understanding now - I can't rely on it immediately after calling `del chkpt_mgr` like I was in the unit tests, but it will be called before the process exits.

torch_xla/experimental/distributed_checkpoint/manager.py
yeounoh: @jonb377 good, can you just add a TODO comment, that need to verify for XLA:GPU

test/spmd/test_xla_distributed_checkpoint.py
yeounoh: also, if you can't rely on _del_ being called, you may wanto throw in `await`.

alanwaketan (approved):
LGTM.

jonb377:
cc @wz337

Issue #5625, opened 2023-09-20
Problem with huggingface imports in kaggle tpu


level14taken:
## ❓ Questions and Help
Hello everyone,

When working in kaggle with xla and huggingface in latest environment, even importing transformers in a different order(which library is imported first or last) or importing some models is causing errors,How do i tackle this?

[kaggle-notebook](https://www.kaggle.com/code/level14taken/notebooka2e632541b/notebook) to reproduce errors.

Additional info:-
This happens when import of libraries in not done in a certain way and it leads to error. This isn't in the notebook so posted it here.
```
2023-09-20 03:44:44.786788: F tensorflow/compiler/xla/xla_client/pjrt_computation_client.cc:81] Non-OK-status: pjrt::LoadPjrtPlugin( "tpu", sys_util::GetEnvString(env::kEnvTpuLibraryPath, "libtpu.so")) status: INVALID_ARGUMENT: Unexpected PJRT_Api size: expected 496, got 512. Check installed software versions.
```

will-cromar:
That error normally happens when there's a version mismatch between the framework and libtpu. There's a better error message coming in PT/XLA 2.1.

I wonder if Transformers is importing another ML framework and causing an issue. Unfortunately, each framework requires a different libtpu, so there are actually three copies on the VM. Can you rerun with the environment variable `TF_CPP_MIN_LOG_LEVEL=0`? I want to see which libtpu is being used.

level14taken:
Hi @will-cromar, check version 4 of [this](https://www.kaggle.com/code/level14taken/notebooka2e632541b/notebook) notebook.

level14taken:
Hi @will-cromar, it looks like the message is missed somehow, can you give a confirmation if this can be done and how or nothing can be done for now. It is sort of urgent for me.

will-cromar:
Hey, sorry for the slow response, and thank you for adding the debug logging. I think I understand what's happening.

Kaggle comes preinstalled with the TPU version of TF. Unfortunately, the TPU build of TF immediately initializes the TPU when you import it, which causes problems when any other framework tries to initialize the TPU. In your case, the parent process holds a lock on the TPU, which prevents any of the child processes from `xmp.spawn` from opening their respective chip.

Note this log when you load your model:

```
[...]

PR #5698, opened 2023-10-11
Don't set $TPU_LIBRARY_PATH during import

will-cromar:
JAX and PyTorch/XLA are both overriding `$TPU_LIBRARY_PATH` during import, leading to confusing issues that depend on import order, e.g. https://github.com/pytorch/xla/issues/5625

- Use the new `get_library_path` function to get that path from the libtpu package.
- Keep `$TPU_LIBRARY_PATH` as an override for compatibility

Corresponding fix in JAX: https://github.com/google/jax/commit/b81a3e1fd774ebdbc3015f1bc977bfacb5d4b745

The issue will be resolved when both JAX and PyTorch/XLA release new packages that include these two changes.

This is a temporary hack. I expect we'll have a better way to get the PJRT Plugin path in the near future.

cc @jyingl3

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -72,27 +72,30 @@ def _aws_ec2_inf_trn_init():
 
 
 def _setup_tpu_vm_library_path() -> bool:
-  """Returns true if $TPU_LIBRARY is set or can be inferred.
+  """Returns true if $TPU_LIBRARY_PATH is set or can be inferred.
 
   We load libtpu.so in the following order of precedence:
 
   1. User-set $TPU_LIBRARY_PATH
   2. libtpu.so included in torch_xla/lib
   3. libtpu-nightly pip package
+
+  Sets $PTXLA_TPU_LIBRARY_PATH if path is inferred by us to prevent conflicts
+  with other frameworks. This env var will be removed in a future version.
   """
-  if 'TPU_LIBRARY_PATH' in os.environ:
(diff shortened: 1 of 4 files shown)


JackCaoG:
I am a bit confuse after this pr, what will happen if user still set `TPU_LIBRARY_PATH ` ?

will-cromar:
We'll still take `TPU_LIBRARY_PATH` as an override. The change here is that we will not _set_ it. Right now, if you import us first, we'll override JAX's libtpu path, since they also treat `TPU_LIBRARY_PATH` as a user override.

Internally, we can use `PTXLA_...` so we stop breaking JAX, but we need a better way to plumb the plugin path.

JackCaoG (approved):

torch_xla/__init__.py
```diff
@@ -72,13 +72,16 @@ def _aws_ec2_inf_trn_init():
...
+  """Returns true if $TPU_LIBRARY_PATH is set or can be inferred.
 
   We load libtpu.so in the following order of precedence:
 
   1. User-set $TPU_LIBRARY_PATH
   2. libtpu.so included in torch_xla/lib
   3. libtpu-nightly pip package
+
+  Sets $PTXLA_TPU_LIBRARY_PATH if path is inferred by us to prevent conflicts
+  with other frameworks. This env var will be removed in a future version.
```
vanbasten23, line 84: I wonder under what condition can we remove this env var.

PR #5703, opened 2023-10-15
Use the bundle with tf.Variable when creating call_module op.

qihqi:
This way the result .pb file in saved_model will no longer contain weights.


lsy323:
@qihqi linter check fails, let's fix it before merging

qihqi:
haoliang merged this would be equivalent: https://github.com/pytorch/xla/pull/5708

PR #5706, opened 2023-10-16
Support SPMD through the xla:// init_method

jonb377:
For distributed checkpointing, a CPU process group is required to coordinate the checkpoint. This requires knowledge of the master worker IP address, which can be cumbersome for the user to configure manually.

In multiprocess, we support auto-discovery of the master worker IP through `tpu.discover_master_worker_ip`. The multiprocess implementation requires manually invoking collectives to broadcast the master IP to all workers. This change adds support for master worker IP discovery using the SPMD APIs.

Note that the `xla://` init_method is a general rendezvous that is not specific to the XLA backend, so this change enables creating the requisite CPU process group through, e.g.:

```
import torch_xla.distributed.xla_backend

xr.use_spmd()
dist.init_process_group(backend='gloo', init_method='xla://')
```

### DIFF
--- a/torch_xla/_internal/rendezvous.py
+++ b/torch_xla/_internal/rendezvous.py
@@ -28,7 +28,7 @@ def pjrt_rendezvous_handler(url: str,
     ) == 'TPU' else 'localhost'
 
   master_port = xu.getenv_as('MASTER_PORT', int, 12355)
-  world_size = xr.world_size()
+  world_size = xr.process_count()
   with _store_lock:
     global _store
     if not _store:
@@ -44,4 +44,7 @@ def pjrt_rendezvous_handler(url: str,
             xr.process_count(),
             is_master=xr.process_index() == 0)
 
-  yield (_store, xr.global_ordinal(), world_size)
+  # In SPMD, the rank is the process index, while in multiprocess it should be
+  # the device ordinal.
+  rank = xr.process_index() if xr.is_spmd() else xr.global_ordinal()
+  yield (_store, rank, world_size)
--- a/torch_xla/_internal/rendezvous.py
+++ b/torch_xla/_internal/rendezvous.py
@@ -28,7 +28,6 @@ def pjrt_rendezvous_handler(url: str,
     ) == 'TPU' else 'localhost'
 
   master_port = xu.getenv_as('MASTER_PORT', int, 12355)
-  world_size = xr.process_count()
   with _store_lock:
     global _store
     if not _store:
@@ -44,7 +43,8 @@ def pjrt_rendezvous_handler(url: str,
             xr.process_count(),
             is_master=xr.process_index() == 0)
 
-  # In SPMD, the rank is the process index, while in multiprocess it should be
-  # the device ordinal.
+  # In SPMD, the world size and rank are determined by the process count and
+  # index, while in multiprocess they are based on the device count and ordinal.
+  world_size = xr.process_count() if xr.is_spmd() else xr.world_size()
   rank = xr.process_index() if xr.is_spmd() else xr.global_ordinal()
   yield (_store, rank, world_size)
--- a/torch_xla/_internal/tpu.py
+++ b/torch_xla/_internal/tpu.py
@@ -270,7 +270,10 @@ def configure_topology(local_rank: int,
 
 
 def discover_master_worker_ip(use_localhost: bool = True) -> str:
-  """Find the IP of the TPU host with TPU:0.
+  """Find the IP of the master TPU host.
+
+  In multiprocess, this is the host with TPU:0.
+  In SPMD mode, this is the host running process 0.
 
   TPU device IDs are nondeterministic and independent from Cloud TPU worker IDs.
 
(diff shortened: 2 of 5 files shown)


torch_xla/_internal/rendezvous.py
```diff
@@ -28,7 +28,7 @@ def pjrt_rendezvous_handler(url: str,
     ) == 'TPU' else 'localhost'
 
   master_port = xu.getenv_as('MASTER_PORT', int, 12355)
-  world_size = xr.world_size()
+  world_size = xr.process_count()
```
will-cromar, line 31: This should stay `xr.world_size` for the multiprocessing case. This is only returned at the very end to set the `torch.distributed` world size correctly for multithreading, so you can make this conditional and move it down with `rank`

jonb377: Ah I forgot about the multithreading case... Sounds good, I'll ensure this method is exactly the same without SPMD. Thanks for checking Will!

jonb377:
Supporting `discover_master_worker_ip` with SPMD is needed for auto checkpointing, please take a look when you get a chance @alanwaketan @yeounoh. I can remove the init_method support from this PR if needed.

torch_xla/_internal/tpu.py
```diff
@@ -268,23 +270,50 @@ def configure_topology(local_rank: int,
...
+
   t = torch.tensor([current_worker_id], device=xm.xla_device())
   xm.collective_broadcast([t])
   xm.mark_step()
 
   master_worker_id = int(t.cpu())
   return worker_ips[master_worker_id]
+
+
+def _spmd_find_master_ip(current_worker_ip: str) -> str:
```
alanwaketan, line 302: This is very convoluted but I guess it works...

alanwaketan (commented):
I think this approach is fine. Let's have a test case for it given the logic is complicated.

torch_xla/_internal/tpu.py
jonb377: Yeah, it's a hack around our lack of direct control over collectives... If you or @yeounoh have any other ideas, I'm definitely open to revisiting this.

torch_xla/runtime.py
```diff
@@ -242,3 +242,14 @@ def is_spmd():
   """Returns if SPMD is set for execution."""
   # TODO(yeounoh) replace this when we fully deprecate the flag.
   return xu.check_env_flag('XLA_USE_SPMD')
+
+
+@requires_pjrt
+def get_master_ip() -> str:
```
jonb377, line 248: cc @will-cromar

jonb377:
> I think this approach is fine. Let's have a test case for it given the logic is complicated.

Added a basic unit test. Full E2E coverage will be through the multihost checkpoint tests.

torch_xla/_internal/rendezvous.py
```diff
@@ -44,4 +43,8 @@ def pjrt_rendezvous_handler(url: str,
             xr.process_count(),
             is_master=xr.process_index() == 0)
 
-  yield (_store, xr.global_ordinal(), world_size)
+  # In SPMD, the world size and rank are determined by the process count and
+  # index, while in multiprocess they are based on the device count and ordinal.
+  world_size = xr.process_count() if xr.is_spmd() else xr.world_size()
+  rank = xr.process_index() if xr.is_spmd() else xr.global_ordinal()
```
yeounoh, line 49: Is the process idx local (per host)?

jonb377: Yeah, with SPMD we have a single process per host. process_index gives the global index of the current host's process in the runtime, so it will be different on each host.

torch_xla/_internal/rendezvous.py
```diff
@@ -44,4 +43,8 @@ def pjrt_rendezvous_handler(url: str,
             xr.process_count(),
             is_master=xr.process_index() == 0)
 
-  yield (_store, xr.global_ordinal(), world_size)
+  # In SPMD, the world size and rank are determined by the process count and
```
yeounoh, line 46: In both cases the world size and rank should be determined by the process count and index -- this is needed due to the fact that `xr.world_size()` returns 1 for a single-replica SPMD case. I think we could instead explain this, saying that "In SPMD we use the process count and index directly, since the xr.world_size() returns 1 without any replication group set." Or you can also add some more in the context of the CPU process group need for SPMD?

torch_xla/_internal/tpu.py
```diff
@@ -268,23 +270,50 @@ def configure_topology(local_rank: int,
...
+  import torch_xla.runtime as xr
+  import torch_xla.experimental.xla_sharding as xs
+  from_cpu_shards = torch_xla._XLAC._global_tensor_from_cpu_shards
+  ip_int = int(ip_address(current_worker_ip))
+  n_dev = xr.global_runtime_device_count()
+  local_ndev = len(torch_xla._XLAC._xla_get_runtime_devices())
+  # Create a global (n_dev x 2) tensor containing all process indices and IPs,
+  # and find the process 0 IP as the master IP.
+  shard = torch.LongTensor([[xr.process_index(), ip_int]])
```
yeounoh, line 311: Could we use global device ordinal instead of the process idx? It's the same thing, but wonder if it's feasible -- that way your search for the IP associated with TPU0 would be more consistent with the code?

jonb377: Unfortunately it's actually different - there's no guaranteed mapping between processes and devices, so process 0 may not control device 0 and we can't pull the IP out of the device 0 slot in the global tensor. In SPMD, I think it's better to rely on the process abstraction, since we never actually need the devices for collectives.

torch_xla/_internal/tpu.py
```diff
@@ -268,23 +270,50 @@ def configure_topology(local_rank: int,
...
+  op_sharding = xs.Mesh(range(n_dev), (n_dev, 1)).get_op_sharding((0, 1))
+  global_tensor = from_cpu_shards([shard] * local_ndev, op_sharding).cpu()
+  try:
+    # Process 0 may not control device 0, so we must do a linear search.
+    master_ip = next(ip for proc, ip in global_tensor.tolist() if proc == 0)
```
yeounoh, line 316: Let's use while loop instead of next & tolist()?

torch_xla/_internal/tpu.py
yeounoh: Thanks @jonb377 , we are not looking for the IP associated with TPU0, but with the rank=0.

yeounoh (approved):
LGTM

PR #5707, opened 2023-10-16
Add support for `_unsafe_index`.

ysiraichi:
This PR adds support for `_unsafe_index` by aliasing `index`.

**Problem:** `_unsafe_index` was hitting the CPU fallback execution flow, which moved the tensors to CPU (including `self`)
- The CPU fallback was calling the dispatcher again: `at::index`
- That, was getting dispatched to the XLA implementation (error, since `self` lives on CPU)

This problem affected a few Torchbench benchmarks running with `openxla` backend: `Super_SloMo`, `doctr_det_predictor`, `pytorch_unet`, ...

### DIFF
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -2033,6 +2033,34 @@ TEST_F(AtenXlaTensorTest, TestScatterReduceMaxInPlace) {
   ExpectCounterChanged("xla::scatter_reduce", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestUnsafeIndex) {
+  for (torch::ScalarType scalar_type :
+       {torch::kFloat, torch::kByte, torch::kChar, torch::kShort, torch::kInt,
+        torch::kLong}) {
+    torch::Tensor a =
+        isFloatingType(scalar_type)
+            ? torch::rand({3, 4}, torch::TensorOptions(scalar_type))
+            : torch::randint(100, {3, 4}, torch::TensorOptions(scalar_type));
+    for (torch::ScalarType index_scalar_type : {torch::kInt, torch::kLong}) {
+      torch::List<torch::optional<torch::Tensor>> indices{
+          torch::tensor({0, 1}, torch::TensorOptions(index_scalar_type)),
+          torch::tensor({2, 3}, torch::TensorOptions(index_scalar_type))};
+      torch::Tensor c0 = torch::_unsafe_index(a, indices);
+      ForEachDevice([&](const torch::Device& device) {
+        torch::Tensor xla_a = CopyToDevice(a, device);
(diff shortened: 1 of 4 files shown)


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -626,6 +626,13 @@ at::Tensor XLANativeFunctions::_unsafe_view(const at::Tensor& self,
   return view_copy_symint(self, c10::fromIntArrayRefSlow(size));
 }
 
+at::Tensor XLANativeFunctions::_unsafe_index(
```
JackCaoG, line 629: do we know when will this method being used(how does it get dispatched to this) and create a test for it?

wonjoolee95: +1 to adding the tests. For more info, the cpp unit tests can verify that this new op you lowered is properly invoked via metric assertions as such https://github.com/pytorch/xla/blob/master/test/cpp/test_aten_xla_tensor_1.cpp#L35.

lezcano: @peterbell10 added this function. These functions are an artefact to signal the compiler to avoid performing any boundary checks on the indices that they use. These are used in decompositions where we know that the indices will be in the right range, like the backward of functions, or a decomposition where we know that the inputs are bounded to the right range.

lezcano: Note that you also have its cousin `_unsafe_index_put`

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -2033,6 +2033,34 @@ TEST_F(AtenXlaTensorTest, TestScatterReduceMaxInPlace) {
   ExpectCounterChanged("xla::scatter_reduce", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestUnsafeIndex) {
```
lezcano, line 2036: Is it possible / do we want to merge this test and that from `aten.index`?

ysiraichi: @JackCaoG @wonjoolee95 @lezcano 
I guess it's possible. But I'd argue against it for consistency. But, wouldn't mind doing this. What do you think?

wonjoolee95: Both sounds okay to me, but this is fine as it is now given the structure of our cpp unit tests.

wonjoolee95 (approved):

JackCaoG (approved):
Thanks!

peterbell10:
> That, was getting dispatched to the XLA implementation (error, since self lives on CPU)

This sounds like a bug, either the fallback creates a CPU tensor with XLA dispatch keys, or the falback mechanism doesn't handle lists of tensors and some of the arguments were not moved to CPU.

PR #5710, opened 2023-10-17
Patch RNG for better memory utilization in dropout layers

yeounoh:
The patch has landed in OpenXLA https://github.com/openxla/xla/pull/6015, we can enable this now, to reduce memory pressure in `nn.Dropout` layer.

DO_NOT_MERGE until the next pin update.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -30,6 +30,7 @@ python_configure(
 # b) get the sha256 hash of the commit by running:
 #    curl -L https://github.com/openxla/xla/archive/<git hash>.tar.gz | sha256sum
 #    and update the sha256 with the result.
+<<<<<<< HEAD
 http_archive(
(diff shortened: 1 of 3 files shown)


JackCaoG (approved):

JackCaoG:
do we want to add a test to check that int8 is used for rng? I think we can check the HLO generated,

yeounoh:
> do we want to add a test to check that int8 is used for rng? I think we can check the HLO generated,

Hey Jack, yea good idea, will add before merging.

yeounoh:
> > do we want to add a test to check that int8 is used for rng? I think we can check the HLO generated,
> 
> Hey Jack, yea good idea, will add before merging.

Verified locally with the new unit test -- cc @JackCaoG

PR #5711, opened 2023-10-18
When exporting StableHLO to SavedModel, also include the original var

haozha111:
When exporting StableHLO to SavedModel, also include the original variable name from Pytorch.


qihqi (approved):

qihqi:
please rebase, I think the change on tf signature is already merged?

haozha111:
> please rebase, I think the change on tf signature is already merged?

done

PR #5712, opened 2023-10-19
Implement mark_sharding as a custom op to support dynamo spmd activation sharding

wonjoolee95:
Implement mark_sharding as a custom op to support dynamo spmd activation sharding.

The PR is a bit messy as it includes fair bit amount refactoring, here is a quick summary:
- In init_python_bindings.cpp
   -  Move the existing `_xla_mark_sharding` logic out to a helper function, so it can be called by the new custom op `xla_mark_sharding_dynamo_custom_op`.
   - Move the torch custom op registration `TORCH_LIBRARY` from `aten_autograd_ops.h` to `init_python_bindings.cpp` since torch custom ops can registered only in one location.
 - Update the existing `torch_xla/experimental/xla_sharding.py` function to accept an additional boolean flag `dynamo_custom_op`. When set to true, it calls this new custom op `xla_mark_sharding_dynamo_custom_op` instead of the existing ``_xla_mark_sharding`. 

Test plan:
python test/spmd/test_dynamo_spmd.py DynamoSpmdInferenceTest.test_mark_sharding_inside_compile


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1626,6 +1626,11 @@ void InitXlaModuleBindings(py::module m) {
     // Register sharded tensor data.
     XLAGraphExecutor::Get()->RegisterTensor(xtensor->data());
   });
+  m.def("_xla_mark_sharding_custom_op",
```
yeounoh, line 1629: nit. for future reference, can we make it explicit by calling `_xla_mark_sharding_dynamo_custom_op`?

test/spmd/test_dynamo_spmd.py
```diff
@@ -171,6 +171,25 @@ def test_dynamo_input_sharding_threashold(self):
     else:
       del os.environ['XLA_DYNAMO_INPUT_SHARDING_CHECK_THRESHOLD']
 
+  def test_mark_sharding_inside_compile(self):
+
+    def fn_simple(x):
+      y = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
+                       dtype=torch.float,
+                       device=xm.xla_device())
+      ys = xs.mark_sharding(y, self._get_mesh((1, self.n_devices)), (0, 1))
```
yeounoh, line 180: Yes, this should test the activation sharding use-case. cc @wonjoolee95

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -1228,4 +1228,11 @@ xla::XlaOp BuildCustomSharding(const xla::XlaOp& input) {
                          {input}, ShapeHelper::ShapeOfXlaOp(input));
 }
 
+xla::XlaOp BuildCustomMarkSharding(const torch::lazy::BackendDevice& device,
+                                   const xla::XlaOp& input,
+                                   const xla::XlaOp sharding) {
+  return xla::CustomCall(input.builder(), /*call_target_name=*/"MarkSharding",
```
yeounoh, line 1234: Could we re-use the existing `CustomSharding` op? `mark_sharding` can be translated into the `CustomSharding` op for an IR, which is the activation node. Unless it's the Dynamo side change to register the custom ops, we can reuse the existing `extern const OpKindWrapper xla_custom_sharding`, no?

wonjoolee95: The original intention here was to have a separate custom op that includes all the logic for `mark_sharding` so Dynamo can properly capture `mark_sharding` as an op. All the logic currently at `xla_mark_sharding` at `init_python_bindings.cpp` would be moved to the lowering logic of this new custom op, so Dynamo can trace and recognize it properly.

torch_xla/csrc/init_python_bindings.cpp
wonjoolee95: Sg, updatd. Also added a new API in `xla_sharding.py` named `mark_sharding_dynamo_custom_op` specifically for interacting with this new custom op.

wonjoolee95:
Did some tests locally, it seems like we need to move rest of `mark_sharding` logic that's currently at `tensor_methods.cpp:: custom_mark_sharding` to the actual lowering logic of the new `CustomMarkSharding` op for dynamo to properly recognize and trace this new op. Now working on this fix.

wonjoolee95:
Unit tests are passing locally, this should be ready for review now, @JackCaoG @yeounoh.

test/spmd/test_dynamo_spmd.py
```diff
@@ -15,16 +15,22 @@
...
     self.fc2 = nn.Linear(128, 1)
     # Add an additional 1x1 layer at the end to ensure the final layer
     # is not sharded.
     self.fc3 = nn.Linear(1, 1)
+    # If mesh is not none, we'll do a mark sharding inside the forward function
+    # to ensure dynamo can recognize and trace it in a torch compile.
+    self.mesh = mesh
 
   def forward(self, x):
+    if self.mesh and 'xla' in str(self.fc2.weight.device):
```
wonjoolee95, line 31: @yeounoh, I updated the unit test for doing `mark_sharding` inside torch compile to be part of the existing `SimpleLinear`. Here, I just do a `mark_sharding` call inside the `forward` function. Please let me know if you think this will suffice.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -651,6 +652,136 @@ std::string GetPyTypeString(py::handle obj) {
...
+      "-> Tensor",
+      torch::dispatch(
+          c10::DispatchKey::XLA,
+          TORCH_FN(torch_xla::aten_autograd_ops::max_pool2d_backward)));
+  m.def(
+      "xla_mark_sharding_dynamo_custom_op(Tensor input, int[][] "
+      "tile_assignment, int[][] group_assignment, int[][] replication_groups, "
+      "int sharding_type) -> ()",
+      torch::dispatch(c10::DispatchKey::XLA,
+                      TORCH_FN(xla_mark_sharding_dynamo_custom_op)));
```
JackCaoG, line 782: Just a random quesiton, will pytorch try to run backward on this custom op?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1559,72 +1690,39 @@ void InitXlaModuleBindings(py::module m) {
...
+          xla_mark_sharding(input, sharding);
+        });
+  m.def("_xla_mark_sharding_dynamo_custom_op",
+        [](const at::Tensor& input, const py::list& tile_assignment,
+           const py::list& group_assignment, const py::list& replication_groups,
+           int sharding_type) {
+          c10::List<at::IntArrayRef> time_assignment_list =
+              c10::List<at::IntArrayRef>();
+          for (auto t : tile_assignment) {
+            time_assignment_list.push_back(
```
JackCaoG, line 1704: typo, `tile_assignment`

torch_xla/experimental/xla_sharding.py
```diff
@@ -82,7 +82,8 @@ def get_axis_name_idx(self, name: str) -> int:
 
   @functools.lru_cache(maxsize=None)
   def get_op_sharding(self,
-                      partition_spec: Tuple) -> torch_xla._XLAC.OpSharding:
+                      partition_spec: Tuple,
+                      flatten=False) -> torch_xla._XLAC.OpSharding:
```
JackCaoG, line 86: nit, `flatten_opsharding` might be a better name

torch_xla/experimental/xla_sharding.py
```diff
@@ -449,9 +456,10 @@ def _translate_named_partition_spec(mesh: Mesh, partition_spec: Tuple):
 
 
 @xr.requires_pjrt
-def mark_sharding(
-    t: Union[torch.Tensor, XLAShardedTensor], mesh: Mesh,
-    partition_spec: Tuple[Union[Tuple, int, str, None]]) -> XLAShardedTensor:
+def mark_sharding(t: Union[torch.Tensor, XLAShardedTensor],
+                  mesh: Mesh,
+                  partition_spec: Tuple[Union[Tuple, int, str, None]],
+                  dynamo_custom_op: bool = False) -> XLAShardedTensor:
```
JackCaoG, line 462: nit maybe `use_dynamo_custom_op `?

torch_xla/experimental/xla_sharding.py
```diff
@@ -497,13 +508,28 @@ def mark_sharding(
...
+  if dynamo_custom_op:
+    tile_assignment, group_assignment, replication_groups, sharding_type = mesh.get_op_sharding(
+        partition_spec, flatten=True)
+
+    if isinstance(t, XLAShardedTensor):
+      torch_xla._XLAC._xla_mark_sharding_dynamo_custom_op(
+          t.global_tensor, tile_assignment, group_assignment,
+          replication_groups, sharding_type)
+      return t
+    torch_xla._XLAC._xla_mark_sharding_dynamo_custom_op(t, tile_assignment,
```
JackCaoG, line 520: can we put this into a else?

miladm:
Thanks @wonjoolee95 for this PR.
As a follow up to this PR, once landed, can you please evaluate the perf gain impact of activation sharding on the llama2 model?

torch_xla/csrc/init_python_bindings.cpp
wonjoolee95: According to https://pytorch.org/docs/stable/notes/extending.html, it looks like we need to implement and override  a specific `backward` function. So in this case, PyTorch should not try to run backward on this custom op.

test/spmd/test_dynamo_spmd.py
```diff
@@ -53,6 +59,22 @@ def test_dynamo_spmd_basic(self):
     # TODO(JackCaoG): add counter checks after ExecuteReplicated also creates
     # a ExecuteMetric.
 
+  def test_dynamo_spmd_basic_with_custom_mark_sharding_op(self):
```
JackCaoG, line 62: nit, maybe rename it something similar to `test_dynamo_spmd_mark_sharding_outside_of_compile`

test/spmd/test_dynamo_spmd.py
```diff
@@ -53,6 +59,22 @@ def test_dynamo_spmd_basic(self):
...
+    xs.mark_sharding(
+        linear.fc2.weight,
+        self._get_mesh((1, self.n_devices)), (1, 0),
+        use_dynamo_custom_op=True)
+    xla_res = linear(xla_x)
+    xm.mark_step()
+
+    dynamo_linear = torch.compile(linear, backend="openxla")
+    dynamo_res = dynamo_linear(xla_x)
+    torch.allclose(xla_res.cpu(), dynamo_res.cpu())
```
JackCaoG, line 76: can we add a counter check? We want to make sure we are not recompiling across differerent runs. You can either add it here or add a separate test.

JackCaoG (commented):
mostly lgtm, minor nits

test/spmd/test_dynamo_spmd.py
wonjoolee95: Updated

JackCaoG (approved):

test/spmd/test_dynamo_spmd.py
yeounoh: SG

torch_xla/experimental/xla_sharding.py
```diff
@@ -104,9 +105,15 @@ def get_op_sharding(self,
     replicate_dims = {i for i, d in enumerate(partition_spec) if d is None}
     group_assignment, replication_groups = _get_group_assignment(
         sharding_type, tile_assignment, len(partition_spec), replicate_dims)
-    return torch_xla._XLAC.OpSharding(tile_assignment.tolist(),
-                                      group_assignment, replication_groups,
-                                      int(sharding_type))
+
+    # If flatten_opsharding = True, return the flattened version of OpSharding
+    if flatten_opsharding:
```
yeounoh, line 110: What is flattening here? Just returning as a tuple? If so, maybe call it `as_tuple`. I am still hesitant to override the return type here... you can try this using the accessor methods instead, since you only need the `sharding_type` value in your work. Let me create a follow-up PR and add you, let's land this for now. Thanks @wonjoolee95

torch_xla/experimental/xla_sharding.py
```diff
@@ -471,6 +479,9 @@ def mark_sharding(
         >> mesh_shape = (4, 2)
         >> partition_spec = (0, None)
 
+        dynamo_custom_op (bool): if set to True, it calls the dynamo custom op variant of mark_sharding
+          to make itself recognizeable and traceable by dynamo.
```
yeounoh, line 483: nit. "recognizeable and traceable" --> "traceable"

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1559,72 +1560,39 @@ void InitXlaModuleBindings(py::module m) {
...
-    }
+  m.def("_xla_mark_sharding",
+        [](const at::Tensor& input, xla::OpSharding sharding) {
+          ShardingUtil::xla_mark_sharding(input, sharding);
+        });
+  m.def("_xla_mark_sharding_dynamo_custom_op",
+        [](const at::Tensor& input, const py::list& tile_assignment,
+           const py::list& group_assignment, const py::list& replication_groups,
+           int sharding_type) {
+          c10::List<at::IntArrayRef> tile_assignment_list =
```
yeounoh, line 1571: Can we move the following data processing logic into `xla_mark_sharding_dynamo_custom_op` and just call ```
xla_mark_sharding_dynamo_custom_op(
              input, tile_assignment_list, group_assignment_list,
              replication_groups_list, sharding_type);
        });
```

similar to what you've done for mark_sharing.

yeounoh (approved):
LGTM, have some comments that we can address in a follow-up PR. Thank you @wonjoolee95

PR #5721, opened 2023-10-23
Make sure printing XLA tensor only execute the HLO once

JackCaoG:


test/test_operations.py
```diff
@@ -1649,6 +1649,40 @@ def test_cached_addcdiv(self):
...
+
+    # case 1 mark_step
+    t1 = torch.randn(1, 4, device=xla_device)
+    xm.mark_step()
+    xm.wait_device_ops()
+    self.assertEqual(met.metric_data('ExecuteTime')[0], 1)
+    for _ in range(3):
+      print(t1)
+    self.assertEqual(met.metric_data('ExecuteTime')[0], 1)
+    self.assertIn('xla::device_data',
```
vanbasten23, line 1665: I wonder what this`self.assertIn('xla::device_data'` is testing for.

test/test_operations.py
```diff
@@ -1649,6 +1649,40 @@ def test_cached_addcdiv(self):
...
+  def test_print_executation(self):
+    xla_device = xm.xla_device()
+    xm.mark_step()
+    met.clear_all()
+
+    # case 1 mark_step
+    t1 = torch.randn(1, 4, device=xla_device)
+    xm.mark_step()
+    xm.wait_device_ops()
+    self.assertEqual(met.metric_data('ExecuteTime')[0], 1)
```
will-cromar, line 1661: side note: we need a less obscure way to get metrics. I assume `met.metric_data(...)[0]` is the number of samples for a metric?

will-cromar (approved):
LGTM. Was there a recent bug that inspired this test?

vanbasten23 (approved):

test/test_operations.py
JackCaoG: make sure t1 is a `device_data` after print, this mean execution has materialize the tensor.

alanwaketan (approved):
LGTM. Good news that we don't have bugs in the code.

JackCaoG:
somehow cpu CI failed, it passed TPU CI on my end. let me try to repo.

test/test_operations.py
wonjoolee95: +1 for this, I think it should be pretty simple, we probably just need to add some getter functions. Yeah currently 0th index is `TotalSamples`. Ex. output looks like:
```
Metric: OutboundData
  TotalSamples: 3
  Accumulator: 12.00B
  ValueRate: 34.00B / second
  Rate: 8.50056 / second
  Percentiles: 1%=4.00B; 5%=4.00B; 10%=4.00B; 20%=4.00B; 50%=4.00B; 80%=4.00B; 90%=4.00B; 95%=4.00B; 99%=4.00B
  ```

wonjoolee95 (approved):

JackCaoG:
oh it failed for `XLA_USE_EAGER_DEBUG_MODE` which is expected. I would skip the test is we are in `XLA_USE_EAGER_DEBUG_MODE` since it will force each op to be execute right after it is called and messed up the `ExecuteTime` counter.

yeounoh (approved):
LGTM

test/test_operations.py
JackCaoG: yea sg

wonjoolee95: Not blocking though, I actually had a local branch that somewhat does this l when I was debugging something else with metrics -- so I can probably follow-up with a PR.

PR #5724, opened 2023-10-24
add op lowering for _prelu_kernel_backward

### LINKED ISSUE #5695: OP lowering for `_prelu_kernel_backward`
## 🚀 Feature

Hello! I found out that there is no xla op for `nn.PReLU`, although  `_prelu_kernel` is [supported](https://github.com/pytorch/xla/blob/v2.1.0/codegen/xla_native_functions.yaml#L269), it looks like its not about  `nn.PReLU` function. For the sake of example, I added `PReLU` functions into default [example](https://github.com/pytorch/xla/blob/v2.1.0/test/test_train_mp_mnist.py#L45) .

```
class MNIST(nn.Module):

  def __init__(self):
    super(MNIST, self).__init__()
    self.conv1 = nn.Conv2d(1, 10, kernel_size=5)
    self.prelu1 = nn.PReLU(10)
    self.bn1 = nn.BatchNorm2d(10)
    self.conv2 = nn.Conv2d(10, 20, kernel_size=5)
    self.prelu2 = nn.PReLU(20)
    self.bn2 = nn.BatchNorm2d(20)
    self.fc1 = nn.Linear(320, 50)
    self.prelu3 = nn.PReLU(50)
    self.fc2 = nn.Linear(50, 10)

  def forward(self, x):
    x = self.prelu1(F.max_pool2d(self.conv1(x), 2))
    x = self.bn1(x)
    x = self.prelu2(F.max_pool2d(self.conv2(x), 2))
    x = self.bn2(x)
    x = torch.flatten(x, 1)
    x = self.prelu3(self.fc1(x))
    x = self.fc2(x)
    return F.log_softmax(x, dim=1)
```

Debug output
```
Counter: aten::_prelu_kernel_backward
  Value: 305

pt-xla-profiler: TransferFromServerTime too frequent: 311 counts during 105 steps
pt-xla-profiler: Op(s) not lowered: aten::_prelu_kernel_backward,  Please open a GitHub issue with the above op lowering requests.
```

Packages version:
```
torch                    2.1.0
torch-xla                2.1.0
```

zpcore:
Fix #5695

Manual lowering for op `_prelu_kernel_backward`, based on https://github.com/pytorch/pytorch/blob/4b324a8717e54faf1664248c3347633f4ff1e428/aten/src/ATen/native/cpu/Activation.cpp#L1419C1-L1430C1.

The op function signature in auto generated file `XLANativeunctions.h` is:
```
static ::std::tuple<at::Tensor,at::Tensor> _prelu_kernel_backward(const at::Tensor & grad_output, const at::Tensor & self, const at::Tensor & weight);
```

### DIFF
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -268,6 +268,7 @@ supported:
   - pow.Tensor_Scalar
   - pow.Tensor_Tensor
   - _prelu_kernel
+  - _prelu_kernel_backward
   - prod
   - prod.dim_int
   - _propagate_xla_data
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -3397,6 +3397,22 @@ TEST_F(AtenXlaTensorTest, TestPrelu) {
   ExpectCounterChanged("xla::_prelu_kernel", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestPreluBackward) {
+  auto testfn = [&](const std::vector<torch::Tensor>& inputs) -> torch::Tensor {
+    return torch::prelu(inputs[0], inputs[1]);
+  };
+  torch::Tensor input = torch::rand(
+      {5, 3}, torch::TensorOptions(torch::kFloat).requires_grad(true));
+  torch::Tensor weight = torch::rand({3}, torch::TensorOptions(torch::kFloat));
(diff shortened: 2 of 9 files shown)


zpcore:
Actually I am not quite understand why `_prelu_kernel_backward` was not decomposed into `where` based on the code [here](https://github.com/pytorch/pytorch/blob/4ed4753ac391b7a551b59455ca21f201bc147631/torch/_decomp/decompositions.py#L287). Can anyone explain the reason?

JackCaoG (approved):

Issue #5722, opened 2023-10-23
XLA dynamo bridge breaks when there are CPU outputs


ysiraichi:
## 🐛 Bug

Running Torchbench benchmarks, such as `hf_GPT2`, with XLA dynamo bridge fails due to non-XLA tensors present in the output of a `GraphModule`. 

## To Reproduce

Run the following program:

```python
import torch
import torch_xla.core.xla_model as xm

device = xm.xla_device()

def foo(x):
    return (x + x).cpu()

x = torch.randn(5, device=device)

expected = foo(x)
print(expected)

actual = torch.compile(backend="openxla")(foo)(x)
print(actual)
```

```python
  File "torch/_dynamo/backends/torchxla.py", line 49, in fwd
    compiled_graph = bridge.extract_compiled_graph(model, args)
  File "xla/torch_xla/core/dynamo_bridge.py", line 533, in extract_compiled_graph
    extract_internal(fused_module), node.args, None)
  File "xla/torch_xla/core/dynamo_bridge.py", line 337, in extract_internal
    dumb_return_handler, xla_args_need_update) = extract_graph_helper(xla_model)
  File "xla/torch_xla/core/dynamo_bridge.py", line 284, in extract_graph_helper
    graph_hash = torch_xla._XLAC._get_graph_hash(args_and_out)
RuntimeError: torch_xla/csrc/aten_xla_bridge.cpp:95 : Check failed: xtensor
*** Begin stack trace ***
        tsl::CurrentStackTrace[abi:cxx11]()
        torch_xla::bridge::GetXlaTensor(at::Tensor const&)
```

## Environment

PyTorch/XLA: 5dfcec90ca2dd33d2361adf981677cc0978186fa (Oct 21)

ysiraichi:
One possible solution is to use the `CapabilityBasedPartitioner` to ignore nodes where `node.meta["val"]` (populated by FX proxies) is on CPU. While this works, we may be also ignoring programs that wouldn't fail (e.g. if we use the CPU tensor as indices).

JackCaoG:
yea this is a bit annoying.. What should happen is that we extract the computation that happens on the XLA device, execute the graph and then do things like `res = res_xla.cpu()` to return the final output.

I think currently what happened is that during 
https://github.com/pytorch/xla/blob/43cd6b50d8a66d8337db28b328e6c80f797fc888/torch_xla/core/dynamo_bridge.py#L253

[...]

PR #5726, opened 2023-10-24
Ignore nodes from dynamo graph that represent non-XLA tensors.

ysiraichi:
Fix: #5722 

This PR adds another condition to `CapabilityBasedPartitioner`: whether its traced value is an XLA tensor or not. This is necessary, since dynamo bridge expects only XLA output tensors for things like computing the graph hash.

**Implementation summary:** we rely on `node.meta["val"]` value populated by FX proxy tensors. It provides us tensor metadata, such as the running device.

### DIFF
--- a/test/dynamo/test_bridge.py
+++ b/test/dynamo/test_bridge.py
@@ -91,6 +91,27 @@ class UpsampleModule(nn.Module):
     return (torch.randn((1, 1, 5)),)
 
 
+class CopyToCPUModule(nn.Module):
+
+  def forward(self, x):
+    return (x + x).cpu()
+
+  def get_random_inputs(self):
+    return (torch.randn((2, 2)),)
+
+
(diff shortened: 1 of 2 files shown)


JackCaoG:
This seems to regress CompileTime and ExecuteTime on other tests.

ysiraichi:
The culprit seems to be the number of partitions derived from the supported nodes:
- Inference of `resnet18` went from 1 to 18 partitions
- Training of `resnet18` went from 3 graphs of 1 partition to 3 graphs of 18, 35, and 1 partition

Maybe we should either wait for [@eellison pass](https://github.com/pytorch/xla/issues/5722#issuecomment-1777699572) or create something similar. What do you think?

miladm:
Are there any updates on this PR and its blocker? @ysiraichi

PR #5728, opened 2023-10-24
Make the git version appears in pip

alanwaketan:
Summary:
This change makes the git version appears in pip by default. For any official release, we should turn that off by setting VERSIONED_XLA_BUILD=0.

Here is how it looks like:
torch                        2.2.0a0+git6ac8451
torch-xla                    2.2.0+git43cd6b5

Test Plan:
Manual.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,8 +10,8 @@
 #     specify the version of PyTorch/XLA, rather than the hard-coded version
 #     in this file; used when we're building binaries for distribution
 #
-#   VERSIONED_XLA_BUILD
-#     creates a versioned build
+#   GIT_VERSIONED_XLA_BUILD
+#     creates a git versioned build
 #
 #   TORCH_XLA_PACKAGE_NAME
 #     change the package name to something other than 'torch_xla'
@@ -101,7 +101,7 @@ def get_git_head_sha(base_dir):
 
 def get_build_version(xla_git_sha):
   version = os.getenv('TORCH_XLA_VERSION', '2.2.0')
-  if _check_env_flag('VERSIONED_XLA_BUILD', default='1'):
+  if _check_env_flag('GIT_VERSIONED_XLA_BUILD', default='TRUE'):
     try:
       version += '+git' + xla_git_sha[:7]
     except Exception:
--- a/setup.py
+++ b/setup.py
@@ -101,9 +101,9 @@ def get_git_head_sha(base_dir):
 
 def get_build_version(xla_git_sha):
   version = os.getenv('TORCH_XLA_VERSION', '2.2.0')
-  if _check_env_flag('VERSIONED_XLA_BUILD', default='0'):
+  if _check_env_flag('VERSIONED_XLA_BUILD', default='1'):
     try:
-      version += '+' + xla_git_sha[:7]
+      version += '+git' + xla_git_sha[:7]
     except Exception:
       pass
   return version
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -101,9 +101,9 @@ def get_git_head_sha(base_dir):
 
 def get_build_version(xla_git_sha):
   version = os.getenv('TORCH_XLA_VERSION', '2.2.0')
-  if _check_env_flag('VERSIONED_XLA_BUILD', default='0'):
+  if _check_env_flag('VERSIONED_XLA_BUILD', default='1'):
     try:
-      version += '+' + xla_git_sha[:7]
+      version += '+git' + xla_git_sha[:7]
```
JackCaoG, line 106: hmm, I guess for the release build we need to turn this `VERSIONED_XLA_BUILD ` to 0?

JackCaoG (approved):

setup.py
alanwaketan: Yea, that's correct.

ManfeiBai: we might want to add a new env FLAG `RELEASED_XLA_BUILD` here, such as:
```
if _check_env_flag('VERSIONED_XLA_BUILD', default='1') and _check_env_flag(' RELEASED_XLA_BUILD=1', default='0')::
   try:
...
```

then add `RELEASED_XLA_BUILD=1` in https://github.com/pytorch/xla/blob/master/infra/tpu-pytorch-releases/artifacts.auto.tfvars for public release version

ManfeiBai (approved):
LGTM

setup.py
alanwaketan: Why adding another env? Can't we just use the VERSIONED_XLA_BUILD directly?

setup.py
```diff
@@ -101,9 +101,9 @@ def get_git_head_sha(base_dir):
 
 def get_build_version(xla_git_sha):
   version = os.getenv('TORCH_XLA_VERSION', '2.2.0')
-  if _check_env_flag('VERSIONED_XLA_BUILD', default='0'):
+  if _check_env_flag('VERSIONED_XLA_BUILD', default='1'):
```
will-cromar, line 104: Since this variable looks otherwise unused, what do you think of changing the name to be more explicit, e.g. `APPEND_GIT_SHA`?

"`VERSIONED_XLA_BUILD` makes me think of releases, which is the opposite of what we want.

alanwaketan: Sure, why not.

alanwaketan:
@will-cromar Can you take another look?

setup.py
```diff
@@ -101,9 +101,9 @@ def get_git_head_sha(base_dir):
 
 def get_build_version(xla_git_sha):
   version = os.getenv('TORCH_XLA_VERSION', '2.2.0')
-  if _check_env_flag('VERSIONED_XLA_BUILD', default='0'):
+  if _check_env_flag('GIT_VERSIONED_XLA_BUILD', default='TRUE'):
```
will-cromar, line 104: Do you know if `nightly_release` this is going to be `TRUE` or `true`?

will-cromar (approved):
Thanks!

setup.py
alanwaketan: It doesn't matter. `_check_env_flag` will convert everything to upper case.

alanwaketan:
Thanks all for the review. Pending CI now.

PR #5729, opened 2023-10-25
Fix the missing parameter error when running mp_imagenet with torchrun

vanbasten23:


test/test_train_mp_imagenet.py
```diff
@@ -375,6 +376,6 @@ def _mp_fn(index, flags):
 
 if __name__ == '__main__':
   if dist.is_torchelastic_launched():
-    _mp_fn(FLAGS)
+    _mp_fn(xu.getenv_as(xenv.LOCAL_RANK, int), FLAGS)
```
jonb377, line 379: This is set by torchrun, right? Just confirming my understanding.

jonb377 (approved):
LGTM, thanks Xiongfei

PR #5731, opened 2023-10-25
Revert "Don't set $TPU_LIBRARY_PATH during import (#5698)"

alanwaketan:
This reverts commit 146f2a0140c5217a31d55cfdebb1efb6bc05b41a.

It regresses the profiler.

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -72,16 +72,13 @@ def _aws_ec2_inf_trn_init():
 
 
 def _setup_tpu_vm_library_path() -> bool:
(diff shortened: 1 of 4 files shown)


JackCaoG (approved):
hmm, I wonder why

alanwaketan:
> hmm, I wonder why

No ideas...

will-cromar (approved):
Do you have any more details? If the profiler depends on `$TPU_LIBRARY_PATH`, we should fix this upstream.

alanwaketan:
> Do you have any more details? If the profiler depends on `$TPU_LIBRARY_PATH`, we should fix this upstream.

Do you have the more details you need in the gchat?

PR #5733, opened 2023-10-25
Support PreemptionSyncManager in XlaCoordinator

jonb377:
To support autocheckpointing upon preemption, we need to access a PreemptionSyncManager to identify sync points when a preemption has occurred.

This change additionally refactors the DistributedRuntime to be owned by the ComputationClient, since in the GPU case the ComputationClient has a direct dependency on the DistributedRuntimeClient.

This change adds the PreemptionSyncManager to the new XlaCoordinator class. The PreemptionSyncManager has the side effect of registering a SIGTERM handler, so it is not enabled by default.


torch_xla/csrc/runtime/distributed_runtime.h
```diff
@@ -3,33 +3,78 @@
...
+  // Returns true if the distributed runtime has already been initialized.
+  static bool IsInitialized() { return dist_runtime_ != nullptr; }
+
+  // Initialize the shared DistributedRuntime object. This creates a
+  // DistributedRuntimeClient on each worker, and on global_rank 0 initializes
+  // the corresponding DistributedRuntimeService.
+  static void Initialize(int global_rank, int world_size,
+                         std::string master_addr, std::string port) {
+    XLA_CHECK(!IsInitialized()) << "DistributedRuntime already initialized";
+    dist_runtime_ = std::unique_ptr<DistributedRuntime>(
```
vanbasten23, line 28: Would it guarantee the DistributedRuntime ctor only run once?

torch_xla/csrc/runtime/distributed_runtime.h
```diff
@@ -3,33 +3,78 @@
...
+        << "Must initialize distributed runtime before shutdown";
+    dist_runtime_ = nullptr;
+  }
+
+  // Retrieve the shared DistributedRuntime object.
+  static DistributedRuntime& Get() {
+    XLA_CHECK(IsInitialized())
+        << "Must initialize distributed runtime before retrieval";
+    return *dist_runtime_;
   }
```
vanbasten23, line 46: I wonder why we make `std::unique_ptr<DistributedRuntime>` a unique_ptr instead of just a DistributedRuntime&

torch_xla/csrc/runtime/distributed_runtime.h
vanbasten23, line 28: umm, previously, the reason why I made `dist_runtime_` a static variable is that it guarantees that
- it only initialize only once
- it's initialized in a thread safe manner under c++11 ([so](https://stackoverflow.com/questions/246564/what-is-the-lifetime-of-a-static-variable-in-a-c-function))
- it runs its destructor even after an exception is thrown ([so](https://stackoverflow.com/questions/6418006/are-static-objects-deleted-when-an-exception-is-thrown-or-just-local-objects)) so we can make sure the distributed runtime service is shut down eventually

(3) is the most important imo.

With the current implementation, I wonder when `dist_runtime_` is destroyed.

torch_xla/csrc/runtime/distributed_runtime.h
```diff
@@ -3,33 +3,78 @@
...
 #include <memory>
 
+#include "torch_xla/csrc/runtime/debug_macros.h"
+#include "tsl/distributed_runtime/preemption/preemption_sync_manager.h"
 #include "xla/pjrt/distributed/distributed.h"
 
 namespace torch_xla {
 namespace runtime {
 
+// DistributedRuntime serves as the point of entry for all operations which
```
vanbasten23, line 13: Thanks for improving this class, Jon!

torch_xla/csrc/runtime/distributed_runtime.h
jonb377: The `unique_ptr` takes ownership of the raw pointer, so the constructor is only called once. Due to the visibility of the private constructor, we can't directly use `make_unique`.

jonb377: `dist_runtime_` is still static, just at the class level. All three should still apply, since `dist_runtime_` still has static [storage duration](https://en.cppreference.com/w/cpp/language/storage_duration).

torch_xla/csrc/runtime/distributed_runtime.h
jonb377: The `unique_ptr` makes ownership explicit, and it will manage destructing the singleton instance when it releases ownership (through reassignment or when it is destructed).

torch_xla/csrc/runtime/distributed_runtime.h
jonb377: Whew, thanks for the tough questions Xiongfei! Really pushing my C++ knowledge to the limits 😛

I've added @will-cromar to this review as well, since he's much more c++ savvy than I am. Will, could you please double check my understanding? Mainly around the singleton instance handling.

torch_xla/csrc/runtime/distributed_runtime.h
vanbasten23: gotcha, thanks for pointing it out!

vanbasten23 (approved):
LGTM!

torch_xla/csrc/runtime/distributed_runtime.h
will-cromar: I have one general question: why don't we tie the lifetime of `DistributedRuntime` to `ComputationClient` by having it hold a `unique_ptr<DistributedRuntime>`? Is it because we never run the destructor on `ComputationClient`? Or does the lifetime of `DistributedRuntime` substantially differ from that of the `PjRtClient`?

I ask because IMO one singleton is better than two.

(If the above change is possible but non-trivial, one of us can do the refactoring in a follow-up)

torch_xla/csrc/runtime/distributed_runtime.h
```diff
@@ -3,33 +3,78 @@
...
+// required the XLA distributed runtime, such as preemption coordination.
 class DistributedRuntime {
```
will-cromar, line 15: I really dislike the naming choice for the upstream `xla::DistributedRuntime`, since it's not actually a distributed runtime. Since this class is becoming more than just a wrapper around `xla::DistributedRuntimeService` and `xla::DistributedRuntimeClient`, what do you think of changing the name to something more intuitive? e.g. `XlaCoordinator`

torch_xla/csrc/runtime/distributed_runtime.h
```diff
@@ -3,33 +3,78 @@
...
+  // through the ReachedSyncPoint API. See the documentation of
+  // tsl::PreemptionSyncManager for the full semantics:
+  // https://github.com/google/tsl/blob/3bbe663/tsl/distributed_runtime/preemption/preemption_sync_manager.h#L34
+  void ActivatePreemptionSyncManager();
+
+  // A pass-throguh API to the PreemptionSyncManager::ReachedSyncPoint.
+  // The PreemptionSyncManager must be activated within the DistributedRuntime.
+  // Returns true when the input step has been identified as a sync point, and
+  // false otherwise.
+  bool ReachedSyncPoint(int step);
```
will-cromar, line 67: Do you think it makes more sense to expose the `tsl::PreemptionSyncManager` directly as we do with the `xla::DistributedRuntimeClient`? Or do we want to restrict access to the underlying object?

torch_xla/csrc/runtime/distributed_runtime.cc
```diff
@@ -50,5 +47,22 @@ std::shared_ptr<xla::DistributedRuntimeClient> DistributedRuntime::GetClient() {
   return dist_runtime_client_;
 }
 
+void DistributedRuntime::ActivatePreemptionSyncManager() {
+  if (preemption_sync_manager_ == nullptr) {
```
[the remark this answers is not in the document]
jonb377: I was hesistant to do that, since it will register a SIGTERM handler which will cause any intentional SIGTERMs to be ignored. Open to revisiting, let me know which approach you think makes more sense!

torch_xla/csrc/runtime/distributed_runtime.h
jonb377: Actually, we should do absolutely that because of the GPU case - we should never shut down the dist runtime without also shutting down the ComputationClient for GPUs.

This helped with testability for TPUs, since it's currently easy to recreate the dist runtime without killing the computation client, but I really just need to recreate the PreemptionSyncManager between tests.

torch_xla/csrc/runtime/distributed_runtime.h
jonb377: Totally agree, XlaCoordinator it is! I'll update the pybinds as well.

torch_xla/csrc/runtime/distributed_runtime.h
jonb377: I considered that, but if the PreemptionSyncManager outlives the DistributedRuntimeClient, the program will segfault... 😢 I figured it's better to keep it hidden to avoid that edge case.

jonb377 (commented):
Thanks for the review @vanbasten23 and @will-cromar! I'll update to address the feedback.

vanbasten23:
Oh one more thing, could you help check if we have a test verifying whether the distributed runtime service is always turned down every time? I'd imagine if we comment out the line `dist_runtime_service_->Shutdown();`, the test `PJRT_DEVICE=GPU torchrun --nnodes 1 --nproc-per-node 2 pytorch/xla/test/pjrt/test_torchrun.py` should time out.

jonb377:
@vanbasten23 @will-cromar I've updated to have the ComputationClient own the XlaCoordinator. Please take a second look when you get a chance!

torch_xla/csrc/runtime/computation_client.h
```diff
@@ -30,6 +30,9 @@
 namespace torch_xla {
 namespace runtime {
 
+// Forward declaration
+class XlaCoordinator;
```
will-cromar, line 34: Why do you need the forward declaration here? Can you `#include` `XlaCoordinator` instead?

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -109,13 +109,17 @@ PjRtComputationClient::PjRtComputationClient() {
...
-        "XLA_COORDINATOR_PORT", DistributedRuntime::default_coordinator_port);
-    std::shared_ptr<xla::DistributedRuntimeClient> distributed_client =
-        DistributedRuntime::getInstance(global_process_rank, master_addr, port)
-            .GetClient();
+        "XLA_COORDINATOR_PORT", XlaCoordinator::kDefaultCoordinatorPort);
+
+    // Use the XlaCoordinator as the distributed key-value store.
+    coordinator_ = std::make_unique<XlaCoordinator>(
+        global_process_rank, global_world_size, master_addr, port);
+    auto distributed_client = coordinator_->GetClient();
```
will-cromar, line 122: nit: replace `auto` here with the actual type since it's not otherwise clear from context

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -9,7 +9,7 @@
 #include "pjrt_computation_client.h"
 #include "torch_xla/csrc/runtime/computation_client.h"
 #include "torch_xla/csrc/runtime/debug_macros.h"
-#include "torch_xla/csrc/runtime/distributed_runtime.h"
+//#include "torch_xla/csrc/runtime/xla_coordinator.h"
```
will-cromar, line 12: Can you uncomment this?

torch_xla/csrc/runtime/xla_coordinator.h
```diff
@@ -0,0 +1,58 @@
+#ifndef PTXLA_RUNTIME_COORDINATOR_H_
+#define PTXLA_RUNTIME_COORDINATOR_H_
+
+#include <memory>
+
+#include "tsl/distributed_runtime/preemption/preemption_sync_manager.h"
+
+// Forward declaration
```
will-cromar, line 8: Same here. Do you need the forward declaration?

torch_xla/csrc/runtime/computation_client.h
```diff
@@ -350,6 +353,17 @@ class ComputationClient {
   // the local devices will be waited for.
   virtual void WaitDeviceOps(const std::vector<std::string>& devices) = 0;
 
+  // Check whether the XlaCoordinator has been initialized.
+  virtual bool CoordinatorInitialized() const = 0;
```
will-cromar, line 357: Do these need to be virtual? It looks like the implementations below don't depend on the underlying runtime client.

will-cromar (approved):
Overall LGTM.

FYI, our style guide cautions against forward declarations of entities in other projects, even if it saves compile time: https://google.github.io/styleguide/cppguide.html#Forward_Declarations

If you forward-declared the `DistributedRuntime` classes to unravel e.g. a macro conflict or circular dependency, please leave a comment explaining why.

test/spmd/test_xla_distributed_checkpoint.py
```diff
@@ -340,6 +343,8 @@ def setUp(self):
     # Initialize the a minimal process group
     dist.init_process_group(
         backend='gloo', init_method='tcp://127.1:8932', world_size=1, rank=0)
```
will-cromar, line 345: TIL `127.1` is [a thing](https://superuser.com/questions/614001/why-can-i-ping-127-1) :dizzy_face: 

Can you change this to `localhost`?

torch_xla/csrc/runtime/computation_client.h
jonb377: I tried that, but since XlaCoordinator includes a pjrt header, it redefined a ton of log macros and broke the build...

torch_xla/csrc/runtime/computation_client.h
jonb377: The XlaCoordinator depends on PJRT, so I kept it separate.

torch_xla/csrc/runtime/pjrt_computation_client.cc
jonb377: Oops, I should have proofread before requesting review, thanks Will!

torch_xla/csrc/runtime/xla_coordinator.h
jonb377: I only need one to avoid the log macro issue, this one can be removed.

test/spmd/test_xla_distributed_checkpoint.py
jonb377: Haha you don't like IP shorthand?! 😱 It's the same as `::1` but for IPv4. I guess it's more niche, I'll replace with `localhost`.

torch_xla/csrc/runtime/computation_client.h
jonb377: Added a comment and removed the redundant forward declaration in `xla_coordinator.h`

torch_xla/csrc/runtime/distributed_runtime.h
jonb377: Refactored into ComputationClient

torch_xla/csrc/runtime/runtime.cc
```diff
@@ -34,8 +36,10 @@ ComputationClient* CreateClient() {
 }  // namespace
 
 ComputationClient* GetComputationClient() {
-  std::call_once(g_computation_client_once,
-                 [&]() { g_computation_client = std::move(CreateClient()); });
+  std::call_once(g_computation_client_once, [&]() {
+    g_computation_client = std::move(CreateClient());
+    std::atexit([]() { delete g_computation_client.load(); });
```
jonb377, line 41: cc @will-cromar, we need to ensure the ComputationClient is deleted before exiting so that all clients can gracefully shutdown before the DistributedRuntimeService is destroyed.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -109,13 +109,18 @@ PjRtComputationClient::PjRtComputationClient() {
     bool async = sys_util::GetEnvBool(env::kEnvPjrtAsyncGpuClient, true);
     int local_process_rank = sys_util::GetEnvInt(env::kEnvPjRtLocalRank, 0);
     int global_process_rank = sys_util::GetEnvInt("RANK", local_process_rank);
+    int local_world_size = sys_util::GetEnvInt("LOCAL_WORLD_SIZE", 1);
+    int global_world_size = sys_util::GetEnvInt("WORLD_SIZE", local_world_size);
     std::string master_addr =
-        runtime::sys_util::GetEnvString("MASTER_ADDR", "localhost");
+        runtime::sys_util::GetEnvString("MASTER_ADDR", "127.0.0.1");
```
jonb377, line 115: cc @vanbasten23, it looks like our CI hosts don't support ipv6, but they resolve `localhost` to `::1`:

```
"os_error":"Address family not supported by protocol","syscall":"socket","target_address":"[::1]:8547"
```

https://github.com/pytorch/xla/actions/runs/6673957659/job/18140734664?pr=5733

torch_xla/csrc/runtime/runtime.cc
will-cromar: Is it possible to do with from `XLAC._prepare_to_exit`? We need to guarantee that `ComputationClient::PrepareToExit()` is not called after this shutdown hook

will-cromar: side note: this looks a lot like just holding a static variable that will be destroyed at the end of the process. I'll look into refactoring how we handle `g_computation_client` after this PR merges.

torch_xla/csrc/runtime/computation_client.h
```diff
@@ -30,6 +30,12 @@
 namespace torch_xla {
 namespace runtime {
 
+// Forward declare XlaCoordinator to avoid logging macro redefinition from the
+// transitively included PJRT header.
+// TODO(jonbolin): We need a way to ensure the right macros are included
```
will-cromar, line 35: logging macros are cursed :frowning_face:

torch_xla/csrc/runtime/runtime.cc
jonb377: I see, both are registered atexit.. I'll make this a bit more controlled and set g_computation_client to nullptr.

I'm not sure why, but [using PrepareToExit](https://github.com/pytorch/xla/commit/92ba0796fd8402b2d65f177454229d89a1bae737) resulted in segfaults in MP tests: https://github.com/pytorch/xla/actions/runs/6673957659/job/18140734664

will-cromar: I suggest changing the pybind to actually delete the client here after calling the client's `PrepareToExit`: https://github.com/pytorch/xla/blob/1f0e972a6dfc5ed3205b23d0fccac8ebb128584f/torch_xla/csrc/init_python_bindings.cpp#L98

jonb377: That's safer, I'll make the change.

yeounoh (approved):
LGTM, thanks!

alanwaketan (commented):
I guess I had a hard time of imagining how this is going to be incorporated to the ckpt mgr.

jonb377:
> I guess I had a hard time of imagining how this is going to be incorporated to the ckpt mgr. @jonb377 Can you point me to some examples?

@alanwaketan CheckpointManager will initialize the PreemptionSyncManager on construction and call into `_sync_point_reached` each step to check for preemption in `should_save` - I'll open a draft PR to illustrate.

jonb377:
@alanwaketan See https://github.com/pytorch/xla/commit/5fdce130367c242ee8531a1dbd5502372584afcb for the intended usage.

jonb377:
Synced up with @will-cromar  - we'll take the initial singleton approach and follow up with a refactor to transfer ownership to the ComputationClient.

alanwaketan:
> I guess I had a hard time of imagining how this is going to be incorporated to the ckpt mgr.



> > I guess I had a hard time of imagining how this is going to be incorporated to the ckpt mgr. @jonb377 Can you point me to some examples?
> 
> @alanwaketan CheckpointManager will initialize the PreemptionSyncManager on construction and call into `_sync_point_reached` each step to check for preemption in `should_save` - I'll open a draft PR to illustrate.

I see, that makes sense.

alanwaketan (approved):
LGTM.

torch_xla/csrc/runtime/distributed_runtime.h
jonb377: I left as-is for now to keep this change minimal, we can revisit in the upcoming refactor.

PR #5736, opened 2023-10-25
Set --xla_latency_hiding_scheduler_rerun to 1

alanwaketan:
Summary:
This flag will rerun the latency hidding scheduler if the default
shared memory limit 95% leads to OOM. Each rerun will choose a value
0.9x of the previous run, and the number of rerun is set to 1 now.
Shared memory limit refers to --xla_tpu_scheduler_percent_shared_memory_limit.
Lower shared memory limit means less communiation and computation overlapping,
and thus worse performance.

Test Plan:
Tested on Llama 2 7B on V4-32.

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -33,6 +33,8 @@ def _setup_xla_flags():
   # shared memory limit 95% leads to OOM. Each rerun will choose a value
   # 0.9x of the previous run, and the number of rerun is set to 1 now.
   # Shared memory limit refers to --xla_tpu_scheduler_percent_shared_memory_limit.
+  # Lower shared memory limit means less communiation and computation overlapping,
+  # and thus worse performance.
   flags = _set_missing_flags(flags,
                              (('xla_latency_hiding_scheduler_rerun', '1'),))
   os.environ['XLA_FLAGS'] = ' '.join(flags)
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -29,6 +29,12 @@ def _setup_xla_flags():
       flags, (('xla_gpu_simplify_all_fp_conversions', 'false'),))
   flags = _set_missing_flags(flags,
                              (('xla_gpu_force_compilation_parallelism', '8'),))
+  # This flag will rerun the latency hidding scheduler if the default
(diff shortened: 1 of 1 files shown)


JackCaoG:
lgtm, let's wait for the manual run passed then merge this pr

torch_xla/__init__.py
```diff
@@ -29,6 +29,14 @@ def _setup_xla_flags():
...
   flags = _set_missing_flags(flags,
                              (('xla_gpu_force_compilation_parallelism', '8'),))
+  # This flag will rerun the latency hidding scheduler if the default
+  # shared memory limit 95% leads to OOM. Each rerun will choose a value
+  # 0.9x of the previous run, and the number of rerun is set to 1 now.
+  # Shared memory limit refers to --xla_tpu_scheduler_percent_shared_memory_limit.
+  # Lower shared memory limit means less communiation and computation overlapping,
+  # and thus worse performance.
+  flags = _set_missing_flags(flags,
+                             (('xla_latency_hiding_scheduler_rerun', '1'),))
```
jonb377, line 39: Nice! I was actually wondering about this, we set `XLA_FLAGS` here, but in other cases we pass the flags through `LIBTPU_INIT_ARGS`. Do you know if there's a difference? If you saw the appropriate log output from the test, it seems both work...

JackCaoG: oh right.. sorry I missed this. so for everything in `compiler/xla/xla.proto` we used `XLA_FLAGS`. `xla_latency_hiding_scheduler_rerun ` is one of those TPU specified flags that needs to be passed in with `LIBTPU_INIT_ARGS`.

alanwaketan: Good catch. I miss this lol.

jonb377: Ah, makes sense that `LIBTPU_INIT_ARGS` would be tpu-specific lol, thanks @JackCaoG. Is there a rule of thumb to tell which flag goes where? I'm thinking in terms of hashing the compilation environment, I suppose we'll just need to ensure both env vars are included.

jonb377 (approved):
LGTM, thanks Jiewen!

alanwaketan:
Thanks for the review, Jon.

PR #5743, opened 2023-10-26
Fix XLA tensor storage device by using `XlaDeviceToAtenDevice`.

ysiraichi:
Discussion: https://github.com/pytorch/pytorch/issues/111506

This PR replaces `backendDeviceToAtenDevice` by `XlaDeviceToAtenDevice` when instantiating a new storage in the newly created XLA tensor.

**Problem:** there are 2 problems:
1. `backendDeviceToAtenDevice` doesn't take into consideration the device type
2. XLA uses `XlaDeviceType` enum as the device type, while eager uses `c10::DeviceType`

**Solution:** use `XlaDeviceToAtenDevice` function, since we know that the input is an XLA device anyway.

### DIFF
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -23,6 +23,16 @@ class AtenXlaTensorTest : public AtenXlaTensorTestBase {};
 
 }  // namespace
 
+TEST_F(AtenXlaTensorTest, TestStorage) {
+  torch::Tensor a = torch::tensor({0.0});
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    XLATensorPtr xla_tensor_a = bridge::GetXlaTensor(xla_a);
+    EXPECT_EQ(xla_a.device(), xla_tensor_a->Storage().device());
+    AllClose(a, xla_a);
+  });
+}
+
 TEST_F(AtenXlaTensorTest, TestEmpty) {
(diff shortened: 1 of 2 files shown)


ysiraichi:
@JackCaoG The failing CI doesn't seem to be related to this PR. Can we re-run it?

JackCaoG:
Let me also give you write access so you can rerun CI too

alanwaketan (commented):
I wonder have you experienced any similar bugs on python level?

This is originally a hack see: https://github.com/pytorch/xla/blob/master/torch_xla/csrc/tensor.h#L334

We don't expect people to actually use `storage_`.

ysiraichi:
As of now, we do hit that bug when running [this example](https://github.com/pytorch/pytorch/issues/111506#issue-1950643605). That said, there are 2 things I'd like to clarify:
- We did decide to solve that bug, specifically, in some other way
- I still think we should be consistent with aten devices, here
    - e.g. `XLATensorImpl` also calls `bridge::XlaDeviceToAtenDevice` for constructing itself

alanwaketan (approved):
LGTM.

alanwaketan:
> As of now, we do hit that bug when running [this example](https://github.com/pytorch/pytorch/issues/111506#issue-1950643605). That said, there are 2 things I'd like to clarify:
> 
> * We did decide to solve that bug, specifically, in some other way
> * I still think we should be consistent with aten devices, here
>   
>   * e.g. `XLATensorImpl` also calls `bridge::XlaDeviceToAtenDevice` for constructing itself

Got it. Hopefully we can remove this workaround on fake storage soon.

PR #5749, opened 2023-10-30
Don't preserve variable names from Pytorch when exporting to SavedModel.

haozha111:
In this change, I remove the `name=k` when exporting to SavedModel, since the pytorch naming might not be accepted by SavedModel APIs and result in a export failure later.


JackCaoG:
Do we still need this pr?

PR #5750, opened 2023-10-30
Destroy the ComputationClient when the program exits

will-cromar:
- Remove `ComputationClient::PrepareToExit()`. Any non-trivial destruction can be added to the destructor. @jonb377 found that in practice, any attempt to use `PrepareToExit` resulted in a segfault, likely due to the sequencing of Python's `atexit` with other teardown.
- The `static` variable initialization inside of `GetComputationClient` is thread safe after C++11: https://codereview.stackexchange.com/a/197494
- Confirmed that `ComputationClient` is destroyed by adding a print to the destructor.

### DIFF
--- a/torch_xla/csrc/runtime/runtime.cc
+++ b/torch_xla/csrc/runtime/runtime.cc
@@ -10,10 +10,11 @@ namespace torch_xla {
 namespace runtime {
 namespace {
 
-std::atomic<ComputationClient*> g_computation_client(nullptr);
-std::once_flag g_computation_client_once;
+std::atomic<bool> g_computation_client_initialized(false);
 
 ComputationClient* CreateClient() {
+  bool was_initialized = g_computation_client_initialized.exchange(true);
+  XLA_CHECK(!was_initialized) << "ComputationClient already initialized";
   if (sys_util::GetEnvBool("XLA_DUMP_FATAL_STACK", false)) {
     tsl::testing::InstallStacktraceHandler();
   }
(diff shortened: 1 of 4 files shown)


jonb377 (approved):
LGTM, thanks Will!

jonb377:
Let's wait for the GPU tests to finish before merging, the two left running are what segfaulted for me.

JackCaoG (approved):

torch_xla/csrc/runtime/runtime.cc
```diff
@@ -34,13 +35,12 @@ ComputationClient* CreateClient() {
 }  // namespace
 
 ComputationClient* GetComputationClient() {
-  std::call_once(g_computation_client_once,
-                 [&]() { g_computation_client = std::move(CreateClient()); });
-  return g_computation_client.load();
+  static auto client = std::unique_ptr<ComputationClient>(CreateClient());
```
vanbasten23, line 38: If the static variable is guaranteed to be initialized once, do you still need the function `GetComputationClientIfInitialized` and the variable `std::atomic<bool> g_computation_client_initialized(false);`?

will-cromar: The client is guaranteed to be initialized once on the first time we call into `GetComputationClient`. There are a couple of calls that want to tell if `client` has been initialized (ie comparing `GetComputationClientIfInitialized` to `nullptr`), but there's not a way I can see to tell if `client` has been initialized. As soon as you call `GetComputationClient`, it gets created. That's why I added the extra flag outside of `GetComputationClient`.

On a second look, I think it would actually be clearer to eliminate `ComputationClient* GetComputationClientIfInitialized` and replace it with `bool ComputationClientIsInitialzed()`.

PR #5751, opened 2023-10-31
fix squeeze op lowering issue when dim is not in sorted order

zpcore:
We notice that torch.ops.aten.sequeeze(tensor, dims) will throw out an error if the dims is not in sorted order after we convert it to Canonical dimensions. Meanwhile if the dims = [-1], it will also throw out an error.

The issue is mostly due to the logic [here](https://github.com/pytorch/xla/blob/63ea76cdf0891d3fc7865757c18f270819cc40f8/torch_xla/csrc/tensor_methods.cpp#L2481C1-L2483C4).

We make the logic fix and add the test case in the pr.

### DIFF
--- a/torch_xla/csrc/data_ops.h
+++ b/torch_xla/csrc/data_ops.h
@@ -49,6 +49,9 @@ xla::XlaOp BuildMaskedFillScalar(xla::XlaOp input, xla::XlaOp mask,
 std::vector<int64_t> BuildSqueezedDimensions(
     absl::Span<const int64_t> dimensions, int64_t squeeze_dim);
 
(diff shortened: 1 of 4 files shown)


torch_xla/csrc/data_ops.h
```diff
@@ -49,6 +49,9 @@ xla::XlaOp BuildMaskedFillScalar(xla::XlaOp input, xla::XlaOp mask,
 std::vector<int64_t> BuildSqueezedDimensions(
```
lsy323, line 49: The above one should be a special case of the 2nd one, shall we just keep the second one?

lsy323 (approved):
LGTM, thanks!

PR #5759, opened 2023-11-02
Make the pjrt gpu allocator configurable

anw90:
This PR aims to make the pjrt gpu allocator configurable.
The default value of  memory allocator fraction is [0.75](https://github.com/openxla/xla/blob/7ab5df624ff1d98804999b03b21abecd14ec57a6/xla/pjrt/gpu/gpu_helpers.h#L41), which is too small.

### DIFF
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -64,6 +64,11 @@ xla::Shape host_output_shape(xla::PjRtBuffer* buffer) {
 
 xla::GpuAllocatorConfig GetGpuAllocatorConfig() {
   auto allocator_config = xla::GpuAllocatorConfig{};
+  if (sys_util::GetEnvString(env::kEnvPjrtAllocatorCudaAsync, "").empty() &&
+      sys_util::GetEnvString(env::kEnvPjrtAllocatorPreallocate, "").empty() &&
+      sys_util::GetEnvString(env::kEnvPjrtAllocatorFraction, "").empty()) {
+    return allocator_config;
+  }
   if (sys_util::GetEnvBool(env::kEnvPjrtAllocatorCudaAsync, false)) {
     allocator_config.kind = xla::GpuAllocatorConfig::Kind::kCudaAsync;
   }
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -70,7 +70,7 @@ xla::GpuAllocatorConfig GetGpuAllocatorConfig() {
   allocator_config.preallocate =
       sys_util::GetEnvBool(env::kEnvPjrtAllocatorPreallocate, true);
   allocator_config.memory_fraction =
-      sys_util::GetEnvDouble(env::kEnvPjrtAllocatorFraction, 0.9);
+      sys_util::GetEnvDouble(env::kEnvPjrtAllocatorFraction, 0.75);
   return allocator_config;
 }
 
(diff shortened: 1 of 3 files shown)


torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -62,6 +62,18 @@ xla::Shape host_output_shape(xla::PjRtBuffer* buffer) {
...
+xla::GpuAllocatorConfig GetGpuAllocatorConfig() {
+  auto allocator_config = xla::GpuAllocatorConfig{};
+  if (sys_util::GetEnvBool(env::kEnvPjrtAllocatorCudaAsync, false)) {
+    allocator_config.kind = xla::GpuAllocatorConfig::Kind::kCudaAsync;
+  }
+  allocator_config.preallocate =
+      sys_util::GetEnvBool(env::kEnvPjrtAllocatorPreallocate, true);
+  allocator_config.memory_fraction =
+      sys_util::GetEnvDouble(env::kEnvPjrtAllocatorFraction, 0.9);
+  return allocator_config;
```
vanbasten23, line 74: This seems to change the default behavior. If none of kEnvPjrtAllocatorCudaAsync, kEnvPjrtAllocatorPreallocate, kEnvPjrtAllocatorFraction is set, could you just return xla::GpuAllocatorConfig{} as before?

vanbasten23:
I wonder what problem you run into when memory allocator fraction is set to 0.75.

anw90:
> I wonder what problem you run into when memory allocator fraction is set to 0.75.

For 80G H100/A100, there is 20G memory waste if the memory allocator fraction is set to 0.75. For distributed training job, NCCL does not requires as much memory.

GPU memory is particularly constrained, especially during LLM training. Therefore, setting memory allocator fraction to 0.9 can improve the memory usage more efficiency.

torch_xla/csrc/runtime/pjrt_computation_client.cc
anw90: If kEnvPjrtAllocatorFraction is set to 0.75, it is the same as xla::GpuAllocatorConfig{}. Should the default value be changed from 0.9 to 0.75?

anw90: In our previous experience with GPUs, 0.9 is considered a reasonable value for the memory fraction.

vanbasten23: Yeah, let's use the default value from xla:gpu https://github.com/openxla/xla/blob/7ab5df624ff1d98804999b03b21abecd14ec57a6/xla/pjrt/gpu/gpu_helpers.h#L41-L60.
With the new flags, it gives you the flexibility to choose the best configuration that suits your needs.

anw90: done

vanbasten23: Actually, if none of kEnvPjrtAllocatorCudaAsync, kEnvPjrtAllocatorPreallocate, kEnvPjrtAllocatorFraction is set, could you just return xla::GpuAllocatorConfig{} as before? 

The reason is that with the current implementation if xla:GPU change their default value, then we have to update ours and we may forget to do so or don't know they have changed the values. By just returning xla::GpuAllocatorConfig{}, we can use whatever default value xla:GPU has set.

anw90: ok. When none of  kEnvPjrtAllocatorCudaAsync, kEnvPjrtAllocatorPreallocate, kEnvPjrtAllocatorFraction is set, just return xla::GpuAllocatorConfig{}.

vanbasten23 (approved):
LGTM

PR #5770, opened 2023-11-06
Add GKE support and various usability improvements in CheckpointManager

jonb377:
This change addresses a few usability issues with CheckpointManager:
- In GKE, TPU workers are identified by hostname rather than IP. This change translates hostname to IP before master IP discovery. This is backwards compatible with raw IP addresses, for which [socket.gethostbyname](https://docs.python.org/3/library/socket.html#socket.gethostbyname) is a no-op.
- The asynchronous worker thread is replaced with a ThreadPoolExecutor to ensure that all async work completes before the main process can exit.
- Previously, the caller was responsible for ensuring the state_dict had complete data before starting the checkpoint. This change adds an additional `_wait_for_data` operation before processing the state_dict to ensure device data is ready.
- If the target checkpoint path doesn't exist, this change will create it on behalf of the user.
- This change pulls the checkpointing logic into a helper function to be shared between synchronous and async checkpointing, guarded by a mutex to ensure only a single checkpoint is ever active.

### DIFF
--- a/torch_xla/experimental/distributed_checkpoint/manager.py
+++ b/torch_xla/experimental/distributed_checkpoint/manager.py
@@ -2,11 +2,11 @@ import fsspec
 import logging
 import os
 import pickle
-import queue
 import threading
 import torch.distributed as dist
 import torch.distributed.checkpoint as dist_cp
 import torch_xla
+import torch_xla.core.xla_model as xm
 import torch_xla.runtime as xr
 import torch_xla.experimental.distributed_checkpoint as xc
 import traceback
@@ -16,6 +16,7 @@ from datetime import datetime
 from collections import deque
 from fsspec.core import url_to_fs
 from os.path import basename
+from concurrent.futures import ThreadPoolExecutor, wait
 from typing import Deque, List, Optional, Union
 from torch.distributed.checkpoint.metadata import STATE_DICT_TYPE
 from ._helpers import _sharded_cpu_state_dict
@@ -81,7 +82,7 @@ class CheckpointManager:
   step_period, as would be the case in auto checkpointing.
 
   This class is inspired by Orbax's CheckpointManager, which can be found here:
-  https://github.com/google/orbax/blob/efc079c4e5b437782a80138913d322cb3ed365c7/checkpoint/orbax/checkpoint/checkpoint_manager.py
+  https://github.com/google/orbax/blob/efc079c/checkpoint/orbax/checkpoint/checkpoint_manager.py
   """
 
   # The base path to write checkpoints to. Each checkpoint taken by the manager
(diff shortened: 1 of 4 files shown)


torch_xla/experimental/distributed_checkpoint/manager.py
```diff
@@ -215,6 +202,35 @@ def _release_oldest_checkpoints(self):
         oldest_chkpt = self._tracked_chkpts.popleft()
         self._delete_chkpt_at_step(oldest_chkpt.step)
 
+  def _wait_for_data(self):
+    xm.mark_step()
```
yeounoh, line 206: you can do `xm.mark_step(wait=True)` instead.

yeounoh (approved):
LGTM

jonb377:
cc @ wz337

torch_xla/experimental/distributed_checkpoint/manager.py
jonb377: I noticed that `xm.mark_step(wait=True)` doesn't actually block on device ops - I guess that's a bug. I'll open a follow-up item for this.

alanwaketan (commented):
With this _wait_for_data change, will the on_demand ckpt slows down the training process a lot? I remember the should_save is called every step. It effectively breaks tracing and computation overlapping.

jonb377:
> With this _wait_for_data change, will the on_demand ckpt slows down the training process a lot? I remember the should_save is called every step. It effectively breaks tracing and computation overlapping.

`wait_for_data` only happens when a checkpoint should be taken, in which case the process would have blocked on data becoming available anyway. We do lose tracing/computation overlap when taking a checkpoint, but that was the case before as well. This just fixes a couple of small issues:
1. Previously when `mark_step` wasn't called before triggering the checkpoint, each tensor would require a compilation/execution, potentially of a large graph
2. An edge case where the checkpoint was taken too quickly and the process didn't block on data availability, so the data was still placeholder and caused an error.

alanwaketan:
> > With this _wait_for_data change, will the on_demand ckpt slows down the training process a lot? I remember the should_save is called every step. It effectively breaks tracing and computation overlapping.
> 
> `wait_for_data` only happens when a checkpoint should be taken, in which case the process would have blocked on data becoming available anyway. We do lose tracing/computation overlap when taking a checkpoint, but that was the case before as well. This just fixes a couple of small issues:
> 
> 1. Previously when `mark_step` wasn't called before triggering the checkpoint, each tensor would require a compilation/execution, potentially of a large graph
> 2. An edge case where the checkpoint was taken too quickly and the process didn't block on data availability, so the data was still placeholder and caused an error.

That makes sense to me. Thanks for confirming!

PR #5776, opened 2023-11-07
[SPMD] suppor DTensor API integration

yeounoh:
This is a follow-up to https://github.com/pytorch/pytorch/issues/92909, DTensor & XLA integration. In this PR, we address
- move PyTorch/XLA SPMD API to `toch_xla/distributed/spmd`
- introduce `xla` DTensor API to `torch_xla/distributed/spmd` so DTensor implementation can access.


yeounoh:
Tested locally, @JackCaoG @jonb377 @alanwaketan I am moving/grouping the SPMD API under `toch_xla/experimental/spmd`. Later we can move the `spmd` package out of `experimental`, when we graduate Beta.

docs/spmd.md
```diff
@@ -130,7 +130,7 @@ In general, SPMD programs should create a single mesh and reuse it for all shard
 Mesh nicely abstracts how the physical device mesh is constructed. Users can arrange devices in any shape and order using the logical mesh. However, one can define a more performant mesh based on the physical topology, especially when it involves Data Center Network (DCN) cross slice connections. HybridMesh creates a mesh which gives good performance out of the box for such multislice environments. It accepts ici\_mesh\_shape and dcn\_mesh\_shape which denote logical mesh shapes of inner and outer network.
 
 ```python
-from torch_xla.experimental.xla_sharding import HybridMesh
```
JackCaoG, line 133: I think @will-cromar did something to aliasing the old experimental PJRT API with the new one(and throw a warning message) when moving pjrt out of experimental. I think we should do the same for SPMD, otherwise we break the backward compatibility

yeounoh: Agreed, I will look into it and add some warning msgs to the `experimental.spmd/__init__.py` when we move out of experimental.

yeounoh:
```
>>> import torch_xla.experimental.xla_sharding as xs
<stdin>:1: DeprecationWarning: Importing from `torch_xla.experimental.xla_sharding` will be deprecated after 2.2 release. Please use `torch_xla.experimental.spmd` instead.
```

JackCaoG: do you want to just move it out of experimental all-togeather? I think we can call it beta in 2.2 release

yeounoh: Ack, sounds good to me.

yeounoh: Done.

yeounoh:
> Tested locally, @JackCaoG @jonb377 @alanwaketan I am moving/grouping the SPMD API under `toch_xla/experimental/spmd`. Later we can move the `spmd` package out of `experimental`, when we graduate Beta.

Had a chat with @JackCaoG , we will move out of experimental and to distributed from BETA.

torch_xla/distributed/spmd/api.py
```diff
@@ -0,0 +1,196 @@
...
+
+import torch
+
+import torch.nn as nn
+from torch.distributed._tensor.device_mesh import DeviceMesh
+from torch.distributed._tensor.placement_types import Placement, Replicate
+
+log = logging.getLogger(__name__)
+
+TORCH_XLA_INITIALIZED = False
```
JackCaoG, line 14: I don't quite understand this `TORCH_XLA_INITIALIZED ` flag

torch_xla/distributed/spmd/api.py
```diff
@@ -0,0 +1,196 @@
...
+
+  @wraps(func)  # pyre-ignore[6]
+  def wrapper(
+      self,
+      *args: Tuple[object],
+      **kwargs: Dict[str, Any]  # type: ignore[misc]
+  ) -> None:
+    if TORCH_XLA_INITIALIZED:
+      # TODO(yeounoh) replace this with xr.use_spmd() when we deprecate the flag.
+      os.environ["XLA_USE_SPMD"] = "1"
```
JackCaoG, line 41: why not just call `xr.use_spmd()` now?

JackCaoG (commented):
I didn't read the whole or too careful since I felt like most of them are just moving the code around. Is that true?

docs/spmd.md
```diff
@@ -46,8 +46,8 @@ import numpy as np
 import torch
 import torch_xla.core.xla_model as xm
 import torch_xla.runtime as xr
-import torch_xla.experimental.xla_sharding as xs
-from torch_xla.experimental.xla_sharding import Mesh
+import torch_xla.distributed.spmd.xla_sharding as xs
```
jonb377, line 49: What do you think about making the import path just `import torch_xla.distributed.spmd as xs`? Not sure if `xs` is still the right abbreviation in that case (maybe we can think of it as `x`la `s`pmd), but it would make the `spmd` package the single entrypoint in userspace.

jonb377 (approved):
LGTM, thanks Yeounoh!

torch_xla/distributed/spmd/api.py
yeounoh: there could be a case where the caller already has initialized some tensors before entering the API. We may want to support both, at least in the implementation. I will leave as a TODO here.

torch_xla/distributed/spmd/api.py
yeounoh: Ah, we don't need this anymore, due to the fact that this code now lives in torch_xla.

docs/spmd.md
yeounoh: Good point, let's do that.

JackCaoG (approved):

alanwaketan (commented):
Moving forward, should we use dtensor directly even within our codebase?

@jonb377 do you want to update HF to reflect this change? We can maybe a couple of days in case this get reverted : ).

JackCaoG:
@alanwaketan for HF my major concern is the backward compatibility. I think We should at least wait until 2.2 release is out before updating.

alanwaketan:
> @alanwaketan for HF my major concern is the backward compatibility. I think We should at least wait until 2.2 release is out before updating.

Sounds good to me.

PR #5777, opened 2023-11-08
Refactor type conversion functions out of `//torch_xla/csrc:tensor`

will-cromar:
This helps prevent circular dependencies. See #5772 for motivation.

- Create `dtype.cpp/h` for type conversions.
- Attempt to make naming more clear and separate "pure" conversions from those that upcast/downcast based on the device.
- Consolidate some giant `switch` statements.


test/cpp/cpp_test_util.cpp
```diff
@@ -307,7 +307,7 @@ std::vector<at::Tensor> Fetch(
   std::vector<at::Tensor> tensors;
   for (auto& literal : literals) {
     tensors.push_back(MakeTensorFromXlaLiteral(
-        literal, TensorTypeFromXlaType(literal.shape().element_type())));
+        literal, MaybeUpcastForHost(literal.shape().element_type())));
```
JackCaoG, line 310: I understand the motivation of this renaming but it made it less obvious to see this function does a conversation from XLA type to Tensor type. `MaybeUpcastXlaTypeToTensorType`? Same for the downcast.

will-cromar: Done. Renamed to `MaybeUpcastToHostTorchType` and `MaybeDowncastToXlaDeviceType` to clarify the return type.

JackCaoG (approved):

PR #5781, opened 2023-11-08
Lower AtenFull op

### LINKED ISSUE #5780: Op lowering for AtenFull
In https://github.com/pytorch/xla/issues/5764 `AtenFull` is listed as an op that needs to be lowered in XLA. This issue is to track the work on lowering this op.

danielvegamyhre:
Fixes #5780 

This is my first attempt at op lowering, and I have a couple questions:

1) Do we want to fallback to the pytorch/aten CPU version of this op if the `layout` and/or `pin_memory` values are not the default? I saw this pattern in other places in in `aten_xla_type.cpp`, and the `tensor_methods::full` function in XLA doesn't support these parameters, so it seemed like the right thing to do, but I'm not sure.

2) I notice `full` was not listed in `xla_native_functions.yaml` (as expected) but there IS an existing test for it in `test_aten_xla_tensor_1.cpp` called `TestFull`, which seems to verify that we use `xla::empty` followed by `xla::fill_` to create a tensor with the fill value. So I just changed this test to check that we are instead using the new `full` method implemented in this PR. I confirmed the test is passing with these changes, and failing without the changes. Is this the correct thing to do here?

### DIFF
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -192,6 +192,7 @@ supported:
   - floor_divide
   - fmod.Scalar
   - fmod.Tensor
+  - full
   - gather
   - gelu
   - gelu_backward
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -139,8 +139,7 @@ TEST_F(AtenXlaTensorTest, TestFull) {
   });
 
   ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
-  ExpectCounterChanged("xla::empty", cpp_test::GetIgnoredCounters());
-  ExpectCounterChanged("xla::fill_", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::full", cpp_test::GetIgnoredCounters());
 }
 
 TEST_F(AtenXlaTensorTest, TestFullLike) {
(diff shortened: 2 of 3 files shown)


wonjoolee95 (approved):
Thanks for working on this, @danielvegamyhre!

> 1. Do we want to fallback to the pytorch/aten CPU version of this op if the `layout` and/or `pin_memory` values are not the default? I saw this pattern in other places in in `aten_xla_type.cpp`, and the `tensor_methods::full` function in XLA doesn't support these parameters, so it seemed like the right thing to do, but I'm not sure.


Yes, this is the correct behavior. We can fallback to CPU for now for `full` like you did in this PR. 

> 2. I notice `full` was not listed in `xla_native_functions.yaml` (as expected) but there IS an existing test for it in `test_aten_xla_tensor_1.cpp` called `TestFull`, which seems to verify that we use `xla::empty` followed by `xla::fill_` to create a tensor with the fill value. So I just changed this test to check that we are instead using the new `full` method implemented in this PR. I confirmed the test is passing with these changes, and failing without the changes. Is this the correct thing to do here?

Good find -- so most likely what's happening here is that when `full` is not lowered, it's decomposed `empty` and `fill`. But since this PR lowers `full`, what you did here -- removing the counter checks for `empty` and `fill` -- is correct, thanks!

LGTM.

miladm:
Awesome! Thank you @danielvegamyhre

PR #5785, opened 2023-11-09
Remove some unused code from `csrc/runtime`

will-cromar:
- `nccl_distributed` and `async_task` are completely unused
- hash functions and `Unique` already exist upstream
- Delete some more random unused or duplicated functions from `util.h`

`MultiWait` is also duplicated upstream, but I have separate plans for that one.


JackCaoG:
I don't think `Unique` and `optional` are the same, when you try to call `set` on a Unique it will give a error. I don't think optional has that property.

will-cromar:
Got it. Switched our code to use the upstream copy of `Unique`.

JackCaoG (approved):

alanwaketan (commented):
Code crusher!

Issue #5768, opened 2023-11-03
How to provide sharding annotation for MpDeviceLoader when data has different dimensions


hanzhi713:
## ❓ Questions and Help

Let's say my dataloader yields a dict when iterating over and the members of this dict has different dimensions
```python
{
    "input_ids": shape = (batch, seq),
    "masks": shape = (batch, seq, seq),
}
```

`pl.MpDeviceLoader` appears to only able to provide one sharding annotation. I'm currently using it like this:
```python
data_loader = pl.MpDeviceLoader(
    data_loader,
    dev,
    input_sharding=xs.ShardingSpec(mesh, ('data', None, None)))
```
Obviously, ('data', None, None) is not valid for `input_ids` which has only 2 dimensions. But this seems to work. I wonder what's the proper way of using `MpDeviceLoader` in this case.

hanzhi713:
cc @alanwaketan

alanwaketan:
cc @yeounoh

jonb377:
@hanzhi713 MpDeviceLoader currently only accepts a single ShardingSpec, which will only be applied to compatible tensors (i.e. tensors whose rank matches the partition spec). In your case, `masks` will be sharded and `input_ids` will not.

We are developing a general sharding-aware dataloader to replace the MpDeviceLoader approach for more advanced use cases, but for now we can generalize this to support multiple ShardingSpecs for different inputs. I've opened #5789 with the patch.

With that change, you can provide multiple ShardingSpec to the MpDeviceLoder `input_sharding`, which would be:

```python
specs = [xs.ShardingSpec(mesh, spec) for spec in [('data', None), ('data', None, None)]
data_loader = pl.MpDeviceLoader(
    data_loader,
    dev,
    input_sharding=specs)
```

PR #5789, opened 2023-11-10
Support list of ShardingSpec in MpDeviceLoader

jonb377:
In https://github.com/pytorch/xla/issues/5768, there is a use case to shard different-rank input tensors. Currently, the MpDeviceLoader sharding only accepts a single spec, which is applied to compatible tensors. With this change, multiple specs can be provided in construction, and the first compatible spec will be applied to each tensor.


torch_xla/core/xla_model.py
```diff
@@ -1012,7 +1012,15 @@ def convert_fn(tensors):
...
     shardings = None
     if input_sharding:
-      shardings = [input_sharding.xla_spec(t) for t in tensors]
+      if isinstance(input_sharding, list):
+        shardings = [None] * len(tensors)
+        for i, tensor in enumerate(tensors):
+          for sharding in input_sharding:
+            if sharding.can_apply(tensor):
+              shardings[i] = sharding.xla_spec(tensor)
+              break
```
yeounoh, line 1021: Maybe we also add some comments about the first match is applied?

yeounoh (approved):
LGTM

torch_xla/core/xla_model.py
jonb377: Done, thanks Yeounoh!

alanwaketan: Is first match something common practice? Or will it be better:
1. either one sharding spec for everything;
2. or one sharding spec for each input and in order.

That the behavior is well defined. Otherwise, this "compatible check" is totally a black box for the user.

jonb377: It's definitely not ideal... Probably better would be to have the user provide sharding specs in a structure matching that of the inputs and avoid the black-box compatibility check, e.g.:

```python
# Inputs of the form:
{"a": torch.randn(16), "b": torch.randn(16, 16), "c": torch.randn(16, 16, 16)}

# Then sharding_spec input would be:
{"a": ShardingSpec(mesh, (0,)), "b": ShardingSpec(mesh, (0, None)), "c": ShardingSpec(mesh, (0, None, None))}
```

This is a bit of a refactor of the TensorToXlaArena code though, so I took the easy way out. I'll go ahead and take a stab at the cleaner approach now.

alanwaketan: No worries. Take your time.

Issue #5764, opened 2023-11-02
Supporting the Core ATen Opset


wonjoolee95:
## 🚀 Feature
As PyTorch/XLA tries to support the PyTorch core ATen opset in its entirety, it requires lowering each core ATen op in PyTorch/XLA. As of today, most of these ops are already lowered. This issue will serve as a tracker for our support of the core ATen opset.

## Unlowered core ATen ops

- [ ] AtenAcos_AutoI32
- [ ] AtenAcosh_AutoI32
- [ ] AtenArangeStartStep
- [ ] AtenAsStrided
- [ ] AtenAsStrided_AutoI32
- [ ] AtenAsin_AutoI32
- [ ] AtenAsinh_AutoI32
- [ ] AtenAtan2_AutoI32
- [ ] AtenAtan_AutoI32
- [ ] AtenAvgPool2dBackward
- [ ] AtenAvgPool2dBackward_AutoI32
- [ ] AtenCdistForward_AutoI32
- [ ] AtenCol2im
- [ ] AtenConvolutionBackward
- [ ] AtenCos_AutoI32
- [ ] AtenCosh_AutoI32
- [ ] AtenEmbeddingDenseBackward
- [ ] AtenEmptyStrided
- [ ] AtenErf_AutoI32
- [ ] AtenExp_AutoI32
- [ ] AtenExpm1_AutoI32
- [ ] AtenFull
- [ ] AtenGelu_AutoI32
- [ ] AtenGridSampler2d
- [ ] AtenGridSampler2d_AutoI32
- [ ] AtenIsinf_AutoI32
- [ ] AtenIsnan_AutoI32
- [ ] AtenMaxPool2dWithIndicesBackward
- [ ] AtenMeanDim_AutoI32
- [ ] AtenMean_AutoI32
- [ ] AtenNativeBatchNormLegitNoStats
- [ ] AtenNativeBatchNormLegitNoStats_AutoI32
- [ ] AtenNativeBatchNormLegit
- [ ] AtenNativeBatchNormLegit_AutoI32
- [ ] AtenNativeGroupNormBackward
- [ ] AtenNativeGroupNorm_AutoI32
- [ ] AtenNativeLayerNormBackward
- [ ] AtenNativeLayerNorm_AutoI32
- [ ] AtenPdistForward
- [ ] AtenPdistForward_AutoI32
- [ ] AtenPixelShuffle
- [ ] AtenPixelShuffle_AutoI32
[...]

PR #5790, opened 2023-11-11
Enable passing down dynamic dimensions from torch to XLA

lsy323:
This is based on #5778. This PR focuses on supporting unbounded dynamism in LTC infra. The e2e would work when the unbounded dynamism support on XLA side is merged, and required op lowering logic is updated in torch_xla.

- Add support in LTC to pass down the unbounded dynamism info down to HLO lowering.
- Guard op lowering logic change for unbounded dynamism support by `EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM=1`
- Fixing a bug in setup.py

Test:
Tested with @sdasgup3's XLA change to propagate unbounded dynamism [branch](https://github.com/openxla/xla/compare/main...sdasgup3:openxla-xla:cd2cf5c34). To make the patch works with torch_xla HEAD, the branch needs to be rebased to the pinned XLA version. [Rebased branch](https://github.com/lsy323/openxla-xla/tree/lsiyuan/sandeep-dynamism-rebased)

The unbounded dynamism can be propagated in the following example:
```
import torch
import torch_xla

device = xm.xla_device()

a = torch.tensor([[1,2],[2,4]], device=device)
torch_xla._XLAC._xla_mark_dynamic(a, 0)
b = torch.tensor([[1,2],[2,4]], device=device)
torch_xla._XLAC._xla_mark_dynamic(b, 0)
c = a * b
hlo_content = torch_xla._XLAC._get_xla_tensors_hlo([c])
print(hlo_content)
```
which gives
```
loModule IrToHlo.5, entry_computation_layout={(s64[?,2]{1,0}, s64[?,2]{1,0})->(s64[?,2]{1,0})}

ENTRY %IrToHlo.5 (p0.1: s64[?,2], p1.2: s64[?,2]) -> (s64[?,2]) {
  %p1.2 = s64[?,2]{1,0} parameter(1)
  %p0.1 = s64[?,2]{1,0} parameter(0)
  %multiply.3 = s64[?,2]{1,0} multiply(s64[?,2]{1,0} %p1.2, s64[?,2]{1,0} %p0.1)
  ROOT %tuple.4 = (s64[?,2]{1,0}) tuple(s64[?,2]{1,0} %multiply.3)
}
```


torch_xla/csrc/elementwise.cpp
```diff
@@ -66,8 +66,16 @@ xla::XlaOp BuildThreshold(xla::XlaOp input, xla::XlaOp output,
 
 xla::XlaOp BuildRelu(xla::XlaOp input) {
   const xla::Shape& input_shape = ShapeHelper::ShapeOfXlaOp(input);
+#ifndef EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM
```
qihqi, line 69: Let's also use env var instead of macros? Macros need to recompile if changed.

qihqi (approved):
Please change the title of PR to something more descriptive: something like "Enable passing down dynamic dimensions from torch to XLA" etc.

Also all C++ use of macro change that to env var like the python one. (so that no need to recompile to enable the feature)

torch_xla/csrc/elementwise.cpp
JackCaoG: +1

torch_xla/csrc/elementwise.cpp
```diff
@@ -66,8 +66,16 @@ xla::XlaOp BuildThreshold(xla::XlaOp input, xla::XlaOp output,
...
   return xla::Max(input, XlaHelpers::ScalarValue<float>(
                              0, input_shape.element_type(), input.builder()));
+#else
+  xla::XlaOp scalar = XlaHelpers::ScalarValue<float>(
+      0, input_shape.element_type(), input.builder());
+  auto promoted = XlaHelpers::Promote(input, scalar);
+
+  return xla::Max(promoted.first, promoted.second);
```
JackCaoG, line 77: can we add a comment to explain about this change?

torch_xla/csrc/helpers.cpp
```diff
@@ -322,6 +329,116 @@ xla::XlaOp XlaHelpers::DynamicReshapeAs(xla::XlaOp input,
              : xla::Reshape(input, shape.dimensions());
 }
 
+#if EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM
```
JackCaoG, line 332: not a fan of macro, if you want to guard these functions you can add a check like `XLA_CHECK(getenv(EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM))`

torch_xla/csrc/helpers.cpp
```diff
@@ -322,6 +329,116 @@ xla::XlaOp XlaHelpers::DynamicReshapeAs(xla::XlaOp input,
...
+xla::XlaOp XlaHelpers::DynamicUnboundedBroadcast(
+    xla::XlaOp input, xla::XlaOp aux_input,
+    absl::Span<const int64_t> aux_input_dimensions) {
+  const xla::Shape& input_shape = ShapeHelper::ShapeOfXlaOp(input);
+  const xla::Shape& aux_input_shape = ShapeHelper::ShapeOfXlaOp(aux_input);
+  bool all_static = true;
+  std::vector<int64_t> output_dimensions;
+  std::vector<bool> output_dynamic;
+  for (auto dim : aux_input_dimensions) {
+    if (aux_input_shape.dimensions(dim) == kUnboundedSize) all_static = false;
```
JackCaoG, line 390: nit, can we do 
```
if (aux_input_shape.dimensions(dim) == kUnboundedSize) {
  all_static = false;
}
```
we don't inline the `if` in this repo

torch_xla/csrc/helpers.cpp
```diff
@@ -578,6 +700,137 @@ std::pair<xla::XlaOp, xla::XlaOp> XlaHelpers::PromoteSecond(xla::XlaOp op1,
...
+    std::vector<int64_t> broadcast_sizes(shape_dims.begin(),
+                                         shape_dims.begin() + size_delta);
+    for (int i = 0; i < size_delta; i++) {
+      if (broadcast_sizes[i] != kUnboundedSize) {
+        get_dim_ops.push_back(
+            XlaHelpers::ScalarValue<int32_t>(broadcast_sizes[i], op.builder()));
+
+        auto s = ShapeHelper::ShapeOfXlaOp(get_dim_ops.back());
+        std::cout << "implicitB shape: " << xla::ShapeUtil::HumanString(s)
+                  << " for size: " << broadcast_sizes[i] << "\n";
```
JackCaoG, line 734: we should use TF_VLOG instead of cout

torch_xla/csrc/helpers.cpp
```diff
@@ -578,6 +700,137 @@ std::pair<xla::XlaOp, xla::XlaOp> XlaHelpers::PromoteSecond(xla::XlaOp op1,
...
+      } else {
+        get_dim_ops.push_back(xla::GetDimensionSize(aux_op, i));
+
+        auto s = ShapeHelper::ShapeOfXlaOp(get_dim_ops.back());
+        std::cout << "implicitB shape: " << xla::ShapeUtil::HumanString(s)
```
JackCaoG, line 739: ditto

torch_xla/csrc/helpers.cpp
```diff
@@ -578,6 +700,137 @@ std::pair<xla::XlaOp, xla::XlaOp> XlaHelpers::PromoteSecond(xla::XlaOp op1,
...
+                  << " for size: ? of index: " << op_shape_index << "\n";
+      }
+    } else {
+      get_dim_ops.push_back(xla::GetDimensionSize(op, op_shape_index));
+
+      auto s = ShapeHelper::ShapeOfXlaOp(get_dim_ops.back());
+      std::cout << "implicitB shape: " << xla::ShapeUtil::HumanString(s)
+                << " for size: ? of index: " << op_shape_index << "\n";
```
JackCaoG, line 816: all of these seems like debug string added instead of meaningful VLOG, I am not sure if we want to keep these.. if we do than give it like `VLOG(7)` so it won't be too noisy

torch_xla/csrc/ir.h
```diff
@@ -138,6 +138,12 @@ class XlaNode : public torch::lazy::Node {
 
   std::string ToString() const override;
 
+  void MarkDynamicDimension(uint32_t dim) { dynamic_dims_.push_back(dim); }
```
JackCaoG, line 141: what if same dim is being called multiple times? Shouldn';t it be a `unordered_set` or boolean array?

torch_xla/csrc/lowering_context.cpp
```diff
@@ -93,16 +93,23 @@ LoweringContext::LoweringContext(
   }
 }
 
+// import xla::Shape.h to inlcude the following defintion.
+static constexpr int64_t kUnboundedSize = std::numeric_limits<int64_t>::min();
 xla::XlaOp LoweringContext::GetParameter(
-    const std::shared_ptr<torch::lazy::BackendData>& data) {
+    const std::shared_ptr<torch::lazy::BackendData>& data,
+    const std::vector<uint32_t>& dynamic_dims) {
```
JackCaoG, line 100: I am abit confuse for this call, so while we are trying to get the parameter version of the data, we also overwrite the `dynamic_dims` field? I guess the right thing to do is to let `data` to carry that information but we don't fully support that at the moment. If that's the case, maybe rename the second parameter to `overwrite_dynamic_dims`. it also should be a `unordered_set` instead of vector....

JackCaoG: ok, maybe `overwrite_unbounded_dynamic_dims` is a more accurate name..

torch_xla/csrc/lowering_context.cpp
```diff
@@ -170,6 +177,20 @@ XlaOpVector LoweringContext::LowerNode(const torch::lazy::Node* node) {
...
+    if (mutable_dynamic->empty()) {
+      for (int i = 0; i < inst->dimensions_size(); i++) {
+        mutable_dynamic->Add(false);
+      }
+    }
+    auto* mutable_dims = inst->mutable_shape()->mutable_dimensions();
+    for (const auto dim : casted->dynamic_dims()) {
+      mutable_dynamic->Set(dim, true);
+      mutable_dims->Set(dim, kUnboundedSize);
+    }
```
JackCaoG, line 193: do we always need to run these or we should guard this in a conditional?

torch_xla/csrc/elementwise.cpp
lsy323: Sounds good, will update to use the env var. The intension was to enable this experimental feature at compile time, so it won't affect the default. lowering workflow

torch_xla/csrc/elementwise.cpp
lsy323: Sure, the xla::Max doesn't do the implicit broadcasting for unbounded dynamism, so we need to promote in torch_xla, will leave a comment there.

torch_xla/csrc/helpers.cpp
lsy323: sg, will update

torch_xla/csrc/helpers.cpp
lsy323: will update, thanks

torch_xla/csrc/helpers.cpp
lsy323: Let me remove all of these debugging prints

torch_xla/csrc/ir.h
lsy323: Good point, let me change to `unordered_set`

torch_xla/csrc/lowering_context.cpp
lsy323: I think the logic is the `dynamic_dims` only takes effect when the parameter is created. Once it's created, the dynamic_dims cannot be changed anymore (We should assert on that).

torch_xla/csrc/helpers.cpp
```diff
@@ -21,6 +21,12 @@
 namespace torch_xla {
 namespace {
 
+static const bool experimental_unbounded_dynamism =
```
qihqi, line 24: Make this a function that returns the values of the flag, and call this function on places this variable is read.

Reason being, if it's a constant then it cannot be changed at runtime, and has to be set before the program starts.

torch_xla/csrc/elementwise.cpp
```diff
@@ -66,8 +70,16 @@ xla::XlaOp BuildThreshold(xla::XlaOp input, xla::XlaOp output,
 
 xla::XlaOp BuildRelu(xla::XlaOp input) {
   const xla::Shape& input_shape = ShapeHelper::ShapeOfXlaOp(input);
-  return xla::Max(input, XlaHelpers::ScalarValue<float>(
-                             0, input_shape.element_type(), input.builder()));
+  xla::XlaOp scalar = XlaHelpers::ScalarValue<float>(
+      0, input_shape.element_type(), input.builder());
+  if (experimental_unbounded_dynamism) {
```
miladm, line 75: Can this path execute in the non export path?
Wdyt we limit this experimental condition to torch.export path?

torch_xla/csrc/helpers.cpp
```diff
@@ -322,6 +331,57 @@ xla::XlaOp XlaHelpers::DynamicReshapeAs(xla::XlaOp input,
...
 
+bool XlaHelpers::IsUnboundedDynamic(const xla::Shape& shape) {
+  XLA_CHECK(experimental_unbounded_dynamism)
+      << "EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM needs to be turned on.";
+  const absl::Span<const int64_t> dims = shape.dimensions();
+  return std::any_of(dims.begin(), dims.end(),
+                     [](int64_t size) { return size == kUnboundedSize; });
+}
+
+xla::XlaOp XlaHelpers::DynamicUnboundedReshape(
```
miladm, line 342: Can you please prepare an opset issue to track supporting this mode of dynamism?

Similar work example: https://github.com/pytorch/xla/issues/5764

torch_xla/csrc/reduction.cpp
```diff
@@ -31,6 +31,9 @@ struct SummationResult {
   xla::XlaOp result;
 };
 
+static const bool experimental_unbounded_dynamism =
+    runtime::sys_util::GetEnvBool("EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM", false);
```
miladm, line 35: env variables causes poor user experience. What's the plan to clean up a better solution (potentially via the upstream torch API level)?

cc @JackCaoG @qihqi @lsy323

miladm: Also, for adding env variables, please keep this file up to date.

https://github.com/pytorch/xla/blob/master/configuration.yaml

miladm (changes requested):
Thanks @lsy323 - I left a few comments.

torch_xla/csrc/reduction.cpp
qihqi: Current thought is to use env variable to limit the code path to "Export Only", as you've pointed out in many places. If there are better mechanisms to accomplish that we can also use something different.

torch_xla/csrc/helpers.cpp
lsy323: updated

torch_xla/csrc/elementwise.cpp
lsy323: Per the discussion below, seems there is no better solution than using a env variable to enable the unbounded dynamism lowering. But we can put more thought on this one.

torch_xla/csrc/helpers.cpp
```diff
@@ -322,6 +327,57 @@ xla::XlaOp XlaHelpers::DynamicReshapeAs(xla::XlaOp input,
...
+    xla::XlaOp input, xla::XlaOp aux_input,
+    absl::Span<const int64_t> output_sizes) {
+  XLA_CHECK(XlaHelpers::IsUnboundedDynamismEnabled())
+      << "EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM needs to be turned on.";
```
JackCaoG, line 342: nit, can we reword it to "set EXPERIMENTAL_XLA_UNBOUNDED_DYNAMISM=1 to run any unbounded dynamism workload". turn on can be a bit confusing if people doesn't realize it is an env var

torch_xla/csrc/ir.cpp
```diff
@@ -174,6 +174,10 @@ xla::Shape XlaNode::GetOpShape(
 std::string XlaNode::ToString() const {
   std::stringstream ss;
   ss << torch::lazy::Node::ToString() << ", xla_shape=" << xla_shape_;
+  ss << ", dynamic_dims: ";
+  for (const auto dim : dynamic_dims_) {
+    ss << dim;
```
JackCaoG, line 179: do you need space or `,` after dim?

torch_xla/csrc/ir.h
```diff
@@ -138,6 +138,15 @@ class XlaNode : public torch::lazy::Node {
...
   std::string ToString() const override;
 
+  void MarkDynamicDimension(uint32_t dim) { dynamic_dims_.insert(dim); }
+
+  const std::unordered_set<uint32_t>& dynamic_dims() const {
+    return dynamic_dims_;
+  }
+
+ protected:
+  std::unordered_set<uint32_t> dynamic_dims_;
```
JackCaoG, line 148: nit `unbounded_dynamic_dims_ ` so we don't confuse with bounded dynamisim work that @vanbasten23 has been working on.

torch_xla/csrc/elementwise.cpp
lsy323: Comment added

torch_xla/stablehlo.py
```diff
@@ -212,6 +212,16 @@ def call_function(self, target, args: Tuple, kwargs: Dict) -> Any:
       new_kwargs['device'] = self._device
     return super().call_function(target, args, new_kwargs)
 
+  def run_node(self, n) -> Any:
```
vanbasten23, line 215: Could you add a test?

torch_xla/csrc/helpers.cpp
lsy323: Sure, I think we have a list internally

JackCaoG (approved):
mostly lgtm, minor nits

torch_xla/csrc/ir.cpp
lsy323: Thanks, let me better format it

torch_xla/csrc/helpers.cpp
lsy323: That makes sense, thanks for catching it.

torch_xla/csrc/ir.h
lsy323: Yes, it's indeed confusing, let me rename

torch_xla/stablehlo.py
lsy323: I added a test, even though it requires unmerged XLA change to work.

torch_xla/csrc/elementwise.cpp
miladm: SG. This issue doesn't block this PR. Let's prepare a proposal in parallel.

miladm (approved):
LGTM. Thanks.

PR #5793, opened 2023-11-13
[PJRT] Use TPU profiler plugin

will-cromar:
The [`tpu_tracer` backend](https://github.com/openxla/xla/blob/34b419269fa7477155086730670e2aa740d80a0d/xla/backends/profiler/tpu/tpu_tracer.cc#L177) uses a static initializer that requires us to set `TPU_LIBRARY_PATH`, which required us to roll back #5731. Instead, use the new PJRT C API profiler extension API to register the TPU profiler.

- Add new API `profiler::RegisterProfilerForPlugin` that we can reuse for other plugins as needed. Mostly copied from JAX's pybind to do the same thing: https://github.com/openxla/xla/blob/34b419269fa7477155086730670e2aa740d80a0d/xla/python/profiler.cc#L90-L105
- If we register both the profielr plugin and the default `tpu_tracer` starting a profiler sessions triggers a crash. Since the issue is (again) a static initializer, define our own `profiler_backends` target that excludes `tpu_tracer` from the build entirely.
  - Piggyback fix: include the GPU device tracer again after https://github.com/openxla/xla/commit/d73a491ea57ea1a57a0a03d3276deff9e367f686. We can remove this after we move GPU support into a plugin.

Confirmed that TPU device traces still show up after this change.

### DIFF
--- a/torch_xla/csrc/runtime/BUILD
+++ b/torch_xla/csrc/runtime/BUILD
@@ -83,6 +83,7 @@ cc_library(
     ":debug_macros",
     ":env_vars",
     ":multi_wait",
+    ":profiler",
     ":stablehlo_helper",
     ":tensor_source",
     ":tf_logging",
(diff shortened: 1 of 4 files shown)


JackCaoG (approved):

alanwaketan (approved):
LGTM.

PR #5794, opened 2023-11-13
Don't set $TPU_LIBRARY_PATH during import

will-cromar:
See #5698 and #5731. Once we remove `tpu_tracer` (#5793), we no longer need to set `TPU_LIBRARY_PATH`.


JackCaoG (approved):
does the existing profiler test enough to cover the testing? or it has to be you manually patch and test?

will-cromar:
Our existing profiler test does not check for device traces, so I had to test manually.

alanwaketan (approved):
LGTM.

PR #5795, opened 2023-11-14
Revert "[HotFix] disable test_set temporarily"

yeounoh:
Reverts pytorch/xla#5792, as it seems to be passing again on the nightly. Verifying in the CI.


JackCaoG:
I think we should first update upstream xla pin to the pr that disabled the test. Let them merge the offending change. And then we reenable the test with new counter update.

PR #5796, opened 2023-11-14
Add a unittest testing each core ATen ops.

qihqi:


wonjoolee95:
This will be helpful to track the ops, thanks Han. FYI we need to add this .py test file to https://github.com/pytorch/xla/blob/master/test/run_tests.sh#L196 for it to be ran by the CI.

qihqi:
> This will be helpful to track the ops, thanks Han. FYI we need to add this .py test file to https://github.com/pytorch/xla/blob/master/test/run_tests.sh#L196 for it to be ran by the CI.

Thanks! done.

test/test_core_aten_ops.py
```diff
@@ -0,0 +1,6883 @@
...
+import unittest
+
+
+class AtenOpTest(unittest.TestCase):
+
+  def test_aten_abs_0(self):
+    args = (torch.randn((10, 10)).to(torch.float32),)
+    kwargs = dict()
+    res = torch.ops.aten.abs(*args, **kwargs)
+    exported = torch.export.export(torch.ops.aten.abs, args, kwargs)
```
JackCaoG, line 17: question.. what happened when we try to export the op that we do not have lowering for?

qihqi: exported_program_to_stablehlo throws exception.

JackCaoG: ok so I guess this test only includes ops that we have lowering? The purpose is to make sure there is no regression?

wonjoolee95: IIRC, this test file includes all core aten ops (even unlowered ops in ptxla). Those that are unlowered are skipped with `@unittest.skip`.

wonjoolee95: But looking at it again, I had a qq, @qihqi -- seems like many op tests here are skipped even for already lowered ops. For example, I see that `test_aten_any_0` is skipped, but we should actually have `any` lowered already -- https://github.com/pytorch/xla/blob/master/codegen/xla_native_functions.yaml#L13. Is there another reason why we're skipping certain op tests?

wonjoolee95: Oh I see, some tests might have problems with the test code itself. Looks like this is happening on the CI, probably just need to skip those as well.

qihqi: Yes we mark skip for now the ones that dont have lowrings (or gives incorrect result). We can they file issues for those and instructions to fix

wonjoolee95 (approved):

JackCaoG (approved):

PR #5800, opened 2023-11-15
Add a persistent cache

jonb377:
This changes adds a new PersistentCache instance which will read and write to disk to enable persistent compilation caching. The PersistentCache is currently not used anywhere, but in a later change, the PersistentCache will replace the Cache in the XlaGraphExecutor when persistent caching is enabled.

### DIFF
--- a/torch_xla/csrc/runtime/cache.h
+++ b/torch_xla/csrc/runtime/cache.h
@@ -133,8 +133,8 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
 
   explicit PersistentCache(int kMaxMemoryCacheSize, std::string cache_dir,
                            bool readonly,
-                           std::function<std::string(TypePtr&)> serialize,
-                           std::function<TypePtr(std::string&)> deserialize)
+                           std::function<std::string(const TypePtr&)> serialize,
+                           std::function<TypePtr(const std::string&)> deserialize)
       : memory_cache_(kMaxMemoryCacheSize),
         cache_dir_(cache_dir),
         readonly_(readonly),
@@ -211,7 +211,7 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
   std::string GetPath(K key) {
     std::stringstream ss;
     ss << key;
-    return cache_dir_ / std::filesystem::path(ss.str());
+    return cache_dir_ / ss.str();
   }
 
   bool Exists(std::string path) {
@@ -225,8 +225,8 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
   }
 
   Cache<K, T, H, E> memory_cache_;
-  std::function<std::string(TypePtr&)> serialize_;
-  std::function<TypePtr(std::string&)> deserialize_;
+  std::function<std::string(const TypePtr&)> serialize_;
+  std::function<TypePtr(const std::string&)> deserialize_;
   std::filesystem::path cache_dir_;
   std::mutex lock_;
   bool readonly_;
--- a/torch_xla/csrc/runtime/cache.h
+++ b/torch_xla/csrc/runtime/cache.h
@@ -131,13 +131,13 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
  public:
   using TypePtr = std::shared_ptr<T>;
 
-  explicit PersistentCache(int kMaxMemoryCacheSize, std::string cache_dir,
-                           bool readonly,
-                           std::function<std::string(const TypePtr&)> serialize,
-                           std::function<TypePtr(const std::string&)> deserialize)
+  explicit PersistentCache(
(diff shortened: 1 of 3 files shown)


torch_xla/csrc/runtime/cache.h
```diff
@@ -106,6 +122,107 @@ class Cache {
   ElementMap element_map_;
 };
 
+// A persistent cache which serializes values to disk. This wraps a Cache
+// instance, so values will only be read from disk once and subsequent reads
+// will go through the wrapped Cache.
+template <typename K, typename T, typename H = std::hash<K>,
+          typename E = std::equal_to<K>>
+class PersistentCache : public AbstractCache<K, T, H, E> {
```
alanwaketan, line 130: Can we specialize the PersistentCache a bit more? Do we really need to carry that many template parameters?

jonb377: I don't see a use case for the H and E templates, so those can be dropped and the default value directly applied the parent class. Keeping the same interface as Cache lets this be a drop-in replacement though, maybe we can follow up to remove the unnecessary templates from both?

alanwaketan: Let me take a closer look (probably after my break...). I need to play around with the patch to awake my C++ muscles...

will-cromar (approved):
Thanks!

torch_xla/csrc/runtime/cache.h
```diff
@@ -106,6 +122,116 @@ class Cache {
...
+ public:
+  using TypePtr = std::shared_ptr<T>;
+
+  explicit PersistentCache(int kMaxMemoryCacheSize, std::string cache_dir,
+                           bool readonly,
+                           std::function<std::string(TypePtr&)> serialize,
```
yeounoh, line 136: nit. do you expect the serializer to change the original data? otherwise, use `const TypePtr&`

torch_xla/csrc/runtime/cache.h
```diff
@@ -106,6 +122,116 @@ class Cache {
...
+    TORCH_LAZY_COUNTER("PersistentCacheHit", 1);
+    // Make sure the memory_cache_ tracks the value to prevent multiple loads
+    return memory_cache_.Add(key, val);
+  }
+
+  void Clear() override {
+    std::lock_guard<std::mutex> slock(lock_);
+    memory_cache_.Clear();
+    // Delete and recreate the cache directory on disk.
+    if (!readonly_) {
```
yeounoh, line 197: Maybe instead of conditioning on `readonly_` here, introduce an optional arg for `Clear` to clear the persistent cache, if `cache_dir` is provided.

yeounoh (approved):
LGTM, thanks @jonb377 left minor comments

torch_xla/csrc/runtime/cache.h
jonb377: I think we should respect the `readonly_` setting here - the reason for it is in shared mounts, we only want the writer to modify the objects on disk.

torch_xla/csrc/runtime/cache.h
```diff
@@ -106,6 +122,116 @@ class Cache {
...
+    ss << key;
+    return cache_dir_ / std::filesystem::path(ss.str());
+  }
+
+  bool Exists(std::string path) {
+    struct stat buffer;
+    return stat(path.c_str(), &buffer) == 0;
+  }
+
+  bool _erase(const K& key) {
```
alanwaketan, line 222: The naming of this function seems a bit weird? May `EraseImpl`?

torch_xla/csrc/runtime/cache.h
```diff
@@ -106,6 +122,116 @@ class Cache {
...
+    memory_cache_.Erase(key);
+    return !readonly_ && std::filesystem::remove(GetPath(key));
+  }
+
+  Cache<K, T, H, E> memory_cache_;
+  std::function<std::string(TypePtr&)> serialize_;
+  std::function<TypePtr(std::string&)> deserialize_;
+  std::filesystem::path cache_dir_;
+  std::mutex lock_;
+  bool readonly_;
```
alanwaketan, line 232: Can we make `readonly_` to be constant?

alanwaketan: It seems readonly_ means the persistent cache is readonly but the memory cache is rw... And it confuses me as I thought the whole cache is readonly. Can we have more documentation about the flag?

alanwaketan (approved):
A few nit, but LGTM in general.

torch_xla/csrc/runtime/cache.h
jonb377: I see, that's a good caveat - readonly refers only to the persistence layer. I'll add some documentation around this.

PR #5803, opened 2023-11-15
Incorporate compilation environment in hash

jonb377:
Include the compilation environment in the graph hash. This will ensure that options which impact the compilation result will change the hash, so the same traced IR graph will result in a different hash when the compilation environment is different.

Currently, the following env vars must be hashed: `XLA_FLAGS`, `LIBTPU_INIT_ARGS`, and `TPU_MEGACORE`. All env vars which impact the hash must be included in `env_hash.cc`, otherwise there can be spurious cache hits!

This PR builds on #5800, but it doesn't depend on it and can be landed separately.

### DIFF
--- a/torch_xla/csrc/runtime/env_hash.cc
+++ b/torch_xla/csrc/runtime/env_hash.cc
@@ -1,4 +1,4 @@
-#include "torch_xla/csrc/runtime/hash.h"
+#include "torch_xla/csrc/runtime/env_hash.h"
 
 #include <iostream>
 #include <sstream>
@@ -45,7 +45,6 @@ torch::lazy::hash_t hash_xla_flags(std::string env_var_name) {
   torch::lazy::hash_t hash = 0;
   // Parse the space-delimited flags once at a time.
   while (std::getline(xla_flag_env, current_flag, ' ')) {
-    // Only consider flags which 
     if (current_flag.rfind(XLA_FLAG_PREFIX, 0) != 0) {
       continue;
     }
@@ -71,7 +70,8 @@ torch::lazy::hash_t hash_xla_flags(std::string env_var_name) {
   return hash;
 }
 
-torch::lazy::hash_t hash_xla_env_vars(std::vector<std::string> flag_vars, std::vector<std::string> raw_vars) {
+torch::lazy::hash_t hash_xla_env_vars(std::vector<std::string> flag_vars,
+                                      std::vector<std::string> raw_vars) {
   torch::lazy::hash_t hash;
   // Parse the flag_vars for XLA flags.
   for (auto& env_var_name : flag_vars) {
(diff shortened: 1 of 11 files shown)


torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -533,6 +534,12 @@ XLAGraphExecutor::SyncTensorCollection XLAGraphExecutor::CollectSyncTensors(
   // The force_ltc_data controls aliasing compilation, so effectively the same
   // graph with on/off force_ltc_data should not match, hash wise.
   coll.hash = torch::lazy::MHash(config.force_ltc_data);
+  // Ensure the compilation environment and git revision are reflected in the
+  // hash.
+  coll.hash = torch::lazy::HashCombine(
+      coll.hash, runtime::GetComputationClient()->HashCompilationEnv());
+  coll.hash =
+      torch::lazy::HashCombine(coll.hash, torch::lazy::StringHash(XLA_GITREV));
```
jonb377, line 542: cc @yeounoh, I wonder if we need to include whether or not we're executing in SPMD mode here. I found that the hash is different for SPMD vs non-SPMD, but I'm not sure if that will be the case generally.

torch_xla/csrc/runtime/env_hash.cc
```diff
@@ -0,0 +1,105 @@
...
+namespace torch_xla {
+namespace runtime {
+namespace hash {
+
+namespace {
+static const std::string XLA_FLAG_PREFIX = "--xla";
+
+// Taken from JAX:
+// https://github.com/google/jax/blob/8ee5811/jax/_src/cache_key.py#L325-L346
+static const std::set<std::string> FlagsToExclude = {
```
JackCaoG, line 17: not that it really matters, but is there a reason to use `set` instead of `unordered_set`?

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: Let's add it. We don't really want SPMD and nonSPMD hash to collide .

JackCaoG (commented):
mostly lgtm, minor nits

torch_xla/csrc/runtime/env_hash.cc
jonb377: No reason, I'll switch to `unordered_set`

torch_xla/csrc/xla_graph_executor.cpp
jonb377: Done

torch_xla/csrc/runtime/env_hash.h
```diff
@@ -0,0 +1,15 @@
...
+
+namespace torch_xla {
+namespace runtime {
+namespace hash {
+
+// Take a hash of XLA flags which impact the compilation result.
+// TODO(jonbolin): We should move away from manually hashing the env vars and
+// instead hash the compilation environment directly when the functionality is
+// supported in the runtime.
+torch::lazy::hash_t HashXlaEnvVars(bool force_rehash = false);
```
yeounoh, line 11: What's `force_reshash`?

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: FWIW It should already be dependent to SPMD via sharding hash.

torch_xla/csrc/runtime/env_hash.h
jonb377: This is for testing - the hash is statically stored, but in the tests I need to be able to change the environment and rehash.

I suppose I'm actually storing the result of the environment hash within the computation client, so this is still only called once. I'll make the hash results here non-static.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -559,6 +595,13 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
   return computations;
 }
 
+torch::lazy::hash_t PjRtComputationClient::HashCompilationEnv() {
+  // TODO(jonbolin): Incorporate CompileOptions into the hash. These are
```
yeounoh, line 599: Once we do, we really don't need the SPMD mode be part of the hash (virtual device on/off).

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: https://github.com/pytorch/xla/pull/5803/files#r1414428856

yeounoh (approved):
LGTM thanks!

PR #5804, opened 2023-11-15
Enable persistent compilation caching

jonb377:
This change enables persistent caching by combining #5800 and #5803. It uses the serialization functionality in PjRtLoadedExecutable to convert the executables to/from strings, which are written to disk by the persistent cache.

The persistent cache is enabled by setting the environment variable `XLA_PERSISTENT_CACHE_PATH` to the desired compilation cache path. An additional environment variable `XLA_PERSISTENT_CACHE_READ_ONLY` can be used to control whether the cache is readonly, which can be useful when the cache is shared across workers in an SPMD setting.

Note that the persistent cache does not perform any eviction, so it is currently up to the user to clean the cache.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -210,6 +210,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/test_input_output_aliases.py"
   run_test "$CDIR/test_torch_distributed_xla_backend.py"
   run_torchrun "$CDIR/pjrt/test_torchrun.py"
+  run_test "$CDIR/test_persistent_cache.py"
   # NOTE: this line below is testing export and don't care about GPU
   PJRT_DEVICE=CPU CPU_NUM_DEVICES=1 run_coverage "$CDIR/test_core_aten_ops.py"
 }
--- a/test/test_persistent_cache.py
+++ b/test/test_persistent_cache.py
@@ -1,0 +1,117 @@
+import functools
+import os
+from subprocess import run, STDOUT, PIPE
+import sys
+import tempfile
+import unittest
+
+import torch_xla.runtime as xr
+
+
+# Wrapper to manage a temporary directory for the wrapped test
+def run_with_tmpdir(f):
+
+  @functools.wraps(f)
+  def run(*args, **kwargs):
+    with tempfile.TemporaryDirectory() as tmpdir:
+      kwargs.setdefault('tmpdir', tmpdir)
+      f(*args, **kwargs)
+
+  return run
+
+
+# Basic command to generate a simple graph and perform metrics assertions
+METRICS_CMD_FMT = r'''
+import os
+import torch
+import torch_xla.core.xla_model as xm
+import torch_xla.debug.metrics as met
+import torch_xla.runtime as xr
+
+if 'TEST_WITH_SPMD' in os.environ:
+  xr.use_spmd()
+
+t = torch.randn(4, 4).to(xm.xla_device())
+s = t @ t
+xm.mark_step()
+
+for counter, value in %s:
+  actual = met.counter_value(counter)
+  assert actual == value, \
+    f'Unexpected value for counter {counter}: expected {value}, got {actual}'
+'''
+
+
+@unittest.skipUnless(xr.device_type() in {'TPU', 'GPU'},
+                     'Device type does not support persistent caching')
+class PersistentCacheTest(unittest.TestCase):
+  """
+  Test suite to verify compilation cache across processes. Tests will run
+  multiple Python subprocesses which use the XLA runtime to populate the cache
+  and perform assertions on the metrics generated.
+  """
+
+  def _run_with_metric_assertions(self, env: dict, metric_expectations: dict):
+    cmd = METRICS_CMD_FMT % list(metric_expectations.items())
+    proc = run([sys.executable, '-c', cmd], env=env, stdout=PIPE, stderr=STDOUT)
+    self.assertEqual(proc.returncode, 0,
+                     f'Non-zero exit code, output:\n{proc.stdout.decode()}')
+
+  def _run_tests(self, tmpdir, use_spmd=False):
+    cache_dir = os.path.join(tmpdir, 'cache')
+    env = os.environ.copy()
+    env['XLA_PERSISTENT_CACHE_PATH'] = cache_dir
+    if use_spmd:
+      env['TEST_WITH_SPMD'] = '1'
+
+    # Use subtests to avoid having to prime the cache for each test.
+    with self.subTest('The first attempt should miss on the persistent cache'):
+      self._run_with_metric_assertions(env, {
+          'PersistentCacheMiss': 1,
+          'PersistentCacheHit': None
+      })
+
+    with self.subTest('A second run should hit the cache'):
+      self._run_with_metric_assertions(env, {
+          'PersistentCacheMiss': None,
+          'PersistentCacheHit': 1
+      })
+
+    with self.subTest('Ignored XLA flags should not impact the hash'):
+      env['XLA_FLAGS'] = f'--xla_dump_disable_metadata'
+      self._run_with_metric_assertions(env, {
+          'PersistentCacheMiss': None,
+          'PersistentCacheHit': 1
+      })
+
+    with self.subTest('Non-ignored LIBTPU_INIT_ARGS should impact the hash'):
+      env['LIBTPU_INIT_ARGS'] = '--xla_enable_async_collective_permute=true'
+      self._run_with_metric_assertions(env, {
+          'PersistentCacheMiss': 1,
+          'PersistentCacheHit': None
+      })
+
+    with self.subTest('Corrupt serialization should not be loaded'):
+      for fname in os.listdir(cache_dir):
+        with open(os.path.join(cache_dir, fname), 'wb') as f:
+          f.write(b'')
+      self._run_with_metric_assertions(
+          env, {
+              'PersistentCacheMiss': None,
+              'PersistentCacheHit': None,
+              'PersistentCacheDeserializeFailure': 1
+          })
+
+  @run_with_tmpdir
+  def test_persistent_cache(self, tmpdir):
+    self._run_tests(tmpdir)
+
+  @unittest.skipUnless(xr.device_type() == 'TPU', 'TPU required for SPMD')
+  @run_with_tmpdir
+  def test_persistent_cache_spmd(self, tmpdir):
+    self._run_tests(tmpdir, use_spmd=True)
+
+
+if __name__ == '__main__':
+  test = unittest.main()
+  sys.exit(0 if test.result.wasSuccessful() else 1)
--- a/test/test_persistent_cache.py
+++ b/test/test_persistent_cache.py
@@ -20,25 +20,60 @@ def run_with_tmpdir(f):
   return run
 
 
-# Basic command to generate a simple graph and perform metrics assertions
-METRICS_CMD_FMT = r'''
+# Command to generate a simple graph and perform correctness/metrics assertions.
+# There are four methods supported by configuring the environment for the test:
+#  - Single-device (default): The program runs a computation on a single XLA
+#    device.
+#  - Unsharded SPMD: Setting TEST_WITH_SPMD will run the computation replicated
+#    across all devices.
+#  - Sharded SPMD: Setting TEST_WITH_SPMD and MARK_SHARDING will shard the
+#    computation across all devices.
+#  - Multiprocess: Setting TEST_WITH_MP will run the same computation across all
+#    devices using multiprocessing.
+TEST_FMT = r'''
 import os
 import torch
 import torch_xla.core.xla_model as xm
 import torch_xla.debug.metrics as met
+import torch_xla.distributed.spmd as xs
 import torch_xla.runtime as xr
+ 
+def test_fn(rank=None):
+  if rank is not None:
+    # In a multiprocess setting, rank will be set to the process rank. For MP,
+    # we need to change the cache dir for each process to avoid a race condition
+    # where one process loads the compilation result of another, which would
+    # break the metrics assertion.
+    os.environ['XLA_PERSISTENT_CACHE_PATH'] = \
+      os.path.join(os.environ['XLA_PERSISTENT_CACHE_PATH'], str(rank))
 
-if 'TEST_WITH_SPMD' in os.environ:
-  xr.use_spmd()
+  t = torch.randn(16)
+  expected = t + t
 
-t = torch.randn(4, 4).to(xm.xla_device())
-s = t @ t
-xm.mark_step()
+  xt = t.to(xm.xla_device())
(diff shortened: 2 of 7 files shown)


torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -593,6 +593,41 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
...
+        << "Failed to retrieve HLO modules from deserialized executable";
+    return nullptr;
+  }
+  XLA_CHECK(hlo_modules->size() == 1)
+      << "Only a single module is supported for persistent computation "
+         "caching. Please unset the XLA_PERSISTENT_CACHE_PATH "
+         "variable to disable persistent caching.";
+  xla::XlaComputation computation((*hlo_modules)[0]->ToProto());
+
+  std::vector<std::string> devices = {GetDefaultDevice()};
```
jonb377, line 626: @JackCaoG I'm not sure what this should be for SPMD mode. Since it's in the runtime layer, this will result in `TPU:0` with SPMD, but should it be `SPMD:0`? It doesn't seem to be causing any issues as is.

jonb377:
Ah bummer, it's not running CI since I'm targeting a different branch... I was hoping to see results for GPU. I guess we'll need to wait until the others land so I can target master.

test/test_persistent_cache.py
```diff
@@ -0,0 +1,117 @@
...
+
+for counter, value in %s:
+  actual = met.counter_value(counter)
+  assert actual == value, \
+    f'Unexpected value for counter {counter}: expected {value}, got {actual}'
+'''
+
+
+@unittest.skipUnless(xr.device_type() in {'TPU', 'GPU'},
+                     'Device type does not support persistent caching')
```
JackCaoG, line 46: why not cpu?

torch_xla/csrc/runtime/pjrt_computation_client.cc
JackCaoG: Hmm I guess for SPMD we only load it once in one process? I don't remember exactually what I did but I am wondering if we compile the program normally without cache, would device return in `PjRtComputation` has `TPU:0` or `SPMD:0`?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -76,6 +76,36 @@ bool ShouldSyncIrValue(const torch::lazy::Value& ir_value) {
...
+          computation->computation);
+    };
+    auto deserialize_fn = [](auto& serialization)
+        -> std::shared_ptr<XLAGraphExecutor::CachedComputation> {
+      auto computation =
+          runtime::GetComputationClient()->DeserializeComputation(
+              serialization);
+      if (!computation) return nullptr;
+      bool is_sharded = bridge::GetDefaultDevice()->toString() ==
+                        GetVirtualDevice().toString();
```
JackCaoG, line 98: why do we need to check current default device type?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -76,6 +76,36 @@ bool ShouldSyncIrValue(const torch::lazy::Value& ir_value) {
   return ir_value->op() != xla_not_supported;
 }
 
+XLAGraphExecutor::ComputationCache* CreateComputationCache() {
+  static const size_t kMaxCacheSize =
+      runtime::sys_util::GetEnvInt("XLA_COMPILATION_CACHE_SIZE", 1024);
+  static const bool readonlyPersistentCache =
+      runtime::sys_util::GetEnvBool("XLA_PERSISTENT_CACHE_RO", false);
```
JackCaoG, line 83: I would prefer to just call be `READ_ONLY` instead of `RO`

torch_xla/csrc/xla_graph_executor.h
```diff
@@ -163,8 +163,14 @@ class XLAGraphExecutor : public torch::lazy::LazyGraphExecutor {
   };
 
   using ComputationCache =
+      runtime::util::AbstractCache<torch::lazy::hash_t, CachedComputation,
+                                   torch::lazy::HashReducer>;
+  using MemoryCache =
       runtime::util::Cache<torch::lazy::hash_t, CachedComputation,
```
JackCaoG, line 169: I am confuse about the difference between `ComputationCache ` and `MemoryCache `

test/run_tests.sh
```diff
@@ -208,6 +208,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/test_input_output_aliases.py"
   run_test "$CDIR/test_torch_distributed_xla_backend.py"
   run_torchrun "$CDIR/pjrt/test_torchrun.py"
+  run_test "$CDIR/test_persistent_cache.py"
```
JackCaoG, line 211: Thanks @jonb377! I think we want to test 3 different cases
1. single device
2. multi-device non spmd
3. multi-device spmd
and in later time we want to at least manually test in pod.

test/test_persistent_cache.py
jonb377: I tested it, but CPU isn't supported (deserialization fails). JAX has a similar restriction: https://github.com/google/jax/blob/234be736c4cdd8da4197078278d35a6a1cde3767/tests/compilation_cache_test.py#L69C41-L69C41

torch_xla/csrc/xla_graph_executor.cpp
jonb377: I'm a bit outdated on how to check if we're using SPMD mode in the C++ layer - I think this should be the equivalent of what used to be `ShardingUtil::UseVirtualDevice`, but let me know if I'm mistaken.

torch_xla/csrc/xla_graph_executor.h
jonb377: `ComputationCache` is used for the umbrella return type, and `MemoryCache` and `PersistentCache` are subtypes of `ComputationCache`

test/run_tests.sh
jonb377: Sure! Right now, this covers 1 and 2. I'll add an MP test to the suite.

torch_xla/csrc/xla_graph_executor.h
jonb377: The types were split out in https://github.com/pytorch/xla/pull/5800

torch_xla/csrc/xla_graph_executor.cpp
jonb377: Turns out we still have a utility for this! Thanks @JackCaoG 

https://github.com/pytorch/xla/blob/6e661304e270c6a1a864700e04d5d44235c43f3b/torch_xla/csrc/device.cpp#L101-L105

test/run_tests.sh
jonb377: Just a heads up, the tests are pretty slow since we have to boot up the runtime multiple times (right now it's ~80seconds, with MP it will be 2min).

torch_xla/csrc/runtime/pjrt_computation_client.cc
jonb377: Looks like `SPMD:0`, I'll change this.

test/run_tests.sh
jonb377: Done - the tests now include:
- Single-device
- Replicated SPMD
- Sharded SPMD
- MP

jonb377:
> @jonb377 do you still want to merge this one?

@JackCaoG Yes, just pending reviews. I'll ping some folks.

test/test_persistent_cache.py
```diff
@@ -0,0 +1,187 @@
...
+                     f'Non-zero exit code, output:\n{proc.stdout.decode()}')
+
+  @run_with_tmpdir
+  def test_persistent_cache(self, tmpdir):
+    env = os.environ.copy()
+    env['XLA_PERSISTENT_CACHE_PATH'] = tmpdir
+
+    # Use subtests to avoid having to prime the cache for each test.
+    with self.subTest('The first attempt should miss on the persistent cache'):
+      self._run_with_metric_assertions(env, {
```
will-cromar, line 102: This looks like a perfect use case for parameterized tests https://github.com/abseil/abseil-py/blob/main/absl/testing/parameterized.py

test/test_persistent_cache.py
```diff
@@ -0,0 +1,187 @@
...
+# Command to generate a simple graph and perform correctness/metrics assertions.
+# There are four methods supported by configuring the environment for the test:
+#  - Single-device (default): The program runs a computation on a single XLA
+#    device.
+#  - Unsharded SPMD: Setting TEST_WITH_SPMD will run the computation replicated
+#    across all devices.
+#  - Sharded SPMD: Setting TEST_WITH_SPMD and MARK_SHARDING will shard the
+#    computation across all devices.
+#  - Multiprocess: Setting TEST_WITH_MP will run the same computation across all
+#    devices using multiprocessing.
```
will-cromar, line 32: Why does this test case live in a string? If it has to be a separate file, can it be a file checked into this repository? 

If it's possible, I would actually rather you untangle the different test cases into functions and launch them directly, either with xmp.spawn or normal `multiprocessing`, rather than writing a file to run with subprocess. That may be slightly repetitive, but that's okay for unit tests -- it's better to be very explicit about test assertions than build the best abstraction https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

test/test_persistent_cache.py
```diff
@@ -0,0 +1,187 @@
...
+    # where one process loads the compilation result of another, which would
+    # break the metrics assertion.
+    os.environ['XLA_PERSISTENT_CACHE_PATH'] = \
+      os.path.join(os.environ['XLA_PERSISTENT_CACHE_PATH'], str(rank))
+
+  t = torch.randn(16)
+  expected = t + t
+
+  xt = t.to(xm.xla_device())
+  if {'TEST_WITH_SPMD', 'MARK_SHARDING'} <= os.environ.keys():
```
will-cromar, line 54: TIL `dict` implements the `<=` operator

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -76,6 +76,34 @@ bool ShouldSyncIrValue(const torch::lazy::Value& ir_value) {
...
 
+XLAGraphExecutor::ComputationCache* CreateComputationCache() {
+  static const size_t kMaxCacheSize =
+      runtime::sys_util::GetEnvInt("XLA_COMPILATION_CACHE_SIZE", 1024);
+  static const bool readonlyPersistentCache =
+      runtime::sys_util::GetEnvBool("XLA_PERSISTENT_CACHE_READ_ONLY", false);
+  static std::string persistentCacheDir =
+      runtime::sys_util::GetEnvString("XLA_PERSISTENT_CACHE_PATH", "");
+  if (!persistentCacheDir.empty()) {
+    auto serialize_fn = [](auto& computation) -> std::string {
```
will-cromar, line 87: nit: can you fill in these `auto`s? In particular, I can't tell what type `computation` is supposed to be

test/test_persistent_cache.py
jonb377: Which part should be parametrized?

jonb377: Actually - I'm just going to rework the entire test when splitting it out to a separate file.

yeounoh (approved):
LGTM

test/test_persistent_cache.py
jonb377: It looks quite nice as a parametrized test!

test/test_persistent_cache.py
jonb377: I decided to directly use `xmp.spawn` and run SPMD and single-device tests using a ProcessPoolExecutor, so the code lives directly in the test file now. Much cleaner!

torch_xla/csrc/xla_graph_executor.cpp
jonb377: I learned while filling in the `auto`s that I also didn't know what type `computation` was lol

test/test_persistent_cache.py
```diff
@@ -0,0 +1,116 @@
...
+  os.environ['XLA_PERSISTENT_CACHE_PATH'] = \
+    os.path.join(os.environ['XLA_PERSISTENT_CACHE_PATH'], str(rank))
+
+  t = torch.randn(16)
+  xt = t.to(xm.xla_device())
+  _assert_correctness_and_metrics(t, xt, metrics)
+
+
+def _single_device_test(metrics):
+  t = torch.randn(16)
```
will-cromar, line 59: small suggestion: you can move these test cases into your test class by making them `@staticmethod`s

will-cromar (approved):
LGTM. Thanks!

test/test_persistent_cache.py
jonb377: I tried that initially, but for the ProcessPoolExecutor it complained that they weren't pickleable 🥲

will-cromar: Hmm, are you sure you didn't use `@classmethod`? We have other tests that use `@staticmethod` to create pickleable test cases

jonb377: Just retried the single device test:

```
class PersistentCacheTest(parameterized.TestCase):
  """
  Test suite to verify compilation cache across processes. Tests will run
  multiple Python subprocesses which use the XLA runtime to populate the cache
  and perform assertions on the metrics generated.
  """

  @staticmethod
  def _single_device_test(metrics):
    t = torch.randn(16)
    xt = t.to(xm.xla_device())
    _assert_correctness_and_metrics(t, xt, metrics)
```

It hit this:

```
  File "/home/ptxla/.local/lib/python3.8/site-packages/absl/testing/parameterized.py", line 321, in bound_param_test
    return test_method(self, *testcase_params)
  File "test_persistent_cache.py", line 23, in run
    f(*args, **kwargs)
  File "test_persistent_cache.py", line 103, in test_persistent_cache
    launch_method(test_fn, ({
  File "test_persistent_cache.py", line 31, in _test_spawn
    pool.submit(fn, *args).result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 444, in result
    return self.__get_result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.8/multiprocessing/queues.py", line 239, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/usr/local/lib/python3.8/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: cannot pickle 'staticmethod' object
```

The _mp_test also hit this when using staticmethod... I wonder what's going wrong lol

jonb377:
MP GPU looks good, but single-device seems to be broken by the same issue addressed in #6023 (cc @vanbasten23). I've disabled single-device and SPMD GPU tests.

PR #5808, opened 2023-11-15
Add benchmark selection by tier 1-3

frgossen:


benchmarks/experiment_runner.py
```diff
@@ -440,9 +446,26 @@ def parse_args(args=None):
...
+}
+
+
 def main():
   args = parse_args()
 
+  # Set filter based on tier
+  if args.filter_by_tier is not None:
+    args.filter = []
+    args.exclude = [r"^$"]
```
cota, line 462: I wouldn't overwrite args.filter nor args.exclude.
For example, one might want to run tier1, except BERT_pytorch (for example because it's known to not work). This could be accomplished with --filter-by-tier=1 --exclude=BERT_pytorch. Similarly, we might want to run tier1 plus a handful of benchmarks from other tiers, e.g. --filter-by-tier=1 --filter=hf_Bert. This should still be possible.

So, how about the following:

1. Set the filter and filter-by-tier options to a `[]` default (Note: [] also evaluates to false, just like None, so the code after this still works).
2. Replace this whole hunk with:
```
for tier in args.filter_by_tier:
  if tier not in FILTER_BY_TIER:
    continue
  args.filter.append(FILTER_BY_TIER[tier])
```

frgossen: Nice. Also added exclusion as discussed.

cota: Thanks. Still needs the default [] in the filter and exclude args, otherwise we might append to None, which cannot happen.
I'd also set the filter|exclude-by-tier args to a default [], so that we can directly iterate over them instead of checking for !None.

frgossen: Thanks!

benchmarks/experiment_runner.py
```diff
@@ -440,11 +452,35 @@ def parse_args(args=None):
...
+  3: r"^(doctr_det_predictor|doctr_reco_predictor|drq|functorch_maml_omniglot)$",
+}
+
+
 def main():
   args = parse_args()
 
-  args.filter = args.filter or [r"."]
-  args.exclude = args.exclude or [r"^$"]
+  args.filter = args.filter or []
```
cota, line 465: Sorry, what I meant is to change the default where the flags are defined, e.g.
parser.add_argument("--filter", ..., default=[])
Otherwise they default to None.

benchmarks/experiment_runner.py
```diff
@@ -440,11 +454,32 @@ def parse_args(args=None):
...
+}
+
+
 def main():
   args = parse_args()
 
-  args.filter = args.filter or [r"."]
-  args.exclude = args.exclude or [r"^$"]
+  # Expand filter/exclude by tier.
+  if args.filter_by_tier is not None:
```
cota, line 468: Now this line can be removed, since filter_by_tier is always iterable (it's an empty list by default)

benchmarks/experiment_runner.py
```diff
@@ -440,11 +454,32 @@ def parse_args(args=None):
...
+    for tier in args.filter_by_tier:
+      if tier not in FILTER_BY_TIER:
+        continue
+      args.filter.append(FILTER_BY_TIER[tier])
+  if args.exclude_by_tier is not None:
```
cota, line 473: remove

benchmarks/experiment_runner.py
```diff
@@ -440,11 +454,32 @@ def parse_args(args=None):
...
+    for tier in args.exclude_by_tier:
+      if tier not in FILTER_BY_TIER:
+        continue
+      args.exclude.append(FILTER_BY_TIER[tier])
+
+  if len(args.filter) == 0:
```
cota, line 479: I'd leave the existing code here, seems more idiomatic:
args.filter = args.filter or [r"."]
args.exclude = args.exclude or [r"^$"]

benchmarks/experiment_runner.py
frgossen: done

benchmarks/experiment_runner.py
frgossen: done

benchmarks/experiment_runner.py
frgossen: done

benchmarks/experiment_runner.py
frgossen: done

cota (approved):
*clapping*

PR #5811, opened 2023-11-16
Re-implement `ComputationClient::WaitDeviceOps`

will-cromar:
- Implement `OperationTracker` to keep track of inflight operations
- Operations are tracked by the `Operation` RAII resource, which we can move into the `PjRtFuture`'s callback. This prevents us from transferring locks between threads (see below).
- Internally, the count of operations for each device is kept by `Counter`.

This PR aims to solve a few problems with the current implementation:

- Moving a `shared_lock` into a lambda and assigning that as a `PjRtFuture`'s callback is surprisingly slow. We currently have to do this asynchronously.
- Currently, we acquire the lock for a device _after_ we start execution. That means we can actually start new executions while another thread is waiting in `WaitDeviceOps`.
- Since the lock is acquired asynchronously, there can occasionally be a gap between `ExecuteReplicated` exiting and actually locking the device. In this scenario, `WaitDeviceOps` will have no effect. This mostly manifests in random unit test failures.
- We also release the device lock for each device before locking the next device. This also allows other threads to launch new operations during `WaitDeviceOps`.
- Apparently, unlocking a mutex from a different thread than the one that locked it is [undefined behavior](https://stackoverflow.com/questions/44207489/handing-over-locked-stdunique-lock-to-new-threads/44207588#44207588).

### DIFF
--- a/torch_xla/csrc/runtime/operation_tracker.cc
+++ b/torch_xla/csrc/runtime/operation_tracker.cc
@@ -62,14 +62,16 @@ void OperationTracker::Counter::Decrement() {
   }
 }
 
-std::unique_lock<std::shared_mutex> OperationTracker::Counter::BlockNewOperations() {
+std::unique_lock<std::shared_mutex>
+OperationTracker::Counter::BlockNewOperations() {
   return std::unique_lock(pending_operations_mu_);
 }
 
 void OperationTracker::Counter::Wait() {
   TF_VLOG(3) << "Waiting.... " << count_;
   std::unique_lock cv_lock(cv_mu_);
-  cv_.wait(cv_lock, [this]{ return count_.load(std::memory_order_acquire) == 0; });
+  cv_.wait(cv_lock,
+           [this] { return count_.load(std::memory_order_acquire) == 0; });
   TF_VLOG(3) << "Done waiting.";
 }
 
--- a/torch_xla/csrc/runtime/operation_tracker.cc
+++ b/torch_xla/csrc/runtime/operation_tracker.cc
@@ -11,7 +11,7 @@ namespace runtime {
 
 OperationTracker::OperationTracker(absl::Span<const std::string> devices) {
   for (auto& device : devices) {
-    op_counters_.emplace(device, std::make_unique<Counter>());
+    op_counters_.try_emplace(device);
   }
 }
 
@@ -24,7 +24,7 @@ OperationTracker::Operation::~Operation() { counter_->Decrement(); }
 
 std::unique_ptr<OperationTracker::Operation> OperationTracker::StartOperation(
     std::string device) {
-  return std::make_unique<Operation>(op_counters_.at(device).get());
+  return std::make_unique<Operation>(&op_counters_.at(device));
 }
 
 void OperationTracker::WaitForDevices(absl::Span<const std::string> devices) {
@@ -33,12 +33,12 @@ void OperationTracker::WaitForDevices(absl::Span<const std::string> devices) {
 
   for (const std::string& device_str : devices) {
     TF_VLOG(5) << "Blocking new operations on " << device_str;
-    auto lock = op_counters_.at(device_str)->BlockNewOperations();
+    auto lock = op_counters_.at(device_str).BlockNewOperations();
     locks.emplace_back(std::move(lock));
 
     TF_VLOG(3) << "Waiting for device execution for " << device_str
                << " to finish";
-    op_counters_.at(device_str)->Wait();
+    op_counters_.at(device_str).Wait();
     TF_VLOG(3) << "Finished operations on device " << device_str;
   }
 }
--- a/torch_xla/csrc/runtime/operation_tracker.cc
+++ b/torch_xla/csrc/runtime/operation_tracker.cc
@@ -11,7 +11,7 @@ namespace runtime {
 
 OperationTracker::OperationTracker(absl::Span<const std::string> devices) {
   for (auto& device : devices) {
-    op_counters_.try_emplace(device);
+    op_counters_.try_emplace(device, device);
   }
 }
 
@@ -48,16 +48,16 @@ void OperationTracker::Counter::Increment() {
   // atomic so we don't need an exclusive lock to prevent data races.
   std::shared_lock lock(pending_operations_mu_);
   auto current = count_.fetch_add(1, std::memory_order_acq_rel) + 1;
-  TF_VLOG(3) << "Increment.... " << current;
+  TF_VLOG(5) << "Incremented operations for " << device_ << " to " << current;
 }
 
 void OperationTracker::Counter::Decrement() {
   auto current = count_.fetch_sub(1, std::memory_order_acq_rel) - 1;
-  TF_VLOG(3) << "Decrement.... " << current;
+  TF_VLOG(5) << "Decremented operations for " << device_ << " to " << current;
 
   if (current == 0) {
     std::unique_lock cv_lock(cv_mu_);
-    TF_VLOG(3) << "notify";
+    TF_VLOG(3) << "All operations complete for " << device_;
     cv_.notify_all();
   }
 }
@@ -68,11 +68,11 @@ OperationTracker::Counter::BlockNewOperations() {
 }
 
 void OperationTracker::Counter::Wait() {
-  TF_VLOG(3) << "Waiting.... " << count_;
+  TF_VLOG(5) << "Waiting for " << count_ << " operations on " << device_;
   std::unique_lock cv_lock(cv_mu_);
   cv_.wait(cv_lock,
            [this] { return count_.load(std::memory_order_acquire) == 0; });
-  TF_VLOG(3) << "Done waiting.";
+  TF_VLOG(5) << "Done waiting for " << device_;
 }
 
 }  // namespace runtime
--- a/torch_xla/csrc/runtime/operation_tracker.cc
+++ b/torch_xla/csrc/runtime/operation_tracker.cc
@@ -50,6 +50,7 @@ void OperationTracker::Counter::Decrement() {
   TF_VLOG(3) << "Decrement.... " << current;
 
   if (current == 0) {
+    std::unique_lock cv_lock(cv_mu_);
     TF_VLOG(3) << "notify";
     cv_.notify_all();
   }
--- a/torch_xla/csrc/runtime/operation_tracker.cc
+++ b/torch_xla/csrc/runtime/operation_tracker.cc
@@ -1,0 +1,65 @@
+#include "torch_xla/csrc/runtime/operation_tracker.h"
+
+#include <mutex>
+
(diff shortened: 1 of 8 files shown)


will-cromar:
Confirmed that the idea here holds water. Performance sanity-check on llama with SPMD:

```
Totally decoded 1007 tokens in 7.06446 seconds
```

This PR is still a WIP while I clean up.

yeounoh:
Can we use this for mark_step(wait=True) as well?

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -704,6 +692,16 @@ PjRtComputationClient::ExecuteReplicated(
   // Required as of cl/518733871
   execute_options.use_major_to_minor_data_layout_for_callbacks = true;
 
+  // Grab the shared lock and block the `WaitDeviceOps` until buffer is
+  // ready. Since this is the SPMD code path. There is no points to grab
+  // devices lock for every individual device.
```
yeounoh, line 697: +1 we do this in pjrt layer for callbacks, too.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -715,6 +713,12 @@ PjRtComputationClient::ExecuteReplicated(
                   ->Execute(std::move(argument_handles), execute_options,
                             returned_futures)
                   .value();
+
+    (*returned_futures)[0].OnReady(
+        std::move([timed, op = std::move(op)](xla::Status unused) mutable {
```
yeounoh, line 718: Think the callback `OnReady` moves its argument implicitly, so no need for `std::move` on the input lambda.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -715,6 +713,12 @@ PjRtComputationClient::ExecuteReplicated(
...
+          timed.reset();
```
yeounoh, line 719: If `timed` is a shared_ptr already, can you just move & capture, insteaed of explicitly resetting?

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -704,6 +692,16 @@ PjRtComputationClient::ExecuteReplicated(
...
+  // ready. Since this is the SPMD code path. There is no points to grab
+  // devices lock for every individual device.
+  TF_VLOG(5) << "ExecuteReplicated acquiring PJRT device lock for "
+             << spmd_device_str;
+  auto op = operation_tracker_.StartOperation(spmd_device_str);
+  // auto lock = lock_device_shared(spmd_device_str);
+  TF_VLOG(5) << "ExecuteReplicated acquiring PJRT device lock for "
+             << spmd_device_str << " Done";
+
   std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures(
```
yeounoh, line 705: You probably don't need to initialize `returned_futures`, iirc -- you can simply pass in an uninitialized one (`std::vector<xla::PjRtFuture<xla::Status>> returned_futures;`) and the runtime client will do the rest.

yeounoh:
This is amazing, thanks @will-cromar  -- left a few comments for clean-up.

yeounoh (approved):
LGTM, thank you @will-cromar -- left minor comments for clean-up.

JackCaoG:
Thanks, will try to take a look today

torch_xla/csrc/runtime/pjrt_computation_client.cc
will-cromar: Yeah, you're right. I was following [this documentation for absl::AnyInvocable](https://github.com/abseil/abseil-cpp/blob/master/absl/functional/any_invocable.h#L83-L86), but that is only relevent if we assign the lambda to a variable.

torch_xla/csrc/runtime/pjrt_computation_client.cc
will-cromar: I was confused by this at first too. The trick is, we have to ensure the `TimedSection` ends before the `Operation` does, otherwise we may allow `WaitDeviceOps` to proceed before we register the `ExecuteReplicated` metrics. The destruction order of captured variables is undefined.

Adding a comment since we both questioned this.

torch_xla/csrc/runtime/pjrt_computation_client.cc
will-cromar: Good catch. I'm always happy to delete a bit of code.

will-cromar:
> Can we use this for mark_step(wait=True) as well?

I went through the code, and all we actually do with `wait` is [wait for the runtime to issue the `Execute` call]( https://github.com/pytorch/xla/blob/6addbde3a0764ae7ae3f898315637d7382733593/torch_xla/csrc/xla_graph_executor.cpp#L360-L362). I agree that also calling `WaitDeviceOps` makes more sense. Since that's a change in behavior, I'd rather send that in a follow-up PR.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -704,8 +693,17 @@ PjRtComputationClient::ExecuteReplicated(
...
 
-  std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures(
-      devices.size());
+  // Grab the shared lock and block the `WaitDeviceOps` until buffer is
+  // ready. Since this is the SPMD code path. There is no points to grab
+  // devices lock for every individual device.
+  TF_VLOG(5) << "ExecuteReplicated acquiring PJRT device lock for "
+             << spmd_device_str;
+  auto op = operation_tracker_.StartOperation(spmd_device_str);
+  // auto lock = lock_device_shared(spmd_device_str);
```
JackCaoG, line 702: remove this line?

torch_xla/csrc/runtime/operation_tracker.cc
```diff
@@ -0,0 +1,79 @@
...
+void OperationTracker::Counter::Increment() {
+  // Block new operations after Wait() is called. count_ is already atomic, so
+  // atomic so we don't need an exclusive lock to prevent data races.
+  std::shared_lock lock(pending_operations_mu_);
+  auto current = count_.fetch_add(1, std::memory_order_acq_rel) + 1;
+  TF_VLOG(5) << "Incremented operations for " << device_ << " to " << current;
+}
+
+void OperationTracker::Counter::Decrement() {
+  auto current = count_.fetch_sub(1, std::memory_order_acq_rel) - 1;
```
JackCaoG, line 55: not exactly sure about `fetch_sub`'s syntax but I am guessing it returns the value before subtraction?

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -704,8 +693,17 @@ PjRtComputationClient::ExecuteReplicated(
...
   execute_options.use_major_to_minor_data_layout_for_callbacks = true;
 
-  std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures(
-      devices.size());
+  // Grab the shared lock and block the `WaitDeviceOps` until buffer is
+  // ready. Since this is the SPMD code path. There is no points to grab
+  // devices lock for every individual device.
+  TF_VLOG(5) << "ExecuteReplicated acquiring PJRT device lock for "
+             << spmd_device_str;
+  auto op = operation_tracker_.StartOperation(spmd_device_str);
```
JackCaoG, line 701: nit, can we rename this to `op_tracker`? `op` is a really confusing name, I thought it is the operations that needs to be executed when I read the `OnReady` code down below.

JackCaoG (approved):
Reading lock related code hurts my brain every time lol. Minor nits, mostly lgtm

torch_xla/csrc/runtime/pjrt_computation_client.cc
will-cromar: oops, my bad

torch_xla/csrc/runtime/operation_tracker.cc
will-cromar: Yes. This is just a quirk of how atomics have to work.

will-cromar: After looking at this again, I'm half inclined to replace the atomic counter with locks and a normal int. I don't expect contention to be a concern because there will be relatively few concurrent users of that mutex, and I can stop anyone else from having to read up on `std::atomic` and memory order. What do you think?

torch_xla/csrc/runtime/pjrt_computation_client.cc
will-cromar: Now that we've solved the easy problems (synchronizing concurrent operations between devices), let's move on to the hard one: naming things.

I don't think `op_tracker` works here, because the parent owner of all of the operation counters is called `OperationTracker`. It seems conventional in C++ to name RAII resources directly after what they represent. E.g., all of the lock objects are just calling `lock` on a mutex, but they don't hold the lock state themselves. I do get the confusion though.

will-cromar:
Hmmm, some of the unit tests started suspiciously crashing after 77f4b14, but everything passed with a8462d5. Maybe one of those changes was not actually a no-op...

torch_xla/csrc/runtime/operation_tracker.cc
JackCaoG: I suspect most people will ever read into this implementation, both ways works for me.

torch_xla/csrc/runtime/pjrt_computation_client.cc
JackCaoG: oh I see.. it is actually an op.... it is very confusing through haha

will-cromar: I'm okay calling this resource (currently `Operation`) an `OperationTracker`. But, then we need a new name for `OperationTracker`. Perhaps `OperationTrackerTracker`?

JackCaoG: lol `OperationTrackerManager`?

will-cromar: Oof, "`TrackerManager`" gives me Java vibes

torch_xla/csrc/runtime/operation_tracker.cc
will-cromar: It turns out using a lock+int counter is actually noticeably slower. I'll just take the win instead of interrogating why...

torch_xla/csrc/runtime/pjrt_computation_client.cc
will-cromar: Settled for "OperationManager" and "OperationTracker" as the names. Let me know if that works for you.

JackCaoG (approved):

PR #5812, opened 2023-11-16
Add profiling capabilities to experiment_runner.py script

golechwierowicz:
This PR adds new capabilities around profiling to the `experiment_runner.py`.
`--profile-cuda` enables the profiler, while `--profile-cuda-dump` specifies the directory to dump the profiling information to.
These two flags are separated because in the next PR I have tracking CPU, and CUDA time separately ready. So it might the case that we want to do either just prof dumping, or just CPU/CUDA tracking, or both.


benchmarks/benchmark_model.py
```diff
@@ -98,15 +98,10 @@ def _prepare_for_eval(self):
   def _prepare_for_train(self):
     self.module.train()
     self.model_iter_fn = self.train
-    if self.benchmark_experiment.dynamo == "openxla":
```
vanbasten23, line 101: any reason to remove it?

benchmarks/experiment_runner.py
```diff
@@ -242,26 +257,43 @@ def timed_run(self, benchmark_experiment, benchmark_model):
...
-      if benchmark_experiment.xla:
-        t_trace += time.perf_counter() - t_trace_start
+        if prof:
+          prof.step()
 
-      self._mark_step(benchmark_experiment)
+        return output
+
+    if enable_prof:
+      with profile(activities=[ProfilerActivity.CUDA, ProfilerActivity.CPU]) as prof:
```
vanbasten23, line 287: cool!

benchmarks/experiment_runner.py
```diff
@@ -232,6 +233,20 @@ def prepare_inputs(self, example_inputs, should_randomize_input):
       inputs_list.append(inputs)
     return inputs_list
 
+  def dump_profile_info(self, prof, model_name):
+    assert prof is not None, 'Expecting profiler to be defined!'
+    if not self._args.profile_cuda_dump:
```
vanbasten23, line 238: what do you think we also return an error here, just so users wouldn't wonder where the dump goes?

benchmarks/util.py
```diff
@@ -1,154 +1,155 @@
...
+    if benchmark_experiment and benchmark_experiment.xla:
+        device = benchmark_experiment.get_device()
+        xm.set_rng_state(1337, str(device))
 
 
 @functools.lru_cache(maxsize=3)
 def is_xla_device_available(devkind):
-  if devkind not in ["CPU", "CUDA", "TPU"]:
-    raise ValueError(devkind)
+    if devkind not in ["CPU", "CUDA", "TPU"]:
```
vanbasten23, line 53: any reason to change the indentation?
If all the other files indents 2 spaces, let's be consistent and also indent 2 spaces

golechwierowicz: bad merge

benchmarks/experiment_runner.py
golechwierowicz: :)

benchmarks/experiment_runner.py
golechwierowicz: I can add additional logging, but returning an error does not make much sense, as it is not an error really.

benchmarks/experiment_runner.py
```diff
@@ -453,6 +486,20 @@ def parse_args(args=None):
       help="JSON string of the model config dict.",
   )
 
+  parser.add_argument(
+      "--profile-cuda",
```
cota, line 490: When does it make sense to pass --profile-cuda and not pass --profile-cuda-dump? If it doesn't, we should merge the two flags.

benchmarks/experiment_runner.py
```diff
@@ -232,6 +233,21 @@ def prepare_inputs(self, example_inputs, should_randomize_input):
       inputs_list.append(inputs)
     return inputs_list
 
+  def dump_profile_info(self, prof, model_name):
+    assert prof is not None, 'Expecting profiler to be defined!'
+    if not self._args.profile_cuda_dump:
+      logger.warning('Profling enabled, but dumping tracing/kernel summary disabled.')
```
cota, line 239: s/Profling/Profiling/

cota: Oh I can do suggestions here as well (Sorry I'm new to reviewing on Github):
```suggestion
      logger.warning('Profiling enabled, but dumping tracing/kernel summary disabled.') 
```

benchmarks/experiment_runner.py
golechwierowicz: Check the desc. I've a PR ready which introduces a flag which will allow for gathering CPU + CUDA time, so you can either use profile-cuda-dump, or new flag or both. Merging profiling instantiation into either of flags, or both complicates a code (as you have to check whether profiler was instantiated already etc.), so that's why it's separated.

Because it's git, and it's hard to do the stacked diffs I'm keeping it locally.

cota: Missed that, thanks! I am new to this reviewing UI -- apologies.
Re: stacked diffs, git can help a bit with --update-refs: https://andrewlock.net/working-with-stacked-branches-in-git-is-easier-with-update-refs/
Note: AFAIK on Github in order to show the right diff's, one has to push all the branches involved to the repo in question -- i.e. one cannot choose a branch in a fork as a diff base. Perhaps that's why this repo has 2000+ branches?

benchmarks/experiment_runner.py
vanbasten23: If the user enables `--profile-cuda`, is the dump the only place where the user can see the profile?

vanbasten23 (approved):
LGTM

PR #5813, opened 2023-11-16
Revert "update dependency errir" and "update run model config call interface, optimizer and result analyze script"

frgossen:
Revert "update dependency errir"
Revert "update run model config call interface, optimizer and result analyze script"


zpcore:
Thanks for pointing out the issue. LGTM

zpcore:
Oh wait, we may not want to merge into `master` branch. This should go to the `benchmark` branch first.

PR #5818, opened 2023-11-17
add torch_xla 2.0 with python 3.10 wheel

ManfeiBai:
cc @zpcore

### DIFF
--- a/infra/tpu-pytorch-releases/artifacts.auto.tfvars
+++ b/infra/tpu-pytorch-releases/artifacts.auto.tfvars
@@ -96,6 +96,12 @@ versioned_builds = [
     accelerator     = "tpu"
   },
   {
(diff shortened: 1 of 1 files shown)


JackCaoG:
why do we need py3.10 whl for 2.0?

ManfeiBai:
> why do we need py3.10 whl for 2.0?

torchbench might need this wheel cc @zpcore

zpcore:
Thanks @ManfeiBai for building the 2.0. Currently I am working on reproducing the torchbench result from 6 months ago. python 3.10 version torch_xla 2.0 will help in reproducing the results since all the other benchmark releated packages I use are built for python 3.10.

zpcore:
I already found the intended version in docker image. Guess I don't need it anymore. Thanks @ManfeiBai for building the wheel.

Issue #5746, opened 2023-10-28
Torchbench `speech_transformer` fails when calling `_get_graph_hash`.


ysiraichi:
## 🐛 Bug

Running the `speech_transformer` benchmark of Torchbench fails with the following error: 

```python
  File "torch/_dynamo/backends/torchxla.py", line 49, in fwd
    compiled_graph = bridge.extract_compiled_graph(model, args)
  File "xla/torch_xla/core/dynamo_bridge.py", line 552, in extract_compiled_graph
    extract_internal(fused_module), node.args, None)
  File "xla/torch_xla/core/dynamo_bridge.py", line 354, in extract_internal
    dumb_return_handler, xla_args_need_update) = extract_graph_helper(xla_model)
  File "xla/torch_xla/core/dynamo_bridge.py", line 285, in extract_graph_helper
    graph_hash = torch_xla._XLAC._get_graph_hash(args_and_out)
RuntimeError: ./torch_xla/csrc/runtime/pjrt_computation_client.h:134 : Check failed: HasValue()
*** Begin stack trace ***
        tsl::CurrentStackTrace[abi:cxx11]()
        torch_xla::runtime::PjRtComputationClient::PjRtData::GetHandle()
        torch::lazy::LazyGraphExecutor::RunPostOrder(std::vector<torch::lazy::Value, std::allocator<torch::lazy::Value> > const&, torch::lazy::LazyGraphExecutor::SyncTensorCollection*)
        torch_xla::XLAGraphExecutor::RunPostOrder(std::vector<torch::lazy::Value, std::allocator<torch::lazy::Value> > const&, torch::lazy::LazyGraphExecutor::SyncTensorCollection*)
        torch_xla::XLAGraphExecutor::GetGraphHash(std::vector<c10::intrusive_ptr<torch_xla::XLATensor, c10::detail::intrusive_target_default_null_type<torch_xla::XLATensor> >, std::allocator<c10::intrusive_ptr<torch_xla::XLATensor, c10::detail::intrusive_target_default_null_type<torch_xla::XLATensor> > > > const&)
```

<details>
<summary>Raw Output</summary>

```
ERROR:common:Backend dynamo failed in warmup()
Traceback (most recent call last):
  File "benchmarks/dynamo/common.py", line 2462, in warmup
    fn(model, example_inputs)
  File "torch/_dynamo/eval_frame.py", line 401, in _fn
    return fn(*args, **kwargs)
[...]

PR #5823, opened 2023-11-18
Sync `xla_args` before computation.

ysiraichi:
Fix: #5746 

This PR synchronizes `FunctionalTensorWrapper` objects, before actually computing them. As a result, `_clear_pending_irs` doesn't delete the computed tensor anymore.

### DIFF
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -455,6 +455,12 @@ class InputCollector(torch.fx.Interpreter):
 
 
 def extract_compiled_graph(xla_model: torch.fx.GraphModule, xla_args):
+  # Synchronize xla_args, so that each FunctionalTensorWrapper argument updates its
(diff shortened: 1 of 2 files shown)


torch_xla/core/dynamo_bridge.py
```diff
@@ -455,6 +455,12 @@ def call_module(self, target, args, kwargs):
 
 
 def extract_compiled_graph(xla_model: torch.fx.GraphModule, xla_args):
+  # Synchronize xla_args, so that each FunctionalTensorWrapper argument updates its
+  # value reference before actually computing it.
+  for a in xla_args:
+    if isinstance(a, torch.Tensor) and torch._is_functional_tensor(a):
+      torch._functionalize_sync(a)
```
JackCaoG, line 462: first time learnt this api.. interesting.

JackCaoG (approved):

PR #5824, opened 2023-11-19
Parallelize device-to-host transfers

jonb377:
For async checkpointing, it is important to unblock training as quickly as possible. Training can only be unblocked when all data has been moved off of device onto the host to free up device memory.

One bottleneck I found was that in TransferFromServer, the tensors are transferred using `ToLiteralSync`, meaning each tensor is transferred sequentially. In benchmarking a 2B parameter model, parallelizing these transfers decreased the time spent in TransferFromServer from 5.1s to 1.8s, ~65% reduction.

There is still significant overhead from copying the resulting xla::Literal into torch.Tensor, but that's for another PR.


will-cromar (approved):
Thanks!

yeounoh (approved):
LGTM

PR #5825, opened 2023-11-20
Distribute Literal->Tensor copies across thread pool

jonb377:
After an xla::Literal has been created in TransferFromServer, it must be copied into an at::Tensor. This incurs a significant amount of overhead (up to 3x the transfer overhead after #5824). This is because the copies still occur synchronously on a single thread.

This change dispatches the copies to a thread pool to speed up the process. When checkpointing a 2B parameter model, the overhead decreases from ~5s to ~611ms.


torch_xla/csrc/tensor_util.cpp
```diff
@@ -796,13 +796,18 @@ std::vector<xla::Literal> ReleaseGilAndTransferData(
 
 std::vector<at::Tensor> XlaDataToTensors(
     absl::Span<const torch::lazy::BackendDataPtr> xla_data,
-    at::ScalarType dest_element_type) {
+    absl::Span<const at::ScalarType> dest_element_type) {
```
will-cromar, line 799: What's the reason for this change?

torch_xla/csrc/tensor_util.cpp
```diff
@@ -796,13 +796,18 @@ std::vector<xla::Literal> ReleaseGilAndTransferData(
...
   std::vector<xla::Literal> literals = ReleaseGilAndTransferData(xla_data);
-  std::vector<at::Tensor> tensors;
-  tensors.reserve(literals.size());
-  for (auto& literal : literals) {
-    tensors.push_back(MakeTensorFromXlaLiteral(literal, dest_element_type));
+  std::vector<at::Tensor> tensors(literals.size());
+  absl::BlockingCounter counter(literals.size());
+  for (size_t i = 0; i < tensors.size(); ++i) {
+    auto copy_fn = [&, i]() {
```
will-cromar, line 804: Can you capture the variables you need explicitly?

torch_xla/csrc/tensor_util.cpp
```diff
@@ -796,13 +796,18 @@ std::vector<xla::Literal> ReleaseGilAndTransferData(
 
 std::vector<at::Tensor> XlaDataToTensors(
     absl::Span<const torch::lazy::BackendDataPtr> xla_data,
-    at::ScalarType dest_element_type) {
+    absl::Span<const at::ScalarType> dest_element_type) {
   std::vector<xla::Literal> literals = ReleaseGilAndTransferData(xla_data);
```
will-cromar, line 800: I wonder if we should just be returning `Tensor`s here

will-cromar (approved):

torch_xla/csrc/tensor_util.cpp
will-cromar: Actually, you need just about every variable in this scope since it's pretty narrow. I take that back.

torch_xla/csrc/tensor_util.cpp
JackCaoG: it seems like we never really call `XlaDataToTensors ` with different `dest_element_type`. @jonb377 are you introducing a new use case? if not we can keep it as a singleton?

JackCaoG (approved):

torch_xla/csrc/tensor_util.cpp
jonb377: The reason to make the dest_elem_types a vector is actually the next change - I'm batching the local shard transfers for many tensors into a single XlaDataToTensors call. I probably should have kept this refactor with the upcoming change... But it makes that PR slightly smaller.

torch_xla/csrc/tensor_util.cpp
jonb377: I'm interested in making TransferFromServer return at::Tensor and cut out the xla::Literal middleman, but that's in the idea phase. Opted to keep this change smaller and just distribute the copy work over more cores.

jonb377:
Thanks for the reviews @will-cromar and @JackCaoG!

yeounoh (approved):
LGTM

PR #5826, opened 2023-11-20
Vectorize local shard retrieval

jonb377:
To capitalize on the improvements in #5824 and #5825, moving tensor shards to CPU should be batched. This change does the following:
- Make `_get_local_shards` and `_get_local_shard_replica_and_indices` operate on lists of tensors instead of individual tensors.
- Updates `_sharded_cpu_state_dict` to use the batched method across all sharded tensors in the state_dict.

With all three changes applied, the amount of time spent transferring the state_dict to CPU for a 2B parameter model decreases from >10s to 3.4s, which unblocks training much more quickly.

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1816,10 +1816,10 @@ void InitXlaModuleBindings(py::module m) {
   // of the shards returned from `_get_local_shards`.
   m.def(
       "_get_local_shard_replica_and_indices",
-      [](const std::vector<at::Tensor>& input)
+      [](const std::vector<at::Tensor>& input_tensors)
           -> std::vector<std::vector<std::pair<int, py::object>>> {
         std::vector<std::vector<std::pair<int, py::object>>> result;
-        for (auto& tensor : input) {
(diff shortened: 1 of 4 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1711,48 +1711,67 @@ void InitXlaModuleBindings(py::module m) {
...
   // slices for each dimension or an Ellipsis object indicating that the tensor
   // is replicated. These indices will not reflect any padding that has been
   // applied to the shards. The order of the returned indices matches the order
   // of the shards returned from `_get_local_shards`.
-  m.def("_get_local_shard_replica_and_indices",
-        [](const at::Tensor& input) -> std::vector<std::pair<int, py::object>> {
-          XLATensorPtr xtensor = bridge::GetXlaTensor(input);
+  m.def(
+      "_get_local_shard_replica_and_indices",
+      [](const std::vector<at::Tensor>& input)
```
yeounoh, line 1770: nit. rename `input` to `inputs` or `input_tensors`

yeounoh (approved):
LGTM

PR #5831, opened 2023-11-20
Use cuda 12.1 and build pytorch with cuda enabled.

vanbasten23:
This PR:
- build pytorch with cuda enabled in https://github.com/pytorch/xla/blob/master/docs/gpu.md#develop-pytorchxla-on-a-gpu-instance-build-pytorchxla-from-source-with-gpu-support
- use cuda 12.1 because the multinode training feature only works for cuda 12+.
- replace previous cuda 12.0 with cuda 12.1 so that we can be consistent with pytorch perhttps://github.com/pytorch/pytorch/blob/main/RELEASE.md

The change has been validated [here](b/310958350).

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,10 +147,11 @@ bucket.
 
 | Version | Cloud TPU VMs Wheel |
 | --- | ----------- |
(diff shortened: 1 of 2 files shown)


JackCaoG (approved):
I take it that you already verified these whls?

vanbasten23:
> I take it that you already verified these whls?

Yeah, you can see the validation result [here](https://github.com/pytorch/xla/pull/b/310958350)

vanbasten23:
Thanks for the review!

PR #5838, opened 2023-11-21
Add stack frame id metadata and correct filename and line number for custom op_name

mrnikwaws:
### Summary
This adds one feature and a bug fix for HLO lowered via the lowering context:
* Populate stack frames information in hlo_module where IR debug is enabled
* Pass a max stack depth to ignore un-needed stack frames

### Testing
* Extend HLO metadata test
* Add new test for confirming stack frame information (TBC)


JackCaoG:
@qihqi do you have cycle to take this one?

mrnikwaws:
As discussed offline this change adds full stack frame lowering and works to correct past problems with incorrect file name and line number when using a TorchDispatch based interceptor for lowering.

The key remaining issue is for lowered tensors declared like this in a TorchDispatch scope:

```
    with CustomOpNameLowering() as c:
      model = model.to(device=xm.xla_device())

      # It appears there is no node to attach metadata to at this line
      #inp = torch.rand(4, 4, device=xm.xla_device()) 

      # This works correctly
      inp = torch.rand(4, 4)
      inp = inp.to(device=xm.xla_device())

      out = model(inp)
```

mrnikwaws:
As discussed offline some lowerings result in multiple XlaNodes.  The latest change add a new method to recursively setting metadata on nodes where it was not set prior.  Since the graph is traversed forward this should only set metadata on nodes added since the last dispatch.  

Unit test passes for all op_name and stack trace checks

torch_xla/csrc/lowering_context.cpp
```diff
@@ -67,11 +80,24 @@ class HloMetadataSetter {
     metadata.set_op_name(absl::StrCat(op_name_prefix, op_type));
 
     if (!nmeta.frame_info.empty()) {
```
qihqi, line 82: Here we are attempting to correct the following 3 fields:
* source_file
* source_line
* stack_frame_id

Instead of having custom computation here, we should have those passed in.

The location of this change should be inside of of the if in line 69.

torch_xla/csrc/ir.h
```diff
@@ -195,6 +194,15 @@ T* NodeCast(const torch::lazy::Node* node, torch::lazy::OpKind op) {
   return const_cast<T*>(casted);
 }
 
+struct CustomOpNameMetaData : public torch::lazy::UserMetaData {
+  CustomOpNameMetaData(const std::string& input_op_name_prefix,
+                       int input_max_stack_depth)
+      : op_name_prefix(input_op_name_prefix),
+        max_stack_depth(input_max_stack_depth) {}
+  std::string op_name_prefix;
```
qihqi, line 202: Let's make this struct to contain the exact things you need to override in HLO proto, i.e. 
source_file, source_line, stack_frame_id etc.

torch_xla/csrc/lowering_context.cpp
qihqi: basically the changes in this function should read something along the lines of:
* if there are custom metadata: pass it down to HLO; else: do what it used to do before. No custom computations should happen here.

torch_xla/csrc/lowering_context.h
```diff
@@ -110,6 +115,14 @@ class LoweringContext : public torch::lazy::LoweringContext {
       parameters_map_;
   std::vector<xla::XlaOp> root_tuple_;
   OutputMap<xla::XlaOp> emitted_outputs_;
+
+  // Stack frame index tables - we accumulate and write these to the HloModule
+  xla::StackFrameIndexProto indexes_;
```
qihqi, line 120: These fields should be unneccesary if we compute the relevant information in Python

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1918,15 +1918,14 @@ void InitXlaModuleBindings(py::module m) {
         [](at::Tensor& self, const at::Tensor& source) -> at::Tensor& {
           return XLANativeFunctions::set_(self, source);
         });
-  m.def("_set_xla_custom_op_name",
-        [](const at::Tensor& input, const std::string& op_name) {
+  m.def("_set_xla_custom_op_name_prefix",
```
qihqi, line 1921: rename to `_set_user_metadata` and take a torch::lazy::UserMetadata if possible.

Let's pass in everything that can be forwarded to HLO here (like filename, line num, whatever Indexes etc); so no computation happens in C++.

qihqi: Can pass in a python dict (which becomes a map here) with all needed content and do creation of CustomOpNameMetadata if you feel exposing CustomOpNameMetadata through pybind11 is too much work.

torch_xla/csrc/lowering_context.cpp
```diff
@@ -263,8 +312,66 @@ void LoweringContext::AddParameter(const torch::lazy::Output& output,
   return;
 }
 
+// TODO: Remove duplicate code copied form openxla/xla MHLO -> HLO
+int64_t LoweringContext::AddStackFrameLocation(
```
qihqi, line 316: Can you put a link to openxla/xla code that has the similar code?

torch_xla/csrc/lowering_context.h
```diff
@@ -110,6 +115,14 @@ class LoweringContext : public torch::lazy::LoweringContext {
...
+
```
qihqi, line 121: I am trying to think a way to not keep these states, and I think way to accomplish that is the following:

1. Here we keep one extra field `std::vector< std::vector< FrameInfo>*> finfos` here. 
2. In lowering_context.cpp; we just store the pointer `&nmeta.frame_info` here.
3. Have a function `void FillStackFrameIds( const std::vector< std::vector<FrameInfo>*>& finfos, StackFrameIndexProto* )` etc that, takes everything and computes Stack frame index; then; these maps can be local variable to that function. This function can be called in `BuildXLA`.
wdyt?

torch_xla/csrc/lowering_context.cpp
mrnikwaws: Yes

torch_xla/csrc/lowering_context.h
mrnikwaws: One concern with the above is: if the torch::lazy::Node graph goes out of scope (is deleted) prior to lowering this will seg fault on frame info vector pointers.  Is the graph guaranteed to stay live until lowering is complete?

mrnikwaws: Also realized this means we don't compute and lower the stack frame id into the HLO metadata at the moment of lowering a given node.  This means that we have to *also* somehow find that node after the fact as we are parsing the frames post node lowering.  I think this is a lot less efficient.  Let me take a look at the state storaga again ...

mrnikwaws: So the storage here is actually pretty efficient: https://github.com/pytorch/xla/pull/5838/files#diff-04a457b5166774690501626cb60a8b2ae595a9abf4359f70f88a584ab750b274R122-R123 is only storing string_view objects (not strings) so this is just two pointers each. AFAIK (i.e. start and end).  So state is only really accumulating in the StackFrameIndex proto itself.  This means at the moment of lowering a node there are exactly two copies of the path and function strings.  If the nodes are quickly deleted after lowering this would limit the memory usage.

mrnikwaws: I understand that we would like to move the code complexity out of this class.  I'd recommend that post shipping the change we make a note and start the process of building this into openxla: https://github.com/openxla/xla/blob/952d3cf39c3e3eeaa790cc1dd53423c8eb27d473/xla/client/xla_builder.h#L70

i.e. extend XlaBuilder / create a new helper that populates the underlying HloProto in the XlaComputation class.  This would be in line with how the rest of the HloProto is constructed and return the LoweringContext to a minimal form

However this will make no difference to memory overhead and needs to be a two-step to meet the 12/1 deadline for 2.1.2 and 2.2 releases.

qihqi (approved):

torch_xla/csrc/lowering_context.cpp
```diff
@@ -51,13 +55,20 @@ class HloMetadataSetter {
...
-    if (xla_node_cast != nullptr && !xla_node_cast->custom_op_name().empty()) {
-      op_name_prefix = xla_node_cast->custom_op_name();
+    std::string op_name_prefix;
+    size_t max_stack_depth = nmeta.frame_info.size();
+
+    if (custom_opname_meta != nullptr) {
+      op_name_prefix = custom_opname_meta->op_name_prefix;
+      max_stack_depth = custom_opname_meta->max_stack_depth;
+    } else {
+      TF_LOG(WARNING) << "No custom opname metadata! op_type=" << op_type;
```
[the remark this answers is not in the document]
JackCaoG: it should not be a spam, let's not use TF_LOG and use VLOG for this and backport the change to 2.2 maybe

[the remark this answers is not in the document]
alanwaketan: It looks like that this case is actually the normal case and shouldn't be warned. Judging by the numbers of ops don't have the metadata...

PR #5907, opened 2023-11-21
update result analyze script based on Ranran's fix

zpcore:


benchmarks/util.py
```diff
@@ -141,14 +141,10 @@ def get_cpu_name():
 
 
 def get_gpu_name():
-  gpu_names = subprocess.check_output(
+  return subprocess.check_output(
```
Liyang90, line 144: Is this change intended or not? It seems to be reverting to older codes.

benchmarks/result_analyzer.py
```diff
@@ -43,7 +54,6 @@ def run(self):
         "accelerator": pd.Series(dtype="str"),
         "accelerator_model": pd.Series(dtype="str"),
         "xla": pd.Series(dtype="str"),
```
Liyang90, line 56: "xla_flags" was removed here. Is it intentional or accidental reverting?

benchmarks/util.py
zpcore: You are right, thanks, I have reverted this line back.

benchmarks/result_analyzer.py
zpcore: Thanks for catching that. I have added this line back.

zpcore:
Add @RissyRan in this PR since she is the main author for the change.

benchmarks/result_analyzer.py
```diff
@@ -58,13 +69,92 @@ def run(self):
...
     for file in jsonl_files:
-      metric_df = self.extract_metrics(file, metric_df)
+      metric_df = self.extract_metrics_csv(file, metric_df)
 
     # additional processing of the metric_df can be done here
 
     self.export_metric_report(metric_df)
 
-  def extract_metrics(self, file, metric_df):
+  def get_calculated_metrics(self, d, tmp):
```
Liyang90, line 78: Could this method args be made more readable? It's okay to use `tmp` in the method call, but in method definition, it's better to reflect the actual object that is passed.

In this case `tmp` is a dict read from a json string representing the raw results of one experiment.

Liyang90 (commented):
LGTM with minor suggestion

miladm:
would be great to merge this PR today if at all possible. thank you @zpcore

frgossen:
Just a note, we should merge this into the main branch

benchmarks/result_analyzer.py
zpcore: Thanks, I will make the update. By the way, we will drop this PR and instead, merge this into main branch directly.

frgossen (changes requested):
This should be merged into the main branch

zpcore:
Created a new PR: https://github.com/pytorch/xla/pull/5920 to replace this one.

frgossen:
Thanks!

PR #5483, opened 2023-08-23
Update artifacts.auto.tfvars for 2.1 release

wonjoolee95:
Update artifacts.auto.tfvars for 2.1 release. The actual branch cut will be on 08/28. We'll continue to sync this branch with master on a daily basis. Setting this up now to see if there will be any errors when we create these new triggers for 2.1.

Tag for 2.1 has also been added -- https://github.com/pytorch/xla/tags.

PR #5683, opened 2023-10-06
Update artifacts.auto.tfvars for cuda 12.1 wheel

ManfeiBai:

PR #5908, opened 2023-11-22
Update artifacts.auto.tfvars for 2.2 release

zpcore:
Add the 2.2 release build similar to 2.1. We will trigger the 2.2 wheel build for xl-ml test. Refer to 2.1 release [PR](https://github.com/pytorch/xla/pull/5483)

Tag for 2.2 has also been added -- https://github.com/pytorch/xla/tags.

### DIFF
--- a/infra/tpu-pytorch-releases/artifacts.auto.tfvars
+++ b/infra/tpu-pytorch-releases/artifacts.auto.tfvars
@@ -55,23 +55,6 @@ xrt_versioned_builds = [
   },
 ]
 
-# TODO: Remove this after the 2.2 release
-xrt_versioned_builds = [
-  {
-    accelerator    = "tpu"
-    python_version = "3.10"
-    pytorch_git_rev = "v2.2.0"
-    package_version = "2.2.0+xrt"
-  },
-  {
-    accelerator  = "cuda"
-    python_version = "3.10"
-    cuda_version = "12.0"
-    pytorch_git_rev = "v2.2.0"
-    package_version = "2.2.0+xrt"
-  },
-]
-
 # Built on push to specific tag.
 versioned_builds = [
   # Remove libtpu from PyPI builds
--- a/infra/tpu-pytorch-releases/artifacts.auto.tfvars
+++ b/infra/tpu-pytorch-releases/artifacts.auto.tfvars
@@ -38,62 +38,40 @@ nightly_builds = [
   }
 ]
 
-# TODO: Remove this after the 2.1 release
-xrt_versioned_builds = [
-  {
-    accelerator    = "tpu"
-    python_version = "3.10"
-    pytorch_git_rev = "v2.1.0"
-    package_version = "2.1.0+xrt"
-  },
-  {
-    accelerator  = "cuda"
-    python_version = "3.10"
-    cuda_version = "12.0"
-    pytorch_git_rev = "v2.1.0"
-    package_version = "2.1.0+xrt"
-  },
-]
-
 # Built on push to specific tag.
 versioned_builds = [
   # Remove libtpu from PyPI builds
   {
-    git_tag         = "v2.2.0"
-    pytorch_git_rev = "v2.2.0"
+    git_tag         = "v2.2.0-rc1"
     package_version = "2.2.0"
     accelerator     = "tpu"
     bundle_libtpu   = "0"
   },
   {
-    git_tag         = "v2.2.0"
-    pytorch_git_rev = "v2.2.0"
+    git_tag         = "v2.2.0-rc1"
     package_version = "2.2.0"
     accelerator     = "tpu"
     python_version  = "3.9"
     bundle_libtpu   = "0"
   },
   {
-    git_tag         = "v2.2.0"
-    pytorch_git_rev = "v2.2.0"
-    package_version = "2.2.0"
+    git_tag         = "v2.2.0-rc1"
+    package_version = "2.2.0-rc1"
     accelerator     = "tpu"
     python_version  = "3.10"
     bundle_libtpu   = "0"
   },
   {
-    git_tag         = "v2.2.0"
-    pytorch_git_rev = "v2.2.0"
-    package_version = "2.2.0"
+    git_tag         = "v2.2.0-rc1"
+    package_version = "2.2.0-rc1"
     accelerator     = "tpu"
     python_version  = "3.11"
     bundle_libtpu   = "0"
   },
   # Bundle libtpu for Kaggle
   {
-    git_tag         = "v2.2.0"
-    pytorch_git_rev = "v2.2.0"
-    package_version = "2.2.0+libtpu"
+    git_tag         = "v2.2.0-rc1"
+    package_version = "2.2.0-rc1+libtpu"
     accelerator     = "tpu"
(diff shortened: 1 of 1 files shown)


zpcore:
I am not sure if we still need `package_version = "2.2.0+xrt"`.

lsy323 (approved):
LGTM, thanks!

infra/tpu-pytorch-releases/artifacts.auto.tfvars
```diff
@@ -108,6 +148,43 @@ versioned_builds = [
     package_version = "1.13"
     accelerator     = "tpu"
   },
+  {
+    git_tag         = "v2.2.0"
+    pytorch_git_rev = "v2.2.0"
+    package_version = "2.2.0",
+    accelerator     = "cuda"
+    cuda_version    = "12.0"
```
will-cromar, line 156: Which CUDA version are we targeting? Should we just pick the same CUDA version `torch` will use in their default wheel?

infra/tpu-pytorch-releases/artifacts.auto.tfvars
```diff
@@ -58,6 +58,46 @@ xrt_versioned_builds = [
 # Built on push to specific tag.
 versioned_builds = [
   # Remove libtpu from PyPI builds
+  {
+    git_tag         = "v2.2.0"
+    pytorch_git_rev = "v2.2.0"
```
will-cromar, line 63: FYI, PyTorch doesn't apply the `v2.2.0` tag until the final release. They'll have `v2.2.0rc0`, `v2.2.0rc1`, etc. I actually suggest we do the same for this release rather than updating the `v2.2.0` tag multiple times.

infra/tpu-pytorch-releases/artifacts.auto.tfvars
zpcore: How about we leave 11.8 and 12.1 here?

lsy323: I think this makes sense, I found Pytorch 2.1.1 is using CUDA 12.1 by default on pypi, let's use 12.1 as well. cc @vanbasten23, to double check our stack works with CUDA 12.1

infra/tpu-pytorch-releases/artifacts.auto.tfvars
zpcore: Thanks, can you elaborate more about the solution? Should we add `v2.2.0rc0`, `v2.2.0rc1` build etc. here in advance?

infra/tpu-pytorch-releases/artifacts.auto.tfvars
zpcore: Hi @lsy323 , do we still need CUDA 11.8 for 2.2 release?

infra/tpu-pytorch-releases/artifacts.auto.tfvars
ManfeiBai: Thanks for this, Will

Hi, @zpcore, FYI, since we want to add trigger for 2.2.0 wheel to test in xl-ml-dashboard, and we won't catch `pytorch_git_rev` tag `v2.2.0` until PyTorch 2.2 release from PyTorch, and we also won't catch any other tag like `v2.2.0rc0`/`v2.2.0rc1` until PyTorch 2.2.0 branch cut date, we could merge this PR now but the trigger could build wheel successfully until they catch the tag

So for wheel on xl-ml-dashboard, let's use nightly wheel of torch and torch_xla until branch cut date

cc @lsy323

infra/tpu-pytorch-releases/artifacts.auto.tfvars
ManfeiBai: Hi, @will-cromar, thanks, do we know how to confirm the CUDA version `torch` will use in their default wheel? would they mention in the blog related to 2.2 release?

since there were many PRs to add specific CUDA version to 2.1 release like https://github.com/pytorch/xla/pull/5683, so there might be some requirements to build 2.2 wheel with these CUDA versions. And if build failed, we could catch the failure early too

For resources, do we want to delete triggers for 2.1 to limit the resource used for release here?

infra/tpu-pytorch-releases/artifacts.auto.tfvars
ManfeiBai: > Thanks, can you elaborate more about the solution? Should we add `v2.2.0rc0`, `v2.2.0rc1` build etc. here in advance?

let's modify `git_tag` and `pytorch_git_rev` to `"v2.2.0rc0"` in this PR, we would modify again once PyTorch has other wheels built

cc @will-cromar, would you mind help to confirm it?

zpcore: Thanks, we have removed `pytorch_git_rev`.

infra/tpu-pytorch-releases/artifacts.auto.tfvars
```diff
@@ -38,26 +38,44 @@ nightly_builds = [
...
-xrt_versioned_builds = [
+# Built on push to specific tag.
+versioned_builds = [
+  # Remove libtpu from PyPI builds
   {
-    accelerator    = "tpu"
-    python_version = "3.10"
-    pytorch_git_rev = "v2.1.0"
-    package_version = "2.1.0+xrt"
+    git_tag         = "v2.2.0-rc1"
```
ManfeiBai, line 45: Hi, @zpcore, thanks for the update, not sure how PyTorch rag 2.2 at branch cut day, in 2.1 release, PyTorch's wheel version is `2.1.0-rc0` at the branch cut day; if we want to catch and start testing 2.2 release from branch cut day, do we want to use "v2.2.0-rc0"?

ManfeiBai (approved):
LGTM

infra/tpu-pytorch-releases/artifacts.auto.tfvars
zpcore: Hi @ManfeiBai , I checked the pytorch repo but didn't find tag end with `-rc0`. Can you help double check if we should start from `-rc0`? They may delete it after the release. Thanks,

ManfeiBai: Thanks for confirming this, @zpcore, you are right, I didn't find rc0, we could start from `-rc1` based on 2.1.0 release history

ManfeiBai (approved):
LGTM

infra/tpu-pytorch-releases/artifacts.auto.tfvars
```diff
@@ -108,6 +126,45 @@ versioned_builds = [
...
+    accelerator     = "cuda"
+    cuda_version    = "12.1"
+  },
+  {
+    git_tag         = "v2.2.0-rc1"
+    package_version = "2.2.0-rc1"
+    accelerator     = "cuda"
+    cuda_version    = "11.8"
+    python_version  = "3.10"
+  },
```
JackCaoG, line 153: @vanbasten23 do we need this?

JackCaoG (approved):

PR #5914, opened 2023-11-22
Fix `as_strided` for inputs smaller than the arguments specification.

ysiraichi:
Fix: #5719

This PR introduces a `base_` attribute for `XLATensor`. It keeps track of the tensor whose storage would be aliased by the outer tensor due to a view operation.

```python
a = torch.rand(5, device=xm.xla_device())  # base_ is undefined
b = a + a  # base_ is undefined
c = b[2:]  # base_ is b
d = c.as_strided((5,), (1,), 0)  # uses b (the base_ tensor), instead of c, as input
                                 # base_ is b
```

### DIFF
--- a/torch_xla/csrc/aten_xla_bridge.h
+++ b/torch_xla/csrc/aten_xla_bridge.h
@@ -148,6 +148,13 @@ auto TupleAtenFromXlaTensors(const std::vector<XLATensorPtr>& tensors) {
   return TupleAtenFromXlaTensorsImpl(tensors, std::make_index_sequence<N>{});
 }
 
+// Returns the deepest base tensor for a given tensor.
+// If the base tensor is not defined, returns the tensor itself.
+const at::Tensor& GetRootBase(const at::Tensor& tensor);
+// Sets the base tensor of a given XLATensor. Convenient function
+// to be used when returning tensors.
+XLATensorPtr SetBaseTensor(XLATensorPtr tensor, const at::Tensor& base);
+
 }  // namespace bridge
 }  // namespace torch_xla
 
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -709,13 +709,13 @@ at::Tensor XLANativeFunctions::as_strided_copy(
                                                       storage_offset);
   }
   // Sets the base tensor as tensor.
-  // Even though this function copies (without aliasing) tensor, it's still treated
-  // as a view function in the functionalization layer.
-  return bridge::AtenFromXlaTensor(
-      bridge::SetBaseTensor(
-          tensor_methods::as_strided(
-              self_tensor, std::move(xsize), std::move(xstride), XlaHelpers::I64Optional(storage_offset)),
-          tensor));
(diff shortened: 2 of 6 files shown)


JackCaoG:
lol do you mind resolve the conflict?

ysiraichi:
@JackCaoG I think this is ready for another round of reviews. Could you take a look at it?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -694,7 +694,12 @@ at::Tensor XLANativeFunctions::as_strided_copy(
     const at::Tensor& self, at::IntArrayRef size, at::IntArrayRef stride,
     c10::optional<int64_t> storage_offset) {
   TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
-  XLATensorPtr self_tensor = bridge::GetXlaTensor(self);
+  // Retrieve the base tensor, if there's one.
+  // This function actually operates on the tensor's storage. Since XLA does not
+  // expose the actual storage, we use the originally allocated tensor.
+  const auto& base = bridge::GetXlaTensor(self)->Base();
+  const auto& tensor = base.defined() ? base : self;
```
JackCaoG, line 701: can we explicitly call out the type here lol. I am trying to reduce the auto in out code base to make code reading easier..

torch_xla/csrc/aten_xla_bridge.h
```diff
@@ -148,6 +148,13 @@ auto TupleAtenFromXlaTensors(const std::vector<XLATensorPtr>& tensors) {
   return TupleAtenFromXlaTensorsImpl(tensors, std::make_index_sequence<N>{});
 }
 
+// Returns the deepest base tensor for a given tensor.
+// If the base tensor is not defined, returns the tensor itself.
+const at::Tensor& GetRootBase(const at::Tensor& tensor);
+// Sets the base tensor of a given XLATensor. Convenient function
+// to be used when returning tensors.
+XLATensorPtr SetBaseTensor(XLATensorPtr tensor, const at::Tensor& base);
```
JackCaoG, line 156: do we ever expect base to be on non-xla device? If not can we add an explict check?

torch_xla/csrc/tensor.h
```diff
@@ -337,6 +340,11 @@ class XLATensor : public torch::lazy::LazyTensor {
   // points to the same storage, and thus alias of each other.
   // FIXME(alanwaketan): Remove this once we have functionalization (bdhirsh).
   c10::Storage storage_;
+  // Base tensor for view and view_copy operations. This is used mainly for
+  // operations such as as_strided, which operates on the allocated storage.
+  // Since XLATensor doesn't actually use the storage, we have to run the
```
JackCaoG, line 345: `doesn't actually use the storage` -> `doesn't expose the storage`

torch_xla/csrc/aten_xla_bridge.h
ysiraichi: Good question. I don't think so, since we got to a XLA dispatched kernel. Will add the check.

JackCaoG (approved):

PR #5919, opened 2023-11-22
Don't fallback for pow

qihqi:


JackCaoG:
op info test started to complain, can you take a look?

test/test_core_aten_ops.py
```diff
@@ -3295,6 +3295,29 @@ def test_aten_pow_Tensor_Scalar_2(self):
     kwargs = dict()
     run_export_and_compare(self, torch.ops.aten.pow.Tensor_Scalar, args, kwargs)
 
+  def test_aten_pow_Scalar_1(self):
+    args = (10000,
+            torch.Tensor([
```
JackCaoG, line 3300: any reason not using randn here?

JackCaoG (approved):

test/test_core_aten_ops.py
qihqi: either way works.

PR #5922, opened 2023-11-26
[dynamo] Move CPU tensor constructor nodes to XLA.

ysiraichi:
Fixes: #5722 
Depends on PyTorch PR: https://github.com/pytorch/pytorch/pull/114539

This PR introduces `XLAConstructorMoverPass`: a FX graph pass inherited from PyTorch's `ConstructorMoverPass`. It turns FX nodes that create CPU tensors into nodes that create XLA tensors.

Affected benchmarks: `hf_GPT2`

### DIFF
--- a/test/dynamo/test_bridge.py
+++ b/test/dynamo/test_bridge.py
@@ -222,12 +222,12 @@ class TorchXLAReuseGraphTest(torch._dynamo.test_case.TestCase):
   test_training_maxpool = make_training_test(MaxPoolModule)
   test_training_upsample = make_training_test(UpsampleModule)
 
-  def _compile_and_check(self, thing, args, backend="openxla"):
-    r = thing(*args)
+  def _compile_and_check(self, fn, args, backend="openxla"):
+    r = fn(*args)
     xm.mark_step()
 
-    compiled_thing = torch.compile(backend=backend)(thing)
-    compiled_r = compiled_thing(*args)
+    compiled_fn = torch.compile(backend=backend)(fn)
+    compiled_r = compiled_fn(*args)
     xm.mark_step()
 
     self.assertEqual(r, compiled_r)
--- a/test/dynamo/test_bridge.py
+++ b/test/dynamo/test_bridge.py
@@ -222,6 +222,16 @@ class TorchXLAReuseGraphTest(torch._dynamo.test_case.TestCase):
   test_training_maxpool = make_training_test(MaxPoolModule)
   test_training_upsample = make_training_test(UpsampleModule)
 
+  def _compile_and_check(self, thing, args, backend="openxla"):
+    r = thing(*args)
+    xm.mark_step()
+
+    compiled_thing = torch.compile(backend=backend)(thing)
+    compiled_r = compiled_thing(*args)
+    xm.mark_step()
+
+    self.assertEqual(r, compiled_r)
+
   def test_non_tensor_args_for_partition(self):
 
     class Emb(torch.nn.Embedding):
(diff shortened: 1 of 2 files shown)


ysiraichi:
This PR will only work with https://github.com/pytorch/pytorch/pull/114539 changes. Waiting for it to land.

JackCaoG:
Thanks, can you give me a ping when upstream pr merged?

ysiraichi:
@JackCaoG The PRs have landed! Is there a way to specify what's the PyTorch minimum version for this PR?

ysiraichi:
@JackCaoG done! I think this PR is good for another round of reviews. Could you take a look at it?

JackCaoG:
will do on Monday

torch_xla/core/dynamo_bridge.py
```diff
@@ -454,6 +458,26 @@ def call_module(self, target, args, kwargs):
     return super().call_module(target, args, kwargs)
 
 
+class XLAConstructorMoverPass(ConstructorMoverPass):
+
+  def __init__(self):
+    super().__init__("xla")
+
+  def allow_cpu_device(self, node: torch.fx.Node):
```
JackCaoG, line 466: can you add a comment here explain what will happen if this fn returns False?

test/dynamo/test_bridge.py
```diff
@@ -233,23 +243,28 @@ def __init__(self):
...
     x = x.unsqueeze(dim=-1)
-    foo(x)
+    self._compile_and_check(foo, (x,))
+
+  def test_factory_copy(self):
+
+    def foo(device):
+      return torch.arange(5, device="cpu").to(device)
+
+    self._compile_and_check(foo, (xm.xla_device(),))
```
JackCaoG, line 267: will this case fail without the `XLAConstructorMoverPass`?

test/dynamo/test_bridge.py
```diff
@@ -222,6 +222,16 @@ class TorchXLAReuseGraphTest(torch._dynamo.test_case.TestCase):
   test_training_maxpool = make_training_test(MaxPoolModule)
   test_training_upsample = make_training_test(UpsampleModule)
 
+  def _compile_and_check(self, thing, args, backend="openxla"):
```
JackCaoG, line 225: nit, `thing` -> `fn`, `compiled_thing` -> `compiled_fn`

torch_xla/core/dynamo_bridge.py
```diff
@@ -464,6 +488,9 @@ def extract_compiled_graph(xla_model: torch.fx.GraphModule, xla_args):
   # This call is critical to make sure xla_args' tensor id show up in graph_input_tensor_ids
   xm.mark_step()
 
+  # Move tensor constructor nodes that do not target XLA devices.
```
JackCaoG, line 491: I am a bit confuse by `move`, where does it move to? Does it auto convert the cpu tensor to xla tensor?

test/dynamo/test_bridge.py
ysiraichi: Yes. The partitioner will split that function into 2: (i) arange; and (ii) _to_copy. The first one will be returning a CPU tensor, which will fail.

torch_xla/core/dynamo_bridge.py
ysiraichi: It replaces `device=cpu` by `device=xla` (1) in the tensor constructor nodes it finds (2).

(1) uses an XLA device being used in the program
(2) checks whether it's "movable", first.

A tensor constructor node is movable iff all of its uses (recursively) don't reach an un-movable node (see [`cannot_be_moved` function](https://github.com/pytorch/pytorch/blob/5de0dff7ea2f0857f2682ce4c1e6ad70779a76a6/torch/_inductor/fx_passes/post_grad.py#L1027)).

ysiraichi:
@JackCaoG thank you for your reviews. I have addressed them and updated the PR. Could you take a look at it, again?

JackCaoG:
Thanks, let me take a look today

JackCaoG (approved):

PR #5924, opened 2023-11-27
Add benchmark noise reducing info.

golechwierowicz:
Add info about knobs making benchmarks more stable across different runs.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -5,6 +5,27 @@ The two main benchmarking scripts are
   - `result_analyzer.py` to aggregate the benchmark result in CSV form.
 
 
+## Reducing benchmark noise 
+
+It is important to keep the benchmark runs safe from external effects 
+to reduce noise. Run:
+
+```
+# Sets the CPU statically to the highest tuneable frequency. 
+# Prevent energy saving features to kick in.
+sudo cpupower frequency-set --governor performance
+
+# Lock GPU clocks to lower frequency to reduce the chance of extra throttling. Choose
+# FREQ based on your GPU info. For example A100 operates on 765MHz (up to 1410 MHz),
+# with memory operating on 1215MHz. Setting the clock a couple hundrend MHz below
+# will most likely prevent thermal effects.
+FREQ=...
+nvidia-smi --lock-gpu-clocks=$FREQ,$FREQ
+
+# Disable autoboost selecting clock rate based on thermal, and power budget effects.
+CUDA_AUTO_BOOST=0
+```
+
 ## Experiment runner
 
 Run the `experiment_runner.py` from the `pytorch` directory, which should be the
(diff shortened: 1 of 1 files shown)


benchmarks/README.md
```diff
@@ -5,6 +5,27 @@ The two main benchmarking scripts are
...
+It is important to keep the benchmark runs safe from external effects 
+to reduce noise. Run:
+
+```
+# Sets the CPU statically to the highest tuneable frequency. 
+# Prevent energy saving features to kick in.
+sudo cpupower frequency-set --governor performance
+
+# Lock GPU clocks to lower frequency to reduce the chance of extra throttling. Choose
+# FREQ based on your GPU info. For example A100 operates on 765MHz (up to 1410 MHz),
```
vanbasten23, line 19: Could you add the link to help users find the 765MHz and 1215MHz?
Also, regarding `a couple hundrend MHz` below, can it be more quantifiable, for example 90%, 80% etc.?

golechwierowicz: It's hard to tell, personally I set both core clock, and mem clock to 1005MHz for A100. So like 80% of the max clock, but I can imagine it differs across GPUs of different architectures. Sometimes you can even set it to 90% so the device time does not look out of the blue.

golechwierowicz: I added a more general procedure of finding out the frequency. PTAL.

benchmarks/README.md
```diff
@@ -5,6 +5,28 @@ The two main benchmarking scripts are
   - `result_analyzer.py` to aggregate the benchmark result in CSV form.
 
 
+## Reducing benchmark noise 
+
+It is important to keep the benchmark runs safe from external effects 
+to reduce noise. Run:
+
+```
+# Sets the CPU statically to the highest tuneable frequency.
```
frgossen, line 14: nit: Maybe move the comments out of the code sections.

frgossen (approved):
Thanks!

zpcore:
Quick questions:
- Is there a recommendation/consensus GPU freq we can use throughout different model testing so that we can compare the results under the same constraint in the same platform? 
- Also want to check if the GPU freq can be set arbitrarily as long as it is below the maximum or there is only limited options we can choose from.

benchmarks/README.md
golechwierowicz: Done.

golechwierowicz:
> Quick questions:
> 
> * Is there a recommendation/consensus GPU freq we can use throughout different model testing so that we can compare the results under the same constraint in the same platform?
> * Also want to check if the GPU freq can be set arbitrarily as long as it is below the maximum or there is only limited options we can choose from.

1. I don't think there is a consensus among the team, it will be good to establish it now. (given everyone runs benchmarks manually, once we have an automated suite it will be unnecessary though)
2. AFAIK it has to below the maximum, above the minimum, and it has to be the multiple of 15. But the entire point is setting it to such a value so that thermal throttling does not kick in, and there is no random slowdown on computation. So setting it to ~80% of the maximum seems like a good idea.

PR #5929, opened 2023-11-28
Update troubleshooting doc with few common env var combination

JackCaoG:
Remove some of the env var that was rarely used in the doc and add some useful debugging tips.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -259,6 +259,10 @@ only be enabled for debugging.
 * ```XLA_SYNC_WAIT```: Forces the XLA tensor sync operation to wait for its completion, before
   moving to the next step.
 
+* ```XLA_USE_EAGER_DEBUG_MODE```: Forces the XLA tensor to execute eagerly, meaning compile and execute the torch operations one
+  by one. This is useful to bypass the long compilation time but overall step time will be a lot slower and memory usage will be higher
+  since all compiler optimizaiton will be skipped.
+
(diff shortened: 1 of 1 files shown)


JackCaoG:
@will-cromar can I get a review for this doc change?

TROUBLESHOOTING.md
```diff
@@ -308,3 +296,19 @@ only be enabled for debugging.
 * ```XLA_DUMP_HLO_GRAPH```: If set to `=1` in case of a compilation or execution error the
   offending HLO graph will be dumped as part of the runtime error raised by `xla_util.cc`.
 
+### Common Debugging Environment Variables Combinations
+
+* Record the graph execution in the IR format
+  ```
+  XLA_SAVE_TENSORS_FMT="hlo" XLA_SAVE_TENSORS_FILE="/tmp/save1.hlo"
```
JackCaoG, line 303: maybe I should also add `XLA_IR_DEBUG=1 and XLA_HLO_DEBUG=1` here..

will-cromar, line 303: Can we make this setting less obscure? I pretty much always have to look these up.

It seems kind of silly to have to set 2 environment variables in each case. We can either check the file extension `.hlo` vs `.ir`, or have two different env vars. Not to mention, the name `XLA_SAVE_TENSORS_FILE` sounds like it's dumping tensor contents, which only makes sense if you understand lazy tensor

JackCaoG: not sure if I will have time to fix this before branch cut, if not I prefer to get this one in. Some instruction is better than no instruction. I was very confuse on how to actually dump HLO and IR when I started at this project.

PR #5930, opened 2023-11-28
Add Pattern Boundary Marking API

lsy323:
This is the first version of the Pattern Boundary Marking API, the user interface, functionality and the algorithm used in the implementation will be improved in the following PRs.

co-authored with @chunnienc, who authored the MLIR pass that groups MLIR graph with xla marker ops outlining the patterns in the graph, into stablehlo composites.

Added op
```
def mark_tensor(x: torch.Tensor,
                name: str,
                pos: int,
                id: int,
                is_input: bool,
                attr: Dict = None):
  """Attach pattern boundary metadata to a XLA Tensor.
  
  Args:
      x: torch.Tensor (On XLA device) - the marked tensor.
      name: str - The name of the pattern, it will be the name of the stablehlo composite op.
      pos: int - Input/output Position of the annotated tensor in the pattern.
      id: int - Unique identifier of the pattern instance.
      is_input: bool - If the annotated tensor is the input to the pattern.
      attr: dict - Attribute of the pattern, it will be passed down to the attribute field
                   in the stablehlo composite.
  """
```

Example Usage:
```
def f(x):
      x = x + 1
      x = torch.ops.xla_pattern_marking.mark_tensor(x, name="p", pos=0, id=0, is_input=True)
      x = x + 2
      x = torch.ops.xla_pattern_marking.mark_tensor(x, name="p", pos=0, id=0, is_input=False, attr={"attr1":"val1"})
      return x
```
Generated StableHLO:
```
module @IrToHlo.15 attributes {mhlo.cross_program_prefetches = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  func.func @main(%arg0: tensor<f32>, %arg1: tensor<5xf32>) -> tensor<5xf32> {
    %0 = stablehlo.constant dense<1.000000e+00> : tensor<5xf32>
    %1 = stablehlo.add %arg1, %0 : tensor<5xf32>
    %2 = stablehlo.custom_call @stablehlo.composite(%1, %arg0) {called_computations = [@p__0.impl], composite.backend_config = {attributes = {attr1 = "val1"}, name = "p"}} : (tensor<5xf32>, tensor<f32>) -> tensor<5xf32>
    return %2 : tensor<5xf32>
  }
  func.func private @p__0.impl(%arg0: tensor<5xf32>, %arg1: tensor<f32>) -> tensor<5xf32> {
    %0 = stablehlo.broadcast_in_dim %arg1, dims = [] : (tensor<f32>) -> tensor<5xf32>
    %1 = stablehlo.add %arg0, %0 : tensor<5xf32>
    return %1 : tensor<5xf32>
  }
}
```

This PR includes the following changes:
1. Add a custom torch op `torch.ops.xla_pattern_marking.mark_tensor`, allowing backend SDK/compiler to outline the boundary of a pattern subgraph by applying boundary metadata annotation to the input/output of the pattern subgraph.
2. The custom op will be lowered to a HLO/StableHLO custom call to `xla_mark_tensor`.
3. An MLIR pass will traverse from the output(s) of each pattern subgraph, the traversal ends when the input markers are reached. The traversed vertices will be grouped into a StableHLO composite call.
4. Introduce a new dependency used for JSON deserialization in C++. It's also a submodule in PyTorch https://github.com/pytorch/pytorch/blob/8933ff35953908545cb410d736a6c4a3a7761c54/.gitmodules#L140

Limitation and future work:
1. The MLIR pass will be extended to support nested patterns.
2. The MLIR pass will be extended to support patterns with multiple outputs.
3. The API interface will be improved based on feedbacks from SDK developers.

cc @GleasonK @miladm


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -648,6 +648,11 @@ void MapXlaEnvVarsToLazy() {
       runtime::sys_util::GetEnvInt("XLA_TRIM_GRAPH_SIZE", 100000);
 }
 
+at::Tensor MarkTensor(const at::Tensor& input, const std::string& info) {
+  auto result = tensor_methods::mark_tensor(bridge::GetXlaTensor(input), info);
```
JackCaoG, line 652: lol call me anti-auto, but can we explicitly spell out he basic types? It will make code reading a lot easier.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,391 @@
...
+  static std::unique_ptr<BoundaryMetadata> Parse(llvm::StringRef str) {
+    auto j = json::parse(str, /*cb=*/nullptr, /*allow_exceptions=*/false);
+    return Build(j);
+  }
+
+ private:
+  static std::unique_ptr<BoundaryMetadata> Build(
+      const nlohmann::basic_json<>& j) {
+    BoundaryMetadata metadata;
+#define FIELD(key, value_type)                           \
```
JackCaoG, line 66: I would avoid macro is possible since it makes compilation error hard to read, is there an easy way to just make it a helper function?

torch_xla/experimental/xla_marker.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@impl(xla_pattern_marking_lib, "mark_tensor", "CompositeExplicitAutograd")
+def mark_tensor(x: torch.Tensor,
+                name: str,
+                pos: int,
+                id: int,
+                is_input: bool,
+                attr: Dict = None):
+  # Do nothing for non-xla tensor.
```
JackCaoG, line 64: do we want this or explicitly error out when calling `mark_tensor` on a non-xla tensor?

JackCaoG (commented):
I didn't review `torch_xla/csrc/runtime/stablehlo_composite_helper.cc` too carefully but otherwise lgtm

torch_xla/csrc/init_python_bindings.cpp
lsy323: Makes sense, let me explicitly specify the type

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: Done. PTAL.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+      }
+
+      for (auto [use_op, operand_number] : uses) {
+        use_op->setOperand(operand_number, original_value);
+      }
+      ops_to_erase.push_back(mark_tensor_op.getOperation());
+    }
+    for (auto* op : ops_to_erase) {
+      op->erase();
+    }
```
GleasonK, line 379: Can we add code comments? I'm not positive what this is doing.. Is this just replacing the uses of the custom call result with its operand? If so something like this should manage the intermediate logic:

```
for (auto op : func_op.getOps<mlir::stablehlo::CustomCallOp>()) {
  if (!IsXlaMarkTensorOp(op.getOperation())) continue;
  for (auto i = 0; i < op.getNumResults(); ++i) {
    Value result = op.getResult(i);
    result.replaceAllUsesWith(op.getOperand(i));
  }
}
```

Note this is coded for the pairwise input-output if we end up going with a design that permits specifying multiple inputs to an abstraction at once.

GleasonK: Also instead of explicitly calling erase we can let DCE during canonicalization take care of that, or just call erase inline - we likely don't need to keep a vector of deleted nodes just to delete later?

Looks like we already have a canonicalize pass that runs with this below.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+
+  void runOnOperation() override {
+    mlir::ModuleOp module_op = getOperation();
+    llvm::SmallVector<mlir::func::FuncOp> func_ops(
+        module_op.getOps<mlir::func::FuncOp>());
+    for (mlir::func::FuncOp& func_op : func_ops) {
+      llvm::DenseMap<const mlir::Operation*, size_t> op_line_num =
+          BuildOperationsLineNumberMap(func_op);
+      for (auto op : func_op.getOps<mlir::stablehlo::CustomCallOp>()) {
+        BuildStableHLOCompositeOp(op.getOperation(), op_line_num);
```
GleasonK, line 119: `if IsXlaMarkTensorOp(op.getOperation)`?

When should this method be called? Only on outputs and only for MarkTensor ops?

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+                                 curr_metadata.pos});
+          continue;
+        }
+      }
+
+      scope_ops_setvec.insert(curr_op);
+      for (mlir::Value value : curr_op->getOperands()) {
+        mlir::Operation* def_op = value.getDefiningOp();
+        if (def_op == nullptr) {
+          // Terminal condition: Global function arg
```
GleasonK, line 230: Should this ever happen? This feels like an error since everything should eventually come through a `mark_begin`?

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+          arg_pos_setvec.insert({value, std::numeric_limits<int64_t>::max()});
+        } else if (llvm::isa<mlir::stablehlo::ConstantOp>(op)) {
+          // Terminal condition: constant
```
GleasonK, line 233: Same question, should constants go through a `mark_begin` call?

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+        } else {
+          processing.push_back(def_op);
+        }
+      }
+    }
+    // Sorts all ops within the boundary by their line numbers in the input
+    // MLIR. The ops will be duplicated to the impl function following this
+    // order.
+    auto scope_ops = scope_ops_setvec.takeVector();
+    for (auto& op : scope_ops) {
```
GleasonK, line 244: This sounds fragile.. is it needed for correctness? I.e. a program could be all be a single line? What would break if it were?

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+          break;
+        default:
+          // Ignored unrecognizable attr json
+          break;
+      }
+    }
+    return builder.getDictionaryAttr(named_attrs);
+  }
+
+  void BuildStableHLOCompositeOp(
```
GleasonK, line 188: The cyclomatic complexity of this function is pretty high, for loops nested in while loops, nested ifs in loops etc. Would be good to break down into smaller functions. Would help with understanding what exactly is happening as well.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+    builder.create<mlir::func::ReturnOp>(impl_func.getBody().getLoc(),
+                                         mapping.lookup(op)->getResults());
+
+    // Adds the new function to symbol table.
+    mlir::SymbolTable symbol_table(module_op);
+    impl_func.setPrivate();
+    symbol_table.insert(impl_func);
+
+    builder.setInsertionPointAfter(op);
+    llvm::SmallVector<mlir::NamedAttribute> call_attrs{
```
GleasonK, line 302: Would suggest making this a separate method: `buildCompositeAsCustomCall`. Will make future refactoring to composite much easier.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+        builder.create<mlir::stablehlo::CustomCallOp>(
+            op->getLoc(), impl_func.getFunctionType().getResults(), args,
+            call_attrs);
+
+    // Updates all users of this op's result(s) to use the results(s) of impl
+    // func call.
+    for (size_t i = 0; i < op->getNumResults(); ++i) {
+      mlir::OpResult result = op->getResult(i);
+      mlir::OpResult new_result = composite_call_op->getResult(i);
+      for (mlir::OpOperand& use : result.getUses()) {
```
GleasonK, line 338: Similar to above, utility methods are probably safer:

```
Value result = op->getResult(i);
result.replaceAllUsesWith(composite_call_op->getResult(i));
```

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,404 @@
...
+    // boundary in the function.
+    llvm::SmallVector<mlir::Location> arg_locs;
+    llvm::SmallVector<mlir::Type> arg_types,
+        result_types(op->getResultTypes().begin(), op->getResultTypes().end());
+    for (auto& arg : args) {
+      arg_types.push_back(arg.getType());
+      arg_locs.push_back(arg.getLoc());
+    }
+
+    mlir::func::FuncOp impl_func = builder.create<mlir::func::FuncOp>(
```
GleasonK, line 279: This portion would make a good separate function:

```
FuncOp buildImplementationFuncOp(...);
```

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
GleasonK: I think that not having _everything_ go through `mark` calls puts us at risk of corrupt abstractions. I.e. what does the following test case do?

```
x = mark_begin(x, "a")
x = x + 1
y = mark_begin(y, "b")
z = x + y
z = mark_end(z, "a")
z = z*2
z = mark_end(z, "b")
```

This has overlapping regions between separate abstractions.. I _think_ that one of the calls to begin would be consumed when creating an abstraction for `"a"`, and the `"b"` call may still be successful since it will trace back to an input argument or constant?

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
GleasonK: Can we write this in a way where this method should be expected to succeed and give an error message if it fails.

Currently it fails silently (if it isn't the right custom call target, if JSON parsing fails, etc). Consider making stronger use of `FailureOr<T>` which can return a failure message or a type on success.

test/stablehlo/test_mark_pattern.py
```diff
@@ -0,0 +1,144 @@
...
+    else:
+      out = [out]
+    stablehlo = xm.get_stablehlo(out)
+    return stablehlo
+
+  def test_basic(self):
+
+    def f(x):
+      x = x + 1
+      x = torch.ops.xla_pattern_marking.mark_tensor(x, "p", 0, 0, True)
```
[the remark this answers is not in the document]
GleasonK: 4. In an ideal world we wouldn't need `id` either. The algorithm could be: Find the first `mark_end` custom_call and traverse the use-def chain upwards until we hit the first `mark_begin` with matching name. This would handle lexical scoping of nested abstractions without the need for IDs

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: Because the StableHLO is lowered from functionalized FX graph, there can be some constants lifted to be MLIR inputs.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: Yes. Otherwise it'd be very hard to determine the insertion order.

It's actually not the "line number" in the input text but the order when iterating through ops in the function in MLIR. So it's guaranteed that each op will have a unique "line number"

GleasonK: Ah, that makes more sense. I'll dig around if theres a more convenient way to do this in MLIR. Can we use a different name than `line_number`? `topological_position`, `encounter_position`, etc?

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: Yes you're right. I'm not aware of the `replaceAllUsesWith` APIs.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: > When should this method be called? Only on outputs and only for MarkTensor ops?
Yes. There is a check right after `GetBoundaryMetadata` in the function.

> Can we write this in a way where this method should be expected to succeed and give an error message if it fails.
This is intended as the composite is optional and the invalid mark_tensor custom_call can still be removed afterwards. Do you suggest to fail the conversion if the composite is not complete?

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: `mark_begin` only accepts tensors, while the lowering can turn Python scalars that into StableHLO constant tensors.

The example is invalid: we have no plans to accept overlapped patterns. The behavior would be undefined or even wrong.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
GleasonK: Ah, yeah. Thinking through it probably better to allow input arguments instead of require that all constants be marked. The alternative IIUC is:

```py
def f():
  x = 3
  x = mark_begin(x, "a")
#  -functionalize->
def f(x):
  x = mark_begin(x, "a")
```

But there are going to be constants used everywhere in a graph and to mark them all would make for some ugly code. Feel free to resolve this comment if my understanding is correct.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: The frontend in ODML SDK will make sure no overlapped patterns are marked in the long term.

chunnienc:
> Gave a first passthrough. If we can make the build composite function simpler to review that would be great, it is a bit of a monolith currently

SG. I will break down the build composite function later.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
GleasonK: Personal preference is yes - From a functional perspective, if a boundary is marked, it is likely intentional. If we must export with a boundary, it is probably better to know during export time as opposed to much later in the pipelines, whether it manifests via poor performance on some hardware, failed pattern match later in compilation, etc.

From a usability perspective - If we silently fail, what is the next step to diagnosing an issue? If we cannot honor that boundary, we should at a minimum give the reason why.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
GleasonK: > mark_begin only accepts pytorch tensors, while the torch_xla

Would be good to settle on one thing, but even so take that suggestion as pseudo code, a hypothetical `torch_xla.mark_begin`.

> The behavior would be undefined or even wrong.

I guess my proposal is that the behavior should be "error"

test/stablehlo/test_mark_pattern.py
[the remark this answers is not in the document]
qihqi: I like these suggestions. To be clear, we can still keep the one MarkTensor Lazy IR node if we want; but the exposed Python API can be mark_begin / mark_end

test/stablehlo/test_mark_pattern.py
```diff
@@ -0,0 +1,144 @@
...
+import torch_xla.experimental.xla_marker
+from torch.utils import _pytree as pytree
+
+
+class XlaMarkPatternTest(unittest.TestCase):
+
+  def run_func_get_stablehlo(self, f, input_args):
+
+    device = xm.xla_device()
+    input_args = pytree.tree_map_only(torch.Tensor,
```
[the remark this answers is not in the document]
lsy323: Yes, using [pytree](torch/utils/_pytree.py) would simplify the logic in ([ref](https://github.com/pytorch/xla/blob/a3b0c6ec5dc299370773a016e787dc5a13dfab9a/benchmarks/util.py#L82C53-L82C53))

test/stablehlo/test_mark_pattern.py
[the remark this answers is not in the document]
chunnienc: > Could remove the need for tracking position via `pos=`

I think it is okay to keep the low level API here to be raw. There can be some tricky requirements from different HLFB frontends e.g. pattern matching requires inserting a additional mark tensor node to the FX graph for each input and output [ref](https://github.com/arfaian/odml-sdk/blob/2f08c60d1d33397614de4edff60eb45cce9b2a99/odml_sdk/experimental/torch/hlfb/mark_pattern.py#L180).  ODML SDK will definitely build additional HLFB layers to wrap the torch_xla low level APIs for different frontends and safety guards.


> In an ideal world we wouldn't need id either

I'd lean toward keeping the id to guarantee the correctness of composites from MLIR pass. TBH I also want an additional field `total_inputs` and `total_outputs` to make sure all the boundary IO are correctly identified during the graph traversal.

torch_xla/csrc/init_python_bindings.cpp
lsy323: Done

torch_xla/experimental/mark_pattern_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+class StableHLOCompositeBuilder:
+
+  def __init__(self,
+               name: str,
+               attrs: Dict[str, Union[int, float, str]] = None):
+
+    self.attrs = attrs
+    self.name = str(name)
+    self.id = uuid.uuid1().hex
```
qihqi, line 20: lets use uuid4

torch_xla/experimental/mark_pattern_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class StableHLOCompositeBuilder:
+
+  def __init__(self,
+               name: str,
+               attrs: Dict[str, Union[int, float, str]] = None):
+
+    self.attrs = attrs
+    self.name = str(name)
```
qihqi, line 19: why copy the name? str is immutable anyways

torch_xla/experimental/mark_pattern_utils.py
```diff
@@ -0,0 +1,86 @@
...
+    self.id = uuid.uuid1().hex
+    self._inputs = {}
```
qihqi, line 21: should these be lists?

torch_xla/experimental/mark_pattern_utils.py
```diff
@@ -0,0 +1,86 @@
...
+  def uid(self):
+    return f"{self.name}__{self.id}"
+
+  def get_input(pos: int) -> torch.Tensor:
+    return self._inputs[pos]
+
+  def get_output(pos: int) -> torch.Tensor:
+    return self._outputs[pos]
+
+  def mark_input(self, x: torch.Tensor, pos: int) -> torch.Tensor:
```
qihqi, line 34: just mark_inputs should be sufficient?

chunnienc:
@GleasonK I refactored the MLIR pass and made it signal failures when running into errors. PTAL

torch_xla/experimental/mark_pattern_utils.py
lsy323: @qihqi Refactored the entire CompositeBuilder, PTAL, thanks!

lsy323:
> Do we need this pr in 2.2 release?

@JackCaoG Not necessarily to to have it in 2.2 release. If we can make it, we can add to the release note, it will be experimental anyway.

torch_xla/experimental/mark_pattern_utils.py
lsy323: refactored the impl

torch_xla/experimental/mark_pattern_utils.py
lsy323: refactored the impl

torch_xla/experimental/mark_pattern_utils.py
lsy323: refactored the impl

lsy323:
CI failed, it should pass after rebasing to head

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: Renamed the lut to `op_order_map`

qihqi (approved):

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2061,6 +2067,22 @@ void InitXlaModuleBindings(py::module m) {
...
+          TORCH_LAZY_COUNTER("XlaMarkTensor", 1);
+          at::Tensor result;
+          {
+            NoGilSection nogil;
+            result = MarkTensor(input, info);
+          }
+          return result;
+        });
+
+  m.def("_xla_mark_dynamic", [](const at::Tensor& input, uint32_t dim) {
```
GleasonK, line 2081: Looks like this got copied accidentally?

torch_xla/csrc/ops/mark_tensor.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+torch::lazy::NodePtr MarkTensor::Clone(torch::lazy::OpList operands) const {
+  return torch::lazy::MakeNode<MarkTensor>(operands.at(0), info_);
+}
+
+XlaOpVector MarkTensor::Lower(LoweringContext* loctx) const {
+  xla::XlaOp input = loctx->GetOutputOp(operand(0));
+  xla::Shape input_shape = ShapeHelper::ShapeOfXlaOp(input);
+  static const std::string opname = "xla_mark_tensor";
+  xla::XlaOp output = xla::CustomCall(
```
GleasonK, line 24: Looking at xla_builder.h, I think everything after shape is optional, and everything after `info_` is default I believe?

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,465 @@
...
+#include "stablehlo/dialect/StablehloOps.h"
+
+namespace torch_xla {
+namespace runtime {
+
+namespace {
+
+using nlohmann::json;
+
+static bool IsXlaMarkTensorOp(mlir::Operation* op) {
```
GleasonK, line 22: Something we've found useful for CustomCall operations is using an adaptor. This does not have to happen now, just wanted to make you aware of this since it makes verification / accessors much easier:
https://github.com/tensorflow/mlir-hlo/blob/753fda7c2a79732eaa1f6eb076b1ac90c72441de/stablehlo/stablehlo/experimental/dialect/StablehloOps.h#L200

Then you have getters which take a custom call and return an optional adaptor if the CC is well-formed:
https://github.com/tensorflow/mlir-hlo/blob/753fda7c2a79732eaa1f6eb076b1ac90c72441de/stablehlo/stablehlo/experimental/dialect/StablehloOps.h#L224

Then you can write methods that take in this MarkTensorOp, and you don't need to verify the custom call everywhere

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -0,0 +1,465 @@
...
+
+  mlir::FailureOr<std::unique_ptr<BoundaryMetadata>> GetBoundaryMetadata(
+      mlir::Operation* op) {
+    if (!IsXlaMarkTensorOp(op)) {
+      return mlir::FailureOr(nullptr);
+    }
+    auto backend_config =
+        op->getAttr("backend_config").dyn_cast<mlir::StringAttr>();
+    if (backend_config == nullptr) {
+      return mlir::FailureOr(nullptr);
```
GleasonK, line 155: Silent pass failures tend to be difficult to track down. Would recommend using `op->emitError()` for each of these failure cases. Usually we implement failure messages at the "leafs" (i.e. this error is propagated up, and if `GetBoundaryMetadata` fails, the callsite also returns failure, but that one can be silent).

```
return op->emitError() << "expected mark tensor operation";
...
return op->emitError() << "expected mark tensor op to have backend_config";
```

GleasonK (approved):
A few minor nits, feel free to implement any of this in a followup.

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: This is not a failure though. `unique_ptr` here is used as an alternative of `optional`. So the semantic of `FailureOr(nullptr)` is "the operation succeeded, but got no boundary metadata".

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
chunnienc: Sounds good, I like this pattern. I created a separate bug to track the progress of this refactor.

torch_xla/csrc/init_python_bindings.cpp
lsy323: Thanks for catching it! Removed

torch_xla/csrc/ops/mark_tensor.cpp
lsy323: Yeah, let me remove the default args.

PR #5624, opened 2023-09-20
Add all-gather/reduce-scatter coalescee for FSDP.

jeffhataws:
This PR adds all-gather and reduce-scatter coalescence support and use that in FSDP. Also it enables using reduce-scatter's scale param in FSDP. This PR is to be used in conjunction with https://github.com/openxla/xla/pull/5740 .

This is a revival of https://github.com/pytorch/xla/pull/4145 . Will need to address the comments.

PR #4145, opened 2022-11-01
Add all-gather and reduce-scatter coalescence support and use that in…

hjm-aws:
… FSDP. Allow using reduce-scatter's scale param in FSDP.

This PR depends on tensorflow PR https://github.com/tensorflow/tensorflow/pull/58377

PR #5938, opened 2023-11-29
Add reduce-scatter coalescing for FSDP/ZeRO1

jeffhataws:
(Replaced by https://github.com/pytorch/xla/pull/5956)

This PR adds reduce-scatter coalescence support and use that in FSDP/ZeRO1. This also enables using reduce-scatter's scale param in FSD.. This PR is companion to https://github.com/pytorch/xla/pull/5624 and to be used in conjunction with openxla https://github.com/openxla/xla/pull/5740 .

This is a revival of https://github.com/pytorch/xla/pull/4145 . Will need to address the comments.

### DIFF
--- a/test/test_torch_distributed_xla_backend.py
+++ b/test/test_torch_distributed_xla_backend.py
@@ -2,7 +2,7 @@ import contextlib
 import functools
 import os
 import re
-from unittest import mock
+from unittest import mock, skipIf
 
 from absl.testing import absltest, parameterized
 import torch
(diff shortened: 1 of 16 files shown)


JackCaoG:
Build seems to fail with oom, let's use a larger machine. can you modify 

https://github.com/pytorch/xla/blob/master/.github/workflows/build_and_test.yml#L22

and add 
```
runner: linux.12xlarge
```
 I think we used larger machine to run the test, but use default machine to build.

PR #5946, opened 2023-11-29
Fix bad error message when `PjRtComputationClient` throws exception

will-cromar:
The `g_computation_client_initialized` doesn't get reset in this case, leading to incorrect behavior in `GetComputationClientIfInitialized`. Make `CreateClient` anonymous so we can't call it twice by accident.

Example from @jonb377:

```
ptxla@t1v-n-f494979e-w-0:/workspaces/work/pytorch/xla$ python -c 'import torch_xla; torch_xla._XLAC._xla_get_runtime_devices()'
RuntimeError: torch_xla/csrc/runtime/runtime.cc:27 : $PJRT_DEVICE is not set.

Error in atexit._run_exitfuncs:
  File "/workspaces/work/pytorch/xla/torch_xla/__init__.py", line 127, in _prepare_to_exit
    _XLAC._prepare_to_exit()
RuntimeError: torch_xla/csrc/runtime/runtime.cc:17 : Check failed: !was_initialized 
*** Begin stack trace ***

        torch_xla::runtime::GetComputationClient()


        PyCFunction_Call
        _PyObject_MakeTpCall
        _PyEval_EvalFrameDefault
        _PyFunction_Vectorcall
        PyVectorcall_Call

        Py_FinalizeEx
        Py_RunMain
        Py_BytesMain
        __libc_start_main
        _start
*** End stack trace ***
ComputationClient already initialized
```

`ComputationClient` was never initialized, so it doesn't make sense to print `ComputationClient already initialized`.

### DIFF
--- a/torch_xla/csrc/runtime/runtime.cc
+++ b/torch_xla/csrc/runtime/runtime.cc
@@ -30,7 +30,7 @@ std::unique_ptr<ComputationClient> CreateClient() {
 
 }  // namespace
 
-ComputationClient* GetComputationClient(bool create = true) {
+ComputationClient* GetComputationClient(bool create) {
   static std::unique_ptr<ComputationClient> client = nullptr;
   if (!client && create) {
     static std::once_flag flag;
--- a/torch_xla/csrc/runtime/runtime.cc
+++ b/torch_xla/csrc/runtime/runtime.cc
@@ -27,7 +27,7 @@ ComputationClient* GetComputationClient() {
 
     XLA_CHECK(client);
 
-    g_computation_client_initialized.exchange(true);
+    g_computation_client_initialized = true;
     return client;
   }();
 
--- a/torch_xla/csrc/runtime/runtime.cc
+++ b/torch_xla/csrc/runtime/runtime.cc
@@ -10,38 +10,37 @@ namespace torch_xla {
 namespace runtime {
 namespace {
 
-std::atomic<bool> g_computation_client_initialized(false);
-
-ComputationClient* CreateClient() {
-  bool was_initialized = g_computation_client_initialized.exchange(true);
-  XLA_CHECK(!was_initialized) << "ComputationClient already initialized";
+std::unique_ptr<ComputationClient> CreateClient() {
   if (sys_util::GetEnvBool("XLA_DUMP_FATAL_STACK", false)) {
     tsl::testing::InstallStacktraceHandler();
   }
 
-  ComputationClient* client;
+  std::unique_ptr<ComputationClient> client = nullptr;
 
   if (sys_util::GetEnvString(env::kEnvPjRtDevice, "") != "") {
-    client = new PjRtComputationClient();
+    client = std::make_unique<PjRtComputationClient>();
   } else {
-    g_computation_client_initialized = false;
     XLA_ERROR() << "$PJRT_DEVICE is not set." << std::endl;
   }
 
-  XLA_CHECK(client != nullptr);
+  XLA_CHECK(client);
 
   return client;
 }
 
 }  // namespace
 
-ComputationClient* GetComputationClient() {
-  static auto client = std::unique_ptr<ComputationClient>(CreateClient());
+ComputationClient* GetComputationClient(bool create = true) {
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/runtime/runtime.cc
```diff
@@ -10,38 +10,37 @@ namespace torch_xla {
...
+  XLA_CHECK(client);
 
   return client;
 }
 
 }  // namespace
 
-ComputationClient* GetComputationClient() {
-  static auto client = std::unique_ptr<ComputationClient>(CreateClient());
+ComputationClient* GetComputationClient(bool create = true) {
```
JackCaoG, line 33: I think you only need default variable in `.h` file

JackCaoG:
what would the error message be after your change?

torch_xla/csrc/runtime/runtime.cc
```diff
@@ -10,38 +10,37 @@ namespace torch_xla {
...
   }
 
-  ComputationClient* client;
+  std::unique_ptr<ComputationClient> client = nullptr;
 
   if (sys_util::GetEnvString(env::kEnvPjRtDevice, "") != "") {
-    client = new PjRtComputationClient();
+    client = std::make_unique<PjRtComputationClient>();
   } else {
-    g_computation_client_initialized = false;
```
jonb377, line 27: Does removing the atomic bool introduce a race condition here, or does the `std::call_once` cover that?

jonb377 (commented):
Thanks Will!

torch_xla/csrc/runtime/runtime.cc
will-cromar: Fixed

will-cromar:
> what would the error message be after your change?

It would just be the first error, `$PJRT_DEVICE is not set.`

jonb377 (approved):

will-cromar:
There seems to be a consistent error during teardown during the GPU tests. Creating a GPU dev environment to see what's going on

will-cromar:
Some of this GPU tests are starting to pass, so it looks like this tweak works.

torch_xla/csrc/runtime/runtime.cc
```diff
@@ -8,35 +8,29 @@
...
-ComputationClient* CreateClient() {
-  bool was_initialized = g_computation_client_initialized.exchange(true);
-  XLA_CHECK(!was_initialized) << "ComputationClient already initialized";
-  if (sys_util::GetEnvBool("XLA_DUMP_FATAL_STACK", false)) {
-    tsl::testing::InstallStacktraceHandler();
-  }
-
-  ComputationClient* client;
+ComputationClient* GetComputationClient() {
+  static std::unique_ptr<ComputationClient> client = []() {
```
jonb377, line 15: What's the difference between this and the original approach, since both rely on the static initializer? It seems like the logic has just been moved out of CreateClient and into the lambda.

will-cromar: Yeah, that's right. This lets me skip checking the case where this function gets called twice, the handling of which causes the bad error message. C++11 statics will ensure that it only completes once, and the function is anonymous now, which will prevent future code from erroneously calling `CreateClient` somewhere else in the future.

jonb377: Ah I see, the lambda is guaranteed to only ever be called once. Do we know how `CreateClient` was called twice in the first place?

will-cromar: I added that check to defend against future errors (likely by future me). There was never a case in the original code where `CreateClient` actually _completed_ twice.

The bad error messaging occurred when the constructor of `PjRtComputationClient` threw an exception. `g_computation_client_initialized` was never reset to `false`, `GetComputationClientIfInitialized` calls `GetComputationClient` during teardown, which re-runs `CreateClient`, which complains because `g_computation_client_initialized` never got reset.

In this PR, I only set `g_computation_client_initialized` _after_ the actual runtime init completed without breaking, since there are guaranteed to be no other concurrent callers of `CreateClient`.

PR #5949, opened 2023-11-30
use a common random seed for test core aten ops

qihqi:


wonjoolee95 (approved):

JackCaoG:
Let's keep playing with this pr until ci can pass consistently. If this takes more than 1 day we should consider disable this test on head for now.

JackCaoG:
bump, master is still flaky. We want to fix this for branch cut and before upstream update their pin to include this test.

qihqi:
`torch._export.verifier.SpecViolationError: Constant tensor _lifted_tensor_constant0 is not in the tensor constants dictionary.` Looks like this error comes from upstream. I'll just mark those as skip for now.

PR #5950, opened 2023-11-30
Add all-gather coalescing for FSDP/ZeRO1

jeffhataws:
This PR adds all-gather coalescence support and use that in FSDP/ZeRO1 (replacing https://github.com/pytorch/xla/pull/5624). This PR is to be used in conjunction with https://github.com/openxla/xla/pull/5740 .

A separate and related PR for reduce-scatter coalescence that also enables using reduce-scatter's scale param in FSDP is https://github.com/pytorch/xla/pull/5938.

This is a revival of https://github.com/pytorch/xla/pull/4145 . Will need to address the comments.

### DIFF
--- a/torch_xla/csrc/cross_replica_reduces.cpp
+++ b/torch_xla/csrc/cross_replica_reduces.cpp
@@ -247,8 +247,10 @@ AllGatherResultCoalesced BuildAllGatherCoalesced(
   // AllGather().
   ReduceContext cc_ctx = GetReduceContext(inputs);
   std::vector<xla::XlaOp> result(inputs.size());
+
   for (auto& type_ctx : cc_ctx.contexts) {
     xla::XlaOp all_gather_result;
+    type_ctx.second.ops[0] = token_handler.GetInput(type_ctx.second.ops[0], &type_ctx.second.operand_shapes[0]);
     if (pin_layout) {
       all_gather_result = xla::AllGather(
           xla::Tuple(inputs[0].builder(), type_ctx.second.ops), dim,
--- a/torch_xla/csrc/cross_replica_reduces.cpp
+++ b/torch_xla/csrc/cross_replica_reduces.cpp
@@ -210,11 +210,36 @@ AllToAllResult BuildAllToAll(xla::XlaOp input, xla::XlaOp token,
   return {reduce_result, token_handler.GetNewToken(reduce_result)};
 }
 
-AllGatherResult BuildAllGather(absl::Span<const xla::XlaOp> inputs,
-                               xla::XlaOp token, int64_t dim,
+AllGatherResult BuildAllGather(xla::XlaOp input, xla::XlaOp token, int64_t dim,
                                int64_t shard_count,
                                const std::vector<std::vector<int64_t>>& groups,
                                bool pin_layout) {
+  std::vector<xla::ReplicaGroup> reduce_groups = CreateReduceGroups(groups);
+  const xla::Shape& input_shape = ShapeHelper::ShapeOfXlaOp(input);
+  TokenHandler token_handler(token);
+  xla::XlaOp all_gather_result;
+  if (pin_layout) {
+    torch::lazy::BackendDevice xla_device = bridge::GetCurrentDevice();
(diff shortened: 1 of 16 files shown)


JackCaoG:
@jeffhataws let me know when you are done addressing comments, I will take another look

torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -309,6 +348,16 @@ ReduceScatterResult BuildReduceScatter(
   return {reduce_result, token_handler.GetNewToken(reduce_result)};
 }
 
+// moved from torch_xla/csrc/ops/all_reduce.cpp
```
JackCaoG, line 351: let's remove this comment and rename this to `GetOperandListWithToken`

torch_xla/csrc/cross_replica_reduces.h
```diff
@@ -65,6 +71,11 @@ AllGatherResult BuildAllGather(xla::XlaOp input, xla::XlaOp token, int64_t dim,
                                const std::vector<std::vector<int64_t>>& groups,
                                bool pin_layout);
 
+AllGatherResultCoalesced BuildAllGatherCoalesced(
+    absl::Span<const xla::XlaOp>, xla::XlaOp token, int64_t dim,
```
JackCaoG, line 75: nit `absl::Span<const xla::XlaOp> inputs `

torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py
```diff
@@ -295,6 +295,7 @@ def __init__(
       sharding_world_size: Optional[int] = None,
       shard_param_on_dim_0: bool = False,
       pin_layout_in_collective_ops: bool = True,
+      coalesce_all_gather_ops: bool = False,
```
JackCaoG, line 298: Do you mind explaining the change in this file? I think `coalesce_all_gather_ops ` is always `False` in our test, did you run into these issues with your own test?

torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -236,6 +236,45 @@ AllGatherResult BuildAllGather(xla::XlaOp input, xla::XlaOp token, int64_t dim,
...
+    bool pin_layout) {
+  std::vector<xla::ReplicaGroup> cc_groups = CreateReduceGroups(groups);
+  TokenHandler token_handler(token);
+  // TODO: We use pseudo-tokens ATM, which are real values. This need to be
+  // switched to use the real XLA Token once support has been added to XLA
+  // AllGather().
+  ReduceContext cc_ctx = GetReduceContext(inputs);
+  std::vector<xla::XlaOp> result(inputs.size());
+
+  for (auto& type_ctx : cc_ctx.contexts) {
```
JackCaoG, line 251: if you want to assume there is only one `type_ctx`, let's not use the for loop and `GetReduceContext` at all. This way we don't need to handle the token per type.

JackCaoG (commented):
mostly lgtm beside the changes in FSDP. If we didn't change the default behavior of all-gather test should pass right?

I will look into reduce scatter one today, let's try to merge these two pr soon.

torch_xla/csrc/cross_replica_reduces.cpp
jeffhataws: fixed

torch_xla/csrc/cross_replica_reduces.h
jeffhataws: fixed

torch_xla/csrc/cross_replica_reduces.cpp
jeffhataws: Let me check with others on this.

JackCaoG (approved):
Thanks! I think we should test allgather_coalesced using resnet on gpu to make sure we don't break it in the future. You can refer to existing test https://github.com/pytorch/xla/blob/2c4983dd1ab7600bd359740cb1f9f06c154d2933/.circleci/common.sh#L136.

we can do that in a separate pr.

PR #5957, opened 2023-12-01
Fix test_dynamo_input_sharding_threashold

JackCaoG:
Not sure what happened but it seem like test is a bit problematic, update the test with comments to explain the expected behavior.


JackCaoG:
@yeounoh can I get a review for this?

yeounoh (approved):
Lgtm

PR #5963, opened 2023-12-01
Reduce benchmark experiment filename string.

ysiraichi:
This PR removes both `accelerator_model` and `xla_flags` from the list of components of the filename string for a given benchmark experiment.

**Problem:** some filesystems (e.g. ext4)  have limitations on filename length (e.g. 255 bytes)
- `accelerator_model` gathers the model of all available GPUs (16 A100 ended up with a 400+ bytes filename)
- `xla_flags` might not get there, but I guess it's better to leave this one out

### DIFF
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -183,9 +183,7 @@ class BenchmarkExperiment:
     d = OrderedDict()
     d["experiment_name"] = self.experiment_name
     d["accelerator"] = self.accelerator
-    d["accelerator_model"] = self.accelerator_model
     d["xla"] = self.xla
-    d["xla_flags"] = self.xla_flags
     d["dynamo"] = self.dynamo
     d["test"] = self.test
     d["batch_size"] = self.batch_size
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -177,13 +177,21 @@ class BenchmarkExperiment:
 
   @property
   def filename_str(self):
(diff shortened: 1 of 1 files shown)


ysiraichi:
cc @JackCaoG @miladm

frgossen:
> This PR removes both accelerator_model and xla_flags from the list of components of the filename string for a given benchmark experiment.
The diff looks like it's doing something else. Can you check?

ysiraichi:
Ah, you are right. Fixed that.

benchmarks/benchmark_experiment.py
```diff
@@ -178,15 +178,13 @@ def get_device(self):
...
-    return "-".join(self.to_dict().values())
+    return "-".join(str(v) for v in self.to_dict().values()).replace(" ", "")
 
   def to_dict(self):
     d = OrderedDict()
     d["experiment_name"] = self.experiment_name
     d["accelerator"] = self.accelerator
-    d["accelerator_model"] = self.accelerator_model
     d["xla"] = self.xla
-    d["xla_flags"] = self.xla_flags
```
frgossen, line 189: I don't think this is the place to remove them. We want them in the reports and if I understand you correctly, you only want to remove them from filenames.

ysiraichi: Agreed.

ysiraichi:
Closed due to [a timeout issue](https://github.com/pytorch/xla/pull/6061#issuecomment-1847632833). Re-opened the PR at #6069.

PR #5965, opened 2023-12-01
Enable TF32 multiplication by default.

golechwierowicz:
This PR enables fast TF32 for PyTorch by default to mirror XLA behaviour.


frgossen (approved):
Thank you, especially for finding this.

vanbasten23:
Thanks for adding this option! Would you also add to the documentation about the sample commands we should run for inductor, pytorch/xla dynamo, etc., so that both Quansight and xla:GPU folks who works on the xla:GPU benchmarking run the same commands?

PR #5969, opened 2023-12-01
Add profiler API for async capture

jonb377:
To more easily support programmatic captures, this changes add a new API `trace_detached` to `torch_xla.debug.profiler` to capture a profile from a background thread.

Tested by using the new API in Llama2 training.

### DIFF
--- a/test/test_profile_mp_mnist.py
+++ b/test/test_profile_mp_mnist.py
@@ -1,6 +1,29 @@
 """Fork of test_train_mp_mnist.py to demonstrate how to profile workloads."""
 import args_parse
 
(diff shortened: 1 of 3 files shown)


jonb377:
> Thanks a lot! Can we add a test to https://github.com/pytorch/xla/blob/master/test/test_profiler.py ? We already have a test for `trace` so I imagine you can mostly just copy paste the test and make small modifications.

Thanks @JackCaoG - I generalized the test_profile_mp_mnist script to support programmatic captures and added a test case

JackCaoG (approved):
Thanks!

PR #5970, opened 2023-12-01
Optimize autocast tests.

vanbasten23:
This PR speeds up the test_autocast by setting up autocast_lists, autocast_lists_extra, autocast_unsupported_lists before each test suite instead of each test.

The test suite test_autocast.py has been really slow. (Takes 82 min for [TestAutocastCuda](https://screenshot.googleplex.com/9pjvMCmhYLzGN8W) on HEAD because it creates a new autocast_lists, autocast_lists_extra, autocast_unsupported_lists for each test.

The test `_run_autocast_outofplace` cast out of place, and tensor.to(dtype) returns a copy of the tensor so each test does not change the input tensor. I print the `op and args` in`for op, args in TestAutocastCuda.get_autocast_list('linalg_fp16')` before and after the test runs and the result seems to be identical.

### DIFF
--- a/test/test_autocast.py
+++ b/test/test_autocast.py
@@ -230,18 +230,13 @@ class AutocastCudaTestUnsupportedLists(object):
 
 class TestAutocastBase(unittest.TestCase):
   
-  count = 0
-
   @classmethod
   def setUpClass(cls):
-    print('xw32 calling setupClass line235, cls=', cls)
     cls.autocast_lists = AutocastTestLists(torch.device(xm.xla_device()))
-    print('xw32 cls.autocast_lists=', cls.autocast_lists)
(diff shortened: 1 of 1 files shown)


JackCaoG (approved):
Nice! I guess the timeout we saw in the pin update is not a real regression then?

ManfeiBai (approved):
LGTM

vanbasten23:
> Nice! I guess the timeout we saw in the pin update is not a real regression then?

I summarized in this [ticket](b/314169482). I ran a commented out test and compared `met.metrics_report()` and HLO, before and after pin update. The results are very close. So I don't think autocast regressed.

yeounoh (commented):
This is great, thanks @vanbasten23

PR #5974, opened 2023-12-01
Enable Cumsum with long input

JackCaoG:
Verified that this runs on TPU as well.


JackCaoG:
kokoro test failure should be fine

alanwaketan (approved):
LGTM.

JackCaoG:
we don't really run CPP test on TPU CI, I will see if I can add a similar test to `test_operation.py` as a follow up.

PR #6019, opened 2023-12-04
Enable reduce scatter test on gpu

JackCaoG:
I think reduce_scatter works on GPU as well.

cc @jeffhataws


JackCaoG:
OK it actually passed, so we should definally enable it. @jeffhataws can you rebase after this pr merged and added a test cast for both `all_gather` and `reduce_scatter` with a list of input tensors? this way we know it actually works.

vanbasten23 (approved):
LGTM!

alanwaketan (commented):
LGTM.

PR #6021, opened 2023-12-04
Disable flaky  test

JackCaoG:
I can't really figure out a way to optionally disable a test so I just disable it for now to unblock the upstream CI. I took a look at the test, if that thing doesn;'t work, there is no way any test would pass. I think we are not losing any coverage... especially this is actually the last test to run, instead of first.

### DIFF
--- a/torch_xla/csrc/runtime/BUILD
+++ b/torch_xla/csrc/runtime/BUILD
@@ -371,27 +371,28 @@ ptxla_cc_test(
     ],
 )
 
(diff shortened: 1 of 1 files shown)


will-cromar (approved):
We should have a way to disable tests for upstream. This code path is exercised by basically every other test, so it's not useful unless for upstream contributors, but it is useful to know if you broke something low-level in this repository. I expect that's true for many other unit tests.

PR #6023, opened 2023-12-04
Not creating the coordinator servie for single process.

vanbasten23:


jonb377:
@vanbasten23 Is the crash only in the case of single-process workloads? I wonder if it's a more general issue that we should handle. Agree that we don't need a coordinator for single-process though.

torch_xla/csrc/runtime/xla_coordinator.cc
```diff
@@ -9,6 +9,9 @@ namespace runtime {
 
 XlaCoordinator::XlaCoordinator(int global_rank, int world_size,
                                std::string master_addr, std::string port) {
+  if (world_size <= 1) {
```
jonb377, line 12: Can we instead avoid creating the client specifically for GPUs in the PjRtComputationClient constructor? Autocheckpointing will still require a coordinator even with world size 1.

vanbasten23: Sure.

vanbasten23:
> @vanbasten23 Is the crash only in the case of single-process workloads? I wonder if it's a more general issue that we should handle. Agree that we don't need a coordinator for distributed kv store in single-process though.

It fails in the persistent cache test https://gist.github.com/vanbasten23/2cb90b2f72a40ef965b965bc12bc5ded. I fixed your comment and let me rerun.

jonb377:
> > @vanbasten23 Is the crash only in the case of single-process workloads? I wonder if it's a more general issue that we should handle. Agree that we don't need a coordinator for distributed kv store in single-process though.
> 
> It fails in the persistent cache test https://gist.github.com/vanbasten23/2cb90b2f72a40ef965b965bc12bc5ded. I fixed your comment and let me rerun.

@vanbasten23 I've merged the persistent cache change, if you want to test the single device test with this fix you can do a rebase on master and modify the single-device test to run on GPU [here](https://github.com/pytorch/xla/commit/fae016687f227a5cd8f03e360e7391567458c71f#diff-ad1ed131a3dfb0e2bfaadd7cb3d4959549097b49c45838449738dd620a0e6fd8R111)

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -177,27 +177,29 @@ PjRtComputationClient::PjRtComputationClient() {
...
-      };
+    auto allowed_devices =
+        std::make_optional<std::set<int>>(std::set{local_process_rank});
+    if (global_world_size > 1) {
+      // Use the XlaCoordinator as the distributed key-value store.
+      coordinator_ = std::make_unique<XlaCoordinator>(
+          global_process_rank, global_world_size, master_addr, port);
+      std::shared_ptr<xla::DistributedRuntimeClient> distributed_client =
+          coordinator_->GetClient();
+      if (distributed_client != nullptr) {
```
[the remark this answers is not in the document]
vanbasten23: I like the idea!

will-cromar (approved):

jonb377 (approved):
LGTM, thanks!

JackCaoG:
Do we need this in 2.2?

PR #6027, opened 2023-12-05
handle dim_size=0 in slice

lsy323:
In the lowering of `slice`, if the `dim_size=0`, will it hit the assertion in the upstream helper function [ref](https://github.com/pytorch/pytorch/blob/main/torch/csrc/lazy/core/helpers.cpp#L29) (args are dim = -1, rank = 0).

minimal repro of the issue
```
x = torch.rand(0,1,512)
x = x.to(xm.xla_device())
y = torch.ops.aten.slice(x, 0,0,-1,2)
```

The expected behavior of slice a empty tensor is to give an empty tensor

We add a special handling case for `dim_size=0` in this PR to workaround the constraint in `GetCanonicalDimensionIndex`. 

~An upstream PR to fix `GetCanonicalDimensionIndex` for this special case will be provided later.~
After having a second thought, it may not be reasonable to have `GetCanonicalDimensionIndex` to support `dim_size=0`, because a dim with size of 0 cannot be indexed. (e.g. t has size `[0,1,2]`, then`t[0,:,:]` and `t[-1,:,:]`) are both invalid, and `GetCanonicalDimensionIndex` is intended to Canonicalize index. So I think it makes more sense to keep the special handling in the `slice` lowering.

### DIFF
--- a/test/cpp/test_aten_xla_tensor_1.cpp
+++ b/test/cpp/test_aten_xla_tensor_1.cpp
@@ -1505,6 +1505,18 @@ TEST_F(AtenXlaTensorTest, TestSlice) {
   });
 }
 
+TEST_F(AtenXlaTensorTest, TestSliceZeroDimSize) {
+  torch::Tensor a =
+      torch::rand({0, 24, 16}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor b = torch::slice(a, 0, 0, -1, 1);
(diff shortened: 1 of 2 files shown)


qihqi (approved):
thanks！

lsy323:
> Do you need to backport this to 2.2?

No need to have this in 2.2

PR #6029, opened 2023-12-05
Add a patch for mismatched batch size.

golechwierowicz:
Created a separate folder to store potential patches, and added a patch I use for A100 to get the right batch size for inference. This patch has to be applied from the torchbench git repository.
In order to use it you have to change the `current_device_name` to the CUDA accelerator name you're currently using.

### DIFF
--- a/benchmarks/patches/mismatched_batch_size.patch
+++ b/benchmarks/patches/mismatched_batch_size.patch
@@ -6,6 +6,7 @@ index 8593ba4c..57fef507 100644
 
          # use the device suggestion on CUDA inference tests, key should be either eval_batch_size or train_batch_size
          device_batch_size_key = f"{self.test}_batch_size"
++        # A patch to making sure batch sizes are comparable. It's needed because xla device string is unrecognized.
 +        current_device_name = 'NVIDIA A100-SXM4-40GB'
          if self.metadata and "devices" in self.metadata and current_device_name in self.metadata["devices"] \
                              and device_batch_size_key in self.metadata["devices"][current_device_name]:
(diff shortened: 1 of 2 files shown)


benchmarks/patches/mismatched_batch_size.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/torchbenchmark/util/model.py b/torchbenchmark/util/model.py
+index 8593ba4c..57fef507 100644
+--- a/torchbenchmark/util/model.py
++++ b/torchbenchmark/util/model.py
+@@ -182,6 +182,7 @@ class BenchmarkModel(metaclass=PostInitProcessor):
+
+         # use the device suggestion on CUDA inference tests, key should be either eval_batch_size or train_batch_size
+         device_batch_size_key = f"{self.test}_batch_size"
++        current_device_name = 'NVIDIA A100-SXM4-40GB'
```
frgossen, line 9: Maybe an in-patch comment on why this is needed?

frgossen (approved):
Thanks for fixing this!

vanbasten23:
Could you also add it to the readme so other users know how to use it?

benchmarks/patches/mismatched_batch_size.patch
golechwierowicz: Done, I additionally added the info how to apply the patch into the README file (cc: @vanbasten23)

miladm:
can we consider the same path for H100 and L4?

@golechwierowicz @frgossen @zpcore

PR #6030, opened 2023-12-05
Fix CUDA profiling dump.

golechwierowicz:
Now we take into consideration the directory to dump kernel profile and traces. Additionally there is also a support for gathering the e2e profiling information across the entire benchmark suite. Previously it was not possible due to conflicting resulting filenames.

### DIFF
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -178,7 +178,7 @@ class BenchmarkExperiment:
   @property
   def filename_str(self):
     return "-".join(
-        str(x) if x is not None else 'null' for x in self.to_dict().values())
+        str(x) if x is not None else 'None' for x in self.to_dict().values())
 
   def to_dict(self):
     d = OrderedDict()
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -177,7 +177,8 @@ class BenchmarkExperiment:
 
   @property
   def filename_str(self):
-    return "-".join(self.to_dict().values())
+    return "-".join(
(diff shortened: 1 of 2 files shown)


benchmarks/benchmark_experiment.py
```diff
@@ -177,7 +177,8 @@ def get_device(self):
 
   @property
   def filename_str(self):
-    return "-".join(self.to_dict().values())
+    return "-".join(
+        str(x) if x is not None else 'null' for x in self.to_dict().values())
```
frgossen, line 181: Think most uses here use "None" strings. Null might be confusing.

benchmarks/experiment_runner.py
```diff
@@ -235,20 +235,25 @@ def prepare_inputs(self, example_inputs, should_randomize_input):
...
     os.makedirs(file_path, exist_ok=True)
-    prof.export_chrome_trace(os.path.join(file_path, "trace.json"))
+    prof.export_chrome_trace(
+        os.path.join(file_path, with_configuration_suffix("trace", "json")))
 
     kernel_dump = prof.key_averages().table(
         sort_by="cuda_time_total", row_limit=500)
-    with open(os.path.join(file_path, "kernel_dump.txt"), "a") as f:
+    with open(
+        os.path.join(file_path, with_configuration_suffix("kernel_dump",
```
frgossen, line 255: This is the PyTorch profile, right? Can we include `pt-profile` or something in the name to disambiguate form the ptxla metrics?

frgossen (approved):
Thanks for adding this nicely. 
Two comments on naming.

benchmarks/experiment_runner.py
golechwierowicz: Done.

PR #6031, opened 2023-12-05
Fix tensor to xla dtype conversion error

yeounoh:
This addresses a minor bug in dtype conversion for TPU. It was surfaced during SPMD unit testing on v5 with `PJRT_DEVICE=TPU_C_API`.


yeounoh:
Tested locally with TPUv4 and v5, the CI shouldn't be affected -- since we don't use TPU. cc @JackCaoG @will-cromar

JackCaoG:
`TPU_C_API ` should be deprecated I think..

will-cromar:
> `TPU_C_API ` should be deprecated I think..

Missed this PR. @JackCaoG is correct. I'll remove `TPU_C_API` for the next release since it's redundant. We don't support SE on TPU anymore.

PR #6033, opened 2023-12-06
benchmarks/result_analyzer: set default output format to csv

cota:
To match our documentation.

Moreover, note that using JSON output can lead to confusion: we read in all jsonl files in the output directory, and therefore after running the script twice, we try to read (and fail) the previously-generated jsonl file. With CSV this is not a problem.


frgossen (approved):
Thanks!

zpcore (approved):

PR #6034, opened 2023-12-06
benchmarks/aggregate: add tests + use oldest benchmarked Inductor as baseline

cota:
See commit logs and updated readme for rationale.


frgossen:
This is great. Thank you!

benchmarks/test/Makefile
```diff
@@ -0,0 +1,21 @@
+TEST_ARGS = $(shell echo $@ | perl -pe 's/([^.]*)\.([^.]*)\.([^.]*)\.test/--accelerator=$$1 --test=$$2 --report=$$3/')
```
frgossen, line 1: @JackCaoG , is this the way to add such a test or is there a more idiomatic way to add such tests in PyTorch/XLA?

JackCaoG: we run out CI in https://github.com/pytorch/xla/blob/1ea23741cadbe754728d2dac42fcfcb2718806dc/.circleci/common.sh#L128

that script will run python tests.

cota: Thanks! Sent https://github.com/pytorch/xla/pull/6035 as a follow-up PR

frgossen (approved):

PR #6035, opened 2023-12-06
test: add benchmarks/test to run_tests.sh

cota:


JackCaoG:
`run_test.sh` is mostly for the python test, we have a `run_test.sh` for cpp test as well https://github.com/pytorch/xla/blob/master/test/cpp/run_tests.sh. I recommend you add a run_test.sh for the benchmark instead of adding it to the python script.

cota:
> `run_test.sh` is mostly for the python test, we have a `run_test.sh` for cpp test as well https://github.com/pytorch/xla/blob/master/test/cpp/run_tests.sh. I recommend you add a run_test.sh for the benchmark instead of adding it to the python script.

Will do, thanks!

cota:
PTAL.

JackCaoG:
@vanbasten23 can you take a look? It seems like CI is not being run in this pr because it is not trying to merge to the master..

test/benchmarks/Makefile
```diff
@@ -11,7 +11,7 @@ ifndef V
 endif
 
 $(TESTS):
-	$(QUIET_AGGREGATE)python3 ../aggregate.py $(TEST_ARGS) \
```
vanbasten23, line 14: Curious because I couldn't see the file benchmarks/test/Makefile in HEAD https://github.com/pytorch/xla. Where is the current file located?

cota: Oh I've added this PR on top of this other PR https://github.com/pytorch/xla/pull/6034 [still waiting for approval] to ease review. I was wondering why CI wasn't running, now I know why!
Will rebase onto master once the parent PR is merged. Will ping here once that happens.

test/benchmarks/a6000.csv
vanbasten23, line 1: I wonder why we need to check in this file

test/benchmarks/run_tests.sh
```diff
@@ -0,0 +1,46 @@
...
+  esac
+done
+shift $(($OPTIND - 1))
+
+function run_make_tests {
+  MAKE_V=""
+  if [ "$VERBOSITY" != "0" ]; then
+    MAKE_V="V=$VERBOSITY"
+  fi
+  make -C $CDIR $MAKE_V all
```
vanbasten23, line 35: Wondering what the benefits are to use Makefile as opposed to running regular python test, for example "PJRT_DEVICE=CUDA python test/test_train_mp_imagenet_fsdp.py --fake_data --use_nested_fsdp --use_small_fake_sample --num_epochs=1".

test/benchmarks/v100.csv
vanbasten23, line 1: do we need to check in this file?

test/benchmarks/Makefile
```diff
@@ -11,7 +11,7 @@ ifndef V
...
+	$(QUIET_AGGREGATE)python3 ../../benchmarks/aggregate.py $(TEST_ARGS) \
```
vanbasten23, line 14: Not quite familiar with the Makefile but I guess it tests by generating a tmp file and compare it with the other checked in files?

vanbasten23:
+@yeounoh , when we add a new test to the CI, does the new test run in the CI before we merge the PR or it only appears after the code is merged?

test/benchmarks/Makefile
cota: Correct. See the example run here:  https://github.com/pytorch/xla/pull/6035/commits/bbf01bebf21e
TESTS expands to all the *.csv files in the directory. The arguments passed to aggregate.py are taken from the file name of the CSV file -- this is what `TEST_ARGS` does.

.circleci/common.sh
```diff
@@ -196,6 +196,13 @@ function run_torch_xla_cpp_tests() {
   popd
 }
 
+function run_torch_xla_benchmark_tests() {
```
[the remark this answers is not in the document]
cota: I did that at first. However @JackCaoG asked me to move these tests to their own run_tests.sh file, since they're not Python tests per se (as the ones in xla/test/run_tests.sh). That's why I added a new xla/test/benchmarks/run_tests.sh file. (Note that we also have xla/test/cpp/run_tests.sh; those tests aren't Python tests either.)

cota:
I just realized that the only way to merge a PR is to squash all commits in a PR (pushing to master directly is not allowed). I really dislike this (why would I make small commits otherwise), so I will split this PR into 3 single-commit PRs. This will hopefully also take care of the lack of CI running on this PR.

cota:
> I just realized that the only way to merge a PR is to squash all commits in a PR (pushing to master directly is not allowed). I really dislike this (why would I make small commits otherwise), so I will split this PR into 3 single-commit PRs. This will hopefully also take care of the lack of CI running on this PR.

See #6043

PR #6045, opened 2023-12-07
benchmarks/aggregate: use typing.List[] instead of builtins.list[]

cota:
CI runs Python 3.8, which doesn't support builtins.list[] -- see https://docs.python.org/3/library/typing.html#typing.List


frgossen (approved):
Thanks!

PR #6046, opened 2023-12-07
Add documentation and python API for persistent cache

jonb377:
This change adds a Python API to activate the persistent cache, rather than controlling it exclusively through environment variables.

### DIFF
--- a/torch_xla/runtime.py
+++ b/torch_xla/runtime.py
@@ -265,3 +265,21 @@ def get_master_ip() -> str:
   if device_type() == 'TPU':
     return tpu.discover_master_worker_ip()
   raise RuntimeError(f'IP discovery not supported for device: {device_type()}')
+
+
+@requires_pjrt
+def initialize_cache(path: str, readonly: bool = False):
+  """Initializes the persistent compilation cache. This API must be called
+  before any computations have been performed.
+
+  Args:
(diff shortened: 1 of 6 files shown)


will-cromar (approved):
Nice! I prefer the Python API to the environment variables. +1 to eliminating env vars as the way we communicate between Python and C++

jonb377:
> Nice! I prefer the Python API to the environment variables. +1 to eliminating env vars as the way we communicate between Python and C++

Let me do that now actually, keeping the env vars in 2.2 would mean a whole deprecation process.

torch_xla/runtime.py
```diff
@@ -265,3 +265,21 @@ def get_master_ip() -> str:
...
+
+  Args:
+    path: The path at which to store the persistent cache.
+    readonly: Whether or not this worker should have write access to the cache.
+  """
+  assert not torch_xla._XLAC._xla_computation_cache_is_initialized(
+  ), "Computation cache has already been initialized"
+
+  # TODO(jonbolin): Consider moving away from environment variables to control
+  # the cache.
```
JackCaoG, line 283: yea I think it is a good idea to just drop env var going forward.

JackCaoG (approved):

yeounoh:
cc @zpcore for visibility

PR #6051, opened 2023-12-07
benchmarks/aggregate: output datetime in UTC

cota:
Using a specific timezone makes sure that our tests are consistent regardless of where they are run.

UTC seems like a reasonable TZ choice.

While at it, fix a bug in pr_latest's PNG code path.


benchmarks/aggregate.py
```diff
@@ -225,13 +225,13 @@ def pr_latest(results_map: dict[str, Any], args, timestamps: List[str]):
     plt.plot(speedups[0], label='Inductor', marker='^')
     plt.plot(speedups[1], label='PytorchXLA', marker='o')
     plt.legend()
-    dates = date.fromtimestamp(float(speedup_timestamps[0]))
+    datestr = datetime.utcfromtimestamp(float(speedup_timestamps[0]))
     if speedup_timestamps[0] != speedup_timestamps[1]:
-      dates = f'{dates} (Inductor)'
-      dates += ', {date.fromtimestamp(float(dates[1]))} (PytorchXLA)'
```
cota, line 231: `dates` (now `datestr`) is not a list (it used to be in my local repo); this is the drive-by fix mentioned in the commit log.

frgossen (commented):
Thanks!

PR #6059, opened 2023-12-08
Add out-of-place all-gather coalesced

jeffhataws:
Similar to https://github.com/pytorch/xla/pull/6058 , we add out-of-place version for all-gather coalesced (which was introduced in https://github.com/pytorch/xla/pull/5950 without an out-of-place version).

### DIFF
--- a/test/test_mp_all_gather.py
+++ b/test/test_mp_all_gather.py
@@ -77,10 +77,7 @@ def _mp_fn(index):
         sys.exit(1)
 
     # Testing with a single replica group and tensor list as input and output!=None (out-of-place)
-    ordinal_tensors = [
-        torch.tensor([i * 1000 + index], dtype=torch.float).to(device)
-        for i in range(input_list_size)
-    ]
+    # Reuse ordinal_tensors from previous test
     output_tensors = [
         torch.zeros([world_size], dtype=torch.float).to(device)
         for i in range(input_list_size)
--- a/test/test_mp_all_gather.py
+++ b/test/test_mp_all_gather.py
@@ -13,7 +13,8 @@ def all_gather(tensor, dim):
 def _mp_fn(index):
   device = xm.xla_device()
   world_size = xm.xrt_world_size()
-  if xm.xla_device_hw(device) in ('TPU', 'GPU', 'CUDA', 'ROCM'):
+  input_list_size = 5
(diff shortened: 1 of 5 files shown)


test/test_mp_all_gather.py
```diff
@@ -57,6 +58,48 @@ def _mp_fn(index):
...
+      expected = i * 1000 + torch.arange(world_size, dtype=torch.float)
+      if not cpu_result.allclose(expected):
+        print(
+            'xm.all_gather() produced wrong reductions for item {i} in result list',
+            file=sys.stderr)
+        print(f'[{index}] {cpu_result}', file=sys.stderr)
+        sys.exit(1)
+
+    # Testing with a single replica group and tensor list as input and output!=None (out-of-place)
+    ordinal_tensors = [
```
JackCaoG, line 80: question, couldn't we just reuse the `ordinal_tensors` from the previous test case? I thought `all_gather` won't touch the original tensor.

torch_xla/core/xla_model.py
```diff
@@ -553,7 +553,7 @@ def all_gather(value, dim=0, groups=None, output=None, pin_layout=True):
     A tensor which has, in the ``dim`` dimension, all the values from the
     participating replicas.
   """
-  if pin_layout and (output == None or xla_device_hw(value.device) == 'NEURON'):
+  if pin_layout and output == None and isinstance(value, torch.Tensor):
```
JackCaoG, line 556: Could you add a comment to explain the check here? I think it means we can't use `_all_gather_using_all_reduce` if input is a list or `output` is not none? I forgot why we added `xla_device_hw(value.device) == 'NEURON'` at first place..

JackCaoG (commented):
mostly lgtm, minor comments

torch_xla/core/xla_model.py
jeffhataws: Will add a note. Yes, _all_gather_using_all_reduce doesn't work with coaelescing (multiple tensor list).

test/test_mp_all_gather.py
jeffhataws: You are right. Fixed in latest commit.

JackCaoG (approved):
Thanks, I will merge once all CI passed

PR #6061, opened 2023-12-08
Enable CUDA support for PyTorch on CI.

ysiraichi:
This PR enables CUDA support for PyTorch for CI builds. Since [it's the recommended way](https://github.com/pytorch/xla/blob/master/docs/gpu.md#develop-pytorchxla-on-a-gpu-instance-build-pytorchxla-from-source-with-gpu-support) of installing PyTorch/XLA, we should also test it.

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -41,8 +41,7 @@ apply_patches
 
 python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
 
-# We always build PyTorch without CUDA support.
-export USE_CUDA=0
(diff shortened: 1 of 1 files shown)


ysiraichi:
@JackCaoG @miladm

JackCaoG:
@ysiraichi can you create a branch from pytorch/xla instead from your fork? Otherwise CI won't use cache and build will likely timeout

ysiraichi:
Closed due to [a timeout issue](https://github.com/pytorch/xla/pull/6061#issuecomment-1847632833). Re-opened the PR at #6070.

PR #6062, opened 2023-12-08
benchmarks/aggregate: round floats before printing them

cota:
So that when we diff them against golden files in tests, tiny differences won't matter.


frgossen (approved):
Thanks!

PR #5960, opened 2023-12-01
Use `subprocess` for checking XLA supported devices.

ysiraichi:
Fix: #5943

In summary:

- Create a self-contained XLA device checking python script: check_xla_device.py
- Use `subprocess` for calling the checking script with different devices

PR #6068, opened 2023-12-08
Use `subprocess` for checking XLA supported devices.

ysiraichi:
Re-opened from: #5960


frgossen (approved):
Thanks for fixing this. The comment helps!

vanbasten23 (approved):
LGTM!

PR #6072, opened 2023-12-08
Remove branch on undefined variable.

ysiraichi:
cc @JackCaoG @miladm

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -276,8 +276,7 @@ class ExperimentRunner:
       if evt.device_type == DeviceType.CPU:
         # in legacy profiler, kernel info is stored in cpu events
         if evt.is_legacy:
-          if not use_device:
-            total_cuda_time += evt.self_cuda_time_total
+          total_cuda_time += evt.self_cuda_time_total
       elif evt.device_type == DeviceType.CUDA:
         # in kineto profiler, there're events with the correct device type (e.g. CUDA)
         total_cuda_time += evt.self_cuda_time_total


benchmarks/experiment_runner.py
```diff
@@ -276,8 +276,7 @@ def collect_profile_to_metrics(self, prof, metrics):
       if evt.device_type == DeviceType.CPU:
         # in legacy profiler, kernel info is stored in cpu events
         if evt.is_legacy:
-          if not use_device:
```
ysiraichi, line 279: `use_device` wasn't defined anywhere.

golechwierowicz: Thanks for noticing this. It means that this path was never exercised before then. Can you raise an exception here instead so we won't be overcounting the profiler time if something changes?

ysiraichi: I did end up hitting this branch. As far as I remember, accounting for this number is correct. Let me take a look at how PyTorch counts that to confirm.

golechwierowicz: On what model did you hit this branch?

golechwierowicz: It might be GPU specific then.  Can you double check that there's not overcounting, post it here, and then we will merge to the main branch?

ysiraichi: This is similar to how PyTorch summarizes its data ([link](https://github.com/pytorch/pytorch/blob/afe6d272c69ae5671ca0df978be8fff7e8e4ed4e/torch/autograd/profiler_util.py#L1025-L1032)). Although it does explicitly checks for `use_device`, it's a redundant check, since [it's already done when we access the attribute](https://github.com/pytorch/pytorch/blob/afe6d272c69ae5671ca0df978be8fff7e8e4ed4e/torch/autograd/profiler_util.py#L567-L576). i.e. if `use_device == True`, `evt.self_cuda_time_total` will be 0.

Apparently, we execute the `not use_device` branch when PyTorch is compiled without support for Kineto.
That said, I ran `alexnet` and didn't notice any signs of overcounting.

golechwierowicz: Thanks for checking this!

golechwierowicz (approved):

vanbasten23 (approved):

PR #6075, opened 2023-12-08
Increase delay in trace_detached test

jonb377:
In #6058, the trace_detached test was flaky. This may be due to the tightness in duration of the profile and delay in the test.

Decreasing the duration of the capture will cause some required values to be missing from the proto, so this change instead increases the delay to allow the capture to be fully processed.


JackCaoG (approved):

jonb377:
Unfortunate, seems like a 1s capture doesn't include some of the expected values. I'll instead increase the delay.

PR #6077, opened 2023-12-08
Set DISABLE_NUMERIC_CC_TOKEN=1 default for Neuron

jeffhataws:
For Neuron we set DISABLE_NUMERIC_CC_TOKEN=1 to avoid extra scalar computation in CC operations. For background see https://github.com/pytorch/xla/pull/3825 .

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -78,10 +78,6 @@ def _summarize_fn_tracker():
 
 
 def _aws_ec2_inf_trn_init():
-  import importlib.util
-  if importlib.util.find_spec("torch_neuronx") is not None:
-    os.environ["DISABLE_NUMERIC_CC_TOKEN"] = "1"
-
   try:
     from torch_neuronx import xla
   except ImportError:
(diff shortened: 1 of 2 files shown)


JackCaoG:
do you need this to backport to 2.2?

jeffhataws:
> do you need this to backport to 2.2?

Yes. This was used since 1.13.

torch_xla/__init__.py
```diff
@@ -78,6 +78,10 @@ def _summarize_fn_tracker():
 
 
 def _aws_ec2_inf_trn_init():
+  import importlib.util
+  if importlib.util.find_spec("torch_neuronx") is not None:
+    os.environ["DISABLE_NUMERIC_CC_TOKEN"] = "1"
```
aws-rhsoln, line 83: can we move it to the else part of the below try except else

jeffhataws: Let me test that out first.

jeffhataws: Tested that setting within neuron.initialize_env gets picked up by GetInput and GetNewToken. Updated. Thanks!

jeffhataws:
Looks like I am able to set this by overriding neuron.initialized_env in torch-neuronx during init. So I will drop this PR for now.

PR #6080, opened 2023-12-08
Re-enable single-device GPU persistent cache test

jonb377:
With #6023 merged, the single-device GPU persistent cache test should work as intened. This change re-enables that test.

Currently a draft PR to verify that the test passes.


jonb377:
cc @vanbasten23, it looks like we're still creating the coordinator in the single-device test case: https://github.com/pytorch/xla/actions/runs/7147022524/job/19466545643?pr=6080

```
RuntimeError: torch_xla/csrc/runtime/xla_coordinator.cc:28 : Check failed: dist_runtime_client_->Connect().ok() 
```

I verified this commit includes #6023, so this should only be the case if one of the WORLD_SIZE env vars is set, right?

vanbasten23:
> cc @vanbasten23, it looks like we're still creating the coordinator in the single-device test case: https://github.com/pytorch/xla/actions/runs/7147022524/job/19466545643?pr=6080
> 
> ```
> RuntimeError: torch_xla/csrc/runtime/xla_coordinator.cc:28 : Check failed: dist_runtime_client_->Connect().ok() 
> ```
> 
> I verified this commit includes #6023, so this should only be the case if one of the WORLD_SIZE env vars is set, right?

FWIW, GPU CI set GPU_NUM_DEVICES=2, not sure if it causes the failure. Perhaps stick a print befoer https://github.com/pytorch/xla/blob/eb3c446f9b70f3de93369336dd20f767b39501cc/torch_xla/csrc/runtime/pjrt_computation_client.cc#L184 to validate?

jonb377:
Ah, good catch @vanbasten23 - it looks like when we `xmp.spawn`, it sets the `LOCAL_WORLD_SIZE` env var in the parent process: https://github.com/pytorch/xla/blob/fc713b7879282b0266efb570c03a801d058fc3f6/torch_xla/_internal/gpu.py#L12-L14

Then subsequent tests carry this env var and trigger the coordinator initialization. Maybe let's move setting `LOCAL_WORLD_SIZE` to within the subprocesses here in `initialize_multiprocess`? https://github.com/pytorch/xla/blob/fc713b7879282b0266efb570c03a801d058fc3f6/torch_xla/_internal/pjrt.py#L112-L115

I'll update this PR with the change and tag you.

vanbasten23:
> Ah, good catch @vanbasten23 - it looks like when we `xmp.spawn`, it sets the `LOCAL_WORLD_SIZE` env var in the parent process:
> 
> https://github.com/pytorch/xla/blob/fc713b7879282b0266efb570c03a801d058fc3f6/torch_xla/_internal/gpu.py#L12-L14
> 
> Then subsequent tests carry this env var and trigger the coordinator initialization. Maybe let's move setting `LOCAL_WORLD_SIZE` to within the subprocesses here in `initialize_multiprocess`?
> 
> https://github.com/pytorch/xla/blob/fc713b7879282b0266efb570c03a801d058fc3f6/torch_xla/_internal/pjrt.py#L112-L115
> 
> I'll update this PR with the change and tag you.

Maybe run the test locally and see if a coordinator is create. I suspect even if we set LOCAL_WORLD_SIZE in the subprocess, when the subprocess creates the pjrt client, it will see LOCAL_WORLD_SIZE(hence GLOBAL_WORLD_SIZE) to be 2 and also create a coordinator.

On another thought, if the test is single-device, can we override the default GPU_NUM_DEVICE and set it to 1 to reflect the single device intention?

jonb377:
> I suspect even if we set LOCAL_WORLD_SIZE in the subprocess, when the subprocess creates the pjrt client, it will see LOCAL_WORLD_SIZE(hence GLOBAL_WORLD_SIZE) to be 2 and also create a coordinator.

That's correct for the MP case, which is the intended behavior. After the MP test completes, the parent process's environment won't contain `LOCAL_WORLD_SIZE` anymore, so no coordinator should be created.

> can we override the default GPU_NUM_DEVICE and set it to 1 to reflect the single device intention

Single-device execution doesn't pass through `xmp.spawn`, so the `LOCAL_WORLD_SIZE` env won't be overridden if it remains active from the MP test.

I just updated the PR to move the env var to the subprocess, let's see how it goes!

vanbasten23:
> By moving it to the subprocess, after the MP test completes, the parent process's environment won't contain LOCAL_WORLD_SIZE anymore, so no coordinator should be created.

I thought only the subprocess creates the pjrt client and the coordinator. When does the parent process create the pjrt client and hence create the coordinator?

jonb377:
> > By moving it to the subprocess, after the MP test completes, the parent process's environment won't contain LOCAL_WORLD_SIZE anymore, so no coordinator should be created.
> 
> I thought only the subprocess creates the pjrt client and the coordinator. When does the parent process create the pjrt client and hence create the coordinator?

In the general single-device case, no subprocesses are launched (i.e. the user won't call `xmp.spawn`). A script like this will launch the client and coordinator in the main process:

```python
import torch, torch_xla.core.xla_model as xm

t = torch.randn(4, 4, device=xm.xla_device())
xm.mark_step()
```

vanbasten23:
> In the general single-device case, no subprocesses are launched (i.e. the user won't call `xmp.spawn`). A script like this will launch the client (and the coordinator if applicable) in the main process:
> 

Right.
I'm not sure if I follow the original problem. If you do `PJRT_DEVICE=CUDA GPU_NUM_DEVICE=1 python single_device_test.py`, why does it set WORLD_SIZE to be greater than 1 and create a coordinator service?

jonb377:
> > In the general single-device case, no subprocesses are launched (i.e. the user won't call `xmp.spawn`). A script like this will launch the client (and the coordinator if applicable) in the main process:
> 
> Right. I'm not sure if I follow the original problem. If you do `PJRT_DEVICE=CUDA GPU_NUM_DEVICE=1 python single_device_test.py`, why does it set WORLD_SIZE to be greater than 1 and create a coordinator service?

The specific issue here comes down to the order: the MP persistent cache test runs before the single-device test. MP uses `xmp.spawn`, which adds `LOCAL_WORLD_SIZE` to the environment of the main test process. This then carries over to the environment of the single-device test, where it incorrectly launches a coordinator expecting two clients to connect despite only one existing.

We could just reverse the order of the tests to avoid the issue, but I would prefer if we can avoid modifying the environment of the parent process, since it's subtle behavior which can be hard to debug.

vanbasten23:
> > > In the general single-device case, no subprocesses are launched (i.e. the user won't call `xmp.spawn`). A script like this will launch the client (and the coordinator if applicable) in the main process:
> > 
> > 
> > Right. I'm not sure if I follow the original problem. If you do `PJRT_DEVICE=CUDA GPU_NUM_DEVICE=1 python single_device_test.py`, why does it set WORLD_SIZE to be greater than 1 and create a coordinator service?
> 
> The specific issue here comes down to the order: the MP persistent cache test runs before the single-device test. MP uses `xmp.spawn`, which adds `LOCAL_WORLD_SIZE` to the environment of the main test process. This then carries over to the environment of the single-device test, where it incorrectly launches a coordinator expecting two clients to connect despite only one existing.
> 
> We could just reverse the order of the tests to avoid the issue, but I would prefer if we can avoid modifying the environment of the parent process, since it's subtle behavior which can be hard to debug.

Gotcha. I thought every test is isolated. What do you think we reset the WORLD_SIZE/LOCAL_WORLD_SIZE in a `def setUp` just to mimic the fact that in the real world before we run a script, the env var such as WORLD_SIZE are not set.

jonb377:
Yeah, I think that may be the way to go unfortunately - it seems the GPU client gets initialized very early compared to the TPU client, so we can't set the env in the child process.

PR #6082, opened 2023-12-09
Add benchmark tests

frgossen:
Add first test case for `experiment_runner.py`

### DIFF
--- a/test/benchmarks/test_experiment_runner.py
+++ b/test/benchmarks/test_experiment_runner.py
@@ -1,0 +1,32 @@
+import unittest
+
+import subprocess
+
+import experiment_runner
+
+EXPERIMENT_RUNNER_PY = experiment_runner.__file__
+
+
+class ExperimentRunnerTest(unittest.TestCase):
+
+  def test_alexnet_dry_run(self):
+    child = subprocess.run([
+        "python", EXPERIMENT_RUNNER_PY, "--dynamo=openxla", "--dynamo=inductor",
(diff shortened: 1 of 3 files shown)


test/benchmarks/test_experiment_runner.py
```diff
@@ -0,0 +1,30 @@
...
+EXPERIMENT_RUNNER_PY = experiment_runner.__file__
+
+
+class ExperimentRunnerTest(unittest.TestCase):
+
+  def test_alexnet_dry_run(self):
+    child = subprocess.run([
+        "python", EXPERIMENT_RUNNER_PY, "--dynamo=openxla", "--dynamo=inductor",
+        "--xla=PJRT", "--xla=None", "--test=eval", "--test=train",
+        "--suite-name=torchbench", "--accelerator=cpu", "--filter=^alexnet$",
```
frgossen, line 16: With cuda, this currently fails for the same reason it fails when benchmarking

ysiraichi: I think it's better to use CPU for these tests, anyway. That said, out of curiosity, what's the reason this benchmark fails?

frgossen: For the cuda initialization issue. It is fixed for me after https://github.com/pytorch/xla/pull/6068 but I am not still sure if this would pass the CI.  
Think we should eventually test both, especially because it failed. Will add a cuda test in a fllow-up.

cota (approved):

PR #6091, opened 2023-12-11
Increase delay in trace_detached test

jonb377:
Cherry-pick #6075 - the programmatic profile test is due to tightness between the test duration and the profile duration. If we see flakiness on the r2.2 branch, we should pull this in.


JackCaoG (approved):

jonb377:
I saw a flake in r2.2 this morning: https://github.com/pytorch/xla/actions/runs/7171632293/job/19528577295

Let's merge when CI finishes.

Issue #5887, opened 2023-11-21
[Core ATen Opset] Lower aten_pow_Tensor_Tensor


wonjoolee95:
In order for PyTorch/XLA to support the PyTorch core ATen opset, it requires lowering each core ATen op in PyTorch/XLA. This issue is used to track the PyTorch/XLA lowering for **aten_pow_Tensor_Tensor**.

Here are some general guidelines to lowering this op:
  - Uncomment `@unittest.skip` or `@unittest.expectFailure` and run the unit test at [test_core_aten_ops.py](https://github.com/pytorch/xla/blob/master/test/test_core_aten_ops.py). Eg: `pytest test/test_core_aten_ops.py -k test_aten_pow_Tensor_Tensor_0`
  - Make code changes until the test passes. Read and follow [fix_lowering_for_core_aten_ops.md](https://github.com/pytorch/xla/blob/master/FIX_LOWERING_FOR_CORE_ATEN_OPS.md) for ideas to fix.
    - There may be multiple unit tests for a single op. For this op, the corresponding unit tests are:
      - test_aten_pow_Tensor_Tensor_0
      - test_aten_pow_Tensor_Tensor_1
      - test_aten_pow_Tensor_Tensor_2
    - Please also uncomment the skips for all these tests and ensure all tests are fixed.
    - Note that sometimes the fix may be to fix the unit tests itself. Please take a look at the corresponding unit tests to make sure the tests are valid.
  - Submit the PR!

For any questions, feel free to leave a comment in this PR.

qihqi:
This one looks like a real bug from XLA:

```
In [18]: args2
Out[18]:
(tensor([[8, 8, 8, 9, 3, 6, 8, 2, 9, 7],
         [7, 1, 7, 2, 4, 8, 5, 5, 0, 9],
         [1, 1, 2, 5, 7, 1, 5, 4, 9, 9],
         [2, 9, 5, 5, 0, 9, 4, 5, 0, 2],
         [2, 9, 0, 5, 3, 3, 2, 2, 6, 6],
         [5, 4, 9, 7, 2, 8, 0, 1, 3, 8],
         [1, 4, 6, 9, 4, 0, 6, 4, 7, 9],
         [1, 4, 5, 2, 1, 1, 8, 7, 6, 0],
         [2, 6, 9, 9, 7, 9, 9, 7, 5, 8],
[...]

PR #6092, opened 2023-12-11
Fix precision issue of pow(int, float)

qihqi:
fixes: https://github.com/pytorch/xla/issues/5887

Root cause: we are using dtype of first arg to promote the scalar argument.

### DIFF
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -2090,11 +2090,19 @@ XLATensorPtr pow(const XLATensorPtr& input, const at::Scalar& exponent) {
   // optimize
   torch::lazy::Value exponent_node =
       XLAGraphExecutor::Get()->GetIrValueForConstant(exponent, input->shape());
-  return input->CreateFrom(Pow(input->GetIrValue(), exponent_node));
+  torch::lazy::NodePtr node = Pow(input->GetIrValue(), exponent_node);
+  auto* xla_node = dynamic_cast<XlaNode*>(node.get());
+  at::ScalarType dtype =
+      TorchTypeFromXlaType(xla_node->xla_shape().element_type());
+  return input->CreateFrom(node, dtype);
 }
 
 XLATensorPtr pow(const XLATensorPtr& input, const XLATensorPtr& exponent) {
-  return input->CreateFrom(Pow(input->GetIrValue(), exponent->GetIrValue()));
+  torch::lazy::NodePtr node = Pow(input->GetIrValue(), exponent->GetIrValue());
+  auto* xla_node = dynamic_cast<XlaNode*>(node.get());
+  at::ScalarType dtype =
+      TorchTypeFromXlaType(xla_node->xla_shape().element_type());
+  return input->CreateFrom(node, dtype);
 }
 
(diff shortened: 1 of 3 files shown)


torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -262,11 +262,7 @@ torch::lazy::Value XLAGraphExecutor::GetDeviceDataIrValue(
 torch::lazy::Value XLAGraphExecutor::GetIrValueForConstant(
     const at::Scalar& value, const xla::Shape& shape) {
   torch::lazy::Value ir_value =
-      ScalarOp(std::move(value), shape.element_type());
-  if (!shape.dimensions().empty()) {
-    ir_value = torch::lazy::MakeNode<Expand>(
-        ir_value, torch::lazy::ToVector<int64_t>(shape.dimensions()));
-  }
+      ScalarOp(std::move(value), XlaTypeFromTorchType(value.type()));
```
qihqi, line 265: hi @JackCaoG, please take look at this part, i think it will make our performance better. Basically, binary ops between tensor + scalar doesnt need to expand the scalar to the shape of tensor (as XLA support broadcast of scalar to tensor). 

This bit is not complete though (there are places that we call Promote scattered around) might be good to remove those too.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -2090,11 +2090,19 @@ XLATensorPtr pow(const XLATensorPtr& input, const at::Scalar& exponent) {
   // optimize
   torch::lazy::Value exponent_node =
       XLAGraphExecutor::Get()->GetIrValueForConstant(exponent, input->shape());
-  return input->CreateFrom(Pow(input->GetIrValue(), exponent_node));
+  torch::lazy::NodePtr node = Pow(input->GetIrValue(), exponent_node);
+  auto* xla_node = dynamic_cast<XlaNode*>(node.get());
+  at::ScalarType dtype =
+      TorchTypeFromXlaType(xla_node->xla_shape().element_type());
+  return input->CreateFrom(node, dtype);
```
JackCaoG, line 2097: why do we need to set the logical element type here? It does not impact the HLO graph. I think it is only being used by `XLATensor::dtype()`, 

https://github.com/pytorch/xla/blob/dbd3e33c45e53dc5725a07a3508f0d40d9e7e32e/torch_xla/csrc/tensor.cpp#L159C16-L163

If we don't set it, it will fallback to `xla_node->xla_shape().element_type()` anyway

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: I felt like this is not the right fix, this functions takes a shape and you now ignore the shape. I would suggest make a function that doesn't take shape and replace the usage of this function one by one.

torch_xla/csrc/tensor_methods.cpp
qihqi: If we don't set it; it will use element type of `input`. So the result will get casted from float to int

qihqi:
Moved to https://github.com/pytorch/xla/pull/6103

PR #6093, opened 2023-12-11
Add cuda bm dry run

frgossen:


test/benchmarks/test_experiment_runner.py
```diff
@@ -0,0 +1,46 @@
...
+        "--dry-run"
+    ],
+                           capture_output=True,
+                           text=True)
+    expected_in_stderr = [
+        "'--experiment-config={\"accelerator\": \"cuda\", \"xla\": \"PJRT\", \"xla_flags\": null, \"dynamo\": \"openxla\", \"test\": \"eval\"}', '--model-config={\"model_name\": \"alexnet\"}'",
+        "'--experiment-config={\"accelerator\": \"cuda\", \"xla\": \"PJRT\", \"xla_flags\": null, \"dynamo\": \"openxla\", \"test\": \"train\"}', '--model-config={\"model_name\": \"alexnet\"}'",
+    ]
+    for expected in expected_in_stderr:
+      self.assertTrue(expected in child.stderr)
```
ysiraichi, line 42: This tests if all the configurations in `expected_in_stderr` are in the actual `stderr`. However, it doesn't test the other way around: if `expected_in_stderr` has all the actually generated configurations. I think it would be nice to have that as well.

ysiraichi (changes requested):

test/benchmarks/test_experiment_runner.py
frgossen: Agreed. Added a check for the number of configs to expect.

test/benchmarks/test_experiment_runner.py
```diff
@@ -0,0 +1,52 @@
...
+                           capture_output=True,
+                           text=True)
+    expected_in_stderr = [
+        "Number of selected experiment configs: 4",
+        "Number of selected model configs: 1",
+        "'--experiment-config={\"accelerator\": \"cuda\", \"xla\": \"PJRT\", \"xla_flags\": null, \"dynamo\": \"openxla\", \"test\": \"eval\"}', '--model-config={\"model_name\": \"alexnet\"}'",
+        "'--experiment-config={\"accelerator\": \"cuda\", \"xla\": \"PJRT\", \"xla_flags\": null, \"dynamo\": \"openxla\", \"test\": \"eval\"}', '--model-config={\"model_name\": \"alexnet\"}'",
+        "'--experiment-config={\"accelerator\": \"cuda\", \"xla\": null, \"xla_flags\": null, \"dynamo\": \"inductor\", \"test\": \"eval\"}', '--model-config={\"model_name\": \"alexnet\"}'",
+        "'--experiment-config={\"accelerator\": \"cuda\", \"xla\": null, \"xla_flags\": null, \"dynamo\": \"inductor\", \"test\": \"train\"}', '--model-config={\"model_name\": \"alexnet\"}'",
+    ]
```
ysiraichi, line 46: Not a blocker, just an idea: PyTorch uses [ezyang's expecttest](https://github.com/ezyang/expecttest) package for tests like this. IMO, it makes writing this kind of test easier.

ysiraichi (approved):
lgtm.

test/benchmarks/test_experiment_runner.py
frgossen: Thanks for the pointer. I will try that for the next tests

cota (approved):

PR #6096, opened 2023-12-11
Revert "Allow benchmark runner config: `openxla` for inference runs. …

frgossen:
…(#5939)"

This reverts commit dcdd66e387a2d8ee1ca3352ccca1b53e54a98354.

Introduced openxla to unfold for both training and inference. This is not intended and leads to duplicate experiment runs.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -66,10 +66,9 @@ python xla/benchmarks/experiment_runner.py                   \
 You can change the flags to add the configurations you are interested in. The
 `experiment_runner.py` will expand the options to all supported configurations.
 For example, in the case above, it will consider all the possible combinations
-among the flags `--dynamo`, `--xla`, and `--test`, 5 of which are supported:
+among the flags `--dynamo`, `--xla`, and `--test`, 4 of which are supported:
 
   - `dynamo=openxla_eval`, `xla=PJRT`, `test=eval`
-  - `dynamo=openxla`, `xla=PJRT`, `test=eval`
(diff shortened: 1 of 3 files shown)


frgossen:
Will introduce tests in https://github.com/pytorch/xla/pull/6093

ysiraichi:
>  This is not intended and leads to duplicate experiment runs.

I remember @JackCaoG mentioning we were trying to get rid of `openxla_eval` ([comment](https://github.com/pytorch/xla/pull/5689#discussion_r1355595444)). Is it not the case?

frgossen:
I agree, replacing `openxla_eval` with `openxla` is generally a good idea. This change is unfortunately not doing that exactly.

ysiraichi:
Correct. #5939 only adds `openxla` as a possible backend for `eval` benchmark runs. i.e. reverting that PR would make it not possible to run models in `eval` mode with `openxla` dynamo backend. I guess the real question is: do we want to disallow that?

ysiraichi:
Yes. `openxla_eval` doesn't use AOTAutograd. i.e. it doesn't get a `nn.Module`-traced dynamo graph. This makes the set models that run/fail is also different.

frgossen:
Thanks for explaining that. So we actually want to have these different combinations. Eventually, we want to get rid of the openxla_eval backend. Is that correct?

ysiraichi:
As far as I understand, that's the idea. Maybe @JackCaoG can confirm this.

PR #6101, opened 2023-12-11
[SPMD] Allow mark_shard scalar tensors

alanwaketan:
Summary:
Support mark_shard scalar tensors.

Test Plan:
python test/spmd/test_xla_sharding.py -v -k test_mark_shard_scalar

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1066,8 +1066,24 @@ class BasicShardingTest(test_xla_sharding_base.XlaShardingTest):
     x = torch.tensor(1.0).to(xm.xla_device())
     self.assertEqual(len(x.shape), 0)
 
-    with self.assertRaises(ValueError):
-      xs.mark_sharding(x, self._get_mesh((1, self.n_devices)), ())
+    # with self.assertRaises(ValueError):
+    xt = xs.mark_sharding(x, self._get_mesh((1, self.n_devices)), ())
+    self.assertEqual(xt, x)
+    self.assertEqual(xt.sharding_type, xs.ShardingType.REPLICATED)
+    self.assertEqual(xt.sharding_spec, "{replicated}")
+
+    shards = xt.local_shards
+    self.assertEqual(len(shards), self.n_devices)
+    # all shards are REPLICATED.
+    for i, shard in enumerate(shards):
+      self.assertEqual(shard.data.device, torch.device('cpu'))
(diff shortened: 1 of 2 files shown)


alanwaketan:
Do we plan to support such use case in the future?

JackCaoG:
what's the expected behavior through?

alanwaketan:
> what's the expected behavior through?

This is for open discussion. I would suggest we don't error out but instead ignore the operation if the tensor is a scalar given scalars can only be replicated.

jonb377:
> Do we plan to support such use case in the future?

The only possible sharding on a scalar is pure replication and any unsharded tensors are implicitly replicated, so I don't see a need from the expressibility of the APIs.

I would say it is worth supporting though, so that mark_sharding can act on all tensors without regard for their rank. Right now, a user would need to special-case for rank-0 tensors when iterating over a set of tensors, while we could hide this behind mark_sharding.

alanwaketan:
Okay, I would go ahead to make mark_shard not error out on this operation.

test/spmd/test_xla_sharding.py
```diff
@@ -1047,6 +1047,29 @@ def test_from_cpu_shards_global_shape(self):
...
+    self.assertEqual(len(shards), self.n_devices)
+    # all shards are REPLICATED.
+    for i, shard in enumerate(shards):
+      self.assertEqual(shard.data.device, torch.device('cpu'))
+      self.assertTrue(torch.allclose(shard.data, torch.tensor(1.0)))
+      self.assertIsInstance(shard.indices, type(Ellipsis))
+      self.assertEqual(shard.replica_id, i)
+
+    # It looks like mesh_shape attribute is never implemented.
+    with self.assertRaises(AttributeError):
```
alanwaketan, line 1070: @yeounoh What's the deal here? Should we implement it?

jonb377 (approved):
LGTM! Thanks for resolving this @alanwaketan

alanwaketan:
Thanks Jon for approving the change.

test/spmd/test_xla_sharding.py
yeounoh: Let's keep it unimplemented for now, I will revisit when integrate with DTensor.

yeounoh:
Awesome, thank you @alanwaketan

yeounoh (commented):
LGTM

PR #6110, opened 2023-12-12
Do manual type conversion for isnan op

wonjoolee95:
Fixes https://github.com/pytorch/xla/issues/5872

---

This was actually a PyTorch/XLA feature gap with PyTorch. According to https://pytorch.org/docs/stable/generated/torch.isnan.html, PyTorch allows integral types as input to `torch.isnan`. So doing this works in PyTorch:
```
>>> x = torch.tensor([1, 2, 3], dtype=int)
>>> x
tensor([1, 2, 3])
>>> x.dtype
torch.int64
>>> torch.isnan(x)
tensor([False, False, False])
```

PyTorch/XLA directly calls `xla::IsNan` in our lowering of this op, and XLA only allows floating types for this op. So doing something similar will cause this failure:
```
RuntimeError: Error while lowering: [] aten::isnan, location=__call__@_ops.py:755, xla_shape=pred[10,10]{1,0}, dynamic_dims: ()
XLA builder error: INVALID_ARGUMENT: Operands to IsNan must be real-valued floating-point, but got S32: 
```

This change does a manual type conversion to F32 if this op is given an integral type.

### DIFF
--- a/test/test_core_aten_ops.py
+++ b/test/test_core_aten_ops.py
@@ -2246,7 +2246,6 @@ class AtenOpTest(unittest.TestCase):
     kwargs = dict()
     run_export_and_compare(self, torch.ops.aten.isnan, args, kwargs)
 
-  @unittest.skip
   def test_aten_isnan_2(self):
     args = (torch.randint(0, 10, (10, 10)).to(torch.int32),)
     kwargs = dict()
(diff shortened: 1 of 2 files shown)


wonjoolee95:
@JackCaoG, this is probably not too significant, but just calling it out to see if we should cherry-pick this onto the 2.2 branch.

JackCaoG:
yea we can backport this if it is a feature gap

qihqi (approved):

PR #6117, opened 2023-12-12
Add comments in benchmark_experiment and generalize workaround for too long strings in `filename_str`

frgossen:
`filename_str` ignored `accelerator_model` and `xla_flags`, which will likely cause silent problems when running experiments comparing different flag combinations. To avoid long filenames, which is why flags were not included, we use a hash instead. This also generalized to all properties listed in the configs.

### DIFF
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -4,7 +4,7 @@ import os
 import torch
 import torch._dynamo as dynamo
 import torch_xla.core.xla_model as xm
-from util import is_xla_device_available, get_accelerator_model
+from util import parse_none_str, is_xla_device_available, get_accelerator_model
 
 logger = logging.getLogger(__name__)
 
@@ -14,21 +14,9 @@ class ExperimentLoader:
   def __init__(self, args):
     self._args = args
 
-  def expand_config_choices(self, config_choices):
-    configs = [{}]
-
-    for key, choices in config_choices.items():
-      tmp_configs = []
-      for config in configs:
-        for choice in choices:
-          tmp_config = config.copy()
-          tmp_config[key] = choice
-          tmp_configs.append(tmp_config)
-      configs = tmp_configs
-
-    return configs
-
   def list_experiment_configs(self):
+
+    # Start with default config.
     config_choices = {
         "accelerator": ["cpu", "cuda", "tpu"],
         "xla": [None, "PJRT", "XRT"],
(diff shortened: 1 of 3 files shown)


benchmarks/benchmark_experiment.py
```diff
@@ -114,34 +116,32 @@ def _add_experiment_env(self, experiment_config):
...
       elif is_xla_device_available("CUDA"):
         process_env["GPU_NUM_DEVICES"] = "1"
-    elif not experiment_config["xla"] and is_xla_device_available(
-        experiment_config["accelerator"].upper()):
+    elif experiment_config["xla"] is None:
       # In non-xla CPU training experiments, an env var is still needed if an
       # xla device exists, or there will be "Missing XLA configuration" error.
-      process_env["PJRT_DEVICE"] = experiment_config["accelerator"].upper()
+      if is_xla_device_available(experiment_config["accelerator"].upper()):
+        process_env["PJRT_DEVICE"] = experiment_config["accelerator"].upper()
```
ysiraichi, line 123: Is this because, even if we aren't using XLA, we still have to populate `PJRT_DEVICE` env var? If so, wouldn't it be better just to:

```python
...
else:
    process_env["PJRT_DEVICE"] = "CPU"  # should be available everywhere
```

benchmarks/benchmark_experiment.py
```diff
@@ -46,33 +36,44 @@ def list_experiment_configs(self):
...
+      config_choices["dynamo"] = list(
+          map(parse_none_str, set(self._args.dynamo)))
     if self._args.test:
       config_choices["test"] = list(set(self._args.test))
     if self._args.xla_flags:
-      config_choices["xla_flags"] = [
-          x if x != "None" else None for x in list(set(self._args.xla_flags))
-      ]
+      config_choices["xla_flags"] = map(parse_none_str,
+                                        set(self._args.xla_flags))
```
ysiraichi, line 51: Missing: `list(map(...), ...)` ?

frgossen: Yes, ty

benchmarks/benchmark_experiment.py
frgossen: The change here merely restructures the code to switch on the xla value. 

The reason for setting PJRT_DEVICE is in the comment "In non-xla CPU training experiments, an env var is still needed if an xla device exists, or there will be Missing XLA configuration error."

I am guessing it would work with setting PJRT_DEVICE to CPU always but even then I think might be cleaner to only set it when there is an XLA device. I can change it but can't really test that case

vanbasten23:
Can you elaborate on `generalize workaround for too long strings in filename_str`? What problem do you see?

frgossen:
> Can you elaborate on `generalize workaround for too long strings in filename_str`? What problem do you see?

Fully ignoring the XLA flags can be confusion when experimenting with different sets of XLA flags, e.g. to test priority-based fusion.

frgossen:
> Can you elaborate on `generalize workaround for too long strings in filename_str`? What problem do you see?

Extended the PR description.

cota (approved):

vanbasten23 (approved):

PR #6119, opened 2023-12-12
Enforce successful imports and cause failure if not set up correctly

frgossen:


cota (approved):

frgossen:
@zpcore , just want to make sure this does not cause any trouble for you before merging it.

zpcore:
Thanks for the clean up, any special actions we need to take in order to set up correctly with this change?

frgossen:
For me it works just like this but I know we had different experiences at times. Let me know if this becomes an issue. You can always add the directory to the python path

PR #6125, opened 2023-12-12
Extract fingerprinting for experiment and model config

frgossen:
Extract fingerprinting, into a function, which was anonymously implemented in two places before. This is used to resume benchmarking from where it stopped. Configurations are identified by their fingerprint.

Clearly distinguish between the subprocess and the main process in the code, trying to make it more readable.

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -44,97 +44,114 @@ class ExperimentRunner:
     self.output_file = os.path.join(self.output_dir, self._args.output_basename)
 
   def run(self):
-    if self._args.experiment_config and self._args.model_config:
-      if self._args.dry_run:
-        logger.warning(f"Dry run with {[sys.executable] + sys.argv}")
-        return
-      experiment_config = json.loads(self._args.experiment_config)
-      model_config = json.loads(self._args.model_config)
-      self.run_single_experiment(experiment_config, model_config)
+    is_main_process = self._args.experiment_config is None and self._args.model_config is None
+    if is_main_process:
+      self.generate_and_run_all_configs()
     else:
-      assert not self._args.experiment_config and not self._args.model_config
-      finished_experiments = set()
-      if os.path.exists(self.output_file):
-        if self._args.no_resume:
-          os.unlink(self.output_file)
-        else:
-          with open(self.output_file, mode="r", encoding="utf-8") as f:
-            jsonlines = f.read().splitlines()
-          for jsonline in jsonlines:
-            tmp = json.loads(jsonline)
-            # the finished experiment batch_size may be altered by model set_up(),
-            # so the dummy experiment will not match it
-            tmp["experiment"]["batch_size"] = self._args.batch_size
-            finished_experiments.add("-".join(
-                str(item) for item in (list(tmp["model"].values()) +
-                                       list(tmp["experiment"].values()))))
+      assert self._args.experiment_config is not None and self._args.model_config is not None
+      self.run_single_config()
 
-      experiment_configs = self.experiment_loader.list_experiment_configs()
-      model_configs = self.model_loader.list_model_configs()
-      logger.warning(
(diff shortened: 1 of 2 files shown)


benchmarks/util.py
```diff
@@ -41,7 +41,7 @@ def reset_rng_state(benchmark_experiment=None):
   torch.manual_seed(1337)
   random.seed(1337)
   np.random.seed(1337)
-  if benchmark_experiment and benchmark_experiment.xla:
+  if benchmark_experiment is not None and benchmark_experiment.xla:
```
cota, line 44: `if foo` is the same as `if foo is not None`, and AFAICT is more idiomatic. Is there a good reason not to leave this as is?

benchmarks/experiment_runner.py
```diff
@@ -44,97 +44,113 @@ def __init__(self, args):
...
 
   def run(self):
-    if self._args.experiment_config and self._args.model_config:
-      if self._args.dry_run:
-        logger.warning(f"Dry run with {[sys.executable] + sys.argv}")
-        return
-      experiment_config = json.loads(self._args.experiment_config)
-      model_config = json.loads(self._args.model_config)
-      self.run_single_experiment(experiment_config, model_config)
+    is_entry_process = self._args.experiment_config is None and self._args.model_config is None
```
vanbasten23, line 47: by "entry_process", do you mean the main process?

benchmarks/experiment_runner.py
```diff
@@ -44,97 +44,113 @@ def __init__(self, args):
...
-          else:
-            e = "SKIP because of incompatible model and experiment configs."
-            logger.warning(e)
+      assert self._args.experiment_config is not None and self._args.model_config is not None
+      self.run_single_config()
+
+  def generate_and_run_all_configs(self):
+    assert self._args.experiment_config is None and self._args.model_config is None
+
+    # Collect fingerprints for configs to skip. These are configs for which we already have results.
```
vanbasten23, line 57: What's a "fingerprint"?

benchmarks/experiment_runner.py
```diff
@@ -144,19 +160,16 @@ def run_single_experiment(self, experiment_config, model_config):
...
                                              benchmark_model)
         output = move_to_device(output, 'cpu')
         outputs.append(output)
         for key, val in run_metrics.items():
-          # metrics from repeated runs are formed into lists in the metrics dict
           if key not in metrics:
             metrics[key] = []
           metrics[key].append(val)
 
-    # additional experiment metrics can be added here
```
vanbasten23, line 158: Can we leave the comment as is, if it's still valid?

vanbasten23:
Would you mind adding more context about what this PR is trying to achieve in the pr description?

benchmarks/util.py
frgossen: I'm not sure what is more idiomatic but it is def not the same. `foo` will evaluate to false for a lot of things, e.g. the empty string, an empty list, ... 
With `is not None` the comparison is stricter and less ambiguous to read (?)

benchmarks/experiment_runner.py
frgossen: Yes, changed the name.

benchmarks/experiment_runner.py
frgossen: Added a comment.

benchmarks/util.py
cota: Yeah I meant in this case you'll get the same result. Indeed one check is more specific than the other.

FWIW when I see "if foo is not None" or similar, that makes me think that foo might indeed be `0`, `[]`, `{}`, etc. In this case I don't think that can happen, hence my suggestion for what I believe is more idiomatic. But in any case this is a nit.

frgossen:
> Would you mind adding more context about what this PR is trying to achieve in the pr description?

Ofc, added a description.

cota (approved):

PR #6133, opened 2023-12-13
Apply default precision from bench models to XLA

golechwierowicz:
This PR introduces a change fixing a disparity between instructions' precision of XLA, and TorchInductor.

For example:
After calling torch.half() or any other buffer/parameter casting TorchInductor will cast all non-fp16 floating types to fp16, and program input shape will correctly be reflected by this. OTOH XLA needs additional environment variables to perform casting upon lowering.
This PR reads the default precision set by TorchBench models for inference, and training and sets the appropriate env vars.


cota (approved):

cota:
Thanks Greg! Please land this and I'll run another nightly with it today.

benchmarks/torchbench_model.py
```diff
@@ -209,6 +211,25 @@ def set_up(self):
     del benchmark
     gc.collect()
 
+  def apply_default_precision_config(self, test, benchmark):
```
frgossen, line 214: Thanks for finding this and adding it to the benchmarks. 
I am wondering if this is the right place to fix this. 
Is this something we can add in the dynamo bridge when lowering to XLA?

cota: This is a workaround. See b/316126405

frgossen: I wonder if we should add these things to benchmarking or if that is just more confusing overall. We should at least add a comment in these cases.

PR #6076, opened 2023-12-08
Fix model initialization.

ysiraichi:
Fix: #6006 #6007 #6011

This PR introduces a set of models that must be first initialized on CUDA, before moving them to XLA. Even though it fixes all those issues, the majority still have some other problems that do not allow them to run successfully:

- `moco`: still fails because it fallbacks to CPU on `allgather` operation
- `nvidia_deeprecommender`: still fails on training
- `pytorch_CycleGAN_and_pix2pix`: PASS
- `simple_gpt`: requires BF16 (same as inductor)
- `simple_gpt_tp_manual`: requires BF16 (same as inductor)
- `timm_efficientdet`: OOM on training (RTX 2060)

cc @JackCaoG @miladm

PR #6134, opened 2023-12-13
Revert "Fix model initialization. (#6076)"

frgossen:
This reverts commit 7786d5dfe92bf4d2ab97989916b8c353a414b52a.

`python benchmarks/experiment_runner.py --dynamo=openxla --dynamo=openxla_eval --dynamo=inductor --xla=PJRT --xla=None --test=eval --test=train --suite-name=torchbench --accelerator=cuda  --filter=^alexnet$  --no-resume` fails for openxla benchmarks.

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -5,8 +5,10 @@ import os
 from os.path import abspath, exists
 import sys
 import torch
-import torch.utils._pytree as pytree
-from torch._dynamo.testing import reduce_to_scalar_loss
+import torch.nn as nn
+from torch._dynamo.testing import collect_results, reduce_to_scalar_loss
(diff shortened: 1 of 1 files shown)


zpcore:
Sorry I didn't catch up with recent PRs. Do we already support automatically match dynamo backend `openxla`, `openxla_eval` with `train` and `eval` based on the cmdline in the description?

frgossen:
We should support openxla_eval + eval, openxla + eval, and openxla + train 
This issue is unrelated though. The flags are expanded to configs, so if the combination is not supported it should just skip it. 
The failure I see is a segfault which did not happen before.

zpcore (approved):
Thanks for the clarification. LGTM! Can you also resolve the conflict?

PR #6138, opened 2023-12-13
benchmarks/aggregate: remove timestamp from "latest" title

cota:
In the nightly scripts, we will use the --title flag to pass in the date.


frgossen (approved):
Thanks!

PR #6139, opened 2023-12-13
Do not fail on lack of default precision set.

golechwierowicz:
I discovered some models from the suite do not have the default precision set so instead of failing the script we just log the case, and do nothing, as no additional machinery should run for the Inductor anyway. Additionally I wrapped the exceptions with the ValueError so the logging message will not pollute with info about str not inheriting from Exception class.

@cota , note that needs to be hooked "somewhere". Not sure where, as there was a revert in https://github.com/pytorch/xla/pull/6134, but in general it can be done prior to moving the model to the device safely.

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -209,18 +209,20 @@ class TorchBenchModel(BenchmarkModel):
     elif test == "train" and hasattr(benchmark, 'DEFAULT_TRAIN_CUDA_PRECISION'):
       precision = benchmark.DEFAULT_TRAIN_CUDA_PRECISION
     else:
-      raise f"Unkown test type {test}!"
+      logger.warning("No default precision set. No patching needed.")
+      return
 
     if precision == "fp16":
       os.environ['XLA_USE_FP16'] = '1'
     elif precision == "amp":
-      raise f"AMP for PT/XLA:GPU is not implemented yet for torchbench models"
+      raise ValueError(
+          f"AMP for PT/XLA:GPU is not implemented yet for torchbench models")
     elif precision == "bf16":
       os.environ['XLA_USE_BF16'] = '1'
     elif precision == "fp32":
       logger.warning("Sticking with the default fp32 precision.")
     else:
-      raise f"Unknown precision: {precision}"
+      raise ValueError(f"Unknown precision: {precision}")
 
   def pick_grad(self):
     # special case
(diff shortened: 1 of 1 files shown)


cota:
Thanks Greg -- I had missed that your change was reverted as part of another revert.

Please land this (including the PR description when merging, otherwise annoyingly we lose that info in the repo) and I'll try to re-land your reverted change.

cota (approved):

benchmarks/torchbench_model.py
```diff
@@ -209,18 +209,20 @@ def apply_default_precision_config(self, test, benchmark):
     elif test == "train" and hasattr(benchmark, 'DEFAULT_TRAIN_CUDA_PRECISION'):
```
frgossen, line 209: Can you add a comment that this is a workaround and should really be fixed in the bridge?

frgossen (approved):
One comment. 
Thanks you for finding these thinsg!

benchmarks/torchbench_model.py
golechwierowicz: Done in https://github.com/pytorch/xla/pull/6139/commits/2791a22abdf55a30766c0b9f343612a386cee532.

cota:
> Please land this (including the PR description when merging, otherwise annoyingly we lose that info in the repo) and I'll try to re-land your reverted change.

Given that Jack is looking into this at the bridge level, I'll wait for him to finish that investigation since that's the right place to fix this.
Thanks again for identifying this issue Greg!

benchmarks/torchbench_model.py
ysiraichi: @frgossen @golechwierowicz Why do we need to set an environment variable for that? Is it related to https://github.com/pytorch/pytorch/issues/115792?

benchmarks/torchbench_model.py
```diff
@@ -204,23 +204,36 @@ def set_up(self):
...
-      raise f"Unkown test type {test}!"
+      logger.warning("No default precision set. No patching needed.")
+      return
 
     if precision == "fp16":
       os.environ['XLA_USE_FP16'] = '1'
     elif precision == "amp":
-      raise f"AMP for PT/XLA:GPU is not implemented yet for torchbench models"
+      raise ValueError(
+          f"AMP for PT/XLA:GPU is not implemented yet for torchbench models")
```
ysiraichi, line 230: @miladm @frgossen @golechwierowicz Isn't AMP supported on PT/XLA:GPU?

PR #6140, opened 2023-12-13
Initialize compilation env hash

jonb377:
The compilation env hash needs to be initialized. This was apparently lost during a rebase of the original PR #5803.


will-cromar (approved):

jonb377:
Local build & test succeeds - merging to unblock @will-cromar and #5677

PR #6148, opened 2023-12-14
Don't explcitly include virtual device in hash

jonb377:
As attempted in #6140, this addresses an issue where the compilation env hash was never initialized. However, this change decouples the virtual device initialization logic from the runtime, allowing `xr.use_spmd` to be called after the runtime has been initialized, which won't break the unit tests.


will-cromar (approved):
This makes sense to me. From the runtime's perspective, SPMD is not a permanent setting. We could choose in the framework to flip back and forth if we wanted to

will-cromar (approved):

yeounoh (approved):
LGTM

PR #6155, opened 2023-12-14
Remove Colab notebooks

will-cromar:
We no longer support TPU Node on Colab in current releases.


JackCaoG:
what happened to people that still use old release and use colabs?

will-cromar:
The old colab notebooks will still be available in our old release branches. I also removed links to these from our Cloud documentation a while ago. I think it's more confusing to keep them around on master when current releases don't support it.

JackCaoG (approved):
Let's add them back when colab supports TPUVM

PR #6156, opened 2023-12-14
Dump pytorch profiles and traces per repeat iteration 

frgossen:
Also generalize the mechanism to dump result files. We can use this to dump other metrics soon.

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -11,6 +11,7 @@ import sys
 import time
 import torch
 import tiers
+from typing import Optional
 import torch_xla.debug.metrics as met
 from tqdm import tqdm
 from torch.profiler import profile, ProfilerActivity
@@ -28,13 +29,13 @@ class ExperimentRunner:
 
   def __init__(self, args):
     self._args = args
-    self.suite_name = self._args.suite_name
 
     self.experiment_loader = ExperimentLoader(self._args)
 
-    if self.suite_name == "torchbench":
+    # Choose model loader.
+    if self._args.suite_name == "torchbench":
       self.model_loader = TorchBenchModelLoader(self._args)
-    elif self.suite_name == "dummy":
+    elif self._args.suite_name == "dummy":
       self.model_loader = ModelLoader(self._args)
     else:
       raise NotImplementedError
@@ -118,6 +119,8 @@ class ExperimentRunner:
 
         # Launch subprocess.
         try:
+          # TODO: See if we can generalize this for all env vars. The experiment
+          # config is currently a bit bloated.
           process_env = benchmark_model.extend_process_env(process_env)
           command = [sys.executable] + sys.argv + [
               f"--experiment-config={json.dumps(experiment_cfg)}"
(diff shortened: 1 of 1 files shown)


benchmarks/util.py
```diff
@@ -41,7 +41,7 @@ def reset_rng_state(benchmark_experiment=None):
   torch.manual_seed(1337)
   random.seed(1337)
   np.random.seed(1337)
-  if benchmark_experiment and benchmark_experiment.xla:
+  if benchmark_experiment is not None and benchmark_experiment.xla is not None:
```
vanbasten23, line 44: if benchmark_experiment and benchmark_experiment.xla are empty, you would do the same? If so, why not keeping it as is as it's more pythonic?

vanbasten23:
Do you think we need to update the README?

golechwierowicz:
So the nice thing about the current approach is that we got one file with a summary of CPU/CUDA times (note the "a" flag for open function). This is quite convenient, because you can open one file and it shows all the iterations. Is it possible to preserve this behaviour? So for summaries we'd have one file, but for traces separate files (we're now preserving only the latest trace, thanks for fixing this!).

benchmarks/experiment_runner.py
```diff
@@ -429,187 +489,170 @@ def parse_log_level(level: str):
...
           logging.CRITICAL,
           logging.ERROR,
           logging.WARNING,
           logging.INFO,
           logging.DEBUG,
       ],
-      type=parse_log_level,
+      type=_parse_log_level,
       help="Specify log level.",
   )
```
ysiraichi, line 502: Maybe it's not an issue for this PR, but I'd think we want string choices instead of integers. A minimal program with only this arguments gives:

```python
usage: log.py [-h] [-l {50,40,30,20,10}]

options:
  -h, --help           show this help message and exit
  -l {50,40,30,20,10}  Specify log level.
```

benchmarks/util.py
frgossen: This is from a base PR. Sorry, I'm still learning how to use github reviews properly.

I'm still curious now and will shamelessly exploit this opportunity to learn something :) Why would `if benchmark_experiment and benchmark_experiment.xla` be more pythonic? 

I don't have a strong opinion on this but my thinking is this: benchmark_experiment is of type `Optional[BenchmarkExperiment]` so I would like to easily distinguish between the `None` case and the `BenchmarkExperiment` case.

frgossen:
> Do you think we need to update the README?

This is not currently documented in the readme. I expect to add more dumps so it might be a good idea to document when this settles.

frgossen:
> So the nice thing about the current approach is that we got one file with a summary of CPU/CUDA times (note the "a" flag for open function). This is quite convenient, because you can open one file and it shows all the iterations. Is it possible to preserve this behaviour? So for summaries we'd have one file, but for traces separate files (we're now preserving only the latest trace, thanks for fixing this!).

Yeah, I saw the "a", so no dump was lost there. I think it did accumulate dumps over multiple runs though, which is not intended, I assume. I will add the accumulation behaviour back so we get both views.

benchmarks/experiment_runner.py
frgossen: Agreed, that does not look nice. Will address that in a follow-up

frgossen: Addressed in https://github.com/pytorch/xla/pull/6188

golechwierowicz (approved):

PR #6159, opened 2023-12-14
benchmarks/aggregate: distinguish between openxla and openxla_eval fo…

cota:
…r inference

While at it, remove an openxla data line from a6000's training CSV to test that the corresponding empty field is reported.


frgossen (approved):
Thanks!

PR #6070, opened 2023-12-08
Enable CUDA support for PyTorch on CI.

ysiraichi:
Re-opened from: #6061

PR #6160, opened 2023-12-14
Move tests that need CUDA to `test_operations.py`.

ysiraichi:
This PR moves tests introduced in #6060 and #6071 from `test_core_aten_ops.py` to `test_operations.py`. This change is needed since `test_core_aten_ops.py` executes only with `PJRT_DEVICE=CPU`.

### DIFF
--- a/test/test_core_aten_ops.py
+++ b/test/test_core_aten_ops.py
@@ -10,19 +10,6 @@ import tempfile
 import unittest
 
 
-def onlyIfTorchSupportsCUDA(fn):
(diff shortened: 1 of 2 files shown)


JackCaoG (approved):

ysiraichi:
I'm temporarily turning CUDA support for PyTorch on, so that we can run these tests. This is causing problems, as suggested in #6076.

ysiraichi:
Maybe it'll be better to wait for #6070

PR #6162, opened 2023-12-14
Unify the way in which result files are dumped

frgossen:
Both PyTorch profiles and output files are now dumped in the same way. This gives us records per repeat iteration.

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -113,8 +113,7 @@ class ExperimentRunner:
         if not self.model_loader.is_compatible(benchmark_model,
                                                benchmark_experiment):
           logger.warning("SKIP incompatible model and experiment configs.")
-          self.save_results(benchmark_experiment, benchmark_model,
-                            {"error": "SKIP"}, None)
+          self._save_results(experiment_cfg, model_cfg, {"error": "SKIP"})
           continue
 
         # Launch subprocess.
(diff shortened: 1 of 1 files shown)


benchmarks/util.py
```diff
@@ -41,7 +41,7 @@ def reset_rng_state(benchmark_experiment=None):
   torch.manual_seed(1337)
   random.seed(1337)
   np.random.seed(1337)
-  if benchmark_experiment and benchmark_experiment.xla:
+  if benchmark_experiment is not None and benchmark_experiment.xla is not None:
```
vanbasten23, line 44: If benchmark_experiment and benchmark_experiment.xla are empty, would you do the same thing? If so, you can keep it as is as `if benchmark_experiment and benchmark_experiment.xla`.

frgossen: This is from a base PR. Sorry, I'm still learning how to use github reviews properly.

I'm still curious now and will shamelessly exploit this opportunity to learn something :) Why would if benchmark_experiment and benchmark_experiment.xla be more pythonic?

I don't have a strong opinion on this but my thinking is this: benchmark_experiment is of type Optional[BenchmarkExperiment] so I would like to easily distinguish between the None case and the BenchmarkExperiment case.

benchmarks/experiment_runner.py
```diff
@@ -144,20 +143,17 @@ def generate_and_run_all_configs(self):
         except subprocess.TimeoutExpired as e:
           self._fwd_captured_stdout_stderr(e.stdout, e.stderr)
           logger.error("TIMEOUT")
-          self.save_results(benchmark_experiment, benchmark_model,
-                            {"error": str(e)}, None)
+          self._save_results(experiment_cfg, model_cfg, {"error": str(e)})
         except subprocess.CalledProcessError as e:
           self._fwd_captured_stdout_stderr(e.stdout, e.stderr)
```
golechwierowicz, line 148: Should this line be also under the last two exceptions?

frgossen: I think no because the last two exceptions mean that no subprocess was launched, so there would not be anything to fwd.

cota (approved):

PR #6163, opened 2023-12-14
benchmarks: save timestamp in json output

cota:
So that benchmarks run from the same invocation get the same timestamp.

I'm adding timestamps to old JSONL output files with this script:

```
import json
import sys

def replace_timestamps(filename, timestamp):
  with open(filename) as read_file:
    for line in read_file:
      try:
        r = json.loads(line.rstrip('\n|\r'))
      except json.JSONDecodeError as e:
        sys.exit(f'Invalid JSONL:\n{line}{e}')
      if 'timestamp' not in r:
        r['timestamp'] = float(timestamp)
      print(json.dumps(r))

def main():
  if len(sys.argv) != 3:
    sys.exit("Usage: add_timestamp.py input_file timestamp")
  filename = sys.argv[1]
  timestamp = sys.argv[2]
  replace_timestamps(filename, timestamp)

if __name__ == '__main__':
  main()
```

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -121,7 +121,11 @@ class ExperimentRunner:
           process_env = benchmark_model.extend_process_env(process_env)
           command = [sys.executable] + sys.argv + [
               f"--experiment-config={json.dumps(experiment_cfg)}"
-          ] + [f"--model-config={json.dumps(model_cfg)}"]
+          ] + [f"--model-config={json.dumps(model_cfg)}"] + [
+              # Note: if "--timestamp foo" is already in sys.argv, we
+              # harmlessly pass "--timestamp foo" again here.
+              f"--timestamp={self._args.timestamp}"
+          ]
           command_str = " ".join(command)
           logger.debug(f"Run `{command_str}`")
           child_process = subprocess.run(
@@ -206,6 +210,7 @@ class ExperimentRunner:
 
     results["metrics"] = metrics
     results["outputs_file"] = outputs_file_name
+    results["timestamp"] = self._args.timestamp
 
     json_str = json.dumps(results, ensure_ascii=False)
     with open(self.output_file, mode="a", encoding="utf-8") as f:
@@ -612,6 +617,13 @@ def parse_args(args=None):
       type=str,
       help="JSON string defining the model configuration. When set an experiment is run with exactly this one configuration.",
   )
+
+  parser.add_argument(
+      "--timestamp",
+      default=time.time(),
+      type=float,
+      help="Timestamp (seconds since the epoch) to assign to the benchmarks.")
+
   return parser.parse_args(args)
 
 
--- a/benchmarks/result_analyzer.py
+++ b/benchmarks/result_analyzer.py
@@ -108,6 +108,8 @@ class ResultAnalyzer:
     runs = []
     for jsonline in jsonlines:
       dataline = json.loads(jsonline)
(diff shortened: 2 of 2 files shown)


benchmarks/result_analyzer.py
```diff
@@ -255,8 +259,8 @@ def parse_args(args=None):
 
   parser.add_argument(
```
frgossen, line 260: Do we still need this flag?

benchmarks/experiment_runner.py
```diff
@@ -175,6 +178,7 @@ def save_results(self, benchmark_experiment, benchmark_model, metrics,
 
     results["metrics"] = metrics
     results["outputs_file"] = outputs_file_name
+    results["timestamp"] = self.timestamp
```
frgossen, line 181: Think you can get it from the args here.

benchmarks/experiment_runner.py
```diff
@@ -28,6 +28,7 @@ class ExperimentRunner:
 
   def __init__(self, args):
     self._args = args
+    self.timestamp = args.timestamp or time.time()
```
frgossen, line 31: Don't think we need this as a member variable.

benchmarks/experiment_runner.py
```diff
@@ -90,6 +91,8 @@ def run(self):
           command = ([sys.executable] + sys.argv +
                      [f"--experiment-config={experiment_config_str}"] +
                      [f"--model-config={model_config_str}"])
+          if not self._args.timestamp:
+            command += [f"--timestamp={self.timestamp}"]
```
frgossen, line 95: This function (`run`) is probably the right place to generate the timestamp.

frgossen (changes requested):
Thanks for fixing this! 
Can we wait landing this to avoid merge conflicts with what I have in flight?

benchmarks/result_analyzer.py
cota: It's useful to have it if we're processing old json files that have no timestamps in them.
It's true though that eventually this flag should probably go away.

benchmarks/experiment_runner.py
cota: (Answering your other comments for this file here.)
This is a bit subtle so maybe I should have added a comment.
We could always use args.timestamp and set time.time() as its default value. Let's call that approach A.
But then in this hunk we might pass --timestamp twice to the child process; think for instance when the original invocation had --timestamp passed to it. It's no big deal (it would be --timestamp foo --timestamp foo), but I thought it would be cleaner not to do that.
Alternatives would be:
Approach B: Look for --timestamp in sys.argv. I really wouldn't bother doing this.
Approach C: The approach you see here. That is, have a --timestamp flag that defaults to None so that we know we have to pass --timestamp to the subprocess. Note that the member variable (or equivalent) is necessary in this case, since from save_results we need to get the timestamp and --timestamp cannot be set to a default [unless we're OK possibly passing --timestamp twice to the child].

I'm OK with either A or B. Let me know if you have a preference.

frgossen: Got it. I had assumed it was propagated through the metrics. Now that you outlined all of these options, I actually like approach A. Happy with either option (except for B maybe)

cota: OK, will do approach A, which, BTW, is what I had implemented originally x_x

frgossen (approved):
Thanks!

PR #6168, opened 2023-12-14
Add default v5 flags

jonb377:
Adds default values for TPU-specific XLA flags on v5e and v5p.

Verifying performance with a v5e run of Llama2 70B.

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -51,7 +51,7 @@ def _setup_libtpu_flags():
         # Limit compiler-injected rematerialization
         'xla_jf_rematerialization_percent_shared_memory_limit': '10000',
         # Enable collective matmul
-        'xla_jf_spmd_threshold_for_windowed_einsum_mib': '0'
+        'xla_jf_spmd_threshold_for_windowed_einsum_mib': '0',
         # Enable async collective fusions
         'xla_tpu_enable_async_collective_fusion': 'true',
         'xla_tpu_enable_async_collective_fusion_fuse_all_gather': 'true',
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -42,6 +42,27 @@ def _setup_libtpu_flags():
   # and thus worse performance.
   flags = _set_missing_flags(flags,
                              (('xla_latency_hiding_scheduler_rerun', '1'),))
+
+  if tpu.version() == 5:
+    default_v5_flags = {
+        # Enable async collectives
+        'xla_enable_async_all_gather': 'true',
+        'xla_enable_async_collective_permute': 'true',
+        # Limit compiler-injected rematerialization
+        'xla_jf_rematerialization_percent_shared_memory_limit': '10000',
+        # Enable collective matmul
+        'xla_jf_spmd_threshold_for_windowed_einsum_mib': '0'
+        # Enable async collective fusions
+        'xla_tpu_enable_async_collective_fusion': 'true',
+        'xla_tpu_enable_async_collective_fusion_fuse_all_gather': 'true',
+        'xla_tpu_enable_async_collective_fusion_multiple_steps': 'true',
+        # Disable net router
+        'xla_tpu_enable_net_router_in_all_gather': 'false',
+        # Disable experimental Reduce+Broadcast->ReduceWindow-Conv fusion
+        'xla_tpu_rwb_fusion': 'false',
+    }
+    flags = _set_missing_flags(flags, default_v5_flags.items())
+
   os.environ['LIBTPU_INIT_ARGS'] = ' '.join(flags)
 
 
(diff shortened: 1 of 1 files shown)


JackCaoG:
do we need these for 2.2 release?

jonb377:
> do we need these for 2.2 release?

It would be nice to include so v5 is performant out-of-the-box on 2.2

torch_xla/__init__.py
```diff
@@ -42,6 +42,27 @@ def _setup_libtpu_flags():
...
   flags = _set_missing_flags(flags,
                              (('xla_latency_hiding_scheduler_rerun', '1'),))
+
+  if tpu.version() == 5:
+    default_v5_flags = {
+        # Enable async collectives
+        'xla_enable_async_all_gather': 'true',
+        'xla_enable_async_collective_permute': 'true',
+        # Limit compiler-injected rematerialization
+        'xla_jf_rematerialization_percent_shared_memory_limit': '10000',
```
alanwaketan, line 52: Do we need this? I don't think MaxText adds this one as well.

torch_xla/__init__.py
```diff
@@ -42,6 +42,27 @@ def _setup_libtpu_flags():
...
+        # Enable collective matmul
+        'xla_jf_spmd_threshold_for_windowed_einsum_mib': '0',
+        # Enable async collective fusions
+        'xla_tpu_enable_async_collective_fusion': 'true',
+        'xla_tpu_enable_async_collective_fusion_fuse_all_gather': 'true',
+        'xla_tpu_enable_async_collective_fusion_multiple_steps': 'true',
+        # Disable net router
+        'xla_tpu_enable_net_router_in_all_gather': 'false',
```
alanwaketan, line 60: Should we drop this one?

torch_xla/__init__.py
```diff
@@ -42,6 +42,27 @@ def _setup_libtpu_flags():
...
+        # Disable experimental Reduce+Broadcast->ReduceWindow-Conv fusion
+        'xla_tpu_rwb_fusion': 'false',
```
alanwaketan, line 62: Does this really give us any performance boost?

alanwaketan (commented):
Thanks Jon for making this change. I have a few questions regarding to some of the flags.

torch_xla/__init__.py
alanwaketan: For reference: https://github.com/google/maxtext/blob/d2bd833a8b21dd9fd2f02832112dbb4a21b26f69/MaxText/configs/experimental/32b.sh#L32

alanwaketan:
BTW, do we need flags for MultiSlice?

jonb377:
Thanks @alanwaketan - let me try matching the MaxText flags exactly. A lot of these were recommended by the XLA team prior to your optimization barrier change, which may make them unnecessary.

jonb377:
I saw ~1.5% performance degradation using the MaxText flags and when omitting the three specified. I think the main driver of this is the rematerialization flag. Trying another run with the remat flag set.

torch_xla/__init__.py
jonb377: I tried a run without all three flags mentioned, and the absolute MFU reduction was 1.5%. Including this one increases MFU by 1%. I have another run to identify which of the other two flags accounts for the other 0.5% gain.

jonb377:
> BTW, do we need flags for MultiSlice?

Comparing to the MaxText flags, the only MultiSlice-related flag without the default value only relates to scanned compilation. I don't think we need to include any in the defaults here, but there can still be gains from tuning flags for the specific MultiSlice environment for a given workload.

alanwaketan:
> I saw ~1.5% performance degradation using the MaxText flags and when omitting the three specified. I think the main driver of this is the rematerialization flag. Trying another run with the remat flag set.

@jonb377 Have you figured out where the ~1.5% comes from?

jonb377:
I found the following:
- Adding only `xla_jf_rematerialization_percent_shared_memory_limit` back yields 1% improvement
- Removing only `xla_tpu_enable_net_router_in_all_gather` from this set yields ~2% improvement (which is actually more performant than our baseline)
- `xla_tpu_rwb_fusion` is inconclusive from the tests I've run so far, but doesn't seem to do much.

I'll do one more test removing only `xla_tpu_rwb_fusion`.

JackCaoG:
This is not really a bug fix, if we want to get this in for 2.2, I would like to get it merge ASAP so we have enough time for testing.

jonb377:
cc @alanwaketan, let's be conservative on the flags then and land today - I'll update this PR to only enable async collectives. There's also still an issue with async AG fusion in the current libtpu pin. Users can always specify `LIBTPU_INIT_ARGS` for the best perf.

JackCaoG (approved):

Issue #5966, opened 2023-12-01
[torchbench] AssertionError: All tensors should be on xla


ysiraichi:
## 🐛 Bug

Running a few torchbench benchmarks, using dynamo+openxla backend, ends up in an assertion failure:

```python
Traceback (most recent call last):
  File "xla/benchmarks/experiment_runner.py", line 601, in <module>
    main()
  File "xla/benchmarks/experiment_runner.py", line 597, in main
    runner.run()
  File "xla/benchmarks/experiment_runner.py", line 65, in run
    self.run_single_experiment(experiment_config, model_config)
  File "xla/benchmarks/experiment_runner.py", line 161, in run_single_experiment
    run_metrics, output = self.timed_run(benchmark_experiment,
  File "xla/benchmarks/experiment_runner.py", line 328, in timed_run
    output = loop()
  File "xla/benchmarks/experiment_runner.py", line 310, in loop
    output = benchmark_model.model_iter_fn(
  File "torch/_dynamo/eval_frame.py", line 410, in _fn
    return fn(*args, **kwargs)
  File "xla/benchmarks/benchmark_model.py", line 154, in eval
    pred = self.module(*inputs)
  File "torch/nn/modules/module.py", line 1510, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "torch/nn/modules/module.py", line 1519, in _call_impl
    return forward_call(*args, **kwargs)
  File "/lib/python3.10/site-packages/detectron2-0.6-py3.10-linux-x86_64.egg/detectron2/modeling/meta_arch/rcnn.py", line 150, in forward
    return self.inference(batched_inputs)
  File "/lib/python3.10/site-packages/detectron2-0.6-py3.10-linux-x86_64.egg/detectron2/modeling/meta_arch/rcnn.py", line 203, in inference
    images = self.preprocess_image(batched_inputs)
  File "/lib/python3.10/site-packages/detectron2-0.6-py3.10-linux-x86_64.egg/detectron2/modeling/meta_arch/rcnn.py", line 229, in preprocess_image
    images = ImageList.from_tensors(
  File "/lib/python3.10/site-packages/detectron2-0.6-py3.10-linux-x86_64.egg/detectron2/structures/image_list.py", line 58, in from_tensors
    @staticmethod
  File "torch/_dynamo/eval_frame.py", line 410, in _fn
    return fn(*args, **kwargs)
  File "torch/_dynamo/external_utils.py", line 17, in inner
[...]

PR #6170, opened 2023-12-15
Ignore non-XLA nodes and their direct dependents.

ysiraichi:
Fix: #5966 

This PR generalizes `FallBackNodeCollector` into `UnsupportedNodesCollector`, improving and solving a few issues the old implementation had:

- nodes which resulted in a non-XLA tensor weren't flagged as fallback
- nodes whose arguments were a container with non-XLA tensors weren't flagged as fallback (e.g. `stack`)
- tracking these "unsupported" nodes weren't really only fallback nodes
    - but, nodes that can't exist in partition boundaries (e.g. arguments and return values)

### DIFF
--- a/test/dynamo/test_bridge.py
+++ b/test/dynamo/test_bridge.py
@@ -266,6 +266,17 @@ class TorchXLAReuseGraphTest(torch._dynamo.test_case.TestCase):
 
     self._compile_and_check(foo, (xm.xla_device(),))
 
+  def test_cpu_input(self):
+
+    def foo(xt, t):
+      # Cannot move t to XLA, since it's also an output.
+      return (xt * 5)[t], t
+
+    device = xm.xla_device()
+    xt = torch.rand(5, device=device)
+    t = torch.randint(0, 5, (3,))
+    self._compile_and_check(foo, (xt, t))
+
 
 if __name__ == "__main__":
   from torch._dynamo.test_case import run_tests
--- a/test/dynamo/test_bridge.py
+++ b/test/dynamo/test_bridge.py
@@ -266,7 +266,7 @@ class TorchXLAReuseGraphTest(torch._dynamo.test_case.TestCase):
 
     self._compile_and_check(foo, (xm.xla_device(),))
 
-  def test_cpu_input(self):
+  def test_cpu_index_and_output(self):
 
     def foo(xt, t):
       # Cannot move t to XLA, since it's also an output.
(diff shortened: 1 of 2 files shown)


ysiraichi:
cc @JackCaoG @miladm

ysiraichi:
Note: this PR can, possibly, create more partitions than before, affecting performance. Maybe we should make sure there are no regressions before actually landing it.

JackCaoG:
I think the definition of the fallback is that `we have to execute this op on fallback devices(uusually)` cpu. for the cases you mentioned
1. nodes which resulted in a non-XLA tensor weren't flagged as fallback
Does the operation being executed on cpu or gpu/tpu?

2. nodes whose arguments were a container with non-XLA tensors weren't flagged as fallback (e.g. stack)
I guess this one is technically a fallback, since op will be executed on cpu. It will show up in dynamo fallback messages, but shouldn't be in the XLA's metrics `aten` counter I think?

I don't fully understand what 3 means, can you give ma an example?

ysiraichi:
The problem was that `FallBackNodeCollector` was doing more than just collecting operations that were executed on CPU. It was also flagging nodes that had non-XLA tensors as fallback (even though they were executed on XLA).

**Reason:** my guess is that it did so, in order to guarantee that the input/output of the generated partitions were all XLA tensors.

Assuming that's the case, it still missed the 2 cases I mentioned:

- **nodes which result in non-XLA tensor:** might end up as output of some partition
- **nodes whose arguments were a container with non-XLA tensors:** arguments might end up as the input of some partition

**Solution:**

- disambiguate fallback from unsupported nodes (from the perspective of the partitioner)
- add the missing 2 cases

@JackCaoG let me know what you think.

JackCaoG:
yea that make sense, is this pr ready for review?

ysiraichi:
Yes, it is.

test/dynamo/test_bridge.py
```diff
@@ -266,6 +266,25 @@ def foo(device):
 
     self._compile_and_check(foo, (xm.xla_device(),))
 
+  def test_cpu_index_and_output(self):
+
+    def foo(xt, t):
+      # Cannot move t to XLA, since it's also an output.
+      return (xt * 5)[t], t
```
JackCaoG, line 273: I am not fully following what this test is trying to test. Can you make comment more explicit? For example without this change how/why will it fail?

torch_xla/core/dynamo_bridge.py
```diff
@@ -420,30 +420,51 @@ def optimized_mod(*args):
...
     else:
-      # if inputs are non-xla tensors, it should be executed on CPU
-      if n.op in ["call_function", "call_module", "call_method"]:
-        args, kwargs = self.fetch_args_kwargs_from_env(n)
-        for arg in args:
-          if isinstance(arg, torch.Tensor) and not is_xla_tensor(arg):
-            self._fallback_ops.append(n)
-            break
+      # Check whether the tensors contained in value are all XLA tensors.
+      def are_all_tensors_xla(value):
```
JackCaoG, line 437: nit  maybe `all_tensors_on_xla_device`?

JackCaoG (commented):
mostly lgtm, minor questions on test cases.

JackCaoG (approved):

PR #6177, opened 2023-12-15
Catch exception from AMP mismatch.

golechwierowicz:
There're a couple of models (yolov3, pytorch_unet, tacotron2) which have AMP mode enabled. If we run the entire benchmarking suite they will crash the main process though. In this PR we catch this, and save the results to the output file.

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -152,6 +152,9 @@ class ExperimentRunner:
           logger.error("ERROR when launching child process")
           self.save_results(benchmark_experiment, benchmark_model,
                             {"error": str(e)}, None)
+        except ValueError as e:
+          self._fwd_captured_stdout_stderr(e.stdout, e.stderr)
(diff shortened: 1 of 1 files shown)


benchmarks/experiment_runner.py
```diff
@@ -125,6 +125,10 @@ def generate_and_run_all_configs(self):
             logger.error("ERROR")
             self.save_results(dummy_benchmark_experiment, dummy_benchmark_model,
                               {"error": str(e)}, None)
+          except ValueError as e:
+            logger.error("ERROR")
```
frgossen, line 129: Can you log the error e, too?

frgossen (approved):
Thanks for making it more robust! One comment

benchmarks/experiment_runner.py
golechwierowicz: Yes, I used the `logger.exception` to do it automagically.

PR #6186, opened 2023-12-15
experiment_runner: revamp options to dump PyTorch/XLA metrics

frgossen:
Currently enabling profiling requires two steps: (1) enabling profiling with `--profile-cuda`, and (2) choose one (or both) options to dump the profiling into: `--profile-cuda-cpu-collect` to dump to the output JSONL file and/or `--profile-cuda-dump` to point to a directory where profiling output files are written to.

This PR revamps these options by replacing these flags with new ones to specify finer-grained profiling, e.g. we can now just `--dump-pytorch-xla-metrics` without dumping pytorch profiles (`--dump-pytorch-profiles`). Moreover, recording profiling info to the JSONL output file (if any profiling is enabled) becomes mandatory and therefore `--profile-cuda-cpu-collect` goes away.

While at it, add comments on metrics collection and group them in the code.

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -18,9 +18,10 @@ from enum import Enum
 from torch.profiler import profile, ProfilerActivity
 from torch.autograd import DeviceType
 from benchmark_model import ModelLoader
+from enum import Enum
 from torchbench_model import TorchBenchModelLoader
 from benchmark_experiment import ExperimentLoader
-from util import reset_rng_state, move_to_device, randomize_input
+from util import reset_rng_state, move_to_device, randomize_input, us_to_s, ns_to_s
 import torch_xla.core.xla_model as xm
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


benchmarks/experiment_runner.py
```diff
@@ -183,22 +188,145 @@ def run_single_config(self):
...
+        # Mark step.
+        self._mark_step(benchmark_experiment)
+        if pytorch_profile is not None:
+          pytorch_profile.step()
+
+      self._synchronize(benchmark_experiment)
+      return output
+
+    # Execute all iterations (with) profiling.
+    enable_pytorch_profiling = self._args.dump_pytorch_profiles or \
```
golechwierowicz, line 253: FYI previous semantics was that you had to run with `--profile-cuda`, and then choose one (or two) of both options: collecting numbers to the `results.jsonl` file, and/or dumping traces with kernel summaries. 

Probably we can get rid of the first one.

golechwierowicz (approved):

.vscode/settings.json
```diff
@@ -18,5 +18,5 @@
         "./bazel-out/_coverage/_coverage_report.dat"
     ],
     "python.formatting.provider": "yapf",
-    "editor.formatOnSave": true
+    "editor.formatOnSave": false
```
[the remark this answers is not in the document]
cota: Fixed, thanks

cota (approved):

benchmarks/experiment_runner.py
cota: Done.

PR #6182, opened 2023-12-15
Re-land: Fix model initialization.

ysiraichi:
Re-land: #6076

cc @JackCaoG @miladm

PR #6195, opened 2023-12-18
Fix Inductor runs.

golechwierowicz:
After the flag `--dynamo=inductor` is issued it breaks because device is uninitialised.


frgossen (commented):
Thanks! We all ran into it late Friday

frgossen (approved):

frgossen:
Was introduced by https://github.com/pytorch/xla/pull/6182

PR #6152, opened 2023-12-14
Apply precision config env vars in the root process.

golechwierowicz:
After some changes to the main branch, `os.environ` was not sufficient to pick up new env vars in the subprocess.
In this PR we apply a necessary workaround in the root process which launches subprocess per each experiment. New flags are passed via `process_env` var.
I tried to keep the `experiment_runner.py` as clean as possible, and abstracted the new env vars via `apply_env` method of `{Benchmark, TorchBench}Model`.

Tested with
`PJRT_DEVICE=CUDA python3 new_xla/benchmarks/experiment_runner.py --dynamo=openxla --xla=PJRT --test=eval --filter='hf_Bert$|BERT_pytorch$' --suite-name=torchbench --accelerator=cuda --progress-bar --output-dirname=/tmp/output --repeat=3 --print-subprocess --no-resume --profile-cuda --profile-cuda-dump=/tmp/dumpz --profile-cuda-cpu-collect`

For `hf_Bert` the fp16 casting works, and it shows in the gemm kernels.

```
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------
                                                   Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg     Self CUDA   Self CUDA %    CUDA total  CUDA time avg    # of Calls
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------
                                    triton_gemm_dot_918         0.00%       0.000us         0.00%       0.000us       0.000us     436.000us        17.83%     436.000us      36.333us            12
                                     triton_gemm_dot_75         0.00%       0.000us         0.00%       0.000us       0.000us     299.000us        12.23%     299.000us      27.182us            11
void cutlass::Kernel<cutlass_80_tensorop_f16_s16816g...         0.00%       0.000us         0.00%       0.000us       0.000us     288.000us        11.78%     288.000us      24.000us            12
                                    triton_gemm_dot_128         0.00%       0.000us         0.00%       0.000us       0.000us     256.000us        10.47%     256.000us     256.000us             1
                                    triton_gemm_dot_872         0.00%       0.000us         0.00%       0.000us       0.000us     144.000us         5.89%     144.000us      12.000us            12
[...]

PR #6199, opened 2023-12-18
benchmarks/torchbench_model: skip benchmarks that fail to load

cota:
In "dfcf306e7 Apply precision config env vars in the root process. (#6152)"
we started running load_benchmark() from experiment_runner's
main process. Unfortunately, load_benchmark() for
some models does exit the calling process, which results
in experiment_runner exiting prematurely.

Work around this issue by adding these models to the deny list,
so that experiment_runner does not die early.

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -56,6 +56,8 @@ DENY_LIST = {
             "accelerator": "tpu"
         },
     ],  # not implemented
+    # self.load_benchmark() exits the main process. See issue #6207.
+    "pytorch_CycleGAN_and_pix2pix": [{}],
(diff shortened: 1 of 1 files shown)


frgossen (approved):
Thanks! If this is a wider spread issue, we should investigate further. Can we add an issue for this one?

cota:
`pytorch_unet` also needs to be added. Will update the PR once the run continues -- there might be more additions to the deny list needed.
Will add an issue once I gather them all.

cota:
I've added the necessary models to the deny list as a workaround. I've also filed issue #6207 and referenced it from the code change so that we do not forget why they were added to the deny list.

golechwierowicz (approved):

PR #6202, opened 2023-12-18
Rename `TransferToServer` to `TransferToDevice`

will-cromar:
There is no more "server" now that we've removed XRT.


JackCaoG (approved):

jonb377 (approved):
Thanks Will!

PR #6210, opened 2023-12-19
benchmarks/aggregate: skip first three data points

cota:
Skipping the first iteration is a good idea but doesn't seem to be sufficient to get reasonable stddev. For instance, yolov3 is quite noisy in the first few iterations, not just the first one:
- Inductor training numbers (in seconds): `79.075, 0.607, 0.115, 0.106, 0.100`
- Pytorch/XLA training: `99.074, 5.638, 5.871, 0.384, 0.367`

I believe it's because the first iteration compiles, and the subsequent two warm up the caches (must be a large model since Inductor takes ~80s to compile it).

In order to reduce stddev in the reports, skip the first three data points.


benchmarks/aggregate.py
```diff
@@ -94,13 +94,17 @@ def process_file(args, results_map: Dict[str, Any], filename: str):
...
       total_times = r['metrics']['total_time'] if 'total_time' in r[
           'metrics'] else []
-      if len(total_times) <= 1:
+      # Skip the first three elements in total_times[]; the first one
+      # includes compilation time, the 2nd and 3rd warm up the caches.
+      skip_elems = 3
+      if len(total_times) <= skip_elems:
         continue
-      # Skip total_times[0] because it includes compilation time.
-      avg = np.average(total_times[1:])
```
golechwierowicz, line 100: George suggested using minimums, it's a worthy path to explore.

golechwierowicz (approved):
We should get minim

benchmarks/aggregate.py
cota: Not sure I understand -- what does "using mininums" mean in this context?
Something we could do is to first discard outliers using statistics (e.g. anything too far beyond a certain amount of standard deviations). After that whether we picked the avg+std or something else should not make a big difference.

Issue #6207, opened 2023-12-19
benchmarks/torchbench_model: some benchmarks fail to load and kill experiment_runner's main process


cota:
## 🐛 Bug

In `dfcf306e7 Apply precision config env vars in the root process.` (#6152)
we started running load_benchmark() from experiment_runner's
main process. Unfortunately, load_benchmark() for
some models does exit the calling process.
This results in experiment_runner exiting prematurely.

## To Reproduce

Try to run under XLA any of the benchmarks added to the deny list in #6199. For example:

```
python xla/benchmarks/experiment_runner.py --dynamo=openxla --dynamo=openxla_eval --xla=PJRT --test=eval --test=train --accelerator=cuda --output-dirname=/tmp/pix2pix --repeat=5 --print-subprocess --suite-name=torchbench --filter='^pytorch_CycleGAN_and_pix2pix$' --log-level=debug ; echo $?
```

Note: `pytorch_CycleGAN_and_pix2pix` also fails early under inductor.

## Expected behavior

The above should print a `0` exit code regardless of whether the benchmark fails to run or not. However, it prints `2`.

## Environment

 - Reproducible on XLA backend [CPU/TPU]: GPU
 - torch_xla version: dfcf306e7 and later.

yeounoh:
@cota thanks for addressing this, can we close this issue now?

cota:
We're working around this issue by temporarily disabling the affected benchmarks. But AFAICT this is still an issue. If you want we can close it -- I won't be working on this in the near future.
Maybe @ysiraichi will? Let's have him decide what to do with this issue.

ysiraichi:
I'm only running `dynamo+openxla` tests, here. And, so far, I can't really reproduce these loading errors:

- `pytorch_CycleGAN_and_pix2pix`: runs eval and train successfully
- `pytorch_unet`: runs eval and train successfully if [we don't throw on AMP](https://github.com/pytorch/xla/blob/259b7dcd072b31134e5b5b26f5e8276b036b3f50/benchmarks/torchbench_model.py#L351-L353)
- `tacotron2`: eval and train fail at execution time (not when it's loading the model)

PR #6226, opened 2023-12-21
benchmarks/torchbench_model: add more benchmarks that fail to load

cota:
Add Inductor to the deny list for two benchmarks.

Note: I should have added these already in "3ccb4ed2a benchmarks/torchbench_model: skip benchmarks that fail to load (#6199)". This fixes that oversight.

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -69,12 +69,8 @@ DENY_LIST = {
     "pytorch_struct": [{
         "test": "eval"
     },],  # not implemented
-    "pytorch_unet": [
-        {
-            # self.load_benchmark() exits the main process. See issue #6207.
-            "xla": "PJRT",
-        },
-    ],
+    # self.load_benchmark() exits the main process. See issue #6207.
+    "pytorch_unet": [{}],
     "resnet50_quantized_qat": [
         {
             "test": "eval",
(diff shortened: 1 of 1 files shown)


benchmarks/torchbench_model.py
```diff
@@ -69,12 +69,8 @@
     "pytorch_struct": [{
         "test": "eval"
     },],  # not implemented
-    "pytorch_unet": [
-        {
-            # self.load_benchmark() exits the main process. See issue #6207.
-            "xla": "PJRT",
-        },
-    ],
+    # self.load_benchmark() exits the main process. See issue #6207.
```
vanbasten23, line 72: Since `"xla": "PJRT",` has been removed, do you still need the comment `# self.load_benchmark() exits the main process. See issue #6207.`?

benchmarks/torchbench_model.py
```diff
@@ -69,12 +69,8 @@
...
+    "pytorch_unet": [{}],
```
vanbasten23, line 73: Does the change mean pytorch_unet can be run on torch_xla?

vanbasten23 (approved):

benchmarks/torchbench_model.py
cota: Yes. The initialization problem still exists, so the comment remains valid. Initialization fails also for Inductor, which is why we're skipping this model entirely.

benchmarks/torchbench_model.py
cota: (This is a deny list, not an allow list.)

cota: See https://github.com/pytorch/xla/pull/6199 for more details on the problem.

cota:
Turns out these errors on Inductor are coming from the AMP mismatch, which raises an exception. Handling this exception is the right fix for this problem -- see #6229 which supersedes this PR.

PR #6228, opened 2023-12-21
Pin expecttest version (#6221)

zpcore:
Cherry pick the pin update to resolve xl-ml test failure.

### DIFF
--- a/test/tpu/xla_test_job.yaml
+++ b/test/tpu/xla_test_job.yaml
@@ -41,7 +41,7 @@ spec:
     - bash
     - -cxe
     - |
-      pip install expecttest
+      pip install expecttest==0.1.6
       pip install rich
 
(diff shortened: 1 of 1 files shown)


JackCaoG:
this only helps the TPU CI through, it won't fix xlml test. they used a different config file I believe.

JackCaoG (approved):

ManfeiBai (approved):
LGTM

zpcore:
> this only helps the TPU CI through, it won't fix xlml test. they used a different config file I believe.

Oh you are right, should we just drop this PR? I will fix in the xl-ml test script.

JackCaoG:
it doesn't hurt to merge this one in, but if we don't run TPUCI on r2.2 then it doesn't really fix anything either

zpcore:
OK, let me merge this PR. Thanks

PR #6229, opened 2023-12-21
experiment_runner: catch AMP exception when building child environment

cota:
In "30458aaae Catch exception from AMP mismatch. (#6177)" we forgot to also handle the AMP mismatch exception when preparing the child environment. Fix it now so that the main process does not die.

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -122,7 +122,13 @@ class ExperimentRunner:
         # Compose child process environment.
         process_env = os.environ.copy()
         benchmark_experiment.update_process_env(process_env)
-        benchmark_model.update_process_env(process_env)
+        try:
+          benchmark_model.update_process_env(process_env)
+        except ValueError as e:
+          logger.error(f"ERROR preparing child env: {e}")
+          self._save_results(benchmark_experiment.to_dict(),
+                             benchmark_model.to_dict(), {"error": str(e)})
+          continue
 
         # Setup HLO dumps.
         if self._args.dump_hlo:


cota:
@golechwierowicz related question: does it make sense to raise the exception even on Inductor runs or only for `if xla`? IOW, could we correctly run Inductor with AMP using experiment_runner, or not? I ask because we used to, but perhaps AMP was not honored even for Inductor?

zpcore:
Thanks for the fixing. May I know in what case will `update_process_env` raise the value error?

miladm (approved):
LGTM

zpcore (approved):
LGTM!

cota:
> Thanks for the fixing. May I know in what case will `update_process_env` raise the value error?

https://github.com/pytorch/xla/blob/22a6e6ea134cc29e49885b9ea294a01ec09fc9d4/benchmarks/torchbench_model.py#L258
This method gets called from the main process for every benchmark. A few of them use AMP; Pytorch/XLA does not yet support it and we raise that exception that we're now catching to keep going (and log the benchmark error). See #6133 and the subsequent fix #6177; the PR here is yet another fix (hopefully the last one for this issue).

golechwierowicz:
> @golechwierowicz related question: does it make sense to raise the exception even on Inductor runs or only for `if xla`? IOW, could we correctly run Inductor with AMP using experiment_runner, or not? I ask because we used to, but perhaps AMP was not honored even for Inductor?

AMP was honored by the Inductor, internally it correctly sets up the AMP context. For PT/XLA we did not because it requires a different handling of the context (there's a different set of methods to use).

PR #6233, opened 2023-12-22
Measure GPU wall time without launch overheads.

golechwierowicz:
This is useful for scoping up the single graph, and seeing how compiler is able to optimise it.
This is is optional i.e not added by default. Once we start setting up another benchmark suite with profiling info etc. this should be run additionally, but separately so CUPTI interface won't interact with pure CUDA events.


cota (approved):

benchmarks/experiment_runner.py
```diff
@@ -242,41 +271,57 @@ def run_once_and_gather_metrics(self, benchmark_experiment, benchmark_model,
...
+            'Run with pure wall time, but also with profiling flags enabled. Falling back to a default wall time.'
+        )
       with profile(activities=[ProfilerActivity.CUDA, ProfilerActivity.CPU
                               ]) as pytorch_profile:
-        output = loop(pytorch_profile)
+        output, _ = loop(pytorch_profile, iter_fn=self._default_iter_fn)
     else:
-      output = loop()
+      if self._args.pure_wall_time:
+        output, pure_wall_timing = loop(iter_fn=self._pure_wall_time_iter_fn)
```
zpcore, line 307: bench.py is intended for benchmarking on CUDA, please prohibit the TPU xla routing from entering the `_pure_wall_time_iter_fn`.

benchmarks/experiment_runner.py
```diff
@@ -192,6 +191,36 @@ def _fwd_captured_stdout_stderr(self, stdout_text: str, stderr_text: str):
...
+    # result.
+    if benchmark_experiment.xla:
+      t_trace_start = time.perf_counter()
+    output = benchmark_model.model_iter_fn(
+        input_tensor, collect_full_output=self._args.collect_full_output)
+    if benchmark_experiment.xla:
+      tracing_time = time.perf_counter() - t_trace_start
+
+    # Mark step.
+    self._mark_step(benchmark_experiment)
```
zpcore, line 208: There is a redundant _mark_step in line 286.

PR #6235, opened 2023-12-26
Remove or simplify hardcoded lists of device types

will-cromar:
Allows new device plugins without registering the device type in our code! See #6242 for broader context of this cleanup.

- Add placeholder `PLUGIN` device type for unknown devices
  - Store the actual device name (according to `PjRtClient::platform_name`) in the `DeviceType`, since we still need to pass the device name string across Python/C++ boundary
  - Replace explicit lists of devices with patterns
- Deprecate `devkind` argument, since we only support one PJRT backend at a time
- Remove explicit references to device types that do not have any special cases (XPU, ROCM)
- Remove deprecated `GPU` device type
- Simplify several test skip conditions that had lists of device types

Tested with CPU plugin in #6253 where I change the platform name to `TEST` to simulate an unknown device type. `test_operations.py` passes with that plugin.

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -72,7 +72,7 @@ def is_xla_tensor(tensor):
 
 
 def parse_xla_device(device):
-  m = re.match(r'(CPU|TPU|GPU|ROCM|CUDA|XPU|NEURON):(\d+)$', device)
+  m = re.match(r'([A-Z]+):(\d+)$', device)
   if m:
     return (m.group(1), int(m.group(2)))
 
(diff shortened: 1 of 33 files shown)


vanbasten23 (approved):
LGTM. Thanks for the simplification!

torch_xla/core/xla_model.py
```diff
@@ -72,7 +72,7 @@ def is_xla_tensor(tensor):
 
 
 def parse_xla_device(device):
-  m = re.match(r'(CPU|TPU|GPU|ROCM|CUDA|XPU|NEURON):(\d+)$', device)
+  m = re.match(r'([A-Z]+):(\d+)$', device)
```
JackCaoG, line 75: trying to think.. where would it fail if user has a typo on `PJRT_DEVICE`?

torch_xla/csrc/device.cpp
```diff
@@ -9,39 +9,49 @@
...
 
-thread_local absl::optional<torch::lazy::BackendDevice> g_current_device;
+// This is set when any device is initialized, so to prevent using non-virtual
+// device and virtual device together.
+static bool spmd_config_is_locked = false;
+
+}  // namespace
+
+std::string DeviceType::XlaDeviceTypeToString(XlaDeviceType hw_type) {
+  XLA_CHECK(hw_type != XlaDeviceType::PLUGIN) << "PLUGIN type name unknown";
```
JackCaoG, line 19: when will this function be called? Are we not expceting this to be called on plug in?

torch_xla/csrc/device.h
```diff
@@ -15,15 +15,27 @@ namespace torch_xla {
...
+  DeviceType(XlaDeviceType xla_device_type)
+      : torch::lazy::BackendDeviceType(static_cast<int>(xla_device_type)),
+        type_name_(XlaDeviceTypeToString(xla_device_type)) {}
+  DeviceType(const std::string& type_name)
+      : torch::lazy::BackendDeviceType(
+            static_cast<int>(StringToXlaDeviceType(type_name))),
+        type_name_(type_name) {}
+
+  // TODO(wcromar): do we even want this default constructor?
+  DeviceType() : DeviceType(XlaDeviceType::CPU){};
```
JackCaoG, line 30: I felt like we should eventually let PyTorch/XLA to auto detect the device based on some rules(libtpu, cuda version etc). Can't really think of a case we want to default construct a CPU device...

torch_xla/csrc/random.cpp
```diff
@@ -20,9 +20,7 @@ std::string GetDefaultGitGeneratorName() {
   XlaDeviceType hw_type =
       static_cast<XlaDeviceType>(bridge::GetCurrentDevice().type());
   switch (hw_type) {
-    case XlaDeviceType::GPU:
     case XlaDeviceType::CUDA:
```
JackCaoG, line 23: can't we throw a explicitly message when people set `PJRT_DEVUCE=GPU`, this is not backward compatible so I want to make sure user knows how to fix it.

JackCaoG: ok I saw below we say we want to remove that warning after 2.2.. fine haha

JackCaoG (approved):

torch_xla/core/xla_model.py
will-cromar: The device name here comes from the `PjRtClient::platform_name`, which is not guaranteed to match the `PJRT_DEVICE` name.

If there's a typo in `PJRT_DEVICE`, the runtime will fail to initialize and throw an error.

torch_xla/csrc/device.cpp
will-cromar: This function is used when constructing a device from the `XlaDeviceType` enum type rather than the device name in a string. E.g. here:https://github.com/pytorch/xla/blob/9a4ef68a1ff5b53c1f793d9b2370836a29dc93dc/torch_xla/csrc/xla_backend_impl.cpp#L183-L184

If all we get is the placeholder `XlaDeviceType::PLUGIN` type, then we don't know the actual `platform_name` of the device type for `toString`. I think this is a relatively rare case in our code, so I can try to remove it. I have a more ambitious refactoring effort going in https://github.com/pytorch/xla/pull/6261.

torch_xla/csrc/device.h
will-cromar: I'm 90% sure this constructor is never called. Logically, the "default" device type is managed by `xla_backend_impl`.

torch_xla/csrc/runtime/ifrt_computation_client.h
```diff
@@ -71,6 +71,11 @@ class IfrtComputationClient : public ComputationClient {
 
   std::string GetDefaultDevice() const override;
 
+  torch_xla::DeviceType GetDeviceType() const override {
+    return torch_xla::DeviceType(
+        absl::AsciiStrToUpper(client_->platform_name()));
```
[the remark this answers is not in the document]
will-cromar: Yeah

torch_xla/csrc/device.h
will-cromar: Actually, we definitely don't need this. Upstream provides a better default constructor: https://github.com/pytorch/pytorch/blob/16d69290c6d037a25e32220b9517597d04dbd0bf/torch/csrc/lazy/backend/backend_device.cpp#L14-L15

I'm just going to delete this now instead of leaving the TODO.

PR #6240, opened 2023-12-27
Experimental GPU PJRT Plugin

will-cromar:
Add experimental CUDA plugin package under `plugins/`.

Manually tested that single device initialization works as expected. Almost nothing else works right now, so not including any CI tests.


plugins/cuda/pyproject.toml
```diff
@@ -0,0 +1,18 @@
+[build-system]
```
vanbasten23, line 1: What's the toml file used for?

will-cromar: This is a replacement for `setup.py` https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

JackCaoG (approved):

vanbasten23 (approved):

PR #6247, opened 2024-01-02
Set all_reduce_token to None when exiting

### LINKED ISSUE #6246: SIGSEGV when exiting the dataloader in the middle of training
A SIGSEGV error occurs when running the following code with `PJRT_DEVICE=CUDA torchrun --nproc_per_node=2 test.py`.
```
import torch
import torch_xla.distributed.parallel_loader as pl

import os
os.environ['PJRT_LOCAL_PROCESS_RANK'] = os.environ['LOCAL_RANK']

device = xm.xla_device()
xm.set_replication(device, [device])
import torch_xla.utils.utils as xu
train_loader = xu.SampleGenerator(
      data=torch.zeros(1, 12),
      sample_count=1024)
train_loader = pl.MpDeviceLoader(train_loader, device)
max_steps = 10
for step, inputs in enumerate(train_loader):
  xm.all_reduce('sum', [inputs], scale=1.0/xm.xrt_world_size())

  if step > max_steps: break
```

This is due to early exit from the dataloader, causing the `xm.mark_step` in pl.MpDeviceLoader to not be executed. As a result, `all_reduce_token` is not set to None, which causes `all_reduce_token` in the global variable `g_all_reduce_tokens` in `torch_xla/csrc/cross_replica_reduces.cpp` to be released only when the program exits.

yitongh:
This pr fixes #6246.

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -148,6 +148,8 @@ _aws_ec2_inf_trn_init()
 
 
 def _prepare_to_exit():
+  device = _XLAC._xla_get_default_device()
+  _XLAC._set_all_reduce_token(device, None)
   _XLAC._prepare_to_exit()
   if int(os.environ.get('PT_XLA_DEBUG', '0')):
     _summarize_fn_tracker()
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -235,6 +235,7 @@ function run_mp_op_tests {
   run_test "$CDIR/test_mp_save.py"
   run_test "$CDIR/test_mp_mesh_reduce.py"
   run_test "$CDIR/test_mp_sync_batch_norm.py"
+  run_test "$CDIR/test_mp_early_exit.py"
   run_pt_xla_debug "$CDIR/debug_tool/test_mp_pt_xla_debug.py"
   run_xla_backend_mp "$CDIR/test_torch_distributed_all_gather_xla_backend.py"
   run_xla_backend_mp "$CDIR/test_torch_distributed_all_reduce_xla_backend.py"
--- a/test/test_mp_early_exit.py
+++ b/test/test_mp_early_exit.py
@@ -1,0 +1,26 @@
+import sys
+import torch
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.parallel_loader as pl
+import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla.utils.utils as xu
+
+
+def _mp_fn(index):
+  device = xm.xla_device()
+  if xm.xla_device_hw(device) in ('TPU', 'GPU', 'CUDA', 'ROCM', 'NEURON'):
+    train_loader = xu.SampleGenerator(
+        data=torch.zeros(1, 12), sample_count=1024)
+    train_loader = pl.MpDeviceLoader(train_loader, device)
+    max_steps = 10
+    for step, inputs in enumerate(train_loader):
+      xm.all_reduce('sum', [inputs], scale=1.0 / xm.xrt_world_size())
+      if step > max_steps:
+        break
+  else:
+    print(f'{device} is not a TPU or GPU device', file=sys.stderr)
+
+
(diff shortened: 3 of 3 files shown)


JackCaoG:
I am curious why does reduce_token matters when we exit the pytorch/xla?

JackCaoG:
oh ok I saw https://github.com/pytorch/xla/issues/6246, my bad. Can you add the test case you mentioned in the issue to a separate test? An example would be https://github.com/pytorch/xla/blob/master/test/test_mp_collective_permute.py which will be run on https://github.com/pytorch/xla/blob/d7c443085f56423bd306eb007882a7cb6f1ac19f/test/run_tests.sh#L231

yitongh:
@JackCaoG I have added a test case, but I cannot run it. The program hangs after setting GPU_NUM_DEVICES=2. Other mp test cases also hang.

JackCaoG:
let's see if CI will be able to run it then.

JackCaoG:
CI seems to be happy, if you can fix the linter I can help you land it.

yitongh:
@JackCaoG I have fixed the linter, sorry forget this.

JackCaoG (approved):

vanbasten23:
> @JackCaoG I have added a test case, but I cannot run it. The program hangs after setting GPU_NUM_DEVICES=2. Other mp test cases also hang.

This seems to be a new issue. I have also observed the same https://github.com/pytorch/xla/issues/6320. Before this change, all the mp test cases don't hang. @yitongh @JackCaoG

JackCaoG:
@vanbasten23 does this pr breaks multi device GPU training or just the test?

JackCaoG:
@vanbasten23 let's revert his pr. @ManfeiBai can you also help reverting this in the 2.2 release branch? Release date is approaching, I don't want to take this risk.

torch_xla/__init__.py
```diff
@@ -148,6 +148,8 @@ def _setup_tpu_vm_library_path() -> bool:
 
 
 def _prepare_to_exit():
+  device = _XLAC._xla_get_default_device()
```
alanwaketan, line 151: Does this always return the same device regardless of different processes in the pool?

JackCaoG: I think it will return the device that belong to current process, assuming each process only has one device.

alanwaketan: Then I'm not sure why it will break other test cases...

yitongh: I think the error is due to the ComputationClient has exited in atexit._run_exitfuncs when using xla_multiprocessing, while the client still exists when using torchrun. Therefore, in the xla_multiprocessing, this invocation will result in the creation of a new PjRtComputationClient, causing a hang.

 So the better solution is setting the all reduce token in PrepareToExit, like this:
```
diff --git a/torch_xla/__init__.py b/torch_xla/__init__.py
index 8d4997e28..d753f8f7c 100644
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -148,8 +148,6 @@ _aws_ec2_inf_trn_init()


 def _prepare_to_exit():
-  device = _XLAC._xla_get_default_device()
-  _XLAC._set_all_reduce_token(device, None)
   _XLAC._prepare_to_exit()
   if int(os.environ.get('PT_XLA_DEBUG', '0')):
     _summarize_fn_tracker()
diff --git a/torch_xla/csrc/init_python_bindings.cpp b/torch_xla/csrc/init_python_bindings.cpp
index 3281f0e9a..b255bb043 100644
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -97,6 +97,8 @@ void PrepareToExit() {
   runtime::ComputationClient* client =
       runtime::GetComputationClientIfInitialized();
   if (client != nullptr) {
+    auto xla_device = GetDeviceOrCurrent("");
+    SetAllReduceToken(xla_device, nullptr);
     XLAGraphExecutor::Get()->WaitDeviceOps({});
   }
 }
```

PR #6253, opened 2024-01-03
Add CPU PJRT plugin for testing

will-cromar:
This plugin is for testing only. There is no plan to remove the CPU client from the main `torch_xla` build.


plugins/cpu/README.md
```diff
@@ -0,0 +1,44 @@
...
+# Log device type
+os.environ['TF_CPP_MIN_LOG_LEVEL'] = '0'
+os.environ['TF_CPP_VMODULE'] = 'pjrt_registry=5'
+
+from torch_xla.experimental import plugins
+import torch_xla_cpu_plugin
+import torch_xla.core.xla_model as xm
+import torch_xla.runtime as xr
+
+# Use dynamic plugin instead of built-in CUDA support
```
vanbasten23, line 38: nit: `CPU` support instead of CUDA support?

plugins/cpu/pjrt_c_api_cpu_version_script.lds
```diff
@@ -0,0 +1,10 @@
+# Only symbols in the global section are available to other frameworks.
```
vanbasten23, line 1: can you add a comment about what this file is for?

plugins/cpu/test_cpu_plugin.cc
```diff
@@ -0,0 +1,25 @@
...
+// Use `test` as the platform name instead of `cpu` so torch_xla treats this
+// as an unknown device.
+PJRT_Error* test_platform_name(PJRT_Client_PlatformName_Args* args) {
+  static const std::string platform_name = "test";
+  args->platform_name = platform_name.c_str();
+  args->platform_name_size = platform_name.size();
+  return nullptr;
+}
+
+const PJRT_Api* GetPjrtApi() {
```
vanbasten23, line 17: Curious what this file does.

plugins/cpu/test_cpu_plugin.cc
```diff
@@ -0,0 +1,25 @@
...
+  // HACK: The CPU client is created as a constexpr, so const-casting is
+  // undefined behavior. Make a non-const copy of the struct so we can override
```
vanbasten23, line 19: Wondering where you "Make a non-const copy of the struct"

plugins/cpu/README.md
will-cromar: Oops! Copy/pasting strikes again

plugins/cpu/pjrt_c_api_cpu_version_script.lds
will-cromar: I think the comment here sums it up. This is a linker script, which defines which symbols are exported for use by other binaries (namely `torch_xla`). https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_chapter/ld_3.html

plugins/cpu/test_cpu_plugin.cc
will-cromar: This is just the implementation for `GetPjrtApi`. The integration guide outlines the expected binary API for plugins: https://github.com/openxla/xla/blob/main/xla/pjrt/c/docs/pjrt_integration_guide.md#step-2-implement-getpjrtapi

plugins/cpu/test_cpu_plugin.cc
will-cromar: `pjrt_api` is a value type, so assignment is implicitly a copy.

JackCaoG (approved):

plugins/cpu/pjrt_c_api_cpu_version_script.lds
vanbasten23: I see.  How can we call this `GetPjrtApi` for example from the script in the plugins/cpu/README.md? Would be helpful to add an example.

vanbasten23 (approved):

vanbasten23 (approved):
Thanks!
LGTM with one question in the comment.

plugins/cpu/pjrt_c_api_cpu_version_script.lds
will-cromar: GetPjrtApi is only called indirectly through OpenXLA. The `PjRtCApiClient` handles all of the C-specific plumbing like this. I will add a more detailed plugin integration guide when this work is more mature.

PR #6257, opened 2024-01-04
Update XLAShardedTensor to make it traceable

wonjoolee95:
As follow-up to https://github.com/pytorch/pytorch/issues/115970 and https://github.com/pytorch/xla/pull/6161, this PR adds `__tensor_flatten__` and `__tensor_unflatten__` to `XLAShardedTensor` to make it traceable by Dynamo. 

Some context on the two methods `__tensor_flatten__` and `__tensor_unflatten__`:
- An example of how these methods are implemented in other tensor subclasses: https://github.com/pytorch/pytorch/blob/main/torch/sparse/semi_structured.py#L208
- `__tensor_flatten__` returns two things:
   - List of `strs` that refer to the var name of the `torch.tensor` in the subclass. In our case, that is the `global_tensor` in `XLAShardedTensor`. 
   - Other metadata about the tensor subclass, such as shape, etc. In our case, that is the sharding spec and partitions.

Test plan:
Unit test and CI

### DIFF
--- a/test/spmd/test_dynamo_spmd.py
+++ b/test/spmd/test_dynamo_spmd.py
@@ -237,13 +237,11 @@ class DynamoSpmdInferenceTest(test_xla_sharding_base.XlaShardingTest):
 
     met.clear_counters()
     device = xm.xla_device()
-    dynamo_fn = torch.compile(
-        my_fn, backend="openxla")
+    dynamo_fn = torch.compile(my_fn, backend="openxla")
     t = torch.tensor([0, 1, 2])
     xla_t = t.to(device)
     xla_result = dynamo_fn(xla_t)
-    print(type(xla_result))
-    # torch.allclose(xla_result.cpu(), dynamo_res.cpu())
+    torch.allclose(xla_result.cpu(), dynamo_res.cpu())
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


wonjoolee95:
The tests are green but it seems like this unit test is not sufficient to test this change as it doesn't fail without this PR. I was hoping to have a small unit test to verify the correctness of this change. 

Question for @yeounoh -- we saw the error with `XLAShardedTensor` being untraceable when we were testing LLaMa, wouldn't this unit test suffice to test the traceability?

yeounoh:
> The tests are green but it seems like this unit test is not sufficient to test this change as it doesn't fail without this PR. I was hoping to have a small unit test to verify the correctness of this change.
> 
> Question for @yeounoh -- we saw the error with `XLAShardedTensor` being untraceable when we were testing LLaMa, wouldn't this unit test suffice to test the traceability?

Yea, it's tricky. Based on our previous experiences with Dynamo & our unit testing; I would still go back to the llama2 run to confirm the error with your fix. What we need is another unit test that also checks the fx.graph for the XLAShardedTensor traces.

wonjoolee95:
No longer needed since activation sharding is completed with https://github.com/pytorch/xla/pull/6524. Closing.

PR #6259, opened 2024-01-04
Add matmul microbenchmark.

golechwierowicz:
With the pure wall time infrastructure came nice
ability to do microbenchmarks easily. I added
the sample microbenchmark for matmul.

Additionally for slower runs it dumps PT profiles/traces of base, and test functions.

### DIFF
--- a/benchmarks/microbench.py
+++ b/benchmarks/microbench.py
@@ -1,0 +1,136 @@
+import os
+
+from dataclasses import dataclass
+from torch.profiler import profile, ProfilerActivity, schedule
+
+WAIT = 2
+WARMUP = 5
+ACTIVE = 3
+DEFAULT_SCHEDULE = schedule(wait=WAIT, warmup=WARMUP, active=ACTIVE)
+PROF_ITERS = WAIT + WARMUP + ACTIVE
+
+
+@dataclass
+class MicrobenchResults:
+  test_name: str
+  testing_speedup: float
+  baseline_wall_ms: float
+  testing_wall_ms: float
+
+  def __str__(self):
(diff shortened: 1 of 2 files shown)


vanbasten23:
Why not merging to the master branch?

Also, can you add the feature to the readme?

golechwierowicz:
> Why not merging to the master branch?
> 
> Also, can you add the feature to the readme?

Hey, it depends on changes in the previous branch, bench.py file which provides necessary methods to make this microbenchmark work.

I'll add the usage instructions in the next PR.

cota:
Note that until the baseline is the master branch, CI will not run.
I encountered this in the past, and couldn't get CI to run even after resetting the baseline to the master branch. I ended up creating a new PR based off master so that CI would run.

benchmarks/microbench.py
```diff
@@ -0,0 +1,136 @@
...
+  return MicrobenchResults(
+      test_name=test_name,
+      testing_speedup=baseline_wall_ms / testing_wall_ms,
+      baseline_wall_ms=baseline_wall_ms,
+      testing_wall_ms=testing_wall_ms,
+  )
+
+
+def _dump_flame(filepath, prof):
+  """Data can be interpreted with running ./flamegraph.pl --title "CUDA time" --countname "us." /tmp/profiler_stacks.txt > perf_viz.svg"""
```
cota, line 96: Where does `flamegraph.pl` come from?

golechwierowicz: Good point, I added the comment mentioning how to do a set up.

cota (approved):

PR #6262, opened 2024-01-05
remove model training using openxla backend

zpcore:
IIUC, there should not exist combination of `openxla` dynamo backend for evaluation.

### DIFF
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -80,7 +80,7 @@ class ExperimentLoader:
       if cfg_xla is None or cfg_test != "eval":
         return False
     elif cfg_dynamo == "openxla":
-      if cfg_xla is None:
+      if cfg_xla is None or cfg_test != "train":
         return False
     elif cfg_dynamo is None:
       pass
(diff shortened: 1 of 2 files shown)


frgossen:
I think @JackCaoG had added this and my understanding is that the plan is to retire openxla_eval

zpcore:
> I think @JackCaoG had added this and my understanding is that the plan is to retire openxla_eval

Thanks,

vanbasten23 (approved):
LGTM assuming openxla_eval is not deprecated per https://github.com/pytorch/xla/pull/6262#issuecomment-1881197725. Suggest to check with Jack about the plan.

frgossen:
I don't understand why this was merged. Shouldn't `openxla` support both training and inference?

zpcore:
> I don't understand why this was merged. Shouldn't `openxla` support both training and inference?

Hi @JackCaoG , can you help confirm if `openxla` backend support both training and inference? And do you have any plan on when to deprecate `openxla_eval`?

JackCaoG:
openxla supports both training and inference. In the long run we want to get rid of `openxla_eval`, but currently `openxla_eval` is likely yield better performance on inference.

JackCaoG:
in the mean time `openxla` to train and inference + `openxla_eval` for inference are valid use cases.

zpcore:
I see, thanks. Let me revert the PR.

Issue #6252, opened 2024-01-03
test/spmd/test_spmd_debugging.py fails when run on a v4-8 TPU


mbzomowski:
## 🐛 Bug

test/spmd/test_spmd_debugging.py fails when run on a v4-8 TPU, with the following output:

```
root@8a626de6d6ae:/ansible/pytorch/xla# python3 -u test/spmd/test_spmd_debugging.py
s                                                           
 TPU 0  TPU 4  TPU 8  TPU 12  TPU 2  TPU 6  TPU 10  TPU 14 
                                                           
                                                           
 TPU 1  TPU 5  TPU 9  TPU 13  TPU 3  TPU 7  TPU 11  TPU 15 
                                                           
Fs              
 TPU 0  TPU 1 
              
              
 TPU 2  TPU 3 
              
Fs              
  TPU [0, 1]  
              
              
  TPU [4, 5]  
              
              
  TPU [8, 9]  
              
              
 TPU [12, 13] 
              
              
  TPU [2, 3]  
              
              
  TPU [6, 7]  
              
              
 TPU [10, 11] 
              
              
 TPU [14, 15] 
              
Fs                  
 TPU [0, 1, 2, 3] 
                  
Fs            
 TPU [0, 1] 
            
            
 TPU [2, 3] 
            
Fs                  
 TPU [0, 1, 2, 3] 
                  
F
======================================================================
FAIL: test_debugging_spmd_multi_host_tiled_tpu (__main__.DebuggingSpmdTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/spmd/test_spmd_debugging.py", line 454, in test_debugging_spmd_multi_host_tiled_tpu
    assert output == fake_output
AssertionError

======================================================================
FAIL: test_debugging_spmd_single_host_tiled_tpu (__main__.DebuggingSpmdTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/spmd/test_spmd_debugging.py", line 109, in test_debugging_spmd_single_host_tiled_tpu
    assert output == fake_output
AssertionError

======================================================================
FAIL: test_multi_host_partial_replication_tpu (__main__.DebuggingSpmdTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/spmd/test_spmd_debugging.py", line 538, in test_multi_host_partial_replication_tpu
    assert output == fake_output
AssertionError

======================================================================
[...]

PR #6263, opened 2024-01-05
Update test_spmd_debugging.py to avoid code test code self

ManfeiBai:
Fix https://github.com/pytorch/xla/issues/6252

- `test/spmd/test_spmd_debugging.py` was added for test spmd debugging tool, to avoid local test failure and circular reasoning, test based on given sharding string
- add GPU tests by change `@unittest.skipIf(`
- enable local test color check before draw table

### DIFF
--- a/test/spmd/test_spmd_debugging.py
+++ b/test/spmd/test_spmd_debugging.py
@@ -123,14 +123,28 @@ class DebuggingSpmdTest(test_xla_sharding_base.XlaShardingTest):
     col = []
     col.append(
         rich.padding.Padding(
-            rich.align.Align('TPU [0, 1, 2, 3]', "center", vertical="middle"),
+            rich.align.Align('TPU [0, 1]', "center", vertical="middle"),
             (1, 1, 1, 1),
             style=rich.style.Style(bgcolor=color, color=text_color)))
     fake_table.add_row(*col)
     col = []
     col.append(
         rich.padding.Padding(
-            rich.align.Align('TPU [4, 5, 6, 7]', "center", vertical="middle"),
+            rich.align.Align('TPU [2, 3]', "center", vertical="middle"),
+            (1, 1, 1, 1),
+            style=rich.style.Style(bgcolor=color, color=text_color)))
+    fake_table.add_row(*col)
+    col = []
+    col.append(
+        rich.padding.Padding(
+            rich.align.Align('TPU [4, 5]', "center", vertical="middle"),
+            (1, 1, 1, 1),
+            style=rich.style.Style(bgcolor=color, color=text_color)))
+    fake_table.add_row(*col)
+    col = []
+    col.append(
+        rich.padding.Padding(
+            rich.align.Align('TPU [6, 7]', "center", vertical="middle"),
             (1, 1, 1, 1),
             style=rich.style.Style(bgcolor=color, color=text_color)))
     fake_table.add_row(*col)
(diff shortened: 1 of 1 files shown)


test/spmd/test_spmd_debugging.py
```diff
@@ -116,6 +119,8 @@ def test_single_host_partial_replication_tpu(self):
     from torch_xla.distributed.spmd.debugging import visualize_tensor_sharding
     device = xm.xla_device()
     num_devices = self.n_devices
+    if num_devices != 8:
+      self.skipTest("skip num_devices!=8 env to avoid circular reasoning")
```
jonb377, line 123: Could we make these checks into test decorators? e.g. `unittest.skipIf(xr.global_runtime_device_count() != 8)`

Also for my understanding, what's the circular reasoning here?

test/spmd/test_spmd_debugging.py
```diff
@@ -255,6 +267,8 @@ def test_single_host_partial_replication_cpu(self):
     from torch_xla.distributed.spmd.debugging import visualize_tensor_sharding
     device = xm.xla_device()
     num_devices = self.n_devices
+    if num_devices != 1:
```
jonb377, line 270: Naive question, but is there a case where `PJRT_DEVICE=CPU` has n_devices > 1?

jonb377 (commented):
Thanks Manfei!

test/spmd/test_spmd_debugging.py
yeounoh: +1, `unittest.skipIf` is the way to go. What's the reason behind this check? `!=8` seems too restrictive -- so if we can generalize to `n` devices, it would be better?

ManfeiBai:
we might want to move code of get local device number to doc example, and generalize test code start from given string

test/spmd/test_spmd_debugging.py
ManfeiBai: thanks, Jon, very good question, checked with TPU Pod:
```
$ gcloud alpha compute tpus tpu-vm ssh ${TPU_NAME}  --project ${PROJECT_ID} --zone  ${ZONE} --worker all --command='PJRT_DEVICE=CPU python3 -c "import torch; import torch_xla; import torch_xla.runtime as xr; print(xr.global_runtime_device_count());"'
Using ssh batch size of 4. Attempting to SSH into 1 nodes with a total of 4 workers.
SSH: Attempting to connect to worker 0...
SSH: Attempting to connect to worker 1...
SSH: Attempting to connect to worker 2...
SSH: Attempting to connect to worker 3...
1
1
1
1
```
looks like `PJRT_DEVICE=CPU` only has n_devices = 1

test/spmd/test_spmd_debugging.py
ManfeiBai: Thanks for both suggestion, has updated to more generalized way to generate expected table without restrictive limitation

and for `circular reasoning`, I want to describe a test situation that we need to provide test example by our function, and provide expected example by our function too,

for this `test/spmd/test_spmd_debugging.py`, if we want to test on given device kind, we need to generate table with our function, and generate expected table with our function too, so to avoid this(circular reasoning), we want to limit the test device kind to 8-devices

let me change to `code test code self` for better description

ManfeiBai:
> we might want to move code of get local device number to doc example, and generalize test code start from given sharding string

done

yeounoh (approved):
LGTM

PR #6266, opened 2024-01-05
Fix global_runtime_device_count() for spmd case

vanbasten23:
This pr fixes global_runtime_device_count() for SPMD case. This fix works on multi-host scenario as well.

Prior this PR, the APIs returns 1. With the PR, global_runtime_device_count() returns all the devices across the host (with torchrun).

Note xr.local_device_count() should still return 1 for SPMD. (fwiw, runtime::GetComputationClient()->GetLocalDevices() gets all the devices on the host even for spmd case, which can be confusing) and xr.global_device_count() still returns 1, as shown in test/spmd/test_xla_spmd_python_api_interaction.py

PR #6275, opened 2024-01-09
fix gpu spmd core dump

mars1248:
This `std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures(devices.size());` will cause the elements in returned_futures to be initialized with null values, which will cause an error on subsequent attempts to access returned_futures[0]

### DIFF
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -679,7 +679,7 @@ PjRtComputationClient::ExecuteReplicated(
 
   using FutureVector = std::vector<xla::PjRtFuture<xla::Status>>;
   std::optional<FutureVector> returned_futures = FutureVector();
-  returned_futures->reserve(devices.size()); 
+  returned_futures->reserve(devices.size());
   std::vector<std::vector<std::unique_ptr<xla::PjRtBuffer>>> results;
   {
     tsl::profiler::TraceMe activity(
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -677,8 +677,8 @@ PjRtComputationClient::ExecuteReplicated(
   TF_VLOG(5) << "ExecuteReplicated acquiring PJRT device lock for "
              << spmd_device_str << " Done";
 
-  using FutureVector = std::vector<xla::PjRtFuture<xla::Status>>;
-  std::optional<FutureVector> returned_futures = FutureVector();
+  std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures =
+      std::vector<xla::PjRtFuture<xla::Status>>();
   returned_futures->reserve(devices.size());
   std::vector<std::vector<std::unique_ptr<xla::PjRtBuffer>>> results;
   {
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -679,7 +679,6 @@ PjRtComputationClient::ExecuteReplicated(
 
   std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures =
       std::vector<xla::PjRtFuture<xla::Status>>();
-  returned_futures->reserve(devices.size());
   std::vector<std::vector<std::unique_ptr<xla::PjRtBuffer>>> results;
   {
     tsl::profiler::TraceMe activity(
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -677,8 +677,9 @@ PjRtComputationClient::ExecuteReplicated(
   TF_VLOG(5) << "ExecuteReplicated acquiring PJRT device lock for "
              << spmd_device_str << " Done";
 
-  std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures(
-      devices.size());
+  using FutureVector = std::vector<xla::PjRtFuture<xla::Status>>;
+  std::optional<FutureVector> returned_futures = FutureVector();
+  returned_futures->reserve(1); 
   std::vector<std::vector<std::unique_ptr<xla::PjRtBuffer>>> results;
   {
     tsl::profiler::TraceMe activity(
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -679,7 +679,7 @@ PjRtComputationClient::ExecuteReplicated(
 
   using FutureVector = std::vector<xla::PjRtFuture<xla::Status>>;
   std::optional<FutureVector> returned_futures = FutureVector();
-  returned_futures->reserve(1); 
+  returned_futures->reserve(devices.size()); 
   std::vector<std::vector<std::unique_ptr<xla::PjRtBuffer>>> results;
   {
     tsl::profiler::TraceMe activity(


will-cromar:
Hmm, I wonder why this didn't cause issues before. From the docs: [`if Execute does not return an error status: *returned_futures will be resized to be the same length as the return vector ...`](https://github.com/openxla/xla/blob/13200274883ac910d5c25d5b82d50e98f42b1b39/xla/pjrt/pjrt_client.h#L1383-L1387). Let's just remove both the iniitial size and the `reserve` call and let PJRT worry about the vector size. In this case, they'll call `reserve` themselves down the line: https://github.com/openxla/xla/blob/13200274883ac910d5c25d5b82d50e98f42b1b39/xla/pjrt/pjrt_stream_executor_client.cc#L2911-L2913

cc @vanbasten23 since you're working on GPU SPMD.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -677,8 +677,9 @@ PjRtComputationClient::ExecuteReplicated(
   TF_VLOG(5) << "ExecuteReplicated acquiring PJRT device lock for "
              << spmd_device_str << " Done";
 
-  std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures(
-      devices.size());
+  using FutureVector = std::vector<xla::PjRtFuture<xla::Status>>;
```
will-cromar, line 680: Please remove the `FutureVector` alias. The triple-nested template is verbose, but I'd rather match [the API we're using here](https://github.com/openxla/xla/blob/13200274883ac910d5c25d5b82d50e98f42b1b39/xla/pjrt/pjrt_client.h#L1396).

vanbasten23:
> Hmm, I wonder why this didn't cause issues before. From the docs: [`if Execute does not return an error status: *returned_futures will be resized to be the same length as the return vector ...`](https://github.com/openxla/xla/blob/13200274883ac910d5c25d5b82d50e98f42b1b39/xla/pjrt/pjrt_client.h#L1383-L1387). Let's just remove both the iniitial size and the `reserve` call and let PJRT worry about the vector size. In this case, they'll call `reserve` themselves down the line: https://github.com/openxla/xla/blob/13200274883ac910d5c25d5b82d50e98f42b1b39/xla/pjrt/pjrt_stream_executor_client.cc#L2911-L2913
> 
> cc @vanbasten23 since you're working on GPU SPMD.

Just to provide some context (we discussed in https://github.com/pytorch/xla/issues/6225): To make spmd on gpu work, we need a prerequisite fix such as https://github.com/pytorch/xla/pull/6266. Then we bumped into error in this PR. Without this fix, running SPMD on GPU would fail at https://github.com/pytorch/xla/blob/645b50eb885131c666d8274818d1276968677e5a/torch_xla/csrc/runtime/pjrt_computation_client.cc#L697 with an error:
```
F0000 00:00:1702341857.871586   15007 pjrt_future.h:229] Check failed: IsValid()
*** Check failure stack trace: ***
    @     0x7fb9f5855589  absl::lts_20230802::log_internal::LogMessageFatal::~LogMessageFatal()
    @     0x7fb9eb243f3c  xla::PjRtFuture<>::OnReady()
    @     0x7fb9eb2457d4  torch_xla::runtime::PjRtComputationClient::ExecuteReplicated()
    @     0x7fb9eba60e39  torch_xla::XLAGraphExecutor::ScheduleSyncTensorsGraph()::{lambda()#1}::operator()()
    @     0x7fbb69978b2a  torch::lazy::MultiWait::Complete()
Aborted (core dumped)
```
But I also wonder why it's not a problem for TPU since SPMD runs fine on TPU (perhaps an issue with the gpu se client?)

vanbasten23:
I'm not sure if I follow why the change would fix the issue even with the comment https://github.com/pytorch/xla/pull/6275#issue-2071795005. With the change in this PR, 
```
std::optional<FutureVector> returned_futures = FutureVector();
returned_futures->reserve(devices.size());
```
`returned_futures`'s size is 0. @mars1248 can you elaborate why this pr fixes the crash?

will-cromar:
> I'm not sure if I follow why the change would fix the issue even with the comment [#6275 (comment)](https://github.com/pytorch/xla/pull/6275#issue-2071795005). With the change in this PR,
> 
> ```
> std::optional<FutureVector> returned_futures = FutureVector();
> returned_futures->reserve(devices.size());
> ```
> 
> `returned_futures`'s size is 0. @mars1248 can you elaborate why this pr fixes the crash?

The SE client pushes the "real" futures to the back of the list: https://github.com/openxla/xla/blob/13200274883ac910d5c25d5b82d50e98f42b1b39/xla/pjrt/pjrt_stream_executor_client.cc#L2935

So if you initialize the list with empty futures, they'll be left empty and you get a segfault.

vanbasten23:
> So if you initialize the list with empty futures, they'll be left empty and you get a segfault.

I see. Perhaps the reason why TPU doesn't have the issue is the pjrt c client doesn't push the futures..

torch_xla/csrc/runtime/pjrt_computation_client.cc
yeounoh: +1 I prefer sticking to non-aliased type names, unless it's used frequently.

yeounoh:
The reason why C API Client doens't have an issue with this is because it resizes internally afterwards. Anyway, thank you @mars1248 for catching this -- will approve after we settle on the aliasing comment.

vanbasten23:
LGTM pending on https://github.com/pytorch/xla/pull/6275#discussion_r1447719594

mars1248:
> IIRC, the reason why C API Client doens't have an issue with this is because it resizes internally afterwards. Anyway, thank you @mars1248 for catching this -- will approve after we settle on the aliasing comment.

@yeounoh I've removed aliasing.Please check it again.
@vanbasten23 I would like to ask why this pr #6266 does not fit in? With this pr and my fix, the gpu spmd is working on my side.

vanbasten23 (approved):

vanbasten23:
> @vanbasten23 I would like to ask why this pr #6266 does not fit in? With this pr and my fix, the gpu spmd is working on my side.

Not sure what you mean by "fit in". Do you mean with your fix #6275 and the pr #6266, the gpu spmd is working?

yeounoh (approved):
Thank you!

vanbasten23:
The test fails with error `virtual memory exhausted: Cannot allocate memory`. @mars1248 could you rebase and trigger the CI again? IIRC, we've fixed a similar issue a few days ago.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -677,8 +677,9 @@ PjRtComputationClient::ExecuteReplicated(
   TF_VLOG(5) << "ExecuteReplicated acquiring PJRT device lock for "
              << spmd_device_str << " Done";
 
-  std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures(
-      devices.size());
+  std::optional<std::vector<xla::PjRtFuture<xla::Status>>> returned_futures =
+      std::vector<xla::PjRtFuture<xla::Status>>();
+  returned_futures->reserve(devices.size());
```
will-cromar, line 682: Can you also remove the `reserve` here? Let's let PJRT handle this vector.

mars1248: ok, I've removed it

mars1248:
> > @vanbasten23 I would like to ask why this pr #6266 does not fit in? With this pr and my fix, the gpu spmd is working on my side.
> 
> Not sure what you mean by "fit in". Do you mean with your fix #6275 and the pr #6266, the gpu spmd is working? If so, I agree with you and we are on the same page.
@vanbasten23 
yes, fix #6275 and the pr #6266 spmd multi-GPU can be trained normally.

vanbasten23:
> @vanbasten23 Yes, fix #6275 and the pr #6266 spmd multi-GPU can be trained normally.So I would like to ask when the pr #6266 can be combined.

#6266 requires your pr (otherwise the test fails. The failure is exactly you are fixing in this pr). Once your pr is merged, #6266 can be merged.

PR #6283, opened 2024-01-10
Fix output type of custom calls while lowering quant/dequant torch op to HLO

sdasgup3:
https://github.com/pytorch/xla/pull/5763 allows lowering the torch quantize/dequantize operations to HLO custom calls. For example, 
the following PyTorch code
```
x = torch.ops.quantized_decomposed.quantize_per_tensor(
    x, 0.4, 2, -128, 127, torch.int8)
x = torch.ops.quantized_decomposed.dequantize_per_tensor(
    x, 0.4, 2, -128, 127, torch.int8)
```

is lowered to the following HLO operations:
```
ENTRY %IrToHlo.5 (p0.1: f32[2,3,4,5]) -> (f32[2,3,4,5]) {
  %p0.1 = f32[2,3,4,5]{3,2,1,0} parameter(0)

  %custom-call.2 = f32[2,3,4,5]{3,2,1,0} custom-call(f32[2,3,4,5]{3,2,1,0} %p0.1), custom_call_target="stablehlo.uniform_quantize", api_version=API_VERSION_TYPED_FFI, backend_config={scale=[0.4],zero_point=[2],storage_type=si8,expressed_type=f32,storage_min=-128,storage_max=127}

  %custom-call.3 = f32[2,3,4,5]{3,2,1,0} custom-call(f32[2,3,4,5]{3,2,1,0} %custom-call.2), custom_call_target="stablehlo.uniform_dequantize", api_version=API_VERSION_TYPED_FFI, backend_config={scale=[0.4],zero_point=[2],storage_type=si8,expressed_type=f32,storage_min=-128,storage_max=127}
  ROOT %tuple.4 = (f32[2,3,4,5]{3,2,1,0}) tuple(f32[2,3,4,5]{3,2,1,0} %custom-call.3)
}
```

Note that the output of custom call corresponding to quantize op has element type `f32`.  The fact that the output of custom_call (for quantize operation) is a float is more of a logical problem as the the result of quantization is generally expected to be in integer domain. Also, note that the choice of output type should not effect the eventual conversion of HLO custom calls to mhlo `uniform.quantize`/`uniform.dequantize` operations. 

Moreover, based on https://github.com/pytorch/pytorch/blob/0b72ce1bd1a4a0596dde4053899b8a9a7999bc47/torch/ao/quantization/fx/_decomposed.py#L164 we set the output element type of dequatize operation to be `f32`

Finally improved the debuggability of the map queries using proper error messages.  

cc @GleasonK

### DIFF
--- a/torch_xla/csrc/ops/dequant_tensor.cpp
+++ b/torch_xla/csrc/ops/dequant_tensor.cpp
@@ -36,13 +36,15 @@ torch::lazy::NodePtr DequantizeTensor::Clone(
 XlaOpVector DequantizeTensor::Lower(LoweringContext* loctx) const {
   xla::XlaOp input = loctx->GetOutputOp(operand(0));
   xla::Shape input_shape = ShapeHelper::ShapeOfXlaOp(input);
+  xla::Shape output_shape = xla::ShapeUtil::MakeShape(xla::PrimitiveType::F32,
+                                                      input_shape.dimensions());
   // TODO(lsy323): Lower to HLO directly once qdtype is added to HLO.
   static const std::string opname = "stablehlo.uniform_dequantize";
   auto qparams = QuantParams(scale_, zero_point_, quant_min_, quant_max_, axis_,
-                             dtype_, input_shape.element_type());
+                             dtype_, xla::PrimitiveType::F32);
 
   xla::XlaOp output = xla::CustomCall(
-      input.builder(), opname, {input}, input_shape,
+      input.builder(), opname, {input}, output_shape,
       qparams.SerializeToAttrDictStr(),
       /*has_side_effect=*/false,
       /*output_operand_aliasing=*/{}, /*literal=*/nullptr,
--- a/torch_xla/csrc/runtime/stablehlo_helper.cc
+++ b/torch_xla/csrc/runtime/stablehlo_helper.cc
@@ -168,14 +168,14 @@ void ConvertStableHloToHlo(mlir::ModuleOp* mlir_module,
                          << getMlirModuleStr(*mlir_module);
 }
 
-const std::unordered_map<std::string, std::string>&
-GetTorchDtypeToStablehloDtypeMap() {
-  static const std::unordered_map<std::string, std::string> m_{
-      {"torch.int8", "si8"},
-      {"torch.uint8", "ui8"},
-      {"torch.int16", "si16"},
-  };
-  return m_;
+const std::string GetTorchDtypeToStablehloDtype(const std::string& dtype) {
+  if (dtype == "torch.int8") return "si8";
+  if (dtype == "torch.uint8") return "ui8";
+  if (dtype == "torch.int16") return "si16";
+  if (dtype == "torch.int32") return "si32";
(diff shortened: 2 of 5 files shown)


torch_xla/csrc/ops/dequant_tensor.cpp
```diff
@@ -36,13 +36,15 @@ torch::lazy::NodePtr DequantizeTensor::Clone(
 XlaOpVector DequantizeTensor::Lower(LoweringContext* loctx) const {
   xla::XlaOp input = loctx->GetOutputOp(operand(0));
   xla::Shape input_shape = ShapeHelper::ShapeOfXlaOp(input);
+  xla::Shape output_shape = xla::ShapeUtil::MakeShape(xla::PrimitiveType::F32,
```
GleasonK, line 39: Should we assert that the input_shape uses f32 storage type? Is this guaranteed?

torch_xla/csrc/runtime/stablehlo_helper.cc
```diff
@@ -192,4 +192,16 @@ GetHloDtypeToStablehloDtypeMap() {
   return m_;
 }
 
+const std::unordered_map<std::string, xla::PrimitiveType>&
+GetTorchIntDtypeToHloDtypeMap() {
+  static const std::unordered_map<std::string, xla::PrimitiveType> m_{
```
GleasonK, line 197: We could implement this as a switch / series of if stmts to avoid a static dictionary, it's a short enough list. Not sure of PT/XLAs stance on static data.

torch_xla/csrc/ops/dequant_tensor.cpp
GleasonK: Edit: I see that dequantize decomposes to f32 in PT. Probably OK for now. Not sure if there's a way to future proof this. I.e. can we query the storage type from elsewhere? The PT op maybe?

GleasonK (approved):

lsy323 (approved):

torch_xla/csrc/ops/dequant_tensor.cpp
sdasgup3: Unfortunately the PT op like the followings:
```
x1 = torch.ops.quantized_decomposed.quantize_per_tensor(
    x, 0.4, 2, -128, 127, torch.int8)
x2 = torch.ops.quantized_decomposed.dequantize_per_tensor(
    x1, 0.4, 2, -128, 127, torch.int8)
```

does not convey the expressed type (`f32` in this case) as its attributes. For quantize PT op, it is determined via the input argument `x` and for de-quantize the expressed type can be determined via the input argument of the operation generating `x1`.  The fact that quantize decomposes to f32 in PT readily provides us output type information for dequantize.

torch_xla/csrc/runtime/stablehlo_helper.cc
sdasgup3: I agree about less static data floating around. Replaced the dictionary with  conditionals. 
@lsy323 If the change LG to you then I can merge once the CI is green.

PR #6284, opened 2024-01-10
Fix some more core aten opset tests

wonjoolee95:
Fixes https://github.com/pytorch/xla/issues/5893, fixes https://github.com/pytorch/xla/issues/5978, fixes https://github.com/pytorch/xla/issues/5977, fixes https://github.com/pytorch/xla/issues/5860, fixes https://github.com/pytorch/xla/issues/5899

aten_sort
- add manual dtype conversion

aten_unsqueeze
- no-op, just enabling it

aten_squeeze_dim
- no-op, just enabling it

aten_div_Tensor
- no-op, just enabling it

aten_sort
- this is a tricky one -- so according to https://pytorch.org/docs/stable/generated/torch.sort.html, aten_sort returns two things: a namedtuple of (values, indices) is returned, where the values are the sorted values and indices are the indices of the elements in the original input tensor. We want to make sure the 0th index tensor (values) want to equal. However, the other tensor (indices) might be different if there are duplicates. Torch.sort has a optional param `stable` that controls this, but for some reason, trying to pass this `kwargs` is a bit tricky in the way we set up our tests. So I just manually updated the test parameter to be a tensor without repeating values by using `torch.randperm`.

Test plan: CI

### DIFF
--- a/test/test_core_aten_ops.py
+++ b/test/test_core_aten_ops.py
@@ -1438,7 +1438,6 @@ class AtenOpTest(unittest.TestCase):
     kwargs = dict()
     run_export_and_compare(self, torch.ops.aten.div.Tensor, args, kwargs)
 
-  @unittest.skip
   def test_aten_div_Tensor_2(self):
     args = (
         torch.randint(0, 10, (10, 10)).to(torch.int32),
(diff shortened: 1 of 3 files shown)


ManfeiBai (approved):
LGTM

PR #6022, opened 2023-12-04
Fix global_device_count(), local_device_count() for single process on CUDA

vanbasten23:
This PR fixes 
- global_device_count() so that it returns all GPU devices across all processes/hosts. The fix also works for [multi-host case](https://gist.github.com/vanbasten23/b544ef039c2fb3548ab5d13399463d1e).
- local_device_count()
for single process on CUDA so that it returns all GPU devices on the current host.

Before this PR, both APIs always return 1 as reported in this [issue](https://github.com/pytorch/xla/issues/5910).).

global_runtime_device_count is not fixed since it seems it's only used in spmd case and it's been fixed in another pr.

Test:
- PJRT_DEVICE=CUDA   python pytorch/xla/test/pjrt/test_runtime.py
- PJRT_DEVICE=CUDA torchrun --nnodes 1 --nproc-per-node 2 pytorch/xla/test/pjrt/test_torchrun.py

Note: [here](b/300100792#comment7) is the behavior of torch.cuda.device_count() on multi-host case

PR #6289, opened 2024-01-10
Support PJRT C API `create_options`

will-cromar:


will-cromar:
Our resnet example actually works on 4 v100s with the plugin now!

will-cromar:
`torch_xla._XLAC._register_pjrt_plugin` is starting to get bloated. If we have to add more to it, I'll look into defining a base class in C++ and allowing overrides in Python. Then, we can just pass the entire `DevicePlugin` to C++.

plugins/cuda/torch_xla_cuda_plugin/__init__.py
```diff
@@ -1,11 +1,32 @@
...
+    local_world_size = xu.getenv_as("LOCAL_WORLD_SIZE", int, 1)
+    global_world_size = xu.getenv_as("WORLD_SIZE", int, local_world_size)
+
+    # The available options are defined in OpenXLA: https://github.com/openxla/xla/blob/1bb2a74be91fabf5f9aa2702b2592b5b022c9052/xla/pjrt/c/pjrt_c_api_gpu_internal.cc#L58-L67
+    return {
+      "platform_name": "gpu",
+      # TODO(wcromar): make this configurable
+      "allocator": "cuda_async" if xu.getenv_as("PJRT_ALLOCATOR_CUDA_ASYNC", bool, False) else "default",
+      "memory_fraction": xu.getenv_as("PJRT_ALLOCATOR_FRACTION", float, .75),
+      "preallocate": xu.getenv_as("PJRT_ALLOCATOR_PREALLOCATE", bool, True),
```
JackCaoG, line 25: are these env var new?

JackCaoG (commented):
do you want to add any test for GPU plugin?

will-cromar:
> do you want to add any test for GPU plugin?

I don't actually have a good build process set up yet for this plugin.  I'll work on that this week.

Unless we find an issue, I'm inclined to just move the whole CI over to use the plugin once the build is automatable. WDYT?

torch_xla/experimental/plugins.py
```diff
@@ -41,6 +41,17 @@ def physical_chip_count(self):
     """
     return 1
 
+  def client_create_options(self) -> dict:
+    return {}
+
+  def requires_xla_coordinator(self) -> bool:
```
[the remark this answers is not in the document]
will-cromar: This is backend-specific. TPU, for example, does not need the coordinator even when `world_size > 1`. Plugins should override this with the correct logic as needed.

torch_xla/csrc/runtime/pjrt_registry.cc
```diff
@@ -54,13 +64,46 @@ InitializePjRt(const std::string& device_type) {
...
+                                                   v);
+          };
+        }
+      }
       const PJRT_Api* c_api = *pjrt::LoadPjrtPlugin(
-          absl::AsciiStrToLower(device_type), *plugin_path);
+          absl::AsciiStrToLower(device_type), plugin->library_path);
       XLA_CHECK_OK(pjrt::InitializePjrtPlugin(device_type));
-      client = xla::GetCApiClient(absl::AsciiStrToUpper(device_type)).value();
+      client = xla::GetCApiClient(absl::AsciiStrToUpper(device_type),
```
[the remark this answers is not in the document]
will-cromar: Here: https://github.com/openxla/xla/blob/main/xla/pjrt/pjrt_c_api_client.h

torch_xla/csrc/runtime/pjrt_registry.cc
```diff
@@ -54,13 +64,46 @@ InitializePjRt(const std::string& device_type) {
   std::unique_ptr<XlaCoordinator> coordinator;
 
   if (sys_util::GetEnvBool(env::kEnvPjrtDynamicPlugins, false)) {
-    std::optional<std::string> plugin_path = GetPjRtPluginPath(device_type);
-    if (plugin_path) {
+    std::optional<PluginEntry> plugin = GetPjRtPlugin(device_type);
```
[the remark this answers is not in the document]
will-cromar: Eventually we'll use this branch for every plugin (ie every case except CPU). This is an experimental option for now.

JackCaoG:
@will-cromar How will the UX be(for both building from source and installing from whl)? I want to make sure we don't require user to do additional steps to use the plug in after we make it default.

will-cromar:
> @will-cromar How will the UX be(for both building from source and installing from whl)? I want to make sure we don't require user to do additional steps to use the plug in after we make it default.

For installing, we can add an extra requirement like we have for `libtpu`. ie `pip install torch_xla[cuda] -f http://index/of/cuda/plugins`. This is what JAX does

For building, you'll may have to build both the plugin and `torch_xla`. Although, in 99% of cases, you can just install the pre-built plugin like we do for TPU today

torch_xla/csrc/runtime/pjrt_registry.cc
```diff
@@ -14,10 +14,16 @@
...
 namespace runtime {
 
-std::unordered_map<std::string, std::string> pjrt_plugins_;
-
 namespace {
 
+struct PluginEntry {
+  std::string library_path;
+  absl::flat_hash_map<std::string, xla::PjRtValueType> create_options;
+  bool init_coordinator;
```
jonb377, line 22: I wonder if we should drop `init_coordinator` and instead always initialize the coordinator when the distributed env vars are set, even on TPU where it's not strictly necessary. As long as torchrun launches the training in a distributed context, the env vars should be set, which I believe covers all GPU use cases since we plan to use torchrun for GPU SPMD (cc @vanbasten23).

On TPU it's not currently required, but we can always detect the environment from the GCE metadata and set the env vars for the user in a distributed context, since we don't require torchrun for multicontroller execution.

jonb377 (commented):
> Note: it may be possible to directly use the torch.distributed store here directly instead.

We briefly chatted about this offline, but perhaps we should push XlaCoordinator's distributed KV store up into the Python layer to replace TCPStore in our XLA backend implementation instead of dropping XlaCoordinator. Autocheckpointing will still require XlaCoordinator even if we move to `torch.distributed`'s kv store.

Just raising this for discussion. Overall this looks great, thanks Will!

torch_xla/experimental/plugins.py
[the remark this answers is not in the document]
vanbasten23: I see your point. It looks to me the gpu plugin will overwrite this

vanbasten23:
When debugging the pjrt c api for gpu, right now I can change pytorch/xla/WORKSPACE to use a local open XLA, add `print` to open XLA, compile everything, then I can see my `print` statements. I wonder if/how I can do that with the plugin.

plugins/cuda/torch_xla_cuda_plugin/__init__.py
```diff
@@ -1,11 +1,32 @@
...
+      "visible_devices": [local_process_rank],
```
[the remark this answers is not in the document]
will-cromar: Yeah, this PR has the same bug that #6022 fixes. I'll port the fix to this `DevicePlugin` once that PR is finalized and merged.

torch_xla/csrc/runtime/pjrt_registry.cc
```diff
@@ -54,13 +64,46 @@ InitializePjRt(const std::string& device_type) {
...
+        std::string port = runtime::sys_util::GetEnvString(
+            "XLA_COORDINATOR_PORT", XlaCoordinator::kDefaultCoordinatorPort);
+
+        if (global_world_size > 1) {
+          // Use the XlaCoordinator as the distributed key-value store.
+          coordinator = std::make_unique<XlaCoordinator>(
+              global_process_rank, global_world_size, master_addr, port);
+          std::shared_ptr<xla::DistributedRuntimeClient> distributed_client =
+              coordinator->GetClient();
+          std::string key_prefix = "gpu:";
```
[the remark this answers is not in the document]
will-cromar: Yeah, that makes sense. I copied this from the current GPU implementation. I'll try to figure out if this value is significant.

will-cromar:
> When debugging the pjrt c api for gpu, right now I can change pytorch/xla/WORKSPACE to use a local open XLA, add `print` to open XLA, compile everything, then I can see my `print` statements. I wonder if/how I can do that with the plugin.

The GPU plugin binary still uses the same bazel workspace, so any changes you make there will apply to both the plugin and `torch_xla`

torch_xla/csrc/runtime/pjrt_registry.cc
will-cromar: > I wonder if we should drop init_coordinator and instead always initialize the coordinator when the distributed env vars are set, even on TPU where it's not strictly necessary.

In this case, I think we still want to keep `requires_xla_coordinator` option. We would just throw an error immediately if we don't have enough information to start the coordinator.

My other idea initially was that we could ask the plugin for the master IP, local rank, global rank, and world size, perhaps just asking `torch.distributed` for those values in the default implementation.

torch_xla/csrc/runtime/pjrt_registry.cc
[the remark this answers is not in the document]
will-cromar: Switching this to `pjrt` didn't make anything explode, so I think we're good.

torch_xla/csrc/runtime/pjrt_registry.cc
jonb377: I see - that makes sense. Just for context why I brought this up, JAX recently started requiring the coordinator to be initialized before the backend can be used (even on TPUs), but I'm not sure on the reason.

Keeping `init_coordinator` sounds fine to me. Thanks Will!

plugins/cuda/torch_xla_cuda_plugin/__init__.py
```diff
@@ -1,11 +1,32 @@
...
+    local_process_rank = xu.getenv_as("LOCAL_RANK", int, 0)
+    global_process_rank = xu.getenv_as("RANK", int, local_process_rank)
+    local_world_size = xu.getenv_as("LOCAL_WORLD_SIZE", int, 1)
+    global_world_size = xu.getenv_as("WORLD_SIZE", int, local_world_size)
+
+    # The available options are defined in OpenXLA: https://github.com/openxla/xla/blob/1bb2a74be91fabf5f9aa2702b2592b5b022c9052/xla/pjrt/c/pjrt_c_api_gpu_internal.cc#L58-L67
+    return {
+      "platform_name": "gpu",
+      # TODO(wcromar): make this configurable
+      "allocator": "cuda_async" if xu.getenv_as("PJRT_ALLOCATOR_CUDA_ASYNC", bool, False) else "default",
```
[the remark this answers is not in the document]
will-cromar: Good catch. Removed some options entirely when the environment variable is not set.

will-cromar:
Registering the plugins right away is going to be inevitably flaky, because they will register their create options immediately. This doesn't allow the user to change settings at all after `import torch_xla`, such as `xr.use_spmd()`. We should instead defer `create_options` until the client is actually created.

I'll go learn how to define the plugin base class in C++ so the `pjrt_client` code can call it directly in a future PR.

will-cromar:
I started work on defining the interface in C++ in #6360.

Originally, I wanted to wait to merge this PR until #6022 goes in, but that is apparently blocked by an XLA bug. What do you all think of merging this PR now to unblock other work? It does not alter the default behavior at all, so it will not be a breaking change.

plugins/cuda/torch_xla_cuda_plugin/__init__.py
```diff
@@ -1,11 +1,46 @@
...
+        "platform_name":
+            "gpu",
+        # TODO(wcromar): make this configurable
+        "allocator":
+            "cuda_async" if xu.getenv_as("PJRT_ALLOCATOR_CUDA_ASYNC", bool,
+                                         False) else "default",
+        "memory_fraction":
+            xu.getenv_as("PJRT_ALLOCATOR_FRACTION", float, None),
+        "preallocate":
+            xu.getenv_as("PJRT_ALLOCATOR_PREALLOCATE", bool, None),
```
JackCaoG, line 33: do we ever expect user to set it?

torch_xla/csrc/runtime/pjrt_registry.cc
```diff
@@ -56,13 +66,37 @@ InitializePjRt(const std::string& device_type) {
...
-    std::optional<std::string> plugin_path = GetPjRtPluginPath(device_type);
-    if (plugin_path) {
+    std::optional<PluginEntry> plugin = GetPjRtPlugin(device_type);
+    if (plugin) {
       TF_VLOG(1) << "Initializing client for PjRt plugin " << device_type;
+
+      std::shared_ptr<xla::KeyValueStoreInterface> kv_store = nullptr;
+      if (plugin->init_coordinator) {
+        int global_process_rank = sys_util::GetEnvInt("RANK", 0);
+        int global_world_size = sys_util::GetEnvInt("WORLD_SIZE", 1);
```
JackCaoG, line 76: didn't we already get it in the `plugin->create_options` in python?

plugins/cuda/torch_xla_cuda_plugin/__init__.py
```diff
@@ -1,11 +1,46 @@
...
+
   def library_path(self) -> str:
     return os.path.join(os.path.dirname(__file__), 'pjrt_c_api_gpu_plugin.so')
 
   def physical_chip_count(self) -> int:
     # TODO: default to actual device count
-    return int(os.getenv('GPU_NUM_DEVICES', '1'))
+    return xu.getenv_as('GPU_NUM_DEVICES', int, 1)
+
+  def client_create_options(self) -> dict:
```
JackCaoG, line 15: who will call this `client_create_options ` method here?

will-cromar: `plugins.py` will call it during plugin registration in this PR. In the follow up, this will be called when the client is created.

torch_xla/csrc/runtime/pjrt_registry.cc
will-cromar: `create_options` doesn't include the rank and world size. In the future, we should actually just be creating the `XlaCoordinator` in Python and using it at `torch.distributed`'s `Store`. See https://github.com/pytorch/xla/pull/6289#pullrequestreview-1826006557

plugins/cuda/torch_xla_cuda_plugin/__init__.py
will-cromar: I copied this from the existing implementation. @vanbasten23 will know if a user requested it.

JackCaoG (approved):

plugins/cuda/torch_xla_cuda_plugin/__init__.py
```diff
@@ -10,4 +11,37 @@ def library_path(self) -> str:
 
   def physical_chip_count(self) -> int:
     # TODO: default to actual device count
-    return int(os.getenv('GPU_NUM_DEVICES', '1'))
+    return xu.getenv_as('GPU_NUM_DEVICES', int, 1)
```
vanbasten23, line 14: Perhaps irrelevant to this PR but just want to confirm that the `# TODO: default to actual device count` still holds, since `GPU_NUM_DEVICES` is not always set and the default value may not be 1.

plugins/cuda/torch_xla_cuda_plugin/__init__.py
```diff
@@ -10,4 +11,37 @@ def library_path(self) -> str:
...
+        "node_id":
+            global_process_rank,
+        "num_nodes":
+            global_world_size,
+    }
+
+    return {k: v for k, v in options.items() if v is not None}
+
+  def requires_xla_coordinator(self) -> bool:
+    return True
```
vanbasten23, line 47: I wonder why it always return True. For single processing, probably we don't need the coordinator? So should it depend on whether it is single processing?

plugins/cuda/torch_xla_cuda_plugin/__init__.py
vanbasten23: I see it's only used in `run_multiprocess`. So it sound using `GPU_NUM_DEVICES ` preserves the current behavior. Looks good to me then.

torch_xla/csrc/runtime/pjrt_registry.cc
vanbasten23, line 76: Sorry I didn't catch it earlier. What if the users start the single-host training in a non-torchrun way such as `PJRT_DEVICE=CUDA GPU_NUM_DEVICES=4 python3 xla/test/test_train_mp_imagenet.py`, then `global_world_size` should default to `local_world_size`? Similar reason for `global_process_rank`

plugins/cuda/torch_xla_cuda_plugin/__init__.py
will-cromar: Yeah, this is the same as the current behavior. Ideally this should check the PCI device IDs like we do for TPUs.

plugins/cuda/torch_xla_cuda_plugin/__init__.py
will-cromar: I think in a previous draft, I caught this case in `InitializePjrt`. But as @jonb377 says, a plugin may always require the coordinator. I'll go ahead and update this to `return global_world_size > 1`.

torch_xla/csrc/runtime/pjrt_registry.cc
will-cromar: This case is slightly wrong. The precedence should be `$PJRT_LOCAL_PROCESS_COUNT`, then `$WORLD_SIZE`, then `1` (default). I'll fix it.

vanbasten23: Shouldn't it be WORLD_SIZE -> PJRT_LOCAL_PROCESS_COUNT or LOCAL_WORLD_SIZE -> 1?
If we use torchrun, then WORLD_SIZE will be set.
If we `GPU_NUM_DEVICES=4 python3 `, non-torchrun for single-host-multi-process, then WORLD_SIZE is not set and we rely on `PJRT_LOCAL_PROCESS_COUNT or LOCAL_WORLD_SIZE`.

vanbasten23 (approved):
LGTM. Thanks!

jonb377 (approved):
LGTM!

PR #6298, opened 2024-01-11
benchmarks/nightly.sh: add -P flag to enable profiling

cota:


cota:
I'll edit this PR tomorrow so that on the plots it's clear that profiling is enabled.

benchmarks/nightly.sh
```diff
@@ -7,8 +7,9 @@ ACCELERATOR=a100
 OUTPUT_DIR=${HOME:?}
 WORKSPACE=$(date --utc +%Y-%m-%d)
 REPEAT=8
+ENABLE_PROFILING=0
 
-while getopts 'A:O:R:W:' OPTION
+while getopts 'A:O:PR:W:' OPTION
```
miladm, line 12: To clarify, is this meant to be `'A:O:P:R:W:'`?

zpcore:
LGTM!

Not release to this PR, in the script when you try:
```
  # Set up torchbench
  cd benchmark
  USE_CUDA=1 python install.py
  cd ..
```
Did you see the model installation process terminated due to fail to install a model? We use `--continue_on_fail` to prevent the issue.

zpcore (approved):

benchmarks/nightly.sh
zpcore: I don't think so, the `-P` seems don't need any input argument.

golechwierowicz (approved):
Thank you, Emilio!

cota:
> I'll edit this PR tomorrow so that on the plots it's clear that profiling is enabled.

Done.
I've also edited the output directory paths so that we do not mix profiling with non-profiling results.

cota:
> Did you see the model installation process terminated due to fail to install a model?

Maybe? I haven't looked at the logs. The idea is to either install all models, or fail. If we fail that means we won't have results for that day, but that's OK -- it's likely that in subsequent days the problem would be fixed.

The rationale for this is that we want the speedups to be consistent over time. That is, between two consecutive days, the benchmark set should be the same -- otherwise the "overall geomean speedup" we'd report could be misleading.

benchmarks/nightly.sh
```diff
@@ -141,11 +161,14 @@ for testname in inference training; do
       for tier in '' 1; do
 	# Note: these (as well as $tier) can be null; read them with ${FOO?}
 	# instead of ${FOO:?}.
-	TITLE_PFX=
+        TITLE_PFX=
```
vanbasten23, line 164: what does PFX stand for?

cota: PREFIX. I've updated the PR with s/PFX/PREFIX/.

Issue #6292, opened 2024-01-11
`test_trace_and_metrics` fail if PyTorch has CUDA support.


ysiraichi:
## 🐛 Bug

```bash
PJRT_DEVICE=CUDA python test/test_profiler.py
```

The command above fails with:

```
2024-01-11 02:59:59.706960: I torch_xla/csrc/runtime/pjrt_computation_client.cc:167] Initializing PjRt GPU client...
2024-01-11 02:59:59.707055: I torch_xla/csrc/runtime/pjrt_computation_client.cc:200] Getting StreamExecutorGpuClient for node_id=0, num_nodes=1
2024-01-11 02:59:59.730538: E external/xla/xla/stream_executor/cuda/cuda_driver.cc:282] failed call to cuInit: CUDA_ERROR_NOT_INITIALIZED: initialization error
2024-01-11 02:59:59.730584: I external/xla/xla/stream_executor/cuda/cuda_diagnostics.cc:137] retrieving CUDA diagnostic information for host: qgpu3
2024-01-11 02:59:59.730600: I external/xla/xla/stream_executor/cuda/cuda_diagnostics.cc:144] hostname: qgpu3
2024-01-11 02:59:59.730688: I external/xla/xla/stream_executor/cuda/cuda_diagnostics.cc:168] libcuda reported version is: 530.30.2
2024-01-11 02:59:59.730731: I external/xla/xla/stream_executor/cuda/cuda_diagnostics.cc:172] kernel reported version is: 530.30.2
2024-01-11 02:59:59.730745: I external/xla/xla/stream_executor/cuda/cuda_diagnostics.cc:253] kernel version seems to match DSO: 530.30.2
Process Process-33:
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/local/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "test/test_profiler.py", line 74, in train_worker
    test_profile_mp_mnist.train_mnist(
  File "xla/test/test_profile_mp_mnist.py", line 102, in train_mnist
    sample_count=600000 // flags.batch_size // xm.xrt_world_size())
  File "xla/torch_xla/core/xla_model.py", line 127, in xrt_world_size
    return runtime.world_size()
  File "xla/torch_xla/runtime.py", line 87, in wrapper
    return fn(*args, **kwargs)
[...]

PR #6302, opened 2024-01-12
Use spawn as the fork method for the profiler test.

ysiraichi:
Fix: #6292 

Summary of the changes:

- `train_worker` function was made top-level: spawn method requires it to be pickle-able
- Creating a new context where the fork method is spawn: needed for avoiding CUDA initialization issues

cc @miladm @JackCaoG

### DIFF
--- a/test/test_profiler.py
+++ b/test/test_profiler.py
@@ -14,6 +14,31 @@ import torch_xla.debug.profiler as xp
 import torch_xla.utils.utils as xu
 
 
+# This function must remain a top-level function. Using spawn
+# as the fork method requires this function to be pickle-able.
+def train_worker(port, training_started):
+  flags = args_parse.parse_common_options(
+      datadir='/tmp/mnist-data',
+      batch_size=16,
+      momentum=0.5,
+      lr=0.01,
+      num_epochs=10)
+  flags.fake_data = True
+  flags.profiler_port = port
+
+  # Disable programmatic profiling
+  flags.profile_step = -1
+  flags.profile_epoch = -1
+  flags.profile_logdir = None
+  flags.profile_duration_ms = -1
+
+  test_profile_mp_mnist.train_mnist(
+      flags,
+      training_started=training_started,
+      dynamic_graph=True,
+      fetch_often=True)
+
+
 class ProfilerTest(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


ysiraichi:
@JackCaoG this is a friendly reminder. Do you have some time to review this PR?

JackCaoG:
lol totally forgot, let me take a look.

JackCaoG (approved):

test/test_profiler.py
```diff
@@ -51,33 +76,15 @@ def _check_trace_namespace_exists(self, path):
                     f'Expected "build_graph" trace in: {path}')
 
   def test_trace_and_metrics(self):
+    # Create a new context for forking processes with the spawn method.
+    # This is necessary so as to avoid CUDA initialization issues when
+    # both PyTorch and PyTorch/XLA were compiled with CUDA support.
+    context = multiprocessing.get_context("spawn")
```
vanbasten23, line 82: @ysiraichi IIUC, the failure happens when we initialize CUDA in the parent process and use CUDA in the child process. I wonder where we initialize CUDA in the parent process before your change in this PR.

ysiraichi: After some investigation, I believe it comes from importing `torch_xla`. Specifically, the following chain:

- `torch_xla`
- `stablehlo`
- `dynamo_bridge`
- `torch._inductor.fx_passes.post_grad`

I guess, one way to solve this issue is to move `ConstructorMoverPass` out of inductor tree.

vanbasten23: Thanks for the reply.

Curious how do you know `torch._inductor.fx_passes.post_grad` initializes a CUDA context.
Also, what do you mean by `move ConstructorMoverPass out of inductor tree`?

ysiraichi: > how do you know torch._inductor.fx_passes.post_grad initializes a CUDA context.

Just by commenting it out, the problem goes away.

> what do you mean by move ConstructorMoverPass out of inductor tree?

This class is declared under inductor module. Importing it means that we have to load the inductor module itself, which initializes a CUDA context. If the class is declared somewhere else (which is possible, since it doesn't really depend on anything of inductor), that initialization goes away

Issue #6036, opened 2023-12-06
2.2 backport PR request list


zpcore:
The issue is to track 2.2 release backport. 

For any PRs you want to backport to 2.2, please reply with following:
- Original PR link
- Reason to backport
- 2.2 backport PR link

JackCaoG:
`Master PR`: https://github.com/pytorch/xla/pull/5956

`Backport PR`: None

Reason:
AWS need this to pr to land to enable the coalesce reduce_scatter

JackCaoG:
Master PR: https://github.com/pytorch/xla/pull/5958
Status: In Review

Backport PR: None

Reason:
Enhance the `PT_XLA_DEBUG` tool to truncate the python frame count.

JackCaoG:
Master PR: https://github.com/pytorch/xla/pull/6039
status: In Review

Backport PR: None

Reason:
Update the trouble shooting doc with pt_xla_debug.

jonb377:
Master PR: #5800 
Status: Merged
Backport PR: None

Reason:
Persistent compilation caching

jonb377:
Master PR: #5804
Status: Merged
Backport PR: None

Reason:
Persistent compilation caching

jonb377:
Master PR: #6046
Status: Merged
Backport PR: None

Reason:
Persistent compilation caching

ManfeiBai:
Master PR: https://github.com/pytorch/xla/pull/6053
status: merged

Backport PR: https://github.com/pytorch/xla/pull/6079
status: in review

Reason: Hot-fix random duplicated trace registered TPU CI failure

jeffhataws:
Master PR: https://github.com/pytorch/xla/pull/6059
status: open

Reason: Add a separate out-of-place all-gather. Add missing test for all-gather coalesce out, and fixed a bug for list size check in output!=None case.

jeffhataws:
Master PR: https://github.com/pytorch/xla/pull/6058
status: open

Reason: Add a separate out-of-place reduce-scatter and accompanying test.

jonb377:
Master PR: https://github.com/pytorch/xla/pull/6075
Status: Merged

Backport PR: https://github.com/pytorch/xla/pull/6091
Status: In review

Reason: Eliminate flakiness in profiler tests

JackCaoG:
Master PR: https://github.com/pytorch/xla/pull/6071#event-11200521507
Status: Merged

Backport PR: None
Status:

[...]

PR #6304, opened 2024-01-13
Correct order of constant vs args

qihqi:


JackCaoG:
Can you reply to https://github.com/pytorch/xla/issues/6036 for backport. I guess this qualified as bug fixes?

test/stablehlo/test_exports.py
```diff
@@ -23,7 +23,7 @@ def __init__(self):
     super().__init__()
 
   def forward(self, a):
-    return a * torch.tensor(3)
+    return a / torch.tensor(3)
```
ManfeiBai, line 26: it looks like a large change, and why do we change logic here from * to / ?

ManfeiBai (approved):
LGTM

test/stablehlo/test_exports.py
qihqi: This is a test case. So the error is that I got the order of constant vs. arg wrong. But because a* b = b*a so my old test case couldnt catch the error. but because a / b != b / a so errors of ordering shows up with this test case.

PR #6309, opened 2024-01-17
Update xla_native_functions.yaml

ManfeiBai:


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1661,6 +1661,14 @@ at::Tensor XLANativeFunctions::xlogy(const at::Tensor& self,
       bridge::GetXlaTensor(self), bridge::GetXlaTensor(other)));
 }
 
+std::tuple<at::Tensor, at::Tensor> XLANativeFunctions::map(
+    const Callable f, const at::Tensor& xs) {
```
ManfeiBai, line 1665: #TODO no Callable used in this file, need add support for Callable here

ManfeiBai: ok, so we use `Callable` here due to match JAX style, and JAX use this seems due to trace to JAXprs,

and what we actually want to do is to match torch.control_flow.map, so we should match what torch used here

but seems torch implemented this in python level and this is a pure python func, so do we have any example that accept python func?

PR #6314, opened 2024-01-18
Improve build process for PJRT plugins

will-cromar:
Note: directly running `setup.py` is now [deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html). Both plugins are now installable by running `pip install plugin/...`, which runs `setuptools`/`setup.py` in the background. We'll have to deal with this problem in our main `setup.py` build soon too.

- Build plugins within each `setup.py`
    - Note: `BazelExtension` and `BuildBazelExtension` are not necessary for the plugins because they don't expose any functions to Python like `_XLAC` does. Instead, I opted to run bazel in a subprocess before running `setup`. According to `setuptools`'s docs, the "right" way to extend `setuptools` is to [add commands](https://setuptools.pypa.io/en/latest/userguide/extension.html), but custom `setup.py` commands are being deprecated (since you have to invoke it through pip). The documentation doesn't make it clear what the right thing to do here is, so I opted for the most straightforward solution. (Not to mention, the documentation about how to actually write custom commands is also sparse.)
- Move some common bazel setup code to `build_util` and symlink it to plugin directories
  - Note: I can't figure out how to use a local file in a `setup.py` with `pip`'s build isolation, so [I disabled isolation with `--no-build-isolation`](https://stackoverflow.com/questions/62889093/what-does-no-build-isolation-do). The only way I see how to do this is to install our build-time library as a package, but it seems silly to create a package that will only be used to build our real packages in the same repository.

The current state of Python packaging is kind of a mess. See these [three](https://gregoryszorc.com/blog/2023/10/30/my-user-experience-porting-off-setup.py/) [blog](https://chriswarrick.com/blog/2023/01/15/how-to-improve-python-packaging/) [posts](https://chriswarrick.com/blog/2024/01/15/python-packaging-one-year-later/) for a pretty good summary of grievances with the current situation. I'm very open to feedback on this PR, since I'm not at all convinced I'm using the new tools the "right" way (if there even is one).

The bright side is, [`pyproject.toml`](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) will make it _much_ easier to get away from `setuptools` if we want to do that. Especially when we have to deal with our main `setup.py`, we should look to see if any of the alternatives better support our use case with less hackery. (I did some research this week, but most alternatives, including PyPA's `hatch` do not support C++ extensions).

### DIFF
--- a/plugins/cuda/pyproject.toml
+++ b/plugins/cuda/pyproject.toml
@@ -15,4 +15,4 @@ requires-python = ">=3.8"
 torch_xla_cuda_plugin = ["lib/*.so"]
 
 [project.entry-points."torch_xla.plugins"]
-gpu = "torch_xla_cuda_plugin:GpuPlugin"
+cuda = "torch_xla_cuda_plugin:CudaPlugin"
--- a/plugins/cuda/pyproject.toml
+++ b/plugins/cuda/pyproject.toml
@@ -12,7 +12,7 @@ description = "PyTorch/XLA CUDA Plugin"
 requires-python = ">=3.8"
 
 [tool.setuptools.package-data]
-torch_xla_cuda_plugin = ["*.so"]
+torch_xla_cuda_plugin = ["lib/*.so"]
 
 [project.entry-points."torch_xla.plugins"]
 gpu = "torch_xla_cuda_plugin:GpuPlugin"
--- a/plugins/cuda/pyproject.toml
+++ b/plugins/cuda/pyproject.toml
@@ -6,9 +6,9 @@ build-backend = "setuptools.build_meta"
 name = "torch_xla_cuda_plugin"
 version = "0.0.1"
 authors = [
-    {name = "Will Cromar", email = "wcromar@google.com"},
+    {name = "PyTorch/XLA Dev Team", email = "pytorch-xla@googlegroups.com"},
 ]
-description = "CUDA Plugin"
+description = "PyTorch/XLA CUDA Plugin"
 requires-python = ">=3.8"
 
 [tool.setuptools.package-data]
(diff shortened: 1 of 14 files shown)


will-cromar:
baf70be adds the CUDA plugin to the CI build, but the default behavior will remain the same. If it unnecessarily extends the build time, I'll revert it.

will-cromar:
The build failed, so I reverted the change. I'll debug this in a future PR.

plugins/cuda/pyproject.toml
```diff
@@ -6,13 +6,13 @@ build-backend = "setuptools.build_meta"
 name = "torch_xla_cuda_plugin"
 version = "0.0.1"
 authors = [
-    {name = "Will Cromar", email = "wcromar@google.com"},
+    {name = "PyTorch/XLA Dev Team", email = "pytorch-xla@googlegroups.com"},
```
JackCaoG, line 9: haha

setup.py
```diff
@@ -201,20 +190,6 @@ def run(self):
   # Copy libtpu.so into torch_xla/lib
   maybe_bundle_libtpu(base_dir)
 
-DEBUG = _check_env_flag('DEBUG')
-IS_DARWIN = (platform.system() == 'Darwin')
-IS_WINDOWS = sys.platform.startswith('win')
-IS_LINUX = (platform.system() == 'Linux')
-GCLOUD_KEY_FILE = os.getenv('GCLOUD_SERVICE_KEY_FILE', default='')
-CACHE_SILO_NAME = os.getenv('SILO_NAME', default='dev')
-BAZEL_JOBS = os.getenv('BAZEL_JOBS', default='')
```
JackCaoG, line 210: are these no longer needed?

will-cromar: The OS flags are irrelevant since we only support linux. The others were moved into local variables

JackCaoG (approved):

PR #6318, opened 2024-01-18
Update doc on new modules.

golechwierowicz:
1. Add section about the verification module.
2. Add section about microbenchmarks.


vanbasten23 (approved):
Thanks @golechwierowicz !

PR #6324, opened 2024-01-18
Use SGD for list of models from PyTorch.

ysiraichi:
This PR replaces Adam for SGD optimizer for a few models that are tight on memory. This list was taken from PyTorch's main repository:

https://github.com/pytorch/pytorch/blob/eba5d5485dd3232737ac789f1614f4cbc09220a4/benchmarks/dynamo/common.py#L214-L252


vanbasten23 (approved):

benchmarks/torchbench_model.py
```diff
@@ -29,6 +29,27 @@
     "detectron2_fcos_r_50_fpn",
 }
 
+# torchbench models that might OOM using Adam.
+# This list was extracted from PyTorch's repository: benchmarks/dynamo/common.py
```
golechwierowicz, line 33: So in the link it is mentioned that they use Adam optimiser in the CI. What optimiser is used in the official PyTorch HUD? I'm asking because we want to be as close as possible.

ysiraichi: Good catch. So, maybe we want to get the previous set: `BENCHMARK_USE_SGD`, which is used outside CI.

PR #6325, opened 2024-01-18
Clean-up CUDA cache.

ysiraichi:
This PR clears CUDA cache on 2 cases:

- `benchmark` instance was deleted (i.e. it was first instantiated with `cuda` as device, but it's not needed anymore)
- inner model has to be moved to an XLA device
    - since #6296, models are first initialized in the original accelerator, and then moved to XLA device

Additionally, it also fixes a bug where the instantiated `benchmark` wouldn't be deleted if it didn't have a default precision set (in `default_precision_flag` property).


vanbasten23:
Sorry for the late review. What issue do we get if we don't clean up the CUDA cache? @ysiraichi

ysiraichi:
I was getting OOM when training `llama_v2_7b_16h` with both inductor and `openxla` backends.

PR #6327, opened 2024-01-18
Update contribution guide to use python setup.py develop.

changm:
Also added a gotcha section based on some things others may hit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,12 +35,12 @@ We recommend you to use our prebuilt Docker image to start your development work
 * Build PyTorch
   ```Shell
   cd /pytorch/
-  python setup.py install
+  python setup.py develop
   ```
 * Build PyTorch/XLA
   ```Shell
   cd xla/
-  python setup.py install
+  python setup.py develop
   ```
 
 ### Build PyTorch/XLA from source with GPU support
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -107,3 +107,7 @@ Then run `test/run_tests.sh` and `test/cpp/run_tests.sh` to verify the setup is
 ### Useful materials
 1. [OP Lowering Guide](https://github.com/pytorch/xla/blob/master/OP_LOWERING_GUIDE.md)
 2. [CODEGEN MIGRATION GUIDE](https://github.com/pytorch/xla/blob/master/CODEGEN_MIGRATION_GUIDE.md)
+
+### Sharp Edges
+
+* If local changes aren't visible, ensure you built both PyTorch and PyTorch/XLA with `python setup.py develop` and not `python setup.py install`. If you're unsure, uninstall existing pytorch/xla with `pip uninstall torch_xla` and `pip uninstall torch`, then rebuild PyTorch and PyTorch/XLA.
```
will-cromar, line 113: I think the issue you ran into was using `develop` and then `install` (or the opposite) which can create problems. I generally `install` PyTorch and `develop` PyTorch/XLA without issue. I would shorten the advice here to:

>  If local changes aren't visible, uninstall existing pytorch/xla with `pip uninstall torch_xla` and `pip uninstall torch`, then rebuild PyTorch and PyTorch/XLA with `python setup.py develop` or `python setup.py install`

will-cromar (approved):
Otherwise LGTM. Thanks!

CONTRIBUTING.md
changm: Thanks updated the text to reflect what you have here.

PR #6166, opened 2023-12-14
Promote int to float for tanh operation (consistent with Pytorch)

ManfeiBai:
Fixes https://github.com/pytorch/xla/issues/5903

---

`test_aten_tanh_2` failed due to:
```
RuntimeError: Error while lowering: [] aten::tanh, location=__call__@_ops.py:755, xla_shape=s32[10,10]{1,0}, dynamic_dims: ()
XLA builder error: INVALID_ARGUMENT: Expected element type in shape to be floating or complex for tanh operation; got S32.: 
```
and tested `torch.tanh` with PyTorch and PyTorch/XLA found tanh support int in torch and didn't support int in torch_xla:
```
# PJRT_DEVICE=TPU python
Python 3.10.13 (main, Sep 11 2023, 13:44:35) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> a = torch.randn((10, 10)).to(torch.float32)
>>> b = torch.tanh(a)
>>> 
>>> c = torch.randn((10, 10)).to(torch.float16)
>>> d = torch.tanh(c)
>>> 
>>> e = torch.randint(0, 10, (10, 10)).to(torch.int32)
>>> f = torch.tanh(e)
>>> 
```
and
```
# PJRT_DEVICE=TPU python
Python 3.10.13 (main, Sep 11 2023, 13:44:35) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch_xla
>>> import torch
>>> import torch_xla.core.xla_model as xm
>>> 
>>> m = torch.randn((10, 10)).to(torch.float32).to(xm.xla_device())
>>> n = torch.tanh(m)
>>> xm.mark_step()
>>> 
>>> o = torch.randn((10, 10)).to(torch.float16).to(xm.xla_device())
>>> p = torch.tanh(o)
>>> xm.mark_step()
>>> 
>>> p
tensor([[-0.4927,  0.6211,  0.7344, -0.6538, -0.1163,  0.6934, -0.5459,  0.8491,
          0.3271,  0.6426],
        [ 0.8691,  0.0098, -0.6626, -0.5479, -0.1429,  0.8721,  0.7822, -0.5654,
         -0.2031, -0.7305],
        [-0.3169, -0.4194, -0.5518, -0.9673, -0.7153,  0.3308, -0.9453,  0.2111,
          0.0225,  0.3533],
[...]

PR #6329, opened 2024-01-19
[Cherry-pick] Update test_spmd_debugging.py to avoid code test code self and Promote int to float for tanh operation (#6263)(#6166)

ManfeiBai:


zpcore (approved):
LGTM!

PR #6333, opened 2024-01-19
Add tabulate mode for printing simple speedups.

golechwierowicz:
In this PR we add:
1. Parametrisation of aggregation of total times from benchmarked results.
2. `tab` format which shows speedups in CLI.

Command
```
python3 aggregate.py --test=inference --format=tab --agg=min --report=latest --format=tabular --accelerator=a100 do_bench_11_01_24.jsonl
```

Sample output
```
╒═════════════════════════╤═══════════╤════════════════╤═════════════════════╕
│ model                   │   speedup │   xla std (ms) │   inductor std (ms) │
╞═════════════════════════╪═══════════╪════════════════╪═════════════════════╡
│ hf_Bert                 │      1.00 │           0.00 │                0.00 │
├─────────────────────────┼───────────┼────────────────┼─────────────────────┤
│ drq                     │      1.00 │           0.00 │                0.00 │
╘═════════════════════════╧═══════════╧════════════════╧═════════════════════╛

Geomean: 1.00x.
```

### DIFF
--- a/benchmarks/aggregate.py
+++ b/benchmarks/aggregate.py
@@ -56,7 +56,38 @@ def skip_model(args, model_name: str):
           re.search("|".join(args.exclude), model_name, re.I))
 
 
-def process_file(args, results_map: Dict[str, Any], filename: str):
+class DatapointStrategy:
+
+  def avg(total_times):
+    # Skip the first three elements in total_times[]; the first one
+    # includes compilation time, the 2nd and 3rd warm up the caches.
+    skip_elems = 3
+    if len(total_times) <= skip_elems:
+      logger.warning(
+          f"Skipping the datapoint calculation due less data than expected. Got {len(total_times)} v. expected {skip_elems}"
+      )
+      return None
+    avg = np.average(total_times[skip_elems:])
+    # Standard deviation of the sample, i.e. N-1 denominator.
+    # Note: avoid NaN when we compute the std with just one sample.
+    std = np.std(
+        total_times[skip_elems:],
+        ddof=1) if len(total_times) > skip_elems + 1 else 0.0
+    return Datapoint(avg, std)
+
+  def min(total_times):
+    if len(total_times) == 0:
+      return None
+    val = np.min(total_times)
+    std = np.std(total_times)
+    # TODO: Rename `Datapoint` avg to something more generic
+    return Datapoint(avg=val, std=std)
+
+
+def process_file(args,
+                 results_map: Dict[str, Any],
+                 filename: str,
+                 datapoint_strategy=DatapointStrategy.avg):
   fields = {
       'experiment': ['accelerator_model', 'batch_size', 'dynamo', 'test'],
       'metrics': [],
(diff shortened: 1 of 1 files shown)


cota:
I've left a few small comments in the code.

Some bigger questions here.

## 1. agg=min mode

Why is it needed?

Note that we create a Datapoint with (min, std), which fails to characterize the distribution of the samples [something that (avg, std) does, which is why Datapoint is {.avg, .std}]. A second effect of agg=min is that on the table we report the speedup of the min's, but cannot provide a sense of the uncertainty of that speedup, other than writing the standard deviation of the sample sets where the min's come from [we cannot do it because the standard deviation is centered on the avg of the samples, not on the min]. I think this can confuse people. Furthermore, why include all runs when computing the std? This will increase dispersion since compilation times >> run times.

I believe just reporting the speedup of the means (after discarding outliers) is more robust and will avoid confusion. We could then just report the speedup and its corresponding error, which is most intuitive.

## 2. Lack of support for other output formats/modes

If it doesn't make sense to support other outputs formats, and not even training-latest, isn't that a sign that this output mode should be a separate script? Consider a small python script that takes the CSV output from `latest` and prints a table out of it. If we want to report results vs. the latest inductor, that's trivial to do; just recompute the speedup vs. the latest inductor, i.e. just a division between the two speedups (they have the same baseline so we can do this). The errors are trivial to compute as well (sqrt of the two relative errors), just like  [compute_speedup](https://github.com/pytorch/xla/blob/e93597de0d95b3962a25e3fde36ad39312c4f9ce/benchmarks/aggregate.py#L123) does. The only downside of this is that the error might be larger than necessary if the old inductor had high error, but so far I have not seen this happen often.

golechwierowicz:
> ## 2. Lack of support for other output formats/modes
> If it doesn't make sense to support other outputs formats, and not even training-latest, isn't that a sign that this output mode should be a separate script? Consider a small python script that takes the CSV output from `latest` and prints a table out of it. If we want to report results vs. the latest inductor, that's trivial to do; just recompute the speedup vs. the latest inductor, i.e. just a division between the two speedups (they have the same baseline so we can do this). The errors are trivial to compute as well (sqrt of the two relative errors), just like [compute_speedup](https://github.com/pytorch/xla/blob/e93597de0d95b3962a25e3fde36ad39312c4f9ce/benchmarks/aggregate.py#L123) does. The only downside of this is that the error might be larger than necessary if the old inductor had high error, but so far I have not seen this happen often.

My plan was to start simple, and add support gradually to training, etc.
That being said, I wanted to have some `code` which calculates a speedup over Inductor in a single run, and I was unsure how open we are to an idea of the script fragmentation. If we're ok (seems like you are already), with having two scripts for seemingly computing speedup then I'd be more than happy to write functionality in this PR in a separate script.

As for the "why min" question, the idea was for this to be used with `--pure-wall-time` option, which does not have compilation issues, and effectively tries to reduce CPU noise from both compilers.

cota:
I see. My point was, if this is a one-off use, a separate script is better. OTOH if we have plans to make this comprehensive, then integrating it into aggregate.py is better IMO.

For the "latest" baseline, we can add a --baseline=latest flag that should simplify this. That way we can decouple the baseline calculation from the chosen report. I will send a patch today.

We can then add to aggregate.py the tabulate mode. Does this sound good?

cota:
> For the "latest" baseline, we can add a --baseline=latest flag that should simplify this. That way we can decouple the baseline calculation from the chosen report. I will send a patch today.

https://github.com/pytorch/xla/pull/6349

golechwierowicz:
Do we still consider a tabulate option as valuable? The problem I'm trying to solve is that csv format is hard to read. Does it still make sense to implement and extend it? (and remove the `min` logic from this PR)

Or do you have more useful idea/work on this?

cota:
Please take a look at this: https://github.com/cota/pytorch-xla/pull/1
It emits the same output as CSV, but with `tabulate`. I also changed the headers a bit to make it less wide on the terminal.
If you like it I'll send it as a PR to the main repo -- it's a draft PR on my fork because its base commit is not yet on master.

golechwierowicz:
> Please take a look at this: [cota#1](https://github.com/cota/pytorch-xla/pull/1) It emits the same output as CSV, but with `tabulate`. I also changed the headers a bit to make it less wide on the terminal. If you like it I'll send it as a PR to the main repo -- it's a draft PR on my fork because its base commit is not yet on master.

Yes! This looks good.

golechwierowicz:
Closing as Emilio already has some code for this.

PR #6335, opened 2024-01-19
Pin to pytorch#117182 (tags/v2.2.0-rc8)

yeounoh:
Pin to https://github.com/pytorch/pytorch/pull/117182 --> tags/v2.2.0-rc8


torch_patches/.torch_pin
```diff
@@ -0,0 +1 @@
+#117182
```
JackCaoG, line 1: please follow the example of https://github.com/pytorch/xla/blob/r2.1/.torch_pin. I think we should just pin it to releases/2.2 branch for now and update it to a tag after release it out.

zpcore:
I think it should be `tags/v2.2.0-rc8`.

ManfeiBai (approved):
vote to `tags/v2.2.0-rc8`

torch_patches/.torch_pin
ManfeiBai: do we wanna update to `tags/v2.2.0-rc8` directly?

yeounoh: Thank you, changed!

torch_patches/.torch_pin
```diff
@@ -0,0 +1 @@
+tags/v2.2.0-rc8
```
JackCaoG, line 1: let's move this outside of `torch_patches`. I think that's what we want to do for release branches.

yeounoh: Done!

JackCaoG (approved):

ManfeiBai (approved):
LGTM, Thanks, Yeounoh!

zpcore (approved):

PR #6349, opened 2024-01-22
aggregate.py: add --baseline option

cota:
This allows us to pick the latest set of Inductor results as the baseline. This is useful when doing a direct comparison between Inductor and XLA for the latest date.


golechwierowicz (approved):

cota:
Note: I've changed the tests to pick flags from inside the test file via a comment, e.g. `# ARGS: --foo=bar --foo2=baz`. When diffing the generated files we ignore that line by passing `-I` to `git diff`.
Rationale: I have coming changes that need more flexibility with the flags and embedding them in the file name gets unwieldy.

Issue #5967, opened 2023-12-01
[torchbench] Check failed: cachedComputation


ysiraichi:
## 🐛 Bug

Running a few torchbench benchmarks, using dynamo+openxla backend, ends up in an assertion failure:

```python
Traceback (most recent call last):
  File "torch/_dynamo/eval_frame.py", line 489, in _fn
    return fn(*args, **kwargs)
  File "benchmarks/dynamo/torchbench.py", line 544, in forward_pass
    return mod(*inputs)
  File "torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "torchbenchmark/models/cm3leon_generate/model.py", line 1113, in forward
    def forward(self, src_tokens):
  File "torch/_dynamo/eval_frame.py", line 489, in _fn
    return fn(*args, **kwargs)
  File "torch/_dynamo/external_utils.py", line 17, in inner
    return fn(*args, **kwargs)
  File "torch/_functorch/aot_autograd.py", line 4963, in forward
    return compiled_fn(full_args)
  File "torch/_functorch/aot_autograd.py", line 2017, in g
    return f(*args)
  File "torch/_functorch/aot_autograd.py", line 3164, in runtime_wrapper
    all_outs = call_func_with_args(
  File "torch/_functorch/aot_autograd.py", line 2041, in call_func_with_args
    out = normalize_as_list(f(args))
  File "torch/_functorch/aot_autograd.py", line 2145, in rng_functionalization_wrapper
    return compiled_fw(args)
  File "torch/_functorch/aot_autograd.py", line 2017, in g
    return f(*args)
  File "torch/_dynamo/backends/torchxla.py", line 51, in fwd
    return compiled_graph(*args)
  File "torch/fx/graph_module.py", line 736, in call_wrapped
    return self._wrapped_call(self, *args, **kwargs)
  File "torch/fx/graph_module.py", line 315, in __call__
    raise e
  File "torch/fx/graph_module.py", line 302, in __call__
    return super(self.cls, obj).__call__(*args, **kwargs)  # type: ignore[misc]
  File "torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
[...]

PR #6352, opened 2024-01-22
disable some models for torchbench

zpcore:
1. Disable `opacus_cifar10` for TPU due to stackdump issue.
2. Disable `cm3leon_generate`, `hf_T5_generate` for GPU due to no install.py in torchbench.


benchmarks/torchbench_model.py
```diff
@@ -60,6 +60,7 @@
 
 # Skip the experiment of a model if any of the experiment configs in the list is fully matched
 DENY_LIST = {
+    "cm3leon_generate": [{}],  # no install.py
```
ysiraichi, line 63: Why are we skipping `cm3leon_generate` and `hf_T5_generate` benchmarks? Although they are skipped in the pytorch/benchmark `install.py` script, as far as I remember, they still do run successfully (at least on inductor).

ysiraichi (changes requested):

benchmarks/torchbench_model.py
zpcore: Thanks, if we don't care about `install.py`, maybe we just keep the inductor run and skip other configs? I tried model `cm3leon_generate` at  in L4, it tooks more than 1 hour to execute the model without returning results. WDYT?

zpcore: Also I notice that if we set `DEFAULT_TRAIN_BSIZE ` when we install models. If we don't install the model, the test will throw out error: 
```
File "/home/ml-auto-solutions/benchmark/torchbenchmark/util/model.py", line 202, in _determine_batch_size
    raise NotImplementedError(f"Model's {'DEFAULT_TRAIN_BSIZE' if self.test == 'train' else 'DEFAULT_EVAL_BSIZE'} is not implemented.")
NotImplementedError: Model's DEFAULT_TRAIN_BSIZE is not implemented.
```

ysiraichi: > maybe we just keep the inductor run and skip other configs?

I guess we could do this. According to #5932, both models timeout (after 30 mins).

> If we don't install the model, the test will throw out error

I don't think that's exactly how it works. A model only has an `install.py` file if there are any required dependecies/action needed before running the model (e.g. installing dependencies with `pip`).

That said, some models don't define `DEFAULT_TRAIN_BSIZE`, which probably means that the model wasn't designed for training (guess).

zpcore: I add a new argument `--strict-compatible` (value is default to False which has no impact to current setup) to choose between whether we want to involve those stackdump or timeout models. We don't what to wait for timeout or stackdump since this is break up the benchmark automation system when we do remote execution.

ysiraichi: Cool. But, maybe we should skip those only when `"xla": "PJRT"`, since inductor runs successfully. What do you think?

benchmarks/torchbench_model.py
```diff
@@ -124,6 +124,22 @@
...
 
+# This strict deny list denies tests that hold for too long and timeoout.
+STRICT_DENY_LIST = {
+    **{
+        "cm3leon_generate": [{
+            "test": "train",
+        }],  # no install.py
+        "hf_T5_generate": [{
+            "test": "train",
+        }],  # no install.py
```
ysiraichi, line 135: - I think these `STRICT_DENY_LIST` items (`{ "test": "train" }`) can be made into `DENY_LIST`, since these models fail to run training even on inductor 
- Shouldn't these be checking for `{ "xla": "PJRT" }`, since those are the ones that timeout (inductor doesn't)?

benchmarks/experiment_runner.py
```diff
@@ -879,6 +880,12 @@ def __str__(self):
       default=False,
       help="""If set, verifies the model output with PT Eager mode, and saves relative error to the output file."""
   )
+  parser.add_argument(
+      "--strict-compatible",
+      action="store_true",
+      default=False,
```
ysiraichi, line 886: I believe that setting `action="store_true"` already sets the default accordingly.

benchmarks/torchbench_model.py
zpcore: I just tried the eager mode for GPU with `xla:null,dynamo:null` also, it passes for those two models. Did you see the failing somewhere?

zpcore: In addition, LTC mode with `xla:PJRT,dynamo:null` with cause timeout.

ysiraichi: - `xla:PJRT,dynamo:null`: both timeout
- `xla:PJRT,dynamo:openxla`: both end up with the same issue https://github.com/pytorch/xla/issues/5967

zpcore: That's true, let's just skip PJRT backend as you mentioned.

benchmarks/experiment_runner.py
zpcore: Hmm, right. I also removed other default=False when action="store_true" in the file.

benchmarks/torchbench_model.py
zpcore: Moved `cm3leon_generate` and `hf_T5_generate` from strict deny list to general list.

benchmarks/torchbench_model.py
```diff
@@ -56,13 +56,28 @@
...
+    ],  # no install.py
+    "hf_T5_generate": [
+        {
+            "test": "train",
+        },
+        {
+            "test": "eval",
+            "xla": "PJRT",
+        },
+    ],  # no install.py
```
ysiraichi, line 80: Given your definition for "strict-compatible", I think `{"test": "eval", "xla": "PJRT"}` should be there, shouldn't it? It's not that they fail, but they do timeout.

ysiraichi: Just to make it explicit: with this setting, we are also skipping `dynamo+openxla` backend. While that doesn't timeout, it does run into the [`cachedComputation` check error](https://github.com/pytorch/xla/issues/5967).

ysiraichi (approved):
Just one more comment, but LGTM otherwise.

benchmarks/torchbench_model.py
zpcore: Thanks for the info!

zpcore: Right, they do timeout, but we want to also kick out the timeout test when we do test automation to prevent job failure.

PR #6354, opened 2024-01-22
Update README.md for cicleci

wonjoolee95:
Update README.md for cicleci

---

We had an issue recently where we rebased the fix PR for a breaking change, which made the commit hash obsolete. As mentioned by @JackCaoG, we have two ways to fix this:
1. Ask upstream to pin to branch name, instead of commit
2. Do not rebase after upstream pin to the commit

I prefer the first option as it's safer. However, it seems like all the .github pins are made to commits in PyTorch -- https://github.com/pytorch/pytorch/tree/main/.github. For now, we'll do the second option to update the documentation.


JackCaoG (approved):

wonjoolee95:
Merging now since it's only documentation changes.

PR #6355, opened 2024-01-22
[SPMD] update test format

ManfeiBai:
fix format in test

### DIFF
--- a/test/spmd/test_spmd_debugging.py
+++ b/test/spmd/test_spmd_debugging.py
@@ -191,10 +191,7 @@ class DebuggingSpmdTest(test_xla_sharding_base.XlaShardingTest):
         pad_edge=False,
         box=rich.box.SQUARE if not use_color else None)
     col = []
-    alltpus = xr.device_type() + ' [0'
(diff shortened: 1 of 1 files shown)


test/spmd/test_spmd_debugging.py
```diff
@@ -193,7 +193,7 @@ def test_single_host_replicated_tpu(self):
     col = []
     alltpus = xr.device_type() + ' [0'
     for i in range(xr.global_runtime_device_count() - 1):
-      alltpus = alltpus + ',' + str(i + 1)
+      alltpus = alltpus + ', ' + str(i + 1)
     alltpus = alltpus + ']'
```
jonb377, line 197: Could we use `str(list(range(xr.global_runtime_device_count())))` instead of building the string manually? The resulting format looks like Python's string representation of lists.

jonb377 (approved):
LGTM!

test/spmd/test_spmd_debugging.py
ManfeiBai: thanks, Jon, good point, updated

JackCaoG (approved):

PR #6356, opened 2024-01-22
Add documentation around CheckpointManager

jonb377:
Add documentation for CheckpointManager in the SPMD docs.

### DIFF
--- a/docs/spmd.md
+++ b/docs/spmd.md
@@ -270,6 +270,56 @@ dist_cp.load_state_dict(
 model.load_state_dict(state_dict["model"])
 ```
 
+#### CheckpointManager
+
+The experimental [CheckpointManager](https://github.com/pytorch/xla/blob/master/torch_xla/experimental/distributed_checkpoint/manager.py#L40)
+interface provides a higher-level API over the `torch.distributed.checkpoint`
+functions to enable a few key features:
+
+- **Managed checkpoints**: Each checkpoint taken by the `CheckpointManager` is
+identified by the step at which it was taken. All steps tracked are accessible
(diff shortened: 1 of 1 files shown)


docs/spmd.md
```diff
@@ -270,6 +270,56 @@ dist_cp.load_state_dict(
 model.load_state_dict(state_dict["model"])
 ```
 
+#### CheckpointManager
+
+The experimental [CheckpointManager](https://github.com/pytorch/xla/blob/master/torch_xla/experimental/distributed_checkpoint/manager.py#L40)
```
JackCaoG, line 275: does this feature tie to the SPMD? If not I felt like it should be in a separate doc.

jonb377: We currently only support `torch.distributed.checkpoint` with SPMD, but CheckpointManager only builds on `torch.distributed.checkpoint` and isn't SPMD-specific.

JackCaoG: Do we have any plan to support non-spmd use cases? How difficult will it be?

alanwaketan: I guess that's nice to have to support non-SPMD case, but I guess the future is to move everything to be on top of SPMD?

jonb377: Sorry missed this yesterday - I don't know of any plans to support non-SPMD use cases like user-mode FSDP, but AFAIK it will require an additional planner implementation.

Since the usage is restricted to SPMD at the moment, I'd say we should keep it in the SPMD doc for now and consider moving it out when the API becomes more stable or we add more use cases, wdyt @JackCaoG?

JackCaoG (approved):

PR #6357, opened 2024-01-22
Provide op lowering for Logit

bhavya01:
For the issue #5989


bhavya01:
For the test failures, I tried running each of the inputs individually and it produces the expected output. 

However, the actual test runs in the inputs in a for loop here https://github.com/pytorch/xla/blob/bc2ebed8dfc63a731c1f3704da0cef0f85f28865/test/test_ops.py#L415. I wonder if it's interfering with how the graph is generated 

@wonjoolee95 Any thoughts?

wonjoolee95:
Hmm, this for loop shouldn't have an effect. What happens if you do some print statements here in the for loop to print out the inputs? Are they same as expected?

bhavya01:
I checked the inputs and the value of the optional variable was changing across iterations. Before this variable wasn't part of the node hash value. The tests pass now with the latest commit.

torch_xla/csrc/elementwise.cpp
```diff
@@ -405,6 +405,23 @@ xla::XlaOp BuildLogSigmoidBackward(xla::XlaOp grad_output, xla::XlaOp input,
...
+  xla::XlaOp zero = xla::Zero(input.builder(), shape.element_type());
+  xla::XlaOp xla_eps =
+      eps.has_value() ? XlaHelpers::ScalarValue<float>(
+                            eps.value(), shape.element_type(), input.builder())
+                      : zero;
+  xla::XlaOp clamped = xla::Clamp(input, xla_eps, one - xla_eps);
+  xla::XlaOp xla_log = xla::Log(clamped / (one - clamped));
+  xla::XlaOp invalid_input = xla::Or(xla::Lt(input, zero), xla::Gt(input, one));
+  xla::XlaOp xla_nan = xla::NanValue(input.builder(), shape.element_type());
+  // // Replace invalid inputs with Nan.
```
wonjoolee95, line 421: nit: extra pair of "//"

wonjoolee95:
Thanks! Glad all the tests are passing now and CI is green. One thing -- since this is a new op, can we add the corresponding CPP op unit test? Example is at https://github.com/pytorch/xla/blob/master/test/cpp/test_aten_xla_tensor_1.cpp.

wonjoolee95:
LGTM! Thanks and congrats on your first PR! Many more to come 😄

wonjoolee95 (approved):

PR #6360, opened 2024-01-23
Define PJRT plugin interface in C++

will-cromar:
- Define base interface `PjRtPlugin` in C++ so the functionality can be called there. In general, implementations should still be in Python, inheriting from `DevicePlugin`.
  - `PyPjRtPlugin` is the [trampoline](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#overriding-virtual-functions-in-python) class that allows python implementations of virtual functions.
- `library_path` and `client_create_options` won't be invoked until client creation time, allowing the user to change settings after importing `torch_xla`.
  - Fixes bug where `TpuPlugin` throws an `EnvironmentError` upon registration.
- It should be possible to merge the two maps of names to PJRT plugins, but I'll need to sort out some issues of lifetimes and ownership first. (e.g. if the Python reference goes out of scope, the remaining C++ object behaves incorrectly)

### DIFF
--- a/torch_xla/csrc/runtime/pjrt_registry.cc
+++ b/torch_xla/csrc/runtime/pjrt_registry.cc
@@ -97,11 +97,10 @@ InitializePjRt(const std::string& device_type) {
           absl::AsciiStrToLower(device_type), plugin->library_path());
       XLA_CHECK_OK(pjrt::InitializePjrtPlugin(device_type));
       auto create_options = plugin->client_create_options();
-      client =
-          xla::GetCApiClient(absl::AsciiStrToUpper(device_type),
-                             {create_options.begin(), create_options.end()},
-                             kv_store)
-              .value();
+      client = xla::GetCApiClient(
+                   absl::AsciiStrToUpper(device_type),
+                   {create_options.begin(), create_options.end()}, kv_store)
+                   .value();
       profiler::RegisterProfilerForPlugin(c_api);
     }
   } else if (device_type == "CPU") {
--- a/torch_xla/csrc/runtime/pjrt_registry.cc
+++ b/torch_xla/csrc/runtime/pjrt_registry.cc
@@ -1,3 +1,5 @@
+#include "torch_xla/csrc/runtime/pjrt_registry.h"
+
 #include "torch_xla/csrc/runtime/debug_macros.h"
 #include "torch_xla/csrc/runtime/env_vars.h"
 #include "torch_xla/csrc/runtime/profiler.h"
@@ -18,13 +20,7 @@ namespace runtime {
 
 namespace {
 
-struct PluginEntry {
-  std::string library_path;
-  absl::flat_hash_map<std::string, xla::PjRtValueType> create_options;
-  bool init_coordinator;
-};
-
-std::unordered_map<std::string, PluginEntry> pjrt_plugins_;
+std::unordered_map<std::string, std::shared_ptr<PjRtPlugin>> pjrt_plugins_;
 
 xla::GpuAllocatorConfig GetGpuAllocatorConfig() {
   auto allocator_config = xla::GpuAllocatorConfig{};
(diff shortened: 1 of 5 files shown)


torch_xla/csrc/runtime/pjrt_registry.cc
```diff
@@ -43,21 +40,18 @@ xla::GpuAllocatorConfig GetGpuAllocatorConfig() {
   return allocator_config;
 }
 
-std::optional<PluginEntry> GetPjRtPlugin(const std::string& device_type) {
+std::shared_ptr<const PjRtPlugin> GetPjRtPlugin(
```
JackCaoG, line 43: any particualr reason for this optional -> shared_ptr change?

torch_xla/experimental/plugins.py
```diff
@@ -84,21 +85,12 @@ def default() -> DevicePlugin:
 
 def register_plugin(name: str, device_plugin: DevicePlugin):
   _plugin_registry[name.upper()] = device_plugin
-  torch_xla._XLAC._register_pjrt_plugin(
-      name, device_plugin.library_path(), device_plugin.client_create_options(),
-      device_plugin.requires_xla_coordinator())
+  torch_xla._XLAC._register_pjrt_plugin(name, device_plugin)
 
 
 def register_installed_plugins():
```
vanbasten23, line 91: pytorch/xla/plugins/cuda/README.md gives an example of how to use `register_plugin`. I wonder under what circumstance do we use `register_installed_plugins()` (I see it's used when we `import torch_xla`. Does it mean if we set XLA_REGISTER_INSTALLED_PLUGINS, then we don't have to use `register_plugin` anymore?)

torch_xla/csrc/runtime/pjrt_registry.cc
```diff
@@ -66,12 +60,12 @@ InitializePjRt(const std::string& device_type) {
...
 
   if (sys_util::GetEnvBool(env::kEnvPjrtDynamicPlugins, false)) {
-    std::optional<PluginEntry> plugin = GetPjRtPlugin(device_type);
+    std::shared_ptr<const PjRtPlugin> plugin = GetPjRtPlugin(device_type);
     if (plugin) {
       TF_VLOG(1) << "Initializing client for PjRt plugin " << device_type;
 
       std::shared_ptr<xla::KeyValueStoreInterface> kv_store = nullptr;
-      if (plugin->init_coordinator) {
+      if (plugin->requires_xla_coordinator()) {
```
vanbasten23, line 68: where can i find the implementation for `requires_xla_coordinator()` for GPU case?

torch_xla/csrc/runtime/pjrt_registry.cc
will-cromar: `PjRtPlugin` has to be a pointer or a reference now that we don't know the concrete type (whereas `PluginEntry` was just a value). My first thought was to make this an optional reference, but apparently [C++ doesn't support that](https://stackoverflow.com/a/26895581). An optional pointer would be unwieldy, because we'd have two layers of indirection/nullability (empty optional, and an optional holding `nullptr`). So I'm just returning `shared_ptr` here and letting `nullptr` represent the empty value.

torch_xla/experimental/plugins.py
will-cromar: I don't expect users to ever call either of these themselves. As long as plugin authors set the entrypoint correctly in their package, registration will happen in the background. Otherwise, plugin authors may add `register_plugin` to their module and run it on import (similar to how `torch.distributed` backend registration works).

torch_xla/csrc/runtime/pjrt_registry.cc
will-cromar: The GPU plugin implementation is in `plugins/cuda` on #6289. In this case, it always returns `True`

jonb377 (approved):
LGTM, I like the C++ to Python interface approach.

JackCaoG (approved):

engineer1109:
Love C++ interface too.

PR #6371, opened 2024-01-24
Add compilation time analysis.

golechwierowicz:


benchmarks/aggregate.py
```diff
@@ -36,6 +36,38 @@
...
+        'metrics'] else []
+    if len(total_times) <= 1:
+      return None
+    compile_and_run_time = total_times[0]
+    run_time = np.average(total_times[1:])
+    std = np.std(total_times[1:])
+    compile_time = compile_and_run_time - run_time
+    # Single sample size for compilation time, std = 0.
+    # Using stddev of the run time distribution.
+    return Datapoint(compile_time, 0)
```
golechwierowicz, line 52: @ecota,

I eventually put 0 in here, but in general I'm not sure what to put here, as n = 1 for compilation estimate.

cota: 0 is the only reasonable value here. stddev of the sample divides by `N-1`, so if N==1 we'd have infinite stddev. Which would be fair enough (we have no clue about the distribution with just one sample), but we're not going to plot infinite error bars :-)

benchmarks/aggregate.py
```diff
@@ -36,6 +36,38 @@
...
+class DatapointSelector:
+
+  def compile(row):
+    total_times = row['metrics']['total_time'] if 'total_time' in row[
+        'metrics'] else []
+    if len(total_times) <= 1:
+      return None
+    compile_and_run_time = total_times[0]
+    run_time = np.average(total_times[1:])
+    std = np.std(total_times[1:])
```
cota, line 48: Let's not compute std then -- AFAICT it's unused.

benchmarks/aggregate.py
```diff
@@ -36,6 +36,38 @@
...
+    compile_time = compile_and_run_time - run_time
+    # Single sample size for compilation time, std = 0.
+    # Using stddev of the run time distribution.
```
cota, line 51: This comment is also obsolete since std is set to 0.

cota:
Other than the small nit above, this LGTM with the caveat that we want to double check that this methodology is sound.

benchmarks/aggregate.py
```diff
@@ -90,21 +122,10 @@ def process_file(args, results_map: Dict[str, Any], filename: str):
...
         continue
-      total_times = r['metrics']['total_time'] if 'total_time' in r[
-          'metrics'] else []
-      # Skip the first three elements in total_times[]; the first one
-      # includes compilation time, the 2nd and 3rd warm up the caches.
-      skip_elems = 3
-      if len(total_times) <= skip_elems:
+
+      dp = getattr(DatapointSelector, args.metric)(r)
+      if dp is None:
```
vanbasten23, line 127: is dp used if it's not None?

benchmarks/aggregate.py
golechwierowicz: Thanks, my reasoning was that if 100% of the data samples (n=1) have the same value then std dev has to be 0, although I felt there some holes in it.

benchmarks/aggregate.py
golechwierowicz: It's used on line 140.

benchmarks/aggregate.py
golechwierowicz: Done.

benchmarks/aggregate.py
golechwierowicz: Done.

cota (approved):

PR #6373, opened 2024-01-24
Add GPU plugin build to CI

will-cromar:
- Install GPU plugin in CI. With the cache warmed, this doesn't affect the build time much. Cold-start builds are _much_ slower -- over an hour and 40 minutes.
  - We'll only ever have to rebuild the plugin during XLA pin updates and potentially when we update XLA patches. Otherwise, this should be entirely cached, since all of the code comes from XLA.
  - The cold-start build times will get better when we can drop the GPU client from the main build. Then, the CI can build the CPU-only `torch_xla` package, which is much smaller.
- `build_utils` symlinks didn't actually do anything, so remove them. Adjust the Python search path instead.
- Remove build isolation too. This only worked locally because of my specific environment.
- Add plugin build to ansible configs. Tested this locally. I think this will get uploaded to our GCS bucket automatically. Sanity-checking this with the TPU CI (which uses ansible).

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -104,7 +104,7 @@ function build_torch_xla() {
   XLA_DIR=$1
   pushd "$XLA_DIR"
   python setup.py install
-  pip install plugins/cuda -v --no-build-isolation
+  pip install plugins/cuda -v
   popd
 }
 
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -104,7 +104,7 @@ function build_torch_xla() {
   XLA_DIR=$1
   pushd "$XLA_DIR"
   python setup.py install
-  pip install plugins/cuda -v --no-build-isolation
+  PYTHONPATH=. pip install plugins/cuda -v --no-build-isolation
   popd
 }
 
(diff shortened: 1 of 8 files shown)


plugins/cpu/setup.py
```diff
@@ -1,3 +1,8 @@
+# add `build_util` to import path
+import os
+import sys
+sys.path.append(os.path.join(os.path.dirname(__file__), '..', '..'))
```
JackCaoG, line 4: what's this line for?

.circleci/common.sh
```diff
@@ -104,6 +104,7 @@ function build_torch_xla() {
   XLA_DIR=$1
   pushd "$XLA_DIR"
   python setup.py install
+  pip install plugins/cuda -v
```
JackCaoG, line 107: will this command both build and install the cuda plugin?

plugins/cpu/setup.py
will-cromar: See the comment and PR description. `setuptools` doesn't include the local project directory in the `PYTHONPATH` by default when using `pip install`/`pip wheel`. Slightly re-ordered lines.

.circleci/common.sh
will-cromar: Yes

JackCaoG (approved):

Issue #6286, opened 2024-01-10
Torchbench Benchmark Running ERROR


zpcore:
Tested with the lastest commit (235b82b5bd5ce6e3fee2f4143675c435613f1046) on TPU V5 and notice the following error:
```
"/usr/local/lib/python3.10/dist-packages/transformers/utils/generic.py:441: UserWarning: torch.utils._pytree._register_pytree_node is deprecated. Please use torch.utils._pytree.register_pytree_node instead.
  _torch_pytree._register_pytree_node(
WARNING:__main__:Enabling fast F
32 multiplication for PyTorch
Traceback (most recent call last):
  File \"/home/piz/xla/benchmarks/experiment_runner.py\", line 878, in <module>
    main()
  File \"/home/piz/xla/benchmarks/experiment_runner.py\", line 874, in main
    runner.run()
  File \"/home/piz/xla/benc
hmarks/experiment_runner.py\", line 58, in run
    self.run_single_config()
  File \"/home/piz/xla/benchmarks/experiment_runner.py\", line 237, in run_single_config
    reset_rng_state(benchmark_experiment)
  File \"/home/piz/xla/benchmarks/util.py\", line 53, in reset_rng_stat
e
    device = benchmark_experiment.get_device()
  File \"/home/piz/xla/benchmarks/benchmark_experiment.py\", line 163, in get_device
    return xm.xla_device(devkind=self.accelerator.upper())
  File \"/home/piz/.local/lib/python3.10/site-packages/torch_xla/core/xla_model.py\",
 line 207, in xla_device
    return runtime.xla_device(n, devkind)
  File \"/home/piz/.local/lib/python3.10/site-packages/torch_xla/runtime.py\", line 88, in wrapper
    return fn(*args, **kwargs)
  File \"/home/piz/.local/lib/python3.10/site-packages/torch_xla/runtime.py\", li
ne 124, in xla_device
    return torch.device(torch_xla._XLAC._xla_get_default_device())
RuntimeError: Bad StatusOr access: UNKNOWN: TPU initialization failed: open(/dev/vfio/0): Device or resource busy: Device or resource busy; Couldn't open iommu group /dev/vfio/0
WARNING: All
 log messages before absl::InitializeLog() is called are written to STDERR
[...]

PR #6375, opened 2024-01-24
Use `benchmark_cls` for checking precision.`

ysiraichi:
This PR makes it so we don't have to call `load_benchmark` only for checking the precision to be used.

cc @miladm @JackCaoG

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -1,3 +1,4 @@
+import functools
 import gc
 import importlib
 import logging
@@ -244,16 +245,18 @@ class TorchBenchModel(BenchmarkModel):
     del benchmark
     self._cleanup()
 
-  def load_benchmark(self):
+  @functools.lru_cache(maxsize=1)
+  def benchmark_cls(self):
     try:
       module = importlib.import_module(
           f"torchbenchmark.models.{self.model_name}")
     except ModuleNotFoundError:
       module = importlib.import_module(
           f"torchbenchmark.models.fb.{self.model_name}")
-    benchmark_cls = getattr(module, "Model", None)
+    return getattr(module, "Model", None)
 
-    cant_change_batch_size = (not getattr(benchmark_cls,
+  def load_benchmark(self):
+    cant_change_batch_size = (not getattr(self.benchmark_cls(),
                                           "ALLOW_CUSTOMIZE_BSIZE", True))
     if cant_change_batch_size:
       self.benchmark_experiment.batch_size = None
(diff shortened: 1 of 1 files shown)


benchmarks/torchbench_model.py
```diff
@@ -244,16 +245,18 @@ def set_up(self):
     del benchmark
     self._cleanup()
 
-  def load_benchmark(self):
+  @functools.lru_cache(maxsize=1)
```
zpcore, line 248: I am afraid this doesn't solve the issue. `lru_cache` may not work across processes. We will still call the function later as a subprocess.  Have you tried this in TPU device?

zpcore:
Refer to the issue here for the context: https://github.com/pytorch/xla/issues/6286
Thanks for making the fix. 

The key point I think is to prevent leaving behind a dangling object which e.g., moved a model to xla device. `del benchmark` doesn't resolve the issue because it has already claimed the PJRT runtime. This will trigger the stackdump error: `RuntimeError: Bad StatusOr access: UNKNOWN: TPU initialization failed: open(/dev/vfio/0): Device or resource busy: Device or resource busy; Couldn't open iommu group /dev/vfio/0`.

zpcore:
Since we only need to detect the precision, we can fetch the information directly without invoking
```
benchmark_cls(
        test=self.benchmark_experiment.test,
        device=device,
        batch_size=self.benchmark_experiment.batch_size,
    )
```
 I think we can call the following `load_benchmark_precision` instead of `load_benchmark` to get the precision directly.
 ```
   def load_benchmark_precision(self):
    try:
      module = importlib.import_module(
          f"torchbenchmark.models.{self.model_name}")
    except ModuleNotFoundError:
      module = importlib.import_module(
          f"torchbenchmark.models.fb.{self.model_name}")
    benchmark_train_precision = getattr(module.Model, "DEFAULT_TRAIN_CUDA_PRECISION", None)
    benchmark_eval_precision = getattr(module.Model, "DEFAULT_EVAL_CUDA_PRECISION", None)
    return benchmark_train_precision, benchmark_eval_precision
 ```
 WDYT?

ysiraichi:
Right. Correct if I'm misunderstanding things, but isn't that exactly what I'm doing here?

zpcore:
> Right. Correct if I'm misunderstanding things, but isn't that exactly what I'm doing here?

Hah, you are right. I didn't notice that you called `benchmark_cls` instead. 

Now it LGTM!

zpcore (approved):

PR #6383, opened 2024-01-25
Correctly set the exit code when an exception is raised in the atexit callback _prepare_to_exit.

vanbasten23:
This PR is intended to correctly set the exit code when an exception is raised in the atexit callback `_prepare_to_exit`. Consequently, the CI can catch such issues whenever an exception happens in the callback, as opposed to silently failing and escaping the CI. See https://github.com/pytorch/xla/issues/6385.

Currently, when an exception is raised in the call back, the exit code is still 0 (success), which makes the CI unable to catch issues under the circumstance. The incorrect exit code is due to https://bugs.python.org/issue27035: python 2 can set the correct exit code (expect to be non-zero) if an exception is raised in the atexit callback but python 3 cannot. But the bug seems still to be open.

In the callback, neither `sys.exit(1)` or `raise Exception` is able to set the exit code correctly so a os._exit(1) is used. It may not be ideal but it's better for us to catch the mishap. Also it should be ok because it's in the less frequently executed path.

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -148,7 +148,6 @@ _aws_ec2_inf_trn_init()
 
 def _prepare_to_exit():
   try:
-    raise ValueError('fake error')
     _XLAC._prepare_to_exit()
     if int(os.environ.get('PT_XLA_DEBUG', '0')):
       _summarize_fn_tracker()
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -147,11 +147,15 @@ _aws_ec2_inf_trn_init()
 
 
 def _prepare_to_exit():
-  device = _XLAC._xla_get_default_device()
-  _XLAC._set_all_reduce_token(device, None)
-  _XLAC._prepare_to_exit()
-  if int(os.environ.get('PT_XLA_DEBUG', '0')):
-    _summarize_fn_tracker()
+  try:
+    device = _XLAC._xla_get_default_device()
+    _XLAC._set_all_reduce_token(device, None)
+    _XLAC._prepare_to_exit()
+    if int(os.environ.get('PT_XLA_DEBUG', '0')):
+      _summarize_fn_tracker()
+  except Exception as e:
+    print('Caught an exception in the atexit callback: ', e)
+    os._exit(1)
 
 
 def _init_xla_lazy_backend():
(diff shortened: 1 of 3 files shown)


vanbasten23:
Ok, by this far, I can repro the failure. I can see the error msg `Check failed: dist_runtime_client_->Connect().ok() ` even though the GPU CI succeeded.

torch_xla/__init__.py
```diff
@@ -147,9 +147,13 @@ def _setup_tpu_vm_library_path() -> bool:
...
 def _prepare_to_exit():
-  _XLAC._prepare_to_exit()
-  if int(os.environ.get('PT_XLA_DEBUG', '0')):
-    _summarize_fn_tracker()
+  try:
+    _XLAC._prepare_to_exit()
+    if int(os.environ.get('PT_XLA_DEBUG', '0')):
+      _summarize_fn_tracker()
+  except Exception as e:
+    print('Caught an exception in the atexit callback: ', e, flush=True)
```
will-cromar, line 155: nit: can you make this `logging.error("...", exc_info=e)`?

will-cromar (approved):

yeounoh (approved):
LGTM thank you!

torch_xla/__init__.py
vanbasten23: updated.

torch_xla/__init__.py
```diff
@@ -147,9 +147,14 @@ def _setup_tpu_vm_library_path() -> bool:
...
-  if int(os.environ.get('PT_XLA_DEBUG', '0')):
-    _summarize_fn_tracker()
+  try:
+    _XLAC._prepare_to_exit()
+    if int(os.environ.get('PT_XLA_DEBUG', '0')):
+      _summarize_fn_tracker()
+  except Exception as e:
+    logging.error(
+        "Caught an exception when exiting the process. Exception: ", exc_info=e)
+    os._exit(1)
```
will-cromar, line 157: sorry, one more micro-nit: make this `exit(1)`

vanbasten23: I tried and it didn't set the exit code correctly. Per https://docs.python.org/3/library/constants.html#exit, calling `exit(1)` raises a `SystemExit` exception. But due to https://bugs.python.org/issue27035, raising an exception won't set the exit code correctly.

Also, per https://docs.python.org/3/library/constants.html#exit, `exit` or `quit` "are useful for the interactive interpreter shell and should not be used in programs.".

I'll add a comment.

PR #6384, opened 2024-01-25
Update DDP example in README.md

bhavya01:
xmp.spawn doesn't expect the world size as a function parameter.


will-cromar (approved):
Thanks!

PR #6386, opened 2024-01-26
[FSDPv2] Add a user guide

alanwaketan:
Summary:
This diff adds a user guide for FSDPv2.

### DIFF
--- a/docs/fsdpv2.md
+++ b/docs/fsdpv2.md
@@ -1,0 +1,58 @@
+# Fully Sharded Data Parallel via SPMD
+
+Fully Sharded Data Parallel via SPMD or FSDPv2 is an utility that re-epxresses the famous FSDP algorithm in SPMD. [This](https://github.com/pytorch/xla/blob/master/torch_xla/experimental/spmd_fully_sharded_data_parallel.py) is
+an experimental feature that aiming to offer a familiar interface for users to enjoy all the benefits that SPMD brings into
+the table. The design doc is [here](https://github.com/pytorch/xla/issues/6379).
+
+Example usage:
+```python3
+import torch
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.spmd as xs
+from torch_xla.experimental.spmd_fully_sharded_data_parallel import SpmdFullyShardedDataParallel as FSDPv2
+
+# Define the mesh following common SPMD practice
+num_devices = xr.global_runtime_device_count()
+mesh_shape = (num_devices, 1)
+device_ids = np.array(range(num_devices))
+# To be noted, the mesh must have an axis named 'fsdp', which the weights and activations will be sharded on.
+mesh = Mesh(device_ids, mesh_shape, ('fsdp', 'model'))
+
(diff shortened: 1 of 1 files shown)


JackCaoG:
Can you also update https://github.com/pytorch/xla/blob/master/docs/source/index.rst so it will show up in https://pytorch.org/xla/release/2.1/index.html?

docs/fsdpv2.md
```diff
@@ -0,0 +1,58 @@
...
+mesh = Mesh(device_ids, mesh_shape, ('fsdp', 'model'))
+
+# As normal FSDP, but an extra mesh is needed.
+model = FSDPv2(my_module, mesh)
+optim = torch.optim.Adam(model.parameters(), lr=0.0001)
+output = model(x, y)
+loss = output.sum()
+loss.backward()
+optim.step()
+```
```
JackCaoG, line 28: should we add a note that don't use `xmp.spawn` for this somewhere in this doc?

alanwaketan: Maybe I just link to the spmd doc? And ask the users to go over it first?

JackCaoG: depending on your target user. If you are trying to convert those who uses FSDP today to use FSDPv2 I think you want to be more explict about how to migrate their existing code to FSDP v2

docs/fsdpv2.md
```diff
@@ -0,0 +1,58 @@
...
+num_devices = xr.global_runtime_device_count()
+mesh_shape = (num_devices, 1)
+device_ids = np.array(range(num_devices))
+# To be noted, the mesh must have an axis named 'fsdp', which the weights and activations will be sharded on.
+mesh = Mesh(device_ids, mesh_shape, ('fsdp', 'model'))
+
+# As normal FSDP, but an extra mesh is needed.
+model = FSDPv2(my_module, mesh)
+optim = torch.optim.Adam(model.parameters(), lr=0.0001)
+output = model(x, y)
```
jonb377, line 24: We'll need to add input sharding here - do you want to add an example with MpDeviceLoader?

docs/fsdpv2.md
alanwaketan: Let's target people that is willing to try SPMD first. Once it's mature, we can then target people that is currently using FSDP.

docs/fsdpv2.md
alanwaketan: Let's manually shard the x here. Good catch!

docs/fsdpv2.md
JackCaoG: sure, it is still good to mention something along the line of "This is using SPMD so you don't need to spawn multiple process per host, for more detail please check this doc"

alanwaketan: Added one line on top:
Please review the [SPMD user guide](./spmd.md) before proceeding.

JackCaoG (approved):

alanwaketan:
Thanks, @JackCaoG!

PR #6392, opened 2024-01-26
aggregate.py: allow picking backends to report on

cota:
With this we also allow picking backends that we ignored until now, namely eager and openxla{,_eval}+lazytensor.

While at it, shorten the printed names of "PytorchXLA" to "XLA" and "WorkloadNumber" to "Workload". This paves the way for an upcoming "tabulate" mode.

### DIFF
--- a/benchmarks/aggregate.py
+++ b/benchmarks/aggregate.py
@@ -26,6 +26,14 @@ logger = logging.getLogger(__name__)
 
 Datapoint = namedtuple('Datapoint', 'avg, std')
 
+_title_map = {
+    'eager': 'Eager',
+    'inductor': 'Inductor',
+    'openxla_eval+dynamo': 'XLA_Eval+Dynamo',
+    'openxla+dynamo': 'XLA+Dynamo',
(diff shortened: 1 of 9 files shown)


frgossen (approved):
Thank you!

cota:
I just removed openxla_eval+lazytensor, which makes no sense (openxla_eval only works with dynamo).
I've also added sanity checks for the passed `--backend` flags.

benchmarks/aggregate.py
```diff
@@ -56,9 +64,26 @@ def skip_model(args, model_name: str):
           re.search("|".join(args.exclude), model_name, re.I))
 
 
+def get_backend_name(dynamo: str, xla: str) -> str:
+  if dynamo == 'inductor':
+    return dynamo
```
golechwierowicz, line 69: nitty nit: return 'inductor', so it's more explicit and you do not have to look at the line above.

golechwierowicz (approved):

benchmarks/aggregate.py
cota: Done, thanks

PR #6402, opened 2024-01-29
[torchbench] Re-land: Force `llama_v2_7b_16h` to use SGD.

ysiraichi:
Relanding #6350. It was reverted by mistake at #6352.

cc @miladm @JackCaoG @zpcore


zpcore (approved):
Thanks for going through all those models and resolve the issue!

PR #6405, opened 2024-01-29
aggregate.py: add tab format

cota:
Note that empty strings ('') and numbers should not be on the same column, for tabulate interprets the whole column as a string, thereby ignoring the floatfmt option.

Here we add tests to make sure that None instead of '' is used with tabulate when we want some empty fields. We also test the inverse when emitting CSV.


benchmarks/aggregate.py
```diff
@@ -277,6 +278,36 @@ def get_pr_titles(args):
...
+    for row in rows:
+      print(','.join([str(f) if f is not None else '' for f in row]))
+  elif args.format == 'tab':
+    print(
+        tabulate(rows, headers=headers, tablefmt='fancy_grid', floatfmt='.2f'))
+
+
 def pr_latest(results_map: Dict[str, Any], args, timestamps: List[str]):
   titles, data_labels = get_pr_titles(args)
   speedups = [[] for _ in titles]
```
golechwierowicz, line 313: Can we add short comment about the structure of speedups array? Indices semantics are unclear at a first glance.

benchmarks/aggregate.py
```diff
@@ -297,11 +328,13 @@ def pr_latest(results_map: Dict[str, Any], args, timestamps: List[str]):
...
-    ]))
+  if args.format == 'csv' or args.format == 'tab':
+    headers = ['Workload'] + [
+        header for title in titles for header in [
+            speedup_header(title, base_backend_name, args), 'StdDev',
+            modelname_header(title, args)
+        ]
+    ]
     # Note: the latest timestamp might not have results for all benchmarks.
     max_len = max([len(l) for l in speedups])
```
golechwierowicz, line 339: `s/max_len/row_size`

?

golechwierowicz (approved):
Thanks!

benchmarks/aggregate.py
cota: Done, thanks.

benchmarks/aggregate.py
cota: Ended up using `num_rows`, otherwise `row_size` might be confused with `num_cols` :-P

cota: Also added a comment explaining why we pad the lists.

golechwierowicz (approved):

PR #6406, opened 2024-01-29
nightly.sh: pass --always to git describe --tags

cota:
For some commits we cannot get a tag via `git describe --tags`:

```
$ git -C pytorch/xla describe --tags
fatal: No tags can describe '2f4275f5...'.
Try --always, or create some tags.
```

With --always, we can get the commit as a fallback.


golechwierowicz (approved):
Thanks! Does it make sense to always use a commit hash then?

cota:
> Thanks! Does it make sense to always use a commit hash then?

With the current solution we get the tag + commit ([example; look at the "XLA" subtitle](https://pantheon.corp.google.com/storage/browser/_details/torchbench-pytorch-xla-benchmark-results/nightly_results/reports/2024-01-11/a100-inference-latest-tier1.svg)), and the commit as a fallback. I think this is the best we can do.

PR #6409, opened 2024-01-29
Avoid fallback for avg_pool -

qihqi:
By supporting divisor overrides and ceil_mode + count_include_pad property.

When count_include_pad is True, the number of paddings also appear in the denominator. But, if ceil_mode is also true, then, when we round, we can introduce extra padding. ** these paddings are NOT counted in the denominator** Therefore, when ceil_mode is true, we need to manually pad to distinguish padding that should count for denominator and those that shouldnt (i.e. those introduced by ceil_mode)

### DIFF
--- a/test/cpp/test_tensor.cpp
+++ b/test/cpp/test_tensor.cpp
@@ -316,7 +316,8 @@ TEST_F(TensorTest, TestAvgPool2D) {
                            /*kernel_size=*/{kernel_size, kernel_size},
                            /*stride=*/{stride, stride},
                            /*padding=*/{padding, padding},
-                           /*ceil_mode=*/false, count_include_pad);
+                           /*ceil_mode=*/false, count_include_pad,
+                           /*divisor_override=*/std::nullopt);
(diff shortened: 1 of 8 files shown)


wonjoolee95:
Changes LGTM, but seems like CI failed:
```
FAIL: test_aten_avg_pool2d_3 (__main__.AtenOpTest) [torch_xla_diff:0.001]
  File "/tmp/pytorch/xla/test/test_core_aten_ops.py", line 72, in run_export_and_compare
    diff_output(
  File "/tmp/pytorch/xla/test/test_core_aten_ops.py", line 33, in diff_output
    testcase.assertTrue(
AssertionError: False is not true
```
Maybe just need to adjust rtol/atol?

wonjoolee95 (approved):

wonjoolee95:
@qihqi, thanks! Could you link the issues that this PR fixes, if there are any?

qihqi:
It looks like from Jan 31 -> Feb 1 we have improved performance of some vision models, this might has helped (as it avoids CPU fallback):

image

PR #6410, opened 2024-01-29
nightly.sh: add flag to set Hugging Face token

### LINKED ISSUE #6401: [torchbench] HF API Token for benchmark execution.
## 🐛 Bug

Some benchmarks are not working (internally) because PyTorch/benchmark expects `HUGGING_FACE_HUB_TOKEN` environment variable (HF API token) to be set:

```python
NotImplementedError: Make sure to set `HUGGING_FACE_HUB_TOKEN` so you can download weights
```

## Affected Benchmarks

- stable_diffusion_text_encoder
- stable_diffusion_unet
- llama_v2_7b_16h

## Environment

 - Reproducible on XLA backend [CPU/TPU]: CUDA
 - torch_xla version: 9e4db963a196378e7e812bcb58f8f1af6a8f452f

cc @miladm @JackCaoG @cota @frgossen

cota:
Fixes #6401.

### DIFF
--- a/benchmarks/nightly.sh
+++ b/benchmarks/nightly.sh
@@ -9,7 +9,7 @@ WORKSPACE=$(date --utc +%Y-%m-%d)
 REPEAT=8
 ENABLE_PROFILING=
 
-while getopts 'A:O:PR:W:' OPTION
+while getopts 'A:O:PR:T:W:' OPTION
 do
   case ${OPTION?} in
     A)
@@ -23,6 +23,12 @@ do
       ;;
     R)
       REPEAT=${OPTARG:?}
+      ;;
+    T)
+      # Avoid printing the token; re-enable printing later.
+      { set +x; } 2>/dev/null
+      export HUGGING_FACE_HUB_TOKEN=${OPTARG:?}
+      set -x
       ;;
     W)
       WORKSPACE=${OPTARG:?}


benchmarks/nightly.sh
```diff
@@ -24,6 +24,12 @@ do
     R)
       REPEAT=${OPTARG:?}
       ;;
+    T)
+      # Avoid printing the token; re-enable printing later.
+      { set +x; } 2>/dev/null
```
frgossen, line 29: Does it really print the token otherwise?

frgossen (approved):
Nice! Ty!

benchmarks/nightly.sh
cota: It really does :)
With `set -x` every bash command gets printed out -- `FOO=BAR` (i.e. bash variable assignment) or `export FOO=BAR` both get printed out.
For good measure I'm also stripping out the token from the logs/json files in case it ever leaks out of this script or through the Torchbench logs.

frgossen: Thanks for checking

golechwierowicz (approved):
So this is supposed to make llama v2 run only, or there are other models benefiting from it? If I look at the [PyTorch HUD](https://hud.pytorch.org/benchmark/torchbench/inductor_no_cudagraphs?startTime=Tue,%2023%20Jan%202024%2009:04:31%20GMT&stopTime=Tue,%2030%20Jan%202024%2009:04:31%20GMT&granularity=hour&mode=inference&dtype=bfloat16&lBranch=main&lCommit=47b5a6b05d00e91182ee3c4a62f9b94e2d21b4eb&rBranch=main&rCommit=bb28965924cd142a9d12f475fa646c45acb8639f) it's not there.

ysiraichi:
stable_diffusion benchmarks also depend on this token.

PR #6412, opened 2024-01-30
Move torch import closer to the place it uses

alanwaketan:
Summary:
Move torch import closer to the place it uses such that we don't have to have torch installed to do `python setup.py clean`.

Test Plan:
Tested locally.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,6 @@ import tempfile
 import zipfile
 
 import build_util
-import torch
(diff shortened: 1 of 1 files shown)


will-cromar (approved):

JackCaoG (approved):

alanwaketan:
Thanks Jack and Will!

PR #6416, opened 2024-01-30
Use PyTorch's dynamo benchmark skip-list.

ysiraichi:
This PR starts using PyTorch's dynamo benchmark skip-list for skipping specific models, based on the experiments' configuration. Besides that, it also:

- Refactors `is_compatible` function
- Refactors the lookup of a given file, previously in `add_torchbench_dir` function
- Replaces the comments on each skipped benchmark by its actual error message
    - I'd say it's more informative, but will revert it if anyone thinks otherwise 

cc @miladm

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -171,7 +171,8 @@ class TorchBenchModelLoader(ModelLoader):
   def add_torchbench_dir(self):
     os.environ["KALDI_ROOT"] = "/tmp"  # avoids some spam
 
-    torchbench_dir = self._find_near_file(("torchbenchmark", "torchbench", "benchmark"))
+    torchbench_dir = self._find_near_file(
+        ("torchbenchmark", "torchbench", "benchmark"))
     assert torchbench_dir is not None, "Torch Benchmark folder not found."
 
     if torchbench_dir is not None:
@@ -192,7 +193,8 @@ class TorchBenchModelLoader(ModelLoader):
     benchmarks_dir = self._find_near_file(("benchmarks",))
     assert benchmarks_dir is not None, "PyTorch benchmarks folder not found."
 
-    skip_file = os.path.join(benchmarks_dir, "dynamo", "torchbench_skip_models.yaml")
+    skip_file = os.path.join(benchmarks_dir, "dynamo",
+                             "torchbench_skip_models.yaml")
     with open(skip_file) as f:
       data = yaml.safe_load(f)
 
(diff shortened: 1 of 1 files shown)


benchmarks/torchbench_model.py
```diff
@@ -64,59 +65,51 @@
...
         "test": "train"
-    },],  # not implemented
+    },],  # Model's DEFAULT_TRAIN_BSIZE is not implemented
     "pyhpc_turbulent_kinetic_energy": [{
         "test": "train"
-    },],  # not implemented
-    "pytorch_struct": [{
-        "test": "eval"
-    },],  # not implemented
+    },],  # Model's DEFAULT_TRAIN_BSIZE is not implemented
```
zpcore, line 112: Thanks for the refining. Does those models removed from the list work fine now?

ysiraichi: Not really. But, as far as I can see, they are not in torchbench list of models anymore

zpcore (approved):

benchmarks/torchbench_model.py
```diff
@@ -218,16 +230,29 @@ def is_compatible(self,
...
+    if name in self.skip["test"].get(benchmark_experiment.test, {}):
+      return False
+
+    if name in self.skip["device"].get(benchmark_experiment.accelerator, {}):
+      return False
+
+    if name in self.skip["multiprocess"]:
+      # No support for multiprocess, yet. So, skip all benchmarks that
+      # only work with it.
+      return False
```
miladm, line 248: I wonder if we should include an "INFO" message when skipping models so it's clear on the logs - unless it gets reported elsewhere. wdyt @zpcore?

ysiraichi: The scripts already report when a model is skipped. However, the reason is not logged. Do you think we should log that, too?

benchmarks/torchbench_model.py
```diff
@@ -169,32 +154,59 @@ def __init__(self, args):
...
     return torchbench_dir
 
+  def get_skip_data(self):
+    """Retrieve the skip data in the PyTorch YAML file.
+
+    Reads the YAML file in PyTorch's dynamo benchmarks directory, and transform
+    its lists of models into sets of models.
+    """
+
+    benchmarks_dir = self._find_near_file(("benchmarks",))
```
zpcore, line 193: It turns out pytorch/benchmarks and xla/benchmarks share the same `benchmarks` name. And when we search for yaml file, the nearest will be the xla/benchmarks.

PR #6417, opened 2024-01-30
Gradient bucketing using a pre-defined bucket size cap

amithrm:


alanwaketan:
Do you mind adding a test case?

amithrm:
Added the test case and rebased @JackCaoG @alanwaketan

torch_xla/core/xla_model.py
```diff
@@ -974,6 +976,56 @@ def wait_device_ops(devices=[]):
...
 
+def bucketed_allreduce(gradients):
+    total = 0
+    tensor_bucket = []
+
+    for grad in gradients:
+      grad_bytes = grad.numel() * grad.element_size()
+
+      # Gradient is larger than bucket_cap, don't bucketize
+      if grad_bytes > bucket_cap:
```
alanwaketan, line 987: Curious why you want to specialize this case?

torch_xla/core/xla_model.py
```diff
@@ -990,14 +1042,13 @@ def reduce_gradients(optimizer, groups=None, pin_layout=True):
   """
   count = xrt_world_size()
   if count > 1:
-    gradients = _fetch_gradients(optimizer)
```
alanwaketan, line 993: Can we keep the original behavior? And maybe use a flag to turn this feature on?

amithrm: OK..let me work on that

torch_xla/core/xla_model.py
amithrm: if the grad_bytes (already in the tensor) is larger than bucket cap, we send it straight away as a single tensor instead of bucketing.

alanwaketan: Right, I understood the logic. But why? Combining it with the bucket introduce some problems?

torch_xla/core/xla_model.py
jeffhataws: Maybe we should introduce an argument "bucket_cap_mb" that turns this on, instead of environmental variable?  bucket_cap_mb=0 turns off bucketing and is the default?

torch_xla/core/xla_model.py
jeffhataws: Yeah, looks like you can get rid of this if statement (until continue), and the "if total > bucket_cap" should take care of this condition when total=0.

torch_xla/core/xla_model.py
```diff
@@ -974,6 +976,56 @@ def wait_device_ops(devices=[]):
...
+            REDUCE_SUM, [grad],
+            scale=1.0 / count,
+            groups=groups,
+            pin_layout=pin_layout)
+        continue
+
+      # Bucketize till the total spills over
+      total += grad_bytes
+      if total > bucket_cap:
+        all_reduce(
```
jeffhataws, line 1009: Need to check "if len(tensor_bucket):" because tensor_bucket can be empty at the start, when grad_bytes > bucket_cap.

torch_xla/core/xla_model.py
```diff
@@ -974,6 +976,56 @@ def wait_device_ops(devices=[]):
   torch_xla._XLAC._xla_wait_device_ops(devices=devices)
 
 
+def bucketed_allreduce(gradients):
```
jeffhataws, line 979: Maybe name it similar to the original function all_reduce? How about all_reduce_bucketized?

Also, do you need to pass "groups" and "pin_layout" also?

torch_xla/core/xla_model.py
amithrm: The issue with combining this with the rest is that the "buffer" allocated in the underlying runtime may not have enough space to fit this large tensor. The idea is to have a large buffer that can fit all the tensors. It can happen that total_bytes is just below the max allowed and this tensor if added to the bucket spills the maximum. Hence should go "alone" without bucketizing

amithrm: See your concerns now !! Fixed the code flow

torch_xla/core/xla_model.py
amithrm: done

alanwaketan (approved):
LGTM. Please address other comments as well.

jeffhataws:
@JackCaoG do you know why the build failed with "ERROR: Error initializing RemoteModule"?

JackCaoG:
It is on a fork hence can't use remote cache. I think we fixed this issue. If you rebase the CI should start running.

jeffhataws:
@JackCaoG looks like build is still failing for some reason after rebasing. Maybe another rebase is needed?

JackCaoG:
The error still seems to be related with the fork. Let me grant both of you the write access, then you can open pr directly.

JackCaoG:
OK I gave @amithrm write access

jeffhataws:
Replaced by https://github.com/pytorch/xla/pull/7216 to avoid the build issues in CI testing.

PR #6420, opened 2024-01-30
Force install `google-cloud-sdk-gke-gcloud-auth-plugin` to enable TPUCI run_e2e_tests

ManfeiBai:
since Jan30, TPU CI start to failed due to there is the existing `google-cloud-cli-gke-gcloud-auth-plugin` before install `google-cloud-sdk-gke-gcloud-auth-plugin`, 

so to bring TPU-CI/CI green, confirm force install `google-cloud-sdk-gke-gcloud-auth-plugin` by `--force-overwrite`

### DIFF
--- a/infra/terraform_modules/xla_docker_build/xla_docker_build.tf
+++ b/infra/terraform_modules/xla_docker_build/xla_docker_build.tf
@@ -131,7 +131,8 @@ locals {
 
           # Try to setup kubectl credentials the new way,
           # see https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
-          apt-get install google-cloud-sdk-gke-gcloud-auth-plugin -o DPkg::options::="--force-overwrite" -y
+          apt-get uninstall google-cloud-cli-gke-gcloud-auth-plugin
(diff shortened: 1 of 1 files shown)


will-cromar (approved):

JackCaoG:
Looking at the log it still does not have `-y`, I am guessing this pr will not take effect until it is merged. @will-cromar should we just merge it?

will-cromar:
Correct, this change will only have an effect after it is merged.

Since the cause of our issue here is that `google-cloud-sdk-gke-gcloud-auth-plugin` is already installed (when it apparently was not before), should we just remove this `apt-get install` entirely?

ManfeiBai:
> Correct, this change will only have an effect after it is merged.
> 
> Since the cause of our issue here is that `google-cloud-sdk-gke-gcloud-auth-plugin` is already installed (when it apparently was not before), should we just remove this `apt-get install` entirely?

thanks, the cause was `google-cloud-cli-gke-gcloud-auth-plugin` installed already, and we want to install a different plugin `google-cloud-sdk-gke-gcloud-auth-plugin`, so remove the old plugin and install the new plugin here

PR #6421, opened 2024-01-30
Fix distributed checkpoint errors when PyTorch CUDA is enabled.

ysiraichi:
This PR fixes the errors found in #6070 CI. Specifically, the tests on `test/spmd/test_xla_distributed_checkpoint.py` fail due to an assertion `tensor.is_cpu`.

**Problem:** PyTorch filesystem writers don't support XLA
- If PyTorch is compiled with CUDA support, we try to move tensors asynchronously to CPU
- However, we only do so for tensors whose device is `cuda`
- Since XLA tensors' device are not `cuda`, those are not moved to CPU (which is why the assertion breaks)

**Solution:** don't allow asynchronous copy
- AFAIU, XLA doesn't support it (XLA to CPU asynchronous copy)
- Compiling PyTorch without CUDA support (previous default) fallbacked to this behavior

@miladm @JackCaoG

### DIFF
--- a/test/spmd/test_xla_distributed_checkpoint.py
+++ b/test/spmd/test_xla_distributed_checkpoint.py
@@ -93,7 +93,10 @@ class EndToEndCheckpointTest(DistributedCheckpointTestBase):
     model_out_state_dict = model_out.state_dict()
     dist_cp.save_state_dict(
         state_dict=model_in_state_dict,
-        storage_writer=dist_cp.FileSystemWriter(chkpt_path),
+        storage_writer=dist_cp.FileSystemWriter(
+            chkpt_path,
+            per_thread_copy_ahead=0,
+        ),
         planner=save_planner,
         no_dist=no_dist,
     )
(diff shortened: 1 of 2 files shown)


jonb377:
Interesting find, thanks for the fix! For my understanding, is the issue on TPU tests when CUDA is enabled in the build?

ysiraichi:
No. But probably, it would also error for TPU. The setup I tested is the following:

- Compile PyTorch with CUDA support (i.e. `USE_CUDA=1`)
- Compile PyTorch/XLA with CUDA support (i.e. `XLA_CUDA=1`)
- Run the test: `PJRT_DEVICE=CUDA python test/spmd/test_xla_distributed_checkpoint`
    - Errors due to `AssertionError: assert tensor.is_cpu`

This PR fixes that.

ysiraichi:
@jonb377 Could you approve this PR, if it looks good for you?

jonb377 (approved):
LGTM, thanks for fixing this!

PR #6428, opened 2024-01-31
Fix model skip yaml file not found issue

zpcore:
1. Also search for path `../../../` in case user place xla under pytorch folder.
2. Both xla and pytorch have a subfolder `benchmarks`.  Use `pytorch/benchmarks` instead to prevent searching under xla.

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -164,7 +164,7 @@ class TorchBenchModelLoader(ModelLoader):
 
     Looks for `names` in the current directory, up to its two direct parents.
     """
-    for dir in ("./", "../", "../../"):
+    for dir in ("./", "../", "../../", "../../../"):
       for name in names:
         path = os.path.join(dir, name)
         if exists(path):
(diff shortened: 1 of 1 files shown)


miladm (approved):
LGTM

benchmarks/torchbench_model.py
```diff
@@ -193,7 +193,7 @@ def get_skip_data(self):
     its lists of models into sets of models.
     """
 
-    benchmarks_dir = self._find_near_file(("benchmarks",))
+    benchmarks_dir = self._find_near_file(("pytorch/benchmarks",))
```
ysiraichi, line 196:
```suggestion
    benchmarks_dir = self._find_near_file(("benchmarks", "pytorch/benchmarks"))
```

zpcore: Hi @ysiraichi, if you have a better fix in mind, can you create PR since you are more familiar with the code you created? I will drop this one. I am a bit concerned about adding `../../../../` in this PR.

ysiraichi: Sure thing.

zpcore: Sorry, just notice there is still an issue with the PR you made for the fix. I just quickly reverted it so it won't block our benchmark automation run.  I will merge this one first for the quick fix...

PR #6434, opened 2024-01-31
Disambiguate `benchmarks` directory search.

ysiraichi:
Follow-up: #6428

This PR disambiguates `pytorch/benchmarks` directory from `xla/benchmarks` directory. In order to do that without relying on the name of the directory that the repo lives, we look for `benchmarks/dynamo`, instead. Works, since the `dynamo` directory only exists on `pytorch/benchmarks`.

cc @miladm

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -193,11 +193,10 @@ class TorchBenchModelLoader(ModelLoader):
     its lists of models into sets of models.
     """
 
-    benchmarks_dir = self._find_near_file(("benchmarks",))
+    benchmarks_dir = self._find_near_file(("benchmarks/dynamo",))
     assert benchmarks_dir is not None, "PyTorch benchmarks folder not found."
 
-    skip_file = os.path.join(benchmarks_dir, "dynamo",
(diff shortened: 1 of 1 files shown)


zpcore (approved):
Thanks, LGTM!

zpcore:
Oh, just noticed that it is still unable to find the `benchmarks/dynamo`, because the file is under `pytorch/`. Let's add 'pytorch/benchmarks' instead... Let me revert this one for now.

ysiraichi:
@zpcore Interesting. How are the repos (PyTorch and PyTorch/XLA) organized in CI (or wherever you are trying to run the benchmarks)?

zpcore:
Hi @ysiraichi , I tried the following organization:
```
|--pytorch
|  --benchmarks
|     --dynamo
|     --...
|--xla
|  --benchmarks  <- I am running under this folder
|      -- 
```
In this case, it will not search under pytorch folder.

ysiraichi:
Hmm... I see. What do you say we leave both of them: `("benchmarks/dynamo", "pytorch/benchmarks")`?

zpcore:
Yes, this should also work. Thanks!

PR #6436, opened 2024-01-31
Remove random tests in test_core_aten_ops

qihqi:
We don't expect torch.rand to produce the same
numbers for 2 devices.


wonjoolee95:
This makes sense, but I think it would be worthwhile to have at least some kind of testing for all core aten ops in our end. So maybe just a execution of the torch_xla version of the op to make sure there is no crash and a lowering exists? Something like: https://github.com/pytorch/xla/pull/6427/files#diff-e64d0ea276eb592135f003cc2a0c5a8d3bda4e8789f19c5ef400c54b56f3abf9. 

But not a blocker, we can merge this first and follow up in a later PR.

wonjoolee95 (approved):

PR #6439, opened 2024-02-01
Update Dockerfile

ManfeiBai:


vanbasten23:
> I think this is also going to overwrite the pytorch build in our nightly images, which is not what we want.

Yeah, that's great point!
Perhaps we make the similar change in the release branch as r2.3, for example, so we get the correct `torch` package.

ManfeiBai:
TPU CI met the same error, so it happen on master branch too, the reason of this error only on TPUCI maybe due to only tests in TPUCI used torchvision?

ManfeiBai:
since https://github.com/pytorch/xla/pull/6458 merged, close this PR

PR #6458, opened 2024-02-02
bring TPU CI bakc to green by reinstall torch/torchvision

ManfeiBai:
the TPU CI has used `package_version` as `2.1.0` for a long time, and we update it to `2.3.0` since PyTorch/XLA has announced 2.2 release

add short term hack `pip3 install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cpu` to the TPU CI script to bring TPU CI green before replaced with the new TPU CI

### DIFF
--- a/infra/ansible/config/pip.yaml
+++ b/infra/ansible/config/pip.yaml
@@ -49,6 +49,5 @@ pip:
   # Packages that will be installed with the `--nodeps` flag.
   pkgs_nodeps:
     release_common:
-      - torch
       - torchvision
       - pillow
--- a/infra/ansible/config/pip.yaml
+++ b/infra/ansible/config/pip.yaml
@@ -49,5 +49,6 @@ pip:
   # Packages that will be installed with the `--nodeps` flag.
   pkgs_nodeps:
     release_common:
+      - torch
       - torchvision
       - pillow
--- a/test/tpu/xla_test_job.yaml
+++ b/test/tpu/xla_test_job.yaml
@@ -41,6 +41,7 @@ spec:
     - bash
     - -cxe
     - |
+      # TODO: new TPUCI please remove this short term hack for `RuntimeError: operator torchvision::nms does not exist`
       pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121
       pip install expecttest==0.1.6
       pip install rich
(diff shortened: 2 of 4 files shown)


ManfeiBai:
ok, according to the log of TPUCI(https://github.com/pytorch/xla/pull/6458/checks?check_run_id=21166603603), rename the `package_version` didn't change a lot, the chaneg would only rename the wheel, and this error might be a real error: `RuntimeError: operator torchvision::nms does not exist`

ManfeiBai:
let's build on TPU and test to see whether we could see this failure or not

infra/ansible/config/pip.yaml
```diff
@@ -49,5 +49,6 @@ pip:
   # Packages that will be installed with the `--nodeps` flag.
   pkgs_nodeps:
     release_common:
+      - torch
```
will-cromar, line 52: See this comment: https://github.com/pytorch/xla/pull/6439#pullrequestreview-1856977923

I agree with @vanbasten23's hack in the short term: https://github.com/pytorch/xla/pull/6439#issuecomment-1922481500

Basically, just add `pip3 install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cpu` to the TPU CI script. This is going to fail some of the time when the last nightly doesn't include required changes from the current head. But, failing some of the time is better than failing all of the time. TPU CI script: https://github.com/pytorch/xla/blob/master/test/tpu/xla_test_job.yaml

A short term hack is fine, since we're trying to replace the current TPU CI anyway. cc @mbzomowski

ManfeiBai: Thanks, Will, make sense and TPUCI passed with this change: https://github.com/pytorch/xla/runs/21241385079

let's add this short term hack and wait for TPU CI replacement too

test/tpu/xla_test_job.yaml
```diff
@@ -41,6 +41,7 @@ spec:
     - bash
     - -cxe
     - |
+      pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121
```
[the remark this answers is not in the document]
ManfeiBai: Thanks, good point, added comment

PR #6460, opened 2024-02-02
Wrap constant arg in XLAExportInterpreter

lsy323:
If the op spec expects a `Tensor` input, we wrap the python primitive type to a torch.tensor. The dtype for tensor respects `torch.get_default_dtype`.

Without this wrapping, the python float will be wrapped before it enters dispatcher, and it doesn't respect the global default dtype.

With this change, `aten.sub(1.0, x)` and `aten.sub(torch.tensor(1.0), x)` will be the same, when they enters the LTC.

It's still unknown where the scalar constant is wrapped into a tensor in upstream. It should be before it enters dispatcher (By investigation and syncing with PyTorch team).

### DIFF
--- a/torch_xla/stablehlo.py
+++ b/torch_xla/stablehlo.py
@@ -233,23 +233,24 @@ class XLAExportInterpreter(torch.fx.Interpreter):
     new_kwargs = dict(kwargs)
     if 'device' in kwargs:
       new_kwargs['device'] = self._device
-    # Note: Use `_disable_current_modes` to alwasys create constant tensor.
-    # Under `fake_tensor_mode` a fake tensor will be created. This is not a
-    # use case for XLAExportInterpreter right now, adding to be future-proof.
-    with torch.utils._python_dispatch._disable_current_modes():
-      # If the op spec expects a `Tensor` input, we wrap the python float/int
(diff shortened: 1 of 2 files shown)


torch_xla/stablehlo.py
```diff
@@ -233,6 +233,22 @@ def call_function(self, target, args: Tuple, kwargs: Dict) -> Any:
...
+      # If the op spec expects a `Tensor` input, we wrap the python primitive
+      # type to a torch.tensor. The dtype for float respects
+      # `torch.get_default_dtype`. Without this wrapping, the python float
+      # will be wrapped before it enters dispatcher, and it doesn't respect
+      # the global default dtype.
+      args_and_specs = tuple(zip(args, target._schema.arguments))
+      args = tuple(
+          map(
+              lambda arg_spec: torch.tensor(arg_spec[0])
+              if isinstance(arg_spec[0], (float, int, bool)) and type(arg_spec[
```
JackCaoG, line 249: why boolean is also included here? I understand float and int has 64 bit dtype issue but I thought there is only one boolean dtype

lsy323: Makes sense, let me remove bool

JackCaoG (approved):

Issue #5897, opened 2023-11-21
[Core ATen Opset] Lower aten_sin


wonjoolee95:
In order for PyTorch/XLA to support the PyTorch core ATen opset, it requires lowering each core ATen op in PyTorch/XLA. This issue is used to track the PyTorch/XLA lowering for **aten_sin**.

Here are some general guidelines to lowering this op:
  - Uncomment `@unittest.skip` or `@unittest.expectFailure` and run the unit test at [test_core_aten_ops.py](https://github.com/pytorch/xla/blob/master/test/test_core_aten_ops.py). Eg: `pytest test/test_core_aten_ops.py -k test_aten_sin_0`
  - Make code changes until the test passes. Read and follow [fix_lowering_for_core_aten_ops.md](https://github.com/pytorch/xla/blob/master/FIX_LOWERING_FOR_CORE_ATEN_OPS.md) for ideas to fix.
    - There may be multiple unit tests for a single op. For this op, the corresponding unit tests are:
      - test_aten_sin_3
    - Please also uncomment the skips for all these tests and ensure all tests are fixed.
    - Note that sometimes the fix may be to fix the unit tests itself. Please take a look at the corresponding unit tests to make sure the tests are valid.
  - Submit the PR!

For any questions, feel free to leave a comment in this PR.

ShatilKhan:
Hi @wonjoolee95 
Please assign me this issue

ShatilKhan:
Hi, thanks for assigning
I'm  new to PyTorch 
I read the docs a bit & found some setup guides, not sure which one to follow
I found this [PyTorch Notebook](https://www.kaggle.com/code/herbison/pytorch-on-cloud-tpus-multicore-example/notebook) on Kaggle
I read the [CONTRIBUTING](https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md) guide as well , should I start here for setting up?
I found the [OP_LOWERING_GUIDE](https://github.com/pytorch/xla/blob/master/OP_LOWERING_GUIDE.md) too , reading through it.
If any maintainers could provide some guidance or tips to getting started that would be very helpful

wonjoolee95:
Sure thing! Glad you'll be working on this, these should be good first issues to start contributing and get familiar with PyTorch/XLA. To answer your questions:

[...]

PR #6462, opened 2024-02-03
Fix some more core aten ops

wonjoolee95:
Fixes https://github.com/pytorch/xla/issues/5848
- `aten::as_strided` is an op that should never be dispatched to torch_xla, as it is a view op and functionalization pass in PyTorch should decompose into `aten::as_strided_copy` (as noted in https://github.com/pytorch/xla/blob/master/codegen/xla_native_functions.yaml#L387, note that functionalization is, by default, enabled in torch_xla and we want that to be the default/norm).Therefore, remove this from the test.

Fixes https://github.com/pytorch/xla/issues/5861
- Introduce a wrapper to pass in kwarg

Fixes https://github.com/pytorch/xla/issues/5859
- Introduce a wrapper to pass in kwarg

Fixes https://github.com/pytorch/xla/issues/5897
- Do manual dtype conversion in aten_sin

### DIFF
--- a/test/test_core_aten_ops.py
+++ b/test/test_core_aten_ops.py
@@ -1371,41 +1371,38 @@ class AtenOpTest(unittest.TestCase):
     kwargs = dict()
     run_export_and_compare(self, torch.ops.aten.div.Scalar, args, kwargs)
 
-  @unittest.skip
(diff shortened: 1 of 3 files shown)


ManfeiBai (approved):
LGTM

PR #6463, opened 2024-02-03
moving torch_xla2 experimental code in-tree

qihqi:
-- skip CIs for it, as it will not be built into torch_xla wheel. 
Will setup different github workflow scripts for it's unit test.


.github/workflows/build_and_test.yml
```diff
@@ -3,10 +3,14 @@ on:
     branches:
       - master
       - r[0-9]+.[0-9]+
+    paths-ignore:
+      - 'experimental/torch_xla2/**'
```
lsy323, line 7: Should be `torch_xla/experimental/torch_xla2/**`? Looks like the CI is still triggered.

lsy323 (approved):
One comment on the GH workflow change. Otherwise LGTM

.github/workflows/build_and_test.yml
qihqi: The files are added to `experimental/` and not `torch_xla/experimental` as the latter still builds into the wheel. 

I am guessing the CI triggered because of changing this file (build_and_test.yml). 

We'll find out if this is the case in next PR.

Issue #6465, opened 2024-02-05
torch.distributed.checkpoint and XLA backend


wintwoo:
Referring to the "SPMD user guide" here: https://github.com/pytorch/xla/blob/535d398b9a7d2952d04abe6395307897352664d2/docs/spmd.md

What's the right approach to use the `torch.distributed.checkpoint` (and the distributed checkpoint manager) on Cloud TPUs?

I understand I should call `init_process_group` to use the `torch.distributed.*` functions. When I use:

```
xr.use_spmd()
dist.init_process_group("xla", init_method="pjrt://")
```

I run into this broken assertion: `AssertionError: XLA backend is not supported with SPMD. Please use a CPU process group instead.`

Do I need to init a CPU process group (with rank, world size etc) in "parallel" with how I normally use `torch_xla.core.xla_model` for distribution on TPUs?

JackCaoG:
@jonb377 can you help? Let's also use this chance to update the user guide if needed.

jonb377:
In SPMD mode, we currently don't allow a process group to control the on-device collectives, since all collectives are inserted by the compiler. However, distributed checkpointing requires a process group to coordinate the checkpoint.

The current recommendation is to use a CPU process group such as `gloo`. The `xla://` init_method is still supported for master IP discovery on TPUs. Your example should work with:

```python
xr.use_spmd()
dist.init_process_group('gloo', init_method='xla://')
```

I missed this in the documentation, I'll add an example of `init_process_group` to the CheckpointManager section.

wintwoo:
Thank you for the clarification!

ManfeiBai:
Hi, Thanks, @wintwoo, does above helped answer all questions?

If yes, I would suggest to mark this issue as fixed; or let's continue discuss for follow-up questions

PR #6469, opened 2024-02-05
Add process group documentation for SPMD

jonb377:
As pointed out in #6465, our documentation is missing discussion of how to initialize the process group in SPMD execution mode.

A process group is required for distributed checkpointing and can be used with various other `torch.distributed` APIs. In SPMD, we don't allow process groups on the XLA backend, since the compiler is responsible for controlling the on-device collectives.

### DIFF
--- a/docs/spmd.md
+++ b/docs/spmd.md
@@ -320,6 +320,28 @@ for step, data in enumerate(dataloader):
         print(f'Checkpoint taken at step {step}')
 ```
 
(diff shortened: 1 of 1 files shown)


vanbasten23:
> cc @vanbasten23 , you might have already done, did we add the SPMD + GPU documentqtaion/section as well?

Good call. I haven't done that yet but let me add some tmr.

PR #6484, opened 2024-02-06
[benchmarks] Add `no-skip` argument.

ysiraichi:
This PR adds and argument for not skipping models.

cc @miladm

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -111,10 +111,17 @@ class ExperimentRunner:
           logger.info(f"SKIP already completed benchmark")
           continue
 
-        # Skip unsupported config.
-        if not self.model_loader.is_compatible(benchmark_model,
-                                               benchmark_experiment,
-                                               self._args.strict_compatible):
+        # Check if we should execute or skip the current configuration.
+        # A configuration SHOULD be skipped if and only if:
+        #
+        #   1. --no-skip was not specified; AND
+        #
+        #   2. the model is not compatible with the experiment configuration
+        #
+        # Otherwise, we should go ahead and execute it.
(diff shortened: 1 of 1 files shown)


benchmarks/experiment_runner.py
```diff
@@ -111,10 +111,17 @@ def generate_and_run_all_configs(self):
...
-                                               self._args.strict_compatible):
+        # Check if we should execute or skip the current configuration.
+        # A configuration SHOULD be skipped if and only if:
+        #
+        #   1. --no-skip was not specified; AND
+        #
+        #   2. the model is not compatible with the experiment configuration
+        #
+        # Otherwise, we should go ahead and execute it.
+        if (not self._args.no_skip and not self.model_loader.is_compatible(
```
miladm, line 122: don't we need to check for `no_skip` in line `99`? hoping to save some time not loading a model we don't intend to run.

zpcore:
Thanks, by the way, I think we can remove `--strict_compatible`  and the strict_deny_list. I tried with the skip list from the yaml file you introduced without `--strict_compatible` argument and it works fine now!

benchmarks/experiment_runner.py
ysiraichi: I don't think we can do that. We still have to _"load the model"_ for updating the environment variables of the child process (which will actually execute the benchmark). That said, this loading won't take many cycles, since we are calling with `dummy=True`: _"instantiate the wrapper `BenchmarkModel`, but don't instantiate the actual torchbench model."_

ysiraichi:
@zpcore Sure. I will open a new PR doing that.

zpcore (approved):
LGTM!

PR #6488, opened 2024-02-07
Update xla_docker_build.tf

ManfeiBai:


ManfeiBai:
wait for TPUCI result, although this change will only affect after merged;

this PR will be closed after gain the result, because this error disappear in the current TPUCI now, maybe the testing env has changed very quickly?

PR #6492, opened 2024-02-07
CI: increase build timeout from 180 to 240 minutes

cota:
Currently, unless the CI build cache is enabled, builds do time out at 180 minutes. Given that the CI build cache is only enabled on branches on the pytorch/xla repo, contributors external to the project (who therefore cannot push a branch to the pytorch/xla repo) cannot get CI to run.

Fix this by increasing the build timeout to 240 minutes.

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -38,7 +38,7 @@ on:
 jobs:
   build:
     runs-on: ${{ inputs.runner }}
-    timeout-minutes: 180
+    timeout-minutes: 240
     outputs:
       docker-image: ${{ steps.upload-docker-image.outputs.docker-image }}
     env:


.github/workflows/_build.yml
```diff
@@ -38,7 +38,7 @@ on:
 jobs:
   build:
     runs-on: ${{ inputs.runner }}
-    timeout-minutes: 180
+    timeout-minutes: 240
```
changm, line 41: Maybe add a comment here about the differences for CI/CD and why it's so long?

changm:
Thanks for fixing this. Non blocking - Any idea why caching is only enabled on branches of pytorch/xla repo? Can we enable caching everywhere?

changm (approved):

cota:
> Any idea why caching is only enabled on branches of pytorch/xla repo? Can we enable caching everywhere?
I believe that's up to the "pytorch" org on Github -- they are the ones who pay for the CI server AFAIK.

Issue #5982, opened 2023-12-02
[Core ATen Opset] Lower aten_embedding


wonjoolee95:
In order for PyTorch/XLA to support the PyTorch core ATen opset, it requires lowering each core ATen op in PyTorch/XLA. This issue is used to track the PyTorch/XLA lowering for **aten_embedding**.

Here are some general guidelines to lowering this op:
  - Uncomment `@unittest.skip` or `@unittest.expectFailure` and run the unit test at [test_core_aten_ops.py](https://raw.githubusercontent.com/pytorch/xla/master/test/test_core_aten_ops.py). Eg: `pytest test/test_core_aten_ops.py -k test_aten_embedding_0`
  - Make code changes until the test passes. Read and follow [fix_lowering_for_core_aten_ops.md](https://github.com/pytorch/xla/blob/master/FIX_LOWERING_FOR_CORE_ATEN_OPS.md) for ideas to fix.
    - There may be multiple unit tests for a single op. For this op, the corresponding unit tests are:
      - test_aten_embedding_1
      - test_aten_embedding_2
    - Please also uncomment the skips for all these tests and ensure all tests are fixed.
    - Note that sometimes the fix may be to fix the unit tests itself. Please take a look at the corresponding unit tests to make sure the tests are valid.
  - Submit the PR!

For any questions, feel free to leave a comment in this PR.

wonjoolee95:
Note to assignee: `_embedding_bag` is also a core aten op (https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L2318C9-L2318C23) and should also be lowered as part of this. Also we should lower the `_embedding_bag_forward_only` according to https://github.com/pytorch/xla/issues/2215.

wonjoolee95:
Hey @bhavya01, this should be a good next op you can work on. We can discuss details in our sync next week.

bhavya01:
Working on implementing EmbeddingBag. Looking at native pytorch implementation, the op handles a bunch of different use cases depending on the inputs. Will open a PR once a basic mean of the embedding vectors is working. https://github.com/pytorch/pytorch/blob/113138aa5575301d914c18bd882d6ab3735aa18a/aten/src/ATen/native/EmbeddingBag.cpp

bhavya01:
Was able to get a basic sum version of embedding bag working using xla::While node. Working on making small changes for mean, max and taking into account other params for the function.

wonjoolee95:
[...]

PR #6495, opened 2024-02-07
Lower the embedding op

bhavya01:
Reference: https://github.com/pytorch/pytorch/blob/113138aa5575301d914c18bd882d6ab3735aa18a/aten/src/ATen/native/Embedding.cpp#L37

The native pytorch implmentation also uses just the indices and weight matrix. Other options seem to be ignored.

### DIFF
--- a/test/cpp/test_aten_xla_tensor_5.cpp
+++ b/test/cpp/test_aten_xla_tensor_5.cpp
@@ -261,6 +261,9 @@ TEST_F(AtenXlaTensorTest, TestEmbedding) {
                                            /*scale_grad_by_freq=*/false,
                                            /*sparse=*/false);
     AllClose(b, xla_b);
+    ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
+    ExpectCounterChanged("xla::embedding_symint",
+                         cpp_test::GetIgnoredCounters());
   });
 }
 
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -115,9 +115,10 @@ void CheckSubOperandTypes(at::ScalarType type1, at::ScalarType type2) {
 
 c10::optional<at::ScalarType> PromoteIntegralType(
     at::ScalarType src_dtype, const c10::optional<at::ScalarType>& opt_dtype) {
-  return opt_dtype.has_value() ? opt_dtype.value()
-         : at::isIntegralType(src_dtype, /*includeBool=*/true) ? at::kLong
-                                                               : opt_dtype;
+  return opt_dtype.has_value()
+             ? opt_dtype.value()
+             : at::isIntegralType(src_dtype, /*includeBool=*/true) ? at::kLong
+                                                                   : opt_dtype;
 }
 
 bool IsTypeWithLargerRangeThanLong(torch::ScalarType dtype) {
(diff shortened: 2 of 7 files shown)


wonjoolee95:
@bhavya01, is this ready to be reviewed? Also can you add a corresponding cpp unit test as it is a new op?

bhavya01:
@wonjoolee95 This PR should be ready for a review for #5982

wonjoolee95:
Do we already have a cpp test for this op? Also wil `embedding_bag` also be lowered in this PR or will that be in a follow-up PR?

wonjoolee95:
@bhavya01, can we ensure that the lowering get invoked correctly by adding the metric check in the `embedding` cpp test like https://github.com/pytorch/xla/blob/master/test/cpp/test_aten_xla_tensor_5.cpp#L246? 

Thanks, let's also work on the `embedding_bag` as we have the customer request on that op. Thanks for working on this!

bhavya01:
Added the check for metrics.  EmbeddingBag is still WIP.

test/cpp/test_aten_xla_tensor_5.cpp
```diff
@@ -261,6 +261,9 @@ TEST_F(AtenXlaTensorTest, TestEmbedding) {
                                            /*scale_grad_by_freq=*/false,
                                            /*sparse=*/false);
     AllClose(b, xla_b);
+    ExpectCounterNotChanged("aten::.*", cpp_test::GetIgnoredCounters());
+    ExpectCounterChanged("xla::embedding_symint",
```
wonjoolee95, line 265: nit: `xla::embedding_syntint` -> `xla::embedding*`. Not blocker, we can update in the EmbeddingBag PR.

wonjoolee95: Also want to confirm one thing -- before implementing this lowering (i.e. without the changes in this PR), does this metric assertion fail as expected? Just wanted to make sure that the `_symint` variant is getting recognized properly. It should, but just wanted to confirm.

But not a blocker, we can follow-up in the EmbeddingBag PR.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -3515,11 +3515,10 @@ at::Tensor XLANativeFunctions::embedding_symint(const at::Tensor& weight,
     return at::native::embedding_symint(weight, indices, padding_idx,
                                         scale_grad_by_freq, sparse);
   }
-  // TODO: for now route to native, which dispatches supported XLA operations.
-  // We need to make use of the TPU embedding core here eventually.
```
wonjoolee95, line 3519: This comment (seems like it existed for a while) `We need to make use of the TPU embedding core here eventually.` makes me a bit cautious about the `Embedding` op. I assume this comment means to lower this op in a performant way, we somehow want to make use of the TPU embedding core? Just leaving this comment as a future reference. Also let's bring this comment back in the code so we are aware. Again, not a blocker, we can include as a follow-up in the next PR.

wonjoolee95 (approved):
Thanks! 

I left a few comments as follow-up items. Not a blocker, so feel free to merge this PR. But let's follow-up on the items in the next PR (preferably the EmbeddingBag PR).

test/cpp/test_aten_xla_tensor_5.cpp
bhavya01: Yes, without the change this check fails.

image

torch_xla/csrc/aten_xla_type.cpp
bhavya01: Sounds good. Will add it back in a follow up PR

PR #6498, opened 2024-02-07
[SPMD] Introduce global mesh

alanwaketan:
Summary:
SPMD expects the mesh to be the same across the board. Therefore, introduce the concept of a global mesh to reduce the need to carry the same mesh in the code.

Test Plan:
python test/spmd/test_xla_sharding.py -v -k test_global_mesh


alanwaketan:
Thanks for the quick review, Jon. Let me do the doc update later. I'm in a rush...

alanwaketan:
Let's ignore the GPU test given it takes forever. I'm in a very tight deadline.

PR #6500, opened 2024-02-08
[FSDPv2] Use the global mesh API

alanwaketan:
Summary:
Make FSDPv2 to use the global mesh API such that users can just set the global mesh.

Test Plan:
python test/spmd/test_fsdp_v2.py -v -k test_fsdp_v2_global_mesh


jonb377 (approved):

alanwaketan:
Thanks, Jon!

PR #6507, opened 2024-02-08
Lowering for _pdist_forward

zpcore:
Resolves https://github.com/pytorch/xla/issues/5885.

### DIFF
--- a/torch_xla/csrc/ops/ops.cpp
+++ b/torch_xla/csrc/ops/ops.cpp
@@ -21,6 +21,7 @@
 #include "torch_xla/csrc/ops/permute.h"
 #include "torch_xla/csrc/ops/softmax_backward.h"
 #include "torch_xla/csrc/ops/sum.h"
+#include "torch_xla/csrc/ops/unsqueeze.h"
 #include "torch_xla/csrc/ops/xla_ops.h"
 #include "torch_xla/csrc/pooling.h"
 #include "torch_xla/csrc/runtime/debug_macros.h"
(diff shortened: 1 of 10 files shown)


zpcore:
Oh, there is test failure in pytorch size for corner case shape [0,2], [1,2] and [3,0]. Let me make the fix.

torch_xla/csrc/ops/ops.cpp
```diff
@@ -566,6 +567,32 @@ torch::lazy::NodePtr Norm(const torch::lazy::Value& input,
...
+                      LoweringContext* loctx) -> XlaOpVector {
+    xla::XlaOp xla_input = loctx->GetOutputOp(node.operand(0));
+    return node.ReturnOp(BuildUpperTriangle(xla_input), loctx);
+  };
+  auto lower_for_shape_fn =
+      [&](absl::Span<const xla::XlaOp> operands) -> xla::XlaOp {
+    return BuildUpperTriangle(operands[0]);
+  };
+  torch::lazy::NodePtr tmp = input - torch::lazy::MakeNode<Unsqueeze>(input, 1);
+  torch::lazy::NodePtr result_matrix = Norm(tmp, p, dtype, {2}, false);
```
wonjoolee95, line 585: The logic LGTM, but typically we organize this logic in such a way that we have a function called `BuildPdistForward` (instead of `BuildUpperTriangle`) to have all the lowering-specific logic in that function instead of here. Can we do that for this as well?

wonjoolee95: Understood, makes sense. I was referring to putting the lines 584-585 in that `BuildPdistForward` as well.

zpcore: I see, this case is a bit tricky. Line 584-585 were handled in node(`torch::lazy::NodePtr`) level using the existing `Unsqueeze` and `Norm` Nodes, while the remaining is handled in lower xla op level since they don't provide `Norm` etc. as we needed here. I didn't see others handle `torch::lazy::NodePtr` in `Build*` functions. That's why I split them out.

wonjoolee95: I see, sg, let's just leave it as is. Thanks!

wonjoolee95 (approved):
LGTM, thanks!

PR #6508, opened 2024-02-09
Add explicit shape check for index_copy

wonjoolee95:
As per https://pytorch.org/docs/stable/generated/torch.Tensor.index_copy_.html, the dim-th dimension of the tensor must have the same size as the length of index. PyTorch behavior: 
```
>>> import torch
>>> x = torch.zeros(5, 3)
>>> t = torch.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=torch.float)
>>> index = torch.tensor([0, 4, 2, 1])
>>> x.index_copy_(0, index, t)
IndexError: index_copy_(): Number of indices (4) should be equal to source.size(dim) (3)
```
Update the torch_xla's `index_copy` lowering to throw a similar error message.

One other check we may want is checking each value in the index tensor to be within the dimensions of the input tensor. But because torch_xla's lowering of `index_copy` uses `xla::Scatter` which requires torch_xla to do some broadcast, doing this bound check for each index seems a bit tricky. So for now, this PR will first add this explicit shape check only.

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1440,7 +1440,7 @@ at::Tensor XLANativeFunctions::index_copy(const at::Tensor& self, int64_t dim,
   // https://pytorch.org/docs/stable/generated/torch.Tensor.index_copy_.html,
   // the dim-th dimension of source tensor must have the same size as the length
   // of index tensor.
-  XLA_CHECK(dim < self_tensor_dims.size() &&
+  XLA_CHECK(dim < source_tensor_dims.size() &&
             source_tensor_dims[dim] == index_tensor_dims[0])
       << "dim-th dimension of the source tensor must have the same size as the "
          "length of index tensor.";
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -1424,11 +1424,30 @@ at::Tensor XLANativeFunctions::index_add(const at::Tensor& self, int64_t dim,
 at::Tensor XLANativeFunctions::index_copy(const at::Tensor& self, int64_t dim,
                                           const at::Tensor& index,
                                           const at::Tensor& source) {
-  TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
   XLATensorPtr self_tensor = bridge::GetXlaTensor(self);
-  return bridge::AtenFromXlaTensor(
-      tensor_methods::index_copy(self_tensor, dim, bridge::GetXlaTensor(index),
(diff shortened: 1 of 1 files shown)


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1424,11 +1424,30 @@ at::Tensor XLANativeFunctions::index_add(const at::Tensor& self, int64_t dim,
...
+  XLATensorPtr index_tensor = bridge::GetXlaTensor(index);
+  absl::Span<const int64_t> index_tensor_dims =
+      index_tensor->shape().get().dimensions();
+
+  // As per
+  // https://pytorch.org/docs/stable/generated/torch.Tensor.index_copy_.html,
+  // the dim-th dimension of source tensor must have the same size as the length
+  // of index tensor.
+  XLA_CHECK(dim < source_tensor_dims.size() &&
+            source_tensor_dims[dim] == index_tensor_dims[0])
```
miladm, line 1444: For out of bound index issues, we need to verify the following test scenario is caught in tests.

```
x = torch.zeros(5, 3)
t = torch.tensor([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=torch.float)
index = torch.tensor([0, 9, 2]) # the second index (i.e. 9) is out of bound as the first dim of x is 5.
x.index_copy_(0, index, t)
```

miladm (changes requested):
thanks @wonjoolee95 added a comment.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1424,11 +1424,30 @@ at::Tensor XLANativeFunctions::index_add(const at::Tensor& self, int64_t dim,
...
-                                 bridge::GetXlaTensor(source)));
+  absl::Span<const int64_t> self_tensor_dims =
+      self_tensor->shape().get().dimensions();
+
+  XLATensorPtr source_tensor = bridge::GetXlaTensor(source);
+  absl::Span<const int64_t> source_tensor_dims =
+      source_tensor->shape().get().dimensions();
+
+  XLATensorPtr index_tensor = bridge::GetXlaTensor(index);
+  absl::Span<const int64_t> index_tensor_dims =
```
alanwaketan, line 1436: Maybe we can just query the length of the index tensor as it should be a vector.

alanwaketan (commented):
Can we add a test case?

wonjoolee95:
This PR just implements an out-of-range check `index_copy`, so we should be able to just fix the failing assertions and merge them. Or we can just close it. cc @miladm

wonjoolee95:
I'll close this for now as it is not relevant, but feel free to re-open if necessary.

PR #6509, opened 2024-02-09
[benchmarks] Increase compilation cache.

ysiraichi:
Fix: #5967 

This PR increases the compilation cache for 2 benchmarks that were failing due to an assertion error: `cachedComputation`.

- cm3leon_generate
- hf_T5_generate

It also changes the `DENY_LIST` so that we only skip these benchmarks if they are running on XLA without dynamo (timeout).

cc @miladm


vanbasten23 (approved):
LGTM! Thanks.

PR #6512, opened 2024-02-09
[benchmarks] Fix execution with AMP precision.

ysiraichi:
This PR moves AMP to be only activated in the forward execution.

cc @miladm

### DIFF
--- a/benchmarks/benchmark_model.py
+++ b/benchmarks/benchmark_model.py
@@ -66,6 +66,8 @@ class BenchmarkModel:
     self.suite_name = suite_name
     self.model_name = model_name
     self.benchmark_experiment = benchmark_experiment
+    self.autocast = contextlib.nullcontext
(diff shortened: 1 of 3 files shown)


zpcore:
Hi @ysiraichi , did you see any issue recently that GPU xla backend also failed to run? I am not sure if this is related to the data format changes.

ysiraichi:
Not really...

vanbasten23:
Overall, LGTM.
One quick question, why do we want to moves AMP to be only activated in the forward execution? If we dont do it, what error do you see?

ysiraichi:
It's not exactly an error. But [this is how AMP is supposed to be used](https://pytorch.org/docs/stable/amp.html#torch.autocast). This is also how PyTorch's benchmarking script uses it (if we want to have comparable numbers).

golechwierowicz (approved):

PR #6514, opened 2024-02-09
Add more pjrt runtime tests for GPU.

vanbasten23:


test/pjrt/test_runtime_single_proc_gpu.py
```diff
@@ -38,12 +38,35 @@ def test_num_local_devices(self):
...
-    self.assertLen(torch_xla._XLAC._xla_get_all_devices(),
-                   xr.global_device_count())
+  def test_num_global_device_count(self):
     self.assertEqual(self.num_cuda_devices, xr.global_device_count())
 
+  def test_local_process_count(self):
+    self.assertEqual(1, xr.local_process_count())
+
+  # xw32: need review here.
+  def test_world_size(self):
```
vanbasten23, line 50: hey @will-cromar , on the single-processing case, is 1 the expected output of xr.world_size()?

torch_xla/runtime.py
```diff
@@ -139,7 +139,7 @@ def global_device_count() -> int:
 
 @requires_pjrt
 def world_size() -> int:
-  """Returns the total number of configured logical devices."""
```
will-cromar, line 142: This is still the number of configured logical devices. For SPMD, that number happens to be equal to the number of processes, since there's one logical device per process.

torch_xla/runtime.py
```diff
@@ -215,7 +215,7 @@ def global_runtime_device_attributes() -> List[Dict[str, object]]:
 @requires_pjrt
 @functools.lru_cache()
 def global_runtime_device_count() -> int:
-  """Returns the total number of runtime devices across all processes/hosts."""
+  """Returns the total number of runtime devices across all processes/hosts. Currently this is only used in SPMD."""
```
will-cromar, line 218: I would say this is _useful_ for SPMD, since otherwise it's the same as `global_device_count`.

test/pjrt/test_runtime_single_proc_gpu.py
will-cromar: Either case, it's the number of logical devices. For SPMD, that's one per process, and normally one per host. For multiprocessing, this is the number of global devices.

Think of this as the number of `torch.device`s in the whole system. SPMD bundles all runtime devices as one `torch.device`, and multiprocessing exposes each one individually.

vanbasten23: Thanks for the explanation. I get confused when reading PyTorch's doc https://pytorch.org/docs/stable/distributed.html which defines `world_size` as `Number of processes participating in the job.`. I really hope both PyTorch and PyTorch/XLA can unify the definition.

Does the expectation below look correct to you?
- Here, in the non-SPMD single-processing test, `xr.world_size()` should return 1.
- In the non-SPMD multi-processing test (the test `test_world_size` in pytorch/xla/test/pjrt/test_runtime_multi_gpu.py in this PR), `xr.world_size()` should return the number of GPU devices.
- In the SPMD case, `xr.world_size()` should return 1.

will-cromar: > I really hope both PyTorch and PyTorch/XLA can unify the definition.

If you except our multithreading hackery for old TPUs (or treat threads as processes for this definition), these definitions do align most of the time. Each process/thread should get one logical device if you start through `xmp.spawn` or `torchrun`.

> Here, in the non-SPMD single-processing test, xr.world_size() should return 1.

I don't think that's true after your fix to expose all available GPUs to a single process by default. You would see devices `['xla:0', 'xla:1', 'xla:2', 'xla:3']`. Unless you don't set `xm.set_replication_devices` --  then `xm.world_size` is 1... I want to simplify this latter condition out, but we need either refactoring or an explicit API change.

vanbasten23: I see. I also spent some time digging the issue and here is what I found:

1. The only place where we use `xr.world_size()` is [here](https://github.com/pytorch/xla/blob/1d4a8da5f8e0507731ef2178aa7d24495faf9bf4/torch_xla/_internal/rendezvous.py#L48) and it's the place where we interact with upstream PyTorch: if it's spmd, we assign `xr.process_count()` to `world_size`; otherwise, we assign `xr.world_size()`. If the `world_size` there will be used by pytorch, then I feel it should follow pytorch's definition of `world_size` ( the number of processes instead of number of devices). What do you think?

2. On TPU, the non-SPMD single-processing, I can see:
```
# PJRT_DEVICE=TPU python
>>> xr.global_device_count()
8
>>> xr.world_size()
1
>>> xr.local_device_count()
8
>>> xr.addressable_device_count()
8
```

So I feel the discussion boils down to how we define world_size, whether it's the number of devices or the number of processes. If it's the latter, both TPU and GPU are already doing the right thing. Otherwise, we may need to change both.

vanbasten23: discussed offline, will assert equal to 1

torch_xla/runtime.py
vanbasten23: Per our discussion. let me know what you think of the new comment.

torch_xla/runtime.py
vanbasten23: done

will-cromar (approved):

vanbasten23:
Thanks for the review!

Issue #5904, opened 2023-11-21
[Core ATen Opset] Lower aten_topk


wonjoolee95:
In order for PyTorch/XLA to support the PyTorch core ATen opset, it requires lowering each core ATen op in PyTorch/XLA. This issue is used to track the PyTorch/XLA lowering for **aten_topk**.

Here are some general guidelines to lowering this op:
  - Uncomment `@unittest.skip` or `@unittest.expectFailure` and run the unit test at [test_core_aten_ops.py](https://github.com/pytorch/xla/blob/master/test/test_core_aten_ops.py). Eg: `pytest test/test_core_aten_ops.py -k test_aten_topk_0`
  - Make code changes until the test passes. Read and follow [fix_lowering_for_core_aten_ops.md](https://github.com/pytorch/xla/blob/master/FIX_LOWERING_FOR_CORE_ATEN_OPS.md) for ideas to fix.
    - There may be multiple unit tests for a single op. For this op, the corresponding unit tests are:
      - test_aten_topk_3
    - Please also uncomment the skips for all these tests and ensure all tests are fixed.
    - Note that sometimes the fix may be to fix the unit tests itself. Please take a look at the corresponding unit tests to make sure the tests are valid.
  - Submit the PR!

For any questions, feel free to leave a comment in this PR.

wonjoolee95:
Had a sync with @ManfeiBai, to make the `aten::topk` core aten op test succeed:
1. Run the core_aten_op unit test and put the failure here. My guess is that native PyTorch converts the output into floating type even when input type if integer, but PyTorch/XLA doesn't do that.
2. If our guess from 1) is correct, then we have to add a manual dtype conversion. However, this op is not codegened, so it might be little tricky. You can probably do a manual dtype conversion here: https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_lower_util.cpp#L358. Or maybe even a layer above. You can take a look and decide which layer to do it.

ManfeiBai:
Thanks, @wonjoolee95,

- tested locally for failure:
```
(torch310) root@cb1bb455239f:~/pytorch/xla# pytest test/test_core_aten_ops.py -k test_aten_topk_0
============================= test session starts ==============================
platform linux -- Python 3.10.13, pytest-7.4.4, pluggy-1.4.0
rootdir: /root/pytorch
configfile: pytest.ini
plugins: hypothesis-6.92.2
collected 522 items / 521 deselected / 1 selected                              

[...]

PR #6519, opened 2024-02-12
Fix some more core aten ops

wonjoolee95:
Fixes https://github.com/pytorch/xla/issues/5904
- Update the test to use randperm, since torch.topk is not *stable* when there are duplicates.

Fixes https://github.com/pytorch/xla/issues/5857
- Update to do manual dtype conversion

Fixes https://github.com/pytorch/xla/issues/5856
- Update to do manual dtype conversion

### DIFF
--- a/torch_xla/csrc/ops/ops_lower_fn.cpp
+++ b/torch_xla/csrc/ops/ops_lower_fn.cpp
@@ -344,11 +344,19 @@ torch_xla::XlaOpVector ClampMinTensor::Lower(LoweringContext* loctx) const {
 
 torch_xla::XlaOpVector Cos::Lower(LoweringContext* loctx) const {
   xla::XlaOp xla_input = loctx->GetOutputOp(operand(0));
+  if (xla::primitive_util::IsIntegralType(XlaHelpers::TypeOfXlaOp(xla_input))) {
+    xla::PrimitiveType input_type = XlaHelpers::TypeOfXlaOp(xla_input);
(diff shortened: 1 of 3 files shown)


ManfeiBai (approved):
LGTM

torch_xla/csrc/ops/ops_lower_fn.cpp
```diff
@@ -344,11 +344,19 @@ torch_xla::XlaOpVector ClampMinTensor::Lower(LoweringContext* loctx) const {
 
 torch_xla::XlaOpVector Cos::Lower(LoweringContext* loctx) const {
   xla::XlaOp xla_input = loctx->GetOutputOp(operand(0));
+  if (xla::primitive_util::IsIntegralType(XlaHelpers::TypeOfXlaOp(xla_input))) {
+    xla::PrimitiveType input_type = XlaHelpers::TypeOfXlaOp(xla_input);
+    xla_input = ConvertTo(xla_input, input_type, xla::PrimitiveType::F32);
```
vanbasten23, line 349: Do you know why we want to convert the input type to F32 when it is the integral type?

wonjoolee95: It was to address a feature gap in torch and torch_xla. In torch, `aten::cos` allows integer types as inputs. However, torch_xla didn't. This PR fixes that. More details at https://github.com/pytorch/xla/pull/6110.

PR #6523, opened 2024-02-12
aggregate.py: fix detection of no datapoints

cota:
The existing check became obsolete once we added the --backends flag in PR #6392; in particular, len(speedups) could be 1 when a single backend is passed to --backends. Fix it and add a test to make sure we emit no output to stdout (note that the warning message goes to stderr).

### DIFF
--- a/benchmarks/aggregate.py
+++ b/benchmarks/aggregate.py
@@ -330,7 +330,7 @@ def pr_latest(results_map: Dict[str, Any], args, timestamps: List[str]):
             list, zip(*sorted(zip(acc_map[label], acc_map[model_label]))))
         speedups[i] = list(map(pr_round, speedups[i]))
         break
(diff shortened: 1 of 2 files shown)


benchmarks/aggregate.py
```diff
@@ -330,7 +330,7 @@ def pr_latest(results_map: Dict[str, Any], args, timestamps: List[str]):
             list, zip(*sorted(zip(acc_map[label], acc_map[model_label]))))
         speedups[i] = list(map(pr_round, speedups[i]))
         break
-  if not speedups[0] or not speedups[1]:
+  if all([not x for x in speedups]):
```
golechwierowicz, line 333: You can drop square parentheses.

golechwierowicz (approved):

PR #6528, opened 2024-02-13
Add default flag `xla_tpu_prefer_async_allgather_to_allreduce`

jonb377:
AllGather decomposition into AllReduce by the compiler is less memory efficient, the result being shared between the forward and backward pass instead of rematerialized. This problem is particularly pronounced with 1D sharding over a 3D topology.

With FSDPv2 being the easiest entrypoint for SPMD, this issue becomes a blocker for performance at large scale, since most device memory will be allocated for the AllGathered parameters, constraining the batch size. On v4, the problem begins to manifest with v4-128's 4x4x4 topology.

This flag only applies with async AllGather enabled. On v4, this is enabled by default with `TPU_MEGACORE=megacore_dense`. On v5, the flag currently won't take effect because async AG requires continuation fusion, which is currently unstable and not enabled by default.

On a v4-128, enabling this option decreases peak memory usage from 17.5GB to 3GB on a 7B parameter model with the same batch size.

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -45,9 +45,20 @@ def _setup_libtpu_flags():
   flags = _set_missing_flags(flags,
                              (('xla_latency_hiding_scheduler_rerun', '1'),))
 
+  # This flag will prevent AllGather decomposition into AllReduce by the
+  # compiler when async AllGather is enabled. Decomposed AllGathers are
+  # persisted in-memory and shared between the forward and backward passes,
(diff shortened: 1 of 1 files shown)


alanwaketan (approved):
LGTM. Thanks Jon for identifying the issue and take a quick fix.

JackCaoG:
hmm, seem safe since real `all_gather` is mostly used in SPMD case, non-spmd they are just `all_reduce` to begin with.

JackCaoG (approved):

PR #6533, opened 2024-02-14
Add limit to max inflight TPU computations

will-cromar:
Requires `XLA_REGISTER_INSTALLED_PLUGINS=1` to use `DevicePlugin` API.

### DIFF
--- a/torch_xla/_internal/tpu.py
+++ b/torch_xla/_internal/tpu.py
@@ -346,6 +346,6 @@ class TpuPlugin(plugins.DevicePlugin):
 
   def client_create_options(self):
     return {
-      'max_inflight_computations':
-          xu.getenv_as('XLA_TPU_MAX_INFLIGHT_COMPUTATIONS', int, 4)
+        'max_inflight_computations':
+            xu.getenv_as('XLA_TPU_MAX_INFLIGHT_COMPUTATIONS', int, 4)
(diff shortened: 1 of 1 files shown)


torch_xla/_internal/tpu.py
```diff
@@ -343,3 +343,9 @@ def configure_multiprocess(self, local_rank, local_world_size):
 
   def physical_chip_count(self):
     return num_available_chips()
+
+  def client_create_options(self):
```
alanwaketan, line 347: who calls this method?

alanwaketan: Thanks, Will. Hard to tell it it's a virtual method in pythong. haha

alanwaketan (approved):
LGTM.

torch_xla/_internal/tpu.py
```diff
@@ -343,3 +343,9 @@ def configure_multiprocess(self, local_rank, local_world_size):
...
+    return {
+        'max_inflight_computations':
+            xu.getenv_as('XLA_TPU_MAX_INFLIGHT_COMPUTATIONS', int, 4)
```
miladm, line 350: shall we track this new variable in https://github.com/pytorch/xla/blob/master/configuration.yaml?

PR #6548, opened 2024-02-16
Add flag for HLO post-optimization check

ManfeiBai:


JackCaoG:
Thanks, @ManfeiBai can you reply to the origional issue for how to run with these flags to unblock user?

PR #6550, opened 2024-02-16
[benchmarks] Fix AMP data-type.

ysiraichi:
This PR fixes the behavior of AMP, aligning it to [that of PyTorch HUD](https://github.com/pytorch/pytorch/blob/3f4dd9bfa4c43d9d314d50573eacac9afeb474bb/benchmarks/dynamo/common.py#L2007-L2013). Instead of setting a default data-type, we use the default one.

cc @miladm

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -383,6 +383,15 @@ class TorchBenchModel(BenchmarkModel):
     return torch.bfloat16
 
   def _get_autocast_with_kwargs(self):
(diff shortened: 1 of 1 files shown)


JackCaoG:
@yeounoh can you review this one? I don't remember why we set the default dtype.

ysiraichi:
I have reset the AMP data-types according to whatever is the default data type in `torch.<device>.amp` (or `torch_xla.amp`).

PR #6552, opened 2024-02-16
Add XLA specific XPlane profiling.

golechwierowicz:
Following internal discussion with @vanbasten23 I'm adding an option to use the native XLA profiler. The exported xplane file can be opened in the internal tooling, and it shows what is happening on the host side. That's in contrast to the native pytorch profile we open via perfetto.

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -24,7 +24,9 @@ from enum import Enum
 from torchbench_model import TorchBenchModelLoader
 from benchmark_experiment import ExperimentLoader
 from util import reset_rng_state, move_to_device, randomize_input, us_to_s, ns_to_s
+
 import torch_xla.core.xla_model as xm
+import torch_xla.debug.profiler as xp
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


benchmarks/experiment_runner.py
```diff
@@ -906,6 +932,11 @@ def main():
     logger.warning('Enabling fast F32 multiplication for PyTorch')
     torch.set_float32_matmul_precision('high')
 
+  if args.profile_xla:
+    logger.info(
+        'Enabling XLA XPlane profiling. Do not benchmark with this option set.')
+    server = xp.start_server(9012)
```
vanbasten23, line 938: consider adding a FLAGS.profiler_xla_port default to 9012

vanbasten23 (approved):
LGTM. Thanks!

benchmarks/experiment_runner.py
golechwierowicz: Do you mean importing some specific constant, or parametrising server port?

vanbasten23: I meant parametrising server port, such as https://github.com/pytorch/xla/blob/dc08cb6893e84f29fc31ed4e44e8998ed680b2a1/test/spmd/test_train_spmd_imagenet.py#L380

PR #6553, opened 2024-02-16
Do not cache input args in dynamo bridge

lsy323:
Do not cache FX Graph args in `GraphInputMatcher`. It's not necessary to cache those tensors, because user inputs will be used in each run.

Without this change, the cached input tensors will cause more device memory usage as new graph is cached.

`xla_args` is the user inputs passed down from Dynamo, the cached tensors are released based on the `tensor_id` of `xla_args`, when each `GraphInputMatcher` is constructed.


lsy323:
Hi @wonjoolee95 @JackCaoG, this PR is ready for review. The CI failed because of the staled GCP token. I ran the dynamo tests locally and they passed.

test/dynamo/test_num_output.py
```diff
@@ -78,10 +78,14 @@ def do_test(self, model_class, expected_num_output):
...
-
+    xla_args_tensor_ids = set(
+        tree_map_only(torch.Tensor,
+                      lambda input: torch_xla._XLAC._xla_get_tensor_id(input),
+                      inputs))
     graph_input_matcher = GraphInputMatcher(tensor_id_to_arg_idx,
                                             graph_input_tensor_ids,
-                                            graph_input_xla_values)
+                                            graph_input_xla_values,
+                                            xla_args_tensor_ids)
```
JackCaoG, line 88: can we add a test in following pr to check `graph_input_matcher. graph_input_xla_values`? Some of them should be None.

JackCaoG (approved):

wonjoolee95 (approved):

test/dynamo/test_num_output.py
lsy323: Added a test for checking the None in the cached tensor list

test/dynamo/test_graph_input_matcher.py
```diff
@@ -0,0 +1,60 @@
+import unittest
```
JackCaoG, line 1: you need to add this test to the `test/run_test.sh`. Let's do that in a follow up pr. I will merge this one to unblock hexllm folks

JackCaoG (approved):

test/dynamo/test_graph_input_matcher.py
lsy323: Ok, sorry missed this comment yesterday, thanks for adding it in a new PR @JackCaoG !

Issue #6399, opened 2024-01-29
[RFC] Flatten basic APIs


will-cromar:
# TL;DR

Move some basic functionality to the top-level `torch_xla` module, including:

*   Getting and counting addressable XLA devices
*   Launching distributed training in various environments
*   Synchronizing the graph

# Introduction

The ergonomics of PyTorch/XLA could use some work.

Reviewing our [Getting Started diffs](https://github.com/pytorch/xla?tab=readme-ov-file#getting-started), converting even the most basic PyTorch code to use XLA requires 3 imports (non-`torch.distributed`) or 4 imports (`torch.distributed`), because the fully qualified names for even basic functions are excessively verbose. For example, to even get the XLA device, the name mentions XLA 3 times: `torch_xla.core.xla_model.xla_device`.

This is most noticeable to me when I'm prototyping a new script or debugging in an interactive REPL or notebook. Although we have limited concrete feedback from new users, I expect the verbosity is a deterrent to someone seeing PyTorch/XLA for the first time.

# Objectives

My proposal is simple: pull fundamental functionality to the top-level `torch_xla` package. I'll define fundamental functionality as the following:

*   The basic functions that you could not reasonably use PyTorch/XLA without
*   The first functions you run when you open a new REPL or notebook

We can leave more advanced functions organized into lower-level packages. The discussion of usability within the PyTorch/XLA team more or less comes down to how we expose complexity gradually to users, rather than dumping all of the complexity at once. Users can then dig deeper as needed to improve performance (e.g. data preloading, SPMD, etc) once they're up and running. There are more proposals coming in that vein. The goal of this proposal specifically is to make the basics clear and accessible.

I'm intentionally taking a conservative approach here. It's much easier to add a visible function to our public API than remove it. For that reason, we should start with the APIs we're most confident will remain consistent in the long-term.

# Proposal

## The easy one: `torch_xla.device()`

There is no way to construct a PyTorch/XLA program without actually moving data and modules to the XLA device. So, there's no reason to bury this function in our API. `torch_xla.device` would effectively just be an alias to the existing `xm.xla_device()`, minus the deprecated `devkind` argument.

## Listing and counting devices

Although these functions are not always required to write PyTorch/XLA programs like `xm.xla_device()`, they are a first-line debugging tool users can use to inspect their environment. I propose the following new functions:

[...]

PR #6571, opened 2024-02-20
Add basic device APIs to the top-level `torch_xla` module.

will-cromar:
Implemented basic device APIs from #6399.

- Add `device`, `devices`, `real_devices`, and `device_count` as described in RFC.
- We already do a substantial amount of setup in `__init__.py` . Add `torch_xla.py` for public functions on `torch_xla` module.

Follow up:
- Update documentation to use new APIs.
- Start deprecating or discouraging usage of old APIs like `xm.xla_device` and `runtime.xla_device`.

### DIFF
--- a/test/test_devices.py
+++ b/test/test_devices.py
@@ -7,20 +7,21 @@ import torch_xla.runtime as xr
 
 
 class TestDevices(parameterized.TestCase):
+
   def setUpClass():
     xr.set_device_type('CPU')
     os.environ['CPU_NUM_DEVICES'] = '4'
 
-  @parameterized.parameters(
-      (None, torch.device('xla:0')),
-      (0, torch.device('xla:0')),
-      (3, torch.device('xla:3')))
+  @parameterized.parameters((None, torch.device('xla:0')),
+                            (0, torch.device('xla:0')),
+                            (3, torch.device('xla:3')))
   def test_device(self, index, expected):
     device = xla.device(n=index)
     self.assertEqual(device, expected)
 
   def test_devices(self):
-    self.assertEqual(xla.devices(), [torch.device(f'xla:{i}') for i in range(4)])
+    self.assertEqual(xla.devices(),
+                     [torch.device(f'xla:{i}') for i in range(4)])
 
   def test_real_devices(self):
     self.assertEqual(xla.real_devices(), [f'CPU:{i}' for i in range(4)])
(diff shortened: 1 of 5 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1188,14 +1188,20 @@ void InitXlaModuleBindings(py::module m) {
...
-  });
+  m.def("_xla_real_devices",
+        [](const std::optional<std::vector<std::string>> devices) {
+          if (!devices) {
+            return runtime::GetComputationClient()->GetLocalDevices();
+          }
+
+          std::vector<std::string> xla_devices;
+          {
+            NoGilSection nogil;
```
vanbasten23, line 1199: Do you know why we release the GIL in this block? Is it for tpu v2/v3 where we allow multiple threads to do some runtime job such as `GetXlaDevices(*devices)`?

will-cromar: I'm not sure to be honest. Maybe `GetLocalDevices` was a blocking call in XRT?

test/test_devices.py
```diff
@@ -0,0 +1,34 @@
...
+import torch
+import torch_xla as xla
+import torch_xla.runtime as xr
+
+
+class TestDevices(parameterized.TestCase):
+
+  def setUpClass():
+    xr.set_device_type('CPU')
+    os.environ['CPU_NUM_DEVICES'] = '4'
```
JackCaoG, line 13: hmm, it is OK for now but shouldn't we also test it on GPU and TPU?

torch_xla/__init__.py
```diff
@@ -182,3 +182,5 @@ def _init_xla_lazy_backend():
 if os.getenv('XLA_REGISTER_INSTALLED_PLUGINS') == '1':
   plugins.use_dynamic_plugins()
   plugins.register_installed_plugins()
+
+from .torch_xla import *
```
JackCaoG, line 186: why is this needed?

test/test_devices.py
will-cromar: This is sufficient IMO. We're really just testing the integration of this module with the runtime client, which has the same API regardless of the underlying device.

As we switch to using these functions by convention, they'll be exercised by almost every other test.

torch_xla/__init__.py
will-cromar: This imports the contents of `torch_xla.py` into `torch_xla/`'s module scope. Otherwise, the functions would be `torch_xla.torch_xla.etc`. This assigns them to `torch_xla.etc`

torch_xla/csrc/init_python_bindings.cpp
JackCaoG: oh lol it is that `torch_xla`

JackCaoG (approved):

Issue #6546, opened 2024-02-16
Pytorch XLA SPMD CheckpointManager Optimizer Loading Error


mfatih7:
Hello

I am using CheckpointManager to save and load checkpoints for models with SPMD.
In my training setup, in the first epoch, the model is trained with lossA only.
After the first epoch, the model is trained using lossA and lossB.
There is no problem while loading checkpoints in the later epochs.
But specifically while loading a checkpoint saved after the first epoch I get the error below.

```
  File "/home/THEUSER/17_featureMatching/run_train_TPU.py", line 101, in <module>
    training_results = train_1_1_each_sample_in_single_batch_TPU_spmd_train_and_val(   
  File "/home/THEUSER/17_featureMatching/process/train_1_1_each_sample_in_single_batch_TPU_spmd.py", line 81, in train_and_val
    loss_checkpoint, proc_time_checkpoint = checkpoint.load_checkpoint( config, device, model, optimizer, chkpt_mgr=chkpt_mgr)    
  File "/home/THEUSER/17_featureMatching/checkpoint.py", line 100, in load_checkpoint
    chkpt_mgr.restore( max(tracked_steps), checkpoint )
  File "/home/THEUSER/env3_8/lib/python3.8/site-packages/torch_xla/experimental/distributed_checkpoint/manager.py", line 319, in restore
    dist_cp.load_state_dict(
  File "/home/THEUSER/env3_8/lib/python3.8/site-packages/torch/distributed/checkpoint/state_dict_loader.py", line 31, in load_state_dict
    return _load_state_dict(
  File "/home/THEUSER/env3_8/lib/python3.8/site-packages/torch/distributed/checkpoint/state_dict_loader.py", line 185, in _load_state_dict
    central_plan = distW.reduce_scatter("plan", local_step, global_step)
  File "/home/THEUSER/env3_8/lib/python3.8/site-packages/torch/distributed/checkpoint/utils.py", line 185, in reduce_scatter
    raise result
torch.distributed.checkpoint.api.CheckpointException: CheckpointException ranks:dict_keys([0])
Traceback (most recent call last): (RANK 0)
  File "/home/THEUSER/env3_8/lib/python3.8/site-packages/torch/distributed/checkpoint/utils.py", line 158, in reduce_scatter
    local_data = map_fun()
  File "/home/THEUSER/env3_8/lib/python3.8/site-packages/torch/distributed/checkpoint/state_dict_loader.py", line 175, in local_step
[...]

PR #6572, opened 2024-02-20
Add optimizer priming for dist chkpt

jonb377:
See also: https://github.com/pytorch/xla/issues/6546

The optimizer state must be primed before it can be restored. Optimizer state isn't materialized until the first `optim.step` call, so to restore optimizer state before resuming training, a dummy step is needed.

This PR introduces the `prime_optimizer` API, which will run a dummy optimizer step with zeroed gradients. The gradient sharding is copied from the parameters to ensure the resulting sharding is the same.

### DIFF
--- a/docs/spmd.md
+++ b/docs/spmd.md
@@ -297,7 +297,7 @@ checkpointing directly to any fsspec-compatible filesystem, including GCS.
 Example usage of the CheckpointManager is below:
 
 ```python
-from torch_xla.experimental.distributed_checkpoint import CheckpointManager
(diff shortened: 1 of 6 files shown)


alanwaketan (approved):
LGTM! Thanks, Jon.

jonb377:
Hi @dasoto, I've found that this approach will not guarantee the same sharding in the optimizer compared to running a full training step (this is due to sharding propagation decisions in the compiler). I believe the AdamW unit test was broken after an openxla pin update, for example.

Since this is an experimental feature, I would be OK to merge after a rebase. cc @JackCaoG @alanwaketan

Issue #4298, opened 2022-12-08
Forward-pass NN model with dynamic input doesn't work on CPU, but work on GPU and TPU.


vanbasten23:
## 🐛 Bug

Forward-pass NN model with dynamic input doesn't work on CPU, but work on GPU and TPU, per [PR](https://github.com/pytorch/xla/pull/4256).

## To Reproduce

On cloudtop docker container created via 
```
$ sudo docker pull gcr.io/tpu-pytorch/xla_base:latest
$ sudo docker run -it gcr.io/tpu-pytorch/xla_base:latest
```

```
$ export XLA_EXPERIMENTAL="nonzero:masked_select"
$ python3 pytorch/xla/test/test_dynamic_shape_models.py TestDynamicShapeModels.test_forward_pass_dynamic_input_correctness
```

The error message we get is 
```
======================================================================
ERROR: test_forward_pass_dynamic_input_correctness (__main__.TestDynamicShapeModels)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/pytorch/xla/test/test_dynamic_shape_models.py", line 51, in test_forward_pass_dynamic_input_correctness
    xm.mark_step()
  File "/opt/conda/lib/python3.7/site-packages/torch_xla-1.14-py3.7-linux-x86_64.egg/torch_xla/core/xla_model.py", line 953, in mark_step
    wait=xu.getenv_as('XLA_SYNC_WAIT', bool, False))
RuntimeError: INVALID_ARGUMENT: From /job:localservice/replica:0/task:0:
2 root error(s) found.
  (0) INVALID_ARGUMENT: Fail to proof the equality of two dimensions at compile time: %reduce.56 = s32[] reduce(s32[10]{0} %convert.50, s32[] %constant.51), dimensions={0}, to_apply=%add_S32.52 vs %reduce.17 = s32[] reduce(s32[10]{0} %convert.11, s32[] %constant.12), dimensions={0}, to_apply=%add_S32.13
	 [[{{node XRTCompile}}]]
	 [[XRTCompile_G3]]
  (1) INVALID_ARGUMENT: Fail to proof the equality of two dimensions at compile time: %reduce.56 = s32[] reduce(s32[10]{0} %convert.50, s32[] %constant.51), dimensions={0}, to_apply=%add_S32.52 vs %reduce.17 = s32[] reduce(s32[10]{0} %convert.11, s32[] %constant.12), dimensions={0}, to_apply=%add_S32.13
[...]

PR #6578, opened 2024-02-21
Replicate test dynamic shapes models with the equivalent version of torch.compile

changm:


test/ds/test_dynamic_shape_models_dynamo.py
```diff
@@ -0,0 +1,125 @@
...
+      num_non_zero_added += 1
+      if num_non_zero_added == num_non_zeros:
+        break
+
+    x_test_nonzero_dev = torch.nonzero(x_test.int()).float()
+    y_test_nonzero_dev = torch.nonzero(y_test.int()).float().squeeze()
+    return x_test_nonzero_dev, y_test_nonzero_dev
+
+if __name__ == '__main__':
+  test = unittest.main(verbosity=FLAGS.verbosity, exit=False)
```
vanbasten23, line 122: you may want to check if the ds feature flag `assert os.environ['XLA_EXPERIMENTAL'] != ''`?

test/ds/test_dynamic_shape_models_dynamo.py
```diff
@@ -0,0 +1,125 @@
...
+    model = Feedforward(num_features, hidden_size=10)
+    self.backwards_pass_with_dynamic_input(model, num_features)
+
+  def test_compiled_backward_pass_with_dynamic_input(self):
+    num_features = 2
+    model = Feedforward(num_features, hidden_size=10)
+    model.compile()
+    self.backwards_pass_with_dynamic_input(model, num_features)
+
+  def backwards_pass_with_dynamic_input(self, model, num_features):
```
vanbasten23, line 68: nit: mind making it private "_backwards_pass_with_dynamic_input"?

test/ds/test_dynamic_shape_models_dynamo.py
```diff
@@ -0,0 +1,125 @@
...
+                                                     num_features)
+      y_pred = model(x_test)
+      criterion(y_pred.squeeze(), y_test).item()
+    print('Test passed.')
+
+  def create_dynamic_test_data(self,
+                               num_test_samples,
+                               num_features,
+                               num_non_zeros=1):
+    x_test = torch.zeros(num_test_samples, num_features)
```
vanbasten23, line 98: need to move the input and the model to xla device?

test/ds/test_dynamic_shape_models_dynamo.py
```diff
@@ -0,0 +1,125 @@
...
+
+    for i in range(0, 100):
+        self.backwards_pass_with_dynamic_input(model, num_features)
+
+  def test_backward_pass_with_dynamic_input(self):
+    num_features = 2
+    model = Feedforward(num_features, hidden_size=10)
+    self.backwards_pass_with_dynamic_input(model, num_features)
+
+  def test_compiled_backward_pass_with_dynamic_input(self):
```
vanbasten23, line 62: what's the difference between this test and "test_benchmark_backwards_pass_with_dynamic_input"?

changm: It's the same, but "test_benchmark_backwards_pass_with_dynamic_input" is used to easily benchmark it.

test/ds/test_dynamic_shape_models_dynamo.py
changm: Maybe I should've given more context, or let me know if this isn't generally OK in this repo. Since the original test didn't work on CPU, I couldn't test this on XLA. I mostly wanted to commit this to try to reproduce the benchmark results you had here - https://github.com/pytorch/xla/blob/master/docs/dynamic_shape.md and see what inductor did, so that when XLA works, this test could be reproduced.

The current test_dynamic_shape_models.py file is skipped on CI/CD since it isn't on a TPU. Does that make sense or are these kinds of tests not allowed in this repo?

test/ds/test_dynamic_shape_models_dynamo.py
changm: ditto from the other comment, this is scaffolding to get a basic test working on inductor, then try to make it work on XLA as a follow up.

test/ds/test_dynamic_shape_models_dynamo.py
changm: Done.

test/ds/test_dynamic_shape_models_dynamo.py
vanbasten23: I see. Fwiw, test_dynamic_shape_models.py is running in the post-submit TPU CI https://github.com/pytorch/xla/blob/49a6eec96a24f7f4030f20e2fae9a43bf72599f9/.github/workflows/tpu_ci.yml#L44. Though, there's a plan to make TPU CI presubmit soon.

> when XLA works, this test could be reproduced.

Could you elaborate this? What's not working in XLA now?

changm: Ahh I see, I was only testing on CPU and the original test skips on CPU:

https://github.com/pytorch/xla/blob/master/test/ds/test_dynamic_shape_models.py#L50

See https://github.com/pytorch/xla/issues/4298

Is there a way to test TPU CI? Otherwise as a follow up, can I try to get this test working on TPU with XLA and torch.compile() ?

changm: I tired testing this on TPU and get the following errors:

```
FAILED test/ds/test_dynamic_shape_models_dynamo.py::TestDynamicShapeModels::test_backward_pass_with_dynamic_input - RuntimeError: torch_xla/csrc/aten_xla_bridge.cpp:95 : Check failed: xtensor
FAILED test/ds/test_dynamic_shape_models_dynamo.py::TestDynamicShapeModels::test_benchmark_backwards_pass_with_dynamic_input - torch._dynamo.exc.InternalTorchDynamoError: 'torch._C._SymNode' object has no attribute 'shape_env'
FAILED test/ds/test_dynamic_shape_models_dynamo.py::TestDynamicShapeModels::test_compiled_backward_pass_with_dynamic_input - torch._dynamo.exc.InternalTorchDynamoError: 'torch._C._SymNode' object has no attribute 'shape_env'
FAILED test/ds/test_dynamic_shape_models_dynamo.py::TestDynamicShapeModels::test_forward_pass_dynamic_input_correctness - torch._dynamo.exc.InternalTorchDynamoError: 'torch._C._SymNode' object has no attribute 'shape_env'
```

So would it be better to:

1) Check this in as with XLA support and disable the test
2) Check this in as a passing test without XLA support
3) Don't check this in at all?

Thanks!

vanbasten23: afaict, we don't usually merge non torch_xla related changes into the repo. So I'd like to understand the intention here. If the intention is to merge the test and let it run in the CI to prevent regression, we should merge it. If the test serves as a toy example to experiment what/how inductor does, then I'm hesitant to merge it. Wdyt?

changm: The intention was, use it as a toy example such that `torch.compile(..)` == the existing test that works with XLA, and make it easier for others to reproduce rather than pulling another branch / having it get stale. If that doesn't fit in this repo, then it's ok to not merge.

That's also why I originally just had it without XLA support in the first PR so that it was running in CI to prevent regression. Does that make sense?

vanbasten23: > That's also why I originally just had it without XLA support in the first PR so that it was running in CI to prevent regression.

Sorry I forgot which first PR you are referring to. By "prevent regression", do you mean the regression of native pytorch dynamo on dynamic input?

changm: Ahh all good, sorry for my miscommunication. I meant the original version of this PR was prevent a regression of the native PyTorch dynamo on dynamic input.

vanbasten23: If it's purely testing pytorch feature on dynamic input, shouldn't it be added to the pytorch repo?

changm: Sorry, this is the untouched version so that it works with dynamo and the test can be updated as we get support for XLA to slowly transform this test to use XLA. Does that make sense?

But generally it sounds like that type of development isn't supported in the PyTorch/XLA repo? Kind of like experimental?

vanbasten23: I see. If it helps make the development work easier, feel free to merge. But please update the pr description about the intention of adding this script and this script come from pytorch (with a link?).

vanbasten23 (approved):

Issue #5995, opened 2023-12-02
[Core ATen Opset] Lower aten_relu


wonjoolee95:
In order for PyTorch/XLA to support the PyTorch core ATen opset, it requires lowering each core ATen op in PyTorch/XLA. This issue is used to track the PyTorch/XLA lowering for **aten_relu**.

Here are some general guidelines to lowering this op:
  - Uncomment `@unittest.skip` or `@unittest.expectFailure` and run the unit test at [test_core_aten_ops.py](https://raw.githubusercontent.com/pytorch/xla/master/test/test_core_aten_ops.py). Eg: `pytest test/test_core_aten_ops.py -k test_aten_relu_0`
  - Make code changes until the test passes. Read and follow [fix_lowering_for_core_aten_ops.md](https://github.com/pytorch/xla/blob/master/FIX_LOWERING_FOR_CORE_ATEN_OPS.md) for ideas to fix.
    - There may be multiple unit tests for a single op. For this op, the corresponding unit tests are:
      - test_aten_relu_2
    - Please also uncomment the skips for all these tests and ensure all tests are fixed.
    - Note that sometimes the fix may be to fix the unit tests itself. Please take a look at the corresponding unit tests to make sure the tests are valid.
  - Submit the PR!

For any questions, feel free to leave a comment in this PR.

Nandika-A:
Hello! can I work in this issue?

kurayami07734:
assign this issue to me

wonjoolee95:
> assign this issue to me

Done!

kurayami07734:
Are there any similar issue which have been merged?
It will be a good reference for me

wonjoolee95:
The description of this issue should have a lot of good information on how to fix these issues. Also you can look at closed issues with label "core aten opset" -- https://github.com/pytorch/xla/issues?q=label%3A%22core+aten+opset%22+is%3Aclosed.

wonjoolee95:
Hey @kurayami07734, I'll take over this issue for now. Let me know if you're interested in contributing to other issues. Thanks!

Issue #5880, opened 2023-11-21
[Core ATen Opset] Lower aten__native_batch_norm_legit


wonjoolee95:
In order for PyTorch/XLA to support the PyTorch core ATen opset, it requires lowering each core ATen op in PyTorch/XLA. This issue is used to track the PyTorch/XLA lowering for **aten__native_batch_norm_legit**.

Here are some general guidelines to lowering this op:
  - Uncomment `@unittest.skip` or `@unittest.expectFailure` and run the unit test at [test_core_aten_ops.py](https://github.com/pytorch/xla/blob/master/test/test_core_aten_ops.py). Eg: `pytest test/test_core_aten_ops.py -k test_aten__native_batch_norm_legit_0`
  - Make code changes until the test passes. Read and follow [fix_lowering_for_core_aten_ops.md](https://github.com/pytorch/xla/blob/master/FIX_LOWERING_FOR_CORE_ATEN_OPS.md) for ideas to fix.
    - There may be multiple unit tests for a single op. For this op, the corresponding unit tests are:
      - test_aten__native_batch_norm_legit_0
      - test_aten__native_batch_norm_legit_1
    - Please also uncomment the skips for all these tests and ensure all tests are fixed.
    - Note that sometimes the fix may be to fix the unit tests itself. Please take a look at the corresponding unit tests to make sure the tests are valid.
  - Submit the PR!

For any questions, feel free to leave a comment in this PR.

wonjoolee95:
The `aten__native_batch_norm_legit ` tests are failing due to:
```
======================================================================
ERROR: test_aten__native_batch_norm_legit_0 (__main__.AtenOpTest) [stablehlo_can_run]
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/wonjoo/pytorch/xla/test/test_core_aten_ops.py", line 68, in run_export_and_compare
    res2 = shlo(*args, **kwargs)
  File "/home/wonjoo/pytorch/xla/torch_xla/stablehlo.py", line 110, in __call__
    return self.evaluate(self._default_method, args)
  File "/home/wonjoo/pytorch/xla/torch_xla/stablehlo.py", line 86, in evaluate
    res = pytree.tree_unflatten(res, out_spec)
  File "/home/wonjoo/miniconda3/lib/python3.10/site-packages/torch/utils/_pytree.py", line 818, in tree_unflatten
    return treespec.unflatten(leaves)
[...]

PR #6580, opened 2024-02-21
Fix some more core aten ops

wonjoo-wj:
Fixes https://github.com/pytorch/xla/issues/5980
- As per https://github.com/pytorch/xla/blob/master/torch_xla/csrc/aten_xla_type.cpp#L698-L699, XLA does not expose tensor storage, so just updating the test values to supported types.

Fixes https://github.com/pytorch/xla/issues/5991
- As per https://github.com/pytorch/xla/issues/5991#issuecomment-1856515770, torch_xla already lowers `aten_nonzero` but it is not enabled by default due to dynamic shape related issues. Given that we have other tests that tests lowering of torch_xla's `nonzero` that is controlled with XLA flags, we'll remove these tests for now.

Fixes https://github.com/pytorch/xla/issues/5883
- No op, just enable them

Fixes https://github.com/pytorch/xla/issues/5995
- No op, just enable them

Remove `_native_batch_norm_legit` tests
- These ops are already lowered in torch_xla, as per [xla_native_functions.yaml#L357-L358](https://github.com/pytorch/xla/blob/master/codegen/xla_native_functions.yaml#L357-L358). However, as per [aten/src/ATen/native/native_functions.yaml#L4263](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L4263), there seems to be an issue with wrongly fetching the schema of the function (also noted in https://github.com/pytorch/xla/issues/5880#issuecomment-1955743675). Given that we already have lowerings for these ops and unit tests, we'll remove these from the core aten op tests for now.


ManfeiBai (approved):
LGTM

wonjoolee95:
Kokoro test failed due to timeout but does not seem related. All the other CIs are green, merging this.

PR #6587, opened 2024-02-22
Add API to donate input buffer for dynamo execution

JackCaoG:
This pr adds an api `torch_xla._XLAC._set_buffer_donation` which can be used to mark the buffer associated with current tensor to be donated in the next execution. This api is currently only enabled for dynamo(torch.compile) use case, ti is a no-op in LazyTensor world(since LTC already has the auto aliasing based on inplace op and dynamo's functionization pass remove all inplace ops).

Example usage should be
```
  def dummy_inplace_add(self, input):
    input += 1
    return

  def test_manual_buffer_donation(self):
    device = xm.xla_device()
    input = torch.randn(5, 5).to(device)
    dummy_inplace_add_compiled = torch.compile(
        self.dummy_inplace_add, backend='openxla')

    met.clear_all()
    # input is a device_data, we should be able to set the buffer donation field.
    self.assertTrue(torch_xla._XLAC._set_buffer_donation(input, True))
    # make sure buffer donation setting is correctly updated
    self.assertTrue(torch_xla._XLAC._get_buffer_donation(input))
	
	for _ in range(100):
      # You don't need to keep calling this function if you function does not cause dynamo recompilation.
      # check below for the reason.
      self.assertTrue(torch_xla._XLAC._set_buffer_donation(input, True))
      dummy_inplace_add_compiled(input)
```

Please note a couple things
1. `_set_buffer_donation` is called on a tensor, but being applied to the buffer associated with this tensor. If the tensor you passed in is not a buffer(for example an intermediate tensor that has not been evulated) this api will be no-op and return false.
2. Buffer aliasing is being set up during compilation time. Torch Dynamo also does not track this field, so even if you change `_set_buffer_donation` after first execution(torch.compile compilation triggered at first execution of the compiled function), aliasing will not change.

### DIFF
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -28,6 +28,25 @@ dynamo_debug = int(os.environ.get('XLA_DYNAMO_DEBUG', '0')) == 1
 ptxla_debug = int(os.environ.get('PT_XLA_DEBUG', '0')) == 1
 
 
+class AliasWithBufferDonorContext(object):
+
+  def __init__(self, should_alias: bool):
+    self.should_alias = should_alias
+
+  def __enter__(self):
+    self.env_inited = 'XLA_SHOULD_ALIAS_WITH_BUFFER_DONOR' in os.environ
+    if self.env_inited:
+      self.env_saved = os.environ['XLA_SHOULD_ALIAS_WITH_BUFFER_DONOR']
+    os.environ[
+        'XLA_SHOULD_ALIAS_WITH_BUFFER_DONOR'] = '1' if self.should_alias else '0'
+
+  def __exit__(self, exc_type, exc_val, exc_tb):
+    if self.env_inited:
+      os.environ['XLA_SHOULD_ALIAS_WITH_BUFFER_DONOR'] = self.env_saved
+    else:
+      del os.environ['XLA_SHOULD_ALIAS_WITH_BUFFER_DONOR']
+
+
 @dataclasses.dataclass
 class GraphInputMatcher:
   """
@@ -307,9 +326,10 @@ def extract_graph_helper(xla_model: torch.fx.GraphModule):
   # calculate graph hash
   dumb_return_handler = DumbReturnHandler(xla_args, args_and_out,
                                           xla_args_need_update_bool)
-  graph_hash = torch_xla._XLAC._get_graph_hash(args_and_out)
-  if dynamo_debug:
-    print("graph_hash", graph_hash)
+  with AliasWithBufferDonorContext(True) as context:
+    graph_hash = torch_xla._XLAC._get_graph_hash(args_and_out)
+    if dynamo_debug:
+      print("graph_hash", graph_hash)
 
   # Collect all device data nodes that is needed to compute the args_and_out
   # and wrap those device data nodes inside a at::tensor(graph_input_xla_values).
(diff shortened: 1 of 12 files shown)


JackCaoG:
I chatted with @bdhirsh , ideally we want to figure out which input tensor can be aliased without user calling this api. There should be a way for us to retrive the aliasing information from functionization pass in aot-autograd.

The context is that If we need to alias an input buffer to output buffer, we need to make sure input buffer can not be accessed by original tensor. For example 
```
a += 1
```
is fine because we know once the in place ops happens, a will point to a different buffer, so it is safe to reuse that buffer to something else.

if it is 
```
b = a + 1
```
I can't alias because a's buffer is still needed. once aliasing happens, original buffer's value will be invalidated. if we alias in the 
```
b = a + 1
```
case. `a`'s value will be incorrect after this computation. Dynamo comes with functionization which will remove all inplace ops from the fx graph passed down, which is a problem in this case.

torch_xla/core/dynamo_bridge.py
```diff
@@ -28,6 +28,25 @@
 ptxla_debug = int(os.environ.get('PT_XLA_DEBUG', '0')) == 1
 
 
+class AliasWithBufferDonorContext(object):
+
+  def __init__(self, should_alias: bool):
+    self.should_alias = should_alias
```
lsy323, line 34: Question: When do we want to set it to false? Seems like we always want to alias buffer if possible

torch_xla/csrc/helpers.cpp
```diff
@@ -928,13 +929,20 @@ xla::StatusOr<xla::XlaComputation> XlaHelpers::WrapXlaComputation(
...
+      // Both input and output will be a tuple so parameter_number will always
+      // be
+      // 0
+      builder.SetUpAlias(/*output_index=*/xla::ShapeIndex({output_index}),
+                         /*param_number=*/0,
+                         /*param_index=*/xla::ShapeIndex({input_index}));
+    }
+  } else if (buffer_donor_indices.size() > 0) {
+    for (size_t i : buffer_donor_indices) {
+      builder.AddBufferDonor(/*param_number=*/0,
```
lsy323, line 943: param_number is always 0 because HLO always generate 1 tuple output?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1368,13 +1439,26 @@ XLAGraphExecutor::SyncTensorsGraphInternal(
...
-      TryRunCachedSync(tensors, &coll, &po_data, tensor_data_vec);
-  if (async != nullptr) {
-    return async;
+
+  std::pair<bool, std::shared_ptr<XLAGraphExecutor::Async>> cache_res =
+      TryRunCachedSync(tensors, &coll, &po_data, tensor_data_vec,
+                       warm_up_cache_only);
+  if (cache_res.first) {
+    // we have a cache hit, execution has been scheduled by TryRunCachedSync.
+    return cache_res.second;
```
lsy323, line 1461: Question: Looks like after adding the `warm_up_cache_only` arg, the `cached_res.second` will be null if `warm_up_cache_only=true`, why do we need this behavior?

lsy323 (approved):
LGTM, left a few questions

torch_xla/core/dynamo_bridge.py
JackCaoG: yea, I guess one way to do it is to let this context manager only set it to true..

JackCaoG: Let me adjust this along with @will-cromar 's comment in https://github.com/pytorch/xla/pull/6616

torch_xla/csrc/helpers.cpp
JackCaoG: yea, output of the origional HLO(assuming we don't wrap it) is always a tuple

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: we do. This is actually kind of funny, from dynamo's persprctive
```
a = a + 1
```
and 
```
a += 1
```
are different so it will generate different dynamo hash and ask XLA to compiles. However from XLA's perspective, the HLO generated for both cases are the same. When `a += 1` compiled by dynamo after `a = a + 1`, it will ask PyTorch/XLA to compile this new graph. Without the change here, we will call `RunCachedGraph` and schedule an execution. The change here was to separate the 
1. check if there is a graph hash hit
2. schedule the graph execution

In the warm_up_cache case, we only want to check if there is a hash match(to avoid recompilation).

JackCaoG:
I will merge this change to unblock the user, fix comments in a follow up pr.

alanwaketan (commented):
Sorry for being late, but looks good to me.

PR #6601, opened 2024-02-23
nightly.sh: fix assignment of timestamp default

cota:
sed(1) does not return a non-zero exit code on match failure. We therefore fail to assign a default timestamp.

Fix it by ending the pipe with grep, which does return a non-zero exit code on match failure. Note that we could use sed's "q" operator to achieve the same result, but I think using grep is preferable here because sed's "q" operator is not that well known.


frgossen (approved):
Thanks!

PR #6532, opened 2024-02-14
[Fori_loop|While_loop] Placeholder lower torch.while_loop with python dispatch for simple addition test case

ManfeiBai:
This is the step one PR:

Due to `torch.while_loop` is defined in [torch/_higher_order_ops/while_loop.py](https://github.com/pytorch/pytorch/blob/abadbbc4b03138ebe5a892e19ae06eaa41b104d0/torch/_higher_order_ops/while_loop.py#L46), not described in [native_functions.yaml](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml), lower `torch.while_loop` via python dispatch like [PR#5763](https://github.com/pytorch/xla/pull/5763)

`TODO`: current status need to patch the `python dispatch` in PyTorch code, we need modify this PR to enable the python dispatch from PyTorch/XLA code

child PR: https://github.com/pytorch/xla/pull/6529

PR resource: https://github.com/pytorch/xla/pull/6527

PR #6529, opened 2024-02-13
[Fori_loop] Support torch.while_loop with multi situations

ManfeiBai:
This is the step three PR:

Add multi while_loop situation test case

father PR: (child of) https://github.com/pytorch/xla/pull/6603

source PR: https://github.com/pytorch/xla/pull/6527

PR #6603, opened 2024-02-23
[Fori_loop|While_loop] Enable fori_loop with add/sub test case

ManfeiBai:
For fori_loop implementation with while_loop, this PR is for lowering body/cond to replace formal placeholder

This is the step two PR, and father PR(#6532), child PR(#6529), source PR(#6563)

---
some issue fixed:
- [x] ~body fn is `-`, not a torch func, will test later~ tried torch.sub(a, b), passed too locally
- [x] current code has changed many logic of lowering, let's move these logics to a new function without affecting the existing functions
- [x] ~input are limited to list/tuple~ this match torch._higher_order_ops.while_loop required
- [x] ~input was trans from list to not list after torch.compile, TODO, add the same logic like torch.compile to use inputs, not like currently create a duplicated tensor in the fori_loop.py file~

### DIFF
--- a/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
+++ b/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
@@ -41,6 +41,7 @@ class WhileLoopTest(unittest.TestCase):
 
     def cond_fn(x):
       limit_value = torch.ones(1, dtype=torch.int32, device=device)
+      limit_value[0] = 10
       return x[0] <= limit_value[0]
 
     def body_fn(x):
--- a/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
+++ b/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
@@ -40,14 +40,13 @@ class WhileLoopTest(unittest.TestCase):
     device = xm.xla_device()
 
     def cond_fn(x):
-      limit_value = torch.tensor([10], dtype=torch.int32, device=device)
-      
+      limit_value = torch.ones(1, dtype=torch.int32, device=device)
       return x[0] <= limit_value[0]
 
     def body_fn(x):
       return (torch.add(x[0], 1),)
 
-    xi = torch.tensor([1], dtype=torch.int32, device=device)
+    xi = torch.tensor([0], dtype=torch.int32, device=device)
     res = while_loop(cond_fn, body_fn, (xi,))
     expected = _fake_while_loop(cond_fn, body_fn, xi)
     self.assertEqual(expected, res)
--- a/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
+++ b/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
@@ -24,8 +24,8 @@ class WhileLoopTest(unittest.TestCase):
     device = xm.xla_device()
 
     def cond_fn(x):
-      ten = torch.ones(1, dtype=torch.int32, device=device)
-      return x[0] >= ten[0]
+      limit_value = torch.ones(1, dtype=torch.int32, device=device)
+      return x[0] >= limit_value[0]
 
     def body_fn(x):
       return (torch.sub(x[0], 1),)
@@ -35,6 +35,22 @@ class WhileLoopTest(unittest.TestCase):
     expected = _fake_while_loop(cond_fn, body_fn, xi)
     self.assertEqual(expected, res)
 
+  def test_while_loop_tpu_addition(self):
+
+    device = xm.xla_device()
+
+    def cond_fn(x):
+      limit_value = torch.tensor([10], dtype=torch.int32, device=device)
+      
+      return x[0] <= limit_value[0]
+
+    def body_fn(x):
+      return (torch.add(x[0], 1),)
+
+    xi = torch.tensor([1], dtype=torch.int32, device=device)
+    res = while_loop(cond_fn, body_fn, (xi,))
+    expected = _fake_while_loop(cond_fn, body_fn, xi)
+    self.assertEqual(expected, res)
 
 if __name__ == '__main__':
   test = unittest.main()
--- a/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
+++ b/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
@@ -28,7 +28,7 @@ class WhileLoopTest(unittest.TestCase):
       return x[0] >= ten[0]
 
     def body_fn(x):
-      return (x[0] - 1,)
+      return (torch.sub(x[0], 1),)
 
     xi = torch.tensor([5], dtype=torch.int32, device=device)
     res = while_loop(cond_fn, body_fn, (xi,))
--- a/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
+++ b/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
@@ -12,29 +12,48 @@ import torch_xla.core.xla_builder as xb
 
 
 def _fake_while_loop(cond_fn, body_fn, operands):
-  while cond_fn(operands):
-    operands = body_fn(operands)
+  while cond_fn(operands[0], operands[1]):
+    operands = body_fn(operands[0], operands[1])
   return operands
 
 
 class WhileLoopTest(unittest.TestCase):
 
-  def test_while_loop_tpu(self):
+  def test_while_loop_tpu_subtraction(self):
 
     device = xm.xla_device()
 
-    def cond_fn(x):
-      limit_value = torch.ones(1, dtype=torch.int32, device=device)
-      return x[0] >= limit_value[0]
+    def cond_fn(init, limit_value):
+      return limit_value[0] <= init[0]
 
-    def body_fn(x):
-      return (torch.sub(x[0], 1),)
+    def body_fn(init, limit_value):
+      one_value = torch.ones(1, dtype=torch.int32, device=device)
+      two_value = limit_value.clone()
+      return (torch.sub(init, one_value), two_value)
 
-    xi = torch.tensor([5], dtype=torch.int32, device=device)
-    res = while_loop(cond_fn, body_fn, (xi,))
-    expected = _fake_while_loop(cond_fn, body_fn, xi)
+    init = torch.tensor([10], dtype=torch.int32, device=device)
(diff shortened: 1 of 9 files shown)


torch_xla/csrc/lowering_context.cpp
```diff
@@ -169,6 +169,33 @@ xla::StatusOr<xla::XlaComputation> LoweringContext::BuildXla() {
   return xla;
 }
 
+xla::StatusOr<xla::XlaComputation> LoweringContext::BuildXlaWithCheck() {
+  xla::StatusOr<xla::XlaComputation> xla;
+  if (!root_tuple_.empty() & (root_tuple_.size() > 1)) {
+    xla::XlaOp root = xla::Tuple(builder(), root_tuple_);
+    xla = builder()->Build(root);
+  } else if (!root_tuple_.empty() & (root_tuple_.size() == 1)) {
```
ManfeiBai, line 177: we need to skip tuple for cond/body computation creation due to match the xla::While format check

JackCaoG:
@amithrm FYI

ManfeiBai:
kokoro failure should be fixed on master branch, let's skip it now

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1022,12 +1050,15 @@ void BuildLoweringContextSubmodule(py::module* m) {
...
       .def("build", &PyLoweringContext::Build)
+      .def("build_for_while", &PyLoweringContext::BuildForWhile)
       .def("hlo", &PyLoweringContext::GetHlo)
       .def("hlo_text", &PyLoweringContext::GetHloText)
       .def("hlo_json", &PyLoweringContext::GetHloJsonText)
       .def("parameter_id_tensor_mapping",
            &PyLoweringContext::GetParameterIdTensorMapping)
-      .def("tensor_parameter_id", &PyLoweringContext::GetTensorParameterId);
+      .def("tensor_parameter_id", &PyLoweringContext::GetTensorParameterId)
+      .def("setnamestring", &PyLoweringContext::SetNameString)
```
yeounoh, line 1060: Do you need this setter/getter for the name? Isn't there always one context/graph in your impl that you can name whatever? Also, "settnamestring" -> "set_name_string"

torch_xla/experimental/fori_loop.py
```diff
@@ -21,22 +21,48 @@ def while_loop(cond_fn, body_fn, operands):
...
+  builder = xb.create_builder('test_while')
+  params = []
+  # secondparams = []
+  for shape in shapes:
+    p = xb.mkparam(builder, len(params), shape)
+    params.append(p)
+    # secondparams.append(xb.Op.tuple([p]))
+
+  # generate cond_fn xlacomputation
+  xm.mark_step()
```
yeounoh, line 36: Let's keep this PR as draft until we experiment with `mark_step()`

ManfeiBai: Thanks, we have a simple test PR for experimental of without `mark_step()`: https://github.com/pytorch/xla/pull/6643, will wait for test results in https://github.com/pytorch/xla/pull/6643

torch_xla/csrc/init_python_bindings.cpp
ManfeiBai: Thanks, updated the function name in newer commit;

yes, we need this setter/getter for the name to recognize body_computation/cond_computation by name given when we create the lowering context; With recognized the body_computation/cond_computation by name, we could skip tuple for cond/body computation creation to match the xla::While [format requirement for cond](https://github.com/openxla/xla/blob/0f46dddc43c1b4989db8806a9b9abd128234e48a/xla/service/shape_inference.cc#L3167), and [error log](https://gist.github.com/ManfeiBai/19b6fc987a15a020a665e5498a446e8b)

and we have tried another solution like below, which modification could not be caught when create computation:
```
body_ctx_builder = body_ctx.builder()
body_ctx_builder.name_ = 'bodyctx'
```

so we choose to add this setter/getter function for xla::While requirement, and please feel free to redirect me to other possible solutions if I missed

qihqi (approved):

torch_xla/csrc/lowering_context.h
```diff
@@ -71,6 +76,11 @@ class LoweringContext : public torch::lazy::LoweringContext {
   // embedded XLA builder (returned by the builder() API).
   xla::StatusOr<xla::XlaComputation> BuildXla();
 
+  // Build the XLA computation capturing all the operations created with the
+  // embedded XLA builder (returned by the builder() API) with check whether
+  // build for cond/body computation or not.
+  xla::StatusOr<xla::XlaComputation> BuildXlaWithCheck();
```
yeounoh, line 82: Can you merge the logic to the original `BuildXla()`, since you can condition on the designated computation names?

torch_xla/experimental/fori_loop.py
```diff
@@ -21,22 +21,46 @@ def while_loop(cond_fn, body_fn, operands):
...
+  # create inputs placeholder
+  kwargs = {}
+  shapes = xb.tensor_shape(operands)
+  builder = xb.create_builder('test_while')
+  params = []
+  # secondparams = []
+  for shape in shapes:
+    p = xb.mkparam(builder, len(params), shape)
+    params.append(p)
+    # secondparams.append(xb.Op.tuple([p]))
```
yeounoh, line 33: Remove commented line

yeounoh (changes requested):
Left some suggestions

torch_xla/csrc/lowering_context.h
ManfeiBai: Thanks, make sense, we used to write a duplicated BuildXla() to avoid affect original `BuildXla()` performance, and tested locally&simply: modify `BuildXla` has no appear affect on performance, merged the logic to the original BuildXla()

torch_xla/experimental/fori_loop.py
ManfeiBai: Thanks, updated

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1027,7 +1033,9 @@ void BuildLoweringContextSubmodule(py::module* m) {
       .def("hlo_json", &PyLoweringContext::GetHloJsonText)
       .def("parameter_id_tensor_mapping",
            &PyLoweringContext::GetParameterIdTensorMapping)
-      .def("tensor_parameter_id", &PyLoweringContext::GetTensorParameterId);
+      .def("tensor_parameter_id", &PyLoweringContext::GetTensorParameterId)
+      .def("set_name_string", &PyLoweringContext::SetNameString)
```
yeounoh, line 1037: Good, thanks!

torch_xla/csrc/lowering_context.cpp
```diff
@@ -154,9 +154,19 @@ void LoweringContext::SetResult(size_t index, xla::XlaOp op) {
 
 xla::StatusOr<xla::XlaComputation> LoweringContext::BuildXla() {
   xla::StatusOr<xla::XlaComputation> xla;
-  if (!root_tuple_.empty()) {
+  if (!root_tuple_.empty() & (root_tuple_.size() > 1)) {
     xla::XlaOp root = xla::Tuple(builder(), root_tuple_);
     xla = builder()->Build(root);
+  } else if (!root_tuple_.empty() & (root_tuple_.size() == 1)) {
```
yeounoh, line 160: I think we should condition on `get_name_string()`. Add this check at the top and build for while loop `if get_name_string() == "condctx" or get_name_string() == "bodyctx"; otherwise, you can keep the original build logic.

Have your logic for while loop build in a separate private method, and call it if ``if get_name_string() == "condctx" or get_name_string() == "bodyctx"` is true. 

So you can keep `BuildXla()` simple.

torch_xla/csrc/lowering_context.h
```diff
@@ -68,7 +72,8 @@ class LoweringContext : public torch::lazy::LoweringContext {
   xla::XlaOp GetOutputOp(const torch::lazy::Output& output);
 
   // Build the XLA computation capturing all the operations created with the
-  // embedded XLA builder (returned by the builder() API).
+  // embedded XLA builder (returned by the builder() API) with check whether
```
yeounoh, line 75: Think we can keep the original comment. Add the while specific comment on the new private method, as described above.

torch_xla/csrc/lowering_context.cpp
ManfeiBai: Thanks, make sense and updated in the newest commit, due to the `logic for while loop` is one simple line code, we run it directly without warping it in a separate private method

torch_xla/csrc/lowering_context.h
ManfeiBai: Thanks, updated

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -12,25 +12,27 @@
...
 
 class WhileLoopTest(unittest.TestCase):
 
   def test_while_loop_tpu(self):
 
+    device = xm.xla_device()
+
     def cond_fn(x):
-      return x.sum() <= 10
+      ten = torch.ones(1, dtype=torch.int32, device=device)
```
miladm, line 27: cosmetic nit: `ten` can be interpreted as number 10 - suggesting a better name plz

miladm (approved):
thanks - wondering if it helps to have both `add` and `sub` tests?

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: Thanks, updated

ManfeiBai:
> add

Thanks for the suggestion, add would be our next step in another PR, we split add and sub due to add test failed at dynamo: https://gist.github.com/ManfeiBai/9e382014c0817899ad8dbf7f1d33089d

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -901,6 +901,22 @@ class PyLoweringContext {
       lowering_ctx.AddResult(root);
     }
     computation = ConsumeValue(lowering_ctx.BuildXla());
+
+    // warp inputs of cond/body_computation
```
yeounoh, line 905: nit. "wrap"

yeounoh (approved):
LGTM

torch_xla/csrc/init_python_bindings.cpp
ManfeiBai: Thanks, updated

ManfeiBai:
add documents

PR #6605, opened 2024-02-23
Automatically detect CUDA if it isn't set and default to upstream torch.cuda number of devices

changm:


changm:
I manually verified that this worked on a GPU VM.

torch_xla/runtime.py
```diff
@@ -44,10 +45,15 @@ def _maybe_select_default_device():
   if torch_xla._found_libtpu and tpu.num_available_chips() > 0:
     logging.warning('libtpu.so and TPU device found. Setting PJRT_DEVICE=TPU.')
     os.environ[xenv.PJRT_DEVICE] = 'TPU'
-  # TODO(wcromar): Detect GPU device
   elif xu.getenv_as(xenv.GPU_NUM_DEVICES, int, 0) > 0:
     logging.warning('GPU_NUM_DEVICES is set. Setting PJRT_DEVICE=CUDA')
     os.environ[xenv.PJRT_DEVICE] = 'CUDA'
+  elif torch.cuda.is_available():
```
will-cromar, line 51: Out of curiosity, does this work if you use `torch` compiled without CUDA? I think this is what our CI uses.

will-cromar (approved):
Nice! LGTM as long as test works in our CI

torch_xla/runtime.py
changm: Not sure, is that actually a viable option? As in the current XLA docs assume that you compiled PyTorch with CUDA:

https://github.com/pytorch/xla/blob/master/docs/gpu.md#develop-pytorchxla-on-a-gpu-instance-build-pytorchxla-from-source-with-gpu-support

Should CI/CD be updated to compile with CUDA?

JackCaoG:
I think @ysiraichi updated our CI to compile pytorch with CUDA as well.

torch_xla/runtime.py
```diff
@@ -44,10 +45,15 @@ def _maybe_select_default_device():
...
+    logging.warning(
+        'Found CUDA without GPU_NUM_DEVICES. Defaulting to PJRT_DEVICE=CUDA with GPU_NUM_DEVICES=1.'
+    )
+    os.environ[xenv.PJRT_DEVICE] = 'CUDA'
+    os.environ[xenv.GPU_NUM_DEVICES] = '1'
```
vanbasten23, line 56: how do you know GPU_NUM_DEVICES is 1?

torch_xla/runtime.py
vanbasten23, line 51: What torch.cuda.is_available() does is that it checks both if there is underlying CUDA device and if torch is compiled with USE_CUDA enabled.

Ideally I was thinking we should the same: check the hardware and if torch_xla is compiled with XLA_CUDA enabled. `torch.cuda.is_available()` checks the former for you but I didn't know how to check the latter.

vanbasten23: btw, do you know how JAX does it?

changm: I briefly looked at JAX. From what I can tell, for C++ classes, they have `#ifdef JAX_GPU_CUDA`. For Python, it looks like they try to find the CUDA plugin - https://github.com/google/jax/blob/516b75dc24d8e974d40fbc34e0945ad4aaeae474/jax_plugins/cuda/__init__.py#L30

Do we need to figure all that out for this PR?

torch_xla/runtime.py
changm: Good point, I didn't :). I just defaulted to 1 naively. Changed to default to the number of devices torch detected.

changm:
Was failing CI/CD because the test harness explicitly sets PJRT_DEVICE. However, now that it's set, and this test only checks when the environment variable isn't set, the test doesn't actually run.

Should I update CI/CD to not set GPU devices by default in a follow up PR? @will-cromar for insight? Thanks!

test/test_gpu_device_detection.py
```diff
@@ -0,0 +1,34 @@
...
+import torch
+import torch.cuda
+import torch_xla
+import torch_xla.core.xla_env_vars as xenv
+import torch_xla.core.xla_model as xm
+import torch_xla.runtime as xr
+
+
+@unittest.skipIf(
+    os.getenv(xenv.PJRT_DEVICE) != None or not torch.cuda.is_available(),
```
vanbasten23, line 13: Wondering if you need to check `os.getenv(xenv.PJRT_DEVICE) != None` since you are doing
```
    os.unsetenv(xenv.PJRT_DEVICE)
    os.unsetenv(xenv.GPU_NUM_DEVICES)
```
As long as it's gpu ci, regardless if PJRT_DEVICE is set, you can run this test, right?

torch_xla/runtime.py
vanbasten23: Great finding! Later, after PyTorch/XLA start using GPU plugin (cc @will-cromar), we can use this way to check if torch_xla is compiled with cuda enabled and add a check at this place. Feel free to do in a follow-up PR.

vanbasten23 (approved):

will-cromar:
> Should I update CI/CD to not set GPU devices by default in a follow up PR? @will-cromar for insight? Thanks!

When `PJRT_DEVICE=CUDA` is set, we either use GPU or crash if it's not available for some reason. For the vast majority of our tests, this is the correct behavior. We don't want them defaulting to CPU without us noticing for example.

In you test, feel free to just unset `PJRT_DEVICE` in the script.

torch_xla/runtime.py
changm: Sounds good, thank you!

test/test_gpu_device_detection.py
changm: No it doesn't run. The problem is setUp() runs after, so it would unset xenv.PJRT_DEVICE when it's explicitly set on CPU CI/CD. I think that CI still has CUDA, so then it runs the GPU tests. In a follow up, I can remove the PJRT_DEVICE=CUDA from CI/CD so that the tests run.

PR #6607, opened 2024-02-24
Add topological sort after composite grouping pass

lsy323:
When the composite has multiple output, we the program order of the MLIR module may not be topological sorted after the composite grouping pass. It happens when the producers and the consumers of the composite overlap in program order. Because the composite is inserted after the first composite output, there will be composite input lies after the composite, violating the topological order

e.g.
```
%input1
%input2
%output1
%input3
%output2
```
after composite insertion
```
%input1
%input2
%output1
composite call (%input1, %input2, %input3)
%input3
%output2
```

### DIFF
--- a/torch_xla/csrc/runtime/stablehlo_composite_helper.cc
+++ b/torch_xla/csrc/runtime/stablehlo_composite_helper.cc
@@ -9,6 +9,7 @@
 #include "absl/strings/str_cat.h"
 #include "mlir/IR/IRMapping.h"
 #include "mlir/Support/LogicalResult.h"
+#include "mlir/Transforms/TopologicalSortUtils.h"
 #include "single_include/nlohmann/json.hpp"
 #include "stablehlo/dialect/StablehloOps.h"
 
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/runtime/stablehlo_composite_helper.cc
```diff
@@ -277,6 +278,11 @@ class BuildStableHLOCompositePass : public mlir::OperationPass<mlir::ModuleOp> {
       }
     }
 
+    if (!mlir::sortTopologically(composite_op->getBlock())) {
+      composite_op->emitError()
+          << "The graph is not acyclic after the composite creation pass.";
```
chunnienc, line 283: The graph is not acyclic after BuildStableHLOCompositePass

chunnienc (approved):
The topological looks great, thanks!

sdasgup3 (approved):

torch_xla/csrc/runtime/stablehlo_composite_helper.cc
lsy323: Yeah, just would like to catch the error if `mlir::sortTopologically` failed.

chunnienc: I mean changing the message from "after the composite creation pass" to "after BuildStableHLOCompositePass" to be more precise lol.

lsy323: sg, updated

PR #6612, opened 2024-02-25
[benchmarks] Set autocast kwargs only if AMP.

ysiraichi:
This PR sets the `dtype` kwarg only if AMP is supposed to be used. Previously, leaving the `dtype` kwarg made it so `nullcontext` complained about it. This would crash every inference benchmark.

cc @miladm

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -364,14 +364,14 @@ class TorchBenchModel(BenchmarkModel):
   def _get_autocast_with_kwargs(self):
     kwargs = {}
 
-    # Set the default data-type based on the accelerator.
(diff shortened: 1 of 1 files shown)


frgossen:
Can you give a bit more detail on why this is needed?

ysiraichi:
Otherwise, we are calling `nullcontext(dtype=...)`, here:

https://github.com/pytorch/xla/blob/3e5ca3c595a4aeef568a32699fdb552ee223c046/benchmarks/benchmark_model.py#L155-L159

ysiraichi:
As far as I understand it, for AMP case, `self.autocast` will be either `torch.amp.autocast` or `torch.cuda.autocast`. Am I missing something?

PR #6615, opened 2024-02-26
Support building StableHLO composite with different attr value types

chunnienc:
Torch dynamo does not support tracking Python dict with values in different types. So the mark_tensor op and StableHLOCompositeBuilder would throw error when there are multiple attr values in different types:
```
torch._dynamo.exc.TorchRuntimeError: Failed running call_function xla.mark_tensor(*(FakeTensor(..., size=(2, 960, 16, 16)),), **{'name': 'torch.nn.functional.interpolate', 'pos': 0, 'id': '554aaacbfd8e4847b7023162ff460da5', 'is_input': False, 'attr': {'scale_factor': 2.0, 'mode': 'nearest'}}):
Tracer cannot infer type of {'scale_factor': 2.0, 'mode': 'nearest'}
:Dictionary inputs to traced functions must have consistent type. Found float and str`
```

This fix turn the attr dictionary into the flatten tuple for dynamo export and reconstruct the dictionary before serializing to JSON. We will need a different serialization if we want to support attr value in Python dict in the future, but it won't have any impacts to users or developers using this HLFB API.

internal ref: https://b.corp.google.com/issues/325677546


torch_xla/experimental/xla_marker.py
```diff
@@ -1,12 +1,13 @@
 import dataclasses
 import json
 from dataclasses import dataclass
-from typing import Dict
+from typing import Dict, Union
```
lsy323, line 4: nit: looks like Union is not used

torch_xla/experimental/xla_marker.py
```diff
@@ -30,7 +31,7 @@ def default(self, obj):
     return super().default(obj)
 
 
-def _assert_valid_composite_attr(attr):
+def assert_valid_composite_attr(attr):
```
lsy323, line 34: nit: Can we keep the '_' since it's internal to this module?

test/stablehlo/test_mark_pattern.py
```diff
@@ -11,6 +11,7 @@
 from torch_xla import stablehlo
 from torch_xla.experimental.mark_pattern_utils import StableHLOCompositeBuilder
 from utils import has_tf_package
+from torch_xla.experimental import xla_marker
```
lsy323, line 14: nit: sort the import

torch_xla/experimental/xla_marker.py
```diff
@@ -1,12 +1,13 @@
...
 import json
 from dataclasses import dataclass
-from typing import Dict
+from typing import Dict, Union
 
 import torch
 from torch.library import impl
 import torch_xla
 from torch_xla.core.xla_model import XLA_LIB
+import torch._dynamo as torchdynamo
```
lsy323, line 10: nit: sort the import

lsy323 (approved):
@chunnienc Thanks! lgtm, left a few nits

PR #6616, opened 2024-02-26
Add api to buffer donation

JackCaoG:


setup.py
```diff
@@ -88,7 +88,7 @@ def get_git_head_sha(base_dir):
 
 
 def get_build_version(xla_git_sha):
-  version = os.getenv('TORCH_XLA_VERSION', '2.2.0')
+  version = os.getenv('TORCH_XLA_VERSION', '2.3.0')
```
will-cromar, line 91: nit: this is correct, but please separate it into a new PR.

test/dynamo/test_dynamo.py
```diff
@@ -286,7 +286,7 @@ def fn_fallback(t):
     xla_dynamo_res = dynamo_fn(t_xla)
     self.assertTrue(torch.allclose(cpu_res, xla_dynamo_res.cpu()))
     self.assertEqual(met.metric_data('CompileTime')[0], 3)
-    self.assertEqual(met.metric_data('ExecuteTime')[0], 9)
+    self.assertEqual(met.metric_data('ExecuteTime')[0], 7)
```
will-cromar, line 289: what does `7` represent?

torch_xla/core/dynamo_bridge.py
```diff
@@ -28,6 +28,25 @@
 ptxla_debug = int(os.environ.get('PT_XLA_DEBUG', '0')) == 1
 
 
+class AliasWithBufferDonorContext(object):
+
+  def __init__(self, should_alias: bool):
```
will-cromar, line 33: Can you make `should_alias` default to `True`? I think that's going to be the case most of the time when you actually invoke this context manager.

torch_xla/core/dynamo_bridge.py
```diff
@@ -28,6 +28,25 @@
...
+    self.should_alias = should_alias
+
+  def __enter__(self):
```
will-cromar, line 36: I strongly recommend using `contextlib.contextmanager` for trivial contexts like this: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

torch_xla/core/dynamo_bridge.py
```diff
@@ -28,6 +28,25 @@
...
+    self.env_inited = 'XLA_SHOULD_ALIAS_WITH_BUFFER_DONOR' in os.environ
+    if self.env_inited:
+      self.env_saved = os.environ['XLA_SHOULD_ALIAS_WITH_BUFFER_DONOR']
+    os.environ[
```
will-cromar, line 40: We really need to break that habit of communicating between Python/C++ with env vars. Setting a variable in C++ doesn't update Python's `os.environ` for example.

Can you add a pybind to update an internal variable instead?

JackCaoG:
oh.. I need to rebase this pr haha. My old pr for aliasing in https://github.com/pytorch/xla/pull/6587. I will adjust your comment there.

PR #6623, opened 2024-02-27
Scrub use of explicit PJRT_DEVICE=CUDA for Python CI/CD and rely on autodetection

changm:
Can't do it for CPP tests since CPP uses a different PJRT C++ logic to directly read the PJRT_DEVICE to initialize.

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -137,7 +137,7 @@ function run_torch_xla_python_tests() {
 
         # single-host-multi-process
         num_devices=$(nvidia-smi --list-gpus | wc -l)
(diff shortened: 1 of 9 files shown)


will-cromar (changes requested):
We should leave `PJRT_DEVICE=CUDA` set for almost all of our tests. We'd rather our CI crash if there was an issue with the GPU client vs silently falling back to CPU. See my comment here: https://github.com/pytorch/xla/pull/6605#issuecomment-1965464819

changm:
Ahh ack ok sorry I misread that versus a different comment. I'll close this then. Thanks!

PR #6625, opened 2024-02-27
Make XLAShardedTensor use default __torch_function__

JackCaoG:
Companion pr for https://github.com/pytorch/pytorch/pull/120632#issuecomment-1967489393, through I have to admit I don't fully understand why XLA test will fail without this..


albanD:
Thanks for the quick PR!
This is completely fine to land this before the PR in PyTorch as this is the default behavior and is a no-op right now.
Just let me know when this propagates and I can rebase the PyTorch side PR to include this change.

albanD:
> I have to admit I don't fully understand why XLA test will fail without this..

Me neither.
The main change I am adding in my PR is disabling this default handler. The main thing it does is automatically wrap anything your torch_dispatch handler returns into your subclass. Meaning that if your torch_dispatch handler returns a plain Tensor, then the torch_function handler will wrap it into your subclass. This is usually not the expected behavior though, hence the fact that we're changing that.

But from a quick look at what you do, it shouldn't change anything here...

JackCaoG:
Hmm dynamo started to complain after adding this `torch_dispatch` thingy..

albanD:
Ho I see the same issue happening on Int16Tensor in the other PR. Let me look into that.

torch_xla/distributed/spmd/xla_sharded_tensor.py
```diff
@@ -108,6 +108,9 @@ def __new__(cls, elem: torch.Tensor, *args, **kwargs):
     r.global_tensor = elem.detach() if r.requires_grad else elem
     return r
 
+  def __torch_function__(cls, func, types, args=(), kwargs=None):
```
albanD, line 111:
```suggestion
  @classmethod
  def __torch_function__(cls, func, types, args=(), kwargs=None):
```

albanD (commented):
Well, this seems to fix it from this basic experience haha https://github.com/pytorch/pytorch/issues/120799

Will investigate the Dynamo failure as well but this should make this PR green!

torch_xla/distributed/spmd/xla_sharded_tensor.py
JackCaoG: ok let me update and give it another try

JackCaoG:
@albanD ok great, seems like this CI will fix the failure. Let me remove the torch pin and merge it to unblock you.

JackCaoG:
@yeounoh can you take a look at this pr and merge it to unblock Alban?

yeounoh (approved):
LGTM thanks @albanD

PR #6632, opened 2024-02-28
[benchmarks] Small fixes for benchmarking script.

ysiraichi:
This PR implements small fixes introduced by past PRs:

- Removes the incorrect `self` argument from the `find_near_file` function that was left-over in the last PR that moved it #6542  
- Replace the variable `model_name` by `TorchBenchModel` attribute of same name
- Select user specified batch size by default

@miladm

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -310,18 +310,14 @@ class TorchBenchModel(BenchmarkModel):
     if cant_change_batch_size:
       self.benchmark_experiment.batch_size = None
 
-    if self.benchmark_experiment.batch_size is not None:
-      batch_size = self.benchmark_experiment.batch_size
-    elif self.is_training() and self.model_name in self.batch_size["training"]:
-      batch_size = self.batch_size["training"][self.model_name]
-    elif self.is_inference(
-    ) and self.model_name in self.batch_size["inference"]:
-      batch_size = self.batch_size["inference"][self.model_name]
-    else:
-      # This should work, since TorchBench relies on class variables:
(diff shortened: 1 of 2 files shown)


cota (approved):

zpcore:
Thanks, I think you forgot to import `exists`.

zpcore:
There are also other issues I see in recent check-ins:
https://github.com/pytorch/xla/blob/0a2e5b5a2d88bddf758f8ec0ef99ffd31c0ab97e/benchmarks/torchbench_model.py#L309, need `self.model_name`.
For the `batch_size`, I can't find `DEFAULT_EVAL_BSIZE` and `DEFAULT_TRAIN_BSIZE` to be use anywhere.

vanbasten23 (approved):

frgossen (approved):

cota (approved):

zpcore (approved):
Thanks for the fixing!

zpcore:
Hi @ysiraichi , do you know where the default batch size from torchbench was used in the code?

ysiraichi:
Do you mean where do we set the default batch size? As far as I understand, torchbench does it automatically when we instantiate a `Model`. Specifically, here:

https://github.com/pytorch/xla/blob/4cde625c0c6608fed05a627abc4feb3a6c1ac9f3/benchmarks/torchbench_model.py#L329-L333

Which ends up in [this function](https://github.com/pytorch/benchmark/blob/55f8e5a977349adbbd4918bf7d25a274405e7e32/torchbenchmark/util/model.py#L189-L210).

zpcore:
Got it, thanks! Now the batch_size setting looks clear to me.

PR #6635, opened 2024-02-28
Update OpenXLA pin to enable libtpu & pjrt plugin for 02282024

yeounoh:
This enables PJRT plugin 0.45, and remove quantization openxla patch in our repo.

Tested locally with ResNet.


qihqi:
afaik we still need quant_dequant_converter @lsy323

PR #6642, opened 2024-02-28
Skip the execution if all Pending IRs are device data

JackCaoG:
Today if we have a bunch of DeviceData IR and called `xm.mark_step`, we will execute a graph looks like
```
    # ENTRY %IrToHlo.4 (p0.1: f32[4,2,2], p1.2: f32[4,2,2]) -> (f32[4,2,2], f32[4,2,2]) {
    # %p0.1 = f32[4,2,2]{2,1,0} parameter(0)
    # %p1.2 = f32[4,2,2]{2,1,0} parameter(1)
    # ROOT %tuple.3 = (f32[4,2,2]{2,1,0}, f32[4,2,2]{2,1,0}) tuple(f32[4,2,2]{2,1,0} %p0.1, f32[4,2,2]{2,1,0} %p1.2)
    # }
```

This is pretty much a no-op graph that will generate new buffer(might or might not be aliased with input) with the same value. Execution of such graph will lead to runtime overhead as well as device execution overhead. For training it is usually not a big deal but for inference this can be very annoying, especially in dynamo bridge we sometimes needs to `mark_step` to sync inputs. My change will force the tensor to drop reference to the DeviceData IR and point to the `XLAData` direcly to save the execution of that tensor.

My change will incur a problem if 2 tensors share the same buffer, in the old design both tensor will get new buffers after execution. With my new change no execution happens and they all point to the same PJRT buffer. However if this input buffer is being aliased with the output(in place update for one tensor), another tensor will point to a deleted buffer. To address this issue I make sure to always allocate new output buffer for cloned tensor.

### DIFF
--- a/test/dynamo/test_dynamo_aliasing.py
+++ b/test/dynamo/test_dynamo_aliasing.py
@@ -42,7 +42,7 @@ class TestDynamoBufferDonationAliasingWithCustomOp(unittest.TestCase):
   def test_manual_buffer_donation(self):
     device = xm.xla_device()
     input = torch.randn(5, 5).to(device)
-    input_cloned = torch.clone(input)
+    input_cloned = input.cpu().to(device)
     dummy_inplace_mul_compiled = torch.compile(
         self.dummy_inplace_mul, backend='openxla')
 
@@ -57,7 +57,7 @@ class TestDynamoBufferDonationAliasingWithCustomOp(unittest.TestCase):
   def test_manual_buffer_donation_for_non_inplce_op(self):
     device = xm.xla_device()
     input = torch.randn(5, 5).to(device)
-    input_cloned = torch.clone(input)
+    input_cloned = input.cpu().to(device)
     dummy_mul_compiled = torch.compile(self.dummy_mul, backend='openxla')
 
     met.clear_all()
@@ -83,7 +83,7 @@ class TestDynamoBufferDonationAliasingWithCustomOp(unittest.TestCase):
 
     device = xm.xla_device()
     input = torch.randn(5, 5).to(device)
-    input_cloned = torch.clone(input)
+    input_cloned = input.cpu().to(device)
     dummy_inplace_add_compiled = torch.compile(dummy_inplace, backend='openxla')
     xm.mark_step()
     met.clear_all()
@@ -111,7 +111,7 @@ class TestDynamoBufferDonationAliasing(unittest.TestCase):
   def test_manual_buffer_donation(self):
     device = xm.xla_device()
     input = torch.randn(5, 5).to(device)
-    input_cloned = torch.clone(input)
+    input_cloned = input.cpu().to(device)
     dummy_inplace_add_compiled = torch.compile(
         self.dummy_inplace_add, backend='openxla')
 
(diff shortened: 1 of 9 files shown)


frgossen (approved):
Nice, ty!

JackCaoG:
lol it breaks `test_shard_hashing`, I think this was where I gave up last time. Let me took another look why that's the case.

JackCaoG:
`test/test_input_output_aliases.py` still fails, I know why it failed but what's weird(input shape and output shape has different layout so it won't alias) is that it also failed even if I build from the master... Not sure what happened.

JackCaoG:
I also need to add more test about `torch.clone` which will make a new DeviceData that share the same XLAData, need to make sure aliasing won't cause issue here.

JackCaoG:
More test failures.. will need to look into this.

JackCaoG:
Finally test is green lol. I don't want to merge this to 2.3 since it is too risky, I will leave this to nightly for now.

test/dynamo/test_dynamo_aliasing.py
```diff
@@ -153,7 +153,7 @@ def dummy_inplace(input):
 
     device = xm.xla_device()
     input = torch.randn(5, 5).to(device)
-    input_cloned = torch.clone(input)
```
alanwaketan, line 156: So, `clone` cannot be used in pt-xla?

test/spmd/test_xla_sharding.py
```diff
@@ -630,8 +635,8 @@ def test_shard_hashing(self):
     self.assertTrue(torch.allclose(xt1 + 0, xt2 + 0))
 
     # Check that hashes are different for the sharded and non-sharded tensors
-    hash1 = torch_xla._XLAC._get_graph_hash([xt1])
-    hash2 = torch_xla._XLAC._get_graph_hash([xt2])
+    hash1 = torch_xla._XLAC._get_graph_hash([xt1 + 0])
```
alanwaketan, line 638: Why?

test/test_metrics.py
```diff
@@ -72,6 +72,7 @@ def test_short_metrics_report_custom_list(self):
     xla_device = xm.xla_device()
     t1 = torch.tensor(100, device=xla_device)
     t2 = t1 * 2
+    t1 += 2
```
alanwaketan, line 75: why?

alanwaketan (commented):
What are the ways to create device data IR?

test/dynamo/test_dynamo_aliasing.py
JackCaoG: clone can be used, but in this test we manually donated `input`'s buffer, which is shared by `input_cloned`. 

actually I made this change prior to my fix, I think cloned should work even with manual buffer donation.. through it is generally a bad idea to donate the buffer while also clone the same tensor..

test/spmd/test_xla_sharding.py
JackCaoG: `xt1` is already a device data, `_get_graph_hash ` will call `CollectSyncTensors` and it will return an empty list(since it think `xt1` don't need to be synced). This test should not use `_get_graph_hash ` as it is an internal API but I am a bit lazy to rewrite the test. 

Forcing it to be `xt1 + 0` will let `CollectSyncTensors` use the interedmate tensor as root and calculate the hash.

test/test_metrics.py
JackCaoG: `t1` will be ignored by the `mark_step`(as an output) after my change since it is already a device data. I think one of the metrics checks was checking something related to t1 so I quickly brought it back to the graph as a output.

alanwaketan (approved):
LGTM. It looks good to me.

JackCaoG:
Thanks @alanwaketan I am going to wait for the branch cut to happen and then merge this change.

PR #6644, opened 2024-02-29
Automatically move CUDA non XLA Tensors to XLA Device and back to CUDA device

changm:
Currently only works for inference. The assumptions don't hold for training with Autograd yet.

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -113,6 +113,9 @@ class DynamoInferenceBasicTest(unittest.TestCase):
     original_device = x.device
     eager_result = self.fn_simple(x, y)
 
+    # Since all tests run in the same process, have to reset the metrics report.
+    met.clear_all()
+
     fn_simple_dynamo = torch.compile(self.fn_simple, backend="openxla")
     res_xla_dynamo = fn_simple_dynamo(x, y)
     self.assertIn('xla::add', met.counter_names())
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -83,7 +83,7 @@ class DynamoInferenceBasicTest(unittest.TestCase):
     res_xla_dynamo = fn_simple_dynamo(xla_x, xla_y)
     self.assertIn('xla::add', met.counter_names())
     self.assertTrue(torch.allclose(res_cpu, res_xla_dynamo.cpu()))
-    # verifiy that tracing is skipped in following runs
+    # verify that tracing is skipped in following runs
     met.clear_counters()
     res_xla_dynamo_2 = fn_simple_dynamo(xla_x, xla_y)
     self.assertNotIn('xla::add', met.counter_names())
@@ -102,6 +102,35 @@ class DynamoInferenceBasicTest(unittest.TestCase):
                      torch_xla._XLAC._get_xla_tensor_debug_info(xla_xy))
     self.assertNotIn('XLAData: None',
                      torch_xla._XLAC._get_xla_tensor_debug_info(xla_y3))
+
+  # Tests that the dynamo bridge automatically moves tensors to XLA device,
+  # then back to the original device.
+  @unittest.skipIf(xr.device_type() != "CUDA",
+                   f"GPU tests should only run on GPU devices.")
+  def test_simple_model_automoves_tensors(self):
+    x = torch.tensor(100.0).to(device="cuda")
+    y = torch.tensor(200.0).to(device="cuda")
+    original_device = x.device
+    eager_result = self.fn_simple(x, y)
+
+    fn_simple_dynamo = torch.compile(self.fn_simple, backend="openxla")
+    res_xla_dynamo = fn_simple_dynamo(x, y)
+    self.assertIn('xla::add', met.counter_names())
+    self.assertTrue(res_xla_dynamo.device == original_device)
+    self.assertTrue(torch.allclose(eager_result, res_xla_dynamo))
+
+    # verify that tracing is skipped in following runs
+    met.clear_counters()
+    res_xla_dynamo_reused = fn_simple_dynamo(x, y)
+    self.assertNotIn('xla::add', met.counter_names())
+    self.assertTrue(res_xla_dynamo_reused.device == original_device)
+    self.assertTrue(torch.allclose(eager_result, res_xla_dynamo_reused))
+
+    # verify that dynamo can handle different inputs
+    res_xla_dynamo_different = fn_simple_dynamo(x + y, y * 3)
+    res_cpu_3 = self.fn_simple(x + y, y * 3)
+    self.assertTrue(res_xla_dynamo_different.device == original_device)
+    self.assertTrue(torch.allclose(res_cpu_3, res_xla_dynamo_different))
 
   def test_fn_without_input(self):
 
@@ -526,7 +555,7 @@ class DynamoTrainingOptimizerTest(unittest.TestCase):
         met.metric_data('RunCachedGraphOutputData')[0], sample_count * 3)
 
 
-class DynamErrorMessageTest(unittest.TestCase):
+class DynamoErrorMessageTest(unittest.TestCase):
 
   def test_mixed_cpu_tensor(self):
     device = xm.xla_device()
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -111,6 +111,65 @@ def get_fallback_ops():
   return fallback_ops
 
 
+# Checks that all input args that are tensors are on the same device.
+def _get_input_arg_device(input_args: tuple) -> torch.device:
+  device = None
+  for arg in input_args:
+    if not isinstance(arg, torch.Tensor):
+      continue
+
+    if device is None:
+      device = arg.device
+    else:
+      assert arg.device == device, "Not all args are on the same device."
+
+    return device
+
+
+# Returns True if all the input args are on a CUDA device.
+def _args_on_cuda(input_args: tuple) -> bool:
+  input_device: torch.device = _get_input_arg_device(input_args)
+  if input_device is None:
+    return False
+
+  return input_device.type == "cuda"
+
+
+# Given an input list, moves the tensors to the given target_device.
+# The output order will be the same as the input. Non tensors will also still
+# be in the list.
+def _maybe_move_tensors_to_device(tensors: tuple,
+                                  target_device: torch.device) -> tuple:
+  if not torch.cuda.is_available():
+    return tensors
+
+  moved_tensors = []
+  cpu_device: torch.device = torch.device("cpu")
+
+  for tensor in tensors:
+    if not isinstance(tensor, torch.Tensor):
+      moved_tensors.append(tensor)
+      continue
+
+    if tensor.device == target_device:
+      moved_tensors.append(tensor)
+      continue
+
+    if dynamo_debug:
+      print("Moving Tensor {} to device {}".format(tensor, target_device))
+
+    # Have to move to CPU before moving it to target device.
+    moved_tensor = tensor.to(cpu_device)
+    moved_tensor = moved_tensor.to(target_device)
+
+    # Explicitly have to copy requires_grad attribute because it's dropped
+    # with torch.to(..)
+    moved_tensor.requires_grad = tensor.requires_grad
+    moved_tensors.append(moved_tensor)
+
+  return tuple(moved_tensors)
+
+
 class Deduper:
 
   def __init__(self):
@@ -375,7 +434,7 @@ def extract_internal(xla_model: torch.fx.GraphModule):
   skip_checking_input_sharding_threashold = xu.getenv_as(
       'XLA_DYNAMO_INPUT_SHARDING_CHECK_THRESHOLD', int, 5)
 
-  def optimized_mod(*args):
+  def optimized_mod(*args: tuple):
     nonlocal xla_model
     nonlocal xla_args_sharding_spec
     nonlocal args_and_out
@@ -386,6 +445,12 @@ def extract_internal(xla_model: torch.fx.GraphModule):
     nonlocal dumb_return_handler
     nonlocal xla_args_need_update
     nonlocal skip_checking_input_sharding_threashold
+
+    original_device: torch.device = _get_input_arg_device(args)
+    is_cuda_args: bool = _args_on_cuda(args)
+
+    if is_cuda_args:
+      args = _maybe_move_tensors_to_device(args, xm.xla_device())
 
     # mark_step needs to be blocking since we want to access args's XLADatas
     # and they can't be placeholder.
@@ -437,6 +502,9 @@ def extract_internal(xla_model: torch.fx.GraphModule):
     result = res[len(xla_args_need_update):]
 
     none_remover.add_nones(result)
+    if is_cuda_args:
+      result = _maybe_move_tensors_to_device(tuple(result), original_device)
+
     if len(result) == 1:
       return result[0]
     else:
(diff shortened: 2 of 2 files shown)


test/dynamo/test_dynamo.py
```diff
@@ -93,6 +93,49 @@ def test_simple_model(self):
...
+    input = torch.randn(4, 3, 224, 224)
+    resnet18 = torchvision.models.resnet18()
+    resnet18.eval()
+    dynamo_resnet18_cpu = torch.compile(resnet18, backend='openxla')
+
+    # input and model weight on cpu
+    with warnings.catch_warnings(record=True) as w:
+      res = dynamo_resnet18_cpu(input)
+      # there should be 18 paramters + 1 input all moved to XLA Device.
+      self.assertTrue(len(w) == 0)
```
vanbasten23, line 133: I wonder what this is testing.

test/dynamo/test_dynamo.py
```diff
@@ -93,6 +93,49 @@ def test_simple_model(self):
     res_cpu_3 = self.fn_simple(x + y, y * 3)
     self.assertTrue(torch.allclose(res_cpu_3, res_xla_dynamo_3.cpu()))
 
+  # Tests that the dynamo bridge automatically moves tensors to XLA device
+  def test_simple_model_automoves_tensors(self):
+    xla_device = xm.xla_device()
+    x = torch.tensor(100.0)
+    y = torch.tensor(200.0)
```
vanbasten23, line 100: Based on the proposal, should you move the tensors/models to cuda first? https://screenshot.googleplex.com/7bc4dGUP7Vev7VE

changm: Can you explain a bit more? My understanding is that this should work both for CPU/CUDA anyway, as in we would silently swizzle the pointer underneath and not do a move. However, XLA still needs the tensors on the XLA device. So this PR doesn't do the swizzling yet versus just get this to automatically move tensors to the device. In a follow up PR, I'd try to swizzle the pointers so its a zero copy move. Does that make sense?

test/dynamo/test_dynamo.py
```diff
@@ -93,6 +93,49 @@ def test_simple_model(self):
     res_cpu_3 = self.fn_simple(x + y, y * 3)
     self.assertTrue(torch.allclose(res_cpu_3, res_xla_dynamo_3.cpu()))
 
+  # Tests that the dynamo bridge automatically moves tensors to XLA device
+  def test_simple_model_automoves_tensors(self):
+    xla_device = xm.xla_device()
```
vanbasten23, line 98: You can remove this var as it's not used in the test.

test/dynamo/test_dynamo.py
vanbasten23: I think there're 2 reasons:
- The main scenario we are optimizing is CUDA instead of CPU as most of people uses CUDA and TPU instead of CPU. @JackCaoG 's design is mainly for CUDA.
- When you later deal with the zero copy move, there might differences when you move from CUDA or CPU to the XLA device. Either way, we want to make sure the CUDA->XLA works.

So I feel it's better to make the input tensor on CUDA instead of CPU.

torch_xla/core/dynamo_bridge.py
```diff
@@ -377,7 +419,7 @@ def extract_internal(xla_model: torch.fx.GraphModule):
   skip_checking_input_sharding_threashold = xu.getenv_as(
       'XLA_DYNAMO_INPUT_SHARDING_CHECK_THRESHOLD', int, 5)
 
-  def optimized_mod(*args):
+  def optimized_mod(*input_args):
```
golechwierowicz, line 422: `optimized_mod` is invoked with every inference call for dynamo path. Can you measure before, and after performance overhead on this diff? Wrapping the entire function around `time.perf_counter()` (basically less exhaustive https://github.com/pytorch/xla/pull/6598) would be sufficient.

changm: Sure, is there a specific workload you wanted to measure on?

test/dynamo/test_dynamo.py
changm: Sorry not quite sure I follow. This test case tests that we automatically move the tensor to the XLA device, so we should not be moving the tensors here at all. The logic here tests that the dynamo bridge automatically moves tensors that are NOT on the XLA device to the XLA device, then back after the bridge exits. How would you test that if you pre-moved the tensors to CUDA?

test/dynamo/test_dynamo.py
changm: Done.

torch_xla/core/dynamo_bridge.py
golechwierowicz: Just anything basic, can be resnet50 for example, but the point is to grasp the overhead.

JackCaoG:
Is this an experimental pr or you want to merge this?

torch_xla/core/dynamo_bridge.py
changm: Test PR here to double check my work - https://github.com/changm/pytorch-xla/pull/3

Resnet 50 model in a loop for 100x, run 3x (note, everything done on CPU in a Cloud VM sooo probably noisy). All times in nanoseconds:

With Patch:
254414249.59
256134671.03
254181391.66

Without
255055284.05
255336177.86
256976810.52

Doesn't seem to be much overhead.

test/dynamo/test_dynamo.py
changm: Ahh you're right, good call. The test was failing on CUDA because I also had to move the tensor from CUDA -> CPU -> XLA:GPU and back. You can't immediately just move from CUDA <-> XLA:GPU. Updated the test to put the tensor on a CUDA device and just run on GPU.

changm:
> Is this an experimental pr or you want to merge this?

Ideally we would merge this, or is there a reason not to?

test/dynamo/test_dynamo.py
```diff
@@ -103,6 +103,52 @@ def test_simple_model(self):
     self.assertNotIn('XLAData: None',
                      torch_xla._XLAC._get_xla_tensor_debug_info(xla_y3))
 
+  # Tests that the dynamo bridge automatically moves tensors to XLA device,
+  # then back to the original device.
+  @unittest.skipIf(xr.device_type() != "CUDA",
```
JackCaoG, line 108: why cuda only? In your implementation you didn;t have any guard on only run data moving logic only for cuda. If the intention is to do this auto-move for cuda, then it should throw an error or something if device is no cpu for example.

torch_xla/core/dynamo_bridge.py
```diff
@@ -387,6 +433,9 @@ def optimized_mod(*args):
     nonlocal xla_args_need_update
     nonlocal skip_checking_input_sharding_threashold
 
+    original_device: torch.device = _get_input_arg_device(input_args)
+    args = _maybe_move_tensors_to_device(input_args, xm.xla_device())
```
JackCaoG, line 437: maybe check if `original_device` is cuda? otherwise we don't need to call `_maybe_move_tensors_to_device ` which will loop through all input tensors.

torch_xla/core/dynamo_bridge.py
```diff
@@ -437,6 +486,8 @@ def optimized_mod(*args):
     result = res[len(xla_args_need_update):]
 
     none_remover.add_nones(result)
+    result = _maybe_move_tensors_to_device(result, original_device)
```
JackCaoG, line 489: ditto, only call this if `original_device ` is cuda

test/dynamo/test_dynamo.py
changm: See @vanbasten23 comments where his review said to update the test for CUDA. I think the intention should be to auto-move for any non-XLA device, and not limit it to CUDA. 

I can also replicate this test for a CPU version as well.

torch_xla/core/dynamo_bridge.py
changm: See other comment, I don't think this should be limited to CUDA only unless you see a strong reason to?

Also, see @golechwierowicz comment thread where we measured the overhead which seemed negligible.

torch_xla/core/dynamo_bridge.py
changm: ditto

JackCaoG: the problem of this `_maybe_move_tensors_to_device ` is it might force the execution to be blocking(since you need to move the result to cpu) instead of async. Prior to this change we just return in the bridge while execution happens in the background.

I want to isolate these potenial dangerous behavior(might cause unintended slowdown) so we don't accidentally introduce regressions. I also mentioned in my UX doc that we might need to find a way to convert the async xla tensor to a async cuda tensor to relief this issue.

torch_xla/core/dynamo_bridge.py
golechwierowicz: Hmm, looks like just noise. I'll add some micro-benchmark which just measures overhead to the repo so we will have it standarised. Thanks!

test/dynamo/test_dynamo.py
changm: Updated to move these tensors to CUDA device.

test/dynamo/test_dynamo.py
changm: Done with CUDA only.

torch_xla/core/dynamo_bridge.py
changm: Updated to only do this if we have a cuda tensor.

test/dynamo/test_dynamo.py
```diff
@@ -103,6 +103,35 @@ def test_simple_model(self):
...
+                   f"GPU tests should only run on GPU devices.")
+  def test_simple_model_automoves_tensors(self):
+    x = torch.tensor(100.0).to(device="cuda")
+    y = torch.tensor(200.0).to(device="cuda")
+    original_device = x.device
+    eager_result = self.fn_simple(x, y)
+
+    fn_simple_dynamo = torch.compile(self.fn_simple, backend="openxla")
+    res_xla_dynamo = fn_simple_dynamo(x, y)
+    self.assertIn('xla::add', met.counter_names())
```
JackCaoG, line 118: nit, do a `met.clear_all()` before `res_xla_dynamo = fn_simple_dynamo(x, y)`.

JackCaoG: context is if you run `python test/dynamo/test_dynamo.py` all tests will be run in the same process.

torch_xla/core/dynamo_bridge.py
```diff
@@ -111,6 +111,65 @@ def get_fallback_ops():
...
+def _get_input_arg_device(input_args: tuple) -> torch.device:
+  device = None
+  for arg in input_args:
+    if not isinstance(arg, torch.Tensor):
+      continue
+
+    if device is None:
+      device = arg.device
+    else:
+      assert arg.device == device, "Not all args are on the same device."
```
JackCaoG, line 124: @ysiraichi I remembered you added a pass to dynamo bridge to handle where some of the inputs are on cpu(by graph partiton). Will the change here break that?

torch_xla/core/dynamo_bridge.py
```diff
@@ -387,6 +446,12 @@ def optimized_mod(*args):
     nonlocal xla_args_need_update
     nonlocal skip_checking_input_sharding_threashold
 
+    original_device: torch.device = _get_input_arg_device(args)
+    is_cuda_args: bool = _args_on_cuda(args)
```
JackCaoG, line 450: `_args_on_cuda ` will call `_get_input_arg_device` which is redundant.

torch_xla/core/dynamo_bridge.py
```diff
@@ -111,6 +111,65 @@ def get_fallback_ops():
...
+  return input_device.type == "cuda"
+
+
+# Given an input list, moves the tensors to the given target_device.
+# The output order will be the same as the input. Non tensors will also still
+# be in the list.
+def _maybe_move_tensors_to_device(tensors: tuple,
+                                  target_device: torch.device) -> tuple:
+  if not torch.cuda.is_available():
+    return tensors
```
ysiraichi, line 144: Why do we need this?

torch_xla/core/dynamo_bridge.py
```diff
@@ -111,6 +111,65 @@ def get_fallback_ops():
...
+    if tensor.device == target_device:
+      moved_tensors.append(tensor)
+      continue
+
+    if dynamo_debug:
+      print("Moving Tensor {} to device {}".format(tensor, target_device))
+
+    # Have to move to CPU before moving it to target device.
+    moved_tensor = tensor.to(cpu_device)
+    moved_tensor = moved_tensor.to(target_device)
```
ysiraichi, line 163: I believe you can copy tensors directly to `target_device`.

torch_xla/core/dynamo_bridge.py
ysiraichi: I think this is fine. If I remember correctly `ConstructorMoverPass` only modified tensor constructor nodes (not arguments).

torch_xla/core/dynamo_bridge.py
ysiraichi: This works with CPU, too, doesn't it?

JackCaoG: We have had a couple discussion about whether to support CPU. My concern is that moving tensors between devices will force the execution to be blocking. The main motivation for this pr was to support execute the part that dynamo can't capture on cuda tensor, so I want to limit the scope of this pr to CUDA tensor only for now. We can support CPU in the future as well.

torch_xla/core/dynamo_bridge.py
JackCaoG: Can we move directly from XLA:GPU to cuda?

ysiraichi: If I remember correctly, we can, since: https://github.com/pytorch/xla/pull/6060

changm: I tried this before and upstream pytorch specifically stated it could not and had to be moved to CPU.

torch_xla/core/dynamo_bridge.py
changm: Correct and I'm just being defensive here to ensure someone doesn't accidentally call this and it supports non-CUDA moves.

torch_xla/core/dynamo_bridge.py
ysiraichi: This is odd. I'm on 99c41baefa44e62242d158e2321d75b87b01cffe

```python
>>> a = torch.arange(10, device="cuda")
>>> a.to(xm.xla_device())
tensor([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], device='xla:0')
```

ysiraichi: Oh. Maybe we can't do the other way around.

test/dynamo/test_dynamo.py
changm: Done.

torch_xla/core/dynamo_bridge.py
changm: So is this ok to leave here? Thanks for double checking :)

torch_xla/core/dynamo_bridge.py
changm: I still think it's a little cleaner to do the redundant call, but removed the call here.

JackCaoG (approved):

JackCaoG (approved):

vanbasten23:
Sorry for being late. It's looking good!

Issue #6535, opened 2024-02-14
Tracking `moving TSL to XLA` on PyTorch/XLA


ManfeiBai:
## 🚀 Feature

OpenXLA proposed to move TSL to XLA, and there would be change in PyTorch/XLA to include this moving: change `@tsl` to `@xla/tsl`

## Motivation

OpenXLA proposal: https://github.com/openxla/community/pull/100

## Pitch

- PyTorch/XLA would confirm locally first in a list;
- PyTorch/XLA would have a list of PRs contain per moving change to run CI test before OpenXLA migrate per change
- PyTorch/XLA include the new change by OpenXLA-pin update after the change

cc @miladm @ddunl

ysiraichi:
@ManfeiBai I'm closing this issue, since it looks like this was already merged.
Feel free to re-open this issue if needed.

PR #6655, opened 2024-03-01
[OpenXLA|TSL] Replace tsl::Status to absl::Status in profiler.h/.cc

ManfeiBai:
pre-change =for https://github.com/pytorch/xla/issues/6535

cc @ddunl


will-cromar (approved):
Should we be using `absl::Status` or `xla::Status` going forward? I see a few references to `xla::Status` remaining. I assuming this is just an alias.

ManfeiBai:
> Should we be using `absl::Status` or `xla::Status` going forward? I see a few references to `xla::Status` remaining. I assuming this is just an alias.

nice point, since `xla::Status` looks like an alias of `absl::Status`: [link](https://github.com/openxla/xla/blob/081f529d5bf11380c116f747865b77d2a1b3ed27/xla/status.h#L26C7-L26C18), let's try to use `xla::Status` too

PR #6662, opened 2024-03-04
fix csv output error

zpcore:
Fix command `python result_analyzer.py --output-format=csv` will return error.


vanbasten23 (approved):
Thanks for the fix.

PR #6670, opened 2024-03-05
[DO NOT REVIEW] Install missing package

vanbasten23:


vanbasten23:
Enabled the coverage by setting `collect-coverage: true` and verified that the fix works

yeounoh (approved):
LGTM

Issue #6322, opened 2024-01-18
[RFC] PyTorch/XLA Auto-Sharding API


yeounoh:
# 🚀 Feature & Motivation

PyTorch/XLA recently launched PyTorch/XLA SPMD ([blog](https://pytorch.org/blog/pytorch-xla-spmd/), [docs/spmd.md](https://github.com/pytorch/xla/blob/master/docs/spmd.md)), as a first-step to automate ML workloads parallelization using GSPMD. In contrast to previous PyTorch user experiences with implementing tensor & model parallelism for their ML workloads, PyTorch/XLA SPMD allows the users to provide a handful of “sharding hints” via the PyTorch/XLA SPMD sharding annotation API, and keep the original model implementation as-is. We presented some exciting CloudTPU PyTorch LLaMA2 training results using SPMD at Google Next ([blog](https://pytorch.org/blog/high-performance-llama-2/)). Another highlight of PyTorch/XLA SPMD was that it enables more advanced and hybrid types of parallelism strategies for the PyTorch users, combining data and tensor parallelism, as well as pipelining. While this is all great and we are happy to release the new feature to the PyTorch community, one challenge still remains as to providing the optimal sharding hints. It turns out that the performance largely depends on the quality of sharding hints provided by the user – and it requires a correct and deep understanding of model architectures and much expertise to come up with optimal sharding hints. 

To address this problem, we propose to integrate PyTorch/XLA SPMD with XLA's auto sharding service that allows the XLA compiler to shard and optimize the whole model without any user input. XLA auto-sharding service is based on a published research work, Alpa ([blog](https://blog.research.google/2022/05/alpa-automated-model-parallel-deep.html)). While this sounds like a leap of faith, it is already being tested and showing some promising results on the Google internal workloads (also see our mini-benchmark results).

# API and Usage Example

To enable auto-sharding, simply call `use_spmd` with the `auto=True` flag:
```python
import torch_xla.runtime as xr 

# Enable XLA SPMD execution mode with auto-sharding.
xr.use_spmd(auto=True)

# Write a PyTorch program without any sharding annotations.
# User can still provide sharding hints, but optionally. 
...
```
This should be it, the PyTorch program should be automatically sharded and executed.

There are optional configuration knobs, though:
```python
# The optional auto_sharding config can be passed to control 
# the auto-sharding behavior.
config = {"auto_sharding" : {"partitioner" : "alpa", "keep_user_sharding" : True }}
[...]

PR #6680, opened 2024-03-06
Make use_spmd to force SPMD mode for easier 3rd-party integration.

yeounoh:
It is actually more easier to force SPMD mode than blocking if there is non-SPMD initialized tensors, for the 3rd-party solution integration. For instance, HuggingFace trainer pre-loads embeddings table and the training initialization is done over multiple scripts. Being able to force SPMD allows the users to call `xr.use_spmd()` more freely, given that the earlier they call, the smaller the one-time overhead of replicating non-SPMD backed tensors.


yeounoh:
This need to land for experimental release of the auto-sharding API https://github.com/pytorch/xla/issues/6322

yeounoh:
This works on HuggingFace by inserting `use_spmd` in `run_clm.py` only:
```
def main():
  import torch_xla
  torch_xla.runtime.use_spmd()
  ...
```
cc @alanwaketan @JackCaoG

yeounoh:
It works, but I also need to fix the unit test.

yeounoh:
All CI tests are green, rebased just now. @alanwaketan could you help review/land this?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1147,7 +1147,35 @@ void InitXlaModuleBindings(py::module m) {
...
+    // non-SPMD backed tensors.
+    torch::lazy::BackendDevice backend_device = bridge::GetCurrentDevice();
+    std::vector<XLATensorPtr> xtensors =
+        XLAGraphExecutor::Get()->GetLiveTensors(&backend_device);
+    for (auto xtensor : xtensors) {
+      XlaDeviceType xla_device_type =
+          static_cast<XlaDeviceType>(xtensor->GetDevice().type());
+      if (xla_device_type != XlaDeviceType::SPMD) {
+        // TODO(yeounoh) this can be further optimized via CopyToDevice.
+        at::Tensor tensor = xtensor->ToTensor(false);
```
alanwaketan, line 1166: So we are copying the tensor and than shard the copied one? So, there is no way to shard the existing tensors directly?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1147,7 +1147,35 @@ void InitXlaModuleBindings(py::module m) {
...
+        xtensor->SetXlaData(
+            TensorToXlaData(tensor, ParseDeviceString("SPMD:0")));
+      }
+    }
+  });
   m.def("_init_computation_client", []() { runtime::GetComputationClient(); });
+  m.def("_xla_get_device_hw_type", [](const at::Tensor& tensor) {
```
alanwaketan, line 1173: I think we have something like this already: https://github.com/pytorch/xla/blob/b6b9c6dabe6359196596a9890e60d6c15c0b7a7d/torch_xla/core/xla_model.py#L237

Can we do some consolidation?

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Good question, some things to note and why I decided to do this way
- it's not re-sharding, as we only need to replicate a single PjRtData across devices.
- the data already fits on a single device memory
- it's a one-time overhead when we initialize the SPMD mode, 
- it's cleaner this way, as we drop the reference to the old PjRtData
- as "TODO(yeounoh) this can be further optimized via CopyToDevice" suggests, I implemented `CopyToDevice` API a while back to support device to device transfer. It's an optimization, but given that this is a one-time overhead at the startup time, I deprioritized this.

I actually have a generic re-sharding API for auto-sharding feature, but that runs a separate computation, can be more efficient, but we may have to worry about when stack gets recycled afterward.

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Oh, thank you -- will look into this!

torch_xla/csrc/init_python_bindings.cpp
yeounoh: cc @JackCaoG

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1147,7 +1147,35 @@ void InitXlaModuleBindings(py::module m) {
...
+        XLAGraphExecutor::Get()->GetLiveTensors(&backend_device);
+    for (auto xtensor : xtensors) {
+      XlaDeviceType xla_device_type =
+          static_cast<XlaDeviceType>(xtensor->GetDevice().type());
+      if (xla_device_type != XlaDeviceType::SPMD) {
+        // TODO(yeounoh) this can be further optimized via CopyToDevice.
+        at::Tensor tensor = xtensor->ToTensor(false);
+        xtensor->SetXlaData(
+            TensorToXlaData(tensor, ParseDeviceString("SPMD:0")));
+      }
```
jonb377, line 1169: We statically track whether to execute in SPMD mode here: https://github.com/pytorch/xla/blob/13dc9df4df52225f428e059f54661d1f9481fd65/torch_xla/csrc/device.cpp#L86

It's initialized on the first `UseVirtualDevice` call. Not sure if it's an issue, but could we `XLA_CHECK(UseVirtualDevice())` at the end to ensure it's being set properly?

torch_xla/csrc/init_python_bindings.cpp
yeounoh: @alanwaketan it turns out that `xla_device_hw` tracked `PJRT_DEVICE` value instead. I think there is room for improvement/consolidation. Will follow-up in a separate PR.

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Oh, that's set outside `force_spmd_device` but in `use_spmd`. Bascially `UseVirtualDevice` looks at `XLA_USE_SPMD` envvar, but in C++ we cannot update the envvar system-wide.

torch_xla/csrc/init_python_bindings.cpp
alanwaketan: Sure.

.vscode/settings.json
```diff
@@ -18,5 +18,15 @@
         "./bazel-out/_coverage/_coverage_report.dat"
     ],
     "python.formatting.provider": "yapf",
-    "editor.formatOnSave": true
-}
\ No newline at end of file
+    "editor.formatOnSave": true,
```
alanwaketan, line 21: I guess you don't want to update this?

alanwaketan (approved):
LGTM. Please address the last comment and comments from others.

torch_xla/csrc/init_python_bindings.cpp
jonb377: Sorry if I'm missing something 😅 

Based on https://github.com/pytorch/xla/blob/aa0641467b335f22a00d97f33ae8be6049e81624/torch_xla/runtime.py#L245-L251, we only call `force_spmd_device` when the config is locked, so `UseVirtualDevice`'s return value has been fixed (to `false` without the env var set). So would the runtime would still expect an unsharded computation?

Here's a program which demonstrates the problem:

```python
import torch_xla.runtime as xr, torch_xla, torch

print(xr.addressable_device_count()) # Calls into `_xla_num_devices` which locks the config
t = torch.randn(2, 2).to('xla')  # Data is on non-SPMD device
xr.use_spmd()  # Now we convert the data to replicated for SPMD

print(t + t) # Does this execute replicated?
```

.vscode/settings.json
yeounoh: Oh no, thanks!

torch_xla/csrc/init_python_bindings.cpp
yeounoh: Ahhhhh, I get it now `static bool` -- "it's initialized on the first call" good catch. Will look into it.

yeounoh: @jonb377 do you know if that needs to be static, or we can just do the envvar lookup? I don't have the context as to why that was declared static. cc @JackCaoG

yeounoh: hm my guess is that's because we can't set envvar system-wide in python... let me work around it.

yeounoh: Resolved, and updated the unit test inspired by @jonb377 's example ❤️ 
```
...

    # execute replicated
    self.assertTrue(
        "{replicated}" in torch_xla._XLAC._get_xla_tensors_hlo([t + t]))
```

yeounoh:
@JackCaoG @wonjoolee95 do you know why this may be failing dynamo tests?
```
FAIL: test_resnet18 (__main__.DynamoTrainingOptimizerTest)
  File "/tmp/pytorch/xla/test/dynamo/test_dynamo.py", line 519, in test_resnet18
    self.assertEqual(met.metric_data('CompileTime')[0], 7)
AssertionError: 6 != 7
```

JackCaoG:
> @JackCaoG @wonjoolee95 do you know why this may be failing dynamo tests?
> 
> ```
> ======================================================================
> FAIL: test_resnet18 (__main__.DynamoTrainingOptimizerTest)
> ----------------------------------------------------------------------
>   File "/tmp/pytorch/xla/test/dynamo/test_dynamo.py", line 519, in test_resnet18
>     self.assertEqual(met.metric_data('CompileTime')[0], 7)
> AssertionError: 6 != 7
> ```

if you rebase the issue will be gone.

torch_xla/csrc/device.cpp
```diff
@@ -81,12 +81,17 @@ bool ShouldUseVirtualDevice() {
...
+bool UseVirtualDevice(bool flip_spmd_config) {
   spmd_config_is_locked = true;
   static bool use_virtual_device = ShouldUseVirtualDevice();
+  if (flip_spmd_config) {
+    use_virtual_device = !use_virtual_device;
+  }
   return use_virtual_device;
 }
 
+void ResetVirtualDevice(bool use_spmd) {}
```
jonb377, line 93: This seems unused

torch_xla/csrc/device.cpp
```diff
@@ -81,12 +81,17 @@ bool ShouldUseVirtualDevice() {
   return use_virtual_device;
 }
 
-bool UseVirtualDevice() {
+bool UseVirtualDevice(bool flip_spmd_config) {
   spmd_config_is_locked = true;
   static bool use_virtual_device = ShouldUseVirtualDevice();
+  if (flip_spmd_config) {
+    use_virtual_device = !use_virtual_device;
+  }
```
jonb377, line 89: `flip_spmd_config` is a bit confusing at first glance since `UseVirtualDevice(true)` could set the value to false, is there a use case to disable the virtual device? Or could we make this `UseVirtualDevice(bool force_spmd = false)` and do

```cpp
if (force_spmd) {
   use_virtual_device = true;
}
```

torch_xla/csrc/init_python_bindings.cpp
jonb377: Hmm I think the static bool is just an optimization to avoid querying the environment on each lookup. I'm not sure how significant the overhead would be without it.

torch_xla/runtime.py
```diff
@@ -244,9 +244,11 @@ def use_spmd():
 
   if torch_xla._XLAC._xla_get_spmd_config_is_locked(
   ) and not xu.check_env_flag("XLA_USE_SPMD"):
-    raise RuntimeError(
+    warnings.warn(
+        "Replicating tensors already initialized on non-virtual XLA device for SPMD. "
```
JackCaoG, line 248: nit, it is better to call out Pytorch/XLA will handle the data movement for the user but it will be slower. Otherwise this warning sounds like the result is incorrect which scare users.

test/spmd/test_xla_spmd_python_api_interaction.py
```diff
@@ -123,8 +123,20 @@ def test_runtime_spmd_api(self):
...
+    t = torch.ones(2, 2).to(xm.xla_device())
+
+    # Should enable SPMD without crashing.
+    xr.use_spmd()
+    self.assertTrue("SPMD" in torch_xla._XLAC._xla_get_device_hw_type(t))
+    self.assertTrue(xr.is_spmd())
+
+    # execute replicated
+    self.assertTrue(
+        "{replicated}" in torch_xla._XLAC._get_xla_tensors_hlo([t + t]))
```
jonb377, line 139: Could we also add a test which initializes tensors on non-SPMD, calls `use_spmd`, and runs an actual computation on the result?

jonb377 (commented):
Thanks Yeounoh!

torch_xla/csrc/device.cpp
```diff
@@ -81,12 +81,17 @@ bool ShouldUseVirtualDevice() {
   return use_virtual_device;
 }
 
-bool UseVirtualDevice() {
+bool UseVirtualDevice(bool flip_spmd_config) {
```
JackCaoG, line 84: this `flip_spmd_config` is a bit too hacky. we used `static bool` because we don't intend to change this field. If we do intend to change it, it is better to put the static bool in the top of this file and introduce a `SetUseVirtualDevice(bool)`. You can fix that in a follow up pr.

JackCaoG (approved):
Left some comments, you can fix them in a follow up pr.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1304,7 +1304,7 @@ XLAGraphExecutor::CompilationResult XLAGraphExecutor::Compile(
     lowering_ctx.AddResult(root);
   }
   // Always execute sharded when running in SPMD mode
-  bool is_sharded = (coll.device == GetVirtualDevice());
+  bool is_sharded = (coll.device == GetVirtualDevice()) || UseVirtualDevice();
```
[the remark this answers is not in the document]
yeounoh: No, the device doesn't get updated since it's immutable in lazy::Tensor. I have a TODO comment for this. So we need both.

torch_xla/csrc/device.cpp
yeounoh: i like this idea better, thanks!

test/spmd/test_xla_spmd_python_api_interaction.py
yeounoh: that test can be done after we make device mutable.

torch_xla/csrc/device.cpp
yeounoh: Thanks, I adoped this as well as Jon's. We can refactor once again after making the device mutable in the tensor.

test/spmd/test_xla_spmd_python_api_interaction.py
yeounoh: I think this can be more properly done once we change the device mutable.

torch_xla/csrc/device.cpp
yeounoh: Yes, removed!

torch_xla/csrc/device.cpp
yeounoh: I like this, done!

torch_xla/csrc/xla_graph_executor.cpp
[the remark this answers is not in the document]
yeounoh: We do, since the tensor device is immutable. I will make it mutable in the upstream in the follow-up pr.

torch_xla/runtime.py
yeounoh: Updated the comment!

yeounoh:
The CI passed, and re-ran the local tests again. There is non-related test failure, which also fails on master:
```
FAIL: test_resnet18 (__main__.DynamoTrainingOptimizerTest)
  File "test/dynamo/test_dynamo.py", line 518, in test_resnet18
    self.assertEqual(met.metric_data('CompileTime')[0], 6)
AssertionError: 7 != 6

Ran 16 tests in 69.818s
```

I will merge after rebasing.

PR #6681, opened 2024-03-06
Add missing shell options to `test/tpu/run_tests.sh`

will-cromar:
Otherwise this script never actually fails when the test cases fail :facepalm:


zpcore (approved):

zpcore:
For the CI, do we want to terminate the runs if error happen or we continue to next test?

will-cromar:
> For the CI, do we want to terminate the runs if error happen or we continue to next test?

IMO we should terminate and return an error

vanbasten23:
Good catch. Previous TPU CI silently fails: https://github.com/pytorch/xla/actions/runs/8179987440/job/22367135194

PR #6683, opened 2024-03-07
[SPMD][DTensor] introduce xla_distribute_module for DTensor integration

yeounoh:
This is to support https://github.com/pytorch/pytorch/issues/92909


yeounoh:
cc @baoleai for visibility

yeounoh:
CI turned green, and locally looks good on TPU and CPU
```
python test/spmd/test_dtensor_integration.py
...
Ran 3 tests in 3.740s

OK
```

alanwaketan (approved):
LGTM, but how does that work with auto_sharding? You still shard the inputs in the test case.

yeounoh:
> LGTM, but how does that work with auto_sharding? You still shard the inputs in the test case.

Good question, we were thinkning about introducing pre-defined partition_fn for autosharding, e.g., `torch_xla.distributed.auto_sharding_policy` (subject to change). It would just be calling `use_spmd(auto=True)` though.

PR #6685, opened 2024-03-07
[Infra] Bump python version to 3.10

alanwaketan:
Summary:
Bump python version to 3.10 to better align witn our TPU VM minimum version.

Test Plan:
N/A

### DIFF
--- a/infra/tpu-pytorch-releases/dev_images.auto.tfvars
+++ b/infra/tpu-pytorch-releases/dev_images.auto.tfvars
@@ -2,6 +2,7 @@ dev_images = [
   {
     accelerator = "tpu"
     extra_tags  = ["tpu"]
+    python_version = "3.10"
   },
   {
     accelerator  = "cuda"


scripts/build_torch_wheels.sh
```diff
@@ -6,7 +6,7 @@ set -x  # Display commands being run.
 PYTHON_VERSION=$1
 RELEASE_VERSION=$2  # rX.Y or nightly
 BUILD_CPP_TESTS="${3:-0}"
-DEFAULT_PYTHON_VERSION=3.8
+DEFAULT_PYTHON_VERSION=3.10
```
will-cromar, line 9: I don't think this script is used anymore

infra/tpu-pytorch-releases/dev_images.tf
```diff
@@ -2,7 +2,7 @@ variable "dev_images" {
   type = list(object({
     accelerator    = string
     arch           = optional(string, "amd64")
-    python_version = optional(string, "3.8")
+    python_version = optional(string, "3.10")
```
will-cromar, line 5: Do you mind just adding `python_version = "3.10"` to the images you want to update (e.g. [here](https://github.com/pytorch/xla/blob/46e2230012c37ef0cdaf4b9fa7bde1d61e2ebe06/infra/tpu-pytorch-releases/dev_images.auto.tfvars#L9)) instead of updating the default? IMO we should remove this default and explicitly mark the python version we want for each artifact, but that's a separate PR

infra/ansible/development.Dockerfile
```diff
@@ -17,3 +17,8 @@ ENV TAGS="bazel,configure_env,install_deps"
 ARG ansible_vars
 RUN ansible-playbook playbook.yaml -e "stage=build" -e "${ansible_vars}" --tags "${TAGS}"
 RUN ansible-playbook playbook.yaml -e "stage=release" -e "${ansible_vars}" --tags "${TAGS}"
+
+RUN useradd --create-home --shell /bin/bash ptxla
```
will-cromar, line 21: Can you move this to a separate PR? This is much riskier than updating the Python version.

will-cromar: I'll see if there's a way to use the TPU as a non-root user in a container in the meantime.

will-cromar (commented):
To test the terraform changes, you can run `terraform plan` from the `tpu-pytorch-releases` directory and post the output here.

.devcontainer/tpu-contributor/devcontainer.json
```diff
@@ -22,5 +22,6 @@
         "ms-python.python"
       ]
     }
-  }
+  },
+  "containerUser": "ptxla"
```
yeounoh, line 26: This may break the builds or test executions, etc.. I think we may need sudo/root privilege for some of the operations. cc @will-cromar

infra/ansible/development.Dockerfile
yeounoh: ah, +1 https://github.com/pytorch/xla/pull/6685/files#r1516655463

.devcontainer/tpu-contributor/devcontainer.json
alanwaketan: Will remove that.

infra/ansible/development.Dockerfile
alanwaketan: Will remove that.

infra/tpu-pytorch-releases/dev_images.tf
alanwaketan: which means I need to discard all the changes in this PR?

alanwaketan:
@will-cromar @yeounoh I discarded all changes and only change the file Will suggested?

will-cromar (approved):
Thanks! Should the CUDA image also be Python 3.10?

alanwaketan:
> Thanks! Should the CUDA image also be Python 3.10?

I have no ideas what's the default python version for gpu vm. cc @vanbasten23 do you have any recommendations?

PR #6690, opened 2024-03-07
Fix TPU CI failures by updating torchvision

mbzomowski:
Fix TPU CI failures by updating torchvision


JackCaoG:
Do we still need this pr?

PR #6691, opened 2024-03-07
Install torch wheel from pytorch to unblock TPU CI

vanbasten23:
https://github.com/pytorch/xla/pull/6681 helps uncover the silent failures in the `TPU Integration Test / tpu-test (push)`. The failure is
```
+ python3 test/test_operations.py -v
  File "test/test_operations.py", line 47, in <module>
    import torchvision
  File "/usr/local/lib/python3.8/site-packages/torchvision/__init__.py", line 6, in <module>
    from torchvision import _meta_registrations, datasets, io, models, ops, transforms, utils
  File "/usr/local/lib/python3.8/site-packages/torchvision/_meta_registrations.py", line 164, in <module>
    def meta_nms(dets, scores, iou_threshold):
  File "/home/runner/.local/lib/python3.8/site-packages/torch/library.py", line [46](https://github.com/pytorch/xla/actions/runs/8180398097/job/22368345687#step:5:47)7, in inner
    handle = entry.abstract_impl.register(func_to_register, source)
  File "/home/runner/.local/lib/python3.8/site-packages/torch/_library/abstract_impl.py", line 30, in register
    if torch._C._dispatch_has_kernel_for_dispatch_key(self.qualname, "Meta"):
RuntimeError: operator torchvision::nms does not exist
```
The failure is due to the torch wheel built by us and torchvison official wheel is not compatible with the torch wheel built by us.

### DIFF
--- a/.github/workflows/tpu_ci.yml
+++ b/.github/workflows/tpu_ci.yml
@@ -32,8 +32,8 @@ jobs:
         run: |
           pip install fsspec
           pip install rich
-          pip uninstall torch
-          pip uninstall torchvision
+          yes | pip uninstall torch
+          yes | pip uninstall torchvision
           pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121
           cd pytorch/xla
           test/tpu/run_tests.sh
--- a/.github/workflows/tpu_ci.yml
+++ b/.github/workflows/tpu_ci.yml
@@ -26,14 +26,14 @@ jobs:
         run: |
           cd pytorch/xla
           python3 setup.py install --user
-      - name: Install torch wheel built by PyTorch
-        run: |
-          pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121 --user
       - name: Run Tests
         env:
           PJRT_DEVICE: TPU
         run: |
           pip install fsspec
           pip install rich
+          pip uninstall torch
+          pip uninstall torchvision
+          pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121
           cd pytorch/xla
           test/tpu/run_tests.sh
--- a/.github/workflows/tpu_ci.yml
+++ b/.github/workflows/tpu_ci.yml
@@ -30,7 +30,7 @@ jobs:
         env:
           PJRT_DEVICE: TPU
         run: |
-          pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121
+          pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121 --user
           pip install fsspec
           pip install rich
           cd pytorch/xla
--- a/.github/workflows/tpu_ci.yml
+++ b/.github/workflows/tpu_ci.yml
@@ -26,6 +26,9 @@ jobs:
         run: |
           cd pytorch/xla
           python3 setup.py install --user
+      - name: Install torch wheel built by PyTorch
+        run: |
+          pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121 --user
       - name: Run Tests
         env:
           PJRT_DEVICE: TPU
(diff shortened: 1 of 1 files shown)


mbzomowski:
Actions > TPU Integration Test > Run workflow > then select your branch, and it'll start the run from your edited workflow

FYI I opened a PR for this: https://github.com/pytorch/xla/pull/6690, which failed: https://github.com/pytorch/xla/actions/runs/8193035259

vanbasten23:
Building torchvision from src is working https://github.com/pytorch/xla/actions/runs/8195031279: we don't have the the `nms` error and we don't have to rely on nightly torch wheels so that we have to break every time we make companion change in pytorch. cc @will-cromar @mbzomowski

vanbasten23:
Per offline discussion, it's less ideal to compile torchvision from src and it's better to stick with the original plan

vanbasten23:
Current failure:
```
+ python3 test/test_operations.py -v
2024-03-08T19:03:57.5462724Z Traceback (most recent call last):
2024-03-08T19:03:57.5463798Z   File "test/test_operations.py", line 31, in <module>
2024-03-08T19:03:57.5464676Z     import torch_xla
2024-03-08T19:03:57.5467031Z   File "/home/runner/.local/lib/python3.8/site-packages/torch_xla-2.3.0+git177eb6e-py3.8-linux-x86_64.egg/torch_xla/__init__.py", line 7, in <module>
2024-03-08T19:03:57.5469219Z     import _XLAC
2024-03-08T19:03:57.5479640Z ImportError: /home/runner/.local/lib/python3.8/site-packages/torch_xla-2.3.0+git177eb6e-py3.8-linux-x86_64.egg/_XLAC.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZN5torch4lazy13MetricFnValueB5cxx11Ed
2024-03-08T19:03:57.8749327Z ##[error]Process completed with exit code 1.
```

vanbasten23:
torch has abi disabled. I think we enabled it when building torch_xla

.github/workflows/tpu_ci.yml
```diff
@@ -30,6 +32,7 @@ jobs:
         env:
           PJRT_DEVICE: TPU
         run: |
+          pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121 --user
```
will-cromar, line 35: I know the error originates from the other file, but we don't need `torchaudio` and there is a CPU index https://download.pytorch.org/whl/nightly/cpu

We don't need to install the CUDA versions on TPU

.github/workflows/tpu_ci.yml
```diff
@@ -10,6 +10,8 @@ jobs:
     runs-on: v4-runner-set
     steps:
       - name: Checkout and Setup PyTorch Repo
+        env:
+          _GLIBCXX_USE_CXX11_ABI: 0
```
will-cromar, line 14: The C++11 ABI switch has been such an unnecessary headache. I don't know why the default for building `torch` doesn't match their published wheels....

will-cromar (approved):
This lgtm as a temporary fix if it's working. Please update to the CPU wheels before merging.

I'm interested to see if https://github.com/pytorch/xla/pull/6704 works as a more stable solution.

vanbasten23:
> This lgtm as a temporary fix if it's working. Please update to the CPU wheels before merging.

Why CPU wheels? The old TPU CI uses the cuda wheel and our github index page also suggest to install the cuda torch wheel. So shouldn't our CI be consistent?

will-cromar:
The CPU wheel is much smaller and will download more quickly. The other TPU CI should also be using CPU wheels. Our release builds will get tested against the final upstream CUDA wheel.

will-cromar:
> If our github index page suggests our users to use cuda torch wheel, should we do the same? Or it doesn't matter much?

These are just nightly builds, so in my mind it doesn't matter.

.github/workflows/tpu_ci.yml
vanbasten23: fixed.

vanbasten23:
Pending on a new TPU CI run. Stay tuned..

mbzomowski:
> Pending on a new TPU CI run. Stay tuned..

https://github.com/pytorch/xla/actions/runs/8240857381

Looks good

mbzomowski (approved):

vanbasten23:
The new TPU CI passed. Thanks for the review.

PR #6694, opened 2024-03-07
[Fori_loop|While_loop] Add Fori loop implemented with while_loop

ManfeiBai:


ManfeiBai:
only 0~20 example situation could pass, I doube the logic of fori_loop should be changed of the cond_fn part

PR #6698, opened 2024-03-08
Remove `torchvision` from `test_operations`

will-cromar:
~Let's see if we can save ourselves a headache and just ignore `torchvision`.~ Nope, we have to keep `torchvision` in the TPU CI for `test_dynamo`.

cc @vanbasten23 @mbzomowski


test/test_operations.py
```diff
@@ -44,7 +44,6 @@
 import torch_xla.core.xla_model as xm
 import torch_xla.core.functions as xf
 import torch_xla.debug.profiler as xp
-import torchvision
```
JackCaoG, line 47: I think test_dynamo also needs torch_vision

Issue #6700, opened 2024-03-08
PyTorch/XLA should not crash runtime during int64 dot product on TPU


sagelywizard:
## Issue description

Currently, PyTorch/XLA will crash the entire runtime (ret check failure?) if you do an int64 dot product on a TPU. Seems like poor UX to me. It'd be preferable to raise a proper Python exception rather than crashing the whole runtime. This is particularly problematic for folks using PyTorch/XLA in a Jupyter notebook, because it'll crash the whole notebook.

## Code example

```
> import torch
> import torch_xla.core.xla_model as xm
> dev = xm.xla_device()
> t1 = torch.tensor([1,2,3], device=dev)
> t2 = torch.tensor([3,4,5], device=dev)
> t1 @ t2
Non-OK-status: status.status() status: UNIMPLEMENTED: While rewriting computation to not contain X64 element types, XLA encountered an HLO for which this rewriting is not implemented: %dot.3 = s64[] dot(s64[3]{0} %p1.2, s64[3]{0} %p0.1), lhs_contracting_dims={0}, rhs_contracting_dims={0}
```

## System Info

- reproducible on XLA backend [CPU/TPU/CUDA]: TPU
- torch_xla version: 2.1.0

JackCaoG:
Thanks @sagelywizard for raising the issue. I think what we could do is
1. verify if this is the issue on nightly
2. fallback to cpu if in64 dot product is still not implemented.

@bhavya01 do you have cycle to help with this?

bhavya01:
Looking at this now

will-cromar:
Crashing on this operation is one thing. Fallback to CPU to avoid the crash seems reasonable to me in this case.

The _way_ we crash is the real issue IMO. Since most of the time, we're running through `xmp.spawn`. When we hit a fatal error, just the child process dies (usually after dumping an error message) and the parent throws a real exception.

But in notebooks, if you run an operation that fails a status check, it crashes the whole notebook process and resets your state. That's a pretty bad user experience.

JackCaoG:
hmm do we know why?

will-cromar:
I reproduced the error to get the line number where we terminate the process:

https://github.com/pytorch/xla/blob/6cc5c3819c09c7b1b4ca4927d7fa65133f95b41c/torch_xla/csrc/runtime/debug_macros.h#L20

[...]

PR #6708, opened 2024-03-09
Fallback to CPU for dot implementation for long tensors

bhavya01:
Fall back to cpu implementation for dot product if both the tensor are int64. #6700


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1122,11 +1122,17 @@ at::Tensor XLANativeFunctions::div(const at::Tensor& self,
 
 at::Tensor XLANativeFunctions::dot(const at::Tensor& self,
                                    const at::Tensor& tensor) {
+  std::cout << "aten_xla_type " << self.scalar_type() << std::endl;
```
JackCaoG, line 1125: nit remove.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1122,11 +1122,17 @@ at::Tensor XLANativeFunctions::div(const at::Tensor& self,
...
   TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
   XLA_CHECK_EQ(self.dim(), 1)
       << "dot: Expected 1-D argument self, but got " << self.dim() << "-D";
   XLA_CHECK_EQ(tensor.dim(), 1)
       << "dot: Expected 1-D argument tensor, but got " << tensor.dim() << "-D";
+  if (self.scalar_type() == at::ScalarType::Long &&
+      tensor.scalar_type() == at::ScalarType::Long) {
+    return at::native::call_fallback_fn<&xla_cpu_fallback, ATEN_OP(dot)>::call(
```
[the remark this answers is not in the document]
JackCaoG: also does this crash only happens in TPU? If so we should also do a conditional fallback only on TPU.

test/cpp/test_aten_xla_tensor_1.cpp
```diff
@@ -467,6 +467,18 @@ TEST_F(AtenXlaTensorTest, TestDot) {
...
+  torch::Tensor a = torch::randint(0, 100, {10});
+  torch::Tensor b = torch::randint(0, 100, {10});
+  torch::Tensor c = torch::dot(a, b);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = CopyToDevice(b, device);
+    torch::Tensor xla_c = torch::dot(xla_a, xla_b);
+    AllClose(c, xla_c);
+  });
+}
```
JackCaoG, line 480: can we add counter checks similar to 
```
  ExpectCounterChanged("xla::mm", cpp_test::GetIgnoredCounters());
```
we do expect `aten` counter to change since it will fallback, but xla counter will also changed since it will first dispatched to XLA.

JackCaoG (commented):
Thanks! Did you verify using user script as well? I think `@` should be dispatched to `dot` but it is always good to double check.

miladm:
is this the only operation missing support for `long` `dtype`?

bhavya01:
> Is this the only operation missing support for `long` `dtype`? I assume this is TPU specific?

On more testing, I found when one of the tensors is a long and other is any integral type, it still doesn't work. Change the implementation to reflect this. Yes, this is only for TPU.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1127,6 +1127,18 @@ at::Tensor XLANativeFunctions::dot(const at::Tensor& self,
...
       << "dot: Expected 1-D argument tensor, but got " << tensor.dim() << "-D";
+  // Fallback to CPU if both tensor types are integral and atleast one of them
+  // is a long, as this multiplication is not supported for TPUs.
+  XlaDeviceType hw_type =
+      static_cast<XlaDeviceType>(bridge::GetCurrentDevice().type());
+  if (CheckTpuDevice(hw_type) &&
+      (at::isIntegralType(self.scalar_type(), /*include_bool=*/true) &&
+       at::isIntegralType(tensor.scalar_type(), /*include_bool=*/true) &&
+       (at::elementSize(self.scalar_type()) == 8 ||
+        at::elementSize(tensor.scalar_type()) == 8))) {
```
JackCaoG, line 1138: this `at::elementSize` is new to me, why couldn't we just check if they are int64? Are you trying to cover the unsigned_int64 case?

bhavya01: Yes, also trying to cover the unsigned_int64 case.

JackCaoG: Can you add a comment here to explain that int64 and uinte64 dot is not supported on TPU yet?

JackCaoG (approved):

PR #6726, opened 2024-03-12
Add IsVirtualDeivce() and refactor

yeounoh:
This addresses the following to support #6322 
- add IsVirtualDeivce() 
- Un-const `tensors` argument to `XLAGraphExecutor::Compile` method, since they can be updated during auto-sharding paass.
- refactor to remove redundant codes

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1939,16 +1939,8 @@ void InitXlaModuleBindings(py::module m) {
           std::vector<std::string> sharding_specs;
           sharding_specs.reserve(tensors.size());
           for (const at::Tensor& tensor : tensors) {
-            XLATensorPtr xtensor = bridge::GetXlaTensor(tensor);
-            XLATensor::ShardingSpecPtr sharding_spec =
-                xtensor ? xtensor->sharding_spec() : nullptr;
-            if (sharding_spec != nullptr) {
-              sharding_specs.push_back(
-                  xla::HloSharding::FromProto(sharding_spec->sharding)
-                      ->ToString());
-            } else {
-              sharding_specs.push_back("");
-            }
+            sharding_specs.push_back(
+                GetXLAShardingSpec(bridge::GetXlaTensor(tensor)));
           }
           return sharding_specs;
         });
--- a/torch_xla/csrc/ops/device_data.h
+++ b/torch_xla/csrc/ops/device_data.h
@@ -36,12 +36,10 @@ class DeviceData : public XlaNode {
   // backend data with a partitioned one in the node operands. Note that
   // this is permitted only if the node holds a placeholder.
   void Assign(std::shared_ptr<torch::lazy::BackendData> data) {
-    // TODO(yeounoh) check if the existing data is a placeholder after we
-    // address the issue where some of the sync tensors spill with device node.
     XLA_CHECK(data->shape() == data_->shape())
         << "Shape mismatch: expected (" << data_->shape().to_string()
         << "), actual (" << data->shape().to_string() << ")";
-    data_ = data;
+    data_.reset(data.get());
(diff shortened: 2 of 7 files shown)


torch_xla/csrc/ops/device_data.h
```diff
@@ -36,12 +36,10 @@ class DeviceData : public XlaNode {
   // backend data with a partitioned one in the node operands. Note that
   // this is permitted only if the node holds a placeholder.
   void Assign(std::shared_ptr<torch::lazy::BackendData> data) {
-    // TODO(yeounoh) check if the existing data is a placeholder after we
-    // address the issue where some of the sync tensors spill with device node.
     XLA_CHECK(data->shape() == data_->shape())
         << "Shape mismatch: expected (" << data_->shape().to_string()
         << "), actual (" << data->shape().to_string() << ")";
-    data_ = data;
+    data_.reset(data.get());
```
JackCaoG, line 42: what's the difference between these two?

torch_xla/csrc/xla_graph_executor.h
```diff
@@ -346,7 +346,7 @@ class XLAGraphExecutor : public torch::lazy::LazyGraphExecutor {
   std::vector<size_t> SetBufferDonors(LoweringContext* lowering_ctx);
 
   // We don't use upstream Compile to have BuildInputOutputAliases.
-  CompilationResult Compile(const std::vector<XLATensorPtr>& tensors,
+  CompilationResult Compile(std::vector<XLATensorPtr>& tensors,
```
JackCaoG, line 349: I can see how this will become problematic for the dynamo, since in dynamo we will first dry run the compilation and don't execute the compiled program. Please leave a TODO somewhere for the dynamo integration with auto-sharding.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1939,16 +1939,8 @@ void InitXlaModuleBindings(py::module m) {
...
-                xtensor ? xtensor->sharding_spec() : nullptr;
-            if (sharding_spec != nullptr) {
-              sharding_specs.push_back(
-                  xla::HloSharding::FromProto(sharding_spec->sharding)
-                      ->ToString());
-            } else {
-              sharding_specs.push_back("");
-            }
+            sharding_specs.push_back(
+                GetXLAShardingSpec(bridge::GetXlaTensor(tensor)));
```
JackCaoG, line 1943: what will `GetXLAShardingSpec ` return when sharding_spec is null?

torch_xla/csrc/ops/device_data.h
yeounoh: Better memory management, this is refactoring to reuse the existing data_ to hold the new data.

torch_xla/csrc/xla_graph_executor.h
yeounoh: Makes sense I will also add a section in the design doc.

torch_xla/csrc/init_python_bindings.cpp
yeounoh: `""` empty string as expected.

torch_xla/csrc/xla_graph_executor.h
yeounoh:
```
// We don't use upstream Compile to have BuildInputOutputAliases.
  // TODO(yeounoh) auto-sharding can change tensors shardings, which needs to be
  // accounted for in Dynamo integration.
  CompilationResult Compile(std::vector<XLATensorPtr>& tensors,
                            absl::Span<const std::string> devices,
                            const SyncTensorCollection& coll,
                            PostOrderData* po_data,
                            const std::vector<torch::lazy::Value>& ir_values);
```

JackCaoG (approved):

PR #6730, opened 2024-03-13
install pinned torchvision for Github Action TPU CI

lsy323:


lsy323:
GHA CI passes with this PR, let's land after removing the code that runs TPU CI on PR push

will-cromar (approved):

will-cromar:
Thanks!

PR #6677, opened 2024-03-06
Pin update March 2024

lsy323:
Update xla pin to HEAD

Summary:
- Update bazel to 6.5.0
- Rename `PJRT_Structure_Base` to `PJRT_Extension_Base` to accommodate change in XLA. 
- Generate custom_call to `mhlo.uniform_dequantize/quantize` to accommodate the change in HLO->MHLO converter patch.
- Reduce the BS of several GPU tests to unblock the pin update from the OOM issue in GPU tests. (Details in comment below).

PR #6740, opened 2024-03-13
Bump JAX version in `setup.py`

will-cromar:
#6677 updated the build date, but did not bump the JAX minor version. We need to install this wheel: https://storage.googleapis.com/jax-releases/nightly/nocuda/jaxlib-0.4.26.dev20240305-cp310-cp310-manylinux2014_x86_64.whl

This breaks the TPU CI since we can't install the right JAX version.


lsy323:
Thanks @will-cromar

lsy323 (approved):

PR #6752, opened 2024-03-14
Register reshape to RegisterAutogradXLA.cpp

wonjoolee95:
Fixes https://github.com/pytorch/pytorch/issues/115816

Previously there is no `reshape` registration in `RegisterAutogradXLA.cpp`, hence this warning message. After this change, we can see it being registered:
```
at::Tensor wrapper_AutogradXLA__reshape(const at::Tensor & self, c10::SymIntArrayRef shape) {
    // No device check
  // DeviceGuard omitted
  return torch_xla::XLANativeFunctions::reshape(self, C10_AS_INTARRAYREF_SLOW(shape));
}
...
    m.impl("reshape",
    TORCH_FN(wrapper_AutogradXLA__reshape));
```

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -3653,13 +3653,17 @@ at::Tensor XLANativeFunctions::pixel_unshuffle(const at::Tensor& self,
       pixel_unshuffle)>::call(self, downscale_factor);
 }
 
-at::Tensor XLANativeFunctions::reshape(const at::Tensor & self, c10::SymIntArrayRef shape) {
+at::Tensor XLANativeFunctions::reshape(const at::Tensor& self,
+                                       at::IntArrayRef shape) {
   // See Note: [Disabling functionalization]
   if (runtime::sys_util::GetEnvBool("XLA_DISABLE_FUNCTIONALIZATION", false)) {
-    return at::native::reshape_symint(self, shape);
+    return at::native::reshape_symint(
+        self,
+        c10::SymIntArrayRef(reinterpret_cast<const c10::SymInt*>(shape.data()),
+                            shape.size()));
   }
-  return at::functionalization::functionalize_aten_op_symint<ATEN_OP(
-      reshape)>::call(self, shape);
+  return at::functionalization::functionalize_aten_op<ATEN_OP(reshape)>::call(
+      self, shape);
 }
 
 at::Tensor XLANativeFunctions::select_backward_symint(
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -3663,11 +3663,13 @@ at::Tensor XLANativeFunctions::reshape(const at::Tensor& self,
                             shape.size()));
   }
   auto inner_tensor = torch::lazy::maybe_unwrap_functional(self);
-  return at::functionalization::functionalize_aten_op_symint<ATEN_OP(
-      reshape)>::call(inner_tensor,
-                      c10::SymIntArrayRef(
-                          reinterpret_cast<const c10::SymInt*>(shape.data()),
-                          shape.size()));
+  return MaybeWrapTensorToFunctional(
+      at::functionalization::functionalize_aten_op_symint<ATEN_OP(
+          reshape)>::call(inner_tensor,
+                          c10::SymIntArrayRef(
+                              reinterpret_cast<const c10::SymInt*>(
+                                  shape.data()),
+                              shape.size())));
 }
 
 at::Tensor XLANativeFunctions::select_backward_symint(
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -3662,8 +3662,9 @@ at::Tensor XLANativeFunctions::reshape(const at::Tensor& self,
         c10::SymIntArrayRef(reinterpret_cast<const c10::SymInt*>(shape.data()),
                             shape.size()));
   }
+  auto inner_tensor = torch::lazy::maybe_unwrap_functional(self);
   return at::functionalization::functionalize_aten_op_symint<ATEN_OP(
-      reshape)>::call(self,
+      reshape)>::call(inner_tensor,
                       c10::SymIntArrayRef(
                           reinterpret_cast<const c10::SymInt*>(shape.data()),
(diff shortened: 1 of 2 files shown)


wonjoolee95:
Fix doesn't seem to be working yet. The warning is gone, but now I see an error:
```
RuntimeError: !at::functionalization::impl::isFunctionalTensor(t) INTERNAL ASSERT FAILED at "/home/wonjoo/pytorch/aten/src/ATen/FunctionalTensorWrapper.cpp":766, please report a bug to PyTorch. The composite op functionalization fallback expects its inputs all not to be functional tensors
```
Updated details at https://github.com/pytorch/pytorch/issues/115816#issuecomment-1998681437.

JackCaoG:
@ysiraichi in case you have some insights

wonjoolee95:
Tried a naive solution to simply unwrap the functional tensor since the error message in https://github.com/pytorch/xla/pull/6752#issuecomment-1998682283 suggests that all tensors needs to be non-functional but this produces the error:
```
  File "/home/wonjoo/debug/reshape-op/reshape_op.py", line 12, in <module>
    tf_data[i] = t_data[i].expand_as(tf_data[i])
RuntimeError: self__storage_saved.value().is_alias_of(result.storage()) INTERNAL ASSERT FAILED at "/home/wonjoo/pytorch/torch/csrc/autograd/generated/VariableType_3.cpp":14700, please report a bug to PyTorch. 
```
Let me get a backtrace of both of these and paste them here soon.

wonjoolee95:
Here is a backtrace of where the error at https://github.com/pytorch/xla/pull/6752#issuecomment-2001012026 is thrown (it's a bit messy but let me know if you need some other info) at https://gist.github.com/wonjoolee95/e608c6204a40dcba8cb4e1b7e2b6c59a, @ysiraichi.

So if I understand this correctly, torch_xla needs to register the op `reshape` in its AutogradXLA key (`ResigisterAutogradXLA.h/cpp` files are auto-generated similar to torch). And to do this, torch_xla needs to declare this `reshape` op under `autograd` here at [xla_native_functions.yaml](https://github.com/pytorch/xla/blob/master/codegen/xla_native_functions.yaml#L421). However, torch_xla is seeing the errors listed at https://github.com/pytorch/xla/pull/6752#issuecomment-2001012026 when we try to do this. I'm a bit lost here -- how can we register the `reshape` op in AutogradXLA key and have a proper forwarding/lowering to the functionalization pass? Cc'ing @ @albanD and @bdhirsh from the other thread (https://github.com/pytorch/pytorch/issues/115816).

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -3653,14 +3653,21 @@ at::Tensor XLANativeFunctions::pixel_unshuffle(const at::Tensor& self,
...
   // See Note: [Disabling functionalization]
   if (runtime::sys_util::GetEnvBool("XLA_DISABLE_FUNCTIONALIZATION", false)) {
-    return at::native::reshape_symint(self, shape);
+    return at::native::reshape_symint(
+        self,
+        c10::SymIntArrayRef(reinterpret_cast<const c10::SymInt*>(shape.data()),
+                            shape.size()));
   }
+  auto inner_tensor = torch::lazy::maybe_unwrap_functional(self);
   return at::functionalization::functionalize_aten_op_symint<ATEN_OP(
```
ysiraichi, line 3666: Hmm. I think we have to call `MaybeWrapTensorToFunctional` on the result of `functionalize_aten_op_symint` (a). The thing is that (a) wraps the tensor arguments with `FunctionalTensorWrapper` at first and, before returning, unwraps it. However, we do need to return a `FunctionalTensorWrapper` (see `AtenFromXlaTensor` function).

ysiraichi (changes requested):

torch_xla/csrc/aten_xla_type.cpp
wonjoolee95: Thanks for the pointer, I've updated the code to do that and seems like the warning is gone, but I'll wait for the CI to verify the unit tests.

wonjoolee95:
I see that a bunch of CI is failing with error: 
```
C++ exception with description "element 0 of tensors does not require grad and does not have a grad_fn
```
This error seems to be thrown at in [torch/csrc/autograd/autograd.cpp](https://github.com/pytorch/pytorch/blob/main/torch/csrc/autograd/autograd.cpp#L109), which is invoked because this PR registers `reshape` as an autograd.

Looking around a bit, if I explicitly set `loss.requires_grad = True`, this error is not thrown. For example:
```
data = torch.tensor([[1.,2.,3.,4.], [5.,6.,7.,8.]], requires_grad=True, device=xla_device)
batch = data.shape[0]
t_data = data.reshape(batch, 2, 2)
tf_data = torch.zeros((batch, 3, 2, 2), device=xla_device)
for i in range(batch):
    tf_data[i] = t_data[i].expand_as(tf_data[i])
loss = torch.sum(tf_data)
print('tf_data\n', tf_data)
print('data\n', data)
# loss.requires_grad = True # if this line is enabled, the error is gone
loss.backward()
``` 
Need to further look into what is the correct backward behavior when `reshape` is registered as an autograd op or if this `loss.requires_grad=True` is expected to be set.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -3653,14 +3653,23 @@ at::Tensor XLANativeFunctions::pixel_unshuffle(const at::Tensor& self,
...
-                                              c10::SymIntArrayRef shape) {
+at::Tensor XLANativeFunctions::reshape(const at::Tensor& self,
+                                       at::IntArrayRef shape) {
   // See Note: [Disabling functionalization]
   if (runtime::sys_util::GetEnvBool("XLA_DISABLE_FUNCTIONALIZATION", false)) {
-    return at::native::reshape_symint(self, shape);
-  }
-  return at::functionalization::functionalize_aten_op_symint<ATEN_OP(
-      reshape)>::call(self, shape);
+    return at::native::reshape_symint(
```
albanD, line 3660: composite implicit ops should never call into the `native::` namespace as it bypasses the dispatcher.
In this case , I guess that this is why you get a Tensor that doesn't require grad: you skipped autograd and so the output never had the proper autograd graph added to it!

wonjoolee95:
This is completed with https://github.com/pytorch/xla/pull/6801, closing this.

PR #6753, opened 2024-03-14
Include auto-sharding configs as part of hash

yeounoh:
This is to support #6322 . We need to include such configs in the hash computation to avoid collision with persistent caching.

### DIFF
--- a/test/spmd/test_xla_auto_sharding.py
+++ b/test/spmd/test_xla_auto_sharding.py
@@ -28,32 +28,29 @@ from torch_xla._internal import tpu
 class XlaAutoShardingTest(test_xla_sharding_base.XlaShardingTest):
 
   # initialized in setUpClass
-  hash_no_spmd = ""
-  hash_with_spmd = ""
+  hash_no_auto = ""
 
   @classmethod
   def setUpClass(cls):
-    cls.init_test_variables()
-    assert cls.hash_no_spmd != cls.hash_with_spmd
-    xr.use_spmd(auto=True)
     super().setUpClass()
+    assert not torch_xla._XLAC._xla_get_auto_sharding()
+    cls.init_test_variables(cls)
+    assert cls.hash_no_auto
+    xr.use_spmd(auto=True)
 
-  def init_test_variables():
-    # initialize before xr.use_spmd()
-    xt_no_spmd = torch.ones(2, 2).to(xm.xla_device())
-    hash_no_spmd = torch_xla._XLAC._get_graph_hash([xt_no_spmd + 0])
-
-    # initialize with SPMD, hasing with sharding annotations
-    # is already being tested in test_xla_sharding.py
-    xr.use_spmd()
-    xt_with_spmd = torch.ones(2, 2).to(xm.xla_device())
-    hash_with_spmd = torch_xla._XLAC._get_graph_hash([xt_with_spmd + 0])
+  def init_test_variables(cls):
+    xt_no_auto = torch.ones(2, 2).to(xm.xla_device())
+    cls.hash_no_auto = torch_xla._XLAC._get_graph_hash([xt_no_auto + 0])
 
   def test_auto_sharding_hashing(self):
     xt = torch.ones(2, 2).to(xm.xla_device())
+    assert torch_xla._XLAC._xla_get_auto_sharding()
     hash_auto_spmd = torch_xla._XLAC._get_graph_hash([xt + 0])
-    self.assertNotEqual(hash_auto_spmd, self.hash_no_spmd)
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1510,6 +1510,14 @@ XLAGraphExecutor::SyncTensorsGraphInternal(
           coll.hash, torch::lazy::Hash(buffer_donor_index));
     }
   }
+  if (ShardingUtil::GetAutoSharding()) {
+    std::string xla_auto_spmd = "XLA_AUTO_SPMD=1";
+    coll.hash = torch::lazy::HashCombine(
+        coll.hash, torch::lazy::StringHash(xla_auto_spmd.c_str()));
```
JackCaoG, line 1516: I think you can just 
```
    coll.hash = torch::lazy::HashCombine(
        coll.hash, torch::lazy::MHash(ShardingUtil::GetAutoSharding()));
    coll.hash = torch::lazy::HashCombine(
        coll.hash, torch::lazy::StringHash(sys_util::GetEnvString("XLA_AUTO_SPMD_MESH", "").c_str()));
```
The order of the hash is fixed, we don't need to prefix them with string.

JackCaoG: given that we have to hash each IR already, adding 2 small hash won't hurt the non-auto-sharding performance.

JackCaoG:
Can you add a similar test as https://github.com/pytorch/xla/blob/master/test/spmd/test_xla_sharding.py#L622-L637 ? We can at least test hash is different when we change `XLA_AUTO_SPMD_MESH `

torch_xla/csrc/xla_graph_executor.cpp
yeounoh: Sounds good.

test/spmd/test_xla_auto_sharding.py
```diff
@@ -27,10 +27,28 @@
...
+  def init_test_variables(cls):
+    xt_no_auto = torch.ones(2, 2).to(xm.xla_device())
+    cls.hash_no_auto = torch_xla._XLAC._get_graph_hash([xt_no_auto + 0])
+
+  def test_auto_sharding_hashing(self):
+    xt = torch.ones(2, 2).to(xm.xla_device())
+    assert torch_xla._XLAC._xla_get_auto_sharding()
+    hash_auto_spmd = torch_xla._XLAC._get_graph_hash([xt + 0])
+    self.assertNotEqual(hash_auto_spmd, self.hash_no_auto)
+    os.environ['XLA_AUTO_SPMD_MESH'] = '2,2'
```
JackCaoG, line 49: should we del this env var after this test?

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -517,6 +517,14 @@ torch::lazy::hash_t XLAGraphExecutor::GetGraphHash(
           res_hash, torch::lazy::Hash(buffer_donor_index));
     }
   }
+  if (ShardingUtil::GetAutoSharding()) {
```
JackCaoG, line 520: let's drop this if statement, consider a case where we have
```
if (feature_a_enabled) {
  coll.hash = combine(coll.hash , True)
}
if (feature_b_enabled) {
  coll.hash = combine(coll.hash , True)
}
```
in this case there will be a hash collision since hash is the same for either `a` or `b` is enabled. Let's always hash the `autosharding` and `XLA_AUTO_SPMD_MESH `.

yeounoh: I was including the flag name to the string for this... but then I realize that hashing True twice would be different from hashing True once. We can do eitehr way... yea let's drop and make it "always"

test/spmd/test_xla_auto_sharding.py
yeounoh: ah good catch!

torch_xla/csrc/xla_graph_executor.cpp
JackCaoG: right, I was not worried about autosharding. I was worried about we have another feature like
```
if(ShardingUtil::GetAutoSharding()) {
}
if(feature_b_enabled) {
}
```
if `feature_b` also has a boolean hashing as well, there might be conflicts.

I think we can either
1. include string name in the hash(which is what you did) and put this hash in a if condition
2. only hash the boolean value but unconditionally hashing it

I think we usually do 2(for example for `force_ltc_tensor` config)

JackCaoG (approved):

PR #6756, opened 2024-03-14
Build torchvision from the pytorch pin in ansible

will-cromar:


JackCaoG (approved):

infra/ansible/roles/fetch_srcs/tasks/main.yaml
```diff
@@ -20,25 +20,6 @@
       dest: "{{ (src_root, 'pytorch/xla') | path_join }}"
       version: "{{ xla_git_rev }}"
 
-- name: Find *.diff files in pytorch/xla/openxla_patches
```
vanbasten23, line 23: I wonder why we don't need the openxla patches anymore.

will-cromar: The patches are applied by bazel automatically now.

PR #6769, opened 2024-03-18
Use AddBufferDonor to replace SetUpAlias for input output aliasing

JackCaoG:
Existing `SetUpAlias` requires us(PyTorch/XLA) to specify the input and output index we want to alias. This can be difficult for the SPMD use case because we don't know the output sharding hence we can't be sure if input and output will have the same buffer size.

`AddBufferDonor` is a newish API added by XLA team. It only requires us to give compiler a list of input buffer that can be reused for the output. It is up to the compiler to determine how to reuse these buffer. I have been user `AddBufferDonor` for the manual buffer donation for dynamo. In this pr I extend the use case to the auto-aliasing in LTC(the way it works is to determine all of the input buffers that does not have a XLATensor associated and conclude an inplace operation happened so the buffer can be reused).

### DIFF
--- a/torch_xla/csrc/xla_graph_executor.cpp
+++ b/torch_xla/csrc/xla_graph_executor.cpp
@@ -415,7 +415,7 @@ void XLAGraphExecutor::MarkStep(const torch::lazy::BackendDevice& device) {
   ResetTrimCounter();
 }
 
-std::vector<size_t> GetBufferDonorIndex(
+std::vector<size_t> GetBufferDonorIndexFromUserConfig(
     const std::vector<torch::lazy::BackendDataPtr>& parameters_data) {
   std::vector<size_t> buffer_donor_indexs;
   for (size_t i = 0; i < parameters_data.size(); ++i) {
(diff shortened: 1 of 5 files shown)


torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1261,59 +1260,14 @@ XLAGraphExecutor::BuildInputOutputAliases(
       // this buffer is not needed after execution since XLATensor will get a
```
yeounoh, line 1260: This is the case we reuse / donate?

JackCaoG: yea, the idea is we don't care which output will this input get aliased to. It is up to the compiler.

yeounoh: Makes sense, we are reusing instead of creating a new one.

yeounoh (approved):
LGTM

JackCaoG:
Let's see if TPUCI will pass before merging.

PR #6775, opened 2024-03-19
aot prototyping

zpcore:


torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -896,7 +925,11 @@ PjRtComputationClient::GetReplicationDevices() {
 }
 
 xla::PjRtDevice* PjRtComputationClient::StringToPjRtDevice(
-    const std::string& device) {
+    const std::string& raw_device) {
+  std::string device = raw_device;
+  if (raw_device=="AOT:0") {
```
miladm, line 930: question: is this a new flag or a recycle of an existing flag in the upstream torch?

torch_xla/csrc/device.h
```diff
@@ -15,7 +15,7 @@ namespace torch_xla {
 // TODO(yeounoh) `SPMD` is a virtual device that defers data `TransferToDevice`
 // until after the paritioning pass. This avoids transfering  the full input
 // tensor to the device.
-enum class XlaDeviceType { CPU, CUDA, TPU, NEURON, SPMD, PLUGIN };
+enum class XlaDeviceType { CPU, CUDA, TPU, NEURON, SPMD, PLUGIN, AOT };
```
miladm, line 18: Do we have a parallel solution on the inductor side?

zpcore: Unfortunately no, AOT cross-device can't be applied to GPU because GPU compile needs to be done on a GPU machines to get auto-tuning results.

torch_xla/csrc/runtime/pjrt_computation_client.cc
zpcore: `AOT` is the newly added device type, similar to `SPMD` device flag. Both only exist in torch_xla.

PR #6783, opened 2024-03-20
Add patch for unbounded dynamism

ghpvnist:
Some binary ops return types whose element type differs from the operand element types (e.g. Ne). While broadcasting, it should use the element type of the operand instead of the inferred type.

The fix for this is currently in flight, and the patch will be removed in the next pin update once those changes land.


lsy323:
LGTM. Thanks @ghpvnist! Would you please update the PR description and specify the patch will be removed in the next pin? Let's merge after CI Passes

lsy323 (approved):

PR #6790, opened 2024-03-20
Re-enable AMP test on GPU CI.

vanbasten23:


JackCaoG (approved):

vanbasten23:
With the above commit, only `xla_op3` fails instead of all GPU tests xla_op1-3.

PR #6794, opened 2024-03-21
[Backport] Introduce jax_import_guard

alanwaketan:
Summary:
Importing JAX will lock the TPU devices and prevent any pytorxh/xla's TPU computations. To address it, we need to acquire the TPU first. Somehow xm.xla_device() is enough to acquire the TPU device.

Test Plan:
python test/test_pallas.py


JackCaoG (approved):

alanwaketan:
Thanks, Jack!

PR #6798, opened 2024-03-21
Triton

bhavya01:
Add Triton kernel support to PyTorch/XLA

The torch_triton.py is adapted from https://github.com/jax-ml/jax-triton/blob/main/jax_triton/triton_lib.py

Added a new CI (build and test) workflow for Triton as we need GPUs with compute capability > 7 for Triton support and updating the existing builds for multiple GPU architectures causes OOM.


torch_xla/csrc/triton_kernels.h
```diff
@@ -0,0 +1,118 @@
+#ifndef TORCH_XLA_CSRC_TRITON_H_
```
miladm, line 1: suggesting to organize triton files under a separate directory plz

test/test_operations.py
```diff
@@ -1970,6 +1970,32 @@ def foo(x):
...
...\x12r\x8a\xe9H\xa3\x17~(\xc3/pT&\xe2\xd2Z\xe4\x19\xac\xb3L4}\xb6xV\xe3i\x0b\x85Sg\x02j\xe3\x96cw\x02\xd3\xe5,\xa3\xf7<,\xa3\xec\x02\x16\r\xac\xc3\xad\x93c\xac\x166\x04Hc\xc3\x0c\x0b\x9eUyy\xea_\xc3\xfdYKi\xd7\x94\xb6\xa6\xac\x8a4\x12\xb2\xa2\x19>\x06\xdc\x08?;\xdb\x93\xe8\xd4\x89N[<,R(^\x1e\xcd\xd8^\xb9&\xcb5\xe4\xe4\n\xb9[\xd1\xads\xbd\xad\xb3\xcb"\xc1U*@R\x1e\x06\xdd\xbd\xb2M\xea\xb4/+\xfd9\xee\xad\'\xb3\x84\xde\xe0\x96\xd3\xab9}\xcd\t\xdd\x03\x0c$d1\xddB\xfa\x04\xe3=56|\x81\xe6K\xf3(\x06\xf2\xb1\xecw\xbd\x88\x16\x97rU\xd2f~\xf9U\xb2\x10\xdd@R\xbd\xcd\xa1)8\xaen\x17\xef>7\xa6\xcd\xba#};\xe8\xc3\xed\xf0\x1bRj\xfa\xa1\x8e\xf8\x0f2\xa4ed]G(\xfd\x18\x96\x04\x8d\x96\xdd[\xce\xf2\xef!\xdd\xbf\x9e)i\x12\x9d\xbe\x97\xf6\xc3\xbd\xa4\xcd\x93\x84\xba=3\x9d\x87\x9a\xd9P\xca\x12E^r\xc9-[\xb3\xef\xe7\xe6~\xe7(k\x89\xe1\xff\xd4\xaa\xcc\xb6#\xea\xd9\xb2i\xefG\xa8\xfb[!\x0e\xbe\x14\xa93cL\x03\xd9a\xa0\xdb\x02\xed\xc3@of\x07\x1a\xe8\xdc\x05d\x9e\x06\xba\x87\x81\xd3\x19\xad\x19\xbd\xc3@\xbf\xadqz\x17\x90:\x1a\xe8\x1f\x06\x06-cp\x17\xb0aT;~?\x92\x11\xf0\xb2F\x1e\x9e\x1bFg\xac\xe1<<9\x80\xb4\xed\x1ayxv$\xe7\x08\x91\x9f\x8e\x1e=y\x84\x1f\x9d|\x86\x1e\x1f\x1d\xd5\xd7\xc7O>\xd7\xbf\xfe\x03\x97\x03\xa4b'
```
miladm, line 1990: I expect we have more than one test kernel going fwd - wdyt we put all kernel payloads for Triton (and Pallas) in a separate yaml file?

cc @jiawenliu64

bhavya01:
The GPU CI test failed for the CUDA version issue. I'll try if the CI tests pass with a newer CUDA version.

bhavya01:
The Triton implementation now works for single outputs. I added a separate CI for the Triton test. Still need to look into caching the builds.

Everything under torch_xla/csrc/triton directory is common with JAX and it might be better to have it as part of XLA repo. Feel free to skip that part while reviewing.

test/test_triton.py
```diff
@@ -0,0 +1,65 @@
...
+  @unittest.skipIf(xr.device_type() != 'CUDA', "This test only works on GPU.")
+  def test_gpu_custom_call_triton_add(self):
+    size = 16
+
+    x = torch.arange(size, dtype=torch.int64).to("xla")
+    y = torch.arange(size, dtype=torch.int64).to("xla")
+    output = torch.empty_like(x)
+    block_size = 8
+    grid = (size // block_size,)
+    payload = torch_triton.triton_call(
```
vanbasten23, line 53: The payload is the triton code, right?

bhavya01: It is a serialized proto that is compressed using zlib. 
image

The ptx field in the proto represents the compiled triton kernel. Let me know if you need more information.

test/test_triton.py
```diff
@@ -0,0 +1,65 @@
...
+
+  @unittest.skipIf(xr.device_type() != 'CUDA', "This test only works on GPU.")
+  def test_gpu_custom_call_triton_add(self):
+    size = 16
+
+    x = torch.arange(size, dtype=torch.int64).to("xla")
+    y = torch.arange(size, dtype=torch.int64).to("xla")
+    output = torch.empty_like(x)
+    block_size = 8
+    grid = (size // block_size,)
```
alanwaketan, line 52: This produce floor? Is this what you need?

test/test_triton.py
```diff
@@ -0,0 +1,65 @@
...
+    payload = torch_triton.triton_call(
+        x, y, output, size, kernel=add_kernel, grid=grid, BLOCK_SIZE=block_size)
+    torch_xla._XLAC._xla_gpu_custom_call_(output, [x, y], payload)
```
alanwaketan, line 55: Can you follow the new signatures of pallas APIs?

torch_xla/experimental/torch_triton.py
```diff
@@ -0,0 +1,205 @@
...
+
+import os
+from typing import Any, Callable, Dict, Tuple, Union
+import zlib
+import torch
+
+import numpy as np
+import triton
+import triton.language as tl
+import torch_xla.csrc.triton._triton as lib_triton
```
alanwaketan, line 13: So all the triton package you built is to access in python? If that's the case, then when can't we just use jax_lib?

torch_xla/experimental/torch_triton.py
```diff
@@ -0,0 +1,205 @@
...
+
+Grid = Union[int, Tuple[int], Tuple[int, int], Tuple[int, int, int]]
+GridOrLambda = Union[Grid, Callable[[Dict[str, Any]], Grid]]
+
+NUM_WARPS = 4
+NUM_STAGES = 3
+NUM_CTAS = 1
+
+
+def normalize_grid(grid: GridOrLambda, metaparams) -> Tuple[int, int, int]:
```
alanwaketan, line 23: It feels like a lot of helpers are actually copied from jax_triton: https://github.com/jax-ml/jax-triton/blob/ebdbde90c8129eb3e0471ee2d49f8281a9ed24f5/jax_triton/triton_lib.py#L74

Is that true?

alanwaketan (commented):
Leave some high level comments. Let's have a more in-depth discussion tomorrow before proceeding.

torch_xla/experimental/torch_triton.py
bhavya01: You're right! I justed verified that just using jaxlib should also work. I was earlier worried being able to register XLA custom call from JAX but that works too as it just uses a static registry object for custom calls. This should make this PR much simpler. I'll update it and request a review again.

alanwaketan: Perfect. Looking forward to it!

infra/tpu-pytorch/infra_triggers.tf
```diff
@@ -2,7 +2,7 @@ module "terraform_apply" {
   source = "../terraform_modules/apply_terraform_trigger"
 
   included_files    = ["infra/**"]
-  branch           = "master"
+  branch           = "main"
```
will-cromar, line 5: This doesn't look right. Our primary branch is still `master`.

bhavya01: Ah right! I merged main branch instead of master that updated this file. I'll revert it

test/test_triton.py
bhavya01: Thanks for pointing out! Updated it to use Triton.cdiv as the original kernel

WORKSPACE
```diff
@@ -24,6 +24,14 @@ http_archive(
     url = "https://github.com/nlohmann/json/archive/refs/tags/v3.11.2.tar.gz",
 )
 
+http_archive(
```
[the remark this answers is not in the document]
bhavya01: Removed with the new code.

.github/workflows/triton.yml
```diff
@@ -0,0 +1,99 @@
+on:
```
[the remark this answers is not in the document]
bhavya01: The problem with using the existing GPU flow is that it build pytorch without cuda dependency which doesn't work with Triton as Triton uses PyTorch to detect if there is a GPU device.

It would be great if we can use this for now while I figure out how to merge the two CI workflows.

.circleci/triton.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
bhavya01: Triton only works with GPUs with compute capability > 7 and the existing CI GPUs are at 5.2. So, we need to rebuild pytorch with CUDA support for the new GPUs.

torch_xla/experimental/torch_triton.py
bhavya01: Yes, it is similar to the one used in jax-triton. Added a reference

.circleci/triton.sh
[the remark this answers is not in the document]
alanwaketan: Comment at the top to suggest why. @will-cromar can you review this part? Appreciate it.

.github/workflows/triton.yml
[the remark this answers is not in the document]
alanwaketan: Same comment as the triton.sh. @will-cromar

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2356,6 +2357,28 @@ void InitXlaModuleBindings(py::module m) {
               bridge::GetXlaTensors(inputs), payload, output_shapes, dtypes);
           return bridge::AtenFromXlaTensors(xtensors);
         });
+  m.def("_xla_gpu_custom_call",
+        [](const std::vector<at::Tensor>& inputs, const std::string& payload,
+           const std::vector<std::vector<int64_t>>& output_shapes,
+           const std::vector<py::object>& output_dtypes)
+            -> std::vector<at::Tensor> {
+          std::vector<at::ScalarType> dtypes;
```
alanwaketan, line 2365: Nit: maybe we can have a helper function for this to unify this method and the tpu one.

torch_xla/csrc/ops/gpu_custom_call.h
```diff
@@ -0,0 +1,26 @@
+#ifndef XLA_TORCH_XLA_CSRC_OPS_GPU_CUSTOM_CALL_H_
+#define XLA_TORCH_XLA_CSRC_OPS_GPU_CUSTOM_CALL_H_
+
+#include "torch_xla/csrc/ir.h"
+
+namespace torch_xla {
+// TODO: Merge GPU and TPU custom call.
```
alanwaketan, line 7: Probably doesn't matter.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -1272,11 +1272,41 @@ xla::XlaOp BuildCustomSharding(const xla::XlaOp& input, const std::string& type,
...
+std::vector<xla::XlaOp> BuildGpuCustomCall(
+    const std::vector<xla::XlaOp>& inputs, const xla::Shape& output_shape,
+    const std::string& payload) {
+  std::vector<xla::Shape> input_shapes;
+  input_shapes.reserve(inputs.size());
+  for (const auto& input : inputs) {
+    input_shapes.push_back(ShapeHelper::ShapeOfXlaOp(input));
+  }
+
+  // std::vector<xla::Shape> output_shapes;
```
alanwaketan, line 1284: What's this? Maybe removing it.

torch_xla/experimental/torch_triton.py
```diff
@@ -0,0 +1,217 @@
...
+
+  call_proto = kernel_call.to_proto(kernel_call_name, serialized_metadata)
+  return zlib.compress(call_proto)
+
+
+def triton_call(
+    *args: Union[torch.Tensor, bool, int, float, np.float32],
+    kernel: triton.JITFunction,
+    grid: GridOrLambda,
+    name: str = "",
```
alanwaketan, line 199: Maybe we can just drop this parameter?

torch_xla/experimental/torch_triton.py
```diff
@@ -0,0 +1,217 @@
...
+    debug: bool = False,
+    serialized_metadata: bytes = b"",
```
alanwaketan, line 201: Let's drop this as well?

alanwaketan (approved):
LGTM. Thanks @bhavya01. It has been a long journey!

test/test_triton.py
```diff
@@ -0,0 +1,68 @@
+import logging
+import torch
+from torch import nn as nn
+import unittest
+
+import torch_xla.experimental.torch_triton as torch_triton
```
alanwaketan, line 6: One thing I forgot to mention. Do you think it's better to name it as triton instead of torch_triton?

.github/workflows/triton.yml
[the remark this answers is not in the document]
will-cromar: Is it feasible to  merge your new tests into the existing workflow and upgrade the GPUs we use there? It's probably time we use more modern GPUs for all of our CI tests anyway.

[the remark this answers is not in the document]
will-cromar: You can upgrade the runner here: https://github.com/pytorch/xla/blob/8a1ada88e4f657c83d660b32e775e564fb8ea320/.github/workflows/build_and_test.yml#L56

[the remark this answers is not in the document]
JackCaoG: Let me check with pytorch folks first..

.circleci/triton.sh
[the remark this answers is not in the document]
will-cromar: Please don't add any new build scripts. These should have been deleted a long time ago, but nobody has had time to do a full refactor. There's already a variable `cuda_compute_capabilities` to update the compute capabilities, which is set here: https://github.com/pytorch/xla/blob/77bbf7f3804dee0d9f7e4cf4a4a1e0ba226dea65/.github/workflows/_build_plugin.yml#L42C98-L42C123

infra/ansible/config/env.yaml
```diff
@@ -42,7 +42,9 @@ build_env:
 
   cuda:
     TF_CUDA_COMPUTE_CAPABILITIES: "{{ cuda_compute_capabilities }}"
+    TORCH_CUDA_ARCH_LIST: "{{ torch_cuda_arch_list | default('') }}"
     XLA_CUDA: 1
+    USE_CUDA: "{{ build_pytorch_with_cuda | default(0) }}"
```
will-cromar, line 47: Discussed offline. Please rebase onto master and undo the build/CI infra changes, and then I'll work with you on a follow up PR to support the OpenXLA CUDA plugin.

test/test_operations.py
```diff
@@ -2711,18 +2711,6 @@ def test_dlpack_pytorch_cuda_to_xla(self):
     t2_cuda.fill_(6)
     self.assertTrue(torch.allclose(xla_t2.cpu(), t2_cuda.cpu()))
 
-    cuda1 = torch.device('cuda:1')
```
vanbasten23, line 2714: this test shouldn't be removed. Neither should the one below.

bhavya01: Ack. I removed it as the new runner only had 1 GPU but now it is reverted. I will add the triton tests separately.

infra/ansible/config/env.yaml
bhavya01: Reverted the build/CI infra changes. Thanks for helping me out with this

PR #6801, opened 2024-03-21
Remove reshape lowering

wonjoolee95:
To fix https://github.com/pytorch/pytorch/issues/115816, we first tried to simply register the `reshape` op to AutogradXLA, but that seems like giving some errors -- https://github.com/pytorch/xla/pull/6752. 

While we're getting some help from PyTorch folks, another way to solve this problem is by removing the reshape op (this PR) and fix the regression. 

Plan is to merge this on Monday, observe the regression for a day, and revert it the next day (Tuesday). Then based on the regressing test cases, we'll fix the regression.

### DIFF
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -380,7 +380,6 @@ supported:
   - narrow_copy
   - pixel_shuffle
   - pixel_unshuffle
-  - reshape
   - select_backward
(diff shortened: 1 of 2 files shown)


JackCaoG (approved):

wonjoolee95:
I'll merge this now and revert tomorrow.

PR #6802, opened 2024-03-22
Make compile count in Dynamo test more general

JackCaoG:
There has been a lot of changes regarding sgd + dynamo. The purpose of the `test_dynamo.py test_resnet18` was to make sure the `CompileTime` does not increase for the PyTorch/XLA over the time. Instead of using a strict `assertEqual` we can use `assertLessEqual` to avoid breaking upstream pytorch prs even if they managed to reduce the compile count for us.

The most recent pr that failed this test is https://github.com/pytorch/pytorch/pull/122349


yeounoh (approved):
LGTM

lsy323 (approved):

PR #6807, opened 2024-03-22
[Fori_loop|While_loop] Add nested test

ManfeiBai:
Add nested test case for while_loop

### DIFF
--- a/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
+++ b/test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
@@ -55,6 +55,42 @@ class WhileLoopTest(unittest.TestCase):
     expected = _fake_while_loop(cond_fn, body_fn, (init, limit_value))
     self.assertEqual(expected, res)
 
+  def test_while_loop_tpu_subtraction_nested(self):
+
+    device = xm.xla_device()
+
+    def cond_fn(init, limit_value):
+      return limit_value[0] <= init[0]
+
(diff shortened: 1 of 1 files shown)


ManfeiBai:
since its only add test case, we would not include it in 2.3 but only in nightly, so remove @lsy323's review here too and remove `backport 2.3` label too

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -55,6 +55,42 @@ def body_fn(init, limit_value):
...
+      two_value = limit_value.clone()
+      return (torch.sub(torch.sub(init, one_value), one_value), two_value)
+
+    init = torch.tensor([10], dtype=torch.int32, device=device)
+    limit_value = torch.tensor([0], dtype=torch.int32, device=device)
+    res = while_loop(cond_fn, body_fn, (init, limit_value))
+    expected = _fake_while_loop(cond_fn, body_fn, (init, limit_value))
+    self.assertEqual(expected, res)
+
+  def test_while_loop_tpu_addition_nested(self):
```
yeounoh, line 76: let's just add substration, it's more or less the same.

yeounoh (approved):
LGTM

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: Thanks, agree, updated

Issue #6336, opened 2024-01-19
[torchbench] Detectron2 benchmarks failing to run.


ysiraichi:
## 🐛 Bug

After #6296, a few detectron2 benchmarks started failing when using XLA:

```bash
python xla/benchmarks/experiment_runner.py \
    --suite-name torchbench --accelerator cuda --repeat 2 \
    --test eval --xla PJRT --dynamo openxla \
    -k detectron2_fasterrcnn_r_50_c4
```

```python
Traceback (most recent call last):
  File "xla/benchmarks/experiment_runner.py", line 906, in <module>
    main()
  File "xla/benchmarks/experiment_runner.py", line 902, in main
    runner.run()
  File "xla/benchmarks/experiment_runner.py", line 59, in run
    self.run_single_config()
  File "xla/benchmarks/experiment_runner.py", line 247, in run_single_config
    metrics, last_output = self.run_once_and_gather_metrics(
  File "xla/benchmarks/experiment_runner.py", line 324, in run_once_and_gather_metrics
    output, _ = loop(iter_fn=self._default_iter_fn)
  File "xla/benchmarks/experiment_runner.py", line 293, in loop
    output, timing, trace = iter_fn(benchmark_experiment, benchmark_model,
  File "xla/benchmarks/experiment_runner.py", line 209, in _default_iter_fn
    output = benchmark_model.model_iter_fn(
  File "xla/benchmarks/benchmark_model.py", line 155, in eval
    pred = self.module(*inputs)
  File "torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/lib/python3.8/site-packages/detectron2/modeling/meta_arch/rcnn.py", line 150, in forward
    return self.inference(batched_inputs)
  File "/lib/python3.8/site-packages/detectron2/modeling/meta_arch/rcnn.py", line 208, in inference
    proposals, _ = self.proposal_generator(images, features, None)
  File "torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/lib/python3.8/site-packages/detectron2/modeling/proposal_generator/rpn.py", line 454, in forward
[...]

PR #6814, opened 2024-03-23
Refactor `nms` into TorchVision variant.

ysiraichi:
This PR refactors the existing `nms` lowering implementation, changing the appropriate parts for complying with [TorchVision semantics](https://pytorch.org/vision/main/generated/torchvision.ops.nms.html#torchvision.ops.nms). In summary:

- Implement new `nms` lowering, based on the old implementation
    - More comments
    - TorchVision semantics
- Adapts TorchVision tests for `nms`
- Move the lowering implementation to _torch_xla/csrc/xla_lower_util.cpp_
    - Alongside other lowering implementations
- Register the new kernel as an XLA dispatch for `torchvision::nms`
    - Make it possible to call the kernel by calling `torchvision.ops.nms(...)` directly
- [Kills](https://github.com/pytorch/xla/issues/6336#issuecomment-2015544866) the old implementation

cc @miladm @JackCaoG


ysiraichi:
I'm having a bit of trouble identifying the cause of the CI failure. Basically, if we set `XLA_USE_EAGER_DEBUG_MODE=1`, the test `TestNMS::test_nms_ref` starts failing with the following error:

```python
RuntimeError: Bad StatusOr access: INVALID_ARGUMENT: Executable expected parameter 0 of size 4000
but got buffer with incompatible size 4004
```

After looking into where this error was coming from, I found out that it was due to PyTorch/XLA trying to figure out the actual size of the output.

My guess is that, since I call `xla::SetDimensionSize` in the lowering, one of the dimensions are marked as dynamic. Which makes the `XLATensorImpl::sym_sizes_` be populated with a `c10::SymInt`. When we try to retrieve the actual size (triggered by [a `to_functional_tensor(t)` call](https://github.com/pytorch/pytorch/blob/14e348b7ad1b3472812f2b077020d80deaf6a787/aten/src/ATen/FunctionalizeFallbackKernel.cpp#L97) on the result of `nms`), PyTorch/XLA tries to figure it out by actually running the computation.

-----

Given all this information, I'm not sure where the "buffer with imcompatible size 4004" is coming from. `test_nms_ref` does create a tensor of size 4000 (1000x4 elements). But, the last 4 is unclear to me.

Another thing I'm puzzled about is why is this test failing only when `XLA_USE_EAGER_DEBUG_MODE` is set? Why wouldn't a normal test run trigger the same error?

@JackCaoG Have you ever seen an error like this? I believe other lowerings also call this same function (e.g. `BuildMaskedSelect`). So, doesn't seem like a problem with `xla::SetDimensionSize` being used.

JackCaoG:
`XLA_USE_EAGER_DEBUG_MODE ` it is a debug thing where it tries to execute the value of the node upon construction. You can think of it as calling `mark_step` every time a IR node is created.

ysiraichi:
Right. So, what is puzzling is why would it fail only when `XLA_USE_EAGER_DEBUG_MODE` is set. Do you have any guesses?

JackCaoG:
not really but feel free to skip the test when `XLA_USE_EAGER_DEBUG_MODE ` is enabled. nms is a rarely used ops that's not even aten, `XLA_USE_EAGER_DEBUG_MODE` is a debug only flag. I think it is OK.

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -1264,4 +1264,247 @@ xla::XlaOp BuildTpuCustomCall(const std::vector<xla::XlaOp>& inputs,
...
+  xla::XlaBuilder* builder = boxes.builder();
+
+  const int64_t COORDINATES = 4;
+  const xla::XlaOp ONE = xla::One(builder, XLAIndexType);
+  const xla::XlaOp ZERO = xla::Zero(builder, XLAIndexType);
+
+  const xla::Shape& boxes_shape = ShapeHelper::ShapeOfXlaOp(boxes);
+  XLA_CHECK_EQ(boxes_shape.rank(), 2);
+  XLA_CHECK_EQ(boxes_shape.dimensions(1), COORDINATES);
+  int64_t num_boxes = boxes_shape.dimensions(0);
```
vanbasten23, line 1350: I wonder if the implementation here is a copy with some modification of the deleted nms_op.cpp, or if it is a brand new implementation.

ysiraichi: I would say that it is a new implementation heavily based on the previous one. So, the problems that I saw with the previous implementation were:

- Almost no comments: couldn't easily tell what some parts of the code were doing
- Copied from another source: it was copied from an old version of tensorflow
- Different signature and semantics: it returned the best `output_size` box indices, even though some of them might have been suppressed

I believe this implementation checks all of the above-mentioned boxes. I added plenty of comments + a few changes that resulted in a more maintainable code (IMHO).

ysiraichi:
@JackCaoG @vanbasten23 I think this PR is ready. Could you take a look at it when you have some time?

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -1264,4 +1264,247 @@ xla::XlaOp BuildTpuCustomCall(const std::vector<xla::XlaOp>& inputs,
...
+                        updated_state, state);
+
+        xla::XlaOp next_box_index = box_index + ONE;
+        return std::vector<xla::XlaOp>{next_box_index, new_state,
+                                       iou_threshold_mask};
+      },
+      init_values, "BoxSelectionLoop", builder));
+}
+
+xla::XlaOp BuildNms(xla::XlaOp boxes, xla::XlaOp scores,
```
vanbasten23, line 1336: Is there any resource that you referenced when coming up with the implementation?

vanbasten23 (approved):

torch_xla/csrc/xla_lower_util.cpp
ysiraichi: I based this implementation in two other implementations:

- The old PyTorch/XLA `nms` implementation
- TorchVision CUDA `nms` implementation

test/test_operations.py
```diff
@@ -2473,6 +2452,84 @@ def test_dropout(self):
...
+    boxes[:, 2:] += boxes[:, :2]
+    boxes[-1, :] = boxes[0, :]
+    x0, y0, x1, y1 = boxes[-1].tolist()
+    iou_thresh += 1e-5
+    boxes[-1, 2] += (x1 - x0) * (1 - iou_thresh) / iou_thresh
+    scores = torch.rand(N)
+    return boxes, scores
+
+  @skipOnEagerDebug
+  def test_nms_ref(self):
```
miladm, line 2487: @ysiraichi can you please add a dynamic shape input test scenario for `nms`? Because this op is dynamic, it's been falling back to CPU. Of course, there is a lot of interest to bring the op to the xla device; though, this requires correct functionality for dynamism.

miladm: more clarity: number of boxes is set to `1000` at the moment. we want that number to be dynamic.

miladm: to be clear the output of this test appears to be dynamic, though the input number of boxes can also be dynamic for `nms`. This test, currently, covers one dynamism scenario (i.e. the output dynamism).

PR #6828, opened 2024-03-27
Using dummy store when user skips barrier to save open file descriptors

barry-jin:
### Notes
Store in PJRT process group is never used if user skips store based barrier, we will use DummyStore if user chooses to skip store based barrier to save open file descriptors.


jeffhataws:
@JackCaoG @will-cromar apologies as I merged before your comments. Let us know if there's any concern. This is for the AWS neuron branch.

@barry-jin  I think you also have a PR for top-of-tree?

PR #6834, opened 2024-03-27
Using dummy store when user skips barrier to save open file descriptors

barry-jin:
### Notes
Store in PJRT process group is never used if user skips store based barrier, we will use DummyStore if user chooses to skip store based barrier to save open file descriptors. This is useful when running at large scale. 

User can choose to use DummyStore by setting `XLA_USE_DUMMY_STORE=0` and `TORCH_DIST_INIT_BARRIER=1`


will-cromar (approved):
Neat, I didn't know this option existed. Do you know if the store is used for anything after `init_process_group`? The built-in init methods all create a store unconditionally, e.g. https://github.com/pytorch/pytorch/blob/cd51496f8bfef5d2bd29263baa3db57c42b1e7b6/torch/distributed/rendezvous.py#L246

This will raise a not-implemented error, which is fine as long as it doesn't break core functionality:

```
>>> DummyStore().set('a', 'b')
RuntimeError: fn INTERNAL ASSERT FAILED at "/workspaces/work/pytorch/torch/csrc/distributed/c10d/init.cpp":169, please report a bug to PyTorch. Not implemented.
```

barry-jin:
> Neat, I didn't know this option existed. Do you know if the store is used for anything after `init_process_group`? The built-in init methods all create a store unconditionally, e.g. https://github.com/pytorch/pytorch/blob/cd51496f8bfef5d2bd29263baa3db57c42b1e7b6/torch/distributed/rendezvous.py#L246
> 
> This will raise a not implemented error, which is fine as long as it doesn't break core functionality:
> 
> ```
> >>> DummyStore().set('a', 'b')
>   File "<stdin>", line 1, in <module>
> RuntimeError: fn INTERNAL ASSERT FAILED at "/workspaces/work/pytorch/torch/csrc/distributed/c10d/init.cpp":169, please report a bug to PyTorch. Not implemented.
> ```

In torch initi_process_group functions, store is only used for store based barrier and for ProcessGroupNCCL to exchange NCCL IDs. In XLA, this store is never used, so it's safe to use DummyStore here. But if the store is used by other software stacks that depends on XLA, then user definitely needs this store. I can add an environment variable for user to choose to use DummyStore to speed up initialization at large scale.

torch_xla/_internal/rendezvous.py
```diff
@@ -34,7 +40,14 @@ def pjrt_rendezvous_handler(url: str,
   with _store_lock:
     global _store
     if not _store:
-      if xu.getenv_as('TORCHELASTIC_USE_AGENT_STORE', str) == 'True':
+      # Create DummyStore when user skips store based barrier by setting TORCH_DIST_INIT_BARRIER=0
+      # and enables XLA_USE_DUMMY_STORE=1. It's safe to do so because store created by _pjrt_rendezvous_handler
+      # is only used as a barrier in process groups. If store is needed, user can set XLA_USE_DUMMY_STORE=0 to
+      # use TCPStore.
+      if xu.getenv_as('TORCH_DIST_INIT_BARRIER', int, 1) == 0 and xu.getenv_as(
+          'XLA_USE_DUMMY_STORE', int, 0) == 1:
```
jeffhataws, line 48: @barry-jin can you add this new XLA envvar to https://github.com/pytorch/xla/blob/f6d6286d595b588449c13673c0a7d49de7cc6f68/configuration.yaml for documentation?

barry-jin: Sure

will-cromar (approved):
Thanks!

ManfeiBai:
Hi, @barry-jin, now we are preparing 2.5 release, and this change will be included, would you mind help to describe more context about this PR's feature/use-case for or inspired from?

barry-jin:
> Hi, @barry-jin, now we are preparing 2.5 release, and this change will be included, would you mind help to describe more context about this PR's feature/use-case for or inspired from?

Hi @ManfeiBai, the context is that in torch `init_process_group`, a store is usually used for distributed communication purposes. By default, `init_process_group` will create a TCPStore, which is not performant at very large scale. While this TCPStore is not used in XLA, so I introduce this feature to allow user to pass in a DummyStore when `init_process_group` to bypass the TCPStore performance issues at large scale.

PR #6835, opened 2024-03-27
Add a dev image for cuda12.3

bhavya01:


infra/ansible/config/cuda_deps.yaml
```diff
@@ -3,12 +3,14 @@
 cuda_deps:
   # List all libcudnn8 versions with `apt list -a libcudnn8`
   libcudnn:
+    "12.3": libcudnn9-cuda-12=9.0.0.312-1
```
vanbasten23, line 6: I wonder where you got the `libcudnn9-cuda-12=9.0.0.312-1`

vanbasten23, line 6: Also, when pytorch publishes their cuda 12.3, we should align the cudnn version with them.

infra/tpu-pytorch-releases/dev_images.auto.tfvars
```diff
@@ -13,5 +13,10 @@ dev_images = [
     accelerator  = "cuda"
     cuda_version = "12.1"
     extra_tags   = ["cuda"]
+  },
+  {
+    accelerator  = "cuda"
```
will-cromar, line 18: Can you set the python version for at least this image to 3.10? 3.8  (default) hits EOL this year.

bhavya01: Done. Thanks for pointing out.

infra/ansible/config/cuda_deps.yaml
bhavya01: Just searching for libcudnn9 versions doesn't give any results using apt list -a libcudnn9. To install cudnn9 nvidia recommends `sudo apt-get -y install cudnn9-cuda-12` 
image

So searching for apt list -a libcudnn9-cuda-12 gave the above version.

will-cromar (approved):
Thanks!

PR #6836, opened 2024-03-27
Add environment XLA_USE_DUMMY_STORE for user to choose to use DummyStore

barry-jin:
Store in PJRT process group is never used if user skips store based barrier. User can set environment variables `TORCH_DIST_INIT_BARRIER=0` and `XLA_USE_DUMMY_STORE=1` to use DummyStore in initialization.

### DIFF
--- a/torch_xla/_internal/rendezvous.py
+++ b/torch_xla/_internal/rendezvous.py
@@ -44,7 +44,7 @@ def pjrt_rendezvous_handler(url: str,
       # process group initialization. In NxD, the store is also used to exchange python
       # objects. So, to use DummyStore, set TORCH_DIST_INIT_BARRIER=0 and NEURON_USE_DUMMY_STORE=1
       if xu.getenv_as('TORCH_DIST_INIT_BARRIER', int, 1) == 0 and xu.getenv_as(
-          'NEURON_USE_DUMMY_STORE', int, 1) == 1:
+          'TORCH_USE_DUMMY_STORE', int, 1) == 1:
         _store = DummyStore()
       elif xu.getenv_as('TORCHELASTIC_USE_AGENT_STORE', str) == 'True':
         attempt = xu.getenv_as('TORCHELASTIC_RESTART_COUNT', int, defval=0)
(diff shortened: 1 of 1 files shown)


torch_xla/_internal/rendezvous.py
```diff
@@ -39,10 +39,12 @@ def pjrt_rendezvous_handler(url: str,
...
-      # to skip store based barriers. It's safe to do this because store
-      # created by _pjrt_rendezvous_handler is only used as a barrier in
-      # process group initialization.
-      if xu.getenv_as('TORCH_DIST_INIT_BARRIER', int, 1) == 0:
+      # to skip store based barriers. It's safe to do this in Neuron NeMo Megatron
+      # because store created by _pjrt_rendezvous_handler is only used as a barrier
+      # process group initialization. In NxD, the store is also used to exchange python
+      # objects. So, to use DummyStore, set TORCH_DIST_INIT_BARRIER=0 and NEURON_USE_DUMMY_STORE=1
+      if xu.getenv_as('TORCH_DIST_INIT_BARRIER', int, 1) == 0 and xu.getenv_as(
+          'NEURON_USE_DUMMY_STORE', int, 1) == 1:
```
jeffhataws, line 47: Can this be changed to be more generic? This PR should go to mainline also, so having something that work across platform should be better.

barry-jin: Sure, will change this environment variable to `TORCH_USE_DUMMY_STORE`

will-cromar: Most of our env var options are prefixed with `XLA_`. Can you make this `XLA_USE_DUMMY_STORE` instead?

barry-jin: Sure

will-cromar (approved):
Thanks!

JackCaoG (approved):

barry-jin:
Thanks for approvals! Can we merge this PR? cc @jeffhataws @will-cromar @JackCaoG

PR #6839, opened 2024-03-27
udpate release tag to rc6

lsy323:
Merge the change to r2.3 release branch instead of master, because our infra can use the config from target branch now. cc @will-cromar

### DIFF
--- a/infra/tpu-pytorch-releases/artifacts.auto.tfvars
+++ b/infra/tpu-pytorch-releases/artifacts.auto.tfvars
@@ -19,10 +19,78 @@ nightly_builds = [
     accelerator  = "cuda"
     cuda_version = "12.1"
   },
+  {
+    accelerator  = "cuda"
+    cuda_version = "12.1"
+    python_version = "3.10"
+  },
+  {
+    accelerator  = "cuda"
+    cuda_version = "12.1"
+    python_version = "3.11"
+  },
 ]
 
 # Built on push to specific tag.
 versioned_builds = [
+  # Remove libtpu from PyPI builds
+  {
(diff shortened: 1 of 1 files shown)


infra/tpu-pytorch-releases/artifacts.auto.tfvars
```diff
@@ -19,10 +19,78 @@ nightly_builds = [
...
+  # Bundle libtpu for Kaggle
+  {
+    git_tag         = "v2.3.0-rc6"
+    package_version = "2.3.0-rc6+libtpu"
+    pytorch_git_rev = "v2.3.0-rc6"
+    accelerator     = "tpu"
+    python_version  = "3.10"
+    bundle_libtpu   = "1"
+  },
+  {
```
will-cromar, line 78: You can remove the 3.8 CUDA build IMO. The EOL is later this year: https://devguide.python.org/versions/

will-cromar (approved):

JackCaoG (approved):

yeounoh (approved):

infra/tpu-pytorch-releases/artifacts.auto.tfvars
lsy323: Sorry I just saw the comment after merging lol. Do we plan to remove 3.8 for all of the build targets, instead of only CUDA?

lsy323: Not sure if we should keep it around for a while, from the link it will be reaching EOL in Oct 2024.

PR #6843, opened 2024-03-28
Update base image in .circleci dockerfile

bhavya01:


.circleci/docker/cloudbuild.yaml
```diff
@@ -6,15 +6,15 @@ steps:
   args: [
```
vanbasten23, line 6: Where is this file used?

bhavya01: I used to create a manual trigger on GCP to build the image. I don't think it is used anywhere else

vanbasten23: I was wondering why we need to change this file.

vanbasten23: If it's not used, you can drop the change in this file.

vanbasten23:
Let me know if the pr is ready for review

PR #6846, opened 2024-03-28
[SPMD] use UNKNOWN as default sharded data sharding type only for auto-sharding

yeounoh:
Using UNKNOWN as default sharded data sharding type degraded SPMD inference performance. I added a comment to do a follow-up investigation, to see if we can completely move to Unknown for non-auto-sharding route as well.

Test plan:
- llama2 inference performance check
- auto-sharding llama2 & gemma checks

### DIFF
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -249,9 +249,10 @@ void XLATensor::SetShardingSpec(const ShardingSpec& sharding, bool overwrite) {
   // Existing annotation must be cleared explicitly. We do not clear and
   // overwrite the existing sharding on the user's behalf. This is a no-op if
   // the same sharding already applied.
-  if (!sharding_spec() || overwrite ||
-      sharding_spec()->sharding.type() == xla::OpSharding::REPLICATED ||
-      sharding_spec()->sharding.type() == xla::OpSharding::UNKNOWN) {
+  ShardingSpecPtr current_sharding = sharding_spec();
+  if (!current_sharding || overwrite ||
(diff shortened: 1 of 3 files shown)


wonjoolee95:
I can confirm that with this PR, the LLaMa 2 inference numbers are back to late 4s (~4.9s), cc @yeounoh

torch_xla/csrc/tensor.cpp
```diff
@@ -249,17 +249,18 @@ void XLATensor::SetShardingSpec(const ShardingSpec& sharding, bool overwrite) {
   // Existing annotation must be cleared explicitly. We do not clear and
   // overwrite the existing sharding on the user's behalf. This is a no-op if
   // the same sharding already applied.
-  if (!sharding_spec() || overwrite ||
-      sharding_spec()->sharding.type() == xla::OpSharding::REPLICATED ||
-      sharding_spec()->sharding.type() == xla::OpSharding::UNKNOWN) {
+  ShardingSpecPtr current_sharding = sharding_spec();
+  if (!current_sharding || overwrite ||
+      current_sharding->sharding.type() == xla::OpSharding::REPLICATED ||
+      current_sharding->sharding.type() == xla::OpSharding::UNKNOWN) {
```
JackCaoG, line 255: this is just a clean up right? I don't see any functionality change but want to double check.

JackCaoG (approved):

torch_xla/csrc/tensor.cpp
yeounoh: It's a clean-up

PR #6673, opened 2024-03-06
[Test] Fix zero1 gpu test

hgt312:
address https://github.com/pytorch/xla/issues/6260

PR #6857, opened 2024-03-29
[zero_redundancy_optimizer] 1. save master weights 2. zero1 fp32 grad…

sssrijan-amazon:
Add changes to save the sharded master weights inside optimizer because it has higher precision. When load checkpoint, we call allgather to generate full parameter and set it back so the weights in optimizer is in sync with weights in model.


jeffhataws:
This is to 2.1 AWS Neuron branch. Will have another PR for the mainline/master branch.

jeffhataws:
We are kind of flying blind with ZeRO1 test being disabled for all platforms. There's effort to revive the GPU test https://github.com/pytorch/xla/pull/6673 .

sssrijan-amazon:
Breaking this PR into two separate PRs. Ref - https://github.com/pytorch/xla/pull/6862

PR #6862, opened 2024-03-29
[ZeroRedundancyOptimizer] Save master weights

sssrijan-amazon:
This change is to save the sharded master weights inside optimizer because it has higher precision. When load checkpoint, we call all-gather
to generate full parameter and set it back so the weights in optimizer is in sync with weights in model.

### DIFF
--- a/torch_xla/distributed/zero_redundancy_optimizer.py
+++ b/torch_xla/distributed/zero_redundancy_optimizer.py
@@ -326,6 +326,14 @@ class ZeroRedundancyOptimizer(Optimizer):
     self.base_optimizer.step(closure=None, **kwargs)
     # Remove shards' grads
     self.base_optimizer.zero_grad(set_to_none=True)
+    self.allgather_weights_and_update_full_parameter()
+
+    # sync back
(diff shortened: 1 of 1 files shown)


jeffhataws:
@alanwaketan @JackCaoG  this PR has been tested internally. We will merge this to r2.1_aws_neuron and have another PR for master.

This is split from https://github.com/pytorch/xla/pull/6857 (1 of 2).

torch_xla/distributed/zero_redundancy_optimizer.py
```diff
@@ -365,16 +372,18 @@ def step(self, closure=None, **kwargs):
...
     state_dict = super().state_dict()
     base_state = self.base_optimizer.state_dict()['state']
     state_dict['base_state'] = base_state
     state_dict['shape_info'] = self.get_shape_info()
+    master_weights = []
+    for sharded_param_group in self.base_optimizer.param_groups:
+      for sharded_param in sharded_param_group['params']:
+        master_weights.append(sharded_param.data)
+
+    state_dict['sharded_master_weights'] = master_weights
```
jeffhataws, line 385: Should we turn off this change by default?

PR #6867, opened 2024-04-01
[Fori_loop|While_loop] Enable while_loop/fori_loop, Add linear/MNIST test case

ManfeiBai:
- add linear model test case with `fori_loop` or `while_loop`
- modify logic of buildforiloop to chaneg cond/body's xla_computation to meet requirement of `xla::While`: `cond's input, body's input/output and init should be the same shape`;
- modify input arguments shape/order to meet `xla::While`'s requirement mentioned above

next plan:
- `init_python_bindings.cpp`: get body xlacomputation arguments' number first then decide items in `additional_inputs_list`, maybe implement in python level


torch_xla/experimental/fori_loop.py
```diff
@@ -10,66 +10,95 @@
...
-    return lower[0] < upper[0]
+  one_value = torch.tensor([1], dtype=torch.int32, device=device)
 
-  def body_fn(upper, lower, *init_val):
-    one_value_i = torch.ones(1, dtype=torch.int32, device=device)
-    res_list = list(user_body_func(*init_val))
-    res_list.insert(0, lower)
-    res_list.insert(0, torch.sub(upper, one_value_i))
-    return res_list
+  if (hasattr(body_fun, 'weight') or hasattr(body_fun, 'bias')):
```
qihqi, line 23: what is the plan to remove this condition?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -913,14 +913,31 @@ class PyLoweringContext {
...
-      for (at::Tensor input_argument : input_arguments) {
-        xla::Shape shape =
-            xla::ShapeUtil::MakeShape(xla::PrimitiveType::S32, {1});
+      int64_t parameter_idx =
+          2;  // parameter_idx start from 2 after used upper and lower
+      for (auto& additional_input_tensor : additional_inputs_list) {
+        XLATensorPtr xtensor = bridge::GetXlaTensor(additional_input_tensor);
+        xla::Shape shape = xtensor->shape().get();
+        xla::XlaOp x = xla::Parameter(local_builder, parameter_idx, shape,
+                                      "UnusedArgumentsPlaceholder");
```
JackCaoG, line 927: I don't quite understand this part, `x` is a local variable that will be released after the for loop, is the intention of calling `xla::Parameter` to init a parameter at position `parameter_idx` with given shape? If so can you add a comment to make it more clear?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -913,14 +913,31 @@ class PyLoweringContext {
...
+    // for xla::while requriement
     if (GetNameString() == "condctx") {
       xla::XlaBuilder* local_builder = lowering_ctx.builder();
-      // hard-code parameter_idx to 2 to skip existing upper/lower arguments
-      int64_t parameter_idx = 2;
-      for (at::Tensor input_argument : input_arguments) {
-        xla::Shape shape =
-            xla::ShapeUtil::MakeShape(xla::PrimitiveType::S32, {1});
+      int64_t parameter_idx =
+          2;  // parameter_idx start from 2 after used upper and lower
```
JackCaoG, line 922: can you move the comment above the code? Thanks!

torch_xla/experimental/fori_loop.py
JackCaoG: actually why do we special casing the weight and bias? Is it only for the linear layer?

torch_xla/experimental/fori_loop.py
```diff
@@ -10,66 +10,95 @@
...
+    output_value = torch.zeros([20], dtype=torch.float32, device=device)
+
+    def cond_fn(upper, lower, one_value, x, input_value, output_value):
+      return lower[0] < upper[0]
+
+    def body_fn(upper, lower, one_value, x, input_value, output_value):
+      new_lower = torch.add(one_value, lower)
+      output_value = body_fun(input_value)
+      weight = body_fun.weight  # not be used actually, initialized as placeholder xlacomputation requirement
+      bias = body_fun.bias  # not be used actually, initialized as placeholder xlacomputation requirement
```
JackCaoG, line 33: nit, can we move the comment above the code?

torch_xla/experimental/fori_loop.py
```diff
@@ -10,66 +10,95 @@
...
+      return upper.clone(), new_lower.clone(), one_value.clone(), torch.add(
+          one_value, x), input_value.clone(), bias.clone(), weight.clone(
+          ), output_value.clone()
```
JackCaoG, line 36: why do we need to clone?

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -82,25 +86,139 @@ def body_fn(init, limit_value):
...
+      return lower[0] < upper[0]
+
+    def body_fn(upper, lower, one_value, x, input_value, output_value):
+      new_lower = torch.add(one_value, lower)
+      output_value = linear_0(input_value)
+      weight = linear_0.weight  # not be used actually, initialized as placeholder xlacomputation requirement
+      bias = linear_0.bias  # not be used actually, initialized as placeholder xlacomputation requirement
+      return upper.clone(), new_lower.clone(), one_value.clone(), torch.add(
+          one_value, x), input_value.clone(), bias.clone(), weight.clone(
+          ), output_value.clone()
```
JackCaoG, line 107: remind me why do we need to return weight and bias in this function?

torch_xla/experimental/fori_loop.py
ManfeiBai: miss `.clone()` here would cause PyTorch ERROR: [`torch.while_loop's body_fn might be aliasing the input!`](https://github.com/pytorch/pytorch/blob/ca6a0e1348ba7dcade1833d983b1b4ca12a5c1e1/torch/_dynamo/variables/higher_order_ops.py#L118), so add `.clone()` to avoid return input arguments directly

torch_xla/experimental/fori_loop.py
ManfeiBai: yes, its only for the linear layer, and plan to remove this condition is: only check whether `additional_inputs`(weight/bias) exist for code of weight/bias in the next step

torch_xla/csrc/init_python_bindings.cpp
ManfeiBai: yes, x was added here to add expected shape/type arguments in body/cond xlacomputation's arguments, to meet `XLA::While` requirement: `parameter of condition and body, the result of the body, and init must all have the same shape`

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: we need to make sure body_fn's xlacomputation's input and output are the same, because input would include weight automatically, so here we return weight and bias from python level to ensure weight and bias are included in ouput too. Add bias to avoid output_value is used as bias in calculation, because bias has the same shape and value as output_value

JackCaoG: This is too confusing, we need to think of a better UX. body_fn also should take `linear_0` as an input instead of calling it from parent scope. 

I think instead of manually return the each parameter in the module, we should just return module's `named_paramter` . User also shouldn't need to manually order the return parameter(this will be super confusing to the user), we should do it in our layer.

miladm:
do we have an ETA for this PR?

JackCaoG:
I felt like the way we determine the HLO input parameter order is not ideal, it only works on simple examples. In this pr Manfei is trying to support Linear and she need to do some hacks to make the order correct. I am trying to figure out if there is a more general way to determine the parameter order, hopefully I can find some time today and tmr to draft a poc out..

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -82,25 +86,139 @@ def body_fn(init, limit_value):
...
+    def cond_fn(upper, lower, one_value, x, input_value, output_value):
+      return lower[0] < upper[0]
+
+    def body_fn(upper, lower, one_value, x, input_value, output_value):
+      new_lower = torch.add(one_value, lower)
+      output_value = linear_0(input_value)
+      weight = linear_0.weight  # not be used actually, initialized as placeholder xlacomputation requirement
+      bias = linear_0.bias  # not be used actually, initialized as placeholder xlacomputation requirement
+      return upper.clone(), new_lower.clone(), one_value.clone(), torch.add(
+          one_value, x), input_value.clone(), bias.clone(), weight.clone(
```
JackCaoG, line 106: why are we returning a `torch.add(one_value, x)` here?

ManfeiBai: It was used here to confirm calculation run expected times as a timer

JackCaoG: I think our test case is too complicated, we should aim to support what pytorch support, similar to https://github.com/pytorch/pytorch/blob/8573d9551a7694b9313310412867ac3b6b751f26/test/functorch/test_control_flow.py#L137-L150.

ManfeiBai:
> I felt like the way we determine the HLO input parameter order is not ideal, it only works on simple examples. In this pr Manfei is trying to support Linear and she need to do some hacks to make the order correct. I am trying to figure out if there is a more general way to determine the parameter order, hopefully I can find some time today and tmr to draft a poc out..

Thanks, @JackCaoG, for comparison, this WIP PR is also trying post-order like `XLAGraphExecutor::Compile` according to your idea and suggestion: https://github.com/pytorch/xla/pull/7031, then add python interface for while_loop to generate xlacomputation for cond/body

JackCaoG:
I have some high level ideas

as long as there is no inplace update to any parameters(which will not be true I guess since we need to decrement the iterator), get_hlo(input1, input2, input3, output) should have parameter in the same order. This is because in order to compute the new `input1`, we need `input1` as result. The compilation of this approach is if somehow `input1` is updated, for example
```
fn(input1, input2, input3, output):
  input1 += input2
  ...
```
`get_hlo(input1, input2, input3, output)` will likely result in the parameter order of `input2, input1, input3, output`. The key is always put all inputs in front of the output.

The same rule should apply for the named_parameters from the `module` , if they are being passed as `additional_input` from the while loop, we just need to stich them to inputs list(not from user code level, in our while loop implementation level). Hopefully this will just work..

ManfeiBai:
thanks for th

Issue #6864, opened 2024-04-01
torchbench hf_* models fail on both TPU and GPU


zpcore:
## 🐛 Bug

Torchbench models like hf_Albert, hf_Bart, hf_Bert, hf_Bert_large, hf_BigBird, hf_DistilBert, hf_GPT2, hf_GPT2_large, hf_Longformer, hf_Reformer, hf_T5, hf_T5_base, hf_T5_generate, hf_T5_large all failed recently wit hte 

## To Reproduce

1.
```
cd /tmp/ && git clone https://github.com/pytorch/benchmark.git
cd /tmp/ && git clone https://github.com/pytorch/pytorch.git
cd /tmp/ && git clone https://github.com/pytorch/xla.git
```
2. 
```
cd benchmark && python install.py models hf_Bert
```
3. 
```
python xla/benchmarks/experiment_runner.py \
--suite-name torchbench    \
--accelerator cuda  \
--xla PJRT   \
--test eval  \
--repeat 2  \
--iterations-per-run 1   \
--print-subprocess \
--no-resume -k hf_Bert
```

Error log:
1. with dynamo backend:
```
ERROR:__main__:ERROR in subprocess
INFO:__main__:Run with --model-config={"model_name": "hf_Bert_large"} --experiment-config={"accelerator": "cuda", "xla": "PJRT", "xla_flags": null, "dynamo": "openxla", "test": "eval"}
WARNING:__main__:Enabling fast F32 multiplication for PyTorch
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1711990897.946354    2258 service.cc:145] XLA service 0x5612ea54bb40 initialized for platform CUDA (this does not guarantee that XLA will be used). Devices:
I0000 00:00:1711990897.946448    2258 service.cc:153]   StreamExecutor device (0): NVIDIA A100-SXM4-40GB, Compute Capability 8.0
I0000 00:00:1711990897.947022    2258 se_gpu_pjrt_client.cc:853] Using BFC allocator.
I0000 00:00:1711990897.947102    2258 gpu_helpers.cc:107] XLA backend allocating 31724126208 bytes on device 0 for BFCAllocator.
I0000 00:00:1711990897.947133    2258 gpu_helpers.cc:147] XLA backend will use up to 10574708736 bytes on device 0 for CollectiveBFCAllocator.
INFO:benchmark_model:Running torch.compile with opts {'backend': 'openxla'}
Traceback (most recent call last):
[...]

PR #6869, opened 2024-04-01
format the output model input

zpcore:
Fix https://github.com/pytorch/xla/issues/6864

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -264,7 +264,8 @@ class TorchBenchModel(BenchmarkModel):
     benchmark = self.load_benchmark()
 
     self.module, self.example_inputs = benchmark.get_module()
-    if isinstance(self.example_inputs, dict) and "input_ids" in self.example_inputs:
+    if isinstance(self.example_inputs,
+                  dict) and "input_ids" in self.example_inputs:
       self.example_inputs = (self.example_inputs['input_ids'],)
     self.benchmark_experiment.batch_size = benchmark.batch_size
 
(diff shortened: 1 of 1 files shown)


ysiraichi:
Does this work for all the models in the issue? I'd imagine we'd have to do something similar to [their old implementation](https://github.com/pytorch/benchmark/blob/43866040439cb9a2d1c5508c72609dce7250fb2d/torchbenchmark/util/framework/huggingface/model_factory.py#L140-L148)

benchmarks/torchbench_model.py
```diff
@@ -264,7 +264,9 @@ def set_up(self):
     benchmark = self.load_benchmark()
 
     self.module, self.example_inputs = benchmark.get_module()
-
+    if isinstance(self.example_inputs,
```
ysiraichi, line 267: Leave a comment here, mentioning the pytorch/benchmark PR. It would be good to know in the future.

ysiraichi:
I have a few more questions:

1. Would this also work with older versions of pytorch/benchmark? 
2. What's the policy when a breaking change is merged on pytorch/benchmark's main branch? Do we follow:
    - the main branch
    - the pinned commit found in the main branch of pytorch/pytorch
    - some other pinned commit

zpcore:
Thanks for pointing this out!

> Would this also work with older versions of pytorch/benchmark?

It should work. Currently only the hf models inputs have been changed into the dict type.

> What's the policy when a breaking change is merged on pytorch/benchmark's main branch?

I don't have a good solution at this moment, but I expect the changes like regarding the input type should be rare. I prefer sync up to the latest pytorch/benchmark repo at this moment. For the long term, once your proposal of merging the pytorch/xla/benchmarks is landed and implemented, we should get rid of this issue with their CI.

zpcore:
Hello, pin to see if someone can make an approval.

vanbasten23 (approved):
Thanks for the fix!

zpcore:
Hi @ysiraichi , I will merge the change. You may need to update the python/benchmark PIN to the head when you run torchbench or error may occur for those HF models.

PR #6870, opened 2024-04-02
[Pallas] Support Flash Attention backward kernels

alanwaketan:
Summary:
This changes refactors custom_kernel.py to support all three new kernels from Pallas that are involved in Flash Attention backward calculations.

The refactoring includes:
1. Adds support for static_argnums which will ignore some positional arguments for jax tracing.
2. Separate jax tracing part out such that we can do the tracing alone.

Test Plan:

### DIFF
--- a/test/test_pallas.py
+++ b/test/test_pallas.py
@@ -294,15 +294,31 @@ class PallasTest(unittest.TestCase):
     def shape_dtype(q, *arg):
       res_shape = list(q.shape)
       res_shape[-1] = MIN_BLOCK_SIZE
-      return [(q.shape, q.dtype), (res_shape, torch.float32), (res_shape, torch.float32)]
-    flash_attention_kernel = make_kernel_from_pallas(
-        _flash_attention_impl, shape_dtype)
+      return [(q.shape, q.dtype), (res_shape, torch.float32),
+              (res_shape, torch.float32)]
+
+    flash_attention_kernel = make_kernel_from_pallas(_flash_attention_impl,
+                                                     shape_dtype)
 
     q = torch.randn(3, 2, 128, 4, dtype=torch.bfloat16).to("xla")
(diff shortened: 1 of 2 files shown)


test/test_pallas.py
```diff
@@ -283,6 +284,158 @@ def add_minus_vectors(x: jax.Array, y: jax.Array) -> jax.Array:
...
+    grad_i = grad_i.unsqueeze(-1).expand(3, 2, 128, MIN_BLOCK_SIZE)
+    grad_q = torch.randn(3, 2, 128, 4).to("xla")
+    torch_xla._XLAC._xla_tpu_custom_call_([grad_q],
+                                          [q, k, v, l, m, grad_o, grad_i],
+                                          payload)
+
+    xm.mark_step()
+
+    # TODO: I don't really know how to test the value. Let's do the shape check for now.
+    self.assertEqual(grad_q.shape, (3, 2, 128, 4))
```
JackCaoG, line 437: if we do the fwd and do `res.backward` then check the grad on q they should match?

alanwaketan: The softmax is done differently. I don't think there is any guarantees.

JackCaoG: result should still be somewhat close right? we can tune down the precision. If the result return by this is dramatically different than the one that was computed using dot attention that seems wrong..

alanwaketan: https://pytorch.org/docs/stable/generated/torch.nn.Softmax.html

Softmax requires all the elements to produce the results, but flash attention chunks the data into blocks and use a technique called tiling to make sure the softmax still serve the functionality to stable the data. Since there are no aggregation, I don't know how the tiling softmax could produce the same results as the regular one.

In JAX, I have to use `atol=1e-01, rtol=1e-01` to do the comparisons...

JackCaoG (approved):

alanwaketan:
Thanks, Jack!

PR #6873, opened 2024-04-02
Fix div lowering and core aten test script enhancement

lsy323:
- Fix a shape inf bug in div lowering
- Enhance aten test script. `torch.allclose` would broadcast the operands (e.g. `torch.allclose([3, 1], [4])` would broadcast both operands to `[3,4]` and then compare). Add shape check before `torch.allclose`


JackCaoG (approved):
@vanbasten23 as you are thinking about doing the same thing. We should also cherry-pick this into 2.3 branch as this is a regression.

wonjoolee95 (approved):

Issue #6854, opened 2024-03-29
[torchxla2] Tensor constructor capturing


qihqi:
## 🚀 Feature

Create lowerings for tensor creation ops so that 

```python
with XLADispatchMode():
   x = create_a_tensor()
   print(x)
```

outputs XLATensor2 instead of regular CPU tensor.

This requires implementing tensor creation ops (https://pytorch.org/docs/stable/torch.html#creation-ops) and add their override
in XLADispatchMode (https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/torch_xla2/tensor.py#L16)

## Additional context

The most important one is `torch.tensor` that takes a Python list and creates a tensor. This is important because it indexing using list will be converted to tensor using it.

i.e. 

```python

x = XLATensor2(...)
y = x[:, [0, 2]] 
```
the second line will call `torch.tensor` to convert `[0, 2]` to tensor first.

`torch.tensor` is not torch dipatch capturable, it need to be captured using `__torch_function__`

So it might be a good idea to also create a XLAFuntionMode() and another context manager that enables/disables both the function mode and dispatcher mode.

The tensor constructors for sparse are not needed ATM.

PR #6874, opened 2024-04-02
[WIP] [torch_xla2] Implement some common tensor constructors

will-cromar:
- Implement some common tensor constructors with similar semantics between JAX and PyTorch
- Add `XLAFunctionMode` to dispatch to JAX implementations

TODO

- [ ] Finish adding unit tests
- [ ] Don't ignore dtypes
- [ ] Wrap `XLADispatchMode` and `XLAFunctionMode` together (maybe a future PR)

#6854

cc @qihqi

### DIFF
--- a/experimental/torch_xla2/test/test_functions.py
+++ b/experimental/torch_xla2/test/test_functions.py
@@ -1,0 +1,26 @@
+from absl.testing import absltest
+from absl.testing import parameterized
+import torch
+import torch_xla2
+import torch_xla2.functions
+import torch_xla2.tensor
+
+class TestTorchFunctions(parameterized.TestCase):
+  @parameterized.parameters(
+    [([[0.1, 1.2], [2.2, 3.1], [4.9, 5.2]],)],
+    [([0, 1],)],
+    [(3.14159,)],
+    [([],)],
+  )
+  def test_tensor(self, args):
+    expected = torch.tensor(*args)
+
+    with torch_xla2.functions.XLAFunctionMode():
+      actual = torch.tensor(*args)
+
+    # TODO: dtype is actually important
+    torch.testing.assert_close(torch_xla2.tensor.j2t(actual), expected, check_dtype=False)
+
+
+if __name__ == '__main__':
+  absltest.main()
--- a/experimental/torch_xla2/torch_xla2/functions.py
+++ b/experimental/torch_xla2/torch_xla2/functions.py
@@ -1,0 +1,28 @@
+"""Tensor constructor overrides"""
+import warnings
+
+import torch
+import jax.numpy as jnp
+
+fns = {
+  torch.tensor: jnp.array,
+  # torch.ones: jnp.ones,
+  # torch.zeros: jnp.zeros,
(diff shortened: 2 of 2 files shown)


experimental/torch_xla2/torch_xla2/functions.py
```diff
@@ -0,0 +1,29 @@
...
+  # torch.logspace: jnp.logspace,
+  # torch.empty: jnp.empty,
+  # torch.eye: jnp.eye,
+  # torch.full: jnp.full,
+}
+
+class XLAFunctionMode(torch.overrides.TorchFunctionMode):
+  def __torch_function__(self, func, types, args=(), kwargs=None):
+    jax_func = fns.get(func)
+    if not jax_func:
```
qihqi, line 23: if jax func dont exist, should just fall through with `return func(*args, **kwargs)`

experimental/torch_xla2/torch_xla2/functions.py
```diff
@@ -0,0 +1,29 @@
...
+      raise NotImplementedError(f'No jax function found for {func.__name__}')
+
+    if kwargs:
+      warnings.warn(f'kwargs not implemented for {kwargs}')
+
+    return tensor.wrap(jax_func(*args))
```
qihqi, line 29: args need to be unpacked with `args = tensor.unwrap(args)`. for tensor constructors it doesn't matter because likely it's input are not other tensors; for other torch functions it does

experimental/torch_xla2/test/test_functions.py
```diff
@@ -0,0 +1,26 @@
...
+    [([],)],
+  )
+  def test_tensor(self, args):
+    expected = torch.tensor(*args)
+
+    with torch_xla2.functions.XLAFunctionMode():
+      actual = torch.tensor(*args)
+
+    # TODO: dtype is actually important
+    torch.testing.assert_close(torch_xla2.tensor.j2t(actual), expected, check_dtype=False)
```
qihqi, line 22: `j2t(actual._elem)` so `actual` would be of type `XLATensor2`. and `actual._elem` is the jax array

will-cromar: What's the difference? I tried it on a basic example and got the same result:

```
>>> x
XLATensor2(..., device='meta', size=(2,))
>>> torch_xla2.tensor.j2t(x)
tensor([1, 2], dtype=torch.int32)
>>> torch_xla2.tensor.j2t(x._elem)
tensor([1, 2], dtype=torch.int32)
```

experimental/torch_xla2/torch_xla2/functions.py
will-cromar: Looking more closely, many of the constructors are explicitly `int`s and not another tensor. So in the function dispatcher, I think we can leave the inputs alone, and then I can add a decorator like `@convert_dtype` to handle any cases where we can end up taking other `Tensor`s.

examples: https://pytorch.org/docs/stable/generated/torch.full.html and https://pytorch.org/docs/stable/generated/torch.ones.html

PR #6875, opened 2024-04-02
Add flash attention as custom op so dynamo can trace it

JackCaoG:


torch_xla/experimental/custom_kernel.py
```diff
@@ -183,3 +185,34 @@ def flash_attention(
...
+                        v: torch.Tensor,
+                        causal: bool = False):
+  return flash_attention(q, k, v, causal=causal)
+
+
+@impl(XLA_LIB, "flash_attention", "CompositeExplicitAutograd")
+def flash_attention_non_xla(q: torch.Tensor,
+                            k: torch.Tensor,
+                            v: torch.Tensor,
+                            causal: bool = False):
```
JackCaoG, line 207: I didn't add `segment_ids ` as it is not implemented in `flash_attention`. Maybe I should add it here... @alanwaketan is `flash_attention` also a `torch.tensor`?

alanwaketan: I don't think we need to support segment_ids. We can safely ignore it. I will drop the comment and the parameter later.

JackCaoG:
@alanwaketan can I get a review for this one?

alanwaketan (approved):
LGTM.

PR #6876, opened 2024-04-02
[Backport] Set the major-minor layout for inputs and outputs

alanwaketan:
Summary:
Mosaic only accepts major-minor layout for both its inputs and outputs. So we need to enforce those layouts by setting the expected input&output shapes in xla::CustomCallWithLayout. After this change, XLA_TPU_LAYOUT is no longer needed.

Test Plan:


alanwaketan:
This should be the last fix for the Pallas to back port.

JackCaoG (approved):

lsy323 (approved):

PR #6797, opened 2024-03-21
Update XLA pin 

yeounoh:
Update XLA pin  to `b9ece5154181608ff1acea9b02a1ee7a526af16b` to incorporate auto-sharding patch.

PR #6879, opened 2024-04-03
Update xla pin to Apr 3

bhavya01:


JackCaoG:
LGTM, @bhavya01 do you mind do a quick sanity check on TPU? We usually compile it locally, and run reset with fake data with `python test/test_train_mp_imagenet.py --fake_data`
On TPUv4 you should see speed similar to https://github.com/pytorch/xla/pull/6797#issuecomment-2015431044

This is to make sure we don't have any noticeable speed regression for the pin update.

setup.py
```diff
@@ -64,7 +64,7 @@
 
 base_dir = os.path.dirname(os.path.abspath(__file__))
 
-_date = '20240322'
+_date = '20240403'
```
JackCaoG, line 67: if you use a commit from 04/03, we should use the libtpu date at 04/04 because the libtpu was build midnight.

bhavya01: Sounds good. Updated.

Resnet numbers still look good
```
 Training Device=xla:0/0 Epoch=1 Step=1340 Loss=0.00141 Rate=1769.13 GlobalRate=923.65 Time=20:39:43
| Training Device=xla:0/1 Epoch=1 Step=1360 Loss=0.00140 Rate=1769.17 GlobalRate=938.59 Time=20:39:45
| Training Device=xla:0/3 Epoch=1 Step=1360 Loss=0.00140 Rate=1769.14 GlobalRate=934.33 Time=20:39:45
| Training Device=xla:0/2 Epoch=1 Step=1360 Loss=0.00140 Rate=1769.13 GlobalRate=936.62 Time=20:39:45
```

JackCaoG (approved):

JackCaoG:
Thanks! you can merge the pr after CI is green.

PR #6881, opened 2024-04-03
[benchmarks] Use Torchbench YAML configuration data.

ysiraichi:
This PR makes the benchmarking scripts to start using the configuration YAML file exposed by PyTorch. Previously, we were using it only for checking models that should be skipped. From this PR onwards, we are using the configuration file for other things, such as:

- Deciding what should be the data-type
- Checking whether a model is a Detectron2 model

cc @miladm


zpcore:
Thanks! Logics look good to me. Can you move function `config_data()` into the front of the file?

zpcore (approved):

vanbasten23 (approved):
Thanks!

PR #6882, opened 2024-04-03
Check pointer when we cast  PjRtLayout to PjRtXlaLayout

vanbasten23:
This PR checks if the `cast` is safe by following the original XLA change https://github.com/openxla/xla/commit/b8109e2f697338e8517d110014ce2160444aa927.

The pin update https://github.com/pytorch/xla/pull/6677 made a change https://github.com/pytorch/xla/blob/895b0c2ea14ae1d5c5daf76056de64df28d1b6f2/torch_xla/csrc/runtime/pjrt_computation_client.cc#L60-L62 which is caused by an XLA change https://github.com/openxla/xla/commit/b8109e2f697338e8517d110014ce2160444aa927. The `GetXlaLayoutUnsafe` is safer because it checks if `tensorflow::down_cast` succeeds and fails early: https://github.com/openxla/xla/blob/956f0ed40f4144a83f9e5aa3a0fe75ab94a81b70/xla/pjrt/pjrt_layout.h#L101-L107

It's unclear it caused the regression but it makes our change consistent with open XLA and it's safer as `dynamic_cast` may fail (return nullptr)

### DIFF
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -57,9 +57,9 @@ std::unordered_map<int, int> build_index_map(
 xla::Shape host_output_shape(xla::PjRtBuffer* buffer) {
   xla::Shape shape = xla::ShapeUtil::MakeShape(
       buffer->element_type(), buffer->logical_dimensions().value());
-  *shape.mutable_layout() =
-      dynamic_cast<const xla::PjRtXlaLayout*>(buffer->layout().get())
-          ->xla_layout();
+  *shape.mutable_layout() = xla::GetXlaLayoutUnsafe(buffer->layout());
+  //     dynamic_cast<const xla::PjRtXlaLayout*>(buffer->layout().get())
+  //         ->xla_layout();
 
   return xla::ShapeUtil::DeviceShapeToHostShape(shape);
 }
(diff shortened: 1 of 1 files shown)


will-cromar (approved):

will-cromar:
Thanks for the fix. From the comments, it looks like the function is temporary. Do you know how to use `PjRtLayout` directly in our code? I don't love that it's embedded in `xla::Shape` (but not always), but I also don't know how to get away from `xla::Shape`.

vanbasten23:
> Thanks for the fix. From the comments, it looks like the function is temporary. Do you know how to use `PjRtLayout` directly in our code? 

I haven't studied it yet. But I am debugging why we get the error `malloc(): invalid next size (unsorted)` and this change looks suspicious to me.

Also, I'm suspicious of the original change https://github.com/openxla/xla/commit/b8109e2f697338e8517d110014ce2160444aa927, especially this part:
```
inline Layout GetXlaLayoutUnsafe(
    const std::unique_ptr<PjRtLayout>& pjrt_layout) {
  PjRtXlaLayout* xla_layout =
      tensorflow::down_cast<PjRtXlaLayout*>(pjrt_layout.get());
  CHECK(xla_layout != nullptr) << "Got unexpected layout type";
  return xla_layout->xla_layout();
}
```
`tensorflow::down_cast` uses `static_cast` internally. Per [static_cast](https://learn.microsoft.com/en-us/cpp/cpp/static-cast-operator?view=msvc-170), even if the `static_cast` fails, the resulting `xla_laybout` may be an invalid pointer but it may not be nullptr. So I'm kind of curious why the author doesn't use dynamic_cast and then check for nullptr.

lsy323:
I think the `GetXlaLayoutUnsafe` is doing the same thing as the current code, except the nullptr check.

Since `GetXlaLayoutUnsafe` is temporary according to the comment, adding an assertion after the `dynamic_cast` seems to be cleaner to me.

vanbasten23:
> I think the `GetXlaLayoutUnsafe` is doing the same thing as the current code, except the nullptr check.
> 
> Since `GetXlaLayoutUnsafe` is temporary according to the comment, adding an assertion after the `dynamic_cast` seems to be cleaner to me.

Yeah, I think the correct thing is to do dynamic_cast followed by a nullptr check due to the reason in https://github.com/pytorch/xla/pull/6882#issuecomment-2035342708.

PR #6884, opened 2024-04-04
[Backport] Add flash attention as custom op so dynamo can trace it (#6875)

JackCaoG:
We intended to add dynamo support for pallas but flash attention support needs to be added.


alanwaketan (approved):
LGTM.

will-cromar:
Can you just hardcode the release branch on this branch?

https://github.com/pytorch/xla/blob/7520c04691360ee30a211a61b97ad429890598af/.github/workflows/tpu_ci.yml#L24

I can update master to use the ansible config so we don't have to do stuff like this twice.

PR #6885, opened 2024-04-04
Don't bundle libtpu in nightly

alanwaketan:
Summary:
Pallas depends on JAX to run, and JAX by default will use the system libtpu and won't use our own bundled libtpu. In order to have JAX and PyTorch/XLA run side by side, we will need to have them use the same libtpu. The solution could either be setting the TPU_LIBRARY_PATH or having both of them use the same system libtpu. The latter yields a much better UX.

Test Plan:
N.A.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,16 +207,16 @@ You can also add `+yyyymmdd` after `torch_xla-nightly` to get the nightly wheel
 of a specified date. To get the companion pytorch and torchvision nightly wheel,
 replace the `torch_xla` with `torch` or `torchvision` on above wheel links.
 
-#### Installing libtpu (before PyTorch/XLA 2.0)
+#### Installing libtpu
 
-For PyTorch/XLA release r2.0 and older and when developing PyTorch/XLA, install
-the `libtpu` pip package with the following command:
+Install the `libtpu` pip package with the following command:
 
 ```
 pip3 install torch_xla[tpuvm]
 ```
 
 This is only required on Cloud TPU VMs.
+On Kaggle, the libtpu will be bundled in torch-xla installed.
 
 </details>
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -207,16 +207,16 @@ You can also add `+yyyymmdd` after `torch_xla-nightly` to get the nightly wheel
 of a specified date. To get the companion pytorch and torchvision nightly wheel,
 replace the `torch_xla` with `torch` or `torchvision` on above wheel links.
 
-#### Installing libtpu (before PyTorch/XLA 2.0)
+#### Installing libtpu
```
will-cromar, line 210: Can you just delete this section? There are more up to date instructions to install libtpu up top.

will-cromar (approved):
> Pallas depends on JAX to run, and JAX by default will use the system libtpu and won't use our own bundled libtpu. In order to have JAX and PyTorch/XLA run side by side, we will need to have them use the same libtpu. The solution could either be setting the TPU_LIBRARY_PATH or having both of them use the same system libtpu. The latter yields a much better UX.

It's worth noting this is a trade-off. To have release versions of JAX and PyTorch installed side-by-side, you must use the bundled libtpu in PyTorch/XLA, unless by some coincidence the two releases depend on the same `libtpu-nightly` version. This is why we use the bundled version in Colab/Kaggle.

Conversely to install JAX as a _dependency_ of PyTorch/XLA (rather than having both side-by-side), you have to use the exact nightly version we specify that matches the nightly version of libtpu we depend on.

Have you already tried overriding just JAX's libtpu path with `TPU_LIBRARY_PATH` before the import?

```
os.environ['TPU_LIBRARY_PATH'] = ...
import jax
```

Unbundling by default SGTM even if that works since we did this for release builds. This would just potentially make pallas possible in Colab/Kaggle. Either way, there's going to be some form of mess or hackery unless pallas can be separated out of JAX, or we can align with JAX on a common release schedule for libtpu.

alanwaketan:
@will-cromar Thanks for approving the change, yea, this works:
```
os.environ['TPU_LIBRARY_PATH'] = ...
```

README.md
alanwaketan: Done.

PR #6886, opened 2024-04-04
[Pallas] Make FlashAttention as torch.autograd.Function

alanwaketan:
Summary:
This pull request makes the flash attention kernel as a torch.autograd.Function such that we can enable backward on the kernel.

Test Plan:
PJRT_DEVICE=TPU python test/test_pallas.py -v -k test_flash_attention_backward

### DIFF
--- a/test/test_pallas.py
+++ b/test/test_pallas.py
@@ -449,9 +449,9 @@ class PallasTest(unittest.TestCase):
     from torch_xla.experimental.custom_kernel import flash_attention
 
     torch.manual_seed(42)
-    q = torch.randn(3, 2, 128, 4, requires_grad=True).to("xla")
-    k = torch.randn(3, 2, 128, 4, requires_grad=True).to("xla")
-    v = torch.randn(3, 2, 128, 4, requires_grad=True).to("xla")
+    q = torch.randn(4, 2, 128, 8, requires_grad=True).to("xla")
+    k = torch.randn(4, 2, 128, 8, requires_grad=True).to("xla")
+    v = torch.randn(4, 2, 128, 8, requires_grad=True).to("xla")
     q.retain_grad()
     k.retain_grad()
     v.retain_grad()
@@ -466,9 +466,9 @@ class PallasTest(unittest.TestCase):
     v_grad = v.grad
 
     torch.manual_seed(42)
-    q = torch.randn(3, 2, 128, 4, requires_grad=True).to("xla")
-    k = torch.randn(3, 2, 128, 4, requires_grad=True).to("xla")
-    v = torch.randn(3, 2, 128, 4, requires_grad=True).to("xla")
+    q = torch.randn(4, 2, 128, 8, requires_grad=True).to("xla")
+    k = torch.randn(4, 2, 128, 8, requires_grad=True).to("xla")
+    v = torch.randn(4, 2, 128, 8, requires_grad=True).to("xla")
     q.retain_grad()
     k.retain_grad()
     v.retain_grad()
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/custom_kernel.py
```diff
@@ -160,47 +160,184 @@ def wrapped_kernel(kernel: Callable,
...
+      l, m = (v[..., 0] for v in aux[-2:])
+
+    ctx.save_for_backward(q, k, v, o, l, m)
+    return o
+
+  @staticmethod
+  def backward(ctx, grad_output):
+    # Import JAX within the function such that we don't need to call the jax_import_guard()
+    # in the global scope which could cause problems for xmp.spawn.
+    jax_import_guard()
```
JackCaoG, line 239: we shouldn't need this right? The forward should have been called at this point and fwd/bwd happens in the same process.

torch_xla/experimental/custom_kernel.py
```diff
@@ -160,47 +160,184 @@ def wrapped_kernel(kernel: Callable,
...
+                      k.shape[2]),
+          sm_scale=1.0,
+          causal=causal,
+          mask_value=FlashAttention.DEFAULT_MASK_VALUE,
+          debug=False,
+          static_argnames=[
+              "block_q_major", "block_k_major", "block_k", "sm_scale", "causal",
+              "mask_value", "debug"
+          ])
+      grad_q = torch.empty(q.shape, dtype=q.dtype).to(xm.xla_device())
```
JackCaoG, line 284: it should be fine for the most part, but I think it is better to do `.to(q.deivce)` instead of move to `xm.xla_device()`. Should we add a check somewhere to make sure all tensors on XLA device?

torch_xla/experimental/custom_kernel.py
```diff
@@ -160,47 +160,184 @@ def wrapped_kernel(kernel: Callable,
...
+
+    q, k, v, o, l, m = ctx.saved_tensors
+    causal = ctx.causal
+    grad_q = grad_k = grad_v = None
+
+    grad_i = torch.sum(
+        o.to(torch.float32) * grad_output.to(torch.float32),
+        axis=-1)  # [batch_size, num_heads, q_seq_len]
+
+    expanded_l = l.unsqueeze(-1).expand(3, 2, 128,
```
alanwaketan, line 250: Oops, I shouldn't hardcode the shape...

test/test_pallas.py
```diff
@@ -467,6 +449,45 @@ def test__flash_attention_bwd_dkv(self):
...
+    v.retain_grad()
+
+    o = self._attention(q, k, v)
+    loss = o.sum()
+    loss.backward()
+    xm.mark_step()
+
+    mse = torch.nn.MSELoss()
+    for i in [(q, q_grad), (k, k_grad), (v, v_grad)]:
+      self.assertTrue(mse(i[0].grad.cpu(), i[1].cpu()) < 1e-4)
```
JackCaoG, line 489: I am not sure what this part is checking, do you mind explaining a bit?

alanwaketan: Since the gradients are a little bit off, it's hard to use torch.allclose. I'm just trying to use MSE to calculate the difference to see if it's close to zero.

torch_xla/experimental/custom_kernel.py
alanwaketan: Right!

torch_xla/experimental/custom_kernel.py
alanwaketan: Fixed the .to(). For the second question, I guess Mosaic already guard it?

JackCaoG (approved):

PR #6889, opened 2024-04-04
Adjust allclose atol for the flash attention TPU test

JackCaoG:
This should fix the TPU CI Failure, confirmed locally.

### DIFF
--- a/test/test_pallas.py
+++ b/test/test_pallas.py
@@ -205,7 +205,7 @@ class PallasTest(unittest.TestCase):
 
     o = flash_attention(q, k, v)
     expected_o = self._attention(q, k, v)
-    self.assertTrue(torch.allclose(o.cpu(), expected_o.cpu()))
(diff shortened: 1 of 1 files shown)


test/test_pallas.py
```diff
@@ -205,7 +205,7 @@ def test_flash_attention_wrapper(self):
 
     o = flash_attention(q, k, v)
     expected_o = self._attention(q, k, v)
-    self.assertTrue(torch.allclose(o.cpu(), expected_o.cpu()))
+    self.assertTrue(torch.allclose(o.cpu(), expected_o.cpu(), atol=1e-07))
```
alanwaketan, line 208: How about using 1e-4? Just in case.

JackCaoG: let's do 1e-5 lol, we also set the precision to highest in the test.

alanwaketan (approved):
LGTM.

JackCaoG:
TPU CI passed, I am just going to merge.

PR #6890, opened 2024-04-04
Manually init absl log to avoid log spam

JackCaoG:
This should get rid of
```
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1712267574.801377  681831 pjrt_api.cc:100] GetPjrtApi was found for tpu at /usr/local/lib/python3.10/site-packages/libtpu/libtpu.so
I0000 00:00:1712267574.801477  681831 pjrt_api.cc:79] PJRT_Api is set for device type tpu
I0000 00:00:1712267574.801484  681831 pjrt_api.cc:146] The PJRT plugin has PJRT API version 0.46. The framework PJRT API version is 0.46.
```
on every init. `torch_xla/csrc/BUILD` was messed by the formater, my only change was to add `"@com_google_absl//absl/log:initialize",`

### DIFF
--- a/torch_xla/csrc/runtime/BUILD
+++ b/torch_xla/csrc/runtime/BUILD
@@ -2,15 +2,9 @@ load(
     "//bazel:rules_def.bzl",
     "ptxla_cc_test",
 )
-
 load(
     "@tsl//tsl/platform/default:cuda_build_defs.bzl",
     "if_cuda_is_configured",
-)
-
-load(
-    "//bazel:rules_def.bzl",
-    "ptxla_cc_test",
 )
 
 licenses(["notice"])  # Apache 2.0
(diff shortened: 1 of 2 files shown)


JackCaoG:
@lsy323 I consider this as a regression, since in 2.0 we fixed this issue once. I am debating whether we should backport this to 2.3. What do you think?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2206,6 +2207,7 @@ void InitXlaModuleBindings(py::module m) {
            !xtensor->CurrentDataHandle()->HasValue();
   });
   m.def("_init_xla_lazy_backend", []() {
+    absl::InitializeLog();
```
will-cromar, line 2210: Hmmm, I'm not confident about this change. IIRC this function errors out if it's called twice, and it is for sure called in the GPU plugin: https://github.com/openxla/xla/blob/df104c0963bf1597358dc759da3fa3d2639de03c/xla/pjrt/c/pjrt_c_api_gpu.cc#L28

If you're considering a backport, a more conservative change would be to add the initialization for TPU only sine we're sure `libtpu` does not init logging: https://github.com/pytorch/xla/blob/0b9513cb15f812252493c362ab67862bd664a54b/torch_xla/csrc/runtime/pjrt_registry.cc#L111

JackCaoG: sg, will update

will-cromar (approved):
What tool did you use to format the bazel build file and should I be using it?

JackCaoG:
> What tool did you use to format the bazel build file and should I be using it?

I need to find out haha, I was just using the vscode with our dev container and the vscode decided to format the bazel file.

PR #6891, opened 2024-04-05
add torch_distributed tests to our TPU CI

JackCaoG:
`get_xla_supported_devices ` will deprecate `devkind ` after 2.3 so remove it to avoid warning for all mp tests.

@will-cromar I think now all mp workload will throw this warning since they will all call `xla_replication_devices`. Do you think we should just fix that in the release branch?


will-cromar (approved):
Yeah, please fix this line in the release branch if it creates log spam. `devkind` does nothing

JackCaoG:
TPU CI is broken right now, I will hold on merging this pr.

PR #6897, opened 2024-04-06
[Pallas] Remove torch.empty in tracing

alanwaketan:
Summary:
Previously we rely on torch.empty to create some empty tensors as the outputs from the Pallas and then make Pallas as in-place ops. However, it turns out that torch.empty will actually allocate memory and therefore it's expansive to use. In this change, I switched to simply pass the shapes and dtypes to construct the graph node.

Test Plan:

Performance benchmarks can be found: http://shortn/_wdmom7I6q7

### DIFF
--- a/test/test_pallas.py
+++ b/test/test_pallas.py
@@ -38,7 +38,8 @@ class PallasTest(unittest.TestCase):
     y = torch.arange(8, dtype=torch.int).to("xla")
     expected_output = x + y
 
-    output = torch_xla._XLAC._xla_tpu_custom_call([x, y], payload, [x.shape], [x.dtype])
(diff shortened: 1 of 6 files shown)


alanwaketan:
This PR is ready for review.

JackCaoG:
I will take a look

JackCaoG (approved):

JackCaoG:
Do we need this in 2.3 release? I don't consider this as the critical fix, wdyt?

alanwaketan:
> Do we need this in 2.3 release? I don't consider this as the critical fix, wdyt?

I think it is, but I will hold it until I finished some e2e modeling tests. Thanks for the review.

PR #6898, opened 2024-04-08
Set all_reduce_token to null when exiting

yitongh:
This PR has the same purpose as #6247 , fixing the #6246 issue. This PR sets the token only when the client still exists, to avoid the hanging issue in mp test. This PR is reopened because the core issue with this token is frequently encountered.


yitongh:
cc @JackCaoG

JackCaoG (approved):

PR #6902, opened 2024-04-08
[torch_xla2] Combined tensor.XLADispatchMode() and functions.XLAFunctionMode()

will-cromar:
See #6874 and #6854

Implement `torch_xla2.mode` to mark code that should use JAX-based lowerings. 

Some other naming ideas:

- `torch_xla2.jax`
- `torch_xla2.lower`

### DIFF
--- a/experimental/torch_xla2/test/test_context.py
+++ b/experimental/torch_xla2/test/test_context.py
@@ -12,8 +12,6 @@ class TestContext(unittest.TestCase):
       self.assertIsInstance(x, tensor.XLATensor2)
       y = x.abs()
       self.assertIsInstance(y, tensor.XLATensor2)
-      # TODO: remove print
-      print(y)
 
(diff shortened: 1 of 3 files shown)


will-cromar:
@qihqi, I added a quick test to confirm that tensors inside this context manager are actually `XlaTensor2`s, and I noticed this warning being printed: `WARNING:root:Falling back to default implementation of abs`

I confirmed (with the prints, will remove before submitting) that I'm getting the right result.

I added that warning in the previous PR so we don't fall back silently, but it will get spammy if `__torch_function__` is called even for tensor methods that are lowered. Am I missing something? Do you think I should remove the warning?

will-cromar:
Reduced the log about unimplemented functions to `info` for now so it's silent by default.

qihqi (approved):

PR #6905, opened 2024-04-09
[torch_xla2] Simplify developer setup steps

will-cromar:
- Add optional dependencies for CPU, CUDA, and TPU. Use `jax[cuda12]` instead of `jax[cuda12_pip]` since it doesn't require an extra index.
- Make use of global options in `requirements.txt`s to install the CPU version of torch automatically.
- Use `tensorflow-cpu` because it's much smaller
- Recommend using VSCode to set up virtual env
- Update package dependencies
- Update name of `test_requirements` to match `dev-requirements`
- Add `ruff` tool setup and adjust to be more like `yapf` settings in main package. `ruff check` shows 42 errors, so we can start fixing these in future PRs.

Tested locally by deleting my old environment and creating a new one with these instructions.

### DIFF
--- a/experimental/torch_xla2/README.md
+++ b/experimental/torch_xla2/README.md
@@ -4,7 +4,8 @@
 
 Currently this is only source-installable. Requires Python version >= 3.10.
 
-### NOTE: 
+### NOTE:
+
 Please don't install torch-xla from instructions in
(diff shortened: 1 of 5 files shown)


qihqi (approved):
this is awesome. Thanks!

PR #6907, opened 2024-04-10
fix pixel_shuffle return empty

zpcore:
Fixes https://github.com/pytorch/xla/issues/5886. Basically lowered the op using the XlaOps instead of going through pytorch decomposition.

Removed `ExpectCounterChanged("xla::permute_copy", cpp_test::GetIgnoredCounters());` because we don't call `xla::permute_copy` in the lowering.

### DIFF
--- a/torch_xla/csrc/xla_lower_util.cpp
+++ b/torch_xla/csrc/xla_lower_util.cpp
@@ -1194,6 +1194,24 @@ xla::XlaOp BuildCdistForward(xla::XlaOp x1, xla::XlaOp x2, xla::XlaOp p,
   }
 }
 
+xla::XlaOp BuildPixelShuffle(xla::XlaOp input, int64_t upscale_factor) {
+  const xla::Shape& input_shape = ShapeHelper::ShapeOfXlaOp(input);
+  absl::Span<const int64_t> dimensions = input_shape.dimensions();
+  int64_t batch_size = dimensions[0];
+  int64_t channels = dimensions[1];
+  int64_t height = dimensions[2];
+  int64_t width = dimensions[3];
+
+  int64_t new_channels = channels / (upscale_factor*upscale_factor);
+  int64_t new_height = height * upscale_factor;
+  int64_t new_width = width * upscale_factor;
+
+  xla::XlaOp tmp = xla::Reshape(input, {batch_size, new_channels, upscale_factor, upscale_factor, height, width});
+  tmp = xla::Transpose(tmp, {0, 1, 4, 2, 5, 3});
+  xla::XlaOp output = xla::Reshape(tmp, {batch_size, new_channels, new_height, new_width});
+  return output;
+}
+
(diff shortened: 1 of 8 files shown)


JackCaoG:
Do you know if pixel_shuffle a view op?

zpcore:
> Do you know if pixel_shuffle a view op?

This is how pytorch do the composition: https://github.com/pytorch/pytorch/blob/58047205ed098c04ec045e66fc39dcc70b60600b/torch/_refs/nn/functional/__init__.py#L1183-L1190. 

XlaOp:Reshape served the functionality of `view` op here. That's why we don't see `view` in our lowering.

zpcore:
No, it's not. Just tested on both CPU and XLA device, the input and output are in different shapes.

JackCaoG:
ok.. yea input and output will be different shape, but if on CPU you modify the source tensor, output tensor won't get modified, it is a not a view op

zpcore:
Oh I see, I should modify the contents to test the view. Thanks!

torch_xla/csrc/xla_lower_util.cpp
```diff
@@ -1194,6 +1194,27 @@ xla::XlaOp BuildCdistForward(xla::XlaOp x1, xla::XlaOp x2, xla::XlaOp p,
...
+  int64_t width = dimensions[3];
+
+  int64_t new_channels = channels / (upscale_factor * upscale_factor);
+  int64_t new_height = height * upscale_factor;
+  int64_t new_width = width * upscale_factor;
+
+  xla::XlaOp tmp =
+      xla::Reshape(input, {batch_size, new_channels, upscale_factor,
+                           upscale_factor, height, width});
+  tmp = xla::Transpose(tmp, {0, 1, 4, 2, 5, 3});
```
wonjoolee95, line 1212: Maybe I just don't understand `pixel_shuffle` enough, but why is this `tranpose` needed here?

zpcore: I am using https://g.co/gemini/share/8ee217185f5f. We permute the matrix so that we can easily do reshape next by stacking those dimensions. Same logic was used here: https://github.com/pytorch/xla/blob/786484d2fd271e519c4f222c1224f836322662a8/experimental/torch_xla2/torch_xla2/_ops.py#L987.

wonjoolee95: Nice! LGTM.

wonjoolee95 (approved):

torch_xla/csrc/xla_lower_util.cpp
zpcore: We permute the matrix so that we can easily do reshape next by stacking those dimensions. Same logic was used here: https://github.com/pytorch/xla/blob/786484d2fd271e519c4f222c1224f836322662a8/experimental/torch_xla2/torch_xla2/_ops.py#L987.

PR #6910, opened 2024-04-10
Add ansible stage for building CUDA plugin

will-cromar:
- New github action to build plugin. Future actions will consume it. This will not affect other workflows for now.
  - Runs directly in dev container and publishes build artifact through GHA. Click into the workflow and check it out!
image
- The main package and CUDA plugin can be built completely separately, so remove the plugin from build.
- Add nightly version tags to CUDA plugin. It should use the same source as the `torch_xla` main package, but this is still better than 0.0.1
- Add `numpy` to the GPU build environment. The latest XLA pin update requires it.

Long term: we can separate the build flow in our CI into two streams. We can build the main `torch_xla` package _without GPU support_, which will be substantially faster. CPU and TPU tests can launch immediately after this. The GPU plugin will build concurrently. Since there are almost never changes to the OpenXLA binary between pin updates, this will be ~instant unless you update or patch XLA. GPU tests will begin when both the main package and plugin are done building.

There is no impact to existing workflows right now.

### DIFF
--- a/plugins/cuda/setup.py
+++ b/plugins/cuda/setup.py
@@ -1,3 +1,4 @@
+import datetime
 import os
 import sys
 
@@ -10,4 +11,7 @@ import setuptools
 build_util.bazel_build('@xla//xla/pjrt/c:pjrt_c_api_gpu_plugin.so',
                        'torch_xla_cuda_plugin/lib', ['--config=cuda'])
 
-setuptools.setup()
+setuptools.setup(
+  # TODO: Use a common version file
+  version=f'2.4.0.dev{datetime.date.today().strftime("%Y%m%d")}'
+)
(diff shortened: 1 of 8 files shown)


JackCaoG:
that separate the GPU plugin build part sounds awesome.

plugins/cuda/setup.py
```diff
@@ -10,4 +11,7 @@
 build_util.bazel_build('@xla//xla/pjrt/c:pjrt_c_api_gpu_plugin.so',
                        'torch_xla_cuda_plugin/lib', ['--config=cuda'])
 
-setuptools.setup()
+setuptools.setup(
+  # TODO: Use a common version file
+  version=f'2.4.0.dev{datetime.date.today().strftime("%Y%m%d")}'
```
JackCaoG, line 16: lol do we need to update this for every release?

JackCaoG (approved):

plugins/cuda/setup.py
will-cromar: Yeah... I do want to refactor this to use a common version file with the main build, but that's a future PR.

will-cromar:
Merging this because we need the `pyproject.toml` changes to prevent the GPU nightly build from breaking tonight.

PR #6913, opened 2024-04-10
Fix runtime error when run dynamo with a profiler scope

JackCaoG:
Without this change, when running dynamo inside a `xp.trace` it will throw a runtime error.


wonjoolee95 (approved):

alanwaketan (commented):
Thanks for fixing this long standing issue!

PR #6914, opened 2024-04-10
Add `xla.sync`

will-cromar:
See #6399 

Leaving out the `wait` parameter for now since we plan to change the behavior.

### DIFF
--- a/test/test_devices.py
+++ b/test/test_devices.py
@@ -4,6 +4,7 @@ from absl.testing import absltest, parameterized
 import torch
 import torch_xla as xla
 import torch_xla.runtime as xr
+import torch_xla.debug.metrics as met
 
 
 class TestDevices(parameterized.TestCase):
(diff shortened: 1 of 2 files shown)


test/test_devices.py
```diff
@@ -29,6 +33,12 @@ def test_real_devices(self):
   def test_device_count(self):
     self.assertEqual(xla.device_count(), 4)
 
+  def test_sync(self):
+    torch.ones((3, 3), device=xla.device())
```
JackCaoG, line 37: nit clear the metrics before `sync` call checking the metrics.

test/test_devices.py
```diff
@@ -12,6 +13,9 @@ def setUpClass():
     xr.set_device_type('CPU')
     os.environ['CPU_NUM_DEVICES'] = '4'
 
+  def tearDown():
```
JackCaoG, line 16: oh is this being called after every test?

JackCaoG (approved):

test/test_devices.py
will-cromar: Yeah, I decided to clear after every test in case we add more cases like me.

PR #6915, opened 2024-04-10
[SPMD] Support manual sharding

alanwaketan:
Summary:
This pull request makes SPMD support the manual sharding type via a new private API called: _mark_manual_sharding. I don't expect users will need to call this function explicitly.

Besides adding support for the sharding annotation, we also need to define the behavior of the data shards. For data, the current behavior is error out.

Test Plan:
PJRT_DEVICE=TPU python test/spmd/test_xla_sharding.py -v -k test__mark_manual_sharding

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1100,7 +1100,7 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
 
     self.assertEqual(id(mesh), id(expected_mesh))
 
-  def test__mark_manual_sharding(self):
+  def test_mark_manual_sharding(self):
     x = torch.randn(3, 2)
     xx = x.to(xm.xla_device())
     xt = xs._mark_manual_sharding(xx)
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1100,6 +1100,28 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
 
     self.assertEqual(id(mesh), id(expected_mesh))
 
+  def test__mark_manual_sharding(self):
+    x = torch.randn(3, 2)
+    xx = x.to(xm.xla_device())
+    xt = xs._mark_manual_sharding(xx)
+
+    hlo = torch_xla._XLAC._get_xla_tensors_hlo([xt.global_tensor])
+    self.assertIn(
+        '%p0.1 = f32[3,2]{0,1} parameter(0), sharding={manual}',
+        hlo)
+
+    self.assertEqual(xt.sharding_type, xs.ShardingType.MANUAL)
+    self.assertEqual(xt.sharding_spec, "{manual}")
+
+    shards = xt.local_shards
+    self.assertEqual(len(shards), 1)
+    # Only one shard on device 0.
+    for i, shard in enumerate(shards):
+      self.assertEqual(shard.data.device, torch.device('cpu'))
+      self.assertTrue(torch.allclose(shard.data, x))
+      self.assertIsInstance(shard.indices, type(Ellipsis))
+      self.assertEqual(shard.replica_id, i)
+
 
 if __name__ == '__main__':
   test = unittest.main()
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1101,24 +1101,22 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertEqual(id(mesh), id(expected_mesh))
 
   def test_mark_manual_sharding(self):
-    x = torch.randn(3, 2)
-    xx = x.to(xm.xla_device())
+    x = torch.zeros(3, 2).to(xm.xla_device())
+    with self.assertRaises(RuntimeError):
+      xt = xs._mark_manual_sharding(x)
+
+    xx = x + 1
     xt = xs._mark_manual_sharding(xx)
 
     hlo = torch_xla._XLAC._get_xla_tensors_hlo([xt.global_tensor])
-    self.assertIn('parameter(0), sharding={manual}', hlo)
-
+    self.assertIn(', sharding={manual}', hlo)
     self.assertEqual(xt.sharding_type, xs.ShardingType.MANUAL)
     self.assertEqual(xt.sharding_spec, "{manual}")
 
-    shards = xt.local_shards
-    self.assertEqual(len(shards), 1)
-    # Only one shard on device 0.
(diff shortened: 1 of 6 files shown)


torch_xla/csrc/xla_sharding_util.h
```diff
@@ -80,11 +80,13 @@ class ShardingUtil {
   // based on the `sharding` spec. REPLICATED sharding should result in shards
   // identical to the input; OTHERS (tiled) sharding result in shards where
   // each data dimension is sharded across devices along the same dimension in
-  // the `tile_assignment`; the returned tensor shards vector is indexed by the
-  // device IDs. There is no data duplication. Shards are not padded in case the
-  // input tensor is not evenly partitionable, unless `padded` is set.
-  // The the returned tensors will be in 1:1 correspondence with the `devices`
-  // vector, so the `i`th result will belong on the `i`th device.
+  // the `tile_assignment`; MANUAL sharding result in shards where only the
+  // first device holds the full data; the returned tensor shards vector is
```
jonb377, line 84: > only the first device holds the full data

Is this by definition of manual sharding?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2034,45 +2034,43 @@ void InitXlaModuleBindings(py::module m) {
...
-          std::vector<at::Tensor> cpu_shards =
-              XlaDataToTensors(WrapXlaData(handles), element_types);
-          // Populate the resulting vector of shards and device strings
-          std::vector<std::vector<std::pair<at::Tensor, std::string>>> result;
-          int shards_per_tensor =
-              runtime::GetComputationClient()->GetLocalDevices().size();
-          result.reserve(cpu_shards.size() / shards_per_tensor);
-          for (int i = 0; i < cpu_shards.size(); i += shards_per_tensor) {
+            std::vector<at::Tensor> cpu_shards =
+                XlaDataToTensors(WrapXlaData(shard_handles), element_types);
```
jonb377, line 2055: Calling `XlaDataToTensors` on each tensor individually will slow down d2h transfers for async checkpointing, since PjRt won't be able to fully utilize transfer parallelization.

Do we expect manually-sharded tensors to contain actual device data generally, or will they usually be IR? If just IR, maybe we can add an assertion to prevent access here.

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -488,8 +489,11 @@ std::vector<at::Tensor> ShardingUtil::ShardTensor(
             shards[i], c10::IntArrayRef(pads.data(), pads.size()), 0);
       }
     }
-  } else if ((sharding.type() == xla::OpSharding::MANUAL) ||
-             (sharding.type() == xla::OpSharding::TUPLE)) {
+  } else if ((sharding.type() == xla::OpSharding::MANUAL)) {
+    // Just put the full tensor on the first device.
+    shards[0] = tensor;
+    shards.resize(1);
```
jonb377, line 495: How does this work for a compuatation, since we need to feed each device some input data?

e.g. based on your unit test, what happens if we run:

```python
    x = torch.randn(3, 2)
    xx = x.to(xm.xla_device())  # xx is device data
    xt = xs._mark_manual_sharding(xx)

    ones = torch.ones(3, 2).to(xm.xla_device()) # ones is replicated to all devices
    print(xx + ones)  # What will happen here?
```

jonb377 (commented):
Interesting! I've been wondering what manual sharding is for.

torch_xla/csrc/xla_sharding_util.h
yeounoh: This is not by definition, but by our implementation choice.

test/spmd/test_xla_sharding.py
```diff
@@ -1100,6 +1100,26 @@ def test_global_mesh(self):
 
     self.assertEqual(id(mesh), id(expected_mesh))
 
+  def test__mark_manual_sharding(self):
+    x = torch.randn(3, 2)
+    xx = x.to(xm.xla_device())
+    xt = xs._mark_manual_sharding(xx)
+
+    hlo = torch_xla._XLAC._get_xla_tensors_hlo([xt.global_tensor])
+    self.assertIn('parameter(0), sharding={manual}', hlo)
```
yeounoh, line 1109: Great!

torch_xla/csrc/init_python_bindings.cpp
yeounoh: I rather keep it functional for both cases -- shouldn't it be asynchronous anyway, not blocking the actual training run?

torch_xla/csrc/xla_sharding_util.cpp
yeounoh: XLA should assume that `xt` is sharded manually, so expected to be plicated as well. The purpose of `MANUAL` is to support custom kernel and prevent XLA to override the manual sharding.

yeounoh (approved):
LGTM, I leave the correctness review of distributed checkpointing with manual sharding to @jonb377  and his unit tests.

test/spmd/test_xla_sharding.py
```diff
@@ -1100,6 +1100,26 @@ def test_global_mesh(self):
 
     self.assertEqual(id(mesh), id(expected_mesh))
 
+  def test__mark_manual_sharding(self):
```
yeounoh, line 1103: nit. even though it's testing the `_` prefixed api, let's keep it as `test_mark_manual_sharding`

torch_xla/csrc/init_python_bindings.cpp
alanwaketan: This is interesting. I was not aware this performance optimization...

torch_xla/csrc/xla_sharding_util.cpp
alanwaketan: Good question. I would expect it behaves as a single device. Let me double check as well.

torch_xla/csrc/xla_sharding_util.h
alanwaketan: @yeounoh Will that be replicated then?

alanwaketan:
Here is the new TPU CI run: https://github.com/pytorch/xla/actions/runs/8652176761

alanwaketan:
TPU test here: https://github.com/pytorch/xla/actions/runs/8654305716

alanwaketan:
All tests passed. I'm going to merge it. Let me know if I need to follow up on anything.

torch_xla/csrc/xla_sharding_util.h
yeounoh: Per our offline discussion, abstain from manual sharding on input data.

PR #6918, opened 2024-04-11
Build TPU package in CI

will-cromar:
It's not strictly just the "TPU" package since it can still use the GPU plugin. Will rename it when I get rid of the GPU-only build.

Move TPU CI into the main build/test workflow and consume the main packages. This should reduce queue times for the TPU CI, since builds won't occupy a TPU and prevent other tasks from running.

Next steps:

- Migrate GPU and CPU tests over to new dependency chain
- Fix whatever the doc push task is doing?
- Publish base image for upstream
- Remove old workflow and all old circle CI scripts :confetti_ball:


will-cromar:
Successful TPU run here: https://github.com/pytorch/xla/actions/runs/8653732236/job/23730116979?pr=6918

will-cromar:
I'll make ansible respect the torch pin in a follow-up. The current TPU CI also does not use the pin, so it's not a regression.

.github/workflows/_build_torch_xla.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Setup gcloud
+        shell: bash
+        run: |
+          echo "${GCLOUD_SERVICE_KEY}" > $GOOGLE_APPLICATION_CREDENTIALS
+      - name: Checkout PyTorch Repo
+        uses: actions/checkout@v4
+        with:
+          repository: pytorch/pytorch
+          path: pytorch
+          # TODO: correct pin
```
JackCaoG, line 39: are we checking out torch pin? I think this is a very important part of the CI..

will-cromar:
Never mind, we did fix it in the current TPU CI. Do you want me to fix it in this PR or a follow up?

JackCaoG (approved):

JackCaoG:
why is GPU build failing after 4s lol

will-cromar:
Infrastructure issue? `Temporarily overriding HOME='/home/ec2-user/actions-runner/_work/_temp/aecbae00-42d3-4e19-adba-8747bad65ab6' before making global git config changes`

I'll try kicking it again. I am trying to speed up the CI, but not by _that_ much

PR #6920, opened 2024-04-12
Move `.torch_pin` and handle in ansible

will-cromar:
Tested locally


infra/ansible/roles/build_srcs/tasks/main.yaml
```diff
@@ -1,3 +1,18 @@
+- name: Read PyTorch pin
+  ansible.builtin.command: cat {{ (src_root, 'pytorch/xla/.torch_pin') | path_join }}
```
JackCaoG, line 2: There are actually 2 places historylly where we put `.torch_pin`, one is `pytorch/xla/.torch_pin`, another one is `pytorch/xla/torch_patches/.torch_pin`. I think it might be better to deprecate the `pytorch/xla/.torch_pin` since we only use this for the release branches.

will-cromar: Yeah, I did see half of the references in this repo were to one place, and half were to the other place. I weakly prefer just deleting the `torch_patches` directly since (as far as I know) we never add patches and our build scripts do not apply them.

will-cromar: Discussed offline. Moving this under `.github` since this mainly exists for CI anyway. Updated all of the docs too.

.circleci/README.md
```diff
@@ -2,16 +2,16 @@
...
+Sometimes a PyTorch/XLA PR needs to be pinned to a specific PyTorch PR to test new featurues, fix breaking changes, etc. Since PyTorch/XLA CircleCI pulls from PyTorch master by default, we need to manually provided a PyTorch pin. In a PyTorch/XLA PR, PyTorch an be manually pinned by creating a `.torch_pin` file at the root of the repository. The `.torch_pin` should have the corresponding PyTorch PR number prefixed by "#". Take a look at [example here](https://github.com/pytorch/xla/pull/3792/commits/40f41fb98b0f2386d287eeac0bae86e873d4a9d8). Before the PyTorch/XLA PR gets merged, the `.torch_pin` must be deleted.
 
 ## Coodinating merges for breaking PyTorch PRs
-When PyTorch PR introduces a breaking change, its PyTorch/XLA CircleCI tests will fail. Steps for fixing and merging such breaking PyTorch change is as following: 
+When PyTorch PR introduces a breaking change, its PyTorch/XLA CircleCI tests will fail. Steps for fixing and merging such breaking PyTorch change is as following:
```
JackCaoG, line 14: nit `CircleCI` -> `github action`? or just say CI

will-cromar: Yeah, I'll just move this whole README to .github too

JackCaoG (approved):
I think this should work but do you mind just test it by adding a pin in this pr and check log if it actually check out the that commit/branch in the CI log?

will-cromar:
Pinned to random commit 9c7c81b89789802d783098c9fd217f41338dfdb4. Looks like this works from the build output:

```
From https://github.com/pytorch/pytorch
     * branch                  9c7c81b89789802d783098c9fd217f41338dfdb4 -> FETCH_HEAD
```

PR #6922, opened 2024-04-12
[SPMD] Support SPMDFullToShardShape

alanwaketan:
Summary:
This pull request supports SPMDFullToShardShape which is a custom op that opens a region for non-partitioned graph in SPMD program. It will stop SPMD auto sharding and partition in that region and therefore allows manual sharding like cc ops.

To implement it, this pull request expands CustomSharding node to accept a new type. To be notice, the output shape of the op needs to be the shard shape of the input, and the node needs to have manual sharding annotation.

Test Plan:
PJRT_DEVICE=TPU python test/spmd/test_xla_sharding.py -v -k test_spmd_full_to_shard_shape

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1139,7 +1139,8 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
     # xx.cpu()
 
     # Replicated shape
-    xt = xs.mark_sharding(x, self._get_mesh((1, self.n_devices)), (None, None))
+    x = torch.zeros(8, 4).to(xm.xla_device())
+    xt = xs.mark_sharding(x, self._get_mesh((self.n_devices, 1)), (None, None))
     xx = torch_xla._XLAC._spmd_full_to_shard_shape(xt.global_tensor)
 
     hlo = torch_xla._XLAC._get_xla_tensors_hlo([xx])
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1118,6 +1118,36 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
     # The following exception cannot be caught somehow.
     # xt.global_tensor.cpu()
 
+  def test_spmd_full_to_shard_shape(self):
+    x = torch.zeros(8, 4).to(xm.xla_device())
+    with self.assertRaises(RuntimeError):
+      x = torch_xla._XLAC._spmd_full_to_shard_shape(x)
+
+    # Sharded shape
+    xt = xs.mark_sharding(x, self._get_mesh((1, self.n_devices)), (0, 1))
+    xx = torch_xla._XLAC._spmd_full_to_shard_shape(xt.global_tensor)
+
+    hlo = torch_xla._XLAC._get_xla_tensors_hlo([xx])
+    self.assertEqual(xx.shape, (8, 4 // self.n_devices))
+    self.assertIn(f'%custom-call.2 = f32[8,{4//self.n_devices}]{{1,0}}', hlo)
+    self.assertIn(f'custom_call_target="SPMDFullToShardShape", sharding={{manual}}', hlo)
+    self.assertEqual(torch_xla._XLAC._get_xla_sharding_spec(xx), "{manual}")
+
+    # It looks like XLA does't like only having manual sharding in the HLO.
+    # It needs to be paired with SPMDFullToShardShape/SPMDShardToFullShape.
+    # The following exception cannot be caught somehow.
+    # xx.cpu()
+
+    # Replicated shape
+    xt = xs.mark_sharding(x, self._get_mesh((1, self.n_devices)), (None, None))
+    xx = torch_xla._XLAC._spmd_full_to_shard_shape(xt.global_tensor)
+
+    hlo = torch_xla._XLAC._get_xla_tensors_hlo([xx])
+    self.assertEqual(xx.shape, (8, 4))
+    self.assertIn(f'%custom-call.2 = f32[8,4]{{1,0}}', hlo)
+    self.assertIn(f'custom_call_target="SPMDFullToShardShape", sharding={{manual}}', hlo)
+    self.assertEqual(torch_xla._XLAC._get_xla_sharding_spec(xx), "{manual}")
+
 
 if __name__ == '__main__':
   test = unittest.main()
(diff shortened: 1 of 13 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -182,6 +182,12 @@ std::vector<XLATensorPtr> GetXlaTensors(const std::vector<at::Tensor>& tensors,
   return xtensors;
 }
 
+bool IsIr(const at::Tensor& tensor) {
```
JackCaoG, line 185: nit, `IsNonDeviceDataIR`

test/spmd/test_xla_sharding.py
```diff
@@ -1118,6 +1118,39 @@ def test_mark_manual_sharding(self):
...
+    self.assertEqual(xx.shape, (8, 4 // self.n_devices))
+    self.assertIn(f'%custom-call.2 = f32[8,{4//self.n_devices}]{{1,0}}', hlo)
+    self.assertIn(
+        f'custom_call_target="SPMDFullToShardShape", sharding={{manual}}', hlo)
+    self.assertEqual(torch_xla._XLAC._get_xla_sharding_spec(xx), "{manual}")
+
+    # It looks like XLA does't like only having manual sharding in the HLO.
+    # It needs to be paired with SPMDFullToShardShape/SPMDShardToFullShape.
+    # The following exception cannot be caught somehow.
+    # xx.cpu()
```
JackCaoG, line 1140: do you intend to keep this `xx.cpu`?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2503,7 +2503,8 @@ void XLANativeFunctions::_propagate_xla_data(const at::Tensor& input,
   XLATensor::ShardingSpecPtr sharding = input_tensor->sharding_spec();
   if (sharding && sharding->sharding.type() != xla::OpSharding::UNKNOWN) {
     tensor_methods::custom_sharding_(output_tensor,
-                                     input_tensor->sharding_spec());
+                                     input_tensor->sharding_spec(),
+                                     CustomSharding::Type::kSharding);
```
JackCaoG, line 2507: so you assume only tensor with `kSharding` will be called with in place ops?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1939,16 +1945,26 @@ void InitXlaModuleBindings(py::module m) {
...
         });
-  m.def("_mark_manual_sharding", [](const at::Tensor& input,
-                                    xla::OpSharding sharding) {
+  m.def("_mark_manual_sharding",
+        [](const at::Tensor& input, xla::OpSharding sharding) {
+          XLA_CHECK(IsIr(input))
+              << "Marking any data tensors as manual is not supported";
+          ShardingUtil::XlaMarkSharding(input, sharding);
+        });
+  m.def("_spmd_full_to_shard_shape", [](const at::Tensor& input) -> at::Tensor {
```
JackCaoG, line 1954: this method is call `xxx_shape` but it actually returns a tensor which is very confusing...

torch_xla/csrc/ops/custom_sharding.h
```diff
@@ -7,14 +7,24 @@ namespace torch_xla {
 
 class CustomSharding : public XlaNode {
  public:
+  enum class Type {
+    kSharding,
+    kSPMDFullToShardShape,
+    kSPMDShardToFullShape,
+  };
```
JackCaoG, line 14: This enum is really confusing, can you add some comment around what they actually does? I was reading the SPMD code again, this op itself only means we want to shard the underlying value and the actual sharding resides in the XlaTensor or Based XLAIR object?

test/spmd/test_xla_sharding.py
alanwaketan: Yea, it's more like a note that this won't work... I was trying to use with self.assertRaises but that doesn't capture the exception... I have noticed this before too. When libtpu crashed, it's hard to catch it in the py level. Not sure why. Maybe you have some better ideas?

torch_xla/csrc/aten_xla_type.cpp
alanwaketan: That's the original design which is to align with the original design of SPMD... So yea.. for kSharding...

torch_xla/csrc/init_python_bindings.cpp
alanwaketan: I just faithfully copy the XlaOP name... This won't be exposed to the user so I guess it's okay to leave it as it is.

torch_xla/csrc/ops/custom_sharding.h
alanwaketan: Right, this is just the name of the custom call. The sharding annotation is in XlaTensor as normal. I can add more explanations.

test/spmd/test_xla_sharding.py
JackCaoG: oh I think I run into similar issue before.. The way I handle it was ugly through https://github.com/pytorch/xla/blob/a7a1357d5ee3b575a7eeba523f74607461eddad1/test/spmd/test_dynamo_spmd.py#L172-L181

torch_xla/csrc/init_python_bindings.cpp
JackCaoG: yea.. as long as it is just a pybind I am fine with it..

torch_xla/csrc/ops/custom_sharding.h
yeounoh: Maybe we can annotate explicilty that this is sharding type for custom call in the enum class name or somethinhg.

test/spmd/test_xla_sharding.py
alanwaketan: C++ crash on pt level can be caught with `self.assertRaise` but not libtpu level.... I'm not sure why... yea, not even with this hack...

alanwaketan: cc @will-cromar Do you know how to catch libtpu exception on py? Appreciate your insights.

torch_xla/csrc/ops/custom_sharding.h
alanwaketan: I guess the current approach sort of does it already? Can you be more specific? @yeounoh

test/spmd/test_xla_sharding.py
will-cromar: I don't think you can. To make a proper runtime error, you have to raise an exception, and Google internal binaries don't generally do that. I wrote about a similar case in https://github.com/pytorch/xla/issues/6700#issuecomment-1986521945

alanwaketan: Thanks, Will. That makes a lot of sense now.

torch_xla/csrc/ops/custom_sharding.h
yeounoh: I agree, `Type` is already defined under `CustomSharding`

yeounoh (approved):
LGTM, thanks!

alanwaketan:
Thanks, Yeounoh!

PR #6925, opened 2024-04-16
[SPMD] Support SPMDShardToFullShape

alanwaketan:
Summary:
This pull request enables SPMDShardToFullShape. The trickiest part is how to get the full shape, and here is a couple of options:
- [ ] Bookkeeping the shape full shape that enters SPMDFullToShardShape. This is not selected given the output could be created on the fly.
- [ ] Constructing the full shape from the local shard and the sharding spec. This is not selected given there is no way to deal with the padding. We can't examine the data during the tracing time.
- [x] Let users pass the full shape in. This is selected because it's just the most sounded path.

Tes Plan:
PJRT_DEVICE=TPU python test/spmd/test_xla_sharding.py -v -k test_manual_sharding_e2e -k test_spmd_shard_to_full_shape

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1152,7 +1152,7 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertEqual(torch_xla._XLAC._get_xla_sharding_spec(xx), "{manual}")
 
   def test_spmd_shard_to_full_shape(self):
-    x = torch.zeros(8, 4).to(xm.xla_device())
+    x = torch.zeros(8, 8).to(xm.xla_device())
     x += 1
     # No sharding spec attached.
     with self.assertRaises(RuntimeError):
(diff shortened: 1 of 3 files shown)


test/spmd/test_xla_sharding.py
```diff
@@ -1151,6 +1151,55 @@ def test_spmd_full_to_shard_shape(self):
...
+      x = torch_xla._XLAC._spmd_shard_to_full_shape(
+          xt.global_tensor,
+          torch_xla._XLAC.OpSharding([], [], [], xs.ShardingType.REPLICATED),
+          x.shape, x.dtype)
+
+    xs.clear_sharding(xt)
+    xt = xs._mark_manual_sharding(xt)
+    xx = torch_xla._XLAC._spmd_shard_to_full_shape(
+        xt.global_tensor,
+        torch_xla._XLAC.OpSharding([], [], [], xs.ShardingType.REPLICATED),
```
yeounoh, line 1175: So it uses the passed sharding type when returns back to SPMD?

test/spmd/test_xla_sharding.py
```diff
@@ -1151,6 +1151,55 @@ def test_spmd_full_to_shard_shape(self):
...
     self.assertEqual(torch_xla._XLAC._get_xla_sharding_spec(xx), "{manual}")
 
+  def test_spmd_shard_to_full_shape(self):
+    x = torch.zeros(8, 8).to(xm.xla_device())
+    x += 1
+    # No sharding spec attached.
+    with self.assertRaises(RuntimeError):
+      x = torch_xla._XLAC._spmd_shard_to_full_shape(
+          x, torch_xla._XLAC.OpSharding([], [], [], xs.ShardingType.REPLICATED),
+          x.shape, x.dtype)
```
yeounoh, line 1161: so we've decided to pass the original full shape as argument, right?

yeounoh (approved):
LGTM, left some questions -- thanks @alanwaketan ❤️

test/spmd/test_xla_sharding.py
alanwaketan: At least I have decided haha, see the PR descriptions for the reasons behind.

test/spmd/test_xla_sharding.py
alanwaketan: Yup!

alanwaketan:
Thanks @yeounoh for the quick review!

PR #6863, opened 2024-03-30
[ZeroRedundancyOptimizer] fp32 grad acc

sssrijan-amazon:
Add support to use grad accumulation hook for back-propagation in ZeroRedundancyOptimizer.

PR #6930, opened 2024-04-16
[ZeRO-1] add args for save master weights and fp32 grad acc

hgt312:
- add args for save master weights/fp32 grad acc
- also doc strings
- change saved master weights format from list to dict (it's for DCP to transform the dict to a useable format)

### DIFF
--- a/torch_xla/distributed/zero_redundancy_optimizer.py
+++ b/torch_xla/distributed/zero_redundancy_optimizer.py
@@ -1,4 +1,5 @@
 import copy
+import logging
 from typing import (Any, Iterator, Optional, Type, Union, List, Dict)
 
 import torch
@@ -49,7 +50,7 @@ class ZeroRedundancyOptimizer(Optimizer):
             ``dtype`` of grad accumulation will be the same as ``optimizer_dtype``. Users can set this
             to True to use higher precision for gradients accumulation. Default: False
         save_master_weights (bool, Optional):
-            if ``True``, also save sharded master weights. Default: False
+            if ``True``, also save sharded master weights. Default: True
         higher_cc_precision (bool, Optional): if ``True``, use higher precision for collective communication
             operators (the same as ``optimizer_dtype``). Default: False
         **defaults: any trailing arguments, which are forwarded to the local
@@ -74,10 +75,15 @@ class ZeroRedundancyOptimizer(Optimizer):
       lazy_init: bool = False,
       coalesce_cc: bool = False,
       use_grad_acc_hook: bool = False,
-      save_master_weights: bool = False,
+      save_master_weights: bool = True,
       higher_cc_precision: bool = False,
       **defaults: Any,
   ):
+    if not save_master_weights:
+      logging.warning(
+          'Not saving master weights may have accuracy issues when resuming training!'
+      )
+
     super().__init__(params, defaults)
 
     self.global_world_size = xm.xrt_world_size()
--- a/torch_xla/distributed/zero_redundancy_optimizer.py
+++ b/torch_xla/distributed/zero_redundancy_optimizer.py
@@ -416,7 +416,7 @@ class ZeroRedundancyOptimizer(Optimizer):
                 pin_layout=self.pin_layout,
                 groups=self.sharding_groups,
             )
-            if padded_param != param.dtype:
+            if padded_param.dtype != param.dtype:
               padded_param = padded_param.to(dtype=param.dtype)
             param.data.copy_(padded_param.data[:param.size(0)])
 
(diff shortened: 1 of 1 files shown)


torch_xla/distributed/zero_redundancy_optimizer.py
```diff
@@ -57,11 +69,13 @@ def __init__(
...
       max_norm: Optional[float] = None,
       pin_layout: bool = True,
-      use_grad_acc_hook: bool = False,
       sharding_groups: Optional[Any] = None,
       grad_norm_groups: Optional[Any] = None,
       lazy_init: bool = False,
       coalesce_cc: bool = False,
+      use_grad_acc_hook: bool = False,
+      save_master_weights: bool = False,
+      higher_cc_precision: bool = False,
```
aws-rhsoln, line 78: why do we need this? If once the optimizer dtype is set, shouldn't all ops in optimizer use the same dtype?

torch_xla/distributed/zero_redundancy_optimizer.py
```diff
@@ -411,7 +435,7 @@ def allgather_weights_and_update_full_parameter(self):
           if param.grad is not None or (self.use_grad_acc_hook and
                                         hasattr(shard, 'main_grad')):
             padded_param = padded_params[index]
-            if param.dtype != self.optimizer_dtype:
+            if padded_param != param.dtype:
```
aws-rhsoln, line 438: bug? We are comparing param to dtype

torch_xla/distributed/zero_redundancy_optimizer.py
hgt312: it controls the cc ops dtype, 2 reason to have this:
- without this, we see performance drop from tutorial, because it's from bf16/bf16 to fp32/fp32
- Haozheng's gpu compatible branch need this, we make it a config in nxd

[reduce-scatter dtype]/[all-gather dtype]
| use_grad_acc_hook/higher_cc_precision | True      | False     |
| ------------------------------------- | --------- | --------- |
| True                                  | fp32/fp32 | fp32/bf16 |
| False                                 | fp32/fp32 | bf16/bf16 |

HahTK: For production version of this we need to ensure the feature matches the expected use cases. I think there are some gaps. Please refer to use case below. 

Also let me know if I missed anything

Checkpoint save : 
+++++++++++
1. Default use case : Save Master Weights and Optimizer State. Do not save : Forward Weights. 
2. Backward compatibility : Save Forward Weights and Optimizer State. Do not save : Master Weights. 
3. There is no use case for saving all 3 master + forward + optimizer but also no reason to prevent it ? 

DEFAULT BEHAVIOR should be (1). 
WARNINGS : We should print warnings for use case (2) and (3).
  * (2) - warn user that this has accuracy issues if they use it to resume training.
  * (3) - warn user that they are wasting disk space and forward weights will be ignored on load.
NOTE : There may exist use cases where customer want (2) because they want to publish lower precision checkpoint for inference or fine-tuning etc. We recommend that be done using an offline conversion script. A convenience feature maybe to automatically dump an *extra* lower precision checkpoint only on the final checkpoint. Some users continue training even from "completed" training runs with new data so the final checkpoint still needs to have the full precision.

Checkpoint load : 
+++++++++++
1. Default use case : Load master weights and Optimizer. Broadcast that to forward weights and ignore forward weights if they were saved.
2. Backward Compatibility : If master weights does not exist, the run should *FAIL* by default. Then expose an option to force a training to resume from lower precision weights

DEFAULT BEHAVIOR should be (1) and FAIL on (2).
WARNINGS : We should inform the user that continuing training from low precision weights will affect accuracy, especially if there were multiple crashes close in time because the errors introduced by each save/restore will compound itself.  We should print the option needed to override the FAIL in the warning message.

torch_xla/distributed/zero_redundancy_optimizer.py
```diff
@@ -394,6 +416,8 @@ def allgather_weights_and_update_full_parameter(self):
                 pin_layout=self.pin_layout,
                 groups=self.sharding_groups,
             )
+            if padded_param != param.dtype:
```
aws-rhsoln, line 419: same bug?

hgt312: nice catch..

torch_xla/distributed/zero_redundancy_optimizer.py
hgt312: agree with these. while these should be done in nxd/nnm level.

for this PR, do we want to use `True` as default for `save_master_weights` and `higher_cc_precision`?

HahTK: Instead of duping code in NNM and NXD, I think the warnings should be in ZERO1. 
IMHO, the default should be to use save_master_weights = True but this means it is a change that must be documented in release notes. 

for `higher_cc_precision` this is a product decision.

aws-rhsoln (approved):

jeffhataws:
@JackCaoG  @alanwaketan  this is followup PR to https://github.com/pytorch/xla/pull/6863 and https://github.com/pytorch/xla/pull/6862 . Will cherry-pick all of these for master soon.

JackCaoG:
Ah, it is to your 2.1 branch. SG, I will just merge it then.

PR #6942, opened 2024-04-19
[Fori_loop|While_loop] Create fori_loop.md

ManfeiBai:
Add doc for `fori_loop`/`while_loop` and add simple user guide for simple test case


yeounoh (approved):
LGTM

PR #6945, opened 2024-04-19
[cherry-pick] [Fori_loop|While_loop] Create fori_loop.md (#6942)

ManfeiBai:
Add doc for fori_loop/while_loop and add simple user guide for simple test case

### DIFF
--- a/docs/fori_loop.md
+++ b/docs/fori_loop.md
@@ -3,7 +3,7 @@
 - if loop range is dynamic, its deeper implementation would be [`jax.lax.while_loop`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.while_loop.html),
 - if loop range is not dynamic, its deeper implementation would be [`jax.lax.scan`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.scan.html),
 
-In 2.3, PyTorch/XLA enable `while_loop` with some simple case, and
+In 2.3, PyTorch/XLA enable `while_loop` with some simple case.
 
 # while_loop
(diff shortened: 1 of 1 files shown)


yeounoh (approved):
LGTM

yeounoh:
If not all the fori_loop PRs are in r2.3, then maybe we shouldn't back-port this... I will leave it up to you and Siyuan. @ManfeiBai @lsy323

JackCaoG:
If user tried to run the code in the doc with 2.3 whl, would it work? if not we should not backport.

ManfeiBai:
> If user tried to run the code in the doc with 2.3 whl, would it work? if not we should not backport.

we have `while_loop` and `fori_loop` in this PR at the same time, and we only include PR for while_loop in 2.3, so I could update this PR to only include `while_loop` version and push local change to this branch again

qihqi (approved):

ManfeiBai:
> If user tried to run the code in the doc with 2.3 whl, would it work? if not we should not backport.

Thanks, tried with 2.3RC12 wheel(torch and torch_xla) and passed test case mentioned in md file, also passed tests in `test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py`

ManfeiBai:
> If not all the fori_loop PRs are in r2.3, then maybe we shouldn't back-port this... I will leave it up to you and Siyuan. @ManfeiBai @lsy323

Thanks, @lsy323, I updated the doc to only mention feature included in 2.3, do we wanna backport this doc change?

lsy323:
@ManfeiBai Backport docs sounds good to me.

lsy323 (approved):

PR #6947, opened 2024-04-19
Build CPP tests in new CI workflow

will-cromar:
- Pre-build CPP tests during build step and upload them as artifacts. This significantly speeds up getting feedback from these tests.
  - If the build options for the `torch_xla` package and the C++ tests, this invalidates the cache. Rather than trying to bring `test/cpp/run_test.sh` in line with `setup.py`, we can just build everything at once
  - Support building C++ tests in ansible so we use exactly the same build options.
- Create a new `test_suite` for CPP tests
- Add placeholder plugin that just returns a library path and default options. This is for C++ tests only for now.
- Remove old workflow.

A typical run in the old CI took about 2.5 hours. This test run took only 1 hour and 10 minutes. Most tests in the new workflow finish before the old _build_ finishes, e.g.

image


.github/scripts/run_tests.sh
```diff
@@ -0,0 +1,108 @@
...
+  export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${TORCH_DIR}/lib
+  if [ -x "$(command -v nvidia-smi)" ]; then
+    CUDA_PLUGIN_DIR=$(python -c "import pkgutil; import os; print(os.path.dirname(pkgutil.get_loader('torch_xla_cuda_plugin').get_filename()))")
+    export PJRT_LIBRARY_PATH=$CUDA_PLUGIN_DIR/lib/pjrt_c_api_gpu_plugin.so
+    export PJRT_DEVICE=LIBRARY
+    export PJRT_DYNAMIC_PLUGINS=1
+  else
+    export PJRT_DEVICE=CPU
+  fi
+  export XLA_EXPERIMENTAL="nonzero:masked_select"
```
JackCaoG, line 43: can you also add `nms`, example in https://github.com/pytorch/xla/blob/023e2c83dcf20b973ad4ec60b27469fcada02af5/test/tpu/run_tests.sh#L14-L15

torch_xla/csrc/runtime/pjrt_registry.cc
```diff
@@ -21,8 +21,24 @@ namespace runtime {
...
+    return sys_util::GetEnvString("PJRT_LIBRARY_PATH", "");
+  }
+
+  const std::unordered_map<std::string, xla::PjRtValueType>
+  client_create_options() const override {
+    return {};
+  }
+
+  bool requires_xla_coordinator() const override { return false; }
+};
```
JackCaoG, line 38: why do we need this plugin?

.github/scripts/run_tests.sh
will-cromar: Done

torch_xla/csrc/runtime/pjrt_registry.cc
will-cromar: This is just a minimal implementation of the plugin interface that lets us pass in a path. We want to take this codepath during init, rather than the statically linked CUDA client: https://github.com/pytorch/xla/blob/b9a9449f205d10769660c428cc68755a6ffe183a/torch_xla/csrc/runtime/pjrt_registry.cc#L63

 Normally the plugin registration is done through Python, so we need a workaround for the C++ tests.

JackCaoG (approved):

PR #6951, opened 2024-04-22
Lower embedding bag forward only

bhavya01:


wonjoolee95 (approved):
LGTM!

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1290,6 +1290,38 @@ at::Tensor XLANativeFunctions::embedding_dense_backward(
...
+  TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  if (mode == 1 || scale_grad_by_freq || sparse || padding_idx != -1) {
+    return at::native::call_fallback_fn<
+        &xla_cpu_fallback,
+        ATEN_OP(_embedding_bag_forward_only)>::call(weight, indices, offsets,
+                                                    scale_grad_by_freq, mode,
+                                                    sparse, per_sample_weights,
+                                                    include_last_offset,
+                                                    padding_idx);
+  }
```
ysiraichi, line 1308: @bhavya01 Is there a reason only `_embedding_bag_forward_only` (instead of also lowering `_embedding_bag`)? What about the fallback condition? Is there a specific reason we are not lowering that, too?

wonjoolee95: If I recall correctly, the code for the other rest of `_embedding_bag` was overly complicated so we measured it as out of scope for this PR. @bhavya01, please correct me if I'm wrong.

wonjoolee95: Hey @ysiraichi, would lowering `_embedding_bag` entirely be something that you need?

ysiraichi: Well, it would be nice having that. I won't be working on it right now, since I have things with higher priority to be done. Anyway, I have [this draft branch](https://github.com/pytorch/xla/compare/ysiraichi/use-torch-decompositions?expand=1) that can help us lowering + maintaining composite operations. It does mainly 2 things:

- Allow us to write lowerings for composite operations in Python
- Check for implemented decompositions (PyTorch + PyTorch/XLA) whenever we hit [the fallback function](https://github.com/pytorch/xla/blob/c1b745e5fd0db62156eab09731711ee51b83020b/torch_xla/csrc/aten_cpu_fallback.cpp#L19). If it finds one, use that to decompose the operation into possibly already decomposed operations
    - At the moment, PyTorch decompositions are only used in dynamo
    - Allow us to use PyTorch decompositions on non-dynamo experiments
    - Allow us to use PyTorch/XLA specific decompositions on both dynamo and non-dynamo experiments

I won't be working on this PR for a while, so if anyone wants to take over it, I don't mind. If this PR gets merged, we would probably have an easier time lowering operations.

PR #6952, opened 2024-04-23
Build upstream CI image on push to master

will-cromar:
This saves us a step of building the image manually and uploading to GCR. Plus, we have to [stop using GCR](https://cloud.google.com/artifact-registry/docs/transition/prepare-gcr-shutdown) at some point anyway

`_build_upstream_image` is a copy of `_build` but GitHub doesn't recognize it.

TODO after this PR: replace the 1.1-lite tag in upstream CI with 1.2-lite to confirm this works.


will-cromar:
Test build passed here: https://github.com/pytorch/xla/actions/runs/8900702561/job/24442860347

Rebasing and adding `if` condition back

JackCaoG (approved):

PR #6954, opened 2024-04-23
Rework docs push

will-cromar:
- Follow [current GHA examples](https://github.com/actions/starter-workflows/blob/main/pages/static.yml) for deploying pages.
- Tweak logic for selecting subdir. As near as I can tell, every release commit was updating master's docs: https://github.com/pytorch/xla/blob/b834e49907aab60e096620664ca8ba328ef92141/.circleci/doc_push.sh#L22-L26
- ~Use token from GHA instead of bot account.~ Nope, org policy forbids us from getting a GHA token with write access
- Run docs build for every event, but only publish for pushes. You can check what the generated docs will look like in the `github-pages` artifact (`releases/$PR_NUMBER/merge`)

### DIFF
--- a/.github/workflows/_docs.yml
+++ b/.github/workflows/_docs.yml
@@ -47,7 +47,8 @@ jobs:
     - name: Merge changes
       shell: bash
       run: |
-        subdir=${{ env.BRANCH_NAME == 'master' && 'master' || format('{0}/{1}', 'release/', env.BRANCH_NAME) }}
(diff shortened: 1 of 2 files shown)


will-cromar:
`The nested job 'deploy' is requesting 'pages: write, id-token: write', but is only allowed 'pages: none, id-token: none'.`

Hmmm, there may be a repository setting we need to change.

JackCaoG (approved):

PR #6958, opened 2024-04-24
Re-land: dynamo expand test with view-replay.

ysiraichi:
This PR is a follow-up to #6950, re-introducing the dynamo expand test while setting the `view_replay_for_aliased_outputs` dynamo configuration that uses the outputs of the functionalization step for reconstructing output views.

cc @miladm @JackCaoG


test/dynamo/test_dynamo.py
```diff
@@ -671,6 +676,22 @@ def foo(x):
     self.assertEqual(expected.dtype, actual.dtype)
     self.assertEqual(expected.device, actual.device)
 
+  def test_return_expand(self):
+    _set_view_replay_for_aliased_outputs()
```
JackCaoG, line 680: do we want to leave this options to always true througout the test? If the intention is to always set this to True for PyTorch/XLA, you should set it in https://github.com/pytorch/xla/blob/master/torch_xla/__init__.py .

torch_patches/.torch_pin
```diff
@@ -0,0 +1 @@
+680a89e2ced8f45b87b7fd0509c07cf6b768be14
```
JackCaoG, line 1: why do we need a pin? I thought Brian's change already merged to main.

ysiraichi: It was reverted.

JackCaoG (approved):

ysiraichi:
@JackCaoG @vanbasten23 I'm puzzled by this CI failure. Doesn't TPU CI build take the `.torch_pin` into account? It looks like it's not

JackCaoG:
yea TPU CI does not take into pin into account yet. I think it is oK to merge. @will-cromar FYI

PR #6964, opened 2024-04-24
Temporarily ignore torch commit in CI test

will-cromar:
It looks like our import is printing something extra to stdout and preventing us from reading the git pin:

```
Run echo "PYTORCH_COMMIT=$(python -c 'import torch_xla.version; print(torch_xla.version.__torch_gitrev__)')" >> $GITHUB_ENV
Error: Unable to process file command 'env' successfully.
Error: Invalid format 'refreshing <module 'torch.ops.quantized_decomposed' from 'torch.ops'> quantize_per_tensor'
```

Ignore the torch commit from the build step so we can run tests again. This means the commit used for the test step may be different than that used for the build step, which is bad. This hack is temporary.

### DIFF
--- a/.github/workflows/_test_python.yml
+++ b/.github/workflows/_test_python.yml
@@ -101,14 +101,18 @@ jobs:
           # TODO: Add these in setup.py
           pip install fsspec
           pip install rich
-      - name: Record PyTorch commit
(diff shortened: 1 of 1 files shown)


will-cromar:
Let's make sure the tests actually run before we merge this

lsy323 (approved):

PR #6966, opened 2024-04-24
Write `torch_xla.version.__torch_gitrev__` to file directly

will-cromar:
This is a little cursed, but it should work more consistently than the old way.


JackCaoG (approved):
Do we know why origional method doesn't work?

PR #6972, opened 2024-04-25
Update test_export_fx_passes.py

avikchaudhuri:
https://github.com/pytorch/pytorch/pull/124898 makes some changes to how `args` and `dynamic_shapes` are matched in `torch.export`, which will fail some XLA tests. This PR avoids those failures.

### DIFF
--- a/test/stablehlo/test_export_fx_passes.py
+++ b/test/stablehlo/test_export_fx_passes.py
@@ -18,7 +18,7 @@ class ExportFxPassTest(unittest.TestCase):
 
   def test_decompose_dynamic_shape_select(self):
     args = (torch.rand((10, 197, 768)), 1, 0)
-    dynamic_shapes = ([{0: Dim("bs")}, None, None],)
+    dynamic_shapes = (({0: Dim("bs")}, None, None),)
     m = wrap_func_as_nn_module(torch.ops.aten.select.int)
     ep = export(m, args, dynamic_shapes=dynamic_shapes)
     out1 = ep.module()(*args)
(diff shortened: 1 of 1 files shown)


test/stablehlo/test_export_fx_passes.py
```diff
@@ -18,7 +18,7 @@ class ExportFxPassTest(unittest.TestCase):
 
   def test_decompose_dynamic_shape_select(self):
     args = (torch.rand((10, 197, 768)), 1, 0)
-    dynamic_shapes = ([{0: Dim("bs")}, None, None],)
+    dynamic_shapes = (({0: Dim("bs")}, None, None),)
```
avikchaudhuri, line 21: `wrap_func_as_nn_module` creates a module with input signature `args: tuple(...)` and thus `dynamic_shapes` needs to match with a `tuple` as well.

test/stablehlo/test_export_fx_passes.py
```diff
@@ -55,7 +55,7 @@ def forward(self, x):
   def test_embedding_indices_flatten(self):
     args = (torch.rand((20, 768)), torch.randint(0, 15,
                                                  (3, 10)).to(torch.int64))
-    dynamic_shapes = ([None, {0: Dim("bs")}],)
+    dynamic_shapes = ((None, {0: Dim("bs")}),)
```
avikchaudhuri, line 58: same as above

avikchaudhuri:
@lsy323 for review?

JackCaoG:
@lsy323 Can you review this one?

avikchaudhuri:
Ping! This is blocking land of https://github.com/pytorch/pytorch/pull/124898

JackCaoG:
CI failed at build step so it is hard for me to see whether we can merge this pr without upstream one. From the change it seems like we do require upstream?

JackCaoG:
Ah.. you are using a fork of the pytorch/xla so we can not pin upstream.. Let me first verify if your changes works. If so we can just disable these two tests and merge the pr. I can enable the test with your change afterward.

JackCaoG:
Ah I confirmed that it worked locally even without upstream change. I will just approve and merge this pr.

JackCaoG (approved):

avikchaudhuri:
Thanks @JackCaoG

PR #6983, opened 2024-04-26
Enable trigger TPUCI run with label `tpuci`

ManfeiBai:


will-cromar:
I think we have to check the labels list on the PR, since we don't watch for labeling events: https://stackoverflow.com/a/74829754

Trying this idea in https://github.com/pytorch/xla/pull/6984

ManfeiBai:
Thanks, agree

ManfeiBai:
> I think we have to check the labels list on the PR, since we don't watch for labeling events: https://stackoverflow.com/a/74829754
> 
> Trying this idea in #6984

Thanks, agree!

PR #6992, opened 2024-04-29
Support `ahead of time (AOT) cross-device compilation`

zpcore:
## Summary
This is the first commit for supporting AOT compilation. Basically this allows us to compile TPU XLA executable on a CPU devices. Below is the command to run AOT compilation:
```
PJRT_DEVICE=AOT-v4:2x2x1  XLA_PERSISTENT_CACHE_PATH=./  python sample.py
```
In the command, we need two things:
1. Setup `PJRT_DEVICE`  to `AOT-...`. The suffix part should be the topology of the targeting TPU device that we want to compile.
2. Enable persistent cache like `XLA_PERSISTENT_CACHE_PATH=./ `. This will generate the XLA executable file in the current folder.

To use the XLA executable on a TPU device, we can use the following command:
```
PJRT_DEVICE=TPU  XLA_PERSISTENT_CACHE_PATH=./  python sample.py
```
If everything works fine, you should be able to see the executable deserialization successfully.

###  Example of `sample.py`:
```
import torch
import torch_xla

a = torch.rand([2,3])
b = torch.rand([2,3])
device = xm.xla_device()
a = a.to(device)
b = b.to(device)

f = torch.hstack([a,b])

# for compile, use torch_xla._XLAC._xla_warm_up_cache([f],[]). you can also use print or mark_step. Those two will trigger errors but can still can produce the executable.
# for execution, use print(f) or mark_step().
```

## Some todo items:
1. Add the unit test for the AOT compilation;
2. Support of dynamo compilation.

### DIFF
--- a/torch_xla/csrc/aten_xla_bridge.cpp
+++ b/torch_xla/csrc/aten_xla_bridge.cpp
@@ -54,6 +54,9 @@ class AtenXlaDeviceMapper {
     if (UseVirtualDevice()) {
       devices_.emplace_back(ParseDeviceString("SPMD:0"));
       devices_ordinals_[devices_.back()] = 0;
+    } else if (UseCrossCompilationDevice()) {
+      devices_.emplace_back(ParseDeviceString("AOT:0"));
+      devices_ordinals_[devices_.back()] = 0;
     } else {
       for (auto& device_str :
            torch_xla::runtime::GetComputationClient()->GetLocalDevices()) {
@@ -338,7 +341,8 @@ torch::lazy::BackendDevice AtenDeviceToXlaDevice(const c10::Device& device) {
 c10::Device XlaDeviceToAtenDevice(const torch::lazy::BackendDevice& device) {
   // TODO(yeounoh) until we expose SPMD virtual device to the frontend, this
   // will just be `XLA:0`.
-  if (device.type() == (int8_t)XlaDeviceType::SPMD) {
+  if (static_cast<XlaDeviceType>(device.type()) == XlaDeviceType::SPMD ||
+      static_cast<XlaDeviceType>(device.type()) == XlaDeviceType::AOT) {
     return c10::Device(at::kXLA, (size_t)0);
   }
   return c10::Device(at::kXLA,
@@ -350,9 +354,14 @@ std::string ToXlaString(const c10::Device& device) {
 }
 
 const torch::lazy::BackendDevice* GetDefaultDevice() {
-  static std::string default_device_spec =
-      UseVirtualDevice() ? "SPMD:0"
-                         : runtime::GetComputationClient()->GetDefaultDevice();
+  std::string default_device_spec;
+  if (UseVirtualDevice()) {
+    default_device_spec = "SPMD:0";
+  } else if (UseCrossCompilationDevice()) {
+    default_device_spec = "AOT:0";
+  } else {
+    default_device_spec = runtime::GetComputationClient()->GetDefaultDevice();
+  }
   XLA_CHECK(!default_device_spec.empty());
   static const torch::lazy::BackendDevice default_device =
       ParseDeviceString(default_device_spec);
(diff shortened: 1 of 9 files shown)


zpcore:
You don't need to pay too much attention to [pjrt_compile_only.cc](https://github.com/pytorch/xla/pull/6992/files#diff-96773c03cc27d0255ac91485a573ab1ff7d60adc6dd16ddff1ac875274b2a79a) and [pjrt_compile_only.h](https://github.com/pytorch/xla/pull/6992/files#diff-78e9faff5e35e37202076619c9c33c7d6b35c520a5f8bab8c6773c219a688c5e). Those are used to construct the `CompileOnlyPjRt[Client/Buffer/Device]`. A lot of functions returns status unimplemented, because those are pure virtual functions in `PjRt[Client/Buffer/Device]` and I have to mention them anyway.

will-cromar (commented):
Let's think about other ways we can frame the UX. It's fine if the interface is minimal in this PR as long as we have an idea of how to follow up.

In general, we should not embed complex information in env vars. I don't even really like setting `PJRT_DEVICE`, which is why we try to select it automatically where possible (similar to JAX). We used to require complex configuration in XRT, and it was a huge point of friction.

First, as far as I can tell, the `PJRT_DEVICE` should remain the _target_ device type. AOT is orthogonal to the device type and to SPMD. (SPMD is orthogonal from the user's perspective with a separate setting `XLA_USE_SPMD`/`xr.enable_spmd`, even if we confuse the concepts ourselves in code). Similarly, I would expect a Python API like `xla.aot.compile_for_device` that takes the target device type/size/shape. We can define presets for known TPU sizes in code. I'd actually try to dodge adding a new env var setting entirely if we can, but if we do, it should correspond closely to a more well-defined Python interface.

My concern with embedding complex strings in env vars is that typos become dangerous and hard to debug. Plus, it's hard to debug where env vars can originate, because they can be set at a system level, in wrapper scripts, or in code.

will-cromar:
Here's where I would go with this PR:

- Instead of creating a partial implementation of the `xla::PjRtClient` and using `PjRtComputationClient` around that, create a partial implementation of `ComputationClient`. I'll call it `PjRtCompilationClient` for now.
-  You can add a branch in `runtime.cc` to select this `PjRtCompilationClient` instead of the other two options:

https://github.com/pytorch/xla/blob/0e1f765c359c80ad02f5a674f8c3710c267fc947/torch_xla/csrc/runtime/runtime.cc#L24-L32

- `CompilationClient` is an API that we control, and it very rarely changes compared to the `xla:PjRtClient`. Maintaining a partial implementation will be easier for that reason.
  - There's also a conceptual issue with `pjrt_compile_only`: PJRT already has appropriate semantics for compilation independent of a functioning client or local device. `pjrt_compile_only` wraps that functionality under APIs that do define semantics for execution, like `PjRtClient`.
- `PjRtCompilationClient` will be analogous to `PjRtComputationClient`, with a few changes. Most of this internal state is irrelevant because it's there for tracking execution and local devices:

https://github.com/pytorch/xla/blob/0e1f765c359c80ad02f5a674f8c3710c267fc947/torch_xla/csrc/runtime/pjrt_computation_client.h#L156-L166

- e.g. `PjRtCompilationClient` should contain a `topology_` directly rather than a `client_` that contains a `topology_`
- Implement any alternative behaviors (like handling un-loaded computations) directly in `PjRtCompilationClient::Compile` and `PjRtCompilationClient::ExecuteComputation`
- I expect you'll need to fill in stub implementations of `Data` and `Computation` to make this work. The data is just  a placeholder, so the implementations should be simple. `PjRtUnloadedComputation` should just become `PjRtCompilationClient::Computation`.
- Remove the AOT device type: the device type unfortunately directly impacts the IR that we actually generate. We already run into issues like this with SPMD. That also means that if `AOT` is a device type, then it cannot be composable with `SPMD`, which is also a distinct device type.

zpcore:
Close the first version of AOT due to the drawback that this will make it hard to conduct openxla Pin update. Use https://github.com/pytorch/xla/pull/7581 instead.

PR #6996, opened 2024-04-29
Manually push to `gh-pages` branch

will-cromar:
`actions/deploy-pages@v4` requires some extra permissions that our robot account does not have.

Testing with a dry run first. `Deploy` should be disabled for PRs before merging.

PR #6997, opened 2024-04-29
Revert "Rework docs push"

will-cromar:
Reverts pytorch/xla#6954

Docs push is broken and I don't see an obvious way to fix forward.


will-cromar:
Actually, #6996 should work

PR #6998, opened 2024-04-29
Support pin pr number in new .torch_pin

JackCaoG:


JackCaoG:
@will-cromar I think something is wrong with the `.torch_pin`.

In `Build upstream CI image`, it failed with
```
/tmp/pytorch /var/lib/jenkins/workspace
+ '[' -e xla/.torch_pin ']'
++ cat xla/.torch_pin
+ git checkout '#123683'
```
I think this means we just don't prefix it with `#`, fine, we can do that.

However same error does not happened in the `Build and test / Build PyTorch/XLA`... I looked at the log and it looks like it still just check out main.

JackCaoG:
oh on a second thought, we should support `#pr_number`. I don't think we can directly checkout a fork..

will-cromar:
I'm not sure where the `#` came from. You should be able to check out a PR with a ref like this: https://stackoverflow.com/a/32447696

JackCaoG:
I think `#` means it is a PR number, without it we can assume it is just a branch name or commit number that can be directly checkout. I was trying to update the example pin pr [`40f41fb` (#3792)](https://github.com/pytorch/xla/pull/3792/commits/40f41fb98b0f2386d287eeac0bae86e873d4a9d8) mentioned in https://github.com/pytorch/xla/blob/master/.github/ci.md#pinning-pytorch-pr-in-pytorchxla-pr

will-cromar:
Yeah, it looked like there was special logic in `apply_patches.sh` (but not `common.sh` lol) to handle PR numbers like this. IMO it's better to just treat `.torch_pin` as a ref instead of applying special logic, since GH already gives us refs for PR numbers. I can update the doc.

JackCaoG:
I think most people are used to `#pr_number` by now, asking everyone to learn how to figure out the new format is not a great experience. Unless there is a strong reason we can't do this I'd prefer to just move the logic from `apply_patches.sh` to `common.sh`..

will-cromar:
I guess I just don't see the purpose of reinventing syntax for something that exists. Generally I've been trying to simplify the bash spaghetti, and it's simpler to interpret this value as a ref. `#` would only have been respected if the pin was at `/torch_patches/.torch_pin` (referenced by `apply_patches.sh`), but not at `/.torch_pin` (referenced by nightly build and `common.sh`).

The current logic for the pin checkout is here, which I mostly copied from `common.sh`:

https://github.com/pytorch/xla/blob/0e032b17f4041fc3cbea6e0e6dac1b55e91a1cee/infra/ansible/roles/build_srcs/tasks/main.yaml#L7-L14

We should not ever touch `common.sh` again until we're ready to update the upstream CI and remove it. Changes in the `ansible` stack will apply throughout the various CI systems and builds now.

JackCaoG:
hmm 
```
+ git checkout pull/123683/head
error: pathspec 'pull/123683/head' did not match any file(s) known to git
```

JackCaoG:
Well, from the old script, it seems like we need to do 
https://github.com/pytorch/xla/blob/73b915b55d96553a0e370b2bab01f47b8c2a9e7c/scripts/apply_patches.sh#L21-L22

I have to admit I don't really know what `fetch` actually do..

infra/ansible/roles/build_srcs/tasks/main.yaml
```diff
@@ -8,8 +8,14 @@
   # ansible.builtin.git wants to fetch the entire history, so check out the pin manually
   ansible.builtin.shell:
     cmd: |
-      git fetch origin {{ torch_pin.stdout }}
-      git checkout --recurse-submodules {{ torch_pin.stdout }}
+      if [[ {{ torch_pin.stdout }} = \#* ]]; then
+        PRNUM="${torch_pin.stdout//[!0-9]/}"
+        git fetch origin "pull/$PRNUM/head:$PRNUM"
```
will-cromar, line 13: You can simplify this a bit:

```
  ...
  git fetch origin "pull/$PRNUM/head
else
  git fetch origin {{ torch_pin.stdout }}
fi
git checkout --recurse-submodules FETCH_HEAD
```

Either ref you fetch will be tagged `FETCH_HEAD` immediately after, so there's no need to create a new branch called `$PRNUM`

JackCaoG:
@will-cromar is anisble really run with `bash`, it error out with
```
  start: '2024-04-30 23:02:46.498667'
  stderr: |-
    /bin/sh: 2: [[: not found
```
which suggested that it is using `sh`, https://stackoverflow.com/questions/3401183/bash-syntax-error-not-found

will-cromar:
The default is `sh` for some reason. You can switch to `bash` though: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/shell_module.html#examples

JackCaoG:
Ah ok, this seems to work now
```
    Switched to branch '123683'
```
Let me do the refactor you suggested

will-cromar (approved):

JackCaoG:
@will-cromar In my previous experiment I found that CI will continue to run if it can't checkout the pin(due to my syntax error in bash script). This seems dangerous, is there a way to force CI to fail if PIN check out script failed?

will-cromar:
Can you try `set -e` in the pin checkout script? I would expect that step to fail the whole build if any command fails. It's already skipped if there is no pin file at all by design.

JackCaoG:
To be safe I am just going to wait for all CI passed before merging. I will update the doc for the pin_update in a separate pr..

PR #7005, opened 2024-04-30
Clean up workspace before test

will-cromar:
Should fix this error:

```
Import check...
  File "/__w/xla/xla/torch_xla/__init__.py", line 152, in <module>
    from .version import __version__
ModuleNotFoundError: No module named 'torch_xla.version'
```

https://github.com/pytorch/xla/actions/runs/8898851270/job/24437801900

This results from combination of questionable decisions from both GitHub and Python. First, GHA does not guarantee that the workspace is actually empty. In this case, `torch_xla/` is still checked out from a previous run. Since there is a local `torch_xla` directory, Python gives this precedence over the installed `torch_xla` package.

Just delete everything in the workspace before running tests to eliminate non-determinism.

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -46,7 +46,7 @@ jobs:
       BAZEL_JOBS: 16
     steps:
       # See https://github.com/actions/checkout/issues/1014#issuecomment-1906802802
-      - name: Fix workspace permissions
+      - name: Clean up workspace
         run: |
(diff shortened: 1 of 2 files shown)


will-cromar:
This is a little repetitive. We can package the environment setup in a reusable template if we start adding more workflows that need it.

JackCaoG (approved):

PR #7006, opened 2024-04-30
Expose python unsafe buffer pointer

vanbasten23:
Similar to what JAX does https://github.com/openxla/xla/blob/f8cafa2b4ae9d7cf96adca75aaad37952be1a28b/xla/python/py_array.cc#L1598, we can use this helper to make sure dlpack does zero copy, by comparing the buffer point such as https://github.com/google/jax/blob/17d0eb30fa609df2a404543925ca50ae92b5c618/tests/array_interoperability_test.py#L84. This will be used in the followup dlpack PR.

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -2482,10 +2482,12 @@ void InitXlaModuleBindings(py::module m) {
               return runtime::GetComputationClient()->UnsafeBufferPointer(
                   UnwrapXlaData(data));
             } else {
-              XLA_ERROR() << "Could not get the buffer pointer.";
+              XLA_ERROR() << "Could not get the buffer pointer for XLATensor "
+                             "with IR that's not DeviceData";
             }
           }
-          XLA_ERROR() << "Could not get the buffer pointer.";
+          XLA_ERROR() << "Could not get the buffer pointer for XLATensor "
+                         "without a data handle or an IR.";
         });
 
   // -------------Dynamo Integration API Start-------------------------
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -2475,8 +2475,6 @@ void InitXlaModuleBindings(py::module m) {
                     xtensor->CurrentDataHandle());
             return runtime::GetComputationClient()->UnsafeBufferPointer(data);
           } else if (xtensor->CurrentIrValue().node != nullptr) {
-            // DeviceData* device_data =
-            // torch_xla::DeviceData::Cast(xtensor->CurrentIrValue().node.get());
             DeviceData* device_data =
                 DeviceData::Cast(xtensor->CurrentIrValue().node.get());
             if (device_data != nullptr) {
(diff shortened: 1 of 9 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2465,6 +2465,29 @@ void InitXlaModuleBindings(py::module m) {
...
+            return runtime::GetComputationClient()->UnsafeBufferPointer(data);
+          } else if (xtensor->CurrentIrValue().node != nullptr) {
+            DeviceData* device_data =
+                DeviceData::Cast(xtensor->CurrentIrValue().node.get());
+            if (device_data != nullptr) {
+              torch::lazy::BackendDataPtr data = device_data->data();
+              return runtime::GetComputationClient()->UnsafeBufferPointer(
+                  UnwrapXlaData(data));
+            } else {
+              XLA_ERROR() << "Could not get the buffer pointer.";
```
JackCaoG, line 2485: it is better to be a bit specified, like 
```
XLA_ERROR() << "Could not get the buffer pointer for XLATensor with IR that's not DeviceData";
```

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2465,6 +2465,29 @@ void InitXlaModuleBindings(py::module m) {
...
+            }
+          }
+          XLA_ERROR() << "Could not get the buffer pointer.";
```
JackCaoG, line 2488: ditto 
```
'Could not get the buffer pointer for XLATensor without XLAData and IR'
```

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -458,6 +458,16 @@ std::vector<ComputationClient::DataPtr> PjRtComputationClient::ReshardData(
   return resharded_results;
 }
 
+std::uintptr_t PjRtComputationClient::UnsafeBufferPointer(
+    const DataPtr handle) {
+  std::shared_ptr<PjRtData> pjrt_data = ReplicateShardedData(handle);
```
JackCaoG, line 463: why do you need to call `ReplicateShardedData`? Is it to handle the SPMD cases?

torch_xla/csrc/unwrap_data.h
```diff
@@ -11,6 +11,9 @@
 
 namespace torch_xla {
 
+runtime::ComputationClient::DataPtr UnwrapXlaData(
```
JackCaoG, line 14: I didn't see implementation for this function, did I miss anything?

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: I don't intend to handle the SPMD case atm. Let me change it.

torch_xla/csrc/unwrap_data.h
vanbasten23: The implementation is at https://github.com/pytorch/xla/blob/2907ab3093753d8a3516c8f67337a9464bf4ed8a/torch_xla/csrc/unwrap_data.cpp#L14. Somehow it's missing in the header..

JackCaoG (approved):

vanbasten23:
Thanks Jack for the review!

PR #7007, opened 2024-04-30
add check for neuron root comm id if not set

aws-kingrj:
We don't want to always overwrite this env var, so add a check to see if it's set

### DIFF
--- a/torch_xla/_internal/neuron.py
+++ b/torch_xla/_internal/neuron.py
@@ -5,8 +5,9 @@ import logging
 def num_local_processes() -> int:
   if 'MASTER_ADDR' not in os.environ:
     logging.warning("MASTER_ADDR not setting, defaulting to localhost")
-  os.environ['NEURON_RT_ROOT_COMM_ID'] = '{}:{}'.format(
-      os.environ.get('MASTER_ADDR', 'localhost'), '62182')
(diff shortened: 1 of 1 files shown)


torch_xla/_internal/neuron.py
```diff
@@ -5,8 +5,9 @@
 def num_local_processes() -> int:
   if 'MASTER_ADDR' not in os.environ:
     logging.warning("MASTER_ADDR not setting, defaulting to localhost")
-  os.environ['NEURON_RT_ROOT_COMM_ID'] = '{}:{}'.format(
-      os.environ.get('MASTER_ADDR', 'localhost'), '62182')
+  if os.environ.get("NEURON_RT_ROOT_COMM_ID", None) is None:
```
will-cromar, line 8: nit: I suggest `dict.setdefault` here instead https://docs.python.org/3.12/library/stdtypes.html#dict.setdefault

will-cromar (approved):

torch_xla/_internal/neuron.py
aws-kingrj: Only added in Python 3.12 though right? we are still using versions 3.8-3.10

will-cromar: `setdefault` has been around for a long time. It's in the docs for 3.8 too https://docs.python.org/3.8/library/stdtypes.html#dict.setdefault

PR #7011, opened 2024-05-01
Fix q dtype in paged attention kernel

wonjoolee95:
Fix q dtype in paged attention kernel:
1. At `if (num_heads // num_kv_heads) % 8 != 0`, we should've set q's dtype to float32.
    - Reference: [jax kernel #1](https://github.com/google/jax/blob/main/jax/experimental/pallas/ops/tpu/paged_attention/paged_attention_kernel.py#L393)
2. When we pass q to the pallas kernel, we need to call `q.to()` to convert it to the calculated dtype above.
     - Reference: [jax kernel #2](https://github.com/google/jax/blob/main/jax/experimental/pallas/ops/tpu/paged_attention/paged_attention_kernel.py#L508)

Test plan:
Paged attention unit tests are passing locally, and Woosuk's example is also passing locally as well.

### DIFF
--- a/torch_xla/experimental/custom_kernel.py
+++ b/torch_xla/experimental/custom_kernel.py
@@ -392,10 +392,10 @@ def paged_attention(q, k_pages, v_pages, lengths, page_indices,
   batch_size, num_heads, head_dim = q.shape
   num_kv_heads, _, page_size, head_dim_k = k_pages.shape
   batch_size_paged_indices, pages_per_sequence = page_indices.shape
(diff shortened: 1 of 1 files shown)


JackCaoG (approved):

wonjoolee95:
Thanks for the reviews! The CI and TPU CI are all passing, I'll merge this now to unblock Woosuk.

alanwaketan (commented):
LGTM.

PR #7013, opened 2024-05-01
Support torch_xla2 benchmarking using torchbench 

zpcore:
## Summary
Integrate the torch_xla2 testing into the benchmarking.

## Details
The PR ports the torch_xla2 running script based on file [here](https://gist.github.com/qihqi/9b952672b4971bb6697919e7108f8871#file-run-torchbench) into the existing benchmark script.

### How to run
To run the torch_xla2 benchmarking, you need to install the torch_xla2 based on instructions [experimental/torch_xla2/README.md](https://github.com/pytorch/xla/tree/2907ab3093753d8a3516c8f67337a9464bf4ed8a/experimental/torch_xla2)

Currently torch_xla2 doesn't work with dynamo. We need to append flag `--torch-xla2` to switch to torch_xla2, e.g.,
```
export JAX_PLATFORMS=TPU;
python experiment_runner.py \
--suite-name=torchbench \
--accelerator=tpu \
--progress-bar  \
--xla=PJRT  \
--test=eval \
--filter=dcgan \
--torch-xla2
```

In practical, we need to make sure JAX version and torch_xla are using the same PJRT version. I did the openxla pin update (backport) for torch_xla in order to use PJRT `0.47` verison. 

## Sample result
Just tried a simple model `dcgan` on TPU v5p-8, the torch_xla2 performance is impressive:

| benchmark      | platform  | torch_xla version |  backend | median_total_time (s) | compile_time (s) |
|------------|---------|-------|------|------|--------------|
| dcgan (eval)      | v5-8   | torch_xla |  LTC  | 0.0022        |  1.5729 |
| dcgan (eval)      | v5-8   | torch_xla |  openxla  | 0.0005        |  1.8661 |
| dcgan (eval)      | v5-8   | torch_xla2 | jax.jit | 0.0003911869862349704       | 1.464099046002957        |

### DIFF
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -22,8 +22,7 @@ class ExperimentLoader:
         "xla": [None, "PJRT", "XRT"],
         "xla_flags": [None],
         "dynamo": [None, "inductor", "openxla_eval", "openxla"],
-        "torch_xla2": [None, "extract_jax",
-                       "torch_export"],  # options only apply to torch_xla2
+        "torch_xla2": [None],  # options only apply to torch_xla2
         "test": ["eval", "train"],
     }
 
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -125,7 +125,14 @@ class ExperimentLoader:
 
 class BenchmarkExperiment:
 
-  def __init__(self, accelerator, xla, xla_flags, dynamo, test, batch_size, use_torch_xla2: bool = False):
+  def __init__(self,
+               accelerator,
+               xla,
+               xla_flags,
+               dynamo,
+               test,
+               batch_size,
+               use_torch_xla2: bool = False):
     self.accelerator = accelerator
     self.xla = xla
     self.xla_flags = xla_flags
@@ -142,7 +149,7 @@ class BenchmarkExperiment:
       process_env.pop("PJRT_DEVICE", None)
       process_env.pop("XRT_TPU_CONFIG", None)
       process_env.pop("XLA_FLAGS", None)
-    
+
     if self.use_torch_xla2:
       process_env.pop("JAX_PLATFORMS", self.accelerator.upper())
 
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -72,6 +72,9 @@ class ExperimentLoader:
         exclude_tags=()):
       return False
 
+    if cfg_dynamo is not None and self._args.torch_xla2:
+      return False
+
     # Check dynamo backend-specifics constraints.
     if cfg_dynamo == "inductor":
       if cfg_accelerator == "tpu" or cfg_xla is not None:
@@ -116,12 +119,13 @@ class ExperimentLoader:
         xla_flags=xla_flags,
         dynamo=dynamo,
         test=test,
-        batch_size=batch_size)
+        batch_size=batch_size,
+        use_torch_xla2=self._args.torch_xla2)
 
 
 class BenchmarkExperiment:
 
-  def __init__(self, accelerator, xla, xla_flags, dynamo, test, batch_size):
+  def __init__(self, accelerator, xla, xla_flags, dynamo, test, batch_size, use_torch_xla2: bool = False):
     self.accelerator = accelerator
     self.xla = xla
     self.xla_flags = xla_flags
@@ -129,6 +133,7 @@ class BenchmarkExperiment:
     self.test = test
     self.batch_size = batch_size
     self.accelerator_model = get_accelerator_model(self.accelerator)
+    self.use_torch_xla2 = use_torch_xla2
 
   def update_process_env(self, process_env):
 
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -124,7 +124,7 @@ class ExperimentLoader:
         accelerator=accelerator,
         xla=xla,
         xla_flags=xla_flags,
-        torch_xla2=self._args.torch_xla2,
+        torch_xla2=torch_xla2,
         dynamo=dynamo,
         test=test,
         batch_size=batch_size)
@@ -200,6 +200,7 @@ class BenchmarkExperiment:
     d["accelerator_model"] = self.accelerator_model
     d["xla"] = self.xla
     d["xla_flags"] = self.xla_flags
+    d["torch_xla2"] = self.torch_xla2
     d["dynamo"] = self.dynamo
     d["test"] = self.test
     d["batch_size"] = self.batch_size
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -125,21 +125,21 @@ class ExperimentLoader:
         accelerator=accelerator,
         xla=xla,
         xla_flags=xla_flags,
-        torch_xla2=torch_xla2,
         dynamo=dynamo,
+        torch_xla2=torch_xla2,
         test=test,
         batch_size=batch_size)
 
 
 class BenchmarkExperiment:
 
-  def __init__(self, accelerator, xla, xla_flags, torch_xla2, dynamo, test,
+  def __init__(self, accelerator, xla, xla_flags, dynamo, torch_xla2, test,
                batch_size):
     self.accelerator = accelerator
     self.xla = xla
     self.xla_flags = xla_flags
-    self.torch_xla2 = torch_xla2
     self.dynamo = dynamo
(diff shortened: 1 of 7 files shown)


benchmarks/benchmark_experiment.py
```diff
@@ -72,6 +72,9 @@ def _is_available(self, experiment_config):
         exclude_tags=()):
       return False
 
+    if cfg_dynamo is not None and self._args.torch_xla2:
```
ysiraichi, line 75: Could you add a comment explaining why this is not available?

benchmarks/experiment_runner.py
```diff
@@ -284,7 +284,7 @@ def run_once_and_gather_metrics(self, benchmark_experiment, benchmark_model,
 
     # Reset state and sync.
     reset_rng_state(benchmark_experiment)
-    self._mark_step(benchmark_experiment)
+    self._mark_step(benchmark_experiment, inputs_list[0])
```
ysiraichi, line 287: I'm not familiar with `torch_xla2`, but shouldn't we `mark_step` on all elements of `inputs_list`?

benchmarks/experiment_runner.py
```diff
@@ -307,7 +307,7 @@ def loop(pytorch_profile=None, iter_fn=None):
           total_timing += timing
 
         # Mark step.
-        self._mark_step(benchmark_experiment)
+        self._mark_step(benchmark_experiment, inputs_list[i])
```
ysiraichi, line 310: Shouldn't this `mark_step` be on `output`?

benchmarks/experiment_runner.py
```diff
@@ -414,9 +414,14 @@ def _prepare_inputs(self, example_inputs, should_randomize_input):
       inputs_list.append(inputs)
     return inputs_list
 
-  def _mark_step(self, benchmark_experiment):
+  def _mark_step(self, benchmark_experiment, tensor_to_check=None):
```
ysiraichi, line 417: Wouldn't it be more convenient to accept a list of tensors?

benchmarks/experiment_runner.py
```diff
@@ -859,6 +864,12 @@ def __str__(self):
       action="append",
       help="Flags to forward to XLA via `XLA_FLAGS` env var.",
   )
+  parser.add_argument(
+      "--torch-xla2",
+      action="store_true",
+      default=False,
```
ysiraichi, line 870: No need for `default` if using `action="store_true"`.

benchmarks/util.py
```diff
@@ -50,7 +50,7 @@ def reset_rng_state(benchmark_experiment=None):
   torch.manual_seed(1337)
   random.seed(1337)
   np.random.seed(1337)
-  if benchmark_experiment is not None and benchmark_experiment.xla is not None:
+  if benchmark_experiment is not None and benchmark_experiment.xla is not None and not benchmark_experiment.use_torch_xla2:
```
ysiraichi, line 53: Don't we need to reset RNG state for `torch_xla2`?

benchmarks/util.py
```diff
@@ -76,8 +76,14 @@ def is_xla_device_available(devkind):
   return r.returncode == 0
 
 
-def move_to_device(item, device):
-  return pytree.tree_map_only(torch.Tensor, lambda t: t.to(device), item)
+def move_to_device(item, device, use_torch_xla2: bool = False):
+  if use_torch_xla2:
+    import torch_xla2
+    import jax
+    move_to_device_func = lambda t: jax.device_put(torch_xla2.tensor.t2j(t))
```
ysiraichi, line 83: I think we also need to check if we are moving to an XLA device.

benchmarks/experiment_runner.py
zpcore: Oh yes, I think you are right!

zpcore: Hmm, just double checked, the inputs_list contains the inputs for all iterations. inputs_list[0] should be the one that feed into the model in the first step. However, I am not sure if we need to `block_until_ready` for all inputs. But to be safe, I will just block all inputs. WDYT?

benchmarks/experiment_runner.py
```diff
@@ -414,9 +414,14 @@ def _prepare_inputs(self, example_inputs, should_randomize_input):
       inputs_list.append(inputs)
     return inputs_list
 
-  def _mark_step(self, benchmark_experiment):
+  def _mark_step(self, benchmark_experiment, tensor_to_check=None):
     if benchmark_experiment.xla:
-      xm.mark_step()
+      if benchmark_experiment.use_torch_xla2:
+        assert tensor_to_check is not None, "torch_xla2 requires input tensor to block_until_ready"
```
vanbasten23, line 420: nit: do you need the assert? Or to block_until_ready on each tensor passed in

benchmarks/benchmark_experiment.py
```diff
@@ -138,6 +150,9 @@ def update_process_env(self, process_env):
       process_env.pop("XRT_TPU_CONFIG", None)
       process_env.pop("XLA_FLAGS", None)
 
+    if self.use_torch_xla2:
+      process_env.pop("JAX_PLATFORMS", self.accelerator.upper())
```
vanbasten23, line 154: Could you add a comment why we need to remove this env var?

benchmarks/benchmark_experiment.py
zpcore: I didn't see dynamo is supported for torch_xla2. @qihqi to confirm.

Added the comment to clarify here.

benchmarks/experiment_runner.py
zpcore: `tensor_to_check` is already a list of tensors. Update the name to `tensors_to_check`.

benchmarks/benchmark_experiment.py
zpcore: Hah, typo here. Insert the env instead of pop. Thanks

benchmarks/experiment_runner.py
zpcore: Oh, we may still need that because `tensor_to_check ` can be None and will error out in the for statement.

benchmarks/experiment_runner.py
ysiraichi: I think that works.
But, if we didn't, wouldn't we be also possibly timing the operations outside the i-th step of the model?

benchmarks/util.py
```diff
@@ -76,8 +76,15 @@ def is_xla_device_available(devkind):
...
 
 
-def move_to_device(item, device):
-  return pytree.tree_map_only(torch.Tensor, lambda t: t.to(device), item)
+def move_to_device(item, device, use_torch_xla2: bool = False):
+  if use_torch_xla2:
+    import torch_xla2
+    import jax
+    move_to_device_func = lambda t: jax.device_put(
+        torch_xla2.tensor.t2j(t), device)
```
ysiraichi, line 84: Is this going to work if `device="cpu"`?

benchmarks/benchmark_experiment.py
qihqi: it hasnt been implemented yet.

benchmarks/benchmark_model.py
```diff
@@ -110,21 +111,40 @@ def conversion_dtype(self):
...
     else:
       raise NotImplementedError
 
+    if self.benchmark_experiment.use_torch_xla2:
+      # for torch_xla2, we export model to FX graph and move weights to JAX device
+      import torch_xla2.export
+      import torch_xla2
+      import jax
+      import jax.numpy as jnp
+      exported = torch.export.export(self.module, self.example_inputs)
```
qihqi, line 131: let's add an option to use torch.export

The other (non-export) way is to use `torch_xla2.extract_jax` without exporting first

report both numbers they are both useful to know.

qihqi (approved):

benchmarks/util.py
qihqi: Not yet. How RNGs will be set in torch_xla2 is still under iteration. We can change once we figured out the API

benchmarks/experiment_runner.py
qihqi: Let's call `jax.block_until_ready` on all inputs.

zpcore: Yes, `_mark_step` internally is calling `block_until_ready` if using torch_xla2.

benchmarks/benchmark_model.py
zpcore: Added

benchmarks/util.py
zpcore: Just checked, CPU also works.

PR #7016, opened 2024-05-01
[ZeRO-1] Sync features from r2.1_aws_neuron branch

hgt312:
Sync features from r2.1_aws_neuron branch:
- use grad hook to make grad accumulation use higher precision
- save master weights
- an option to change cc ops' dtype

This PR also includes https://github.com/pytorch/xla/pull/6673


hgt312:
@jeffhataws

jeffhataws:
Hi @JackCaoG , it's strange that after including https://github.com/pytorch/xla/pull/7112 it is still failing. Maybe we need another change?

jeffhataws:
https://github.com/pytorch/xla/pull/7132 replaces this PR.

Issue #6979, opened 2024-04-26
Support non-traceable Custom Ops


thong3le:
## 🚀 Feature
`torch.export` supports exporting blackbox custom ops, however, we fails to export it to StableHLO using `exported_program_to_stablehlo` API 
https://pytorch.org/tutorials/intermediate/torch_export_tutorial.html#custom-ops

## Motivation
if we have non-traceable python codes in the custom ops, we can't export it to stablehlo program. This means we won't be able to cover as much of the model when exporting through StableHLO.

## Pitch

Here is the example pytorch codes
```
import torch
from torch.library import Library, impl, impl_abstract

m = Library("my_custom_library", "DEF")
m.define("custom_op(Tensor input) -> Tensor")

@impl(m, "custom_op", "CompositeExplicitAutograd")
def custom_op(x):
    raise Exception("DON'T GO HERE")
    return torch.relu(x)

@impl_abstract("my_custom_library::custom_op")
def custom_op_meta(x):
    return torch.empty_like(x)

class CustomOpExample(torch.nn.Module):
    def forward(self, x):
        x = torch.sin(x)
        x = torch.ops.my_custom_library.custom_op(x)
        x = torch.cos(x)
        return x

em = torch.export.export(CustomOpExample(), (torch.randn(3, 3),))
em.graph_module.graph.print_tabular()

from torch_xla.stablehlo import exported_program_to_stablehlo
stablehlo_program = exported_program_to_stablehlo(em)
print(stablehlo_program.get_stablehlo_text())
```
As you can see, `torch.export` runs fine and give us this fx graph, without caring what is inside `custom_op` impl.

```
opcode         name       target                               args          kwargs
-------------  ---------  -----------------------------------  ------------  --------
placeholder    arg0_1     arg0_1                               ()            {}
call_function  sin        aten.sin.default                     (arg0_1,)     {}
call_function  custom_op  my_custom_library.custom_op.default  (sin,)        {}
call_function  cos        aten.cos.default                     (custom_op,)  {}
output         output     output                               ((cos,),)     {}
```

`exported_program_to_stablehlo` fails because it runs the `custom_op` and hits `Exception`.

When I comment out the line `raise Exception("DON'T GO HERE")`, `exported_program_to_stablehlo` works fine, however it traces into `custom_op` by converting `relu` to `stablehlo.maximum`,

```
[...]

PR #7017, opened 2024-05-02
Support export custom op to stablehlo custom call

lsy323:
This PR resolves https://github.com/pytorch/xla/issues/6979

Support export custom op to stablehlo custom call, there are 2 user journeys

1. Register customer op to be wraped as stablehlo custom call during export, via `StableHLOExportOptions`
```
from torch_xla.stablehlo import allow_custom_op_in_graph

m = Library("my_custom_library", "DEF")
m.define("custom_op(Tensor input) -> Tensor")

class M(torch.nn.Module):

  def forward(self, x):
    x = torch.sin(x)
    x = torch.ops.my_custom_library.custom_op(x)
    x = torch.cos(x)
    return x

options = StableHLOExportOptions()
options.custom_ops_allowed_in_graph.add("my_custom_library")
ep = torch.export.export(M(), (torch.randn(3, 3),))
shlo_module = exported_program_to_stablehlo(ep, options)
```
Example output
```
module @IrToHlo.6 attributes {mhlo.cross_program_prefetches = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  func.func @main(%arg0: tensor<3x3xf32>) -> tensor<3x3xf32> {
    %0 = stablehlo.sine %arg0 : tensor<3x3xf32>
    %1 = stablehlo.custom_call @my_custom_library.custom_op.default(%0) {api_version = 0 : i32} : (tensor<3x3xf32>) -> tensor<3x3xf32>
    %2 = stablehlo.cosine %1 : tensor<3x3xf32>
    return %2 : tensor<3x3xf32>
  }
}
```
2. Call the `stablehlo_custom_call` API in the XLA dispatch impl of the custom op, with more control on the generated custom call
```
m.define("custom_op3(Tensor input) -> Tensor")

@impl(m, "custom_op3", "XLA")
def custom_op3_xla(x):
  res = stablehlo_custom_call((x,), "custom_op3", [x.shape[1:]],
                                  [torch.int8], True, "backend_config", 1)
  return res

class M(torch.nn.Module):

  def forward(self, x):
    x = torch.sin(x)
    x = torch.ops.my_custom_library.custom_op3(x)
    x = x + 1
    return x

ep = torch.export.export(M(), (torch.randn(3, 3),))
shlo_module = exported_program_to_stablehlo(ep)
shlo_text = shlo_module.get_stablehlo_text()
```
example ouptut
```
module @IrToHlo.10 attributes {mhlo.cross_program_prefetches = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  func.func @main(%arg0: tensor<3x3xf32>) -> tensor<3xi8> {
    %c = stablehlo.constant dense<1> : tensor<3xi8>
    %0 = stablehlo.sine %arg0 : tensor<3x3xf32>
    %1 = stablehlo.custom_call @custom_op3(%0) {backend_config = "backend_config", has_side_effect = true} : (tensor<3x3xf32>) -> tensor<3xi8>
    %2 = stablehlo.add %1, %c : tensor<3xi8>
    return %2 : tensor<3xi8>
  }
}
```

Test: Added unit tests to cover multiple inputs/outputs, and 2 user jouneys.

Future work:
1. Register `stablehlo_custom_call` as a torch op, currently the op schema failed with parameter type of `ScalarType[]`.
2. Merge the `tpu_custom_call` code path and `custom_call` code path. Currently there are duplicated code in 2 paths. `tpu_custom_call` should be a special use case of custom_call. Need to extend custom_call to support  `xla::CustomCallWithLayout` cc @alanwaketan 

cc @GleasonK


torch_xla/stablehlo.py
```diff
@@ -212,6 +214,22 @@ class StableHLOModelBundle:
   stablehlo_funcs: List[StableHLOFunc]
 
 
+# Ops that will be mapped to stablehlo.custom_call in the exported StableHLO graph.
+custom_ops_allowed_in_graph: Set[Callable] = set()
```
qihqi, line 218: Instead of a global, put this into the Config object that exported_program_to_stablehlo receives.

Also let's do per library granularity for now instead of per op.

user registers 'my_custom_library' entirely

torch_xla/stablehlo.py
```diff
@@ -262,6 +280,11 @@ def run_node(self, n) -> Any:
         ]
         self._mark_dynamic(res, dynamic_dims)
       return res
+    if n.op == 'call_function' and n.target in custom_ops_allowed_in_graph:
```
qihqi, line 283: ..then this like becomes `... and n.target.namespace in custom_ops_allowed_in_graph` etc

torch_xla/stablehlo.py
lsy323: Done, put it in the config object.

qihqi (approved):

xinli-centml:
Hi @lsy323, thanks for the amazing work!

We are investigating issue #8385 and found that `ouput_oprand_aliasing` [here](https://github.com/pytorch/xla/blob/73f54ba5bd7fb83d7ba81fe6f5e05fb6ee815d6f/torch_xla/csrc/ops/custom_call.cpp#L61) is silently dropped when calling the custom XLA OP. 

Do you know if there is a way to restore aliasing information here? If not, do you recall what the limitation is that caused the asliasing information to be dropped here?

PR #7023, opened 2024-05-03
Handle multiple inplace update input output aliasing

JackCaoG:
Fix the bug where if inplace operation is being applied multiple times, the aliasing won;t happened.

Consider the case without this pr
```
# Tensor ID 1, alias ID 1
t1 = torch.randn(5,5).to('xla:0')
# Tensor ID 2, alias ID 1
t1 += 1
# Tensor ID 3, alias ID 2
t1 * = 3
```
During the `mark_step` time we check that input buffer has `tensor ID` 1, and the output `alias id` is 2, hence it will skip donating input buffer of size (5,5). 

https://github.com/pytorch/xla/blob/d1235858628417ed7abc0d61e6e9be50df3e1a87/torch_xla/csrc/xla_graph_executor.cpp#L1249-L1253

https://github.com/pytorch/xla/blob/d1235858628417ed7abc0d61e6e9be50df3e1a87/torch_xla/csrc/xla_graph_executor.cpp#L1261-L1269

Alias ID should track the tensor ID of the input buffer, not the tensor ID of last base.

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -29,6 +29,7 @@
 #include "torch_xla/csrc/helpers.h"
 #include "torch_xla/csrc/ops/as_strided.h"
 #include "torch_xla/csrc/ops/as_strided_view_update.h"
+#include "torch_xla/csrc/ops/device_data.h"
 #include "torch_xla/csrc/ops/diagonal_view_update.h"
 #include "torch_xla/csrc/ops/einsum_utilities.h"
 #include "torch_xla/csrc/ops/index_ops.h"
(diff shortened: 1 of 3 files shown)


JackCaoG:
@alanwaketan @wonjoolee95 I think this one is ready for review.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2538,7 +2539,38 @@ void XLANativeFunctions::_propagate_xla_data(const at::Tensor& input,
   // 1) Aid XLA's InputOutputAlias.
   auto input_tensor = bridge::GetXlaTensor(input);
   auto output_tensor = bridge::GetXlaTensor(output);
-  output_tensor->data()->alias_id = input_tensor->GetUniqueId();
+  if (input_tensor->CurrentDataHandle() != nullptr ||
```
alanwaketan, line 2542: I guess we can always use alias_id?

JackCaoG: haha that's what I thought but actually no. Look at my example down below
```
    // x.tensor_id = 1, x.alias_id = 1
    x = torch.randn(5,5).to(xla_device())
    // x.tensor_id = 2, x.alias_id should be 1
    x += 1
    xm.mark_step()
    // x.tensor_id =3, x.alias_id should be 2 since input tensor id will be 2
    // for this graph
    x *= 1 of 1
    xm.mark_step()
```

if we always use `alias_id`, the `alias_id` of `x` in second would be `1`, but we need it to be 2.

JackCaoG: in the second execution, input tensor id is `2`, we need the alias ID to always match the input tensor ID. In other world we should not carry `alias_id` across `mark_step`.

JackCaoG: This is a bit tricky, even the underlying buffer is aliased, we still create a new PjrtBuffer object for `x` after the first `mark_step`. That DeviceData object(wrap about pjrtbuffer) will have `data_info ` with `tensor_id` 2, since `x`'s tensor id is 2 after the first `mark_step`.

alanwaketan: I see so the problem is mark_step? Will the right solution to clear out alias_id after mark_step?

alanwaketan (approved):
LGTM.

torch_xla/csrc/aten_xla_type.cpp
JackCaoG: we can do that too(reset alias_id to tensor id after processed the input_output_alias info). That might make this code less confuse haha.

alanwaketan: That sounds like a good follow up, but feel free to skip it.

jeffhataws:
Will this go into 2.4? Any chance it can be backported to 2.3?

JackCaoG:
This will be part of the 2.4, we don't do dot releases so it is unlikely for this one to be in the 2.3 release.

PR #7025, opened 2024-05-04
[WIP] Add dlpack support

vanbasten23:


vanbasten23:
Currently, the tests fails with 
```
2024-05-07 17:55:28.847281: F external/xla/xla/pjrt/pjrt_stream_executor_client.cc:1322] Check failed: holds_[i] == 0 (1 vs. 0)
Aborted (core dumped)
```
callstack https://gist.github.com/vanbasten23/a42b196b2fcd5e985a083752386dd3f8

test/test_operations.py
```diff
@@ -2441,6 +2445,115 @@ def test_unsafe_buffer_pointer(self):
     self.assertGreaterEqual(buf_ptr_2, 0)
 
 
+class TestDLPack(parameterized.TestCase):
+
+  def _test_dlpack_capsule_conversion_helper(self, xla_tensor):
+    dlpt = xdlpack.to_dlpack(xla_tensor)  # dlpt1 has type PyCapsule
+    got = xdlpack.from_dlpack(dlpt)
```
JackCaoG, line 2452: nit, `xla_tensor2` or `xla_tensor_dlpacked`. `got` is a bit confuinsg

torch_xla/csrc/dl_convertor.cpp
```diff
@@ -0,0 +1,368 @@
...
+        DeviceData::Cast(xtensor->CurrentIrValue().node.get());
+    if (device_data != nullptr) {
+      return UnwrapXlaData(device_data->data());
+    }
+    TF_VLOG(4) << "The xla tensor has IR value but does not have device data.";
+  }
+  TF_VLOG(4)
+      << "The xla tensor either has no current data handle or has no IR value.";
+  return nullptr;
+}
```
JackCaoG, line 41: maybe put this in `torch_xla/csrc/tensor_util.h`

torch_xla/csrc/dl_convertor.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+struct TorchXLADLMTensor {
+  ~TorchXLADLMTensor();
+  std::unique_ptr<xla::PjRtBuffer::ExternalReference> external_reference;
+  std::shared_ptr<xla::PjRtBuffer> buffer_reference;
+  // at::Tensor source_tensor;
```
JackCaoG, line 47: remove?

torch_xla/csrc/dl_convertor.cpp
```diff
@@ -0,0 +1,368 @@
...
+                   xla::CudaId());
+      return runtime::GetComputationClient()->LookupAddressableDevice(
+          context.device_id);
+    default:
+      return tsl::errors::InvalidArgument(
+          "Unknown/unsupported DLPack device type %d", context.device_type);
+  }
+}
+
+absl::StatusOr<xla::PrimitiveType> DLDataTypeToPrimitiveType(DLDataType type) {
```
JackCaoG, line 206: is there a similar function in Openxla? I am trying to avoid having duplicate code in our code base.

torch_xla/csrc/dl_convertor.cpp
```diff
@@ -0,0 +1,368 @@
...
+    // AcquireExternalReference may block
+    auto external_ref = pjrt_buffer->AcquireExternalReference();
+    XLA_CHECK_OK(external_ref.status());
+    pack->external_reference = std::move(external_ref.value());
+    xla::PjRtFuture<> future = pjrt_buffer->GetReadyFuture();
+    absl::Status status = future.Await();
+    XLA_CHECK_OK(status);
+  }
+  pack->buffer_reference = pjrt_buffer;
+  // pack->source_tensor = input;
```
JackCaoG, line 166: nit: remove

torch_xla/csrc/dl_convertor.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+  if (pjrt_buffer->IsTuple()) {
+    XLA_ERROR() << "Unimplemented. BufferToDLPackManagedTensor is not "
+                   "implemented for tuple buffers.";
+  }
+  if (pjrt_buffer->has_dynamic_dimensions()) {
+    XLA_ERROR() << "Unimplemented. DynamicShape is not implemented in DLPack.";
+  }
+
+  auto pack = std::make_unique<TorchXLADLMTensor>();
```
JackCaoG, line 154: why is it called pack?

test/test_operations.py
```diff
@@ -2441,6 +2445,115 @@ def test_unsafe_buffer_pointer(self):
...
+    self._test_dlpack_capsule_conversion_helper(xla_tensor_1)
+
+    # xtensor->CurrentDataHandle() == nullptr but xtensor->CurrentIrValue().node != nullptr and device_data != nullptr
+    xla_tensor_2 = torch.arange(5, dtype=dtype).to(xla_device)
+    self._test_dlpack_capsule_conversion_helper(xla_tensor_2)
+
+    xla_tensor_3 = torch.arange(5, dtype=dtype, device=xm.xla_device())
+    xm.mark_step()
+    # Without the `wait_device_ops()`, the pjrt buffer (pjrt_data->buffer) at https://github.com/pytorch/xla/blob/e3fc03314dab5f44e3ed9ccbba6c15fbca3285cd/torch_xla/csrc/runtime/pjrt_computation_client.cc#L467 will be nullptr.
+    xm.wait_device_ops()
```
JackCaoG, line 2496: I felt like this ux is not great, Can we just let the C++ code to get the PJRT_BUFFEr anyway and if it is empty, call `wait_device_ops` in C++. Asking user to always do `wait_device_ops` is a bit too much.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -463,12 +470,22 @@ std::uintptr_t PjRtComputationClient::UnsafeBufferPointer(
...
   return ptr.value();
 }
 
+std::shared_ptr<xla::PjRtBuffer> PjRtComputationClient::GetPjRtBuffer(
+    const DataPtr handle) {
+  std::shared_ptr<PjRtData> pjrt_data =
+      std::dynamic_pointer_cast<PjRtData>(handle);
+  XLA_CHECK(pjrt_data) << "handle must be PjRtData, got " << handle->ToString();
+  return pjrt_data->buffer;
+}
```
JackCaoG, line 487: why this method has to be a class method of `PjRtComputationClient`? It doesn;t access any class private members. If it is just a helper, you don't need this to b a class method. @will-cromar wdyt?

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -185,6 +185,13 @@ ComputationClient::DataPtr PjRtComputationClient::CreateDataPlaceholder(
   return std::make_shared<PjRtData>(std::move(device), std::move(shape));
 }
 
+ComputationClient::DataPtr PjRtComputationClient::CreateData(
+    std::string device, xla::Shape shape,
+    std::shared_ptr<xla::PjRtBuffer> pjrt_buffer) {
+  return std::make_shared<PjRtData>(std::move(device), std::move(shape),
+                                    pjrt_buffer);
+}
```
JackCaoG, line 193: I think you just need a helper, not the class method.

JackCaoG (changes requested):
mostly lgtm, minor nits.

test/test_operations.py
vanbasten23: done

torch_xla/csrc/dl_convertor.cpp
vanbasten23: Yeah [there](https://github.com/openxla/xla/blob/d63deb9250b9c212445290bd08c6effb5b6d0a2b/xla/python/dlpack.cc#L132) is. It was the first thing that I tried by reusing their function. But it didn't work because it requires me to take many unrelated dependencies such as nanobind, llvm stuff..

torch_xla/csrc/dl_convertor.cpp
vanbasten23: I think it is because the original struct is named `DLPackTensor` instead of `TorchXLADLMTensor`. I renamed TorchXLADLMTensor to DLPackTensor

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: The `struct PjRtData` is private to PjRtComputationClient

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: The `struct PjRtData` is private to PjRtComputationClient so I can't make it a helper.

vanbasten23:
hey @will-cromar @JackCaoG , it seems the CI skips tests that require PyTorch CUDA support, for example
```
test_aten_move_cuda_to_xla (__main__.TestGeneric) ... skipped 'requires PyTorch CUDA support'
test_aten_move_scalar_cuda_to_xla (__main__.TestGeneric) ... skipped 'requires PyTorch CUDA support'
```
so the tests I added in this PR are not actually run by the CI.

What do you think of adding a new CI workflow that builds pytorch with CUDA, builds pytorch/xla, and runs the test? By preserving the existing CI workflow (build pytorch with CUDA disabled), people can still get faster feedback.

will-cromar:
Yeah, that's a loose end from all of the refactoring I did. I agree with adding a new workflow or branch of the workflow that adds CUDA separately. 

If it's okay with you, the easiest option would be to download the pre-built nightly CUDA wheel from https://download.pytorch.org/whl/nightly/cu121. This will be faster and require less maintenance, but it will also break periodically if there's a breaking change on head since the last nightly build.

torch_xla/csrc/dl_convertor.cpp
```diff
@@ -0,0 +1,345 @@
...
+      get_data_handle(input);
+  XLA_CHECK(handle != nullptr)
+      << "Could not extract a valid data handle from the input tensor";
+
+  std::shared_ptr<xla::PjRtBuffer> pjrt_buffer =
+      runtime::GetComputationClient()->GetPjRtBuffer(handle);
+  XLA_CHECK(pjrt_buffer != nullptr) << "Could not get a valid pjrt_buffer";
+
+  if (pjrt_buffer->IsTuple()) {
+    XLA_ERROR() << "Unimplemented. BufferToDLPackManagedTensor is not "
```
will-cromar, line 125: micro-nit: use `XLA_CHECK(!pjrt_buffer->IsTuple()) << "message"` here and below

torch_xla/csrc/runtime/pjrt_computation_client.cc
will-cromar: Does the input `pjrt_buffer` depend on the current instance of `PjRtClient`? ie do you need to check here that this buffer belongs to this client? If you don't depend on any members of `PjRtComputationClient`, can you make this a static method to make that clearer?

Also, this should be private and probably not part of `ComputationClient`. The public API of `PjRtComputationClient` shouldn't operate on PJRT primitives. It should be a complete wrapper to leave the door open to other runtime interfaces (namely IFRT).

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -463,12 +470,31 @@ std::uintptr_t PjRtComputationClient::UnsafeBufferPointer(
...
   xla::StatusOr<std::uintptr_t> ptr =
       client_->UnsafeBufferPointer(pjrt_data->buffer.get());
   XLA_CHECK(ptr.ok());
   return ptr.value();
 }
 
+std::shared_ptr<xla::PjRtBuffer> PjRtComputationClient::GetPjRtBuffer(
+    const DataPtr handle) {
+  std::shared_ptr<PjRtData> pjrt_data =
+      std::dynamic_pointer_cast<PjRtData>(handle);
```
will-cromar, line 484: Do you want to handle sharded data?

torch_xla/csrc/runtime/ifrt_computation_client.h
```diff
@@ -84,6 +91,15 @@ class IfrtComputationClient : public ComputationClient {
         absl::AsciiStrToUpper(client_->platform_name()));
   };
 
+  xla::PjRtPlatformId GetPlatformID() const override {
+    return client_->platform_id();
```
will-cromar, line 95: Heads up: this method doesn't exist in the PJRT C API: https://github.com/openxla/xla/blob/80b854f9ef3e8b913ed9ca2d930c81e32c6d02da/xla/pjrt/pjrt_c_api_client.h#L209-L211

Since dlpack is a very special case, what do you think of just adding a `is_cuda` or `supports_dlpack` attribute to `PjRtComputationClient` and setting it during init?

will-cromar (commented):
Thanks! This is really neat

torch_xla/csrc/dl_convertor.cpp
vanbasten23: done

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: Not at this point.

torch_xla/csrc/runtime/ifrt_computation_client.h
vanbasten23: > Heads up: this method doesn't exist in the PJRT C API

The `client_` here has type `xla::ifrt::PjRtClient` which implements `platform_id` at https://github.com/openxla/xla/blob/62568f0ef380c883defd44a060722dd62e81df1b/xla/python/pjrt_ifrt/pjrt_client.h#L147-L150. I wonder how the `xla::ifrt::PjRtClient` relate to the PJRT C API `PjRtCApiTopologyDescription`.

>  what do you think of just adding a is_cuda or supports_dlpack attribute to PjRtComputationClient and setting it during init?

Can you elaborate more? Do you mean adding a `supports_dlpack ` to `ComputationClient` and set it true for `PjRtComputationClient ` and false to `IfRtComputationClient`?

torch_xla/csrc/runtime/pjrt_computation_client.cc
vanbasten23: Right, the `pjrt_buffer` doesn't depend on the current instance of `PjRtClient`. I have made it a static method.

I also made it not a part of `ComputationClient`. Not sure how I can make it private and call it in the dl_convertor.cpp at the same time.

ysiraichi (commented):
Really great work, @vanbasten23. I have left a few comments in the PR. I do have a couple of questions, though:

1. Could you add an overview of the functions involved in the DLPack-XLA (to/from) execution paths to the PR description? I feel like that would make your PR easier to follow.

2. Before converting CUDA to DLPack to XLA, don't we need to call `torch.cuda.synchronize()`? That's because IIRC, some CUDA computation is also lazy.

vanbasten23:
> Really great work, @vanbasten23. I have left a few comments in the PR. I do have a couple of questions, though:
> 
> 1. Could you add an overview of the functions involved in the DLPack-XLA (to/from) execution paths to the PR description? I feel like that would make your PR easier to follow.

Added the execution paths.

> 2. Before converting CUDA to DLPack to XLA, don't we need to call `torch.cuda.synchronize()`? That's because IIRC, some CUDA computation is also lazy.

Good callout. I couldn't find this in the pytorch's documentation. But from their test pytorch/test/test_dlpack.py, it seems it is only needed when `stream` is used. Since this PR is not supporting stream, so I think it is not needed.

test/test_operations.py
```diff
@@ -2441,6 +2445,113 @@ def test_unsafe_buffer_pointer(self):
...
+    # xtensor->CurrentDataHandle() == nullptr but xtensor->CurrentIrValue().node != nullptr and device_data != nullptr
+    self._test_dlpack_capsule_conversion_helper(xla_tensor_1)
+
+    # xtensor->CurrentDataHandle() == nullptr but xtensor->CurrentIrValue().node != nullptr and device_data != nullptr
+    xla_tensor_2 = torch.arange(5, dtype=dtype).to(xla_device)
+    self._test_dlpack_capsule_conversion_helper(xla_tensor_2)
+
+    xla_tensor_3 = torch.arange(5, dtype=dtype, device=xm.xla_device())
+    xm.mark_step()
+    self._test_dlpack_capsule_conversion_helper(xla_tensor_3)
```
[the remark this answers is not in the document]
vanbasten23: xla_tensor_0 vs xla_tensor_1: the former results in  `xtensor->CurrentDataHandle() != nullptr` with a mark_step. The latter (without mark_step) results in `xtensor->CurrentDataHandle() == nullptr but xtensor->CurrentIrValue().node != nullptr and device_data != nullptr`

xla_tensor_2 vs xla_tensor_3: the former create the tensor on cpu and uses `XLANativeFunctions::_to_copy` to move to the XLA device. The latter create a `arange_out` IR node.

They may overlap a bit but I included them to be complete.

test/test_operations.py
```diff
@@ -2441,6 +2445,113 @@ def test_unsafe_buffer_pointer(self):
...
+
+  @onlyIfTorchSupportsCUDA
+  @onlyIfPJRTDeviceIsCUDA
+  def test_dlpack_roundtrip_bool(self):
+    xla_tensor = torch.ones(1, dtype=torch.bool).to(xm.xla_device())
+    self._test_dlpack_capsule_conversion_helper(xla_tensor)
```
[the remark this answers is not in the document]
vanbasten23: The test `test_dlpack_roundtrip` skips the bool type because it'd otherwise fail with `"arange_cpu" not implemented for 'Bool'`

test/test_operations.py
```diff
@@ -2441,6 +2445,113 @@ def test_unsafe_buffer_pointer(self):
...
+
+  @onlyIfTorchSupportsCUDA
+  @onlyIfPJRTDeviceIsCUDA
+  def test_dlpack_pytorch_cuda_to_xla(self):
+    t1_cuda = torch.arange(5).cuda()
+    dlt1 = torch.utils.dlpack.to_dlpack(t1_cuda)
+    xla_t1 = xdlpack.from_dlpack(dlt1)
+    t1_cuda[0] = t1_cuda[0] + 20
+    self.assertTrue(torch.allclose(xla_t1.cpu(), t1_cuda.cpu()))
```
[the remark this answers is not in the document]
vanbasten23: Done

test/test_operations.py
```diff
@@ -2441,6 +2445,113 @@ def test_unsafe_buffer_pointer(self):
...
+
+    t2_cuda = torch.tensor(5).cuda()
+    dlt2 = torch.utils.dlpack.to_dlpack(t2_cuda)
+    xla_t2 = xdlpack.from_dlpack(dlt2)
+    t2_cuda.fill_(6)
+    self.assertTrue(torch.allclose(xla_t2.cpu(), t2_cuda.cpu()))
```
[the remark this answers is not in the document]
vanbasten23: done

test/test_operations.py
```diff
@@ -2441,6 +2445,113 @@ def test_unsafe_buffer_pointer(self):
...
+
+  @onlyIfTorchSupportsCUDA
+  @onlyIfPJRTDeviceIsCUDA
+  def test_dlpack_xla_to_pytorch_cuda(self):
+    xla_t1 = torch.arange(5).to(xm.xla_device())
+    dlt1 = xdlpack.to_dlpack(xla_t1)
+    cuda_t1 = torch.utils.dlpack.from_dlpack(dlt1)
+    cuda_t1[0] = cuda_t1[0] + 20
+    self.assertTrue(torch.allclose(xla_t1.cpu(), cuda_t1.cpu()))
```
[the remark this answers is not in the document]
vanbasten23: done

test/test_operations.py
```diff
@@ -2441,6 +2445,113 @@ def test_unsafe_buffer_pointer(self):
...
+        lambda: xdlpack.from_dlpack(t3.__dlpack__()))
+
+    t4 = cuda_t[1, :]
+    xla_t4 = xdlpack.from_dlpack(t4.__dlpack__())
+    self.assertTrue(torch.allclose(t4.cpu(), xla_t4.cpu()))
+
+    t5 = cuda_t[1]
+    xla_t5 = xdlpack.from_dlpack(t5.__dlpack__())
+    self.assertTrue(torch.allclose(t5.cpu(), xla_t5.cpu()))
+
```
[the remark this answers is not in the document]
vanbasten23: done

torch_xla/csrc/tensor_util.cpp
```diff
@@ -931,4 +932,25 @@ xla::PrimitiveType GetShapeDimensionType(
   return xla::PrimitiveType::S32;
 }
 
+std::shared_ptr<runtime::ComputationClient::Data> get_data_handle(
+    const at::Tensor& input) {
+  XLATensorPtr xtensor = bridge::GetXlaTensor(input);
+  XLA_CHECK(xtensor) << "The input has to be an XLA tensor.";
```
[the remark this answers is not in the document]
vanbasten23: removed it

test/test_operations.py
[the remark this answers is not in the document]
ysiraichi: And, I suppose, `xla_tensor_1` vs `xla_tensor_2` is tensor vs scalar, correct?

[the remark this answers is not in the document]
ysiraichi: I think it would be nice to have what you described as comments to each of them. Wdyt?

test/test_operations.py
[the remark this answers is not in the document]
ysiraichi: So, maybe it would be better to use `all_types_and_half()` instead of `all_types_and_complex_and(torch.half, ...)`

test/test_operations.py
```diff
@@ -2441,6 +2445,127 @@ def test_unsafe_buffer_pointer(self):
...
+    self._test_dlpack_capsule_conversion_helper(xla_tensor)
+
+  @onlyIfTorchSupportsCUDA
+  @onlyIfPJRTDeviceIsCUDA
+  def test_dlpack_pytorch_cuda_to_xla(self):
+    t1_cuda = torch.arange(5).cuda()
+    dlt1 = torch.utils.dlpack.to_dlpack(t1_cuda)
+    xla_t1 = xdlpack.from_dlpack(dlt1)
+    self.assertEqual(xla_t1.device.type, 'xla')
+    self.assertEqual(xla_t1.device.index, t1_cuda.device.index)
```
ysiraichi, line 2510: Is this correct? As far as I remember, even with multiple GPUs, XLA will have only one device: `xla:0`. Meanwhile, PyTorch may have `cuda:0` and `cuda:1`.

ysiraichi:
> it seems the CI skips tests that require PyTorch CUDA support, for example
> 
> ```
> test_aten_move_cuda_to_xla (__main__.TestGeneric) ... skipped 'requires PyTorch CUDA support'
> test_aten_move_scalar_cuda_to_xla (__main__.TestGeneric) ... skipped 'requires PyTorch CUDA support'
> ```

@will-cromar Do you mean that, at some point, we stopped using `.circleci/build.sh`? I remember adding `USE_CUDA=1` to it in #6070.

ysiraichi:
> it seems it is only needed when stream is used. Since this PR is not supporting stream, so I think it is not needed.

```python
>>> a = torch.rand(1024, 1024, 1024, device="cuda")

# This returns instantly
>>> for i in range(1000):
>>>    a = a @ a

# Does NOT return instantly. Can hear the GPU fans going up.
>>> torch.cuda.synchronize()
```

I think that, at first, we can just call `torch.cuda.synchronize()` and then do the transition using DLPack. That said, a better way to handle this would be to use the same stream object (in PyTorch and XLA), so that no explicit synchronization is actually needed.

will-cromar:
> > it seems the CI skips tests that require PyTorch CUDA support, for example
> > ```
> > test_aten_move_cuda_to_xla (__main__.TestGeneric) ... skipped 'requires PyTorch CUDA support'
> > test_aten_move_scalar_cuda_to_xla (__main__.TestGeneric) ... skipped 'requires PyTorch CUDA support'
> > ```
> 
> @will-cromar Do you mean that, at some point, we stopped using `.circleci/build.sh`? I remember adding `USE_CUDA=1` to it in #6070.

Yeah, that's right. The TPU CI, CPU/GPU CI, and nightly build were all using different build scripts. While splitting the XLA CUDA build from the `torch_xla` build, I moved them all to the ansible config at `infra/ansible`.

torch_xla/csrc/runtime/ifrt_computation_client.h
will-cromar: The PJRT C API is a subset of the PJRT C++ API, so some methods of the C++ -> C wrapper are unimplemented. I would add a field to `PjRtComputationClient` to indicate if dlpack is expected to work with the underlying `PjRtClient` (only true when `PJRT_DEVICE=CUDA`), since that depends on the specific device type, which we should not rely on after initialization.

I'm not concerned about IFRT right now. Also, remember that `xla::PjRtClient` is different than `xla::ifrt::PjRtClient`.

test/test_operations.py
vanbasten23: That's a very good point. I thought about it so I added a test:
```
cuda1 = torch.device('cuda:1')
    t3_cuda = torch.tensor(5, device=cuda1)
    dlt3 = torch.utils.dlpack.to_dlpack(t3_cuda)
    xla_t3 = xdlpack.from_dlpack(dlt3)
    self.assertEqual(xla_t3.device.type, 'xla')
    self.assertEqual(xla_t3.device.index, t3_cuda.device.index, msg='both value should 1. xla_t3.device should be xla:1.')
    print('xw32 line2528 xla_t3.device.index=', xla_t3.device.index, ', xla_t3.device=', xla_t3.device, ', t3_cuda.device.index=', t3_cuda.device.index)
```
Here it prints `xw32 line2528 xla_t3.device.index= 1 , xla_t3.device= xla:1 , t3_cuda.device.index= 1`. So we should be good here.
Moreover, when we do `xla_tensor.device`, LTC's https://github.com/pytorch/pytorch/blob/799f1460af159a1417bb3ef1f088347cb8d89434/torch/csrc/lazy/core/tensor.cpp#L104 will be called and the device from the [data](https://github.com/pytorch/pytorch/blob/799f1460af159a1417bb3ef1f088347cb8d89434/torch/csrc/lazy/core/tensor.h#L27) will be used. The device from the `data` is set in the dlpack_convertor.cc. I think that's why it works.

torch_xla/csrc/runtime/ifrt_computation_client.h
vanbasten23: I see. If the concern is dlpack should only be used when PJRT_DEVICE=CUDA, then there is a check in this PR https://github.com/pytorch/xla/pull/7025/files#diff-cf3922091c803bce3341e4e55b2c54d277812059adeafa297eb1c7b444213b2aR45-R53 to make sure dlpack is used for CUDA instead of TPU.

test/test_operations.py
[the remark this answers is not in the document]
vanbasten23: SG. It's much cleaner.

test/test_operations.py
ysiraichi: Wow. You are right. I tried calling `xm.get_xla_supported_devices()` on a multi-gpu machine and got `[xla:0, xla:1]`. I remember this wasn't the case some time ago, e.g. last year. I remember even asking about this and being told that we could only do that by using multiple process. Do you know what changed?

vanbasten23:
> Yeah, that's a loose end from all of the refactoring I did. I agree with adding a new workflow or branch of the workflow that adds CUDA separately.
> 
> If it's okay with you, the easiest option would be to download the pre-built nightly CUDA wheel from https://download.pytorch.org/whl/nightly/cu121. This will be faster and require less maintenance, but it will also break periodically if there's a breaking change on head since the last nightly build.

Thanks for the suggestion. I've seen a few incompatible torch wheel and torch_xla wheel recently so I feel it is not very uncommon. If they are incompatible, our CI will be red for a day or two and we'll have to communicate to the team which is kind of a hassle lol. So I'll strive for building pytorch with cuda from source.

torch_xla/csrc/dl_convertor.cpp
```diff
@@ -0,0 +1,341 @@
...
+#include "torch_xla/csrc/runtime/tf_logging.h"
+#include "torch_xla/csrc/tensor.h"
+#include "torch_xla/csrc/tensor_util.h"
+#include "torch_xla/csrc/unwrap_data.h"
+#include "xla/pjrt/pjrt_client.h"
+#include "xla/pjrt/pjrt_future.h"
+#include "xla/pjrt/pjrt_layout.h"
+#include "xla/status.h"
+
+namespace torch_xla {
```
ysiraichi, line 21: Just checking: the only thing in this file that you copied from XLA was the function `DLDataTypeToPrimitiveType`, correct?

vanbasten23: There are a few relevant functions, such as PrimitiveTypeToDLDataType, DLDeviceForDevice, DeviceForDLDevice, StridesToLayout

ysiraichi: Could add a comment to all copied functions, so that we can easily keep track of?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2508,6 +2540,24 @@ void InitXlaModuleBindings(py::module m) {
...
+    // "dltensor", dlPack_Capsule_Destructor));
+    return PyCapsule_New(dlMTensor, "dltensor", dlPack_Capsule_Destructor);
+  });
+  // m.def("_to_dlpack", &tensor_toDLPack, ""); //
+
+  // from a dlpack tensor to an XLA tensor
+  m.def("_from_dlpack", [](py::handle ext_data) -> at::Tensor {
+    return tensor_fromDLPack(ext_data.ptr());
+  });
+
```
ysiraichi, line 2560: Could you leave a comment for each of them, stating what are the assumptions you made when creating those functions? e.g. before calling `_from_dlpack(ext_data)`, we should synchronize the CUDA device if `ext_data` is the result of a CUDA computation. Same thing for `_to_dlpack(input)`.

ysiraichi (approved):
I think we need more comments, e.g.:
- where each function was copied from? 
- how to use the added API correctly?

Otherwise, LGTM.

JackCaoG:
I will try to take a look tmr

torch_xla/csrc/dl_convertor.cpp
vanbasten23: Added a comment to each function.

torch_xla/csrc/init_python_bindings.cpp
vanbasten23: Added it.

vanbasten23:
> I think we need more comments, e.g.:
> 
> * where each function was copied from?
> * how to use the added API correctly?
> 
> Otherwise, LGTM.

Thanks for the review. I've added the comments you suggested.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2508,6 +2540,29 @@ void InitXlaModuleBindings(py::module m) {
                          "without a data handle or an IR.";
         });
 
+  // from an XLA tensor to a dlpack tensor.
+  // If ext_data is the result of an CUDA computation, we should synchronize
+  // (waits for all kernels in all streams on a CUDA device to complete) if the
+  // current stream is different from the ext_data's stream. Otherwise, we may
+  // risk of getting incorrect results.
```
ysiraichi, line 2547: I believe this is `_from_dlpack`'s comment.

vanbasten23:
hey @JackCaoG could you take a look at the PR when you are available? Thanks!

JackCaoG (approved):

vanbasten23:
Thanks for the review Jack and Yukio!

torch_xla/csrc/dl_convertor.cpp
```diff
@@ -0,0 +1,345 @@
...
+  xla::StatusOr<std::unique_ptr<xla::PjRtBuffer>> pjrt_buffer =
+      device->client()->CreateViewOfDeviceBuffer(
+          static_cast<char*>(dlmt->dl_tensor.data) +
+              dlmt->dl_tensor.byte_offset,
+          shape, device, on_delete_callback);
+  XLA_CHECK_OK(pjrt_buffer.status()) << "Failed to create a pjrt buffer.";
+  XLA_CHECK(pjrt_buffer.value() != nullptr) << "pjrt buffer is null.";
+
+  runtime::ComputationClient::DataPtr data =
+      runtime::PjRtComputationClient::CreateData(
```
zpcore, line 336: QQ: Does this only apply to PjRtComputationClient?

zpcore: Background: I am working on a new ComputationClient and here reports the error:
```
torch_xla/csrc/dl_convertor.cpp:337:16: error: 'torch_xla::runtime::PjRtComputationClient' has not been declared
  337 |       runtime::PjRtComputationClient::CreateData(
      |                ^~~~~~~~~~~~~~~~~~~~~
```
Looks like we have to use PjRtComputationClient with this file.

ysiraichi: I think so. As far as I understand it, the DLPack API only works with XLA:CUDA, i.e. PJRT.

PR #7032, opened 2024-05-06
Megablox `tgmm` kernel implementation

miladm:
Megablox `tgmm` kernel implementation

### DIFF
--- a/test/test_megablox.py
+++ b/test/test_megablox.py
@@ -71,6 +71,7 @@ class MegabloxTest(unittest.TestCase):
     self.tests_cases = []
     self.tests_cases.append({
         'dtype': torch.float32,
+        'jax_dtype': jnp.float32,
         'm': 128,
         'k': 128,
         'n': 128,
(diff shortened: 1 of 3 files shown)


miladm:
Last step before review / merge:
The reference tgmm implementation should uses self._reference_gmm and run a grad call
(JAX code for ref: https://github.com/google/jax/blob/main/tests/pallas/gmm_test.py#L374-L384)

cc @wonjoolee95

wonjoolee95:
Picking this back up since gmm PR is now merged. While trying to compare the output from our tgmm kernel and jax kernel, our implementation of kernel seems to be wrong. The values from the two kernels differ by a lot, and some of the test cases cause our kernel to seg fault (while they all pass in the jax tgmm kernel) -- https://gist.github.com/wonjoolee95/335c33d5c8c0fab64f293a7b80a8706f. 

I'll need to review how the tgmm kernel works and work on a fix.

PR #7034, opened 2024-05-07
[Pallas] Improve FlashAttention segment_ids test case

alanwaketan:
Summary:
Make the test case more useful. The original test case will just create a mask that select everything, lol.

Test Plan:


JackCaoG (approved):

alanwaketan:
Thanks, Jack!

PR #7039, opened 2024-05-08
add a metrics to track persistent cache loading time

JackCaoG:


torch_xla/csrc/runtime/cache.h
```diff
@@ -162,6 +162,7 @@ class PersistentCache : public AbstractCache<K, T, H, E> {
   // if the key is tracked in memory, and if not it will check for a persisted
   // version on disk.
   TypePtr Get(const K& key) override {
+    TORCH_LAZY_TIMED("PersistentCacheGet");
```
jonb377, line 165: This will also trigger for binaries that are already loaded and cached in-memory, what do you think of moving it to ~L176 to only capture the load+deserialize logic? Maybe renamed to `TORCH_LAZY_TIMED("PersistentCacheLoad")`

JackCaoG: sg, let me update

PR #7044, opened 2024-05-10
[FSDPv2] Support MultiSlice

alanwaketan:
Summary:
This pull request adds the multi-slice support for FSDPv2. Basically, the default setup is to use the dcn axis as the data axis, and it means we only do data parallel over multi-slices. In the future, we could also support FSDP over mutli-slices.

Test Plan:
PJRT_DEVICE=TPU python test/spmd/test_fsdp_v2.py

### DIFF
--- a/test/spmd/test_fsdp_v2.py
+++ b/test/spmd/test_fsdp_v2.py
@@ -142,23 +142,23 @@ class FSDPv2Test(test_xla_sharding_base.XlaShardingTest):
   @unittest.skipIf(xr.device_type() != 'TPU', "This test only works on TPU.")
   def test_fsdp_v2_multi_slice(self):
     model = self.SimpleLinear().to(xm.xla_device())
-    mesh = self._get_mesh((2, self.n_devices // 2, 1), None, ('data', 'fsdp', 'tensor'))
+    mesh = self._get_mesh((2, self.n_devices // 2, 1), None,
+                          ('data', 'fsdp', 'tensor'))
     model = FSDPv2(model, mesh=mesh, extra_data_axis="data")
 
     # Make sure all weights are sharded.
     annotation = '{devices=[2,1,2]0,2,1,3 last_tile_dim_replicate}'
     self.assertEqual(annotation,
-                      torch_xla._XLAC._get_xla_sharding_spec(model.fc1.weight))
+                     torch_xla._XLAC._get_xla_sharding_spec(model.fc1.weight))
     self.assertEqual(annotation,
-                      torch_xla._XLAC._get_xla_sharding_spec(model.fc2.weight))
+                     torch_xla._XLAC._get_xla_sharding_spec(model.fc2.weight))
 
     x = torch.randn(16, 128).to(xm.xla_device())
     xs.mark_sharding(x, mesh, (('data', 'fsdp'), None))
     output = model(x)
     # Make sure output are sharded.
     annotation = '{devices=[4,1]0,2,1,3}'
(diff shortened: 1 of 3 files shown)


torch_xla/experimental/spmd_fully_sharded_data_parallel.py
```diff
@@ -24,6 +24,8 @@ def _prepare_spmd_partition_spec(param):
   # TODO: should we shard on the maximal dim for param? Then we need
   # another helper for the output.
   partition_spec[0] = "fsdp"
+  if extra_data_axis:
+    partition_spec[0] = ("fsdp", extra_data_axis)
```
jonb377, line 28: We usually have this reversed for DCN - `(extra_data_axis, 'fsdp')`. The axes should be in order of increasing network intensity in the mesh, and the order in the partition spec will impact the sharding.

torch_xla/experimental/spmd_fully_sharded_data_parallel.py
```diff
@@ -44,10 +46,12 @@ class SpmdFullyShardedDataParallel(nn.Module):
   def __init__(
       self,
       module: nn.Module,
+      *,
       mesh: Optional[spmd.Mesh] = None,
       shard_output: Optional[Callable] = None,
       auto_wrap_policy: Optional[Callable] = None,
       auto_wrapper_callable: Optional[Callable] = None,
+      extra_data_axis: Optional[str] = None,
```
jonb377, line 54: What do you think of calling it `replica_axis` instead of `extra_data_axis`?

test/spmd/test_fsdp_v2.py
```diff
@@ -139,6 +139,59 @@ def test_fsdp_v2_cpu_model(self):
...
+    self.assertEqual(annotation,
+                     torch_xla._XLAC._get_xla_sharding_spec(model.fc1.weight))
+    self.assertEqual(annotation,
+                     torch_xla._XLAC._get_xla_sharding_spec(model.fc2.weight))
+
+    x = torch.randn(16, 128).to(xm.xla_device())
+    xs.mark_sharding(x, mesh, (('data', 'fsdp'), None))
+    output = model(x)
+    # Make sure output are sharded.
+    annotation = '{devices=[4,1]0,2,1,3}'
```
jonb377, line 160: This would be different than `x`'s sharding - x should have `[4,1]0,1,2,3` with the iota mesh. I left a comment below, I think we should reverse the order in `_prepare_spmd_partition_spec`.

jonb377 (approved):
LGTM, thanks Jiewen!

alanwaketan:
@JackCaoG The TPU CI doesn't seem running even with the label.

test/spmd/test_fsdp_v2.py
alanwaketan: My mistake. Thanks for pointing it out.

torch_xla/experimental/spmd_fully_sharded_data_parallel.py
alanwaketan: I think replica_axis is too tied to the underneath technology while users may only be familiar with data paralle, fsdp, and tensor parallel.

JackCaoG:
yea I think they only check the label when CI is being run. It is OK if you have any changes and repush it will run, otherwise we can let head to check.

alanwaketan:
I'm landing it. If the master TPU CI breaks, let's deal with that later.

PR #7047, opened 2024-05-10
[test] Fori loop simple case test post order linear model wrapped body clean version while loop only

ManfeiBai:
script PR contain
- buidlfori_loop & tests
- _get_xla_computation & tests
- target linear


ManfeiBai:
continue tracking in https://github.com/pytorch/xla/pull/7094, let's close this pr for now

PR #7048, opened 2024-05-11
Add simple example for how to use torch_xla

JackCaoG:
TODO:
1. add readme
2. add these to TPU tests


qihqi (approved):

examples/train_resnet_base.py
```diff
@@ -0,0 +1,72 @@
...
+        sample_count=train_dataset_len // self.batch_size //
+        xm.xrt_world_size())
+
+    self.device = xm.xla_device()
+    self.train_device_loader = pl.MpDeviceLoader(train_loader, self.device)
+    self.model = torchvision.models.resnet50().to(self.device)
+    self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
+    self.loss_fn = nn.CrossEntropyLoss()
+
+  def run_optimizer(self):
```
will-cromar, line 41: Is the only reason we need a class here to override `optimizer.step` here?

examples/train_resnet_base.py
```diff
@@ -0,0 +1,72 @@
...
+time.ctime()
+
+
+def _train_update(step, loss, tracker, epoch):
+  print(f'epoch: {epoch}, step: {step}, loss: {loss}, rate: {tracker.rate()}')
+
+
+class TrainResNetBase():
+
+  def __init__(self):
```
will-cromar, line 23: If this stays a class, use a `dataclass` here so we can update the batch size etc without having to write out the whole constructor.

Also, IMO, Python's constructors are ugly and I like to hide them as much as possible.

examples/train_resnet_base.py
```diff
@@ -0,0 +1,72 @@
...
+      self.model.train()
+      for step, (data, target) in enumerate(loader):
+        self.optimizer.zero_grad()
+        output = self.model(data)
+        loss = self.loss_fn(output, target)
+        loss.backward()
+        self.run_optimizer()
+        tracker.add(self.batch_size)
+        if step % 10 == 0:
+          xm.add_step_closure(_train_update, args=(step, loss, tracker, epoch))
```
will-cromar, line 57: This is a good opportunity to be the change we want to see in the world and use `xla.sync()` instead. There's no performance reason to use a step closure here, and I think `with xla.sync()` is clearer.

examples/train_resnet_base.py
```diff
@@ -0,0 +1,72 @@
...
+        if self.num_steps == step:
```
will-cromar, line 58: nit: prefer `islice`ing the loader instead of adding a separate stopping condition

https://stackoverflow.com/a/5234170

examples/train_resnet_base.py
will-cromar:
```
    with xla.sync():
        self.optimizer.zero_grad()
        output = self.model(data)
        loss = self.loss_fn(output, target)
        loss.backward()
        self.run_optimizer()
    
    tracker.add(self.batch_size)
    if step % 10 == 0: print(...)
```

examples/train_resnet_base.py
```diff
@@ -0,0 +1,72 @@
...
+
+import os
+import time
+
+import torch
+import torchvision
+import torch.optim as optim
+import torch.nn as nn
+
+time.ctime()
```
will-cromar, line 14: What is this statement for?

examples/train_resnet_base.py
```diff
@@ -0,0 +1,72 @@
...
+        tracker.add(self.batch_size)
+        if step % 10 == 0:
+          xm.add_step_closure(_train_update, args=(step, loss, tracker, epoch))
+        if self.num_steps == step:
+          break
+
+    for epoch in range(1, self.num_epochs + 1):
+      xm.master_print('Epoch {} train begin {}'.format(
+          epoch, time.strftime('%l:%M%p %Z on %b %d, %Y')))
+      train_loop_fn(self.train_device_loader, epoch)
```
will-cromar, line 64: Can you just inline the train loop function here?

examples/train_resnet_base.py
```diff
@@ -0,0 +1,72 @@
+from torch_xla import runtime as xr
```
will-cromar, line 1: Can you move this file to `torch_xla/examples` so it's included in our package? Then we can run examples like `python -m torch_xla.examples.train_resnet_ddp` directly from the installed package.

will-cromar (commented):
Thanks for starting this PR. Our various resnet examples were getting unwieldy, and I like that this is more straightforward.

My only concern here is that the `TrainResNetBase` might be too unwieldy. It sort of looks like a stripped down `accelerate` wrapper or `LightningModule`. IMO these examples should not necessarily focus on reusable code, but more on showcasing the order of execution in a training loop in a really clear way. Flax's examples do this well: https://github.com/google/flax/blob/main/examples/README.md (they have more model definition than we do because the ecosystem doesn't have an equivalent of `torch{vision,text,audio}`)

Ideally we should be able to strip this down to a really simple function like `train_resnet` and different variations do different setup/teardown. I do see that there is at least one case where a child method accesses some common internal state (`run_optimizer` needs `optimizer`), so that may not be possible here.

examples/train_resnet_base.py
JackCaoG: yea

examples/train_resnet_base.py
JackCaoG: sure, is `sync` change already merged?

examples/train_resnet_base.py
JackCaoG: oh, forgot to remove..

examples/train_resnet_base.py
JackCaoG: Hmm dynamo works like `torch.compile(trainer, backend="openxla")`. For normal case I can inline them.

examples/train_resnet_base.py
JackCaoG: do we want to include them as part of our package through?

JackCaoG:
My thinking is that I want to make it really clear how simple it is to enable some features (like xla_ddp, profiling). Having them in a separate file makes it very clear what code user needs to add to enable those features.

examples/train_resnet_base.py
JackCaoG: I actually don't know how dataclass works haha. If you don't mind, can you starting a pr after this one to fix it?

examples/train_resnet_base.py
JackCaoG: actually..hmm, if we want to do that we might want a way to disable `mark_step` from the loader, otherwise it will be a bit confusing when they turn on `PT_XLA_DEBUG`..

will-cromar (approved):

examples/train_resnet_base.py
JackCaoG: Let me think about this in a follow up pr. Calling `mark_step` is not free, one of the optimizations in inference is to remove any additional `mark_step` that's not required. If we were to add `sync` here, we want to somehow disable the `mark_step` in the loader.

examples/train_resnet_base.py
JackCaoG: sure I can move.

JackCaoG: oh wait... we depends on torchvision in this example... I don't want to force that dependency in our package.. Let's pause this move until we figure out how to deal with that..

PR #7052, opened 2024-05-13
Pin update 20240513

qihqi:

PR #7059, opened 2024-05-14
reenable disabled pt2e test

lsy323:
Fixes the disabled test in https://github.com/pytorch/xla/pull/7052

Workaround https://github.com/pytorch/pytorch/issues/126189

Update:
Add the step of running example inputs through pt2e graph before convert resolves the issue. Detail at https://github.com/pytorch/pytorch/issues/126189#issuecomment-2110933920

cc @sdasgup3 @paulinesho


sdasgup3 (approved):

sdasgup3:
@lsy323 Do we need this if we can run the model using sample inputs before quantization.

lsy323:
> @lsy323 Do we need this if we can run the model using sample inputs before quantization. 

No, I'll just update the test script.

Issue #6989, opened 2024-04-29
Loading large sharded model weights


adithyaiyer1999:
## ❓ Questions and Help

Hi!

I'm trying to enable checkpointing of our training runs, which involves saving the model weights, and the optimzer states, and then when i want to continue from the last checkpoint - load both into the model.

My problem is - if the model is really big, loading the entire model and opt state is impossible due to memory contraints (on v4 tpus) - so I am attempting to load my sharded models. 

The way i save the model weights, inside each worker/processis as follows :

```
# ckpt_path is path name saved with a string with a number inidicating which process 
state_dict = self.model.state_dict()
        cpu_state_dict = {
                key: value.cpu()
                for key, value in state_dict.items()
            }
os.makedirs(os.path.dirname(ckpt_path), exist_ok=True)
xm.save(ckpt, ckpt_path, master_only=False)
```

This creates say 8 separate model shards for say a TPU-v4-16. When i load back, I'm trying to use:

```
state_dict = torch.load(path)
state_dict = state_dict["model"]
xm.send_cpu_data_to_device(state_dict, xm.xla_device())
self.model.load_state_dict(state_dict)
```

Im mostly referring to https://github.com/pytorch/xla/blob/master/docs/fsdp.md , but even here, the load -> then shard. I want to directly load sharded weights into each process. Currently, Im running into the following error when i do model.forward:
```
indices should be either on cpu or on the same device as the indexed tensor (XLA). When using XLA, the indexed tensor must be an XLA tensor.
```

Essentially, although my model weights are xla tensors, the maybe seem to be in the wrong device? i've verified that everything is in xla device, although when i print `xm.xla_device()`, it always shows `xla:0` (this makes sense because each process can only see it's device).

Any best practices on loading from shards for really large models?

Thanks!

JackCaoG:
You seems to be using FSDP, @jonb377 @alanwaketan any experience you can share for loading the sharded weights?

jonb377:
Gerson has proposed loading the model layer-by-layer to limit host memory usage: https://github.com/pytorch/xla/issues/6822#issuecomment-2021286687

I'm not sure how the optimizer is handled for FSDPv1 checkpoints though, @gkroiz any thoughts?

gkroiz:
@adithyaiyer1999 to fix the error you are seeing when loading a sharded checkpoint, could you try the following in order(if you haven't tried already):
[...]

PR #7061, opened 2024-05-14
add example for fsdp

JackCaoG:
@gkroiz FYI, for some reason I can't add you as a reviewer.

### DIFF
--- a/examples/train_resnet_fsdp_auto_wrap.py
+++ b/examples/train_resnet_fsdp_auto_wrap.py
@@ -35,12 +35,11 @@ class TrainResNetXLAFSDP(TrainResNetBase):
           })
     else:
       raise Exception(f"Invalid auto-wrap policy: {auto_wrap_policy}")
-    fsdp_wrap = lambda m: FSDP(
-        m,
+    self.model = FSDP(
+        self.model,
         compute_dtype=torch.float32,
         pin_layout_in_collective_ops=True,
         auto_wrap_policy=auto_wrap_policy)
-    self.model = fsdp_wrap(self.model)
     self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
 
 
--- a/examples/train_resnet_fsdp_auto_wrap.py
+++ b/examples/train_resnet_fsdp_auto_wrap.py
@@ -1,0 +1,53 @@
+from train_resnet_base import TrainResNetBase
+from functools import partial
+
+import torch
+import torchvision
+import torch.optim as optim
+import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla.core.xla_model as xm
+from torch_xla.distributed.fsdp import XlaFullyShardedDataParallel as FSDP, checkpoint_module
+from torch_xla.distributed.fsdp.wrap import (size_based_auto_wrap_policy,
+                                             transformer_auto_wrap_policy)
+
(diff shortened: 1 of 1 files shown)


alanwaketan:
Given FSDPv1 is more or less deprecated, I don't think we should add it as an example. We should just focus on SPMD instead.

gkroiz:
Thanks for adding this @JackCaoG, I have a few questions.

1) Do we want to add a simple demonstration of checkpointing? I know there was some recent questions around this, namely #6989

2) How does this files purpose differ to the FSDP example code in https://github.com/pytorch/xla/blob/master/test/test_train_mp_mnist_fsdp_with_ckpt.py

JackCaoG:
> Thanks for adding this @JackCaoG, I have a few questions.
> 
> 1. Do we want to add a simple demonstration of checkpointing? I know there was some recent questions around this, namely [Loading large sharded model weights #6989](https://github.com/pytorch/xla/issues/6989)
> 2. How does this file's purpose differ to the FSDP example code in [`master`/test/test_train_mp_mnist_fsdp_with_ckpt.py](https://github.com/pytorch/xla/blob/master/test/test_train_mp_mnist_fsdp_with_ckpt.py?rgh-link-date=2024-05-14T21%3A01%3A30Z)

Old example is kind of long, I want to make it very clear what's the minimal codes to run a feature. If user want to do a deeper dive we should directly them to the more completed example.

JackCaoG:
> Given FSDPv1 is more or less deprecated, I don't think we should add it as an example. We should just focus on SPMD instead.

I see your point, but there are users that want to use FSDP v1. It is similar that we don't really want user to use native DDP but I provide the example anyway. I think we can update our README to specify what do we recommened, in that section we can make it clear that we recommend `xla_ddp` and `fsdp_v2`.

JackCaoG:
@gkroiz for checkponting and gradident checkpointing, I think we should provide a similar simple examples. Let me know if you are interested in contributing.

gkroiz:
> @gkroiz for checkponting and gradident checkpointing, I think we should provide a similar simple examples. Let me know if you are interested in contributing.

Happy to review, probably don't have time in the near future.

examples/train_resnet_fsdp_auto_wrap.py
```diff
@@ -0,0 +1,53 @@
...
+      auto_wrap_policy = partial(
+          transformer_auto_wrap_policy,
+          transformer_layer_cls={
+              torchvision.models.resnet.BasicBlock,
+              torchvision.models.resnet.Bottleneck,
+              torchvision.models.vision_transformer.EncoderBlock,
+          })
+    else:
+      raise Exception(f"Invalid auto-wrap policy: {auto_wrap_policy}")
+    fsdp_wrap = lambda m: FSDP(
```
Liyang90, line 38: Why is the lambda functions needed? One case I can think of would be using it as a callable argument of `auto_wrapper_callable`. But since it is not the case, can we use `FSDP()` directly on the model for simplicity?

Liyang90 (approved):

examples/train_resnet_fsdp_auto_wrap.py
JackCaoG: good catch, I just copied it from the example. Let me clean it up

gkroiz (approved):
LGTM

PR #7063, opened 2024-05-14
add DDP with SPMD example

JackCaoG:


examples/train_resnet_spmd_data_parallel.py
```diff
@@ -0,0 +1,45 @@
...
+class TrainResNetXLASpmdDDP(TrainResNetBase):
+
+  def __init__(self):
+    super().__init__()
+    # Shard along batch dimension only
+    num_devices = xr.global_runtime_device_count()
+    device_ids = np.arange(num_devices)
+    mesh_shape = (num_devices, 1, 1, 1)
+    # We know data is 4d and 0th dimension is the batch dimension
+    input_mesh = xs.Mesh(device_ids, mesh_shape, ('B', 'C', 'W', 'H'))
```
jonb377, line 28: I think we should revisit the mesh shape for these examples - right now it's based on the input tensor shape, but in general we want to reuse the mesh across input and model parameters.

We could do something like MaxText or the shardings example, with `data`, `fsdp`, and `tensor` axes, and recombine them using the partition spec (in this case only with the `data` axis):

```python
mesh_shape = (num_devices,)
mesh = xs.Mesh(device_ids, mesh_shape, ('data',))

# Shard the input's batch dimension along the `data` axis, no sharding along other dimensions
input_sharding=xs.ShardingSpec(mesh, ('data', None, None, None))
```

jonb377 (approved):

examples/train_resnet_spmd_data_parallel.py
JackCaoG: Great! Let me update the example.

alanwaketan (approved):
LGTM.

PR #7064, opened 2024-05-15
[test] Fori loop simple case test post order linear model wrapped body clean version while loop only new args

ManfeiBai:
this PR contain modification to enable simple mnist test case run


ManfeiBai:
CI test keep queued for long time, created clean version with less commits in https://github.com/pytorch/xla/pull/7094 for CI test results

ManfeiBai:
since PR is old, and would track code in https://github.com/pytorch/xla/pull/7094, let's close this PR for now

Issue #6833, opened 2024-03-27
[torchbench] `detectron2_fcos_r_50_fpn` fails to run inference.


ysiraichi:
## 🐛 Bug

Running the upstreamed benchmarking scripts with the following command results in an unexpected error.

```bash
python xla/benchmarks/experiment_runner.py \
       --suite-name torchbench \
       --accelerator cuda \
       --xla PJRT \
       --dynamo openxla --dynamo None \
       --test eval \
       --repeat 8 --iterations-per-run 1 \
       --print-subprocess \
       --no-resume -k speech_transformer
```

```python
Traceback (most recent call last):
  File ""xla/benchmarks/experiment_runner.py"", line 945, in <module>
    main()
  File ""xla/benchmarks/experiment_runner.py"", line 941, in main
    runner.run()
  File ""xla/benchmarks/experiment_runner.py"", line 61, in run
    self.run_single_config()
  File ""xla/benchmarks/experiment_runner.py"", line 256, in run_single_config
    metrics, last_output = self.run_once_and_gather_metrics(
  File ""xla/benchmarks/experiment_runner.py"", line 345, in run_once_and_gather_metrics
    output, _ = loop(iter_fn=self._default_iter_fn)
  File ""xla/benchmarks/experiment_runner.py"", line 302, in loop
    output, timing, trace = iter_fn(benchmark_experiment, benchmark_model,
  File ""xla/benchmarks/experiment_runner.py"", line 218, in _default_iter_fn
    output = benchmark_model.model_iter_fn(
  File ""/home/ysiraichi/pytorch/torch/_dynamo/eval_frame.py"", line 390, in _fn
    return fn(*args, **kwargs)
  File ""/home/ysiraichi/pytorch/xla/benchmarks/benchmark_model.py"", line 170, in eval
    pred = self.module(*inputs)
  File ""/home/ysiraichi/pytorch/torch/nn/modules/module.py"", line 1527, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File ""/home/ysiraichi/pytorch/torch/nn/modules/module.py"", line 1536, in _call_impl
    return forward_call(*args, **kwargs)
  File ""/home/ysiraichi/.local/lib/python3.8/site-packages/detectron2/modeling/meta_arch/dense_detector.py"", line 95, in forward
    images = self.preprocess_image(batched_inputs)
[...]

Issue #6556, opened 2024-02-16
[torchbench] `Super_SloMo` failing on inference with dynamo after `bfloat16` conversion.


ysiraichi:
## 🐛 Bug

After converting the `Super_SloMo` model to `bfloat16`, running inference on dynamo raises the following error:

```bash
python xla/benchmarks/experiment_runner.py \
    --suite-name torchbench --accelerator cuda \
    --xla PJRT --dynamo openxla --test eval \
    --no-resume --print-subprocess \
    -k Super_SloMo
```

```python
Traceback (most recent call last):
  File "xla/benchmarks/experiment_runner.py", line 914, in <module>
    main()
  File "xla/benchmarks/experiment_runner.py", line 910, in main
    runner.run()
  File "xla/benchmarks/experiment_runner.py", line 59, in run
    self.run_single_config()
  File "xla/benchmarks/experiment_runner.py", line 254, in run_single_config
    metrics, last_output = self.run_once_and_gather_metrics(
  File "xla/benchmarks/experiment_runner.py", line 331, in run_once_and_gather_metrics
    output, _ = loop(iter_fn=self._default_iter_fn)
  File "xla/benchmarks/experiment_runner.py", line 300, in loop
    output, timing, trace = iter_fn(benchmark_experiment, benchmark_model,
  File "xla/benchmarks/experiment_runner.py", line 216, in _default_iter_fn
    output = benchmark_model.model_iter_fn(
  File "torch/_dynamo/eval_frame.py", line 454, in _fn
    return fn(*args, **kwargs)
  File "xla/benchmarks/benchmark_model.py", line 168, in eval
    def eval(self, inputs, collect_full_output=False):
  File "torch/_dynamo/eval_frame.py", line 454, in _fn
    return fn(*args, **kwargs)
  File "torch/_dynamo/external_utils.py", line 25, in inner
    return fn(*args, **kwargs)
  File "torch/_functorch/aot_autograd.py", line 893, in forward
    return compiled_fn(full_args)
  File "torch/_functorch/_aot_autograd/utils.py", line 79, in g
    return f(*args)
  File "torch/_functorch/_aot_autograd/runtime_wrappers.py", line 101, in runtime_wrapper
    all_outs = call_func_at_runtime_with_args(
  File "torch/_functorch/_aot_autograd/utils.py", line 103, in call_func_at_runtime_with_args
    out = normalize_as_list(f(args))
[...]

PR #6518, opened 2024-02-10
[benchmarks] Default to `bfloat16` (inference) and AMP (training) precision.

ysiraichi:
Fix: #6483 

This PR makes `bfloat16` the default data-type for inference, and AMP the default execution mode for training. This follows the execution found in the PyTorch HUD.

cc @miladm

Issue #6511, opened 2024-02-09
Autocast fails on dynamo.


ysiraichi:
## 🐛 Bug

```python
@torch.compile(backend="openxla")
def foo(x):
    with torch_xla.amp.autocast(xm.xla_device(), dtype=torch.bfloat16):
        y = x * 5
    return y
x = torch.rand(5, device=xm.xla_device())
foo(x)
```

```
Traceback (most recent call last):
  File "examples/autocast.py", line 12, in <module>
    foo(x)
  File "torch/_dynamo/eval_frame.py", line 454, in _fn
    return fn(*args, **kwargs)
  File "examples/autocast.py", line 7, in foo
    with autocast(xm.xla_device(), dtype=torch.bfloat16):
  File "examples/autocast.py", line 7, in torch_dynamo_resume_in_foo_at_7
    with autocast(xm.xla_device(), dtype=torch.bfloat16):
  File "torch/_dynamo/convert_frame.py", line 904, in catch_errors
    return callback(frame, cache_entry, hooks, frame_state, skip=1)
  File "torch/_dynamo/convert_frame.py", line 769, in _convert_frame
    result = inner_convert(
  File "torch/_dynamo/convert_frame.py", line 398, in _convert_frame_assert
    return _compile(
  File "/usr/local/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "torch/_dynamo/convert_frame.py", line 696, in _compile
    raise InternalTorchDynamoError(str(e)).with_traceback(
  File "torch/_dynamo/convert_frame.py", line 669, in _compile
    guarded_code = compile_inner(code, one_graph, hooks, transform)
  File "torch/_dynamo/utils.py", line 249, in time_wrapper
    r = func(*args, **kwargs)
  File "torch/_dynamo/convert_frame.py", line 542, in compile_inner
    out_code = transform_code_object(code, transform)
  File "torch/_dynamo/bytecode_transformation.py", line 1033, in transform_code_object
    transformations(instructions, code_options)
  File "torch/_dynamo/convert_frame.py", line 163, in _fn
    return fn(*args, **kwargs)
  File "torch/_dynamo/convert_frame.py", line 507, in transform
    tracer.run()
  File "torch/_dynamo/symbolic_convert.py", line 2130, in run
    super().run()
  File "torch/_dynamo/symbolic_convert.py", line 793, in run
    and self.step()
  File "torch/_dynamo/symbolic_convert.py", line 756, in step
[...]

PR #7067, opened 2024-05-15
[benchmarks] Fix AMP setup for torchbench models.

ysiraichi:
Fix: #6556 (and, possibly #6833)

This PR fixes the benchmarks script when running with AMP. Previously, we were calling `torch.amp.autocast(..., device_type="xla")` for both XLA:CUDA and XLA:TPU. However, we should be using `torch.cuda.amp.autocast` for XLA:CUDA (see [this](https://github.com/pytorch/xla/blob/1fa1f858342df29ce7cdf2d18cabb259869f652f/torch_xla/csrc/tensor_impl.cpp#L74-L83) for more details).

**Context:** after #6518, `Super_Slomo` inference started being run using AMP. However, due to #6511, that PR tried to mimic `torch_xla.amp.autocast` behavior, using `torch.amp.autocast`.

cc @miladm @JackCaoG @vanbasten23 @zpcore

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -224,7 +224,7 @@ class TorchBenchModel(BenchmarkModel):
     gc.collect()
 
     # If we are using CUDA, clean-up its cache left-over.
-    if self.benchmark_experiment.accelerator == "cuda":
+    if self.use_accelerator_cuda():
       torch.cuda.empty_cache()
 
   def set_up(self):
@@ -253,7 +253,7 @@ class TorchBenchModel(BenchmarkModel):
     if self.benchmark_experiment.xla:
       # First, move the model and the inputs to CPU.
       # This avoids having dupplicated data on CUDA.
-      if self.benchmark_experiment.accelerator == "cuda":
+      if self.use_accelerator_cuda():
         self.module = self.module.to("cpu")
         self.example_inputs = move_to_device(self.example_inputs, "cpu")
         self._cleanup()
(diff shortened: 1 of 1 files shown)


ysiraichi:
Confirmed it also fixes #6833.

JackCaoG:
Hmm according to https://github.com/pytorch/xla/blob/master/docs/amp.md we should be able to use `autocast ` for both TPU and GPU, is that no longer the case?

ysiraichi:
That document is correct. Problem is that I didn't notice XLA:CUDA is supposed to run with CUDA autocast, i.e. `torch.amp.autocast("cuda")`. Instead, I was running both XLA:CUDA and XLA:TPU with XLA autocast, i.e. `torch.amp.autocast("xla")`. This behavior is implemented in `torch_xla.amp.autocast`. However, since it currently doesn't work (#6511) with dynamo, I was using `torch.amp.autocast` directly.

JackCaoG (approved):

benchmarks/torchbench_model.py
```diff
@@ -350,23 +357,27 @@ def conversion_dtype(self):
...
-      if self.benchmark_experiment.accelerator == "cuda":
+      # TODO: Should call device specific autocast implementations.
+      # Specifically, we should be using:
+      #   - torch.cuda.amp.autocast for inductor
+      #   - torch_xla.amp.autocast for PyTorch/XLA experiments.
+      # PyTorch/XLA autocast does not run with dynamo, though:
+      # https://github.com/pytorch/xla/issues/6511
+      if self.is_accelerator_cuda():
+        # For inductor and XLA:CUDA, we use CUDA autocast.
+        autocast = torch.cuda.amp.autocast
```
vanbasten23, line 369: I guess `torch.cuda.amp.autocast` is the same as `torch.amp.autocast("cuda")`?

vanbasten23, line 369: do you need to set `kwargs["device_type"] = "xla"` for XLA:GPU case?

ysiraichi: Not really. `torch.cuda.amp.autocast` already does that.

Issue #6751, opened 2024-03-14
[RFC] `torch_xla.step` context manager


will-cromar:
# TL;DR

* In most cases, replace `mark_step` with a context manager/decorator `torch_xla.step` to explicitly mark code that should be traced and then executed.
* Both ensures that `mark_step` is always executed upon exit and provides an explicit context to raise errors.

# Introduction

Users should be able to pick up and use PyTorch/XLA without having to understand the execution model. As much as possible, we should bury the implementation details such that you can wrap existing PyTorch code and have it "just work".

`xm.mark_step` is the most obvious example where we directly expose implementation details of how PyTorch/XLA works to the user: we actually require users to decide where to place synchronization barriers. These manual barriers are not required in JAX and TensorFlow, even though both of them implement a similar lazy execution model.

In practice, our current solution is to hide `xm.mark_step` in our preloading data loader implementation that calls it 1) after each batch is loaded and 2) at the end of iteration. If a user is sufficiently careful, they don't have to see this implementation detail at all. Take the following training loop for example:

```python
for batch in loader:
  # Run model
  xm.optimizer_step() # Note: this may be optimizer.step() when using DDP
```

Training loops like the one above will run without issue. However, even slight deviations from this pattern can cause inscrutable problems such as the one below. 

# Weird `xm.mark_step` behaviors

## Checkpointing before `xm.mark_step`

Take the following common pattern, where the master saves a checkpoint every n steps:

```python
for batch in loader:
  # Run model
  xm.optimizer_step()
  if is_master and step % 100 == 0:
    xm.save(model)
```

Can you spot the error? The above code will hang forever without printing an error at all, because the execution diverges between the master and non-master replicas. Compare the order of operations:

<table>
  <tr>
   <td>Replica 0
   </td>
   <td>Others
   </td>
  </tr>
  <tr>
   <td>

<pre class="prettyprint">load data
mark step
run step
checkpoint, hang forever waiting for other replicas</pre>

   </td>
   <td>

<pre class="prettyprint">load data
mark step
run step
load data
mark step, hang forever waiting for replica 0</pre>

   </td>
  </tr>
</table>

Concretely, a version of this same bug took hours to debug when it appeared in both [Accelerate](https://github.com/huggingface/accelerate/pull/1393) and [Lightning](https://github.com/Lightning-AI/pytorch-lightning/pull/17882). The solution in each case was to add an additional `xm.mark_step` on replica 0 before the checkpoint.

[...]

PR #7068, opened 2024-05-15
Add `xla.step` context manager

will-cromar:
See #6751 

- This implementation is intentionally minimal to start with. The main improvement compared to `sync` is that exceptions are handled sanely.
- Update README example to use `xla.step`. Remove `ParallelLoader` because it mostly does not make a difference for MP, and we should keep our starting point as simple as possible.

### DIFF
--- a/test/test_devices.py
+++ b/test/test_devices.py
@@ -40,6 +40,57 @@ class TestDevices(parameterized.TestCase):
 
     self.assertEqual(met.counter_value('MarkStep'), 1)
 
+  def test_step(self):
+    with torch.step():
+      torch.ones((3, 3), device=xla.device())
+
+    self.assertEqual(met.counter_value('MarkStep'), 1)
+
+  def test_step_exception(self):
+    try:
+      with torch.step():
+        torch.ones((3, 3), device=xla.device())
+        raise RuntimeError("Expected error")
+    except:
+      pass
+
+    self.assertEqual(met.counter_value('MarkStep'), 1)
+
+  # Should roughly match example given in README
+  def test_trivial_model(self):
+    class TrivialModel(nn.Module):
+      def __init__(self):
+        super().__init__()
+        self.linear = nn.Linear(10, 10)
+
+      def forward(self, x):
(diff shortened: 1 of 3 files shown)


test/test_devices.py
```diff
@@ -40,6 +43,56 @@ def test_sync(self):
...
+
+    batch_size = 16
+    num_samples = 100
+
+    input_data = torch.randn(num_samples, 10)
+    target_data = torch.randn(num_samples, 10)
+
+    # Create a DataLoader
+    dataset = TensorDataset(input_data, target_data)
+    loader = DataLoader(dataset, batch_size=batch_size, shuffle=True)
```
JackCaoG, line 82: does DataLoader don't take device as argument?

torch_xla/torch_xla.py
```diff
@@ -50,3 +52,17 @@ def device_count() -> int:
 def sync():
   """Launches all pending graph operations."""
   xm.mark_step()
+
+
+@contextlib.contextmanager
+def step():
```
JackCaoG, line 58: the reason I find `step` can be a bit confusing is that we don't call `mark_step` upon entering the `step`.

```
with xla.step():
  y = x + z

y += 1
```
`step` as a context kind of suggest execution will only cover what happened in side the context manger but that's actually not the case.

test/test_devices.py
will-cromar: No. In normal PyTorch, you have to move the data with `tensor.to`: https://pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html#training-on-gpu

torch_xla/torch_xla.py
will-cromar: I agree. This should either print a warning if there are pending operations, or just mark_step twice. What do you think is better?

JackCaoG: let's try `mark_step` twice and benchmark it with one of the examples on resneto50 with fakedata.

JackCaoG (approved):

PR #7072, opened 2024-05-16
Add VSCode devcontainer instructions to CONTRIBUTING.md

will-cromar:
Tested these instructions because I broke my build environment and had to start from scratch.

Remove extra GPU instructions because this should be handled by dev container.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,7 @@ We recommend you to use our prebuilt Docker image to start your development work
   ```bash
   git clone --recursive --depth=1 https://github.com/pytorch/pytorch.git
   git clone https://github.com/pytorch/xla.git pytorch/xla
+  # Optional: use git@github.com:pytorch/xla.git instead if you prefer to use SSH with key forwarding
   ```
 
 * Link (or copy) VSCode configuration to your workspace directory:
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -42,10 +87,6 @@ We recommend you to use our prebuilt Docker image to start your development work
   python setup.py develop
   ```
 
-### Build PyTorch/XLA from source with GPU support
-
-Please refer to this [guide](https://github.com/pytorch/xla/blob/master/docs/gpu.md#develop-pytorchxla-on-a-gpu-instance-build-pytorchxla-from-source-with-gpu-support).
```
JackCaoG, line 47: this is actually useful, user has to run
```
echo "export PATH=\$PATH:/usr/local/cuda-12.1/bin" >> ~/.bashrc
echo "export LD_LIBRARY_PATH=\$LD_LIBRARY_PATH:/usr/local/cuda-12.1/lib64" >> ~/.bashrc
source ~/.bashrc
```
to build on GPU docker.

will-cromar: I've needed to do that before inside containers in GKE (which uses a different container runtime), but never with Docker with `nvidia-container-toolkit` installed. Are you sure that's still necessary?

JackCaoG: Last time I tried(~1 month ago) I still need it with the GPU docker.

will-cromar: Got it. I'll revert this change for now and give it an update next time I break my GPU build machine

JackCaoG (approved):

will-cromar:
I also added some small fixes to the common VSCode config:

- Add `yapf` formatter extension (no longer included in main Python extension
- Specify `clang-format` as formatter for C++
- Instructions to symlink correct configs
- Make git ignore submodules (avoids adding all of PyTorch's unrelated submodules to the sidebar by default)
- Make Pylance ignore generated code/build artifacts so it doesn't OOM

PR #7075, opened 2024-05-17
Fix usage of extract_jax

zpcore:


benchmarks/experiment_runner.py
```diff
@@ -285,6 +285,7 @@ def run_once_and_gather_metrics(self, benchmark_experiment, benchmark_model,
     # Reset state and sync.
     reset_rng_state(benchmark_experiment)
     if benchmark_experiment.torch_xla2:
+      import jax
```
ysiraichi, line 288: Why do we need to import `jax`, here?

zpcore: Thanks for catching that, relocated to the right place.

qihqi (approved):

Issue #7084, opened 2024-05-20
[torchbench] `timm_nfnet` training failing on non-dynamo.


ysiraichi:
After #7067, `timm_nfnet` started failing with the following error:

```bash
python xla/benchmarks/experiment_runner.py \
    --suite-name torchbench --accelerator cuda --repeat 8 --iterations-per-run 1 \
    --xla PJRT --dynamo None --test train \
    --filter timm_nfnet
```

```python
Traceback (most recent call last):
  File "xla/benchmarks/experiment_runner.py", line 960, in <module>
    main()
  File "xla/benchmarks/experiment_runner.py", line 956, in main
    runner.run()
  File "xla/benchmarks/experiment_runner.py", line 61, in run
    self.run_single_config()
  File "xla/benchmarks/experiment_runner.py", line 256, in run_single_config
    metrics, last_output = self.run_once_and_gather_metrics(
  File "xla/benchmarks/experiment_runner.py", line 349, in run_once_and_gather_metrics
    output, _ = loop(iter_fn=self._default_iter_fn)
  File "xla/benchmarks/experiment_runner.py", line 306, in loop
    output, timing, trace = iter_fn(benchmark_experiment, benchmark_model,
  File "xla/benchmarks/experiment_runner.py", line 224, in _default_iter_fn
    self._mark_step(benchmark_experiment, output)
  File "xla/benchmarks/experiment_runner.py", line 428, in _mark_step
    xm.mark_step()
  File "xla/torch_xla/core/xla_model.py", line 1055, in mark_step
    torch_xla._XLAC._xla_step_marker(
RuntimeError: Bad StatusOr access: INTERNAL: during context [Unknown]: Seen floating point types of different precisions in %multiply.3753 = f32[128,3072,6,6]{3,2,1,0} multiply(f16[128,3072,6,6]{3,2,1,0} %multiply.3730, f32[128,3072,6,6]{3,2,1,0} %add.3752), but mixed precision is
disallowed.
```

## Environment

 - Reproducible on XLA backend [CPU/TPU]: CUDA
 - torch_xla version: 62c3ba652ea09e2076a27f200ad755541f37daeb

cc @miladm @JackCaoG @vanbasten23 @zpcore

ysiraichi:
Apparently, fixing `gelu_backward` wasn't enough for fixing `timm_nfnet`.

PR #7090, opened 2024-05-21
Add data-type promotion to `gelu_backward`.

ysiraichi:
Fix: #7084 

This PR adds data-type promotion to `gelu_backward` operation. Previously, there was none. So, the kernel implicitly expected the arguments to be of the same data-type. This might not be the case when using AMP.

cc @miladm @JackCaoG

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1982,6 +1982,22 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
     for dtype in test_dtypes:
       test(dtype)
 
+  def test_gelu_backward_different_types(self):
+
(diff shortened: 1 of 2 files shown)


JackCaoG (approved):

vanbasten23:
Curious, how did you find out it was gelu_backward based on the error message in https://github.com/pytorch/xla/issues/7084#issue-2306723082 (which I don't see any hint of gelu_backward)?

ysiraichi:
Since it was a non-dynamo bug, it was thanks to `XLA_USE_EAGER_DEBUG_MODE=1` that I found the bug.

Issue #7083, opened 2024-05-20
[torchbench] `timm_efficientdet` training failing on non-dynamo.


ysiraichi:
After #7067, `timm_efficientdet` started failing with the following error:

```bash
python xla/benchmarks/experiment_runner.py \
    --suite-name torchbench --accelerator cuda --repeat 8 --iterations-per-run 1 \
    --xla PJRT --dynamo None --test train \
    --filter timm_efficientdet
```

```python
Traceback (most recent call last):
  File "xla/benchmarks/experiment_runner.py", line 945, in <module>
    main()
  File "xla/benchmarks/experiment_runner.py", line 941, in main
    runner.run()
  File "xla/benchmarks/experiment_runner.py", line 61, in run
    self.run_single_config()
  File "xla/benchmarks/experiment_runner.py", line 256, in run_single_config
    metrics, last_output = self.run_once_and_gather_metrics(
  File "xla/benchmarks/experiment_runner.py", line 345, in run_once_and_gather_metrics
    output, _ = loop(iter_fn=self._default_iter_fn)
  File "xla/benchmarks/experiment_runner.py", line 302, in loop
    output, timing, trace = iter_fn(benchmark_experiment, benchmark_model,
  File "xla/benchmarks/experiment_runner.py", line 218, in _default_iter_fn
    output = benchmark_model.model_iter_fn(
  File "xla/benchmarks/torchbench_model.py", line 411, in train
    super().train(inputs, collect_full_output=collect_full_output)
  File "xla/benchmarks/benchmark_model.py", line 160, in train
    loss.backward()
  File "torch/_tensor.py", line 523, in backward
    torch.autograd.backward(
  File "torch/autograd/__init__.py", line 267, in backward
    _engine_run_backward(
  File "torch/autograd/graph.py", line 767, in _engine_run_backward
    return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Bad StatusOr access: INTERNAL: during context [Unknown]: Seen floating point types of different precisions in %concatenate.7662 = f32[1,88,10,10,2]{4,3,2,1,0} concatenate(f16[1,88,10,10,1]{4,3,2,1,0} %reshape.7660, f32[1,88,10,10,1]{4,3,2,1,0} %reshape.7661), dimensions={4}, but mixed precision is disallowed.
```

## Environment

 - Reproducible on XLA backend [CPU/TPU]: CUDA
[...]

PR #7091, opened 2024-05-22
Add data-type promotion to `stack`.

ysiraichi:
Fix: #7083 

This PR adds data-type promotion to `stack` operation. Previously, there was none. So, the kernel implicitly expected the arguments to be of the same data-type. This might not be the case when using AMP.

cc @miladm @JackCaoG

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -3160,8 +3160,12 @@ at::Tensor XLANativeFunctions::squeeze_copy(const at::Tensor& self,
 
 at::Tensor XLANativeFunctions::stack(at::TensorList tensors, int64_t dim) {
   TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  at::ScalarType result_type = at::native::result_type(tensors);
+  std::vector<at::Tensor> c_tensors(tensors.size());
+  std::transform(tensors.begin(), tensors.end(), c_tensors.begin(),
+                 [=](const at::Tensor& t) { return t.to(result_type); });
   return bridge::AtenFromXlaTensor(
-      tensor_methods::stack(bridge::GetXlaTensors(tensors), dim));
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -3158,8 +3158,12 @@ at::Tensor XLANativeFunctions::squeeze_copy(const at::Tensor& self,
 
 at::Tensor XLANativeFunctions::stack(at::TensorList tensors, int64_t dim) {
   TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  at::ScalarType result_type = at::native::result_type(tensors);
+  std::vector<at::Tensor> c_tensors(tensors.size());
```
JackCaoG, line 3162: is stack expecting input tensor to be CPU? `std::vector<at::Tensor> c_tensors` will return a list of tenosrs on CPU right?

ysiraichi: No. They are casted tensors, on XLA.

JackCaoG: Then I am abit confused. Reading your code, you init the `c_tensors` vector which I assume they will be cpu tensors since you didn;t provide the device type. In the later code you only update the dtype of these `c_tensors`, I don't know when are they moved to the XLA device.

ysiraichi: Here's a summary of what this code is doing: considering the arguments `tensors` (a list of XLA tensors) and `dim`, the function:

1. Computes the common data-type of all tensors: `result_type`
2. Converts each tensor to the common data-type, storing the result in `c_tensors` (as in "cast tensors")
3. Calls `tensor_methods::stack` with the casted tensors

JackCaoG: Oh I see. `transform` is called with `tensors.begin()`..

JackCaoG (approved):

PR #7093, opened 2024-05-22
Add torch_xla2 `export_program_to_stablehlo` API with unbounded dynamism support

GleasonK:
This PR adds a new StableHLO export API for torch_xla2:

```python
# Module: torch_xla2.export
def export_program_to_stablehlo(exported : torch.ExportedProgram) -> jax.Exported: ...
```

This **is not** a perfect drop-in replacement for `torch_xla.stablehlo.export_program_to_stablehlo` but largely follows the same contract of "IN: torch exported, OUT: StableHLO thing".

The complexity of this PR comes down to `extract_avals`, a function which takes an exported program, and tries to "abstractify" user input values, for example: 

```
# Trivial
torch.tensor(10).to(torch.float64) --> jax.ShapeDtypeStruct([10], dtype=jnp.float64)

# More complicated
FakeTensorValue(..., [s0, 5]) --> jax.ShapeDtypeStruct([jax.symbolic_shape('s0', constraints=...), 5], dtype=jnp.float64)
```

Some cool features of this API:
- High dynamic shape coverage, likely more than `torch_xla.stablehlo.export_program_to_stablehlo`. While HLO has full dynamism parity, I believe there are still more bridge modifications needed to propagate shape throughout. I was able to export ViT and wav2vec models with dynamic batch dims (these are the only two models I tried).
- Dynamo guards are embedded in the exported IR. This is done via JAX's `custom_call @shape_assert` which has support in TF2XLA execution paths (like savedmodel / tf.serving). These custom_call's will either canonicalize away as we refine shapes, or error if we attempt to refine/compile with incompatible shapes.

Tests:
- Added some testpoints to `test_exported.py`
- Forked `test/stablehlo/test_unbounded_dynamism.py` and swapped `export_program_to_stablehlo` impls. We can easily merge the two, but I'm not sure if we're trying to keep torch_xla2 fully independent, in which case maybe forking is OK. All tests passed. A few needed slight regex modifications, which are noted at the top of the new test file (seems more correct now?).
- Tested new dtype mappings (I had some typos which took a bit to debug, would have been easier with some tests).

Possible next steps:
- Make this a drop-in replacement for `torch_xla.stablehlo.export_program_to_stablehlo`
- Add a SavedModel export API which builds on top of this (or make a jax2tf wrapper, unclear what's best, I kind of like how PT/XLA does this today).

Also, a few QoL changes as I got this part of the repo set up for the first time:
- Edited `experimental/torch_xla2/README.md` to include repo clone step for clarity
- Changed base JAX dependency to `"jax[cpu]>=0.4.24",` to avoid downloading CUDA and friends by default
- Added some more `torch.dtype -> jax.dtype` mappings to match with JAX's [`mlir.py` mappings](https://github.com/google/jax/blob/9c01fc5f0f2d1335810772c79ba0a1c329d3d908/jax/_src/interpreters/mlir.py#L154).

### DIFF
--- a/experimental/torch_xla2/test/test_exports.py
+++ b/experimental/torch_xla2/test/test_exports.py
@@ -93,42 +93,8 @@ class ExportTest(unittest.TestCase):
 
     # Look for dynamic shape artifacts
     self.assertIn("func.func public @main(%arg0: tensor<?x3x200x200xf32>", module_str)
-    self.assertRegex(module_str, r"shape_assertion.*s0 >= 3")
-    self.assertRegex(module_str, r"shape_assertion.*s0 <= 10")
     self.assertIn("stablehlo.dynamic_broadcast_in_dim", module_str)
     self.assertIn("stablehlo.dynamic_gather", module_str)
-
-  def test_complex_constraint(self):
-    """Test a model with a slightly more complex constraint, where the input
-    shapes are determined by an equation of the other, in this case input shapes
-    are s0{<=10} and s0*2.
-    """
-    class ConcatAddModel(torch.nn.Module):
-      def __init__(self):
-        super().__init__()
-
-      def forward(self, a, b):
-        a = torch.concat([a, a], dim=0)
-        return a + b
-
-    # Arg shapes are a=s0{<=10}, b=s0*2
-    model = ConcatAddModel()
-    args = (torch.rand(2),torch.rand(4))
-    sym_a = torch.export.Dim("a", max=10)
-    sym_b = sym_a*2
-    dynamic_shapes = ({0: sym_a}, {0: sym_b})
-
-    with torch.no_grad():
-      exported = torch.export.export(model, args=args, dynamic_shapes=dynamic_shapes)
-    stablehlo = torch_xla2.export.exported_program_to_stablehlo(exported)
-    module_str = str(stablehlo.mlir_module())
-
-    self.assertIn("stablehlo.constant dense<2>", module_str)
-    self.assertRegex(module_str, r"shape_assertion.*s0 <= 10")
-    self.assertRegex(module_str, r"shape_assertion.*2*s0")
-    self.assertRegex(module_str, r'stablehlo.concatenate.*tensor<\?xf32>')
-    self.assertRegex(module_str, r'stablehlo.add.*tensor<\?xf32>')
-    print(stablehlo.mlir_module())
 
   def test_export_dtypes(self):
     DTYPE_TO_MLIR_STR = {
--- a/experimental/torch_xla2/test/test_exports.py
+++ b/experimental/torch_xla2/test/test_exports.py
@@ -54,9 +54,6 @@ class ExportTest(unittest.TestCase):
     self.assertIn("func.func private @clip(%arg0: tensor<500xf32>", module_str)
     self.assertIn("stablehlo.minimum", module_str)
 
-    # Test with dynamic export
-
-
   def test_constant(self):
 
     # Check Accuracy
--- a/experimental/torch_xla2/test/test_exports.py
+++ b/experimental/torch_xla2/test/test_exports.py
@@ -135,7 +135,6 @@ class ExportTest(unittest.TestCase):
         ##   jnp.tensor(dtype=None) maps to f64
         continue
       arg = (torch.randn(10).to(torch_dtype),)
-      print(arg)
       with torch.no_grad():
         exported = torch.export.export(model, arg)
       stablehlo = torch_xla2.export.exported_program_to_stablehlo(exported)
--- a/experimental/torch_xla2/test/test_exports.py
+++ b/experimental/torch_xla2/test/test_exports.py
@@ -93,10 +93,42 @@ class ExportTest(unittest.TestCase):
 
     # Look for dynamic shape artifacts
     self.assertIn("func.func public @main(%arg0: tensor<?x3x200x200xf32>", module_str)
-    self.assertRegex(module_str, "shape_assertion.*s0 >= 3")
-    self.assertRegex(module_str, "shape_assertion.*s0 <= 10")
+    self.assertRegex(module_str, r"shape_assertion.*s0 >= 3")
+    self.assertRegex(module_str, r"shape_assertion.*s0 <= 10")
     self.assertIn("stablehlo.dynamic_broadcast_in_dim", module_str)
     self.assertIn("stablehlo.dynamic_gather", module_str)
+
+  def test_complex_constraint(self):
+    """Test a model with a slightly more complex constraint, where the input
+    shapes are determined by an equation of the other, in this case input shapes
+    are s0{<=10} and s0*2.
+    """
+    class ConcatAddModel(torch.nn.Module):
+      def __init__(self):
+        super().__init__()
+
+      def forward(self, a, b):
+        a = torch.concat([a, a], dim=0)
+        return a + b
+
+    # Arg shapes are a=s0{<=10}, b=s0*2
+    model = ConcatAddModel()
+    args = (torch.rand(2),torch.rand(4))
+    sym_a = torch.export.Dim("a", max=10)
+    sym_b = sym_a*2
+    dynamic_shapes = ({0: sym_a}, {0: sym_b})
+
+    with torch.no_grad():
+      exported = torch.export.export(model, args=args, dynamic_shapes=dynamic_shapes)
+    stablehlo = torch_xla2.export.exported_program_to_stablehlo(exported)
+    module_str = str(stablehlo.mlir_module())
+
+    self.assertIn("stablehlo.constant dense<2>", module_str)
+    self.assertRegex(module_str, r"shape_assertion.*s0 <= 10")
+    self.assertRegex(module_str, r"shape_assertion.*2*s0")
+    self.assertRegex(module_str, r'stablehlo.concatenate.*tensor<\?xf32>')
+    self.assertRegex(module_str, r'stablehlo.add.*tensor<\?xf32>')
+    print(stablehlo.mlir_module())
 
   def test_export_dtypes(self):
     DTYPE_TO_MLIR_STR = {
--- a/experimental/torch_xla2/test/test_exports.py
+++ b/experimental/torch_xla2/test/test_exports.py
@@ -34,33 +34,115 @@ class ExportTest(unittest.TestCase):
 
   def test_interpolate(self):
 
+    # Check Accuracy
     arg = (torch.randn(3, 3, 200, 200),)
     model = Interpolate()
-
     ans = model(*arg)
 
     with torch.no_grad():
       exported = torch.export.export(model, arg)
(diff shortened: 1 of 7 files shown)


qihqi (approved):
Thanks for this very high quality PR, LGTM!

experimental/torch_xla2/torch_xla2/export.py
```diff
@@ -104,3 +107,130 @@ def func(states, inputs):
...
+
+    def _build_symbolic_constraints(symbol_name, torch_constraint):
+      """Convert torch SymInt constraints to string for JAX symbolic_shape
+      Using sympy may be overkill here, currently PyTorch only uses ValueRanges
+      which allow specifying the min and the max of a value, for example:
+        torch.export.Dim("a", min=5, max=10)
+         ==> ("a >= 5", "a <= 10",)
+      """
+      if not isinstance(torch_constraint, torch.utils._sympy.value_ranges.ValueRanges) or torch_constraint.is_bool:
+        raise TypeError(f"No symbolic constraint handler for: {constraint}")
```
mlevesquedion, line 165: I believe this will `NameError` before the `TypeError` can be raised? (`constraint` is not defined)

experimental/torch_xla2/torch_xla2/export.py
```diff
@@ -104,3 +107,130 @@ def func(states, inputs):
...
+
+      constraints = []
+      symbol = sympy.Symbol(symbol_name)
+      if (torch_constraint.lower != 2):
```
mlevesquedion, line 169: Nit: parens around condition are not needed, here and below.

experimental/torch_xla2/torch_xla2/export.py
```diff
@@ -104,3 +107,130 @@ def func(states, inputs):
...
+    # have its own scope.
+    symbolic_shapes = {}
+    symbol_variables = [(s,v) for s,v in range_constraints.items() if s.is_symbol]
+    symbol_exprs = [(s,v) for s,v in range_constraints.items() if not s.is_symbol]
+    for sym, constraint in symbol_variables:
+      symbolic_shape = _build_symbolic_shape(sym, constraint, symbolic_shapes)
+      symbolic_shapes[str(sym)] = symbolic_shape
+    for sym, constraint in symbol_exprs:
+      symbolic_shape = _build_symbolic_shape(sym, constraint, symbolic_shapes)
+      symbolic_shapes[str(sym)] = symbolic_shape
```
mlevesquedion, line 210: Can the for loops be merged? The bodies seem to be the same.

experimental/torch_xla2/torch_xla2/export.py
```diff
@@ -3,10 +3,13 @@
 import copy
 from typing import Any, Dict, Tuple
 import torch
-from torch_xla2.ops import ops_registry
-from torch_xla2 import tensor
 from torch.utils import _pytree as pytree
-
+from torch_xla2 import tensor
+from torch_xla2.ops import ops_registry
+from jax.experimental import export
```
mlevesquedion, line 9: IMO it would be preferable to import the module and refer to `export` by its fully qualified name. Same comment in other places, there are many `export` functions and `export` is a very generic name so having the full name helps with readability.

experimental/torch_xla2/test/test_unbounded_dynamism.py
```diff
@@ -0,0 +1,662 @@
...
+    def __init__(self):
+      super().__init__()
+      
+    def forward(self, *args):
+      return f(*args)
+  return M().eval()
+
+class UnboundedDynamismExportTest(unittest.TestCase):
+
+  def test_add(self):
```
mlevesquedion, line 47: Most of these tests look very similar. Consider extracting the duplicated bits to a function.

experimental/torch_xla2/test/test_unbounded_dynamism.py
```diff
@@ -0,0 +1,662 @@
...
+    args = (torch.rand((10, 197, 768)), torch.rand((10, 197, 768)))
+    dynamic_shapes = (({0: Dim("dim")}, {0: Dim("dim")}),)
+    m = wrap_func_as_nn_module(torch.ops.aten.add.Tensor)
+    ep = export(m, args=args, dynamic_shapes=dynamic_shapes)
+    shlo_module = exported_program_to_stablehlo(ep)
+    shlo_text = shlo_module.get_stablehlo_text()
+    self.assertTrue(
+        re.search(
+            r'tensor<\?x197x768xf32>.*tensor<\?x197x768xf32>.*->.*tensor<\?x197x768xf32>',
+            shlo_text) is not None)
```
mlevesquedion, line 57: Would `assertRegex` work here?

mlevesquedion (commented):
Would it help to have a few higher level tests (in the style of lit tests)? Would be nice to be able to see the mapping at a high level and confirm the overall logic. (Such tests may be brittle though because they are exposed to a very large API surface and some external implementation details -- with that in mind I still think it would make sense to have a few integration tests.)

experimental/torch_xla2/test/test_exports.py
```diff
@@ -34,33 +34,112 @@ def setUp(self):
...
+    self.assertTrue(torch.allclose(ans, ans2, atol=1e-3))
+
+    # Convert to StableHLO
+    stablehlo = torch_xla2.export.exported_program_to_stablehlo(exported)
+    module_str = str(stablehlo.mlir_module())
+    self.assertIn("func.func public @main", module_str)
+    self.assertIn("func.func private @clip(%arg0: tensor<500xf32>", module_str)
+    self.assertIn("stablehlo.minimum", module_str)
+
+    # Test with dynamic export
```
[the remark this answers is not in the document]
GleasonK: made an entire separate test point and forgot to delete comment

experimental/torch_xla2/test/test_exports.py
```diff
@@ -34,33 +34,112 @@ def setUp(self):
...
+      torch.complex128    : "complex<f64>",
+      None                : None,
+    }
+
+    model = TensorConstant()
+    for torch_dtype in torch_xla2.tensor.TORCH_DTYPE_TO_JAX.keys():
+      if torch_dtype == None:
+        ## TODO: Figure out what the None mapping should be, seems like:
+        ##   torch.tensor(dtype=None) maps to f32
+        ##   jnp.tensor(dtype=None) maps to f64
```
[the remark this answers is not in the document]
GleasonK: I didn't write the `None:None` mapping, so going to leave this open, in case there's some other implications to mapping.

Also the way the mapping is written now `None->jax.f32` would imply that `jax.f32->None` in the other direction. Technically not wrong but I'd rather directly map to the torch.f32 type as we currently are

experimental/torch_xla2/torch_xla2/export.py
GleasonK: how you figured this out visually is beyond me. good catch :)

experimental/torch_xla2/torch_xla2/export.py
GleasonK: Yeah just need to enforce that symbol_variable comes first:

```
    symbol_variables = [(s,v) for s,v in range_constraints.items() if s.is_symbol]
    symbol_exprs = [(s,v) for s,v in range_constraints.items() if not s.is_symbol]
    for sym, constraint in symbol_variables + symbol_exprs:
```

Alt would be using sort with comparator, but I'm not sure if there's a good way to sort on booleans (+not sure I like implying an ordering of bools..), so perhaps this is the best

experimental/torch_xla2/test/test_unbounded_dynamism.py
GleasonK: This file is forked from a `test_unbounded_dynamism.py` file that exists in the main torch_xla tests (keeping separate since there's little overlap in testing currently). Want it to be as comparable as possible, so using this formatting. But yes, I use assertRegex elsewhere in the new tests

experimental/torch_xla2/torch_xla2/export.py
GleasonK: done

experimental/torch_xla2/test/test_exports.py
```diff
@@ -34,33 +34,112 @@ def setUp(self):
...
+    module_str = str(stablehlo.mlir_module())
+
+    # Look for dynamic shape artifacts
+    self.assertIn("func.func public @main(%arg0: tensor<?x3x200x200xf32>", module_str)
+    self.assertIn("stablehlo.dynamic_broadcast_in_dim", module_str)
+    self.assertIn("stablehlo.dynamic_gather", module_str)
+
+  def test_export_dtypes(self):
+    DTYPE_TO_MLIR_STR = {
+      # NO_MAPPING        : jnp.float0 (signless scalar int)
```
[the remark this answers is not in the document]
GleasonK: Keeps the two tables aligned, which I think has value given the dependency between the two. I.e. the lack of mappings was discovered by doing this and then looking at torch support

experimental/torch_xla2/torch_xla2/export.py
GleasonK: let me be a cpp dev 😅

GleasonK: fixed

experimental/torch_xla2/test/test_unbounded_dynamism.py
GleasonK: this was the forked test, going to keep it as-is to be aligned with original (easy diffing).

experimental/torch_xla2/test/test_exports.py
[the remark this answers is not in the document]
qihqi: So in torch, when `dtype` is None it default to `torch.get_default_dtype()` and we have implemented that for tensor constructors here: (https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/torch_xla2/ops/op_base.py#L77)

The above is configurable with `torch.set_default_dtype` and defaults to f32.

Jax has a different configuration: jax_enable_x64 which we enable here: https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/torch_xla2/__init__.py#L7

The Jax configuration's semantics is: by default it's False and everything is at most 32 bits. (meaning, even if doing `x.astype(jnp.float64)`  (same for int64) won't give you float64 but just give you an warning saying it's not enabled. When enabled, you are able to get 64 bit, but also the constructors defaults to 64 bits. 

So the current strategy is to 1. enable jax x64 so at least we can get it if requested. and 2. explicitly read torch.get_default_dtype and pass it on.

experimental/torch_xla2/torch_xla2/export.py
mlevesquedion: Ah, I see. I think sorting here would be a good fit actually because we want to enforce an ordering and the ordering for booleans is `False < True`. I like what you did with `symbol_variables + symbol_exprs` though.

mlevesquedion (approved):

PR #7094, opened 2024-05-22
[Fori_loop|While_loop] Enable while_loop/fori_loop, Add linear/MNIST test case

ManfeiBai:
- Modify BuildForiLoop for while_loop/fori_loop
- Modify test code for while_loop/fori_loop with simple version
- Add linear_layer_model and MNIST_model(without BN layer) test case


test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
+        y = self.fc2(y)
+        return iteri - 1, x.clone(), F.log_softmax(y, dim=1)
+
+    mnist = MNIST()
+    mnist.to(device)
+    bs=16
+    l_in_0 = torch.randn(bs, 1, 28, 28, dtype=torch.float32, device=device)
+    l_out = torch.randn(bs, 10, dtype=torch.float32, device=device)
+    iteri = torch.tensor(3, dtype=torch.int64, device=device)
+    _, _, res = mnist(iteri, l_in_0, l_out)
```
JackCaoG, line 243: can we init the `l_out ` inside the mnist forward and returns it? This way the only additional output we provide is the  count.

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
+      return iteri - 1, x, torch.add(x, 1)
 
-    init = torch.tensor([10], dtype=torch.int32, device=device)
-    limit_value = torch.tensor([0], dtype=torch.int32, device=device)
-    res = while_loop(cond_fn, body_fn, (init, limit_value))
-    expected = _fake_while_loop(cond_fn, body_fn, (init, limit_value))
-    self.assertEqual(expected, res)
+    init_val = torch.tensor(3)
+    out_val = torch.tensor(15)
+    iteri = torch.tensor(10)
```
JackCaoG, line 76: I am not sure why do we need to test the cpu torch tensor

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
+
+  def test_while_loop_tpu_addition_nested(self):
+    xm.mark_step()
+    device = xm.xla_device()
+
+    def cond_fn(iteri, x):
+      return iteri > 0
+
+    def body_fn(iteri, x):
+      return iteri - 1, torch.add(torch.add(x, 1), 1)
```
JackCaoG, line 121: we don't need separate test for add and subtract, you can just keep one.

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
-    def cond_fn(init, limit_value):
-      return limit_value[0] >= init[0]
+    n_epochs = 3
+    batch_size_train = 8
+    batch_size_test = 10
+    learning_rate = 0.01
+    momentum = 0.5
+    log_interval = 10
+    random_seed = 1
+    torch.backends.cudnn.enabled = False
```
JackCaoG, line 141: why is this needed?

JackCaoG, line 141: None of these seems to be used

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
+    torch.manual_seed(random_seed)
 
-    def body_fn(init, limit_value):
-      one_value = torch.ones(1, dtype=torch.int32, device=device)
-      return (torch.add(init, one_value), limit_value.clone())
+    ### load data
+    test_loader = xu.SampleGenerator(
+    data=(torch.zeros(8, 1, 28,28), torch.zeros(8, dtype=torch.int64)),
+    sample_count=1000 // 8 // xm.xrt_world_size())
```
JackCaoG, line 147: you don;'t need the loader, you only need to run one step so just give it one input

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
+    _, res = linear_model(iteri, l_in_0)
+
+    # === expected result after 2 iteration to be compared ===
+    _, expected = linear_model.forward_compare(iteri, l_in_0)
+    _, expected = linear_model.forward_compare(iteri, expected)
+
+    self.assertTrue(torch.all(torch.eq(res, expected)))
+
+  def test_while_loop_tpu_MNIST_inside_loop(self):
+    xm.mark_step()
```
JackCaoG, line 181: ditto, you shouldn't need the `mark_step` here

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
-      return limit_value[0] <= init[0]
+    n_epochs = 3
+    batch_size_train = 8
+    batch_size_test = 10
+    learning_rate = 0.01
+    momentum = 0.5
+    log_interval = 10
+    random_seed = 1
+    torch.backends.cudnn.enabled = False
+    torch.manual_seed(random_seed)
```
JackCaoG, line 193: ditto, remove

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
 
-    def body_fn(init, limit_value):
-      one_value = torch.ones(1, dtype=torch.int32, device=device)
-      two_value = limit_value.clone()
-      return (torch.sub(torch.sub(init, one_value), one_value), two_value)
+    ### load data
+    test_loader = xu.SampleGenerator(
```
JackCaoG, line 196: remove loader

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
+        def body_fn(iteri, x, y):
+          y = F.relu(F.max_pool2d(self.conv1(x), 2))
+          # z = self.bn1(z) # torch.while_loop's body_fn might be modifying the input!
+          y = F.relu(F.max_pool2d(self.conv2(y), 2))
+          # y = self.bn2(y)
+          y = torch.flatten(y, 1)
+          y = F.relu(self.fc1(y))
+          y = self.fc2(y)
+
+          return iteri - 1, x.clone(), F.log_softmax(y, dim=1)
```
JackCaoG, line 223: hmm, if you always return the `x`, wouldn't the result always be the same no matter how many steps you run?

JackCaoG (changes requested):
only review the test part, left some comments

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -10,7 +10,17 @@
 from torch._higher_order_ops.while_loop import while_loop
```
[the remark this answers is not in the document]
ManfeiBai: thanks, updated

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -10,7 +10,17 @@
...
 import torch_xla.core.xla_model as xm
 import torch_xla.core.xla_builder as xb
+import torch_xla.utils.utils as xu
 
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+
+def _fake_while_loop_second(cond_fn, body_fn, operands):
```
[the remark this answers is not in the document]
ManfeiBai: thanks for reminder, delete the duplicated one

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
+    def cond_fn(iteri, x, y):
+      return iteri > 0
 
+    def body_fn(iteri, x, y):
+      return iteri - 1, x, torch.sub(x, 1)
 
-class WhileLoopTest(unittest.TestCase):
+    init_val = torch.tensor(10)
+    out_val = torch.tensor(15)
+    iteri = torch.tensor(3)
```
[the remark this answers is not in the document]
ManfeiBai: I wrote compare test case with pure torch to compare result as expected, would delete these pure torch test cases for final version

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
+    res = while_loop(cond_fn, body_fn, (iteri, init_val, out_val))
+    expected = _fake_while_loop_second(cond_fn, body_fn, (iteri, init_val, out_val))
+    self.assertEqual(res, expected)
 
   def test_while_loop_tpu_subtraction(self):
-
+    xm.mark_step()
```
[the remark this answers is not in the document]
ManfeiBai: Thanks, updated

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
```diff
@@ -19,88 +29,311 @@ def _fake_while_loop(cond_fn, body_fn, operands):
...
     device = xm.xla_device()
 
-    def cond_fn(init, limit_value):
-      return limit_value[0] <= init[0]
+    def cond_fn(iteri, x):
+      return iteri > 0
+
+    def body_fn(iteri, x):
+      return iteri - 1, torch.sub(x, 1)
```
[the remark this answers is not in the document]
ManfeiBai: first test use torch, and second test use torch_xla, would delete pure torch test case in the final version

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: thanks, would delete pure torch test case in the final version

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: thanks, updated

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: thanks, they are unused, would delete and update

ManfeiBai: thanks, yes, would delete and update

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: thanks, deleted and updated

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: thanks, deleted and updated

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: thanks, deleted and updated

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: thanks, deleted and updated

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: yes, we have two different test situation now:

- `repeat work`: this result is expected to be the same in each step run for `test_while_loop_tpu_MNIST_inside_loop` test case, because input and output are different shape;

- `duplicate layer`: we have test case `test_while_loop_tpu_simple_linear_inside_loop`, in which input and output has the same shape, so current output would be the next step's input, and this test case simulate situation that duplicated layers

test/test_fori_loop_with_while_loop_simple_add_dispatch_in_torch.py
ManfeiBai: thanks, good idea, due to hard to gain result before finish one run, so current plan would be: 
run body_fn with `pure python while`, get result, then use this result as `l_out`

before implementation, did quick benchmark for this idea with [code](https://gist.github.com/ManfeiBai/482deb202488009a2dea3326b7e4dea6):
```
- iteri = 3: 
  while_loop(3.675 s), pure while(3.107 s), pure_while_then_while_loop_plan(3.724 s)
- iteri = 10: 
  while_loop(3.670s), pure while(3.119s), pure_while_then_while_loop_plan(3.932s)
- iteri = 10: 
  while_loop(3.744 s), pure while(3.110 s), pure_while_then_while_loop_plan(3.749 s)
- iteri = 10: 
  while_loop(3.878 s), pure while(7.053 s), pure_while_then_while_loop_plan(3.809 s)
```
so this plan looks like worth to try

JackCaoG:
@ManfeiBai give me a ping when CI is green

ManfeiBai:
fork PR failed, track in https://github.com/pytorch/xla/pull/7157

ManfeiBai:
thanks for review and comments, since CI failed to start, track test result in https://github.com/pytorch/xla/pull/7157

PR #6940, opened 2024-04-18
Adding megablox gmm standalone

miladm:
In this PR, we add megablox kernel. The current implementation adds a new file `megablox_gmm`. I plan to merge it to `custom_kernel` with the rest of kernels.

PR #7099, opened 2024-05-23
[Pallas] Refactor the gmm kernel

alanwaketan:
Summary:
This is an effort to refactor the code from https://github.com/pytorch/xla/pull/6940 and aims to remove useless code in that part. It reduces the amount of code from ~400 lines to ~50 lines. However, a bummer is the original gmm kernel doesn't work at all... It assumes groups_sizes is a cpu tensor. That means we need to materialize this input in order to use this gmm kernel, and that will introduce graph breaks in the computation. I will need yet another follow up to make this code actually functional... Good news is the test cases seem functional, yay...

Test Plan:
python test/test_megablox.py


JackCaoG (approved):

alanwaketan:
No need to try tpu ci as I mentioned .... This thing is not functional on tpu....

wonjoolee95 (commented):
LGTM, thanks!

PR #7107, opened 2024-05-24
[Pallas] Introduce _make_group_metadata

alanwaketan:
Summary:
_make_group_metadata is a helper function to assist gmm. Before we use the JAX version that cannot be stitched to our HLO. Now with this new torch version, it allows us to contain it in our HLO. However, we still need to lower two ops: https://pytorch.org/docs/stable/generated/torch.repeat_interleave.html and https://pytorch.org/docs/stable/generated/torch.histc.html. Yet we need the JAX version of the repeat to make the op shape static: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.repeat.html.

Test Plan:
python test/test_gmm.py


wonjoolee95 (approved):
LGTM! 

I didn't really go into the details of `_make_group_metadata`, but I guess `test_make_group_metadata` assures that our version works same as JAX's version so LGTM, thanks!

alanwaketan:
Thanks, Wonjoo!

PR #7108, opened 2024-05-24
[OpenXLA-pin-update trigger] Create _openxla_pin_update_weekly.yml

ManfeiBai:
Add Github Action to create workflow enable create OpenXLA-pin update PR weekly Monday 9:00AM automatically


JackCaoG:
@will-cromar can you take a look?

.github/workflows/_openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,52 @@
+name: Weekly Code Update PR
```
will-cromar, line 1: nit: remove the leading `_` underscore in the file name because this isn't a template

.github/workflows/_openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  schedule:
```
will-cromar, line 3: Can you also add `workflow_dispatch` so this can be run manually if needed? https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

.github/workflows/_openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Get current date
+        run: echo "CURRENT_DATE=$(date +'%Y%m%d')" >> $GITHUB_ENV
+
+      - name: Install sed
+        run: sudo apt-get update && sudo apt-get install -y sed
+
+      - name: Update date in setup.py
+        run: |
+          sed -i "s/_date =.*/_date = '${{ env.CURRENT_DATE }}'/" ${FILE_PATH}
```
will-cromar, line 21: Do you have a way of handling the JAX version too? https://github.com/pytorch/xla/blob/22e912ea15c1b98495eacd14aabff8832493ae3e/setup.py#L70

.github/workflows/_openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,52 @@
...
+          sed -i "s/xla_hash =.*/xla_hash = '${{ env.GITHUB_ENV }}'/" ${FILE_PATH}
+          git add ${FILE_PATH}
+          git commit -m "Weekly update WORKSPACE with (${{ env.GITHUB_ENV }})"
+          git push origin HEAD:weekly-update
+        env:
+          FILE_PATH: WORKSPACE
+
+      - name: Create Pull Request
+        uses: actions/create-pull-request@v4
+        with:
```
will-cromar, line 45: Is there a way to tag this with the `tpuci` label?

.github/workflows/_openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          git remote add target https://github.com/openxla/xla.git
+          git fetch target
+          echo "HEAD_COMMIT=$(git rev-parse target/main)" >> $GITHUB_ENV
+
+      - name: Update OpenXLA commit in WORKSPACE
+        run: |
+          sed -i "s/xla_hash =.*/xla_hash = '${{ env.GITHUB_ENV }}'/" ${FILE_PATH}
+          git add ${FILE_PATH}
+          git commit -m "Weekly update WORKSPACE with (${{ env.GITHUB_ENV }})"
```
will-cromar, line 38: Do you need multiple commits/pushes?

.github/workflows/_openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Install sed
+        run: sudo apt-get update && sudo apt-get install -y sed
+
+      - name: Update date in setup.py
+        run: |
+          sed -i "s/_date =.*/_date = '${{ env.CURRENT_DATE }}'/" ${FILE_PATH}
+          git add ${FILE_PATH}
+          git commit -m "Weekly update setup.py to (${{ env.CURRENT_DATE }})"
+          git push origin HEAD:weekly-update
```
will-cromar, line 24: Can you create a new branch instead for each update? Currently this will fail on the second push, because `weekly-update` will fall out of date with `master`. And you don't want to force-push over the old branch if that PR is still open.

will-cromar (commented):
Thanks! This is a neat idea

.github/workflows/_openxla_pin_update_weekly.yml
ManfeiBai: Thanks, updated

.github/workflows/_openxla_pin_update_weekly.yml
ManfeiBai: Thanks, updated

.github/workflows/_openxla_pin_update_weekly.yml
ManfeiBai: thanks, udpated

.github/workflows/_openxla_pin_update_weekly.yml
ManfeiBai: Thanks, updated

.github/workflows/_openxla_pin_update_weekly.yml
ManfeiBai: Thanks, might not need, commented the first commit/push, wait for confirmation after merge to continue permanent change

.github/workflows/_openxla_pin_update_weekly.yml
ManfeiBai: Thanks, make sense, updated

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,79 @@
+name: Weekly Code Update PR
+on:
+  workflow_dispatch: # Manual trigger
+  schedule:
+    - cron: '0 9 * * 1'  # Each Monday 9:00AM (PDT time)
```
will-cromar, line 5: minor suggestion: comment out the cron schedule until you have a successful manual run or two.

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,79 @@
...
+          sed -i "s/_jax_version =.*/_jax_version = f'${{ env.JAX_VERSION }}.dev${{ env.CURRENT_DATE }}'/" ${FILE_PATH}
+          git add ${FILE_PATH}
+          # git commit -m "Weekly update setup.py to (${{ env.CURRENT_DATE }})"
+          # git push origin HEAD
+        env:
+          FILE_PATH: setup.py
+
+      - name: Fetch HEAD commit from OpenXLA github repo
+        run: |
+          git remote add target https://github.com/openxla/xla.git
```
will-cromar, line 44: There is actually a way to get commit hashes without adding/cloning a remote! Check out `git ls-remote`. `git ls-remote https://github.com/openxla/xla.git HEAD | awk '{print $1}'` should do what we need here.

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Create new branch
+        run: git checkout -b weekly-update-${{ env.CURRENT_DATE }}
+
+      - name: Install sed
+        run: sudo apt-get update && sudo apt-get install -y sed
+
+      - name: Get JAX version for specified date
+        run: |
+          JAX_VERSION=$(curl -s https://storage.googleapis.com/jax-releases/jax_releases.json | jq -r '.[] | select(.released_on == "${{ env.CURRENT_DATE }}") | .version')
```
will-cromar, line 25: Is `https://storage.googleapis.com/jax-releases/jax_releases.json` the correct path? I don't see a JSON file when I `gsutil ls gs://jax-releases`.

```
$ curl https://storage.googleapis.com/jax-releases/jax_releases.json
<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: jax-releases/jax_releases.json</Details></Error
```

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,79 @@
...
+        env:
+          FILE_PATH: setup.py
+
+      - name: Fetch HEAD commit from OpenXLA github repo
+        run: |
+          git remote add target https://github.com/openxla/xla.git
+          git fetch target
+          echo "HEAD_COMMIT=$(git rev-parse target/main)" >> $GITHUB_ENV
+
+      - name: Update OpenXLA commit in WORKSPACE
```
will-cromar, line 48: Can you combine this with "Update date in setup.py"? I don't think `FILE_PATH` needs to get added to the environment either way. You can just inline those paths since they're short and well known.

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Create Pull Request
+        uses: actions/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: weekly-update
+          base: main
+          title: "Weekly Code Update - ${{ env.CURRENT_DATE }}"
+          body: "This PR contains the weekly OpenXLA-pin update at commit: (HEAD commit: ${{ env.HEAD_COMMIT }}), with libtpu version at ${{ env.CURRENT_DATE }}."
+
+      - name: Add TPUCI label
```
will-cromar, line 66: `create-pull-request` has a `labels` option: https://github.com/marketplace/actions/create-pull-request#action-inputs

If the label is not on the PR when it is created, the TPU CI will not run.

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,79 @@
...
+          git commit -m "Weekly update WORKSPACE with (${{ env.GITHUB_ENV }})"
+          git push origin HEAD
+        env:
+          FILE_PATH: WORKSPACE
+
+      - name: Create Pull Request
+        uses: actions/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: weekly-update
```
will-cromar, line 61: Make sure to update the branch here too

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: Thanks, updated

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: COOL! thanks, updated

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: thanks, updated

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: nice catch, thanks, updated

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: good suggestion, have we used `create-pull-request` before? wanna confirm before use

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: since we have high frequency pin update right now, and its hard to gain jax version based on date;

so we might want to choose the latest jax dev version: learn from: https://github.com/pytorch/xla/pull/7797/files

confirm jax version match looks like is part of the pin-update work, so the assignee could help to confirm check the jax version now;

before jax has seperate release for torchxla and associated with date, this solution could be the temp solution, which is close to current workflow too

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: thanks, added labels `tpuci` in current PR, please correct me if I understand wrong

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,74 @@
...
+            exit 0
+          fi
+          echo "JAX_VERSION=$JAX_VERSION" >> $GITHUB_ENV
+
+      - name: Update date in setup.py
+        run: |
+          sed -i "s/_date =.*/_date = '${{ env.CURRENT_DATE }}'/" setup.py
+          sed -i "s/_jax_version =.*/_jax_version = f'${{ env.JAX_VERSION }}.dev${{ env.CURRENT_DATE }}'/" setup.py
+          git add setup.py
+          # git commit -m "Weekly update setup.py to (${{ env.CURRENT_DATE }})"
```
will-cromar, line 38: Can you remove these commented lines?

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Create new branch
+        run: git checkout -b weekly-update-${{ env.CURRENT_DATE }}
+
+      - name: Install sed
+        run: sudo apt-get update && sudo apt-get install -y sed
+
+      - name: Get JAX version for specified date
+        run: |
+          version_line=$(curl -s https://raw.githubusercontent.com/google/jax/main/jax/version.py | sed -n '24p')
```
will-cromar, line 25: Hahaha, I like this. I'm okay to merge this so we can see if the workflow works at all, but if the line number where `_version` is defined ever changes, this will break. Please leave a comment so we know where to look if the JAX version doesn't update.

.github/workflows/openxla_pin_update_weekly.yml
will-cromar: If the `tpuci` label is not on the PR when it is created, nothing will happen. I haven't use `create-pull-request` myself before, but I'm just going off of the documentation.  Adding `labels: tpuci` to the step above should work.

.github/workflows/openxla_pin_update_weekly.yml
```diff
@@ -0,0 +1,74 @@
...
+          # git commit -m "Weekly update setup.py to (${{ env.CURRENT_DATE }})"
+          # git push origin HEAD
+
+      - name: Fetch HEAD commit from OpenXLA github repo
+        run: |
+          echo "HEAD_COMMIT=$(git ls-remote https://github.com/openxla/xla.git HEAD | awk '{print $1}')" >> $GITHUB_ENV
+
+      - name: Update OpenXLA commit in WORKSPACE
+        run: |
+          sed -i "s/xla_hash =.*/xla_hash = '${{ env.GITHUB_ENV }}'/" WORKSPACE
```
will-cromar, line 47: This looks like a typo. `env.GITHUB_ENV` should be `env.HEAD_COMMIT`, right? `$GITHUB_ENV` holds the name of the file where workflow env vars are stored.

will-cromar (commented):
Otherwise LGTM

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: Thanks, good idea, updated with comments

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: Thanks, updated above with `labels: tpuci`

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: Thanks, updated

.github/workflows/openxla_pin_update_weekly.yml
ManfeiBai: Thanks, nice catch, updated to `env.HEAD_COMMIT`

.github/workflows/openxla_pin_update_weekly.yml
will-cromar: You can remove this last step since the PR will be created with that label

will-cromar (approved):

PR #7110, opened 2024-05-24
Update XLA pin to 2024/05/24 and fix Hermetic Python integration

vam-google:
### Hermetic Python
XLA is now built with [Hermetic Python](https://rules-python.readthedocs.io/en/latest/toolchains.html#workspace-toolchain-registration) from rules_python. Hermetic Python is essential for ensuring stability and reproduciblity of the build across different platforms, environments and bazel-built projects of ML ecosystem.

From now on all `bazel` commands called for test or build will always ignore system Python (or any other kind of Python installation a user may have, like pyenv) and all the Python packages you may have installed there. Instead it would use hermetic Python installation, meaning bazel will pull its own independent from the rest of the environment Python interpreter, the Python dependencies will be controlled by corresponding `requirements_lock_<py_version>.txt` files (like `requirements_lock_3_11.txt`) and they will be isolated from the system ones.

The version of Python used by the `bazel` build by default will match the version the system Python (note only versions will be matched, the actual interpreter used by `bazel` will still be a different one, just of the same version as your system one). 

#### Updating hermetic Python requirements
To update the requirements first make sure you have all the direct unlocked dependencies listed in `requirements.in` file then run 
```
bazel run //:requirements.update
``` 
The command above will update requirements_lock.txt file for the  Python versions same as your system Python

To update requirements lock for a specific version of Python, e.g. 3.12 run:
```
bazel run //:requirements.update --repo_env=HERMETIC_PYTHON_VERSION=3.12
```

#### The pytorch/dist/*.whl file is now REQUIRED
The pytorch/xla build also assumes that you have ran previously `python setup.py bdist_wheel` for the `pytorch` repository that `pytorch/xla` depends on. Specifically the build will look for a pytorch wheel (.whl file) inside `pytorch/dist` directory. 

Note, any pytorch wheels you may have installed locally in your system python will be ignored by the build, instead the wheel from the dist directory will be used each time. This is similar to how the rest of the build already depends on various `.so` files from `pytorch/build` directory, which is also created either by `python setup.py bdist_wheel` or `python setup.py develop` commands. 

The wheel file is needed because it will be automatically picked up by Hermetic Python (since system-installed pytorch package will be ignored).

Note, if you make any changes in pytorch itself, for those changes to be picked up by the pytorch/xla build the pytorch wheel must be rebuilt (similarly to how you already had to rebulid C++ parts of pytorch for them to be picked up by pytorch/xla bazel build).

### Other Changes
This PR also integrates other unrelated to Hermetic Python chagnes from latest XLA:
- `xla::ifrt::XlaProgram` renamed to `xla::ifrt::XlaProgram`
- `mlir/Transforms/TopologicalSortUtils.h` moved to `mlir/Analysis/TopologicalSortUtils.h`
- `xla_gpu_simplify_all_fp_conversions` flag removed, as it was removed from XLA and also XLA started to error out (instead of ignoring) when you pass non-supported (anymore) flags to it, so removing the flag from test command became necessary


ManfeiBai (approved):
LGTM, let's merge after all CI test green

requirements_lock_3_10.txt
```diff
@@ -0,0 +1,153 @@
+#
+# This file is autogenerated by pip-compile with Python 3.10
```
JackCaoG, line 2: in the future how do we update/generate these requirements lock files?

vam-google: `bazel run //:requirements.update` to update requirements matching your system python version.
`bazel run //:requirements.update --repo_env=HERMETIC_PYTHON_VERSION=3.8` to update for a specific version.

The JAX's https://jax.readthedocs.io/en/latest/developer.html#managing-hermetic-python describes everythign in detail.

Once this is pushed I'll make a xla-specific, shortened version of that and update your documentation as well.

alanwaketan:
Rebuilding PyTorch takes forever for a debug build. For Python files, can we use develop again?

JackCaoG:
@alanwaketan I think you can continue using the `develop`, you just need to do a `bdist_wheel` once and then `develop`. Unless you change the C++ file you don't have to do `bdist_wheel` again(I suspect even if you do change the C++ file you don't need to bdist_wheel since pytorch/xla only look up the pytorch whl for the lazy tensor code gen).

alanwaketan:
Thanks, Jack!

Issue #7095, opened 2024-05-22
`upsample_bilinear2d` returns unexpected data-type.


ysiraichi:
## 🐛 Bug

```python
def foo(x):
    return torch.nn.functional.upsample_bilinear(x, scale_factor=2)

a = torch.rand(1, 3, 10, 10, dtype=torch.half)
Xa = a.to(xm.xla_device())

# torch.float16
print(foo(a).dtype)

# torch.float32
print(foo(Xa).dtype)
```

## Environment

 - Reproducible on XLA backend [CPU/TPU/CUDA]: CUDA
 - torch_xla version: 8d35eb05693fdc52500cef32ae4a69669e27a8b2

## Additional context

This seems to happen due to the fact that we are computing on `F32` regardless of what the original input data type is.

https://github.com/pytorch/xla/blob/f336317ed044584a85b84e673a5b78058c4b8bae/torch_xla/csrc/resize_ops.cpp#L56-L59

cc @miladm @JackCaoG

ysiraichi:
Oh, wait. I think I got something wrong. Let me check it, again.

ysiraichi:
Note: the offending code was extracted from elsewhere:

https://github.com/pytorch/xla/blob/f336317ed044584a85b84e673a5b78058c4b8bae/torch_xla/csrc/resize_ops.cpp#L22-L23

ysiraichi:
@JackCaoG  I would imagine, not only we should return a tensor casted to the input data-type, but we also should do the computation in `f16`, since we are not using AMP. Let me know what you think.

JackCaoG:
I think this op is copied from tf, and from the comment in https://github.com/tensorflow/tensorflow/commit/f8b35e00afe09c8606bcb0441a51be8bd38168d2 TF output shape is always f32.. I guess we don't need to follow that rule.

ysiraichi:
I'm thinking that the way to go here is to compute and return using the input data-type, and not `f32`.
@JackCaoG how does that sound?

PR #7111, opened 2024-05-24
`upsample_bilinear`: fix output data-type.

ysiraichi:
Fix: #7095 

This PR removes the data-type conversion inherited from [the TF implementation](https://github.com/tensorflow/tensorflow/commit/f8b35e00afe09c8606bcb0441a51be8bd38168d2) when `is_kernel_bilinear` is set. We still do it on integer data-types, though.

Note: `upsample_bilinear` call on integral tensors (except `torch.byte` type) is not currently implemented on upstream PyTorch. It fails with:

```
  File "examples/upsample.py", line 16, in <module>
    out = foo(a)
  File "examples/upsample.py", line 7, in foo
    return torch.nn.functional.upsample_bilinear(x, scale_factor=2)
  File "torch/nn/functional.py", line 4203, in upsample_bilinear
    return interpolate(input, size, scale_factor, mode="bilinear", align_corners=True)
  File "torch/nn/functional.py", line 4082, in interpolate
    return torch._C._nn.upsample_bilinear2d(input, output_size, align_corners, scale_factors)
RuntimeError: "upsample_bilinear2d_channels_last" not implemented for 'Long'
```

cc @miladm @JackCaoG


JackCaoG (approved):

JackCaoG:
Seems like you need to resolve the conflict first.

JackCaoG:
I am going to revert this pr as it breaks the TPU CI. considering makeing this change GPU only.

PR #7114, opened 2024-05-24
implement Repeat  with fixed output shape

JackCaoG:


torch_xla/experimental/custom_kernel.py
```diff
@@ -655,6 +655,53 @@ def _make_group_metadata(
...
+  device = input.device
+
+  # to better understand this code, let's assume
+  # input.size() = [10]
+  # repeats = [0, 1, 2, 0, 4, 0, 6, 7, 8, 9]
+  # total_repeat_length = 20
+
+  # shift the repeats by one
+  # tensor([0, 0, 1, 2, 0, 4, 0, 6, 7, 8])
+  exclusive_repeats = torch.roll(repeats, shifts=1)
```
alanwaketan, line 677: Why do we want to shift right? If we don't shift, I assume we don't need to -1 at L700, and it will still output the correct answer?

torch_xla/experimental/custom_kernel.py
```diff
@@ -655,6 +655,53 @@ def _make_group_metadata(
...
+  block_split_indicators = torch.zeros(
+      total_repeat_length, dtype=torch.int64, device=device)
+  block_split_indicators.scatter_add_(0, valid_indices,
+                                      torch.ones_like(block_split_indicators))
+  # out_of_bound indices also scatter to index 0, need to offset them
+  block_split_indicators[0] -= out_of_bound_count
+
+  # value in gather_indices represents the index in the input.
+  # tensor([1, 2, 2, 4, 4, 4, 4, 6, 6, 6, 6, 6, 6, 7, 7, 7, 7, 7, 7, 7])
+  gather_indices = torch.cumsum(block_split_indicators, dim=0) - 1
```
alanwaketan, line 700: Trying to reason about this step... but really hard to understand... Can you explain a little bit to me? I couldn't get why the block_split_indicators can be converted to  gather_indices...

alanwaketan: I can understand for each indicator, the index needs to bump. Trying to understand the value to bump.

alanwaketan (approved):

alanwaketan:
Let's skip the GPU tests to move fast.

alanwaketan:
Thanks, Jack!

PR #7115, opened 2024-05-24
[Pallas] Support _histogram

alanwaketan:
Summary:
This pull request implements a limited version of https://pytorch.org/docs/stable/generated/torch.histc.html to support _make_group_metadata. In the future, we can consider properly lower the op or just make this python version more completed.

Test Plan:
PJRT_DEVICE=TPU python test/test_gmm.py

### DIFF
--- a/test/test_gmm.py
+++ b/test/test_gmm.py
@@ -221,11 +221,11 @@ class MegabloxTest(unittest.TestCase):
           max=test_grid['max'],
       )
 
-    self.assertTrue(
-        torch.all(torch_chart == chart.cpu()))
+    self.assertTrue(torch.all(torch_chart == chart.cpu()))
 
   def test_histogram_raise(self):
-    with self.assertRaisesRegex(AssertionError, "input must be of torch.int32 dtype."):
+    with self.assertRaisesRegex(AssertionError,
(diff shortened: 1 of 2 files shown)


test/test_gmm.py
```diff
@@ -183,6 +183,62 @@ def test_make_group_metadata(self):
...
+    for test_grid in test_grids:
+      torch_chart = torch.histc(
+          torch.tensor(test_grid['input'], dtype=torch.float),
+          bins=test_grid['max'] - test_grid['min'] + 1,
+          min=test_grid['min'],
+          max=test_grid['max'],
+      )
+
+      chart, _ = _histogram(
+          torch.tensor(test_grid['input'], dtype=torch.int32).to("xla"),
```
JackCaoG, line 219: hmm `torch` one takes float tensor and your implementation takes int32. Is this intentional?

JackCaoG (approved):

test/test_gmm.py
alanwaketan: Yea, it's intentional. torch one is more generic.

alanwaketan:
Let's skip the GPU tests to move fast.

Thanks Jack for approving the PR.

PR #7116, opened 2024-05-24
Add function for retrieving fallback operations.

ysiraichi:
This PR introduces `GetFallbackOperations` function, which sweeps through the fallback counter map `_cpu_fallback_counters` and returns the name of the functions whose counter is not zero, i.e. name of the fallbacks that were run.

There was a bug within the dynamo bridge that only classified as fallback operation those that came from the `aten` namespace. Other fallback operations, such as those that came from `torchvision::` were being passed on to the next steps, even though they were fallback operations.

This bug is one of the reasons of the issue #5967.

cc @miladm @JackCaoG

### DIFF
--- a/test/test_metrics.py
+++ b/test/test_metrics.py
@@ -205,6 +205,54 @@ class MetricsTest(unittest.TestCase):
     torch_xla._XLAC._xla_increment_counter('FakeCounter', 2)
     self.assertEqual(met.counter_value('FakeCounter'), 2)
 
+  def test_get_fallback_ops(self):
+
+    def getAndAssertFallbackOpsLenEquals(count):
+      fallback_ops = met.executed_fallback_ops()
+      fallback_ops_number = len(fallback_ops)
+      self.assertEqual(
+          fallback_ops_number,
+          count,
+          msg=f"found {fallback_ops_number}: {fallback_ops}")
+      return fallback_ops
+
+    # Reset all metrics, and make sure we don't start with any fallback ops.
+    met.clear_all()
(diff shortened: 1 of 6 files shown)


torch_xla/debug/metrics.py
```diff
@@ -79,3 +79,8 @@ def short_metrics_report(counter_names: list = None, metric_names: list = None):
         'TransferToDeviceTime', 'TransferFromDeviceTime'
     ]
   return torch_xla._XLAC._short_xla_metrics_report(counter_names, metric_names)
+
+
+def executed_fallback_ops():
```
JackCaoG, line 84: cc @alanwaketan we can start using this new api which is better than do a string assertIn from the metrics_report,

test/test_metrics.py
```diff
@@ -205,6 +205,54 @@ def test_pybind_increment_counter(self):
...
+
+    # Reset all metrics, and make sure we also don't retrieve any
+    # fallback operations.
+    met.clear_all()
+    getAndAssertFallbackOpsLenEquals(0)
+
+    # Run torchvision operations as fallback.
+    import torchvision
+    scores = torch.rand(N).to(xm.xla_device())
+    # NMS doesn't have a PyTorch/XLA implementation without dynamic shapes.
```
JackCaoG, line 246: maybe do an assert that dynamic shape is not enabled?

JackCaoG (approved):

alanwaketan (approved):
Cool. LGTM.

alanwaketan (approved):
Cool. LGTM.

PR #7125, opened 2024-05-28
Make from_dlpack handle cuda synchronization implicitly for input tensors that have __dlpack__ and __dlpack_device__ attributes.

vanbasten23:
from_dlpack should leverage `__dlpack__` and `__dlpack_device__` attributes of a tensor. That way, from_dlpack will handle the cuda synchronization implicity. Similar approaches can be found in [pytorch](https://github.com/pytorch/pytorch/blob/6d43c89f37160cbbce98a7161060cd45b61619e6/torch/utils/dlpack.py) and [jax](https://github.com/google/jax/blob/4461b7fe451cdb7c2e3e51bac6e1c0f78a4515bd/jax/_src/dlpack.py#L274-L278).

This PR fixes the direction when we convert an external tensor such as cuda tensor to an XLA tensor via from_dlpack. For the other direction (convert an XLA tensor to CUDA tensor) requires a change in upstream pytorch.

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2655,10 +2655,20 @@ class TestDLPack(parameterized.TestCase):
 
   @onlyIfTorchSupportsCUDA
   @onlyIfPJRTDeviceIsCUDA
-  def test_dlpack_xla_to_pytorch_cuda(self):
+  def test_dlpack_xla_to_pytorch_cuda_legacy(self):
     xla_t1 = torch.arange(5).to(xm.xla_device())
     dlt1 = xdlpack.to_dlpack(xla_t1)
     cuda_t1 = torch.utils.dlpack.from_dlpack(dlt1)
+    self.assertEqual(cuda_t1.device.type, 'cuda')
+    self.assertEqual(cuda_t1.device.index, xla_t1.device.index)
+    cuda_t1[0] = cuda_t1[0] + 20
+    self.assertTrue(torch.allclose(xla_t1.cpu(), cuda_t1.cpu()))
+
+  @onlyIfTorchSupportsCUDA
+  @onlyIfPJRTDeviceIsCUDA
+  def test_dlpack_xla_to_pytorch_cuda_protocol_conversion(self):
+    xla_t1 = torch.arange(5).to(xm.xla_device())
+    cuda_t1 = torch.utils.dlpack.from_dlpack(xla_t1)
     self.assertEqual(cuda_t1.device.type, 'cuda')
     self.assertEqual(cuda_t1.device.index, xla_t1.device.index)
     cuda_t1[0] = cuda_t1[0] + 20
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2592,7 +2592,7 @@ class TestDLPack(parameterized.TestCase):
 
   @onlyIfTorchSupportsCUDA
   @onlyIfPJRTDeviceIsCUDA
-  def test_dlpack_pytorch_cuda_to_xla_legacy(self):
+  def test_dlpack_pytorch_cuda_to_xla(self):
     t1_cuda = torch.arange(5).cuda()
     dlt1 = torch.utils.dlpack.to_dlpack(t1_cuda)
     xla_t1 = xdlpack.from_dlpack(dlt1)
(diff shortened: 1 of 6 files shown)


JackCaoG:
Let me know when this is ready for review

vanbasten23:
> Let me know when this is ready for review

Thanks. It's ready for review.

test/test_operations.py
```diff
@@ -2528,7 +2528,7 @@ def test_dlpack_roundtrip_bool(self):
 
   @onlyIfTorchSupportsCUDA
   @onlyIfPJRTDeviceIsCUDA
-  def test_dlpack_pytorch_cuda_to_xla(self):
+  def test_dlpack_pytorch_cuda_to_xla_legacy(self):
```
JackCaoG, line 2531: why legacy?

torch_xla/utils/dlpack.py
```diff
@@ -1,10 +1,33 @@
...
+    kDLExtDev = 12,
+    kDLOneAPI = 14,
 
 def from_dlpack(ext_tensor: Any):
-  return torch_xla._XLAC._from_dlpack(ext_tensor)
+  if hasattr(ext_tensor, '__dlpack_device__') and hasattr(ext_tensor, '__dlpack__'):
+    device_type, device_id = ext_tensor.__dlpack_device__()
+    if device_type == DLDeviceType.kDLGPU:
+      stream = torch_xla._XLAC._get_stream_for_cuda_device(device_id)
+      dlpack = ext_tensor.__dlpack__(stream=stream)
```
JackCaoG, line 27: what does this step do?

vanbasten23: For example, if we convert a CUDA tensor to XLA tensor, this line will synchronize the CUDA stream is necessary and call `to_dlpack` (torch_tensor.__dlpack__ defined here https://github.com/pytorch/pytorch/blob/b0ef363972203b163cddc95e4c6054b8221c2300/torch/_tensor.py#L1444-L1445).

test/test_operations.py
vanbasten23: I think initially, the dlpack protocol didn't require the tensor to have `__dlpack_device__` and `__dlpack__` and it was added later. Here is what JAX does https://github.com/google/jax/blob/ede94c3c81e1022af8bee3f60d6e5e0d65647c2a/jax/_src/dlpack.py#L278 (one legacy path and one non-legacy path). PyTorch calls it `protocol_conversion` as in https://github.com/pytorch/pytorch/blob/b0ef363972203b163cddc95e4c6054b8221c2300/test/test_dlpack.py#L50-L51. Let me rename it to conform with Pytorch.

test/test_operations.py
```diff
@@ -2621,6 +2621,37 @@ def test_dlpack_pytorch_cuda_to_xla(self):
     t3_cuda.fill_(6)
     self.assertTrue(torch.allclose(xla_t3.cpu(), t3_cuda.cpu()))
 
+  @onlyIfTorchSupportsCUDA
+  @onlyIfPJRTDeviceIsCUDA
+  def test_dlpack_pytorch_cuda_to_xla_protocol_conversion(self):
```
JackCaoG, line 2626: what does `protocol_conversion` mean?

vanbasten23: The dlpack protocol refers to the array  object need to implement `__dlpack__` and `__delpack_device_`, stated [here](https://screenshot.googleplex.com/BcK2RhyYJqdo3qD) and [here](https://screenshot.googleplex.com/6VmRVuSPHf7jBrc).

 Also, for this test scenario, I'm using a same naming convention as upstream pytorch https://github.com/pytorch/pytorch/blob/b0ef363972203b163cddc95e4c6054b8221c2300/test/test_dlpack.py#L50

torch_xla/utils/dlpack.py
```diff
@@ -1,10 +1,36 @@
 from typing import Any
+import enum
 import torch_xla
 
 
 def to_dlpack(xla_tensor: Any):
   return torch_xla._XLAC._to_dlpack(xla_tensor)
 
 
+class DLDeviceType(enum.IntEnum):
```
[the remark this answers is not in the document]
JackCaoG: it is fine, not blocking.

vanbasten23:
Thanks Jack for the review!

torch_xla/utils/dlpack.py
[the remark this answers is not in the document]
vanbasten23: Thanks. It's fixed in https://github.com/pytorch/xla/pull/7163.

PR #7128, opened 2024-05-28
Add `jlibrary` module to torch_xla2 for preserving library abstractions during export

GleasonK:
It is common to want to export torch->stablehlo with hardware-specific abstractions for better kernel support / mapping to compiler IR. This PR introduces the `jlibrary` APIs: `register_torch_composite` and `register_jax_composite`:

```python
def register_torch_composite(composite_name, impl, *ops, **jit_args): 
  """Register a composite using a JAX implementation.
    composite_name - The name of the library op to use in the exported composite
    impl           - A JAX lowering for the library operation
    *ops           - Variadic torch.ops to lower using `impl`.
    **jit_args     - Additional parameters to forward to JAX jit.
    """
  pass

def register_jax_composite(composite_name, impl, *ops, **jit_args):
  """Register a composite using a PyTorch implementation.
    composite_name - The name of the library op to use in the exported composite
    impl           - A torch method lowering for the library operation, can be the torch.library `@impl`.
    *ops           - Variadic torch.ops to lower using `impl`.
    **jit_args     - Additional parameters to forward to JAX jit.
    """
  pass
```

An example usage is provided in `test_libraries.py`, from a high level, to preserve custom library ops, or aten ops that your hardware has support for, use `jlibrary` to register their lowerings, and they will appear in StableHLO when using `torch_xla2.export` APIs.

Interested in feedback on:
- Module name (`jlibrary` since its next to `jtensor` and `jaten` and its a ptxla2 version of `torch.library`)
- Module location (should this be in `ops`?)
- Method names `register_torch_composite` takes a pytorch decomposition, `register_jax_composite` takes a jax decomposition (like jaten)

Followup tasks:
- Update lowering to export as `stablehlo.composites` instead of `call`.
- Add a means of capturing composite attributes (can this be inferred from library op? state_dict?).

### DIFF
--- a/experimental/torch_xla2/test/test_libraries.py
+++ b/experimental/torch_xla2/test/test_libraries.py
@@ -1,0 +1,80 @@
+import unittest
+import jax
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.library import Library, impl, impl_abstract
+import torch_xla2
+from torch_xla2 import tensor
+from torch_xla2.ops import jaten
+from torch_xla2.ops import jlibrary
+
+
+# Create a `mylib` library which has a basic SDPA op.
+m = Library("mylib", "DEF")
+m.define("scaled_dot_product_attention(Tensor q, Tensor k, Tensor v) -> Tensor")
+
+@impl(m, "scaled_dot_product_attention", "CompositeExplicitAutograd")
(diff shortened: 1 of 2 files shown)


qihqi (approved):
Very neat! Thanks!

PR #7129, opened 2024-05-28
Adding a bunch of ops to make torchbench's BERT_pytorch pass

qihqi:
python benchmarks/experiment_runner.py --suite-name=torchbench --experiment-config='{"accelerator":"CPU","xla":"PJRT","dynamo":null,"test":"eval","torch_xla2":"torch_export","xla_flags":null,"keep_model_data_on_cuda":false}' --model-config='{"model_name":"BERT_pytorch"}'
now passes.

Not all those ops are necessary to make this one model to pass.


experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -1989,3 +1990,48 @@ def _aten_native_batch_norm(input, weight, bias, running_mean, running_var, trai
     return torch.ops.aten._native_batch_norm_legit(input, weight, bias, running_mean, running_var, training, momentum, eps)
   else:
     return torch.ops.aten._native_batch_norm_legit_no_training(input, weight, bias, running_mean, running_var, momentum, eps)
+
+
+@op(torch.ops.aten.normal, needs_env=True)
```
zpcore, line 1995: Do you know when should we use `needs_env` for those ops?

zpcore (approved):
Cool, LGTM!

experimental/torch_xla2/torch_xla2/ops/jaten.py
qihqi: if the ops also want to have an keyword argument `env` passed to it, it should ask with needs_env. Usually random numbers generators needs it

PR #7134, opened 2024-05-29
[FSDPv2] Shard on the maximal dim of weights

alanwaketan:
Summary:
This pull request makes FSDPv2 to shard on the maximal dim of weights instead of the 0th dim.

Test Plan:
XLA_USE_SPMD=1 PJRT_DEVICE=TPU python test/spmd/test_fsdp_v2.py

### DIFF
--- a/torch_xla/experimental/spmd_fully_sharded_data_parallel.py
+++ b/torch_xla/experimental/spmd_fully_sharded_data_parallel.py
@@ -13,7 +13,9 @@ import torch_xla.distributed.spmd as spmd
 from torch_xla.distributed.fsdp.wrap import recursive_wrap
 
 
-def _prepare_spmd_partition_spec(param, extra_data_axis=None, shard_maximal=False):
+def _prepare_spmd_partition_spec(param,
+                                 extra_data_axis=None,
+                                 shard_maximal=False):
   shape = param.shape
   partition_spec = [None] * len(shape)
   # Skip scalar tensors and it replicated.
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/spmd_fully_sharded_data_parallel.py
```diff
@@ -113,7 +118,8 @@ def __init__(
     for param in module.parameters():
       if torch_xla._XLAC._get_xla_sharding_spec(param) != "":
         continue
-      spmd.mark_sharding(param, mesh, _prepare_spmd_partition_spec(param))
+      spmd.mark_sharding(
+          param, mesh, _prepare_spmd_partition_spec(param, shard_maximal=True))
```
JackCaoG, line 122: should we make it configureable?

JackCaoG (approved):

torch_xla/experimental/spmd_fully_sharded_data_parallel.py
alanwaketan: No, not necessary... It shouldn't matter to the user...

alanwaketan:

torch_xla/experimental/spmd_fully_sharded_data_parallel.py
JackCaoG: ok I think I get it.. We need to do a all-gather anyway before entering the layer and only one dimension is being sharded.

alanwaketan: Yea, that's right. If the 0th dim is like size of 8 (MoE) and we are sharding it on v5p-2048, it will be a disaster.

alanwaketan:
Skip GPU to move fast.

PR #7073, opened 2024-05-16
Add CI workflow for tests that requires pytorch CUDA.

vanbasten23:
This PR adds a new CI workflow that build pytorch with CUDA enabled from source, build pytorch/xla with CUDA enabled from source, then run tests. The intention is to run tests that requires pytorch with CUDA.

In detail, this PR add 2 more jobs to .github/workflows/build_and_test.yml
1. build-torch-with-cuda-xla-with-cuda: build pytorch with cuda and build pytorch/xla (It's ok to build torch_xla with cuda enabled now because it takes 2537s to build pytorch and takes 352s to build torch_xla).
3. test-cuda-with-pytorch-cuda-enabled: only run the tests that requires pytorch CUDA.

PR #7140, opened 2024-05-29
Add a CI workflow for tests that requires pytorch CUDA.

vanbasten23:
This PR adds a new CI workflow that build pytorch with CUDA enabled from source, then run tests that requires torch CUDA. Today, the tests requiring torch CUDA are skipped.

In detail, this PR add 2 more jobs to .github/workflows/build_and_test.yml
1. build-torch-with-cuda: build pytorch with cuda
2. test-cuda-with-pytorch-cuda-enabled: only run the tests that requires pytorch CUDA.

### DIFF
--- a/.github/workflows/_build_torch_with_cuda.yml
+++ b/.github/workflows/_build_torch_with_cuda.yml
@@ -28,6 +28,7 @@ jobs:
       BAZEL_JOBS: 16
       BAZEL_REMOTE_CACHE: 1
       BUILD_CPP_TESTS: 1
+      _GLIBCXX_USE_CXX11_ABI: 0
     steps:
       # See https://github.com/actions/checkout/issues/1014#issuecomment-1906802802
       - name: Clean up workspace
--- a/.github/workflows/_build_torch_with_cuda.yml
+++ b/.github/workflows/_build_torch_with_cuda.yml
@@ -34,8 +34,6 @@ jobs:
         run: |
           echo "PATH=$PATH:/usr/local/cuda-12.1/bin" >> $GITHUB_ENV
           echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-12.1/lib64" >> $GITHUB_ENV
-
-          echo 'xw32 got torch-commit:' ${{ inputs.torch-commit }}
       - name: Check GPU
         run: nvidia-smi
       - name: Checkout PyTorch Repo
--- a/.github/workflows/_build_torch_with_cuda.yml
+++ b/.github/workflows/_build_torch_with_cuda.yml
@@ -29,6 +29,11 @@ jobs:
       BAZEL_REMOTE_CACHE: 1
       BUILD_CPP_TESTS: 1
     steps:
+      # See https://github.com/actions/checkout/issues/1014#issuecomment-1906802802
+      - name: Clean up workspace
+        run: |
+          ls -la
+          rm -rvf ${GITHUB_WORKSPACE}/*
       - name: Setup gcloud
         shell: bash
         run: |
--- a/.github/workflows/_build_torch_with_cuda.yml
+++ b/.github/workflows/_build_torch_with_cuda.yml
@@ -16,6 +16,7 @@ jobs:
     runs-on: ${{ inputs.runner }}
     container:
       image: ${{ inputs.dev-image }}
+      options: "--gpus all --shm-size 16g"
     env:
       _GLIBCXX_USE_CXX11_ABI: 0
     steps:
@@ -29,6 +30,8 @@ jobs:
         run: |
           echo "PATH=$PATH:/usr/local/cuda-12.1/bin" >> $GITHUB_ENV
           echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-12.1/lib64" >> $GITHUB_ENV
+      - name: Check GPU
+        run: nvidia-smi
       - name: Checkout PyTorch Repo
         uses: actions/checkout@v4
         with:
@@ -40,7 +43,6 @@ jobs:
         run: |
           cd pytorch
           USE_CUDA=1 python setup.py bdist_wheel
-          ls dist/
       - name: Upload wheel
         uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/_build_torch_with_cuda.yml
+++ b/.github/workflows/_build_torch_with_cuda.yml
@@ -6,6 +6,10 @@ on:
         required: true
         type: string
         description: Base image for builds
+      torch-commit:
+          required: true
+          type: string
+          description: torch-commit
       runner:
         required: false
         type: string
@@ -30,6 +34,8 @@ jobs:
         run: |
           echo "PATH=$PATH:/usr/local/cuda-12.1/bin" >> $GITHUB_ENV
           echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-12.1/lib64" >> $GITHUB_ENV
+
+          echo 'xw32 got torch-commit:' ${{ inputs.torch-commit }}
       - name: Check GPU
         run: nvidia-smi
       - name: Checkout PyTorch Repo
@@ -37,6 +43,7 @@ jobs:
         with:
           repository: pytorch/pytorch
           path: pytorch
+          ref: ${{ inputs.torch-commit }}
           submodules: recursive
       - name: Build
         shell: bash
--- a/.github/workflows/_build_torch_with_cuda.yml
+++ b/.github/workflows/_build_torch_with_cuda.yml
@@ -24,6 +24,11 @@ jobs:
         run: |
           ls -la
           rm -rvf ${GITHUB_WORKSPACE}/*
+      - name: Setup CUDA environment
+        shell: bash
+        run: |
+          echo "PATH=$PATH:/usr/local/cuda-12.1/bin" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-12.1/lib64" >> $GITHUB_ENV
       - name: Checkout PyTorch Repo
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/_build_torch_with_cuda.yml
+++ b/.github/workflows/_build_torch_with_cuda.yml
@@ -1,4 +1,4 @@
-name: build-torch-with-cuda-xla-with-cuda
+name: build-torch-with-cuda
 on:
   workflow_call:
     inputs:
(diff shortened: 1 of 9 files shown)


vanbasten23:
The original comment is at https://github.com/pytorch/xla/pull/7073

.github/workflows/_build_torch_with_cuda.yml
```diff
@@ -0,0 +1,69 @@
...
+      BAZEL_REMOTE_CACHE: 1
+      BUILD_CPP_TESTS: 1
+      _GLIBCXX_USE_CXX11_ABI: 0
+    steps:
+      # See https://github.com/actions/checkout/issues/1014#issuecomment-1906802802
+      - name: Clean up workspace
+        run: |
+          ls -la
+          rm -rvf ${GITHUB_WORKSPACE}/*
+      - name: Setup gcloud
```
will-cromar, line 38: Skip the `gcloud` secret here since you're not building anything with bazel

.github/workflows/_build_torch_with_cuda.yml
```diff
@@ -0,0 +1,69 @@
...
+        shell: bash
+        run: |
+          echo "${GCLOUD_SERVICE_KEY}" > $GOOGLE_APPLICATION_CREDENTIALS
+      - name: Setup CUDA environment
+        shell: bash
+        run: |
+          echo "PATH=$PATH:/usr/local/cuda-12.1/bin" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-12.1/lib64" >> $GITHUB_ENV
+      - name: Check GPU
```
will-cromar, line 47: You don't need a GPU present for the GPU build

.github/workflows/_build_torch_with_cuda.yml
```diff
@@ -0,0 +1,69 @@
...
+        run: nvidia-smi
+      - name: Checkout PyTorch Repo
+        uses: actions/checkout@v4
+        with:
+          repository: pytorch/pytorch
+          path: pytorch
+          submodules: recursive
+      - name: Checkout PyTorch/XLA Repo
```
will-cromar, line 55: You can skip the pytorch/xla checkout, right?

.github/workflows/_build_torch_with_cuda.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Build
+        shell: bash
+        run: |
+          cd pytorch
+          USE_CUDA=1 python setup.py bdist_wheel
+          ls dist/
+      - name: Upload wheel
+        uses: actions/upload-artifact@v4
+        with:
+          name: torch-with-cuda-xla-with-cuda-wheels
```
will-cromar, line 68: I suggest renaming this to make the workflow (`torch-with-cuda`) since there are no longer `torch_xla` wheels bundled here.

.github/workflows/_test_requiring_torch_cuda.yml
```diff
@@ -0,0 +1,129 @@
...
+      BAZEL_JOBS: 16
+      BAZEL_REMOTE_CACHE: 1
+    steps:
+      # See https://github.com/actions/checkout/issues/1014#issuecomment-1906802802
+      # TODO: need to find a way to reuse these steps.
+      - name: Clean up workspace
+        run: |
+          ls -la
+          rm -rvf ${GITHUB_WORKSPACE}/*
+      - name: Setup gcloud
```
will-cromar, line 51: Same here. This won't require the `gcloud` secret.

.github/workflows/_test_requiring_torch_cuda.yml
```diff
@@ -0,0 +1,129 @@
...
+          pip install fsspec
+          pip install rich
+
+          echo "Import check..."
+          python -c "import torch, torch_xla, torchvision"
+          echo "Import check done."
+          echo "Check if CUDA is available for PyTorch..."
+          python -c "import torch; assert torch.cuda.is_available()"
+          echo "CUDA is available for PyTorch."
+      - name: Record PyTorch commit
```
will-cromar, line 100: Just be aware, there's a small chance of version skew in this workflow. You could end up with a broken workflow if there's a breaking change between when `_bulid_torch_xla` and `_build_torch_with_cuda` launch:

- _build_torch_xla launches, pulls pytorch, etc
- breaking change merged to pytorch
- _build_torch_with_cuda, pulls pytorch, etc

It's up to you if you want to find a way to fix that now or just wait to see if it's an issue in practice. There should really only a window of seconds between when workflow launch.

The permanent fix would probably be to add a parent workflow step that "records" the PyTorch commit hash before either of the builds run. Maybe something to address for whoever factors out the setup steps.

infra/ansible/roles/build_srcs/tasks/main.yaml
```diff
@@ -27,8 +27,7 @@
     cmd: python setup.py bdist_wheel
     chdir: "{{ (src_root, 'pytorch') | path_join }}"
     creates: "{{ (src_root, 'pytorch/dist/*.whl') | path_join }}"
-  # Set `USE_CUDA=0` as PyTorch cannot be used with GPU in eager and XLA mode.
-  environment: "{{ env_vars | combine({'USE_CUDA': 0}) }}"
+  environment: "{{ env_vars }}"
```
will-cromar, line 30: You can probably revert the ansible changes, right?

.github/workflows/_test_requiring_torch_cuda.yml
```diff
@@ -0,0 +1,129 @@
+name: xla-test-requiring-torch-cuda
```
will-cromar, line 1: minor suggestion: maybe name this one after the test objective? e.g. `test_cuda_interop` or something like that

will-cromar (approved):
Overall LGTM. Thanks!

.github/workflows/_build_torch_with_cuda.yml
vanbasten23: Good catch!

.github/workflows/_test_requiring_torch_cuda.yml
vanbasten23: Thanks for the suggestion. It turned out more features than interoperability need torch CUDA, for example triton, etc.

.github/workflows/_test_requiring_torch_cuda.yml
vanbasten23: Thanks for the catch!

.github/workflows/_test_requiring_torch_cuda.yml
vanbasten23: It's great point. I'll see if it's an issue in practice. If so, I'll fix it accordingly.

infra/ansible/roles/build_srcs/tasks/main.yaml
vanbasten23: Right, as I don't use the ansible to build the torch wheel.

vanbasten23:
actually, if I build torch with CUDA enabled on a machine without GPU, it fails with error segfault and OOM (https://gist.github.com/vanbasten23/faba932d31baad28196176ebd5b80948), even if I added the CUDA info to the PATH and LD_LIBRARY_PATH (otherwise, building torch with CUDA fails with https://gist.github.com/vanbasten23/338017e9b182cdf6fba49c75aabd2268)

.github/workflows/_test_requiring_torch_cuda.yml
vanbasten23: Actually I saw it in this PR:
```
  File "/usr/local/lib/python3.10/site-packages/torch_xla/__init__.py", line 7, in <module>
    import _XLAC
ImportError: /usr/local/lib/python3.10/site-packages/_XLAC.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN3c[101](https://github.com/pytorch/xla/actions/runs/9308352858/job/25626741718?pr=7140#step:11:102)5SmallVectorBaseIjE8grow_podEPvmm
```
suggesting the torch wheel and torch_xla wheel are not compatible.

So I added a parent job to get the commit and pass it to the reusable workflow `build-torch-xla` and `build-torch-with-cuda` according to https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-outputs-from-a-reusable-workflow

.github/workflows/_get_torch_commit.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          repository: pytorch/pytorch
+          path: pytorch
+          submodules: recursive
+      - id: get_torch_commit
+        name: Get torch commit
+        run: |
+          cd pytorch
+          torch_commit=$(git rev-parse HEAD) 
+          echo "torch_commit=$torch_commit" >> "$GITHUB_OUTPUT"
```
will-cromar, line 31: I just learned about the [`git ls-remote`](https://www.git-scm.com/docs/git-ls-remote) command, which looks perfect here. `git ls-remote https://github.com/pytorch/pytorch.git HEAD | awk '{print $1}'` should work, and then you don't have to clone.

If this ends up being a single step and we don't re-use it, is it possible to actually put that directly into `build_and_test.yaml`?

.github/workflows/_build_torch_xla.yml
```diff
@@ -37,6 +41,7 @@ jobs:
         with:
           repository: pytorch/pytorch
           path: pytorch
+          ref: ${{ inputs.torch-commit }}
```
will-cromar, line 44: Can you make the same change to `_test.yml`? This is way cleaner than what I did:

https://github.com/pytorch/xla/blob/8471826a52417e010a21bbde70a28f426f7a2ad0/.github/workflows/_test.yml#L131-L144

will-cromar (commented):
Thanks for adding the `_get_torch_commit` step! That's much cleaner than the way I was getting the `torch` commit for `_test.yml` :sweat_smile:

.github/workflows/build_and_test.yml
```diff
@@ -20,14 +20,30 @@ concurrency:
...
 
+  build-torch-with-cuda:
+    name: "Build PyTorch with CUDA"
+    uses: ./.github/workflows/_build_torch_with_cuda.yml
+    needs: get-torch-commit
+    with:
+      # note that to build a torch wheel with CUDA enabled, we do not need a GPU runner.
+      dev-image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.10_cuda_12.1
+      torch-commit: ${{needs.get-torch-commit.outputs.torch_commit}}
+      runner: linux.8xlarge.nvidia.gpu
```
will-cromar, line 45: A non-GPU machine should work here, as long as you have the CUDA docker image.

vanbasten23: Yeah, I tried it on a non-GPU machine but it didn't work. I put the detail here https://github.com/pytorch/xla/pull/7140#issuecomment-2140757431.

.github/workflows/_build_torch_xla.yml
vanbasten23: Do you mind if I do it in a follow up PR?

vanbasten23:
Thanks Will for the review!

vanbasten23:
cc @bhavya01 this should unblock you.

.github/workflows/_build_torch_xla.yml
will-cromar: Yeah, works for me.

PR #7144, opened 2024-05-29
isolate the PJRT dependency between torch_xla and torch_xla2

zpcore:
We should not trigger error like:
```
F0529 21:08:47.961712 3737043 pjrt_registry.cc:139] Non-OK-status: tpu_status status: INVALID_ARGUMENT: Mismatched PJRT plugin PJRT API version (0.51) and framework PJRT API version 0.53).
```
when we install the JAX dependency from torch_xla2.

Basically I get rid of checkings for xla devices using torch_xla related functions. In addition, we initiate benchmark models to CPU device first before moving to jax device.

### DIFF
--- a/benchmarks/benchmark_experiment.py
+++ b/benchmarks/benchmark_experiment.py
@@ -80,6 +80,7 @@ class ExperimentLoader:
     # torch_xla2 doesn't support dynamo at this time.
     if cfg_dynamo is not None and cfg_torch_xla2:
       return False
+    use_xla2 = True if cfg_torch_xla2 else False
 
     # Check dynamo backend-specifics constraints.
     if cfg_dynamo == "inductor":
@@ -97,8 +98,8 @@ class ExperimentLoader:
       raise NotImplementedError
 
     # Check XLA device available if requested.
-    if cfg_xla is not None and not is_xla_device_available(
-        cfg_accelerator.upper()):
+    if not use_xla2 and cfg_xla is not None and not is_xla_device_available(
+        cfg_accelerator.upper(), use_xla2):
       return False
 
     # Check accelerator constraints.
@@ -151,8 +152,10 @@ class BenchmarkExperiment:
       process_env.pop("XRT_TPU_CONFIG", None)
       process_env.pop("XLA_FLAGS", None)
 
+    use_xla2 = False
     if self.torch_xla2:
       process_env["JAX_PLATFORMS"] = self.accelerator.lower()
+      use_xla2 = True
 
     if self.xla == "PJRT":
       process_env["PJRT_DEVICE"] = self.accelerator.upper()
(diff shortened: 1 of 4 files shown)


benchmarks/util.py
```diff
@@ -50,13 +50,13 @@ def reset_rng_state(benchmark_experiment=None):
   torch.manual_seed(1337)
   random.seed(1337)
   np.random.seed(1337)
-  if benchmark_experiment is not None and benchmark_experiment.xla is not None and benchmark_experiment.torch_xla2 is not None:
+  if benchmark_experiment is not None and benchmark_experiment.xla is not None and benchmark_experiment.torch_xla2 is None:
```
vanbasten23, line 53: if we don't use torch_xla2, we don't need to do `xm.set_rng_state(1337, str(device))`?

benchmarks/check_xla_device.py
```diff
@@ -1,12 +1,24 @@
 import sys
-assert len(sys.argv) == 2
+
+assert len(sys.argv) in (2, 3)
```
vanbasten23, line 3: could we make a function `use_torch_xla2`?

benchmarks/benchmark_experiment.py
```diff
@@ -80,6 +80,7 @@ def _is_available(self, experiment_config):
     # torch_xla2 doesn't support dynamo at this time.
     if cfg_dynamo is not None and cfg_torch_xla2:
       return False
+    use_xla2 = True if cfg_torch_xla2 else False
```
vanbasten23, line 83: nit: line81 could also use `use_xla2`

benchmarks/benchmark_experiment.py
```diff
@@ -165,13 +168,16 @@ def update_process_env(self, process_env):
...
-      if is_xla_device_available(self.accelerator.upper()):
+      if is_xla_device_available(self.accelerator.upper(), use_xla2):
         process_env["PJRT_DEVICE"] = self.accelerator.upper()
 
     if self.xla_flags:
       process_env["XLA_FLAGS"] = self.xla_flags
 
   def get_device(self):
+    if self.torch_xla2:
+      # Initiate the model in CPU first for xla2. We will move the model to jax device later.
```
vanbasten23, line 179: can you add the reason to the comment?

benchmarks/check_xla_device.py
```diff
@@ -1,12 +1,24 @@
...
 import os
 os.environ["PJRT_DEVICE"] = devkind
 
-import torch_xla.core.xla_model as xm
-devlist = xm.get_xla_supported_devices(devkind=devkind)
+if not use_xla2:
+  import torch_xla.core.xla_model as xm
+  devlist = xm.get_xla_supported_devices(devkind=devkind)
+else:
+  # torch_xla2 needs jax to detect the device
```
vanbasten23, line 16: This is a very useful comment! Thanks

benchmarks/util.py
zpcore: I am using `benchmark_experiment.torch_xla2 is None` in the condition. If we use torch_xla2, we don't set the rng state. 

We will work on a solution to set rng state in jax device later.

benchmarks/benchmark_experiment.py
zpcore: Good catch!

benchmarks/benchmark_experiment.py
zpcore: Done

benchmarks/check_xla_device.py
zpcore: Done

benchmarks/util.py
vanbasten23: SG. Better to add a TODO

vanbasten23 (approved):

PR #7151, opened 2024-05-30
[Pallas] Introduce gmm_backward

alanwaketan:
Summary:
This pull request introduces a helper for gmm_backward. I'm still debuting if we need to make gmm as a autograd.function given we will do manual back-propagation in Mixtral.

Test Plan:
python test/test_gmm.py

### DIFF
--- a/test/test_gmm.py
+++ b/test/test_gmm.py
@@ -7,7 +7,7 @@ import torch
 import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.debug.metrics as met
-from torch_xla.experimental.custom_kernel import gmm, _make_group_metadata, _histogram, tgmm
+from torch_xla.experimental.custom_kernel import gmm, _make_group_metadata, _histogram, tgmm, gmm_backward
 from torch_xla import runtime as xr
 from torch_xla._internal import tpu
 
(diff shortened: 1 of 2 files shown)


test/test_gmm.py
```diff
@@ -344,6 +344,36 @@ def test_tgmm_bf16(self):
     # Make sure tgmm doesn't fallback.
     self.assertNotIn("aten::", met.short_metrics_report())
 
+  @unittest.skipIf(xr.device_type() != 'TPU', "This test only works on TPU.")
+  def test_gmm_backward(self):
+    self._init_test_cases()
```
JackCaoG, line 349: nit, you might need a `met.clear_all()` here.

JackCaoG (approved):
What do you mean by "manual backprop in mixtral"?

test/test_gmm.py
alanwaketan: Let me fix it in the next PR. Don't want to waste CI cycles.

alanwaketan:
Skip GPU tests to move fast.

alanwaketan:
> What do you mean by "manual backprop in mixtral"?

Just we need to override the MoE back prop to accommondate gmm backward and manual sharding. You will know what that means once the code is ready. Thanks for approving this change.

PR #7152, opened 2024-05-30
[Pallas] Introduce GMM(torch.autograd.Function)

alanwaketan:
Summary:
This pull request make GMM as a torch.autograd.Function such that we can use torch.autograd.backward instead of manual backpropagation.

Test Plan:
python test/test_gmm.py

### DIFF
--- a/test/test_gmm.py
+++ b/test/test_gmm.py
@@ -7,7 +7,7 @@ import torch
 import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.debug.metrics as met
-from torch_xla.experimental.custom_kernel import gmm, _make_group_metadata, _histogram, tgmm, gmm_backward
+from torch_xla.experimental.custom_kernel import gmm, _make_group_metadata, _histogram, tgmm, gmm_backward, GMM
 from torch_xla import runtime as xr
 from torch_xla._internal import tpu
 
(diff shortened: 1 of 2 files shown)


test/test_gmm.py
```diff
@@ -374,6 +374,81 @@ def test_gmm_backward(self):
     # Make sure gmm doesn't fallback.
     self.assertNotIn("aten::", met.short_metrics_report())
 
+  @unittest.skipIf(xr.device_type() != 'TPU', "This test only works on TPU.")
```
JackCaoG, line 377: do you need TPU version check here?

alanwaketan: lol, good question.

alanwaketan: v2 is pretty happy on the tree.

test/test_gmm.py
```diff
@@ -374,6 +374,81 @@ def test_gmm_backward(self):
...
+      rhs.retain_grad()
+
+      ref_out = self._reference_gmm(lhs, rhs, group_sizes)
+      ref_out.sum().backward()
+
+      torch.manual_seed(42)
+      lhs_xla = torch.rand(m, k, dtype=lhs_dtype, requires_grad=True).to("xla")
+      rhs_xla = torch.rand(
+          num_groups, k, n, dtype=rhs_dtype, requires_grad=True).to("xla")
+      lhs_xla.retain_grad()
```
JackCaoG, line 401: just curious how is this different than `lhs_xla. requires_grad  = True`?

test/test_gmm.py
```diff
@@ -374,6 +374,81 @@ def test_gmm_backward(self):
...
+      out.sum().backward()
+
+      self.assertTrue(torch.allclose(ref_out, out.cpu()))
+      self.assertTrue(torch.allclose(lhs.grad, lhs_xla.grad.cpu()))
+      self.assertTrue(torch.allclose(rhs.grad, rhs_xla.grad.cpu()))
+
+    # Make sure gmm doesn't fallback.
+    self.assertNotIn("aten::", met.short_metrics_report())
+
+  @unittest.skipIf(xr.device_type() != 'TPU', "This test only works on TPU.")
```
JackCaoG, line 414: ditto, otherwise I think it will break v2 CI

JackCaoG (approved):

test/test_gmm.py
alanwaketan: It make the .grad appears. Otherwise, it's None.

test/test_gmm.py
JackCaoG: interesting.. I thought pallas is not supported on v2.

test/test_gmm.py
alanwaketan: Actually, the CI somehow is happy. Maybe it just don't like the others.

test/test_gmm.py
alanwaketan: gmm is just mm... The kernel is simple... Other fuses softmax, etc...

alanwaketan:
Thanks Jack for approving.

alanwaketan:
Skip GPU tests to move fast.

PR #7155, opened 2024-05-30
Use dest_offsets directly in LoadPlanner

jonb377:
In [create_read_items_for_chunk_list](https://github.com/pytorch/pytorch/blob/3174e6cb8e2d37210c7569e51dc6a9522110e0f3/torch/distributed/checkpoint/planner_helpers.py#L149), the ReadItem's `dest_offsets` are set to index into the local shard. Despite this, we performed an  translation from global to local indices. This caused issues restoring checkpoints with padded tensors, i.e. where the global tensor size is not evenly divisible by the shard size.

We should instead rely on `dest_offsets` directly to narrow the local shards.

### DIFF
--- a/test/spmd/test_xla_distributed_checkpoint.py
+++ b/test/spmd/test_xla_distributed_checkpoint.py
@@ -156,7 +156,7 @@ class EndToEndCheckpointTest(DistributedCheckpointTestBase):
         load_planner=SPMDLoadPlanner())
 
   @unittest.skipIf(xr.global_runtime_device_count() == 1,
-  "Multiple devices needed to change mesh")
+                   "Multiple devices needed to change mesh")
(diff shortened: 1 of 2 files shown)


dasoto:
Any reason this has not been merged?

As context this also fixed an issue we have by saving a distributed checkpoint on a 8x16 and then loading it in other topology.

jonb377:
@alanwaketan Do you have a sec to review? I'll also add @JackCaoG

JackCaoG (approved):

jonb377:
Yes, we should cherry-pick this. I'll open a PR.

alanwaketan (commented):
LGTM. Sorry for missing this PR.

PR #7159, opened 2024-05-30
Backport doc fix

lsy323:


ManfeiBai:
thanks, Siyuan, iirc, we might not want to touch r2.3 after release, do you plan to push to a new branch related to r2.3? but I'm not sure whether we still could push the doc from the new branch

will-cromar (approved):

lsy323:
> thanks, Siyuan, iirc, we might not want to touch r2.3 after release, do you plan to push to a new branch related to r2.3? but I'm not sure whether we still could push the doc from the new branch

I don't think pushing to 2.3 branch after release is the best practice. But I don't think there is an alternative to fix the doc other than pushing to 2.3 branch.

PR #7162, opened 2024-05-31
torchbench CI

zpcore:
Add the CI test for two torchbench models. This is a sanity test for benchmarking script.

Torchbench test requires 1) install torchvision, audio, test dependencies; 2) checkout the pytorch/benchmark repo and Install the torchbench models. Those two steps are introduced in the CI for this PR.


infra/ansible/roles/build_srcs/tasks/main.yaml
```diff
@@ -137,6 +137,46 @@
   ansible.builtin.pip:
     name: "{{ torchvision_wheels.files | map(attribute='path') }}"
 
+- name: Read Torchaudio pin
```
lsy323, line 140: This will make torchaudio/text whl in all build artifacts (e.g. release) as main.yaml is shared for different build. cc @will-cromar for if there is a better way to make these only apply for the test container.

lsy323 (commented):
Left 1 note on making the torchaudio/text whl only available to test image. But I'm not sure how to do that. Otherwise LGTM

zpcore:
Hi @vanbasten23 , I saw you added the torch cuda build in the flow. Now do we need to use CUDA torch in GPU tests or the CUDA build is serving a different purpose?

zpcore:
Hi @will-cromar , @JackCaoG , friendly pin to see if there are any feedbacks on the PR.

vanbasten23:
> Now do we need to use CUDA torch in GPU tests or the CUDA build is serving a different purpose?

We don't need the CUDA torch for GPU tests because most of our GPU tests https://github.com/pytorch/xla/blob/cdb4aec6f6d168024c0ac25996652d1ed2ab6417/.github/workflows/build_and_test.yml#L67-L79 doesn't need CUDA torch. This is intentional because building torch with CUDA takes much longer time and we want our developers to get the test result quicker. cc @will-cromar 

The CUDA build is used to run the tests that only requires CUDA torch https://github.com/pytorch/xla/blob/cdb4aec6f6d168024c0ac25996652d1ed2ab6417/.github/workflows/build_and_test.yml#L81-L90.

infra/ansible/roles/build_srcs/tasks/main.yaml
zpcore: Disscussed with @will-cromar offline, he suggests we move the installation into run_tests.sh to prevent adding extra dependencies in our docker build.

zpcore: Relocated all the torchbench dependecies installation from .yml into the testing script.

.github/scripts/run_tests.sh
```diff
@@ -92,17 +102,20 @@ fi
...
   run_torch_xla_benchmark_tests $XLA_DIR
+  run_torch_xla_torchbench_tests $PYTORCH_DIR $XLA_DIR
 else
   # run tests separately.
   if [[ "$RUN_PYTHON_TESTS" == "python_tests" ]]; then
-    run_torch_xla_python_tests $PYTORCH_DIR $XLA_DIR $USE_COVERAGE
+    run_torch_xla_python_tests $XLA_DIR $USE_COVERAGE
   elif [[ "$RUN_BENCHMARK_TESTS" == "benchmark_tests" ]]; then
     run_torch_xla_benchmark_tests $XLA_DIR
+  elif [[ "$RUN_TORCHBENCH_TESTS" == "torchbench_tests" ]]; then
```
will-cromar, line 116: Can you combine these with the `benchmark_tests`? That suite is very small and only takes ~1 minute to run.

.github/scripts/run_tests.sh
```diff
@@ -1,34 +1,32 @@
 set -ex
 
 function run_torch_xla_python_tests() {
-  PYTORCH_DIR=$1
-  XLA_DIR=$2
-  USE_COVERAGE="${3:-0}"
+  XLA_DIR=$1
```
will-cromar, line 4: What's the reason for removing the `PYTORCH_DIR` argument?

zpcore: I am wondering why we need PYTORCH_DIR in those tests. The PYTORCH_DIR passed in to those bash scripts are not absolute path. It's just something like `pytorch/`.

.github/scripts/run_tests.sh
zpcore: Combined

.github/scripts/run_tests.sh
zpcore: And we didn't use PYTORCH_DIR variable in the function. So I cleaned them up.

zpcore:
Friendly pin to see if there are other feedbacks. @JackCaoG @will-cromar @lsy323 @anbasten23

.github/scripts/run_tests.sh
will-cromar: I thought that variable was being used by one of the downstream scripts. If the tests still pass, this is fine though.

will-cromar (approved):

zpcore:
The GPU test (python tests, torch mp op) is also failing on heads.  I will merge this one for now.

PR #7168, opened 2024-06-03
Re-land: `upsample_bilinear`: fix output data-type.

ysiraichi:
Re-land: #7111 

Previously, it was reverted because it broke TPU CI. That happened due to the newly added test in that PR. Specifically, because it called `upsample_bilinear` with f64 data-type. Since the bug was only on non-TPU and non-Neuron device types, we simply skip the test for those. It's safe to do that because there's a short-circuit that avoids the bug for those devices.

cc @miladm @JackCaoG

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2118,6 +2118,54 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
     self.assertEqual(out, Xout.cpu())
     self.assertEqual("f16", torch_xla._XLAC._get_xla_tensor_shape_type(Xout))
 
+  def test_upsample_bilinear_double(self):
+    # Originally, the upsample_bilinear implementation (in resize_ops.cpp)
+    # was copied from TF. The computation was done intentionally on F32 and
+    # not cast back[1]. However, that didn't reflect in the returned tensor.
+    # Basically, what would happen is:
+    #
+    # 1. A tensor of data-type other than F32 is created:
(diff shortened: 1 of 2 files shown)


ysiraichi:
I thought `tpuci` enabled TPU CI. Doesn't look like it.
@JackCaoG Is there any way to enable TPU CI?

JackCaoG:
haha, you need to first set the `tpu-ci` tag and then rerun the CI. This way TPU ci will get rerun, I just trigger a rerun

ysiraichi:
Hmmm. Let me try rebasing the PR.

ysiraichi:
@JackCaoG Could you take a look at this PR whenever you have some time?

test/test_operations.py
```diff
@@ -2118,6 +2118,59 @@ def fn(inp, s):
     self.assertEqual(out, Xout.cpu())
     self.assertEqual("f16", torch_xla._XLAC._get_xla_tensor_shape_type(Xout))
 
+  # We skip TPU for 2 reasons:
+  #   1. upsample_bilinear on f64 tensors doesn't work on TPUs
```
JackCaoG, line 2122: will f64 works prior to your change on TPU? I don't want this to become a regression.

ysiraichi: My change doesn't affect TPU. The lowering for `upsample_bilinear` is different for CUDA and TPU. This change only affects non-TPU and non-Neuron devices.

JackCaoG (approved):

PR #7173, opened 2024-06-03
Integrate dlpack to dynamo.

vanbasten23:
This PR integrate the DLPack API to dynamo so that when we move a tensor between CUDA and XLA we don't have to go through CPU anymore.

Test:
PJRT_DEVICE=CUDA python pytorch/xla/test/dynamo/test_dynamo.py -k test_simple_model_automoves_tensors


torch_xla/core/dynamo_bridge.py
```diff
@@ -126,18 +128,36 @@ def _args_on_cuda(input_args: tuple) -> bool:
...
+  assert stream is None or type(stream) is int
+  external_stream = torch.cuda.ExternalStream(stream)
+  current_stream = torch.cuda.current_stream()
+  if external_stream != current_stream:
+    event = torch.cuda.Event()
+    event.record(current_stream)
+    external_stream.wait_event(event)
+  dlpack = torch_xla_dlpack.to_dlpack(tensor)
+  cuda_tensor = torch.utils.dlpack.from_dlpack(dlpack)
+  return cuda_tensor
```
ysiraichi, line 151: I feel like this would be useful outside the dynamo bridge, too. What do you think about moving this to `torch_xla.dlpack.as_eager()` or something like this?

torch_xla/core/dynamo_bridge.py
```diff
@@ -126,18 +128,36 @@ def _args_on_cuda(input_args: tuple) -> bool:
...
 
 # Given an input list, moves the tensors to the given target_device.
 # The output order will be the same as the input. Non tensors will also still
 # be in the list.
 def _maybe_move_tensors_to_device(tensors: tuple,
                                   target_device: torch.device) -> tuple:
-  if not torch.cuda.is_available():
-    return tensors
```
ysiraichi, line 136: Why are we getting rid of this?

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -709,7 +709,8 @@ PjRtComputationClient::ExecuteComputation(
     const PjRtData* pjrt_data = dynamic_cast<PjRtData*>(argument.get());
 
     XLA_CHECK(pjrt_device == pjrt_data->buffer->device())
-        << pjrt_device->DebugString() << " vs "
+        << "The device currently being used : " << pjrt_device->DebugString()
+        << " is different from the device where the buffer resides: "
```
ysiraichi, line 713: Cool, better error message!

ysiraichi (commented):
Overall, LGTM.
I do think we need better testing for this use case, though. What do you think about running all dynamo tests with this flag set?

torch_xla/core/dynamo_bridge.py
vanbasten23: For moving XLA tensors to CUDA tensors, if torch.cuda is NOT available, it should fail rather than returning XLA tensors.
Similar arguments applies to the other direction.

If we only support the conversion between CUDA and XLA, I think an assert will be more appropriate. Wdyt?

torch_xla/core/dynamo_bridge.py
vanbasten23: Done.
OTOH, the logic here borrows from https://github.com/pytorch/pytorch/blob/b0ef363972203b163cddc95e4c6054b8221c2300/torch/utils/dlpack.py#L114-L115. So if https://github.com/pytorch/pytorch/pull/128176 gets merged, we don't need this logic here anymore.

vanbasten23:
> Overall, LGTM. I do think we need better testing for this use case, though. What do you think about running all dynamo tests with this flag set?

Currently it works for inference but not for training. If we run all dynamo test, then we need to change all the test. OTOH, all this PR does is to move CUDA tensor to the XLA device at the beginning of the dynamo bridge, all the rest should remain the same. So do we really need to run all dynamo tests with the flag?

test/dynamo/test_dynamo.py
```diff
@@ -154,14 +156,22 @@ def test_simple_model(self):
...
+  @parameterized.parameters(
+      "0",
+      "1",
+  )
+  def test_simple_model_automoves_tensors(self, zero_copy_enabled):
+    os.environ.update({
+        xenv.ZERO_COPY_ENABLED: zero_copy_enabled,
+    })
+    x = torch.tensor(100.0).to(device="cuda:0")
+    y = torch.tensor(200.0).to(device="cuda:0")
```
JackCaoG, line 168: shouldn't you check that output is also on `cuda:0`?

JackCaoG: and somehow verified that computation is run using dynamo not fallback so something

torch_xla/core/dynamo_bridge.py
```diff
@@ -147,14 +146,19 @@ def _maybe_move_tensors_to_device(tensors: tuple,
...
 
-    assert target_device is not None, "Moving tensors to None device not supported"
-
     if dynamo_debug:
       print("Moving Tensor {} to device {}".format(tensor, target_device))
 
-    # Have to move to CPU before moving it to target device.
-    moved_tensor = tensor.to(cpu_device)
-    moved_tensor = moved_tensor.to(target_device)
+    zero_copy_enabled = xu.getenv_as(xenv.ZERO_COPY_ENABLED, bool, defval=False)
```
JackCaoG, line 152: is there a reason this has to be a env var? Biggest reason we use env var is we need to communcate something between python and C++ layers, so it is the master process need to pass some information(like rank) to the child process. In your case this is really just a config somewhere and should not be set as env var.

JackCaoG: for you case I think you can just always use dlpack, is there any reason we don't want to use dlpack to convert XLA:GPU and cuad?

test/dynamo/test_dynamo.py
vanbasten23: > shouldn't you check that output is also on cuda:0?

The test already checks that https://github.com/pytorch/xla/blob/3f54fa21613419680b7c3478ac2320a6612ffa62/test/dynamo/test_dynamo.py#L169

> somehow verified that computation is run using dynamo not fallback so something

The test checks tracing is skipped in following runs https://github.com/pytorch/xla/blob/3f54fa21613419680b7c3478ac2320a6612ffa62/test/dynamo/test_dynamo.py#L172-L175. Do you think it's enough?

torch_xla/core/dynamo_bridge.py
vanbasten23: The flag is only temporary. It's use to do a/b test: how much performance when we move tensor through CPU vs how much performance we can get by using dlpack.

So the temporary  be removed later once the a/b testing is done.

test/dynamo/test_dynamo.py
ysiraichi: Since we are looking for fallbacks, what about using `torch_xla._XLAC._get_executed_fallback_ops`?

ysiraichi:
> Currently it works for inference but not for training. If we run all dynamo test, then we need to change all the test. OTOH, all this PR does is to move CUDA tensor to the XLA device at the beginning of the dynamo bridge, all the rest should remain the same. So do we really need to run all dynamo tests with the flag?

In this case, we could have a decorator or something like this for selecting a few tests to run with zero-copy. For example, there is `DynamoInferenceBasicTest` at _test/dynamo/test_dynamo.py_.

While this is a small change (quantity), it may end up changing the execution behavior in ways that we might not think of. I think it would make this PR more robust.

vanbasten23:
> > Currently it works for inference but not for training. If we run all dynamo test, then we need to change all the test. OTOH, all this PR does is to move CUDA tensor to the XLA device at the beginning of the dynamo bridge, all the rest should remain the same. So do we really need to run all dynamo tests with the flag?
> 
> In this case, we could have a decorator or something like this for selecting a few tests to run with zero-copy. For example, there is `DynamoInferenceBasicTest` at _test/dynamo/test_dynamo.py_.

Sure, I've added test for `DynamoInferenceBasicTest`

test/dynamo/test_dynamo.py
vanbasten23: SG. I added a check `self.assertEqual(torch_xla._XLAC._get_executed_fallback_ops(), [])`

test/dynamo/test_dynamo.py
```diff
@@ -218,72 +249,92 @@ def forward(self, index, copy_tensor, input_tensor, op_name):
...
+  @parameterized.parameters(
+      True,
+      False,
+  )
+  def test_simple_model_with_different_input_shape(self, initialize_on_cuda):
+    met.clear_all()
+    device = self._choose_proper_device(initialize_on_cuda)
+    # We need to make `dim` depend on `initialize_on_cuda` because the compilation cache
+    # does not clean itself between the parameterized tests.
+    dim = 5 + int(initialize_on_cuda)
```
ysiraichi, line 309: Do you mean (i) dynamo's cache or (ii) XLA cache? 

- If it's (i), we can just reset it
- If it's (ii), can't we just  reset it, too? If not, I would argue that we don't need to worry about it, since that's not what's being tested, here

What do you think?

torch_xla/core/dynamo_bridge.py
ysiraichi: As far as I can see, I don't think this function is called unless we are using CUDA tensors (i.e. `torch.cuda.is_available() == True`). So, I guess this wasn't needed in the first place.

torch_xla/utils/dlpack.py
```diff
@@ -21,3 +23,28 @@ def from_dlpack(ext_tensor: Any):
...
+  is_xla_cuda = True if xu.getenv_as("PJRT_DEVICE", str,
+                                     "").lower() == "cuda" else False
+  assert is_xla_cuda, "The XLA tensor is not on CUDA"
+  # consumer is torch, producer is torch_xla
+
+  # Similar logic as torch.utils.dlpack.from_dlpack
+  # https://github.com/pytorch/pytorch/blob/b0ef363972203b163cddc95e4c6054b8221c2300/torch/utils/dlpack.py#L114-L115
+  device_id = tensor.device.index
+  stream = torch_xla._XLAC._get_stream_for_cuda_device(device_id)
+  stream = 1 if stream == 0 else stream
```
ysiraichi, line 40: Could you explain what's going on, here? A comment would be nice.

ysiraichi (approved):
LGTM.

I left a few minor comments.
Thank you for taking your time to adapt existing dynamo tests. They look great!

test/dynamo/test_dynamo.py
vanbasten23: It's (ii) and I don't think there exists a way to reset the XLA compilation cache afaik. 

> I would argue that we don't need to worry about it, since that's not what's being tested, here

Actually, the existing test https://github.com/pytorch/xla/blob/ea2a6f774299abea15450670049429655829784c/test/dynamo/test_dynamo.py#L276 tests the compilation cache. Without the change, the test will fail because of the reason in the comment. That's why I have to change it..

torch_xla/core/dynamo_bridge.py
vanbasten23: Exactly.

torch_xla/utils/dlpack.py
vanbasten23: Sure. Per https://github.com/pytorch/pytorch/blob/b0ef363972203b163cddc95e4c6054b8221c2300/torch/utils/dlpack.py#L114-L115, "The array API specify that the default legacy stream must be passed with a value of 1 for CUDA". I added a comment.

test/dynamo/test_dynamo.py
ysiraichi: Ah, I see. I guess that works. You could also only run this check in one of the runs (e.g. if `initialize_on_cuda == True`). But, I think this is also ok.

vanbasten23:
Thanks for the review!

Issue #7169, opened 2024-06-03
Persistent Cache will not recompile when `XLA_IR_DEBUG` and `XLA_HLO_DEBUG` changed


JackCaoG:
## 🐛 Bug

When the persistent cache is enabled and `XLA_IR_DEBUG` and `XLA_HLO_DEBUG` env var changed, PyTorch/XLA will not recompile a graph with the HLO metadata

## To Reproduce

run the same pytorch/xla problem with persistent cache but flip the `XLA_IR_DEBUG` and `XLA_HLO_DEBUG`.

## Expected behavior

We recompile
## Environment

 - Reproducible on XLA backend: all
 - torch_xla version: nightly

cc @jonb377

JackCaoG:
This is not necessary a problem for non-persisent cache case but a bit annoying with persistent cahce. I found out this by using the cache without `XLA_HLO_DEBUG` first and then found that I can't profiler to work with the hints.

JackCaoG:
I think the easy solution is to let our compilation hash to also include `XLA_IR_DEBUG` and `XLA_HLO_DEBUG`. However there is still a catch, we do not cache on the content on the HLO metadata. This means if we use persistent cache and change the tracing annotation we still won't recompile.

That being said I don't think we should cache the HLO metadata, it will make tracing slower.

jonb377:
Hmm we can add `XLA_IR_DEBUG` and `XLA_HLO_DEBUG` to this list to change the hash: https://github.com/pytorch/xla/blob/7938bb5da6c993609aff614ccfa5b722a339d158/torch_xla/csrc/runtime/env_hash.cc#L96

Do these env vars actually change the compilation result though?

JackCaoG:
Computation wise it's the same. It only affect the metadata we see during profiling.

jonb377:
Sure, I can open a PR for that. I'm still concerned about your second point:

> This means if we use persistent cache and change the tracing annotation we still won't recompile.

Should we instead just disable the persistent cache if these env vars are set?

JackCaoG:
throwing some kind of warning message should be enough.

PR #7175, opened 2024-06-03
Warn when using persistent cache with debug env vars

jonb377:
See also https://github.com/pytorch/xla/issues/7169

When using the persistent compilation cache, modifications to the HLO metadata will not impact the graph hash, and the persistent cache will load executables compiled with potentially outdated metadata.

This change:
- Modifies the graph hash when `XLA_HLO_DEBUG` or `XLA_IR_DEBUG` is set
- Prints a warning when initializing the persistent cache with `XLA_HLO_DEBUG` or `XLA_IR_DEBUG` set.

### DIFF
--- a/torch_xla/csrc/runtime/env_hash.cc
+++ b/torch_xla/csrc/runtime/env_hash.cc
@@ -93,7 +93,7 @@ torch::lazy::hash_t HashXlaEnvVars() {
   // Both XLA_FLAGS and LIBTPU_INIT_ARGS contain XLA flags which impact
   // the compilation result.
   static std::vector<std::string> flag_vars = {"XLA_FLAGS", "LIBTPU_INIT_ARGS"};
-  static std::vector<std::string> raw_vars = {"TPU_MEGACORE"};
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/runtime/env_hash.cc
```diff
@@ -93,7 +93,7 @@ torch::lazy::hash_t HashXlaEnvVars() {
   // Both XLA_FLAGS and LIBTPU_INIT_ARGS contain XLA flags which impact
   // the compilation result.
   static std::vector<std::string> flag_vars = {"XLA_FLAGS", "LIBTPU_INIT_ARGS"};
-  static std::vector<std::string> raw_vars = {"TPU_MEGACORE"};
+  static std::vector<std::string> raw_vars = {"TPU_MEGACORE", "XLA_HLO_DEBUG"};
```
JackCaoG, line 96: can we also include `XLA_IR_DEBUG`? This flag will add additional meta data to the HLO.

jonb377: Ah I see, I didn't think `XLA_IR_DEBUG` would touch the HLO, I'll include both env vars here.

JackCaoG (approved):

PR #7193, opened 2024-06-04
Add options in benchmark script to keep model on CUDA

vanbasten23:
- **add how to run benchmark script for a single config without spawning processes.**
- **Add options in benchmark script to keep model on CUDA.**

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -74,6 +74,14 @@ among the flags `--dynamo`, `--xla`, and `--test`, 4 of which are supported:
   - `dynamo=inductor`, `xla=None`, `test=train`
 
 
+## Run benchmarking for a single configuration
+
+The section `Experiment runner` above shows how to run the benchmarking script for a combination of configurations. For each configuration,
+the script starts a process and run the benchmarking. This section shows how to run the benchmarking for a single configuration without spawning new processes.
+
+
+
+
 ## Verification module
 
 Verification flag, enabled by running the experiment runner script with `--verify`
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -79,7 +79,16 @@ among the flags `--dynamo`, `--xla`, and `--test`, 4 of which are supported:
 The section `Experiment runner` above shows how to run the benchmarking script for a combination of configurations. For each configuration,
 the script starts a process and run the benchmarking. This section shows how to run the benchmarking for a single configuration without spawning new processes.
 
-
+```
+cd pytorch
+python xla/benchmarks/experiment_runner.py \
+    --suite-name=torchbench \
+    --accelerator=cuda \
+    --progress-bar  \
+    --model-config=\{\"model_name\":\"BERT_pytorch\"\} \
+    --experiment-config=\{\"accelerator\":\"cuda\",\"xla\":\"PJRT\",\"xla_flags\":null,\"dynamo\":\"openxla\",\"torch_xla2\":null,\"test\":\"train\"\} \
+    --repeat 1
+```
 
 
 ## Verification module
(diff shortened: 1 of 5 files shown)


benchmarks/experiment_runner.py
```diff
@@ -942,6 +947,7 @@ def main():
 
   logging.basicConfig(level=args.log_level.value, force=True)
   logger.debug(f"Parsed args: {args}")
+  # args.xla_take_cuda_model_and_data available
```
zpcore, line 950: nit: remove.

zpcore:
You probably want to add the `xla_take_cuda_model_and_data` attribute into the result_analyzer.py. Without it, the test script may fail. Also this may mess up with the performance result.

benchmarks/README.md
```diff
@@ -74,6 +74,23 @@ among the flags `--dynamo`, `--xla`, and `--test`, 4 of which are supported:
...
 
 
+## Run benchmarking for a single configuration
+
+The section `Experiment runner` above shows how to run the benchmarking script for a combination of configurations. For each configuration,
+the script starts a process and run the benchmarking. This section shows how to run the benchmarking for a single configuration without spawning new processes.
+
+```
+cd pytorch
+python xla/benchmarks/experiment_runner.py \
```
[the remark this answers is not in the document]
vanbasten23: Much better. Thanks!

benchmarks/torchbench_model.py
```diff
@@ -255,7 +255,8 @@ def set_up(self):
     if self.benchmark_experiment.xla:
       # First, move the model and the inputs to CPU.
       # This avoids having dupplicated data on CUDA.
-      if self.is_accelerator_cuda():
+      keep_model_data_on_cuda = self.benchmark_experiment.xla_take_cuda_model_and_data
+      if self.is_accelerator_cuda() and not keep_model_data_on_cuda:
```
ysiraichi, line 259: Could you leave a comment here, explaining why we need this flag?

ysiraichi (commented):
Overall, LGTM. I have 2 extra comments, though:

1. The argument is called `xla_take_cuda_model_and_data`, but you are using it as `keep_model_data_on_cuda`. Maybe, it's better to stick to one of them to avoid confusion
2. Could you write somewhere the reasoning behind this flag? For example: (i) what does it do to the benchmark execution? and (ii) why are we introducing this flag?

benchmarks/torchbench_model.py
vanbasten23: Done, added it in the help message

zpcore:
Can you do a try of the following with the PR:
```
python xla/benchmarks/experiment_runner.py ...
python xla/benchmarks/result_analyzer.py --output-format=csv
python xla/benchmarks/result_analyzer.py --output-format=jsonl
```
Just in case they need the `keep_model_data_on_cuda` key  in https://github.com/pytorch/xla/blob/cdb4aec6f6d168024c0ac25996652d1ed2ab6417/benchmarks/result_analyzer.py#L49 and https://github.com/pytorch/xla/blob/cdb4aec6f6d168024c0ac25996652d1ed2ab6417/benchmarks/result_analyzer.py#L173.

vanbasten23:
> Can you do a try of the following with the PR:
> 
> ```
> python xla/benchmarks/experiment_runner.py ...
> python xla/benchmarks/result_analyzer.py --output-format=csv
> python xla/benchmarks/result_analyzer.py --output-format=jsonl
> ```
> 
> Just in case they need the `keep_model_data_on_cuda` key in
> 
> https://github.com/pytorch/xla/blob/cdb4aec6f6d168024c0ac25996652d1ed2ab6417/benchmarks/result_analyzer.py#L49
> 
> and
> https://github.com/pytorch/xla/blob/cdb4aec6f6d168024c0ac25996652d1ed2ab6417/benchmarks/result_analyzer.py#L173
> 
> .

I see. Good point. I've fixed it and make sure it generates meaningful results.

zpcore (approved):
Thanks, LGTM! Let's wait for the CI.

vanbasten23:
Thanks for the review Pei and Yukio!

PR #7200, opened 2024-06-05
openxla pin update to 06/05

zpcore:
NIT: Also relocate the commented `local_repository` right after http_archive or it will report xla not found.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -50,7 +50,7 @@ new_local_repository(
 #    curl -L https://github.com/openxla/xla/archive/<git hash>.tar.gz | sha256sum
 #    and update the sha256 with the result.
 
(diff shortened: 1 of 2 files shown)


ManfeiBai (approved):
Thanks, LGTM, please merge after all CI green

WORKSPACE
```diff
@@ -70,6 +70,19 @@ http_archive(
     ],
 )
 
+# For development, one often wants to make changes to the OpenXLA repository as well
```
JackCaoG, line 73: why is this moved?

zpcore: local_repository needs to be stated before `load("@xla//third_party/py:python_init_rules.bzl", "python_init_rules")...` etc. Those are newly added from hermetic python.

PR #7201, opened 2024-06-05
Add int8 per channel weight-only quantized matmul

lsy323:
Add the first xla quantized ops for per-channel weight-only quantized matmul. 

The math is `out[bf16] = matmul(act[bf16], weight[s8]) * scaler[bf16]`, the same as what was adopted in the XLA llama [quant implementation](https://github.com/pytorch-tpu/llama/blob/llama2-google-next-inference/llama/xla_model_parallel.py#L440-L442)

User experience:
- Call quantized op with already quantized weight in model code
- Swap the `nn.Linear` Module with the added quantized module in model code

More details about user experience can be found in the added README.

Changes:
- Added custom torch op and `nn.Module` for the quantized op
- Added user guide

Test:
- Test the lowered HLO is doing what we expect
- Test it works with Dynamo
- Test numerical correctness

int4 and blockwise quant support will be added in following PRs

### DIFF
--- a/docs/quantized_ops.md
+++ b/docs/quantized_ops.md
@@ -1,10 +1,9 @@
 Quantized Operations for XLA device (Experimental feature)
 --------------------------
 
-This document describes how to enable quantization on XLA device by the quantized operations.
+This document outlines how to utilize quantized operations to enable quantization on XLA devices.
 
-XLA Quantized ops provide a high level abstraction on the quantized operations. (e.g. blockwise int4 quantized matmul) 
-The XLA quantized ops are equiavlent to the quantized CUDA kernels in the CUDA world ([example](https://github.com/vllm-project/vllm/blob/main/csrc/quantization/gptq/q_gemm.cu)).
+XLA Quantized ops offer a high-level abstraction for quantized operations (e.g., blockwise int4 quantized matrix multiplication). These ops are analogous to quantized CUDA kernels ([example](https://github.com/vllm-project/vllm/blob/main/csrc/quantization/gptq/q_gemm.cu)) in the CUDA ecosystem, providing similar functionality and performance benefits within the XLA framework.
 
 **NOTE:** Currently this is classified as experimental feature. It's API specifics 
 will change in the next (2.5) release.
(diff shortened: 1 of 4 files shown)


docs/quantized_ops.md
```diff
@@ -0,0 +1,114 @@
...
+import torch_xla.experimental.xla_quantized_matmul
+
+N_INPUT_FEATURES=10
+N_OUTPUT_FEATURES=20
+x = torch.randn((3, N_INPUT_FEATURES), dtype=torch.bfloat16)
+w_int = torch.randint(-128, 127, (N_OUTPUT_FEATURES, N_INPUT_FEATURES), dtype=torch.int8)
+scaler = torch.randn((N_OUTPUT_FEATURES,), dtype=torch.bfloat16)
+
+# Call with torch CPU tensor
+matmul_output = torch.ops.xla.quantized_matmul(x, w_int, scaler)
```
JackCaoG, line 36: why do we want to show the cpu tensor part? Do we ever expect user to use it with CPU tensor?

docs/quantized_ops.md
```diff
@@ -0,0 +1,114 @@
...
+### Module Swap
+
+Alternatively, users can also use the `nn.Module` that wraps the XLA quantized ops and do module swap in the model code:
+
+```Python
+orig_model = MyModel()
+# Quantize the model and get quantized weights
+q_weights = quantize(orig_model)
+# Process the quantized weight to the format that XLA quantized op expects.
+q_weights_for_xla = process_for_xla(q_weights)
```
JackCaoG, line 87: seems to suggest user has to define their own `quantize` and `process_for_xla`?

torch_xla/experimental/xla_quantized_matmul.py
```diff
@@ -0,0 +1,89 @@
+import torch
+import torch.nn.functional as F
+import torch_xla
+from torch.library import impl
+from torch_xla.core.xla_model import XLA_LIB
+
```
JackCaoG, line 6: do you expect to add mores quantized ops? I recommend call this file `xla_quantized_ops` if you plan to add more ops. Having one file for every custom ops we add is not a good idea.

JackCaoG (commented):
mostly lgtm, minor nits

docs/quantized_ops.md
lsy323: Yes, the is the ux for run the op in eager mode, it makes debugging easier I think

docs/quantized_ops.md
lsy323: Yes, those 2 parts are out of the scope of quantize ops

torch_xla/experimental/xla_quantized_matmul.py
lsy323: I think only putting matmul related codes in this file is cleaner. For embedding, we can have another file like `xla_quantized_embedding.py`

JackCaoG (approved):

docs/quantized_ops.md
JackCaoG: please update the comment to say something like `this is mostly for debug`

torch_xla/experimental/xla_quantized_matmul.py
lsy323: That makes sense, the only concern is that the file will become huge. To have user just import 1 file for all ops, we can have a `quantized_ops.py` and import all of our quantized op files there.

docs/quantized_ops.md
lsy323: Done

JackCaoG (approved):

PR #7204, opened 2024-06-05
Allow user to configure distributed runtime service.

vanbasten23:
On GKE, the multi-host GPU training always fails after 5 min into the training, without a clear error message (see the log on the 2 hosts: https://gist.github.com/vanbasten23/bdced1d993fa604025d318e9d68a2f6b https://gist.github.com/vanbasten23/cc4096a869e0278eb414642dd7662166). It fails on larger model such as LLM but succeeds on smaller model such as resnet.

By following JAX's [example](http://google3/learning/brain/research/jax/lib/jax_service_python.cc;l=135-145;rcl=617156835), we allow user to configure max_missing_heartbeats, heartbeat_interval, distributed_shutdown_timeout. I suspect the failure is due to one of the configuration's default value being too small.

### DIFF
--- a/torch_xla/csrc/runtime/BUILD
+++ b/torch_xla/csrc/runtime/BUILD
@@ -237,6 +237,7 @@ cc_library(
     deps = [
         ":debug_macros",
         ":sys_util",
+        ":env_vars",
         "@xla//xla/tsl/distributed_runtime/preemption:preemption_sync_manager",
         "@xla//xla/pjrt/distributed",
     ],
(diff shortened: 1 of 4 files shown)


jonb377 (approved):
LGTM!

From the gist, it looks like the failure happens just after compilation. I wonder if the dist service is missing during compilation given the longer compile time of LLMs.

will-cromar (approved):

vanbasten23:
> From the gist, it looks like the failure happens just after compilation. 

How do you know the compilation has finished?

> I wonder if the dist service is missing heartbeats during compilation given the longer compile time of LLMs.

Yeah, it's also my guess on why it fails. I'd need this PR to verify the hypothesis.

jonb377:
> How do you know the compilation has finished?

`22%|â–ˆâ–ˆâ–       | 2/9 [04:35<18:49, 161.33s/it]` It spent >161s in iteration 1, so at least iteration 1 finished compiling. There will be a second compilation at iteration 2 due to the optimizer state though, so that compilation may still be ongoing when the error is hit.

vanbasten23:
Thanks Jon and Will for the review!

PR #7205, opened 2024-06-05
Make clip_grad_norm_ follow the input's dtype

alanwaketan:
Summary:
The correct clipped value uses fp32 regardless of the input dtype. Let's make it follow the input dtype.

Test Plan:
python test/test_operations.py -v -k test_clip_grad_norm_

### DIFF
--- a/torch_xla/_patched_functions.py
+++ b/torch_xla/_patched_functions.py
@@ -31,9 +31,10 @@ def clip_grad_norm_(parameters: _tensor_or_tensors,
   parameters = list(filter(lambda p: p.grad is not None, parameters))
   max_norm = float(max_norm)
   norm_type = float(norm_type)
-  if len(parameters) == 0:
-    return torch.tensor(0.)
+  dtype = parameters[0].grad.dtype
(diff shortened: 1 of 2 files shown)


torch_xla/_patched_functions.py
```diff
@@ -31,9 +31,10 @@ def clip_grad_norm_(parameters: _tensor_or_tensors,
   parameters = list(filter(lambda p: p.grad is not None, parameters))
   max_norm = float(max_norm)
   norm_type = float(norm_type)
-  if len(parameters) == 0:
-    return torch.tensor(0.)
+  dtype = parameters[0].grad.dtype
```
JackCaoG, line 34: oh man.... I totally forgot we have this patch.

JackCaoG (approved):

torch_xla/_patched_functions.py
alanwaketan: Not sure if it actually is needed though... haha

alanwaketan:
Thanks, Jack!

Issue #6242, opened 2023-12-28
Improve PJRT C API support for GPUs and custom hardware


will-cromar:
## 🚀 Feature

- Provide a consistent API for external packages to register available [PJRT C API Plugins](https://github.com/openxla/xla/blob/main/xla/pjrt/c/docs/pjrt_integration_guide.md). 
- TPU and GPU should use the same plugin API. 

## Motivation

At the time of writing, PyTorch/XLA supports three PJRT Plugins: TPU, Intel XPU, and AWS Neuron. Each of these plugins required updates to PyTorch/XLA itself, and any future plugins will require the same. Hardware vendors should not have to make changes in this repository just to register their devices.

### High-level API for configuration and initialization

A working PJRT C API implementation alone is not enough to cleanly integrate with PyTorch/XLA. We will need to provide at least the following additional functionality in addition to what is available through runtime APIs:

- Plugin location: The low-level ComputationClient will also need to know where to find the PJRT Plugin binary, either by looking in locations known by convention (e.g. bundled libtpu binaries) or by calling the plugin package (e.g. `libtpu.get_library_path()`). 
- `local_rank` and `local_world_size`: the concept of "local" devices exists in PyTorch, but not in PJRT.
- `host_index`: PJRT only has a concept of "process", not host.
- Multi-process initialization: to assign each device to a separate process, PyTorch/XLA performs additional setup. Note: this assignment must be done _before_ the PJRT client is created.

This functionality can be bundled under a common interface, implemented for each plugin and registered with PyTorch/XLA before client initialization. #5644 contains a (likely incomplete) `DevicePlugin` interface and an example implementation for TPU.

### Hardcoded device types

Currently, the available device types have to be hardcoded across PyTorch/XLA, meaning for every new device, vendors must send a PR to us. There are lists of devices in the following locations:

- [`XlaDeviceType` enumeration](https://github.com/pytorch/xla/blob/efa6fcfdac5368330a0770e9019649eba08b5f56/torch_xla/csrc/device.h#L18)
- [Conversion from `XlaDeviceType` to string](https://github.com/pytorch/xla/blob/efa6fcfdac5368330a0770e9019649eba08b5f56/torch_xla/csrc/device.cpp#L14-L34)
- [Parsing device type from string generated above](https://github.com/pytorch/xla/blob/efa6fcfdac5368330a0770e9019649eba08b5f56/torch_xla/core/xla_model.py#L75) in Python

[...]

PR #7209, opened 2024-06-06
Add stub plugins for XPU and Neuron

will-cromar:
We're currently planning to change the default behavior to look for installed plugin packages instead of hardcoding each device type in `pjrt_registry.cc`. This PR adds stub implementations for XPU and Neuron that match the current behavior in the `torch_xla` package.

Long-term, these plugins should move to separate packages similar to `plugins/cuda`. I'll add some more documentation soon.

See #6242 for context.

cc @aws-kingrj

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -302,7 +302,11 @@ setup(
         'console_scripts': [
             'stablehlo-to-saved-model = torch_xla.tf_saved_model_integration:main'
         ],
-        'torch_xla.plugins': ['tpu = torch_xla._internal.tpu:TpuPlugin',],
+        'torch_xla.plugins': [
+            'tpu = torch_xla._internal.tpu:TpuPlugin',
+            'neuron = torch_xla._internal.neuron:NeuronPlugin',
+            'xpu = torch_xla._internal.xpu:XpuPlugin'
+        ],
     },
     extras_require={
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -302,7 +302,11 @@ def run(self):
         'console_scripts': [
             'stablehlo-to-saved-model = torch_xla.tf_saved_model_integration:main'
         ],
-        'torch_xla.plugins': ['tpu = torch_xla._internal.tpu:TpuPlugin',],
+        'torch_xla.plugins': [
+            'tpu = torch_xla._internal.tpu:TpuPlugin',
+            'neuron = torch_xla._internal.neuron:NeuronPlugin',
```
JackCaoG, line 307: so for every plug in we still need to modify the pytorch/xla to make them discoverable?

JackCaoG (approved):

setup.py
will-cromar: No, these are just defined in our `setup.py` because the main package contains the plugin implementations. Any package can add more entry points separately, e.g.

https://github.com/pytorch/xla/blob/de283c986f8514b538037b192f64be7fa744592e/plugins/cuda/pyproject.toml#L17-L18

PR #7211, opened 2024-06-06
Updating ml_framework_name and version to pytorch/xla2 in jax_configs. 

aman2930:
These updates will be used to create the PjRt_client. Also, enabling the telemetry by setting ENABLE_RUNTIME_UPTIME_TELEMETRY=1.

### DIFF
--- a/experimental/torch_xla2/torch_xla2/__init__.py
+++ b/experimental/torch_xla2/torch_xla2/__init__.py
@@ -1,4 +1,6 @@
 import jax
+import jax._src import config
+import os
 import torch
 from torch._functorch import make_functional
 from torch.utils import _pytree as pytree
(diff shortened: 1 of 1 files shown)


experimental/torch_xla2/torch_xla2/__init__.py
```diff
@@ -1,14 +1,24 @@
...
+from jax._src import config
+import os
 import torch
 from torch._functorch import make_functional
 from torch.utils import _pytree as pytree
 from torch_xla2 import export, tensor, tf_integration
 
 jax.config.update('jax_enable_x64', True)
 
+config.update(
```
qihqi, line 11: qq. just jax.config.update won't work? must use config from _src?

qihqi: btw if you want to land now you can add a `if jax.version >= 0.4.29'` so it will work when jax releases on Monday.

qihqi (approved):

experimental/torch_xla2/torch_xla2/__init__.py
aman2930: > btw if you want to land now you can add a `if jax.version >= 0.4.29'` so it will work when jax releases on Monday.

So does it mean that if I don't add `if jax.version >= 0.4.29'`, it won't work even if jax releases on Monday?

PR #7220, opened 2024-06-07
Sample training UX

qihqi:
Add examples to show case llama training.

### DIFF
--- a/examples/train_decoder_only_base2.py
+++ b/examples/train_decoder_only_base2.py
@@ -1,0 +1,107 @@
+from decoder_only_model import DecoderOnlyConfig, DecoderOnlyModel
+
+
+import time
+import itertools
+
+import torch
+import torch_xla.utils.utils as xu
+import torch.optim as optim
+import torch.nn as nn
+import torch_xla2
+import torch_xla2.interop
+
+
+USE_FSDPv2 = False
+
+class OptimizerStep:
+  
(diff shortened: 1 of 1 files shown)


examples/train_decoder_only_base2.py
```diff
@@ -0,0 +1,107 @@
...
+
+    self.state = self.optimizer.state_dict()
+    # if USE_FSDPv2:
+    #     self.model = FSDPv2(model)
+    #     self.state = FSDPv2.shard_state(self.state)
+
+    def single_step(data, target, state):
+        self.optimizer.load_state_dict(state)
+        self.optimizer.zero_grad()
+        res = torch.func.functional_call(self.model, state, data)
```
JackCaoG, line 33: is this state the model parameter? Where is it coming from?

examples/train_decoder_only_base2.py
```diff
@@ -0,0 +1,107 @@
...
+        loss = self.loss_fn(res, target)
+        loss.backward()
+        self.optimizer.step()
+        return loss, self.optimizer.state_dict()
+
+    self.single_step = torch_xla2.interop.jax_jit(single_step)
+
+  def run_optimizer(self, data, target):
+    loss, self.state = self.single_step(data, target, self.state)
```
JackCaoG, line 42: I understand why we need a `self.state` now... but I still felt like it is confusing to normal user.

PR #7221, opened 2024-06-07
Modularize zero optimizer step function

karakusc:
Refactor and (slightly) generalize `step` function of the zero redundancy optimizer, by breaking it into the following three high-level steps:
1. `_reduce_gradients`
2. `_clip_grad_norm`
3. `_update_parameters`

This makes it easier to sub-class and override specific behaviors of the zero optimizer in different contexts. 

In addition, we introduce an additional optional parameter `sharding_scheme`, which allows us to customize steps 1 and 3 above, if needed.

### DIFF
--- a/torch_xla/distributed/zero_redundancy_optimizer.py
+++ b/torch_xla/distributed/zero_redundancy_optimizer.py
@@ -344,12 +344,12 @@ class ZeroRedundancyOptimizer(Optimizer):
       return kwargs["sharding_scheme"]
     else:
       return [
-               {
-                 "scale_factor": 1.0,
-                 "sharding_group": self.sharding_groups,
-                 "group_size": self.local_world_size,
-               },
-             ]
+          {
+              "scale_factor": 1.0,
+              "sharding_group": self.sharding_groups,
+              "group_size": self.local_world_size,
+          },
+      ]
 
   def _reduce_gradients(self, **kwargs):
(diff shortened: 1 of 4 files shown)


torch_xla/distributed/zero_redundancy_optimizer.py
```diff
@@ -411,25 +440,31 @@ def allgather_weights_and_update_full_parameter(self):
         if param.grad is not None or (self.use_grad_acc_hook and
                                       hasattr(shard, 'main_grad')):
           shard_data = shard.data
-          shard_data = shard_data.to(dtype=param.dtype)
+          if not self.higher_cc_precision:
```
hgt312, line 443: now we always use all gather with dtype==param.dtype, do not change this line

hgt312 (approved):
Overall LGTM

hgt312:
and use `yapf --recursive -i *.py test/ scripts/ torch_xla/ benchmarks/` to format

torch_xla/distributed/zero_redundancy_optimizer.py
karakusc: Done

hgt312 (approved):

jeffhataws (approved):

jeffhataws:
This change matches https://github.com/pytorch/xla/pull/7233 which is merged to mainline.

PR #7225, opened 2024-06-09
Fix wrong env flag for using auto sharding in documentation

lausannel:
Could you please consider updating the flag for enabling autosharding to `XLA_AUTO_SPMD` in the doc? It appears that `XLA_SPMD_AUTO` might be a typo. Thank you!

https://github.com/pytorch/xla/blob/5338a3b0ee7560c137006fe1f976be2c6c864dab/torch_xla/runtime.py#L257-L260

https://github.com/pytorch/xla/blob/5338a3b0ee7560c137006fe1f976be2c6c864dab/torch_xla/csrc/xla_sharding_util.cpp#L893-L898


JackCaoG (approved):
nice catch!

PR #7228, opened 2024-06-10
Use std::optional

cyyever:
Since we are moving to std::optional in PyTorch.


JackCaoG:
Test failure is irrelevant, also see the same failure on head. mering.

JackCaoG (approved):

PR #7231, opened 2024-06-10
[SPMD] Support reduce-scatter in manual sharding

alanwaketan:
Summary:
This PR is to add an experimental suport of cc ops in manual sharding zones. This one adds reduce-scatter as the initial step. The key here is to channel_id, replica_groups, and use_global_device_ids.

Test Plan:
PJRT_DEVICE=TPU XLA_USE_SPMD=1 python test/spmd/test_xla_sharding.py -v -k test_spmd_reduce_scatter

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1557,6 +1557,16 @@ void InitXlaModuleBindings(py::module m) {
           result_tuple[1] = new_token;
           return result_tuple;
         });
+  m.def("_xla_spmd_reduce_scatter",
+        [](const std::string& reduce_type, const at::Tensor& input,
+           double scale, int64_t scatter_dim, int64_t shard_count, const py::list& groups) {
+          std::vector<std::vector<int64_t>> replica_groups =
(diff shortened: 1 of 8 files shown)


alanwaketan:
Why the hell do we run spmd tests on cpu?

JackCaoG:
You can skip your test for CPU. For the most sharding related test they actually passed on CPU. CPU test is the only CI that upstream runs against us.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1557,6 +1557,17 @@ void InitXlaModuleBindings(py::module m) {
           result_tuple[1] = new_token;
           return result_tuple;
         });
+  m.def(
+      "_xla_spmd_reduce_scatter",
```
JackCaoG, line 1561: so the only difference is this one does not have token?

alanwaketan: That's one difference. Others are mentioned in the description.

JackCaoG (approved):

alanwaketan:
The GPU test failure doesn't seem to be related.

PR #7233, opened 2024-06-10
Modularize zero step function and make it customizable

karakusc:
Refactor and (slightly) generalize step function of the zero redundancy optimizer, by breaking it into the following three high-level steps:

1. `_reduce_gradients`
2. `_clip_grad_norm`
3. `_update_parameters`
This makes it easier to sub-class and override specific behaviors of the zero optimizer in different contexts.

In addition, we introduce an additional optional parameter `sharding_scheme`, which allows us to customize steps 1 and 3 above, if needed.

### DIFF
--- a/torch_xla/distributed/zero_redundancy_optimizer.py
+++ b/torch_xla/distributed/zero_redundancy_optimizer.py
@@ -324,17 +324,20 @@ class ZeroRedundancyOptimizer(Optimizer):
         if p.grad is not None:
           p.grad.detach().mul_(clip_value)
 
-  @torch.no_grad()
-  def step(self, closure=None, **kwargs):
-    """
-    Performs a single optimizer step and syncs parameters across all ranks.
-    """
-    assert self.inited, "must call init_zero() first"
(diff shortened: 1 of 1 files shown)


torch_xla/distributed/zero_redundancy_optimizer.py
```diff
@@ -324,17 +324,20 @@ def _clip_grad_norm(
...
+    if "sharding_scheme" in kwargs:
+    ¦ return kwargs["sharding_scheme"]
+    else:
+    ¦ return [
+    ¦   ¦ {
+    ¦   ¦   ¦ "scale_factor": 1.0,
+    ¦   ¦   ¦ "sharding_group": self.sharding_groups,
+    ¦   ¦   ¦ "group_size": self.local_world_size,
+    ¦   ¦ },
+    ¦ ]
```
JackCaoG, line 337: you seems to copy some formatter metadata in the pr.

karakusc: Thanks, removed

jeffhataws (approved):

PR #7237, opened 2024-06-11
Fix __getitem__

qihqi:


experimental/torch_xla2/torch_xla2/ops/jtorch.py
```diff
@@ -163,4 +163,13 @@ def scaled_dot_product_attention(
...
+   return _sdpa_reference(query, key, value, attn_mask, dropout_p, is_causal, scale)
+
+@register_function(torch.Tensor.__getitem__)
+def getitem(self, indexes):
+  if isinstance(indexes, list) and isinstance(indexes[0], int):
+    # list of int, i.e. x[[1, 2]] NOT x[1, 2] (the second would be tuple of int)
+    indexes = (indexes, )
+  elif isinstance(indexes, list):
+    indexes = tuple(indexes)
+  return self[indexes]
```
lsy323, line 175: qq: which function `self[indexes]` will call? I don't see `__getitem__` in `XLATensor2` class, does it call the `__getitem__` in `torch.Tensor` base class?

lsy323 (approved):

experimental/torch_xla2/torch_xla2/ops/jtorch.py
qihqi: self is of type jax.Array.

PR #7246, opened 2024-06-11
add torch_xla.experimental.compile for eager mode

JackCaoG:
This should only be used for the eager mode. The `compile` pretty much enable the LTC before entering the function and disable it again. 

TODO
1. add more unit test


JackCaoG:
ok test added, should be ready for review.

torch_xla/experimental/eager.py
```diff
@@ -1,9 +1,48 @@
...
+    assert torch_xla._XLAC._get_use_eager_mode() == True
+    torch_xla._XLAC._set_use_eager_mode(False)
+    # clear the pending graph if any
+    torch_xla.sync()
+    try:
+      # Target Function Execution
+      result = func(*args, **kwargs)
+    except Exception as e:
+      # Handle exceptions (if needed)
+      print(f"Error in target function: {e}")
```
qihqi, line 40: Here the exception is tracing exception right?

torch_xla/experimental/eager.py
```diff
@@ -1,9 +1,48 @@
...
+      raise  # Re-raise the exception
+    # Sync the graph generated by the target function.
+    torch_xla.sync()
```
qihqi, line 43: there actaully runs the graph and you might get exceptions here too.

torch_xla/experimental/eager.py
```diff
@@ -1,9 +1,48 @@
...
+
+def compile(func):
+  """Compile the func with Lazy Tensor.
+
+  Return the optimized function that takes exact same input. Compile will
+  run the target func under the tracing mode using Lazy tensor.
+  """
+
+  @functools.wraps(func)  # Keep function's name, docstring, etc.
+  def wrapper(*args, **kwargs):
```
qihqi, line 29: so is the mechanism for caching the graph is already there right so no need to do anything extra?

qihqi (approved):
lgtm, few qqs:

torch_xla/experimental/eager.py
JackCaoG: yea, execution is async so we won't be able to catch it here.

torch_xla/experimental/eager.py
JackCaoG: right, the way that LTC works is that async execution happens in a separate thread and the runtime error will be set in the unlocker. Next time when we try to get the device lock it will find that exception and throw
https://github.com/pytorch/xla/blob/master/torch_xla/csrc/xla_graph_executor.cpp#L861-L864

That being said I agree with you that there is no harm to put `sync` in the `try` region. Let me update that in the following pr.

torch_xla/experimental/eager.py
JackCaoG: we still need to trace the whole model(run all python code), we just skip XLA compilation and lowering to HLO part.

This `compile` does not modify the function bytecode or transform the function in anyway besides enabling the tracing mode. It is actually more accurate to call it `trace` but `compile` is more align with pytorch API.

PR #7251, opened 2024-06-12
Lower upsample_nearest_3d and its backward

wonjoolee95:
Lower upsample_nearest_3d and its backward

TODO
- Tests

### DIFF
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -351,6 +351,8 @@ supported:
   - upsample_bilinear2d_backward
   - upsample_nearest2d
   - upsample_nearest2d_backward
+  - upsample_nearest3d
+  - upsample_nearest3d_backward
   - var.correction
   - var_mean.correction
   - view_as_complex_copy
--- a/torch_xla/csrc/ops/upsample_nearest3d.cpp
+++ b/torch_xla/csrc/ops/upsample_nearest3d.cpp
@@ -1,0 +1,39 @@
+#include "torch_xla/csrc/ops/upsample_nearest3d.h"
+
+#include "absl/strings/str_join.h"
+#include "torch_xla/csrc/lowering_context.h"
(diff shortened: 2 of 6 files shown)


sdsuster:
hi, any update on this PR draft?

wonjoolee95:
Apologies for the delay, I got side-tracked. I noticed that you tried merging this PR on your own fork (https://github.com/sdsuster/xla/pull/1). Did that work when you tried using it on your end?

sdsuster:
the build didn't work, I'm trying to fix it

sdsuster:
Hi @wonjoolee95, I've done implementing the lowering both backward (currently only 2D , and forward(3D). which I will finalize it tomorrow. 

Since im new and not familiar with open-xla, I have question-related xla::CustomCall. I saw that 2d backward implementation called CustomCall instead of implementing a lowered backward function. My question is, where is the target function is defined, is it calling the native function from pytorch (CPU) or calling custom function defined in TPU side?

JackCaoG:
on the 2d case if it is TPU we used the custom call, otherwise we call another implementation BuildResize https://github.com/pytorch/xla/blob/ac13bf23ef811f058bb5fa70699bb79c78564e16/torch_xla/csrc/resize_ops.cpp#L274-L292

that being said checkout https://github.com/pytorch/xla/blob/master/torch_xla/_internal/decomp_registration.py#L3-L4. For `upsample_trilinear3d` I just hook it into the pytorch implementation. On the TPU it is likely to be less efficient but on cpu it is a lot easier to implement.

sdsuster:
Could you check my recents commits
https://github.com/pytorch/xla/compare/v2.4.0...sdsuster:xla:r2.4_merge_upscale_3d_nearest?expand=1

it has been tested both on CPU and TPU. however, 3D upsampling appears to be a resource-intensive (memory-exhausting) operation.

PR #7256, opened 2024-06-12
support eager mode with torch.compile

JackCaoG:


wonjoolee95 (approved):

test/run_tests.sh
```diff
@@ -199,12 +199,13 @@ function run_xla_op_tests1 {
   run_save_tensor_hlo "$CDIR/spmd/test_spmd_graph_dump.py"
 }
 
-# DO NOT MODIFY
```
vanbasten23, line 202: I think Yeounoh added it because GPU test `run_xla_op_tests2` is running so long that all the new tests should be added to run_xla_op_tests3 or later test. This is to reduce the overall CPU/GPU tests.

JackCaoG: right, but I think you fixed the AMP taking too long issue, now this test suite actually finished very quickly

Issue #7253, opened 2024-06-12
[RFC] PyTorch/XLA eager mode as default


JackCaoG:
# Context

## Objective

_In this RFC I will talk about the roadmap to enable eager mode as the default computation mode for PyTorch/XLA users and how to enable graph compilation in this mode._

## Background

PyTorch/XLA has been using tracing mode as the default mode since the project started. All of the torch operation users issued will be accumulated in the background and sent to the XLA for compilation and execution upon a `mark_step` call.

The upside of this approach is that users don’t need to change their model code too much. As long as the user adds a `mark_step` at the right place everything should just work. However from the user feedback in the last couple years this approach creates too much confusion and frustration for the user. Both PyTorch and JAX took the approach of using eager mode as default and asking users to specify the function that they want to compile. PyTorch/XLA should take the same approach.

# Design

## Eager mode

There is no real eager mode in TPU. However we can fake the eager mode by compiling and executing each torch operation. This approach was way too slow a couple years ago due to XRT not being able to execute small executions very efficiently but with PJRT the performance is much better.

The whole eager mode still builds on top of the existing Lazy tensor framework, but becomes invisible to the user. A couple things we need to do to accommodate the eager mode are_

1. Increase the compilation cache from 1024 to 2048 since each torch op will also reside in the compilation cache. We also need to recompile every torch op for different input shapes.
2. Increase the max execution we can queue in the PJRT level since now we will execute a lot more small computations.

## Compile

For the compile part we currently have 2 options, lazy tensor and torch dynamo(torch.compile).

For lazy tensor based compile I will add a new API_

```
torch_xla.experimental.compile(fn) -> compiled_fn
```

Which under the hood just enables the tracing mode upon running the function and executes the traced graph before returning. Here is the [implementation](https://github.com/pytorch/xla/pull/7246/files#diff-1e2407471d3328b83dabbeb29cdf3ef468a201d3d4aecac8f4cd46f76751b8c1). For `torch.compile` we can just use the existing API.

# Example UX

```python
import torch_xla
torch_xla.experimental.eager_mode(True)

Class TrainDecoderOnlyBase():
  def __init__():
    train_loader = MyLoader()
    self.model = DecoderOnlyModel(self.config).to(torch_xla.device())
    # if run with dynamo, use
    # self.step_fn = torch.compile(self.step_fn, backend="openxla")
    self.step_fn = torch_xla.experimental.compile(self.step_fn)

  def step_fn(self, data, target):
[...]

PR #7263, opened 2024-06-12
Increase max in flight operation to accommodate eager mode

JackCaoG:
this should help https://github.com/pytorch/xla/issues/7253


JackCaoG:
I wonder if this is needed, after I increase it I only see ~ 2% better performance on eager mode. @will-cromar wdyt?

will-cromar (approved):

PR #7278, opened 2024-06-14
Avoid log spam

JackCaoG:


will-cromar:
I still see this in the logs for CPU:

```
test_mp_decorator (__main__.MpDecoratorTest) ... WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1718392474.370713     806 cpu_client.cc:467] TfrtCpuClient created.
```

I don't see the log for GPU though.

will-cromar (approved):

PR #7279, opened 2024-06-14
Use default github token for docs preview

will-cromar:
Fixes PRs from forks!


JackCaoG:
Yay it passed

JackCaoG:
oh nvm.. only linter runs..

PR #7295, opened 2024-06-16
Refactor `should_save` Method in Checkpoint Manager to Avoid Saving at Step 0

huzama:
## Description

This pull request modifies the `should_save` method in the checkpoint manager to prevent saving at step 0. In the original implementation, the method would save at step 0 due to the condition `step % self.save_interval == 0`. The new condition, `step % self.save_interval == self.save_interval - 1`, ensures that saving occurs at the last step before a multiple of the save interval, thus avoiding a save at step 0.

### Detailed Changes
- **Original Condition:** `return step % self.save_interval == 0 or preemption_detected`
- **Modified Condition:** `return step % self.save_interval == self.save_interval - 1 or preemption_detected`

### Benefits
- **Avoid Initial Save:** This change prevents unnecessary saves at the very start of the process (step 0), which can be redundant or undesirable in certain workflows.

### DIFF
--- a/torch_xla/experimental/distributed_checkpoint/manager.py
+++ b/torch_xla/experimental/distributed_checkpoint/manager.py
@@ -251,7 +251,7 @@ class CheckpointManager:
           f"Preemption sync point reached at step {step}. Triggering a checkpoint."
       )
       preemption_detected = True
-    return step % self.save_interval == 0 or preemption_detected
+    return step % self.save_interval == self.save_interval - 1 or preemption_detected
 
   def save(self,
            step,


jonb377:
The existing behavior is consistent with [Orbax](https://github.com/google/orbax/blob/2c0e0036e2c09309327b3274d5db314efbc9d84f/checkpoint/orbax/checkpoint/checkpoint_manager.py#L912-L915) (which will save on step 0) but not with [HuggingFace's Trainer](https://github.com/huggingface/transformers/blob/02300273e220932a449a47ebbe453e7789be454b/src/transformers/trainer_callback.py#L549) (which explicitly skips step 0).

To avoid checkpointing on step 0 with CheckpointManager, the training script can be modified:

```python
# Will save on step 0
if chkpt_mgr.save(step, state_dict):
   print(f'took a checkpoint at step {step}')

# Will not save on step 0
if step > 0 and chkpt_mgr.save(step, state_dict):
   print(f'took a checkpoint at step {step}')
```

To keep the CheckpointManager more general, I would be inclined to keep the current behavior which allows for saving step 0. However, if the above workaround doesn't support your use case, I'd be happy to discuss this further. Please let me know what you think @huzama!

huzama:
Thank you, @jonb377, for your insights. I have adjusted my workflow accordingly. Initially, I was puzzled about the rationale behind saving at step 0, but now I understand that it helps keep the CheckpointManager more general.

PR #7296, opened 2024-06-16
Fix: Ensure checkpoint deletion only occurs on rank 0

huzama:
**Description:**

This pull request addresses an issue where checkpoint deletions were occurring on multiple ranks during distributed training, causing input/output errors. The update ensures that checkpoint deletions are only performed on rank 0 to prevent these errors.

**Error Trace:**

```
[rank23]:   File "/home/huzama/.local/lib/python3.10/site-packages/torch_xla/experimental/distributed_checkpoint/manager.py", line 270, in save
[rank23]:     self._save(step, state_dict)
[rank23]:   File "/home/huzama/.local/lib/python3.10/site-packages/torch_xla/experimental/distributed_checkpoint/manager.py", line 220, in _save
[rank23]:     self._delete_chkpt_at_step(step)
[rank23]:   File "/home/huzama/.local/lib/python3.10/site-packages/torch_xla/experimental/distributed_checkpoint/manager.py", line 193, in _delete_chkpt_at_step
[rank23]:     fs.rm(raw_path, recursive=True)
[rank23]:   File "/home/huzama/.local/lib/python3.10/site-packages/fsspec/implementations/local.py", line 185, in rm
[rank23]:     shutil.rmtree(p)
[rank23]:   File "/usr/lib/python3.10/shutil.py", line 731, in rmtree
[rank23]:     onerror(os.rmdir, path, sys.exc_info())
[rank23]:   File "/usr/lib/python3.10/shutil.py", line 729, in rmtree
[rank23]:     os.rmdir(path)
[rank23]: OSError: [Errno 5] Input/output error: 'PATH'
```

**Code Update:**

*Old Code:*
```python
self._delete_chkpt_at_step(step)
```

*New Code:*
```python
if dist.get_rank(self.pg) == 0:
    self._delete_chkpt_at_step(step)
```

### DIFF
--- a/torch_xla/experimental/distributed_checkpoint/manager.py
+++ b/torch_xla/experimental/distributed_checkpoint/manager.py
@@ -190,17 +190,18 @@ class CheckpointManager:
     return os.path.join(self.base_path, str(step))
 
   def _delete_chkpt_at_step(self, step):
-    path = self._get_path(step)
-    fs, raw_path = url_to_fs(path)
-    if fs.exists(raw_path):
-      fs.rm(raw_path, recursive=True)
+    if dist.get_rank(self.pg) == 0:
+      path = self._get_path(step)
+      fs, raw_path = url_to_fs(path)
+      if fs.exists(raw_path):
+        fs.rm(raw_path, recursive=True)
 
   def _release_oldest_checkpoints(self):
     """
     Delete oldest checkpoints until the number of tracked checkpoints is below
     self.max_to_keep. This operation is only execution on the rank 0 process.
     """
-    if dist.get_rank(self.pg) == 0 and self.max_to_keep > 0:
+    if self.max_to_keep > 0:
       while len(self._tracked_chkpts) > self.max_to_keep:
         oldest_chkpt = self._tracked_chkpts.popleft()
         self._delete_chkpt_at_step(oldest_chkpt.step)
@@ -220,8 +221,7 @@ class CheckpointManager:
     with self._save_mutex:
       path = self._get_path(step)
       # Delete any existing checkpoint at the current step.
-      if dist.get_rank(self.pg) == 0:
-        self._delete_chkpt_at_step(step)
+      self._delete_chkpt_at_step(step)
       dist_cp.save(
           state_dict=state_dict,
           storage_writer=FsspecWriter(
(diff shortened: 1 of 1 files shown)


JackCaoG:
@jonb377 do you mind taking a look?

jonb377 (approved):
Good find, thanks for the patch @huzama!

JackCaoG:
Is there an easy way to add a test for this or this won't regress?

jonb377:
The issue only happens with multiple processes attempting to delete the checkpoint at once, so with SPMD we'd require multiple hosts to trigger the codepath.

Alternatively, we could just remove the logic to delete existing steps. That would be more consistent with Orbax's behavior and would eliminate the race condition, but IMO could be more error prone if the user doesn't adjust the step number per-epoch.

@huzama I'm curious to hear your perspective as well - what's the use case where you're hitting this codepath?

huzama:
Thank you @jonb377 for approving the changes. I've been experiencing a race condition issue in the Checkpoint Manager for the past month. I have a few thoughts and suggestions regarding this problem:

1. **Rank Check in Function**: 
    - Should we add a rank check in this function to prevent potential bugs if it is called elsewhere? 
https://github.com/pytorch/xla/blob/e8601ea3bc70c5a57732f75b9352ebaff863071b/torch_xla/experimental/distributed_checkpoint/manager.py#L192-L196

2. **Storage Type Check**:
    - There is no mechanism to verify whether the user is saving to local storage or shared connected storage. If a user saves the checkpoint to local storage, the `.metadata` file is saved on rank 0 only, causing errors for all other ranks.
    - Can we implement a fix for this?

jonb377:
1. That sounds good to me, can we move the rank check there in this PR? You can also remove the rank check from here (keep the `max_to_keep > 0` condition): https://github.com/pytorch/xla/blob/e8601ea3bc70c5a57732f75b9352ebaff863071b/torch_xla/experimental/distributed_checkpoint/manager.py#L203
2. That's an interesting challenge, `.metadata` is managed by the upstream checkpointer, and it's only written by the coordinator host (rank 0). It would be difficult to detect whether it's local storage or not, since some local paths are really shared mounts (like with `gcsfuse`). I think we would need the upstream to do some validation on the coordinator process that all checkpoint shards are locally accessible to catch this.

jonb377 (approved):
Thanks @huzama!

PR #7307, opened 2024-06-18
Wrap adding logging options in try / catch.

qihqi:
Turns out CUDA PJRT does not support that option yet. Which is fine because we are not logging GPU usage anyways, but it should not fail.

### DIFF
--- a/experimental/torch_xla2/torch_xla2/__init__.py
+++ b/experimental/torch_xla2/torch_xla2/__init__.py
@@ -10,18 +10,29 @@ __all__ = [
   'extract_jax',
 ]
 
-
+from jax._src import xla_bridge
+os.environ.setdefault('ENABLE_RUNTIME_UPTIME_TELEMETRY', '1')
 jax.config.update('jax_enable_x64', True)
-jax.config.update(
-  'jax_pjrt_client_create_options',
-  f'ml_framework_name:PyTorch/XLA2;ml_framework_version:{"v0.0.1"}'
-)
+old_pjrt_options = jax.config.jax_pjrt_client_create_options
+
+try:
+  jax.config.update(
+    'jax_pjrt_client_create_options',
+    f'ml_framework_name:PyTorch/XLA2;ml_framework_version:{"v0.0.1"}'
+  )
+  xla_bridge._clear_backends()
+  jax.devices()  # open PJRT  to see if it opens
+except RuntimeError:
+  jax.config.update(
+    'jax_pjrt_client_create_options', old_pjrt_options
+  )
+  xla_bridge._clear_backends()
+  jax.devices()  # open PJRT  to see if it opens
+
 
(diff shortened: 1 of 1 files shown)


experimental/torch_xla2/torch_xla2/__init__.py
```diff
@@ -10,19 +10,30 @@
...
-)
+old_pjrt_options = jax.config.jax_pjrt_client_create_options
+
+try:
+  jax.config.update(
+    'jax_pjrt_client_create_options',
+    f'ml_framework_name:PyTorch/XLA2;ml_framework_version:{"v0.0.1"}'
+  )
+  xla_bridge._clear_backends()
+  jax.devices()  # open PJRT  to see if it opens
```
will-cromar, line 24: This could be a problem in the future. If you import `torch_xla2` while another process holds the device, your second process may crash even before it tries to run anything. It could also be a problem if we add tests that create a child process to test JAX initialization for the same reason, which I may do to test `torch.distributed` init.

Is JAX going to support this option on GPU (or at least not crash if this option is given on GPU) in the near future? If so, I think this is an okay hack for now and we can fix it later.

will-cromar:
Would it make sense to check `jax.default_backend` explicitly here and only set these options on TPU? I just found it while working on something else: https://jax.readthedocs.io/en/latest/_autosummary/jax.default_backend.html

qihqi:
> Would it make sense to check `jax.default_backend` explicitly here and only set these options on TPU? I just found it while working on something else: [jax.readthedocs.io/en/latest/_autosummary/jax.default_backend.html](https://jax.readthedocs.io/en/latest/_autosummary/jax.default_backend.html)

It would make sense if `default_backend` doesnt open PJRT, but in reality it does. So checking it first means we still need to do _clear_backend to have the  options populated.

experimental/torch_xla2/torch_xla2/__init__.py
qihqi: Yes. PJRT for CUDA need to take those args and ignore. PJRT for CPU already does that.

experimental/torch_xla2/torch_xla2/__init__.py
```diff
@@ -10,19 +10,30 @@
...
 
-
+from jax._src import xla_bridge
+os.environ.setdefault('ENABLE_RUNTIME_UPTIME_TELEMETRY', '1')
 jax.config.update('jax_enable_x64', True)
-jax.config.update(
-  'jax_pjrt_client_create_options',
-  f'ml_framework_name:PyTorch/XLA2;ml_framework_version:{"v0.0.1"}'
-)
+old_pjrt_options = jax.config.jax_pjrt_client_create_options
```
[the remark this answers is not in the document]
qihqi: nop, it's empty.

PR #7311, opened 2024-06-18
[torch_xla2] Prototype of `torch.distributed` backend

will-cromar:
Rough implementation of `torch.distributed` using `jax.lax` collective ops in a `ProcessGroup`. This API is very likely to change before any stable release to better support traceability.

### DIFF
--- a/.github/workflows/torch_xla2.yml
+++ b/.github/workflows/torch_xla2.yml
@@ -41,4 +41,4 @@ jobs:
         shell: bash
         run: |
           pytest test/
-          pytest -n 0 test_dist/
+          XLA_FLAGS=--xla_force_host_platform_device_count=4 pytest -n 0 test_dist/
--- a/.github/workflows/torch_xla2.yml
+++ b/.github/workflows/torch_xla2.yml
@@ -41,3 +41,4 @@ jobs:
         shell: bash
         run: |
           pytest test/
+          pytest -n 0 test_dist/
--- a/experimental/torch_xla2/test/test_distributed.py
+++ b/experimental/torch_xla2/test/test_distributed.py
@@ -1,0 +1,42 @@
+import os
+import jax
+import numpy as np
+
+import pytest
+import torch.distributed as dist
+import torch_xla2
+import torch_xla2.distributed
+
+
+@pytest.fixture(scope='module')
+def multi_cpu():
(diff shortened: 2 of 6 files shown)


will-cromar:
@qihqi I have a rough implementation of `torch.distributed` working. I just need to figure out how to either make `pytest` launch a new process for `test_distributed` or re-initialize the PJRT client so I can set the number of CPU cores to 4. Right now, other tests are initializing the runtime with 1 core so none of these test cases are actually distributed.

qihqi:
> @qihqi I have a rough implementation of `torch.distributed` working. I just need to figure out how to either make `pytest` launch a new process for `test_distributed` or re-initialize the PJRT client so I can set the number of CPU cores to 4. Right now, other tests in the CI are initializing the runtime with 1 core so none of these test cases are actually distributed.
> 
> If I run just this test:
> 
> ```
>  pytest -n 1 experimental/torch_xla2/test/test_distributed.py 
> ================================================================================== test session starts ==================================================================================
> platform linux -- Python 3.11.8, pytest-8.1.1, pluggy-1.4.0
> rootdir: /usr/local/google/home/wcromar/build/tx2/xla/experimental/torch_xla2
> configfile: pyproject.toml
> plugins: xdist-3.5.0, hypothesis-6.103.0
> 1 worker [7 items]     
> .......                                                                                                                                                                           [100%]
> =================================================================================== warnings summary ====================================================================================
> ...
> ============================================================================= 7 passed, 1 warning in 4.38s ==============================================================================
> ```

For now, probably just put it in a different folder, say test_dist/ and use plain `python` to run it.

will-cromar:
> For now, probably just put it in a different folder, say test_dist/ and use plain `python` to run it.

Done. Long term, we should follow JAX's example and mark tests that require multiple accelerators vs single shards, e.g. https://github.com/google/jax/blob/e119fe933b66c09625acde6e6db85e859cf21d5a/.github/workflows/cloud-tpu-ci-nightly.yml#L91-L94

Since a recent change immediately inits PJRT when we import `torch_xla2`, I moved the CPU cores runtime flag to the test workflow rather than trying to be clever with a fixture. This test should actually also work on a `v4-8` or any other environment with 4 accelerators FWIW. It'll need further changes to generalize.

qihqi (approved):

PR #7318, opened 2024-06-20
Introduce CUDA OpenXLA fallback.

ysiraichi:
This PR introduces OpenXLA fallback on PyTorch GPU eager. Instead of running fallback operations (i.e. whenever a operation has no lowering implemented) on CPU, we now make it possible to run them on GPU. This makes sense specially when using XLA:CUDA devices.

In summary, this PR introduces the following changes:

- Rename `xla_cpu_fallback` into `xla_fallback`
    - Changes every call site that manually invokes the fallback
- Implement `cuda_fallback` function
    - A version of `at::native::cpu_fallback`, but with a few changes (called out before each function)
    - Ideally, it would be better to generalize `at::native::cpu_fallback` implementation inside PyTorch, though
- Add `XLA_FALLBACK_CUDA` flag for using this feature
- Add tests for fallback operations that are found in torchbench

cc @miladm @JackCaoG @vanbasten23


ysiraichi:
I'm still running torchbench. Will report back when it is over.

test/test_ops.py
```diff
@@ -1,374 +1,415 @@
...
+  name: str
+  variant_test_name: str = ""
 
-  def __new__(cls, name, variant_test_name=""):
-    return super(AllowedOpInfoEntry, cls).__new__(cls, name, variant_test_name)
+
+@dataclass
+class AllowedFallbackOpInfoEntry(AllowedOpInfoEntry):
+  fallback_ops: List[str] = field(default_factory=list)
+  allow_sample: Optional[Callable[[SampleInput], bool]] = None
```
vanbasten23, line 28: What does `allow_sample` mean?

test/test_ops.py
```diff
@@ -1,374 +1,415 @@
...
+
+    # In theory, this should work.
+    # However, the problem is the way we prepare the reference (CPU) inputs:
+    # we clone them. If they were a view, they are not anymore.
+    #
+    # AllowedOpInfoEntry('as_strided', 'partial_views'),
+)
+
+allowed_fallback_opinfo = get_allowed_ops_map(
+    AllowedFallbackOpInfoEntry(
```
vanbasten23, line 377: I wonder how these ops are selected. 
Also, there are existing ops that can potentially falls back such as `nonzero`, do you also plan to test it?

test/test_ops.py
```diff
@@ -423,6 +464,49 @@ def test_reference_eager(self, device, dtype, op):
...
+    for xla_fallback_cuda in ("0", "1"):
+      os.environ[xenv.XLA_FALLBACK_CUDA] = xla_fallback_cuda
+
+      with self.subTest(XLA_FALLBACK_CUDA=xla_fallback_cuda):
+        for i in range(samples_to_run):
+          sample_input = sample_inputs[i]
+
+          # Clear all metrics.
+          # This should also clear the fallbacks.
+          xmetrics.clear_all()
```
vanbasten23, line 499: nit: might be better to be consistent with other files `import torch_xla.debug.metrics as met` and `met.clear_all()`

test/test_ops.py
```diff
@@ -423,6 +464,49 @@ def test_reference_eager(self, device, dtype, op):
...
+    MAX_SAMPLES_TO_RUN = 2
+
+    # Expected fallback operation names.
+    expected_fallbacks = fallback_info.fallback_ops
+
+    # Filter so that only allowed samples are run.
+    sample_inputs = [
+        sample for sample in op.sample_inputs(
+            device, dtype, requires_grad=dtype.is_floating_point) if
+        fallback_info.allow_sample is None or fallback_info.allow_sample(sample)
```
vanbasten23, line 485: if `allow_sample is None` means samples are allowed, is it cleaner to set them all to True?

torch_xla/csrc/BUILD
```diff
@@ -211,18 +213,6 @@ cc_library(
     ],
 )
 
-ptxla_cc_library(
```
vanbasten23, line 214: What's this change for?

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,362 @@
...
 #include "torch_xla/csrc/function_call_tracker.h"
 #include "torch_xla/csrc/runtime/debug_macros.h"
 #include "torch_xla/csrc/runtime/metrics.h"
 #include "torch_xla/csrc/runtime/tf_logging.h"
+#include "torch_xla/csrc/xla_graph_executor.h"
+
+// Forward declaration of PyTorch CUDA functions.
+// Source: c10/cuda/CUDAFunctions.h
+//
+// These are needed in order to synchronize the CUDA device after running
```
vanbasten23, line 23: Great comment!

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,362 @@
...
+}
+
+// Change: use of std::any_of instead of iterating with a for-loop.
+static bool validate_tensor_list(const c10::List<at::Tensor>& tensorlist) {
+  return std::any_of(tensorlist.begin(), tensorlist.end(),
+                     [](const at::Tensor& tensor) { return tensor.defined(); });
+}
+
+// Retrieve the inner XLATensorPtr, and check it lives inside CUDA.
+static XLATensorPtr get_xla_tensor(const at::Tensor& tensor) {
```
vanbasten23, line 81: nit: `get_xla_cuda_tensor` could be a better name since you assert the device to be cuda?

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,362 @@
...
+// just one more device. Therefore, we copied the needed functions in this file.
+//
+// Before each modified function below, we shall specify what has changed,
+// if there was any.
+
+bool UseCUDAFallback() {
+  return runtime::sys_util::GetEnvBool("XLA_FALLBACK_CUDA", false);
+}
+
+// Change: use of std::any_of instead of iterating with a for-loop.
```
vanbasten23, line 74: nit: the comment is stale now?

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,362 @@
...
+}
+
+// Copy back the results from CUDA to XLA.
+// Assumes that we have already synchronized CUDA.
+static at::Tensor to_xla(const at::Tensor& tensor) {
+  return torch_xla::fromDLPack(at::toDLPack(tensor));
+}
+
+// Synchronizes the CUDA device being used by PyTorch.
+static void torch_cuda_synchronize(at::DeviceIndex common_device) {
```
vanbasten23, line 147: What does the param `common_device` mean? The device to be synchronized?

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,362 @@
...
+  c10::cuda::set_device(current);
+}
+
+// Former 'cpu_fallback'.
+// Changes:
+//
+//   1. Track the device index being used. Rationale: we synchronize the device
+//      before crossing device borders for correctness.
+//
+void cuda_fallback(const c10::OperatorHandle& op, torch::jit::Stack* stack,
```
vanbasten23, line 162: I found similar implementations in PyTorch, one in pytorch/aten/src/ATen/native/CPUFallback.cpp and the other one in pytorch/torch/csrc/lazy/ts_backend/ts_eager_fallback.cpp. How is this implementation different from those?

test/test_ops.py
ysiraichi: It filters the sample list, looking for a specific one. I will leave a comment there.

test/test_ops.py
ysiraichi: > I wonder how these ops are selected.

I selected the fallback operations that show-up when running torchbench.

>  Also, there are existing ops that can potentially falls back such as `nonzero`, do you also plan to test it?

Ah, right. That one does appear in torchbench, but I think I forgot about that one. I will add it.

test/test_ops.py
ysiraichi: Yeah. Maybe that's a better default. Thanks

torch_xla/csrc/BUILD
ysiraichi: Now, _aten_cpu_fallback.cpp_ needs functions from:

- _aten_xla_bridge.cpp_
- _xla_graph_executor.cpp_
- _dl_convertor.cpp_

So, I thought it would be easier to merge it into the main library.

torch_xla/csrc/aten_cpu_fallback.cpp
ysiraichi: No. This is the change that I applied to that function.

torch_xla/csrc/aten_cpu_fallback.cpp
ysiraichi: The main difference is that we (a) are falling back to CUDA. Here are more details regarding these 3 implementations:

- _CPUFallback.cpp_ (b) looks more updated than _ts_eager_fallback.cpp_ (c), e.g. handles mutation in tensor lists
- **Device support:** even though (c) does support other devices, its conversion uses `tensor.to(device)`, which copies the tensor. In contrast, (a) we simply share the storage of the tensors
- **Device synchronization:** (a) needs further device synchronization, since we are not calling `tensor.to` method

With all that said, I do agree the functions are all similar to each other. A better approach would be to generalize (b), adding support for some customization.

ysiraichi:
The CI error is a bit tricky to solve.

**Problem:** I'm using some CUDA functions defined inside PyTorch, which requires linking _libc10_cuda.so_ to the test binaries. However, since (in CI) PyTorch isn't being compiled with CUDA support, that won't work.

While I could condition compilation of that code with C++ macros (e.g. using `XLA_CUDA` definition), that would mean that we never compile that code in CI, since PyTorch/XLA is compiled without that flag set + PyTorch is compiled without CUDA support (in that specific CI action).

**Possible Solution:** create a phony implementation for the CUDA functions I'm using, and compile it to another library. Then, if we don't find the library _libc10_cuda.so_, we link this other library.

Notice that this is only needed for the test binaries.

@JackCaoG @vanbasten23 @lezcano 
What do you think?

lezcano:
We could also always compile PyTorch with CUDA support in CI.

vanbasten23:
> The CI error is a bit tricky to solve.
> 
> **Problem:** I'm using some CUDA functions defined inside PyTorch, which requires linking _libc10_cuda.so_ to the test binaries. However, since (in CI) PyTorch isn't being compiled with CUDA support, that won't work.
> 
> While I could condition compilation of that code with C++ macros (e.g. using `XLA_CUDA` definition), that would mean that we never compile that code in CI, since PyTorch/XLA is compiled without that flag set + PyTorch is compiled without CUDA support (in that specific CI action).
> 
> **Possible Solution:** create a phony implementation for the CUDA functions I'm using, and compile it to another library. Then, if we don't find the library _libc10_cuda.so_, we link this other library.
> 
> Notice that this is only needed for the test binaries.
> 
> @JackCaoG @vanbasten23 @lezcano What do you think?

If it's only the test binary that requires pytorch built with CUDA, there is a way to achieve it. In our CI, there is a workflow that build pytorch with CUDA, build torch_xla with CUDA, and run only those tests that requires pytorch with CUDA:
image
You can add your tests to https://github.com/pytorch/xla/blob/bb27cb2a49ee1413cb824af5c9a5e9bacabd0b04/.github/workflows/_test_requiring_torch_cuda.yml#L104.

vanbasten23:
For the problem 1 "Problem1: C++ test binaries need all references to be resolved", you mentioned the "Solution: Create a fallback implementation of the CUDA functions". Could you point to me where is the fallback implementation of the CUDA functions?

miladm:
@zpcore to upgrade the XLA:GPU benchmarking to adopt CUDA fallback setting after this PR lands.

torch_xla/core/xla_env_vars.py
```diff
@@ -31,3 +31,4 @@
 WORLD_SIZE = 'WORLD_SIZE'
 LOCAL_WORLD_SIZE = 'LOCAL_WORLD_SIZE'
 ZERO_COPY_ENABLED = 'ZERO_COPY_ENABLED'
+XLA_FALLBACK_CUDA = 'XLA_FALLBACK_CUDA'
```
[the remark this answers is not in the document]
ysiraichi: I think it makes sense to defaul XLA:CUDA executions on CUDA fallback, while XLA:CPU and XLA:TPU remain on CPU fallback.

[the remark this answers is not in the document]
miladm: As discussed offline, more specifically, after this PR, it would be good to:

Step1: Reversing the function of `XLA_FALLBACK_CUDA` so the default would be CUDA fallback enabled (e.g. define `DISABLE_XLA_FALLBACK_CUDA`). 

Step2: define a mechanism to remove the env variable and have a native solution that just works, and avoids user experience hiccups.

torch_xla/csrc/aten_cuda_functions.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+namespace c10::cuda {
+
+c10::DeviceIndex current_device() noexcept { return -1; }
+
+void set_device(c10::DeviceIndex) { fail("c10::cuda::set_device()"); }
+
+void device_synchronize() { fail("c10::cuda::device_synchronize()"); }
+
+}  // namespace c10::cuda
```
ysiraichi, line 18: @vanbasten23 this is the fallback implementation.

test/cpp/BUILD
```diff
@@ -137,6 +142,7 @@ ptxla_cc_test(
             ":torch_xla_test",
             "//torch_xla/csrc/runtime:metrics",
             "//torch_xla/csrc:tensor",
+            "//torch_xla/csrc:aten_cuda_functions",
```
ysiraichi, line 145: I'm using the fallback implementation for solving the undefined references in C++ tests. I think this should be reasonable, since we don't test fallback on C++ tests.

ysiraichi:
@will-cromar I'm having a hard time figuring out how to make this PR work with CI. Specifically: compile + run fallback operations test (at _test_ops.py_).

**Context:** I'm calling a few PyTorch CUDA functions inside a function in _aten_cpu_fallback.cpp_. The implementation of these functions live in `libc10_cuda.so`.

**Problem:** In the CI action we compile PyTorch/XLA, we actually compile PyTorch and PyTorch/XLA **without** CUDA support. In other words, `libc10_cuda.so` is not created.
- When we try to import `torch_xla` in that same CI action, it fails because `_XLAC` has undefined references to the CUDA functions
- We can't conditionally compile (i.e. `#ifdef XLA_CUDA`) the CUDA functions, since it would mean CUDA OpenXLA fallback never gets compiled

**Proposed Solution:** have 2 libraries: `_XLAC_cpu` (no CUDA OpenXLA fallback) and `_XLAC_cuda`.
- Conditionally import either of them, depending on whether PyTorch was compiled with support for CUDA
- Create an alias like so: `import _XLAC_cpu as _XLAC` for backwards compatibility

I know this is not a pretty solution, so do you have any suggestions?

will-cromar:
Hey @ysiraichi, I'll spend some more time going over this PR tomorrow to try to understand it better.

We were just preparing to remove the separate GPU variant of the main `torch_xla` package by moving the GPU runtime implementation to a PJRT plugin. PyPI doesn't support any sort of platform tag that would let us release separate stable TPU and GPU variants of the main package. We need to figure out how to build one variant of the `torch_xla` package so everyone can just `pip install torch_xla`.

Most of the team that is building from source is doing so on TPUs realistically, so it is a nice convenience to not have to build the CUDA version of PyTorch first. Obviously adding the CUDA `torch` build to the critical path on the CI will be a significant overhead as well. But if we can use a pre-built PyTorch package somehow, I actually don't mind if we use the regular CUDA `torch` package as a build dependency, since my main concern is how slow that build is. cc @JackCaoG since we've talked about this possibility a few times but never had a good enough reason to add this option

I don't fully understand after skimming the PR why we need `libc10_cuda` at build time. Can that be dynamically loaded as needed?

ysiraichi:
> I don't fully understand after skimming the PR why we need libc10_cuda at build time. Can that be dynamically loaded as needed?

It can be loaded at runtime. However, it can't be loaded conditionally. At least, not like this.

Loading conditionally ("as needed") was, in fact, the solution that I was proposing. We could have a separate library with a phony implementation of these CUDA functions. Then, import it only if we are in an environment where PyTorch has no CUDA support.

Let me have a first implementation. We can remove it, if that's not what we want.

test/test_ops.py
vanbasten23: SampleInput is the sample list that you are referring to?

ysiraichi: `SampleInput` is the class that represents one sample.

ysiraichi:
I have worked on this for a while, now, trying a bunch of things. Unfortunately, none of them worked. Here's the current state of things:

**What I tried:**
- Created a new Python library `_XLAC_cuda_functions.so` that holds the definition for the `c10::cuda` functions I need
    - **Idea:** introduce a definition to those `c10::cuda` functions whenever PyTorch doesn't have CUDA support
- Modified the _torch_xla/\_\_init\_\_.py_ so that we import this library if `not torch.cuda.available()`
    - If CUDA is available, we rely on `import torch` to load `libc10_cuda.so`, which brings definition to `c10::cuda` functions

**What is happening:**
- Even though I'm able to import the new `_XLAC_cuda_functions.so` library, I'm still getting undefined reference for `c10::cuda` functions when `import torch_xla` is called

I'm not sure why this is not working given that:

```bash
$ nm -CD _XLAC.cpython-310-x86_64-linux-gnu.so | grep c10::cuda
                 U c10::cuda::set_device(signed char)
                 U c10::cuda::current_device()
                 U c10::cuda::device_synchronize()

$ nm -CD _XLAC_cuda_functions.cpython-310-x86_64-linux-gnu.so | grep c10::cuda
000000000002c0b1 T c10::cuda::set_device(signed char)
000000000002c09e T c10::cuda::current_device()
000000000002c0cd T c10::cuda::device_synchronize()
```

```bash
# This works!
$ LD_PRELOAD=./_XLAC_cuda_functions.cpython-310-x86_64-linux-gnu.so python -c "import torch_xla"

# This doesn't work...
$ python -c "import _XLAC_cuda_functions; import torch_xla"
  File "xla/torch_xla/__init__.py", line 11, in <module>
    import _XLAC
ImportError: xla/_XLAC.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN3c104cuda14current_deviceEv
```

@JackCaoG @vanbasten23 @lezcano @will-cromar 
Any thoughts?

isuruf:
python imports the libraries with `RTLD_LOCAL` which means the symbols from `_XLAC_cuda_functions` are not added to the global symbol table. You need to set `RTLD_GLOBAL` before importing `_XLAC_cuda_functions`.

```
import sys, os
prev = sys.getdlopenflags()
sys.setdlopenflags(os.RTLD_GLOBAL)
import _XLAC_cuda_functions
sys.setdlopenflags(prev)

import torch_xla
```

ysiraichi:
@vanbasten23 @JackCaoG @will-cromar
I think I finally found a solution to the CUDA functions problem, thanks to @isuruf. Here's a summary of the recent changes in this PR:

- **_aten_cuda_functions.cpp_** implements the CUDA functions used by _aten_cpu_fallback.cpu_ that, unless PyTorch is compiled with CUDA, aren't defined anywhere
- In addition to _\_XLAC.so_, we build **_\_XLAC_cuda_functions.so_**, which is composed of only the aforementioned _aten_cuda_functions.cpp_
- Inside _torch_xla/\_\_init\_\_.py_, we conditionally load the newly added _\_XLAC_cuda_functions.so_ library, just after importing `torch`
    - We only load it if `not torch.cuda.is_available()`, fulfilling the reference in _\_XLAC.so_
    - Note: we load it using the `os.RTDL_GLOBAL` options, which loads the library symbols globally

With these modifications, we are able to load the dummy implementation of the CUDA functions conditionally. And, in case we start using a PyTorch instance with CUDA support, we don't load it and rely on `libtorch_python.so`, which depends on `libc10_cuda.so`.

Let me know what you think.
Whenever you have some time, could you give this PR another go?

torch_xla/csrc/aten_cuda_functions.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include <cassert>
+#include <stdexcept>
+
+static void fail(const char* name) {
+  throw std::runtime_error("PyTorch was compiled without CUDA support.");
+}
+
+namespace c10::cuda {
+
+c10::DeviceIndex current_device() { fail("c10::cuda::current_device()"); }
```
vanbasten23, line 14: This mean, if pytorch is built without CUDA, then these phony definition will be used. Without it, doing `import torch_xla` will fail with something like `ImportError: xla/_XLAC.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN3c104cuda14current_deviceEv`?

BUILD
```diff
@@ -46,6 +46,22 @@ cc_binary(
     ]),
 )
 
+cc_binary(
+    name = "_XLAC_cuda_functions.so",
+    copts = [
+        "-fopenmp",
```
vanbasten23, line 52: I wonder how the copts and linkopts are determined

torch_xla/__init__.py
```diff
@@ -1,10 +1,22 @@
...
+import sys
 import tempfile
 import warnings
 
 import torch
+
+if not torch.cuda.is_available():
+  # Load _XLAC_cuda_functions to RTLD_GLOBAL, so that it can be used by _XLAC.
+  flags = sys.getdlopenflags()
+  sys.setdlopenflags(flags | os.RTLD_NOW | os.RTLD_GLOBAL)
```
vanbasten23, line 13: why use `os.RTLD_NOW` to perform all necessary relocations when dlopen is called, as I don't see in isuruf's example?

torch_xla/__init__.py
```diff
@@ -1,10 +1,22 @@
...
 import os
 import re
+import sys
 import tempfile
 import warnings
 
 import torch
+
+if not torch.cuda.is_available():
+  # Load _XLAC_cuda_functions to RTLD_GLOBAL, so that it can be used by _XLAC.
```
vanbasten23, line 11: can you point me to the place where _XLAC_cuda_functions is used by _XLAC?

vanbasten23: oh I guess you mean the phone definition will be made available when we do `import _XLAC` below

torch_xla/__init__.py
```diff
@@ -1,10 +1,22 @@
...
+  flags = sys.getdlopenflags()
+  sys.setdlopenflags(flags | os.RTLD_NOW | os.RTLD_GLOBAL)
+
+  import _XLAC_cuda_functions
+
+  # Then, restore the original flags.
+  sys.setdlopenflags(flags)
```
vanbasten23, line 18: Do you know why we need to restore the original flags?

torch_xla/csrc/aten_cpu_fallback.cpp
vanbasten23: nit: maybe rephrase `Change:` to `Change made:`? `Change:` sounds it is something we want to change next lol

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,355 @@
...
+
+  // This fallback only works if all XLA:CUDA tensor arguments are
+  // on the same CUDA device.
+  //
+  // We keep track of said device, so that after actually running
+  // the operation on PyTorch CUDA eager-mode, we synchronize the
+  // device.
+  //
+  // This variable is updated over the course of 'to_cuda' calls.
+  c10::DeviceIndex common_device = -1;
```
vanbasten23, line 170: I'm confused about the `common_device`. Do we ever change the value after this line?

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,355 @@
...
+        need_convert_tensors.push_back(opt_tensors[i].value());
+        need_convert_tensors_index.push_back(i);
+      }
+      auto cuda_tensors = to_cuda(need_convert_tensors, &common_device);
+      for (const auto i : c10::irange(need_convert_tensors_index.size())) {
+        auto idx = need_convert_tensors_index[i];
+        opt_tensors[idx] = cuda_tensors[i];
+      }
+      (*stack)[arguments_begin + idx] = c10::IValue(opt_tensors);
+    }
```
vanbasten23, line 209: In the cpu implementation, there is a 
```
else if (ivalue.isDevice()) {
      tgt_device = ivalue.toDevice();
      (*stack)[arguments_begin + idx] = c10::IValue(c10::Device(kCPU));
    }
```
why don't we need it?

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,355 @@
...
+  TORCH_CHECK(
+      common_device != -1,
+      "No defined tensors found for CUDA fallback: ", op.operator_name());
+  torch_cuda_synchronize(common_device);
+
+  // Step 3: We need to take special care to handle mutable aliases properly:
+  // If any input tensors are mutable aliases, we need to
+  // directly copy the updated data on the CUDA tensors back to the original
+  // inputs.
+  for (const auto i : c10::irange(tensor_args_indices.size())) {
```
vanbasten23, line 233: Does torch_xla has a concept of mutable aliases?
Also, do you know if there is a specific test cases for step3?

torch_xla/csrc/aten_cpu_fallback.cpp
vanbasten23: Would the original `tgt_device` be easier to understand?

vanbasten23:
Mostly LGTM with minor comments.

Amazing work!

torch_xla/csrc/aten_cuda_functions.cpp
ysiraichi: Exactly.

BUILD
ysiraichi: To be honest, I just copied them from `_XLAC`. I guess I could get rid of them, though. What do you think?

torch_xla/__init__.py
ysiraichi: Internally we discussed that it would be better, just to be safe. That's because `dlopen` needs one of `RTLD_NOW` or `RTLD_LAZY`.

torch_xla/__init__.py
ysiraichi: Yes. Exactly.

torch_xla/__init__.py
ysiraichi: I'd guess that, in general, we don't really want to load things and make them available globally (i.e. some encapsulation for loaded functions).

torch_xla/csrc/aten_cpu_fallback.cpp
ysiraichi: We do change inside `to_cuda` function. There, we check the device of every XLA tensor, so that we are able to synchronize the computation later.

> Would the original `tgt_device` be easier to understand?

Since they have different types, and are used in different ways, I thought that it would be a bit confusing to name it `tgt_device`.

torch_xla/csrc/aten_cpu_fallback.cpp
ysiraichi: Right. I thought we didn't need it, since we would always be on XLA. But, I guess it's important to have it just to be safe.

torch_xla/csrc/aten_cpu_fallback.cpp
ysiraichi: It doesn't. The functional layer abstracts it for PyTorch/XLA. That's why we have to propagate the results back to the input arguments that were mutated.

ysiraichi: > Also, do you know if there is a specific test cases for step3?

Not sure we have those in PyTorch/XLA. For that, we would need an operation that mutates at least one of the inputs, e.g. operations that have the `out` parameter.

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -1,34 +1,386 @@
...
+  // time we convert an XLA tensor argument into a CUDA tensor.
+  c10::Device common_device;
+
+  // CUDA device index where the tensors live in.
+  //
+  // This is used for synchronizing the device where the fallback operation
+  // was called. This should ensure completion of the CUDA computation, in
+  // order to be used by another XLA computation.
+  c10::DeviceIndex index;
+};
```
ysiraichi, line 88: This struct helps making sure only one device is used in the CUDA OpenXLA fallback.

- XLA tensor devices are checked against `common_device`
- CUDA device `index` is also checked just to be safe

torch_xla/csrc/aten_cpu_fallback.cpp
ysiraichi: Done. Let me know what you think.

BUILD
vanbasten23: Well, if it works now, feel free to keep it :p

torch_xla/csrc/aten_cpu_fallback.cpp
vanbasten23: I can see your point. Maybe add some comment such as: common_device refers to the device that all tensors should be on; Ideally, all the tensors should be on the same device

ysiraichi: Just did that.

```c++
  // Common device for all XLA tensors.
  //
  // CUDA OpenXLA fallback is supported only when all XLA tensors live in
  // the same XLA device. This field should be updated and checked every
  // time we convert an XLA tensor argument into a CUDA tensor.
  c10::Device common_device;
```

ysiraichi:
Running TorchBench with `--verify` flag, showed no new accuracy problems on inference (the flag doesn't work with training). Therefore, I will go on and merge this PR.

PR #7321, opened 2024-06-20
Initialize benchmarks with device type.

ysiraichi:
This PR makes benchmark initialization more consistent. This is in line with https://github.com/pytorch/benchmark/pull/2292, which makes `moco` check for `xla` device for initializing its process group.

cc @miladm @vanbasten23 @zpcore


ysiraichi:
I haven't tested this PR on a TPU machine (I don't have access to one). @zpcore Could you test if it works on TPU benchmark runs?

qihqi (approved):

zpcore (approved):
Works fine for TPU, thanks!

PR #7328, opened 2024-06-21
Handle random seed for eager mode

JackCaoG:
`seed_ir_value` was reset by `mark_step` but we don't call `mark_step` in eager mode. In eager mode it make more sense to manually rotate the `devctx->seed` and always create a new seed IR


JackCaoG:
ok this pr is ready for review

wonjoolee95 (approved):

PR #7329, opened 2024-06-21
Avoid OOM when build PyTorch with CUDA enabled.

vanbasten23:
- Use MAX_JOBS to avoid OOM when build PyTorch with CUDA enabled. Use similar way as what PyTorch does https://screenshot.googleplex.com/GBuQUkxGbTRpSGo. According to the [post](https://discuss.pytorch.org/t/build-pytorch-with-cuda-11-6/149918), it could potentially avoid the OOM.
- Skip other unrelated CI workflow. Need to revert before merging the PR


JackCaoG (approved):

vanbasten23:
Thanks Jack for the review!

PR #7332, opened 2024-06-24
Update contributing doc after walkthrough

tengyifei:
Since the docker container runs as root, if we `chown -R root:root .`, that will cause vscode to be unable to load any files the workspace due to permission denied. That's probably not reasonable for local development. Instead I changed to trust repositories from another user.

Added torchvision install instructions since some tests imports torchvision.

The libtpu install instructions tell us to install `torch_xla[tpu]`. IIUC that will just overwrite whatever code we just installed locally with an official build, which is probably not what we want. I changed that to use the "extras" feature from setup.py.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,10 +12,13 @@ We recommend you to use our prebuilt Docker image to start your development work
 
 ### Visual Studio Code Dev Container
 
-* Create an empty directory (optionally on a remote host via SSH) and open it in VSCode. Then, clone PyTorch and PyTorch/XLA:
+* Create an empty directory (optionally on a remote host via SSH) and open it in VSCode. Then, clone
+  PyTorch, TorchVision, and PyTorch/XLA:
 
   ```bash
   git clone --recursive --depth=1 https://github.com/pytorch/pytorch.git
+  # Optional: install TorchVision if you need to run tests that involve vision modules
+  git clone --recursive --depth=1 https://github.com/pytorch/vision.git
   git clone https://github.com/pytorch/xla.git pytorch/xla
   # Optional: use git@github.com:pytorch/xla.git instead if you prefer to use SSH with key forwarding
   ```
@@ -29,22 +32,35 @@ We recommend you to use our prebuilt Docker image to start your development work
   ln -s pytorch/xla/.clang-format .clang-format
   ```
 
-* From VSCode's command menu, run `Reopen in Container` to open your workspace in one of our pre-built Docker containers. Select the correct container config based on your local accelerator (default to `tpu-contributor` if you are not sure).
+* From VSCode's command menu, run `Reopen in Container` from the command palette
+  (F1 key) to open your workspace in one of our pre-built Docker containers.
+  Select the correct container config based on your local accelerator (default to
+  `tpu-contributor` if you are not sure).
 
-* Since you are running as root in this container, change ownership of all files:
+  * If you cannot find `Reopen in Container`, make sure the `Dev Containers`
+    VSCode extension is installed, then open the `pytorch/xla` folder in VSCode.
+
+* Since you are running as root in this container, teach `git` to recognize the
+  repositories you just cloned (outside of docker) as safe:
 
   ```bash
-  chown -R root:root .
+  git config --global --add safe.directory /workspaces/torch/pytorch
+  git config --global --add safe.directory /workspaces/torch/pytorch/xla
+  git config --global --add safe.directory /workspaces/torch/vision
   ```
 
-* Build PyTorch and PyTorch/XLA:
+* Build PyTorch, TorchVision, and PyTorch/XLA:
 
   ```bash
-  cd pytorch/
+  cd pytorch
   # pytorch/xla requires pytorch wheel to be presented under pytorch/dist
   python setup.py bdist_wheel
   python setup.py install
-  cd xla/
+  cd ..
+  cd vision
+  python setup.py develop
+  cd ..
+  cd pytorch/xla
   python setup.py develop
   # Optional: if you're using TPU, install libtpu
   pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html


CONTRIBUTING.md
```diff
@@ -31,23 +33,30 @@ We recommend you to use our prebuilt Docker image to start your development work
...
   python setup.py install
-  cd xla/
+  popd
+  pushd vision
   python setup.py develop
-  # Optional: if you're using TPU, install libtpu
-  pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
+  popd
+  cd pytorch/xla
+  pip install -e .
```
vanbasten23, line 57: Is that the same as `python setup.py develop`?

CONTRIBUTING.md
```diff
@@ -12,10 +12,12 @@ We recommend you to use our prebuilt Docker image to start your development work
 
 ### Visual Studio Code Dev Container
 
-* Create an empty directory (optionally on a remote host via SSH) and open it in VSCode. Then, clone PyTorch and PyTorch/XLA:
+* Create an empty directory (optionally on a remote host via SSH) and open it in VSCode. Then, clone
+  PyTorch, TorchVision, and PyTorch/XLA:
 
   ```bash
   git clone --recursive --depth=1 https://github.com/pytorch/pytorch.git
+  git clone --recursive --depth=1 https://github.com/pytorch/vision.git
```
JackCaoG, line 20: Let's add a comment say this is optional.

CONTRIBUTING.md
```diff
@@ -31,23 +33,30 @@ We recommend you to use our prebuilt Docker image to start your development work
 
 * From VSCode's command menu, run `Reopen in Container` to open your workspace in one of our pre-built Docker containers. Select the correct container config based on your local accelerator (default to `tpu-contributor` if you are not sure).
 
-* Since you are running as root in this container, change ownership of all files:
+* Since you are running as root in this container, teach `git` to recognize the repositories you
+  just cloned (outside of docker) as safe:
 
   ```bash
-  chown -R root:root .
```
JackCaoG, line 37: @will-cromar

CONTRIBUTING.md
```diff
@@ -31,23 +33,30 @@ We recommend you to use our prebuilt Docker image to start your development work
...
+  git config --global --add safe.directory /workspaces/torch/pytorch/xla
+  git config --global --add safe.directory /workspaces/torch/vision
   ```
 
-* Build PyTorch and PyTorch/XLA:
+* Build PyTorch, TorchVision, and PyTorch/XLA:
 
   ```bash
-  cd pytorch/
+  pushd pytorch
```
JackCaoG, line 48: why use `pushd` instead of `cd`?

CONTRIBUTING.md
```diff
@@ -31,23 +33,30 @@ We recommend you to use our prebuilt Docker image to start your development work
 
 * From VSCode's command menu, run `Reopen in Container` to open your workspace in one of our pre-built Docker containers. Select the correct container config based on your local accelerator (default to `tpu-contributor` if you are not sure).
```
vanbasten23, line 34: Somehow I couldn't find "Reopen in Container" so I can't verify the whole process.

CONTRIBUTING.md
tengyifei: Yeah, from [1] I believe they are equivalent. The `pip` method lets us install the `tpu` extras and I updated both for consistency.

[1]: https://stackoverflow.com/questions/45426794/setup-in-virtualenv-pip-install-e-vs-python-setup-py-install

CONTRIBUTING.md
tengyifei: Usually I prefer `pushd` when navigating deep folders but I think this case there's no obvious advantage, so I'll replace with the more familiar `cd`.

CONTRIBUTING.md
tengyifei: I think @will-cromar  added this section. I've done many things with vscode so I likely had the right extension. Maybe you need this one: [2]?

[2]: https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers

vanbasten23: I have the extension `Dev Containers`.  Is `Reopen in Container` the correct name?

tengyifei: Yes! (and select the appropriate json config)

vanbasten23: I know what's missing. One should open the pytorch/xla folder first in vscode, then `Reopen in Container` will show up after pressing F1, then choose the config. Could you add the missing info to the doc?

CONTRIBUTING.md
JackCaoG: I still prefer `python setup.py develop` since it is what we used in above examples.

CONTRIBUTING.md
```diff
@@ -31,23 +34,30 @@ We recommend you to use our prebuilt Docker image to start your development work
...
-  cd xla/
+  cd ..
+  cd vision
   python setup.py develop
-  # Optional: if you're using TPU, install libtpu
-  pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
+  cd ..
+  cd pytorch/xla
+  pip install -e .
+  # Optional: if you're using TPU, install TPU extras
```
JackCaoG, line 59: `install TPU extras` -> `install libtpu`

CONTRIBUTING.md
```diff
@@ -31,23 +34,30 @@ We recommend you to use our prebuilt Docker image to start your development work
...
+  pip install -e .[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
```
JackCaoG, line 60: can we just use `pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html` I think both will works but I prefer to keep the command to be same (for local install or system install pytorch/xla)

vanbasten23:
Mostly lgtm with one minor comment

vanbasten23 (approved):

CONTRIBUTING.md
tengyifei: Done

CONTRIBUTING.md
tengyifei: SGTM

PR #7333, opened 2024-06-24
Add support as_strided

ManfeiBai:


ManfeiBai:
fixed in https://github.com/pytorch/xla/pull/7323, would close this PR

PR #7341, opened 2024-06-24
Enable eager spmd

JackCaoG:


alanwaketan:
How could SPMD possibly work for eager mode?

JackCaoG:
> How could SPMD possibly work for eager mode?

consider eager mode as calling `mark_step` after every pytorch op.

alanwaketan:
> > How could SPMD possibly work for eager mode?
> 
> consider eager mode as calling `mark_step` after every pytorch op.

Then how sharding propogation and auto partition work? I assume they don't carry states from last graph?

alanwaketan:
Okay, that's fair.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -2742,7 +2742,9 @@ void XLANativeFunctions::_propagate_xla_data(const at::Tensor& input,
 
   // 2) Aid SPMD.
   XLATensor::ShardingSpecPtr sharding = input_tensor->sharding_spec();
-  if (sharding && sharding->sharding.type() != xla::OpSharding::UNKNOWN) {
+  // don't propagate sharding in eager mode.
+  if (!XLAGraphExecutor::Get()->UseEagerMode() && sharding &&
```
alanwaketan, line 2746: May I ask why?

JackCaoG: It complained about the output tensor already has a sharding and we can't propagate to it. This happens in the backward. I didn't spend enough time to debug it but I don't expect user to actually run eager mode with step fn(forward and backward), I only expect them to run it with some data preprocessing on device so I just quickly unblock myself.

alanwaketan (approved):
LGTM. Thanks, Jack!

PR #7576, opened 2024-06-26
[SPMD] Support manual all-reduce

alanwaketan:
Summary:
This is to add manual all-reduce support to SPMD and it currently only supports one input tensor. For array support, we can do that in python layer instead.

Test Plan:
python ./test/spmd/test_xla_sharding.py -v -k test_spmd_all_reduce

### DIFF
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -392,6 +392,12 @@ void all_reduce(const std::vector<XLATensorPtr>& inputs,
   }
 }
 
+XLATensorPtr all_reduce(const XLATensorPtr& input, AllReduceType reduce_type,
+                        double scale, std::vector<std::vector<int64_t>> groups) {
+    return input->CreateFrom(torch::lazy::MakeNode<AllReduce>(
+      reduce_type, input->GetIrValue(), scale, std::move(groups)));
+}
+
 std::pair<XLATensorPtr, torch::lazy::Value> reduce_scatter(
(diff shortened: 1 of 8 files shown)


torch_xla/csrc/tensor_methods.cpp
```diff
@@ -392,6 +392,13 @@ void all_reduce(const std::vector<XLATensorPtr>& inputs,
   }
 }
 
+XLATensorPtr all_reduce(const XLATensorPtr& input, AllReduceType reduce_type,
```
JackCaoG, line 395: can you call it `all_reduce _no_token`, the only difference in signature is it does not take `pin_layout` but the main difference in the op is that it does not set token.. It is better to reflect that in the name.

JackCaoG (approved):
approve to unblock, but I think we should fix the tensor method name

JackCaoG:
for array support  do you plan to call `all_reduce` multiple times? In our C++ implementation I think we group tensors by dtype and call all_rduce once per dtype.

torch_xla/csrc/tensor_methods.cpp
alanwaketan: Sure. I can follow up with that.

alanwaketan:
> for array support do you plan to call `all_reduce` multiple times? In our C++ implementation I think we group tensors by dtype and call all_rduce once per dtype.

I don't think that's necessary. I'm thinking the compiler should be smart enough to fuse all-reduces if the fusion is necessary.

alanwaketan:
Thanks Jack for approving.

PR #7580, opened 2024-06-26
[torch_xla2] Functional collectives

will-cromar:
- Move collective op implementations out of `ProcessGroup` and into registered `__torch_dispatch__` ops.
- Rewrite ProcessGroup implementation using functional collectives
  -  `torch.distributed` dispatches to `c10d.*` (e.g. `torch.ops.c10d.allreduce_`) instead of the functional op directly. he signatures of `c10d.*` differ from `_c10d_functional.*_inplace` differ somewhat (str reduce op vs enum, tensor vs list of tensor), so we can't just reuse the exact same function.
  - When I tried to implement `c10d.*` directly, all non-tensor objects were wrapped in `ScriptObject`, causing a bunch of errors. I could not figure out how to unwrap them.
  - The default implementation of these ops is a wrapper around PG, e.g. https://github.com/pytorch/pytorch/blob/fdd0a7f9b4fb80c4d4870569909505a9beb6ccb3/torch/csrc/distributed/c10d/Ops.cpp#L162-L180
- Some functional collectives ignore the process group given a group name. Others look up e.g. the world size by the PG group name in the background (namely `all_gather`). It's unclear if that's the long-term intended behavior. Let's keep the ProcessGroup for now.

Aside from the traceable op implementations being cleaner, dynamo rewrites `torch.distributed` calls into their functional equivalents. E.g.

```
@torch.compile(backend=my_backend)
def cc(index):
  dist.all_reduce(index)
  return index
```

generates

```
opcode         name         target                                args                   kwargs
-------------  -----------  ------------------------------------  ---------------------  --------
placeholder    arg0_1       arg0_1                                ()                     {}
call_function  all_reduce   _c10d_functional.all_reduce.default   (arg0_1, 'sum', '0')   {}
call_function  wait_tensor  _c10d_functional.wait_tensor.default  (all_reduce,)          {}
call_function  copy         aten.copy.default                     (arg0_1, wait_tensor)  {}
output         output       output                                ((copy, copy),)        {}
```

cc @qihqi 

Depends on #7311

### DIFF
--- a/experimental/torch_xla2/torch_xla2/ops/jaten.py
+++ b/experimental/torch_xla2/torch_xla2/ops/jaten.py
@@ -101,7 +101,8 @@ def _c10d_broadcast(self, src: int, group_name: str):
 
 
 @op(torch.ops._c10d_functional.wait_tensor)
-def _wait_tensor(tensor):
+def _c10d_wait_tensor(tensor):
+  # Async tensor is aleady `wait`ed by dispatcher
   return tensor
 
 
--- a/experimental/torch_xla2/torch_xla2/ops/jaten.py
+++ b/experimental/torch_xla2/torch_xla2/ops/jaten.py
@@ -68,44 +68,6 @@ def _handle_int64_trig(self, func):
   return res
 
 
-@op(torch.ops._c10d_functional.all_gather_into_tensor)
-def _c10d_all_gather(input, group_size: int, group_name: str):
-  return jax.lax.all_gather(input, "torch_dist")
-
-
-# TODO(wcromar): move c10d ops since they're not actually aten
-@op(torch.ops._c10d_functional.all_reduce)
-def _c10d_all_reduce(self, reduceOp: str, group_name: str):
-  match reduceOp:
-    case 'sum':
-      res = jax.lax.psum(self, axis_name="torch_dist")
-    case 'avg':
-      res = jax.lax.pmean(self, axis_name="torch_dist")
-    case 'min':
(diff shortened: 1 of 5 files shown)


experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -64,6 +65,49 @@ def _handle_int64_trig(self, func):
...
+      jax.lax.axis_index("torch_dist") == src,
+      self,
+      jnp.zeros_like(self),
+  )
+  return jax.lax.psum(masked, "torch_dist")
+
+
+@op(torch.ops._c10d_functional.wait_tensor)
+def _wait_tensor(tensor):
+  return tensor
```
qihqi, line 102: qq why not .wait here?

will-cromar: To be honest, I don't actually understand why this works either.

There's no wait here because this is a JAX implementation, so some part of unwrapping the `Tensor` into an `Array` must get rid of the async wrapper. That's something I want to revisit. The dispatcher should probably be aware of `AsyncCollectiveTensor`s and `wait` them so our op implementation don't have to care (see how `copy_` has to have special handling right now).

will-cromar: The dispatcher waits during `t2j` before calling a JAX implementation.

qihqi (approved):

PR #7581, opened 2024-06-26
refining the AOT compilation

zpcore:
This is a follow up PR to refine https://github.com/pytorch/xla/pull/6992. 

In this PR, I created the `PjRtCompilationClient` to serve the ahead of time compilation. In this way, we don't need to create the `CompileOnlyPjRtClient`, `CompileOnlyPjRtDevice` etc. This make it easier when during openxla pin update.

Instructions on how to run AOT compilation has been updated. We need to specify two extra flags: `XLA_REGISTER_INSTALLED_PLUGINS` and `XLA_PERSISTENT_CACHE_PATH` as follows:
```
PJRT_DEVICE=TPU  XLA_REGISTER_INSTALLED_PLUGINS=0  XLA_PERSISTENT_CACHE_PATH=./  python sample.py
```  
sample.py:
```python
import torch
import torch_xla

torch_xla._XLAC._xla_set_virtual_topology("v4:2x2x1")  # <----- define virtual topology here

a = torch.rand([2,3])
b = torch.rand([2,3])
device = xm.xla_device() # this will trigger the TPU device not found
a = a.to(device)


b = b.to(device)
f = torch.hstack([a,b])

torch_xla._XLAC._xla_warm_up_cache([f],[])  # <----- call this to avoid real computation.
```


torch_xla/csrc/runtime/pjrt_compile_only.h
```diff
@@ -0,0 +1,48 @@
+#ifndef XLA_CLIENT_PJRT_COMPILE_ONLY_H_
+#define XLA_CLIENT_PJRT_COMPILE_ONLY_H_
+
+#include "xla/pjrt/pjrt_client.h"
+#include "xla/pjrt/pjrt_future.h"
+
+namespace xla {
+
+class CompileOnlyPjRtBuffer final : public PjRtBuffer {
```
will-cromar, line 9: Thanks for reducing the dependency on PJRT. Why do you still need an implementation of PjRtBuffer? Could `PjRtCompilationClient::PjRtData` just not hold a buffer?

torch_xla/csrc/runtime/runtime.cc
```diff
@@ -12,6 +13,8 @@ namespace runtime {
 
 std::atomic<bool> g_computation_client_initialized(false);
 
+std::string aot_topology = "";
```
will-cromar, line 16: Since this state is only ever used by `PjRtCompilationClient`, would it make sense to make this a static field on `PjRtCompilationClient`?

I think it's probably even better to just imperatively initialize this client when `set_virtual_topology()` happens... but that also would introduce some extra edge cases for initialization. For the POC, a global var is okay, but I think a static field makes more sense to make it clear what uses that state.

torch_xla/csrc/runtime/pjrt_compilation_client.cc
```diff
@@ -0,0 +1,707 @@
...
+using xla::internal::XlaBuilderFriend;
+
+namespace torch_xla {
+namespace runtime {
+
+namespace {
+
+// Builds a map from the device's global ordinal to its index in the `devices`
+// array.
+std::unordered_map<int, int> build_index_map(
```
will-cromar, line 43: I see some duplication of helpers here and in the PJRT/IFRT client implementations. It would be cleaner to factor it out rather than have 3 copies, but both AOT and IFRT are just in a concept stage...

I'll let @JackCaoG have the final word on readability so I don't have to decide.

torch_xla/csrc/runtime/pjrt_compilation_client.cc
```diff
@@ -0,0 +1,707 @@
...
+  }
+
+  return strs;
+}
+
+PjRtCompilationClient::PjRtCompilationClient(
+    std::string& virtual_topology_str) {
+  std::string device_type = sys_util::GetEnvString(env::kEnvPjRtDevice, "");
+
+  // With compilation only, client_ and coordinator_ are nullptr. The point we
```
will-cromar, line 99: It doesn't seem right to call InitializePjrt and throw away the results. If you're planning to hardcode this for only TPU support, just put `pjrt::LoadPjrtPlugin("tpu", etc)` here and skip everything else. (Also, the branch you added over in `pjrt_registry` won't fire by default because dynamic plugins are enabled by default now)

If you want to support general devices, then you would need to add a new function in `pjrt_registry` like `InitializeTopology` that loads the plugin and returns the topology.

torch_xla/csrc/runtime/pjrt_compile_only.h
zpcore: I have tried to use PjRtData without a buffer before. The issue is that Pytorch will produce the correct IR but the HLO is incorrect from lowering_ctx.BuildXla(). This is due to all tensors being mapped to the same null buffer and the HLO will get rid of the redundant tensors with the same null buffer. The solution is to create our own `PjRtCompileOnlyBuffer` that doesn’t interact with the real buffer. As long as our buffer pointers are not null and pointing to different addresses, the HLO will be produced correctly.

The details are mentioned in the `AOT compilation` design doc. However, I think I can further simplify the buffer class in this case and move them inside the pjrt_compilation_client.h.

zpcore: I just simplified the PjRtBuffer into :
```
struct Buffer {
  xla::Shape shape;
  Buffer(xla::Shape shape): shape(shape) {}
};
```
This way we can get rid off the openxla buffer decenency while still maintaining the PjRtData buffer attribute.

torch_xla/csrc/runtime/pjrt_compilation_client.cc
zpcore: That's why I mentioned ` XLA_REGISTER_INSTALLED_PLUGINS=0` in the PR description to skip the dynamic plug in.

Simply to use `pjrt::LoadPjrtPlugin("tpu", etc)` seems to be a good solution. Let me address that.

torch_xla/csrc/runtime/runtime.cc
zpcore: Moving `aot_topology` to `pjrt_compilation_client.h` may cause circular dependency issue because  setting the aot_topology need to check `g_computation_client_initialized` in the current setup. 

Let's use the global var for now since this makes it clear in runtime.cc that we have three kinds of clients all together.

miladm:
@zpcore do we plan to merge this PR? 2.6 release?

cc @JackCaoG @tengyifei

JackCaoG:
@miladm most likely no, this pr still needs some polish and we might not have the right set of people to review and work on it now.

jeffhataws:
How will this work with torch.compile?

zpcore:
> How will this work with torch.compile?

I think it should work. Below is the regular compilation flow of PyTorch/XLA and the proposed Cross-device AOT Compilation:
image

In regular compilation, lazy tensor (LTC) and dynamo backend follow the different path below:
- Nondynamo: code → trace → IR → Hashkey → XlaComputation → (HLO) →  PjRtLoadedCompuatation  (XLA executable to cache)
- Dynamo: code → fx.graph → IR  →  Hashkey → XlaComputation → (HLO)
→  PjRtLoadedCompuatation (XLA executable to cache)

AOT sits after the HLO generation. I think in theory it should work with little changes.

PR #7583, opened 2024-06-27
Add documentation on fixing a torchbench model on torchxla2

qihqi:
* Also add embedding_bag op (not efficient) as an example
* Add facilities for logging op by op and accuracy test


ManfeiBai (approved):
LGTM

Issue #6008, opened 2023-12-03
[torchbench] `dlrm` fails to run on training.


ysiraichi:
## 🐛 Bug

Running the upstreamed benchmarking scripts with the following command results in an unexpected error.

```bash
python xla/benchmarks/experiment_runner.py \
       --suite-name torchbench \
       --accelerator cuda \
       --xla PJRT --xla None \
       --dynamo openxla --dynamo None \
       --test train \
       --repeat 30 --iterations-per-run 5 \
       --print-subprocess \
       --no-resume -k dlrm
```

```python
Traceback (most recent call last):
  File ""xla/benchmarks/experiment_runner.py", line 601, in <module>
    main()
  File "xla/benchmarks/experiment_runner.py", line 597, in main
    runner.run()
  File "xla/benchmarks/experiment_runner.py", line 65, in run
    self.run_single_experiment(experiment_config, model_config)
  File "xla/benchmarks/experiment_runner.py", line 161, in run_single_experiment
    run_metrics, output = self.timed_run(benchmark_experiment,
  File "xla/benchmarks/experiment_runner.py", line 328, in timed_run
    output = loop()
  File "xla/benchmarks/experiment_runner.py", line 310, in loop
    output = benchmark_model.model_iter_fn(
  File "torch/_dynamo/eval_frame.py", line 489, in _fn
    return fn(*args, **kwargs)
  File "xla/benchmarks/torchbench_model.py", line 247, in train
    super().train(inputs, collect_full_output=collect_full_output)
  File "xla/benchmarks/benchmark_model.py", line 142, in train
    self._optimizer_zero_grad()
  File "xla/benchmarks/benchmark_model.py", line 145, in resume_in_train
    loss.backward()
  File "torch/_tensor.py", line 503, in backward
    torch.autograd.backward(
  File "torch/autograd/__init__.py", line 266, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
  File "torch/autograd/function.py", line 289, in apply
    return user_fn(self, *args)
  File "torch/_functorch/aot_autograd.py", line 4201, in backward
    out = call_compiled_backward()
  File "torch/_functorch/aot_autograd.py", line 4167, in call_compiled_backward
    out = call_func_with_args(
  File "torch/_functorch/aot_autograd.py", line 2016, in call_func_with_args
    out = normalize_as_list(f(args))
  File "torch/_dynamo/eval_frame.py", line 489, in _fn
[...]

PR #7584, opened 2024-06-27
Fallback `_embedding_bag_backward` and force `sparse=false`.

ysiraichi:
Fix: #6008 

This PR, in addition to https://github.com/pytorch/pytorch/pull/129691, add support for embedding bag calls when its `sparse` parameter is true.

**Problem:** when `sparse=true`, `_embedding_bag_backward` called `at::_sparse_coo_tensor_unsafe_symint` function which returns a sparse tensor. Since PyTorch/XLA does not support sparse tensors, this resulted in a dispatch error (see [the original issue](https://github.com/pytorch/xla/issues/6008)).

**Solution:** although, ideally we should support sparse tensors, in the short-term we decided (in an offline discussion) to fallback to the dense backwards function.

cc @miladm @JackCaoG


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1489,6 +1489,24 @@ XLANativeFunctions::_embedding_bag_forward_only(
...
+  TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  TORCH_WARN(
+      "XLA does not support EmbeddingBag sparse backward function. "
+      "Falling back to the dense function.");
+  return at::native::call_fallback_fn<&xla_cpu_fallback,
+                                      ATEN_OP(_embedding_bag_backward)>::
+      call(grad, indices_, offsets_, offset2bag, bag_size_, max_indices_,
+           num_weights, scale_grad_by_freq, mode, /*sparse=*/false,
+           per_sample_weights_opt, padding_idx);
+}
```
JackCaoG, line 1508: can we add a test to `test/test_operations.py`, run the embedding bag fwd and bwd and make sure it doesn't crash? this way we can prevent it from regressing.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1489,6 +1489,24 @@ XLANativeFunctions::_embedding_bag_forward_only(
...
 
+at::Tensor XLANativeFunctions::_embedding_bag_backward(
+    const at::Tensor& grad, const at::Tensor& indices_,
+    const at::Tensor& offsets_, const at::Tensor& offset2bag,
+    const at::Tensor& bag_size_, const at::Tensor& max_indices_,
+    int64_t num_weights, bool scale_grad_by_freq, int64_t mode, bool sparse,
+    const std::optional<at::Tensor>& per_sample_weights_opt,
+    int64_t padding_idx) {
+  TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  TORCH_WARN(
```
bhavya01, line 1500: Should warn about this only if sparse is `true`?

bhavya01 (approved):

torch_xla/csrc/aten_xla_type.cpp
ysiraichi: Ah. You are right!

test/test_operations.py
```diff
@@ -2198,6 +2198,82 @@ def foo(x, is_xla=False):
...
+      fresh = tensor
+      # Maybe move to the specified device.
+      if device is not None:
+        fresh = fresh.to(device)
+      # Clone if not cloned already by the previous device move.
+      if fresh.data_ptr() == tensor.data_ptr():
+        fresh = tensor.clone()
+      # Make this tensor a leaf tensor by detaching and reseting its
+      # requires_grad property.
+      fresh = fresh.detach()
```
vanbasten23, line 2229: Very cool. I wonder why we need to make the tensor a leaf tensor. What would happen if we don't do it?

vanbasten23 (approved):

test/test_operations.py
ysiraichi: It's so we can grab its `tensor.grad` and compare.

PR #7585, opened 2024-06-27
should_donate should not propagate to the output data

JackCaoG:
`torch.ops.xla.dynamo_set_buffer_donor_(input)` will set the buffer of the input tensor(we assume input is a XLAData) to be dondated. This is mostly used in the dynamo inference to set the big tensors(like kv cache) to be aliased with output. This is because in dynamo we can not infer the aliasing relationship directly.

Currently for
```
def fn_a(input):
  torch.ops.xla.dynamo_set_buffer_donor_(input)
  input += 1
  return input

def fn_b(input):
  input *= 100
  return input

a = torch.randn(5,5).to(torch_xla.device())
compiled_a = torch.compile(fn_a, backend="openxla")
compiled_b = torch.compile(fn_b, backend="openxla")

res_1 = compiled_a(input)
res_2 = compiled_b(res_1)
```

res_1's XLAData will also has the `should_donate_buffer` set to true, as a result, `compiled_b` will also donate the input buffer even it does not call `torch.ops.xla.dynamo_set_buffer_donor_`. I think this behavior is confusing and we should change it.

### DIFF
--- a/test/dynamo/test_dynamo_aliasing.py
+++ b/test/dynamo/test_dynamo_aliasing.py
@@ -123,6 +123,7 @@ class TestDynamoBufferDonationAliasing(unittest.TestCase):
     self.assertEqual(met.counter_value('XlaSetBufferDonation'), 1)
     dummy_inplace_add_compiled(input)
     torch.allclose(input_cloned.cpu() + 1, input.cpu())
+    self.assertFalse(torch_xla._XLAC._get_buffer_donation(input))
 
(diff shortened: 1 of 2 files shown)


JackCaoG:
hmm
```
2024-06-27T22:33:21.0350937Z     ERROR: An error occurred during the fetch of repository 'zlib':
```

JackCaoG:
This pr should be ready for review.

PR #7590, opened 2024-06-28
get_memory_info can handle default arg

JackCaoG:


torch_xla/core/xla_model.py
```diff
@@ -1488,15 +1488,18 @@ class MemoryInfo(TypedDict):
...
-    device: The device whose memory information are requested.
+    device: Optional[torch.device] The device whose memory information are requested.
+    If not passed will use the default device.
 
   Returns:
     MemoryInfo dict with memory usage for the given device.
   """
+  if device == None:
+    device = xla_device()
   return torch_xla._XLAC._xla_memory_info(str(device))
```
will-cromar, line 1503: nit: this can be

```
  return torch_xla._XLAC._xla_memory_info(str(device or xla_device()))
```

will-cromar (approved):

torch_xla/core/xla_model.py
JackCaoG: lol so many python tricks. I prefer to leave it like this if that is OK with you. I think your implementation is cleaner but hard to read at first glance haha.

Issue #7266, opened 2024-06-13
Error: Check Failed: `it != inputs_.end()` When Using `torch` and `torch_xla` Nightly Version (Post-20240527) with SPMD


huzama:
## ❓ Questions and Help

I am encountering an issue with an SPMD LLAMA forward pass script. When running the script on a Google Cloud v3-8 instance using `torch` and `torch_xla` nightly version (post-20240527), I receive the following error:

```
File "/home/huzama/.local/lib/python3.10/site-packages/transformers/models/llama/modeling_llama.py", line 184, in apply_rotary_pos_emb
    k_embed = (k * cos) + (rotate_half(k) * sin)
RuntimeError: torch_xla/csrc/aten_xla_type.cpp:161 : Check failed: it != inputs_.end() *** Begin stack trace ***
        tsl::CurrentStackTrace()
        torch_xla::XLANativeFunctions::mul(at::Tensor const&, at::Tensor const&)

        at::_ops::mul_Tensor::redispatch(c10::DispatchKeySet, at::Tensor const&, at::Tensor const&)
        at::_ops::mul_Tensor::call(at::Tensor const&, at::Tensor const&)
        PyNumber_Multiply
        _PyEval_EvalFrameDefault
```

Interestingly, running the exact same script within a Jupyter Notebook in the same environment works perfectly.

Here is the script:

```python
import numpy as np
import torch
import torch_xla.core.xla_model as xm
import torch_xla.distributed.spmd as xs
import torch_xla.runtime as xr
from transformers import LlamaForCausalLM

xr.use_spmd()

device = "xla"

model = LlamaForCausalLM.from_pretrained("meta-llama/Meta-Llama-3-8B")
model = model.to(device)

spmd_mesh = xs.Mesh(np.array(range(8)), (4, 2), ("data", "model"))

for name, param in model.named_parameters():
    if len(param.shape) == 2:
        xs.mark_sharding(param, spmd_mesh, ("model", "data"))

model(torch.randint(0, 50256, (1, 512)).to(device))

xm.mark_step()
```

To reiterate, this error occurs when running as a Python script, but not within a Jupyter Notebook.

### Details:
- **Environment**: Google Cloud v3-8 instance
- **Python Version**: 3.10
- **Libraries**: 
  - `torch` (nightly version after `20240527`)
  - `torch_xla` (nightly version after `20240527`)
- **Model**: LlamaForCausalLM from the `transformers` library

JackCaoG:
[...]

PR #7130, opened 2024-05-28
`mul`: convert inputs to result type.

ysiraichi:
Fix: #7084 

This PR fixes a data-type related problem for `mul` operation. It does so by creating a structure `OpConfig` that behaves similarly to `DoBinaryOp`. The difference is that it takes care of pre/post-processing of inputs and outputs, casting them to the correct data-type.

#### Problem

```python
t = torch.rand(10, dtype=torch.half).to(xm.xla_device())
s = torch.tensor(10, dtype=torch.double).to(xm.xla_device())
out = torch.mul_(t, s)
```

- `Tensor.mul_` is dispatched to its `CompositeExplicitAutograd` kernel
    - It wraps the scalar into a tensor, and calls `torch.mul` (functional version)
- `DoBinaryOp` is called
    - Computes `at::result_type` (let's call it `common_dtype`) and passes it on to `bin_op`
    - Note that `UnwrapNumber` does nothing, since `s` is a tensor with `is_wrapped_number_` unset
- The computed `common_dtype` is passed on to `tensor_methods::mul`
    - Creates an IR node with data-type `common_dtype`
    - Does nothing with its inputs
- Later, when `BuildMul` is called, we have 2 `XlaOp` with different data-types: `f16` and `f64`
    - `BuildMul` promotes `f16` to `f64`
- The output is `common_dtype` (`torch.float16`), but the actal `XlaOp` is `f64`

#### Solution

Following PyTorch behavior [[1], [2], [3]], I created `OpConfig`: a structure that let us specify common pre/post-processing on inputs and outputs.

#### Affected Models

- `timm_nfnet` (training+nondynamo)

cc @miladm @JackCaoG @lezcano 

[1]: https://github.com/pytorch/pytorch/blob/1be7e4086ad7c09e51c4c7854628a8adffb38f40/aten/src/ATen/native/BinaryOps.cpp#L166-L170  
[2]: https://github.com/pytorch/pytorch/blob/1be7e4086ad7c09e51c4c7854628a8adffb38f40/aten/src/ATen/TensorIterator.cpp#L994-L1000
[3]: https://github.com/pytorch/pytorch/blob/1be7e4086ad7c09e51c4c7854628a8adffb38f40/aten/src/ATen/TensorIterator.cpp#L979-L985

PR #7602, opened 2024-07-01
Reland: `mul`: convert inputs to result type.

ysiraichi:
Fix: #7266 

This PR tries to re-land #7130, which was reverted due to regressions on XLA:TPU. On top of that, it also fixes a subtle bug that went under the radar. Although we couldn't find a minimal reproducer, internal CI for TPU was reportedly broken due to this PR.

 Therefore, the idea is to have this PR merged, and check the internal CI for TPU results.

cc @miladm @vanbasten23 @JackCaoG

### DIFF
--- a/torch_xla/csrc/aten_xla_type.cpp
+++ b/torch_xla/csrc/aten_xla_type.cpp
@@ -158,7 +158,7 @@ class OpConfig {
         inputs.begin(), inputs.end(), [](const at::Tensor& tensor) {
           return bridge::TryGetXlaTensor(tensor);
         });
-    XLA_CHECK(it != inputs_.end());
+    XLA_CHECK(it != inputs.end());
     // Transform the inputs into a list of XLATensorPtr.
     // For that, either get their corresponding XLATensorPtr, or use the found
     // XLA tensor's BackendDevice for creating a new one.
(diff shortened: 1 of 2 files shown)


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -158,7 +158,7 @@ class OpConfig {
         inputs.begin(), inputs.end(), [](const at::Tensor& tensor) {
           return bridge::TryGetXlaTensor(tensor);
         });
-    XLA_CHECK(it != inputs_.end());
+    XLA_CHECK(it != inputs.end());
```
ysiraichi, line 161: Previously, incorrectly checking against `inputs_` instead of `inputs`.

JackCaoG: man! I should catch this during reviewing haha

JackCaoG (approved):

torch_xla/csrc/aten_xla_type.cpp
ysiraichi: Must admit that it was the best naming choice, though.

JackCaoG: nice catch 🤣 , I will check the TPU CI after this pr is merged.

ysiraichi: Again, sorry for the typo there. That was really bad.

Issue #7161, opened 2024-05-31
A large number of Tensors (>8000) in the graph will trigger an spmd sharding error


mars1248:
## To Reproduce

test.sh
```
rm -rf ./hlo_logs
#export XLA_FLAGS="--xla_gpu_enable_async_all_gather=true \

#export XLA_FLAGS="--xla_dump_to=./hlo_logs"
export XLA_FLAGS="--xla_dump_to=./hlo_logs \
    --xla_gpu_enable_analytical_latency_estimator=true \
    --xla_cpu_enable_fast_math=false \
    --xla_gpu_simplify_all_fp_conversions=false \
    --xla_gpu_force_compilation_parallelism=64  \
    --xla_gpu_enable_pipelined_collectives=true \
    --xla_gpu_enable_pipelined_all_reduce=true \
    --xla_gpu_enable_async_collectives=true \
    --xla_disable_hlo_passes=post-scheduling-passes,gpu-schedule-postprocessing \
    --xla_gpu_enable_triton_gemm=false \
"

export PJRT_ALLOCATOR_PREALLOCATE=false
export PJRT_ALLOCATOR_FRACTION=0.75
export PJRT_ALLOCATOR_CUDA_ASYNC=false
export PT_XLA_DEBUG=1
#export TF_CPP_MIN_LOG_LEVEL=0
#export TF_CPP_VMODULE="lazy_graph_executor=4,xla_graph_executor=5,nccl_collective_thunk=5,gpu_executable=5,gpu_compiler=5,service=5,collectives=5,xla_graph_executor=5,pjrt_computation_client=5,pjrt_stream_executor_client=5"
export XLA_IR_DEBUG=1
export XLA_HLO_DEBUG=1
export XLA_SAVE_TENSORS_FILE=debug.txt
#XLA_USE_SPMD=1 \
export GPU_NUM_DEVICES=8 \
export PJRT_DEVICE=CUDA \
#CUDA_VISIBLE_DEVICES=1,2,3,4 \
#python test_activation_local.py
python test_multi_param_layer.py
```
test_multi_param_layer.py
```
from typing import Dict, List, Optional, Tuple, Union
import torch
import torch_xla
import torch_xla.runtime as xr
import torch_xla.core.xla_model as xm
import torch_xla.experimental.xla_sharding as xs
from torch_xla.experimental.xla_sharding import Mesh
from torch_xla.amp import autocast, GradScaler
import numpy as np
import torch.optim as optim
[...]

PR #7604, opened 2024-07-01
Handle the parameter wrapping for SPMD

JackCaoG:
should fix https://github.com/pytorch/xla/issues/7161

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -117,6 +117,11 @@ function run_pt_xla_debug {
   PT_XLA_DEBUG=1 PT_XLA_DEBUG_FILE="/tmp/pt_xla_debug.txt" run_test "$@"
 }
 
+function run_parameter_warpping {
+  echo "Running in parameter wrapping mode: $@"
+  XLA_PARAMETER_WRAPPING_THREADSHOLD=1 run_test "$@"
+}
+
 function run_pt_xla_debug_level1 {
   echo "Running in save tensor file mode: $@"
   PT_XLA_DEBUG_LEVEL=1 PT_XLA_DEBUG_FILE="/tmp/pt_xla_debug.txt" run_test "$@"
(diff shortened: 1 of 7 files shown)


alanwaketan:
Let me know when it's ready for review.

JackCaoG:
not sure why TPU CI is skipped, but I think this pr is ready for review.

torch_xla/csrc/helpers.cpp
```diff
@@ -1007,6 +1018,26 @@ xla::StatusOr<xla::XlaComputation> XlaHelpers::WrapXlaComputation(
...
+      // we assume that parameter is ordered.
+      XLA_CHECK_EQ(index, param_shardings.size());
+      param_shardings.push_back(*xla::HloSharding::FromProto(instr.sharding()));
+      TF_VLOG(5) << "index = " << index
+                 << " shape = " << xla::Shape(instr.shape()).ToString()
+                 << " sharding = "
+                 << xla::HloSharding::FromProto(instr.sharding())->ToString();
+    }
+  }
+  return std::move(param_shardings);
```
alanwaketan, line 1038: I don't think we need to do std::move for return value.

test/spmd/test_spmd_parameter_wrapping.py
```diff
@@ -0,0 +1,56 @@
...
+    auto_wrap_policy = functools.partial(
+        transformer_auto_wrap_policy,
+        transformer_layer_cls={torch.nn.Linear},
+    )
+    linears = FSDPv2(linears, auto_wrap_policy=auto_wrap_policy)
+    input = torch.randn(100, 10, device=device)
+    output = linears(input)
+    torch_xla.sync()
+    xm.wait_device_ops()
+    self.assertEqual(output.shape, torch.Size([100, 40]))
```
alanwaketan, line 51: Is this check sufficient? Without this change, what will be the output shape?

JackCaoG: I was just lazy and want to make sure this does not crash haha. I can add a unit test for check the value.

alanwaketan: I guess my point is I cannot link the test to the change as the change attach the sharding to inputs but then you are checking the output.

alanwaketan (approved):

test/run_tests.sh
```diff
@@ -117,6 +117,11 @@ function run_pt_xla_debug {
   PT_XLA_DEBUG=1 PT_XLA_DEBUG_FILE="/tmp/pt_xla_debug.txt" run_test "$@"
 }
 
+function run_parameter_warpping {
+  echo "Running in parameter wrapping mode: $@"
+  XLA_PARAMETER_WRAPPING_THREADSHOLD=1 run_test "$@"
```
JackCaoG, line 122: now I think about it maybe it is better to set this env var as part of the test

JackCaoG:
Let me fix the review comments in a follow up pr so we can land this change for @alanwaketan to try.

PR #7605, opened 2024-07-02
Add blockwise quantized dot support

lsy323:
Add blockwise quantized dot support for 8-bit and 4-bit weight

Test:
- Added unit tests

### DIFF
--- a/test/quantized_ops/test_quantized_matmul.py
+++ b/test/quantized_ops/test_quantized_matmul.py
@@ -182,28 +182,31 @@ class QuantizedTest(unittest.TestCase):
     out_features = 8
     block_size = 2
     batch_size = 3
-    weight = torch.randint(
-        -8, 7,
-        (input_features // block_size, block_size, out_features)).to(torch.int8)
-    weight_scaler = torch.ones(input_features // block_size, out_features)
-    x = torch.rand(batch_size, input_features)
+    for n_bit in [4]:
+      with self.subTest(n_bit=n_bit):
+        weight = torch.randint(
+            -8, 7,
+            (input_features // block_size, block_size, out_features)).to(torch.int8)
+        weight_scaler = torch.ones(input_features // block_size, out_features)
+        x = torch.rand(batch_size, input_features)
 
-    # Fake quantize output.
-    w_dq = (weight * weight_scaler.unsqueeze(1)).reshape(
-        input_features, out_features)
-    fake_quant_out = torch.matmul(x, w_dq)
-    # Eager output.
-    torch_out = torch.ops.xla.quantized_matmul(
-        x, weight, weight_scaler, block_size=block_size)
-    # XLA Output.
-    x = x.to(device)
-    weight = weight.to(device)
-    weight_scaler = weight_scaler.to(device)
-    xla_out = torch.ops.xla.quantized_matmul(
-        x, weight, weight_scaler, block_size=block_size)
-    self.assertGreater(
-        self._calc_cosine_dist(fake_quant_out, torch_out), 0.99999)
-    self.assertTrue(torch.allclose(torch_out, xla_out.cpu()))
+        # Fake quantize output.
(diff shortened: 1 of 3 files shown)


ManfeiBai (approved):

torch_xla/experimental/xla_quantized_matmul.py
```diff
@@ -21,83 +21,132 @@ def _check_per_channel_quant_weight_dtype_shapes(input_dim, output_dim, w,
...
+    _check_per_channel_quant_weight_dtype_shapes(x.shape[-1], scaler.shape[0],
+                                                 w, scaler)
+    return F.linear(x, w) * scaler
+  else:
+    # Blockwise quant.
+    _check_blockwise_quant_weight_dtype_shapes(x.shape[-1], w.shape[-1],
+                                               block_size, w, scaler)
+    x = x.reshape(*x.shape[:-1], x.shape[-1] // block_size, block_size)
+    out = torch.einsum('scn,...sc->...sn', w, x)
+    out = torch.einsum('sn,...sn->...n', scaler, out)
```
JackCaoG, line 79: lol I have to admit I don't fully understand the math here but I will assume this is correct given test is passing

torch_xla/experimental/xla_quantized_matmul.py
```diff
@@ -21,83 +21,132 @@ def _check_per_channel_quant_weight_dtype_shapes(input_dim, output_dim, w,
...
     '''
-    Weight shape: [output_channel, input_channel]
-    Weight scaler shape: [output_channel]
+    Weight shape:
+      per-channel quant: [out_channel, in_channel].
+      block_wise quant: [in_channel / block_size, block_size, out_channel].
+
+    Weight scaler shape:
+      per-channel quant: [out_channel,].
+      blockwise quant: [in_channel / block_size, out_channel].
```
JackCaoG, line 130: nit, can fix in a follow up. I find it a bit confusing to use `wegiht shape` here because I thought it is a paramter. I think it is better to follow pytorch documentation pattern
```
weight (Tensor): for per-channel quant, it should be a tensor with shape [out_channel, in_channel] .....
```

JackCaoG (approved):

torch_xla/experimental/xla_quantized_matmul.py
lsy323: The internal doc go/ptxla-quantized-ops should explain the math.

torch_xla/experimental/xla_quantized_matmul.py
lsy323: sounds good, let me update.

lsy323:
> We should also have a doc to explain how to use these quantized ops under https://github.com/pytorch/xla/tree/master/docs

It's in https://github.com/pytorch/xla/blob/master/docs/quantized_ops.md

test/quantized_ops/test_quantized_matmul.py
```diff
@@ -146,6 +177,54 @@ def test_int4_per_channel_linear_module(self):
...
+        m = M(6, 8)
+        x = torch.randn(3, 6)
+        out_fp = m(x)
+        m.replace_with_xla_quantized_matmul(n_bit=8, block_size=2)
+        out_quant = m(x)
+        self.assertGreater(self._calc_cosine_dist(out_fp, out_quant), 0.99)
+
+        # Dot with int4 weight is only supported on TPU
+        if not (n_bit == 4 and xr.device_type() != 'TPU'):
+          m = m.to(device)
```
miladm, line 222: what's the behavior on `CUDA` device?

lsy323: Because int4 only runs on XLA:TPU

miladm: offline discussion summary: `int4` works only on XLA:TPU today; XLA:CPU does not support quant ops at the moment. XLA:GPU level of support is unclear as it is not tested currently.

PR #7609, opened 2024-07-02
Run paged_attention test with megacore mode only on TPU v4

bhavya01:


test/test_pallas.py
```diff
@@ -545,9 +545,10 @@ def test_paged_attention_wrapper(self):
             atol=1e-5,
             rtol=1e-5))
 
-  @unittest.skipIf(xr.device_type() != 'TPU' or tpu.version() < 4,
-                   "This test only works on TPUv4+.")
+  @unittest.skipIf(xr.device_type() != 'TPU' or tpu.version() != 4,
```
JackCaoG, line 548: is there a way to tell the difference between v5e and v5p?

bhavya01: I think we should be able to extract it using tpu.get_tpu_env() and checking for accelerator type.

JackCaoG (approved):

test/test_pallas.py
JackCaoG: for now we don't have v5p CI so this is probably fine

wonjoolee95 (approved):

PR #7610, opened 2024-07-02
Lower empty_like, empty_permute, empty_strided

lsy323:
JAX empty only return all zero tensors, so we cannot compare the numerical value with the `torch.empty` in the tests

PyTorch dispatcher only captures `aten.empty_permute` for `aten.empty_like`, so we lower `aten.empty_like` instead.

torch functions `torch.empty_like` calls into `aten.empty_like`, so lowering `aten.empty_permute` covers `torch.empty_like` as well.

Test
Added unit test

### DIFF
--- a/experimental/torch_xla2/test/test_functions.py
+++ b/experimental/torch_xla2/test/test_functions.py
@@ -30,15 +30,23 @@ class TestTorchFunctions(parameterized.TestCase):
       ('full_2d', lambda: torch.full((2, 3), 3.141592)),
       ('full_2d_dtype', lambda: torch.full(
           (2, 3), 3.141592, dtype=torch.float16)),
+      ('empty', lambda: torch.empty(2, 3), False),
+      ('empty_dtype', lambda: torch.empty(2, 3, dtype=torch.float16), False),
+      ('empty_like', lambda: torch.empty_like(torch.rand(2, 3)), False),
+      ('empty_like_dtype', lambda: torch.empty_like(torch.tensor([[1, 2], [3, 4]]), dtype=torch.float16), False),
   )
-  def test_tensor_constructor(self, func: Callable[[], torch.Tensor]):
+  def test_tensor_constructor(self, func: Callable[[], torch.Tensor], expect_same_value=True):
     expected = func()
 
     with self.env:
       actual = func()
       self.assertIsInstance(actual, torch_xla2.tensor.XLATensor2)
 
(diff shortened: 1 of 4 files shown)


lsy323:
cc @jonb377

will-cromar:
`empty` already exists: https://github.com/pytorch/xla/blob/8f5de773350095004297deae21886ec020f3fd8b/experimental/torch_xla2/torch_xla2/ops/jaten.py#L409-L412

and we skip the value check for some non-deterministic ops in the opinfo tests: https://github.com/pytorch/xla/blob/8f5de773350095004297deae21886ec020f3fd8b/experimental/torch_xla2/test/test_ops.py#L508-L509

`emtpy_like` is still missing though. I'm still fuzzy on the best practices for implementing an op through dispatch vs the function mode vs a decomposition tbh. Going off of what @qihqi has implemented recently, can you write empty_like as a decomposition like this, and then enable the opinfo test?

https://github.com/pytorch/xla/blob/8f5de773350095004297deae21886ec020f3fd8b/experimental/torch_xla2/torch_xla2/decompositions.py#L102-L104

lsy323:
> rand_like(self, **kwargs): 
>      dtype = kwargs.get('dtype') 
>      return torch.rand(s

@will-cromar Thanks for reminding on the existing lowering for `aten.empty`! The comments are indeed inspiring!

For `empty_like`, dispatcher only captures `aten.empty_permuted`. Torch function `torch.empty_like` also calls into `aten.empty_like`, so lowering `aten.empty_permuted` covers all cases.

If dispatcher can capture the `aten.empty_like`, I think we should register the decomp for `aten.empty_like`

My 2 cents on dispatch vs the function mode vs a decomposition:

If a torch function doesn't have a corresponding `aten` op (There is no dedicated aten op for the torch function, for example, `torch.tensor`). For such cases register under function mode makes more sense.

If an `aten` op can be composed by other aten ops, and direct lowering won't make the lowered graph cleaner. In this case we should register a decomposition, this also avoid duplicated code in `jaten.py`.

will-cromar:
> For empty_like, dispatcher only captures aten.empty_permuted.

Yeah, I noticed this too, which is weird because there is actually a dispatch key `torch.ops.aten.empty_like`. Thanks for implementing the other variants of `empty`!

For the other tensor constructors I added to `jtorch` (other than `torch.tensor()`), I'll actually go ahead and move them to `jaten` since I think they make more sense there.

will-cromar (approved):

PR #7627, opened 2024-07-03
Update jax to a stable version(0.4.29) for r2.4

bhavya01:
JAX nightly wheels are automatically deleted after a while. To avoid breaking pallas functionality, we need to depend on a stable JAX version.

0.4.29 release was created on June 10 and it is the closest to the openxla pin of June 12.


JackCaoG:
@bhavya01 were you able to manully verifeid that it works? If not try to rerun CI with TPU CI

JackCaoG (approved):

Issue #7342, opened 2024-06-24
OpenXLA CUDA fallback: Tracking Issue


ysiraichi:
PR #7318 introduced OpenXLA CUDA fallback. However, that's an opt-in feature that relies on the environment variable: `XLA_FALLBACK_CUDA`. Even though that works fine, as [this comment suggests](https://github.com/pytorch/xla/pull/7318#discussion_r1651377797), it doesn't provide a good UX. Therefore, we should improve it further by adding the following changes:

- [ ] Make OpenXLA CUDA fallback the default fallback, making the environment negating the environment variable to something like `DISABLE_XLA_FALLBACK_CUDA` #7630
- [ ] Deprecate the environment variable, by making PyTorch/XLA smart enough to pick the appropriate fallback

cc @miladm @JackCaoG @vanbasten23

ManfeiBai:
Hi, @vanbasten23, is that ok to assign this ticket to you?

ysiraichi:
I will take on this one. @vanbasten23 let me know if you are already working on any of them.

ysiraichi:
Closing this for now.
 
We don't actually do anything other than check if, in summary, (i) a CUDA implementation is available; and (ii) we are using a CUDA device. That said, it seems like there still isn't a need for anything other than that.

PR #7630, opened 2024-07-03
Make CUDA OpenXLA fallback the default.

ysiraichi:
Partially fix: #7342 

This PR changes the default device for running OpenXLA fallback operation from CPU to CUDA. So, instead of specifying `XLA_FALLBACK_CUDA=1`, in order to run fallback operations on CUDA the user must make sure that the following is true:

1. `XLA_FALLBACK_CPU` (newly introduced) is **not** set
2. The `DeviceType` of the current `ComputationClient` **is** CUDA
3. PyTorch was compiled with CUDA support

I have also changed `test_fallback` function so that it won't (un)set `XLA_FALLBACK_CPU` flag. Instead, it just runs the fallback operation. But, from this PR onwards, PyTorch/XLA should be able to detect when it can't do CUDA OpenXLA fallback automatically.

That said, one can force CPU fallback execution with the newly introduced `XLA_FALLBACK_CPU` environment variable.

cc @miladm @vanbasten23 @JackCaoG


torch_xla/csrc/aten_cuda_functions.cpp
```diff
@@ -11,6 +11,8 @@ static void fail(const char* name) {
 
 namespace c10::cuda {
 
+DeviceIndex device_count() noexcept { return 0; }
```
miladm, line 14: IIRC, we unit test this function in python layer - do we need a unit test for the cpp layer?

torch_xla/csrc/device.cpp
```diff
@@ -55,6 +55,10 @@ std::string DeviceType::toString() const {
   return absl::StrCat(type_name_, ":");
 }
 
+XlaDeviceType DeviceType::getType() const {
```
miladm, line 58: ditto

miladm (approved):
@ysiraichi 

+ please add `XLA_FALLBACK_CPU` to https://github.com/pytorch/xla/blob/master/configuration.yaml and explain its current/latest function

+ have you tested this PR on your torchbench runs to verify functionality?

approving to unblock - let's confirm the above items please.

miladm:
cc @zpcore

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -51,8 +52,33 @@ std::vector<std::string> GetFallbackOperations() {
 // Before each modified function below, we shall specify what has changed,
 // if there was any.
```
JackCaoG, line 54: nit, we can fix in a follow up. We should rename this file to `aten_fallback.cpp`

JackCaoG (approved):

zpcore:
> cc @zpcore

Thanks, looks like the fallback is enabled by default. We should see the performance update tomorrow.

torch_xla/csrc/device.cpp
ysiraichi: Do we need a unit test for this function?

ysiraichi:
As @miladm pointed out, I will still run benchmarks with this PR. So, this might get landed only tomorrow.

ysiraichi:
I confirmed this PR doesn't introduce any new regressions.

@miladm I will merge this PR, and add the `XLA_FALLBACK_CPU` description with the file renaming (suggested by @JackCaoG) in a follow-up PR.

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -51,8 +52,33 @@ std::vector<std::string> GetFallbackOperations() {
...
 
-bool UseCUDAFallback() {
-  return runtime::sys_util::GetEnvBool("XLA_FALLBACK_CUDA", false);
+// Decide whether to run OpenXLA fallback operations on CUDA.
+bool UseCUDAFallback(const c10::OperatorHandle& op) {
```
vanbasten23, line 56: Since we have another type of cuda fallback, how about the name "OpenXLAFallbackOnCUDA"

torch_xla/csrc/aten_cuda_functions.cpp
vanbasten23, line 14: It returns 0 because it's a phony implementation? 
Also, could you add a comment in this file describing it's phony implementation and what is the purpose of this file?

vanbasten23 (approved):
LGTM with minor comments

torch_xla/csrc/aten_cpu_fallback.cpp
```diff
@@ -15,11 +15,24 @@
...
+static std::unordered_set<std::string> _force_fallback_on_cpu{
+    // This operation is a simple memory access that transforms the given
+    // 1-element tensor into a Scalar.
+    //
+    // Although it makes sense to run this operation on CPU (since the
+    // output will get copied back to CPU anyway), this also fixes a
+    // particular issue with moco benchmark.
+    // More details: https://github.com/pytorch/xla/issues/7647
+    "aten::_local_scalar_dense",
+};
```
ysiraichi, line 34: Just to be completely transparent: `aten::_local_scalar_dense` is here for 2 reasons:
1. It just makes sense to run it on CPU, since the output (a `Scalar`) also lives on CPU
2. As a temporary fix to #7647

ysiraichi:
@miladm @JackCaoG @vanbasten23 @zpcore In case you want to take another look at this PR, I added the following changes:

- Check whether an operation has a CUDA kernel, before actually fallbacking to it
- Have a set with operations that should fallback on CPU
    - Temporary work-around #7647 

I will merge this one tomorrow.

PR #7631, opened 2024-07-04
fix eager mode spmd module loading with fsdpv2

JackCaoG:
The problem is `module.to` is always a `empty` + `to_copy`. in eager mode `empty` is a `expand` and will be evaluated directly. However this evulation will result in a `replicated ` sharding spec. In `fsdpv2` code, we incorrectly believe this tensor is intentionally replicated and fsdpv2 won't shard it.


alanwaketan (approved):
LGTM.

PR #7641, opened 2024-07-08
Move all docs under source for img to show

JackCaoG:
Let's see if this is going to work...

The problem is that `.md` files are in `docs`, but the `rst` files are under `docs/source`. When we `mdinclude` the md files, the path for image becomes incorrect.


JackCaoG:
well image can be seen now, but the index is kind of mess up.. let me fix

JackCaoG:
I think this change becomes too complicaed, if https://github.com/pytorch/xla/pull/7644 works I prefer to merge that one instead.

PR #7642, opened 2024-07-09
Use torch_xla.experimental.compile for all examples

JackCaoG:
Here is my plan, @will-cromar let me know what you think

1. ask users to use `torch_xla.experimental.compile`(and torch_xla.step()) to wrap their step fn which will mark_step the outside region for them
2. ask users to turn on eager mode later

Pretty much we can stage the effort for this ux migration. For this to work I need to make `torch_xla.step` handles eager mode, I will do that in a follow up pr.

### DIFF
--- a/examples/train_resnet_base.py
+++ b/examples/train_resnet_base.py
@@ -33,6 +33,7 @@ class TrainResNetBase():
     self.model = torchvision.models.resnet50().to(self.device)
     self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
     self.loss_fn = nn.CrossEntropyLoss()
+    self.compiled_step_fn = torch_xla.experimental.compile(self.step_fn)
 
   def _train_update(self, step, loss, tracker, epoch):
     print(f'epoch: {epoch}, step: {step}, loss: {loss}, rate: {tracker.rate()}')
@@ -40,16 +41,19 @@ class TrainResNetBase():
   def run_optimizer(self):
     self.optimizer.step()
 
+  def step_fn(self, data, target):
+    self.optimizer.zero_grad()
+    output = self.model(data)
+    loss = self.loss_fn(output, target)
+    loss.backward()
+    self.run_optimizer()
+
   def train_loop_fn(self, loader, epoch):
     tracker = xm.RateTracker()
     self.model.train()
     loader = itertools.islice(loader, self.num_steps)
     for step, (data, target) in enumerate(loader):
-      self.optimizer.zero_grad()
-      output = self.model(data)
-      loss = self.loss_fn(output, target)
-      loss.backward()
-      self.run_optimizer()
+      loss = self.compiled_step_fn(data, target)
       tracker.add(self.batch_size)
       if step % 10 == 0:
         xm.add_step_closure(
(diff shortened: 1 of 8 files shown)


JackCaoG:
@will-cromar I think this is ready for review

examples/train_resnet_base.py
```diff
@@ -33,23 +33,27 @@ def __init__(self):
     self.model = torchvision.models.resnet50().to(self.device)
     self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
     self.loss_fn = nn.CrossEntropyLoss()
+    self.compiled_step_fn = torch_xla.experimental.compile(self.step_fn)
```
will-cromar, line 36: I'm still not sure I understand why compile is separate from `torch_xla.step`. The main difference that I can see is that you flip eager off then on, which I think you can just add to `torch_xla.step`. `contextlib.contextmanager` already handles the rest of the plumbing you have there such as wrapping and exception handling.

https://github.com/pytorch/xla/blob/3078d163b2cd0c130d06f89041b6fbb46ce5798e/torch_xla/experimental/eager.py#L41-L59

vs 

https://github.com/pytorch/xla/blob/3078d163b2cd0c130d06f89041b6fbb46ce5798e/torch_xla/torch_xla.py#L57-L71

https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

https://docs.python.org/3/library/contextlib.html#contextlib.ContextDecorator

`step` already has a cautionary note in the docstring that we will be changing the public API over time.

JackCaoG: I guess I don't know how do you turn a context manger to a wrapper around the function? All I want was that api looks like
```
torch_xla.experimental.compile()
```
I think it can share the same implementation as the step.

JackCaoG: oh ok I guess you meant to use `@functools.wraps(func)` on top of the step.. let me try that..

JackCaoG: ok they are actually different, `step` is a context manager that doesn't take fn and arguments as input. It also doesn't return the fn's output back.

will-cromar: > ok they are actually different, step is a context manager that doesn't take fn and arguments as input. It also doesn't return the fn's output back.

Context managers and decorators are ~interchangable in python thanks to `ContextDecorator`:

```
@torch_xla.experimental.compile
def f(x):
  return 2 * x

@torch_xla.step()
def g(x):
  return 2 * x
```

The only semantic difference is you have to instantiate the context manager first (`step()`). It appropriately wraps the functions inputs, outputs, docstring, signature, types, etc too. If we want a "no arguments" form of `step`, then I can make that work. Even something silly like this works:

```
stepwithnoargs = torch_xla.step() # not recommended for readability lol

@stepwithnoargs
def g(x):
  """docstring"""
  return 2 * x
```

Python has almost exactly what we want in the standard library already, so let's use it.

JackCaoG: so many new python hacks haha, let me read into this a bit

JackCaoG: I understand how `decorators` and `context manager` are interchangeable now, I am still not sure how to implement the api I want `new_fn = torch.experimental.compile(func)` with the `decorator`, this is a bit mind twisting.  Are you going to make `step` takes an optional argument and branch from there? I will update `step` to handle `eager mode` and you can probably refactor from there.

will-cromar (approved):

PR #7644, opened 2024-07-09
Easier way to make img show up in our doc page

JackCaoG:
This might be an easier way to make image to show up.

The fundermental problem is

1. we want img to show up in markdown
2. we want img to show up in docs
3. we can only specify relative path for image import
4. `md` and `rst` are in different directory.

I had done multiple refactors but eventually I found it easier to just keep 2 copy of images...


JackCaoG:
ok this doesn't work.. only `_static` dirs got copied over, I can just rename the dirs

JackCaoG:
ok I checked that image can be properly shown now.

will-cromar (approved):

PR #7648, opened 2024-07-09
Introduce torch_xla.launch()

zpcore:
Draft of using `torch_xla.launch()` for `torchrun` and `xmp.spawn()`

Unify `_mp_fn(xu.getenv_as(xenv.LOCAL_RANK, int), FLAGS)` (call with torchrun) or   
`xmp.spawn(_mp_fn, args=(FLAGS,), nprocs=FLAGS.num_cores)`  
with 
```
torch_xla.launch(_mp_fn, args=(FLAGS,) , nprocs=FLAGS.num_cores)
```

### DIFF
--- a/examples/data_parallel/train_resnet_ddp.py
+++ b/examples/data_parallel/train_resnet_ddp.py
@@ -8,13 +8,15 @@ import torch.distributed as dist
 from torch.nn.parallel import DistributedDataParallel as DDP
 import torch.optim as optim
 import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla
+import torch_xla.core.xla_model as xm
 
 
 class TrainResNetDDP(TrainResNetBase):
 
   def __init__(self):
-    super().__init__()
     dist.init_process_group('xla', init_method='xla://')
+    super().__init__()
     self.model = DDP(
         self.model, gradient_as_bucket_view=True, broadcast_buffers=False)
     self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
@@ -26,5 +28,7 @@ def _mp_fn(index):
 
 
 if __name__ == '__main__':
-  print('consider using train_resnet_spmd_data_parallel.py instead to get better performance')
-  xmp.spawn(_mp_fn, args=())
+  print(
+      'consider using train_resnet_spmd_data_parallel.py instead to get better performance'
+  )
+  torch_xla.launch(_mp_fn)
--- a/torch_xla/torch_xla.py
+++ b/torch_xla/torch_xla.py
@@ -1,9 +1,14 @@
 import contextlib
-from typing import List
+from typing import Callable, List, Tuple
 
 import torch
+import torch.distributed as dist
 import torch_xla
 import torch_xla.core.xla_model as xm
+import torch_xla.core.xla_env_vars as xenv
+import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla.runtime as xr
+import torch_xla.utils.utils as xu
 
 
 def device(index: int = None) -> torch.device:
@@ -80,3 +85,27 @@ def manual_seed(seed, device=None):
       If missing the default device seed will be set.
   """
   xm.set_rng_state(seed, device)
+
+
+def launch(
+    fn: Callable,
+    args: Tuple = (),
+    start_method: str = 'spawn',
+    debug_single_process: bool = False,
+):
+  """ Entry to launch multiprocess.
+
+  Raises:
+    NotImplementedError: SPMD is not supported yet.
+  """
+  if xr.is_spmd():
+    # TODO(piz): SPMD is specified differently from mp. Skip for now.
+    raise NotImplementedError(
+        'launch function does not support SPMD at this time')
+
+  nprocs = 1 if debug_single_process else None
+
+  if dist.is_torchelastic_launched():
+    fn(xu.getenv_as(xenv.LOCAL_RANK, int), *args)
+  else:
+    xmp.spawn(fn, args=args, nprocs=nprocs, start_method=start_method)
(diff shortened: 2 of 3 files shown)


torch_xla/launch.py
```diff
@@ -0,0 +1,49 @@
...
+import torch.distributed as dist
+import torch_xla
+import torch_xla.core.xla_env_vars as xenv
+import torch_xla.distributed.xla_multiprocessing as xmp
+import torch_xla.utils.utils as xu
+import torch_xla.runtime as xr
+
+
+def spawn_wrapper(index, fn, *args, **kwargs):
+  dist.init_process_group('xla', init_method='xla://')
```
will-cromar, line 12: PyTorch's documentation shows `init_process_group` inside of the user's main function, e.g. https://pytorch.org/tutorials/intermediate/dist_tuto.html

We should stick to that pattern and not `init_process_group` ourselves to be safe.

torch_xla/launch.py
```diff
@@ -0,0 +1,49 @@
...
+  fn(index, *args, **kwargs)
+
+
+def launch(
+    fn,
+    *args,
+    **kwargs,
```
will-cromar, line 19: `spawn` does not actually take `kwargs`, so for consistency, you should not take `kwargs` for `fn` in this wrapper. https://pytorch.org/docs/stable/multiprocessing.html#spawning-subprocesses

Also, I don't recommend mixing variable-length arguments for both `launch` and `fn`. You can get around that by making `args` a singular tuple, similar to `spawn`.

torch_xla/launch.py
```diff
@@ -0,0 +1,49 @@
...
+
+  Raises:
+    NotImplementedError: SPMD is not supported yet.
+  """
+  if xr.is_spmd():
+    # TODO(piz): SPMD is specified differently from mp. Skip for now.
+    raise NotImplementedError(
+        'launch function does not support SPMD at this time')
+
+  nprocs = kwargs.pop('nprocs', None)
```
will-cromar, line 31: Put only the options that are implemented here directly as keyword arguments. Namely, `spawn_method` and `nprocs`

The only valid states of `nprocs` are 1 and `num_devices`. You can make the illegal states impossible by replacing it with a bool (e.g. `debug_single_process`) that determines whether we pass `1` to `xmp.spawn` or not.

`join` and `daemon` are not implemented in `xmp.spawn` and were just there to mirror torch's `spawn`.

examples/data_parallel/train_resnet_ddp.py
```diff
@@ -26,5 +27,7 @@ def _mp_fn(index):
 
 
 if __name__ == '__main__':
-  print('consider using train_resnet_spmd_data_parallel.py instead to get better performance')
-  xmp.spawn(_mp_fn, args=())
+  print(
+      'consider using train_resnet_spmd_data_parallel.py instead to get better performance'
+  )
+  torch_xla.launch(_mp_fn, nprocs=1)
```
will-cromar, line 33: Why `nprocs=1`?

torch_xla/launch.py
```diff
@@ -0,0 +1,49 @@
...
+import torch_xla.utils.utils as xu
+import torch_xla.runtime as xr
+
+
+def spawn_wrapper(index, fn, *args, **kwargs):
+  dist.init_process_group('xla', init_method='xla://')
+  fn(index, *args, **kwargs)
+
+
+def launch(
```
will-cromar, line 16: nit: put this in `torch_xla.py` with the other top-level functions if you can

examples/data_parallel/train_resnet_ddp.py
zpcore: I can leave it to nprocs=None as it is before.

torch_xla/launch.py
zpcore: Regarding `The only valid states of nprocs are 1 and num_devices` , is this also true for GPU?

torch_xla/launch.py
```diff
@@ -0,0 +1,46 @@
...
+  """
+  if xr.is_spmd():
+    # TODO(piz): SPMD is specified differently from mp. Skip for now.
+    raise NotImplementedError(
+        'launch function does not support SPMD at this time')
+
+  if debug_single_process:
+    nprocs = 1
+  else:
+    nprocs = xr.global_runtime_device_count()
```
will-cromar, line 36: Set this to `None` and `spawn` will do the right thing without initializing the runtime

torch_xla/launch.py
zpcore: Cleaned up, it's now much simpler.

examples/data_parallel/train_resnet_ddp.py
```diff
@@ -8,13 +8,15 @@
...
+import torch_xla
+import torch_xla.core.xla_model as xm
 
 
 class TrainResNetDDP(TrainResNetBase):
 
   def __init__(self):
-    super().__init__()
     dist.init_process_group('xla', init_method='xla://')
+    super().__init__()
```
will-cromar, line 19: nit: should the super init stay at the top?

torch_xla/torch_xla.py
```diff
@@ -80,3 +85,27 @@ def manual_seed(seed, device=None):
       If missing the default device seed will be set.
   """
   xm.set_rng_state(seed, device)
+
+
+def launch(
+    fn: Callable,
+    args: Tuple = (),
```
will-cromar, line 92: Can you add a TODO to make `args` a ParamSpec? I want to take a pass to improve our type annotations, but it'll involve adding `typing` backports as a dependency. `Tuple` is correct and works for this PR.

https://docs.python.org/3/library/typing.html#typing.ParamSpec

will-cromar (approved):
Nice! Thank you for implementing this :smile:

torch_xla/torch_xla.py
zpcore: Oh, I just pressed the merge without noticing the comment. Will update in another PR.

examples/data_parallel/train_resnet_ddp.py
zpcore: No, super().__init__() will call some functions from runtime, which needs init the TPU backend first.

PR #7672, opened 2024-07-12
support gmm as a custom op for dynamo

JackCaoG:
PyTorch actually will not run functionization inside a custom op. This means we should avoid using the in place ops for all of the custom op implementation. For this op I made `gmm` functional.

From @bdhirsh 
```
short answer is that functionalization runs on every "primitive" op (both in aten and custom) directly. so if it sees a custom op that is marked as not-mutating-its-inputs, it will assume that op is functional
```


qihqi (approved):

torch_xla/experimental/custom_kernel.py
```diff
@@ -7,6 +7,7 @@
 import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.distributed.spmd as xs
+import torch_xla.debug.metrics as met
```
alanwaketan, line 10: Do you add this in accident?

alanwaketan (commented):
LGTM.

torch_xla/experimental/custom_kernel.py
JackCaoG: lol forgot to delete

PR #7674, opened 2024-07-12
Lower aten::_linalg_eigh

tengyifei:
There's an xla::SelfAdjointEig function so we lower it to that.

I discovered that the XLA implementation of eigenvalue decomposition is not as numerically stable as numpy or torch, despite passing a small tolerance and large max_iter. The unit test thus uses a hardcoded tensor value copied from

https://android.googlesource.com/platform/external/tensorflow/+/f2a058296dd/tensorflow/compiler/xla/client/lib/self_adjoint_eig_test.cc#149


tengyifei:
cc @vanbasten23

test/cpp/test_aten_xla_tensor_2.cpp
```diff
@@ -511,6 +511,46 @@ TEST_F(AtenXlaTensorTest, TestLinalgVectorNormInDimsKeepDtype) {
...
+      torch::Tensor xla_input_reconstructed =
+          torch::mm(torch::mm(xla_eigenvectors,
+                              torch::diag(xla_eigenvalues)
+                                  .toType(c10::ScalarType::ComplexFloat)),
+                    xla_eigenvectors.t().conj());
+      AllClose(input_reconstructed, input);
+      AllClose(xla_input_reconstructed, input);
+    });
+  }
+  ExpectCounterNotChanged("aten::_linalg_eigh", cpp_test::GetIgnoredCounters());
```
JackCaoG, line 550: change this to `aten::.*` to be more general

torch_xla/csrc/ops/eigh.cpp
```diff
@@ -0,0 +1,59 @@
...
+          [&]() { return NodeOutputShape(input); },
+          /*num_outputs=*/2, torch::lazy::MHash(uplo)) {
+  XLA_CHECK(uplo == "L" || uplo == "U") << "Expected L or U, got: " << uplo;
+  uplo_ = uplo[0];
+}
+
+torch::lazy::NodePtr Eigh::Clone(torch::lazy::OpList operands) const {
+  return torch::lazy::MakeNode<Eigh>(operands.at(0),
+                                     c10::string_view{&uplo_, 1});
+}
```
JackCaoG, line 45: we don't need this anymore, can just delete it.

JackCaoG (approved):

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -741,6 +741,22 @@ at::Tensor& XLANativeFunctions::_index_put_impl_(
                                                    accumulate);
 }
 
+std::tuple<at::Tensor, at::Tensor> XLANativeFunctions::_linalg_eigh(
+    const at::Tensor& self, c10::string_view uplo, bool compute_v) {
+  TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  if (!compute_v) {
+    // Fallback to aten in case of `eigvalsh`, which does not compute
+    // eigenvectors but requires numerically stable gradients.
+    return at::native::call_fallback_fn<&xla_fallback,
```
vanbasten23, line 750: I understand we only need to lower torch.linalg.eigh. But in case we need to lower eigvalsh later, would `requires numerically stable gradients` be a blocker so we have to fall back to aten?

torch_xla/csrc/ops/eigh.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "torch_xla/csrc/torch_util.h"
+#include "xla/client/lib/self_adjoint_eig.h"
+
+namespace torch_xla {
+
+namespace {
+
+std::array<xla::XlaOp, 2> LowerImpl(xla::XlaOp input, bool lower) {
+  auto [eigenvectors, eigenvalues] =
+      xla::SelfAdjointEig(input, lower, /* max_iter */ 64, /* tol */ 1e-6);
```
vanbasten23, line 18: What are the max_iter and tol (for opaque) used for? 
Also, could you add a comment on why changing the default value?

vanbasten23 (approved):

torch_xla/csrc/ops/eigh.cpp
tengyifei: When testing I discovered that the default settings lead to a very low accuracy in the reconstructed matrix (e.g. let's say we decompose `A` to `A' = V @ Q @ V_T`, then `A` and `A'` have a difference in elements above 0.1.

After looking at what JAX does I think it can be simpler to align with JAX: https://github.com/google/jax/blob/a8b425cac50c842f66f36903dfb93fe6ad5a2a5b/jax/_src/lax/linalg.py#L726. Looks like they use the same tolerance but higher max_iter.

torch_xla/csrc/aten_xla_type.cpp
tengyifei: So I read the PyTorch docs again and I think I misunderstood it initially. What PyTorch doc suggests is that the gradients of the _eigenvectors_ are unstable. Therefore, if the user calls `eigvalsh`, they will only get _eigenvalues_ and thus the gradients will be stable. I removed this misleading comment.

If we want to support `eigvalsh` later the simplest way is probably discarding the eigenvectors from XLA and also figuring out what to return for the second `at::Tensor` tuple member.

Issue #7652, opened 2024-07-09
[API Usability] Internalize `parse_xla_device`


zpcore:
### Clean up torch_xla.core.xla_model.

Internalize `parse_xla_device`.

```[tasklist]
### Tasks
- [ ] Clean up and use upstream torch.distributed.
- [ ] Update the PyTorch/XLA documentation.
- [ ] Update test examples.
```

PR #7675, opened 2024-07-12
internalize _parse_xla_device

zpcore:
Resolves https://github.com/pytorch/xla/issues/7652

Along with some nit fixes.


torch_xla/core/xla_model.py
```diff
@@ -77,6 +77,8 @@ def is_xla_tensor(tensor):
 
 
 def parse_xla_device(device):
+  #TODO(zpcore): remove this function for release 2.5.
+  logging.warning("function parse_xla_device will be deprecated in release 2.5")
   m = re.match(r'([A-Z]+):(\d+)$', device)
```
JackCaoG, line 82: in here you can just call the internal version.

zpcore: Oh, nice catch!

will-cromar (commented):
Would the deprecation wrapper here work? https://github.com/pytorch/xla/blob/master/torch_xla/experimental/deprecation.py

zpcore:
> Would the deprecation wrapper here work? https://github.com/pytorch/xla/blob/master/torch_xla/experimental/deprecation.py

Also did some updates to the deprecation decorator function. Previously it will print the warning message whenever we  `import torch_xla`.

will-cromar:
> > Would the deprecation wrapper here work? https://github.com/pytorch/xla/blob/master/torch_xla/experimental/deprecation.py
> 
> Also did some updates to the deprecation decorator function. Previously it will print the warning message whenever we `import torch_xla`.

This was actually not written as a decorator. It essentially aliases a function from the new location to the deprecated location. We don't need to keep the old implementation around in most cases. Check out some of the old usage that Jack removed in https://github.com/pytorch/xla/pull/7579

zpcore:
> > > Would the deprecation wrapper here work? https://github.com/pytorch/xla/blob/master/torch_xla/experimental/deprecation.py
> > 
> > 
> > Also did some updates to the deprecation decorator function. Previously it will print the warning message whenever we `import torch_xla`.
> 
> This was actually not written as a decorator. It essentially aliases a function from the new location to the deprecated location. We don't need to keep the old implementation around in most cases. Check out some of the old usage that Jack removed in #7579

Oh, I see. Just fixed wrapper it back.

torch_xla/core/xla_model.py
```diff
@@ -40,6 +42,12 @@
 
 XLA_LIB = Library("xla", "DEF")
 
+aliases = [
+    iutils.parse_xla_device,
+]
+for alias in aliases:
+  register_deprecated(torch_xla.core, alias)
```
will-cromar, line 49: Thanks! I think this should actually be `torch_xla.core.xla_model` looking at the implementation of `register_deprecated`. Please double-check that you get the expected message when you call `torch_xla.core.xla_model.parse_xla-device(...)`

will-cromar (commented):
Otherwise LGTM

torch_xla/core/xla_model.py
zpcore: Updated to `parse_xla_device = deprecated(torch_xla.core, iutils.parse_xla_device)` instead.

will-cromar: The error message still says `torch_xla.core.parse_xla_device` instead of `torch_xla.core.xla_model.parse_xla_device`:

```
root@t1v-n-bf2f726f-w-0:/workspaces/ptxla# python
...
>>> xm.parse_xla_device('TPU:0')
WARNING:root:torch_xla.core.parse_xla_device is deprecated. Use torch_xla._internal.utils.parse_xla_device instead.
('TPU', 0)
```

You'll actually need to pass the `xla_model` module into `register_deprecated`. That was the only way I could find to print the qualified name of the aliased method consistently, but feel free to refactor it if you find a better way.

I think we need to add a unit test here to confirm that the right warning is getting printed, since you're going to be adding a lot of these aliases and this is an easy mistake to make. You can create a new unit test file for all of these public to internal aliases and we can remove it later.

torch_xla/core/xla_model.py
```diff
@@ -19,6 +19,8 @@
 import torch_xla.utils.utils as xu
 import torch_xla.utils.closures as xc
 import os
+from torch_xla.experimental.deprecation import deprecated
+import torch_xla._internal.utils as iutils
```
will-cromar, line 23: I'm not sure about `iutils` as the name of this module. IMO `_utils` is more idiomatic.  Another option is to move all of `torch_xla.utils.utils` to `_internal` and continue to use `xu` as the name for our internal utilities. We can put that off to a future PR.

torch_xla/core/xla_model.py
zpcore: This should work:
```
import torch_xla._internal.utils as _utils
from . import xla_model as this_module
parse_xla_device = deprecated(this_module, _utils.parse_xla_device)
```

torch_xla/core/xla_model.py
zpcore: Use _utils for now. Will relocate later.

will-cromar (approved):
Thanks!

Issue #7614, opened 2024-07-02
Dynamo persistent cache real-time look-up


wonjoolee95:
## 🚀 Feature
As described in https://github.com/pytorch/pytorch/issues/125958, we are integrating with vLLM on TPUs. We see that in the warm up phase of the vLLM, it needs to pre-compile ~30 different input shape combinations. PyTorch/XLA does not support dynamic shapes today so torch.compile will keep compiling the model code which slows down the development speed (waiting for 10 minutes before warm up is finished). PyTorch/XLA already cache the XLA compilation but torch.compile itself is pretty expensive.

This feature request pitches to  achieve the similar effect of dynamic shapes by persistent caching and real time look up of the compiled program.

## Details
To do this, in high-level, we need to do the following:
- Turn on the dynamo dynamic shape mode, dynamo will start passing the inputs with dynamic shapes to PyTorch/XLA
- PyTorch/XLA can then try to figure out if this shape is compiled in XLA
- If it is, we can map the different input shape to different compiled binaries

## Open questions
- Does persistent FxGraph caching work with PyTorch/XLA? Details at https://github.com/pytorch/pytorch/issues/125958#issuecomment-2204040977. 
- How can we properly map the different input shape to different compiled binaries?

cc @JackCaoG @WoosukKwon

JackCaoG:
couple thing you need to check 
1. whether parameters are always passed as input to the `opt_mode` function. It used to be that only the data passed but dynamo expect backend to remember the parameters of the model. We did a mapping in https://github.com/pytorch/xla/blob/7ec577847a1f29f91e88e05bcab9e9f2c6f5cbad/torch_xla/core/dynamo_bridge.py#L495. If parameters are always passed then this is easier, we can just do a hashing on the parameter shapes and use that as the key to find the corresponding hashs for the compiled program.
2. Is the hash for the compiled binary the only thing we need to switch between different shapes. do we need a copy of https://github.com/pytorch/xla/blob/7ec577847a1f29f91e88e05bcab9e9f2c6f5cbad/torch_xla/core/dynamo_bridge.py#L433-L435 for every compiled binary? If so how do we obey it?


I think you can start play with it by turn on the dynamic mode(it is disabled for XLA somewhere in upstream I think) and see what kind of fx graphs and inputs we get.

wonjoolee95:
[...]

PR #7676, opened 2024-07-12
Add support for dynamic shape in dynamo

wonjoolee95:
Fixes https://github.com/pytorch/xla/issues/7614

---

TODO
- Remove debugging code and add comments 
- Add unit tests
- Handle error case when TorchDynamo passes us int types

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -362,7 +362,6 @@ class DynamoInferenceBasicTest(parameterized.TestCase):
     met.clear_all()
     dynamo_resnet18 = torch.compile(device_resnet18, backend='openxla')
     for data, _ in loader:
-      # dynamo_resnet18 = torch.compile(device_resnet18, backend='openxla')
       output = dynamo_resnet18(data)
       output_cpu = resnet18(data.cpu())
       self.assertTrue(
@@ -375,7 +374,6 @@ class DynamoInferenceBasicTest(parameterized.TestCase):
     self.assertEqual(
         met.metric_data('RunCachedGraphOutputData')[0], sample_count)
 
-  @skipOnTpu
   @parameterized.parameters(
       True,
       False,
@@ -408,14 +406,17 @@ class DynamoInferenceBasicTest(parameterized.TestCase):
         'DynamoExtractCompiledGraph')
 
     loader_new_shape = self.get_loader(device, sample_count, batch_size=2)
-    diffs = []
     for data, _ in loader_new_shape:
       output_new_shape = dynamo_resnet18(data)
       output_cpu_new_shape = resnet18(data.cpu())
       # # TPU has some precision issues, skipping allclose check
-      # if not _is_on_tpu():
-      #   self.assertTrue(
-      #       torch.allclose(output_cpu_new_shape, output_new_shape.cpu(), rtol=1e-05, atol=1e-05))
+      if not _is_on_tpu():
+        self.assertTrue(
+            torch.allclose(
+                output_cpu_new_shape,
+                output_new_shape.cpu(),
+                rtol=1e-05,
+                atol=1e-05))
 
     self.assertEqual(
         met.counter_value('DynamoExtractCompiledGraph'),
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -769,6 +769,51 @@ class DynamoOperationsTests(test_utils.XlaTestCase):
 
     self.assertEqual(expected, actual.cpu())
 
+class DynamoDynamicShapeTest(unittest.TestCase):
+
+  def simple_add(self, a, b):
+    c = a + b
+    return c
+
+  def test_dynamic_shape(self):
+    met.clear_all()
+    device = xm.xla_device()
+
+    compiled_fn = torch.compile(
+        self.simple_add, backend="openxla", fullgraph=True, dynamic=True)
+    
+    a_cpu = torch.randn(3, 4)
+    a_xla = a_cpu.to(device)
+    b_cpu = torch.ones(4)
+    b_xla = b_cpu.to(device)
+    print(f'--------------------')
+    res_cpu = compiled_fn(a_cpu, b_cpu)
+    res_xla = compiled_fn(a_xla, b_xla)
+    print(f'[Testing] {res_cpu=}')
+    print(f'[Testing] {res_xla=}')
+    self.assertTrue(torch.all(torch.eq(res_cpu, res_xla.cpu())))
+    print(f"{met.metric_data('CompileTime')[0]=}")
+    print(f'--------------------')
+
+    c_cpu = torch.randn(5, 6)
+    c_xla = c_cpu.to(device)
+    d_cpu = torch.ones(6)
+    d_xla = d_cpu.to(device)
+    print(f'--------------------')
+    res_cpu_2 = compiled_fn(c_cpu, d_cpu)
+    res_xla_2 = compiled_fn(c_xla, d_xla)
+    print(f'[Testing] {res_cpu_2=}')
+    print(f'[Testing] {res_xla_2=}')
+    self.assertTrue(torch.all(torch.eq(res_cpu_2, res_xla_2.cpu())))
+    print(f"{met.metric_data('CompileTime')[0]=}")
+    print(f'--------------------')
+
+    print('Finished!')
+
+    # self.assertTrue(torch.all(torch.eq(t.cpu(), torch.tensor([10, 11, 12]))))
+
+    # compile_count = met.metric_data('CompileTime')[0]
+
 
 if __name__ == '__main__':
   test = unittest.main()
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -334,7 +334,7 @@ class DynamoInferenceBasicTest(parameterized.TestCase):
             atol=1e-05))
 
   def get_loader(self, device, sample_count, batch_size=4):
-    batch_size = xu.getenv_as('BATCH_SIZE', int, batch_size)
+    batch_size = xu.getenv_as('BATCH_SIZE', int, defval=batch_size)
     loader = xu.SampleGenerator(
         data=(torch.randn(batch_size, 3, 224, 224, device=device),
               torch.zeros(batch_size, dtype=torch.int64, device=device)),
@@ -349,7 +349,7 @@ class DynamoInferenceBasicTest(parameterized.TestCase):
   def test_resnet18(self, initialize_on_cuda):
     device = self._choose_proper_device(initialize_on_cuda)
     sample_count = xu.getenv_as('SAMPLE_COUNT', int, defval=10)
-    loader = self.get_loader(device, sample_count)
+    loader = self.get_loader(device, sample_count, batch_size=4)
     resnet18 = torchvision.models.resnet18()
     resnet18.eval()
     device_resnet18 = torchvision.models.resnet18()
@@ -399,15 +399,27 @@ class DynamoInferenceBasicTest(parameterized.TestCase):
     for data, _ in loader:
       output = dynamo_resnet18(data)
       output_cpu = resnet18(data.cpu())
-      self.assertTrue(
-          torch.allclose(output_cpu, output.cpu(), rtol=1e-05, atol=1e-05))
+      # TPU has some precision issues, skipping allclose check
+      if not _is_on_tpu():
+        self.assertTrue(
+            torch.allclose(output_cpu, output.cpu(), rtol=1e-05, atol=1e-05))
 
-    loader_new_shape = self.get_loader(device, sample_count, batch_size=8)
-    for data, _ in loader:
-      output = dynamo_resnet18(data)
-      output_cpu = resnet18(data.cpu())
-      self.assertTrue(
-          torch.allclose(output_cpu, output.cpu(), rtol=1e-05, atol=1e-05))
+    previous_extract_compile_count = met.counter_value(
+        'DynamoExtractCompiledGraph')
+
+    loader_new_shape = self.get_loader(device, sample_count, batch_size=2)
+    diffs = []
+    for data, _ in loader_new_shape:
+      output_new_shape = dynamo_resnet18(data)
+      output_cpu_new_shape = resnet18(data.cpu())
+      # # TPU has some precision issues, skipping allclose check
+      # if not _is_on_tpu():
+      #   self.assertTrue(
+      #       torch.allclose(output_cpu_new_shape, output_new_shape.cpu(), rtol=1e-05, atol=1e-05))
+
+    self.assertEqual(
+        met.counter_value('DynamoExtractCompiledGraph'),
+        previous_extract_compile_count)
 
   def test_resnet18_lazy_vs_dynamo(self):
     sample_count = xu.getenv_as('SAMPLE_COUNT', int, defval=10)
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -333,8 +333,8 @@ class DynamoInferenceBasicTest(parameterized.TestCase):
             rtol=1e-05,
             atol=1e-05))
 
-  def get_loader(self, device, sample_count):
-    batch_size = xu.getenv_as('BATCH_SIZE', int, defval=4)
+  def get_loader(self, device, sample_count, batch_size=4):
+    batch_size = xu.getenv_as('BATCH_SIZE', int, batch_size)
     loader = xu.SampleGenerator(
         data=(torch.randn(batch_size, 3, 224, 224, device=device),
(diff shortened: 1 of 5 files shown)


wonjoolee95:
With the current changes, the following code generates correct results without recompiling the graph:
```
    ###
    # torch.compile dynamic shape ON
    torch._dynamo.config.automatic_dynamic_shapes = True
    compiled_fn = torch.compile(fn, backend='openxla', dynamic=True)
    a = torch.randn(3, 4, device=device)
    b = torch.ones(4, device=device)
    ret = compiled_fn(a, b)
    xm.mark_step()
    print(f'[Testing] {ret=}')
    print(f'--------------------')

    c = torch.randn(4, 5, device=device)
    d = torch.ones(5, device=device)
    ret2 = compiled_fn(c, d)
    xm.mark_step()
    print(f'[Testing] {ret2=}')
    print(f'--------------------')
```
As for next steps, I'll clean up some code and add some unit tests.

test/dynamo/test_dynamo.py
```diff
@@ -769,6 +769,35 @@ def foo(x):
 
     self.assertEqual(expected, actual.cpu())
 
+class DynamoDynamicShapeTest(unittest.TestCase):
+
+  def simple_add(self, a, b):
+    c = a + b
+    return c
```
JackCaoG, line 776: we need a more real-world case. Can you try the resnet18 and uses different batch size?

wonjoolee95: Good point, updated! Any suggestions on what metrics we want to assert on? With this change, now we directly call `optimized_mod` when shape (batch size, in this case) changes. However, we still call `extract_graph_helper` in `optimized_mod` which will trigger the compile.

JackCaoG:
seems like a bunch of test failed and a lot of them are real failures. @wonjoolee95 let me know if you need help debugging them

test/dynamo/test_dynamo.py
```diff
@@ -374,6 +375,40 @@ def test_resnet18(self, initialize_on_cuda):
     self.assertEqual(
         met.metric_data('RunCachedGraphOutputData')[0], sample_count)
 
+  @skipOnTpu
```
JackCaoG, line 378: don't skip on TPU, just don't do the `allclose` check, you can still check the counter etc.

torch_xla/core/dynamo_bridge.py
```diff
@@ -430,13 +430,22 @@ def extract_internal(xla_model: torch.fx.GraphModule):
         print(torch_xla._XLAC._get_xla_tensor_debug_info(xla_arg))
   # Don't reset the scope as we might be under some profiler trace scope.
   xm.mark_step(reset_scope=False)
+
+  # [Note: Dynamo real-time input-shape cache look-up]
+  # We maintain a mapping of input shapes to outputs of extract_graph_helper.
+  # When dynamic=True in torch.compile call, TorchDynamo will not trigger a
+  # new recompile. Then TorchXLA needs to figure out if these input shapes have
+  # been seen before.
+  input_shape_mappings = {}
```
JackCaoG, line 439: let's add a type hint here to make it more clear

torch_xla/core/dynamo_bridge.py
```diff
@@ -448,6 +457,35 @@ def optimized_mod(*args: tuple):
...
+    # When dynamic=True in torch.compile call, TorchDynamo will directly
+    # call optimized_mod without compiling. Hence, xla_model's xla_args will
+    # be different from optimized_mod's args. So we manually set them here.
+    xla_model.xla_args = args
+
+    # See [Note: Dynamo real-time input-shape cache look-up] above.
+    arg_input_shapes = []
+    for arg in args:
+      if isinstance(arg, torch.Tensor):
+        arg_input_shapes.append(tuple(arg.shape))
```
JackCaoG, line 469: can we only do this if `dynamic=True`?

test/dynamo/test_dynamo.py
wonjoolee95: Updated

torch_xla/core/dynamo_bridge.py
wonjoolee95: Added

torch_xla/core/dynamo_bridge.py
wonjoolee95: Updated

test/dynamo/test_dynamo.py
```diff
@@ -374,6 +374,51 @@ def test_resnet18(self, initialize_on_cuda):
...
+        'DynamoExtractCompiledGraph')
+
+    loader_new_shape = self.get_loader(device, sample_count, batch_size=2)
+    for data, _ in loader_new_shape:
+      output_new_shape = dynamo_resnet18(data)
+      output_cpu_new_shape = resnet18(data.cpu())
+      # # TPU has some precision issues, skipping allclose check
+      # if not _is_on_tpu():
+      #   self.assertTrue(
+      #       torch.allclose(output_cpu_new_shape, output_new_shape.cpu(), rtol=1e-05, atol=1e-05))
```
wonjoolee95, line 416: This part is odd. When I run these tests, the `allclose` fails because in some iteration of the data loader with this new_shape, the differences are as big as `0.2`.

test/dynamo/test_dynamo.py
JackCaoG: the test is still marked with `@skipontpu`

torch_xla/core/dynamo_bridge.py
```diff
@@ -447,6 +459,33 @@ def optimized_mod(*args: tuple):
     nonlocal dumb_return_handler
     nonlocal xla_args_need_update
     nonlocal skip_checking_input_sharding_threashold
+    nonlocal input_shape_mappings
+
+    if not torch._dynamo.config.assume_static_by_default:
```
JackCaoG, line 464: do we need to cache this value in `extracted_internal`? for example if user do 
```
torch.compile(, dynamic=True)
torch.compile(, dynamic=False)
```
what will happen?

wonjoolee95: That's a good point. According to my testing, it seems like the flag to the first `torch.compile` call determines the value of `torch._dynamo.config.assume_static_by_default` throughout the program.

Ex. 1: `dynamic=False` in the first `torch.compile` call
Code:
```
    compiled_add_fn = torch.compile(fn, backend='openxla', dynamic=False)
    a = torch.randn(3, 4, device=device)
    b = torch.ones(4, device=device)
    c = compiled_add_fn(a, b)
    print(f'{c=}')

    compiled_add_fn_2 = torch.compile(simple_sub, backend='openxla', dynamic=True)
    d = torch.randn(3, 4, device=device)
    e = torch.ones(4, device=device)
    f = compiled_add_fn(d, e)
    print(f'{f=}')

    a2 = torch.randn(3, 4, device=device)
    b2 = torch.ones(4, device=device)
    c2 = compiled_add_fn(a2, b2)
    print(f'{c2=}')
```
Result: the value of `torch._dynamo.config.assume_static_by_default` is False for all 3 `optimized_mod` calls

Ex. 2: `dynamic=True` in the first `torch.compile` call
Code:
```
    compiled_add_fn = torch.compile(fn, backend='openxla', dynamic=True)
    a = torch.randn(3, 4, device=device)
    b = torch.ones(4, device=device)
    c = compiled_add_fn(a, b)
    print(f'{c=}')

    compiled_add_fn_2 = torch.compile(simple_sub, backend='openxla', dynamic=False)
    d = torch.randn(3, 4, device=device)
    e = torch.ones(4, device=device)
    f = compiled_add_fn(d, e)
    print(f'{f=}')

    a2 = torch.randn(3, 4, device=device)
    b2 = torch.ones(4, device=device)
    c2 = compiled_add_fn(a2, b2)
    print(f'{c2=}')
```
Result: the value of `torch._dynamo.config.assume_static_by_default` is True for all 3 `optimized_mod` calls

This looks like the behavior of `torch._dynamo.config`, as the values are initialized and set when the first `torch.compile` is called. I don't expect this to be a common scenario, but I can follow up with Edward in the PyTorch GitHub issue about this.

wonjoolee95: Ok, I found out why my test above is behaving wrong -- it was an error in my test script.

To come back to the original question:
> Do we need to cache this value in extract_internal? For example if user do
> torch.compile(, dynamic=True)
> torch.compile(, dynamic=False)
> What will happen?
We don't need to cache `torch._dynamo.config.assume_static_by_default`. This value is per `torch.compile` call. On your example, the second `torch.compile` will correctly have `torch._dynamo.config.assume_static_by_default=True`.

test/dynamo/test_dynamo.py
wonjoolee95: This is fixed with the explicit `mark_step` call within `else` statement under `torch._dynamo.config.assume_static_by_default`.

test/dynamo/test_dynamo.py
```diff
@@ -374,6 +374,54 @@ def test_resnet18(self, initialize_on_cuda):
...
+            torch.allclose(output_cpu, output.cpu(), rtol=1e-05, atol=1e-05))
+
+    previous_extract_compile_count = met.counter_value(
+        'DynamoExtractCompiledGraph')
+
+    loader_new_shape = self.get_loader(device, sample_count, batch_size=2)
+    for data, _ in loader_new_shape:
+      output_new_shape = dynamo_resnet18(data)
+      output_cpu_new_shape = resnet18(data.cpu())
+      # # TPU has some precision issues, skipping allclose check
```
JackCaoG, line 412: remove one #

test/dynamo/test_dynamo.py
```diff
@@ -374,6 +374,54 @@ def test_resnet18(self, initialize_on_cuda):
...
+      if not _is_on_tpu():
+        self.assertTrue(
+            torch.allclose(
+                output_cpu_new_shape,
+                output_new_shape.cpu(),
+                rtol=1e-05,
+                atol=1e-05))
+
```
JackCaoG, line 420: maybe also check the `CompileTime` and `ExecuteTime` here

torch_xla/core/dynamo_bridge.py
```diff
@@ -430,9 +439,32 @@ def extract_internal(xla_model: torch.fx.GraphModule):
...
+  # [Note: Dynamo real-time input-shape cache look-up]
+  # We maintain a mapping of input shapes to outputs of extract_graph_helper.
+  # When dynamic=True in torch.compile call, TorchDynamo will not trigger a
+  # new recompile. Then TorchXLA needs to figure out if these input shapes have
+  # been seen before.
+  # Keys: tuple of input shapes
+  # Values: tuple of (xla_args_sharding_spec, args_and_out, graph_hash,
+  # arg_index_to_need_update_index, none_remover, graph_input_matcher,
+  # dumb_return_handler, xla_args_need_update).
+  input_shape_mappings: dict[tuple[int, ...], tuple[object, ...]] = {}
```
JackCaoG, line 452: ust `typing.Dict` and `typing.Tuple` otherwise the python 3.8 CI in upstream will fail

torch_xla/core/dynamo_bridge.py
```diff
@@ -447,6 +479,27 @@ def optimized_mod(*args: tuple):
...
+    if not torch._dynamo.config.assume_static_by_default:
+      xla_model.xla_args = args
+      arg_input_shapes = _get_arg_input_shapes(xla_model.xla_args)
+      if arg_input_shapes in input_shape_mappings:
+        (xla_args_sharding_spec, args_and_out, graph_hash,
+         arg_index_to_need_update_index, none_remover, graph_input_matcher,
+         dumb_return_handler,
+         xla_args_need_update) = input_shape_mappings[arg_input_shapes]
+      else:
+        xm.mark_step()
```
JackCaoG, line 494: add a comment here to explain why `mark_step` is needed

torch_xla/core/dynamo_bridge.py
```diff
@@ -447,6 +479,27 @@ def optimized_mod(*args: tuple):
...
+        (xla_args_sharding_spec, args_and_out, graph_hash,
+         arg_index_to_need_update_index, none_remover, graph_input_matcher,
+         dumb_return_handler,
+         xla_args_need_update) = extract_graph_helper(xla_model)
+        input_shape_mappings[arg_input_shapes] = (
+            xla_args_sharding_spec, args_and_out, graph_hash,
+            arg_index_to_need_update_index, none_remover, graph_input_matcher,
+            dumb_return_handler, xla_args_need_update)
```
JackCaoG, line 502: I think you don't need this here

torch_xla/core/dynamo_bridge.py
```diff
@@ -529,6 +583,10 @@ def __init__(self, module):
     self._unsupported_nodes = []
 
   def run_node(self, n: torch.fx.Node):
+    # We need to restore this metric count later, so save it in a separate variable
+    dynamo_extract_graph_helper_metric_count = metrics.counter_value(
+        'DynamoExtractCompiledGraph')
```
JackCaoG, line 588: will `run_node` call `extract_compiled_graph` too?

JackCaoG:
@ysiraichi FYI

torch_xla/core/dynamo_bridge.py
wonjoolee95: Added

torch_xla/core/dynamo_bridge.py
wonjoolee95: IIUC, we actually need this here. And we actually don't need this same logic in `extract_internal` above (removed this in the newest commit). The reason is when `dynamic=True`, only `optimized_mod` is called. Other functions (including `extract_internal`) are not called.

torch_xla/core/dynamo_bridge.py
wonjoolee95: It's hard to see from documentations. However, when I try comparing metrics before/after `run_node`, from what I can see, it's not calling `extract_compiled_graph`.

torch_xla/core/dynamo_bridge.py
JackCaoG: ok then you will run into the same old problem right?
first time
```
extract_graph_helper -> optimized_mod
```
in this case you do the compile, but you do not cache the `input_shape_mappings `

when `optimized_mod ` is called the first tiem you will need to call `extract_graph_helper` again which is wasteful.

JackCaoG: you should just do the caching(input_shape_mappings[arg_input_shapes] =) inside the `extract_graph_helper `

test/dynamo/test_dynamo.py
JackCaoG: also can you make another test to test the case of
```
fn(shape_a)
fn(shape_b)
fn(shape_c)
fn(shape_a)
```
want to make sure we don't forgot the old shapes that's cached.

torch_xla/core/dynamo_bridge.py
JackCaoG: ok then I am confused what this `dynamo_extract_graph_helper_metric_count ` is doing here

wonjoolee95: This code (`run_node`) is executed when we're fetching the fallback ops. And in this code below, we clear our metric counters via `metrics.clear_counters()`. So we need a way to restore this counter, so we can verify `extract_compiled_graph` only gets called once in our unit tests.

JackCaoG: Ah I see, I can fix it later. I think the right thing to do is to define a region where counter does not incremented.

wonjoolee95:
The PR should be in a reasonable state, now just seeing 2 failures the `GPU tests requiring torch CUDA` tests:
```
#1: DynamoInferenceBasicTest.test_dynamic_shape_resnet180 (True):
Input tensor is not an XLA tensor: CUDAFloatType

#2: DynamoInferenceBasicTest.test_resnet180 (True)
  File "/__w/xla/xla/pytorch/xla/test/dynamo/test_dynamo.py", line 370, in test_resnet18
    self.assertEqual(met.metric_data('CompileTime')[0], 1)
TypeError: 'NoneType' object is not subscriptable
```
For the first error, the stack trace points to: https://github.com/pytorch/xla/blob/4ba63ffd87098968ef314f0b19256b5a2f32404f/torch_xla/core/dynamo_bridge.py#L292-L295 

It seems like we may want to do an additional `isinstance(arg, torch.Tensor)` check here.

JackCaoG:
I will pick this up and try to fix error today

torch_xla/core/dynamo_bridge.py
JackCaoG: let me fix this too..

JackCaoG:
@alanwaketan There are a few places I want to fix but maybe we should just merge this pr to unblock Woosuk now. I am also running some benchmarks

alanwaketan (approved):
Approved to unblock.

PR #7679, opened 2024-07-12
deprecate xrt_world_size

zpcore:
Deprecate `torch_xla.xla_model.xrt_world_size` and use `torch_xla.runtime.world_size` instead.

Add the `run_once` decorator to function `runtime.using_pjrt` since we only need to run this once per process. This helps get rid of dynamo compilation issue with xm.all_reduce.


will-cromar (commented):
Can you use `xr.world_size` instead? I don't think we need to rename the function in `xla_model`

https://github.com/pytorch/xla/blob/5b8e8e01a8436e3b54bb43560385de9e04ca3af6/torch_xla/runtime.py#L148-L152

zpcore:
> Can you use `xr.world_size` instead? I don't think we need to rename the function in `xla_model`
> 
> https://github.com/pytorch/xla/blob/5b8e8e01a8436e3b54bb43560385de9e04ca3af6/torch_xla/runtime.py#L148-L152

`xrt_world_size` uses `_WORLD_SIZE` besides `runtime.world_size()`. There is a concern that removing `_WORLD_SIZE` will cause recompilation.  @alanwaketan for insights.

alanwaketan:
You will know if there is a recompilation from the test.

zpcore:
Hi @will-cromar , I tried functools.lru_cache and it crashes in multiprocess. I notice that if we use the functools.warps and assign attributes to the functions to be wrapped, it will cause crash. Probably lru_cache uses the func attributes in this case.

I create `run_once` and moved global var into runtime.py in order to deprecate xrt_world_size.

torch_xla/runtime.py
```diff
@@ -12,10 +12,41 @@
...
+  # Dynamo doesn't support XRT or multithreaded runtime. See Note [V3-8 Threading]
+  if not runtime.using_pjrt() or runtime.addressable_device_count() > 1:
+    return
+
+  if _WORLD_SIZE is None:
+    _WORLD_SIZE = runtime.world_size()
+    _ORDINAL = runtime.global_ordinal()
+
+
+def run_once(func):
```
will-cromar, line 39: What is this `run_once` for? It's a neat idea, but I see you opted for global variables for world size and ordinal

torch_xla/runtime.py
```diff
@@ -70,6 +101,7 @@ def device_type() -> Optional[str]:
   return pjrt_device.split('_')[0] if pjrt_device else pjrt_device
 
 
+@run_once
 def using_pjrt() -> bool:
```
will-cromar, line 105: Hah, this function also needs to get deprecated since I assume this is always `True`

torch_xla/runtime.py
zpcore: Without the `@run_once` for `using_pjrt`, the test `test_mp_replication` will fail with build the dynamic graph for dynamo compile.

torch_xla/runtime.py
zpcore: We still need to call `__maybe_select_default_device()`, this is the point why I call using_pjrt() only once.

torch_xla/runtime.py
will-cromar: Do you know why that is? Is it because now all calls to get e.g. `world_size` go through functions wrapped in `requires_pjrt`, which in turn actually is checking an env var (`device_type` and `_maybe_select_default_device`)? Whereas before the call would have been stopped by `xm.world_size`.

will-cromar (commented):
General question: if `run_once` makes dynamo happy for `using_pjrt`, do you know why it does't work for `world_size` and `global_ordinal`?

torch_xla/runtime.py
```diff
@@ -12,10 +12,41 @@
...
+  if not runtime.using_pjrt() or runtime.addressable_device_count() > 1:
+    return
+
+  if _WORLD_SIZE is None:
+    _WORLD_SIZE = runtime.world_size()
+    _ORDINAL = runtime.global_ordinal()
+
+
+def run_once(func):
+  result = []
```
will-cromar, line 40: nit: please use a `nonlocal` value instead of a list here https://stackoverflow.com/a/1261961

I just learned about this too :smile:

torch_xla/runtime.py
will-cromar, line 39: Can you move `run_once` to the internal utilities? It's a handy utility for us, but we don't need to make it part of our API

zpcore:
> General question: if `run_once` makes dynamo happy for `using_pjrt`, do you know why it does't work for `world_size` and `global_ordinal`?


Take unittest `test/test_mp_replication.py` as the example.
- Why it work for `using_pjrt`:
  Without `run_once`, we will see this function get called: https://github.com/pytorch/xla/blob/42aa7bdc3444b567f92e2069a01fcadd599a9b30/torch_xla/runtime.py#L36-L38  
  , which checks for the env var and makes it dynamic. Adding the using_pjrt, the function has already been executed and got skipped when we do dynamo compile.
  
- Why we need global var like `_WORLD_SIZE` (actually `_ORDINAL` seems not necessary):
  Without it, we will see error like:
  ```
  from user code:
   File "/home/piz/pytorch/xla/test/test_mp_replication.py", line 11, in all_reduce
    return xm.all_reduce(xm.REDUCE_SUM, tensor)
  File "/home/piz/pytorch/xla/torch_xla/core/xla_model.py", line 428, in all_reduce
    if runtime.world_size() == 1 and not xu.getenv_as('XLA_ALWAYS_ALLREDUCE',
  File "/home/piz/pytorch/xla/torch_xla/runtime.py", line 128, in wrapper
    return fn(*args, **kwargs)
  File "/home/piz/pytorch/xla/torch_xla/runtime.py", line 186, in world_size
    if torch_xla._XLAC._xla_get_replication_devices_count() == 0:

Set TORCH_LOGS="+dynamo" and TORCHDYNAMO_VERBOSE=1 for more information
  ```
  The C API binding ` _xla_get_replication_devices_count` will be called and failed the dynamo compile.

torch_xla/runtime.py
zpcore: Check my reply here: https://github.com/pytorch/xla/pull/7679#issuecomment-2246520735

will-cromar (approved):
Thanks for the explanation! Filed a follow-up bug to clean up `use_pjrt` and `requires_pjrt` and fix a concrete usability issue at https://github.com/pytorch/xla/issues/7730

JackCaoG:
TPU CI failure seems relevant, can we fix forward or revert this pr?

zpcore:
I forgot to update the TPU CI test. Let me make a follow up PR now.

docs/amp.md
```diff
@@ -77,7 +77,7 @@ for input, target in data:
     # Exits the context manager before backward pass
     scaler.scale(loss).backward()
     gradients = xm._fetch_gradients(optimizer)
-    xm.all_reduce('sum', gradients, scale=1.0 / xm.xrt_world_size())
+    xm.all_reduce('sum', gradients, scale=1.0 / xr.world_size())
```
miladm, line 80: ditto

docs/ddp.md
```diff
@@ -96,7 +97,7 @@ def demo_basic(rank):
     # xla specific APIs to get rank, world_size.
     new_rank = xm.get_ordinal()
     assert new_rank == rank
-    world_size = xm.xrt_world_size()
+    world_size = xr.world_size()
```
miladm, line 100: ditto

docs/ddp.md
```diff
@@ -36,7 +36,7 @@ dist.init_process_group("xla", rank=rank, world_size=world_size)
 
 ```
 new_rank = xm.get_ordinal()
-world_size = xm.xrt_world_size()
+world_size = xr.world_size()
```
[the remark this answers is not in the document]
zpcore: `xm.world_size()` is also deprecated. They all point to the same thing. We should only use `xr.world_size()`.

PR #7680, opened 2024-07-12
Optimize execution for ops that have multiple output in eager mode

JackCaoG:
In eager mode the execution happens when we create an XLATensor with `IR`, we will use the `IR` as the root to build/execute the graph. 

This is mostly fine but for ops that has multiple outputs(like `native_batch_norm`), most of the outputs share a good amounts of common HLOs. It will be much faster to execute all of them in a single graph. The eager mode in PyTorch/XLA can't really execute HLO one by one, so the goal is to execute once(ideally) for each pytorch op.

The change in this pr will
1. delay the eager execution for some ops when they creating new XLAtensor with IRs
2. execute the HLO for all XLAtensors after they are created.

I will take another round to check I didn't mess up anything but would appreciate if someone can look closely at my change inside `tensor_method.cpp`.

### DIFF
--- a/torch_xla/csrc/tensor_methods.cpp
+++ b/torch_xla/csrc/tensor_methods.cpp
@@ -587,8 +587,14 @@ std::vector<XLATensorPtr> custom_call(
   std::vector<XLATensorPtr> outputs;
   outputs.reserve(output_shapes.size());
   for (size_t i = 0; i < output_shapes.size(); ++i) {
-    outputs.push_back(
-        inputs[0]->CreateFrom(torch::lazy::Value(node, i), output_dtypes[i]));
+    outputs.push_back(inputs[0]->CreateFrom(torch::lazy::Value(node, i),
+                                            output_dtypes[i],
+                                            /*delay_eager_executation=*/true));
+  }
+  XLAGraphExecutor* graph_executor = XLAGraphExecutor::Get();
+  if (graph_executor->UseEagerMode()) {
+    // Execute the HLO that will run the `customcall` and in one graph
+    graph_executor->ApplyEagerSync(outputs);
   }
   return outputs;
 }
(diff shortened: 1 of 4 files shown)


torch_xla/csrc/tensor_methods.cpp
```diff
@@ -747,12 +775,24 @@ void adam_optimizer_step_(const XLATensorPtr& found_inf, XLATensorPtr& step,
...
   if (amsgrad) {
-    max_exp_avg_sq->SetInPlaceIrValue(torch::lazy::Value(node, 4));
+    max_exp_avg_sq->SetInPlaceIrValue(torch::lazy::Value(node, 4),
+                                      /*delay_eager_executation=*/true);
+  }
+  XLAGraphExecutor* graph_executor = XLAGraphExecutor::Get();
+  if (graph_executor->UseEagerMode()) {
+    // Execute the HLO that will run the `adam_optimizer_step_` and in one hlo
+    std::vector<XLATensorPtr> tensors_to_sync = {step, param, exp_avg,
+                                                 exp_avg_sq, max_exp_avg_sq};
```
JackCaoG, line 794: should conditionally add `max_exp_avg_sq `

JackCaoG:
I also intentionally didn't handle the collectives. Collective will return a all_reduce token which we actually don't want to execute in eager case. I will handle that in a separate pr.

aws-rhsoln:
Curious how much perf boost do we expect when we fuse them into a single graph?

JackCaoG:
> Curious how much perf boost do we expect when we fuse them into a single graph?

for a test code
```
torch_xla.experimental.eager_mode(True)

device = torch_xla.device()
m = nn.BatchNorm2d(16).to(device)
m.train()
input = torch.randn(16, 16, 1024, 1024, device=device)

start = time.time()
for _ in range(20):
  input = m(input)
xm.wait_device_ops()
end = time.time()
duration = end - start
print(f"total time = {duration}")
```
without my change `total time = 0.46190381050109863`, with this change `total time = 14.28174352645874`. I actually don;t know why it is 28x faster, but I did verified that in HLO without my change `BatchNorm2d` will compute the result one by one.

JackCaoG:
@alanwaketan @wonjoolee95 This one is ready for review.

wonjoolee95 (approved):

PR #7686, opened 2024-07-15
Extend timeout for `tpu-info` test

will-cromar:
Init can take up to ~14 seconds on v2 in my testing. Extend the test timeout for TPU init up to 20 seconds to be safe. This should fix flakiness in this test on v2 and v5lite.

Remove extra env var settings that are not needed on recently-created TPUs.


will-cromar:
Tested these changes locally on a v2-8.

JackCaoG (approved):

PR #7696, opened 2024-07-16
[torch_xla2] Fix some duplicate and incorrect op registrations

will-cromar:
Fix a few op issues I've found while working on DDP (e.g. #7651) and `privateuse1` dispatch:

- Incorrect op registration under `torch.full` (instead of `torch.ops.aten.full`)
- `torch.ops.aten.copy` takes two arguments
- Duplicates for `torch.ops.aten.to.device`, `torch.ops.aten.max_pool2d_with_indices_backward`, `torch.ops.aten._local_scalar_dense`, `torch.ops.aten.tensor_split.sections`
- Forward `to.dtype_layout` to `to.dtype`

Also add warning to catch duplicates going forward.


will-cromar:
Oops, switching this back to draft while I fix the test failures.

qihqi (approved):

PR #7699, opened 2024-07-16
Update tuple type annotation

JackCaoG:
This should unblock pin update in https://github.com/pytorch/pytorch/pull/130771


ManfeiBai (approved):
LGTM

PR #7700, opened 2024-07-17
Add docs for eager mode

JackCaoG:


docs/eager.md
```diff
@@ -0,0 +1,122 @@
...
+import torchvision
+
+device = torch_xla.device()
+model = torchvision.models.resnet18().to(device)
+input = torch.randn(64, 3, 224, 224).to(device)
+
+# model tracing
+res = model(input)
+
+# model execution, you can also use `xm.mark_step`
```
will-cromar, line 20: Do you think we should be mentioning `mark_step` in new documentation? Ultimately we should deprecated it IMO

docs/eager.md
```diff
@@ -0,0 +1,122 @@
...
+```
+Note that
+
+1. Currently user has to manually enable the eager mode by `torch_xla.experimental.eager_mode(True)`.
+2. The region of the code that wants to be compiled should be wrapped by `torch_xla.experimental.compile`.
+
+The implementation of the `torch_xla.experimental.compile` is actually pretty straight forward, it disable the eager mode when entering the target function and start tracing. It will call the `torch_xla.sync()` when target function returns and reenable the eager mode. You can expect the same perfomrance by using the `eager` + `compile` API compared to the existing `mark_step/sync` approach.
+
+
+### Infernce
```
will-cromar, line 58: infernce -> inference

docs/eager.md
```diff
@@ -0,0 +1,122 @@
...
+def step_fn(self, data, target):
+    self.optimizer.zero_grad()
+    logits = self.model(data)
+    loss = self.loss_fn(
+        logits.view(-1, self.config.vocab_size), target.view(-1))
+    loss.backward()
+    optimizer.step()
+    return loss
+
+self.compiled_step_fn = torch_xla.experimental.compile(self.step_fn)
```
will-cromar, line 79: nit: remove `self` from this code sample because your function is not defined as a class method

will-cromar, line 79: re: naming, I don't think `compiled_step_fn` makes sense here. It makes it look like this function is imperatively compiled when you call `compile`, when what's really happening is that you will trace and JIT-compile the function when it's called later. `jit` and `trace` already mean something else in this ecosystem though...

docs/eager.md
JackCaoG: my thinking is most people don't know what `torch_xla.sync` is at this point. We should at least mention it so they know `mark_step` and `sync` are the same thing.

docs/eager.md
JackCaoG: well that's the trade of I made when I try to align the api with the upstream, I think it creates more confusion to called it `jited_step_fn`. I will just rename it to `step_fn`

will-cromar: Do you want to just use `compile` as a decorator here if you're following the upstream convention?

will-cromar (approved):

docs/eager.md
JackCaoG: seem like upstream support both, I can support both too https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html

PR #7701, opened 2024-07-17
fix type annotation for 3.8

JackCaoG:


JackCaoG:
upstream update is green https://github.com/pytorch/pytorch/pull/130896, this one is ready for review.

qihqi (approved):

zpcore (approved):
Thanks for the fix!

Nit: We can use `from typing import OrderedDict` to be consistent with other typings.

JackCaoG:
> Thanks for the fix!
> 
> Nit: We can use `from typing import OrderedDict` to be consistent with other typings.

There is a naming conflict I believe, because we also do
```
from collections import OrderedDict
```

will-cromar (approved):
Thanks!

PR #7703, opened 2024-07-17
add depreacated decorator

zpcore:
Introduce the `mark_deprecated` decorator to map deprecated function to the new function. The reason is that existing deprecated function only maps to function with the same name. It's impossible to map function like `xrt_world_size` to `world_size`.

Instead of passing old function to the deprecation wrapper, I think use the decorator above the old function is the most straightward.

Example to use:
```
import torch_xla
from torch_xla.experimental.deprecation import mark_deprecated

@mark_deprecated(torch_xla.runtime.world_size)
def xrt_world_size(defval=1):
  """Retrieves the number of devices which is taking part of the replication.

  Args:
    defval (int, optional): The default value to be returned in case there is no
      replication information available.
      Default: 1

  Returns:
    The number of devices which is taking part of the replication.
  """
  global _WORLD_SIZE
  if _WORLD_SIZE is not None:
    return _WORLD_SIZE

  return runtime.world_size()
```

Output:
```
>>> import torch_xla
>>> torch_xla.core.xla_model.xrt_world_size()
WARNING:root:torch_xla.core.xla_model.xrt_world_size is deprecated. Use torch_xla.runtime.world_size instead.
1
```

### DIFF
--- a/torch_xla/experimental/deprecation.py
+++ b/torch_xla/experimental/deprecation.py
@@ -25,8 +25,8 @@ def mark_deprecated(module, new: FN) -> FN:
   """Decorator to mark a function as deprecated and map to new function.
 
   Args:
-    module: current module of the deprecated function is in. Assume current module name is X, you can use `from . import X` and pass X here.
-    new: new function that we map to. Need to include the path the new functioin is in.
+    module: current module of the deprecated function that is in. Assume current module name is X, you can use `from . import X` and pass X here.
+    new: new function that we map to. Need to include the path the new function that is in.
 
   Returns:
     Wrapper of the new function.
--- a/torch_xla/experimental/deprecation.py
+++ b/torch_xla/experimental/deprecation.py
@@ -21,5 +21,34 @@ def deprecated(module, new: FN) -> FN:
   return wrapped
 
 
+def mark_deprecated(module, new: FN) -> FN:
+  """Decorator to mark a function as deprecated and map to new function.
+
+  Args:
+    module: current module of the deprecated function is in. Assume current module name is X, you can use `from . import X` and pass X here.
+    new: new function that we map to. Need to include the path the new functioin is in.
+
+  Returns:
+    Wrapper of the new function.
+  """
+  already_warned = [False]
+
+  def decorator(func):
+
+    @functools.wraps(new)
+    def wrapped(*args, **kwargs):
+      if not already_warned[0]:
+        logging.warning(
+            f'{module.__name__}.{func.__name__} is deprecated. Use {new.__module__}.{new.__name__} instead.'
+        )
+        already_warned[0] = True
+
+      return new(*args, **kwargs)
+
+    return wrapped
+
+  return decorator
+
+
 def register_deprecated(module, new: FN):
   setattr(module, new.__name__, deprecated(module, new))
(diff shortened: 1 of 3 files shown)


will-cromar:
The decorator makes sense here, but not because `It's impossible to map function like xrt_world_size to world_size.`. For example, see how these three functions in `experimental.pjrt` were renamed after being deprecated:

https://github.com/pytorch/xla/blob/b1f2167db09fad5642c4f4e70465d335e927817c/torch_xla/experimental/pjrt.py#L28-L31

You would need the decorator if the _implementation_ of the old function is actually different than the new one. In this case, the signatures of `xrt_world_size` (`(int) -> int`) and `world_size` (`() -> int`) differ.

Having said that, you can make `mark_deprecated` [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Remember that a decorator is a function that takes one function as an argument and returns a function. If you use `functools.partial` to fill the first argument of `deprecated` (`(module, fn) -> fn`), the partial callable is `(fn) -> fn` (ie a decorator. For any module, you can create a decorator like this:

```
mark_deprecated = functools.partial(deprecated, this_module)
```

Or, you can put this idea to work in `experimental.deprecation` to reduce the duplicated logic in your implementation:

```
def mark_deprecated(module):
  def f(func):
    return deprecated(module, func)
  return f
```

zpcore:
> device_attributes

Thanks for the point.  It's mostly about the warning message. If we did something like:
```
 rendezvous = deprecated(this_module, xm.xla_rendezvous) 
```
It will warn like `this_module.xla_rendezvous is deprecated with xm.xla_rendezvous`, while there is no function named `xla_rendezvous` originally in this module.

will-cromar:
> It will warn like this_module.xla_rendezvous is deprecated with xm.xla_rendezvous, while there is no function named xla_rendezvous originally in this module.

I see, good catch. It looks like `deprecated` is missing a parameter to print the correct warning then. You can add something like this:

```
def deprecated(module, new_fn, old_name=None):
  old_name = old_name or newfn.__name__
  ...
     logging.warning(f'{module.__name__}.{old_name} is deprecated. Use {new.__module__}.{new.__name__} instead.)
  ...
```

torch_xla/experimental/deprecation.py
```diff
@@ -21,5 +22,22 @@ def wrapped(*args, **kwargs):
   return wrapped
 
 
+def mark_deprecated(module, new: FN) -> FN:
```
will-cromar, line 25: Thanks! Can you add a test?

torch_xla/experimental/deprecation.py
```diff
@@ -21,5 +22,22 @@ def wrapped(*args, **kwargs):
...
+  Args:
+    module: current module of the deprecated function that is in. Assume current module name is X, you can use `from . import X` and pass X here.
+    new: new function that we map to. Need to include the path the new function that is in.
+
+  Returns:
+    Wrapper of the new function.
+  """
+
+  def decorator(func):
+    return deprecated(module, new, old_name=func.__name__)
```
will-cromar, line 37: Since you have a reference to both the new and old functions here, does `func.__module__` work in place of `module`? No worries if it doesn't work

zpcore: Nice catch! Updated to __module__.

torch_xla/experimental/deprecation.py
zpcore: Added

will-cromar (approved):
Thanks!

PR #7708, opened 2024-07-17
[Backport] Support reduce-scatter in manual sharding

alanwaketan:
Summary:
This PR is to add experimental support of cc ops in manual sharding zones. This one adds reduce-scatter as the initial step. The key here is to add channel_id, replica_groups, and use_global_device_ids in the lowering.

Test Plan:
PJRT_DEVICE=TPU XLA_USE_SPMD=1 python test/spmd/test_xla_sharding.py -v -k test_spmd_reduce_scatter

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1213,6 +1213,46 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
     self.assertEqual(xxx.shape, (8, 8))
     self.assertTrue(torch.allclose(x.cpu() + 1, xxx.cpu()))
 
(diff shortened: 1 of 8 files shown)


JackCaoG:
Release data is next week, I really don't think we should backport more feature prs.

alanwaketan:
I forgot this didn't get pick up. It's right after the cutoff. Sorry.

alanwaketan:
It's fine I can just ask users to try MoE on nightly. cc @JackCaoG

JackCaoG:
Thanks! I also felt like it will be fine as you are still actively adding features for MOE, nightly will be a better fit anyway.

PR #7709, opened 2024-07-17
Unify the logics to check eager mode

JackCaoG:
Prior to my eager mode changes, eager mode is a debug only mode that's controlled by the `XLA_USE_EAGER_DEBUG_MODE ` env var, now since we want to make it more offical I want to control it explictly with the API instead of the env var. @aws-rhsoln let me know what you think. Eventually I want to deprecate the env var(at least not call it debug mode).

### DIFF
--- a/torch_xla/__init__.py
+++ b/torch_xla/__init__.py
@@ -243,6 +243,10 @@ if os.getenv('XLA_REGISTER_INSTALLED_PLUGINS',
   plugins.use_dynamic_plugins()
   plugins.register_installed_plugins()
 
+if os.getenv('XLA_USE_EAGER_DEBUG_MODE', '0') == '1':
+  from .experimental import eager_mode
+  eager_mode(True)
+
 from .torch_xla import *
 
 # register all custom kenels and decomp by default
--- a/torch_xla/csrc/tensor.cpp
+++ b/torch_xla/csrc/tensor.cpp
@@ -86,8 +86,7 @@ XLATensorPtr XLATensor::Create(
       XLATensor(std::move(ir_value), device, logical_element_type));
   XLAGraphExecutor* graph_executor = XLAGraphExecutor::Get();
   graph_executor->RegisterTensor(xtensor->data());
-  if ((UseEagerDebugMode() || graph_executor->UseEagerMode()) &&
-      !delay_eager_executation) {
+  if ((graph_executor->UseEagerMode()) && !delay_eager_executation) {
     std::vector<XLATensorPtr> xtensors({xtensor});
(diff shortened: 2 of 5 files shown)


aws-rhsoln:
Do you think its valuable to add a message letting the users know that eager mode run without torch.compile is going to be slow, can provide some basic reasoning as to why that would be the case. 
Need to let users know that graph based mode is still going to be faster (be it using compile or lazy eval).

JackCaoG:
@aws-rhsoln check my doc at https://github.com/pytorch/xla/pull/7700, I am still thinking about how to message users better to let them know eager mode exists. but yea I agree that we have to make it clear that compile is the way to go for perfomrance.

JackCaoG:
@aws-rhsoln for some reason I can't make you a reviewer, do you mind also taking a look at this pr?

torch_xla/__init__.py
```diff
@@ -243,6 +243,10 @@ def _init_xla_lazy_backend():
   plugins.use_dynamic_plugins()
   plugins.register_installed_plugins()
 
+if os.getenv('XLA_USE_EAGER_DEBUG_MODE', '0') == '1':
```
aws-rhsoln, line 246: Is this just for backward compatibility, since we already have the api now?

torch_xla/csrc/tensor.cpp
```diff
@@ -86,8 +86,7 @@ XLATensorPtr XLATensor::Create(
       XLATensor(std::move(ir_value), device, logical_element_type));
   XLAGraphExecutor* graph_executor = XLAGraphExecutor::Get();
   graph_executor->RegisterTensor(xtensor->data());
-  if ((UseEagerDebugMode() || graph_executor->UseEagerMode()) &&
-      !delay_eager_executation) {
+  if ((graph_executor->UseEagerMode()) && !delay_eager_executation) {
```
aws-rhsoln, line 89: Nit: Extra brackets around `graph_executor->UseEagerMode()`

aws-rhsoln (approved):
Just one nit, lgtm otherwise! Thanks!

alanwaketan (approved):
LGTM.

torch_xla/__init__.py
JackCaoG: yea, it is mostly for backward compatibility. I think I can also tweak it and use it in test to test both eager and non-eager.

torch_xla/csrc/tensor.cpp
JackCaoG: nice catch

PR #7712, opened 2024-07-19
Add op support from `nan_to_num` to `movedim`

miladm:
Add op support for `nan_to_num`, `msqrt`, `mul`, `mv`, `movedim`


ManfeiBai (approved):
Thanks, LGTM

PR #7718, opened 2024-07-19
Fix the SPMD doc strng

JackCaoG:
SPMD example doc string is not being displayed correctly in https://pytorch.org/xla/master/index.html, I tried to follow upstream pattern(https://github.com/pytorch/pytorch/blob/main/torch/autograd/functional.py#L291-L312) and see if I can fix it...


JackCaoG:
image
 examples are displayed correctly

will-cromar (approved):

Issue #7697, opened 2024-07-16
Invalid version identifier in filenames of nightly builds


fellhorn:
## 🐛 Bug

[pip 24.1 deprecated legacy version identifiers](https://pip.pypa.io/en/stable/news/#deprecations-and-removals) and no longer allows installing the current nightly wheels directly. Other python package managers, like e.g. `uv` never supported these identifiers and always required renaming the wheel.

Additionally the version identifier in the wheel is different than the one in the filename.

### uv

```sh
> uv pip install torch_xla@https://storage.googleapis.com/pytorch-xla-releases/wheels/cuda/12.1/torch_xla-nightly-cp311-cp311-linux_x86_64.whl
error: The wheel filename "torch_xla-nightly-cp311-cp311-linux_x86_64.whl" has an invalid version part: expected version to start with a number, but no leading ASCII digits were found
```

### pip

Broken:
```sh
> pip install pip==24.1.2
...
> pip install https://storage.googleapis.com/pytorch-xla-releases/wheels/cuda/12.1/torch_xla-nightly-cp311-cp311-linux_x86_64.whl
ERROR: Invalid requirement: 'torch-xla==nightly': Expected end or semicolon (after name and no valid version specifier)
    torch-xla==nightly
             ^
```

For others potentially finding this issue and need a workaround:

:green_circle:   Works with `torch_xla@` format
```sh
> pip install pip==24.1.2
...
> pip install torch_xla@https://storage.googleapis.com/pytorch-xla-releases/wheels/cuda/12.1/torch_xla-nightly-cp311-cp311-linux_x86_64.whl
...
Installing collected packages: torch_xla
Successfully installed torch_xla-2.5.0+git41d998d
```

:green_circle:   Works with older pip versions

```sh
> pip install "pip<24"
...
> pip install https://storage.googleapis.com/pytorch-xla-releases/wheels/cuda/12.1/torch_xla-nightly-cp311-cp311-linux_x86_64.whl
...
Installing collected packages: torch_xla
Successfully installed torch_xla-2.5.0+git41d998d
```

## Expected behavior

I would expect the version identifier in the filename to match the one in the wheel and be a valid identifier. This should allow installation with `uv` and modern `pip` versions.

## Potential solutions

Ideas:

* Use last release + day identifier like `torch_xla-2.5.0+nightly20240716-cp311-cp311-linux_x86_64.whl`
[...]

PR #7721, opened 2024-07-22
Build CUDA plugin before `torch` and `torch_xla` in nightly build

will-cromar:
The nightly CUDA package build is failing with this error:

```
Step #2 - "build_xla_docker_image":       Error in repository_rule: invalid repository name '@pypi_torch_nightly+20240722_cp311_cp311_linux_x86_64_whl'
```

AFAICT this is happening because the `build_srcs` step is producing incorrectly-named wheels (see #7697). The GPU plugin build has zero dependency on PyTorch at all -- it builds completely separately in our CI. I believe it is unfortunately getting pulled in transitively through OpenXLAs build rules: https://github.com/pytorch/xla/blob/5ea4535f0699f366adb554183a65ebf7dc34a8be/WORKSPACE#L100

Move the plugin build to happen first as a temporary fix since there's no dependency between these steps.

Triggering the nightly build manually to see if this works

### DIFF
--- a/infra/ansible/Dockerfile
+++ b/infra/ansible/Dockerfile
@@ -8,6 +8,8 @@ RUN pip install ansible
 COPY . /ansible
 
 ARG ansible_vars
+# HACK: install build dependencies only, but skip build step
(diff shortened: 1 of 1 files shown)


will-cromar:
The test build passed. This PR should unbreak the GPU build.

JackCaoG (approved):

PR #7724, opened 2024-07-23
Re-implement and refactor the verifier.

ysiraichi:
This PR re-implements the existing verifier, making the following improvements:

- Enabling the verification of more models on inference
    - Previously, it expected the models to return a single output tensor
- Enabling the verification of training
    - There was a plumbing bug in `TorchBenchModel.train`
    - Running training for a few iterations
- Model cleanup
    - Delete each used model so that we don't run out of memory on large models
- Use PyTorch functions for checking whether the accuracy is acceptable
    - Instead of using MRE, we use the [same thing PyTorch does](https://github.com/pytorch/pytorch/blob/69b1999586eb2df741e7fa4e37a8611818113bb0/torch/_dynamo/utils.py#L1361-L1373)

In order to do so, here's a summary of the changes:

- Introduce `BenchmarkModel` methods: `tolerance()`, `use_cosine_similarity`, and `skip_verifier()`
    - The logic is taken [from PyTorch](https://github.com/pytorch/pytorch/blob/69b1999586eb2df741e7fa4e37a8611818113bb0/benchmarks/dynamo/torchbench.py#L434-L447), which also uses _torchbench.yaml_
- Introduce `force_dtype` parameter when loading a model
    - So that we can run models on eager `fp64`
- More meaningful verification codes
- Move `reset_rng_state` and `cleanup` to _util.py_
- Change how we access the YAML configuration file, replacing the raw strings

cc @miladm @zpcore

### DIFF
--- a/benchmarks/util.py
+++ b/benchmarks/util.py
@@ -77,12 +77,14 @@ def move_to_device(item, device, torch_xla2: bool = False):
     move_to_device_func = lambda t: jax.device_put(
         torch_xla2.tensor.t2j(t), device)
   else:
+
     def move_to_device_func(tensor: torch.Tensor) -> torch.Tensor:
       # If `tensor` is an XLA tensor, first move it to CPU. We need to do
       # that if we want to move the tensor to, say, CUDA.
       if tensor.device.type == "xla":
         return tensor.cpu().to(device)
       return tensor.to(device)
+
   return pytree.tree_map_only(torch.Tensor, move_to_device_func, item)
 
 
--- a/benchmarks/util.py
+++ b/benchmarks/util.py
@@ -1,6 +1,7 @@
 import argparse
 from contextlib import contextmanager
 import functools
+import gc
 import logging
 import os
 from os.path import abspath, exists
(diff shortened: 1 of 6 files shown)


benchmarks/util.py
```diff
@@ -76,11 +77,18 @@ def move_to_device(item, device, torch_xla2: bool = False):
...
         torch_xla2.tensor.t2j(t), device)
   else:
-    move_to_device_func = lambda t: t.to(device)
+
+    def move_to_device_func(tensor: torch.Tensor) -> torch.Tensor:
+      # If `tensor` is an XLA tensor, first move it to CPU. We need to do
+      # that if we want to move the tensor to, say, CUDA.
+      if tensor.device.type == "xla":
+        return tensor.cpu().to(device)
+      return tensor.to(device)
```
zpcore, line 86: Is this because of the `CUDA fallback` you implemented recently?

benchmarks/util.py
```diff
@@ -169,3 +177,26 @@ def find_near_file(names: str):
       if exists(path):
         return abspath(path)
   return None
+
+
+def reset_rng_state(benchmark_experiment: "BenchmarkExperiment"):
```
zpcore, line 182: Thanks for fixing the circular import issue.

benchmarks/util.py
ysiraichi: Not really. It's just that, if we want to do XLA -> CUDA, we need to go through the CPU (unless we use DLPack). Previously, I think we only did CUDA -> XLA, but moving to CPU explicitly.

zpcore:
Thanks for adding the verifier!

zpcore (approved):

PR #7729, opened 2024-07-23
[torch_xla2] Add decorator to promote int inputs

will-cromar:
Several ops are expected to return a `float32` when recieving any int input, e.g.

```
>>> torch.sin(torch.tensor(1, dtype=torch.int8)).dtype
torch.float32
>>> torch.sin(torch.tensor(1, dtype=torch.int16)).dtype
torch.float32
>>> torch.sin(torch.tensor(1, dtype=torch.int64)).dtype
torch.float32
```

JAX has different type promotion rules, such as promoting int64 to float64. Wrap PyTorch's common type promotion rule in a decorator to make it more reusable.

### DIFF
--- a/experimental/torch_xla2/torch_xla2/ops/op_base.py
+++ b/experimental/torch_xla2/torch_xla2/ops/op_base.py
@@ -1,9 +1,11 @@
 import functools
+import jax
+import jax.numpy as jnp
 import torch
 from torch_xla2.ops import mappings
 from torch_xla2 import types
 
-from typing import Optional, ParamSpec
+from typing import Callable, Concatenate, Optional, ParamSpec
 
 
 class InplaceOp:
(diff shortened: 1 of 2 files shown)


experimental/torch_xla2/torch_xla2/ops/op_base.py
```diff
@@ -54,3 +55,16 @@ def wrapper(*args: P.args,
...
 
   return decorator
+
+
+def promote_int_input(f: Callable[Concatenate[jax.Array, P], types.JaxValue]):
+   """If the first argument is an int array, promote it to float32."""
+   @functools.wraps(f)
+   def wrapper(x: jax.Array, *args: P.args, **kwargs: P.kwargs):
+      if x.dtype in [jnp.int8, jnp.int16, jnp.int32, jnp.int64]:
+        x = x.astype(jnp.float32)
```
qihqi, line 65: Actually it seems to just go to the default dtype:
In [2]: torch.set_default_dtype(torch.float64)

In [3]: torch.sin(torch.tensor(1, dtype=torch.int64))
Out[3]: tensor(0.8415)

In [4]: torch.sin(torch.tensor(1, dtype=torch.int64)).dtype
Out[4]: torch.float64

In [5]: torch.sin(torch.tensor(1, dtype=torch.int32)).dtype
Out[5]: torch.float64

qihqi (approved):
change float32 to torch.get_default_dtype and good to go

experimental/torch_xla2/torch_xla2/ops/op_base.py
will-cromar: Good catch. Fixed

PR #7734, opened 2024-07-24
Remove `tpuvm` dependency from r2.4 branch

will-cromar:
Otherwise we get this error:

```
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         Can't have direct dependency: libtpu-nightly@                          
         https://storage.googleapis.com/cloud-tpu-tpuvm-artifacts/wheels/libtpu-
         nightly/libtpu_nightly-0.1.dev20240612-py3-none-any.whl ; extra ==     
         "tpuvm". See https://packaging.python.org/specifications/core-metadata 
         for more information. 
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -313,8 +313,6 @@ setup(
         # pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
         'tpu': [f'libtpu-nightly=={_libtpu_version}'],
         # On nightly, install libtpu with `pip install torch_xla[tpuvm]`
-        # Remove from release branches since this is not allowed by PyPI.
-        'tpuvm': [f'libtpu-nightly @ {_libtpu_storage_path}'],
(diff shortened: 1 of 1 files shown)


will-cromar:
@JackCaoG what do you think of just removing this dep on master too? We'll have to update any tests/usage that uses `torch_xla[tpuvm]` to use `torch_xla[tpu] -f ...`

JackCaoG (approved):

JackCaoG:
I forgot why we switched to `tpu` from `tpuvm` but if this is the future direction we should just deprecate old usage.. only thing I am not sure if can we add some warning messages when users run with `[tpuvm]`

will-cromar:
PyPI rejecting the `tpuvm` form (direct dependency on HTTP URL) is the reason we added `tpu` instead. FWIW, `tpuvm` has not been in stable releases since 2.1.0 for exactly this reason. This would only affect us and potentially any users of the nightly package.

PR #7735, opened 2024-07-24
Fix the crash when symint is part of the output with dynamic torch.compile

JackCaoG:
when we compile the program with `torch.compile(backend="openxla")`, the output of the fx graph might contains int. The logic of the dynamo bridge currently expects all output to be tensor and will crash when calling pybind.


JackCaoG:
@alanwaketan can you take a look at this pr? If everything looks OK you can also merge it directly.

alanwaketan (approved):
LGTM.

Issue #7731, opened 2024-07-23
libtpu not installed with nightly build


davidaknowles:
## ❓ Questions and Help

Hi - I followed the instructions for the nightly build install (on a v4-8), i.e.
```
pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
pip3 install https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-nightly-cp310-cp310-linux_x86_64.whl
```
which ran successfully (pip claims), but this isn't giving me access to the TPU. When I try to run the test code from the Troubleshooting page I get an error that `libtpu` wasn't found. Am I missing an installation step?

JackCaoG:
suspect this is a pip issue, if you do
```
root@t1v-n-bafa368c-w-0:/ansible# python
Python 3.10.14 (main, Apr 24 2024, 12:45:30) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import libtpu
>>> libtpu.__file__
'/usr/local/lib/python3.10/site-packages/libtpu/__init__.py'
```
what would you see?

davidaknowles:
```
daknowles@t1v-n-57a06759-w-0:~$ python
Python 3.10.12 (main, Mar 22 2024, 16:50:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import libtpu
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'libtpu'
>>> 
```

JackCaoG:
cpu part is for the torchvision, it actually has nothing to do with the libtpu. To install the libtpu you can run
```
pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
```

Let me move this up in our README, it is mentioned in https://github.com/pytorch/xla#python-packages but it is hard to find. I will fix it tmr.

davidaknowles:
OK yup that fixed it, thanks! I saw that line in the README but assumed that would install stable not nightly.

PR #7737, opened 2024-07-24
Make libtpu install more explict

JackCaoG:
fix https://github.com/pytorch/xla/issues/7731

The libtpu install can be more explicit

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ To install PyTorch/XLA nightly build in a new TPU VM:
 ```
 pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
 pip3 install https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-nightly-cp310-cp310-linux_x86_64.whl
+pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
 ```
 
 ### GPU Plugin (beta)
@@ -148,7 +149,7 @@ Our comprehensive user guides are available at:
 
 PyTorch/XLA releases starting with version r2.1 will be available on PyPI. You
 can now install the main build with `pip install torch_xla`. To also install the
-Cloud TPU plugin, install the optional `tpu` dependencies after installing the main build with
+Cloud TPU plugin corresponding to your installed `torch_xla`, install the optional `tpu` dependencies after installing the main build with
 
 ```
 pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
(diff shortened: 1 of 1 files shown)


JackCaoG:
@will-cromar can you take a look at this pr?

README.md
```diff
@@ -34,6 +34,7 @@ To install PyTorch/XLA nightly build in a new TPU VM:
 ```
 pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
 pip3 install https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-nightly-cp310-cp310-linux_x86_64.whl
+pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
```
will-cromar, line 37: You should be able to do `pip install 'torch_xla[tpu] @ https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-nightly-cp310-cp310-linux_x86_64.whl' -f https://storage.googleapis.com/libtpu-releases/index.html` 

The advantage here is that `pip` will fail if there's a dependency issue (reasonably likely since you hardcoded python 3.10 in the command) rather than reinstalling a non-nightly `torch_xla` from PyPI. You should also make a note to change the Python version if necessary.

README.md
```diff
@@ -34,6 +34,7 @@ To install PyTorch/XLA nightly build in a new TPU VM:
 ```
 pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
```
will-cromar, line 35: Regarding https://github.com/pytorch/xla/issues/7697#issuecomment-2251576498, can you also add an optional command at the end to install compatible `torchaudio`, `torchtext`, and `torchvision` from the PyTorch nightly index?

JackCaoG: we rarely ever need torchaudio or torchtext. I think we put torchvision here because I use resnet in my examples.

JackCaoG:
@will-cromar can you take another look?

README.md
will-cromar: I'll follow up on this since it's tangential to this PR's topic. IMO it's still useful to include the other official ecosystem packages. It's very easy to break nightly `torch_xla` by installing a stable `torchtext` that pulls in an incompatible stable `torch`. We've made this mistake numerous times in our own test cases.

vanbasten23: If we want to install `torchaudio`, `torchtext`, etc. from the PyTorch nightly index, the command from PyTorch index page https://screenshot.googleplex.com/8xKE2QXabRsqq9D is `pip3 install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cpu`. They should be compatible already.

PR #7743, opened 2024-07-25
fix TPU test regarding word_size

zpcore:


will-cromar (approved):

will-cromar:
LGTM assuming GPU tests pass too

PR #7745, opened 2024-07-25
add test for dynamo + traceable collectives

JackCaoG:


will-cromar (commented):
Do we want to use the experimental API for these directly? https://github.com/pytorch/pytorch/blob/main/torch/distributed/_functional_collectives.py

We really should get rid of `xm.all_*`

alanwaketan (approved):
I think we have test cases for it:
https://github.com/pytorch/xla/blob/master/test/test_mp_replication.py
https://github.com/pytorch/xla/blob/master/test/test_mp_all_gather.py

JackCaoG:
Ah ok, what I want to test is that
1. Dynamo can recognize the traceable collective and does not fallback, which is why I add multiple ops into one fn and check there is only one execution
2. check the `dynamic=True` case

I felt like might be easier to just do it in a separate test.

PR #7746, opened 2024-07-25
Support `@xla.step` decorator without parentheses

will-cromar:
Previously, you had to instantiate `step`:

```
@xla.step()
def f():
   pass
```

Now either option works:

```
@xla.step
def f():
   pass
```

Add explicit feature mode option to revert to old behavior.

Follow-up from #7642

### DIFF
--- a/torch_xla/torch_xla.py
+++ b/torch_xla/torch_xla.py
@@ -61,7 +61,7 @@ def sync():
   xm.mark_step()
 
 
-def step(f: Optional[Callable] = None, *, eager_mode=False):
+def step(f: Optional[Callable] = None):
   """Wraps code that should be dispatched to the runtime.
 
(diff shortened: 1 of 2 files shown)


torch_xla/torch_xla.py
```diff
@@ -59,24 +61,31 @@ def sync():
   xm.mark_step()
 
 
-@contextlib.contextmanager
-def step():
+def step(f: Optional[Callable] = None, *, eager_mode=False):
```
JackCaoG, line 64: within a `step` user would not want to set `eager` to True right?

will-cromar: Haha I added this as an example usage of an argument. I'll go ahead and remove it -- you're right that it doesn't logically make sense here.

JackCaoG (approved):

PR #7749, opened 2024-07-25
[benchmarks] Turn CUDAGraphs on for inductor.

ysiraichi:
This PR turns CUDAGraphs on for inductor configuration.

cc @miladm @JackCaoG @zpcore

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -268,6 +268,11 @@ class ExperimentRunner:
     benchmark_model = self.model_loader.load_model(model_config,
                                                    benchmark_experiment)
 
+    # Turn on CUDAGraphs if we are running inductor
+    if benchmark_experiment.is_inductor():
+      from torch._inductor import config as inductor_config
+      inductor_config.triton.cudagraphs = True
+
     # Repeat the experiment and accumulate metrics.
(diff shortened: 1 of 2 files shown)


benchmarks/experiment_runner.py
```diff
@@ -268,6 +268,11 @@ def run_single_config(self):
     benchmark_model = self.model_loader.load_model(model_config,
                                                    benchmark_experiment)
 
+    # Turn on CUDAGraphs if we are running inductor
+    if benchmark_experiment.is_inductor():
+      from torch._inductor import config as inductor_config
+      inductor_config.triton.cudagraphs = True
```
zpcore, line 274: Is this a common way to enable the cudagraphs? I searched that it can be passed through the compile options: https://pytorch.org/docs/stable/generated/torch.compile.html.

ysiraichi: It's [how they do it](https://github.com/pytorch/pytorch/blob/79f0c4dc04c7976b734767d64c4833932219dcfb/benchmarks/dynamo/common.py#L4425-L4431) in their benchmarking scripts:

zpcore: Thanks for the point!

zpcore (approved):

PR #7750, opened 2024-07-25
Refactor torch_xla.experimental.compile and move it out of experimental

JackCaoG:


torch_xla/torch_xla.py
```diff
@@ -61,13 +61,44 @@ def sync():
   xm.mark_step()
 
 
-def step(f: Optional[Callable] = None):
```
will-cromar, line 64: Do you want `step` to pass through the callable argument?

will-cromar (approved):

torch_xla/torch_xla.py
JackCaoG: my thinking is that `step` does not take `f` in the 2.4 release, and I want people to use the `torch_xla.compile` api directly since they are really the same thing but rebranding. I didn't remove `step` because there might be people use it in 2.4.

Issue #7739, opened 2024-07-24
PyTorch/XLA usability progress tracking


zpcore:
## Description

Status tracking for the progress of improving the usability.

## Action items for APIs

Below are APIs we plan to clean up to imporve the usability. The list is mostly based on the [doc](https://docs.google.com/document/d/1NZJJFk0ICN8kus7U2nIO3moYcUGg4neXA8UkotbJmmY/edit?tab=t.0#heading=h.x9snb54sjlu9) from @will-cromar . 

| APIs            | Actions        |    Complete Date    |     Related issues |Related Design   |    PR
|------------|-------------|--------------------|------------------|----------------|-------------|
| xla_model.parse_xla_device()     | internalize          | 2024-07-18  |  |  |https://github.com/pytorch/xla/pull/7675
| torch_xla.launch()      | new API  introduce        | 2024-07-12 |   | [Improve multiprocess with torch_xla.launch()](https://docs.google.com/document/d/1rnBzKpX-XWkMJE423Y66eT3njL3pNtYFOJp2sXI1AQ4/edit?resourcekey=0-N7dAx6ErffDiCc3sQU6DgA&tab=t.0#heading=h.yqh3poxaddbt) | https://github.com/pytorch/xla/pull/7648| 
| xla_model.xrt_world_size()    | deprecate with runtime.world_size() , remove `defval` argument       | 2024-07-24  |  |  | https://github.com/pytorch/xla/pull/7679
|xla_model.get_ordinal() | deprecate with runtime.global_ordinal() | 2024-07-24  |  |  | https://github.com/pytorch/xla/pull/7679
|xla_model.get_local_ordinal() | deprecate with runtime.global_ordinal() | 2024-07-24  |  |  | https://github.com/pytorch/xla/pull/7679
| using_pjrt() |  deprecate | | https://github.com/pytorch/xla/issues/7730|
| requires_pjrt() |  deprecate | | https://github.com/pytorch/xla/issues/7730|
|xla_real_devices| deprecate |
| xla_device_hw | deprecate | 
| xla_replication_devices | replace |
| set_replication | replace|
|unlazy | delete | 
|RateTracker| internalize |
|ToXlaTensorArena | internalize |
|check_view_sharing | delete | 
| add_step_closure | replace |
|mark_step| replace | | https://github.com/pytorch/xla/issues/6751|
|reduce_gradients| internalize |
|optimizer_step| replace? |
| save | replace/upstream |
[...]

PR #7753, opened 2024-07-26
clean up deprecated

zpcore:
Some clean up fix related to deprecated APIs:

1. Do not use `parse_xla_device`, `xm.get_ordinal` and `xm.world_size` from deprecated functions internally.
2. Update ddp related doc to use `xr.global_ordinal`. E.g. ,https://pytorch.org/xla/master/multi_process_distributed.html
3. Update the deprecation message to involve the version related info as request by @miladm

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -199,7 +199,7 @@ def xla_replication_devices(local_devices):
   real_devices = xla_real_devices(local_devices)
   device_types = set()
   for device in real_devices:
-    xdev = parse_xla_device(device)
+    xdev = _utils.parse_xla_device(device)
     device_types.add(xdev[0])
   if len(device_types) != 1:
     # No replication if the device set spawns multiple device types.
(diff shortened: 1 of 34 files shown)


qihqi (approved):

zpcore:
Does everyone agree that those deprecated functions will be removed in 2.6 release for the grace period?

zpcore:
CI CUDA built failure issue exists in our commit also. Seems to be an upstream issue. I will merge this once all other tests pass.

torch_xla/core/xla_model.py
```diff
@@ -37,9 +37,14 @@
 XLA_LIB = Library("xla", "DEF")
 
 from . import xla_model as this_module
-xrt_world_size = deprecated(this_module, torch_xla.runtime.world_size)
-get_ordinal = deprecated(this_module, torch_xla.runtime.global_ordinal)
-parse_xla_device = deprecated(this_module, _utils.parse_xla_device)
+xrt_world_size = deprecated(this_module, torch_xla.runtime.world_size,
+                            'xrt_world_size() will be removed in release 2.6.')
```
miladm, line 41: a few comments to improve our deprecation approach:

+ Do we have an existing set of action issues defined for future API removals and maintain that list on github to tackle when the time comes
+ Can the message suggest what to replace the API with?
+ Please confirm the API transition is safe and bug free by adding tests that compare old API and new API functionality
+ This message hard codes the version number - I expect the version number can vary from one API to another - can we have the version be parameterized for easier maintenance?
+ We need a RFC that outlines our backward guarantees 

cc @JackCaoG @jiawenliu64 @zpcore @qihqi

zpcore: What about using this table (https://github.com/pytorch/xla/issues/7739) for tracking?  We can add a column for deprecate timeline.

PR #7757, opened 2024-07-26
Update function doc strings

JackCaoG:


JackCaoG:
I check the generated output and it looks reasonable.

docs/source/index.rst
```diff
@@ -77,6 +77,7 @@ distributed
 .. automodule:: torch_xla.distributed.parallel_loader
 .. autoclass:: ParallelLoader
```
will-cromar, line 78: Let's remove `ParallelLoader` from the docs. You should almost never use it instead of `MpDeviceLoader`

JackCaoG:
@will-cromar can you take another look at this pr?

torch_xla/runtime.py
```diff
@@ -313,8 +324,8 @@ def initialize_cache(path: str, readonly: bool = False):
   before any computations have been performed.
 
   Args:
-    path: The path at which to store the persistent cache.
-    readonly: Whether or not this worker should have write access to the cache.
+    path (str): The path at which to store the persistent cache.
```
will-cromar, line 327: I don't think it's necessary to include the type in the docstring when it's already annotated. (`readonly` is an optional bool, not just a bool)

will-cromar (approved):

torch_xla/runtime.py
JackCaoG: yea then the type annotation is also incorrect haha, let me fix that in a follow up pr.

PR #7758, opened 2024-07-26
openxla pin update

zpcore:
Update libtpu to 07/26 nightly.


zpcore:
It seems like those patches are no longer needed.

zpcore:
Below patch changes are no longer needed.
<details>
<summary>openxla diff</summary>
```
diff --git a/xla/mlir_hlo/WORKSPACE b/xla/mlir_hlo/WORKSPACE
index ae7ca4dd02..0f59210633 100644
--- a/xla/mlir_hlo/WORKSPACE
+++ b/xla/mlir_hlo/WORKSPACE
@@ -38,7 +38,10 @@ http_archive(
     build_file_content = "# empty",
     sha256 = LLVM_SHA256,
     strip_prefix = "llvm-project-" + LLVM_COMMIT,
-    urls = ["https://github.com/llvm/llvm-project/archive/{commit}.tar.gz".format(commit = LLVM_COMMIT)],
+    urls = [
+        "https://github.com/llvm/llvm-project/archive/{commit}.tar.gz".format(commit = LLVM_COMMIT),
+        "https://storage.googleapis.com/tpu-pytorch/llvm-raw/{commit}.tar.gz".format(commit = LLVM_COMMIT),
+    ],
 )
 
 load("@llvm-raw//utils/bazel:configure.bzl", "llvm_configure")
diff --git a/xla/service/cpu/runtime_fp16.h b/xla/service/cpu/runtime_fp16.h
index c86d6dc37f..f7bf449ade 100644
--- a/xla/service/cpu/runtime_fp16.h
+++ b/xla/service/cpu/runtime_fp16.h
@@ -18,12 +18,7 @@ limitations under the License.
 
 #include <stdint.h>
 
-// _Float16 always gets us the correct ABI type, so use that if available.
-// AArch64 GCC defines __FLT16_MANT_DIG__ even when _Float16 is not available.
-#if defined(__FLT16_MANT_DIG__) && \
-    (defined(__clang__) || !(defined(__GNUC__) && defined(__aarch64__)))
-using XlaF16ABIType = _Float16;
-#elif defined(__x86_64__)
+#if defined(__x86_64__)
 // Older versions of Clang don't have _Float16. Since both float and _Float16
 // are passed in the same register we can use the wider type and careful casting
 // to conform to x86_64 psABI. This only works with the assumption that we're
diff --git a/xla/service/gpu/gpu_executable.cc b/xla/service/gpu/gpu_executable.cc
index bf9774711f..4160796d5c 100644
--- a/xla/service/gpu/gpu_executable.cc
+++ b/xla/service/gpu/gpu_executable.cc
@@ -816,8 +816,7 @@ absl::StatusOr<ExecutionOutput> GpuExecutable::ExecuteAsyncOnStreamImpl(
 #endif  // GOOGLE_CUDA || TENSORFLOW_USE_ROCM
 
   // Force synchronous execution if the allocator requires it.
-  const bool block_host_until_done =
-      !memory_allocator->AllowsAsynchronousDeallocation();
+  const bool block_host_until_done = true;
 
   // Lock the GPU with a shared lock so that we don't interfere with autotuning
   // that may be running during JIT compilation while allowing multiple XLA
```
</details>

JackCaoG (approved):

openxla_patches/f16_abi_clang.diff
```diff
@@ -1,19 +0,0 @@
...
-
- #include <stdint.h>
-
--// _Float16 always gets us the correct ABI type, so use that if available.
--// AArch64 GCC defines __FLT16_MANT_DIG__ even when _Float16 is not available.
--#if defined(__FLT16_MANT_DIG__) && \
--    (defined(__clang__) || !(defined(__GNUC__) && defined(__aarch64__)))
--using XlaF16ABIType = _Float16;
--#elif defined(__x86_64__)
-+#if defined(__x86_64__)
```
JackCaoG, line 16: I remember this one is needed because of upstream CI failure. does openxla already fix this too?

openxla_patches/gpu_race_condition.diff
```diff
@@ -1,14 +0,0 @@
...
-index 9279bd877..fab926a7c 100644
---- a/xla/service/gpu/gpu_executable.cc
-+++ b/xla/service/gpu/gpu_executable.cc
-@@ -669,8 +669,7 @@ absl::StatusOr<ExecutionOutput> GpuExecutable::ExecuteAsyncOnStreamImpl(
- #endif  // GOOGLE_CUDA || TENSORFLOW_USE_ROCM
-
-   // Force synchronous execution if the allocator requires it.
--  const bool block_host_until_done =
--      !memory_allocator->AllowsAsynchronousDeallocation();
-+  const bool block_host_until_done = true;
```
JackCaoG, line 11: this one is for a GPU repo, did openxla somehow fixed it?

openxla_patches/f16_abi_clang.diff
zpcore: This seems related to the arm build. I am rebuilding on my mac laptop to see if any issue pop up.

openxla_patches/gpu_race_condition.diff
zpcore: The CI shows the GPU test pass, will this be a concern? Just in case the CI doesn't catch the issue.

WORKSPACE
```diff
@@ -60,9 +60,7 @@ http_archive(
     ],
     patch_tool = "patch",
     patches = [
-        "//openxla_patches:cache_urls.diff",
```
will-cromar, line 63: @JackCaoG do you remember what this patch is for? I looked back in the history and couldn't find a reason. That GCS folder (`gs://tpu-pytorch/llvm-raw`) was last updated in 2023, so I assume this is obsolete?

lsy323 (approved):

openxla_patches/f16_abi_clang.diff
zpcore: I am having issue compiling the pytorch on M1 in the first step. Not sure if anyone have tried this for XLA before...

openxla_patches/gpu_race_condition.diff
JackCaoG: it was another user that complains about some inconisistnet behavior if this is enabled, I would just leave it here if nothing changed in openxla side.

WORKSPACE
JackCaoG: I think it was back then the llvm image is not stable so we kept a copy, we can remove it..

openxla_patches/f16_abi_clang.diff
JackCaoG: you can open a pr on pytorch and update the XLA pin in that PR target this pr to verify if you can remove this patch.

zpcore: Opened one here: https://github.com/pytorch/pytorch/pull/132063

openxla_patches/gpu_race_condition.diff
zpcore: Got it, I will just not bother it.

PR #7769, opened 2024-07-30
Disable open PJRT for xal2 if OS env disable flag is true

FanhaiLu1:
Add disable_xla2_PJRT_test to let engineer have flexibility to  turn off jax PIRT test. By default, it won't impact existing logic. 

**Background:**

In ray multiple cases, call jax devices both in head and workers caused exhausted TPU exception. There are 2 ways to fix it:

1.  xla2 init jax function part to a function and don't call it in init state, call the function after all the call inited 
2. have a flag to disable jax function call in jax __init__.py


FanhaiLu1:
@qihqi Can you take a look and approve this PR?

qihqi (approved):

PR #7770, opened 2024-07-30
Implement torch.ops._c10d_functional.broadcast

JackCaoG:


torch_xla/core/c10d_registration.py
```diff
@@ -0,0 +1,25 @@
...
+
+# "broadcast(Tensor self, int src, str tag, int[] ranks, int group_size) -> Tensor",
+@torch.library.impl(c10d_lib, "broadcast", "XLA")
+def broadcast_xla(self: torch.Tensor,
+                  src: int,
+                  tag: str,
+                  ranks: Optional[List] = None,
+                  group_size: Optional[int] = None) -> torch.Tensor:
+  assert group_size == None, "currently does not support group_size"
+  # xm.collective_broadcast perform an inplace update, but
```
will-cromar, line 18: Can you clone the input and do the inplace `xm.collective_broadcast` on that instead of copying the implementation?

Either way, we should do proper functional implementations of these ops in C++ along with `all_reduce` (cc @zpcore since we talked about this today)

torch_xla/core/c10d_registration.py
```diff
@@ -0,0 +1,25 @@
+import torch
```
will-cromar, line 1: Since this is an implementation and not a new public API, can you put an underscore or move this to `_internal`?

torch_xla/core/c10d_registration.py
```diff
@@ -0,0 +1,25 @@
...
+import torch_xla.runtime as xr
+import torch_xla.core.xla_model as xm
+from torch.library import Library
+from typing import List, Optional, TypedDict
+
+c10d_lib = Library("_c10d_functional", "IMPL")
```
will-cromar, line 7: What is this `Library` for? I don't remember needing this when I was playing with `@torch.library.impl` in `torch_xla2`

test/dynamo/test_traceable_collectives.py
```diff
@@ -12,6 +12,11 @@ def dummy_collective_fn(input):
   return res_tensor
 
 
+def collective_broadcast_and_cos(input, src):
+  res_tensor = torch.ops._c10d_functional.broadcast(input, src, "default")
```
will-cromar, line 16: Does it make more sense to test the Python API? https://github.com/pytorch/pytorch/blob/32c57e78edc46aa71ed19e013741c65b3d777fe9/torch/distributed/_functional_collectives.py#L143-L154

These functions (rather, the `_inplace versions`) are what Dynamo actually inserts as a replacement for the legacy `dist.*` collectives.

will-cromar (approved):

torch_xla/core/c10d_registration.py
JackCaoG: let me move this into _internal.

torch_xla/core/c10d_registration.py
JackCaoG: I think they are the same thing, but using different apis

JackCaoG: I think it just an easy way to say I want to register the op under `torch.ops. _c10d_functional`, I copied from upstream haha

will-cromar: Does this work if you comment this line out? I don't think it's necessary in addition to `torch.library.impl`

torch_xla/core/c10d_registration.py
JackCaoG: I really want to just get rid of the old `collective_broadcast`, it is an inplace op so it should be `collective_broadcast_`. Given that @zpcore is going to register theses op properly I think we should delete `xm.collective_broadcast` at some point so it is better not to depend this implementation on that one.

test/dynamo/test_traceable_collectives.py
JackCaoG: so I tried a couple things, first I need to init the process group, but then 
```
res_tensor = torch.distributed.broadcast(input, src)
```
return None, I am guessing I need to pass group and tag correctly. Furthermore when I turn on `torch.compile` it failed with 
```
  File "/workspaces/dk4/pytorch/xla/torch_xla/_internal/pjrt.py", line 174, in run_multiprocess
    replica_results = list(
  File "/workspaces/dk4/pytorch/xla/torch_xla/_internal/pjrt.py", line 175, in <genexpr>
  File "/usr/local/lib/python3.10/concurrent/futures/process.py", line 575, in _chain_from_iterable_of_lists
    for element in iterable:
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 621, in result_iterator
    yield _result_or_cancel(fs.pop())
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 319, in _result_or_cancel
    return fut.result(timeout)
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    return self.__get_result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
AssertionError: expected FunctionType found instancemethod <instancemethod  at 0x7f8d8456bf10>

from user code:
   File "/workspaces/dk4/pytorch/xla/torch_xla/distributed/xla_backend.py", line 98, in broadcast
    root_tensor = tensors[opts.rootTensor]
```

I think it got routed to https://github.com/pytorch/xla/blob/master/torch_xla/distributed/xla_backend.py#L97-L104 instead of my functional version. I will keep test as this to unblock VLLM for now.

torch_xla/core/c10d_registration.py
JackCaoG: if you check down below I did
```
@torch.library.impl(c10d_lib, "broadcast", "XLA")
```
I guess I can do
```
@torch.library.impl(Library("_c10d_functional", "IMPL"), "broadcast", "XLA")
```
but then this `@torch.library.impl` might be called for multuple ops so I figured it is better to leave it there.

will-cromar: You can reference the ops you need directly by name. See my PR: https://github.com/pytorch/xla/pull/7705/files#diff-8e0394772809eec997864870faa616ae1a5123c0776204e34d0036eb94f69750R67

See the documented example (`aten::sin`): https://pytorch.org/docs/stable/library.html#torch.library.impl

Here, it would be something like `"_c10d_functional::broadcast"`

JackCaoG: updated

PR #7771, opened 2024-07-30
Make persistent cache loading failure more explict

JackCaoG:


JackCaoG:
failure is irrelevant

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -660,8 +660,11 @@ ComputationClient::ComputationPtr PjRtComputationClient::DeserializeComputation(
     const std::string& serialized) {
   auto executable_or = client_->DeserializeExecutable(serialized, std::nullopt);
   if (!executable_or.ok()) {
-    TF_LOG(WARNING) << "Failed to deserialize executable: "
-                    << executable_or.status();
+    TF_LOG(WARNING)
+        << "Failed to deserialize executable: " << executable_or.status()
+        << ". This usually means that you are trying to load from "
+           "persistent cache dir but failed. If you just updated the torch_xla "
+           "or libtpu version please clear the cache dir and recompile";
```
jonb377, line 667: What's the failure case you hit? [torch_xla](https://github.com/pytorch/xla/blob/622f4f9376796b4668cdf71cdb6a71013fe1d4f1/torch_xla/csrc/xla_graph_executor.cpp#L644) and [libtpu](https://github.com/pytorch/xla/blob/622f4f9376796b4668cdf71cdb6a71013fe1d4f1/torch_xla/csrc/runtime/pjrt_computation_client.cc#L82) versions should both already influence the graph hash.

JackCaoG: it is that one user updated the torch+torch_xla version and saw this warning message. He was not able to understand what it means so I thought it is easier to explictly ask them to clear the persistent cache.

jonb377: Huh, strange that it would have a cache hit in that case, but there's no harm in improving the error message. Thanks Jack!

jonb377 (approved):

torch_xla/csrc/runtime/pjrt_computation_client.cc
JackCaoG: oh you have a good point... if pytorch/xla version change the graph hash should also changed... now I don't know why it failed to read...

JackCaoG:
@jonb377 what happens when we try to use persistent cache in a multi process context? Will each process overwrite each other's cache?

jonb377:
If each process has the same cache dir, each will try to write to the cache (unless the cache is set to [readonly](https://github.com/pytorch/xla/blob/622f4f9376796b4668cdf71cdb6a71013fe1d4f1/torch_xla/runtime.py#L317) for the process). Maybe multiple concurrent writes is corrupting the serialized binary? I haven't hit this in the past, but my testing was primarily for single-process SPMD.

JackCaoG:
ok that's a great point, I should modify the warning message to suggest this might be the source of the desearlization failure.

PR #7776, opened 2024-07-30
Support fullgraph for torch_xla.compile

JackCaoG:
with `torch_xla.compile(fn, full_graph=True)` we will output an error message when we see graph breaks(early access of the value of tensor or cpu fallback) inside the passed in `fn`

### DIFF
--- a/test/eager/test_eager_with_xla_compile.py
+++ b/test/eager/test_eager_with_xla_compile.py
@@ -20,6 +20,12 @@ class EagerWithXLACompileTest(unittest.TestCase):
   def dummy_cos_sin(self, tensor):
     return torch.cos(torch.sin(tensor))
 
+  def dummy_graph_break(self, t):
+    if t[0][0] > 0:
+      return torch.sin(t)
+    else:
+      return torch.cos(t)
+
   def test_eager_with_compile_basic(self):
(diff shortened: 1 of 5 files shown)


test/eager/test_eager_with_xla_compile.py
```diff
@@ -45,21 +51,34 @@ def test_eager_with_compile_basic(self):
...
-  self.assertTrue(
-      torch.allclose(res * 0.3,
-                     self.dummy_cos_sin(self.dummy_cos_sin(t1)) * 0.3))
-  xm.wait_device_ops()
-  self.assertEqual(met.metric_data("ExecuteTime")[0], 2)
+    with self.assertRaises(Exception) as context:
+      t2_compiled = torch_xla.compile(
+          self.dummy_graph_break, full_graph=True)(
+              t1)
+    self.assertIn(
```
will-cromar, line 79: `assertRaisesRegex` is what you want here https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

will-cromar (approved):
Neat!

JackCaoG:
Let me merge it to unblock my other pr, I will fix review comment in a separate follow up pr too.

PR #7777, opened 2024-07-30
[benchmarks] Fix verifier error handling and OOM.

ysiraichi:
In summary, this PR:

- Skips the actual experiment when running the verifier, i.e. goes directly into the verifier code, instead of running the experiment first (the experiment is still run in the end of the verifier, though)
- Handles errors when running the model on `fp64`, turning cosine similarity verification ON on any exception
- Renames a few confusing `VerificationCode` enum names

cc @miladm @JackCaoG @zpcore

### DIFF
--- a/benchmarks/experiment_runner.py
+++ b/benchmarks/experiment_runner.py
@@ -184,7 +184,7 @@ class ExperimentRunner:
               benchmark_experiment.to_dict(),
               benchmark_model.to_dict(),
               metrics={"error": str(e)},
-              verification_code=VerificationCode.VERIFIER_SKIPPED_UNEXPECTEDLY,
+              verification_code=VerificationCode.VERIFIER_DIDNT_RUN,
           )
         except subprocess.CalledProcessError as e:
           self._fwd_captured_stdout_stderr(e.stdout, e.stderr)
@@ -193,7 +193,7 @@ class ExperimentRunner:
               benchmark_experiment.to_dict(),
               benchmark_model.to_dict(),
               metrics={"error": e.stderr},
-              verification_code=VerificationCode.VERIFIER_SKIPPED_UNEXPECTEDLY,
+              verification_code=VerificationCode.VERIFIER_DIDNT_RUN,
           )
         except subprocess.SubprocessError as e:
           logger.error("ERROR when launching child process")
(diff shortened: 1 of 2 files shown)


benchmarks/verifier.py
```diff
@@ -76,9 +89,12 @@ def verify(
         tol=tolerance,
     ):
       return VerificationCode.FAIL
-  except:
-    traceback.print_exc()
-    return VerificationCode.VERIFIER_FAILED
+  except VerificationException:
+    raise
+  except Exception as e:
+    # If anythin went wrong (other than an explicit VerificationException), raise
```
zpcore, line 95: small typo.

benchmarks/experiment_runner.py
```diff
@@ -262,59 +263,56 @@ def run_single_config(self):
...
+            self,
+            experiment_config,
+            model_config,
+            tolerance=model.tolerance(),
+            use_cosine_similarity=model.use_cosine_similarity())
+      except VerificationException as e:
+        verification_code = e.code
+        # Record the error in the metrics dictionary.
+        # Similar to what's done when the whole experiment fails.
+        accumulated_metrics["error"] = traceback.format_exc()
```
zpcore, line 312: If we use `verify`, we will not capture any performance metrics expect the verify result, did I understand it correctly?

benchmarks/experiment_runner.py
```diff
@@ -262,59 +263,56 @@ def run_single_config(self):
...
+
+    # Initialize output variables
+    accumulated_metrics = OrderedDict()
+    verification_code = VerificationCode.VERIFIER_SKIPPED
 
     # Turn on CUDAGraphs if we are running inductor
-    if benchmark_experiment.is_inductor():
+    if experiment.is_inductor():
       from torch._inductor import config as inductor_config
       inductor_config.triton.cudagraphs = True
```
zpcore, line 280: Will `inductor_config.triton.cudagraphs` impact following runs if we set the flag once?

benchmarks/experiment_runner.py
ysiraichi: Yes. That's correct.

benchmarks/experiment_runner.py
ysiraichi: Yes. I believe so.

zpcore (approved):
LGTM, thanks!

PR #7793, opened 2024-07-31
Attach callback to a Tensor's underlying buffer

will-cromar:
- New experimental high-level API `torch_xla.experimental.callback.on_ready_callback` to install callback (`(Tensor) -> None`) on Tensor, to be called when underlying buffer is ready
  - If Tensor is not assigned a buffer, throw an error
  - @JackCaoG and I chose this route instead of an explicit `Future` because `Tensor`s are already async results in our case, and returning `Futures` would require expanding the API in several places.
- Low-level API takes `void()` to make Pybind plumbing easier
- Since this is a Python callable to be called asynchronously from C++, there is high potential for deadlocks, as `callback` does need to reacquire the GIL from the main thread
  - I tried to fix any cases I could find. If a deadlock happens to you, a `thread apply all bt` from `gdb` would be extremely helpful for me to debug.
- `PrepareToExit` now waits for pending runtime operations to complete. Otherwise, we start destroying the runtime client before operations are done. If the main thread is waiting for a callback to complete, a deadlock occurs.

### DIFF
--- a/test/test_callback.py
+++ b/test/test_callback.py
@@ -24,6 +24,7 @@ class TestExperimentalCallback(absltest.TestCase):
       event.set()
 
     callback.on_ready_callback(c, cb)
+    event.wait(3)
 
 
 if __name__ == "__main__":
--- a/test/test_callback.py
+++ b/test/test_callback.py
@@ -1,0 +1,30 @@
+import threading
+
+from absl.testing import absltest
+import torch
+import torch_xla
+from torch_xla.experimental import callback
+
+
+class TestExperimentalCallback(absltest.TestCase):
+
+  @staticmethod
+  @torch_xla.compile
+  def executable():
+    a, b = torch.randn((100, 100), device=torch_xla.device()), torch.randn(
+        (100, 100), device=torch_xla.device())
+    return a @ b
+
+  def test_callback(self):
+    event = threading.Event()
+    c = self.executable()
+
+    def cb(tensor):
+      self.assertIs(c, tensor)
+      event.set()
+
+    callback.on_ready_callback(c, cb)
+
+
+if __name__ == "__main__":
+  absltest.main()
--- a/test/test_callback.py
+++ b/test/test_callback.py
@@ -21,6 +21,9 @@ class TestExperimentalCallback(absltest.TestCase):
 
     def cb(tensor):
       self.assertIs(c, tensor)
+      # TODO: check that result is both assigned and completed
+      self.assertNotIn("Data Handle: None",
+                       torch_xla._XLAC._get_xla_tensor_debug_info(tensor))
       event.set()
 
     callback.on_ready_callback(c, cb)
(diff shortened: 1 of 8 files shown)


test/test_callback.py
```diff
@@ -0,0 +1,31 @@
...
+    a, b = torch.randn((100, 100), device=torch_xla.device()), torch.randn(
+        (100, 100), device=torch_xla.device())
+    return a @ b
+
+  def test_callback(self):
+    event = threading.Event()
+    c = self.executable()
+
+    def cb(tensor):
+      self.assertIs(c, tensor)
```
JackCaoG, line 23: this doesn't really check if the computation has finished. Can you check the output of https://github.com/pytorch/xla/blob/master/torch_xla/csrc/init_python_bindings.cpp#L1214-L1217 instead? It will output whether the PJRT buffer is a placeholder or a real value.

test/test_callback.py
```diff
@@ -0,0 +1,31 @@
...
+  def executable():
+    a, b = torch.randn((100, 100), device=torch_xla.device()), torch.randn(
+        (100, 100), device=torch_xla.device())
+    return a @ b
+
+  def test_callback(self):
+    event = threading.Event()
+    c = self.executable()
+
+    def cb(tensor):
```
JackCaoG, line 22: is it difficult to support the cases where `call_back` takes more than just the tensor value? `call_back` might be a general python function that `tensor` is just one of the inputs.

will-cromar: You should `functools.partial` in that case to add other arguments to the callable https://docs.python.org/3/library/functools.html#functools.partial

JackCaoG (commented):
mostly lgtm, minor nits

test/test_callback.py
will-cromar: Fair enough. I just tried this, but I don't think the debug string is really testing that the data is ready either. I see:

```
# before
XLATensor {
TensorID: 5
Device: CPU:0
XLA Shape: f32[100,100]
ShardingSpec: None
IR: None
XLAData: 
  Data Device: CPU:0
  Data Shape: f32[100,100]
  Data Handle: None
Tensor on host: None
}
```

vs

```
# after
XLATensor {
TensorID: 5
Device: CPU:0
XLA Shape: f32[100,100]
ShardingSpec: None
IR: None
XLAData: 
  Data Device: CPU:0
  Data Shape: f32[100,100]
  Data Handle: 139615763718176
Tensor on host: None
}
```

That only tells us that a `PjRtBuffer` has been assigned and not that the buffer is ready. Do you think it's worth adding the buffer state to the debug string, or is it worth adding a new private API like `_is_buffer_ready`?

JackCaoG: oh right... I forgot that we will have a handle but that handle won't be ready. I think either approach will be fine and we can do that in a follow up pr.

JackCaoG (approved):

PR #7807, opened 2024-08-05
torch::lazy::MakeNode -> torch_xla::MakeNode

JackCaoG:
I am trying to implement the dynamic shape detection within a `torch_xla.compile` region, in order to do that I need to do some special handling after every IR Node is being created.

My first thought is to add this logic in `XlaNode` constructor but soon realize this will not work too well because
```
Permute::Permute(const torch::lazy::Value& input, std::vector<int64_t> dims)
    : XlaNode(
          torch::lazy::OpKind(at::aten::permute), {input},
          [&]() { return NodeOutputShape(input, dims); },
          /*num_outputs=*/1, torch::lazy::MHash(dims)),
      dims_(std::move(dims)) {}

torch::lazy::NodePtr Permute::Clone(torch::lazy::OpList operands) const {
  return torch::lazy::MakeNode<Permute>(operands.at(0), dims_);
}
```

XlaNode constructor will be called before the child class fully initialized. If I do `this->ToString()` it will actually call the parent class's ToString which is a bit annoying.

In this pr I replaced `torch::lazy::MakeNode` with the `torch_xla::MakeNode` so I can inject my debugging logic there.

### DIFF
--- a/CODEGEN_MIGRATION_GUIDE.md
+++ b/CODEGEN_MIGRATION_GUIDE.md
@@ -88,7 +88,7 @@ at::Tensor XLANativeFunctions::abs(const at::Tensor & self) {
 
   torch::lazy::NodePtr node = torch::lazy::ReuseNode<Abs>(lazy_self->GetIrValue());
   if (!node) {
-    node = torch::lazy::MakeNode<Abs>(lazy_self->GetIrValue());
+    node = torch_xla::MakeNode<Abs>(lazy_self->GetIrValue());
     CacheNode(node);
   }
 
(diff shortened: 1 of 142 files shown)


JackCaoG:
@alanwaketan @wonjoolee95 can I get a review for this one?

wonjoolee95 (approved):
This reminds me when we migrated `torch_xla::Node` to `torch:lazy:Node` long time ago 😆

JackCaoG:
> This reminds me when we migrated `torch_xla::Node` to `torch:lazy:Node` long time ago 😆

yea I am sad that we have to move back to our own implementation lol

PR #7809, opened 2024-08-06
[benchmarks] Guard empty metrics on verifier runs.

ysiraichi:
This PR fixes _result_analyzer.py_ by guarding its metric processing function call `get_calculated_metrics`, checking whether it's empty or not.

Empty metrics was allowed on _results.jsonl_ starting on #7777, where verifier-only runs were enabled.

cc @miladm @zpcore

### DIFF
--- a/benchmarks/result_analyzer.py
+++ b/benchmarks/result_analyzer.py
@@ -157,7 +157,8 @@ class ResultAnalyzer:
 
       if "error" not in dataline["metrics"]:
         d["dimensions"]["run_status"] = "success"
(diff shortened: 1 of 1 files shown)


zpcore (approved):
LGTM!

zpcore:
We can add the `python result_analyzer.py ...` in the test: https://github.com/pytorch/xla/blob/b4335f8ece35685646dbc433f629ee2434d4566e/test/benchmarks/run_torchbench_tests.sh#L74-L96 just to be safe.

ysiraichi:
I will leave that to  a follow-up PR.

PR #7814, opened 2024-08-07
Remove stale `cloud-tpu-client` dependency

will-cromar:
Removing `cloud-tpu-client` since we only used this for 2VM code (thanks @sagelywizard for catching this).

Adding explicit `numpy` dep because we import it all over the place.


sagelywizard:
Thank you! :)

JackCaoG (approved):

PR #7815, opened 2024-08-07
Revert "make StepMarker's wait block for async device execution"

JackCaoG:
Reverts pytorch/xla#7794

PR #7816, opened 2024-08-07
Use local TPU log dir in CI

will-cromar:
Hopefully fixes this log spam: `Could not open the log file '/tmp/tpu_logs/tpu_driver.v4-runner-set-7q76q-runner-ndh9h.runner.log.INFO.20240806-180726.2021': Permission denied`


will-cromar:
This fixes the log spam and fails at the parallel loader test that #7815 already fixed

JackCaoG (approved):

PR #7817, opened 2024-08-07
Initial pr to support dynamic shape detection

JackCaoG:
The implementation in this pr is not ideal because it can only handle a single graph for a compiled program. This is usually too strict for the real world program because in training we usually at least have 2 graphs since in the step0 the optimizer state is not initialized. 

I think what I need to do next is to provide a `max_dynamic_graph_allowed` arguments and use a TRIE to keep tracked of all the IR being generated and when they diverage.


JackCaoG:
rioght now if I used this feature in https://github.com/pytorch/xla/blob/master/examples/train_decoder_only_base.py it will error out because second step graph is different from the first step one(most likely due to the optimizer state).

JackCaoG:
Close in favor of https://github.com/pytorch/xla/pull/7918

PR #7834, opened 2024-08-12
[WIP] Skip AOTAutograd view-reconstruction for inference runs.

ysiraichi:
This PR re-implements part of the AOTAutograd view-reconstruction logic, so that we are able to skip it on inference runs. Most of the discussion can be found in https://github.com/pytorch/xla/issues/5430. This should avoid falling back to non-dynamo tracing outside the compiled graph.

cc @miladm @JackCaoG

### DIFF
--- a/torch_xla/core/dynamo_bridge.py
+++ b/torch_xla/core/dynamo_bridge.py
@@ -741,12 +741,6 @@ class MaybeReconstructOutputs:
           info.functional_tensor.tensor,
         )
       else:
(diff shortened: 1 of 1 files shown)


ysiraichi:
This PR is still being tested on Torchbench. It should be compiled with [this branch of PyTorch](https://github.com/pytorch/pytorch/compare/main...ysiraichi:pytorch:skip-aot-view-reconstruction?expand=1).

PR #7848, opened 2024-08-13
Introduce the `scan` interface

tengyifei:
We define the interface of the `scan` function and add test to verify its behavior. It is written referencing the JAX implementation.

The `scan` function is currently implemented as a for loop to keep each PR size small. In the next PR, we will replace the forward pass with an optimization implementation that lowers to an HLO `While` op.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -195,6 +195,7 @@ function run_xla_op_tests1 {
 function run_xla_op_tests2 {
   run_test "$CDIR/pjrt/test_dtypes.py"
   run_test "$CDIR/test_while_loop.py"
+  run_test "$CDIR/test_scan.py"
   run_test "$CDIR/test_autocast.py"
   run_test "$CDIR/eager/test_eager.py"
   run_test "$CDIR/eager/test_eager_with_xla_compile.py"
--- a/test/test_scan.py
+++ b/test/test_scan.py
@@ -1,0 +1,108 @@
+import sys
+import unittest
+import torch_xla
+import torch
+from torch_xla.experimental.scan import scan
+from torch.utils._pytree import tree_map, tree_flatten, tree_iter
+
+from test.test_utils import XlaTestCase
+
+
+def _loopy_scan(fn, init, xs):
+  """A simple scan implemented with for loops serving as reference
+  implementation."""
+  carry = init
+  ys = []
+  xs_len = len(next(iter(tree_iter(xs))))
+  for i in range(xs_len):
+    carry, y = fn(carry, tree_map(lambda x: x[i], xs))
+    ys.append(y)
+  ys = tree_map(lambda *x: torch.stack(x), *ys)
+  return carry, ys
+
+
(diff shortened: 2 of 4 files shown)


test/run_tests.sh
```diff
@@ -194,6 +194,7 @@ function run_xla_op_tests1 {
 function run_xla_op_tests2 {
   run_test "$CDIR/pjrt/test_dtypes.py"
   run_test "$CDIR/test_while_loop.py"
+  run_test "$CDIR/test_scan.py"
```
JackCaoG, line 197: maybe add it to the `test/tpu/run_test.sh` as well

test/test_scan.py
```diff
@@ -0,0 +1,108 @@
...
+  for i in range(xs_len):
+    carry, y = fn(carry, tree_map(lambda x: x[i], xs))
+    ys.append(y)
+  ys = tree_map(lambda *x: torch.stack(x), *ys)
+  return carry, ys
+
+
+class ScanTest(XlaTestCase):
+
+  def __init__(self, methodName: str = "runTest") -> None:
```
JackCaoG, line 26: what's this `methodName ` used for?

torch_xla/experimental/scan.py
```diff
@@ -0,0 +1,95 @@
...
+
+  if xs_length is None:
+    raise ValueError(f"`xs` {xs} is an empty PyTree.")
+
+  carry = init
+  ys = []
+
+  for i in range(xs_length):
+    carry, y = fn(carry, tree_map(lambda x: x[i], xs))
+    ys.append(y)
```
JackCaoG, line 90: why do we need per layer activation `ys`?

JackCaoG: is it for the backward? for the user the output of the scan forward should be the last activation, not the activations stacking together,

tengyifei: > why do we need per layer activation ys?

In the prototype `apply_layers` (public URL: https://colab.research.google.com/drive/18pPBRDPwSE6S9ZzvMkpVJuAlfJ1ezgsl#scrollTo=B6Q49K72Z9V2&line=6&uniqifier=1), I'll be using the `carry` for tracking per layer activation. The `ys` is unused. `apply_layers` currently passes a bunch of zeros and then discards `ys`, thus cutting that part out of the graph. I believe the JAX version of `scan` supports a `fn` returning `None` for this reason. I can look into aligning with that behavior of JAX either as a follow up PR or in this PR depending on your preference.

> is it for the backward?

Based on the previous explanation, I'm currently not using this for backwards. Although now that you mention this, it sounds like an interesting possibility. My design doc contains a description of how backwards is supported in the prototype.

test/test_scan.py
tengyifei: It looks like part of the method signature of a Python `unittest` test case. Although after consulting the docs, looks like we can use `setUp` instead.

JackCaoG (approved):

PR #7860, opened 2024-08-15
Support `dist.all_gather` related collective ops

zpcore:
Add dynamo/nondynamo support for `torch.distributed.all_reduce` and `torch.distributed.all_gather_into_tensor`.

Motivation: We want to deprecate the collective ops in xla_model.py and be consist with the torch.distributed.

Issue: dist.all_reduct doesn't work with dynamo openxla backend at this time.

### DIFF
--- a/test/pjrt/test_collective_ops_tpu.py
+++ b/test/pjrt/test_collective_ops_tpu.py
@@ -225,7 +225,6 @@ class TestDistCollectiveOpsTpu(parameterized.TestCase):
         self._all_gather_into_tensor, use_dynamo=use_dynamo)
     w = xr.world_size()
     expected = torch.arange(w, dtype=torch.float)
-    expected = torch.tensor([0, 1, 2, 3], dtype=torch.float)
     for index, val in results.items():
       torch.allclose(val, expected)
 
--- a/test/pjrt/test_collective_ops_tpu.py
+++ b/test/pjrt/test_collective_ops_tpu.py
@@ -142,6 +142,11 @@ class TestXMCollectiveOpsTpu(parameterized.TestCase):
 # Test for collective ops from torch.distributed
 class TestDistCollectiveOpsTpu(parameterized.TestCase):
 
+  # TODO(zpcore): fix the openxla dynamo issue for inplace copy
+  @staticmethod
+  def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
+    return gm.forward
+
   @staticmethod
   def _all_reduce(use_dynamo: bool):
     met.clear_all()
--- a/test/pjrt/test_collective_ops_tpu.py
+++ b/test/pjrt/test_collective_ops_tpu.py
@@ -3,6 +3,7 @@ from typing import List
 import torch
 import torch.nn as nn
 import torch.distributed as dist
+import torch.utils._pytree as pytree
 from absl.testing import absltest, parameterized
 from unittest import mock
 import torch_xla
@@ -141,7 +142,7 @@ class TestXMCollectiveOpsTpu(parameterized.TestCase):
 # Test for collective ops from torch.distributed
 class TestDistCollectiveOpsTpu(parameterized.TestCase):
 
-  # TODO(zpcore): fix the openxla dynamo issue for dist.all_reduce
+  # TODO(zpcore): fix the openxla dynamo issue for dist.all_reduce, dist.all_gather
   @staticmethod
   def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
     return gm.forward
@@ -150,29 +151,26 @@ class TestDistCollectiveOpsTpu(parameterized.TestCase):
   def _all_reduce(use_dynamo: bool):
     met.clear_all()
 
-    def _dynamo_func_wrapper(input):
+    def _dist_all_reduce(input):
       dist.all_reduce(input, dist.ReduceOp.SUM)
       return input
 
     dist.init_process_group("xla", init_method='xla://')
     device = xm.xla_device()
-    if xm.xla_device_hw(device) not in ('TPU', 'CUDA', 'NEURON'):
-      print(f'skip this test for hw {xm.xla_device_hw(device)}')
-      return
     ordinal = torch.tensor([xr.global_ordinal()],
                            dtype=torch.float,
                            device=device)
     input = ordinal
+
+    f = torch.compile(
+        _dist_all_reduce, backend=TestDistCollectiveOpsTpu.my_compiler
+    ) if use_dynamo else _dist_all_reduce
+    f(input)
+    torch_xla.sync()
     if not use_dynamo:
-      dist.all_reduce(input)
-      xm.mark_step()
       assert 'xla::AllReduceInPlace' in met.counter_names(
       ) or 'xla::AllReduce' in met.counter_names()
     else:
-      compiled_collective = torch.compile(
-          _dynamo_func_wrapper, backend=TestDistCollectiveOpsTpu.my_compiler)
-      compiled_collective(input)
-      torch_xla.sync()
       assert 'xla::all_reduce' in met.counter_names()
     return input.cpu()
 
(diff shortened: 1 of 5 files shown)


zpcore:
Hi @JackCaoG , I commented out `assert met.metric_data("ExecuteTime")[0] == 1` in the test_traceable_collectives.py since the value changed to 3 for all_gather ops due to upstream changes. Shall we enable the test first so we can run the test?

zpcore:
I have no idea how this works: https://github.com/pytorch/xla/blob/37312c199ecb4b90b3df52349c9eb686200a08bf/torch_xla/distributed/xla_backend.py#L71-L72 

`xm.all_reduce` should return a tensor instead of modifying `inputs` argument. 

I will probably give up supporting `dist.all_reduce` in this PR.

test/dynamo/test_traceable_collectives.py
```diff
@@ -43,8 +55,35 @@ def _mp_fn(index):
...
+        [world_size * world_size / 2] * world_size, dtype=torch.float) + 3.0
+    torch_xla.sync()
+    torch.allclose(res_tensor.cpu(), expected_tensor)
+    # TODO (JackCaoG): fix CI issue and uncomment below.
+    # assert met.metric_data("ExecuteTime")[0] == 1
 
 
 if __name__ == '__main__':
   torch_xla.launch(_mp_fn, args=())
+  torch_xla.launch(_mp_fn_dist, args=(), debug_single_process=False)
```
will-cromar, line 89: Please write a new unit test for your functionality. This test checks a different code path (that `xm.all_reduce` calls a functional collective). You may use https://github.com/pytorch/xla/blob/master/test/pjrt/test_collective_ops_tpu.py as an example. Note that each test case calls one collective op.

torch_xla/distributed/xla_backend.py
```diff
@@ -46,6 +46,13 @@ def __init__(self, prefix_store, rank, size, timeout):
   def getBackendName(self):
     return 'xla'
 
+  def _set_group_name(self, name: str) -> None:
+    self._group_name = name
+
+  @property
+  def group_name(self):
```
will-cromar, line 53: Leave comment here explaining why we need to do this. It's very likely a PyTorch bug, since there's already a default implementation here: https://github.com/pytorch/pytorch/blob/7b1988f9222f3dec5cc2012afce84218199748ae/torch/csrc/distributed/c10d/ProcessGroup.cpp#L148-L152

We can file that bug separately, and clean this up later.

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -533,6 +533,17 @@ std::pair<std::vector<XLATensorPtr>, torch::lazy::Value> all_gather_coalesced(
   return {result, torch::lazy::Value(node, inputs.size())};
 }
 
+std::vector<XLATensorPtr> all_gather_coalesced_wrapper(
```
will-cromar, line 536: Can you handle coalesced all gather in a future PR?

test/dynamo/test_traceable_collectives.py
zpcore: Added the testcase and included the metric counter.

torch_xla/distributed/xla_backend.py
```diff
@@ -46,6 +46,16 @@ def __init__(self, prefix_store, rank, size, timeout):
   def getBackendName(self):
     return 'xla'
 
+  # pytorch's progress group is unable to retrive the group size from python level. It should
```
will-cromar, line 49: nit: "process group"

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -132,5 +138,96 @@ def test_all_to_all(self, pin_layout):
...
+    else:
+      compiled_collective = torch.compile(
+          _dynamo_func_wrapper, backend='openxla')
+      output_tensor = compiled_collective(input)
+      torch_xla.sync()
+      assert 'xla::all_gather_into_tensor' in met.counter_names()
+    return output_tensor.cpu()
+
+  @parameterized.named_parameters(('dynamo', True), ('nondynamo', False))
+  @mock.patch.object(xr, 'world_size', return_value=tpu.num_local_processes())
```
will-cromar, line 213: Why is world size mocked out?

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -132,5 +138,96 @@ def test_all_to_all(self, pin_layout):
...
+  def test_all_reduce(self, use_dynamo, mock_world_size):
+    results = pjrt.run_multiprocess(self._all_reduce, use_dynamo=use_dynamo)
+    w = xr.world_size()
+    expected = torch.tensor([(w - 1) * w / 2], dtype=torch.float)
```
will-cromar, line 217: Haha, `[(w - 1) * w / 2]` is indeed correct here, but it's better to be clear than performant for something that is not in the hot path at run time. Prefer something explicit like `sum(range(etc))`, especially for testing purposes

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -132,5 +138,96 @@ def test_all_to_all(self, pin_layout):
...
+    for index, val in results.items():
+      torch.allclose(val, expected)
```
will-cromar, line 219: This is not an assertion -- it's returning an unused boolean. You were probably looking for `torch.testing.assert_close`.

In fact, since that's not a test case method (like `self.assert*`), you can simplify the logic here a bit by putting `assert_close` inside of `_all_reduce` where it's easier to get the real world size.

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -132,5 +138,96 @@ def test_all_to_all(self, pin_layout):
...
+    else:
+      compiled_collective = torch.compile(
+          _dynamo_func_wrapper, backend=TestDistCollectiveOpsTpu.my_compiler)
+      compiled_collective(input)
+      torch_xla.sync()
+      assert 'xla::all_reduce' in met.counter_names()
+    return input.cpu()
+
+  @staticmethod
+  def _all_gather_into_tensor(use_dynamo: bool):
```
will-cromar, line 180: You can simplify this.

First, what we're really testing here is that the same code works for both dynamo and eager, so no need to rewrite the contents of `_dynamo_func_wrapper`. Just give it some generic name, and have a case like this:

```
f = torch.compile(_dist_allgather, ...) if use_dynamo else _dist_allgather
f(input)
torch_xla.sync()
```

You call the same callable either way, and you only conditionally `compile` it.

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -132,5 +138,96 @@ def test_all_to_all(self, pin_layout):
...
+    met.clear_all()
+
+    def _dynamo_func_wrapper(input):
+      output_tensor = torch.empty((1, xr.world_size()), device=device)
+      dist.all_gather_into_tensor(output_tensor, input, None)
+      return output_tensor
+
+    dist.init_process_group("xla", init_method='xla://')
+    device = xm.xla_device()
+    if xm.xla_device_hw(device) not in ('TPU', 'CUDA', 'NEURON'):
```
will-cromar, line 190: This case is not necessary since this is in a TPU-specific test file.

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -132,5 +138,96 @@ def test_all_to_all(self, pin_layout):
...
+      return
+    ordinal = torch.tensor([xr.global_ordinal()],
+                           dtype=torch.float,
+                           device=device)
+    input = ordinal
+
+    if not use_dynamo:
+      output_tensor = torch.empty((1, xr.world_size()), device=device)
+      dist.all_gather_into_tensor(output_tensor, input, None)
+      xm.mark_step()
```
will-cromar, line 201: `torch_xla.sync`

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -132,5 +138,96 @@ def test_all_to_all(self, pin_layout):
...
+# Test for collective ops from torch.distributed
+class TestDistCollectiveOpsTpu(parameterized.TestCase):
+
+  # TODO(zpcore): fix the openxla dynamo issue for dist.all_reduce
+  @staticmethod
+  def my_compiler(gm: torch.fx.GraphModule, example_inputs: List[torch.Tensor]):
+    return gm.forward
+
+  @staticmethod
+  def _all_reduce(use_dynamo: bool):
```
will-cromar, line 150: See my comments on allgather

test/pjrt/test_collective_ops_tpu.py
zpcore: It's a work around to call `pjrt.run_multiprocess` multiple times due to all `@parameterized.named_parameters` wrapped test functions are launched by reusing the same process one by one.

Without the mock, this error will be triggered: https://github.com/pytorch/xla/blob/ae28308bed8ef2a2086b3d7993fce766b34daf3b/torch_xla/_internal/pjrt.py#L144-L146
If we are not calling xr.world_size(), then `_xla_get_replication_devices_count` and `global_device_count` will not be called: https://github.com/pytorch/xla/blob/ae28308bed8ef2a2086b3d7993fce766b34daf3b/torch_xla/runtime.py#L136-L145  thus we will not trigger the `std::atomic<bool> g_computation_client_initialized`.

zpcore: Just tested that `tpu.num_local_processes()` or `tpu.num_expected_global_devices()` will break the dynamo graph. I think stick to `runtime.world_size()` and use the mock object will be most straightforward way for the unit test. WDYT?

will-cromar: Hmm, that doesn't sound right

`Runtime is already initialized` happens if you do `xr.world_size` in the same process where you call `launch`, so that's expected. But, you really only need it for your test assertion, which you can put with the other test assertions in `self._all_reduce`.

The only dynamo compiled function is `_dist_allgather`. Where are `tpu.num_local_processes()` or `tpu.num_expected_global_devices()` called in there? Also I'm 90% sure this mock does not apply to anything in the child process spawned by `pjrt.run_multiprocess`

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -132,5 +139,126 @@ def test_all_to_all(self, pin_layout):
...
+      ) or 'xla::AllGatherOut' in met.counter_names()
+    else:
+      assert 'xla::all_gather_into_tensor' in met.counter_names()
+    return output_tensor.cpu()
+
+  @staticmethod
+  def _all_gather(use_dynamo: bool):
+    met.clear_all()
+
+    def _dist_allgather(input):
```
will-cromar, line 209: nit: make these inner functions have consistent naming patterns. Since they're used in such a limited context, you can also just given them all the same terse name like `callable`, `cc`, `program`, `model` etc.

test/pjrt/test_collective_ops_tpu.py
zpcore: OK, I think the thumb of rule to prevent the `Runtime is already initialized...` error is:
Never initialize PJRT client (e.g., call xr.world_size()) before calling `pjrt.run_multiprocess()`.

For parameterized test that calls `pjrt.run_multiprocess()`, do not call `xr.world_size()` in the main test process. Because `pjrt.run_multiprocess()` will be triggered in the next parameter test, which violates rule.

I have removed the xr.world_size() mock object and use `tpu.num_expected_global_devices()` in the test process instead.

will-cromar (approved):

rpsilva-aws:
> Motivation: We want to deprecate the collective ops in xla_model.py and be consist with the torch.distributed.

@zpcore This makes sense, and we would like to help close that gap. Do you have a more descriptive motivation and/or opened issue to migrate the remaining collective ops? I tried finding, but checking in first before I create one. cc: @miladm @tengyifei

Issue #7859, opened 2024-08-15
unsqueeze returns wrong results when input is a sliced tensor with non-trivial offset


yoavhacohen:
## 🐛 Bug

unsqueeze returns wrong results when input is a sliced tensor with non-trivial offset.
NOTE: This bug only happens when 

## To Reproduce

Run this from the command line:
```bash
export XLA_DISABLE_FUNCTIONALIZATION=1
```

Run this code:

```
import torch

try:
    # noinspection PyUnresolvedReferences
    import torch_xla
    import torch_xla.core.xla_model as xm
    import torch_xla.distributed.parallel_loader as pl
    import torch_xla.runtime as xr
    from torch_xla.utils import checkpoint as xla_checkpoint
    _torch_xla_available = True
except ImportError:
    _torch_xla_available = False

if _torch_xla_available:
    device = xm.xla_device()
else:
    assert torch.cuda.is_available()
    device = torch.device("cuda")
    
x = torch.randn(size=[2], device=device)
a1, b1 = x.chunk(2)
a2, b2 = x[0:1], x[1:2]
a3, b3 = x[0].unsqueeze(0), x[1].unsqueeze(0)
a4, b4= x[0, None], x[1, None]

def is_ok(a, b):
    return "OK" if a.squeeze() == x[0] and b.squeeze() == x[1] else "Not OK"

print(f"x: {x}")
print(f"a1, b1: {a1}, {b1}, a1.squeeze(), b1.squeeze(): {a1.squeeze()}, {b1.squeeze()} - {is_ok(a1, b1)}")
print(f"a2, b2: {a2}, {b2}, a2.squeeze(), b2.squeeze(): {a2.squeeze()}, {b2.squeeze()} - {is_ok(a2, b2)}")
print(f"a3, b3: {a3}, {b3}, a3.squeeze(), b3.squeeze(): {a3.squeeze()}, {b3.squeeze()} - {is_ok(a3, b3)}")
print(f"a4, b4: {a4}, {b4}, a4.squeeze(), b4.squeeze(): {a4.squeeze()}, {b4.squeeze()} - {is_ok(a4, b4)}")
```

The actual result is:
```
x: tensor([ 1.3556, -0.1277], device='xla:0')
a1, b1: tensor([1.3556], device='xla:0'), tensor([-0.1277], device='xla:0'), a1.squeeze(), b1.squeeze(): 1.3555867671966553, 1.3555867671966553 - Not OK
[...]

PR #7864, opened 2024-08-15
As_stride should aplly on input tensor, not the base tensor

JackCaoG:
This should fix https://github.com/pytorch/xla/issues/7859

at least in the non-functionization case(in the FUNCTIONALIZATION case `as_strided_copy` will be call). the as_strided call should be applied on the input tensor, not the base tensor.

@bdhirsh Do you know when user do
```
x.view().as_strided()
```
in the FUNCTIONALIZATION and non-FUNCTIONALIZATION case, should the `as_strided` be applied on `x` or `x.view`? in the 

```
x = torch.randn(size=[2])
a1, b1 = x.chunk(2)
res = b1.seqeeze()
```
case without FUNCTIONALIZATION, I see that it is a `b1` is a
```
IR {
  %0 = f32[2]{0} xla::device_data(), xla_shape=f32[2]{0}
  %1 = f32[1]{0} xla::select(%0), xla_shape=f32[1]{0}, ROOT=0
}
```
and then `seqeeze` becomes a `as_strided`  with `[], []`. In this case `as_strided` should apply on `b1`, not the x.

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2704,6 +2704,8 @@ class TestGeneric(test_utils.XlaTestCase):
     self.assertEqual(xdata.batch_sizes.device, torch.device('cpu'))
     self.assertEqual(xdata.data.device, xla_device)
 
+  @skipIfFunctionalizationDisabled(
+      "https://github.com/pytorch/xla/pull/7864#issuecomment-2294034008")
   def test_as_strided_input_larger(self):
     size = (5, 5)
     device = xm.xla_device()
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -1006,6 +1006,19 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
     b = torch.ones([2, 2])
     self.runAtenTest((a, b), func)
 
+  def test_multi_view(self):
+
+    def func(x):
+      a1, b1 = x.chunk(2)
+      a2, b2 = x[0:1], x[1:2]
+      a3, b3 = x[0].unsqueeze(0), x[1].unsqueeze(0)
+      a4, b4 = x[0, None], x[1, None]
+      return a1.squeeze(), b1.squeeze(), a2.squeeze(), b2.squeeze(), a3.squeeze(
+      ), b3.squeeze(), a4.squeeze(), b4.squeeze()
+
+    x = torch.randn(size=[2])
+    self.runAtenTest(x, func)
+
   # TODO - upstream behavior has changed and results in expected DestroyXlaTensor
   # counter as of 11/13/2023. Re-enable after reviewing the change.
   # @skipIfFunctionalizationDisabled("metrics differ")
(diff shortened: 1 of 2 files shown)


qihqi (approved):

test/test_operations.py
```diff
@@ -1006,6 +1006,19 @@ def func(a, b):
...
     self.runAtenTest((a, b), func)
 
+  def test_multi_view(self):
+
+    def func(x):
+      a1, b1 = x.chunk(2)
+      a2, b2 = x[0:1], x[1:2]
+      a3, b3 = x[0].unsqueeze(0), x[1].unsqueeze(0)
+      a4, b4 = x[0, None], x[1, None]
+      return a1.squeeze(), b1.squeeze(), a2.squeeze(), b2.squeeze(), a3.squeeze(
```
alanwaketan, line 1016: What will this return? I'm curious on how squeezing a [1] tensor works.... just another [1] tensor?

JackCaoG: it actually will return a scalar
```
>>> import torch
>>> x = torch.randn([2])
>>> x
tensor([-0.0293, -0.5956])
>>> x[0:1]
tensor([-0.0293])
>>> x[0:1].squeeze()
tensor(-0.0293)
```

alanwaketan: Perfect!

alanwaketan (approved):

JackCaoG:
failure is real, need to look into them tmr.

ysiraichi:
Read my comment: [here](https://github.com/pytorch/xla/issues/7859#issuecomment-2294023908)
Basically, we are choosing which issue to fix. Either we fix:

1. Calls to `as_strided` on a view tensor without specifying `storage_offset`; or
2. Calls to `as_strided` on a view tensor where we need more storage than what the view tensor has

In order to fix both, we need to start keeping track of the `storage_offset` of view tensors.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -4167,15 +4167,13 @@ at::Tensor XLANativeFunctions::as_strided(
     const at::Tensor& self, at::IntArrayRef size, at::IntArrayRef stride,
     std::optional<int64_t> storage_offset) {
   TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
-  const auto& base = bridge::GetXlaTensor(self)->Base();
-  const auto& tensor = base.defined() ? base : self;
-  XLATensorPtr self_tensor = bridge::GetXlaTensor(tensor);
+  XLATensorPtr self_tensor = bridge::GetXlaTensor(self);
```
ysiraichi, line 4170: Since we are explicitly not supporting larger tensor use-case, what do you think about raising an error earlier (at tracing time), instead of later (at lowering time)?

ysiraichi (approved):

torch_xla/csrc/aten_xla_type.cpp
JackCaoG: ah, is there an error message/check I can borrow?

ysiraichi: Hmmm. I don't think so. I think you have to check yourself whether the storage needed for the size-stride-storageoffset specification fits the current storage (i.e. number of elements in the tensor)

ysiraichi: There's a `at::detail::computeStorageNbytes` function in PyTorch.

JackCaoG: ok let me do that in a follow up pr, I want to merge this pr to unblock the customer.

PR #7870, opened 2024-08-16
[torchxla2] Fix README venv setup

gmittert:
python -m venv doesn't take a `create` position command, just a list of ENV_DIR to output to.


qihqi (approved):

qihqi:
good catch! thanks!

PR #7722, opened 2024-07-22
update wheel name

zpcore:
Resolves https://github.com/pytorch/xla/issues/7697.

Input name like `torch_xla-2.5.0+git41d998d` will be changed from `torch_xla-nightly+20240722-cp` to `torch_xla-2.5.0+nightly20240722-cp`

PR #7872, opened 2024-08-17
Revert "update wheel name"

zpcore:
Reverts pytorch/xla#7722.

@JackCaoG  noticed some issues regarding the TPU installation:  https://github.com/pytorch/xla#tpu. Revert the PR first to unblock.


JackCaoG:
this is not the real issue, we should update the readme

PR #7874, opened 2024-08-17
Op info test for `special.i1 .. special.modified_bessel_i0`

### LINKED ISSUE #7536: Op info test for `special.i1 .. special.modified_bessel_i0`
## Fix the Op info test for `special.i1 .. special.modified_bessel_i0`

1. Find the lines 303 to 307 of [test_ops.py](test/test_ops.py) and remove
   `special.i1` .. `special.modified_bessel_i0` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ayshiff:
Implementation of the following ops:

- `special_i1`
- `special_i1e`
- `special_laguerre_polynomial_l`
- `special_modified_bessel_i0`
- `erfc `

Fixes #7536


experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -2377,6 +2377,152 @@ def _aten_i0e(self):
...
+
+    def zero_abs(x):
+      return jnp.ones_like(x)
+
+    def default(x):
+      def f(k, carry):
+        p, q = carry
+        return (q, ((k * 2 + (jnp.ones_like(x) - x)) * q - k * p) / (k + 1))
+
+      _, q = jax.lax.fori_loop(1, n_i, f, init_val=(1.0, jnp.ones_like(x) - x))
```
qihqi, line 2416: nit: is it possible to use `jax.lax.scan` instead of fori here?

Reason: `jax.lax.scan` is differentiable so one can use `jax.grad` on it.

ayshiff: I've been trying to implement the equivalent logic using `jax.lax.scan`, but I'm running into some challenges.

#### What I've Tried:

* My initial approach was to create an indices array using `jnp.arange` within the scan loop, but this led to `ConcretizationTypeError` because the stop value (`n_i`) could be a `ShapedArray`.

```py
def default(x):
  def f(carry, k):
    p, q = carry
    return (q, ((k * 2.0 + (jnp.ones_like(x) - x)) * q - k * p) / (k + 1.0)), None
      
  indices = jnp.arange(1, stop=n_i)

  (_, q), _ = jax.lax.scan(f, init=(1.0, 1.0 - x), xs=indices)
  return q
```

* I also tried including the index in the carry value of the scan operation, but I haven't been able to get it to work correctly.

#### Note:

* I was able to successfully use `scan` for the `_aten_special_modified_bessel_i0` implementation

I can keep looking for potential solutions or workarounds. Any insights or suggestions would be greatly appreciated!

qihqi: Yeah seems like scan only works well when the array is given. I guess we could mark `n` as constant but that seems less ideal than fori; as it will trigger recompile more often. 

Lets keep the loop here.

qihqi (approved):

PR #7876, opened 2024-08-18
[developer] Add a script to automate building torch_xla

tengyifei:
This script replicates our CONTRIBUTING.md instructions. It builds PyTorch, TorchVision (optional), and PyTorch/XLA. Then it runs a test to verify the installation.

This is useful after git pulling all three repos.

### DIFF
--- a/scripts/build_developer.sh
+++ b/scripts/build_developer.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -e  # Fail on any error.
+set -x  # Display commands being run.
+
+# Change to pytorch directory.
+cd "$(dirname "$(readlink -f "$0")")"
+cd ../../
+
+python3 setup.py bdist_wheel
+python3 setup.py install
+cd ..
+
+# Optionally install torchvision.
+if [ -d "vision" ]; then
+  cd vision
+  python3 setup.py develop
+fi
+
+cd ..
+cd pytorch/xla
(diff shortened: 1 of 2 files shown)


scripts/build_developer.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
will-cromar, line 1: This is fine with me, as long as we keep it this simple and we continue to keep automated processes using the configs in `infra/ansible`.

scripts/build_developer.sh
```diff
@@ -0,0 +1,30 @@
...
+
+set -e  # Fail on any error.
+set -x  # Display commands being run.
+
+# Change to pytorch directory.
+cd "$(dirname "$(readlink -f "$0")")"
+cd ../../
+
+python3 setup.py bdist_wheel
+python3 setup.py install
```
will-cromar, line 11: Can you install the package produced by the last line directly? ie `pip install dist/*.whl`

qihqi (approved):

scripts/build_developer.sh
tengyifei: I'll keep this the same as the CONTRIBUTING.md for now: https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md, which has worked for people for a while.

PR #7884, opened 2024-08-20
DYNAMO RNG seed update optimization

JackCaoG:
RNG seed had 2 forms in torch_xla, a xla tensor and a scalar. In the dynamo bridge case we want the tensor form for the execution but for the seed update we don't want to do `tensor.item()` since it will cause a `transferFromDevice`. We can directly access the scalar form of the rng seed

### DIFF
--- a/test/dynamo/test_dynamo.py
+++ b/test/dynamo/test_dynamo.py
@@ -54,12 +54,18 @@ class DynamRandomOpTest(unittest.TestCase):
     return torch.randn(5, 5, device=a.device) + a
 
   def test_random_op_different_result_each_run(self):
+    xm.wait_device_ops()
+    met.clear_all()
     dynamo_random_op = torch.compile(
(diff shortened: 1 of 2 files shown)


lsy323 (approved):

test/dynamo/test_dynamo.py
```diff
@@ -54,12 +54,18 @@ def random_op(self, a):
     return torch.randn(5, 5, device=a.device) + a
 
   def test_random_op_different_result_each_run(self):
+    xm.wait_device_ops()
```
lsy323, line 57: Just curious why do we need a barrier only for this test

lsy323 (approved):

test/dynamo/test_dynamo.py
JackCaoG: it is just to be safe for unit tests, there might be async operations that update the metrics. We want to wait for all async ops to finish and then clear the metrics

Issue #7530, opened 2024-06-25
Op info test for `qr .. randn_like`


qihqi:
## Fix the Op info test for `qr .. randn_like`

1. Find the lines 273 to 277 of [test_ops.py](test/test_ops.py) and remove
   `qr` .. `randn_like` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

danielvegamyhre:
Note on `randn_like` op: it appears the op is implemented in torch_xla2 [here](https://github.com/pytorch/xla/blob/ac7fd44f520d586c3d95fc929134c20946e60806/experimental/torch_xla2/torch_xla2/ops/jaten.py#L2252) but tests fail for float32 and int64 dtypes:

```
FAILED test/test_ops.py::TestOpInfoCPU::test_reference_eager_randint_like_cpu_float32 - torch_xla2.tensor.OperatorNotFound: Operator with name aten::randint_like has no lowering

FAILED test/test_ops.py::TestOpInfoCPU::test_reference_eager_randint_like_cpu_int64 - torch_xla2.tensor.OperatorNotFound: Operator with name aten::randint_like has no lowering
```

ManfeiBai:
thanks for contribution, looks like we have `quantile` not been treated yet, so reopen this issue before totally close it

danielvegamyhre:
> thanks for contribution, looks like we have `quantile` not been treated yet, so reopen this issue before totally close it

I will finish quantile soon. Thanks!

danielvegamyhre:
Update: while debugging `quantile` I found of the sub-ops it dispatches to (`aten::gather`) is returning an incorrect result, which I am now investigating:

```
 DISPATCH: aten::gather
> ~/dev/xla/experimental/torch_xla2/torch_xla2/tensor.py(180)debug_accuracy()
-> for ex, real in zip(flattened_expected_out, flattened_current_out):
(Pdb) l
175       expected_out = func(*args_torch, **kwargs_torch)
176  
177       flattened_current_out, _ = torch_pytree.tree_flatten(out_torch)
[...]

PR #7885, opened 2024-08-20
[torch_xla2] Add support for QR op

danielvegamyhre:
This change adds support for QR matrix decomposition op in torch_xla2.

Note: the [op db](https://github.com/pytorch/xla/blob/ac7fd44f520d586c3d95fc929134c20946e60806/experimental/torch_xla2/test/test_ops.py#L5) used in the version of PyTorch being used in torch_xla2 tests is using [torch.qr](https://pytorch.org/docs/stable/generated/torch.qr.html) which is going to be deprecated and removed in a future release, so we need to support the newer [torch.linalg.qr](https://pytorch.org/docs/stable/generated/torch.linalg.qr.html#torch.linalg.qr) whose params align with [jax.numpy.linalg.qr](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.qr.html).

### DIFF
--- a/experimental/torch_xla2/test/test_ops.py
+++ b/experimental/torch_xla2/test/test_ops.py
@@ -247,7 +247,6 @@ skiplist = {
     "polygamma",
     "prod",
     "put",
-    "qr",
     "quantile",
     "renorm",
     "repeat_interleave",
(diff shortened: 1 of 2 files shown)


danielvegamyhre:
Part of #7530

qihqi (approved):

qihqi:
thanks!

PR #7888, opened 2024-08-20
Allows Llama 3 example to run on TPUs with multiple VMs (16+) (XLA2)

zmelumian:
On torch_xla2 experimental code, there's an example of llama_training

currently, the code does work on TPUs with multiple VMs

I added support for it via sharding

I witnessed significant reduction in compilation time, however I noticed that the reported runtime does not differ much from the single-VM scenario.


qihqi (approved):
this is awesome! thanks!

PR #7901, opened 2024-08-23
scan and scan_layers

tengyifei:
Add the lowering of scan to HLO While op.

Introduce scan_layers which can sequentially apply a bunch of layers
using scan underneath.

Beef up unit tests including linear layers and decoders.


tengyifei:
@JackCaoG ty. i followed your example and got it working.

test/test_apply_layers.py
```diff
@@ -0,0 +1,149 @@
...
+
+  def setUp(self):
+    super().setUp()
+
+    self.device = torch_xla.device()
+
+  def test_empty_layers(self):
+    layers = []
+    input_data = torch.randn(64).to(self.device)
+    torch_xla.sync()
```
JackCaoG, line 30: you don't need this line I guess, `to(device)` will not generate ir. it only generate ir if you do `torch.randn(64, device=device)`

test/test_apply_layers.py
```diff
@@ -0,0 +1,149 @@
...
+    output = apply_layers(layers, input_data.clone())
+    super().compareResults(output, input_data, abs_err=0.0001, rel_err=0.01)
```
JackCaoG, line 32: why not `torch.allclose` similar to other tests?

test/test_apply_layers.py
```diff
@@ -0,0 +1,149 @@
...
+
+  def test_linear_layers(self):
+    # We want to apply these layers sequentially
+    import torch.nn as nn
```
JackCaoG, line 36: might as well just move it to the top given you might want to expand tests in the future

test/test_apply_layers.py
```diff
@@ -0,0 +1,149 @@
...
+    scan_layers = deepcopy(layers)
+    loop_layers = deepcopy(layers)
+    torch_xla.sync()
+
+    output = apply_layers(scan_layers, input_data.clone())
+    output.sum().backward()
+
+    # Test that the result is the same as for loop.
+    loop_output = input_data.clone()
+    from copy import deepcopy
```
JackCaoG, line 50: move this to the top, you also don't need to import twice

test/test_apply_layers.py
```diff
@@ -0,0 +1,149 @@
...
+    torch_xla.sync()
+
+    super().compareResults(loop_output, output, abs_err=0.0001, rel_err=0.01)
+
+    loop_output.sum().backward()
+    torch_xla.sync()
+
+    # Test that the gradients are the same too.
+    for layer_scan, layer_loop in zip(scan_layers, loop_layers):
+      super().compareResults(
```
JackCaoG, line 62: same, why not `torch.allclose`?

test/test_apply_layers.py
```diff
@@ -0,0 +1,149 @@
...
+          layer_scan.weight.grad,
+          layer_loop.weight.grad,
```
JackCaoG, line 64: add another check that they are not the same tensor. In case in the future someone messed up the deepcopy then this check become meaningless.

test/test_apply_layers.py
```diff
@@ -0,0 +1,149 @@
...
+      def forward_scan(
+          self,
+          input_ids: torch.Tensor,
+      ) -> torch.Tensor:
+        inputs_embeds = self.decoder.embed_tokens(input_ids)
+        # embed positions
+        assert isinstance(inputs_embeds, torch.Tensor)
+        # decoder layers
+        hidden_states = apply_layers(self.decoder.layers, inputs_embeds)
+        hidden_states = self.decoder.norm(hidden_states)
```
JackCaoG, line 101: why is there a norm here? is it the last layer?

test/test_apply_layers.py
```diff
@@ -0,0 +1,149 @@
...
+           param_scan), (name2,
+                         param_loop) in zip(layer_scan.named_parameters(),
+                                            layer_loop.named_parameters()):
+        assert name == name2
+        if param_scan.grad is not None or param_loop.grad is not None:
+          super().compareResults(
+              param_scan.grad, param_loop.grad, abs_err=0.1, rel_err=0.05)
+          print(f"Pass: {name} {param_scan.shape}")
+
+
```
JackCaoG, line 146: add another test to test the HLO using `torch_xla._XLAC._get_xla_tensors_hlo` (check while is in the HLO)

test/test_operations.py
```diff
@@ -2597,6 +2598,29 @@ def test_api(self):
     mapping = ctx.parameter_id_tensor_mapping()
     self.assertEqual(len(mapping), 2)
 
+  def test_get_parameters_scalar(self):
+    """Scalar tensors parameters may be shared in the HLO graph if their
+    numerical values are equal. `parameter_id_tensor_mapping` needs to handle
+    that appropriately.
+    """
+
+    device = xm.xla_device()
```
JackCaoG, line 2607: `torch_xla.device()` is prefereed

test/test_operations.py
```diff
@@ -2597,6 +2598,29 @@ def test_api(self):
...
+      tensors.append(torch.tensor(i, device=device))
+    result = reduce(lambda a, b: a + b, tensors)
+    ctx = torch_xla._XLAC.lowering.LoweringContext()
+    ctx.build([result])
+    mapping = ctx.parameter_id_tensor_mapping()
+
+    import json
+    hlo_json = json.loads(ctx.hlo_json())
+    num_parameters = len(hlo_json["hostProgramShape"]["parameters"])
+    self.assertEqual(len(mapping), num_parameters)
```
JackCaoG, line 2622: so you expect both value to be 10?

test/test_scan.py
```diff
@@ -1,11 +1,14 @@
...
+
 import torch
+from torch.utils._pytree import tree_map, tree_flatten, tree_iter, tree_leaves, PyTree
+
+import torch_xla
 from torch_xla.experimental.scan import scan
-from torch.utils._pytree import tree_map, tree_flatten, tree_iter
 
-from test_utils import XlaTestCase
+from test_utils import XlaTestCase  # type:ignore
```
JackCaoG, line 11: did some formatter complained?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1077,7 +1076,9 @@ class PyLoweringContext {
       at::ScalarType dtype =
           MaybeUpcastToHostTorchType(literal.shape().element_type());
       at::Tensor input = MakeTensorFromXlaLiteral(literal, dtype);
-      results[param_ids[i]] = input;
+      std::optional param_id = lowering_ctx.GetParameterId(device_data[i]);
+      XLA_CHECK(param_id.has_value());
```
JackCaoG, line 1080: when would it not has value?

torch_xla/experimental/apply_layers.py
```diff
@@ -0,0 +1,72 @@
...
+  return final_carry
+
+
+def _extract_weights_dict(module: nn.Module):
+  """
+  Extracts the parameters (weights and biases) from a PyTorch module and
+  stores them in a dictionary.
+  """
+  weights_dict = {
+      name: param.clone() for name, param in module.named_parameters()
```
JackCaoG, line 61: hmm why clone here?

torch_xla/experimental/apply_layers.py
```diff
@@ -0,0 +1,72 @@
...
+  first layer to perform the computation of all other layers.
+  """
+  # Handle empty layers case.
+  try:
+    next(iter(layers))
+  except StopIteration:
+    return input_data
+
+  # Extract and stack the parameters into a pytree.
+  params = [_extract_weights_dict(layer) for layer in layers]
```
JackCaoG, line 29: what if it is a dropout layer that parameters are more than just tensors?

torch_xla/experimental/apply_layers.py
```diff
@@ -0,0 +1,72 @@
...
+  stacked_params = tree_map(lambda *tensors: torch.stack(tensors, dim=0),
+                            *params)
+
+  # Use the first layer as the example/template layer.
+  from copy import deepcopy
+  example_layer = deepcopy(next(iter(layers)))
+
+  # Hollow out the weights and biases in the example layer.
+  example_layer = example_layer.to_empty(device=None)
```
JackCaoG, line 38: is this not going to impact the cloned arg?

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+  return tree_unflatten(carry, carry_spec), tree_unflatten(
+      fn_carry_history, carry_spec), tree_unflatten(ys, fn_output_y_spec)
+
+
+@pytreeify
+class Scan(torch.autograd.Function):
+
+  @staticmethod
+  def forward(ctx, fn, init, xs):
+    # Forward pass, save inputs for backward
```
JackCaoG, line 363: This will most likely break the gradient checkpointing.

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+  fake_carry, _ = tree_flatten(fake_carry_pytree)
+  fake_x, _ = tree_flatten(fake_x_pytree)
+  fn_output_carry_pytree, fn_output_y_pytree = flat_fn(*(fake_carry + fake_x))
+
+  # Later we'll use `fn_output_carry_spec` etc to turn flattened outputs back to a PyTree.
+  fn_output_carry, fn_output_carry_spec = tree_flatten(fn_output_carry_pytree)
+  assert fn_output_carry_spec == carry_spec
+  fn_output_y, fn_output_y_spec = tree_flatten(fn_output_y_pytree)
+  flat_y_len = len(fn_output_y)
+  fn_outputs = fn_output_carry + fn_output_y
```
JackCaoG, line 194: what if there are in place updates to the tensor but it is not being return from the function?

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+    tensors_list = ctx.saved_tensors
+    carry_history = tree_unflatten(tensors_list[:flat_carry_len], carry_spec)
+    xs = tree_unflatten(tensors_list[flat_carry_len:], xs_spec)
+
+    def detach_tensor(inp: torch.Tensor) -> torch.Tensor:
+      x = inp.clone().detach()
+      x.requires_grad = inp.requires_grad
+      return x
+
+    def step_fn(grad_carry, pytree: Tuple[torch.Tensor, torch.Tensor,
```
JackCaoG, line 390: call it `backward_single_step` might be more accurate

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+                                          torch.Tensor]):
+      grad_y, carry, x = pytree
```
JackCaoG, line 392: is this a typo?

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+
+  @staticmethod
+  def forward(ctx, fn, init, xs):
+    # Forward pass, save inputs for backward
+    ctx._fn = fn
+    with torch._C._AutoDispatchBelowAutograd():
+      carry, carry_history, ys = _scan_impl(fn, init, xs)
+    flat_carry_history, carry_spec = tree_flatten(carry_history)
+    flat_xs, xs_spec = tree_flatten(xs)
+    ctx.save_for_backward(*flat_carry_history, *flat_xs)
```
JackCaoG, line 369: are `flat_carry_history ` `flat_xs ` always everything we need to save for the backward?

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+      grad_y, carry, x = pytree
+      # Compute the backward of a single scan iteration
+      detached_inputs = tree_map(detach_tensor, (carry, x))
+      with torch.enable_grad():
+        # `outputs` is a tuple of two PyTrees, the carry and y.
+        # In order to use `autograd.backward`, we need to flatten all
+        # PyTrees. The flattened `grad_carry` should match up with `carry`
+        # and similar for `y`.
+        outputs = fn(*detached_inputs)
+        output_carry, output_y = outputs
```
JackCaoG, line 401: wait... you are retracing the fwd here?

tengyifei: Correct. I'm not sure if the CSE pass can combine this with the same fn in the fwd pass. If it can't, then the bwd of scan will be slower.

JackCaoG: ehh, what you do here is pretty much force the gradident accumulation(through most likely cancel by the CSE since there is no optimization barrier), this sounds like a bad idea

miladm:
@tengyifei is this PR a 2.5 candidate?

tengyifei:
@miladm yes, I'd like to backport this to 2.5 after addressing the comments etc.

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+  # (lists or tuples of) tensors. So we will craft a `flat_fn` that takes in
+  # flattened PyTrees, internally recreates the desired tree structure, then calls `fn`.
+  def flat_fn(*seq: torch.Tensor) -> Tuple[PyTree, PyTree]:
+    carry = seq[:flat_carry_len]
+    x = seq[flat_carry_len:]
+    carry_pytree = tree_unflatten(carry, carry_spec)
+    x_pytree = tree_unflatten(x, xs_spec)
+    return fn(carry_pytree, x_pytree)
+
+  # Abstractly trace and lower `fn`.
```
tengyifei, line 165: Note to self: To further ensure the layers perform identically structured computations, we can trace each layer and compare their HLO. Tracing should be faster than end-to-end compilation. In practice this means scan would trace i times given input sequence of length i.

We can still avoid re-tracing the backward operations multiple times. We assume that identical forward graphs will produce identical backward graphs.

torch_xla/experimental/apply_layers.py
```diff
@@ -0,0 +1,72 @@
+from typing import Iterable
+
+import torch
+import torch.nn as nn
+from torch.utils._pytree import tree_map
+
+from torch_xla.experimental.scan import scan
+
+
+def apply_layers(layers: Iterable[torch.nn.Module], input_data):
```
tengyifei, line 10: Note to self: The apply_layers function will check that the layers have identical structure in terms of weights and biases:
- They all have the same set of dictionary keys (parameter names)
- The parameters have the same shapes.

test/test_apply_layers.py
tengyifei: Done

test/test_apply_layers.py
tengyifei: It looks like someone added `compareResults` as a helper which can also print the differences and dump the graph, so I thought it's pretty useful.

test/test_apply_layers.py
tengyifei: Done

test/test_apply_layers.py
tengyifei: Done

test/test_apply_layers.py
tengyifei: explained above

test/test_apply_layers.py
tengyifei: Done

test/test_apply_layers.py
tengyifei: Correct. The example decoder also has a `norm` at the end:

```

  def forward(
      self,
      input_ids: torch.LongTensor = None,
  ) -> torch.Tensor:
    inputs_embeds = self.embed_tokens(input_ids)

    # embed positions
    hidden_states = inputs_embeds

    # decoder layers
    for idx, decoder_layer in enumerate(self.layers):
      layer_outputs = decoder_layer(hidden_states,)
      hidden_states = layer_outputs

    hidden_states = self.norm(hidden_states)
    # [B, S, H] -> [B, S, V]
    return self.output(hidden_states)
```

test/test_apply_layers.py
tengyifei: done

test/test_operations.py
tengyifei: done

test/test_scan.py
tengyifei: PyLance can't find this class for some reason.

torch_xla/csrc/init_python_bindings.cpp
tengyifei: When `GetParameterId` receives a `BackendData` that is not a parameter in this lowering context, it will return `std::nullopt`. However, this loop is only iterating over parameters (line 1071, `const std::vector<torch::lazy::BackendDataPtr>& device_data = lowering_ctx.GetParametersData();`), so we will expect all `BackendData` there to have an ID. Seems good to enforce this invariant.

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+  # `1:` to skip `num_iters`
+  carry, xs, fn_carry_history, ys = unpack(outputs[1:])
+
+  # Unflatten tensors back to PyTrees
+  return tree_unflatten(carry, carry_spec), tree_unflatten(
+      fn_carry_history, carry_spec), tree_unflatten(ys, fn_output_y_spec)
+
+
+@pytreeify
+class Scan(torch.autograd.Function):
```
tengyifei, line 359: Note to self: enforce checkpointing by adding the right optimization barriers, and document this constraint. This may be improved in the future. This is fine since checkpointing the decoder block in LLMs is a common use case.

torch_xla/experimental/scan.py
```diff
@@ -82,14 +88,335 @@ def scan(fn, init, xs):
...
+  #
+  # We need to concatenate all three into one big list prior to
+  # entering `body_fn` and `cond_fn`, and split them back to three
+  # objects which is easier to work with after that. This pair of
+  # functions is for that purpose.
+  T = TypeVar('T')
+
+  def pack(carry: Sequence[T], xs: Sequence[T], fn_carry_history: Sequence[T],
+           ys: Sequence[T]) -> Tuple[T, ...]:
+    return tuple(carry) + tuple(xs) + tuple(fn_carry_history) + tuple(ys)
```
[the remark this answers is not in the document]
tengyifei: TIL, nice tip

torch_xla/experimental/scan.py
tengyifei: I don't think so -- `pytree` is a tuple of the output grad at current step (`grad_y`), carry at the current step (`carry`), and input at current step (`x`)

torch_xla/experimental/apply_layers.py
tengyifei: If there is a dropout layer that references other parameters (for example, the dropout probability), then those tensors will be captured as an additional input to the HLO graph. As implemented now, `apply_layers` and `scan` will trace the first layer, and use the same captured tensor for subsequent layers. This will be a problem if the user passes different dropout probabilities for say a sequence of dropout layers -- we'll instead incorrectly just keep using the first dropout's probability. I'll have to dig deeper and find a solution for this.

torch_xla/experimental/scan.py
tengyifei: I tested this and we give the wrong answer: https://github.com/tengyifei/playground/blob/master/scan_with_in_place_updates.ipynb

In the notebook, I wrote an approach to detect and prevent in place updates like that. TLDR is we'll have to trace every forward of each layer and verify that they're the same.

torch_xla/experimental/apply_layers.py
tengyifei: Could you clarify this question -- I thought `to_empty` is going to destroy the value inside `example_layer`, so I `deepcopy` it before to backup.

torch_xla/experimental/apply_layers.py
tengyifei: ah, good catch. not needed

torch_xla/experimental/scan.py
tengyifei: Yes. If the user `fn` references other tensors, they'll be captured as additional inputs to the HLO computation. I'll update the documentation to mention that we'll explicitly checkpoint `fn` in this iteration, and add the right barriers.

torch_xla/experimental/scan.py
tengyifei: Ack

torch_xla/experimental/scan.py
tengyifei: I believe the barriers won't be needed. This is the recommendation from JAX, which says you don't want to wrap inputs into a barrier if the checkpointed function is to be used in a `scan`.

[1]: https://jax.readthedocs.io/en/latest/_autosummary/jax.checkpoint.html

torch_xla/experimental/scan.py
tengyifei: I believe the barriers won't be needed. This is the recommendation from JAX, which says you don't want to wrap inputs into a barrier if the checkpointed function is to be used in a `scan`.

[1]: https://jax.readthedocs.io/en/latest/_autosummary/jax.checkpoint.html

torch_xla/experimental/scan.py
tengyifei: Chatted offline with Jack -- we'll just note this limitation in comments.

torch_xla/experimental/scan.py
tengyifei: I believe the barriers won't be needed. This is the recommendation from JAX, which says you don't want to wrap inputs into a barrier if the checkpointed function is to be used in a `scan`.

[1]: https://jax.readthedocs.io/en/latest/_autosummary/jax.checkpoint.html

torch_xla/experimental/scan.py
tengyifei: See `prevent_cse` documentation in https://jax.readthedocs.io/en/latest/_autosummary/jax.checkpoint.html

tengyifei: This is now avoided since we trace `fn` symbolically into separate fwd and bwd graphs, and weave the bwd into the backward of scan.

torch_xla/experimental/scan.py
tengyifei: In the latest commit, this will now fail with an assertion error instead of silently giving wrong results.

torch_xla/experimental/scan.py
tengyifei: Resolving this too, since we're now saving all the intermediate activations produced by the forward of `fn`, utilizing the graph capturing of AOTAutograd.

test/test_apply_layers.py
```diff
@@ -0,0 +1,273 @@
+import sys
```
JackCaoG, line 1: move all scan test to a folder maybe

test/test_apply_layers.py
```diff
@@ -0,0 +1,273 @@
...
+
+  def assert_different_tensor(self, a: torch.Tensor, b: torch.Tensor):
+    assert a is not b, f"Expected {a} and {b} to be different tensors"
+    assert a.data is not b.data, f"Expected {a} and {b} to have different storage"
+
+  def assert_while_found_in_hlo(self, tensor: torch.Tensor):
+    hlo_text = torch_xla._XLAC._get_xla_tensors_hlo([tensor])
+    assert "while(" in hlo_text
+    assert "condition=" in hlo_text
+    assert "body=" in hlo_text
```
JackCaoG, line 36: nit, `self.assertIn`

test/test_apply_layers.py
```diff
@@ -0,0 +1,273 @@
...
+        input_data = torch.randn(64).to(self.device)
+        torch_xla.sync(wait=True)
+
+    scan_layers = deepcopy(layers)
+    loop_layers = deepcopy(layers)
+    torch_xla.sync()
+
+    output = apply_layers(scan_layers, input_data.clone())
+    output.sum().backward()
+
```
JackCaoG, line 61: add another sync here. In here you are fusing the scan and non-scan into the same graph. It is mostly OK but not as clean.

test/test_apply_layers.py
```diff
@@ -0,0 +1,273 @@
...
+    scan_layers = deepcopy(layers)
+    loop_layers = deepcopy(layers)
+    torch_xla.sync()
+
+    # Also make input data some non-trivial graph instead of just device data.
+    input_data = (torch.randn(64).to(self.device) * 100,
+                  torch.randn(64).to(self.device) * 200,
+                  torch.randn(64).to(self.device) * 300)
+    a = torch.randn(64).to(self.device)
```
JackCaoG, line 121: hmm, we need to check if this `input is an IR` will work in the SPMD case, it is a bit complicated we can chat more offline..

test/test_apply_layers.py
```diff
@@ -0,0 +1,273 @@
...
+      def forward_scan(
+          self,
+          input_ids: torch.Tensor,
+      ) -> torch.Tensor:
+        inputs_embeds = self.decoder.embed_tokens(input_ids)
+        # embed positions
+        assert isinstance(inputs_embeds, torch.Tensor)
+        # decoder layers
+        hidden_states = apply_layers(self.decoder.layers, inputs_embeds)
+        hidden_states = self.decoder.norm(hidden_states)
```
JackCaoG, line 187: note to myself, I remember we need to add `optimizer_barrier` somewhere in the scan implementation.

test/test_scan.py
```diff
@@ -80,26 +126,297 @@ def test_scan_incompatible_length(self):
...
+    super().compareResults(final_carry, final_carry_remat)
+    super().compareResults(ys, ys_remat)
+    torch_xla.sync()
+
+    SINE_OP = re.compile(r" sine\(f32\b")
+
+    def count_number_of_sines(partition_fn):
+      own_carry = carry.clone().detach().requires_grad_()
+      own_xs = xs.clone().detach().requires_grad_()
+      final_carry, ys = scan(fn, own_carry, own_xs, partition_fn=partition_fn)
```
JackCaoG, line 281: this partition_fn is a bit confusing.

torch_xla/experimental/apply_layers.py
```diff
@@ -0,0 +1,126 @@
...
+import torch
+import torch.nn as nn
+from torch.utils._pytree import tree_map
+from functorch.compile import default_partition
+
+import torch_xla
+from torch_xla.experimental.scan import scan
+
+
+def apply_layers(layers: Iterable[torch.nn.Module],
```
JackCaoG, line 12: the name `apply_layers` does not really reflect that we are trying to pack the layers into a while loop, is this name coming from JAX?

torch_xla/experimental/apply_layers.py
```diff
@@ -0,0 +1,126 @@
...
+                 input_data,
+                 partition_fn=default_partition):
+  """Applies each layer in `layers` to `input_data` sequentially.
```
JackCaoG, line 15: please follow https://github.com/pytorch/xla/blob/master/torch_xla/distributed/spmd/xla_sharding.py#L147-L158 for the comment format, this way it can be picked up by the docstring later when we want to make it user facing.

JackCaoG: after render https://pytorch.org/xla/master/learn/api-guide.html#pytorch-xla-api

torch_xla/experimental/apply_layers.py
```diff
@@ -0,0 +1,126 @@
...
+
+  ## partition_fn
+
+  Since this function uses AOTAutograd to trace `fn`, you may override what computation
+  happen in the forward and backward passes by specifying different partition functions.
+  `default_partition` implies no activation checkpointing. You may specify
+  `functorch.compile.min_cut_rematerialization_partition` to use min-cut based activation
+  checkpointing. You may also write your own partitioner to insert any custom logic such
+  as host offloading of activations.
+  """
```
JackCaoG, line 41: I understand `partition_fn` is a more accurate name, but from a UX perspective you might want to just make it a book to enable gradient checkpointing or not.

torch_xla/experimental/apply_layers.py
```diff
@@ -0,0 +1,126 @@
...
+  # Define the function to apply at each step
+  def one_layer(carry, params_buffers):
+    # Apply the current layer's weights and biases to the example layer,
+    # then run the resulting layer.
+    output = torch.func.functional_call(  # type: ignore
+        example_layer, params_buffers, carry, strict=True)
+    # TODO(yifeit): it should be possible to return `None` as opposed to
+    # `torch.zeros(1)`, for additional clarity. There is no extra
+    # computation since we discard `ys` right after.
+    return output, torch.zeros(1, device=torch_xla.device(), requires_grad=True)
```
JackCaoG, line 74: I still don't understand what's this `torch.zeros` for

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+
+def value_and_grad_partitioned(
+    fn: Callable[[Carry, X], tuple[Carry, Y]],
+    init: Carry,
+    xs: X,
+    partition_fn=default_partition) -> tuple[Callable, Callable]:
+  """
+  Given a user `fn` to be scanned over the major axis of the input `xs` PyTree
+  and an initial carry object `init`, symbolically traces `fn` and returns two
+  functions, `forward` and `backward`, where `forward` packs intermediate
```
JackCaoG, line 149: `pack` do you mean `torch.stack` or something else?

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+  # Make some fake tensors to trace the user function and obtain the
+  # forward and backward graphs. Note that the init/carry fake tensor
+  # always requires grad. That's because even if the user passed in some
+  # `init` that does not require grad, we still want gradients to flow
+  # through the `carry` from one iteration of the user function to the
+  # next. In summary, the `carry` argument used to trace a user function
+  # to get a correct backward pass always requires grad.
+  def make_fake_tensor(v: torch.Tensor, requires_grad=True) -> torch.Tensor:
+    return torch.empty_like(
+        v, dtype=v.dtype, device=v.device, requires_grad=requires_grad)
```
JackCaoG, line 177: to enable the SPMD, you need to copy the sharding annotation from `v` to the result tensor. It will be more complicated if `v` is another IR

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+    return grad_carry, grad_x
+
+  return forward, backward
+
+
+def _make_get_graph_compiler():
+  """
+  Creates a compiler that records the graph, and a getter
+  function to retrieve them.
+  """
```
JackCaoG, line 226: lol this seems so hacky, but OK

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+    bw_compiler, get_bwd = _make_get_graph_compiler()
+    fn_compiled = aot_function(
+        fn,
+        fw_compiler=fw_compiler,
+        bw_compiler=bw_compiler,
+        partition_fn=partition_fn)
+    _, bwd_out_spec = tree_flatten((fake_carry_pytree, fake_x_pytree))
+    out = fn_compiled(fake_carry_pytree, fake_x_pytree)
+    out, fwd_out_spec = tree_flatten(out)
+    torch.autograd.backward(out, tree_map(lambda v: torch.ones_like(v), out))
```
JackCaoG, line 194: do we need this to trigger the actual bwd so the graph can be captured?

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+class Scan(torch.autograd.Function):
+
+  @staticmethod
+  def forward(ctx, forward, backward, init, xs):
+    # Forward pass, save activations for backward
+    ctx._backward = backward
+    with torch.no_grad():
+      carry, ys = _scan_impl_pytree(forward, init, xs)
+      ys, activations = ys
+    ctx.save_for_backward(*activations)
```
JackCaoG, line 252: is this stacked `activations`?

JackCaoG: I am trying to figure out where the gradient checkpointing happens when you use a different partion function.

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+    gradient_skipper = _GradientSkipper(ctx, starting_index=4)
+    grad_init = gradient_skipper.process(grad_init)
+    grad_xs = gradient_skipper.process(grad_xs)
+    return None, None, grad_init, grad_xs
+
+
+class _GradientSkipper:
+  """
+  Helper object to replace zero gradients with `None` gradients, if
+  the `needs_input_grad` value for the corresponding input tensor is False.
```
JackCaoG, line 275: why is this needed?

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+  ensure that each iteration of `fn` within the scan yields different random
+  numbers, despite running the same HLO operations. JAX requires the user to
+  explicitly fork the RNG state and pass it to `fn`. In PyTorch, the RNG state
+  is an implicit global variable. Therefore, we take a slightly different
+  approach:
+
+  - Identify usage of RNG state via `_get_tensors_xla_device_data_node`.
+  - Create N different copies of the RNG state contained in a tensor.
+  - While building the `While` op body, index into the RNG state tensor at the
+    current iteration and provide that seed value to `fn`.
```
JackCaoG, line 426: Ah this is because if you use aot_autograd to capture graph the random seed is not part of the graph...

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+  # a smaller computation (`fn_computation`). This is a mapping from
+  # `fn_computation` param ID to While computation param ID.
+  fn_param_id_to_while_param_id: Dict[int, int] = {}
+
+  # Add carry and x.
+  for real, fake in ((init, fake_carry), (xs, fake_x)):
+    for val, fake_val in zip(real, fake):
+      idx = builder.add_param(val)
+      param_id = fn_ctx.tensor_parameter_id(fake_val)
+      if param_id != -1:
```
JackCaoG, line 501: when will it be -1?

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+    seed_idx = ids.index(seed_info_id)
+    seed_parameter_id = fn_ctx.tensor_parameter_id(i_values[seed_idx])
+    assert seed_parameter_id != -1, "`fn` uses random seed, but random seed is not \
+      a parameter to the traced HLO graph"
+
+    # Replace the single seed value with a tensor of seeds, one per iteration.
+    seed_tensor = hoisted_vars[seed_parameter_id]
+    assert seed_tensor.dtype == torch.int64
+    hoisted_vars[seed_parameter_id] = torch.randint(
+        0, 2**62, (num_iters,), dtype=torch.int64, device=torch_xla.device())
```
JackCaoG, line 530: in the future we should have a way to let user control seeds they want to use. It seems like right now if they do 
```
torch.manual_seed(xxx)
```
it will kind of work too as that will control the output of the `torch.randint` but this is buried too deep for most users...

JackCaoG (commented):
mostly lgtm

torch_xla/experimental/scan.py
rpsilva-aws: That is for operations that are not on the device data (OpKindWrapper). I think there is a non-obvious issue/confusion that the mapping gives us all "parameters" for that computation, including inlined constants, but it does not have a parameter ID associated with it for other APIs. Do we have this dependency somewhere to expose these? Those are technically not actual parameters in the computation.

On a similar topic, I wonder that in case of SPMD:
```
    std::vector<xla::Literal> literals =
        runtime::GetComputationClient()->TransferFromDevice(
            UnwrapXlaData(device_data));
```
will be significantly more expensive than having the mapping only traverse ParametersData() and return tensors with `AtenFromXlaTensor` that align with `tensor_parameter_id`, rather than reconstructing a new tensor for all leaf operations. When I tried using sharded tensors with this mapping, it caused a significant synchronous bottleneck - but this can be solved/modified separately.

torch_xla/experimental/scan.py
rpsilva-aws: @JackCaoG Does copying `_get_xla_op_sharding` suffice there? Can you illustrate on "It will be more complicated if v is another IR", since we have a similar application for the while loop.

JackCaoG:
```
device = torch_xla.device()
t1 = torch.tensor(100, device=device)
xs.mark_sharding(t1, mesh, spec)
t2 = t1 * 2

scan(layers, t2)
```
in this case t2 will not have the sharding spec because it has not been propagate by XLA.

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+        v, dtype=v.dtype, device=v.device, requires_grad=requires_grad)
+
+  fake_carry_pytree = tree_map(make_fake_tensor, init)
+  fake_x_pytree = tree_map(
+      lambda v: make_fake_tensor(v[0], requires_grad=v.requires_grad), xs)
+
+  with torch.enable_grad():
+    fw_compiler, get_fwd = _make_get_graph_compiler()
+    bw_compiler, get_bwd = _make_get_graph_compiler()
+    fn_compiled = aot_function(
```
rpsilva-aws, line 186: Will this fail if there are tensors within `fn` that were not provided as parameters, since they are not 'fake' tensors? Say, for model parameters if we have a module inside `fn` that we wish to also trace fwd/bwd on. Is this targeted as a follow-up?

torch_xla/experimental/scan.py
tengyifei: Yes. It appears that the backward capturing is lazy. If I don't do autograd.backward, AOTAutograd doesn't capture the backward.

Fortunately all this does is creating a bunch of IRs that are later thrown away so no expensive computation actually happens, I think...

torch_xla/experimental/apply_layers.py
tengyifei: This is now handled correctly now that I'm splitting the RNG for each layer.

test/test_apply_layers.py
tengyifei: *nod* i think this is a good point

test/test_apply_layers.py
tengyifei: I believe that's not necessary because scan already strongly constrains the execution ordering. In JAX, I found these comments saying that when checkpoint is used in `scan`, `prevent_cse` is not needed. I can chat offline about this too.

test/test_scan.py
tengyifei: Added a comment.

torch_xla/experimental/apply_layers.py
tengyifei: Nope. I invented it.

tengyifei: Not sure what's a better name. In flax (the neural network library on top of jax), it's also called `scan`, but it's in a different names space: `flax.nn.scan` vs `jax.lax.scan`. The flax scan handles modules and the jax scan handles tensors.

tengyifei: Ok, I asked an LLM for ideas and I think we can go with `scan_layers`: it's a version of `scan` that handles nn layers

torch_xla/experimental/apply_layers.py
tengyifei: Here's my plan:

- Need to keep `partition_fn` around so I could also inject offloading logic as I partition the graph.
- Later, integrate with dynamo to get rid of this extra function altogether. Dynamo can decompose `torch.utils.checkpoint` into proper fw/bw graphs.
- When scan moves out of experimental, this arg will go away.

torch_xla/experimental/scan.py
```diff
@@ -82,14 +132,514 @@ def scan(fn, init, xs):
...
+  return carry, ys
+
+
+def value_and_grad_partitioned(
+    fn: Callable[[Carry, X], tuple[Carry, Y]],
+    init: Carry,
+    xs: X,
+    partition_fn=default_partition) -> tuple[Callable, Callable]:
+  """
+  Given a user `fn` to be scanned over the major axis of the input `xs` PyTree
```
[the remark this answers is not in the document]
tengyifei: yep. I'll rename this to "leading dimension", since pytorch prefers "dim" while jax prefers "axis".

torch_xla/experimental/scan.py
tengyifei: It just puts them in a large tuple.

Whereas the original `fn` outputs `(carry, y)`,
the transformed `forward` outputs `(carry, (y, activations))`

When we scan the transformed `forward` over the input, we'll get back not just the output, but also the stacked activations that could be stored in `ctx` and later used in the backward pass.

torch_xla/experimental/scan.py
tengyifei: ack. this will go away if we model `scan` as a lower of the dynamo `scan` op

torch_xla/experimental/scan.py
tengyifei: *nod* added a TODO

torch_xla/experimental/scan.py
tengyifei: It's not! :)

I think it was needed at one point where I didn't handle `None` value in PyTrees nicely. Now it's not needed.

tengyifei: Removed

torch_xla/experimental/scan.py
tengyifei: Yes, it will. That's why I added `scan_layers` (previously named `apply_layers`) in this PR to extract module parameters and functionalize the module.

torch_xla/experimental/apply_layers.py
tengyifei: done

test/tpu/run_tests.sh
```diff
@@ -26,7 +26,8 @@ python3 test/pjrt/test_dtypes.py
 python3 test/pjrt/test_dynamic_plugin_tpu.py
 python3 test/test_while_loop.py
 python3 test/test_scan.py
-python3 test/test_pallas.py -v
+python3 test/test_apply_layers.py
```
JackCaoG, line 29: I onlu see `test_scan_layers`

tengyifei: oops. fixed now

tengyifei:
This is ready for another look. For ease of review I added two additional commits for stuff we talked about offline:

- [1] Test that the XLA compiler can propagate SPMD sharding annotations just fine through the `While` op and the `Body` computations
- [2] Change the name of the `PyLoweringContext` to `FnComputation` when it appears in the HLO

[1]: https://github.com/pytorch/xla/pull/7901/commits/7e4cf9abf944c24a44b4942d49250736423d3fef
[2]: https://github.com/pytorch/xla/pull/7901/commits/98222b25c4db4ff3a46dc7e2e4315e72bdaacdaa

JackCaoG:
> This is ready for another look. For ease of review I added two additional commits for stuff we talked about offline:
> 
> * [1](https://github.com/pytorch/xla/pull/7901/commits/7e4cf9abf944c24a44b4942d49250736423d3fef) Test that the XLA compiler can propagate SPMD sharding annotations just fine through the `While` op and the `Body` computations
> * [2](https://github.com/pytorch/xla/pull/7901/commits/98222b25c4db4ff3a46dc7e2e4315e72bdaacdaa) Change the name of the `PyLoweringContext` to `FnComputation` when it appears in the HLO

please move the SPMD stuff to a different pr, I will try to finish reviewing this pr today( github needs a better way to stack prs....).

torch_xla/experimental/scan.py
tengyifei: Honestly I don't really understand what OpKindWrapper is so I'm not sure if there's something wrong here. If there is, appreciate advice how I should fix it. I can tell you my high level observation: when lowering LazyTensor IR to HLO, inlined constants and tensors created within the function are hoisted to HLO parameters:

```
    def fn(carry, x):
      foo = torch.zeros(8)
      return carry, x + foo
```

lowers into an HLO like this:

```
    HloModule FnComputation, entry_computation_layout={
      (f32[8], f32[8], f32[8]) -> (f32[8], f32[8])
    }
```

Now when I build the While computation, I must provide some tensor for the third parameter, which is in fact `torch.zeros(8)`. This `tensor_parameter_id` seems to provide that functionality. It also seems to materialize the tensor data on the CPU (calls `std::vector<xla::Literal> literals = runtime::GetComputationClient()->TransferFromDevice(UnwrapXlaData(device_data));`).

tengyifei:
> please move the SPMD stuff to a different pr
sure, done

torch_xla/experimental/scan.py
```diff
@@ -33,33 +73,65 @@ def scan(fn, init, xs):
...
-    xs: the input PyTree to scan over. If `xs` is a tensor, then `fn` will get slices along
-        the leading dimension (`xs[i]`). If `xs` is some other PyTree (e.g. tuple of
-        tensor), `fn` will get PyTrees of slices. In that case the leading dimension size
-        of the leaves in the PyTree must be the same.
+    xs: The input PyTree to scan over. If `xs` is a tensor, then `fn` will get slices along
+      the leading dimension (`xs[i]`). If `xs` is some other PyTree (e.g. tuple of
+      tensor), `fn` will get PyTrees of slices. In that case the leading dimension size
+      of the leaves in the PyTree must be the same.
+        
+    partition_fn: Since `scan` uses AOTAutograd to trace `fn`, you may override what
```
JackCaoG, line 101: if this is optional, you want to mark it as optional in the doc string

JackCaoG (approved):

torch_xla/experimental/scan.py
tengyifei: done

tengyifei:
Thanks for the review. I'll merge this first. I'm looking at caching the fn computation properly but that'll take some time. I might be able to send that PR over (and the SPMD one) for next Monday.

ydwu4:
Hi! PyTorch also has a scan operator https://github.com/pytorch/pytorch/blob/main/torch/_higher_order_ops/scan.py. Wondering if we want to consolidate the efforts and what's the plan going forward?

tengyifei:
@ydwu4 does the PyTorch scan operator support autograd? I remember there's an issue tracking autograd support. It would be great to use upstream op but without autograd support we can't use that in training.

ydwu4:
Sounds good. On PyTorch side, we've been prioritizing to get inference working e2e and will get to autograd next half. It will be great if we could reduce fragmentation with a single front end op.

tengyifei:
That's a great idea.

If you looked at the current scan impl in PyTorch/XLA, it uses `AOTAutograd` to derive a backward graph to implement the backward pass of `scan`. That API has a lot of limitations and IIUC dynamo is the well supported frontend.

PR #7912, opened 2024-08-26
Allow MpDeviceLoader to shard dictionaries of tensors

bhavya01:


test/test_mp_input_sharding.py
```diff
@@ -0,0 +1,98 @@
...
+  @unittest.skipUnless(xr.global_runtime_device_count() > 1,
+                       "Multiple devices required for tupled partition spec")
+  def test_multiple_inputs(self):
+    device = xm.xla_device()
+    batch = {'x': torch.randn((16, 128)), 'y': torch.randn((16, 128, 128))}
+    train_loader = self.fake_dataloader(batch)
+    num_devices = xr.global_runtime_device_count()
+    mesh_shape = (num_devices, 1)
+    device_ids = np.arange(num_devices)
+    mesh = Mesh(device_ids, mesh_shape, ('x', 'y'))
```
JackCaoG, line 38: fyi you can use https://github.com/pytorch/xla/blob/master/torch_xla/distributed/spmd/xla_sharding.py#L163-L189 directly

JackCaoG:
please run both v2 and v4 TPUCI in this pr.

test/test_mp_input_sharding.py
```diff
@@ -0,0 +1,98 @@
+import sys
```
JackCaoG, line 1: move this under `test/spmd`

torch_xla/distributed/parallel_loader.py
```diff
@@ -155,27 +157,64 @@ def _loader_worker(self):
...
+      batch -> List(torch.Tensor), List(Dict(str: torch.Tensor)): Input batch
+        present in the cpu memory
+      device: TPU device where the batch should be moved
+    
+    Returns:
+      result -> List(torch.Tensor), Dict(str: torch.Tensor): Returns a dict if the
+        input batch is a dict. Otherwise, returns a list of torch.Tensor.
+    """
+    result = None
+    if isinstance(batches[0], dict) and isinstance(self._input_sharding, dict):
```
JackCaoG, line 187: wouldn't it be better to do :
```
if isinstance(batches[0], dict):
  assert self._input_sharding is None or isinstance(self._input_sharding, dict), "error message you want"
```
?

torch_xla/distributed/parallel_loader.py
```diff
@@ -155,27 +157,64 @@ def _loader_worker(self):
...
+            continue
+          # xla_tensor is a list of tensors.
+          xla_tensor = xm.send_cpu_data_to_device(tensor, device,
+                                                  self._input_sharding[key])
+          xla_batch[key] = xla_tensor[0]
+        if len(missing_keys) != 0:
+          # Returning exception as raisng in the dataloading thread doesn't surface the problem in the main thread.
+          return [
+              KeyError(f"Keys: {missing_keys} are missing from input_sharding.")
+          ]
```
JackCaoG, line 205: why not raise directly?

JackCaoG (commented):
can you also update https://github.com/pytorch/xla/blob/master/docs/spmd_advanced.md#sharding-aware-host-to-device-data-loading?

torch_xla/distributed/parallel_loader.py
bhavya01: This part of the code executes in a separate thread. If an exception happens in the thread, the main thread is just blocked on the _worker thread to load data into the input_batches queue. To explicitly raise the error to the user, I think it might be better if we add the exception to the input batch queue and check for it while passing the data to the main thread in PerDeviceLoader.next function

JackCaoG (approved):
I think we should also mention this in the release note

PR #7913, opened 2024-08-26
add dynamo config skip_input_data_check

JackCaoG:
This is introduce the config `skip_input_data_check ` which will skip an parameter check in dynamo bridge. This code is in the hot path of the dynamo execution and will invoke a pybind + constructuing a boolean array with the same zie of the input parameters.

Enabling this will save us ~0.1ms launch overhead.


test/run_tests.sh
```diff
@@ -182,6 +182,7 @@ function run_xla_op_tests1 {
   run_test "$CDIR/dynamo/test_bridge.py"
   run_test "$CDIR/dynamo/test_num_output.py"
   run_test "$CDIR/dynamo/test_graph_input_matcher.py"
+    run_test "$CDIR/dynamo/test_dynamo_config.py"
```
lsy323, line 185: nit: indent

lsy323 (approved):

test/run_tests.sh
JackCaoG: opps.. let me fix..

Issue #7904, opened 2024-08-23
torch_xla.distributed.parallel_loader doesn't shard data


davidaknowles:
## ❓ Questions and Help

Maybe this is a misunderstanding on my part, but I assumed part of `MpDeviceLoader`s job was to split/shard data across devices. However the test below shows it doesn't do this: all 4 devices on my v4 receive all 12 datapoints. What am I missing here? Thanks. 

```
import torch
import torch_xla.core.xla_model as xm
from torch_xla import runtime as xr
import torch_xla.distributed.parallel_loader as pl
import torch_xla.distributed.xla_multiprocessing as xmp

def test_parallel_loader(rank):

    data = torch.arange(12).reshape(-1, 1)
    dataset = torch.utils.data.TensorDataset(data)
    dataloader = torch.utils.data.DataLoader(dataset, batch_size=2)
    
    parallel_loader = pl.MpDeviceLoader(dataloader, xm.xla_device())
    results = sum([batch[0].tolist() for batch in parallel_loader], [])

    print(f"Device {rank} received data: {results}")
    
    expected_data_size = len(data) // xr.world_size()
    print(f"Device {rank} received {len(results)} datapoints, expected {expected_data_size}")

if __name__ == "__main__":
    xmp.spawn(test_parallel_loader, args=()) 
```

JackCaoG:
Let me take a look

bhavya01:
In this case, it is working as expected. The MpDeviceLoader in each spawned process will asynchronously get batches from the torch dataloader and put them on the TPU. If you use SPMD and just spawn one process and specify `input_sharding` such that it shard on the batch size, then you should see that batch is spread out over multiple devices.

JackCaoG:
To add on what @bhavya01 ParallelLoader itself does not handle distributing the correct data to each worker, it is the dataloder it wrapped needs to do that.

In our MP example it is 
https://github.com/pytorch/xla/blob/3d860bf4f5ad0f20665245ba470ac7f5886f50f7/test/test_train_mp_imagenet.py#L222-L226
doing this work. In the GSPMD case it is the sharding we passed to the `ParallelLoader` does that distribution. In my example I only used fake data loader so this is not an issue, I can fix that to make it more clear.

davidaknowles:
[...]

PR #7914, opened 2024-08-26
Update comments in data parallel example to use sampler

JackCaoG:
fix https://github.com/pytorch/xla/issues/7904

### DIFF
--- a/examples/data_parallel/train_resnet_ddp.py
+++ b/examples/data_parallel/train_resnet_ddp.py
@@ -20,6 +20,28 @@ class TrainResNetDDP(TrainResNetBase):
     self.model = DDP(
         self.model, gradient_as_bucket_view=True, broadcast_buffers=False)
     self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
+    # below code is commented out because in this example we used a fake data
+    # loader that does not take sampler. However this logic is needed if you
+    # want each process to handle different parts of the data.
+    '''
+    train_sampler = None
+    if xr.world_size() > 1:
+      train_sampler = torch.utils.data.distributed.DistributedSampler(
+          train_dataset,
+          num_replicas=xr.world_size(),
+          rank=xr.global_ordinal(),
+          shuffle=True)
+    train_loader = torch.utils.data.DataLoader(
+        train_dataset,
+        batch_size=FLAGS.batch_size,
+        sampler=train_sampler,
+        drop_last=FLAGS.drop_last,
+        shuffle=False if train_sampler else True,
+        num_workers=FLAGS.num_workers,
+        persistent_workers=FLAGS.persistent_workers,
+        prefetch_factor=FLAGS.prefetch_factor)
+    self.train_device_loader = pl.MpDeviceLoader(train_loader, self.device)
+    '''
 
 
 def _mp_fn(index):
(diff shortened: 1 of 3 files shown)


examples/data_parallel/train_resnet_ddp.py
```diff
@@ -20,6 +20,28 @@ def __init__(self):
...
     self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
+    # below code is commented out because in this example we used a fake data
+    # loader that does not take sampler. However this logic is needed if you
+    # want each process to handle different parts of the data.
+    '''
+    train_sampler = None
+    if xr.world_size() > 1:
+      train_sampler = torch.utils.data.distributed.DistributedSampler(
+          train_dataset,
+          num_replicas=xr.world_size(),
```
will-cromar, line 31: nit: dist.world_size

examples/data_parallel/train_resnet_ddp.py
```diff
@@ -20,6 +20,28 @@ def __init__(self):
...
+          rank=xr.global_ordinal(),
```
will-cromar, line 32: nit: dist.get_rank

examples/data_parallel/train_resnet_ddp.py
```diff
@@ -20,6 +20,28 @@ def __init__(self):
     self.model = DDP(
         self.model, gradient_as_bucket_view=True, broadcast_buffers=False)
     self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
+    # below code is commented out because in this example we used a fake data
+    # loader that does not take sampler. However this logic is needed if you
+    # want each process to handle different parts of the data.
+    '''
+    train_sampler = None
+    if xr.world_size() > 1:
```
will-cromar, line 28: also dist.world_size

examples/data_parallel/train_resnet_ddp.py
```diff
@@ -20,6 +20,28 @@ def __init__(self):
     self.model = DDP(
         self.model, gradient_as_bucket_view=True, broadcast_buffers=False)
     self.optimizer = optim.SGD(self.model.parameters(), weight_decay=1e-4)
+    # below code is commented out because in this example we used a fake data
+    # loader that does not take sampler. However this logic is needed if you
+    # want each process to handle different parts of the data.
+    '''
```
will-cromar, line 26: Is it clearer to just apply this sampler to the fake dataset anyway?

JackCaoG: fake dataset is a XLA util https://github.com/pytorch/xla/blob/master/examples/train_resnet_base.py#L25-L29, it does not take sampler.

will-cromar: Oh, I just kind of assumed that `SampleGenerator` was an idiomatic `Dataset`. Can you try just making it inherit from [`IterableDataset`](https://github.com/pytorch/xla/blob/00e406325f5c7c0897680dcb38ca6b168ce165ed/torch_xla/utils/utils.py#L56-L57) since it has `__iter__` and `__len__` already? You should then be able to wrap it in a standard sampler and data loader.

If that doesn't work, then one of us can follow up. Our examples should be as close to PyTorch as possible.

PR #7918, opened 2024-08-27
`DynamicShapeDetector` with trie implementation.

ysiraichi:
This PR finishes the implementation started in #7817. In this PR, we implement the `DynamicShapeDetector` using a trie. It is used for detecting different traces (sequence of PyTorch operations called), which we are assuming is due to dynamic shapes.

**How does it work?** 
- Wrap a function with `_XLAC._dynamic_shape_detector_start_session` call in the beginning, and `_XLAC._dynamic_shape_detector_end_session` call in the end of a function
- When the function is run, the detector will start keeping track of the created IR nodes
- If, at the N-th call, we trace a different sequence of operations from the previous calls, we increment the number of traces
- If the number of traces is greater-than `max_allowed_traces_per_function`, we raise an error

**Implementation Details:**
- Build the trie incrementally, with the help of `TrieBuilder` (state of tracing)
- At every traced operation, we update the `TrieBuilder` (similar to states of a DFA)
- If, at any point, we have to update the actual `TrieNode`, it means this is part of a new trace (i.e. sequence of operations)

cc @miladm @JackCaoG

### DIFF
--- a/test/test_dynamic_shapes_detector.py
+++ b/test/test_dynamic_shapes_detector.py
@@ -4,20 +4,16 @@ import test_utils
 import unittest
 
 
-def set_max_allowed_traces(traces):
-  torch_xla._XLAC._dynamic_shape_detector_set_max_allowed_traces(traces)
-
-
 class TestDynamicShapeDetector(test_utils.XlaTestCase):
 
-  def _run_and_compare(self, f, optf=None, args=None):
+  def _run_and_compare(self, f, args=None, allowed_traces=None):
     """Run f and its torch_xla.compile wrapped version, comparing the equality
     of their results.
 
     If no optf is provided, we create a new one by wrapping it with
     torch_xla.compile ourselves.
     """
-    optf = optf or torch_xla.compile(f, detect_dynamic_shape=True)
+    optf = torch_xla.compile(f, allowed_traces=allowed_traces)
     args = args or []
 
     out = f(*args)
@@ -28,21 +24,17 @@ class TestDynamicShapeDetector(test_utils.XlaTestCase):
   def test_single(self):
     # Test: trace a function once, when only one trace is allowed.
 
-    set_max_allowed_traces(1)
-
     def foo(x):
       return x + x
 
     inp = torch.rand(10, device=torch_xla.device())
-    self._run_and_compare(foo, args=(inp,))
+    self._run_and_compare(foo, args=(inp,), allowed_traces=1)
 
   def test_many_traces(self):
     # Test: multiple traces of a function.
     #
     # Steps 0~2 and 5: create new traces.
     # Steps 3 and 4: ensure we have already traced these paths.
-
-    set_max_allowed_traces(4)
 
     def foo(x, step):
       r0 = x + x + x
@@ -55,34 +47,31 @@ class TestDynamicShapeDetector(test_utils.XlaTestCase):
         return r * 4
       return r0
 
-    optfoo = torch_xla.compile(foo, detect_dynamic_shape=True)
     inp = torch.rand(10, device=torch_xla.device())
 
     for i in range(6):
-      self._run_and_compare(foo, args=(inp, i))
+      self._run_and_compare(foo, args=(inp, i), allowed_traces=4)
 
   def test_trace_limit_exceeded_different_input_shape(self):
     # Test: catch trace limit exceeded error when running the function with a
     # function with different shape.
 
-    set_max_allowed_traces(1)
+    allowed_traces = 1
 
     def foo(x):
       return x + x
 
-    optfoo = torch_xla.compile(foo, detect_dynamic_shape=True)
-
     inp1 = torch.rand(10, device=torch_xla.device())
-    self._run_and_compare(foo, optfoo, args=(inp1,))
+    self._run_and_compare(foo, args=(inp1,), allowed_traces=allowed_traces)
 
     msg = """\
-torch_xla/csrc/dynamic_shape_detector.cpp:47 : Maximum number of different traces allowed per function exceeded: 1
+.* Maximum number of different traces allowed per function exceeded: 1
 Got: [] aten::expand, xla_shape=f32[10]{0}, dynamic_dims: (), size=(10)
 Expected: [] aten::add, xla_shape=f32[10]{0}, dynamic_dims: ()"""
 
     with self.assertRaises(RuntimeError, msg=msg):
       inp2 = torch.rand(5, device=torch_xla.device())
-      self._run_and_compare(foo, optfoo, args=(inp2,))
+      self._run_and_compare(foo, args=(inp2,), allowed_traces=allowed_traces)
 
   def test_trace_limit_exceeded_common_sequence_mismatch(self):
     # Test: catch trace limit exceeded error when the common sequence (i.e. compressed
@@ -95,7 +84,7 @@ Expected: [] aten::add, xla_shape=f32[10]{0}, dynamic_dims: ()"""
     # (ii) mul operation.
     # However, it fails since we have reached the limit.
 
-    set_max_allowed_traces(1)
+    allowed_traces = 1
 
     def foo(x, step):
       if step == 0:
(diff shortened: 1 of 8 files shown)


ysiraichi:
@JackCaoG  When I was writing the tests for this PR, I thought that `torch_xla.compile` could return a class instance (in a future PR), so that we can track the number of recorded traces for each function. What do you think?

torch_xla/csrc/dynamic_shape_detector.cpp
```diff
@@ -0,0 +1,254 @@
...
+  XLA_CHECK(current_session_ != nullptr);
+
+  try {
+    builder_.AddValue({hash, str}, AllowNewTrace());
+  } catch (const std::exception& e) {
+    // AddValue might raise an exception if AllowNewTrace() is false. Catch it
+    // here, so that we can correctly return the builder to the root of the
+    // trie.
+    //
+    // TODO(ysiraichi): we should actually rollback this trace.
```
ysiraichi, line 238: As is, it should work as we expect. However, a better approach (for a future PR) would be to rollback the changes introduced since the last session start.

JackCaoG:
Thanks I will try to take a look today.

torch_xla/torch_xla.py
```diff
@@ -83,9 +84,12 @@ def step():
...
 
 
-def compile(f: Optional[Callable] = None,
-            full_graph: Optional[bool] = False,
-            name: Optional[str] = None):
+def compile(
+    f: Optional[Callable] = None,
+    full_graph: Optional[bool] = False,
+    name: Optional[str] = None,
+    detect_dynamic_shape=False,
```
JackCaoG, line 91: I think instead of a boolean value `detect_dynamic_shape`, we can have it to be something like `max_dynamic_shape_graph_allowed`, then you can map it to `_dynamic_shape_detector_set_max_allowed_traces ` directly.

torch_xla/torch_xla.py
```diff
@@ -125,6 +129,8 @@ def foo2(x):
     elif hasattr(f, '__str__'):
       name = f.__str__()
 
+  current_id = uuid.uuid4().__str__()
```
JackCaoG, line 132: I thought about it a bit and find this is not very ideal. This way if we do
```
def f():
  xxx   

torch.compile(f)
torch.compile(f)
```
we will get 2 uuid. I think we should try to hash the passed in function pointer so we can dedup.

JackCaoG: I think we should use function pointer if it is not None (currently when it is used as a decorator with @, the fn will be none I think).

test/test_dynamic_shapes_detector.py
```diff
@@ -0,0 +1,233 @@
...
+    def foo(x):
+      return x + x
+
+    optfoo = torch_xla.compile(foo, detect_dynamic_shape=True)
+
+    inp1 = torch.rand(10, device=torch_xla.device())
+    self._run_and_compare(foo, optfoo, args=(inp1,))
+
+    msg = """\
+torch_xla/csrc/dynamic_shape_detector.cpp:47 : Maximum number of different traces allowed per function exceeded: 1
```
JackCaoG, line 79: `cpp:47` is too specified, we can remove the line number and file name from the check to make it more general.

torch_xla/csrc/dynamic_shape_detector.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  if (common_sequence_.size() > matched) {
+    ostr << "Expected: " << common_sequence_[matched].str << std::endl;
+  } else {
+    ostr << "Expected either of:" << std::endl;
+    for (auto& pair : children_) {
+      ostr << "  - " << pair.second->common_sequence_.front().str << std::endl;
+    }
+  }
+
```
JackCaoG, line 40: I think we should also dump the current python trace, which would help people to figure out where to fix. Check https://github.com/pytorch/xla/blob/master/torch_xla/csrc/debug_util.cpp#L125-L131 for the example.

torch_xla/csrc/dynamic_shape_detector.cpp
JackCaoG: if we hit a error likely the program will just end so maybe it is fine.

JackCaoG (commented):
implementation lgtm, minor comments on the ux.

torch_xla/torch_xla.py
ysiraichi: What do you think about `allowed_traces` (because we can have multiple graphs when tracing a function once)?

torch_xla/csrc/dynamic_shape_detector.cpp
ysiraichi: Here's an example running `test_trace_limit_exceeded_common_sequence_mismatch`:

```python
  File "/usr/local/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/local/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/usr/local/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "xla/test/test_dynamic_shapes_detector.py", line 104, in test_trace_limit_exceeded_common_sequence_mismatch
    self._run_and_compare(foo, args=(inp, 2), allowed_traces=allowed_traces)
  File "xla/test/test_dynamic_shapes_detector.py", line 20, in _run_and_compare
    optout = optf(*args)
  File "/usr/local/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "xla/test/test_dynamic_shapes_detector.py", line 93, in foo
    return x * 5
RuntimeError: torch_xla/csrc/dynamic_shape_detector.cpp:41 : Maximum number of different traces allowed per function exceeded: 1
Got: [] aten::mul, xla_shape=f32[10]{0}, dynamic_dims: ()
Expected: [] aten::add, xla_shape=f32[10]{0}, dynamic_dims: ()
```

We already have the python trace, since this check is being done incrementally, every time a new IR node is created.

ysiraichi:
Summary of the changes:

- If a function is given, make its `current_id` dependent on its name and id
- Add API for removing C++ session entries: `DynamicShapeDetector::RemoveSessionIfExists`
- Keep track of the compiled functions that are still alive
    - Different local-scoped functions with the same name and id may exist
- Add `allowed_traces` optional parameter + documentation
- Remove file and line information from the expected error messages

ysiraichi:
@JackCaoG This PR is ready for another round of reviews. Could you take a look at it?

torch_xla/torch_xla.py
```diff
@@ -83,9 +85,17 @@ def step():
...
+# Keeps track of the alive functions. This allow us to remove session entries in the
+# C++ side for functions that are no longer alive.
+_compiled_id_to_functions_ref = weakref.WeakValueDictionary()
+
+
+def compile(
+    f: Optional[Callable] = None,
+    full_graph: Optional[bool] = False,
+    name: Optional[str] = None,
+    allowed_traces: Optional[int] = None,
```
JackCaoG, line 97: I think `traces` is too much of an implementation detail. Since we already have the `full_graph` above, how about `num_different_graph_allowed`?

ysiraichi: Wouldn't that be kind of confusing, leading the user to think of the number of HLO graphs?

ysiraichi: Ok. Couldn't think of something better, so went with your suggestion.

torch_xla/csrc/dynamic_shape_detector.cpp
JackCaoG: Ok this is nice, I guess we can go a step further where for every TRIE node we also store the python stack trace for the current node. This way when we raise the runtime error, we can also show that here is the python stack we expects, but now we hit here. It would be easier to debug this way.

we can implement this as a follow up through.

torch_xla/torch_xla.py
ysiraichi: Hmm. That's not the case if we have a fallback operation in the middle, is it?

torch_xla/torch_xla.py
```diff
@@ -134,28 +166,35 @@ def _compile():
...
+    if name is not None:
       torch_xla._XLAC._set_current_graph_name(name)
 
     # if full_graph sets to true execution can not happen before the sync below
     torch_xla._XLAC._set_allow_execution(not full_graph)
 
+    if num_different_graphs_allowed is not None:
+      torch_xla._XLAC._dynamic_shape_detector_set_max_num_different_graphs_allowed(
+          num_different_graphs_allowed)
+      torch_xla._XLAC._dynamic_shape_detector_start_session(current_id)
```
JackCaoG, line 183: I felt like it is better to regsiter the session with `num_different_graphs_allowed ` outside of the `_compile ` and in here we just need to start the session. We can do that in the follow up

JackCaoG (approved):

torch_xla/torch_xla.py
JackCaoG: oh I see, I generally expect user to also turn on `full_graph` to True.. I guess `trace` is more correct but I just find it expose too much underlying implementation details.

PR #7921, opened 2024-08-29
Support gmm and tgmm trace_pallas caching

JackCaoG:


JackCaoG:
still need to add a test for the cache miss case.

torch_xla/experimental/custom_kernel.py
```diff
@@ -116,11 +120,27 @@ def trace_pallas(kernel: Callable,
     else:
       jax_args.append(arg)
 
+  hash_key = ()
+  if use_cache:
+    global trace_pallas_arg_to_payload
+    # implcit assumption here that everything in kwargs is hashable and not a tensor,
+    # which is true for the gmm and tgmm.
+    hash_key = (kernel, static_argnums, tuple(static_argnames), tuple(jax_args),
```
alanwaketan, line 128: How does this work with different objects but with the same size, dtype and device?

alanwaketan: That's interesting. I guess if it works it works. Then why don't just use @cache?

JackCaoG: my understanding is that `@cache` cache the input, inputs of this functions are xla tensor, I felt like `cache` will try to access the value of those tensors. in here I only cache the JAX meta tensor.

Also let me reverify this with the real moe models.

alanwaketan: I see. That's fair.

alanwaketan (approved):

JackCaoG:
verified in the profile that `trace_pallas` is cached.

Issue #7919, opened 2024-08-28
Error while trying to save a sharded model checkpoint on multi-host TPU


dudulightricks:
## To Reproduce

Here is a short example to reproduce the error, running on vp-16 TPU pod:

```
import numpy as np
import torch_xla.core.xla_model as xm
import torch_xla.runtime as xr
import torch_xla.distributed.spmd as xs
from torch_xla.distributed.spmd import Mesh
import torch.distributed as dist
import torch_xla.distributed.xla_backend
import time

from torch import nn
import torch.distributed.checkpoint as dist_cp
import torch_xla.experimental.distributed_checkpoint as xc

SAVING_PATH="/tmp/example4"

xr.use_spmd()
dist.init_process_group('gloo', init_method='xla://')

num_devices = xr.global_runtime_device_count()
mesh_shape = (num_devices//2, 2)
device_ids = np.array(range(num_devices))
mesh = Mesh(device_ids, mesh_shape, ('x', 'y'))

model = nn.Linear(in_features=2, out_features=2, device=xm.xla_device())
model_sharded = xs.mark_sharding(model.weight, mesh, (None, 'y'))

print(f"model weight {model.weight} model bias {model.bias}")

state_dict = {
    "my_model": model.state_dict(),
}

dist_cp.save(
    state_dict=state_dict,
    storage_writer=dist_cp.FileSystemWriter(SAVING_PATH),
    planner=xc.SPMDSavePlanner(),
)

time.sleep(20)

state_dict = {
    "my_model": model.state_dict(),
}

dist_cp.load(
    state_dict=state_dict,
    storage_reader=dist_cp.FileSystemReader(SAVING_PATH),
    planner=xc.SPMDLoadPlanner(),
)
model.load_state_dict(state_dict["my_model"])

print(f"model weight after load: {model.weight} bias: {model.bias}")
```

1. Run this code on multi-host TPU (vp-16 in our case)

## Expected behavior

The model weights should be printed, and should be the same as before the save.

## Stack trace

The code failing with an error:

```
[rank0]:     return func(*args, **kwargs)
[rank0]:   File "/home/user/.local/lib/python3.10/site-packages/torch/distributed/checkpoint/state_dict_loader.py", line 168, in load
[rank0]:     _load_state_dict(
[rank0]:   File "/home/user/.local/lib/python3.10/site-packages/torch/distributed/checkpoint/state_dict_loader.py", line 220, in _load_state_dict
[...]

PR #7929, opened 2024-08-29
Update dedup logic to use dedup_save_plans

jonb377:
`dedup_tensors` comes with a flaw which will erase all WriteItems for shards contained within a single save plan. This surfaced in https://github.com/pytorch/xla/issues/7919.

Further, the upstream deduplication logic now supports load balancing through the `dedup_save_plans` API.

### DIFF
--- a/torch_xla/experimental/distributed_checkpoint/planners.py
+++ b/torch_xla/experimental/distributed_checkpoint/planners.py
@@ -36,9 +36,10 @@ from torch.distributed.checkpoint.utils import find_state_dict_object
 from torch.utils._pytree import tree_map
 from torch_xla.distributed.spmd import XLAShardedTensor, XLAShard
 from torch_xla.experimental.distributed_checkpoint._helpers import (
-    FLATTEN_MAPPING, flatten_state_dict, dedup_tensors, _is_sharded_tensor,
+    FLATTEN_MAPPING, flatten_state_dict, _is_sharded_tensor,
     set_element, narrow_tensor_by_index, _unwrap_xla_sharded_tensor, _CpuShards)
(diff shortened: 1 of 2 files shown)


jonb377:
@JackCaoG This takes a dep on the private `dedup_save_plans` API, should we work with the upstream to make it public or pull a copy of the implementation into our repo instead?

JackCaoG:
it is like one experimental api depends on another api.. it is OK but we should communicate with the upstream and see what's their plan with this api.

torch_xla/experimental/distributed_checkpoint/planners.py
```diff
@@ -36,9 +36,10 @@
 from torch.utils._pytree import tree_map
 from torch_xla.distributed.spmd import XLAShardedTensor, XLAShard
 from torch_xla.experimental.distributed_checkpoint._helpers import (
-    FLATTEN_MAPPING, flatten_state_dict, dedup_tensors, _is_sharded_tensor,
-    set_element, narrow_tensor_by_index, _unwrap_xla_sharded_tensor, _CpuShards)
+    FLATTEN_MAPPING, flatten_state_dict, _is_sharded_tensor, set_element,
```
JackCaoG, line 39: do we need a test?

jonb377: Oh I added one but missed it with git, let me update the PR

JackCaoG (approved):

dudulightricks:
@jonb377 @JackCaoG Hey, just a reminder. We are still using this change on our side. Not urgent but needed.

ysiraichi:
@jonb377 What's the status of this PR? Is there a reason why it wasn't merged?

PR #7930, opened 2024-08-29
[benchmarks] Fix processing of PyTorch profiler events.

ysiraichi:
This PR fixes the benchmarking scripts when processing the recorded events from the PyTorch profiler. This modifies the scripts so that running with `--profile-cuda-and-cpu` and/or `--profile-cuda-and-cpu-individual-ops` works after https://github.com/pytorch/pytorch/pull/123247.

cc @miladm @zpcore


zpcore (approved):
LGTM!

PR #7934, opened 2024-08-30
Double loader_prefetch_size and device_prefetch_size

tengyifei:
This doubles Model FLOPs Utilization (MFU) of Llama-7B training on `v5p-512` TPUs with global batch size 1024.


alanwaketan (approved):

tengyifei:
Removed bug number

PR #7937, opened 2024-08-30
Update README.md with install torch/torch_xla wheel from a specific date

ManfeiBai:
Update README.md with install torch/torch_xla wheel from a specific date


README.md
```diff
@@ -33,9 +33,20 @@ To install PyTorch/XLA nightly build in a new TPU VM:
 
 ```
 pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
+
```
JackCaoG, line 36: there is already a section called `Use nightly build after 08/20/2024`

ManfeiBai: thanks, found it, sure, let's close this PR

PR #7916, opened 2024-08-27
Update Openxla-pin to Sep13, libtpu-pin to Sep13, jax to 0.4.33_nightly

ManfeiBai:
update OpenXLA-pin update to Sep13's commit: `54052bcbe0cda5b0a7be7c8e22eb2bfef025786e`

and also update libtpu version to Sep13

limit TPU CI tests `XLA_EXPERIMENTAL=nonzero:masked_select:nms python3 test/ds/test_dynamic_shapes.py -v` to only TPU v4 CI

PR #7946, opened 2024-09-03
[test] OpenXLA-pin update

ManfeiBai:


ManfeiBai:
failed too, will continue in https://github.com/pytorch/xla/pull/7916

PR #7949, opened 2024-09-03
Skip dynamo collective tests on TPU v2 CI

will-cromar:
Should fix test timeouts caused by `TestDistCollectiveOpsTpu.test_all_gather_dynamo`

image


JackCaoG (approved):

will-cromar:
:facepalm: This failed about a minute after I sent it for review. But it did get past the error in my screenshot. I'll follow up with another fix.

PR #7950, opened 2024-09-03
support dist.reduce_scatter

zpcore:
Support of torch.distributed.reduce_scatter for both dynamo and nondynamo case.


will-cromar (approved):
Thanks!

PR #7956, opened 2024-09-05
support dist.broadcast

zpcore:
Support torch.distributed.broadcast for both dynamo and nondynamo.


torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -309,6 +309,28 @@ AllGatherResultCoalesced BuildAllGatherCoalesced(
...
+  const torch::lazy::BackendDevice& device = xinput->GetDevice();
+  if (device.ordinal() == src) {
+    mask = at::ones_like(input);
+  } else {
+    mask = at::zeros_like(input);
+  }
+  XLATensorPtr xmask = bridge::GetXlaTensor(mask);
+  auto masked_input = tensor_methods::mul(xinput, xmask);
+  auto result = tensor_methods::all_reduce(masked_input, AllReduceType::kSum,
+                                           1.0, {}, true);
```
will-cromar, line 326: nit: name the non-obvious arguments at the end here. Assuming these two are scale and replica groups, `/*scale=*/1, /*groups=*/{}` (double check the names).

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -139,7 +140,7 @@ def test_all_to_all(self, pin_layout):
                                              list(range(world_size))]])
 
 
-@absltest.skipIf(lambda: tpu.num_logical_cores_per_chip() >= 2,
+@absltest.skipIf(tpu.num_logical_cores_per_chip() >= 2,
```
will-cromar, line 143: :facepalm: thanks

torch_xla/_internal/c10d_registration.py
```diff
@@ -1,22 +0,0 @@
-import torch
-import torch_xla.runtime as xr
-import torch_xla.core.xla_model as xm
-from typing import List, Optional, TypedDict
-
-
-# "broadcast(Tensor self, int src, str tag, int[] ranks, int group_size) -> Tensor",
-@torch.library.impl("_c10d_functional::broadcast", "XLA")
```
will-cromar, line 8: @JackCaoG FYI

torch_xla/csrc/cross_replica_reduces.cpp
```diff
@@ -309,6 +309,28 @@ AllGatherResultCoalesced BuildAllGatherCoalesced(
...
 }
 
+at::Tensor collective_broadcast(const at::Tensor& input, int64_t src,
+                                std::string) {
+  XLATensorPtr xinput = bridge::GetXlaTensor(input);
+  TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  at::Tensor mask;
+  const torch::lazy::BackendDevice& device = xinput->GetDevice();
+  if (device.ordinal() == src) {
+    mask = at::ones_like(input);
```
will-cromar, line 319: Is there an equivalent to `torch.no_grad()` in C++?  That's the only difference I see between the original python version and this one

zpcore: Searched the doc and we can use the following scope for tensor operation without grad:
```
  {
    at::NoGradGuard no_grad;
    // tensor operations
   }
```

zpcore: Anyone knows why we set no grad here: https://github.com/pytorch/xla/blob/c0501f01514216b3b86f66cfe9b15dbe1d23a0d6/torch_xla/_internal/c10d_registration.py#L17?  @JackCaoG

PR #7960, opened 2024-09-05
[torch_xla2] Fix type promotions by `aten::copy_` and `aten::rand_like`

will-cromar:
Ensure `copy_` doesn't modify the target tensor's dtype and add a test case. For example:

```
>>> import torch
>>> t1 = torch.ones((3, 3), dtype=torch.int32)
>>> t2 = torch.zeros((3, 3), dtype=torch.float32)
>>> t1.copy_(t2)
tensor([[0, 0, 0], # values are copied
        [0, 0, 0],
        [0, 0, 0]], dtype=torch.int32) # but dtype still int32
```

Remove a `div_` test case that doesn't match `torch` behavior, but relies on incorrect `copy_` behavior:

```
>>> import torch
>>> x = torch.tensor([1, 2, 3], dtype=torch.int32)
>>> y = torch.tensor([4, 5, 6], dtype=torch.int32)
>>> x.div_(y)
RuntimeError: result type Float can't be cast to the desired output type Int
```

Tweak a distributed test case that relied on `div_` casting dtypes.

`rand_like` was not propagating `dtype` at all.

### DIFF
--- a/experimental/torch_xla2/test/test_core_aten_ops.py
+++ b/experimental/torch_xla2/test/test_core_aten_ops.py
@@ -7,20 +7,17 @@ from . import test_base
 from torch.utils import _pytree as pytree
 
 
-def diff_output(testcase, output1, output2, rtol, atol, equal_nan=True):
+def diff_output(testcase, output1, output2, rtol, atol, equal_nan=True, check_dtype=False):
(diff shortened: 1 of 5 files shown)


ManfeiBai (approved):
Thanks, LGTM

PR #7963, opened 2024-09-05
Unpin `scipy` in upstream CI image

will-cromar:
Tested with a local build this time and it passes.

This version of scipy is ancient: https://pypi.org/project/scipy/1.6.3/


ManfeiBai (approved):
Thanks, LGTM

PR #7971, opened 2024-09-06
Fix nprocs description for xmp.spawn; error if nprocs != 1 or None

jeffhataws:
If you specify xmp.spawn argument nprocs != 1 or None, currently there's a warning but execution continues. This can be a source of confusion.

This change fixes the descriptions for this argument and make it return an error.


torch_xla/core/xla_env_vars.py
```diff
@@ -2,6 +2,7 @@
 TPU_NUM_DEVICES = 'TPU_NUM_DEVICES'
 GPU_NUM_DEVICES = 'GPU_NUM_DEVICES'
 CPU_NUM_DEVICES = 'CPU_NUM_DEVICES'
+NEURONCORE_NUM_DEVICES = 'NEURONCORE_NUM_DEVICES'
```
JackCaoG, line 5: this env var is not being used anywhere? If we really need it can it just be `NEURON` since the device type is `NEURON`.

jeffhataws: Removed. Thanks.

JackCaoG (approved):

PR #7995, opened 2024-09-11
Use a different str for device instead of meta

### LINKED ISSUE #7966: Report real device for XLATensor2
Currently we use `meta` as device for the tensor subclass `XLATensor2` to avoid allocating tensor. However, there should be a way to retain device information and still not allocate tensor; such as what `FakeTensor` was able to.

qihqi:
Implements this idea: https://github.com/pytorch/xla/pull/7705 but without the C++ piece.

We could use anything for the device (incl just a string: https://github.com/albanD/subclass_zoo/blob/main/new_device.py); however it's advantagous to use a torch.device with an existing registered key. (we could use 'xla:0' too). 

The caveat is that we need to capture tensor constructors in jtorch instead of jaten (like it used to be) because otherwise torch will check if such device module is loaded or not (example for capturing 'cuda').

fixes #7966

### DIFF
--- a/experimental/torch_xla2/torch_xla2/ops/jaten.py
+++ b/experimental/torch_xla2/torch_xla2/ops/jaten.py
@@ -2,6 +2,7 @@
 
 import sys
 from typing import Optional, Sequence
+import functools
 
 import jax
 from jax import numpy as jnp
@@ -38,6 +39,7 @@ mutation_ops_to_functional = {
   torch.ops.aten.squeeze_: torch.ops.aten.squeeze,
   torch.ops.aten.bernoulli_: torch.ops.aten.bernoulli.p,
   torch.ops.aten.clamp_: torch.ops.aten.clamp,
+  torch.ops.aten.random_: torch.ops.aten.uniform,
 }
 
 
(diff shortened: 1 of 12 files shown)


experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -1634,6 +1645,7 @@ def _aten_arange(
   requires_grad=False,
   device=None,
   pin_memory=False,
+  **kwargs
 ):
```
ManfeiBai, line 1649: do we need add `**kwargs` here to accept `**kwargs` from `res2 = func(input2, *args2, **kwargs2)`?

I seems met one situation that: func in `jaten.py` file didn't received `kwargs2` from `res2 = func(input2, *args2, **kwargs2)`

ManfeiBai (approved):
Thanks, amazing change, LGTM

experimental/torch_xla2/torch_xla2/ops/jaten.py
qihqi: yeah this is not needed. removed.

PR #7998, opened 2024-09-12
Remove contiguity assertion from `XLATensorImpl`.

ysiraichi:
This PR reflects the changes in the PyTorch PR https://github.com/pytorch/pytorch/pull/135237. Basically, it removes the contiguity assertion due to a possible metadata change in the functionalization kernel.

Specifically, at the end of each functionalization kernel, `at::functionalization::impl::set_sizes_strides_offset` is called. It modifies the metadata of the output tensors, so that they match the metadata of eager. Such a metadata update might set `TensorImpl::is_contiguous_` to false.

cc @miladm @JackCaoG @alanwaketan

### DIFF
--- a/torch_xla/csrc/tensor_impl.cpp
+++ b/torch_xla/csrc/tensor_impl.cpp
@@ -173,8 +173,8 @@ int64_t XLATensorImpl::numel_custom() const {
 }
 
 bool XLATensorImpl::is_contiguous_custom(at::MemoryFormat memory_format) const {
-  // Only check that the storage is already contiguous.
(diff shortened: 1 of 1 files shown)


JackCaoG:
any chance you can add a test?

ysiraichi:
Problem is, I don't think anything will change (e.g. strides and storage offset) before merging https://github.com/pytorch/pytorch/pull/135498. So, not sure what kind of test would make sense, here.

JackCaoG (approved):

PR #8002, opened 2024-09-12
add `histogram` .. `linalg.eigvals`

ManfeiBai:
add `histogram` .. `linalg.eigvals`

### DIFF
--- a/experimental/torch_xla2/test/test_ops.py
+++ b/experimental/torch_xla2/test/test_ops.py
@@ -76,7 +76,6 @@ skiplist = {
     "linalg.det",
     "linalg.eig",
     "linalg.eigh",
-    "linalg.eigvals",
     "linalg.eigvalsh",
     "linalg.householder_product",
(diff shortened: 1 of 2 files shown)


experimental/torch_xla2/test/test_ops.py
```diff
@@ -303,6 +299,7 @@
   'uniform',
   # Dropout is not deterministic https://pytorch.org/docs/stable/generated/torch.nn.functional.feature_alpha_dropout.html
   'nn.functional.feature_alpha_dropout',
+  'histc', # result is not stable for single negative scalar
```
qihqi, line 302: try to remove it? so there is some randomness, in my lastest PR I added torch.manual_seed(0) and it seems to pass after that

ManfeiBai: thanks, updated

qihqi (approved):

Issue #7958, opened 2024-09-05
Op info test for `logdet`


ManfeiBai:
## Fix the Op info test for `logdet`

1. Find `"logdet",` [test_ops.py](test/test_ops.py) and remove
   `logdet ` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

yenkwang:
It is fixed at [3a9f53f](https://github.com/pytorch/xla/commit/3a9f53f4943db599ca5813ac676bf2fbc51d3036). We can close the issue.

PR #8004, opened 2024-09-13
Op info test for logdet, linalg.slogdet (#7503) (#7958)

yenkwang:
Op info test for logdet, linalg.slogdet


ManfeiBai (approved):
Thanks, LGTM

PR #8006, opened 2024-09-13
[benchmarks] Disallow `XRT` for XLA:CUDA.

ysiraichi:
This PR filters out "XRT" as an option for `xla` configuration on non-TPU experiments.

cc @miladm @zpcore


zpcore (approved):
Thanks, actually we doesn't support `xrt` anymore. `xrt` has already been cleared in runtime.py etc..

PR #8008, opened 2024-09-13
Set xla_tpu_enable_flash_attention=false to enable libtpu pin update

bhavya01:
This flag enables FlashAttention HLO pass that pattern matches attention and rewrites it as flash attention. This pattern matching is causing issues for our standard dot product attention. Turning it off till we fix the issue with pattern matching.

Also need to update the payload for pallas test.


ManfeiBai (approved):
Thanks, LGTM

PR #8011, opened 2024-09-14
Limit `XLA_EXPERIMENTAL=nonzero:masked_select:nms python3 test/ds/test_dynamic_shapes.py -v` only run on v4

ManfeiBai:


ManfeiBai:
included in https://github.com/pytorch/xla/pull/7916

PR #8016, opened 2024-09-15
[hard] add `masked.median`

ManfeiBai:
`pytest test` failed with error: `RuntimeError: Object of type 'XLATensor2' is not an instance of 'sequence'` at current branch


ManfeiBai:
`masked.median` is decomposed with `nanmedian.dim`,  and `nanmedian.dim` will return two result according to https://pytorch.org/docs/stable/generated/torch.nanmedian.html, but my logic only return one value,

so the logic of `nanmedian.dim` need a sequence of tensor return, when I only return one tensor, it shows error `RuntimeError: Object of type 'XLATensor2' is not an instance of 'sequence'`

Issue #7514, opened 2024-06-25
Op info test for `nanquantile .. native_layer_norm`


qihqi:
## Fix the Op info test for `nanquantile .. native_layer_norm`

1. Find the lines 188 to 192 of [test_ops.py](test/test_ops.py) and remove
   `nanquantile` .. `native_layer_norm` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
cc @matinehakhlaghinia, its would be the first issue to try to add torchxla2 ops

matinehAkhlaghinia:
> cc @matinehAkhlaghinia, its would be the first issue to try to add torchxla2 ops

Sounds good, I can work on this

ManfeiBai:
Hi, @matinehAkhlaghinia, thanks for contribution

please feel free to continue to add `nanquantile` and `native_layer_norm`, and we could mention them here too

matinehAkhlaghinia:
Hi, @ManfeiBai, this issue is done now. What would be a good second issue to pick up?

ManfeiBai:
Thanks for your amazing contribution, @matinehAkhlaghinia! These five ops would be the second issue to try: https://github.com/pytorch/xla/issues/7410, https://github.com/pytorch/xla/issues/7413, https://github.com/pytorch/xla/issues/7488, https://github.com/pytorch/xla/issues/7489, https://github.com/pytorch/xla/issues/7490

PR #8021, opened 2024-09-16
Add fixes for narrow_copy and narrow ops

matinehAkhlaghinia:
Adds support for nansum, narrow_copy and narrow. This is a partial fix for [#7514](https://github.com/pytorch/xla/issues/7514).

- nansum: the tests passed by default when I removed it.
- narrow_copy: the lowering was missing.
- narrow: this was decomposed to slice and the slice implementation was not wrapping the negative dimensions correctly. Wrapping the dimension fixed the tests.


ManfeiBai (approved):
Thanks, LGTM

PR #8023, opened 2024-09-16
[benchmarks] Add docs file.

ysiraichi:


docs/torchbench.md
```diff
@@ -0,0 +1,203 @@
...
+creates 2 new files. For more detailed information, check out [this post][9] explaining
+PyTorch profiling:
+
+1. Chrome trace: same as calling `profiler.export_chrome_trace`
+2. Profile: same as calling `profiler.key_averages().table(...)`, a summarized table with
+   the timing averages
+
+**PyTorch/XLA Metrics:** (repeat-specific) the flag `--dump-pytorch-xla-metrics` creates a
+new file, dumping PyTorch/XLA metrics, such as graph compiling and execution information.
+
```
miladm, line 194: Can we have a section on Nighly CI runs - TPU and GPU

miladm, line 194: Can we include a troubleshooting section? this should include the pitfalls we ran into as well as approaches to debug the benchmark efficiently.

ysiraichi: Sure. But I don't really have any information on those nightly CI runs. What should it contain?

amjames: @miladm I don't think it is appropriate to document any resolved issue here. If there are any features that can be *used* to debug different performance issues we should absolutely make sure they are captured.

miladm: @zpcore can you please help Yukio with this section?

miladm: additional: ideally, we also want to run verifier as part of our tests

miladm:
@ysiraichi 
* can we have a section called "tips and tricks" covering some of the ways we found helping with the performance gain of torch_xla. This can include flags like functionalization, cuda fallback, cuda garphs, etc.

PR #8029, opened 2024-09-17
Implement empty_like in torch aten ops

nupurbaghel:
This op is not part of the skip list so no test count changes.
Fixes https://github.com/pytorch/xla/issues/7386

Did a quick verification by comparing outputs from torch and jax:
```
$a=torch.empty((2,3), dtype=torch.int32)
$print(torch.empty_like(a))
tensor([[88391536,    32347, 88390000],
        [   32347, 88391088,    32347]], dtype=torch.int32)

$b=jnp.empty((2,3), dtype=jnp.int32)
$jnp.empty_like(b)
Array([[0, 0, 0],
       [0, 0, 0]], dtype=int32)
```


ManfeiBai (approved):
Thanks, LGTM

PR #8030, opened 2024-09-17
[torchxla2] filter passed ops

ManfeiBai:
filter current passed ops


ManfeiBai:
duplicated in https://github.com/pytorch/xla/pull/8035

PR #8032, opened 2024-09-17
Forward `XLATensorImpl::is_contiguous_custom` to `TensorImpl`.

ysiraichi:
This PR fixes #7998. Instead of always returning true, we forward this call to the base class `TensorImpl::is_contiguous_custom()`.

The reason is that after https://github.com/pytorch/pytorch/pull/135498 is merged, XLA tensors' metadata might stop reflecting on the actual XLA storage. Which means that the tensors' strides might not always be contiguous. Whenever that happens, `tensor.is_contiguous()` call should be consistent with the tensors' strides.

cc @miladm @JackCaoG @alanwaketan

### DIFF
--- a/torch_xla/csrc/tensor_impl.cpp
+++ b/torch_xla/csrc/tensor_impl.cpp
@@ -175,7 +175,7 @@ int64_t XLATensorImpl::numel_custom() const {
 bool XLATensorImpl::is_contiguous_custom(at::MemoryFormat memory_format) const {
   // Storage is always contiguous, but the tensor metadata is_contiguous_ might
   // be false due to the update in the functionalization layer..
-  return true;
+  return c10::TensorImpl::is_contiguous_custom(memory_format);
 }
 
 void XLATensorImpl::SetupSizeProperties() {
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2783,6 +2783,28 @@ class TestGeneric(test_utils.XlaTestCase):
     buf_ptr_3 = torch_xla._XLAC._unsafe_buffer_pointer(xla_tensor_3)
     self.assertGreaterEqual(buf_ptr_3, 0)
 
+  def test_consistent_strides(self):
+    def stride_is_contiguous(tensor):
+      sizes_and_strides = list(sorted(zip(tensor.shape, tensor.stride()), key=lambda t: t[1]))
+      if sizes_and_strides[0][1] != 1:
+        return False
+      for i, (size, stride) in enumerate(sizes_and_strides[:-1]):
+        if stride[i + 1] != stride[i] * size[i]:
+          return False
+      return True
+
+    def assert_consistent(tensor):
+      self.assertEquals(tensor.is_contiguous(), stride_is_contiguous(tensor))
+
+    a = torch.rand(10).to(xm.xla_device())
+    assert_consistent(a)
+
+    b = a[::2]
+    assert_consistent(b)
+
+    c = b[1:]
+    assert_consistent(c)
+
 
 class TestDLPack(parameterized.TestCase):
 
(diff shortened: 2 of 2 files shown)


JackCaoG:
test failure seems real?

JackCaoG:
yea `test_memory_format_preserved_after_permute_xla ` still failing.

test/spmd/test_xla_sharding.py
```diff
@@ -1068,7 +1068,7 @@ def test_backward_optimization_barrier(self):
 
     hlo = torch_xla._XLAC._get_xla_tensors_hlo([model.fc2.weight.grad])
     self.assertIn(
-        '%opt-barrier.37 = (f32[1,64]{0,1}, f32[1]{0}, f32[2,64]{1,0}) opt-barrier((f32[1,64]{0,1}, f32[1]{0}, f32[2,64]{1,0}) %tuple.36)',
+        '%opt-barrier.38 = (f32[1,64]{1,0}, f32[1]{0}, f32[2,64]{1,0}) opt-barrier((f32[1,64]{1,0}, f32[1]{0}, f32[2,64]{1,0}) %tuple.37)',
```
ysiraichi, line 1071: This was needed for fixing a CI failure in https://github.com/pytorch/pytorch/pull/135237. @JackCaoG let me know if you think this should not be happening.

torch_xla/experimental/scan.py
ysiraichi, line 1: The changes in this file are needed because it seems that PyTorch is running these tests with Python 3.8. Thus, the subscript operator is not allowed for types.

JackCaoG:
no, I don't want to add features to 2.5 releases at this point.

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -173,9 +173,16 @@ int64_t XLATensorImpl::numel_custom() const {
 }
 
 bool XLATensorImpl::is_contiguous_custom(at::MemoryFormat memory_format) const {
+  // If functionalization is disabled, the tensors' metadata aren't being
+  // updated w.r.t. the output of meta functions. Therefore, we fallback to the
+  // old behavior returning true, always.
+  if (runtime::sys_util::GetEnvBool("XLA_DISABLE_FUNCTIONALIZATION", false)) {
```
JackCaoG, line 179: maybe just add a `IsFunctionalizationDisabled`(use a static bool to store the value of this env var) in https://github.com/pytorch/xla/blob/ea8c47a345a29c6a1b1bdf4ee38a9159b07a980f/torch_xla/csrc/xla_graph_executor.h or in `tensor.h` instead of keep getting the from the env var

torch_xla/experimental/scan.py
JackCaoG: hopefully https://github.com/pytorch/pytorch/pull/135278 can be merged and we don't have this issue anymore in the future...

JackCaoG (approved):

ysiraichi:
I don't think the approach in this PR will be enough for propagating the metadata. I think I will have to modify the [`functionalizeFallback` function](https://github.com/pytorch/pytorch/blob/5033a1ca0dd22dae34a8939add33dbebfe0fd31d/aten/src/ATen/FunctionalizeFallbackKernel.cpp#L31) instead, so that all operations propagate the metadata correctly. This will probably add more overhead to the execution, though. @JackCaoG what do you think?

JackCaoG:
would be good to measure those overheads, my experience is that most of the C++ ops are pretty fast except creating large objects and calculating hash. It might be not too bad in this case too.

miladm:
@ysiraichi do we know what else is remaining for this PR to become ready for merging?

ysiraichi:
Before merging this branch, we need to first merge its PyTorch counterpart https://github.com/pytorch/pytorch/pull/135498. In order to merge that one, we need to propagate the Tensor metadata for all operations: manually written kernels and codegen'd ones. Note that this might break previous assumptions that XLA tensors `contiguous()` calls would always return true.

PR #8033, opened 2024-09-17
Fix unbind op.

cornmander:
The test passed with the previous implementation but I created a slightly more efficient one anyway.

Partial fix for issue #7541.


qihqi (approved):

qihqi:
thanks!

PR #8034, opened 2024-09-17
Add profiler annotation for the decoderonly example

JackCaoG:
Originally I want to keep this file pure pytorch but many times I use this example to do real debugging and I have to readd annotations every time which is a bit annoying.


alanwaketan (approved):
LGTM.

JackCaoG:
Ok I realized I can't do this because profile annotation is not `torch.compile` friendly and I use this file in a unit test. closing this issue.

PR #8035, opened 2024-09-17
Uncomment tests that are passing.

cornmander:
I unskipped the tests in the skiplist and checked which tests were actually passing.

I changed the conditional  ```ops_to_test = [
    test for test in op_db
    if ([test.name](http://test.name/) not in skiplist and
        test.variant_test_name not in variant_test_name_to_skip)
] ``` to ONLY test the skipped tests and then re-commented the ones that failed.


ManfeiBai (approved):
Thanks, LGTM

Issue #7510, opened 2024-06-25
Op info test for `masked.norm .. masked_select`


qihqi:
## Fix the Op info test for `masked.norm .. masked_select`

1. Find the lines 168 to 172 of [test_ops.py](test/test_ops.py) and remove
   `masked.norm` .. `masked_select` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
Hi, @anishfish2, would you mind help to reply this PR? this PR would be your first issue, and after you comment, I would be able to assign it to you

anishfish2:
Claiming this issue!

ManfeiBai:
let's reopen this ticket for `masked_scatter` and `masked_select`

ManfeiBai:
Thanks for the contriburtion!

PR #8037, opened 2024-09-18
Fixed masked_scatter and masked_select

anishfish2:
Implementing masked_scatter and masked_select op. This PR should mark conclusion of all ops in #7510.


ManfeiBai (approved):
Thanks, LGTM

Issue #7378, opened 2024-06-25
Op info test for `diag_embed`


qihqi:
## Fix the Op info test for `diag_embed`

1. Find the line 44 of [test_ops.py](test/test_ops.py) and remove
   `diag_embed` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
hi, cc @tqfan28, this github issue would be one of the second issue to try, please feel free to reply this issue, so that we could assign this issue to you

ManfeiBai:
https://github.com/pytorch/xla/pull/8036

ManfeiBai:
hi, this op passed test in https://github.com/pytorch/xla/pull/8036

PR #8038, opened 2024-09-18
Implement torch diagflat, and movedim as well as it is also needed as a helper function (#7378)

simonteozw:
Need some guidance as to whether or not this was the right approach. Added movedim as the tests failed with `torch_xla2.tensor.OperatorNotFound: Operator with name aten::movedim.int has no lowering`

### DIFF
--- a/experimental/torch_xla2/test/test_ops.py
+++ b/experimental/torch_xla2/test/test_ops.py
@@ -26,7 +26,6 @@ skiplist = {
     "combinations",
     "complex",
     "diag_embed",
-    "diagflat",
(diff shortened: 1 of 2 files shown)


ManfeiBai:
> Need some guidance as to whether or not this was the right approach. Added movedim as the tests failed with `torch_xla2.tensor.OperatorNotFound: Operator with name aten::movedim.int has no lowering`

Saw you have added implementation for 'movedim', do we still met above error now?

ManfeiBai (approved):
Thanks, LGTM

PR #8039, opened 2024-09-18
parallel_loader: Fix TPU memory leak when calling __iter__.

dudulightricks:
A TPU memory leak occurs when calling __iter__ on the MpDeviceLoader object, even if called only once. However, the memory growth becomes more noticeable and critical when __iter__ is called repeatedly, eventually leading to a crash. This issue is caused by the threads not being properly terminated because the close() method was not invoked.

This commit resolves the issue by ensuring that close() is called, which properly shuts down the threads and prevents memory from leaking.

### DIFF
--- a/torch_xla/distributed/parallel_loader.py
+++ b/torch_xla/distributed/parallel_loader.py
@@ -95,12 +95,14 @@ class ParallelLoader(object):
     self._done = False
     self._queues = dict()
     self._input_sharding = input_sharding
+    self._threads = []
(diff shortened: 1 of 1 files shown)


dudulightricks:
@JackCaoG

will-cromar (approved):

dudulightricks:
@will-cromar Thanks. Fixed.

dudulightricks:
@will-cromar Merging?

PR #8040, opened 2024-09-18
Fix quantile, nanquantile and gather

matinehAkhlaghinia:
Adding fixes for nanquantile as part of [#7514](https://github.com/pytorch/xla/issues/7514). 

- Also remove quantile from skiplist, as once I fixed nanquantile, it was easy to fix quantile as well.
- nanquantile depended on gather implementation which was not working correctly, once I fixed that, the tests related to gather also started passing so removing that from skiplist too.
- nanquantile and quantile depended on a bunch of other ops, such as logical_not_, ceil_, unsqueeze_ and transpose_, so added those to the mutation_ops_to_functional mapping.


ManfeiBai (approved):
Thanks, LGTM

experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -2043,6 +2042,10 @@ def _aten_frexp(input):
 # aten.gather
 @op(torch.ops.aten.gather)
 def _aten_gather(input, dim, index):
+  if input.ndim == 0:
```
danielvegamyhre, line 2045: Just curious, how did you debug the gather implementation? I also had traced `quantile` problems to the `gather` implementation but had not yet started investigating further. I was going to compare it to other gather implementations in jax/pytorch as a starting point, what did you end up doing?

PR #8046, opened 2024-09-19
Introduce multi-node SPMD initialization for Neuron

rpsilva-aws:
In this PR, we adapt to account for a new initialization path that supports multi-node SPMD in Neuron. In order to minimize this change, we retain the `xla.init()` API, but introduce a reinitialization for PJRT alone once SPMD is enabled. Since enabling SPMD follows the initial Neuron initialization, we require reconfiguring once this is enabled, and if the user did not explicitly set `XLA_USE_SPMD` (via `is_spmd()`, as it is currently recommended). Under the hood, both APIs will guarantee that the environment is correctly configured when SPMD is enabled.


torch_xla/runtime.py
```diff
@@ -253,6 +253,14 @@ def use_spmd(auto: Optional[bool] = False):
     torch_xla._XLAC._xla_set_auto_sharding()
     os.environ["XLA_AUTO_SPMD"] = "1"
 
+  if os.environ[xenv.PJRT_DEVICE] == 'NEURON':
```
will-cromar, line 256: nit: use `device_type` here: https://pytorch.org/xla/release/r2.4/index.html#torch_xla.runtime.device_type

will-cromar (commented):
Otherwise LGTM

PR #8048, opened 2024-09-23
[torch_xla2] Fix nn.functional.conv2d and conv3d

dvhg:
Replicate PyTorch behavior when using single number padding. Ref: https://pytorch.org/docs/stable/generated/torch.nn.functional.conv2d.html#torch.nn.functional.conv2d


ManfeiBai (approved):
Thanks, LGTM

PR #8056, opened 2024-09-24
Support spmd distx init and compute dtype

lausannel:
This PR does the following:
1. Supports using `torchdistx` for model initialization to reduce memory usage
2. Supports different compute dtypes

### DIFF
--- a/torch_xla/experimental/spmd_fully_sharded_data_parallel.py
+++ b/torch_xla/experimental/spmd_fully_sharded_data_parallel.py
@@ -11,6 +11,9 @@ import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.distributed.spmd as spmd
 from torch_xla.distributed.fsdp.wrap import recursive_wrap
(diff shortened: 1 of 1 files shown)


JackCaoG:
AFAIK torchdistx is not being maintained, has that changed?

lausannel:
You're correct. It's worth noting that `torchdistx` is still in use, and a similar implementation logic is employed in FSDPv1.

https://github.com/pytorch/xla/blob/72fda76ed910d81f9704c19f625818c3e52aa253/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py#L486-L490

JackCaoG:
actually I would prefer not to add this logic here. I think pytorch has been trying to use the meta tensor to solve the parameter issue, I prefer not to add additional logics to a library that has been deprecated.

lausannel:
I have removed all the torchdistx-related content as per your suggestion. Could you please review the changes again?

PR #8045, opened 2024-09-19
[torch_xla2] Fix scatter and scatter_add

dvhg:
This fixes the op test for `scatter` and `scatter_add`. `scatter_reduce` is still broken as `include_self = False` is unsupported.

PR #8058, opened 2024-09-24
[torch_xla2] Fix op_info test for scatter_reduce

dvhg:
Following up on #8045. 

Note: Behavior for `reduce='mean'` with integer types doesn't seem well defined in torch docs. The current implementation passes tests but could be fragile.


ManfeiBai (approved):
Thanks, LGTM

PR #8064, opened 2024-09-24
Support dist.all_to_all_single

zpcore:
Support to use torch.distributed.all_to_all_single to both dynamo and nondynamo case.

Note that there is a function signature mismatch between torch's [all_to_all_single](https://github.com/pytorch/pytorch/blob/758d78790164bfb041555daed380de96e06f78a3/torch/distributed/distributed_c10d.py#L3880) and xla op's [AllToAll](https://openxla.org/xla/operation_semantics#alltoall).  To leverage the AllToAll op, we doesn't support specifying the `input_split_sizes` and `output_split_sizes` at this time. Check test_collective_ops_tpu.py for the usage.


qihqi (approved):

will-cromar (approved):

will-cromar:
Thanks!

PR #8066, opened 2024-09-25
Use Tuple instead of tuple in scan.py

tengyifei:


ManfeiBai:
Thanks, could we merge to `r2.5` branch to unblock upstream test? current PR is changing `master` branch

Issue #7479, opened 2024-06-25
Op info test for `linalg.eig`


qihqi:
## Fix the Op info test for `linalg.eig`

1. Find the line 114 of [test_ops.py](test/test_ops.py) and remove
   `linalg.eig` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
Hi, @shitongWang, thanks for the contribution, this issue would be part of the second issues to try, please feel free to comment, so that we could assign this ticket to you

ManfeiBai:
Hi, @anishfish2, thanks for your contribution! this is one of the second 5 ops issues, and please feel free to reply this issue, so that we could assign this to you

anishfish2:
Claiming this issue!

ManfeiBai:
synced offline, due to jnp and torch complex64 and complex128 difference, would suggest to specific atol/rtol used to comfirm difference of `linalg.eig` from torchxla2 and jax, like https://github.com/pytorch/xla/blob/d79a37c408301bf7f845c74e32e0c973138fdbeb/experimental/torch_xla2/test/test_ops.py#L198C1-L198C69

anishfish2:
Sounds good, I'll try this!

PR #8068, opened 2024-09-25
Avoids error removing linalg.eig from skiplist by modifying atol and rtol

anishfish2:
Fixes linalg.eig in #7479.


ManfeiBai (approved):
Thanks, LGTM

Issue #7524, opened 2024-06-25
[hard] Op info test for `nn.functional.max_pool2d .. nn.functional.max_unpool3d`


qihqi:
## Fix the Op info test for `nn.functional.max_pool2d .. nn.functional.max_unpool3d`

1. Find the lines 238 to 242 of [test_ops.py](test/test_ops.py) and remove
   `nn.functional.max_pool2d` .. `nn.functional.max_unpool3d` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
hi, @hosseinsarshar, this issue would be the first issue to try, and please feel free to reply in this ticket, so that we would assign this issue to you

hosseinsarshar:
thanks @ManfeiBai - please go ahead and assigned it to me.

ManfeiBai:
close current issue and move `nn.functional.max_pool2d` and `nn.functional.max_pool3d` to https://github.com/pytorch/xla/issues/8086

PR #8085, opened 2024-09-27
 Reduced the max_unpoolxd logic into one function #7524

hosseinsarshar:
Reduced the max_unpool2d and max_unpool3d into a single function -


ManfeiBai (approved):
Thanks, LGTM

PR #8091, opened 2024-09-28
Implement `torch.ops.aten.embedding_renorm_`

guyao:
Add lowering `torch.ops.aten.embedding_renorm_` to support embedding.


ManfeiBai (approved):
Thanks, LGTM

PR #8202, opened 2024-10-01
Allow MpDeviceLoader to shard dictionaries of tensor 

bhavya01:
Reland https://github.com/pytorch/xla/pull/7912

The new code is only used when input_sharding is a dictionary of ShardingSpec. So, it should be backward compatible. Before this PR, it is expected to be a ShardingSpec

### DIFF
--- a/docs/spmd_advanced.md
+++ b/docs/spmd_advanced.md
@@ -14,6 +14,24 @@ train_loader = pl.MpDeviceLoader(
          input_sharding=xs.ShardingSpec(input_mesh, ('data', None, None, None)))
 ```
 
+It is also possible to specify a different `input_sharding` for each element of the batch if they are different shapes:
+
+```python
+# if batch = next(train_loader) looks like
+# {'x': <tensor of shape [s1, s2, s3, s4]>, 'y': <tensor for shape [s1, s2]>}
+
+# MpDeviceLoader returns ParallelLoader.per_device_loader as iterator
+train_loader = pl.MpDeviceLoader(
+         train_loader,  # wraps PyTorch DataLoader
+         device,
+	       # specify different sharding for each input of the batch.
+         input_sharding={
+          'x': xs.ShardingSpec(input_mesh, ('data', None, None, None)), 
+          'y': xs.ShardingSpec(input_mesh, ('data', None))
+        }
+)
+```
+
 ### Virtual Device Optimization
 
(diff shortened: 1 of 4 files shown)


bhavya01:
@ManfeiBai: I will cherry pick this into 2.5 release.
@zpcore: This should make SD2 script compatible with future nighty versions.

JackCaoG:
I will take a look today

docs/spmd_advanced.md
```diff
@@ -14,6 +14,24 @@ train_loader = pl.MpDeviceLoader(
...
+
+```python
+# if batch = next(train_loader) looks like
+# {'x': <tensor of shape [s1, s2, s3, s4]>, 'y': <tensor for shape [s1, s2]>}
+
+# MpDeviceLoader returns ParallelLoader.per_device_loader as iterator
+train_loader = pl.MpDeviceLoader(
+         train_loader,  # wraps PyTorch DataLoader
+         device,
+	       # specify different sharding for each input of the batch.
```
JackCaoG, line 27: fix indent

torch_xla/distributed/parallel_loader.py
```diff
@@ -46,6 +46,8 @@ def next(self):
         self._batches_yielded += 1
 
     item = self._loader.next_item(self._device)
+    if isinstance(item, Exception):
+      raise item
```
JackCaoG, line 50: Interesting. I was trying to handle the exception differently. I added an error queue in https://github.com/pytorch/xla/pull/8205/files#diff-0633db2f65b315c31a311c41cdc6223a146d2e9233e51907bba45ee000734e87R100

JackCaoG (approved):
I still prefer to separate the error handling instead of sending an exception into the data_queue. I will approve to unblock, I can do a follow up here.

torch_xla/distributed/parallel_loader.py
bhavya01: Agreed. Having a separate exception queue is cleaner. Do you mind if I copy it from your PR as I am going to cherry pick into 2.5 release

JackCaoG: Let's do this. merge this one, backport it. Then I will merge my change. Then you can update. This way we avoid merge conflicts. I also want to get this one into release ASAP.

PR #8205, opened 2024-10-02
Raise error when minibatch is used in SPMD dataloading and per host b…

JackCaoG:
…atch size is not divisble by data mesh

When `minibatch` is used in sharding spec, each device is loading it own part of the data and reassemble it during the runtime. 

Imagenine a case the mesh is (4,1) and the batch size we used is 3, each device will see a data like
```
(100,200,300,0)
(400,500,600,0)
```
and runtime will reassemble a incorrect final data.

If `mini batch` is not used it is OK, because each device will see
```
(100, 200, 300,400)
(500, 600, 0, 0)
```
and we will drop padding at the end.

### DIFF
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -1309,9 +1309,8 @@ def send_cpu_data_to_device(
                                               data_mesh_dim) != 0:
           raise RuntimeError(
               "When minibatch is configured, batch dimension of the tensor " +
-              "must be divisible by data mesh dimension.input data shape " + 
-              f"={tensor.size()}, mesh data dimension = {data_mesh_dim}"
-          )
+              "must be divisible by data mesh dimension.input data shape " +
+              f"={tensor.size()}, mesh data dimension = {data_mesh_dim}")
 
     xtensors = torch_xla._XLAC._xla_tensors_from_aten(tensors, devices,
                                                       shardings)
--- a/torch_xla/core/xla_model.py
+++ b/torch_xla/core/xla_model.py
@@ -1299,6 +1299,20 @@ def send_cpu_data_to_device(
     shardings = None
     if input_sharding:
       shardings = [input_sharding.xla_spec(t) for t in tensors]
+    if input_sharding and input_sharding.minibatch:
+      # when minibatch is configured we must make sure batch dimension of
+      # the tensor is divisible by the data mesh dimension.
+      data_mesh_dim = input_sharding.mesh.mesh_shape[0]
+      for tensor, sharding in zip(tensors, shardings):
+        # assume batch dimension is 0
+        if sharding and tensor.dim() > 0 and (tensor.size()[0] %
+                                              data_mesh_dim) != 0:
+          raise RuntimeError(
+              "When minibatch is configured, batch dimension of the tensor " +
+              "must be divisible by data mesh dimension.input data shape " + 
+              f"={tensor.size()}, mesh data dimension = {data_mesh_dim}"
+          )
+
     xtensors = torch_xla._XLAC._xla_tensors_from_aten(tensors, devices,
                                                       shardings)
     return xtensors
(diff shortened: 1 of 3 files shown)


torch_xla/core/xla_model.py
```diff
@@ -1299,6 +1299,19 @@ def convert_fn(tensors):
     shardings = None
     if input_sharding:
       shardings = [input_sharding.xla_spec(t) for t in tensors]
+    if input_sharding and input_sharding.minibatch:
+      # when minibatch is configured we must make sure batch dimension of
+      # the tensor is divisible by the data mesh dimension.
+      data_mesh_dim = input_sharding.mesh.mesh_shape[0]
+      for tensor, sharding in zip(tensors, shardings):
+        # assume batch dimension is 0
+        if sharding and tensor.dim() > 0 and (tensor.size()[0] %
```
jonb377, line 1308: Trying to understand the issue a bit more - `minibatch` means each host is loading data for its local devices, and the result will be distributed across the local devices and viewed as a global tensor.

So if each host loads 3 examples and has 4 devices, we'll end up with host0 having `[1,2,3,0]` after padding its local example and host1 having `[4,5,6,0]` after padding its own.

Then the final global tensor has data `[1,2,3,0,4,5]` (after concat/unpadding) instead of the desired `[1,2,3,4,5,6]`?

If that's the case, then we should compare the local batch size to the local device count instead of the local shard to the mesh shape?

(I don't recall the full semantics of the API here, but I believe we take in a tensor of the local batch and output a tensor representing the global batch when `minibatch=True`)

JackCaoG: oh you right.. but the issue is that we might have a 2d mesh of `(4,4,None,None)`. I was trying to fix that issue here.. Now I think about it, if that's the case what should I do...

JackCaoG: easier solution is we alwasy assume 1d mesh for the data loading, which I believe is almost always the case.

jonb377: I think @khatwanimohit tried to explicitly disallow higher-dim meshes (ie we only support all devices along a single axis): https://github.com/pytorch/xla/blob/365841da4b3497a832b54045eb0c3ac39c4a1e47/torch_xla/csrc/init_python_bindings.cpp#L1299-L1301

jonb377: > easier solution is we alwasy assume 1d mesh for the data loading, which I believe is almost always the case.

Yeah, IIRC that's also the approach taken in MaxText - load everything with pure batch-dim sharding, and let the compiler shuffle the data around to match the model sharding. Mohit is the expert here though!

JackCaoG: sg, will update

JackCaoG:
ok let me resolve the merge conflicts

JackCaoG:
@jonb377 can you take another look?

torch_xla/core/xla_model.py
```diff
@@ -1299,6 +1299,22 @@ def convert_fn(tensors):
...
+      # when minibatch is configured we must make sure batch dimension of
+      # the tensor is divisible by the local runtime device count.
+      for tensor, sharding in zip(tensors, shardings):
+        # assume batch dimension is 0
+        local_runtime_device_count = torch_xla.runtime.addressable_runtime_device_count(
+        )
+        if sharding and tensor.dim() > 0 and (tensor.size()[0] %
+                                              local_runtime_device_count) != 0:
+          raise RuntimeError(
+              "When minibatch is configured, batch dimension of the tensor " +
```
jonb377, line 1312: Maybe clarify `the per-host batch size must be divisible...`? These concepts are kind of confusing since we're mapping host-level sharding into device-level and representing it as a global tensor, so there's three different batch dimensions to consider.

jonb377 (approved):
Thanks Jack!

torch_xla/core/xla_model.py
JackCaoG: Thanks, let me update in a follow up pr.

PR #8229, opened 2024-10-07
[test] update XLA commit to JAX 0.4.33 same one

ManfeiBai:
update OpenXLA commit to JAX 0.4.33 used one: https://github.com/jax-ml/jax/blob/80e1c94de63e7f89667cdf35f38d8fe298e97a50/third_party/xla/workspace.bzl#L24C15-L24C55

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -50,7 +50,7 @@ new_local_repository(
 #    curl -L https://github.com/openxla/xla/archive/<git hash>.tar.gz | sha256sum
 #    and update the sha256 with the result.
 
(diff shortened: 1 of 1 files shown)


ManfeiBai:
seems failed with OpenXLA-commit update

lsy323:
The compilation failure are caused by tsl migration, should be easy to be fixed by including the correct header files.

However, since our internal test grid didn't complain the Sep 13 OpenXLA pin and Sep 16 Libtpu, I'm fine with proceed with the current OpenXLA Pin.

PR #8232, opened 2024-10-08
Add diagonal_scatter aten ops

nupurbaghel:
Fixes [#7381](https://github.com/pytorch/xla/issues/7381). Passes 2 additional tests for diagonal_scatter.

There was no direct equivalent in JAX for this op. Initially considered using [jnp.diag_indices](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.diag_indices.html#jax-numpy-diag-indices) for generating diagonal indices but it does not account for offset and assumes square arrays which was not the case for some tests.

Additionally, added some logic for moving `input` axes since `diagonal` function makes certain ordering assumptions and moving `src` axes was not very straightforward. Added more details in comments.

### DIFF
--- a/experimental/torch_xla2/test/test_ops.py
+++ b/experimental/torch_xla2/test/test_ops.py
@@ -24,7 +24,6 @@ skiplist = {
     "cholesky_solve",
     "complex",
     "diagonal_copy",
-    "diagonal_scatter",
     "digamma",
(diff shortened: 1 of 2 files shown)


ManfeiBai (approved):
Thanks, LGTM

PR #8242, opened 2024-10-09
Fix `test_arange_nan` after introduction of checks to `out=` functions.

ysiraichi:
This PR aligns with https://github.com/pytorch/pytorch/pull/137525, which runs meta functions for `out=` variants.

cc @JackCaoG


ysiraichi:
Apparently, there are more CI failures than just `test_arange_nan`. This is part of a bigger problem in PyTorch. This will probably be blocked until we solve that.

PR #8251, opened 2024-10-10
Op info test for  `linalg.solve_ex .. linalg.tensorinv` (#7504)

yenkwang:
Implement ops for `linalg.solve_ex .. linalg.tensorinv` (#7504)


ManfeiBai (approved):
Thanks, LGTM

Issue #7542, opened 2024-06-25
Op info test for `uniform .. var_mean`


qihqi:
## Fix the Op info test for `uniform .. var_mean`

1. Find the lines 333 to 337 of [test_ops.py](test/test_ops.py) and remove
   `uniform` .. `var_mean` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
current target should be: "unique_consecutive", "unique", "unravel_index", "var_mean"

ManfeiBai:
HI, @mrguenther, this will be the first issue to try, please feel free to reply so that we would assign it to you

mrguenther:
Sounds good; feel free to assign it to me.

mrguenther:
This can be closed when my pending PRs are merged.

PR #8258, opened 2024-10-14
Add `unique`, `unique_consecutive`

mrguenther:
Add implementations for `torch.unique` and `torch.unique_consecutive`.

Issue: #7542

### DIFF
--- a/experimental/torch_xla2/test/test_ops.py
+++ b/experimental/torch_xla2/test/test_ops.py
@@ -115,7 +115,6 @@ skiplist = {
     "svd_lowrank",
     "unfold_copy",
     "unfold",
-    "unique_consecutive",
     "unravel_index",
     "var_mean",
     "argwhere",
--- a/experimental/torch_xla2/test/test_ops.py
+++ b/experimental/torch_xla2/test/test_ops.py
@@ -116,7 +116,6 @@ skiplist = {
     "unfold_copy",
     "unfold",
     "unique_consecutive",
-    "unique",
     "unravel_index",
     "var_mean",
     "argwhere",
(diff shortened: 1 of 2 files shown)


ManfeiBai:
Thanks, looks like we have a real error for py3.9: 'FAILED test/test_ops.py::TestOpInfoCPU::test_reference_eager_unique_cpu_int64 - AssertionError: The values for attribute 'shape' do not match: torch.Size([5, 1]) != torch.Size([5]).'

Please let know if this take more than one hour to investigate too

mrguenther:
I'm pretty sure the Python 3.9 version isn't executing my implementation at all; I was able to reproduce the test failure in a Python 3.9 environment, but when I explicitly called the lowering for `torch.ops.aten.unique_consecutive`, I got the correct result.
```
>>> sys.version_info
sys.version_info(major=3, minor=9, micro=20, releaselevel='final', serial=0)
>>>
>>> torch.unique_consecutive(torch.zeros((5, 5), dtype=torch.float32), return_inverse=True, return_counts=False, dim=-2)
(tensor([[0., 0., 0., 0., 0.]]), tensor([0, 0, 0, 0, 0]))
>>>
>>> torch_xla2.ops.all_aten_jax_ops()[torch.ops.aten.unique_consecutive](jnp.zeros((5, 5), dtype=jnp.float32), return_inverse=True, return_counts=False, dim=-2)
(Array([[0., 0., 0., 0., 0.]], dtype=float32), Array([0, 0, 0, 0, 0], dtype=int32), Array([5], dtype=int32))
```
That would suggest it's an issue with the overload registration, but only in Python 3.9. Do you have an idea what might be wrong?

mrguenther:
Wait, never mind. It's `unique` that's failing, not `unique_consecutive`. That would certainly explain why I can't get my `unique_consecutive` implementation to produce the same output as the test failure. :P

mrguenther:
The Python 3.9 bug should be fixed now.

qihqi (approved):

Issue #7507, opened 2024-06-25
Op info test for `logspace .. masked.amax`


qihqi:
## Fix the Op info test for `logspace .. masked.amax`

1. Find the lines 153 to 157 of [test_ops.py](test/test_ops.py) and remove
   `logspace` .. `masked.amax` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
actual work will be "logspace", "lu", "lu_solve", "lu_unpack"

ManfeiBai:
Hi, @barney-s, thanks for contribution! This issue would be a new first issue, please feel free to reply this issue, so that we could assign it to you

barney-s:
Please assign it to me.

barney-s:
### Looking at `logspace`

`logspace` is decomposing to `linspace`.  We will fix `linspace` https://github.com/pytorch/xla/issues/7505 and revisit this again.

Script:
```python
import torch
import torch_xla2

env = torch_xla2.default_env()
env.config.debug_print_each_op = True
env.config.debug_accuracy_for_each_op = True

with env:
  y = torch.tensor([1, 5, 10])
  print(torch.logspace(start=-10, end=10, steps=5))
```

Output:

```
FUNCTION: tensor
/usr/local/google/home/barni/workspace/pytorch/xla/experimental/torch_xla2/torch_xla2/ops/jtorch.py:48: UserWarning: Explicitly requested dtype <class 'jax.numpy.int64'> requested in array is not available, and will be truncated to dtype int32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/google/jax#current-gotchas for more.
  return jnp.array(
FUNCTION: logspace
 DISPATCH: aten::logspace
  FUNCTION: linspace
   DISPATCH: aten::linspace
    FUNCTION: arange
     DISPATCH: aten::arange.start_step
    FUNCTION: lt
     DISPATCH: aten::lt.Scalar

```

barney-s:
[...]

PR #8262, opened 2024-10-15
Implement lu_unpack in jax

barney-s:
Implement lu_unpack in jax

- For Lower and Upper matrices, use jnp.tril, jnp.triu
- Shape both triangle matrices and add 1's to the lower triangle to
  match the pytorch behavior
- For 2D permutation matrix start with an identity matrix and mutate it
  based on pivots
  - start with sequential indices and apply the pivot operations to it
  - finally use the pivoted indices to index the identity matrix to
    generate the final permutation matrix for that pivot.
- For 2D inputs and 1D pivot the above logic would work
- For >=3d inputs, we first reshape the inputs to become 3D and then
  call vmap along the first dim with the 2d logic for each 2d matrix

Fixes: #7507

### DIFF
--- a/experimental/torch_xla2/torch_xla2/ops/jaten.py
+++ b/experimental/torch_xla2/torch_xla2/ops/jaten.py
@@ -4789,3 +4789,105 @@ def _euclidean_direct(x1, x2):
   dist = jnp.sqrt(dist_sq).astype(jnp.float32)
 
   return dist
+
+@op(torch.ops.aten.lu_unpack)
+def _aten_lu_unpack(LU_data, LU_pivots, unpack_data=True, unpack_pivots=True):
+  # lu_unpack doesnt exist in jax.
+  # Get commonly used data shape variables
+  n = LU_data.shape[-2]
+  m = LU_data.shape[-1]
+  dim = min(n,m)
+
+  ### Compute the Lower and Upper triangle
+  if unpack_data:
+    # Extract lower triangle
+    L = jnp.tril(LU_data, k=-1)
+
+    #emulate pytorch behavior: Add ones to the diagonal of L
+    eye = jnp.eye(n, m, dtype=LU_data.dtype)
+    L = L + eye
+
+    # emulate pytorch behavior: Reshape lower triangle to match pivot
+    start_indices = jnp.zeros(len(LU_data.shape), dtype=int)
+    limit_indices = list(LU_data.shape)
+    limit_indices[-1] = dim
+    L = jax.lax.slice(L, start_indices, limit_indices) 
+
+    # Extract upper triangle
+    U = jnp.triu(LU_data)
+
+    # emulate pytorch behavior: Reshape upper triangle to match pivot
+    start_indices = jnp.zeros(len(LU_data.shape), dtype=int)
+    limit_indices = list(LU_data.shape)
+    limit_indices[-2] = dim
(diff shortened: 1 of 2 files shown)


experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -4560,3 +4560,100 @@ def _euclidean_direct(x1, x2):
...
+          print("debug: pp_row:", j, pp_row)
+          pp = pp.at[j].set(pp_row)
+        print("debug: (i, row_idx):", pi)
+        print("debug: (row_idx, col_idx):", pp)
+        P = P.at[..., i, row_idx].set(pp)
+        print("debug: p1:", P)
+        P = P.at[..., col_idx, row_idx].set(pi)
+        print("debug: p2:", P)
+    else:
+      _pivots = LU_pivots - 1           # pivots are offset by 1 in jax
```
qihqi, line 4636: this branch looks more correct according to this explanation:
```
Pivot Matrix Representation

PyTorch uses a compact representation for the permutation matrix. Instead of storing the full matrix, it stores a 1-indexed vector called pivots.

Each element pivots[i] indicates the row that the i-th row was swapped with during the LU decomposition process.
Importantly, it uses 1-indexing, meaning pivots[i] = j signifies that the i-th row was swapped with the (j-1)-th row.
Your Example: [2, 2] -> [[0, 1], [1, 0]]

Let's analyze your example step-by-step:

Input pivots = [2, 2]: This vector tells us:

In the 1st step, the 1st row was swapped with the (2-1) = 1st row (essentially, no swap).
In the 2nd step, the 2nd row was swapped with the (2-1) = 1st row.
Constructing the Permutation Matrix:  We start with an identity matrix and apply the swaps indicated by the pivots vector:

Initial Identity Matrix:

[[1, 0],
 [0, 1]] 
Step 1 (no swap): The matrix remains unchanged.

Step 2 (swap rows 2 and 1):

[[0, 1],
 [1, 0]]
```

experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -4560,3 +4560,100 @@ def _euclidean_direct(x1, x2):
...
+      # tile the 2D identity matrix in other dims
+      tile_shape = list(LU_data.shape)
+      tile_shape[-1] = 1
+      tile_shape[-2] = 1
+      P = jnp.tile(jnp.identity(n, dtype=jnp.float32), tile_shape)
+      #print("debug: start permutation matrix:", P)
+
+      indices = [*range(n)]
+      for i in range(_pivots.size):
+        indices[i], indices[_pivots[i]] = indices[_pivots[i]], indices[i]
```
qihqi, line 4648: then, this line should be 

```
indices[i] = _pivots[i]
```

experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -4560,3 +4560,100 @@ def _euclidean_direct(x1, x2):
...
+        #print("[debug]: i, pivot[i], indices:", i, _pivots[i], indices)
+      P = P[jnp.array(indices)]
+      P = jnp.transpose(P)
```
qihqi, line 4651: why transpose here?

qihqi (approved):

PR #8275, opened 2024-10-17
Fix support for `torch.var_mean`.

mrguenther:
The default argument `correction=None` did not match the Torch API (default 1) and wasn't supported by the Jax API (where it's not nullable). Changing the default to 1 fixed the failing test case.

Issue: #7542


mrguenther:
To clarify, `correction=None` does match the internal function's default value, but it doesn't match the documented default behavior of `torch.var_mean`. I assume Torch replaces `None` with `1` by default.

mrguenther:
I added that `None` check per my above comment.

qihqi (approved):

Issue #7505, opened 2024-06-25
Op info test for `linalg.tensorsolve .. log_normal`


qihqi:
## Fix the Op info test for `linalg.tensorsolve .. log_normal`

1. Find the lines 143 to 147 of [test_ops.py](test/test_ops.py) and remove
   `linalg.tensorsolve` .. `log_normal` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

barney-s:
@ManfeiBai - Iam working on this

barney-s:
## trying linspace:

One of the failing test logs:
```
testcase = <test.test_ops.TestOpInfoCPU testMethod=test_reference_eager_linspace_tensor_overload_cpu_float32>, output1 = tensor([-2.]), output2 = tensor(-2.)
rtol = 1e-05, atol = 0.001, equal_nan = True, check_output = True


E         AssertionError: The values for attribute 'shape' do not match: torch.Size([]) != torch.Size([1]).
E         To execute this test, run the following from the base repo dir:
E              python test/test_ops.py -k TestOpInfoCPU.test_reference_eager_linspace_tensor_overload_cpu_float32
E         This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0

----------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------
[DEBUG] sample_input:  SampleInput(input=-2.0, args=(tensor(-3), 0), kwargs={'dtype': torch.int64, 'device': 'cpu'}, broadcasts_input=False, name='')
[DEBUG] sample_input:  SampleInput(input=-2.0, args=(-3, 0), kwargs={'dtype': torch.int64, 'device': 'cpu'}, broadcasts_input=False, name='')
[DEBUG] sample_input:  SampleInput(input=-2.0, args=(tensor(-3), 0), kwargs={'dtype': torch.int64, 'device': 'cpu'}, broadcasts_input=False, name='')
[...]

PR #8287, opened 2024-10-18
enable tensorsolve

barney-s:
Enable tensorsolve
- use jnp tensorsolve
- handle moving dims/axes and reshaping b before calling jnp variant

Fixes: #7505


ManfeiBai (approved):
Thanks, LGTM

PR #8309, opened 2024-10-24
Align rrelu_with_noise schema to reflect noise mutation

IvanKobzarev:
pytorch PR is https://github.com/pytorch/pytorch/pull/138503

rrelu_with_noise actually mutates noise, but this was not reflected in schema.
As a result compilation did not capture its mutation.

This PR is to align xla override for rrelu_with_noise to remove `const` for noise argument


IvanKobzarev:
@qihqi @miladm @JackCaoG Could you please take a look at this PR.

qihqi (approved):
this is meant to be land together with https://github.com/pytorch/pytorch/pull/138503
but because of them being 2 different repos, it has to be done one by one. 
The current plan is to land this one, land the torch one + xla pin update.

IvanKobzarev:
@qihqi Thanks for approving. This build error looks like caused from missing `const` (what we actually fixing).
Is there a way how to specify in xla PR to use some specific revision of pytorch or CI always takes pytorch HEAD?

PR #8328, opened 2024-10-27
Extend paged attention to support query_len>1

vanbasten23:
This PR extends the existing paged attention kernel to support query_len>1. Additionally, it upgrades the flash attention from v1 to v2.

Test plan:
- python pytorch/xla/test/test_pallas.py -v -k PallasTest.test_paged_attention_multi_queries_wrapper
- python pytorch/xla/test/test_tpu_paged_attention_kernel.py 2>&1 | tee out.txt

cc: @miladm

### DIFF
--- a/torch_xla/experimental/custom_kernel.py
+++ b/torch_xla/experimental/custom_kernel.py
@@ -482,13 +482,14 @@ def flash_attention(
   return FlashAttention.apply(q, k, v, causal, q_segment_ids, kv_segment_ids,
                               sm_scale, ab, partition_spec, mesh)
 
+
 def _multi_queries_paged_attention_nonkernel(
     q,  # [batch_size, query_len, num_heads, head_size]
     k_pages,  # [num_kv_heads, total_num_pages, page_size, head_size]
     v_pages,  # [num_kv_heads, total_num_pages, page_size, head_size]
     lengths,  # seq_lengths, [batch_size]. nb batch_size = len(seq_lens)
     page_indices,  # [batch_size, pages_per_sequence]
-) -> torch.Tensor: # [batch_size, query_len, num_heads, head_dim]
+) -> torch.Tensor:  # [batch_size, query_len, num_heads, head_dim]
   print('Running the nonkernel version of multi-queries paged attention.')
   batch_size, query_len, num_query_heads, head_size = q.shape
   num_kv_heads, total_num_pages, page_size, _ = k_pages.shape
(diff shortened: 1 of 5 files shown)


torch_xla/experimental/custom_kernel.py
```diff
@@ -483,6 +483,127 @@ def flash_attention(
...
+
+def multi_queries_paged_attention(
+    q,  # [batch_size, query_len, num_heads, head_size]
+    k_pages,  # [num_kv_heads, total_num_pages, page_size, head_size]
+    v_pages,  # [num_kv_heads, total_num_pages, page_size, head_size]
+    lengths,  # seq_lengths, [batch_size]. nb batch_size = len(seq_lens)
+    page_indices,  # [batch_size, pages_per_sequence]
+    num_kv_pages_per_compute_block,
+    num_queries_per_compute_block,
+    use_kernel=True,
```
vanbasten23, line 553: hey @WoosukKwon, this is the integration point between vLLM and torch_xla. I'm thinking if vLLM can switch this flag `use_kernel` perhaps by using some flags. I want to use the nonkernel version as a per baseline. Do you know if it possible?

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
```diff
@@ -0,0 +1,612 @@
...
+  l_scratch_ref[q_head_idx_per_kv] = l_next
+  m_scratch_ref[q_head_idx_per_kv] = m_next
+
+  l_next_inv_safe = jnp.where(l_next == 0.0, 1.0, 1.0 / l_next)
+  acc_scratch_ref[q_head_idx_per_kv] *= l_broadcast(l_corr * l_next_inv_safe)
+  # Note Matmul operandlhs must have a shape divisible by (16, 1)
+  o_curr = jax.lax.dot(p.astype(v.dtype), v, preferred_element_type=jnp.float32)
+  acc_scratch_ref[q_head_idx_per_kv] += o_curr * l_broadcast(l_next_inv_safe)
+
+  o_ref[0, q_head_idx_per_kv] = acc_scratch_ref[q_head_idx_per_kv].astype(
```
Liyang90, line 199: Could you add a `TODO` here for performance improvement later. Copying to final o, l, m refs can be done only at the last kv_block for each q block and head, using proper `@pl.when(kv_blk_idx == ...)`. Right now it copies at every kv_blk_idx.

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
```diff
@@ -0,0 +1,612 @@
...
+  s = jnp.einsum(
+      'hd,td->ht', q, k,
+      preferred_element_type=jnp.float32)  # [block_q, block_k]
+  assert s.shape == (num_queries_per_compute_block,
+                     kv_seq_len_per_kv_compute_blk)
+
+  q_index = q_blk_idx * num_queries_per_compute_block
+  kv_index = kv_blk_idx * kv_seq_len_per_kv_compute_blk
+  kv_len = lengths_ref[b]
+  row_ids = (kv_len - query_len) + q_index + jax.lax.broadcasted_iota(
```
Liyang90, line 151: Here, we assume the input query corresponds to the last (q_len) of the input kv. For example, if the input q_len is 8, and kv_len is 24, we assume the query corresponds to the kv at index [16. 24), and applies the causal mask accordingly.

@WoosukKwon please let us know if this assumption is valid or nor for the use cases in vLLM.

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
```diff
@@ -0,0 +1,612 @@
...
+      pl.program_id(3),
+  )
+  num_q_blks = pl.num_programs(2)
+  b_q_ref, num_q_heads_per_kv_head, num_queries_per_compute_block, head_dim = q_ref.shape
+  num_kv_heads, total_num_pages, page_size, head_dim = k_pages_hbm_ref.shape
+  compute_blk_size_kv = page_size * num_kv_pages_per_compute_block
+  kv_len = lengths_ref[b]
+
+  # Get the K and V for the current batch and current kv head.
+  @pl.when(kv_blk_idx * compute_blk_size_kv < kv_len)
```
Liyang90, line 254: Could you add a `TODO` here for performance improvement later.

Right now it does the work when the kv_blk is within the valid kv_len, but with q_len dim added in the q, and the causal mask, we could also skip the work when we know the causal mask would mask all (e.g. when the whole kv_blk is after the whole q_blk). This would also require change to the `compute_block_indices()` below.

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
```diff
@@ -0,0 +1,612 @@
...
+  # in-spec. Note q.shape=[batch_size, num_q_heads, query_len, head_dim]
+  # Map from grid idx.
+  print(f'xw32 line591 {num_q_heads_per_kv_head=}')
+
+  def qo_index_map(batch_index, kv_head_index, q_seq_blk_idx, *_):
+    return (batch_index, kv_head_index, q_seq_blk_idx, 0)
+
+  q_block_spec = pl.BlockSpec(
+      # q_ref.shape=[1, num_q_heads_per_kv_head, num_queries_per_compute_block, head_dim]
+      (1, num_q_heads_per_kv_head, num_queries_per_compute_block, head_dim),
```
Liyang90, line 503: Can this be `(None, num_q_heads_per_kv_head, num_queries_per_compute_block, head_dim)` to drop the first dim of size 1 inside the kernel?

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
```diff
@@ -0,0 +1,612 @@
...
+          in_specs=in_specs,
+          out_specs=out_specs,
+          grid=grid,
+          scratch_shapes=scratch_shapes,
+      ),
+      compiler_params=pltpu.TPUCompilerParams(
+          dimension_semantics=(
+              "parallel",
+              "parallel",
+              "parallel",
```
Liyang90, line 592: As the kv blk prefetching for next iteration is done inside the kernel according to the `compute_block_indices`, these grid dims cannot be `"parallel"`. That's why the original implementation has to manually split out a `"parallel"` grid for megacore.

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
vanbasten23: Right. It should improve the perf. Added a todo.

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
vanbasten23: Yeah, I think it can. I'm thinking if we should possibly split on the batch dimension later, similar to what the flash attention pallas kernel does today. If so, we'd need to add this batch dim back.

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
vanbasten23: Good call! Thanks.

torch_xla/experimental/custom_kernel.py
```diff
@@ -540,7 +663,8 @@ def paged_attention(q,
       ], payload, [q.shape, output_shape, output_shape],
       [q_dtype_for_kernel_launch, torch.float32, torch.float32])
 
-  return output.reshape(batch_size, num_heads, head_dim).to(q.dtype)
+  return output.reshape(batch_size, num_heads,
+                        head_dim).to(q_dtype_for_kernel_launch)
 
 
 def _calculate_num_tiles(x: int, tx: int) -> int:
```
vanbasten23, line 670: Let me also add the `XLA_LIB.define(` part.

vanbasten23: done

torch_xla/experimental/custom_kernel.py
vanbasten23: For dynamo, it's similar. The integration point is at `def multi_queries_paged_attention_xla(` in the same file.

torch_xla/experimental/pallas_kernels/multi_queries_paged_attention_kernel.py
WoosukKwon: Yes that's the desired behavior. Thanks for checking it out with me!

Liyang90 (approved):

PR #8333, opened 2024-10-29
custom_kernel: fix shape mismatch by sharding segment_ids in flash attn.

dudulightricks:
when adding the sharding support in this module, seqment_ids weren't take into count which causes a failure with shape mismatch when using them in sharded flash attention.


miladm (commented):
Thanks for the contribution! Looks like your use case runs into computational inaccuracies which may suggest `enable_manual_sharding` API calls need correction. I suggest adding a test case to further verify / debug the issue in a small example.

Here is a reference for [kernel tests](https://github.com/pytorch/xla/blob/master/test/test_pallas_spmd.py) you can refer to.

dudulightricks:
@miladm Hi! We have added a test that currently fails (16% of the values are correct, the others are not). I hope it will help you understand whats wrong.

miladm:
@dudulightricks Thanks for submitting the test code. We had a review of your code internally. It seems if you shard the KV and Q segment_id's the code won't attend the query to all kv elements in the matmul - hence the numerical inconsistency. Have you tried sharding the query segment_id only?

dudulightricks:
@miladm I just did and the test still fails, but why would something like this happen anyway? We are sharding the model and the data and expect consistency in the results in any sharding case. Can't we trust the result in any sharding case?

PR #8335, opened 2024-10-29
[Neuron] Fix the XLADevice Neuron mappings for SPMD downcasts

rpsilva-aws:
Currently, the downcasting mappings based on the XLA device types does not account for virtual devices (SPMD), when requiring Neuron specific casts. The same applies to the lowering and native functions. This PR fixes it by deriving the proper hardware device.

### DIFF
--- a/torch_xla/csrc/device.cpp
+++ b/torch_xla/csrc/device.cpp
@@ -116,4 +116,16 @@ bool CheckTpuDevice(XlaDeviceType hw_type) {
   return false;
 }
 
+bool CheckNeuronDevice(XlaDeviceType hw_type) {
+  if (hw_type == XlaDeviceType::NEURON) {
+    return true;
+  }
+
+  std::string pjrt_device = runtime::sys_util::GetEnvString("PJRT_DEVICE", "");
+  if (hw_type == XlaDeviceType::SPMD) {
+    return pjrt_device == "NEURON";
+  }
+  return false;
+}
+
 }  // namespace torch_xla
--- a/torch_xla/csrc/dtype.cpp
+++ b/torch_xla/csrc/dtype.cpp
@@ -129,7 +129,7 @@ xla::PrimitiveType MaybeDowncastToXlaDeviceType(
       if (UseBF16()) {
         return xla::PrimitiveType::BF16;
       }
-      if (DowncastBF16() || hw_type == XlaDeviceType::NEURON) {
+      if (DowncastBF16() || CheckNeuronDevice(hw_type)) {
         return xla::PrimitiveType::F32;
       }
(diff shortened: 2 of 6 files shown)


rpsilva-aws:
@aws-rhsoln @jeffhataws

torch_xla/csrc/device.cpp
```diff
@@ -116,4 +116,16 @@ bool CheckTpuDevice(XlaDeviceType hw_type) {
   return false;
 }
 
+bool CheckNeuronDevice(XlaDeviceType hw_type) {
+  if (hw_type == XlaDeviceType::NEURON) {
+    return true;
+  }
+
+  std::string pjrt_device = runtime::sys_util::GetEnvString("PJRT_DEVICE", "");
+  if (hw_type == XlaDeviceType::SPMD) {
```
aws-rhsoln, line 125: why check for SPMD, all neuron flow should have tht cast?

aws-nm9: So just check for pjrt_device == "NEURON"? This preserves the old check and additionally checks for this condition but yes we can only check for the pjrt_device and make that work.

jeffhataws (approved):

aws-nm9 (approved):

torch_xla/csrc/dtype.cpp
```diff
@@ -129,19 +129,19 @@ xla::PrimitiveType MaybeDowncastToXlaDeviceType(
...
         return xla::PrimitiveType::F32;
       }
       return xla::PrimitiveType::F64;
     case xla::PrimitiveType::F32:
       return UseBF16() || DowncastBF16() ? xla::PrimitiveType::BF16
                                          : xla::PrimitiveType::F32;
     case xla::PrimitiveType::U16:
-      return hw_type != XlaDeviceType::NEURON ? xla::PrimitiveType::U16
-                                              : xla::PrimitiveType::U32;
+      return CheckNeuronDevice(hw_type) ? xla::PrimitiveType::U32
```
aws-kingrj, line 140: Does the deprecating of the XLA casting env variables affect this at all?

aws-nm9: No, this change only affects the check for the device type, not the downcast flag check(or the deprecation of those flags). Since the pre-existing logic incorporated the deprecated flags, the current one does the same.

aws-kingrj: Thanks

aws-kingrj (approved):

qihqi (approved):

miladm:
cc @tengyifei to help bring this feature in the 2.6 release.

PR #8341, opened 2024-10-30
Fixing missing sharding_scheme param

slabhs:
### Change Summary:

- **Method Signature Updated**: The `sharding_scheme` parameter in `allgather_weights_and_update_full_parameter` is made optional by setting a default value of `None`.

  ```python
  # Original
  def allgather_weights_and_update_full_parameter(self, sharding_scheme):
      # method body

  # Change
  def allgather_weights_and_update_full_parameter(self, sharding_scheme=None):
      if sharding_scheme is None:
          sharding_scheme = self._get_sharding_scheme({})
      # method body
  ```

- **Issue Addressed**: This change fixes a bug where the `load_state_dict` method calls `allgather_weights_and_update_full_parameter` without providing a `sharding_scheme`, which would cause a `TypeError`.

  ```python
  # In both versions within load_state_dict
  self.allgather_weights_and_update_full_parameter()
  ```

### Impact:

- **Bug Fix**: By making `sharding_scheme` optional and handling the `None` case inside the method, the second version prevents the `TypeError` and ensures that `load_state_dict` functions correctly without requiring additional arguments.

### DIFF
--- a/torch_xla/distributed/zero_redundancy_optimizer.py
+++ b/torch_xla/distributed/zero_redundancy_optimizer.py
@@ -433,7 +433,7 @@ class ZeroRedundancyOptimizer(Optimizer):
 
     # All gather the new weights across the ranks and assign them to the full parameters
     if sharding_scheme is None:
-        sharding_scheme = self._get_sharding_scheme({})
+      sharding_scheme = self._get_sharding_scheme({})
     sharded_data = []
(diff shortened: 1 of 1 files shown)


jeffhataws (approved):

hgt312 (approved):
LGTM

PR #8294, opened 2024-10-21
DLPack: add test using PyTorch DLPack functions.

ysiraichi:
Taking over: #7213

This PR introduces a test using PyTorch's `to_dlpack` and `from_dlpack` functions. In order to use the former, PyTorch's tensor `__dlpack__` and `__dlpack_device__` methods had to be modified. The actual changes are in https://github.com/pytorch/pytorch/pull/138470.

PR #8344, opened 2024-10-30
Remove torch_pin that is accidentally introduced here:

qihqi:
remove torch_pin introduced here: https://github.com/pytorch/xla/pull/8294


ManfeiBai (approved):
Thanks, LGTM

PR #7213, opened 2024-06-06
Add a test to convert xla gpu tensor to cuda tensor by using the protocol.

vanbasten23:
Needs https://github.com/pytorch/pytorch/pull/128176

Test plan: PJRT_DEVICE=CUDA python pytorch/xla/test/test_operations.py -k test_dlpack_xla_to_pytorch_cuda_protocol_conversion

PR #8348, opened 2024-11-01
DLPack: fix test using PyTorch API.

ysiraichi:
This PR fixes the test introduced in #7213, where it called PyTorch's `to_dlpack()` function before actually calling `from_dlpack()`. Meaning that both `Tensor.__dlpack__` and `Tensor.__dlpack_device__` functions weren't getting called, [which were the functions that that test was supposed to be testing](https://github.com/pytorch/pytorch/pull/138470).

This should be merged after https://github.com/pytorch/pytorch/pull/138470.

cc @JackCaoG @vanbasten23 @qihqi

### DIFF
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2993,8 +2993,7 @@ class TestDLPack(parameterized.TestCase):
   @onlyIfPJRTDeviceIsCUDA
   def test_dlpack_xla_to_pytorch_cuda_protocol_conversion(self):
     xla_t1 = torch.arange(5).to(xm.xla_device())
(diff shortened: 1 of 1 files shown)


ysiraichi:
At first, I won't add the `.torch_pin` file, so as to make sure the test is failing without https://github.com/pytorch/pytorch/pull/138470. Then, I will add it, so as to make sure it is the reason the test passes.

JackCaoG:
cuda test was disabled on head due to some build issue.

ysiraichi:
Oh, I see. Is it ok if we merge this one? https://github.com/pytorch/pytorch/pull/138470 was already merged.

JackCaoG (approved):

Issue #7495, opened 2024-06-25
Op info test for `linalg.matrix_power`


qihqi:
## Fix the Op info test for `linalg.matrix_power`

1. Find the line 130 of [test_ops.py](test/test_ops.py) and remove
   `linalg.matrix_power` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
hi, @danielvegamyhre, this issue would be part of the second issues to try, pelase feel free to reply in this issue, so that we could assign it to you

simonteozw:
Hi @ManfeiBai was looking through and would be happy to try this issue!

PR #8352, opened 2024-11-03
Remove unnecessary functions and print statements (#7495)

simonteozw:
Discovered aten::eye was unnecessary when working on issue #7495


ManfeiBai (approved):
Thanks, why we notice its unnecessary?

ManfeiBai:
> Thanks, why we notice its unnecessary?

Thanks, saw the reason in chat: `basically matrix_power failed because aten_eye called arange an incorrect number of times.`

PR #8307, opened 2024-10-23
Add nightly C++11 ABI builds

tengyifei:
Initially, PyTorch/XLA was built with C++11 ABI, which was a more modern ABI (e.g. small string optimizations). However, that conflicts with the ABI of upstream PyTorch, so https://github.com/pytorch/xla/pull/5650 disabled the C++11 ABI. It turns out that C++11 ABI improves LazyTensor tracing performance drastically for some large models. We can't go back to C++11 ABI due to conflict with upstream, so this PR adds additional nightly build variants that enable C++11 ABI next to the existing wheels and dockers.

The docker images will be tagged like "nightly_3.10_tpuvm_cxx11_20241023" instead of "nightly_3.10_tpuvm_20241023".

The wheels will be named like
"torch_xla-2.6.0.dev20241023.cxx11-cp310-cp310-linux_x86_64.whl" instead of "torch_xla-2.6.0.dev20241023-cp310-cp310-linux_x86_64.whl".

This PR also adds support for building C++11 ABI variants for stable branches, but we don't activate it yet.

Fixes #8306.

PR #8357, opened 2024-11-04
Correct the cxx11_abi condition

tengyifei:
The cxx11_abi is a `string` variable that is either "0" or "1". Unfortunately, when specifying it in the `when:` ansible stanza, a string value of "0" may get interpreted as true. This causes our nightly wheels from Oct 28 to Nov 04 to be renamed with a "cxx11" suffix, regardless if they actually used C++11 ABI.


ManfeiBai (approved):
Thanks, LGTM

tengyifei:
c.f. older PR that introduced C++11 ABI builds: https://github.com/pytorch/xla/pull/8307/files

PR #8361, opened 2024-11-06
Expose lazy all number special scalars

rpsilva-aws:
Expose the getter for the explicit flag that treats all scalars as special - namely not routing the value to device data.


ManfeiBai (approved):
LGTM

PR #8363, opened 2024-11-07
Align rrelu_with_noise schema to reflect noise mutation [branch]

IvanKobzarev:
Creating instead of https://github.com/pytorch/xla/pull/8309

As fork branch can not be used in torch xla commit pin.

pytorch PR is https://github.com/pytorch/pytorch/pull/138503

rrelu_with_noise actually mutates noise, but this was not reflected in schema.
As a result compilation did not capture its mutation.

This PR is to align xla override for rrelu_with_noise to remove const for noise argument


qihqi (approved):

qihqi:
Merging later after the torch side PR merges first, then torch_pin file should be removed.

PR #8374, opened 2024-11-12
Remove unnecessary prints in PagedAttention

WoosukKwon:
Currently, the log is printed whenever the paged attention op is compiled (for every layer), which is not needed for end users.

@vanbasten23 Let me know if there's a better way to log this.


vanbasten23 (approved):

vanbasten23:
hey @WoosukKwon , could you point me to your vLLM and the new paged attention integration PR in vLLM?

PR #8347, opened 2024-10-31
XLA downcast for S64 and U64 for Neuron

aws-nm9:
This PR downcasts the S64 and U64 datatypes for Neuron devices, to prevent unsupported datatypes from being lowered to the compiler.

PR #8375, opened 2024-11-12
Revert S64 and U64 type change for Neuron device due to hardcoded dtype checks in torch-xla

aws-nm9:
An earlier [PR 8347](https://github.com/pytorch/xla/pull/8347) implemented downcasting for 64-bit ops for Neuron devices. However, due to few hardcoded dtype checks in torch-xla (in [tensor_ops](https://github.com/pytorch/xla/blob/5a4641b4434d3122085b34fe4d9e0a29aa85cb41/torch_xla/csrc/tensor_ops.cpp#L200) and [index_ops](https://github.com/pytorch/xla/blob/c4f2771849e0827badf934500d29cedc83d73b8d/torch_xla/csrc/ops/index_ops.cpp#L368)), this change needs to be reverted to unblock compilation failure. 

Since this change was only implemented for Neuron devices, it (or rather reverting it) would not have any impact on others.

### DIFF
--- a/torch_xla/csrc/dtype.cpp
+++ b/torch_xla/csrc/dtype.cpp
@@ -143,11 +143,9 @@ xla::PrimitiveType MaybeDowncastToXlaDeviceType(
       return CheckNeuronDevice(hw_type) ? xla::PrimitiveType::S32
                                         : xla::PrimitiveType::S16;
     case xla::PrimitiveType::S64:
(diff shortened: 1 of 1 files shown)


jeffhataws (approved):

JackCaoG (approved):

aws-nm9:
If the PR is approved, would it be possible to merge it soon so the previous change doesn't block anyone? Thank you!

PR #8411, opened 2024-11-23
Update XLA pin to 11/22

JackCaoG:
@tengyifei this will conflict with the libtpu change you had, but this one is to unblock a VLLM related experiment.


tengyifei:
that's alright -- seems a small conflict

tengyifei (approved):

PR #8416, opened 2024-11-26
Use regular `torch.Tensor` for CPU tensors

qihqi:
This is to test the water with encouraging `enable_globally()` usage pattern. So ideally torch_xla2 should not conflict with regular (non-compute) torch usage (such as dataloading). So we will make the type conversion explicit.

### DIFF
--- a/experimental/torch_xla2/test/llama/test_llama.py
+++ b/experimental/torch_xla2/test/llama/test_llama.py
@@ -12,101 +12,101 @@ from torch.utils import _pytree as pytree
 class LlamaTest(test_base.TestCase):
 
   def test_can_run(self):
-    sample_args = (
-        torch.randint(0, 32000, (1, 2048)),
-        torch.arange(0, 2048),
-    )
-    sample_args = pytree.tree_map(tensor.t2j, sample_args)
+    with torch_xla2.default_env():
+      sample_args = (
+          torch.randint(0, 32000, (1, 2048), device='jax:0'),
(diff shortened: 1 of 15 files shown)


experimental/torch_xla2/test/llama/test_llama.py
```diff
@@ -12,101 +12,101 @@
...
-        vocab_size=32000,
-        n_layers=2,
-        n_heads=4,
-        dim=256,
-    )
-    m = model_exportable.Transformer(model_args)
-    context_length = 2048
-    input_shape_prefill = (1, context_length)
-    input_shape_decode = (1, 1)
+      model_args = model_exportable.ModelArgs(
```
ManfeiBai, line 37: why we have indent here for this function? for `with torch_xla2.default_env():`?

experimental/torch_xla2/torch_xla2/ops/jaten.py
```diff
@@ -133,6 +133,7 @@ def _aten_copy(x, y, memory_format=None):
 
 
 @op(torch.ops.aten.clone)
+@op(torch.ops.aten.clone.default)
```
ManfeiBai, line 136: what is the `.default` change?

ManfeiBai (approved):
Thanks, LGTM

experimental/torch_xla2/torch_xla2/ops/jaten.py
qihqi: actually not needed, removed.

PR #8419, opened 2024-11-26
[scan] Add a test under SPMD

tengyifei:
Verifies that the GSPMD sharding annotation propagation pass can propagate through a While op and through the Body computation just fine.

### DIFF
--- a/test/scan/test_scan_spmd.py
+++ b/test/scan/test_scan_spmd.py
@@ -1,0 +1,51 @@
+import sys
+import unittest
+
+import torch
+import torch_xla
+from torch_xla.experimental.scan import scan
+from torch_xla.distributed.spmd import mark_sharding, set_global_mesh, get_1d_mesh
+import torch_xla.runtime as xr
+
+
+class ScanSpmdTest(unittest.TestCase):
+
+  def setUp(self):
+    # Activate SPMD
+    xr.use_spmd()
+
+    # Set up a simple SPMD mesh for these tests.
+    self.spmd_mesh = get_1d_mesh(axis_name="model")
+    set_global_mesh(self.spmd_mesh)
+    self.device = torch_xla.device()
+
+  @unittest.skipUnless(xr.global_runtime_device_count() >= 4,
+                       "Multiple devices required")
+  def test_scan_cumsum(self):
+    """This test uses `scan` to implement `torch.cumsum`."""
+
+    def fn(carry, x):
+      new_carry = carry + x
+      y = new_carry
+      return new_carry, y
+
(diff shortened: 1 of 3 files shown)


JackCaoG:
@yaochengji FYI

JackCaoG:
I will take a look today.

test/scan/test_scan_spmd.py
```diff
@@ -0,0 +1,86 @@
...
+import torch_xla.runtime as xr
+
+
+class ScanSpmdTest(unittest.TestCase):
+
+  def setUp(self):
+    # Set up a simple SPMD mesh for these tests.
+    num_devices = xr.global_runtime_device_count()
+    mesh_shape = (num_devices,)
+    self.spmd_mesh = Mesh(list(range(num_devices)), mesh_shape, ('model',))
```
JackCaoG, line 20: I think you can just do `xs.get_1d_mesh(axis_name="model")`

test/scan/test_scan_spmd.py
```diff
@@ -0,0 +1,86 @@
...
+    set_global_mesh(self.spmd_mesh)
+    xr.use_spmd()
```
JackCaoG, line 22: move this to the top of the setup, or even on the beginning on this file outside of test.

test/scan/test_scan_spmd.py
```diff
@@ -0,0 +1,86 @@
...
+    assert start != -1
+    start += len(begin_magic)
+    end = hlo_content.find(end_magic, start)
+    assert end != -1
+    output_sharding = hlo_content[start:end].strip().splitlines()
+
+    # There should be 4 tensors in output sharding: init, xs, final_carry, ys.
+    self.assertEqual(len(output_sharding), 4)
+    for sharding in output_sharding:
+      self.assertIn('devices=[', sharding)
```
JackCaoG, line 78: this is fine, but you could of just check the output sharding of the tensor. https://github.com/pytorch/xla/blob/master/torch_xla/csrc/init_python_bindings.cpp#L1280-L1283 is handy

JackCaoG (approved):

test/scan/test_scan_spmd.py
tengyifei: done

test/scan/test_scan_spmd.py
tengyifei: done

test/scan/test_scan_spmd.py
tengyifei: nice

PR #8431, opened 2024-12-03
Fix an edge case issue when storing output in the paged attention.

vanbasten23:
Fix an edge case for the condition to run `def store_to_output()` in the internal flash attention.

(1) If kv_len=512, kv_seq_len_per_kv_compute_blk=256, then last kv_blk_idx that we need to store_to_output is 1.
(2) If kv_len=513, kv_seq_len_per_kv_compute_blk=256, then last kv_blk_idx that we need to store_to_output is 2.

Without the fix, the case (1) would fail.

Test plan: `python pytorch/xla/test/test_tpu_paged_attention_kernel.py PagedAttentionKernelTest.test_paged_attention_store_to_output_correctly`


Liyang90 (approved):

vanbasten23:
Thanks for the review, Liyang!

PR #8448, opened 2024-12-04
Skip the computation if the current block lies above the causal mask diagonal.

vanbasten23:


Liyang90 (approved):

vanbasten23:
Thanks for the review!

PR #8450, opened 2024-12-04
Support S32/U32 indices for BWD embedding and index ops & Neuron implicit downcast

rpsilva-aws:
In this PR, we extend tensor operations to allow S32 indices. This follows suits with other operations, in order to add flexibility and potentially performance benefits for accelerator backends. Reference for embedding dense bwd: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/Embedding.cpp#L117

I re-structure a good portion of the tests to validate both types.

In addition, we also re-introduce the implicit downcasting for Neuron S64/U64 types, since the Neuron compiler does not support 64 bits.

### DIFF
--- a/.github/workflows/_tpu_ci.yml
+++ b/.github/workflows/_tpu_ci.yml
@@ -25,7 +25,7 @@ jobs:
           pip install rich
           # Jax nightly is needed for pallas tests.
           pip install torch_xla[pallas] -f https://storage.googleapis.com/jax-releases/jax_nightly_releases.html -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
-          pip install torch_xla[tpu] -f https://storage.googleapis.com/libtpu-releases/index.html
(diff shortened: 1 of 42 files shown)


miladm:
is this PR ready for review @rpsilva-aws?

rpsilva-aws:
@miladm Yes, but there seem to be some issues with the test infra - I need to validate with the TPU docker. I can ping once done/fixed.

tengyifei:
Can just ping me again when the "Build and test" has passed

rpsilva-aws:
Unfortunately, part of this PR has a dependency on https://github.com/pytorch/pytorch/pull/142160.

Instead, creating 2 PRs:
- https://github.com/pytorch/xla/pull/8462 (no dependency, partially Neuron-specific)
- https://github.com/pytorch/xla/pull/8463 (PT dependency)

PR #8453, opened 2024-12-04
[LoweringContext] Support an optimized parameter mapping for SPMD

rpsilva-aws:
Currently, the existing parameter mapping for the lowering context is not well suited for SPMD. In case of large models, it will cause a large synchronous bottleneck when transferring all device data to the host. This is caused by each ReplicateShardedData computation that gathers and reassembles each sharded data across multiple devices. This is by design, since it is expected to collect all parameters regardless of their allocation.

In this PR, we introduce a new mapping that does not invoke the sharded replication, but instead uses references to the device data. This is generally sufficient and preferred in most cases, where the user only wants to access the validate parameters (those that are not returned as `-1` from `tensor_parameter_id`, as 'fake' parameters).


rpsilva-aws:
Note: Conflicts with #8418. Will handle rebase order based on merge timing.

test/test_operations.py
```diff
@@ -2642,14 +2642,14 @@ def test_api(self):
...
     self.assertIn('MyCustomName', hlo_text)
-    self.assertIn('opcode: "parameter"', hlo_text)
-    self.assertIn('opcode: "parameter"', hlo_text)
+    self.assertTrue(hlo_text.count('opcode: "parameter"'), 2)
     self.assertIn('opcode: "add"', hlo_text)
     mapping = ctx.parameter_id_tensor_mapping()
-    self.assertEqual(len(mapping), 2)
+    num_params = len(mapping)
+    self.assertEqual(num_params, 2)
+    self.assertTrue(met.metric_data("TransferFromDeviceTime"))
```
tengyifei, line 2652: Should we clear `met` before checking metric data/running this test case?

tengyifei: Alternatively, check `TransferFromDeviceTime` is zero before `parameter_id_tensor_mapping`, and then checking it's larger after `parameter_id_tensor_mapping`.

test/test_operations.py
```diff
@@ -2642,14 +2642,14 @@ def test_api(self):
 
     ctx = torch_xla._XLAC.lowering.LoweringContext("MyCustomName")
     ctx.build([result])
-    hlo = ctx.hlo()
```
tengyifei, line 2645: I think it's still good to preserve this call to test it doesn't throw an exception. Maybe `_ = ctx.hlo()` to be more explicit.

test/run_tests.sh
```diff
@@ -247,6 +247,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
   run_test "$CDIR/spmd/test_mp_input_sharding.py"
+  run_test "$CDIR/spmd/test_spmd_lowering_context.py"
```
tengyifei, line 250: I learned that we also need to add this test to `test/tpu/run_tests.sh`. Otherwise it won't be exercised in TPU CI, which IIUC is the only CI that exercises SPMD.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1096,10 +1096,39 @@ class PyLoweringContext {
     return results;
   }
```
tengyifei, line 1098: Could you add another line in the `GetParameterIdTensorMapping` doc comments to the effect of "this will copy the tensor data from device to host"

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1096,10 +1096,39 @@ class PyLoweringContext {
...
 
+  // Returns a mapping from HLO parameter IDs to their corresponding
+  // device-backed Tensors. This version only returns parameters that were
+  // explicitly allocated on device data, accessible via GetTensorParameterId().
+  // Unlike GetParameterIdTensorMapping(), it avoids transferring data from
+  // device to host, making it more efficient especially for SPMD scenarios
+  // where data may be sharded.
```
tengyifei, line 1104: maybe "where transferring data involves costly collectives".

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1096,10 +1096,39 @@ class PyLoweringContext {
...
+    }
+    return param_to_tensor;
+  }
+
   // Get the parameter identifier of a given tensor. If the tensor is not a
   // parameter this will always return -1. This is useful in conjunction with
   // GetParameterIdTensorMapping to identify which values can be baked into
-  // the graph and which values must remain parameters.
+  // the graph and which values must remain parameters. Note that in
+  // conjunction with GetDeviceParameterIdTensorMapping, all tensors are
```
tengyifei, line 1130: Not sure I understand "in conjunction with GetDeviceParameterIdTensorMapping". Could you elaborate on the doc comments? Is it meant to say that all tensors returned by GetDeviceParameterIdTensorMapping will have a valid parameter ID?

test/test_operations.py
```diff
@@ -2642,14 +2642,14 @@ def test_api(self):
...
     self.assertIn('MyCustomName', hlo_text)
-    self.assertIn('opcode: "parameter"', hlo_text)
-    self.assertIn('opcode: "parameter"', hlo_text)
+    self.assertTrue(hlo_text.count('opcode: "parameter"'), 2)
     self.assertIn('opcode: "add"', hlo_text)
     mapping = ctx.parameter_id_tensor_mapping()
-    self.assertEqual(len(mapping), 2)
+    num_params = len(mapping)
+    self.assertEqual(num_params, 2)
+    self.assertTrue(met.metric_data("TransferFromDeviceTime"))
```
tengyifei, line 2653: Could we also test that `ctx.device_parameter_id_tensor_mapping()` returns 2 entries here?

tengyifei (changes requested):
Nice!

test/test_operations.py
rpsilva-aws: We have it in https://github.com/pytorch/xla/pull/8453/files#diff-b50b5ec9105ce166bf891d5d50080c03fef30396e6b63355ed69ae009064d3d6R22 already.

torch_xla/csrc/init_python_bindings.cpp
rpsilva-aws: Correct, but removed it. In general, there is no difference between the two mappings when it interplays with this function. In general, it is odd to specify a vague use case in the docstring, but just extended it to also reference the newer one.

test/run_tests.sh
rpsilva-aws: Nice, done.

test/test_operations.py
rpsilva-aws: I added to the older file, which you meant.

tengyifei (approved):

rpsilva-aws:
Replacing with https://github.com/pytorch/xla/pull/8460 to fix the merge commit, wrong local branch.

PR #8456, opened 2024-12-05
Stricter testing of sharding annotations after scan

tengyifei:
We would like to test that the output of a `torch_xla.experimental.scan` operation are sharded correctly if their inputs are sharded. In other words, the XLA compiler should be able to propagate the sharding annotation.

For `init` and `final_carry`, the sharding spec should look like `devices=[4]0,1,2,3`.

For `xs` and `ys`, the sharding spec should look like `devices=[1,4]0,1,2,3`.

### DIFF
--- a/test/scan/test_scan_spmd.py
+++ b/test/scan/test_scan_spmd.py
@@ -39,10 +39,19 @@ class ScanSpmdTest(unittest.TestCase):
     # Check the input and output sharding. Note that we do this after
     # `torch_xla.sync()` to ensure the output tensors are materialized and
     # have taken on sharding annotations propagated by the compiler.
-    for tensor in [init, xs, final_carry, ys]:
-      self.assertIn('ShardingSpec: {devices=[',
+    N = xr.global_runtime_device_count()
+    for tensor in [init, final_carry]:
+      self.assertIn(f'devices=[{N}]0,',
+                    torch_xla._XLAC._get_xla_sharding_spec(tensor))
+      self.assertIn('OpSharding: {'
+                    f'devices=[{N}]0,',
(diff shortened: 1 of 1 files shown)


vanbasten23 (approved):

rpsilva-aws (approved):

test/scan/test_scan_spmd.py
```diff
@@ -39,10 +39,21 @@ def fn(carry, x):
     # Check the input and output sharding. Note that we do this after
     # `torch_xla.sync()` to ensure the output tensors are materialized and
     # have taken on sharding annotations propagated by the compiler.
-    for tensor in [init, xs, final_carry, ys]:
-      self.assertIn('ShardingSpec: {devices=[',
+    N = xr.global_runtime_device_count()
+    for tensor in [init, final_carry]:
+      self.assertIn('ShardingSpec: {'
```
rpsilva-aws, line 44: nits: using `_get_xla_sharding_spec` and `_get_xla_op_sharding` instead of `_get_xla_tensor_debug_info`, easier to parse.

tengyifei: TIL

tengyifei: Turns out I can't do that because there's no Python bindings (e.g. comparator) exposed for xla::OpSharding at all.

tengyifei: At least replaced sharding spec parsing with _get_xla_sharding_spec

rpsilva-aws: Oh right, `_get_xla_sharding_spec` returns a string, and `_get_xla_op_sharding` the actual xla::OpSharding binding. The former is confusing...

rpsilva-aws (approved):

PR #8460, opened 2024-12-05
[LoweringContext] Support an optimized parameter mapping for SPMD

rpsilva-aws:
Currently, the existing parameter mapping for the lowering context is not well suited for SPMD. In case of large models, it will cause a large synchronous bottleneck when transferring all device data to the host. This is caused by each ReplicateShardedData computation that gathers and reassembles each sharded data across multiple devices. This is by design, since it is expected to collect all parameters regardless of their allocation.

In this PR, we introduce a new mapping that does not invoke the sharded replication, but instead uses references to the device data. This is generally sufficient and preferred in most cases, where the user only wants to access the validate parameters (those that are not returned as `-1` from `tensor_parameter_id`, as 'fake' parameters).


rpsilva-aws:
Re-opened from https://github.com/pytorch/xla/pull/8453, cleaned up the merge commit.

tengyifei (approved):

PR #8461, opened 2024-12-06
Update libtpu pin to 2024/12/05 / 0.0.5 stable

tengyifei:
The stable libtpu is qualified with more tests compared to nightly tpu. In particular, it has also been qualified with all JAX unit tests, which is beneficial since we depend on JAX.

This PR switches to stable libtpu at HEAD. Later I'll cut a 2.6 branch and that will also get the stable libtpu.

Depending on a stable libtpu means we don't need as much churn of updating the pin. If we later find a urgent need to adopt a specific libtpu CL, we could still switch back to a nightly at HEAD without issues.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,10 +64,11 @@ import build_util
 
 base_dir = os.path.dirname(os.path.abspath(__file__))
 
-_date = '20241122'
-_libtpu_version = f'0.0.5.dev{_date}'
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -64,10 +64,11 @@
 
 base_dir = os.path.dirname(os.path.abspath(__file__))
 
-_date = '20241122'
```
lsy323, line 67: nits: Maybe adding a flag for using the stable version, and add the `.dev{_date}` conditionally. This might make future pin update using nightly easier.

lsy323 (approved):

setup.py
tengyifei: oops, github auto-merge got too happy. i'll upload a fixup PR

PR #8467, opened 2024-12-06
Compute and hash buffer_donor_indices for step marker

### LINKED ISSUE #8464: IR graph hash should include `buffer_donor_indices` returned from `XLAGraphExecutor::SetBufferDonors()`
## 🐛 Bug

Currently `XLAGraphExecutor::SetBufferDonors()` is called from `XLAGraphExecutor::Compile()`, after IR graph hash `coll.hash` has been finalized. As a result, computations where buffer aliasing can be done and those where aliasing cannot be done are hashed to the same thing. This results in aliasing errors.

## To Reproduce

Small example:

```
import torch
import torch_xla

def main():
    device = xm.xla_device()
    t0 = torch.tensor([1], device=device)
    t1 = torch.tensor([2], device=device)
    xm.mark_step()

    t1.add_(t0)
    xm.mark_step()

    t2 = t1 + t0
    xm.mark_step()
    
    print(f"Results: t1={t1.item()}, t2={t2.item()}")

if __name__ == "__main__":
    main()
```

Here both mutating and non-mutating operation of `t1 + t0` result in the same IR hash, so final step marker where `t2` is computed causes `t2` to alias `t1`. Our PJRT device plugin has been recently updated to not allow donated buffers to be used. But a previous implementation prints:
```
Results: t1=4, t2=4
```

## Expected behavior

The expected output should be:
```
Results: t1=3, t2=4
```

## Environment

 - Reproducible on XLA backend [CPU/TPU/CUDA]: AWS trn1
 - torch_xla version: 2.5.1

mcuiaws:
Refactor SetBufferDonors() so that the logic for computing donor
indices in a separate function GetBufferDonors().
SyncTensorsGraphInternal() now calls GetBufferDonors(), and
those indices are hashed first, and the passed into Compile().

This fixes #8464.

### DIFF
--- a/test/test_input_output_aliases.py
+++ b/test/test_input_output_aliases.py
@@ -143,6 +143,25 @@ class InputOutputAliasesTest(unittest.TestCase):
         alias_count == 1.0
     ), f"Expect 1 input-output alias pair for gradient accumulation, got {alias_count}"
 
+  def test_separate_graphs(self):
+    """
+    Test that paramater aliasing differences should produce different graphs.
+    """
+    xla_device = xm.xla_device()
+    t0 = torch.tensor([1], device=xla_device)
+    t1 = torch.tensor([2], device=xla_device)
+    xm.mark_step()
+
+    t1.add_(t0)
+    xm.mark_step()
+
+    # This needs to be a separate graph, otherwise t1 can be corrupted
+    # or result in PJRT error.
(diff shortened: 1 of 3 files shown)


jeffhataws (approved):

tengyifei (approved):

tengyifei:
Thanks!

PR #8472, opened 2024-12-09
Update torch_xla wheels to have the suffix +cxx11 instead of .cxx11

bhavya01:
Currently while running 
```pip3 install https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-2.6.0.dev20241028.cxx11-cp310-cp310-linux_x86_64.whl```
 I get this error 
```ERROR: Invalid requirement: 'torch-xla==2.6.0.dev20241209.cxx11': Expected end or semicolon (after version specifier) torch-xla==2.6.0.dev20241209.cxx11```

Renaming the wheels to this format fixes it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,11 @@ pip3 install torch==2.6.0.dev20241028+cpu.cxx11.abi --index-url https://download
 pip3 install https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-2.6.0.dev20241028.cxx11-cp310-cp310-linux_x86_64.whl
 ```
 
+**As of 12/11/2024, the torch_xla C++11 ABI wheel is named differently and can be installed as follows:**
+```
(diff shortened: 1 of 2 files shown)


tengyifei:
I see. Could you update the README in https://github.com/pytorch/xla/blob/master/README.md#use-nightly-build-with-c11-abi-after-10282024 too? Maybe add a section that from Dec 9 onwards, we'll again use a different syntax for nightly installation.

tengyifei (changes requested):
I see. Could you update the README in https://github.com/pytorch/xla/blob/master/README.md#use-nightly-build-with-c11-abi-after-10282024 too? Maybe add a section that from Dec 9 onwards, we'll again use a different syntax for nightly installation.

tengyifei (approved):

ManfeiBai (approved):
Thanks, LGTM

PR #8476, opened 2024-12-10
Fix #8475: add libtpu-wheels registry when building docker

### LINKED ISSUE #8475: Nightly ptxla docker incorrectly contains torch_xla 2.5.1
Step to reproduce:

```
docker run -it us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.10_tpuvm pip freeze
```

Expected: output should contain `torch-xla @ some sha`

Actual: output contains `torch-xla==2.5.1`

This affects both pre-C++11 ABI and C++11 ABI builds.

tengyifei:
When building docker images, we install `torch_xla[tpu]` on there. If the old registry is used, then pip will only find an old version of libtpu and will then resolve to an older version torch_xla (2.5.1). I went over the repo and made sure every `libtpu-releases` registry is followed by `libtpu-wheels`.

Fixes #8475


ManfeiBai (approved):
Thanks, LGTM

PR #8478, opened 2024-12-10
[AdHoc] Test fix: clear metrics and simplify the replicated counter

rpsilva-aws:
In this PR, we remove the unnecessary metric checks in the lowering context SPMD test.


tengyifei (approved):

tengyifei:
Discussed offline, the metric failure was due to a race condition. Let's remove the metric check since this test is meant to validate the sharding spec.

tengyifei (changes requested):
Discussed offline, the metric failure was due to a race condition. Let's remove the metric check since this test is meant to validate the sharding spec.

tengyifei (approved):

lsy323 (approved):

PR #8480, opened 2024-12-10
Update libtpu/xla/jax pin

lsy323:


test/spmd/test_spmd_lowering_context.py
```diff
@@ -27,6 +27,7 @@ def _get_computation_hlo_txt(self, ctx):
     return xb.get_computation_hlo(comp)
```
tengyifei, line 27: fyi: https://github.com/pytorch/xla/pull/8478/files is merged so this change isn't needed to make TPU pass now

tengyifei (approved):

PR #8481, opened 2024-12-11
Update test_train_spmd_linear_model.py to show clear loss value

ManfeiBai:
linear model return one dimension value for each input will cause loss not clear, we notice all `0` in loss regression;

so increase return value dimension for used torch.nn.CrossEntropyLoss() to figure out

local passed log: https://gist.github.com/ManfeiBai/df6362c9a76890847127374e60fd6846

### DIFF
--- a/test/spmd/test_train_spmd_linear_model.py
+++ b/test/spmd/test_train_spmd_linear_model.py
@@ -121,6 +121,7 @@ def train():
         optimizer.step()
       xm.mark_step()
       if step % 10 == 0:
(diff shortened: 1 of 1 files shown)


tengyifei (changes requested):
Can we add an assertion on line 124 that checks loss is not 0?

ManfeiBai:
> Can we add an assertion on line 124 that checks loss is not 0?

Thanks, nice catch, updated with assertion

tengyifei (approved):

PR #8483, opened 2024-12-11
Torch functional tensors needs sync'ing before transferring to CPU.

### LINKED ISSUE #8482: `xm.save()` needs to synchronize functionalized tensors before calling `_xla_get_cpu_tensors()`
## 🐛 Bug

When xla functionalization is enabled, multiple live XlaTensors may be aliases of each other. A call to `SyncTensorsGraph()` can change the one one alias but not the other, breaking the aliasing relationship temporarily. The aliasing relationship is usually restored on any subsequent tensor operation from the torch side. However, `xm.save()` calls `_xla_get_cpu_tensors()` immediately after sync'ing the tensor graph, and incorrect values could be transferred.

## To Reproduce

Here's a small test case that shows wrong value being returned by `_xla_get_cpu_tensors()`:

```
import os

os.environ["XLA_DISABLE_FUNCTIONALIZATION"] = "0"
os.environ["XLA_ENABLE_PARAM_ALIASING"] = "0"

import torch
import torch_xla

xla_to_cpu = torch_xla._XLAC._xla_get_cpu_tensors

def main():
    device = xm.xla_device()
    t1 = torch.tensor([1], device=device)
    t2 = t1.detach()
    xm.mark_step()

    t2.add_(t2)
    xm.mark_step()

    # mark_step() causes t1 and t2 to be out of sync on the XLA side.
    # _functionalize_sync() is needed to get them back in sync.
    # torch._sync(t1)

    t1_cpu, t2_cpu = xla_to_cpu([t1, t2])
    assert t1.item() == t1_cpu.item(), f"{t1.item()} vs {t1_cpu.item()}"

if __name__ == "__main__":
    main()
```

In this example I explicitly disabled `XLA_ENABLE_PARAM_ALIASING` to show that the wrong value is being returned. In our PJRT implementation, donated PjRtBuffers are not allowed to be read, and a CHECK failure occurs in `PjRtComputationClient::TransferFromDevice()`

## Expected behavior

The program should not assert.

## Environment

 - torch_xla version: 2.5.1

mcuiaws:
This patch fixes #8482.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -252,6 +252,7 @@ function run_xla_op_tests3 {
   # NOTE: this line below is testing export and don't care about GPU
   PJRT_DEVICE=CPU CPU_NUM_DEVICES=1 run_coverage "$CDIR/test_core_aten_ops.py"
   run_test "$CDIR/test_pallas.py"
+  XLA_DISABLE_FUNCTIONALIZATION=0 run_test "$CDIR/test_functionalization.py"
 
   # CUDA tests
   if [ -x "$(command -v nvidia-smi)" ]; then
--- a/test/test_functionalization.py
+++ b/test/test_functionalization.py
@@ -1,0 +1,39 @@
+import io
+import os
+
+os.environ["XLA_DISABLE_FUNCTIONALIZATION"] = "0"
+os.environ["XLA_ENABLE_PARAM_ALIASING"] = "0"
+
+import torch
+import torch_xla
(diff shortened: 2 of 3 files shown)


jeffhataws (commented):
@mcuiaws will you add a test based on the testcase in https://github.com/pytorch/xla/issues/8482 ?

jeffhataws (approved):

mcuiaws:
Superseded by #8489

PR #8484, opened 2024-12-11
xm.save() should not set sync_xla_data=True when sync'ing.

### LINKED ISSUE #8422: `xm.save()` should not trigger buffer aliasing
## 🐛 Bug

When `xm.save()` is called on a subset of tensors, it calls down to `_xla_sync_multi(sync_xla_data=True)` which eventually calls down to `XLAGraphExecutor::SyncTensorsGraph()` with `sync_ltc_data=true` as if it's doing a step marker. This causes buffer aliasing to be performed by `XLAGraphExecutor::Compile()`. However, according to the detailed comments in [xla_graph_executor.cpp#L1336](https://github.com/pytorch/xla/blob/v2.5.1/torch_xla/csrc/xla_graph_executor.cpp#L1336), buffer aliasing should only be performed at a step marker when *all live tensors* are being sync'd. This is obviously not the case for `xm.save()`.

As a result, the `xm.mark_step()` following `xm.save()` will have parameters which refer to donated buffers which are deleted by PJRT, triggering [this](https://github.com/pytorch/xla/blob/v2.5.1/torch_xla/csrc/runtime/pjrt_computation_client.h#L192) `XLA_CHECK()` failure in `PjRtData::GetHandle()`.

## To Reproduce

A small example to reproduce.

```
import os
import torch
import torch_xla

def main():
    xla_device = xm.xla_device()
    t0 = torch.tensor([1], device=xla_device)
    t1 = torch.tensor([2], device=xla_device)
    xm.mark_step()

    t2 = t0 + t1
    t1.add_(1)

    # Save the new value of t1 should not result in the old value
    # being donated...
    xm.save(t1, os.devnull)

    # otherwise this mark_step could crash, or compute the wrong value for t2.
    xm.mark_step()
    assert t2.item() == 3

if __name__ == "__main__":
    main()
```

## Environment

 - torch_xla version: 2.5.1

mcuiaws:
Setting sync_xla_data=True performs tensor graph sync as if it's a mark step, which triggers buffer aliasing to be performed. However, according to the comments in https://github.com/pytorch/xla/blame/v2.5.1/torch_xla/csrc/xla_graph_executor.cpp#L1336, it's not safe to do so unless all live tensors are being sync'd.

This fixes #8422

### DIFF
--- a/test/test_input_output_aliases.py
+++ b/test/test_input_output_aliases.py
@@ -1,3 +1,4 @@
+import os
 import sys
 
 import torch
@@ -161,6 +162,28 @@ class InputOutputAliasesTest(unittest.TestCase):
     xm.mark_step()
 
     self.assertEqual(t1.item(), 3)
+
+  def test_xm_save_no_aliasing(self):
+    """
+    Test that xm.save() does not perform aliasing.
+    """
+    xla_device = xm.xla_device()
+    t0 = torch.tensor([1], device=xla_device)
+    t1 = torch.tensor([2], device=xla_device)
+    xm.mark_step()
+
+    t2 = t0 + t1
+    t1.add_(1)
+
+    # Save the new value of t1 should not result in the old value
+    # being donated...
+    xm.save(t1, os.devnull)
+
+    # otherwise this mark_step could crash, or compute the wrong value
+    # for t2.
+    xm.mark_step()
+
+    self.assertEqual(t2.item(), 3)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


jeffhataws (approved):

test/test_input_output_aliases.py
```diff
@@ -143,6 +144,28 @@ def try_grad_accum(model, device, train_x, train_label, accum_steps):
         alias_count == 1.0
     ), f"Expect 1 input-output alias pair for gradient accumulation, got {alias_count}"
 
+  def test_xm_save_no_aliasing(self):
```
tengyifei, line 147: I think this test is added in https://github.com/pytorch/xla/pull/8467/files already

tengyifei (approved):
Need to resolve conflicts before merging.

test/test_input_output_aliases.py
mcuiaws: rebased.

PR #6838, opened 2024-03-27
Make r2.3 to be compatible with PyPI

lsy323:
This applies https://github.com/pytorch/xla/pull/6145/ to master branch.

PR #3603, opened 2022-05-23
Pin upstream pytorch for pytorch/xla 1.12

JackCaoG:
This make sure that when ci is run for pytorch/xla r1.12 branch, it pick the right upstream.

PR #8488, opened 2024-12-12
More branch cut changes

tengyifei:
Pin PyTorch commit to 2.6 branch. Modeled after https://github.com/pytorch/xla/pull/3603

Makes the wheels compatible with PyPI.

Modeled after https://github.com/pytorch/xla/pull/6838/files. That PR also replaces tpuvm with tpu, but those changes look already applied in master.


ManfeiBai (approved):
Thanks, LGTM

Issue #8396, opened 2024-11-19
DDP doesn't synchronize weights 


darudimimi:
## 🐛 Bug

I am trying to run a basic example for XLA DDP on a CUDA backend. I have one node with two CUDA devices (RTX 4090). My code (presented below) is similar to https://pytorch.org/xla/master/multi_process_distributed.html. However, it seems that DDP does not synchronize the model weights across threads. To the best of my knowledge during DataParallel training, the model weights should be synchronized after each gradient step.

## To Reproduce
I used the prebuilt docker image according to https://pytorch.org/xla/master/gpu.html#docker.

My .bashrc in Docker:

```
export PATH=$PATH:/usr/local/cuda-12.1/bin
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-12.1/lib64
export GPU_NUM_DEVICES=2 PJRT_DEVICE=CUDA
```

Here is my code:

```
import os
import sys
import tempfile
import torch
import torch.distributed as dist
import torch.nn as nn
import torch.optim as optim

from torch.nn.parallel import DistributedDataParallel as DDP

import torch_xla
import torch_xla.core.xla_model as xm
import torch_xla.runtime as xr
import torch_xla.distributed.xla_backend
import torch_xla.distributed.xla_multiprocessing as xmp

import time
import copy

def setup(rank, world_size):
    os.environ['MASTER_ADDR'] = 'localhost'
    os.environ['MASTER_PORT'] = '12357'

    # initialize the xla process group
    dist.init_process_group("xla", rank=rank, world_size=world_size)
    # dist.init_process_group('xla', init_method='xla://')

def cleanup():
    dist.destroy_process_group()

class ToyModel(nn.Module):
    def __init__(self):
        super(ToyModel, self).__init__()
        self.net1 = nn.Linear(10, 10)
        self.relu = nn.ReLU()
        self.net2 = nn.Linear(10, 5)

    def forward(self, x):
        return self.net2(self.relu(self.net1(x)))

def demo_basic(rank):
    # xla specific APIs to get rank, world_size.
    new_rank = xr.global_ordinal()
    assert new_rank == rank
    world_size = xr.world_size()
    setup(rank, world_size)

    # create model and move it to XLA device
    device = xm.xla_device()
    print(f"Running basic DDP example on rank {rank}. world_size: {world_size} device: {device}")
    model = ToyModel().to(device)
[...]

Issue #8455, opened 2024-12-05
2.6 backport PR request list


tengyifei:
This is a tracker for backport/cherry-picks into 2.6. For any PRs you want to backport to 2.6, please reply with following:

- Original PR link (this PR should merge into `master`)
- Reason to backport
- 2.6 backport PR link (a separate PR should be created, and that PR should merge into `r2.6`)

This process is similar to the backport request thread in 2.5 release https://github.com/pytorch/xla/issues/7977

Please note the criterion for cherrypicking into 2.6: https://github.com/pytorch/xla/issues/7203

tengyifei:
@lsy323 libtpu pin update to 0.0.6: https://github.com/pytorch/xla/pull/8480.
Reason: pick stable libtpu for Trillium
Backport PR link: @tengyifei will upload one later

yaochengji:
Fix a DDP graph capture issue https://github.com/pytorch/xla/pull/8489
Reason: The DDP result before this patch is wrong
Backport PR link: https://github.com/pytorch/xla/pull/8500

mcuiaws:
Original PR: Compute and hash buffer_donor_indices for step marker #8467
Reason: Fixes tensor corruption issue
Backport PR: #8503

mcuiaws:
Original PR: xm.save() should not set sync_xla_data=True when sync'ing. #8484
Reason: Fixes tensor corruption issues, easily reproducible by running huggingface tutorials.
Backport PR: #8504

mcuiaws:
Original PR: Add xm.xla_device_kind() to return XLA device kind string. #8493
Reason: Requested by Neuron customers. Feature available in JAX but not PyTorch/XLA. Should be very low risk.
Backport PR: #8506

jeffhataws:
Original PR:  When modifying IR node, make sure to not lose the read_only bit https://github.com/pytorch/xla/pull/8505
Reason: Fixed a bug where 0-dimensional tensors result in aliasing errors
Backport PR: TBD

mcuiaws:
> Original PR: When modifying IR node, make sure to not lose the read_only bit #8505 Reason: Fixed a bug where 0-dimensional tensors result in aliasing errors Backport PR: TBD

Backport PR: #8508

avizon-aws:
Cherry-Pick PR for softmax autocast: https://github.com/pytorch/xla/pull/8511

savitha-aws:
[...]

PR #8489, opened 2024-12-12
Fix a DDP graph capture issue

yaochengji:
There're issues in torch/xla DDP. https://github.com/pytorch/xla/issues/8396. After investigation, one of the issues are caused by the incorrect graph capture caused by the combination of inplace-update and non aten op.

Changes:
1. Modify `TryGetXlaTensor` to make sure we can always have the most updated version of tensors
2. Modify the learning rate of ddp test from 1e-4 to 1e-1 as 1e-4 is too small to verify the DDP correctness
3. Remove the `gradient_as_bucket_view = True` requirement in all the docs and code
4. Add the test_inplace_update.py file, note that `test_non_aten_op_after_partial_update` cannot get passed before the `TryGetXlaTensor` patch


yaochengji:
cc @qihqi @tengyifei

jeffhataws:
@tengyifei  will you enable TPU testing on this PR?

tengyifei:
@yaochengji could you force push a commit? I just added a `tpuci` label but we need to retrigger the jobs to run TPU tests.

tengyifei (approved):
Thanks!

yaochengji:
@tengyifei thanks for your review. Do I still need to force push as it is already merged?

tengyifei:
oops, i merged too soon :) looks like tip-of-tree build is green, so nothing needed on your part now!

jeffhataws:
@yaochengji  will you add this change to https://github.com/pytorch/xla/issues/8455 for cherry-pick to 2.6? Let me know if I can help.

yaochengji:
@jeffhataws it's added, thanks.

PR #8491, opened 2024-12-13
Lower cummax op

### LINKED ISSUE #8230: Lower cummax
## 🚀 Feature
Lower cummax https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L1918-L1922

zyy-martin:
Implement cummax op. Fixes #8230

Cummax returns a tuple of (values, indices).  This PR uses the `ReduceWindowWithGeneralPadding` XLA op (padding + window reducer for cumulative values) to implement it similar to cumsum/cumprod. The only difference is that the indices will be passed to the `ReduceWindowWithGeneralPadding` op along with the values, and a new reducer is implemented to look at both the value and index of input and return a tuple of (max_value, max_index).


qihqi (approved):

zyy-martin:
I exclude the cummax from the python op test because I believe the scalar value generated from the sample input would fail, as will existing ops like cumprod and cumsum. The following script will fail for example:

```
import torch
import torch_xla

a = torch.tensor(7.358110427856445, device=xm.xla_device())
print(f'[xla]: {torch.cumsum(a, 0)}')
```

zyy-martin:
Thanks for the review! Can someone with write access merge the PR? Thanks!

PR #8501, opened 2024-12-18
Remove SPMD logging to indicate that we're using virtual device

bhavya01:
ShouldUseVirtualDevice is called pretty liberally throughout ptxla code and it creates a lot of log spam

### DIFF
--- a/torch_xla/csrc/device.cpp
+++ b/torch_xla/csrc/device.cpp
@@ -81,9 +81,6 @@ bool ShouldUseVirtualDevice() {
   bool use_virtual_device =
       runtime::sys_util::GetEnvBool("XLA_USE_SPMD", false) ||
       runtime::sys_util::GetEnvBool("XLA_AUTO_SPMD", false);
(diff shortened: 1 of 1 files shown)


torch_xla/csrc/device.cpp
```diff
@@ -81,9 +81,6 @@ bool ShouldUseVirtualDevice() {
   bool use_virtual_device =
       runtime::sys_util::GetEnvBool("XLA_USE_SPMD", false) ||
       runtime::sys_util::GetEnvBool("XLA_AUTO_SPMD", false);
-  if (use_virtual_device) {
-    TF_LOG(INFO) << "Using SPMD virtual device optimization";
-  }
   return use_virtual_device;
```
ManfeiBai, line 84: this change is reasonable for me, also wanna suggest to bring it back with cleaner log history version in the future due to its a good signal for user to notice the code has started running

ManfeiBai (approved):
Thanks, LGTM

Issue #8486, opened 2024-12-12
2 questions for the composite op feature


Zantares:
## ❓ Questions and Help
Glad to see that the [composite op feature](https://github.com/pytorch/xla/blob/master/docs/source/features/stablehlo.md#preserving-high-level-pytorch-operations-in-stablehlo-by-generating-stablehlocomposite) is added to Torch-XLA. I have tried this feature and got some questions, hope to get answers/suggestions here:
1. Some redundant IRs (start from `custom_call`) can't be erased after created the composite op, e.g. `Gelu`:
```python
import torch
import torch_xla
import torch_xla.core.xla_model as xm

from torch_xla import stablehlo
from torch_xla.experimental.mark_pattern_utils import StableHLOCompositeBuilder

class Example(torch.nn.Module):
    def __init__(self):
        super(Example, self).__init__()
        self.gelu = torch.nn.GELU(approximate="none")
        self.composite_op = StableHLOCompositeBuilder("composite.gelu", {"approximate": "none"})

    def forward(self, x):
        x = self.composite_op.mark_inputs(x)
        y = self.gelu(x)
        y = self.composite_op.mark_outputs(y)
        return y

x = torch.randn(10, device=xm.xla_device())
model = Example().to(xm.xla_device())
print(model(x))

input_args = (x, )
exported = torch.export.export(model, input_args)
# print(exported.graph)
stablehlo_gm = stablehlo.exported_program_to_stablehlo(exported)
stablehlo = stablehlo_gm.get_stablehlo_text()
print(stablehlo)
```
The generated StableHLO is:
```mlir
module @IrToHlo.16 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  func.func @main(%arg0: tensor<10xf32>) -> tensor<10xf32> {
    %cst = stablehlo.constant dense<0.707106769> : tensor<10xf32>
    %0 = stablehlo.multiply %arg0, %cst : tensor<10xf32>
    %1 = stablehlo.custom_call @mhlo.erf(%0) {mhlo.attributes = {}, mhlo.version = 1 : i64} : (tensor<10xf32>) -> tensor<10xf32>
    %2 = stablehlo.composite "composite.gelu" %arg0 {composite_attributes = {approximate = "none"}, decomposition = @composite.gelu.impl} : (tensor<10xf32>) -> tensor<10xf32>
[...]

PR #8502, opened 2024-12-18
Draft: Lowering Aten op to composite op instead of small ops

Zantares:
This PR is to solve the 2nd question in this issue: [supports composite op in training](https://github.com/pytorch/xla/issues/8486).

## Motivation
**Composite op** is beneficial for performance optimization and we aim to apply it to **training** too. . According to the response in the issue, the community has no plan to extend this to training currently... Thus, I created this draft PR to demonstrate our intention.

## Detail
This PR alters the Aten op lowering process when there isn't a 1:1 mapping to XLA op. It uses composite call instead of small XLA ops. Later, in the optimization process, the composite call can be easily replaced with a custom kernel or decomposed.

This is still a draft PR and only `Gelu` is implemented as an example. If it gets accepted, here are some further suggestions:
1. Keep both the decomposed ops and the composite call implementation. Use a new env setting (e.g. `XLA_COMPOSITE_OP`) to enable this feature. Also, add an op list setting to define which ops can be composed.
2. Only retain the composite call implementation as it can be easily decomposed by StableHLO pass. User can control the behavior by turning the decompose pass on or off.

## Example
```python
import torch

device = xm.xla_device()
gelu = torch.nn.GELU(approximate="none")

x = torch.tensor([2.0], requires_grad=True, device=device)
y = gelu(x ** 2)
y.backward()

print(x.grad)
```
With this PR, the generated StableHLO is:
```mlir
module @SyncTensorsGraph.43 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  func.func private @composite.gelu_backward.14(%arg0: tensor<1xf32>, %arg1: tensor<1xf32>) -> tensor<1xf32> {
    %cst = stablehlo.constant dense<0.398942292> : tensor<1xf32>
    %cst_0 = stablehlo.constant dense<-5.000000e-01> : tensor<1xf32>
    %cst_1 = stablehlo.constant dense<5.000000e-01> : tensor<1xf32>
    %cst_2 = stablehlo.constant dense<1.000000e+00> : tensor<1xf32>
    %cst_3 = stablehlo.constant dense<0.707106769> : tensor<1xf32>
    %0 = stablehlo.multiply %arg1, %cst_3 : tensor<1xf32>
    %1 = stablehlo.custom_call @mhlo.erf(%0) {mhlo.attributes = {}, mhlo.version = 1 : i64} : (tensor<1xf32>) -> tensor<1xf32>
    %2 = stablehlo.add %1, %cst_2 : tensor<1xf32>
    %3 = stablehlo.multiply %2, %cst_1 : tensor<1xf32>
    %4 = stablehlo.multiply %arg1, %arg1 : tensor<1xf32>
    %5 = stablehlo.multiply %4, %cst_0 : tensor<1xf32>
    %6 = stablehlo.exponential %5 : tensor<1xf32>
    %7 = stablehlo.multiply %arg1, %6 : tensor<1xf32>
    %8 = stablehlo.multiply %7, %cst : tensor<1xf32>
    %9 = stablehlo.add %3, %8 : tensor<1xf32>
    %10 = stablehlo.multiply %arg0, %9 : tensor<1xf32>
    return %10 : tensor<1xf32>
  }
  func.func @main(%arg0: tensor<f32>, %arg1: tensor<1xf32>) -> tensor<1xf32> {
    %cst = stablehlo.constant dense<1.000000e+00> : tensor<1xf32>
    %cst_0 = stablehlo.constant dense<2.000000e+00> : tensor<1xf32>
    %0 = stablehlo.power %arg1, %cst_0 : tensor<1xf32>
    %1 = stablehlo.composite "composite.gelu_backward" %cst, %0 {composite_attributes = {approximate = "none"}, decomposition = @composite.gelu_backward.14, version = 1 : i32} : (tensor<1xf32>, tensor<1xf32>) -> tensor<1xf32>
    %2 = stablehlo.power %arg1, %cst : tensor<1xf32>
    %3 = stablehlo.reshape %arg0 : (tensor<f32>) -> tensor<1xf32>
    %4 = stablehlo.multiply %2, %3 : tensor<1xf32>
    %5 = stablehlo.multiply %1, %4 : tensor<1xf32>
    return %5 : tensor<1xf32>
  }
}
```

### DIFF
--- a/torch_xla/csrc/ops/ops.cpp
+++ b/torch_xla/csrc/ops/ops.cpp
@@ -692,7 +692,22 @@ torch::lazy::NodePtr Gelu(const torch::lazy::Value& input) {
   auto lower_fn = [](const XlaNode& node,
                      LoweringContext* loctx) -> XlaOpVector {
     xla::XlaOp xla_input = loctx->GetOutputOp(node.operand(0));
-    return node.ReturnOp(BuildGelu(xla_input), loctx);
+
+    // Building composite computation.
+    const std::string name = "composite.gelu";
+    const std::string attr = "{approximate = \"none\"}";
+    xla::XlaBuilder builder(name);
+    xla::XlaOp arg = xla::Parameter(
+        &builder, 0, ShapeHelper::ShapeOfXlaOp(xla_input), "arg");
+    xla::XlaOp ret = BuildGelu(arg);
+    xla::XlaComputation computation = ConsumeValue(builder.Build(ret));
+
+    // Building call to computation.
+    std::vector<xla::XlaOp> inputs{xla_input};
+    xla::XlaOp output = xla::CompositeCall(loctx->builder(), computation, inputs, name,
+                                           attr, /*version=*/1);
+
+    return node.ReturnOp(output, loctx);
   };
   return GenericOp(torch::lazy::OpKind(at::aten::gelu), {input},
                    GetXlaShape(input), std::move(lower_fn));
@@ -704,7 +719,25 @@ torch::lazy::NodePtr GeluBackward(const torch::lazy::Value& grad_output,
                      LoweringContext* loctx) -> XlaOpVector {
     xla::XlaOp xla_grad_output = loctx->GetOutputOp(node.operand(0));
     xla::XlaOp xla_input = loctx->GetOutputOp(node.operand(1));
-    return node.ReturnOp(BuildGeluBackward(xla_grad_output, xla_input), loctx);
+
+    // Building composite computation.
+    const std::string name = "composite.gelu_backward";
+    const std::string attr = "{approximate = \"none\"}";
+    xla::XlaBuilder builder(name);
+    xla::XlaOp arg_grad_output =
+        xla::Parameter(&builder, 0, ShapeHelper::ShapeOfXlaOp(xla_grad_output),
+                       "arg_grad_output");
+    xla::XlaOp arg_input = xla::Parameter(
+        &builder, 1, ShapeHelper::ShapeOfXlaOp(xla_input), "arg_input");
+    xla::XlaOp ret = BuildGeluBackward(arg_grad_output, arg_input);
+    xla::XlaComputation computation = ConsumeValue(builder.Build(ret));
+
+    // Building call to computation.
+    std::vector<xla::XlaOp> inputs{xla_grad_output, xla_input};
+    xla::XlaOp output = xla::CompositeCall(loctx->builder(), computation, inputs, name,
+                                           attr, /*version=*/1);
+
+    return node.ReturnOp(output, loctx);
   };
   return GenericOp(torch::lazy::OpKind(at::aten::gelu_backward),
                    {grad_output, input}, GetXlaShape(input),


torch_xla/csrc/ops/ops.cpp
```diff
@@ -692,7 +692,22 @@ torch::lazy::NodePtr Gelu(const torch::lazy::Value& input) {
   auto lower_fn = [](const XlaNode& node,
                      LoweringContext* loctx) -> XlaOpVector {
     xla::XlaOp xla_input = loctx->GetOutputOp(node.operand(0));
-    return node.ReturnOp(BuildGelu(xla_input), loctx);
+
+    // Building composite computation.
+    const std::string name = "composite.gelu";
+    const std::string attr = "{approximate = \"none\"}";
```
lsy323, line 698: This is a dummy str for testing purpose?

torch_xla/csrc/ops/ops.cpp
```diff
@@ -692,7 +692,22 @@ torch::lazy::NodePtr Gelu(const torch::lazy::Value& input) {
...
+    xla::XlaBuilder builder(name);
+    xla::XlaOp arg = xla::Parameter(
+        &builder, 0, ShapeHelper::ShapeOfXlaOp(xla_input), "arg");
+    xla::XlaOp ret = BuildGelu(arg);
+    xla::XlaComputation computation = ConsumeValue(builder.Build(ret));
+
+    // Building call to computation.
+    std::vector<xla::XlaOp> inputs{xla_input};
+    xla::XlaOp output = xla::CompositeCall(loctx->builder(), computation, inputs, name,
+                                           attr, /*version=*/1);
```
lsy323, line 708: ditto for version

lsy323:
Hi @Zantares, thanks for the PR! As long as the normal compilation/execution flow is not affected, I think this actually makes the HLO/StableHLO cleaner.

Not sure if you have access to TPU to see if the code example in the PR description would run on TPU as well. I applied TPU CI tag, could you please push again to see if TPU CI pass?

lsy323:
cc @GleasonK in case you know if the composite HLO op affects complication flow

torch_xla/csrc/ops/ops.cpp
Zantares: This is a real op attribution for `GELU`: https://pytorch.org/docs/stable/generated/torch.nn.GELU.html#torch.nn.GELU

The available value of `approximate` is `none` or `tanh`. The lowering process checks this attribution and decides the sub lower function [here](https://github.com/pytorch/xla/blob/49de8bcb8fbe6af064f8a5a525ef13b9d55cf6c3/torch_xla/csrc/tensor_methods.cpp#L1706-L1712). As my changes are in the sub lower function, I manually set this attribution.

It's a common process for composite op which has **attributions** (defined as non-tensor inputs for composite op, e.g. `dim` for `Softmax`).

Zantares:
> Hi @Zantares, thanks for the PR! As long as the normal compilation/execution flow is not affected, I think this actually makes the HLO/StableHLO cleaner.
> 
> Not sure if you have access to TPU to see if the code example in the PR description would run on TPU as well. I applied TPU CI tag, could you please push again to see if TPU CI pass?

Thanks! I will fix the format error and push it again to trigger TPU CI.

torch_xla/csrc/ops/ops.cpp
Zantares: Yes this is for testing, I learned this setting from this [XLA UT](https://github.com/openxla/xla/blob/main/xla/hlo/builder/xla_builder_test.cc#L345-L348). I can remove it if it makes no scense.

torch_xla/csrc/ops/ops.cpp
Zantares: Maybe I can get the attribution from XlaOp instead of manually setting strings, I will try.

torch_xla/csrc/ops/ops.cpp
```diff
@@ -692,7 +692,22 @@ torch::lazy::NodePtr Gelu(const torch::lazy::Value& input) {
   auto lower_fn = [](const XlaNode& node,
                      LoweringContext* loctx) -> XlaOpVector {
     xla::XlaOp xla_input = loctx->GetOutputOp(node.operand(0));
-    return node.ReturnOp(BuildGelu(xla_input), loctx);
+
+    // Building composite computation.
+    const std::string name = "composite.gelu";
```
GleasonK, line 697: The purpose of namespacing is to be able to tell who the maintainer or origin of a given composite is. If something changes about GELU (new attr, etc) who is on the hook to maintain it (for composites this is usually intended to be a vendor who has a library `nvidia.some_op`, `aws.some_op`, `litert.some_op` etc). The name `composite` doesn't answer this question.

For this I'd recommend `ptxla.gelu` or `aten.gelu` as names.

GleasonK: As an aside, I'd be careful representing converting _every_ (or even many) aten operations using composites. They have a maintenance overhead in terms of needing to consider forward/backward compatibilty (i.e. the above thing about "what if gelu changes, who fixes?"), for some ops like gelu/softmax its _probably_ ok, they don't tend to change much and usually look somewhat uniform.

For other aten ops that have very specific HW support, I'd recommend an approach that decentralizes composite ownership/maintenance, i.e. FX graph rewrite-as-composite API or make composite builder work for these use cases.

torch_xla/csrc/ops/ops.cpp
GleasonK: What information is important for optimizations on this op? One option could be to have the composites implied value be `none`, and not generate a composite if `tanh` is used, of if the value is needed, then this looks good as-is, not a great API for making these in XLA, since they're an MLIR-first feature currently

torch_xla/csrc/ops/ops.cpp
GleasonK: (cc @lsy323)

Zantares: Thanks for the suggestions, here are some of my thoughts:
* The namespace `ptxla` sounds quite good, and I will address it.
* Regarding the compatibility/ownership of these ops in this PR, I believe it still belongs to`torch-xla`. The reason is that I didn't re-implement these ops from scratch, I simply wrapped the original implementation (such as `BuildGelu` and  `LowerSoftmax`) with a composite call. W/O this PR, those implementations would need to be fixed if there were any changes in the semantics. At the current stage, I don't have any plans to introduce new composite ops that don't have an original implementation in `torch-xla`
* Since this PR is aimed at resolving training issues, I'm uncertain whether the composite builder will work or not. Judging from the discussions in the [attached issues](https://github.com/pytorch/xla/issues/8486), it appears that it might not work for training purposes. Could you please share some examples or guidance on:
  >  FX graph rewrite-as-composite API or make composite builder work for these use cases.

torch_xla/csrc/ops/ops.cpp
Zantares: Actually, I believe that the `approximate` attribute here serves little purpose other than enhancing readability. Nevertheless, I've retained all the relevant information according to the public guidance [stablehlo.composite](https://github.com/pytorch/xla/blob/r2.6/docs/source/features/stablehlo.md#preserving-high-level-pytorch-operations-in-stablehlo-by-generating-stablehlocomposite). If no more further comments or feedback, I can remove this information.

torch_xla/csrc/ops/ops.cpp
Zantares: Namespace is addressed.

torch_xla/csrc/ops/ops.cpp
Zantares: `version` is removed.

Zantares:
> Hello, @Zantares, sorry for the delay, we are deprecating the current workflow of generating StableHLO from PyTorch programs. In #8544 we have the updated document to export PyTorch program using torch_xla2.
> 
> I would encourage you to explore along the new workflow, however I have no problem landing this change since there is no side effect.

Thanks. I'll go ahead and check out the new path. Before that, I have a quick question here: I have followed the official docs <[OP Lowering Guide](https://pytorch.org/xla/master/contribute/op_lowering.html#op-lowering-guide)> and <[Codegen migration Guide](https://pytorch.org/xla/master/contribute/codegen_migration.html#codegen-migration-guide)> to understand the lowering process, will the new `torch_xla2` change the entire process? If so, hoping to see similar Guide Docs for the new process.

torch_xla/csrc/ops/ops.cpp
Zantares: `approximate` for `Gelu` and `GeluBackward` is removed.

lsy323:
> Zantares

Hi @Zantares, here it the guide https://github.com/pytorch/xla/tree/master/experimental/torch_xla2

Zantares:
Close this PR as there is new solution for this propasal.

PR #8505, opened 2024-12-18
When modifying IR node, make sure to not lose the read_only bit

### LINKED ISSUE #8499: 0-dimensional tensors result in aliasing errors
## 🐛 Bug

When zero-dimenional tensors are created, the XLA device_data comes out of a cache is supposed to be read_only. However, across `mark_step()` the read_only bit is dropped. Subsequently, buffer aliasing can corrupt the cached value.

## To Reproduce
Small test case:
```
import torch

def main():
    """
    Test that device data in DataCache are not aliased.
    """
    xla_device = xm.xla_device()

    t0 = torch.tensor(42, device=xla_device)
    # drops the read-only bit on t0's device_data
    xm.mark_step()

    # cached value of 42 is corrupted
    t0.add_(1)
    xm.mark_step()

    # t1 get the cached device_data, which is corrupted.
    t1 = torch.tensor(42, device=xla_device)
    xm.mark_step()

    t1.add_(1)
    # XLA crashes here because parameter is donated buffer...
    xm.mark_step()

    # If it doesn't crash, the value here would be 44.
    assert t1.item() == 43


if __name__ == '__main__':
    main()
```

## Environment

 - torch_xla version: 2.5.1

mcuiaws:
This fixes #8499

### DIFF
--- a/test/test_input_output_aliases.py
+++ b/test/test_input_output_aliases.py
@@ -185,6 +185,31 @@ class InputOutputAliasesTest(unittest.TestCase):
 
     self.assertEqual(t2.item(), 3)
 
+  def test_device_data_cache_no_aliasing(self):
+    """
+    Test that device data in DataCache are not aliased.
+    """
+    xla_device = xm.xla_device()
+
+    t0 = torch.tensor(42, device=xla_device)
+    # drops the read-only bit on t0's device_data
+    xm.mark_step()
+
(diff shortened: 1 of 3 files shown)


tengyifei:
@mcuiaws could you force-push a commit to test on TPUs

jeffhataws (approved):

PR #8509, opened 2024-12-20
Added softmax to autocast policy

avizon-aws:
Softmax operation was not added to the autocast policy previously which led to its computation in BF16. Since this is a non linear operation, we want to preserve precision and perform it in FP32. This PR adds its support as well as a HLO based unit test.


lsy323:
Hi @avizon-aws, thank you for sending the PR! Could you force push again to trigger TPU CI?

lsy323 (approved):

lsy323:
@avizon-aws LGTM, thanks!

PR #8513, opened 2024-12-21
Prepare to make 2.6.0 rc3 wheel

tengyifei:
PyTorch upstream has tagged rc3: https://github.com/pytorch/pytorch/tree/v2.6.0-rc3

Update CI to pull rc3.

After this commit, I'll add an rc3 tag and run tests.


ManfeiBai (approved):
Thanks, LGTM

PR #8524, opened 2025-01-02
[scan] use device_parameter_id_tensor_mapping

tengyifei:
Previously scan uses `parameter_id_tensor_mapping` to fetch tensors hoisted to HLO parameters e.g. the fn being scanned may create additional tensors while its running. `parameter_id_tensor_mapping` will fetch those tensors back to host as XLA literals and create new tensors wrapping those, resulting in additional host RAM usage.

PR #8460 added `device_parameter_id_tensor_mapping` that returns the actual device backed tensors instead of another copy. So we'll use that and test that this avoids host transfers.


rpsilva-aws (approved):
Nice, lgtm.

bhavya01 (approved):

PR #8526, opened 2025-01-03
Add back the implementation of adaptive_pool_2d

qihqi:


tengyifei:
I'm not too familiar so wondering is this covered by a test?

tengyifei (approved):

tengyifei:
Chatted offline, tested in https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/test/test_core_aten_ops.py#L204 and this PR overrides a decomposition in https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py#L2401

PR #8536, opened 2025-01-06
pin update

lsy323:


lsy323:
> Waiting for the CI to pass. Locally the pallas test failed for me. I am just investigating that now.

Hi @bhavya01 fyi the TPU CI passed on CI. There is a cpu failing test

bhavya01 (approved):
Thanks! I see the pallas issue on v6e. Will debug separately and not block the pin update

PR #8552, opened 2025-01-10
Always run TPU tests on new PRs

tengyifei:
so far, we only run TPU tests on PRs with the `tpuci` label

the problem is this label is easily forgotten and also sometimes it doesn't work for some reason


vanbasten23 (approved):

vanbasten23:
I vaguely remember the reason why it checks the label is due to the limited TPU resources. It's something to look out for in case it happens again. The current tpu ci (before your change) runs for each PRs after they're merged.

tengyifei:
I'll work on expanding the TPU resource

PR #8558, opened 2025-01-13
Set gradient_as_bucket_view=True in test_train_mp_imagenet

tengyifei:
This example is tested in CI on TPUv3, and OOMs unless we train with this memory saving option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ If you're using `DistributedDataParallel`, make the following changes:
 +  dist.init_process_group("xla", init_method='xla://')
 +
 +  model.to(xm.xla_device())
-+  ddp_model = DDP(model)
++  ddp_model = DDP(model, gradient_as_bucket_view=True)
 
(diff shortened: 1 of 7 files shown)


yaochengji:
Do you think the OOM is due to memory leak?

tengyifei:
> Do you think the OOM is due to memory leak?

I'm not sure. I was able to train the model on v3 with batch size = 16 for the same number of epochs given `gradient_as_bucket_view=False`, so perhaps it's not a memory leak, but it's hard to be sure without taking a long profile.

Separately this test failure blocks 2.6 release CI green-ness so I think it's useful to merge it.

yaochengji (approved):

tengyifei:
> mention setting gradient_as_bucket_view to True in all places using DDP, including Docs and code

Done

mikegre-google (approved):

yaochengji (approved):

PR #8561, opened 2025-01-13
Improve and refine MLP tests for extensibility and A/B testing

rpsilva-aws:
In this PR, we include various fixes, improvements and extensions, namely:
- Exposing the MLP test to other tests (to allow us to A/B test convergences: requirement for the grad acc tests)
- Improving the asserts, and extending to losses and outputs
- Using the appropriate flag for the LR and log steps
- Improving the model layout with nn.Sequential
- Enhance the coverage, by actually utilizing the checkpointing flag, and including a sanity test for CPU
- Decouple to simplify the A/B coverage
- Fix imports

### DIFF
--- a/test/neuron/run_tests.sh
+++ b/test/neuron/run_tests.sh
@@ -214,6 +214,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   #run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
   run_test "$CDIR/spmd/test_train_spmd_linear_model.py"
+  run_test "$CDIR/spmd/test_train_spmd_linear_model.py" "$@" --use_gradient_checkpointing
   run_test "$CDIR/spmd/test_xla_spmd_python_api_interaction.py"
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   run_test "$CDIR/spmd/test_fsdp_v2.py"
--- a/test/neuron/run_tests.sh
+++ b/test/neuron/run_tests.sh
@@ -214,7 +214,6 @@ function run_xla_op_tests3 {
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   #run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
   run_test "$CDIR/spmd/test_train_spmd_linear_model.py"
-  run_test "$CDIR/spmd/test_train_spmd_linear_model.py" "$@" --use_gradient_checkpointing
   run_test "$CDIR/spmd/test_xla_spmd_python_api_interaction.py"
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   run_test "$CDIR/spmd/test_fsdp_v2.py"
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -231,7 +231,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
   run_test "$CDIR/spmd/test_mp_input_sharding.py"
-  run_test "$CDIR/spmd/test_train_spmd_linear_model.py"
+  run_test "$CDIR/spmd/test_train_spmd_linear_model.py" "$@" --skip-gradient-checkpointing
   run_save_tensor_hlo run_test "$CDIR/spmd/test_spmd_lowering_context.py"
   run_test "$CDIR/test_operations_hlo.py" "$@" --verbosity=$VERBOSITY
   run_test "$CDIR/test_input_output_aliases.py"
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -231,6 +231,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
   run_test "$CDIR/spmd/test_mp_input_sharding.py"
+  run_test "$CDIR/spmd/test_train_spmd_linear_model.py"
   run_save_tensor_hlo run_test "$CDIR/spmd/test_spmd_lowering_context.py"
   run_test "$CDIR/test_operations_hlo.py" "$@" --verbosity=$VERBOSITY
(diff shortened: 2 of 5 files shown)


rpsilva-aws:
@tengyifei @ManfeiBai I found myself having to largely improve/enhance the MLP tests, since I wanted to reuse this test for A/B convergence validation with/without gradient accumulation (using experimental grad acc loops) and with the same batch sizes. PTAL.

test/neuron/run_tests.sh
```diff
@@ -213,7 +213,7 @@ function run_xla_op_tests3 {
   #run_test "$CDIR/spmd/test_dtensor_integration2.py"
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   #run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
-  run_test "$CDIR/spmd/test_train_spmd_linear_model.py"
```
tengyifei, line 216: Did you intend to remove this test?

test/run_tests.sh
```diff
@@ -231,6 +231,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
   run_test "$CDIR/spmd/test_mp_input_sharding.py"
+  run_test "$CDIR/spmd/test_train_spmd_linear_model.py"
```
tengyifei, line 234: Should we have `test_train_spmd_linear_model_grad_checkpointing` here?

Should we have `test_train_spmd_linear_model_grad_checkpointing` in TPU tests as well?

tengyifei (approved):

test/neuron/run_tests.sh
rpsilva-aws: Yes, it's basically a subset of the checkpoint one. We cross check their convergence, while ensuring that the losses/output is not 0. We only have `test_train_spmd_linear_model` for CPU as standalone, since it doesn't run with grad checkpointing there.

test/run_tests.sh
rpsilva-aws: See above. CPU is the only case we use `test_train_spmd_linear_model` instead of `test_train_spmd_linear_model_grad_checkpointing`. I didn't see much value in having `test_train_spmd_linear_model` as standalone in addition to the checkpoint one. Perhaps I could rephrase the test file name, since it runs baseline as part of its A/B assertions.

tengyifei (changes requested):
Synced offline. Let's add a `--skip-gradient-checkpointing` CLI arg to the train testing script or similar to skip the gradient checkpointing on CPU, in order to avoid the confusing `test_train_spmd_linear_model_grad_checkpointing` name.

rpsilva-aws:
> Synced offline. Let's add a `--skip-gradient-checkpointing` CLI arg to the train testing script or similar to skip the gradient checkpointing on CPU, in order to avoid the confusing `test_train_spmd_linear_model_grad_checkpointing` name.

I was just about to write this, done!

test/run_tests.sh
```diff
@@ -231,6 +231,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/spmd/test_xla_auto_sharding.py"
   run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
   run_test "$CDIR/spmd/test_mp_input_sharding.py"
+  run_test "$CDIR/spmd/test_train_spmd_linear_model.py" "$@" --skip-gradient-checkpointing
```
tengyifei, line 234: Could you add this test to `test/tpu/run_tests.sh` (and i presume neuron) too?

rpsilva-aws: It was already on those, and now neuron and TPU will run with gradient checkpointing as an additional coverage.

rpsilva-aws: - https://github.com/pytorch/xla/blob/master/test/tpu/run_tests.sh#L17
- https://github.com/pytorch/xla/blob/master/test/neuron/run_tests.sh#L217

tengyifei: ack, thanks

tengyifei (approved):

test/utils/train_spmd_linear_model.py
```diff
@@ -0,0 +1,152 @@
+import sys
```
lsy323, line 1: qq is this just a duplicated test file?

lsy323 (changes requested):
Looks like the test file is duplicated under `test/utils`?

test/utils/train_spmd_linear_model.py
rpsilva-aws: This was mostly the former `test_train_spmd_linear_model.py`, minus all the different improvements (you can see the different commits). There is no duplication in the code with this PR.

rpsilva-aws:
> Looks like the test file is duplicated under `test/utils`?

I don't see it. The core of the functionality was moved to utils - it's a standalone training script. The `test_` file can run simultaneous executions of it, to, intentionally, run A/B testing.

lsy323 (approved):

lsy323:
> > Looks like the test file is duplicated under `test/utils`?
> 
> I don't see it. The core of the functionality was moved to utils - it's a standalone training script. The `test_` file can run simultaneous executions of it, to, intentionally, run A/B testing. The test/utils file is not for direct testing, but users can run it as a standalone script, and so can tests.

Hi @rpsilva-aws thank you for the explanation. I still don't get the A/B testing part. But if the body can be reused in other tests then it's fine

rpsilva-aws:
> > > Looks like the test file is duplicated under `test/utils`?
> > 
> > 
> > I don't see it. The core of the functionality was moved to utils - it's a standalone training script. The `test_` file can run simultaneous executions of it, to, intentionally, run A/B testing. The test/utils file is not for direct testing, but users can run it as a standalone script, and so can tests.
> 
> Hi @rpsilva-aws thank you for the explanation. I still don't get the A/B testing part. But if the body can be reused in other tests then it's fine

Thanks. You can see it in this PR here: https://github.com/pytorch/xla/pull/8561/files#diff-09c6a280d1c6fc8053d5a17e919e29fe51a75fd593f6c2012496b6e970312c25R44

Essentially, it allows it to A/B test different functionality. It adds a bit more value than running each standalone test, and just checking that the losses/output are not 0. Alternatively, we'd need to assert against an hardcoded set of expected values. We want to reuse this for testing gradient accumulation with and without XLA's while loop.

PR #8562, opened 2025-01-14
[scan] Reduce memory usage

tengyifei:
This change optimizes the memory usage of scan by turning some copies into aliases. Specifically, it optimizes intermediate activations which are aliases of inputs. This is a common occurrence.

In principle, we could have `forward` return all the intermediate activations, including those that are aliases to an input tensor. However, those inputs will then be duplicated as part of the output of a `scan` call, because we want to save all activations during the forward pass of a `scan`. The XLA compiler can't optimize away this duplication likely because they're behind a DynamicSlice + DynamicUpdateSlice, so we end up doubling the memory usage from those inputs.

To reduce memory usage, we can have `forward` return the activations that don't alias to inputs, called `partial_activations`. The autograd implementation of `scan` will call `alias_input` to add back activations that are aliases of input tensors outside of a scan, turning the partial activations back to full activations.

### DIFF
--- a/torch_xla/experimental/scan.py
+++ b/torch_xla/experimental/scan.py
@@ -46,6 +46,7 @@ from functorch.compile import aot_function, make_boxed_func, default_partition  
 import torch_xla
 import torch_xla.core.xla_builder as xb
 from torch_xla.experimental.pytreeify import pytreeify
+import torch_xla.debug.profiler as xp
 
 Carry = TypeVar('Carry')
 X = TypeVar('X')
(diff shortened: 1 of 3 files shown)


torch_xla/experimental/scan.py
```diff
@@ -253,24 +263,92 @@ def fn_no_output_aliasing(*args):
...
+  Returns:
+
+    A mapping from index into the flatttened
+    input pytree to the index into the list of intermediate activations.
+
+  """
+  flat_carry, _ = tree_flatten(fake_carry_pytree)
+  flat_x, _ = tree_flatten(fake_x_pytree)
+  _actual_out, activations = split(fwd_graph(*flat_carry, *flat_x), num_out)
+  input_ids = (id(v) for v in tree_iter(flat_x))
```
qihqi, line 344: probably inline those 2?

generator expressions can only be iterated once, so once you assign to a variable, you have to be careful to not use it twice (you didn't). But that is a mental burden to the reader.

qihqi (approved):

torch_xla/experimental/scan.py
tengyifei: TIL. Done.

PR #8565, opened 2025-01-14
Lower cummin op

### LINKED ISSUE #8231: Lower cummin
## 🚀 Feature
Lower cummin https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L1942-L1946

zyy-martin:
Lower cummin op. Fixes #8231.

Reused most of the changes from #8491 except for using min and argmin.

### DIFF
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -172,6 +172,7 @@ supported:
   - count_nonzero.dim_IntList
   - cross
   - cummax
+  - cummin
   - cumprod
   - cumsum
   - detach_copy
--- a/test.py
+++ b/test.py
@@ -1,0 +1,28 @@
+import torch
+import torch_xla
+import torch_xla.core.xla_model as xm
+
+
+"""
+   t_inp, t_args, t_kwargs = sample_input.input, sample_input.args, sample_input.kwargs
+    cpu_inp, cpu_args, cpu_kwargs = cpu(sample_input)
+
+    actual = torch_fn(t_inp, *t_args, **t_kwargs)
+    expected = torch_fn(cpu_inp, *cpu_args, **cpu_kwargs)
+
+[test_reference_eager] sample_input: SampleInput(input=7.358110427856445, args=(0,), kwargs={}, broadcasts_input=False, name='')
+[test_reference_eager] sample_input: SampleInput(input=7, args=(0,), kwargs={}, broadcasts_input=False, name='')
+"""
+
+a = torch.tensor(7.358110427856445, device=xm.xla_device())
+print(f'[xla]: {torch.cumsum(a, 0)}')
+
+b = torch.tensor(7.358110427856445)
+print(f'[cpu]: {torch.cumsum(b, 0)}')
+
+# import torch_xla.debug.metrics as met
+
+# # For short report that only contains a few key metrics.
+# print(met.short_metrics_report())
+# # For full report that includes all metrics.
+# print(met.metrics_report())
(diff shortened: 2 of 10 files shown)


test/test_ops.py
```diff
@@ -289,6 +288,7 @@ def get_allowed_ops_map(
     # AllowedOpInfoEntry('cosh'),
     # AllowedOpInfoEntry('cov'),
     # AllowedOpInfoEntry('cummax'),
+    # AllowedOpInfoEntry('cummin'),
```
yaochengji, line 291: @zyy-martin Thanks for your contributing.

Could you add a test for the op execution result's correctness?

ysiraichi: On this same note, could you add a test where `dim` is a 0-sized dimension? 
Since this is using the same lowering as `cummax`, it could also have the same problems. ref: #8610

zyy-martin: It has the same problem as `cummax`, as well as old ops such as `cumprod` and `cumsum`. I assume that was the reason those tests were excluded in the first place.

We can consider modify the op tests to have the option to exclude scalars/0-dimension values to cover other test cases. The C++ tests in this PR have verified the correctness though.

zyy-martin: Right now the SampleInput includes scalers by default here https://github.com/pytorch/pytorch/blob/dddf52b1b91d473e249829dad6c705c13624b35f/torch/testing/_internal/common_methods_invocations.py#L6942. I believe we can override the sample function in our test_ops.py to exclude the zero-dimension test cases. As for why the current implementation does not support scalar, i think we will need more investigation.

yaochengji: Although the test suite only support 0-sized dim, we can write simple tests from scratch to compare the result of torch/xla with native pytorch on tpu.

ysiraichi, line 291: Moving the discussion about 0-sized dimensions here, so as not to flood the other thread.
Do you think it would be hard to add support to 0-sized dimensions? 

I'm asking for that, because I think this is a BC-breaking change. Before your PR, `cummin` would fallback to CUDA, which would work with 0-sized dimensions. Now, it crashes (#8610).

That said, since other operations (e.g. `cummax`) also have this problem, I think we could merge this and work out a solution to this issue afterwards.

zyy-martin: > Do you think it would be hard to add support to 0-sized dimensions?

I am not sure, the error suggests it might have come from the XLA backend: https://github.com/openxla/xla/blob/e795171be9897e1356b1d98588e4fe784e2fc1bb/xla/service/shape_inference.cc#L202. will need more time to figure out a fix.



> That said, since other operations (e.g. `cummax`) also have this problem, I think we could merge this and work out a solution to this issue afterwards.

Agree. It also applies to core ops like `cumprod`/`cumsum` which were lowered since the beginning of ptxla.

ysiraichi: Since the XLA backend doesn't work with 0-sized dimensions, maybe we could catch this while tracing (since the shapes should be static). Then, return 2 0-sized tensors.

That said, I believe this could be left for a future PR (maybe also solving the other ops that have the same issue).

ysiraichi: I also think it would be nice to have at least one test in the Python side, as @yaochengji suggested.
Other than that, LGTM.

pgmoka: I agree with @yaochengji here to include a test for result correctness

pgmoka (changes requested):
LGTM minus the request to please add result correctness check.

PR #8570, opened 2025-01-14
Update iam.auto.tfvars

tengyifei:
Remove individual users which gets outdated. Replace with two groups, `cloud-tpus-grpadm` and `pytorchxla-dev`.


zpcore (approved):
LGTM! I will add for `pytorchxla-general@google.com` after merging the PR.

PR #8571, opened 2025-01-14
Re-introduce "XLA_USE_32BIT_LONG" flag

rpsilva-aws:
In this PR, we reintroduce the `XLA_USE_32BIT_LONG` flag to give customers the flexibility to use INT32 for their workloads. This was previously removed in 2.5 in https://github.com/pytorch/xla/pull/7582. PyTorch's inept INT32 support across the board (https://github.com/pytorch/pytorch/issues/141994) remains an issue, but given that some hw compilers do not support 64-bit types, this helps unblock customers who require this implicit XLA downcasting.

There were previously no tests for `XLA_USE_32BIT_LONG`, so simultaneously refined the test to not require specifying the flag prior to calling the test, and extending the tests to capture more cases.


rpsilva-aws:
cc: @jeffhataws @ManfeiBai @tengyifei

rpsilva-aws:
Discussed with @jeffhataws: we will do a slow roll out of the S64/U64 implicit downcast in the next releases. For 2.6, we will have it as an opt-in, as opposed to an opt-out.

jeffhataws (approved):

qihqi (approved):

rpsilva-aws:
Note that the way forward that we will be tracking, is to upstream this to PyTorch via https://github.com/pytorch/pytorch/issues/141994. Once that is done, we can remove this flag, since that'll be inherited when generating the IRs. Until then, there is quite a lot of value with this, since supported implicit downcasts provide out-of-the-box support for customers who wish downcast these types. Neuron is one case, since we avoid implicit 64-bit truncation on the compiler, by pushing those conversions (safely - it'll error if not safe, e.g. x = torch.tensor(1); x.to(int64)) to the lowering.
JAX doesn't give this flexibility today, but we will likely add that there: https://github.com/jax-ml/jax/issues/24682

test/test_data_type.py
```diff
@@ -1,73 +1,96 @@
...
-      assert 'f32' in device_data_hlo, device_data_hlo
-    else:
-      assert 'f64' in device_data_hlo, device_data_hlo
+    device_data_hlo = hlo_text.split('\n')[2]
+    self.assertIn('xla::device_data', device_data_hlo)
+    self.assertIn(expected_type, device_data_hlo)
+
+  def test_datatype_use_bf16(self):
+    self._set_env(XLA_USE_BF16='1')
+    self._test_datatype(torch.double, 'bf16', torch.floor_divide)
```
bhavya01, line 46: XLA_USE_BF16, XLA_USE_FP16 and the downcast env variables are not used anywhere. I don't think that we need to test for those.

rpsilva-aws: `XLA_USE_BF16` is still used: https://github.com/pytorch/xla/blob/3d1d1b9ec12c6cdaf323bf25611c1af72078106f/torch_xla/csrc/dtype.cpp#L11

I agree for `XLA_USE_FP16`, it seems that we had lingering tests for it, albeit already deprecated.

rpsilva-aws: Removed both flags for FP16, but kept BF16 since we still use it. PTAL.

jeffhataws: @bhavya01 @rpsilva-aws @tengyifei let's aim to deprecate XLA_USE_BF16/XLA_DOWNCAST_BF16 in 2.7.

jeffhataws (approved):

bhavya01 (approved):

test/test_data_type.py
bhavya01: Ah, I was under the impression that we had removed XLA_USE_BF16 (#7582) but we brought it back later. The plan to deprecate it 2.7 sounds good. Thanks!

PR #8573, opened 2025-01-15
Prepare to make rc5 wheel

tengyifei:
PyTorch upstream has tagged rc5: https://github.com/pytorch/pytorch/tree/v2.6.0-rc5

Update CI to pull rc5.

After this commit, I'll add an rc5 tag and run tests.

Also we remove broken CUDA builds.


ManfeiBai (approved):
Thanks, LGTM

PR #8575, opened 2025-01-15
Update terraform providers for tpu-pytorch

tengyifei:
The terraform infra enforcer for the `tpu-pytorch` project started failing in https://github.com/pytorch/xla/runs/35623339584. After searching the error, I found https://github.com/hashicorp/terraform-provider-google/issues/12584#issuecomment-2399758544 which suggests updating the google terraform provider, then changing some configs.

This change updates the google terraform provider.


tengyifei:
I'm not sure how to test this. I ran `terraform apply` locally and it printed

```

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

```

So it appears that by updating terraform, we may have fixed the build error in https://github.com/pytorch/xla/runs/35623339584

zpcore:
Thanks, it’s weird that after the apply there is 0 changed while the CI is highlighting the difference. Have you already applied the change instead of terraform plan? I think we can try merge the PR though and see.

zpcore (approved):
Let’s merge and it should do no harm unless we terraform apply.

PR #8578, opened 2025-01-15
Fix mark_sharding logic

yaochengji:
Always call `SetSharding` to ensure the `CustomSharding` op has the correct
  sharding, especially if a view is updated afterward. Updating a view can
  modify the IR, potentially leading to the sharding being applied to the
  updated view instead of the original `CustomSharding` op.

  For example, consider the following IR:
  ```
  x0 = custom_sharding(input)
  x1 = view_update(x0)
  ```
  In this case, we want to ensure the sharding is applied to `x0`, not `x1`.

  While this solution may add a sharding spec to non-CustomSharding ops like
  `x1`, the XLA compiler will safely ignore it.


JackCaoG (approved):

JackCaoG:
@yaochengji you might want to directly open branch from master instead of a fork. Otherwise the build cache in CI won't be enable and CI will take longer.

yaochengji:
@JackCaoG got it, thanks!

yaochengji:
Update: this PR is available in the nightly wheel.

PR #8579, opened 2025-01-15
Improve build_developer.sh

tengyifei:
- Remove existing torch and torch_xla installations. When these packages are installed from PyPI, it looks like `python setup.py` can't overwrite them in all cases.

- Remove old wheels.

- Install pallas specific dependencies.

Tested: I ran this on some VMs.

### DIFF
--- a/scripts/build_developer.sh
+++ b/scripts/build_developer.sh
@@ -7,6 +7,12 @@ set -x  # Display commands being run.
 cd "$(dirname "$(readlink -f "$0")")"
 cd ../../
 
+# First remove any left over old wheels
+# and old installation
+pip uninstall torch -y
+python3 setup.py clean
+
+# Install pytorch
 python3 setup.py bdist_wheel
 python3 setup.py install
 cd ..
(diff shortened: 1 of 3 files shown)


scripts/build_developer.sh
```diff
@@ -27,6 +35,11 @@ pip install torch_xla[tpu] \
   -f https://storage.googleapis.com/libtpu-wheels/index.html \
   -f https://storage.googleapis.com/libtpu-releases/index.html
 
+# Install Pallas dependencies
```
yaochengji, line 38: Do we also need to mention installing torch_xla[pallas] in README?

scripts/build_developer.sh
```diff
@@ -7,6 +7,12 @@ set -x  # Display commands being run.
 cd "$(dirname "$(readlink -f "$0")")"
 cd ../../
 
+# First remove any left over old wheels
+# and old installation
+pip uninstall torch -y
+rm dist/*
```
rpsilva-aws, line 13: Would `python3 setup.py clear` entirely or at least partially (`rm dist/*`) suffice for the cleanup, whilst ensuring the environment is clean? Similar to torch_xla's.

lsy323 (approved):

scripts/build_developer.sh
tengyifei: Yes, that's a good idea.

scripts/build_developer.sh
tengyifei: Looks like at least that works for pytorch, but for torch_xla it seems to do bazel clean instead.

PR #8582, opened 2025-01-16
Prepare to make rc6 wheel

tengyifei:
PyTorch upstream has tagged rc6: https://github.com/pytorch/pytorch/tree/v2.6.0-rc6

Update CI to pull rc6.

After this commit, I'll add an rc6 tag and run tests.


ManfeiBai (approved):
Thanks, LGTM

PR #8584, opened 2025-01-16
Introducing experimental gradient accumulation API

rpsilva-aws:
In this PR, we introduce `experimental.gradient_accumulation` which leverages XLA's `While` op to accumulate gradients.

```
Training loop with traditional gradient accumulation
===> Preparing data..
Epoch 0 step 8 loss 1.1098170280456543
Epoch 0 step 16 loss 1.1719611883163452
Epoch 0 step 24 loss 3.453134536743164
Epoch 0 step 32 loss 2.518792152404785
Epoch 0 step 40 loss 6.67546272277832
Epoch 0 step 48 loss 4.609560012817383
Epoch 0 step 56 loss 5.953202247619629
Epoch 0 step 64 loss 1.325960636138916
Training loop with XLA's `While` gradient accumulation
===> Preparing data..
Epoch 0 step 8 loss 1.1098170280456543
Epoch 0 step 16 loss 1.1719611883163452
Epoch 0 step 24 loss 3.453134536743164
Epoch 0 step 32 loss 2.518792152404785
Epoch 0 step 40 loss 6.67546272277832
Epoch 0 step 48 loss 4.609560012817383
Epoch 0 step 56 loss 5.953202247619629
Epoch 0 step 64 loss 1.325960636138916
```

```
Accumulates gradients over multiple training steps using XLA's `While`
   operator to iterate over the leading dimension of the iterable tensors.
   The backward computation of the model is implicitly executed following the
   train_step operations.
Notes:
  The model tracing will happen entirely within the loop. Hence, it is
  assumed that `train_step` is purposefully encapsulated inside of the
  loop. Hence, it is not recommended to have any operation involving the
  model parameters outside of `train_step`.
Args:
  train_step: Training function that takes iterable tensors and carried
        tensors, and returns either a loss tensor or a tuple of (loss,
        *carried_outputs). The iterable tensor inputs to this function should
        disregard the leading dimension.
  iterable_tensors: Input tensors to iterate over. All tensors must have the
        same first dimension size which determines number of iterations. The
        underlying loop in the gradient accumulation will iterate through the
        leading dimension of these tensors.
  model: PyTorch model whose parameters will be updated. Note that the entire
          model computation will be traced and generated from within the loop.
  carried_tensors: Optional tensors passed and updated between iterations.
Returns:
  (accumulated_loss, carried_tensor0, carried_tensor1, ...): A tuple including
  the `accumulated_loss` and the same unpacked `carried_tensors` that were
  provided as inputs. In addition, the model parameter gradients, if
  applicable, contain the accumulated gradients.
```

### DIFF
--- a/test/spmd/test_train_spmd_linear_model.py
+++ b/test/spmd/test_train_spmd_linear_model.py
@@ -10,7 +10,10 @@ import test_xla_sharding_base
 
 parent_folder = os.path.dirname(os.path.dirname(__file__))
 sys.path.append(parent_folder)
+
+# TODO - Unify the MLP SPMD linear training files.
 from utils.train_spmd_linear_model import train_and_evaluate
+from utils.train_spmd_linear_model_grad_acc import train_and_evaluate_grad_acc
 
 # CPU does not support optimization barriers, and hence we use this to disable
 # the gradient checkpointing A/B test run for it.
@@ -41,8 +44,7 @@ class TestSPMDLinearModel(test_xla_sharding_base.XlaShardingTest):
         checkpointing_losses = train_and_evaluate()
         # Verify that the runs match with and without checkpointing.
         assert all(
-            torch.allclose(
-                baseline_loss, checkpointing_loss, rtol=1e-3, atol=1e-7)
+            torch.allclose(baseline_loss, checkpointing_loss)
             for baseline_loss, checkpointing_loss in zip(
                 baseline_losses, checkpointing_losses))
 
@@ -58,17 +60,17 @@ class TestSPMDLinearModelGradientAccumulation(
     COMMON_GRAD_ACC_ARGS = ["--gradient_accumulation_steps", "8"]
     print('Training loop with traditional gradient accumulation')
     with extended_argv(COMMON_GRAD_ACC_ARGS):
-      baseline_grad_acc_losses = train_and_evaluate()
+      baseline_grad_acc_losses = train_and_evaluate_grad_acc()
 
     print('Training loop with XLA\'s `While` gradient accumulation')
     with extended_argv(COMMON_GRAD_ACC_ARGS +
                        ["--use_gradient_accumulation_loop"]):
-      loop_grad_acc_losses = train_and_evaluate()
+      loop_grad_acc_losses = train_and_evaluate_grad_acc()
 
     # Verify that the model losses are not zero, and that the runs match.
     assert all(loss != 0 for loss in baseline_grad_acc_losses)
     assert all(
-        torch.allclose(baseline_loss, checkpointing_loss, rtol=1e-3, atol=1e-7)
+        torch.allclose(baseline_loss, checkpointing_loss)
         for baseline_loss, checkpointing_loss in zip(baseline_grad_acc_losses,
                                                      loop_grad_acc_losses))
 
--- a/test/spmd/test_train_spmd_linear_model.py
+++ b/test/spmd/test_train_spmd_linear_model.py
@@ -32,21 +32,17 @@ class TestSPMDLinearModel(test_xla_sharding_base.XlaShardingTest):
   def test_basic(self):
     print('Training loop with baseline')
     with extended_argv([]):
-      baseline_losses, baseline_result = train_and_evaluate()
+      baseline_losses = train_and_evaluate()
     # Verify that the model losses are not zero.
     assert all(loss != 0 for loss in baseline_losses)
-    # Verify that the model produces non-zero outputs.
-    assert not torch.any(baseline_result == 0)
-
     if not SKIP_GRADIENT_CHECKPOINTING:
       print('Training loop with gradient checkpointing')
       with extended_argv(['--use_gradient_checkpointing']):
-        checkpointing_losses, checkpointing_result = train_and_evaluate()
+        checkpointing_losses = train_and_evaluate()
         # Verify that the runs match with and without checkpointing.
-        assert torch.allclose(
-            baseline_result, checkpointing_result, rtol=1e-06, atol=1e-09)
         assert all(
-            torch.allclose(baseline_loss, checkpointing_loss)
+            torch.allclose(
+                baseline_loss, checkpointing_loss, rtol=1e-3, atol=1e-7)
             for baseline_loss, checkpointing_loss in zip(
                 baseline_losses, checkpointing_losses))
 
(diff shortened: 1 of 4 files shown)


rpsilva-aws:
@jeffhataws @tengyifei

torch_xla/experimental/gradient_accumulation.py
```diff
@@ -0,0 +1,387 @@
...
+    >>> # Note: This is a partial example, since it is dependent on the
+    >>> # training model. Please refer to existing tests.
+    >>> 
+    >>> from torch_xla.experimental.gradient_accumulation import (
+    >>>    gradient_accumulation
+    >>> )
+    >>> 
+    >>> def train_step(input, label, other_tensor):
+    >>>   output = model(input_id)
+    >>>   loss = loss_fn(output, label)
```
aws-rhsoln, line 75: I think a pytorch user would be used to writing loss.backwrd themselves. This can get confusing where we compute the backward for them.

torch_xla/experimental/gradient_accumulation.py
```diff
@@ -0,0 +1,387 @@
...
+    else:
+      loss, *carried_tensors = result
+    loss /= context.num_gradient_steps
+    gradients = torch.autograd.grad(loss, model_parameters)
+    acc_grads = [prev_grad + grad for prev_grad, grad in zip(grads, gradients)]
+    return (iteri, loss, *iterable_tensors, *carried_tensors, *params,
+            *acc_grads)
+
+  # Initialize the gradients to zero.
+  grads = [
```
aws-rhsoln, line 372: Instread of creating grads, why can't we create param.grad and then collect all the param.grad in a list and pass that to the while loop. This way the loss.backward would be able to reuse the param.grad and we won't have to explicitly call autograd

torch_xla/experimental/gradient_accumulation.py
```diff
@@ -0,0 +1,387 @@
...
+      torch.zeros(p.size()).to(p.device).requires_grad_(p.requires_grad)
+      for p in model_parameters
+      if p.requires_grad
+  ]
+
+  # Apply gradients to parameters
+  result = _gradient_accumulation_impl(context, body_fn, iterable_tensors,
+                                       model_parameters, grads, carried_tensors)
+
+  for param, grad in zip(model_parameters,
```
aws-rhsoln, line 382: If we use param.grad inside the loop, maybe this copy might not be required.

torch_xla/experimental/gradient_accumulation.py
```diff
@@ -0,0 +1,387 @@
...
+        curr_iter.builder(), context.num_gradient_steps, dtype=xb.Type.S32)
+
+  def _compute_output_indices(
+      context: GradientAccumulationContext) -> List[int]:
+    # Start with loss index
+    indices = [1]
+    # Add indices for carried tensors
+    carried_start = context.num_internal_tensors + context.num_iterable_tensors
+    carried_end = carried_start + context.num_carried_tensors
+    indices.extend(range(carried_start, carried_end))
```
aws-rhsoln, line 329: Is it guaranteed to have this order always?

rpsilva-aws: Yes, it honors the order for iterable tensors and carried tensors.

tengyifei:
@rpsilva-aws do you plan on merging this into r2.6?

rpsilva-aws:
@tengyifei Ideally, yes. It's perfectly fine for the 3-layer MLP, but we're seeing a small difference for Llama runs (difference being, from a previous local patch set that was just before cleaning some of the code), so we're just quickly identifying what it is.

torch_xla/experimental/gradient_accumulation.py
rpsilva-aws: The `train_step` is what gets fed to the lowering context, so loss.backward() ends up being missed in the body computation.
The recommendation here ends up being that we want the while_loop/fori_loop/scan - since at that point, the entire body is controlled by the user. Given that we have the loss.backward() issue, instead not doing in-place accumulation of the gradients internally - it's why we ended up with this grad acc API. Once we have a solid scan/while loop implementation, we can then migrate to it.

torch_xla/experimental/gradient_accumulation.py
rpsilva-aws: Unfortunately, it'll pull in the initialization of `param.grad` inside of the body, namely because these are 0'ed initialized tensors (broadcasted initialization). In which case, the scope of the body computation will be effectively taking in an empty gradient, and computing the newer one. In this case, it doesn't actually accumulate the gradients through all the computation iterations. In this case, since we don't do in-place, but instead carry the gradients, we'll just accumulate (add) on each body computation, bypassing that issue altogether.

We could do this, if we were able to entirely fake the model, but we would require the placeholder work that would prevent us from OOMing.

torch_xla/experimental/gradient_accumulation.py
```diff
@@ -0,0 +1,387 @@
...
+      loss, *carried_tensors = result
+    loss /= context.num_gradient_steps
+    gradients = torch.autograd.grad(loss, model_parameters)
+    acc_grads = [prev_grad + grad for prev_grad, grad in zip(grads, gradients)]
+    return (iteri, loss, *iterable_tensors, *carried_tensors, *params,
+            *acc_grads)
+
+  # Initialize the gradients to zero.
+  grads = [
+      torch.zeros(p.size()).to(p.device).requires_grad_(p.requires_grad)
```
rpsilva-aws, line 373: TODO: Need to propagate the op sharding.

tengyifei:
Okay, please aim to sort out all critical issues by Jan 21 if you're aiming for 2.6 so that we could review and cherrypick it by Jan 22. 2.6 release is quicking drawing in and I would like a few days to test all the builds.

torch_xla/experimental/gradient_accumulation.py
```diff
@@ -0,0 +1,402 @@
...
+
+  init_grads = []
+  # Initialize the gradients to zero.
+  for param in model_parameters:
+    if not param.requires_grad:
+      continue
+    if param.grad:
+      grad = param.grad
+    else:
+      grad = torch.zeros(param.size()).to(param.device).requires_grad_(True)
```
aws-rhsoln, line 385: A grad object doesn't need requires_grad to be true ? Its the param object that might need it but not the grad object

rpsilva-aws: Indeed, no higher order specifics here, similar to the distributed checkpointing.

jeffhataws (approved):

rpsilva-aws:
Tests are succeeding with the default `allclose` for the gradient accumulation. It's only the gradient checkpointing (existing test that I added in the previous PR) that has a relative difference of `4.6153e-4` for the second loss step (all other above 1e-5), likely due to the fact that we reorganized the train loop (some added no-op reshape ops). In any case, this is a safe change, since the added scope is specific to the test that is successful, but test-specific refactoring is having a minor hiccup with the former checkpointing test.

test/spmd/test_train_spmd_linear_model.py
```diff
@@ -30,26 +35,49 @@ class TestSPMDLinearModel(test_xla_sharding_base.XlaShardingTest):
...
-    # Verify that the model produces non-zero outputs.
-    assert not torch.any(baseline_result == 0)
-
     if not SKIP_GRADIENT_CHECKPOINTING:
       print('Training loop with gradient checkpointing')
       with extended_argv(['--use_gradient_checkpointing']):
-        checkpointing_losses, checkpointing_result = train_and_evaluate()
+        checkpointing_losses = train_and_evaluate()
         # Verify that the runs match with and without checkpointing.
-        assert torch.allclose(baseline_result, checkpointing_result)
```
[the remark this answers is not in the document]
rpsilva-aws: An artifact from the TPU validation issue (possibly with the reshape + gather expected from grad acc) not being within the 1e-05 range for one of the losses. Since I brought back the original MLP training file, this should no longer have any impact. Thanks, brought it back.

test/spmd/test_train_spmd_linear_model.py
```diff
@@ -30,26 +35,49 @@ class TestSPMDLinearModel(test_xla_sharding_base.XlaShardingTest):
   def test_basic(self):
     print('Training loop with baseline')
     with extended_argv([]):
-      baseline_losses, baseline_result = train_and_evaluate()
+      baseline_losses = train_and_evaluate()
     # Verify that the model losses are not zero.
     assert all(loss != 0 for loss in baseline_losses)
-    # Verify that the model produces non-zero outputs.
-    assert not torch.any(baseline_result == 0)
```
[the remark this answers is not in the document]
rpsilva-aws: Likewise, added it back.

tengyifei (approved):

rpsilva-aws:
There was some setup issue with CPU's `xla_op1`. It is not relevant in this PR (well-scoped API), which succeed in all previous runs. The test is in `xla_op3`: https://github.com/pytorch/xla/actions/runs/12913551256/job/36016265400?pr=8584#step:13:676.
Pending TPU run.

tengyifei:
Looks like everything passed after a retry

PR #8589, opened 2025-01-17
Re-introduce "XLA_USE_32BIT_LONG" flag

rpsilva-aws:
## Cherry pick of https://github.com/pytorch/xla/pull/8571

In this PR, we reintroduce the `XLA_USE_32BIT_LONG` flag to give customers the flexibility to use INT32 for their workloads. This was previously removed in 2.5 in https://github.com/pytorch/xla/pull/7582. PyTorch's inept INT32 support across the board (https://github.com/pytorch/pytorch/issues/141994) remains an issue, but given that some hw compilers do not support 64-bit types, this helps unblock customers who require this implicit XLA downcasting.

There were previously no tests for `XLA_USE_32BIT_LONG`, so simultaneously refined the test to not require specifying the flag prior to calling the test, and extending the tests to capture more cases.


rpsilva-aws:
@jeffhataws @tengyifei PTAL

tengyifei (approved):

Issue #8577, opened 2025-01-15
Bring back PyTorch/XLA GPU tests/builds


tengyifei:
## 🐛 Bug

PyTorch/XLA on GPUs builds have been failing since Oct 21, 2024.

In order to bring back GPU builds and tests, the first challenge is to build PyTorch/XLA with clang and hermetic CUDA [1]. After that, there may be more tests to fix.

This bug tracks work/discussions needed to bring back GPU builds.

[1]: https://openxla.org/xla/hermetic_cuda#supported_hermetic_cuda_cudnn_versions

tengyifei:
Chengji has a wip branch for hermetic CUDA: https://github.com/yaochengji/xla/tree/chengji/clang-herm

tengyifei:
cc @ysiraichi

ysiraichi:
Update: I have tried Chengji's branch, but the build kept failing with:

```
failed: undeclared inclusion(s) in rule '@zlib//:zlib':
this rule is missing dependency declarations for the following files included by 'zutil.c':
  '/usr/lib/clang/17/include/stddef.h'
  '/usr/lib/clang/17/include/__stddef_max_align_t.h'
  '/usr/lib/clang/17/include/limits.h'
  '/usr/lib/clang/17/include/stdarg.h'
```

Still investigating it.

ysiraichi:
Here's a more verbose update on transitioning PyTorch/XLA to use OpenXLA hermetic CUDA. In summary, these are the things I have added to the build system ([branch diff](https://github.com/pytorch/xla/compare/master...ysiraichi:xla:hermetic-cuda?expand=1)):

- Followed [the documentation](https://openxla.org/xla/hermetic_cuda#supported_hermetic_cuda_cudnn_versions), and added the necessary code to `WORKSPACE` and `.bazelrc`
- Added a few more things that I could find [in the JAX repo](https://github.com/jax-ml/jax)
- Explicitly set `clang-17` as `CC` and `CXX`

Even after all these steps, I am still hitting the error above.

#### Reproducing the Error

- Get [my `hermetic-cuda` branch](https://github.com/ysiraichi/xla/tree/hermetic-cuda)
- Go to the following directory: `plugins/cuda`
- Run the following command:

```bash
$ bazel build @xla//xla/pjrt/c:pjrt_c_api_gpu_plugin.so --symlink_prefix=$(pwd)/bazel- --config=cuda
```

#### My Thoughts

[...]

PR #8286, opened 2024-10-18
disable GPU tests and build

JackCaoG:
This is to unblock the openXLA pin update. More details in https://github.com/pytorch/xla/issues/8199

PR #8593, opened 2025-01-21
Fix CUDA plugin CI.

ysiraichi:
Fix: #8577 

This PR reverts #8286, and bumps CUDA version to 12.3. The latter is needed for successfully compiling GPU dependent source code that makes use of `CUgraphConditionalHandle` (not available in 12.1) driver API typedef.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -48,7 +48,7 @@ jobs:
     needs: get-torch-commit
     with:
       # note that to build a torch wheel with CUDA enabled, we do not need a GPU runner.
-      dev-image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.10_cuda_12.3
+      dev-image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.10_cuda_12.1
       torch-commit: ${{needs.get-torch-commit.outputs.torch_commit}}
       runner: linux.24xlarge
 
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -42,25 +42,23 @@ jobs:
     secrets:
       gcloud-service-key: ${{ secrets.GCLOUD_SERVICE_KEY }}
 
-  # Disable due to https://github.com/pytorch/xla/issues/8199
-  # build-torch-with-cuda:
-  #   name: "Build PyTorch with CUDA"
-  #   uses: ./.github/workflows/_build_torch_with_cuda.yml
-  #   needs: get-torch-commit
-  #   with:
-  #     # note that to build a torch wheel with CUDA enabled, we do not need a GPU runner.
-  #     dev-image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.10_cuda_12.1
-  #     torch-commit: ${{needs.get-torch-commit.outputs.torch_commit}}
-  #     runner: linux.24xlarge
+  build-torch-with-cuda:
+    name: "Build PyTorch with CUDA"
+    uses: ./.github/workflows/_build_torch_with_cuda.yml
+    needs: get-torch-commit
+    with:
+      # note that to build a torch wheel with CUDA enabled, we do not need a GPU runner.
+      dev-image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.10_cuda_12.3
+      torch-commit: ${{needs.get-torch-commit.outputs.torch_commit}}
+      runner: linux.24xlarge
 
-  # Disable due to https://github.com/pytorch/xla/issues/8199
-  # build-cuda-plugin:
-  #   name: "Build XLA CUDA plugin"
-  #   uses: ./.github/workflows/_build_plugin.yml
-  #   with:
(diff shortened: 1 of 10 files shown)


ysiraichi:
It doesn't look widespread (haven't seen in other PRs). I will try rebasing this PR.

tengyifei:
@ysiraichi from https://github.com/pytorch/pytorch/issues/138609#issuecomment-2491806288, it looks like PyTorch upstream decided to release with some specific set of CUDA versions (see issue). Can we use one of their chosen versions, for example CUDA 12.4 instead of CUDA 12.3?

ysiraichi:
Problem is: I didn't find a docker image with CUDA 12.4. Also, I'm not sure how to create one, since it seems something internal.

tengyifei:
> Problem is: I didn't find a docker image with CUDA 12.4. Also, I'm not sure how to create one, since it seems something internal.

Could you clarify this challenge? Do you mean that you were hoping to find a `torch_xla` CUDA 12.4 docker build?

ysiraichi:
As far as I understand, PyTorch/XLA CI relies on docker images (see `dev-image`). My point is that there is no docker image with CUDA 12.4 in that registry.

https://github.com/pytorch/xla/blob/fbbdfca10324ffd95fc1ef2041a2f3fc7b6ae133/.github/workflows/build_and_test.yml#L44-L52

tengyifei:
> As far as I understand, PyTorch/XLA CI relies on docker images (see dev-image). My point is that there is no docker image with CUDA 12.4 in that registry.

@ysiraichi got it. thanks for the explanation. I think using CUDA 12.3 for now is a-okay. IIUC, most of the time we're only using torch CPU + torch_xla GPU in any case.

tengyifei:
LMK when I should review. It looks like there are still some failed tests.

ysiraichi:
After discussing this issue with @lsy323, I think the issue could be due to the older GPU in these instances. Is it okay if I try them out with a `linux.g4dn.12xlarge.nvidia.gpu`?

**Current:** `g3.8xlarge` 2x Tesla M60 (CC 5.2)
**New:** `g4dn.12xlarge` 4x Tesla T4 (CC 7.5)

@miladm @tengyifei 
Let me know what you think.

tengyifei:
Seems reasonable to me -- for comparison, we have created similar  or more advanced dev VMs before

ysiraichi:
There are some GPU tests and Triton tests that are still failing, right now. I think we should merge this PR, skipping them for now. I will make sure to open issues for each of them.

@miladm @tengyifei What do you think?

tengyifei:
> I think we should merge this PR, skipping them for now. I will make sure to open issues for each of them.

That sounds good.

.github/workflows/build_and_test.yml
```diff
@@ -42,25 +42,23 @@ jobs:
...
-  #   uses: ./.github/workflows/_build_torch_with_cuda.yml
-  #   needs: get-torch-commit
-  #   with:
-  #     # note that to build a torch wheel with CUDA enabled, we do not need a GPU runner.
-  #     dev-image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.10_cuda_12.1
-  #     torch-commit: ${{needs.get-torch-commit.outputs.torch_commit}}
-  #     runner: linux.24xlarge
+  build-torch-with-cuda:
+    name: "Build PyTorch with CUDA"
+    uses: ./.github/workflows/_build_torch_with_cuda.yml
```
tengyifei, line 47: I didn't know that we also need to build PyTorch with CUDA. Does building PyTorch with CPU not work for some reason? If we need to build PyTorch with CUDA, do you know if `12.3` is a supported CUDA version to build PyTorch with? From https://github.com/pytorch/pytorch/issues/138609 it looks like upstream picked 12.4 or 12.6

ysiraichi: Regarding CUDA version, I don't think it's a supported version. However, I believe it should compile fine. That said, I think it would be better to change it to a supported version.

tengyifei: Got it. I think we can address this as a follow up.

tengyifei: Could you add a TODO here mentioning https://github.com/pytorch/xla/issues/8700

ysiraichi:
@tengyifei All tests seem to be passing! Could you take a look at this PR?

test/torch_distributed/test_ddp.py
```diff
@@ -38,6 +39,8 @@ def _ddp_correctness(rank,
   def test_ddp_correctness(self):
     torch_xla.launch(self._ddp_correctness, args=(False, FLAGS.debug))
 
+  # Ref: https://github.com/pytorch/xla/pull/8593
```
tengyifei, line 42: Did we file separate issue(s) to fix these failures?

tengyifei (approved):
Thanks!

test/torch_distributed/test_ddp.py
ysiraichi: Not yet. But I will do that right now!

ysiraichi:
Oops. I think GitHub is waiting for either @ManfeiBai @lsy323 or @zpcore to accept it.

lsy323 (approved):

PR #8601, opened 2025-01-22
Add two placeholder example directories

tengyifei:
This way we can link them from the blog post and later fill them with actual examples.


ManfeiBai (approved):
Thanks, LGTM

PR #8609, opened 2025-01-22
use env to skip PJRT initialize

zpcore:
We skip the PJRT Megascale initialization by controlling the env. 

This is a temporary fix, and is supposed to be rolled back.

### DIFF
--- a/torch_xla/experimental/custom_kernel.py
+++ b/torch_xla/experimental/custom_kernel.py
@@ -138,10 +138,12 @@ def trace_pallas(kernel: Callable,
       return trace_pallas_arg_to_payload[hash_key], tensor_args
 
   # Here we ignore the kwargs for execution as most of the time, the kwargs is only used in traced code.
+  os.environ['SKIP_MEGASCALE_PJRT_CLIENT'] = 'true'
   ir = jax.jit(
       kernel, static_argnums=static_argnums,
       static_argnames=static_argnames).lower(*jax_args, **kwargs).compiler_ir()
   payload = _extract_backend_config(ir)
+  os.environ.pop('SKIP_MEGASCALE_PJRT_CLIENT', None)
 
   if use_cache:
     # if we reach here it means we have a cache miss.


torch_xla/experimental/custom_kernel.py
```diff
@@ -138,10 +138,12 @@ def trace_pallas(kernel: Callable,
       return trace_pallas_arg_to_payload[hash_key], tensor_args
 
   # Here we ignore the kwargs for execution as most of the time, the kwargs is only used in traced code.
+  os.environ['SKIP_MEGASCALE_PJRT_CLIENT'] = 'true'
```
tengyifei, line 141: Can we link to a tracking bug for its removal?

tengyifei (approved):
Suggest landing this after the libtpu change is approved.

torch_xla/experimental/custom_kernel.py
zpcore: Let use b/376703567 to track the progress. Good thing is that the PR should do no harm regardless of the libtpu cl status.

tengyifei:
@zpcore cherrypicking is fine with me.

zpcore:
The test failed. I think it is due to https://github.com/pytorch/pytorch/pull/142859.

They have reverted the PR.

tengyifei:
Ack

tengyifei:
@zpcore thanks. next step is to follow the process in https://github.com/pytorch/xla/issues/8455 to create a cherrypick PR

bhavya01:
Retrospective LGTM!

miladm:
Thanks @zpcore - can we add enough details to PR descriptions to help folks without context understand the intent of the contribution more clearly please?

zpcore:
> Thanks @zpcore - can we add enough details to PR descriptions to help folks without context understand the intent of the contribution more clearly please?


The issue in multipod run is that MegascaleXLA(MXLA) will trigger device discovery when we initialize PJRT runtime with the TPU backend. With the introduction of Pallas kernel, we did an extra MXLA trigger when call `jax.jit()`.  Thus there are more than one `device discovery` been executed. Everytime we execute `device discovery`, all device will be assigned a ID. This will cause device ID mismatch after the second `device discovery` thus cause the device communication hang.

The hacky way to fix the issue is to use enviroment variable to control the `device discovery` won't be triggered when call `jax.jit`. This PR works with the fix we made internally in libtpu source code:
```c++
 const char* skip_megascale_pjrt_client = std::getenv("SKIP_MEGASCALE_PJRT_CLIENT");
  bool skip_megascale = false;
  if (skip_megascale_pjrt_client != nullptr) {
    skip_megascale = true;
  }
  if (absl::GetFlag(FLAGS_megascale_num_slices) != 1 && !skip_megascale) {
    client = xla::MegaScalePjRtClient::CreateMegaScalePjRtClient(
        std::move(tpu_client));
    ...
  }
```
With this fix, MegaScalePjRtClient will only be triggered in place (e.g.,) https://github.com/pytorch/xla/blob/82d35047258bf9bcfb78c499dda82f9510cedb13/torch_xla/core/xla_model.py#L93,
where we call `runtime::GetComputationClient()` and initialize the client.

Issue #8522, opened 2024-12-29
SPMD with scalar parameter


giladxw:
## 🐛 Bug

In torch_xla.experimental.spmd_fully_sharded_data_parallel, the function "_prepare_spmd_partition_spec" handles 
scalar wrongly, instead of returning a tuple it returns a list:
[this](https://github.com/pytorch/xla/blob/40efdb7b6571ce92797b5ba42619b79c1b147b3e/torch_xla/experimental/spmd_fully_sharded_data_parallel.py#L23) should be:
```python
  if len(partition_spec) == 0:
    return tuple(partition_spec)
```
 otherwise we get the error:
```python
  File "/usr/local/lib/python3.10/site-packages/torch_xla/experimental/spmd_fully_sharded_data_parallel.py", line 121, in __init__
    spmd.mark_sharding(
  File "/usr/local/lib/python3.10/site-packages/torch_xla/runtime.py", line 95, in wrapper
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/torch_xla/distributed/spmd/xla_sharding.py", line 582, in mark_sharding
    op_sharding = mesh.get_op_sharding(partition_spec)
TypeError: unhashable type: 'list'
```

## To Reproduce

test SpmdFullyShardedDataParallel (i.e. FSDPv2) with any model that contains a scalar parameter

## Environment

 - Reproducible on XLA backend [CPU/TPU/CUDA]: ALL
 - torch_xla version: 2.4.0, 2.5.1

radna0:
@giladxw , I'm not on the xla team but from what I see, [this](https://github.com/pytorch/xla/blob/40efdb7b6571ce92797b5ba42619b79c1b147b3e/torch_xla/experimental/spmd_fully_sharded_data_parallel.py#L34) already handles that, I think it's correct that it should return a tuple even if length is 0, But is it not unlikely that you will pass a param that has shape of 0?

ysiraichi:
Fixed by #8613.

PR #8613, opened 2025-01-22
Make partition_spec as tuple to fix issue #8522

vealocia:
make partition_spec as tuple in case any scalar exists in model's parameters or intermediate outputs #8522


tengyifei (approved):

tengyifei:
@zpcore i made all PR automatically run TPU checks so we don't need the tpuci label now

tengyifei:
Thanks!

PR #8624, opened 2025-01-24
Update README.md to recommend torch_xla cxx11 builds by default.

bhavya01:
Torch builds are compiled with cxx11 by default. Following the change, I think that we should also change our default builds to compiled with cxx11 abi. 

For now, I am just updating the README to recommend cxx11 torch_xla builds and removing the sections about pre-cxx11 builds to avoid confusion. 

Follow up PR will change the default builds to be compiled with cxx11 abi

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ To install PyTorch/XLA nightly build in a new TPU VM:
 
 ```sh
 pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -185,54 +185,17 @@ GPU and nightly builds are available in our public GCS bucket.
...
-```
+#### Use nightly build
 
-The torch wheel version `2.6.0.dev20240925+cpu` can be found at https://download.pytorch.org/whl/nightly/torch/.
-</details>
-
-#### Use nightly build after 08/20/2024
-
-You can also add `yyyymmdd` after `torch_xla-2.6.0.dev` (or the latest dev version)
+You can also add `yyyymmdd` after `torch_xla-2.7.0.dev` (or the latest dev version)
```
tengyifei, line 190: Should this be `+cxx11` too?

tengyifei (approved):

PR #8630, opened 2025-01-27
Update libtpu to 0.0.7.1

tengyifei:
This special cherrypick lets us train on multiple pods without hangs.


zpcore:
Do you know if 0.0.8 build contains the fix or not? I am trying to update to 0.0.8 but there is a TPU ci failure.

tengyifei:
@zpcore unfortunately 0.0.8 doesn't contain this fix. But I'm updating the libtpu version for r2.6 which has been using 0.0.7. If you're interested in picking up the fix in the `master` branch, perhaps we could use a nightly libtpu?

ManfeiBai (approved):
Thanks, LGTM

PR #8644, opened 2025-01-29
Fix metrics saver when a file location is specified

yaochengji:


test/run_tests.sh
```diff
@@ -167,6 +167,7 @@ function run_xla_op_tests1 {
   run_test "$CDIR/test_python_ops.py"
   run_test "$CDIR/test_ops.py"
   run_test "$CDIR/test_metrics.py"
+  XLA_METRICS_FILE=/tmp/metrics.txt run_test "$CDIR/test_metrics.py"
```
tengyifei, line 170: Could we test that the metric file is absent before the run, and present after the run?

yaochengji: Thanks for the suggestion!

tengyifei (approved):

PR #8656, opened 2025-01-30
Add a stub for host offloading docs

tengyifei:
Because the blog post is going out soon and I'm out of time, at least put some general description so people have something to read about offloading.


ManfeiBai (approved):
Thanks, LGTM

PR #8657, opened 2025-01-31
[torchax] fix einsum op conversion

yaochengji:
As jax.Array also supports __len__, we need to check whether it's a jax.Array first.


qihqi (approved):
thanks!

PR #8665, opened 2025-02-03
Transition to Hermetic CUDA

ysiraichi:
This PR transitions PyTorch/XLA to use [OpenXLA Hermetic CUDA](https://openxla.org/xla/hermetic_cuda) for compiling PyTorch/XLA CUDA plugin. Such a change is necessary since the non-hermetic CUDA compilation was flagged as deprecated.

In order to do that, the following changes were necessary:

- Implement the changes in the [Hermetic CUDA documentation](https://openxla.org/xla/hermetic_cuda)
- Change the default compiler to Clang
- Remove the anonymous namespace that wrapped MLIR passes that called `mlir::TypeID::get<[PassName]>()`, e.g. `BuildStableHLOCompositePass` in _stablehlo_composite_helper.cc_

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -27,37 +27,10 @@ build --config=short_logs
 # PyTorch/XLA uses exceptions to communicate with Python.
 build --copt=-fexceptions
 
-# Force GCC because clang/bazel has issues.
-build --action_env=CC=gcc
-build --action_env=CXX=g++
-build --spawn_strategy=standalone
+build --spawn_strategy=sandboxed
 
-###########################################################################
-
-build:clang --action_env=CC=/usr/bin/clang-17
-build:clang --action_env=CXX=/usr/bin/clang++-17
-
-# clang requires the sandbox mode. Without this, `bazel build` generates an
-# error like:
-#
-# ERROR: .../external/llvm-project/llvm/BUILD.bazel:2006:11: Compiling
-# llvm/lib/CodeGenTypes/LowLevelType.cpp failed: undeclared inclusion(s) in
-# rule '@llvm-project//llvm:CodeGenTypes':
-# this rule is missing dependency declarations for the following files
-# included by 'llvm/lib/CodeGenTypes/LowLevelType.cpp':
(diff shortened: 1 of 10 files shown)


ysiraichi:
**Update:** builds were successful, but tests are failing
- Wait for #8593

ysiraichi:
**Update:** some GPU tests were failing due to CUDNN version mismatch
- Tried modifying hermetic CUDA's CUDNN version, but version 9.0.0 is not available
- Created PR for updating the VM's CUDNN version #8718 

This PR is blocked until #8718 is merged, and a new VM is built.

miladm:
can we add a description and value prop for this bug @ysiraichi?

.bazelrc
```diff
@@ -251,8 +238,10 @@ build:linux --copt="-Werror=unused-result"
 build:linux --copt="-Wswitch"
 build:linux --copt="-Werror=switch"
 # Required for building with clang
-build:linux --copt="-Wno-error=unused-but-set-variable"
+build:linux --copt="-Qunused-arguments"
+build:linux --copt="-Wno-unused-command-line-argument"
 
 # Only include debug info for files not under XLA.
 build:dbg -c dbg
+build:dbg --copt=-gdwarf-4
```
ysiraichi, line 246: We need this if we want to debug Clang builds with `gdb` (GCC 10).

.bazelrc
```diff
@@ -27,37 +27,10 @@ build --config=short_logs
 # PyTorch/XLA uses exceptions to communicate with Python.
 build --copt=-fexceptions
 
-# Force GCC because clang/bazel has issues.
-build --action_env=CC=gcc
-build --action_env=CXX=g++
-build --spawn_strategy=standalone
+build --spawn_strategy=sandboxed
```
zhanyong-wan, line 30: I made the same change but had to revert it (https://github.com/pytorch/xla/pull/9212) as it slowed the local build drastically...

ysiraichi:
Closing this, for now, as we are shifting focus from XLA:CUDA.

PR #8677, opened 2025-02-04
update pin

lsy323:


miladm:
@lsy323 can you plz add a description?

openxla_patches/count_down.diff
```diff
@@ -0,0 +1,14 @@
+diff --git a/xla/backends/cpu/runtime/convolution_thunk_internal.h b/xla/backends/cpu/runtime/convolution_thunk_internal.h
+index 84fed6bb78..9835f12e4e 100644
+--- a/xla/backends/cpu/runtime/convolution_thunk_internal.h
++++ b/xla/backends/cpu/runtime/convolution_thunk_internal.h
+@@ -342,7 +342,8 @@ void EigenGenericConv2D(
+         Eigen::Index start = task_index * task_size;
+         Eigen::Index end = std::min(start + task_size, feature_group_count);
+         for (Eigen::Index i = start; i < end; ++i) {
+-          auto on_done = [count_down]() mutable { count_down.CountDown(); };
++          // auto on_done = [count_down]() mutable { count_down.CountDown(); };
```
ManfeiBai, line 10: do we still want to keep this comment or add TODO for notice will bring it back in the future?

setup.py
```diff
@@ -332,9 +332,6 @@ def run(self):
         'tpu': [
             f'libtpu=={_libtpu_version}',
             'tpu-info',
-            # This special version removes `libtpu.so` from any `libtpu-nightly` installations,
-            # since we have migrated to using the `libtpu.so` from the `libtpu` package.
-            "libtpu-nightly==0.1.dev20241010+nightly.cleanup"
```
ManfeiBai, line 337: cool, so we don't need libtpu-nightly wheel installed too from now?

ManfeiBai (approved):
Thanks!

openxla_patches/count_down.diff
lsy323: As mentioned in PR description, once we have a newer compiler the patch can go away.

setup.py
lsy323: I don't think we need it anymore.

PR #8680, opened 2025-02-05
fix a bug in tensor.flatten

yaochengji:


torchax/torchax/tensor.py
```diff
@@ -371,6 +371,8 @@ def _handle_tensor_constructor(self, func, args, kwargs):
           return func(*args, **kwargs)
       with jax.default_device(jax_device):
         op = self._ops.get(func)
+        if op is None and isinstance(func, torch._ops.OpOverload):
```
yaochengji, line 374: These two lines are irrelevant to the flatten bug.

qihqi (approved):

PR #8681, opened 2025-02-05
fixed the type miss match issue with symint in SPMD regime and dynamo

hosseinsarshar:
In an edge case to enable SPMD with vLLM - I faced an issue where `symints` are added to the `xla_args` when mark_dynamic is called on particular tensors - this leads to graph creation failure as `torch_xla._XLAC._get_xla_sharding_specs` only accepts [`vector<at::Tensor>`](https://github.com/maxwillzq/xla/blob/e22c9ce169ebfe50f4f583970dfe395519dfcbf8/torch_xla/csrc/init_python_bindings.cpp#L1476) therefore this PR changes the `args` which might contain symints to `xla_args_tensor_only` which was defined a [few lines above](https://github.com/hosseinsarshar/xla/blob/9cc39b04ba1ad3c2569ac2b4745809c6f2258dab/torch_xla/_dynamo/dynamo_bridge.py#L540C5-L540C25).

cc @qihqi


hosseinsarshar:
cc @miladm

PR #8682, opened 2025-02-06
enable tensor.copy_ with torch native cpu tensor as source

yaochengji:


torchax/test/test_functions.py
```diff
@@ -66,6 +66,13 @@ def test_flatten(self):
       a = a.flatten(0, 1)
       self.assertEqual(tuple(a.shape), (6, 4))
 
+  def test_copy_(self):
+    with self.env:
+      a = torch.zeros((2, 3), device="cpu")
+      b = torch.ones((2, 3))
+      b.copy_(a)
+      self.assertEqual(a, b.cpu())
```
qihqi, line 74: maybe `self.assertTrue( torch.allclose(a, b.cpu())`

qihqi (approved):

qihqi (approved):

torchax/test/test_functions.py
```diff
@@ -66,6 +66,13 @@ def test_flatten(self):
       a = a.flatten(0, 1)
       self.assertEqual(tuple(a.shape), (6, 4))
 
+  def test_copy_(self):
+    with self.env:
```
yaochengji, line 70: It is because `self.env.config.use_torch_native_for_cpu_tensor = False`, then `a` is a jax tensor.

But I use a new env for `test_copy_`, still doesn't work. @qihqi

Issue #3070, opened 2021-08-04
lower complex number operations (view_as_real, view_as_complex, conj, abs)


pitneitemeier:
## 🚀 Feature
Multiple operations on complex numbers default to cpu (visible by aten:: in metrics report) or just do not work.
Specifically I have found:
defaulting to cpu:
- torch.real(x), x.real
- torch.imag(x) x.imag
- torch.conj(x)

not working:
- torch.abs(x) when x.requires_grad==True
      prints: abs does not support automatic differentiation for outputs with complex dtype

There are probably many more that I have not found and i am happy to add to this list if other people finde more.

This is my First Issue so feel free to tell me if any Information is missing, or additional steps I can take.

JackCaoG:
Lowering `real` and `imag` are pretty easy but lowering them as a view operator is a bit tircky. @bdhirsh Do you think it would make sense to wait for your functionalization change. `Real` and `imag` also changes dtype of the tensor (`complexfloat` -> `f32` for example) which makes it a bit more tricky.

SarthakYadav:
Also facing this issue: cast op is failing with complexfloat to f32 conversion (did not check whether it works with no_grad though).

jeffhataws:
Hi @JackCaoG  do you know if this is already fixed?

JackCaoG:
hmm I am not sure, if you use nightly do you see this issue?

miladm:
@ysiraichi is this issue something we can prioritize for this week?

ysiraichi:
Sure. I can look into this.

ysiraichi:
@miladm Here's what I found after some investigation:

- `torch.real(tensor)` and `torch.imag(tensor)` are already lowered
- `torch.conj(tensor)`, which calls `aten._conj_copy(tensor)` is not lowered yet
    - Should be ready to go until the end of this week

Still haven't looked at `abs`, though. But, it looks like a problem in PyTorch core.

miladm:
Thanks for the update @ysiraichi

Issue #4032, opened 2022-09-21
Einsum still falls back in some cases


steventk-g:
## 🐛 Bug

Two issues

1. `xla::Einsum` does not take more than 2 inputs.

2. `xla::Einsum` does not support the same inputs as the PyTorch implementation of einsum, forcing us to fall back sometimes, e.g. in the case of `TestEinsumPyTorchLowerRepeatedAxisBackward`.

Elaboration on the second problem:

In order to execute `einsum_backward`, we need to execute `einsum` using "backward" equations that we derive from the original input equation. For `ijj,k->ik` we need to execute einsum with `ik,k->ijj` and `ijj,ik->k`. However, `ik,k->ijj` will not work on `xla::Einsum`. We run into the following issue: `INVALID_ARGUMENT: Transpose dimensions [0,0] are not a permutation of the operand dimensions (operand shape is f32[2])`. So today, we fall back to the ATen Native implementation. 

#### Problem 1

Pass 3 or more xla tensors into torch.einsum.

#### Problem 2
When I try 
`xla::Einsum(x, y, "ik,k->ijj", xla::PrecisionConfig::Precision::PrecisionConfig_Precision_DEFAULT, xla::PrimitiveType::F32);`

I get
`INVALID_ARGUMENT: Transpose dimensions [0,0] are not a permutation of the operand dimensions (operand shape is f32[2]).`

#### For problem 1:
Try 
```
import torch
import torch_xla
x = torch.rand(4, requires_grad=True, device= xm.xla_device())
y = torch.rand(4, requires_grad=True, device= xm.xla_device())
z = torch.rand(4, requires_grad=True, device= xm.xla_device())
torch.einsum('i,j,k->ijk"', x, y, z)
```
Notice that `xla::Einsum` is not used (e.g. by printing IR or HLO)

#### For problem 2:
Remove the fallback logic from `XLANativeFunctions::einsum` and the

a. Run `TestEinsumPyTorchLowerRepeatedAxisBackward`

OR

b. Try 
```
import torch
import torch_xla
x = torch.rand(2, 4, requires_grad=True, device= xm.xla_device())
y = torch.rand(4, requires_grad=True, device= xm.xla_device())
torch.einsum('ik,k->ijj"', x, y)
```

The program should crash inside of the `xla::Einsum` function

## Expected behavior

1. I would expect XLA to handle more than 2 input tensors, or for PyTorch to break down the inputs into smaller einsum ops for dispatch.
[...]

PR #8686, opened 2025-02-06
Lower `_conj_copy` and `alias` operation.

ysiraichi:
Fix: #3070 

This PR adds a lowering for `_conj_copy`. This operation is called by `torch.conj`, and  was being executed using the fallback path. With this PR, `torch.conj` and its decomposed functions do not fallback.

### DIFF
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -26,6 +26,7 @@ full_codegen:
   - clamp.Tensor
   - clamp_max.Tensor
   - clamp_min.Tensor
+  - _conj_copy
   - cos
   - cosh
   - elu
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -139,6 +139,7 @@ supported:
   - add.Scalar
   - add.Tensor
   - addmm
+  - alias
   - alias_copy
   - arange.start_out
   - as_strided_copy
--- a/test/cpp/test_aten_xla_tensor_2.cpp
+++ b/test/cpp/test_aten_xla_tensor_2.cpp
@@ -2420,6 +2420,26 @@ TEST_F(AtenXlaTensorTest, TestAcoshInPlace) {
   ExpectCounterChanged("xla::acosh", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestAlias) {
+  torch::Tensor a = torch::rand({2, 2}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor b = torch::alias(a);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = torch::alias(xla_a);
+    AllClose(b, xla_b, /*rtol=*/1e-3, /*atol=*/0e-5);
+  });
+}
+
+TEST_F(AtenXlaTensorTest, TestConj) {
+  torch::Tensor a = torch::rand({2, 2}, torch::TensorOptions(torch::kComplexFloat));
+  torch::Tensor b = torch::conj(a);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = torch::conj(xla_a);
+    AllClose(b, xla_b, /*rtol=*/1e-3, /*atol=*/0e-5);
+  });
+}
+
 TEST_F(AtenXlaTensorTest, TestCos) {
   torch::Tensor a = torch::rand({2, 2}, torch::TensorOptions(torch::kFloat));
   torch::Tensor b = torch::cos(a);
--- a/test/cpp/test_aten_xla_tensor_2.cpp
+++ b/test/cpp/test_aten_xla_tensor_2.cpp
@@ -2431,7 +2431,8 @@ TEST_F(AtenXlaTensorTest, TestAlias) {
 }
 
 TEST_F(AtenXlaTensorTest, TestConj) {
-  torch::Tensor a = torch::rand({2, 2}, torch::TensorOptions(torch::kComplexFloat));
+  torch::Tensor a =
+      torch::rand({2, 2}, torch::TensorOptions(torch::kComplexFloat));
   torch::Tensor b = torch::conj(a);
   ForEachDevice([&](const torch::Device& device) {
     torch::Tensor xla_a = CopyToDevice(a, device);
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2384,6 +2384,16 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
 
     self.assertEqual(actual, expected)
 
+  def test_conj_no_fallback(self):
+    met.clear_all()
+
+    tensor = torch.tensor([1 + 2j])
+    expected = torch.conj(tensor)
+    actual = torch.conj(tensor.to(xm.xla_device()))
+
+    self.assertEqual(actual, expected)
+    self.assertEqual(len(met.executed_fallback_ops()), 0)
+
 
 class MNISTComparator(nn.Module):
 
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2396,8 +2396,11 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
     actual = run("cpu")
     expected = run(xm.xla_device())
 
-    self.assertEqual(met.executed_fallback_ops(), [], message="expected no fallback operations.")
-    self.assertEqual(actual, expected.cpu(), message="XLA results should match CPU results.")
+    self.assertEqual(
+        met.executed_fallback_ops(), [],
+        message="expected no fallback operations.")
+    self.assertEqual(
+        actual, expected.cpu(), message="XLA results should match CPU results.")
 
 
 class MNISTComparator(nn.Module):
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2387,11 +2387,14 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
   def test_conj_no_fallback(self):
     met.clear_all()
 
-    tensor = torch.tensor([1 + 2j])
-    expected = torch.conj(tensor)
-    actual = torch.conj(tensor.to(xm.xla_device()))
+    def run(device):
+      tensor = torch.tensor([1 + 2j], device=device)
+      return torch.conj(tensor)
 
-    self.assertEqual(actual, expected)
+    actual = run("cpu")
+    expected = run(xm.xla_device())
+
+    self.assertEqual(actual, expected.cpu())
     self.assertEqual(len(met.executed_fallback_ops()), 0)
 
 
--- a/test/test_operations.py
+++ b/test/test_operations.py
@@ -2384,18 +2384,20 @@ class TestAtenXlaTensor(test_utils.XlaTestCase):
 
     self.assertEqual(actual, expected)
 
-  def test_conj_no_fallback(self):
+  def test_conj(self):
+    # Leave the factory out of the fallback count.
+    tensor = torch.rand(2, 2, dtype=torch.complex64)
+
     met.clear_all()
 
     def run(device):
-      tensor = torch.tensor([1 + 2j], device=device)
-      return torch.conj(tensor)
+      return torch.conj(tensor.to(device))
 
     actual = run("cpu")
     expected = run(xm.xla_device())
 
-    self.assertEqual(actual, expected.cpu())
-    self.assertEqual(len(met.executed_fallback_ops()), 0)
+    self.assertEqual(met.executed_fallback_ops(), [], message="expected no fallback operations.")
+    self.assertEqual(actual, expected.cpu(), message="XLA results should match CPU results.")
 
 
 class MNISTComparator(nn.Module):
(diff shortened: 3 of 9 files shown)


ysiraichi:
**Update:** I have found the solution to the previous error. This PR should pass all tests, now

- `alias()` operation was not lowered
    - Its default implementation instantiates a new `TensorImpl` (not `XLATensorImpl) using the same storage
    - This explained the non-XLA tensor with XLA device
- The conjugate fallback wasn't working as expected
    - It was calling `clone()` in order to materialize the conjugate
        - Its default implementation called `copy_`, which did materialize the conjugate
    - `clone()` was a lowered operation, which didn't materialize the conjugate
        - Had to explicitly do so by creating a new IR node `ConjCopy`

ysiraichi:
**Update:** tests are all passing, now

@tengyifei @pgmoka @lsy323 Could you review it whenever you have some time?

torch_xla/csrc/tensor_methods.h
pgmoka, line 1: My understanding is that tensor_methods doesn't have a test associated to it. Is this something that we should add?

It strikes me that changing the `clone` call doesn't cause a test to break somewhere.

test/test_operations.py
```diff
@@ -2384,6 +2384,19 @@ def test_cummax_0_sized_dimension(self):
 
     self.assertEqual(actual, expected)
 
+  def test_conj_no_fallback(self):
```
pgmoka, line 2387: Not an AI for this PR, but something for a potential future bug. `_no_fallback` tests are likely quite useful. We have seen bugs related to this such as https://github.com/pytorch/xla/issues/4032 and https://github.com/pytorch/xla/issues/8713.

It might be worth investigate if we can do a generalized tests to make this easy for when we lower future operations

test/test_operations.py
pgmoka, line 1: I currently don't see a test for alias. Since we are lowering it, should we not add a test for it?

test/test_operations.py
pgmoka, line 2387: It seems to me that this test is checking two behaviors:
1) Is the lowered operation being used?
2) Is the value of the lowered operation what is expected?

I think these two should be two separate tests for ease of debugging

torch_xla/csrc/ops/ops_xla_shape_fn.h
pgmoka, line 1: I couldn't find a test file for this module. Should we create a bug for this?

torch_xla/csrc/aten_xla_type.cpp
pgmoka, line 1: Should we not at a test for alias in `test_aten_xla_tensor ` as per https://github.com/pytorch/xla/blob/ca248c7e0aaf505e0e3daf82219af48185cbd96a/OP_LOWERING_GUIDE.md#unit-test?

codegen/xla_native_functions.yaml
```diff
@@ -138,6 +139,7 @@ supported:
   - add.Scalar
   - add.Tensor
   - addmm
+  - alias
```
pgmoka, line 142: Should we change the title of the PR to `Lower _conj_copy and alias operations` to better represent the sum total of the changes?

pgmoka (changes requested):
Apologies for the delay on this review. I mainly have questions around tests.

torch_xla/csrc/tensor_methods.h
ysiraichi: That's a good question. I would say we don't need it because we are already testing it by testing the operation itself. In general, I don't think `tensor_methods` are used independently, i.e. called anywhere else except by their corresponding `XLANativeFunction::<op>` implementation.

ysiraichi: The `clone` call doesn't break anything because the newly introduced parameter `is_conj` has a default value that does not trigger the added `resolve_conj()` call.

test/test_operations.py
ysiraichi: @pgmoka I think it would be nice if we had an issue that combines the ones you cited. Could you create one?

ysiraichi: What do you think of adding an error message instead?

torch_xla/csrc/ops/ops_xla_shape_fn.h
ysiraichi: That sounds good. But, in the end, I think this would already get tested in the Python operations above.

test/test_operations.py
ysiraichi: I will add only the C++ tests, since there are no Python bindings for `alias()`.

test/test_operations.py
pgmoka: I think two different tests is preferable. It is easier to debug unit tests when each case individually encompasses a singular behavior.

With https://github.com/pytorch/xla/issues/8725 as a future bug, I think it also makes sense to keep these tests separate as the operation lowering behavior tests might be refactored in the future. Having this be a separate test will provide for an easier example case to base the refactor on.

torch_xla/csrc/tensor_methods.h
pgmoka: I see. If `tensor_methods` is being covered by other tests, that is probably okay as "xla/torch_xla/csrc
/aten_xla_type.cpp" is very closely making the calls here.

pgmoka: Do we have a test that covers the case where we trigger `resolve_conj()`? If it is not, can we add a test case for it?

ysiraichi: Yes! That was actually needed for the test I added to work without functionalization.

ysiraichi: By the way, I've just realized I don't actually call `resolve_conj()` inside `clone`. But, I create a `ConjCopy` IR node (which, semantically, is the same).

test/test_operations.py
ysiraichi: I have 2 reasons to prefer only 1 test:

1. Both tests would have almost the same code, except for the `self.assertEquals(...)`. Duplicating it would make it harder to maintain.
2. Although this test also tests for the actual output, it does so just to make sure the results are equal. The main thing being tested here is whether the lowerings are actually used. The tests for the actual operation can be found in [test_ops.py](https://github.com/pytorch/xla/blob/0c4e82d60a5842da76c45c199d8ced46f7136cdb/test/test_ops.py#L81).

Let me know if you still think I should split this test.

torch_xla/csrc/tensor_methods.h
pgmoka: Gotcha. That escaped me. Thank you!

test/test_operations.py
pgmoka: > 1. Both tests would have almost the same code, except for the self.assertEquals(...). Duplicating it would make it harder to maintain.

I wouldn't be too worried about the code duplication here as we are looking at a small amount of code, and the meat of the code is in a helper method which could easily be removed from the method.

> Although this test also tests for the actual output, it does so just to make sure the results are equal. The main thing being tested here is whether the lowerings are actually used. The tests for the actual operation can be found in [test_ops.py](https://github.com/pytorch/xla/blob/0c4e82d60a5842da76c45c199d8ced46f7136cdb/test/test_ops.py#L81).

I think this is a good point, and a reasonable justification for not splitting the tests. Since we already have a test, there is no reason to duplicate.

Overall, I think given that justification, adding error message here should be sufficient

ysiraichi:
This TPU CI failure looks unrelated to me. Let me undo my changes and see if the issue persists.

torch_xla/csrc/ops/ops_lower_fn.cpp
```diff
@@ -342,6 +342,11 @@ torch_xla::XlaOpVector ClampMinTensor::Lower(LoweringContext* loctx) const {
   return ReturnOp(xla::Max(xla_input, xla_other), loctx);
 }
 
+/*torch_xla::XlaOpVector ConjCopy::Lower(LoweringContext* loctx) const {*/
```
tengyifei, line 345: Extra comment?

torch_xla/csrc/tensor_methods.cpp
```diff
@@ -1229,8 +1229,12 @@ XLATensorPtr clamp(const XLATensorPtr& input,
       Clamp(input->GetIrValue(), min_max.min, min_max.max));
 }
 
-XLATensorPtr clone(const XLATensorPtr& input) {
-  XLATensorPtr cloned = input->CreateFrom(input->GetIrValue());
+XLATensorPtr clone(const XLATensorPtr& input, bool is_conj) {
```
tengyifei, line 1232: Why does the clone method implementation depend on conj? This seems like a layering violation.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -1230,8 +1230,8 @@ at::Tensor XLANativeFunctions::clone(
     const at::Tensor& self,
     std::optional<at::MemoryFormat> /* memory_format */) {
   TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
-  return bridge::AtenFromXlaTensor(
-      tensor_methods::clone(bridge::GetXlaTensor(self)));
+  return bridge::AtenFromXlaTensor(tensor_methods::clone(
+      bridge::GetXlaTensor(self), /* is_conj= */ self.is_conj()));
```
tengyifei, line 1234: What does "is_conj" mean? If I call conj on a tensor, I would expect that to compute the complex conjugate. Why does a tensor need to know if someone did a conj op on it?

tengyifei:
I left some questions. Apologies if they don't make sense. I'm just trying to understand this PR.

torch_xla/csrc/ops/ops_lower_fn.cpp
ysiraichi: I'm trying to confirm whether the TPU CI failures are unrelated to my changes.

torch_xla/csrc/tensor_methods.cpp
ysiraichi: Because that's where conjugate is supposed to be materialized. PyTorch's conjugate is implemented by a dispatch-key, making it lazy. It's only reflected to the actual data on `clone()` calls.

ysiraichi:
@tengyifei Thank you for your review. In fact, you saw lots of comments because I was trying to figure out whether the TPU CI failure was this PR's fault or not. I will ping you for another review once I have that figured out.

ysiraichi:
Ok. This flakiness issue was kind of frustrating. TPU CI seems to be okay with my changes, now.
@tengyifei Could you take a look at this PR whenever you have some time?

ysiraichi:
I will rebase these changes and wait for CI once more.

ysiraichi:
I'm pretty sure this error is unrelated to this PR.

tengyifei:
that's my bad: https://github.com/pytorch/xla/issues/8747

torch_xla/csrc/ops/ops_xla_shape_fn.h
pgmoka: I think my opinion here would be based on test coverage. Some of these operations I think seem complex enough to warrant a small unit test. If we did something like [paramatized tests](https://github.com/google/googletest/blob/main/docs/advanced.md#how-to-write-value-parameterized-tests) could be a clean way to add tests that made additional tests easier to implement.

ysiraichi:
No problem. But I think that this only highlights more #8745.

ysiraichi:
CI is green.
@tengyifei @pgmoka Could you review this PR whenever you have some time?

test/cpp/test_aten_xla_tensor_2.cpp
```diff
@@ -2420,6 +2420,27 @@ TEST_F(AtenXlaTensorTest, TestAcoshInPlace) {
   ExpectCounterChanged("xla::acosh", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestAlias) {
+  torch::Tensor a = torch::rand({2, 2}, torch::TensorOptions(torch::kFloat));
+  torch::Tensor b = torch::alias(a);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_a = CopyToDevice(a, device);
+    torch::Tensor xla_b = torch::alias(xla_a);
+    AllClose(b, xla_b, /*rtol=*/1e-3, /*atol=*/0e-5);
```
pgmoka, line 2429: I see on this file there are many usages of AllClose with different rtol, and atol standards. This seems to have developed organically over time. Would it be worth it to create an issue to standardize these as a global variable to keep one standard for all tests?

pgmoka (commented):
LGTM minus final comments. Once addressed, feel free to @ me again or CC me for a final look. I would also be curious for an LGTM for one of the other folks in the PR.

test/cpp/test_aten_xla_tensor_2.cpp
ysiraichi: I think so. I would go even further as to say that we should factor the common parts of these tests. There's just a bunch of copy-pasting going on.

ysiraichi:
@tengyifei @pgmoka Could you stamp this CI, please?

torch_xla/csrc/tensor_methods.cpp
tengyifei: I see. Could we factor the `ir = torch::lazy::Value(torch_xla::MakeNode<ConjCopy>(ir));` into a separate `conj_copy` function instead of adding a `is_conj` flag to `Clone`? I understand the need to match PyTorch semantics, but it would still be good to avoid mixing concerns.

tengyifei (changes requested):

torch_xla/csrc/aten_xla_type.cpp
tengyifei: Got it

test/cpp/test_aten_xla_tensor_2.cpp
pgmoka: Created issue for tracking: https://github.com/pytorch/xla/issues/8776

pgmoka (commented):
LGTM. Please address other comments before submitting

tengyifei (approved):
Sorry we have some flaky tests again

pgmoka (approved):
Approved

PR #8693, opened 2025-02-09
Add 5D support for flash_attention

qihqi:


torch_xla/experimental/custom_kernel.py
```diff
@@ -257,6 +257,8 @@ def fa_custom_forward(
       ab = xs.enable_manual_sharding(
           ab, partition_spec, mesh=mesh).global_tensor
 
+          
+
```
pgmoka, line 261: NIT: Remove unnecessary additional spaces

torch_xla/experimental/custom_kernel.py
```diff
@@ -279,6 +281,14 @@ def fa_custom_forward(
     segment_ids, q_segment_ids_fa, kv_segment_ids_fa = FlashAttention.prepare_segment_ids(
         q_segment_ids, kv_segment_ids)
 
+  # support 5D inputs
+  if len(q_full_shape) == 5:
```
pgmoka, line 285: Is this reshaping only applicable for dimensions of 5, or can it be generalized? Is it the case that higher than 4 dimensions require reshaping?

torch_xla/experimental/custom_kernel.py
```diff
@@ -279,6 +281,14 @@ def fa_custom_forward(
     segment_ids, q_segment_ids_fa, kv_segment_ids_fa = FlashAttention.prepare_segment_ids(
         q_segment_ids, kv_segment_ids)
 
+  # support 5D inputs
```
pgmoka, line 284: I think this comment is unnecessary as the rest of the code makes this self-evident. What might be more useful as a comment is to answer why the following is necessary for 5d inputs.

Applicable to following instances.

torch_xla/experimental/custom_kernel.py
pgmoka, line 1: I can't find a test associated with this file. Do we have a unit test that covers it? If yes, please add a case for 5 dimension fowarding.

If not, perhaps creating a bug for this would be appropriate.

pgmoka (commented):
Please add more information to the PR description on why this change is necessary to support 5d shapes.

PR #8698, opened 2025-02-11
Support splitting physical axis in SPMD mesh

### LINKED ISSUE #8381: Support splitting physical axis in HybridMesh
## 🚀 Feature

`HybridMesh` is a utility to generate a mapping from accelerator device IDs to logical mesh coordinates. Today it doesn't support splitting a physical axis. So you can't use e.g. ICI mesh shape `(64, 4)` with TPU topology `16 x 16` (4 < 16).

## Motivation

The motivation is to scale a model to multiple pods of Trillium TPUs. For example, we may want to use:

- ICI mesh shape: `(64, 4)`
- DCN mesh shape: `(2, 1)`

over two pods of Trillium 16x16 TPUs. That requires splitting one of the `16` physical axis to `4x4` in order to map to an ICI axis size of `4`.

## Pitch

To do this we can probably reference what JAX does these days, since parts of `HybridMesh` was copied from JAX.

tengyifei:
Previously, you cannot create a logical [64, 4] mesh that is backed by a physical [16, 16] torus. That requires splitting one physical axis to support two logical axes.

This is now supported by copying a bunch of code from JAX, updating the copies we had before.

Fixes #8381

### DIFF
--- a/torch_xla/distributed/spmd/xla_sharding.py
+++ b/torch_xla/distributed/spmd/xla_sharding.py
@@ -1,3 +1,6 @@
+import collections
+from collections.abc import Generator, MutableMapping
+import math
 import os
 from collections import OrderedDict, defaultdict
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


tengyifei:
I couldn't add a test because there's no large enough TPUs in CI to demonstrate breaking a physical axis. But I did cloud build this branch and tested the docker with torchprime.

qihqi (approved):

PR #8717, opened 2025-02-18
[SPMD] Add API to disable global SPMD config

lsy323:


test/spmd/test_disable_spmd.py
```diff
@@ -0,0 +1,123 @@
...
+
+    for i in range(num_iter):
+      single_ddp_output = self._run_single_process_ddp(
+          ddp_model_replicas, input_shape=ddp_input_shape)
+      single_ddp_outputs.append(single_ddp_output)
+      self.assertEqual(
+          met.metric_data('TransferToDeviceTime')[0],
+          expected_transfer_to_device_counter[i])
+      spmd_output = self._run_spmd(spmd_model, spmd_input_shape, mesh)
+      spmd_outputs.append(spmd_output)
```
yaochengji, line 118: Is there any interaction between the SPMD and non-SPMD parts?

lsy323: No, I cannot do that since we cannot create global tensor from device shards until https://github.com/pytorch/xla/pull/8716 is merged. But I do tested them together locally and it works.

Seems we should land https://github.com/pytorch/xla/pull/8716 first.

torch_xla/runtime.py
```diff
@@ -220,8 +220,17 @@ def addressable_runtime_device_count() -> int:
...
+  Revert the environment variable set by `use_spmd`, reset  
+  """
+  os.environ["XLA_USE_SPMD"] = "0"
+  torch_xla._XLAC._set_spmd_mode(False)
+
+
 # TODO(yeounoh) introduce SPMD configuration.
-def use_spmd(auto: Optional[bool] = False):
+def use_spmd(auto: Optional[bool] = False,
+             force_tensors_on_spmd_device: Optional[bool] = False):
```
tengyifei, line 233: nit: the word "force" is meaningless here IMO. I think a more descriptive name could be `replicate_existing_tensors`.

tengyifei, line 233: If `force_tensors_on_spmd_device` defaults to False, would this end up being a backward compatibility breaking change? IIUC we used to always replicate existing tensors unconditionally.

torch_xla/runtime.py
```diff
@@ -220,8 +220,17 @@ def addressable_runtime_device_count() -> int:
...
   """API to enable SPMD mode. This is a recommended way to enable SPMD.
 
   This forces SPMD mode if some tensors are already initialized on non-SPMD
```
tengyifei, line 236: This comment is probably out of date. Also I think whoever wrote this comment had in mind an idea of forcefully enabling SPMD mode, like that's an aggressive act. But with your PR there's nothing forceful about this anymore. IIUC people now have a clear option of replicating the existing tensors, or not replicating them (and keeping them on non-SPMD devices).

torch_xla/csrc/aten_xla_bridge.cpp
```diff
@@ -365,23 +376,20 @@ std::string ToXlaString(const c10::Device& device) {
...
 
 c10::Device AtenDefaultDevice() {
   return XlaDeviceToAtenDevice(*GetDefaultDevice());
 }
 
 torch::lazy::BackendDevice GetCurrentDevice() {
-  if (!g_current_device) {
-    g_current_device = *GetDefaultDevice();
-  }
+  g_current_device = *GetDefaultDevice();
```
tengyifei, line 392: If we're always overwriting `g_current_device`, wondering is it possible to rid of this global variable?

torch_xla/csrc/device.cpp
```diff
@@ -78,19 +78,19 @@ torch::lazy::BackendDevice GetVirtualDevice() {
 }
 
 bool ShouldUseVirtualDevice() {
-  bool use_virtual_device =
+  bool g_use_virtual_device =
```
tengyifei, line 81: This `g_use_virtual_device` variable shadows a global variable of the same name.

tengyifei (changes requested):

lsy323:
`test_runtime_spmd_api` is failing because, I did some investigation and here are the findings:

`AtenXlaDeviceMapper` has 2 states: a) Contains all local devices under non-spmd mode (e.g. TPU:0, TPU1...). b) Contains `SPMD:0` under SPMD mode.

In this PR, we switch the 2 states as `use_spmd` and `disable_spmd` are called. However, in the current `use_spmd` logic, the existing live tensors will be moved onto SPMD virtual device, the `BackendDataHandle` is moved to SPMD virtual device, but the underlying backend device in LazyTensor cannot be updated due to constness [ref](https://github.com/pytorch/pytorch/blob/main/torch/csrc/lazy/core/tensor.h#L55).

In a concrete example:

1. Create some tensors under non-spmd mode, has device (XLA:0), `AtenXlaDeviceMapper` is initialized with devices `XLA:0, XLA:1..`
2. `use_spmd` is called, `AtenXlaDeviceMapper` is switch to state b); The Lazy Tensor state of non-SPMD tensors are still on XLA:0, the underlying `BackendDataHandle` is on SPMD:0.

Currently, it works by having physical devices in `AtenXlaDeviceMapper` even if SPMD is turned on after `AtenXlaDeviceMapper` is initialized. This doesn't seem to be expected but it works now.

My conclusion from the above:

Our current implementation is based on the fact that the state of `AtenXlaDeviceMapper` won't change once it's initialized, this contradicts with scenario of switching between spmd and non-spmd mode. To move forward, I think we need to make  `use_spmd` working properly with a stateful `AtenXlaDeviceMapper`.

test/spmd/test_disable_spmd.py
```diff
@@ -0,0 +1,123 @@
...
+import unittest
+
+import numpy as np
+import torch
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.debug.metrics as met
+import torch_xla.distributed.spmd as xs
+import torch_xla.runtime as xr
+local_bs = 4096 * 8
```
pgmoka, line 12: NIT: It is clearer if there is a space between imports, and global variables

Issue #8711, opened 2025-02-15
Support explicit aliasing / buffer donation


rpsilva-aws:
## 🚀 Feature

We propose introducing a new API that allows users to explicitly control buffer donation in XLA computations, particularly useful for memory optimization in large models. This feature provides fine-grained control over tensor memory management without depending on functionalization's aliasing behavior.

When working with IR device data nodes (e.g., x) and derived tensors (e.g., y = x + 1), users can now explicitly annotate buffer donations through torch_xla._XLAC.donate_buffer(...) . This provides similar memory benefits to functionalization's x += 1 but with more explicit control, annotated in the underlying lazy graph executor.

A) X donation annotation

In this case, we'd have the signature requiring a tensor we wish to donate. The invariant requires this to be a materialized device data IR node. Hence, this parameter input is included in the final XLA computation's buffer donors, whether or not it was previously included by default (propagated aliasing). Changes to this tensor, with or without functionalization, will  retain the same donation intent at the end.

B) X-Y donation annotation

In this case, we'd have the signature requiring a source and destination tensor. This would leverage the functionalization aliasing for XLATensors, by mutating the destination's aliasing to the source's tensor ID (simulating an in-place op). We expect the same invariant as A), in addition to requiring the provided tensors to have the same shape and type. Accessing the source donated tensor will throw an error.

C) Computation buffer donation propagation

At the moment, if we create a user computation, even if setting the [buffer donation](https://github.com/openxla/xla/blob/1c7d5f62e3b4c90a4b3484429c5042c69ceb5c5b/xla/service/hlo.proto#L479) entry, it will not be honored. Only once we explicitly mark step, that the final computation and respective HLO proto is generated that annotates the buffer donation indices - very simply, by mapping the parameter input IDs to all the live tensor alias IDs. This would require us to maintain and write heuristics for propagating local user computation buffer donation indices to the global context.

------

We're leaning towards A, since B would need an exhaustive sets of tests when worked interchangeably with functionalization, in addition to A being more evidently honored when building the buffer donation indices. B on the other hand, would still depend on any remaining aliasing propagation, so it wouldn't necessarily guarantee the parameter input donation - which is main motivation behind this issue. C would require significantly more effort to support this.

## Motivation

We currently do not support function argument in-place mutations on the scan and gradient accumulation APIs. It is a constraint that function arguments (e.g. gradients, weights) should not be in-place mutated at the application level. Hence, we have:
```
def body_fn(x, y):
[...]

PR #8721, opened 2025-02-19
Extend buffer donation aliasing APIs

rpsilva-aws:
Enables https://github.com/pytorch/xla/issues/8711. In particular, it alleviates the fallback aliasing logic for user config. Previously, it would only be applied if auto LTC did not infer any needed aliasing which is a semantically suboptimal surface area. In this PR, we modify this behavior to always accommodate explicitly donated buffers, working simultaneously with LTC, IF enabled.

In this PR, it is out of scope to address the per-device concerns associated with having global (DeviceArenaContext) metadata for all the devices with respect to aliasing and other dynamo execution -specific logic.

### DIFF
--- a/test/dynamo/test_dynamo_aliasing.py
+++ b/test/dynamo/test_dynamo_aliasing.py
@@ -1,3 +1,4 @@
+import sys
 import unittest
 
 import torch
--- a/test/test_input_output_aliases.py
+++ b/test/test_input_output_aliases.py
@@ -8,7 +8,29 @@ import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla.debug.metrics as met
 import unittest
+import contextlib
 import copy
+
+
+def create_xla_config_context(set_func, get_func):
+
+  @contextlib.contextmanager
+  def config_context(value):
+    original_value = get_func()
+    set_func(value)
+    try:
+      assert get_func() == value
+      yield
+    finally:
+      set_func(original_value)
+
+  return config_context
+
+
+alias_with_buffer_donor_config_context = create_xla_config_context(
+    torch_xla._XLAC._xla_set_enable_alias_with_buffer_donor_config,
+    torch_xla._XLAC._xla_get_enable_alias_with_buffer_donor_config,
+)
 
 
 # TODO(alanwaketan): add test for views.
@@ -209,6 +231,78 @@ class InputOutputAliasesTest(unittest.TestCase):
 
     # ...if it doesn't crash, the value here would be 44.
     self.assertEqual(t1.item(), 43)
+
+  def test_user_config_donation_with_ltc_donation(self):
+    with alias_with_buffer_donor_config_context(True):
+      met.clear_all()
+      xla_device = xm.xla_device()
+      t0 = torch.randn(4, 2, 2).to(xla_device)
+      t1 = torch.randn(4, 2, 2).to(xla_device)
+      self.assertTrue(torch_xla._XLAC._set_buffer_donation(t0, True))
+      self.assertTrue(torch_xla._XLAC._get_buffer_donation(t0))
+      self.assertFalse(torch_xla._XLAC._get_buffer_donation(t1))
+      t3 = t0 + t1
+      t1 += 2
+      xm.mark_step(wait=True)
+
+      # We surface the C++ runtime error by checking that the backend data is
+      # no longer present for the IR node.
+      self.assertTrue(torch_xla._XLAC._is_placecholder(t0))
+      self.assertEqual(met.metric_data("InputOutputAliasCount")[1], 2.0)
+
+  def test_user_config_donation_with_ltc_donation_overlap(self):
+    with alias_with_buffer_donor_config_context(True):
+      met.clear_all()
+      xla_device = xm.xla_device()
+      t0 = torch.randn(4, 2, 2).to(xla_device)
+      self.assertTrue(torch_xla._XLAC._set_buffer_donation(t0, True))
+      self.assertTrue(torch_xla._XLAC._get_buffer_donation(t0))
+      t0 += 2
+      xm.mark_step()
+
+      self.assertEqual(met.metric_data("InputOutputAliasCount")[1], 1.0)
+
+  def test_user_config_donation(self):
+    with alias_with_buffer_donor_config_context(True):
+      met.clear_all()
+      xla_device = xm.xla_device()
+      t0 = torch.randn(4, 2, 2).to(xla_device)
+      self.assertTrue(torch_xla._XLAC._set_buffer_donation(t0, True))
+      self.assertTrue(torch_xla._XLAC._get_buffer_donation(t0))
+      self.assertIn('XlaSetBufferDonation', met.counter_names())
+      self.assertEqual(met.counter_value('XlaSetBufferDonation'), 1)
+      t1 = t0 + 1
+      # We use _xla_sync_multi to explicitly disable sync_xla_data, which will
+      # in turn avoid using LTC aliasings. This ensures that the resulting
+      # aliasings are due to the buffer donation.
+      torch_xla._XLAC._xla_sync_multi([t0, t1], [str(xla_device)], True, False)
+
+      self.assertEqual(met.metric_data("InputOutputAliasCount")[1], 1.0)
+
+  def test_user_config_donation_inplace_aliasing(self):
+    with alias_with_buffer_donor_config_context(True):
+      met.clear_all()
+      xla_device = xm.xla_device()
+      t0 = torch.randn(4, 2, 2).to(xla_device)
+      self.assertTrue(torch_xla._XLAC._set_buffer_donation(t0, True))
+      self.assertTrue(torch_xla._XLAC._get_buffer_donation(t0))
+      t0 *= 2
+      # We use _xla_sync_multi to explicitly disable sync_xla_data, which will
+      # in turn avoid using LTC aliasings. This ensures that the resulting
+      # aliasings are due to the buffer donation.
+      torch_xla._XLAC._xla_sync_multi([t0], [str(xla_device)], True, False)
+
+      self.assertEqual(met.metric_data("InputOutputAliasCount")[1], 1.0)
+
+  def test_user_config_donation_no_op_mark_step(self):
+    with alias_with_buffer_donor_config_context(True):
+      xla_device = xm.xla_device()
+      t0 = torch.randn(4, 2, 2).to(xla_device)
+      self.assertTrue(torch_xla._XLAC._set_buffer_donation(t0, True))
+      xm.mark_step()
+      self.assertTrue(torch_xla._XLAC._get_buffer_donation(t0))
+      xm.mark_step()
+      self.assertTrue(torch_xla._XLAC._get_buffer_donation(t0))
 
 
 if __name__ == '__main__':
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1906,14 +1906,15 @@ void InitXlaModuleBindings(py::module m) {
       [](const std::string& device) { return GetRngSeed(device); },
       py::arg("device") = "");
   m.def(
-      "_xla_set_should_alias_with_buffer_donor_config",
-      [](bool should_alias, const std::string& device_str) {
+      "_xla_set_enable_alias_with_buffer_donor_config",
+      [](bool enable_user_config_alias, const std::string& device_str) {
         torch::lazy::BackendDevice device = GetDeviceOrCurrent(device_str);
-        XLAGraphExecutor::Get()->SetAliasWithBufferDonorConfig(should_alias);
+        XLAGraphExecutor::Get()->SetAliasWithBufferDonorConfig(
+            enable_user_config_alias);
       },
-      py::arg("should_alias") = false, py::arg("device") = "");
+      py::arg("enable_user_config_alias") = false, py::arg("device") = "");
   m.def(
-      "_xla_get_should_alias_with_buffer_donor_config",
+      "_xla_get_enable_alias_with_buffer_donor_config",
       [](const std::string& device_str) {
         torch::lazy::BackendDevice device = GetDeviceOrCurrent(device_str);
         return XLAGraphExecutor::Get()->GetAliasWithBufferDonorConfig();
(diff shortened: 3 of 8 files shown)


torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1287,56 +1297,70 @@ std::vector<size_t> GetBufferDonorIndexForStepMarker(
...
-      // But, when we issue a step barrier (force_ltc_data == true) we have to
-      // turn everything into DEVICE_DATA, so we can activate aliasing.
-      buffer_donor_indices = GetBufferDonorIndexForStepMarker(
-          tensors, coll.indices, parameters_data);
-    } else if (GetAliasWithBufferDonorConfig()) {
-      // only alias based on buffer donor if LTC can't auto infer the input
-      // output aliasing.
-      buffer_donor_indices = GetBufferDonorIndexFromUserConfig(parameters_data);
-    }
+  if (ShardingUtil::GetAutoSharding()) {
```
ysiraichi, line 1303: Could we move the refactoring in another PR?

ysiraichi (commented):
Overall, LGTM. I do think it would be better to leave non-trivial refactorings for another PR. It would make reviewing easier.

That said, I'm not entirely familiar with this buffer donor system. Maybe @tengyifei @lsy323 @vanbasten23 @bhavya01 could help review.

test/dynamo/test_dynamo_aliasing.py
```diff
@@ -1,3 +1,4 @@
+import sys
```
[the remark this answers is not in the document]
rpsilva-aws: It is used, but the import was missing.

test/test_input_output_aliases.py
```diff
@@ -210,7 +239,104 @@ def test_device_data_cache_no_aliasing(self):
...
+  test_cases = loader.getTestCaseNames(InputOutputAliasesTest)
+  failed = False
+  for test_name in test_cases:
+    test = InputOutputAliasesTest(test_name)
+    runner = unittest.TextTestRunner(failfast=True)
+    result = runner.run(test)
+    if not result.wasSuccessful():
+      failed = True
+
+  sys.exit(1 if failed else 0)
```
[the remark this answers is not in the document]
ysiraichi: Do you mean that we would have tests running in parallel? As far as I understand it, the tests are run in sequence, aren't they?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2737,36 +2753,43 @@ void InitXlaModuleBindings(py::module m) {
...
-            data->set_should_donate_buffer(should_donate);
-            buffer_donation_updated = true;
-          } else if (xtensor->CurrentIrValue().node != nullptr) {
-            torch::lazy::NodePtr node = xtensor->CurrentIrValue().node;
-            auto device_data = torch_xla::DeviceData::Cast(node.get());
-            if (device_data != nullptr) {
-              device_data->set_buffer_donation(should_donate);
-              buffer_donation_updated = true;
+        [](const std::vector<at::Tensor>& tensors,
+           bool should_donate) -> std::vector<bool> {
```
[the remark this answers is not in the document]
ysiraichi: Do you mean that in these real world examples this modification would improve performance in a non-negligible way? If so, I think I would want to see some kind of profiling, first.

torch_xla/csrc/xla_graph_executor.h
```diff
@@ -276,7 +289,7 @@ class XLAGraphExecutor : public torch::lazy::LazyGraphExecutor {
     torch::lazy::Value IrValueFromScalar(
         const at::Scalar& value, at::ScalarType scalar_type,
         const torch::lazy::BackendDevice& device) final;
-    bool should_alias_with_buffer_donor = false;
+    BufferDonorAliasConfig buffer_donor_alias_config{};
```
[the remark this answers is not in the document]
ysiraichi: That makes sense. We can consider adding more when the need arises, but for now, managing 2 fields separately still seems reasonable. (it also makes reviewing easier)

test/test_input_output_aliases.py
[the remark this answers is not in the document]
rpsilva-aws: I thought so too, but I see lingering graph executor states across tests. This at least enforces that we treat each test fn as a suite, so I don't think it's a functional concern to switch it to this. We have the same complications when trying to use env vars (on/off) in the same test file, hence why most tests ended up to just specifying the env var prior to running the test (e.g. https://github.com/pytorch/xla/blob/master/test/run_tests.sh#L155) which I wanted to avoid doing here.

torch_xla/csrc/init_python_bindings.cpp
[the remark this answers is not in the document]
rpsilva-aws: We can, but from a user interface point alone, why do you think it's easier to call it multiple times? Especially considering that the use case is often over a set of tensors (deciding to donate gradients, weights, optimizer states, etc). I think this is a more general question, since we could say the same about any binding that takes in a vector of tensors. We wouldn't need to make this change in this PR and collect more evidence if you're hesitant to doing this change.

test/test_input_output_aliases.py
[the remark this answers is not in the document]
ysiraichi: > I see lingering graph executor states across tests.

I think they do linger because we don't reset them in the `tearDown()` method.

> I don't think it's a functional concern to switch it to this.

I agree that it's not a functional concern. But I would say that having only this test with this custom runner might be a maintainability concern. Do you think we could keep the old state?

If you think we should adopt this strategy, I think it would be nice to have a separate issue so that people could comment in.

torch_xla/__init__.py
```diff
@@ -195,6 +195,9 @@ def _check_deprecated_env_var():
 if os.environ.get('TF_CPP_MIN_LOG_LEVEL') == '0':
   logger.setLevel(logging.INFO)
 
+if 'XLA_ENABLE_PARAM_ALIASING' in os.environ:
+  _XLAC.set_enable_parameter_aliasing(os.environ['XLA_ENABLE_PARAM_ALIASING'])
+
```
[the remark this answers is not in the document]
rpsilva-aws: That's reasonable, I'll separate it.

test/test_input_output_aliases.py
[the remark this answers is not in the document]
rpsilva-aws: I have also tried with setUp and tearDown to revert to the prior state, but didn't work out. I can do a bit more experimentation here. Unfortunately, this blocks this PR, but let me time box it.

torch_xla/csrc/init_python_bindings.cpp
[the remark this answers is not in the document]
ysiraichi: It's not that I'm hesitant to this change. It's just that this PR is introducing new behavior. So, I think it would be easier to review and understand it if this PR had only the set of changes for that.

This Python binding + the refactoring aren't necessary for this new behavior. They deserve their own issues (pitching that idea) and PRs.

That said, my opinion is that there seems not to be a compelling reason to do this Python binding change. However, I'm not really familiar with how it's used out in the wild. So, let's wait to hear what the others have to say.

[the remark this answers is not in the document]
rpsilva-aws: Absolutely. Sounds good, let's separate it and I can come back to it with any potential impact data.

rpsilva-aws:
As suggested by @ysiraichi, so we don't block on this needed PR, deferring:

> - Switch to using a batch call for the donation - since in practice, there should be a large number of tensors that can be collected and simultaneously annotated. Note that the dynamo execution API remain unchanged, as these are intentionally not modified in this PR.

In addition to separating the following as separate PRs (I'll do so shortly after closing this one):
- Refining the parameter disabling user behavior (https://github.com/pytorch/xla/pull/8721#discussion_r1964297682)
- Gradient accumulation API changes

ysiraichi:
> I'll do so shortly after closing this one

Just so we are clear: are you closing this one? Or do you mean after this gets merged?

rpsilva-aws:
> > I'll do so shortly after closing this one
> 
> Just so we are clear: are you closing this one? Or do you mean after this gets merged?

I will do these as separate PRs after this one is merged.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1290,53 +1290,73 @@ std::vector<size_t> XLAGraphExecutor::GetBufferDonors(
...
+  // any duplicates when merging the two indices vector.
+  std::vector<size_t> buffer_donor_indices;
+  buffer_donor_indices.reserve(
+      std::max(ltc_buffer_donor_indices.size(),
+               user_config_buffer_donor_indices.size()));
+  std::set_union(ltc_buffer_donor_indices.cbegin(),
+                 ltc_buffer_donor_indices.cend(),
+                 user_config_buffer_donor_indices.cbegin(),
+                 user_config_buffer_donor_indices.cend(),
+                 std::back_inserter(buffer_donor_indices));
```
ysiraichi, line 1353: Was this supposed to be a refactoring PR? Reason I'm asking is because these lines are changing the old behavior.

ysiraichi (commented):
Just one comment. Looks good otherwise.
Could you also change the PR description so we know exactly what changes this PR is introducing?

torch_xla/csrc/xla_graph_executor.cpp
rpsilva-aws: No, this is the actual PR that resolves the PFR. We're extending the buffer donation to interchangeably include user config buffer donations with LTC aliasing.

test/test_input_output_aliases.py
```diff
@@ -210,6 +232,69 @@ def test_device_data_cache_no_aliasing(self):
...
+      t0 = torch.randn(4, 2, 2).to(xla_device)
+      t1 = torch.randn(4, 2, 2).to(xla_device)
+      self.assertTrue(torch_xla._XLAC._set_buffer_donation(t0, True))
+      self.assertTrue(torch_xla._XLAC._get_buffer_donation(t0))
+      self.assertFalse(torch_xla._XLAC._get_buffer_donation(t1))
+      t3 = t0 + t1
+      t1 += 2
+      xm.mark_step()
+
+      self.assertEqual(met.metric_data("InputOutputAliasCount")[1], 2.0)
```
tengyifei, line 248: Is it possible to test that t0 can no longer be used?

test/test_input_output_aliases.py
```diff
@@ -210,6 +232,69 @@ def test_device_data_cache_no_aliasing(self):
...
+
+  def test_user_config_donation_with_ltc_donation_overlap(self):
+    with alias_with_buffer_donor_config_context(True):
```
tengyifei, line 251: What does this config context do?

test/test_input_output_aliases.py
```diff
@@ -210,6 +232,69 @@ def test_device_data_cache_no_aliasing(self):
...
+      met.clear_all()
+      xla_device = xm.xla_device()
+      t0 = torch.randn(4, 2, 2).to(xla_device)
+      self.assertTrue(torch_xla._XLAC._set_buffer_donation(t0, True))
+      self.assertTrue(torch_xla._XLAC._get_buffer_donation(t0))
+      self.assertIn('XlaSetBufferDonation', met.counter_names())
+      self.assertEqual(met.counter_value('XlaSetBufferDonation'), 1)
+      t1 = t0 + 1
+      torch_xla._XLAC._xla_sync_multi([t0, t1], [str(xla_device)], True, False)
```
tengyifei, line 272: What does sync_multi do? Is it the same as mark_step?

torch_xla/csrc/xla_graph_executor.h
```diff
@@ -95,8 +95,7 @@ class XLAGraphExecutor : public torch::lazy::LazyGraphExecutor {
   torch::lazy::BackendDataPtr GetBaseSeedData(
       const torch::lazy::BackendDevice& device);
 
-  void SetAliasWithBufferDonorConfig(bool should_alias);
-
+  void SetAliasWithBufferDonorConfig(bool enable_alias);
```
tengyifei, line 98: I see you've renamed "should" to "enable". Does this imply some feature change?

tengyifei (commented):
Apologies for the delay. I'll be frank I don't understand the internal details that well. In fact nobody on the team does. So I've asked some questions on the PR.

test/test_input_output_aliases.py
rpsilva-aws: It enables the general behavior of buffer donation. In other words, this (`_xla_set_enable_alias_with_buffer_donor_config`) needs to be set, so that the tensors (with the associated XLA device data backend) that were explicitly marked to be donated (`_set_buffer_donation`) can be donated in the next execution (i.e. mark step) [1]. It's effectively a general opt-in for the entire donation behavior (arguably we wouldn't need it since `_set_buffer_donation` is already assuming that the user does intend to explicitly donate it, but I just kept it as is).

https://github.com/pytorch/xla/issues/8711 describes the problem and use case, but Jack introduced this in the context of dynamo execution (https://github.com/pytorch/xla/pull/6587). Since it's an opt-in, we want to give users the flexibility to explicitly annotate the donation (precisely when they don't need the tensor after mark step). This is particularly helpful with the gradient accumulation (and scan), since we can't do in-place mutations (aliasing is uncaptured by LTC), so this closes that allows us to mitigate / close that gap. The main change in this PR, to resolve the PFR, is that we enable LTC and buffer donation to coexist (and remove the previous odd constrained fallback behavior: if no LTC aliasings + user config is enabled -> donate the buffers).

[1] XLA HLO donation ref: https://github.com/openxla/xla/blob/1c7d5f62e3b4c90a4b3484429c5042c69ceb5c5b/xla/service/hlo.proto#L479

test/test_input_output_aliases.py
rpsilva-aws: Mostly, but more flexible. It allows me to specify `sync_xla_data=False`, so that I can ensure that the aliasing is due to the user config (explicit buffer donation), and not auto LTC aliasings.

torch_xla/csrc/xla_graph_executor.h
rpsilva-aws: I understand using "should" for donating buffers (`_set_buffer_donation`), because it may not necessarily donate it, since it depends whether this config is enabled (`_xla_set_enable_alias_with_buffer_donor_config`). I renamed the parameters for `_xla_set_enable_alias_with_buffer_donor_config`/`SetAliasWithBufferDonorConfig`, since we are enabling/allowing explicitly marked tensors (`_set_buffer_donation`) to be donated, so I figured that renaming it to "enable" is more suitable.

rpsilva-aws:
> Apologies for the delay. I'll be frank I don't understand the internal details that well. In fact nobody on the team does. So I've asked some questions on the PR.

No worries. Thanks for taking a look - let me know if I was able to answer these. I'll revise with the test change asap.

rpsilva-aws:
@tengyifei Done, PTAL

rpsilva-aws:
@tengyifei @ysiraichi Hey folks, can you PTAL.

test/test_input_output_aliases.py
tengyifei: Got it. So we need both the config and an explicit donation call.

torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -1290,53 +1290,73 @@ std::vector<size_t> XLAGraphExecutor::GetBufferDonors(
...
+    // Also in the first example, unless we are doing a step barrier and hence
+    // include all live tensors, if the B value is not part of the graph, it
+    // will later fetch the new value of A, which is incorrect.
+    // But, when we issue a step barrier (force_ltc_data == true) we have to
+    // turn everything into DEVICE_DATA, so we can activate aliasing.
+    ltc_buffer_donor_indices = GetBufferDonorIndexForStepMarker(
+        tensors, coll.indices, parameters_data);
+  }
+
+  std::vector<size_t> user_config_buffer_donor_indices;
```
tengyifei, line 1334: This "user config" name is really unfortunate. It should just be called "explicit donation". I was going to suggest renaming it until I realized this is an existing feature that you exposed.

tengyifei (approved):

torch_xla/csrc/xla_graph_executor.cpp
rpsilva-aws: Yes, unfortunately these are tied to existing API bindings, so they could already be used by users. We could rename, but we'd have to do it over a few release cycles with a deprecation warning.

ysiraichi (approved):
LGTM.

Issue #8678, opened 2025-02-04
Introduce a `mark_sharding` that also shards the backward


tengyifei:
## 🚀 Feature

PyTorch/XLA `xs.mark_sharding` is an in-place operation that adds sharding annotation to an XLA tensor. However, gradients to be applied to the tensor are not annotated with sharding annotations.

## Motivation

In some cases, GSPMD fails to propagate sharding annotation from the tensor to its gradient. It's useful to shard both tensor and its gradient with the same sharding annotation.

## Pitch

We could write a `torch.autograd.Function` implementation to do this.

## Additional context

JAX `mark_sharding` shards the gradients too.

cc @bhavya01

miladm:
cc @lsy323 for viz as he is working on `disabling_spmd()` API. It seems the two solutions are looking at a new subset of live tensors.

bhavya01:
Also going to add `xs.mark_sharding_and_gradients` before closing this issue.

rpsilva-aws:
Can you explain in which cases the sharding is not present for the gradients? We mentioned "in some cases", but does that point to a particular issue? I have only ever seen this happening in very particular cases, e.g., we pre-initialized gradients and then re-assign it to the parameter (after computing the gradients); we also saw this when adding backward hooks for SP, where we needed to explicitly shard the gradients. In which scenario would we not want to propagate the sharding to the gradients?

tengyifei:
>  In which scenario would we not want to propagate the sharding to the gradients?

I cannot think of a single scenario where we don't want the same sharding on the gradients. In fact `jax.lax.with_sharding_constraints` just does this by default.

rpsilva-aws:
Odd. I was hoping that this would be the assumption behind `mark_sharding` too, but I could also see the concern of changing the existing behavior if there are actual legitimate cases for not wanting it on the gradients (though I've only ever seen it in those 2 specific cases). It would be nice to eventually converge the UX and the behavior, and maybe present users with a parameter in `mark_sharding`/flag/other that does otherwise (not propagate it).

tengyifei:
Looks like the remaining item is to add a wrapper over the autograd Function introduced in #8723 , and unit test it

tengyifei:
@pgmoka if you would like to work on this you can change the assignee I think

iwknow:
> Looks like the remaining item is to add a wrapper over the autograd Function introduced in [#8723](https://github.com/pytorch/xla/pull/8723) , and unit test it

[...]

PR #8723, opened 2025-02-19
Add autograd function for mark_sharding

bhavya01:
Add an autograd function to mark_sharding on intermediate tensors during the forward pass and the gradients of the intermediate tensors during the backward pass.

### DIFF
--- a/torch_xla/distributed/spmd/xla_sharding.py
+++ b/torch_xla/distributed/spmd/xla_sharding.py
@@ -1126,16 +1126,29 @@ def _generate_logical_mesh(
 
 
 class MarkShardingFunction(torch.autograd.Function):
-    @staticmethod
-    def forward(ctx, torch_tensor: torch.Tensor, partition_spec) -> torch.Tensor:
-        mesh = get_global_mesh()
-        mark_sharding(torch_tensor, mesh, partition_spec)
-        ctx.partition_spec = partition_spec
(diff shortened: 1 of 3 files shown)


bhavya01:
Addresses #8678

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -1123,3 +1123,32 @@ def _generate_logical_mesh(
...
   return logical_mesh
+
+
+class MarkShardingFunction(torch.autograd.Function):
+  """
+  Autograd function to mark_sharding on intermediate tensors and the gradient
+  of the intermediate tensors during backward pass.
+
+  Usage:
+  new_tensor = MarkShardingFunction.apply(tensor, mesh, ('axis_1', 'axis_2'))
```
tengyifei, line 1134: WDYT adding a Python function wrapper in the `spmd` module? I think it could be convenient to have something like:

```
t = xs.mark_sharding_and_gradients(t, partition_spec)
```

next to the usual `mark_sharding` function, and we'll also be able to mention this function in the documentation of the `mark_sharding`, highlighting that the former doesn't shard gradients.

bhavya01: Yup, definitely a better UX. Will send a PR today.

PR #8730, opened 2025-02-21
write _shard_map; refactor flash attention to support 5d inputs.

qihqi:
Jax's shard_map works by enabling manual sharding on inputs, and disabling it on outputs.
here we introduce _shard_map to simulate this behavior. This is sufficient to support use cases of calling pallas.
It is not sufficient to support other use cases of shard_map, such as desire to use manual collectives. Because current collective implementation, such as `xm.all_gather` checks fail with manual sharding.

### DIFF
--- a/test/test_pallas_spmd.py
+++ b/test/test_pallas_spmd.py
@@ -80,6 +80,25 @@ class PallasTest(unittest.TestCase):
   @unittest.skipIf(xr.device_type() != 'TPU' or tpu.version() < 3,
                    "This test only works on TPUv3+.")
   @with_jax_high_precision
+  def test_flash_attention_spmd_data_parallel(self):
+    n_devices = xr.global_runtime_device_count()
+    xs.set_global_mesh(xs.Mesh(range(n_devices), (n_devices // 2, 2, 1, 1, 1)))
+
+    q = torch.randn(4, 2, 2, 128, 4).to("xla")
+    k = torch.randn(4, 2, 2, 128, 4).to("xla")
+    v = torch.randn(4, 2, 2, 128, 4).to("xla")
+
+    o = flash_attention(q, k, v, partition_spec=range(n_devices))
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(o),
+        f"{{devices=[{n_devices//2},2,1,1,1]0,1,2,3,4}}")
+
+    expected_o = self._attention(q, k, v)
+    self.assertTrue(torch.allclose(o.cpu(), expected_o.cpu(), atol=1e-05))
+
+  @unittest.skipIf(xr.device_type() != 'TPU' or tpu.version() < 3,
+                   "This test only works on TPUv3+.")
+  @with_jax_high_precision
   def test_flash_attention_spmd_data_parallel_kv_and_ab_padding(self):
     n_devices = xr.global_runtime_device_count()
     xs.set_global_mesh(xs.Mesh(range(n_devices), (n_devices, 1, 1, 1)))
--- a/test/test_pallas_spmd.py
+++ b/test/test_pallas_spmd.py
@@ -84,14 +84,16 @@ class PallasTest(unittest.TestCase):
   def test_flash_attention_spmd_data_parallel_5d(self):
     n_devices = xr.global_runtime_device_count()
     xs.set_global_mesh(
-      xs.Mesh(range(n_devices), (n_devices // 2, 2, 1, 1, 1),
-      ('fsdp', 'dp', 'a', 'b', 'c')))
+        xs.Mesh(
+            range(n_devices), (n_devices // 2, 2, 1, 1, 1),
+            ('fsdp', 'dp', 'a', 'b', 'c')))
 
     q = torch.randn(4, 2, 2, 128, 4).to("xla")
     k = torch.randn(4, 2, 2, 128, 4).to("xla")
     v = torch.randn(4, 2, 2, 128, 4).to("xla")
 
-    o = flash_attention(q, k, v, partition_spec=('fsdp', 'dp', None, None, None))
+    o = flash_attention(
+        q, k, v, partition_spec=('fsdp', 'dp', None, None, None))
     dev_ids = ','.join(map(str, range(n_devices)))
     self.assertEqual(
         torch_xla._XLAC._get_xla_sharding_spec(o),
@@ -144,7 +146,7 @@ class PallasTest(unittest.TestCase):
     q_grad = q.grad
     k_grad = k.grad
     v_grad = v.grad
-    
+
     dev_ids = ','.join(map(str, range(n_devices)))
     self.assertEqual(
         torch_xla._XLAC._get_xla_sharding_spec(q_grad),
--- a/test/test_pallas_spmd.py
+++ b/test/test_pallas_spmd.py
@@ -65,14 +65,15 @@ class PallasTest(unittest.TestCase):
     n_devices = xr.global_runtime_device_count()
     xs.set_global_mesh(xs.Mesh(range(n_devices), (n_devices, 1, 1, 1)))
 
-    q = torch.randn(4, 2, 128, 4).to("xla")
-    k = torch.randn(4, 2, 128, 4).to("xla")
-    v = torch.randn(4, 2, 128, 4).to("xla")
+    q = torch.randn(8, 2, 128, 8).to("xla")
+    k = torch.randn(8, 2, 128, 8).to("xla")
+    v = torch.randn(8, 2, 128, 8).to("xla")
 
-    o = flash_attention(q, k, v, partition_spec=range(n_devices))
+    o = flash_attention(q, k, v, partition_spec=(0, None, None, None))
+    dev_ids = ','.join(map(str, range(n_devices)))
     self.assertEqual(
         torch_xla._XLAC._get_xla_sharding_spec(o),
-        f"{{devices=[{n_devices},1,1,1]0,1,2,3}}")
+        f"{{devices=[{n_devices},1,1,1]{dev_ids}}}")
 
     expected_o = self._attention(q, k, v)
     self.assertTrue(torch.allclose(o.cpu(), expected_o.cpu(), atol=1e-05))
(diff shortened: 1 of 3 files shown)


torch_xla/experimental/custom_kernel.py
```diff
@@ -14,6 +14,87 @@
...
+    mesh,
+    input_specs,
+    output_specs
+):
+    """Map a function over shards of data.
+
+    Note:
+      ``shard_map`` is an experimental API, and still subject to change. For an
+      introduction to sharded data, refer to :ref:`sharded-computation`. For a more
+      in-depth look at using ``shard_map``, refer to `SPMD multi-device parallelism with shard_map`_.
```
pgmoka, line 28: > refer to `SPMD multi-device parallelism with shard_map`
Is this missing a link?

torch_xla/experimental/custom_kernel.py
```diff
@@ -324,27 +380,91 @@ def fa_custom_forward(
...
+    ab: Optional[torch.Tensor], partition_spec: str, mesh: str,
+    ctx_grad: List[bool]
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor,
+           torch.Tensor, torch.Tensor]:
+  partition_spec = eval(partition_spec)
+  mesh = xs.get_global_mesh() or Mesh.from_str(mesh)
+
+  q_full_shape = None
+
+  # Suprisingly, any tensor that is input to the custom_op decorated function will show
```
pgmoka, line 406: Is the an Issue to verify this?

torch_xla/experimental/custom_kernel.py
```diff
@@ -14,6 +14,87 @@
...
+            if axis_sharding is None:
+                new_size = axis_size
+            else:
+                if isinstance(axis_sharding, str):
+                    mesh_mult = mesh_name_to_size[axis_sharding]
+                else:
+                    # tuple or list
+                    mesh_mult = math.prod(mesh_name_to_size[a] for a in axis_sharding)
+                    
+                if mesh_mult is not None:
```
pgmoka, line 72: Just to clear my understanding. I can see this happening in the`mesh_name_to_size[axis_sharding]` case. Can this happen in the math.prod case?

qihqi: Good point. It won't be meaningful but yes it can happen.

torch_xla/experimental/custom_kernel.py
qihqi: yes, done.

torch_xla/experimental/custom_kernel.py
qihqi: not that i am aware of... Also I don't know why it matter to be honest :(

torch_xla/experimental/custom_kernel.py
```diff
@@ -15,6 +15,98 @@
...
+    res = func(*new_args)
+    if isinstance(res, tuple):
+      res_updated = []
+      for i, r in enumerate(res):
+        if isinstance(r, torch.Tensor):
+          assert str(r.device).startswith('xla'), f'{i}th device is {r.device}'
+          assert len(r.shape) == len(
+              output_specs[i]
+          ), f'{i}th shape is {r.shape}, sharding is {output_specs[i]}'
+      return tuple(
```
tengyifei, line 96: This long expression is pretty hard to read. Maybe we could write it like

```
for a, spec in zip(res, output_specs):
  if isinstance(a, torch.Tensor) and spec is not None:
    res_updated.append(xs.disable_manual_sharding(a, spec, _full_shape(a, spec), mesh=mesh).global_tensor)
  else:
    res_updated.append(a)
return tuple(a)
```

torch_xla/experimental/custom_kernel.py
```diff
@@ -370,25 +503,40 @@ def _pad_to_block_size(
...
+    kv_full_shape: List[int], ab_full_shape: Optional[List[int]],
+    ctx_grad: List[bool]
 ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
 
   from jax.experimental.pallas.ops.tpu.flash_attention import _flash_attention_bwd_dq, _flash_attention_bwd_dkv
   grad_q = grad_k = grad_v = grad_ab = segment_ids = None
 
+  num_batches = None
+  batch_size = None
+  if len(q.shape) == 5:
```
tengyifei, line 521: Is it possible to generalize is over an arbitrary number of leading free dimensions? By which I mean:

```
# Returns a pair of reshape functions. Returns a pair of no-op if q is already 4D
reshape_to_4d, undo_reshape_to_4d = maybe_reshape(q.shape)
grad_output = reshape_to_4d(grad_output)
...


# Later
grad_q = undo_reshape_to_4d(grad_q)
...
```

And then within `maybe_reshape`, we'd remember two tuples, `free_dims`, `used_dims`, which correspond to the independent dimensions and the `[SEQ, SEQ, ATTN]` part respectively.

Then the logic has less branching and maybe we could also unit test it.

tengyifei (changes requested):
Makes sense to refactor flash_attention to be shard_map(local_flash_attention)

test/test_pallas_spmd.py
```diff
@@ -65,14 +65,39 @@ def test_flash_attention_spmd_data_parallel(self):
...
+    dev_ids = ','.join(map(str, range(n_devices)))
     self.assertEqual(
         torch_xla._XLAC._get_xla_sharding_spec(o),
-        f"{{devices=[{n_devices},1,1,1]0,1,2,3}}")
+        f"{{devices=[{n_devices},1,1,1]{dev_ids}}}")
+
+    expected_o = self._attention(q, k, v)
+    self.assertTrue(torch.allclose(o.cpu(), expected_o.cpu(), atol=1e-05))
+
+  @unittest.skipIf(xr.device_type() != 'TPU' or tpu.version() < 3,
```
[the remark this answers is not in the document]
qihqi: If we do that, and this check fails, then the test is skipped. In that case we actually would like the test to fail  (maybe then disable it ourselves if we cant get machines with >=4 chips, but that is an active decision).

torch_xla/experimental/custom_kernel.py
```diff
@@ -15,6 +15,98 @@
 _XLA_USE_BF16 = os.environ.get("XLA_USE_BF16", "0") == "1"
 
 
+def _shard_map(func, mesh, input_specs, output_specs):
+  """Map a function over shards of data.
+
+    Note:
+      ``shard_map`` is an experimental API, and still subject to change. For an
+      introduction to sharded data, refer to :ref:`sharded-computation`. For a more
```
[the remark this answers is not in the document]
qihqi: removed. the comment was copied from jax side

test/test_pallas_spmd.py
```diff
@@ -65,14 +65,39 @@ def test_flash_attention_spmd_data_parallel(self):
...
+  def test_flash_attention_spmd_data_parallel_5d(self):
+    n_devices = xr.global_runtime_device_count()
+    xs.set_global_mesh(
+        xs.Mesh(
+            range(n_devices), (n_devices // 2, 2, 1, 1, 1),
+            ('fsdp', 'dp', 'a', 'b', 'c')))
+
+    q = torch.randn(4, 2, 2, 128, 4).to("xla")
+    k = torch.randn(4, 2, 2, 128, 4).to("xla")
+    v = torch.randn(4, 2, 2, 128, 4).to("xla")
```
pgmoka, line 93: In `test_flash_attention_spmd_data_parallel`, , `(8, 2, 128, 8)` is used. Would the equivalent here be something like `(8, 2, 2, 128, 8)`?

test/test_pallas_spmd.py
```diff
@@ -65,14 +65,39 @@ def test_flash_attention_spmd_data_parallel(self):
...
+
+    o = flash_attention(
+        q, k, v, partition_spec=('fsdp', 'dp', None, None, None))
+    dev_ids = ','.join(map(str, range(n_devices)))
+    self.assertEqual(
+        torch_xla._XLAC._get_xla_sharding_spec(o),
+        f"{{devices=[{n_devices//2},2,1,1,1]{dev_ids}}}")
```
pgmoka, line 100: I would add an error string here. This seems complex enough that it is not self-apparent

test/test_pallas_spmd.py
```diff
@@ -169,9 +197,11 @@ def test_flash_attention_wrapper_segment_ids_spmd(self):
         segment_ids_xla,
         segment_ids.to("xla"),
         partition_spec=("data", None, None, None))
+    n_devices = xr.global_runtime_device_count()
+    dev_ids = ','.join(map(str, range(n_devices)))
     self.assertEqual(
         torch_xla._XLAC._get_xla_sharding_spec(o),
-        f"{{devices=[{xr.global_runtime_device_count()},1,1,1]0,1,2,3}}")
+        f"{{devices=[{xr.global_runtime_device_count()},1,1,1]{dev_ids}}}")
```
pgmoka, line 204: I would add an error string here. This seems complex enough that it is not self-apparent

torch_xla/experimental/custom_kernel.py
qihqi: good point.

torch_xla/experimental/custom_kernel.py
```diff
@@ -319,32 +389,95 @@ def fa_custom_forward(
...
+    if l is not None:
+      l = l.reshape(num_batches, batch_size, *l.shape[1:])
+    if m is not None:
+      m = m.reshape(num_batches, batch_size, *m.shape[1:])
+
+  return o, l, m
+
+
+@custom_op("xla::fa_custom_forward", mutates_args=())
+def fa_custom_forward(
```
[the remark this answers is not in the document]
qihqi: We did not change this function's behavior. Just make it to accept a larger range of input. So this is bug fix in some sense.

test/test_pallas_spmd.py
qihqi: (8, 2, 2, 128, 8) would also work.
In general, the axis to shard has to be multiple of number of devices in this axis.
In this case the mesh is (n_devices // 2, 2, 1,1,1) so it's 4,2, 1,1,1; therefore having 4,2 as leading dim is sufficient.

For TPUs with 4 devices it would be 2,2,1,1,1 which 4,2 is still multiple-of. Therefore this configuration is general.

test/test_pallas_spmd.py
qihqi: hmmm pretty hard to write a more meaningful error message other than "hey your sharding spec is wrong".

torch_xla/experimental/custom_kernel.py
```diff
@@ -15,6 +15,98 @@
...
+          mesh_mult = mesh_name_to_size[axis_sharding]
+        else:
+          # tuple or list
+          mesh_mult = math.prod(mesh_name_to_size[a]
+                                for a in axis_sharding
+                                if mesh_name_to_size[a] is not None)
+
+        if mesh_mult is not None:
+          new_size = axis_size * mesh_mult
+      result_shape.append(new_size)
```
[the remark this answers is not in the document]
qihqi: did an attempt, ptal

test/test_pallas_spmd.py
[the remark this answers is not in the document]
tengyifei: Ack

test/test_pallas_spmd.py
```diff
@@ -264,12 +302,12 @@ def test_cross_flash_attention_wrapper_segment_ids_spmd(self):
...
-    k = torch.randn(3, 2, 128, 4)
-    v = torch.randn(3, 2, 128, 4)
-    zeros = torch.zeros(3, 32)
-    kv_segment_ids = torch.cat([zeros, zeros + 1, zeros + 2, zeros + 3], dim=1)
-    q_segment_ids = torch.ones(3, q.shape[2], dtype=torch.float32)
+    q = torch.randn(8, 2, 1024, 4)
+    k = torch.randn(8, 2, 128, 4)
+    v = torch.randn(8, 2, 128, 4)
+    zeros = torch.zeros(8, 32)
+    kv_segment_ids = torch.cat([zeros, zeros + 1, zeros + 2, zeros + 4], dim=1)
```
tengyifei, line 309: I'm not familiar with the kv segment ids. Wondering did you intend to change zero+3 to zero+4?

tengyifei (approved):

tengyifei:
Still need to fix TPU test which looks relevant

test/test_pallas_spmd.py
qihqi: that was unintended indeed. But it shouldn't matter.

torch_xla/experimental/custom_kernel.py
```diff
@@ -15,6 +15,98 @@
...
+        of func. kwarg is not supported yet
+      out_specs: a pytree with :class:`~tuple[tuple[str]]`, with the same length
+        as the number of outputs
+
+    Returns:
+      A callable that applies the input function ``f`` across data sharded according to
+      the ``mesh`` and ``out_specs``.
+
+    Reference:
+      This function is identical Jax's shard_map:
```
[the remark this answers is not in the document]
qihqi: done

test/test_pallas_spmd.py
qihqi: ditto

PR #8733, opened 2025-02-21
Improve the ragged kernel benchmarking script.

vanbasten23:


test/benchmarks/test_ragged_paged_attention_benchmark.py
```diff
@@ -100,18 +98,16 @@ def benchmark(args):
   ]
   num_heads = (4, 4)
   head_dim = 128
-  dtype = jnp.float32
+  dtype = jnp.bfloat16
   page_size = 16
   num_pages = 32768
-  num_queries_per_block = 128
+  num_queries_per_block = 16
```
lsy323, line 104: nit: Making these params as cmd args for easier benchmarking automation?

lsy323 (approved):

test/benchmarks/test_ragged_paged_attention_benchmark.py
```diff
@@ -78,13 +75,14 @@ def _ref_ragged_paged_attention(
...
-  if x <= 0:
-    raise ValueError(f"x must be positive. Got {x}")
-  return 2**int(np.ceil(np.log2(x)))
+def _get_closest_of_multiple(x, base):
+  return (x + base - 1) // base * base
 
 
 def benchmark(args):
+  if args.profile:
+    assert args.profile_path is not None, "Need to provide a profile path to profile."
```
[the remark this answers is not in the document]
vanbasten23: I see your point. How about only use `args.profile_path`?

test/benchmarks/test_ragged_paged_attention_benchmark.py
```diff
@@ -150,28 +159,68 @@ def benchmark(args):
...
+            v_pages_xla,
+            kv_lens_xla,
+            page_indices_xla,
+            cu_q_lens_xla,
+            num_seqs,
+            num_queries_per_block=num_queries_per_block,
+            num_kv_pages_per_block=num_kv_pages_per_block,
+            use_kernel=True,
+        )
+      elif args.kernel == "ragged-paged-attention":
```
pgmoka, line 223: Readability sugetions: rather than having these two similar calls, create a dictionary such as:
 `paged_attn_op = {"ragged-paged-attention-with-torch-xla-dynamo": compiled_paged_attention, "ragged-paged-attention-with-torch-xla-nondynamo": torch.ops.xla.ragged_paged_attention}`

and then call `paged_attn_op[args.kernel](args...)`

I also think that this makes is so you don't need to create temp like "queries_xla", "k_pages_xla", "v_pages_xla", "kv_lens_xla", "page_indices_xla", and "cu_q_lens_xla"

test/benchmarks/test_ragged_paged_attention_benchmark.py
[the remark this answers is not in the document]
yaochengji: Sounds good!

test/benchmarks/test_ragged_paged_attention_benchmark.py
vanbasten23: Things like "queries_xla" are still needed because we need the torch tensor instead of the jnp arrays..

vanbasten23:
The GPU test failure is irrelevant to this PR.

vanbasten23:
Thanks for the review!

test/benchmarks/test_ragged_paged_attention_benchmark.py
pgmoka: I see. I think the dictionary approach should still be valid to make things a little clearer

PR #8734, opened 2025-02-21
Lower `as_strided_copy` use fast path with `slice` 

zpcore:
When we execute the following two code snippets regarding `flash attention kernel` in custom_kernel.py, they suppose to produce the same result.
**a**. 
```
l, m = (v[..., 0] for v in aux[-2:])
```
**b**. 
```
l, m = aux[-2:]
l = torch.ops.aten.slice(l, -1, 0, 1)
m = torch.ops.aten.slice(m, -1, 0, 1)
```
Both will be lowered through https://github.com/pytorch/xla/blob/1acc987aec541978cd52ea77a027f9028d8148d3/torch_xla/csrc/aten_xla_type.cpp#L882 , the difference is that input argument `stride` and `size` will be one element fewer in code **a** compared with code **b**. With such argument difference, code **a** will be fallback into `aten::take` and this can trigger the following error when we call with SPMD:
```
F0223 07:18:45.157172  842998 hlo_sharding.cc:1024] Check failed: !IsManual() 
```

I plan to check in `test_as_stride_use_slice.py` in this PR. 
Note 1. Failing test `test_scan_layer_aot` is not enabled until https://github.com/pytorch/xla/issues/8742 is resolved.
2. Failing test `test_scan_weight_layer_aot` is not enabled unti https://github.com/pytorch/xla/issues/8753 is resolved

### DIFF
--- a/test/cpp/test_aten_xla_tensor_3.cpp
+++ b/test/cpp/test_aten_xla_tensor_3.cpp
@@ -922,6 +922,9 @@ TEST_F(AtenXlaTensorTest, TestAsStrided) {
         torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
     AllClose(output, xla_output);
   });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterNotChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
 }
 
 TEST_F(AtenXlaTensorTest, TestAsStridedInPlace) {
@@ -938,6 +941,9 @@ TEST_F(AtenXlaTensorTest, TestAsStridedInPlace) {
     AllClose(output, xla_output);
     AllClose(input, xla_input);
   });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterNotChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
 }
 
 TEST_F(AtenXlaTensorTest, TestAsStridedWithOffset) {
@@ -956,6 +962,9 @@ TEST_F(AtenXlaTensorTest, TestAsStridedWithOffset) {
                           /*storage_offset=*/storage_offset);
     AllClose(output, xla_output);
   });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterNotChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
 }
 
 TEST_F(AtenXlaTensorTest, TestAsStridedWithInplaceCopy) {
@@ -970,6 +979,9 @@ TEST_F(AtenXlaTensorTest, TestAsStridedWithInplaceCopy) {
     xla_output.as_strided(size, stride).copy_(xla_grad);
     AllClose(output, xla_output);
   });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterNotChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
 }
 
 TEST_F(AtenXlaTensorTest, TestEmptyStrided) {
@@ -982,6 +994,138 @@ TEST_F(AtenXlaTensorTest, TestEmptyStrided) {
     EXPECT_EQ(output.sizes(), xla_output.sizes());
     EXPECT_EQ(output.strides(), xla_output.strides());
   });
+}
+
+TEST_F(AtenXlaTensorTest, TestAsStridedUseSlice) {
+  torch::Tensor input =
+      torch::rand({16, 32, 24}, torch::TensorOptions(torch::kFloat));
+  std::vector<int64_t> size = {16, 8, 24};
+  std::vector<int64_t> stride = {768, 48, 1};
+  torch::Tensor output =
+      torch::as_strided(input, /*size=*/size, /*stride=*/stride);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_input = CopyToDevice(input, device);
+    torch::Tensor xla_output =
+        torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
+    AllClose(output, xla_output);
+  });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterNotChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
+}
+
+TEST_F(AtenXlaTensorTest, TestAsStridedMismatchLastDimUseSlice) {
+  torch::Tensor input =
+      torch::rand({16, 32, 24}, torch::TensorOptions(torch::kFloat));
+  std::vector<int64_t> size = {16, 32}; // 16, 32, 24
+  std::vector<int64_t> stride = {768, 24}; // 768, 24, 1
+  torch::Tensor output =
+      torch::as_strided(input, /*size=*/size, /*stride=*/stride);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_input = CopyToDevice(input, device);
+    torch::Tensor xla_output =
+        torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
+    AllClose(output, xla_output);
+  });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterNotChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
+}
+
+TEST_F(AtenXlaTensorTest, TestAsStridedMismatchMiddleDimUseSlice) {
+  torch::lazy::MetricsArena::Get()->ResetMetrics();
+  runtime::metrics::ClearMetrics();
+  torch::Tensor input =
+      torch::rand({6, 4, 2, 4}, torch::TensorOptions(torch::kFloat));
+  std::vector<int64_t> size = {6, 2, 4};
+  std::vector<int64_t> stride = {32, 4, 1};
+  torch::Tensor output =
+      torch::as_strided(input, /*size=*/size, /*stride=*/stride);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_input = CopyToDevice(input, device);
+    torch::Tensor xla_output =
+        torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
+    AllClose(output, xla_output);
+  });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterNotChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
+}
+
+
+
+TEST_F(AtenXlaTensorTest, TestAsStridedMismatchDimWithOffset) {
+  torch::lazy::MetricsArena::Get()->ResetMetrics();
+  runtime::metrics::ClearMetrics();
+  torch::Tensor input =
+      torch::rand({6, 4, 2, 4}, torch::TensorOptions(torch::kFloat));
+  std::vector<int64_t> size = {6, 2, 4};
+  std::vector<int64_t> stride = {32, 4, 1};
+  torch::Tensor output =
+      torch::as_strided(input, /*size=*/size, /*stride=*/stride, 1);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_input = CopyToDevice(input, device);
+    torch::Tensor xla_output =
+        torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride, 1);
+    AllClose(output, xla_output);
+  });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
+}
+
+TEST_F(AtenXlaTensorTest, TestAsStridedMultipleMismatchDimWithOffset) {
+  torch::lazy::MetricsArena::Get()->ResetMetrics();
+  runtime::metrics::ClearMetrics();
+  torch::Tensor input =
+      torch::rand({6, 4, 2, 4}, torch::TensorOptions(torch::kFloat));
+  std::vector<int64_t> size = {3, 2, 4};
+  std::vector<int64_t> stride = {16, 4, 1};
+  torch::Tensor output =
+      torch::as_strided(input, /*size=*/size, /*stride=*/stride);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_input = CopyToDevice(input, device);
+    torch::Tensor xla_output =
+        torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
+    AllClose(output, xla_output);
+  });
+  for (auto& name : torch_xla::runtime::metrics::GetCounterNames()) {
+    std::cout << name << std::endl;
+  }
+  std::cout << std::endl;
+  for (auto& name : torch::lazy::GetCounterNames()) {
+    std::cout << name << std::endl;
+  }
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
+}
+
+TEST_F(AtenXlaTensorTest, TestAsStridedMultipleDimMismatch) {
+  torch::lazy::MetricsArena::Get()->ResetMetrics();
+  runtime::metrics::ClearMetrics();
+  torch::Tensor input =
+      torch::rand({6, 4, 2, 4}, torch::TensorOptions(torch::kFloat));
+  std::vector<int64_t> size = {6, 4, 1, 2};
+  std::vector<int64_t> stride = {32, 8, 8, 2};
+  torch::Tensor output =
+      torch::as_strided(input, /*size=*/size, /*stride=*/stride);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_input = CopyToDevice(input, device);
+    torch::Tensor xla_output =
+        torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
+    AllClose(output, xla_output);
+  });
+  for (auto& name : torch_xla::runtime::metrics::GetCounterNames()) {
+    std::cout << name << std::endl;
+  }
+  std::cout << std::endl;
+  for (auto& name : torch::lazy::GetCounterNames()) {
+    std::cout << name << std::endl;
+  }
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
 }
 
 TEST_F(AtenXlaTensorTest, TestAvgPool2DBackward) {
--- a/test/cpp/test_aten_xla_tensor_3.cpp
+++ b/test/cpp/test_aten_xla_tensor_3.cpp
@@ -1087,13 +1087,6 @@ TEST_F(AtenXlaTensorTest, TestAsStridedMultipleMismatchDimWithOffset) {
         torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
     AllClose(output, xla_output);
   });
-  for (auto& name : torch_xla::runtime::metrics::GetCounterNames()) {
-    std::cout << name << std::endl;
-  }
-  std::cout << std::endl;
-  for (auto& name : torch::lazy::GetCounterNames()) {
-    std::cout << name << std::endl;
-  }
   ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
   ExpectCounterChanged("xla::take", cpp_test::GetIgnoredCounters());
   ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
@@ -1114,13 +1107,6 @@ TEST_F(AtenXlaTensorTest, TestAsStridedMultipleDimMismatch) {
         torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
     AllClose(output, xla_output);
   });
-  for (auto& name : torch_xla::runtime::metrics::GetCounterNames()) {
-    std::cout << name << std::endl;
-  }
-  std::cout << std::endl;
-  for (auto& name : torch::lazy::GetCounterNames()) {
-    std::cout << name << std::endl;
-  }
   ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
   ExpectCounterChanged("xla::take", cpp_test::GetIgnoredCounters());
   ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
--- a/test/cpp/test_aten_xla_tensor_3.cpp
+++ b/test/cpp/test_aten_xla_tensor_3.cpp
@@ -1014,6 +1014,24 @@ TEST_F(AtenXlaTensorTest, TestAsStridedUseSlice) {
   ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
 }
 
+TEST_F(AtenXlaTensorTest, TestAsStridedUseSliceSizeReduce) {
+  torch::Tensor input =
+      torch::rand({16, 32, 24}, torch::TensorOptions(torch::kFloat));
+  std::vector<int64_t> size = {16, 8, 24};
+  std::vector<int64_t> stride = {768, 24, 1};
+  torch::Tensor output =
+      torch::as_strided(input, /*size=*/size, /*stride=*/stride);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_input = CopyToDevice(input, device);
+    torch::Tensor xla_output =
+        torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
+    AllClose(output, xla_output);
+  });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
+  ExpectCounterNotChanged("xla::take", cpp_test::GetIgnoredCounters());
+  ExpectCounterChanged("xla::as_strided_copy", cpp_test::GetIgnoredCounters());
+}
+
 TEST_F(AtenXlaTensorTest, TestAsStridedMismatchLastDimUseSlice) {
   torch::Tensor input =
       torch::rand({16, 32, 24}, torch::TensorOptions(torch::kFloat));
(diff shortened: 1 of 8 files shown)


torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -883,14 +883,109 @@ at::Tensor XLANativeFunctions::as_strided_copy(
...
   const at::Tensor& base = bridge::GetXlaTensor(self)->Base();
   at::Tensor tensor = base.defined() ? base : self;
 
-  // Fast path: PyTorch/XLA implementation for as_strided works only with
-  // non-overlapping and dense tensors.
+  // Optimize: decide if we can use `slice` to replace `as_strided` to avoid the
+  // copy. To use `slice`, the following conditions must be satisfied:
+  // - `stride` must be: [..., dim[-2]*dim[-1], dim[-1], 1] and one of them
+  //   (assume stride[X]) can be multiplied with a constant K.
+  // - `size` must be: [dim[0], dim[1], ...], and size[X]*K <= dim[X]. å
```
pgmoka, line 898: NIT: Typo ` å` at end of bullet point

test/cpp/test_aten_xla_tensor_3.cpp
pgmoka, line 1: I am not sure if this might be applicable. For the new tests, could we do something like https://pypi.org/project/parameterized/ or some other table drive test methodology to simplify tests?

torch_xla/experimental/custom_kernel.py
```diff
@@ -333,6 +333,12 @@ def fa_custom_forward(
     assert isinstance(o, list)
     o, *aux = o
     l, m = (v[..., 0] for v in aux[-2:])
+    # TODO(tengyifei,zpcore): check to use the following l, m output once
+    # https://github.com/pytorch/xla/issues/8683 is fixed
+
+    # l, m = aux[-2:]
+    # l = torch.ops.aten.slice(l, -1, 0, 1)
+    # m = torch.ops.aten.slice(m, -1, 0, 1)
```
pgmoka, line 341: What is the purpose of these lines? If it is no longer necessary, we could remove them

test/test_as_stride_use_slice.py
pgmoka, line 1: QQ: With the new `test/test_as_stride_use_slice.py` file I don't see any modifications to BUILD. I see the test is called by main. Should there not be a modification to https://github.com/pytorch/xla/blob/master/test/cpp/BUILD to create a test target to this file?

torch_xla/experimental/custom_kernel.py
zpcore: Removed since we have already confirmed that with the proper lowering, they produce the same output.

test/cpp/test_aten_xla_tensor_3.cpp
zpcore: Thanks, that's for the python test.  I didn't find the parameterized test in C++ for the repo. Did you happen to find some?

test/test_as_stride_use_slice.py
zpcore: Python TPU test is triggered by this file: https://github.com/pytorch/xla/blob/79d1e86b7e99dc055f30f3a69aa71471c16f8dfe/test/tpu/run_tests.sh. I didn't add enable test there because the python test is failing at this time...

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+@custom_strided_and_slice_backward.register_fake
+def custom_strided_and_slice_backward_fake(
+    grad_output: torch.Tensor, input: torch.Tensor, use_aten_slice: bool,
+    needs_input_grad: bool) -> torch.Tensor:
+  return torch.empty_like(input)
+
+
+class AsStridedTest(parameterized.TestCase):
+
+  def fake_fa_wrapper(self, use_scan):
```
tengyifei, line 146: nit: for better readability, we can move this to a separate test class like `ScanFlashAttentionTest`. Then this test function can be relocated somewhere around line 68 and we don't have to jump up and down when working on the code anymore :)

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+  def __init__(self):
+    super(StridedAndSlice, self).__init__()
+
+  def forward(self, input, use_aten_slice=True, use_custom_op=False):
+    assert input.dim() > 1
+    if use_custom_op:
+      return custom_strided_and_slice_forward(input, use_aten_slice)
+    else:
+      if not use_aten_slice:
+        output = input[..., 0].squeeze(-1)
```
tengyifei, line 80: we don't need to squeeze again when using the `input[..., 0]` because the `0` indexing will take away a dimension. The reason we need to squeeze in `aten.slice` code path is that that one is doing something like `input[..., 0:1]`.

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+    input = torch.randn((2, 4, 256, 256), device="cpu").requires_grad_()
+    if use_xla:
+      ss.to(xm.xla_device())
+      input = input.to(xm.xla_device())
+    return ss(input, use_aten_slice)
+
+  def custom_op_strided_wrapper(self, input, use_aten_slice):
+    return StridedAndSliceWithCustomOp.apply(input, use_aten_slice)
+
+  @unittest.skipIf(xr.device_type() != 'TPU' or tpu.version() < 3,
```
tengyifei, line 165: Ditto, we can reorganize this to around line 68 or below.

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+                   "This test only works on TPUv3+.")
+  @parameterized.named_parameters(("use_scan_True", True),
+                                  ("use_scan_False", False))
+  def test_scan_layer_aot(self, use_scan):
+    output = self.fake_fa_wrapper(use_scan)
+    torch_xla.sync()
+    self.assertFalse(torch.isnan(output).any())
```
tengyifei, line 172: Does this test pass if you comment out this line of NaN assertion? I filed https://github.com/pytorch/xla/issues/8742 to track the NaN error. I think it's worth running this test in CI if the NaN is the only problem. Then we could comment out this line and add a `# TODO(https://github.com/pytorch/xla/issues/8742): Fix NaN` here.

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+      ("use_aten_slice_False", False),
+  )
+  def test_pure_as_strided(self, use_aten_slice):
+    torch.manual_seed(12)
+    cpu_output = self.pure_strided_wrapper(
+        use_xla=False, use_aten_slice=use_aten_slice)
+    torch.manual_seed(12)
+    xla_output = self.pure_strided_wrapper(
+        use_xla=True, use_aten_slice=use_aten_slice)
+    self.assertTrue(torch.allclose(cpu_output, xla_output.cpu(), atol=1e-4))
```
tengyifei, line 188: `torch.testing.assert_close(cpu_output, xla_output.cpu())`

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+
+  @unittest.skipIf(xr.device_type() != 'TPU' or tpu.version() < 3,
```
tengyifei, line 190: It should be possible to run this test on all hardware yeah? It doesn't use flash attention.

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+                   "This test only works on TPUv3+.")
+  @parameterized.named_parameters(("use_scan_True", True),
+                                  ("use_scan_False", False))
+  def test_scan_layer_aot(self, use_scan):
+    output = self.fake_fa_wrapper(use_scan)
+    torch_xla.sync()
+    self.assertFalse(torch.isnan(output).any())
+
+  # compare torch native against xla aten.slice/aten.as_strided
+  @unittest.skipIf(xr.device_type() != 'TPU' or tpu.version() < 3,
```
tengyifei, line 175: It should be possible to run this test on all hardware yeah? It doesn't use flash attention.

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+                   "This test only works on TPUv3+.")
+  @parameterized.named_parameters(("use_scan_True", True),
+                                  ("use_scan_False", False))
+  def test_scan_layer_aot(self, use_scan):
+    output = self.fake_fa_wrapper(use_scan)
+    torch_xla.sync()
+    self.assertFalse(torch.isnan(output).any())
+
+  # compare torch native against xla aten.slice/aten.as_strided
```
tengyifei, line 174: nit: this comment is better moved to a method doc comment. Then it will show up on CI when failed:

```
def test_pure_as_strided(self, use_aten_slice):
  """Compare torch native against xla aten.slice/aten.as_strided"""
```

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -883,14 +883,113 @@ at::Tensor XLANativeFunctions::as_strided_copy(
...
   TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  XLA_CHECK(size.size() == stride.size())
+      << "mismatch in length of size and stride";
   // Retrieve the base tensor, if there's one.
   // This function actually operates on the tensor's storage. Since XLA does not
   // expose the actual storage, we use the originally allocated tensor.
   const at::Tensor& base = bridge::GetXlaTensor(self)->Base();
   at::Tensor tensor = base.defined() ? base : self;
 
-  // Fast path: PyTorch/XLA implementation for as_strided works only with
```
tengyifei, line 892: Is this line of deleted comment supposed to annotate line 993?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -883,14 +883,113 @@ at::Tensor XLANativeFunctions::as_strided_copy(
...
+        if (skip_dim == -1) {
+          skip_dim = i;
+          K = stride[j] / stride_mul;
+          if (tensor_dim.size() == stride.size() + 1) {
+            // tensor_dim and strict element can potentially shift by one.
+            j++;
+          }
+        } else {
+          // multiple X index found
+          can_replace = false;
```
tengyifei, line 924: A cleaner way is to factor this fast-path into its own function:

```
namespace {

at::Tensor as_strided_eliminate_one_dim_fast_path(
    std::optional<int64_t> storage_offset) {
 ...
// Do this instead of setting `can_replace = false;` and break
return at::Tensor();
}

}  // namespace
```

This way we can early exit at any point in this function. We also don't need to keep checking `can_replace`.

When calling this function, we can do

```
at::Tensor maybe_fast_path = as_strided_eliminate_one_dim_fast_path(...);
if (maybe_fast_path.defined()) {
  return maybe_fast_path;
}

// Rest of the as_strided_copy logic
```

tengyifei (changes requested):

test/cpp/test_aten_xla_tensor_3.cpp
```diff
@@ -984,6 +996,122 @@ TEST_F(AtenXlaTensorTest, TestEmptyStrided) {
...
+  std::vector<int64_t> stride = {768, 48, 1};
+  torch::Tensor output =
+      torch::as_strided(input, /*size=*/size, /*stride=*/stride);
+  ForEachDevice([&](const torch::Device& device) {
+    torch::Tensor xla_input = CopyToDevice(input, device);
+    torch::Tensor xla_output =
+        torch::as_strided(xla_input, /*size=*/size, /*stride=*/stride);
+    AllClose(output, xla_output);
+  });
+  ExpectCounterNotChanged("aten::*", cpp_test::GetIgnoredCounters());
```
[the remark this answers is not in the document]
zpcore: Yes, I remember saw `aten::` counter for some of the newly added tests before the fast path lowering is introduced.

test/test_as_stride_use_slice.py
zpcore: Oh, forgot to remove that.

test/test_as_stride_use_slice.py
zpcore: I think so, we can remove it. Just in case it won't work on old version.

test/test_as_stride_use_slice.py
zpcore: Yes, removed the version constraint

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,233 @@
...
+    value_states = input.clone()
+    attn_output = flash_attention(
+        query_states,
+        key_states,
+        value_states,
+        causal=True,
+        partition_spec=("fsdp", "tensor", None, None),
+    )
+    # below statement is unnecessary for testing the scan and flash attention
+    # kernel
```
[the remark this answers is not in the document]
zpcore: Keep the self.fc reference here since this produces a different error with scan.

test/test_as_stride_use_slice.py
zpcore: Added

torch_xla/csrc/aten_xla_type.cpp
zpcore: Yes, added back

torch_xla/csrc/aten_xla_type.cpp
zpcore: Sound good! Refactored.

test/cpp/test_aten_xla_tensor_3.cpp
pgmoka: I was citing that case as an example. Looking around, I found `INSTANTIATE_TEST_SUITE_P`. Two useful sources on the topic are:
- https://www.sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest.
- https://github.com/google/googletest/blob/main/docs/advanced.md#how-to-write-value-parameterized-tests

I have noticed that we are not using it on our repo at the moment. It might be quite helpful given the nature of testing new operations, but that can be a question for another day.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
...
+    }
+  }
+}
+
 at::Tensor XLANativeFunctions::as_strided_copy(
     const at::Tensor& self, at::IntArrayRef size, at::IntArrayRef stride,
     std::optional<int64_t> storage_offset) {
   TORCH_LAZY_FN_COUNTER_TIMED_TRACING("xla::");
+  XLA_CHECK(size.size() == stride.size())
+      << "mismatch in length of size and stride";
```
pgmoka, line 978: NIT: It would be nice if the error message contained the information that caused the error. https://cplusplus.com/reference/cstdio/printf/ could be helpful for this something like:
```
XLA_CHECK(size.size() == stride.size()) << printf("length of size is %d, expect it to be equal to stride length %d", size.size(), stride.size())
```

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
...
+    std::optional<int64_t> storage_offset) {
+  // Optimize: decide if we can use `slice` to replace `as_strided` to avoid the
+  // copy. To use `slice`, the following conditions must be satisfied:
+  // - `stride` must be: [..., dim[-2]*dim[-1], dim[-1], 1] and one of them
+  //   (assume stride[X]) can be multiplied with a constant K.
+  // - `size` must be: [dim[0], dim[1], ...], and size[X]*K <= dim[X].
+  // In theory we can shuffle element in `stride` and `size` and this can result
+  //   in the transpose of dimensions, but we don't consider this case here.
+  auto tensor_dim = tensor.sizes();
+  if ((tensor_dim.size() == stride.size() ||
```
tengyifei, line 893: We can early exit when the conditions don't apply:

```
if ((tensor_dim.size() != stride.size() && (tensor_dim.size() != stride.size() + 1)) {
  return at::Tensor();
}
if (storage_offset.has_value() && (*storage_offset != 0)) {
  return at::Tensor();
}
```

tengyifei (changes requested):

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
   return out;
 }
 
+at::Tensor as_strided_eliminate_one_dim_fast_path(
```
tengyifei, line 882: We can make this function `static` to hide it from the symbols exposed by the library. It's generally a good practice to make all C++ functions `static` or put them in an anonymous namespace unless we want it to be used by other `.cpp` files.

torch_xla/csrc/aten_xla_type.cpp
zpcore: Sounds good, reformated

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
...
+        if (skip_dim == -1) {
+          skip_dim = i;
+          K = stride[j] / stride_mul;
+          if (tensor_dim.size() == stride.size() + 1) {
+            // tensor_dim and stride element can potentially shift by one.
+            j++;
+          }
+        } else {
+          // multiple X index found
+          return bridge::AtenFromXlaTensor(nullptr);
```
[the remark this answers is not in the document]
zpcore: Fixed

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
   return out;
 }
 
+at::Tensor as_strided_eliminate_one_dim_fast_path(
+    const at::Tensor& tensor, at::IntArrayRef size, at::IntArrayRef stride,
+    std::optional<int64_t> storage_offset) {
+  // Optimize: decide if we can use `slice` to replace `as_strided` to avoid the
+  // copy. To use `slice`, the following conditions must be satisfied:
+  // - `stride` must be: [..., dim[-2]*dim[-1], dim[-1], 1] and one of them
+  //   (assume stride[X]) can be multiplied with a constant K.
```
tengyifei, line 888: What does "can be multiplied with a constant K" mean? Suggest updating the comment to the following:

```
// - `stride` must be: [..., K*dim[-3]*dim[-2]*dim[-1], dim[-2]*dim[-1], dim[-1], 1]
//   where K > 1. We want to find the dimension where the stride does not match
//   the cumulative product of the dimension sizes from right to left. Let's call this
//   dimension X.
```

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
...
+    long stride_mul = 1;
+    long skip_dim = -1;
+    int K = 0;
+    // check `stride`
+    for (long i = l - 1, j = std::min(i, (long)stride.size() - 1); j >= 0;
+         i--, j--) {
+      if (stride_mul != stride[j]) {
+        if (skip_dim == -1) {
+          skip_dim = i;
+          K = stride[j] / stride_mul;
```
tengyifei: Should we check that `stride[j]` can be evenly divided by `stride_mul` and exit if the remainder is not 0?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
...
+          if (size[i] != tensor_dim[j]) {
+            return bridge::AtenFromXlaTensor(nullptr);
+          }
+        }
+      }
+      return bridge::AtenFromXlaTensor(tensor_methods::squeeze(
+          tensor_methods::slice(bridge::GetXlaTensor(tensor), skip_dim, 0, 1,
+                                1),
+          skip_dim));
+    } else if (tensor_dim.size() == stride.size()) {
```
tengyifei: nit: we don't need an `else` here. we can just have to `if (...)` at the same indentation level

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
...
+      long reduce_size_location = -1;
+      for (auto i = 0; i < l; i++) {
```
tengyifei: nit: prefer `long` over `auto`, since nobody can understand what will be type of the variable `i` if we use auto like this due to the complicated C++ rules

tengyifei: nit: prefer `long` over `auto`, since nobody can understand what will be type of the variable `i` if we use auto like this due to the complicated C++ rules

torch_xla/csrc/aten_xla_type.cpp
zpcore: Yes, added the info.

test/cpp/test_aten_xla_tensor_3.cpp
zpcore: We can stress this in a separate PR since we can apply the style to many other tests.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
...
+        if (skip_dim != -1) {
+          if (skip_dim != reduce_size_location ||
+              size[reduce_size_location] * K >
+                  tensor_dim[reduce_size_location]) {
+            return bridge::AtenFromXlaTensor(nullptr);
+          }
+        } else {
+          // we have one dim size reduced but without any step jump regarding
+          // stride.
+          K = 1;
```
tengyifei: I don't understand this case. Does this happen when I'm doing `a[..., 0]`, i.e. i want to get rid of the last dim?

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -879,16 +879,116 @@ at::Tensor& XLANativeFunctions::arange_out(const at::Scalar& start,
...
+        }
+      } else {
+        // size remains the same as tensor, we can't return the same tensor
```
tengyifei: This can also be rewritten to an early exit style:

```
if (reduce_size_location == -1) {
  // ...
  return at::Tensor();
}
```

In summary, it's clearer to hoist any single line "return ..." code path as early as possible. Then we don't need a deep indentation nor do we need an `else` block. In well written complex C++ logic we'll often see something like this:

```
if (foo) {
  return ...;
}
if (bar) {
  return ...;
}
if (baz) {
  return ...;
}
// Now the real stuff begins
...
```

tengyifei (changes requested):

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,230 @@
...
+from torch_xla._internal import tpu
+from torch_xla.experimental.scan_layers import scan_layers
+import torch_xla.distributed.spmd as xs
+from torch_xla.experimental.custom_kernel import flash_attention
+
+from functorch.compile import aot_function, make_boxed_func
+from torch.library import custom_op
+
+
+class AttentionLayers(torch.nn.Module):
```
tengyifei, line 21: nit: I think the name of AttentionLayers and AttentionModule is reversed. Like the smaller one that just wraps flash_attention could be called AttentionModule, and the larger one that consists of a few layers of attention could be called AttentionLayers.

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,230 @@
...
+
+  def pure_strided_wrapper(self, use_xla, use_aten_slice):
+    ss = StridedAndSlice().to("cpu")
+    input = torch.randn((2, 4, 256, 256), device="cpu").requires_grad_()
+    if use_xla:
+      ss.to(xm.xla_device())
+      input = input.to(xm.xla_device())
+    return ss(input, use_aten_slice)
+
+  @unittest.skipIf(xr.device_type() != 'TPU', "This test only works on TPU")
```
tengyifei, line 156: Can we run this test on the CPU too?

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,230 @@
...
+    """compare torch native against xla aten.slice/aten.as_strided"""
+    torch.manual_seed(12)
+    cpu_output = self.pure_strided_wrapper(
+        use_xla=False, use_aten_slice=use_aten_slice)
+    torch.manual_seed(12)
+    xla_output = self.pure_strided_wrapper(
+        use_xla=True, use_aten_slice=use_aten_slice)
+    torch.testing.assert_close(cpu_output, xla_output.cpu())
+
+  @unittest.skipIf(xr.device_type() != 'TPU', "This test only works on TPU")
```
tengyifei, line 171: Can we run this test on the CPU too?

tengyifei: Basically, only stuff like flash attention pallas kernel and SPMD tests require TPU/GPU.

torch_xla/csrc/aten_xla_type.cpp
zpcore: It doesn't need to be 'evenly divided' for j as long as index before j of stride matches with the cumulative product of tensor dim.

test/test_as_stride_use_slice.py
zpcore: Yes, make sense

torch_xla/csrc/aten_xla_type.cpp
zpcore: Nice catch, this should already been captured under `tensor_dim.size() == stride.size() + 1` case. Updated

test/test_as_stride_use_slice.py
zpcore: The whole test has SPMD enabled. Maybe we should have a seperate test file for CPU run without SPMD been triggered at the first place.

torch_xla/csrc/aten_xla_type.cpp
zpcore: Oh, sorry we still need that, this is not for `a[..., 0]` case but can happen when:
```
tensor_dim=[2,3,4]
stride=[12,4,1]
size=[1,3,4]   --> even the stride satisfies the prod cumulative condition, the size get reduced.
```

tengyifei: Got it. If we don't yet have a test for this case, would be good to add one.

test/test_as_stride_use_slice.py
tengyifei: I see. It's probably fine to run the test with SPMD on CPU then. We'll just get 1 device instead of N devices.

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,230 @@
...
+    # self.assertFalse(torch.isnan(output).any())
+
+
+if __name__ == '__main__':
+  logging.getLogger().setLevel(logging.INFO)
+
+  xr.use_spmd()
+  n_devices = xr.global_runtime_device_count()
+  xs.set_global_mesh(
+      xs.HybridMesh(
```
tengyifei, line 223: This probably cannot be a HybridMesh as that's specific to TPUs. I wonder if it works if you use a `get_1d_mesh`

torch_xla/csrc/aten_xla_type.cpp
zpcore: Add `AtenXlaTensorTest.TestAsStridedUseSliceSizeReduce` to stress.

test/test_as_stride_use_slice.py
zpcore: Right, updated

test/test_as_stride_use_slice.py
zpcore: Removed TPU constraints for some tests.

test/test_as_stride_use_slice.py
zpcore: Done

test/test_as_stride_use_slice.py
```diff
@@ -0,0 +1,247 @@
...
+        has_model_weight=has_model_weight_scan, use_scan=False)
+    torch_xla.sync()
+    # TODO(https://github.com/pytorch/xla/issues/8742): Fix NaN
+    # TODO(https://github.com/pytorch/xla/issues/8753): Fix assertion
+    # torch.manual_seed(12)
+    # torch_xla.manual_seed(12)
+    # scan_output = self.fake_fa_wrapper(
+    #     has_model_weight=has_model_weight_scan, use_scan=True)
+    # torch_xla.sync()
+    # torch.testing.assert_close(output.cpu(), scan_output.cpu())
```
zpcore, line 236: @tengyifei , I added the test here to stress the model weight issue in scan.

tengyifei (approved):

test/cpp/test_aten_xla_tensor_3.cpp
pgmoka: I agree. I am thinking of creating an issue for this

pgmoka: Issue created: https://github.com/pytorch/xla/issues/8772

Issue #8371, opened 2024-11-12
TPU Trillium Base Docker Image cannot initialize 


hsebik:
## TPU initialization is failed

When I started tpu v6e-4 TPU Vm with v2-alpha-tpuv6e base image, with pip enviroment and xla updates I can clearly initialized tpus. However when I start to dockerize my pipelie, it fails to initialize TPUs. I tried so much tpu xla base images but I could not achieve to initialize. This happens everytime get device from torch_xla.core.xla_model.xla_device().

I have checked this base images. I guess v2-alpha-tpuv6e configuration is crucial, is there any related base docker image?

> us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.10_tpuvm_20241028

> us-central1-docker.pkg.dev/deeplearning-images/reproducibility/pytorch-tpu-diffusers:v4

## To Reproduce
# DevDockerfile
```dockerfile
FROM us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.10_tpuvm_20241028

# Set environment variables to avoid prompts during installation
ENV DEBIAN_FRONTEND=noninteractive
ENV PYTHONUNBUFFERED=1

RUN apt-get update && apt-get install -y \
    vim \
    curl \
    git \
    bash \
    wget \
    libopenblas-base \
    && rm -rf /var/lib/apt/lists/*

RUN pip3 install  --no-cache-dir --pre torch==2.6.0.dev20241028+cpu torchvision==0.20.0.dev20241028+cpu --index-url https://download.pytorch.org/whl/nightly/cpu
RUN pip install "torch_xla[tpu] @ https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-2.6.0.dev20241028-cp310-cp310-linux_x86_64.whl" -f https://storage.googleapis.com/libtpu-releases/index.html
RUN pip install torch_xla[pallas] -f https://storage.googleapis.com/jax-releases/jax_nightly_releases.html -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
COPY . .
CMD ["python3", "app.py"]
``` 

#app.py 
```python
# Quite simple to reproduce
import torch_xla.core.xla_model as xm

#Hangs in here not initilize tpu.
device = xm.xla_device()
``` 

Both file are in same directory. Generate docker with 
`docker build -f DevDockerfile -t tpu .`
Then run with privileged.
[...]

PR #8735, opened 2025-02-21
Add --network=host to TPU docker build command

bhavya01:
Without `--network=host` command, the docker container cannot access the TPU devices and throws error like:

```
RuntimeError: Bad StatusOr access: UNKNOWN: TPU initialization failed: Failed to connect to [::]:8353
```


zpcore:
Thanks for update the doc. There is a concern that we shouldn't export the network as in https://github.com/pytorch/xla/issues/8371. AFAIK, `--network=host` is needed since v6e. I have no other solution though.

zpcore (approved):
Thanks!

PR #8736, opened 2025-02-22
Change num_seqs type from int to torch.Tensor

vanbasten23:
In vLLM, I see excessive dynamo recompile ([output](https://gist.github.com/vanbasten23/553dbe3d58d8c1530f5daa10155634eb)), for example `L['attn_metadata'].num_seqs == 10 `. Per https://pytorch.org/docs/main/torch.compiler_troubleshooting.html#wrapping-constants-with-tensors, we can wrap the int with tensors to reduce the compilation. So this PR changes the num_seqs type from int to torch.Tensors.

### DIFF
--- a/torch_xla/experimental/custom_kernel.py
+++ b/torch_xla/experimental/custom_kernel.py
@@ -773,13 +773,14 @@ def ragged_paged_attention(
     kv_lens,  # i32[num_tokens]
     page_indices,  # i32[num_tokens, pages_per_sequence]
     cu_q_lens,  # i32[num_tokens + 1]
-    num_seqs,  # int
+    num_seqs,  # i32[]
     num_kv_pages_per_block,
     num_queries_per_block,
     use_kernel=True,
     # TODO(jevinjiang, xiowei): add attn_logits_soft_cap.
     # attn_logits_soft_cap: float | None = None,
 ):  # [batch_size, query_len, num_heads, head_dim]:
+  num_seqs = num_seqs.item()
   assert len(q.shape) == 3, "q should have 3 dimensions."
   if not use_kernel:
     return _ragged_paged_attention_nonkernel(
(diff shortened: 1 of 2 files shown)


torch_xla/experimental/custom_kernel.py
```diff
@@ -773,13 +773,14 @@ def ragged_paged_attention(
...
     cu_q_lens,  # i32[num_tokens + 1]
-    num_seqs,  # int
+    num_seqs,  # i32[]
     num_kv_pages_per_block,
     num_queries_per_block,
     use_kernel=True,
     # TODO(jevinjiang, xiowei): add attn_logits_soft_cap.
     # attn_logits_soft_cap: float | None = None,
 ):  # [batch_size, query_len, num_heads, head_dim]:
+  num_seqs = num_seqs.item()
```
yaochengji, line 783: Why is there a `.item()` here?

vanbasten23: because the argument `num_seqs` has type torch.Tensor, but the jax kernel needs to take an int type. So I need to convert the torch.Tensor to int.
I tried to also make the kernel to also take a jnp.array [here](https://github.com/pytorch/xla/blob/1acc987aec541978cd52ea77a027f9028d8148d3/torch_xla/experimental/pallas_kernels/ragged_paged_attention_kernel.py#L736) but it's not trivial to make the change in the kernel.

yaochengji: There will be a tiny graph here. What will happen if we combine this with torch.compile?

vanbasten23: You meant a graph break? The correctness tests seems fine and it has a torch.compile test. But I'm not sure perf-wise what will happen.

vanbasten23: Each scheduler iteration will have different num_seqs

vanbasten23: I did some experiment https://paste.googleplex.com/4746446904229888 and it seems `.item()` won't create a graph break. I attached the HLO in the link above

yaochengji (approved):
LGTM, thanks!

Issue #8713, opened 2025-02-15
torch.einsum is incorrectly decomposed when wrapped inside a custom op


tengyifei:
## 🐛 Bug

This code

```
import torch
from torch import Tensor
from torch.library import custom_op

@custom_op("xla::custom_linear_forward123", schema="(Tensor input, Tensor weight) -> Tensor", mutates_args=())
def custom_linear_forward123(input: Tensor, weight: Tensor):
    return torch.einsum('...n,mn->...m', input, weight)
```

is a no-op wrapper around `einsum`. However, it changes the behavior significantly in PyTorch/XLA. The einsum is decomposed into a number of expensive transpose ops.

## To Reproduce

Run the notebook https://github.com/tengyifei/playground/blob/master/aot-einsum-3.ipynb locally.

## Expected behavior

The lowering of `custom_linear_forward123` contains an `einsum` op.

## Actual behavior

The lowering of `custom_linear_forward123` becomes

```
IR {
  %0 = f32[] prim::Constant(), xla_shape=f32[]
  %1 = f32[3,3]{1,0} aten::expand(%0), xla_shape=f32[3,3]{1,0}
  %2 = f32[3,3,1]{2,1,0} aten::as_strided(%1), xla_shape=f32[3,3,1]{2,1,0}
  %3 = f32[3,3,1]{2,1,0} aten::as_strided(%2), xla_shape=f32[3,3,1]{2,1,0}
  %4 = f32[1,3,3]{2,1,0} aten::view(%3), xla_shape=f32[1,3,3]{2,1,0}
  %5 = f32[] prim::Constant(), xla_shape=f32[]
  %6 = f32[3,3]{1,0} aten::expand(%5), xla_shape=f32[3,3]{1,0}
  %7 = f32[3,3,1]{2,1,0} aten::as_strided(%6), xla_shape=f32[3,3,1]{2,1,0}
  %8 = f32[3,3,1]{2,1,0} aten::as_strided(%7), xla_shape=f32[3,3,1]{2,1,0}
  %9 = f32[1,3,3]{2,1,0} aten::view(%8), xla_shape=f32[1,3,3]{2,1,0}
  %10 = f32[1,3,3]{2,1,0} aten::matmul(%9, %4), xla_shape=f32[1,3,3]{2,1,0}
  %11 = f32[3,1,3]{2,1,0} aten::view(%10), xla_shape=f32[3,1,3]{2,1,0}
[...]

PR #8739, opened 2025-02-24
Introduce apply_xla_patch_to_nn_linear and test that in a scan

tengyifei:
In order to propagate sharding annotations in 2D sharding, linear layers should be implemented with einsum instead of tranposes/reshapes. Additionally, they need to continue to function inside scan/scan_layers.

For this to work we need three pieces:

- I added a `apply_xla_patch_to_nn_linear` function to replace the implementation of `nn.Linear` with einsum (calling XLAPatchedLinear).
- The XLAPatchedLinear implementation should be wrapped in torch custom ops. That's because AOTAutograd used by scan will decompose all einsums into transposes/reshapes, unless we use `@custom_op` to mark a function as opaque to AOTAutograd.
- Even after wrapping them with `@custom_op`, the einsum is still decomposed into transposes/reshapes due to
https://github.com/pytorch/xla/issues/8713. That's a bug/PyTorch limitation. To workaround this, I added a `_xla_einsum` C++ function that directly builds an einsum given XLA tensors, skipping over any PyTorch dispatcher complexity.

Added a test that demonstrates how `nn.Linear` layers by default flattens any non-contracting dims, and how we could avoid that with `apply_xla_patch_to_nn_linear`.

### DIFF
--- a/torch_xla/distributed/spmd/xla_sharding.py
+++ b/torch_xla/distributed/spmd/xla_sharding.py
@@ -1,15 +1,17 @@
 import collections
 from collections.abc import Generator, MutableMapping
 import math
-import os
 from collections import OrderedDict, defaultdict
 from dataclasses import dataclass, field
 import torch
+from torch import Tensor
+from torch.library import custom_op
 import torch_xla
 import torch_xla.core.xla_model as xm
 import torch_xla._internal.utils as _utils
 from torch_xla.distributed.spmd import XLAShardedTensor, XLAShard
 import torch_xla.runtime as xr
+import torch_xla.debug.profiler as xp
 
 import numpy as np
 import functools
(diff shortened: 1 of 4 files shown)


torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -663,17 +665,106 @@ def apply(self, t: torch.Tensor):
...
+def _einsum_linear_backward(grad_output: Tensor, input: Tensor, weight: Tensor,
+                            bias: Optional[Tensor],
+                            needs_input_grad_input: bool,
+                            needs_input_grad_weight: bool,
+                            needs_input_grad_bias: bool):
+  with xp.Trace('einsum_linear_backward'):
+    grad_input = grad_weight = grad_bias = None
+
+    if needs_input_grad_input:
+      grad_input = torch_xla._XLAC._xla_einsum('...m,mn->...n',
```
zpcore, line 711: Good to know how this is computed!

zpcore (approved):
LGTM!

pgmoka (commented):
LGTM

Issue #8729, opened 2025-02-20
Checkpointing is not compatible with torch.autograd.grad


rpsilva-aws:
## 🐛 Bug

Currently, checkpointing does not work with torch.autograd.grad:
```
    if not torch.autograd._is_checkpoint_valid():
      raise RuntimeError(
          "Checkpointing is not compatible with .grad() or when an `inputs` parameter"
          " is passed to .backward(). Please use .backward() and do not pass its `inputs`"
          " argument.")
```

The PR was added by Jack some time ago: https://github.com/pytorch/xla/commit/16498c3458b#diff-29c25bb9605e07d3044ee5b3794d34a8a7f8d848caf2fc1376df32782e97d6fc, mostly from https://github.com/pytorch/pytorch/blob/main/torch/utils/checkpoint.py, but with optimization barrier.

However, there have been some efforts to close some of the missing gaps (e.g. https://github.com/pytorch/pytorch/commit/7a411952fbb82cec38da936a7d863da49726699f), adding two different checkpointing variants with `use_reentrant`.

Particularly (https://github.com/pytorch/pytorch/blob/main/torch/utils/checkpoint.py#L440):
```
        use_reentrant(bool):
            specify whether to use the activation checkpoint variant that
            requires reentrant autograd. This parameter should be passed
            explicitly. In version 2.5 we will raise an exception if
            ``use_reentrant`` is not passed. If ``use_reentrant=False``,
            ``checkpoint`` will use an implementation that does not require
            reentrant autograd. This allows ``checkpoint`` to support additional
            functionality, such as working as expected with
            ``torch.autograd.grad`` and support for keyword arguments input into
            the checkpointed function.
```

## To Reproduce

Add `--use_gradient_checkpointing` to `COMMON_GRAD_ACC_ARGS` and re-run the TestSPMDLinearModelGradientAccumulation test with `--use_gradient_accumulation_loop`.

ysiraichi:
Thank you for filing this issue. Let me try and reproduce this error.

ysiraichi:
I managed to reproduce it.
@JackCaoG do you remember why does PyTorch/XLA does not support `use_reentrant=False`?

JackCaoG:
I think `use_reentrant` is a thing that come after I copied the checkpoint code, hence I didn't support it. I guess the answer is I am not aware of any strong reason we don't support it besides I didn't have cycle back then.

PR #8768, opened 2025-02-27
Use .backward() with in-place grad mutations for the GA API

rpsilva-aws:
In this PR, we switch to using `.backward()` instead of `torch.autograd.grad` due to https://github.com/pytorch/xla/issues/8729. We resolve this similarly to scan (https://github.com/rpsilva-aws/xla/blob/master/torch_xla/experimental/scan.py#L232), except in our case, since we have full control over the backward pass, we can explicitly do clone over the input gradients, knowingly that the underlying tensors will be updated in-place.

In short, my understanding of why we need to clone is because once we do an in-place mutation, the resulting IR node will not be a device data. Since we rely on the lowering context to collect all parameters in the mapping (`device_parameter_id_tensor_mapping`), this IR will be exempted, since it'll be an IR op (e.g. %add). This doesn't allow us to capture it as a parameter to other XLA while computations (condition and init). This is a known issue with this and the scan API, and we should eventually find a more robust way around it. If we did so, we could offer a more general and all-purpose while_loop/fori_loop API for users (and use that as part of scan and this gradient accumulation API).

### DIFF
--- a/torch_xla/experimental/gradient_accumulation.py
+++ b/torch_xla/experimental/gradient_accumulation.py
@@ -181,6 +181,15 @@ def _gradient_accumulation_impl(context, body_fn, iterable_tensors, params,
   grads = [param.grad for param in params]
   body_fn_inputs = (init_iterator, init_loss, *fake_iterable_tensors,
                     *fake_carried_tensors, *params, *grads)
+  # TODO - Fake the gradients once we are able to create placeholder tensors.
+  # Since the body is expected to do an in-place mutation of the gradients, we
+  # clone the gradients and use that as an input to the body. This will ensure
+  # that we retain a device data IR node in the graph. The cloned gradient will
+  # be updated to denote an IR operation (e.g. %add), and that can not be
+  # captured as a device data input for the other required computations, namely
+  # the condition and init for the XLA while loop.
+  for param in params:
+    param.grad = param.grad.clone()
   body_result = body_fn(init_iterator, init_loss, tuple(fake_iterable_tensors),
                         tuple(fake_carried_tensors), tuple(params),
                         tuple(grads))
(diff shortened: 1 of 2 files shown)


rpsilva-aws:
Canceled the workflow, this is pending on https://github.com/pytorch/xla/pull/8758.

torch_xla/experimental/gradient_accumulation.py
```diff
@@ -375,10 +384,9 @@ def body_fn(iteri: torch.Tensor, _: torch.Tensor,
     else:
       loss, *carried_tensors = result
     loss /= context.num_gradient_steps
-    gradients = torch.autograd.grad(loss, model_parameters)
-    acc_grads = [prev_grad + grad for prev_grad, grad in zip(grads, gradients)]
-    return (iteri, loss, *iterable_tensors, *carried_tensors, *params,
-            *acc_grads)
+    loss.backward()
+    grads = [param.grad for param in params]
+    return (iteri, loss, *iterable_tensors, *carried_tensors, *params, *grads)
```
bhavya01, line 389: Why do we need the grads variable in the body_fn if we are doing an in-place update? params argument should have everything we need right?

rpsilva-aws: Indeed. When tracing the body, it'll identify the XLA device data inputs for the gradients, and so, we need to make sure that those are part of the output (T -> T) for both the body and condition - so I added those explicitly since it was clearer when working with `torch.autograd.grad`. Now that we have changed to `.backward()` and in-place, we should ideally leverage the hoisted vars to achieve this automatically. This can be out-of-scope in this PR (unblocking chkpt), since we should eventually revamp/unify most of the internals with while_loop/fori_loop/scan as we better understand the existing LTC limitations/complications with the XLA while op. I can create a follow-up issue for this, thanks for raising it.

rpsilva-aws:
Kind reminder @tengyifei @bhavya01 - I'll do a follow-up to leverage @tengyifei 's https://github.com/pytorch/xla/pull/8785, so the clone would simultaneously be for just staging (separating the IR that is fed to the body ctx, and the one that is used in the mapping).

bhavya01 (approved):

Issue #8742, opened 2025-02-25
Scan + flash attention kernel = NaN


tengyifei:
## 🐛 Bug

When scanning over modules that contain flash attention pallas kernels, the output contains NaN.

## To Reproduce

Run the test in [1]. This test will fail and the output of `fake_fa_wrapper` contains NaN when `use_scan=True`

## Expected behavior

The output of `fake_fa_wrapper` should be the same regardless if `use_scan=True` or `False`.

[1]: https://github.com/pytorch/xla/pull/8734/files#diff-fd0b937d2fd231d227ae1901d62379ae35208c19b3e90f4188a1e272764eaae4R172

tengyifei:
cc @zpcore

ysiraichi:
cc @vanbasten23

tengyifei:
Interestingly this problem happens only in the combination of scan + flash attention + SPMD.

tengyifei:
## First finding: scan result is non-deterministic despite the same HLO

In this experiment I dump the HLO from both the for loop and the scan variants. We'll see that the HLOs remain the same across runs. However, the output from the scan version keeps changing. The tensor is probably invalid, or there are memory errors.

[test_as_stride_bug_2.log](https://github.com/user-attachments/files/19016647/test_as_stride_bug_2.log)
[test_as_stride_bug.log](https://github.com/user-attachments/files/19016646/test_as_stride_bug.log)

tengyifei:
## Second finding: error goes away if we don't activate SPMD

In this experiment I turned off SPMD in both for loop and scan. We can see that the for loop output stays unchanged (as expected), and the scan output is correct as well. So we know scan + FA works but not scan + FA + SPMD.

[test_as_stride_bug_non_spmd.log](https://github.com/user-attachments/files/19016677/test_as_stride_bug_non_spmd.log)

tengyifei:
I ran the HLOs internally (http://shortn/_Zx4Bw7BQOu) and I suspect the problem comes from the additional HLO input parameters used in the scan version.

For loop HLO: https://gist.github.com/tengyifei/db356669e93f6da43945c10b064a5136

Scan HLO: https://gist.github.com/tengyifei/02158b0de24540ce16cc0f73f7f26f3b

We'll see that the Scan HLO has three extra input parameters. Those are apparently hoisted vars (i.e., the decoder layer ends up referencing some global tensor of that shape).

tengyifei:
[...]

PR #8769, opened 2025-02-28
[scan] Make sure inputs into `fn` are not `device_data` IR nodes

tengyifei:
Fixes https://github.com/pytorch/xla/issues/8742.

This is to workaround a limitation of `mark_sharding`, which replaces the innards of the tensors it operates on. In other words, `mark_sharding` is an in-place operation as opposed to a transform like found in JAX.

When `fn` contains a `mark_sharding` and the `mark_sharding` operates on one of the carry or xs fake tensors, the original device data will be discarded and a new one will be created in its place. As a result, `fn` will appear to create a few empty tensors internally that are unrelated to the carry and xs fake tensors, and the carry and xs will appear completely unused.

See https://github.com/pytorch/xla/issues/8742 for the bug. In short, if an input into the layer to be scanned is a device data, and that layer does a `mark_sharding` on said input, then the graph capturing in `scan` will fail.

The workaround here is simple and cursed: multiply any `device_data` by 1. This will make sure these tensor don't hold device data IR nodes and will defeat the device data replacement of `mark_sharding`.

Fortunately, XLA simplifies away the multiplication (see [1]) so this should become a no-op by the time it hits the TPU.

[1]: https://github.com/openxla/xla/blob/869f57d0082d7adbb9efc10cc18f51a562fc7bf3/xla/hlo/transforms/simplifiers/algebraic_simplifier.cc#L4755-L4770


test/scan/test_scan_pallas.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class ScanFlashAttentionTest(parameterized.TestCase):
+
+  def fake_fa_wrapper(self, has_model_weight, use_scan):
+    torch.manual_seed(12)
+    torch_xla.manual_seed(12)
+    hidden_states = torch.randn((2, 4, 256, 256)).requires_grad_().to('xla')
+    with xm.xla_device():
+      dm = AttentionLayers(has_model_weight, num_layer=3, use_scan=use_scan)
```
bhavya01, line 76: nit: Rename dm to model?

torch_xla/experimental/scan.py
```diff
@@ -564,7 +564,42 @@ def make_fake_tensor(v: torch.Tensor) -> torch.Tensor:
...
+  def defeat_device_data(v: torch.Tensor) -> torch.Tensor:
+    """
+    Make sure inputs into `fn` are not `device_data` IR nodes.
+
+    This is to workaround a limitation of `mark_sharding`, which replaces
+    the innards of the tensors it operates on. In other words, `mark_sharding`
+    is an in-place operation as opposed to a transform like found in JAX.
+
+    When `fn` contains a `mark_sharding` and the `mark_sharding` operates on one
+    of the carry or xs fake tensors, the original device data will be discarded
```
bhavya01, line 577: Can we elaborate on `device data` will be discarded? What about mark_sharding creates this illusion of new tensor being created? Is it setting new IR value or creating new XLA node?

Probably too much LTC detail but will be very helpful to understand why this happens. I am also trying to learn more about this. Will let you know if I find anything.

test/scan/test_scan_pallas.py
tengyifei: Yes. BTW, this was moved from another filed. I should've called that out. But I'll also rename.

torch_xla/experimental/scan.py
tengyifei: Yes.  I've updated the explanation:

    When `fn` contains a `mark_sharding` and the `mark_sharding` operates on one
    of the carry or xs fake tensors, the original device data will be discarded
    and a new one will be created in its place. That's because `mark_sharding` has
    different code paths depending on if the IR has or doesn't have device data.
    If the IR is an intermediate operation like add or matmul, `mark_sharding` will
    update the sharding annotation. If the IR holds data, `mark_sharding` will
    transfer the data to the TPU in a sharded manner, and update the data object
    in the IR to point to a sharded data object, as can be seen in [2].

    As a result, `fn` will appear to create a few empty tensors internally that
    are unrelated to the carry and xs fake tensors, and the carry and xs will
    appear completely unused.

[2]: https://github.com/pytorch/xla/blob/2675e6892c6f955fc2baf88d85dfdfa72062273c/torch_xla/csrc/xla_sharding_util.cpp#L799-L846

tengyifei: `[2]: https://github.com/pytorch/xla/blob/2675e6892c6f955fc2baf88d85dfdfa72062273c/torch_xla/csrc/xla_sharding_util.cpp#L799-L846`

qihqi (approved):

PR #8787, opened 2025-03-04
Manually register einsum xla

pgmoka:
Do manual registration of XLANativeFunctions::einsum for XLA.

This is necessary because currently PyTorch overwrites the key AutogradXLA registration with a its XLA key registration. While ideally we would be able to resolve this problem, this work around resolves the issue from our end. It is also not possible to use full code generation due to https://github.com/pytorch/xla/pull/8739.

This manual registration relies on the `XLANativeFunctions::einsum` function from `xla/torch_xla/csrc/aten_xla_type.cpp`


pgmoka:
As the overwrite is written, there is no meaningful unit test we can add as we rely on the generated XLANativeFunctions function. I could do something like what `nms_kernel` is doing, and refer to tensor_methods::einsum, but this will require rewriting the conditions from https://github.com/pytorch/xla/blob/17270e2f685a4e071ff027340a9d628bf526a87a/torch_xla/csrc/aten_xla_type.cpp#L1531. This cleanest way to do something like that would be to create a utility function shared by both `XLANativeFunctions::einsum` and our overwrite, and then that might let us test the overwrite. This can be the next step here, or if we think it is unnecessary, we can just do the implementation from this PR.

@tengyifei @ysiraichi: Do y'all have any opinions on this?

pgmoka:
CC: @lsy323

ysiraichi:
One thing you can do is to check for `xla::einsum` in the XLA counters. I believe that, before your PR, it wouldn't be in there, since the `CompositeImplicitAutograd` kernel was called.

pgmoka:
> Not sure how all these commits got into this branch. Usually I rebase the branch on top of the latest master and then force push. This way I only have a single commit in the PR.

I honestly don't know how this happened either. I think I messed something up while fetching the current master to rebase to branch with. I needed to do this to get the latest changes related to `_einsum`. The final state is what I wanted, but it creates this unfortunate commit map on the PR

pgmoka:
Too many conflicts. I accidentally merged from master rather than rebasing, and it caused a bunch of issues. My changes are small enough that I will just carry on in a separate PR. I apologize to the reviewers for the noise

PR #8800, opened 2025-03-05
show case use jax.grad in torch_xla

zpcore:
Demonstrate that we can run jax.grad with `call_jax`. This can be helpful to port the Splash Attention kernel from Jax into torch_xla directly.

### DIFF
--- a/test/test_jax_interop.py
+++ b/test/test_jax_interop.py
@@ -12,18 +12,61 @@ class TestJaxInterop(absltest.TestCase):
     """
     Test that we can call a JAX function from PyTorch/XLA lazy tensor tracing.
     """
-
     dev = xm.xla_device()
-    a = torch.ones((3, 3), device=dev)
+    a = torch.randn((3, 3), device=dev, requires_grad=True)
+    b = torch.randn((3, 3), device=dev, requires_grad=True)
 
-    def f(a, b):
-      import jax.numpy as jnp
-      return a + jnp.sin(b)
+    ##############pytorch function##############
+    def f_torch(a, b):
+      return torch.sum(torch.cos(a) + torch.sin(b))
 
-    b = xb.call_jax(f, (a, a), {}, 'my_test')
+    def f_backward_torch(f, a, b):
+      out = f(a, b)
+      print(out)
+      out.backward()
+      return a.grad, b.grad
+
+    _a = a.clone().requires_grad_(True)
+    _b = b.clone().requires_grad_(True)
+    out_torch = f_torch(_a, _b)
+
+    _a.retain_grad()
+    _b.retain_grad()
+    out_grad_torch = f_backward_torch(f_torch, _a, _b)
+
     torch_xla.sync()
-    torch.testing.assert_close(
-        b, torch.sin(torch.ones(3, 3)) + 1, check_device=False)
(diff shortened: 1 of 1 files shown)


test/test_jax_interop.py
```diff
@@ -12,18 +12,61 @@ def test_call_jax(self):
     """
     Test that we can call a JAX function from PyTorch/XLA lazy tensor tracing.
     """
-
     dev = xm.xla_device()
-    a = torch.ones((3, 3), device=dev)
+    a = torch.randn((3, 3), device=dev, requires_grad=True)
```
tengyifei, line 16: Can we create a separate test for testing calling `jax.grad` from torch_xla? Usually it's a good practice to have a base test for the simplest case (no grad) and then other tests for more advanced cases.

zpcore: Sounds good, brought the original test back.

test/test_jax_interop.py
```diff
@@ -25,6 +24,67 @@ def f(a, b):
     torch.testing.assert_close(
         b, torch.sin(torch.ones(3, 3)) + 1, check_device=False)
 
+  def test_call_jax_grad(self):
+    """
+    Test that we can call a JAX function and retrive grad from PyTorch/XLA lazy
```
tengyifei, line 29: nit: retrieve

test/test_jax_interop.py
```diff
@@ -25,6 +24,67 @@ def f(a, b):
...
+    def f_torch(a, b):
+      return torch.sum(torch.cos(a) + torch.sin(b))
+
+    def f_backward_torch(f, a, b):
+      out = f(a, b)
+      print(out)
+      out.backward()
+      return a.grad, b.grad
+
+    _a = a.clone().requires_grad_(True)
```
tengyifei, line 46: What's the purpose of cloning from `a` to `_a`? Usually variable names like `_a` is reserved for variables that are never read. Also I wonder if we could avoid the cloning and just use `a`. If you wanted to avoid accumulating gradient into `a`, then the right incantation is `a.clone().detach().requires_grad_(True)`.

test/test_jax_interop.py
```diff
@@ -25,6 +24,67 @@ def f(a, b):
...
+    _b = b.clone().requires_grad_(True)
+    out_torch = f_torch(_a, _b)
+
+    _a.retain_grad()
+    _b.retain_grad()
+    out_grad_torch = f_backward_torch(f_torch, _a, _b)
+
+    torch_xla.sync()
+    out_torch.detach()
```
tengyifei, line 55: This detach is a no-op. `detach` returns a new tensor and keeps the original tensor unchanged. This line can probably be removed.

test/test_jax_interop.py
```diff
@@ -25,6 +24,67 @@ def f(a, b):
...
+    out_grad_torch = [g.detach() for g in out_grad_torch]
```
tengyifei, line 56: If the only thing we're doing next is comparing these tensor against the call_jax version, then we probably don't need to detach either. Detach returns a new tensor and also stops the gradient flow. If we're no longer doing any gradient propagation, then it's useless.

tengyifei (changes requested):

PR #8801, opened 2025-03-05
Manually register einsum on xla

pgmoka:
Do manual registration of XLANativeFunctions::einsum for XLA.

This is necessary because currently PyTorch overwrites the key AutogradXLA registration with a its XLA key registration. While ideally we would be able to resolve this problem, this work around resolves the issue from our end. It is also not possible to use full code generation due to https://github.com/pytorch/xla/pull/8739.

This manual registration relies on the `XLANativeFunctions::einsum` function from `xla/torch_xla/csrc/aten_xla_type.cpp`.

Originally being worked on https://github.com/pytorch/xla/pull/8787. Abandoned due to conflicts being complex enough starting a new PR made more sense.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -249,11 +249,12 @@ function run_xla_op_tests3 {
   run_test "$CDIR/test_persistent_cache.py"
   run_test "$CDIR/test_devices.py"
   run_device_detection_test "$CDIR/test_gpu_device_detection.py"
+  run_test "$CDIR/test_manual_xla_registration.py"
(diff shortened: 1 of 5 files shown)


pgmoka:
@ysiraichi You mentioned in https://github.com/pytorch/xla/pull/8787 using the XLA counters. I am not familiar with this. Could you send me a resource, or an example of its usage?

pgmoka:
I am looking into the error here. I will also be looking at checking that there are no re-shapes on tests

tengyifei (approved):
Nice!

Issue #8706, opened 2025-02-13
Fix skipped tests on GPU CI.


ysiraichi:
This is a tracking issue for fixing the tests that had to be skipped in the PR that brought back GPU CI #8593.

- [ ] [`TestXrtDistributedDataParallel.test_ddp_correctness_with_gradient_as_bucket_view`][1]: #8841 
- [x] [`TritonTest`][2]: #8803

[1]: https://github.com/pytorch/xla/blob/42edbe18bcd256869bc52e2562727c0f8516b2e6/test/torch_distributed/test_ddp.py#L44
[2]: https://github.com/pytorch/xla/blob/42edbe18bcd256869bc52e2562727c0f8516b2e6/test/test_triton.py#L247

ysiraichi:
Running `torch_distributed/test_ddp.py` test on my 1-GPU machine I see `test_ddp_correctness_large_net` failing, but `test_ddp_correctness_with_gradient_as_bucket_view` passing. I will try to run this once again in a multi-GPU machine.

ysiraichi:
I didn't realize that `test_ddp_correctness_with_gradient_as_bucket_view` is actually pretty recent. It was introduced in #8521 (January 2, 2025). The test actually fails there, too.

ysiraichi:
Closing in favor of #8841

PR #8803, opened 2025-03-06
Re-enable triton tests.

ysiraichi:
Partial Fix: #8706

This PR modifies the experimental triton support, in order to support newer versions of triton. This is necessary because PyTorch uses a newer triton version (with the old triton versions it would error when inductor was imported).

These changes allow us to re-enable the `test_triton.py` test in CI.

### DIFF
--- a/test/test_triton.py
+++ b/test/test_triton.py
@@ -242,8 +242,6 @@ def _attn_fwd(
   tl.store(O_block_ptr, acc.to(Out.type.element_ty))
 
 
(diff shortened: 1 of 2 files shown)


bhavya01 (approved):
This is great!

zpcore:
Hi @ysiraichi , the triton test fails in the head: https://github.com/pytorch/xla/actions/runs/13822990358/job/38674985477? Can you help make a mix? We will cherry pick the fix into r2.7 branch. Thanks

PR #8804, opened 2025-03-06
Misc fixes

qihqi:


tengyifei (approved):
I remember there's some way these are tested but if this is not covered by an existing test plz add one

tengyifei:
screenshot-2025-03-06-16-45-39

tengyifei:
I'm going to turn off required status checks because otherwise no torchax PR can be merged

PR #8810, opened 2025-03-09
[1/N] Initial implementation of local SPMD support

lsy323:
Before this PR, user must uses all global devices in their SPMD program, this limits the flexibility of running MPMD + SPMD under multi-host environment.

This PR enables local SPMD by setting environment variable `XLA_USE_LOCAL_SPMD`. Local SPMD and global SPMD can be switched in the same python program by switching on and off the environment variable.

### Usage
Example usage is:
```python
xr.use_spmd()
os.environ['XLA_USE_LOCAL_SPMD'] = '1'
local_mesh = Mesh(local_device_ids, mesh_shape, axis_names)
# Run local SPMD program
...

os.environ['XLA_USE_LOCAL_SPMD'] = '0'
global_mesh = Mesh(global_device_ids, global_mesh_shape, axis_names)
# Run global SPMD program
...
```

An example script is [here](https://github.com/lsy323/learning_machine/blob/lsiyuan/local_spmd/local_spmd/vae_example.py)

To make a local SPMD work in multi-host setting, we need to ensure:
1. In the lowered HLO graph, the global ordinals are logical indices (starts from zero), instead of physical device ids.
2. During XLA compilation, XLA needs to be configured to target at only local devices.

#### SPMD Python API:
- Allow create mesh with only local devices.
- Update `_get_tile_assignment` to support local device mesh, so that the device id will start from 0 in the HLO sharding annotation

#### Sharing Utilities:
- Support local SPMD in `GetShardReplicaAndIndicesForDevices`. This is achieved by deriving global ordinal from physical device and SPMD sharding annotation.

#### PJRT computation client:
- (Env var is used here) Configure XLA compilation option to only use local devices for the compiled program when local SPMD is enabled.

### Test:
- Added c++ test for sharding util change for local spmd use case
- Verified on multi-host TPU VM, on an example [script](https://github.com/lsy323/learning_machine/blob/lsiyuan/local_spmd/local_spmd/vae_example.py) of running a small VAE with local SPMD, with different input resolutions on different host. Note that the global SPMD and local SPMD are not bridged, we need to have an API to create global SPMD tensor from local SPMD.

### Future work:
- Avoid using env var for local SPMD control.

### DIFF
--- a/test/cpp/test_xla_sharding.cpp
+++ b/test/cpp/test_xla_sharding.cpp
@@ -222,6 +222,113 @@ TEST_F(XLAShardingTest, ShardTensor) {
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({10, 1, 4, 4, 2}));
 }
 
+TEST_F(XLAShardingTest, ShardTensorLocalMesh) {
+  // Test sharding with a local mesh.
+  std::vector<std::string> devices = {"TPU:8",  "TPU:9",  "TPU:10", "TPU:11",
+                                      "TPU:12", "TPU:13", "TPU:14", "TPU:15"};
+
+  // 1D tiled
+  at::Tensor tensor = at::ones({8}, at::TensorOptions(at::kFloat));
+  xla::Shape tensor_shape =
+      CreateComputationShapeFromTensor(tensor, bridge::GetDefaultDevice());
+  xla::OpSharding sharding =
+      xla::HloSharding::Tile1D(
+          CreateComputationShapeFromTensor(tensor, bridge::GetDefaultDevice()),
+          devices.size())
+          .ToProto();
+  auto sharding_spec =
+      std::make_shared<XLATensor::ShardingSpec>(sharding, tensor_shape);
+  auto shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                          /*padded=*/false);
+  EXPECT_EQ(shards.size(), 8);
+  for (auto shard : shards) {
+    EXPECT_EQ(shard.sizes(), c10::ArrayRef<long>({1}));
+  }
+
+  // 2D tiled, The first dim is halved and the last replicated. The last shard
+  // size should be smaller in dim=1 because it's not evenly divisible.
+  tensor = at::ones({8, 7, 4}, at::TensorOptions(at::kFloat));
+  tensor_shape =
+      CreateComputationShapeFromTensor(tensor, bridge::GetDefaultDevice());
+  xla::Array2D<int64_t> mesh({
+      {0, 1, 2, 3},
+      {4, 5, 6, 7},
+  });
+  sharding = xla::HloSharding::Tile(mesh).ToProto();
+  sharding_spec =
+      std::make_shared<XLATensor::ShardingSpec>(sharding, tensor_shape);
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/false);
+  EXPECT_EQ(shards.size(), 8);
+  EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({4, 2, 4}));
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({4, 1, 4}));
+
+  // 3D tiled, the first dim is replicated and the last halved. The last shard
+  // size should be smaller in dim=1 because it's not evenly divisible.
+  xla::Array3D<int64_t> cube({{{0, 1}, {2, 3}, {4, 5}, {6, 7}}});
+  sharding_spec->sharding = xla::HloSharding::Tile(cube).ToProto();
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/false);
+  EXPECT_EQ(shards.size(), 8);
+  EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({8, 2, 2}));
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 1, 2}));
+
+  // Replicated, all shards should be identical.
+  sharding_spec->sharding = xla::HloSharding::Replicate().ToProto();
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/false);
+  EXPECT_EQ(shards.size(), 8);
+  EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({8, 7, 4}));
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({8, 7, 4}));
+
+  // 4D tiled, the first and second dims are replicated and the last halved. The
+  // last shard size should be smaller in dim=2 because it's not evenly
+  // divisible.
+  tensor = at::ones({1, 8, 7, 4}, at::TensorOptions(at::kFloat));
+  tensor_shape =
+      CreateComputationShapeFromTensor(tensor, bridge::GetDefaultDevice());
+  xla::Array4D<int64_t> tesseract({{{{0, 1}, {2, 3}, {4, 5}, {6, 7}}}});
+  sharding = xla::HloSharding::Tile(tesseract).ToProto();
+  sharding_spec =
+      std::make_shared<XLATensor::ShardingSpec>(sharding, tensor_shape);
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/false);
+  EXPECT_EQ(shards.size(), 8);
+  EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({1, 8, 2, 2}));
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({1, 8, 1, 2}));
+
+  // 4D tiled and padded, all shard sizes should be idential.
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/true);
+  EXPECT_EQ(shards.size(), 8);
+  EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({1, 8, 2, 2}));
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({1, 8, 2, 2}));
+
+  // 5D tiled, the first and second dims are replicated and the last halved. The
+  // last shard size should be smaller in dim=2 because it's not evenly
+  // divisible.
+  tensor = at::ones({10, 1, 8, 7, 4}, at::TensorOptions(at::kFloat));
+  tensor_shape =
+      CreateComputationShapeFromTensor(tensor, bridge::GetDefaultDevice());
+  xla::Array<int64_t> hypercube(std::vector<int64_t>{1, 1, 2, 2, 2});
+  hypercube.FillIota(0);
+  sharding = xla::HloSharding::Tile(hypercube).ToProto();
+  sharding_spec =
+      std::make_shared<XLATensor::ShardingSpec>(sharding, tensor_shape);
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/false);
+  EXPECT_EQ(shards.size(), 8);
+  EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({10, 1, 4, 4, 2}));
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({10, 1, 4, 3, 2}));
+
+  // 5D tiled and padded, all shard sizes should be identical.
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/true);
+  EXPECT_EQ(shards.size(), 8);
+  EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({10, 1, 4, 4, 2}));
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({10, 1, 4, 4, 2}));
+}
+
 TEST_F(XLAShardingTest, ShardTensorMultiHost) {
   std::vector<std::string> devices = {"TPU:4", "TPU:5", "TPU:6", "TPU:7"};
 
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1482,7 +1482,7 @@ void InitXlaModuleBindings(py::module m) {
     if (UseVirtualDevice()) {
       return 1;
     } else {
-      return runtime::GetComputationClient()->GetNumDevices();
+      return runtime::GetComputationClient()->GetNumLocalDevices();
     }
   });
   m.def("_xla_get_all_devices", []() {
@@ -1500,12 +1500,15 @@ void InitXlaModuleBindings(py::module m) {
   m.def("_xla_get_runtime_devices",
         []() { return runtime::GetComputationClient()->GetLocalDevices(); });
   m.def("_xla_num_runtime_devices", []() -> int64_t {
-    return runtime::GetComputationClient()->GetNumDevices();
+    return runtime::GetComputationClient()->GetNumLocalDevices();
   });
   m.def("_xla_get_all_runtime_devices", []() {
     std::vector<std::string> all_devices =
         runtime::GetComputationClient()->GetAllDevices();
     return all_devices;
+  });
+  m.def("_xla_num_global_devices", []() -> int64_t {
+    return runtime::GetComputationClient()->GetNumGlobalDevices();
   });
   m.def(
       "_xla_real_devices",
--- a/torch_xla/csrc/runtime/ifrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/ifrt_computation_client.cc
@@ -613,8 +613,12 @@ IfrtComputationClient::ExecuteReplicated(
   return data_handles;
 }
 
-size_t IfrtComputationClient::GetNumDevices() const {
+size_t IfrtComputationClient::GetNumLocalDevices() const {
   return client_->addressable_device_count();
+}
+
+size_t IfrtComputationClient::GetNumGlobalDevices() const {
+  return client_->device_count();
 }
 
 std::string IfrtComputationClient::GetDefaultDevice() const {
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -334,6 +334,7 @@ ComputationClient::DataPtr PjRtComputationClient::CopyToDevice(
 std::shared_ptr<PjRtComputationClient::PjRtData>
 PjRtComputationClient::ReplicateShardedData(
     const ComputationClient::DataPtr& handle) {
+  std::cout << "PjRtComputationClient::ReplicateShardedData" << std::endl;
   if (auto unsharded_data = std::dynamic_pointer_cast<PjRtData>(handle)) {
     return unsharded_data;
   } else if (auto sharded_data =
@@ -347,7 +348,9 @@ PjRtComputationClient::ReplicateShardedData(
     }
     xla::XlaBuilder builder("ReplicateShardedData");
     xla::Shape shape = sharded_data->shape();
-    builder.SetSharding(sharded_data->GetSharding());
+    xla::OpSharding sharding = sharded_data->GetSharding();
+    builder.SetSharding(sharding);
+    size_t num_partitions = sharding.tile_assignment_devices().size();
 
     // perform a simple identity calculation to reassemble the input as
     // replicated output.
@@ -371,6 +374,7 @@ PjRtComputationClient::ReplicateShardedData(
                          GetCompilationDevices(device, {}), &shape,
                          /*should_wrap_parameter=*/false,
                          /*is_sharded=*/true,
+                         /*computation_num_partitions*/ num_partitions,
                          /*allow_spmd_sharding_propagation_to_output=*/false});
     std::vector<
         std::shared_ptr<torch_xla::runtime::ComputationClient::Computation>>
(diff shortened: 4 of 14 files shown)


lsy323:
I tried hard avoid using env var to control local SPMD enablement. (In this [commit](https://github.com/pytorch/xla/pull/8810/commits/c4aa8542eb00fda59a9281043d5ec2dce7517820#diff-50e54e3e70907c0585ca3656238e13dd3d01feec6bb5be0bd51366420857529b)

My thought is:

1. We need to find a way in `PjrtComputationClient::Compile` to determine the target devices of the XLA program we want to compile and execute.

2. The information of sharding is stored in 2 places:
a) `xla::OpSharding` in the lowered `xla::XlaOp` which has sharding annotation 
b) SPMD mesh

Determining the target device from SPMD mesh is the much more straightforward. However, it's not passed down to LTC and LTC cannot be easily extended to pass the SPMD mesh from Python layer (Because torch_xla work on tensors, attaching mesh to each sharded tensor seems to be redundant)

Therefore we have to derive the target devices from the `xla::OpSharding`.

3. I attempted to derive the SPMD target devices information from the `xla::OpSharding` of each lowered `xla::XlaOp` / `xla::Parameter`. [code ref](https://github.com/pytorch/xla/pull/8810/commits/c4aa8542eb00fda59a9281043d5ec2dce7517820#diff-18148b64fc4fce0cb21b9e79d5fb1f0090e8bfc644843b2be36f5aba637a06b0R352-R367)

lsy323:
The GPU test failing is irrelevant, head is failing with the same error.

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -548,8 +565,9 @@ def mark_sharding(
       >>> xs.mark_sharding(linear.weight, mesh, (None, 1)) # 2-way model parallel
   """
   num_devices = xr.global_runtime_device_count()
+  num_local_devices = xr.addressable_runtime_device_count()
   assert num_devices > 0, "This requires XLA supported device(s)."
-  assert mesh.size() == num_devices, \
+  assert mesh.size() == num_devices or mesh.size() == num_local_devices, \
```
tengyifei, line 570: This seem like a smell. Why do we only check the mesh size during `mark_sharding`? Furthermore, should we check that the mesh is a local mesh only when `XLA_USE_LOCAL_SPMD==1`?

tengyifei: Seems that these checks should be inside the `Mesh` constructor IMO

torch_xla/csrc/xla_sharding_util.cpp
```diff
@@ -85,10 +85,11 @@ std::vector<int64_t> TileAssignmentDimensions(
 // order of the output corresponds to the order of the `devices`, which can be
 // arbitrarily set by the caller.
 std::unordered_map<int, int> build_index_map(
-    const std::vector<std::string>& devices) {
+    const std::vector<std::string>& devices, size_t num_mesh_devices) {
   std::unordered_map<int, int> device_index;
   for (int i = 0; i < devices.size(); ++i) {
-    int global_ordinal = ParseDeviceString(devices[i]).ordinal();
+    int global_ordinal =
+        ParseDeviceString(devices[i]).ordinal() % num_mesh_devices;
```
tengyifei, line 92: Could you explain what's the `%` doing here? Like why do we take the remainder

tengyifei:
> We need to find a way in PjrtComputationClient::Compile to determine the target devices

Do you know how JAX figures out which target devices to run the graph on? E.g. if I create two smaller meshes `[0, 1]` and `[2, 3]` in JAX in a `v6e-4`, and jit a computation that uses both meshes, I suppose can JAX figure out that all 4 devices are involved in the computation, and launch the graph on all 4 devices (although some ops will only use 2 out of 4 devices)?

test/cpp/test_xla_sharding.cpp
```diff
@@ -222,6 +222,113 @@ TEST_F(XLAShardingTest, ShardTensor) {
   EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({10, 1, 4, 4, 2}));
 }
 
+TEST_F(XLAShardingTest, ShardTensorLocalMesh) {
+  // Test sharding with a local mesh.
+  std::vector<std::string> devices = {"TPU:8",  "TPU:9",  "TPU:10", "TPU:11",
+                                      "TPU:12", "TPU:13", "TPU:14", "TPU:15"};
+
+  // 1D tiled
```
pgmoka, line 230: For the different tiled methods, could we not do parasitized testing to significantly save code, and make things more readable?

INSTANTIATE_TEST_CASE_P reference: https://www.sandordargo.com/blog/2019/04/24/parameterized-testing-with-gtest

test/cpp/test_xla_sharding.cpp
```diff
@@ -222,6 +222,113 @@ TEST_F(XLAShardingTest, ShardTensor) {
...
+      {4, 5, 6, 7},
+  });
+  sharding = xla::HloSharding::Tile(mesh).ToProto();
+  sharding_spec =
+      std::make_shared<XLATensor::ShardingSpec>(sharding, tensor_shape);
+  shards = ShardingUtil::ShardTensor(tensor, sharding_spec, devices,
+                                     /*padded=*/false);
+  EXPECT_EQ(shards.size(), 8);
+  EXPECT_EQ(shards[0].sizes(), c10::ArrayRef<long>({4, 2, 4}));
+  EXPECT_EQ(shards[7].sizes(), c10::ArrayRef<long>({4, 1, 4}));
```
pgmoka, line 264: Could you add a little on the error messages? What is this checking for?

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -63,12 +63,15 @@ def __init__(self,
...
+    # size of device_ids matches mesh_shape
     assert (len(device_ids) == np.prod(mesh_shape))
+    # device ids are unique
     assert len(device_ids) == len(np.unique(device_ids))
     self.device_ids = device_ids
     self.mesh_shape = mesh_shape
     self.axis_names = axis_names
-    assert all(d < self.size() for d in device_ids)
+    # device ids are continous
+    assert all(d < self.size() for d in device_ids - np.min(device_ids))
```
pgmoka, line 74: NIT: I think it would be easier to read if all asserts were in the same block. This seems like it can come before hand. Is there any reason for it to be after the class assignments?

torch_xla/distributed/spmd/xla_sharding.py
lsy323: This is the mesh constructor : )

torch_xla/csrc/xla_sharding_util.cpp
lsy323: This is a good question! Tried to explain in the PR description but I should have left some comment in the code.

From the logic that `device_index` map is used ([ref](https://github.com/pytorch/xla/blob/b3f387571439438f64b4bb1ae2619b8776936ece/torch_xla/csrc/xla_sharding_util.cpp#L385-L427)), The `global_ordinal` here represents the global ordinal in the mesh, instead of the physical device indices. Ordinal in the mesh and the physical device indices are the same unless we are not using all devices in a SPMD program. So here I used the remainder to get the global ordinal in the mesh.

For example, in a v6e-8 VM with 2 hosts. And we have a mesh size of 4 on each host with its local devices. In the 2nd host (with device TPU:4, TPU:5, TPU:6, TPU:7), the global ordinal should be 0, 1, 2, 3 in the dev

Using the remainder assumes 2 things:

1. Physical device ids are continuous among host (From my observation it's true)
2. In multi-host, if having multiple meshes, the physical ids are continuous within each mesh. (This condition is checked in the mesh constructor in this pr [ref](https://github.com/pytorch/xla/pull/8810/files#diff-3dcff2b7395bbf1f8a09170775388ef686a1e5f593b3c3889996d78c93a9c394R74))

lsy323: Let me add a comment in the code for the rationale.

lsy323:
> > We need to find a way in PjrtComputationClient::Compile to determine the target devices
> 
> Do you know how JAX figures out which target devices to run the graph on? E.g. if I create two smaller meshes `[0, 1]` and `[2, 3]` in JAX in a `v6e-4`, and jit a computation that uses both meshes, I suppose can JAX figure out that all 4 devices are involved in the computation, and launch the graph on all 4 devices (although some ops will only use 2 out of 4 devices)?

I'm not sure if it's possible to use multiple meshed in a single computation. I think in the same SPMD program, we have to use only 1 mesh. I can have a try and update here later.

torch_xla/distributed/spmd/xla_sharding.py
lsy323: Because it calls self.size() which depends on the mesh_shape I think.

test/cpp/test_xla_sharding.cpp
lsy323: Thank you for the suggestion! But I personally think it's hard to parameterize in this case. We need to programmatically derive the mesh and expected output from the test parameters. It's hard to generalize in this case (1d tile, 2d tile and etc). I feel like in the current implementation, even if it's long, but it's more readable.

test/cpp/test_xla_sharding.cpp
lsy323: Actually it's already mentioned in https://github.com/pytorch/xla/pull/8810/files/d3feb5f7097f082897086f0f7add7e06b13e488a#diff-f819f5cf655ccf581542ac85e9eba1ab11730ec87826819ad74c3525de3149dcR248-R249

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -559,7 +559,10 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
           .set_allow_spmd_sharding_propagation_to_output(
               {instance.allow_spmd_sharding_propagation_to_output});
 
-      int num_partitions = client_->device_count();
+      int num_partitions = GetNumGlobalDevices();
+      if (runtime::sys_util::GetEnvBool("XLA_USE_LOCAL_SPMD", false)) {
```
rpsilva-aws, line 563: nit: We could do `bool use_local_spmd = runtime::sys_util::GetEnvBool("XLA_USE_LOCAL_SPMD", false);` and use in both places below.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -589,11 +592,18 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
       }
 
       // TODO(244391366) verify this is correct for the collectives ops
-      xla::DeviceAssignment device_assignment(1, client_->device_count());
+      xla::DeviceAssignment device_assignment(1, num_partitions);
       // DeviceAssignment values must be the PjRtDevice ID, so we need to
       // unwind the global ordinal mapping.
```
rpsilva-aws, line 597: We can adapt the comment, since we don't only "unwind the global ordinal mapping" now.

torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -589,11 +592,18 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
...
-      for (const auto& [device_id, global_ordinal] : global_ordinals_) {
-        device_assignment(0, global_ordinal) = device_id;
+      if (runtime::sys_util::GetEnvBool("XLA_USE_LOCAL_SPMD", false)) {
+        auto local_pjrt_devices = client_->addressable_devices();
```
rpsilva-aws, line 599: nit `const auto&`

torch_xla/csrc/runtime/ifrt_computation_client.cc
```diff
@@ -613,10 +613,14 @@ IfrtComputationClient::ExecuteReplicated(
   return data_handles;
 }
 
-size_t IfrtComputationClient::GetNumDevices() const {
+size_t IfrtComputationClient::GetNumLocalDevices() const {
```
rpsilva-aws, line 616: nit: We could also consider naming to include "addressable"/"visible". Technically, we can have N local devices for a process, but not necessarily all the host's devices - which could be considered to be local devices. If we have 2 processes in one host, each with one addressable device, we can think of how we want the terminology to play out here (no strong opinions).

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -1500,13 +1500,16 @@ void InitXlaModuleBindings(py::module m) {
...
-    return runtime::GetComputationClient()->GetNumDevices();
+    return runtime::GetComputationClient()->GetNumLocalDevices();
   });
   m.def("_xla_get_all_runtime_devices", []() {
     std::vector<std::string> all_devices =
         runtime::GetComputationClient()->GetAllDevices();
     return all_devices;
   });
+  m.def("_xla_num_global_devices", []() -> int64_t {
+    return runtime::GetComputationClient()->GetNumGlobalDevices();
```
rpsilva-aws, line 1511: What's the difference with this and `_xla_get_all_runtime_devices`?

torch_xla/csrc/tensor_impl.cpp
```diff
@@ -57,7 +57,7 @@ struct XLAGuardImpl : public c10::impl::DeviceGuardImplInterface {
       return 0;
     }
 
-    return client->GetNumDevices();
+    return client->GetNumLocalDevices();
```
rpsilva-aws, line 60: Q: Where do we use this?

torch_xla/distributed/spmd/xla_sharding.py
lsy323: We should check if the mesh only contains local devices under local spmd, updated the condition

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -377,6 +380,20 @@ def _get_sharding_type(partition_spec: Tuple[Union[int, None]],
...
+  """
+  Normalize the device mesh to start from 0.
+  
+  This is needed when mesh doesn't include all global devices
+  (e.g. In multi-host setup, each host has a mesh containing local devices).
+  Because HLO graph always use logical device ids in the sharding annotation,
+  we need to normalize the physical device ids to generate the correct HLO
+  sharding annotation.
+  """
+  device_id_min = np.min(device_mesh)
```
rpsilva-aws, line 403: nit: We could avoid the copy if `device_id_min == 0` by exiting early.

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -63,12 +63,15 @@ def __init__(self,
...
     assert axis_names is None or (len(set(axis_names)) == len(axis_names))
+    # size of device_ids matches mesh_shape
     assert (len(device_ids) == np.prod(mesh_shape))
+    # device ids are unique
     assert len(device_ids) == len(np.unique(device_ids))
     self.device_ids = device_ids
     self.mesh_shape = mesh_shape
     self.axis_names = axis_names
-    assert all(d < self.size() for d in device_ids)
+    # device ids are continous
```
rpsilva-aws, line 73: Do we need to check if they're contiguous, and is that an issue? For example, we have device ids [7,6,5,4] or [7,4,6,5].

rpsilva-aws:
Nice, this is great! We were in need of a similar use case for MPMD+SPMD (mostly hardcoded) :)

> Note that the global SPMD and local SPMD are not bridged, we need to have an API to create global SPMD tensor from local SPMD.

This is interesting, do we have a plan for this? We were stumbling on this too.

nit: We could consider adding an helper `xs` method for generating local SPMD meshes (similar to how we have one with 1D mesh), that includes the following taken from the ref above, given a `partition_spec` and `mesh_shape`:
```
process_id = xr.process_index()
num_local_devices = xr.addressable_runtime_device_count()

device_id_start = process_id * num_local_devices
device_ids = np.arange(device_id_start, device_id_start + num_local_devices)
mesh = Mesh(device_ids, mesh_shape, partition_spec)
```

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -63,12 +63,24 @@ def __init__(self,
...
+    if os.environ['XLA_USE_LOCAL_SPMD'] == '1':
+      # In local SPMD mesh only contains local devices.
+      min_device_idx = xr.process_index() * xr.addressable_runtime_device_count(
```
rpsilva-aws, line 76: This assumes homogeneous distribution of addressable devices - is this a requirement? Say we have a different amount of addressable devices per MPMD, e.g. [0,1], [2,3,4,5,6,7].

lsy323: Yes, this is a requirement for now

torch_xla/csrc/tensor_impl.cpp
lsy323: I just rename the API to make explicitly say it's number of local devices, orignal name is confusing.

torch_xla/csrc/init_python_bindings.cpp
lsy323: this returns a `int64_t`

test/cpp/test_xla_sharding.cpp
pgmoka: I suppose making paramatied functions in Python might be easier than C++. If we want to keep the way we are doing this, making these different tests might make things easier to test.

Currently if this test fails, it is hard to tell which case it is related to.

pgmoka (commented):
Not much to add. I am interested in follow-ups to other comments

tengyifei:
> I'm not sure if it's possible to use multiple meshed in a single computation. I think in the same SPMD program, we have to use only 1 mesh. I can have a try and update here later.

Ok also to answer my own question, JAX does these things:

- When compiling, it looks at the input/output/intermediate shardings to find the device assignment in all those, and validates that they're the same devices. [1]. That means one cannot use two non-overlapping meshes in a single jitted computation.
- Once it has done that, it just determines the number of partitions based on the number of devices [2].

So this roughly translates to every sharded tensor having a mesh and the device assignment is derived from their meshes.

> (Because torch_xla work on tensors, attaching mesh to each sharded tensor seems to be redundant)
> Therefore we have to derive the target devices from the xla::OpSharding.

Is it possible to put a `DeviceAssignment` object next to the `xla::OpSharding` object? The `DeviceAssignment` object would hold a vector of device IDs. It sounds like we already store `xla::OpSharding` for each sharded tensor, so storing an extra field shouldn't be that much overhead and can handle the situations where `xla::OpSharding` is insufficient.

[1]: https://github.com/jax-ml/jax/blob/de9ad6bad9d81061c6ce74c31b51b14a93a5fc43/jax/_src/interpreters/pxla.py#L1726

[2]: https://github.com/jax-ml/jax/blob/de9ad6bad9d81061c6ce74c31b51b14a93a5fc43/jax/_src/interpreters/pxla.py#L2700-L2701

torch_xla/distributed/spmd/xla_sharding.py
tengyifei: > This is the mesh constructor : )

Wait, this is not the mesh constructor. This is the `mark_sharding` function. It seems very strange that we're doing these localized checks in the `mark_sharding` function. IMO this check should be moved to the `Mesh` constructor.

tengyifei:
Sorry for the late comment -- I'm honestly a bit worried of the subtle distinction between local device IDs and global device IDs, particularly the `normalize_logical_mesh` function and the logic scatter across Python and C++. I think I understand what's happening enough to have a suggestion. Here's my understanding -- LMK if I got any part wrong:

- In order to do "local SPMD", we need to tell XLA that `num_partitions == 4` or whatever the number of local addressable devices. We also need to set the IDs in any `OpSharding` proto to be local IDs i.e. begin at 0. Finally, we need to create a `xla::DeviceAssignment` that maps these device IDs to `PjRtDevice::id()`s. It's a bit confusing because the PJRT device ID is what everyone else thinks of as "device IDs" when it comes to XLA, and the PJRT device ID is in fact sparse (e.g. it can be 100001 in case of multi-slice). So some part of our code refers the "local/global device ID" as "global ordinals" instead, which is really just the index of a device in the mesh. XLA thinks in terms of the `PjRtDevice::id()`s and our mesh ordinal is just an intermediate contract. In the extreme case, we could shuffle them arbitrarily and as long as we provide an appropriate `xla::DeviceAssignment`, that will result in the correct collectives.
- Currently, the `mark_sharding` call implementations also lack abstraction. They directly build a `xla::OpSharding` proto and the same `xla::OpSharding` proto is later given to the XLA compiler, which requires that IDs in there start from 0. IIUC that's what forces us to `normalize_logical_mesh`. If XLA didn't have this requirement, we could avoid `normalize_logical_mesh` and just need to have the right `xla::DeviceAssignment`.
- As a result, we have many kinds of device ID concepts. There's a "global device ID" that is dense and goes from 0 to the number of chips in the environment. There's a "local device ID" that's derived by subtracting a bunch of device ID by their minimum, whose correctness is guaranteed by a subtle check in `xs.Mesh` that requires device IDs to be locally addressable iff in local SPMD mode. Not to mention there's a third device ID which is the non-dense `PjRTDevice::id()`. The worse part is that they're all untyped integers.

I'm wondering if it makes sense to do a prefactor, to hide the `xla::OpSharding` proto from the public API. I'm wondering could we have a `xs.Device` C++ type that's bridge to Python similar to how `OpSharding` is bridged. This type will store the `PjRTDevice::id()` which identifies the device. It could also store some information that's returned by `xr.global_runtime_device_attributes()`. In any case, we'd create a first-class `Device` object that unambiguously identifies a device without requiring any contextual information. In contrast, if we used an integer, we'd always need to ask ourselves whether this is a global int or local int or pjrt int etc etc.

Once we have a first-class device object, I think instead of storing `xla::OpSharding`, we'd probably create a `torch_xla::OpSharding` type that stores the same things except that it stores `std::shared_ptr<Device>` instead of ID. If the cost of a few hundred shared_ptr is significant, we could also have it store an index into some global `std::vector<Device>` table cached inside the computation client.

Once we've done this refactor, the Python layer can reason in terms of `torch_xla::OpSharding` objects instead of `xla::OpSharding`. Then it's not forced to normalize the device IDs. We'll still have two kinds of device IDs (the global dense ID and the sparse `PjRTDevice::id()`, but that's better than having three kinds of device IDs).

If this feature is no longer urgently required, I wonder is it possible to do this refactor? I think it could also let us support other kinds of SPMD (e.g. using 2 chips out of 4 chips) instead of being restricted to either full local SPMD or global SPMD.

PR #8817, opened 2025-03-11
Add information about codegen in codegen/xla_native_functions.yaml

pgmoka:
Add some information on `codegen` and `supported` configs based on https://github.com/pytorch/xla/issues/8713.

Info on generation also comes from https://github.com/pytorch/pytorch/blob/main/torchgen/gen_lazy_tensor.py.

### DIFF
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -1,6 +1,6 @@
 backend: XLA
 cpp_namespace: torch_xla
-# full_codegen is the prefered method of codegenration. Through this config
+# full_codegen is the prefered method of code generation. Through this config
 # ops get implementations (and IR classes) generated. See
 # https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md
 # for more details on differences on what gets generated or not.
@@ -112,9 +112,9 @@ ir_gen:
   - leaky_relu_backward
 # supported requires the ops to have hand-implemented everything but the native
 # function's declaration and registration. This follows closer the
-# instructions from docs/source/contribute/op_lowering.md and the codegen
-# pattern from https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md#pytorchxla-old-op-lowering-files.
-# Prefer using codegen.
+# instructions from https://github.com/pytorch/xla/blob/master/docs/source/contribute/op_lowering.md
+# and the codegen pattern from https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md#pytorchxla-old-op-lowering-files.
+# Prefer using full_codegen.
 supported:
   - __ilshift__.Scalar
   - __ilshift__.Tensor
@@ -434,6 +434,8 @@ symint:
   - view
 # autograd adds the operation at the "AutogradXLA" dispatch key level. This is
 # important for functions that do not have backpropagation implemented at the
+# Autograd dispatch key level, or that the implementation provided by PyTorch
+# upstream at the Autograd dispatch key level is sub-optimal for our use cases
 autograd:
   - einsum
   - max_pool2d
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -97,8 +97,8 @@ full_codegen:
   - tril
   - triu
   - trunc
-# ir_gen, only IR non_native is used to register kernels not declared from
-# this file
+# ir_gen: Operations in ir_gen generate only the declaration of their respective
+# lazy IR node classes. Their native functions are not generated.
 ir_gen:
   - _adaptive_avg_pool2d
   - _adaptive_avg_pool2d_backward
@@ -110,8 +110,8 @@ ir_gen:
   - bitwise_or.Tensor
   - bitwise_xor.Tensor
   - leaky_relu_backward
-# supported requires the ops to have hand-implementations but still get
-# codegenned declarations and registrations. This follows closer the
+# supported requires the ops to have hand-implemented everything but the native
+# function's declaration and registration. This follows closer the
 # instructions from docs/source/contribute/op_lowering.md and the codegen
 # pattern from https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md#pytorchxla-old-op-lowering-files.
 # Prefer using codegen.
--- a/codegen/xla_native_functions.yaml
+++ b/codegen/xla_native_functions.yaml
@@ -1,5 +1,9 @@
 backend: XLA
 cpp_namespace: torch_xla
+# full_codegen is the prefered method of codegenration. Through this config
+# ops get implementations (and IR classes) generated. See
+# https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md
+# for more details on differences on what gets generated or not.
 full_codegen:
   - acos
   - acosh
@@ -104,6 +108,11 @@ ir_gen:
   - bitwise_or.Tensor
   - bitwise_xor.Tensor
   - leaky_relu_backward
+# supported requires the ops to have hand-implementations but still get
+# codegenned declarations and registrations. This follows closer the
+# instructions from docs/source/contribute/op_lowering.md and the codegen
+# pattern from https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md#pytorchxla-old-op-lowering-files.
+# Prefer using codegen.
 supported:
   - __ilshift__.Scalar
   - __ilshift__.Tensor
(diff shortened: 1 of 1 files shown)


pgmoka:
@tengyifei @ysiraichi my knowledge on the other forms of generation (`symint`, and `autograd`) is a little spotty. I would be interested if y'all have anything to add here.

codegen/xla_native_functions.yaml
```diff
@@ -93,6 +97,8 @@ full_codegen:
   - tril
   - triu
   - trunc
+# ir_gen, only IR non_native is used to register kernels not declared from
+# this file
```
ysiraichi, line 101: I think we could have a better description, here. Something like: _"Operations in `ir_gen` generate only the declaration of their respective lazy IR node classes. Their native functions are not generated."_

codegen/xla_native_functions.yaml
```diff
@@ -104,6 +110,11 @@ ir_gen:
   - bitwise_or.Tensor
   - bitwise_xor.Tensor
   - leaky_relu_backward
+# supported requires the ops to have hand-implementations but still get
+# codegenned declarations and registrations. This follows closer the
```
ysiraichi, line 114: I think we could be more specific, here: _"... requires the ops to have hand-implemented everything but the native function's declaration and registration."_

ysiraichi (commented):
This looks nice. Besides the comments, I have 2 suggestions:
- Add `symint` and `autograd` descriptions
- Mention what gets generated in which file

Let me know if you need help on gathering this information.

pgmoka:
> This looks nice. Besides the comments, I have 2 suggestions:
> 
> * Add `symint` and `autograd` descriptions
> * Mention what gets generated in which file
> 
> Let me know if you need help on gathering this information.

@ysiraichi I haven't been able to find a great definite source on these which is why I haven't written about them. If you have links, I would appreciate it.

codegen/xla_native_functions.yaml
pgmoka: I like this suggestion. Added.

codegen/xla_native_functions.yaml
pgmoka: I like this suggestion. Added.

codegen/xla_native_functions.yaml
```diff
@@ -1,5 +1,9 @@
 backend: XLA
 cpp_namespace: torch_xla
+# full_codegen is the prefered method of codegenration. Through this config
```
tengyifei, line 3: nit: code generation

codegen/xla_native_functions.yaml
```diff
@@ -1,5 +1,9 @@
...
+# ops get implementations (and IR classes) generated. See
```
tengyifei, line 4: Could you talk about what _isn't_ generated, i.e. what the developer needs to fill out?

codegen/xla_native_functions.yaml
```diff
@@ -104,6 +110,11 @@ ir_gen:
   - bitwise_or.Tensor
   - bitwise_xor.Tensor
   - leaky_relu_backward
+# supported requires the ops to have hand-implemented everything but the native
+# function's declaration and registration. This follows closer the
+# instructions from docs/source/contribute/op_lowering.md and the codegen
+# pattern from https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md#pytorchxla-old-op-lowering-files.
+# Prefer using codegen.
```
tengyifei, line 117: `full_codegen`?

codegen/xla_native_functions.yaml
```diff
@@ -93,6 +97,8 @@ full_codegen:
   - tril
   - triu
   - trunc
+# ir_gen: Operations in ir_gen generate only the declaration of their respective
+# lazy IR node classes. Their native functions are not generated.
```
tengyifei, line 101: Double checking my understanding, this means `ir_gen` doesn't generate native functions, but `supported` does generate native functions? Or is there a difference between native function _declaration_ and native function _definition_ (i.e. implementation)?

codegen/xla_native_functions.yaml
```diff
@@ -421,6 +432,8 @@ symint:
   # See Note: [Disabling functionalization]
   - expand
   - view
+# autograd adds the operation at the "AutogradXLA" dispatch key level. This is
+# important for functions that do not have backpropagation implemented at the
```
tengyifei, line 436: [...] implemented at the `Autograd` dispatch key level, or that the implementation provided by PyTorch upstream at the `Autograd` dispatch key level is sub-optimal for our use cases.

tengyifei (changes requested):

codegen/xla_native_functions.yaml
ysiraichi: > this means `ir_gen` doesn't generate native functions, but `supported` does generate native functions?

Yes.

> Or is there a difference between native function declaration and native function definition?

`ir_gen` generates the lazy IR node class declaration. Some of the class' functions still have to be manually written (e.g. `XXX::Lower()`).

`supported` generates the `XLANativeFunction::XXX` declaration + registration to the XLA dispatch-key. We still have to manually write the definition of that function.

ysiraichi:
Here's a small summary about each component inside `xla_native_functions.yaml` file:

- `full_codegen`
    - **Generates:** 
        - Native function declaration and definition (at _XLANativeFunctions.{h,cpp}_)
        - IR node class declaration (at _LazyIR.h_)
        - Kernel registration (at _RegisterXLA.cpp_)
    - **Needs:**
        - `<Op>OutputShape()` declaration and definition (at _ops_xla_shape_fn.{h,cpp}_)
        - `<Op>::Lower()` definition (at _ops_lower_fn.cpp_)
- `supported`
    - **Generates:** 
        - Native function declaration (at _XLANativeFunctions.h_)
        - Kernel registration (at _RegisterXLA.cpp_)
    - **Needs:**
        - Native function definition (at _aten_xla_type.cpp_)
- `ir_gen` (at the moment, always used with `supported`)
    - **Generates:** 
        - IR node class declaration (at _LazyIR.h_)
- `symint` (at the moment, always used with `supported`)
    - Same as `supported` (or `full_codegen`, if used instead), but uses the operation overload that takes `SymInt`s instead of plain integers
- `autograd`
    - **Generates:** 
        - Native function declaration (at _XLANativeFunctions.h_)
        - Kernel registration (at _RegisterAutogradXLA.cpp_)
    - **Needs:**
        - Native function definition (at _aten_xla_type.cpp_)

_**Note:** this is my own understanding by looking at the source code._

codegen/xla_native_functions.yaml
pgmoka: I think https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md is a good resource for that. Do you have something else in mind that you think should be added here?

codegen/xla_native_functions.yaml
pgmoka: Fixed typo

codegen/xla_native_functions.yaml
pgmoka: Yes. Good catch

codegen/xla_native_functions.yaml
pgmoka: Appended suggestion

codegen/xla_native_functions.yaml
pgmoka: Are there any recommendations of something to add here from this conversation?

pgmoka:
> * symint

I have added a description to symint based on this and https://github.com/pytorch/xla/pull/3978. Please let me know what you think of it @ysiraichi

codegen/xla_native_functions.yaml
tengyifei: Let's include Yukio's explanation into the yaml docs

codegen/xla_native_functions.yaml
tengyifei: I see. Maybe add a `# See https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md` at the first line of the yaml

tengyifei: Oh, you already did

tengyifei (approved):
LGTM modulo the open comments

pgmoka:
> I think it would be nice to write down what's getting generated in which files. And, as @tengyifei mentioned, what's needed for finishing the lowering.

I think this information should be in https://github.com/pytorch/xla/blob/master/docs/source/contribute/op_lowering.md and https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md. I would recommend link to those documents, and then supplementing them if we believe there are gaps. We could create an issue to add more information to https://github.com/pytorch/xla/blob/master/docs/source/contribute/op_lowering.md on `symint` and `autograd`, but I would do this in a follow-up PR. Let me know what you think @ysiraichi.

pgmoka:
I have added links to the referred docs in the head of the file.

tengyifei (approved):

ysiraichi (approved):

PR #8823, opened 2025-03-12
Remove xla_mark_sharding_dynamo_custom_op

tengyifei:
AFAIK this is never used anywhere. There's a `torch.ops.xla.dynamo_mark_sharding` Python function that seems to be the way to mark sharding under dynamo.


lsy323 (approved):
LGTM, thanks!

Issue #8820, opened 2025-03-12
Triton test failed in CI: ImportError: cannot import name 'AttrsDescriptor'


tengyifei:
See https://github.com/pytorch/xla/actions/runs/13799623710/job/38601675184?pr=8807

```
======================================================================
ERROR: test_gpu_custom_call_triton_add (__main__.TritonTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/xla/xla/pytorch/xla/test/test_triton.py", line 256, in test_gpu_custom_call_triton_add
    payload = xla_triton.triton_call(
  File "/usr/local/lib/python3.10/site-packages/torch_xla/experimental/triton.py", line 211, in triton_call
    return triton_kernel_call_lowering(array_args, kernel, compiled_kernel,
  File "/usr/local/lib/python3.10/site-packages/torch_xla/experimental/triton.py", line 142, in triton_kernel_call_lowering
    from triton.backends.compiler import AttrsDescriptor
ImportError: cannot import name 'AttrsDescriptor' from 'triton.backends.compiler' (/usr/local/lib/python3.10/site-packages/triton/backends/compiler.py)

======================================================================
ERROR: test_gpu_custom_call_triton_flash_attention (__main__.TritonTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/xla/xla/pytorch/xla/test/test_triton.py", line 292, in test_gpu_custom_call_triton_flash_attention
    payload = xla_triton.triton_call(
  File "/usr/local/lib/python3.10/site-packages/torch_xla/experimental/triton.py", line 211, in triton_call
    return triton_kernel_call_lowering(array_args, kernel, compiled_kernel,
  File "/usr/local/lib/python3.10/site-packages/torch_xla/experimental/triton.py", line 142, in triton_kernel_call_lowering
    from triton.backends.compiler import AttrsDescriptor
ImportError: cannot import name 'AttrsDescriptor' from 'triton.backends.compiler' (/usr/local/lib/python3.10/site-packages/triton/backends/compiler.py)

```

ysiraichi:
Apparently, this was caused by a triton pin update inside PyTorch https://github.com/pytorch/pytorch/pull/148971. This wouldn't have happened if we did have a PyTorch pin, though. Somewhat related to #8745.

PR #8825, opened 2025-03-12
Fix for triton pin update.

ysiraichi:
Fix: #8820

Recently, PyTorch updated its triton pin https://github.com/pytorch/pytorch/pull/148971, making it so our triton tests fail. This PR makes the necessary modifications for accommodating the pin update.

In summary, this PR creates 2 functions `spec_and_divisible_by_16` and `spec_and_equals_1` which are copied from [the former `JITFunction._get_config()` function](https://github.com/triton-lang/triton/blob/da40a1e984bf57c4708daf603eb427442025f99b/python/triton/runtime/jit.py#L187-L198). That function (and their analogous versions) was removed from the recent versions of triton.

cc @bhavya01


bhavya01 (approved):

bhavya01 (commented):
Looks like the CI tests failed.

PR #8826, opened 2025-03-12
Expose mark_sharding_with_gradients as a public API.

iwknow:
This change concludes https://github.com/pytorch/xla/issues/8678

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -4,9 +4,7 @@ import unittest
 from unittest.mock import patch
 import math
 import numpy as np
-import os
 import sys
-import re
 
 import torch
 from torch import nn
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1527,12 +1527,12 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
     torch.testing.assert_close(wg1, wg2)
     torch.testing.assert_close(bg1, bg2)
 
-
   def test_mark_sharding_with_gradients_basic(self):
     partition_spec = (0, 1)
     xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
-                        dtype=torch.float,
-                        device=xm.xla_device(), requires_grad=True)
+                       dtype=torch.float,
+                       device=xm.xla_device(),
+                       requires_grad=True)
     mesh = self._get_mesh((1, self.n_devices))
     xst1 = xs.mark_sharding_with_gradients(xt1, mesh, partition_spec)
     output = xst1.sum()
@@ -1543,8 +1543,9 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
   def test_mark_sharding_with_gradients_annotation(self):
     partition_spec = (0, 1)
     xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
-                        dtype=torch.float,
-                        device=xm.xla_device(), requires_grad=True)
+                       dtype=torch.float,
+                       device=xm.xla_device(),
+                       requires_grad=True)
     mesh = self._get_mesh((1, self.n_devices))
     xst1 = xs.mark_sharding_with_gradients(xt1, mesh, partition_spec)
     output = xst1.sum()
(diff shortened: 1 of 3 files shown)


iwknow:
it seems that i cannot trigger the re-run of the testing workflow. is it a permission issue? or do i miss anything?

tengyifei:
@iwknow for security only repo writers can run workload. i just ran it for you

iwknow:
I am a little bit confused about what to expect here. in my test, i have
```
xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
                       dtype=torch.float,
                       device=xm.xla_device(),
                       requires_grad=True)
xst1 = xs.mark_sharding_with_gradients(xt1, mesh, partition_spec)
xst1.retain_grad()
output = xst1.sum()
output.retain_grad()
output.backward()
```

there are three tensors: xt1, xst1, output and their corresponding gradients. I expect `xst1`, `xst1.grad`, `output`, `output.grad`, and `xt1.grad` have "sharding" section in their hlo.  However, my experiment shows that `xt1`, `xt1.grad`, `xst1`, `output` have the "sharding" section in their hlo but `xst1.grad` and `output.grad` doesn't have the "sharding". is this expected? or is that related to `retain_grad` (by default, these grads are not retained)? or do i miss something?

test/spmd/test_xla_sharding.py
```diff
@@ -1526,6 +1527,40 @@ def test_xla_patched_linear(self):
...
+                       requires_grad=True)
+    mesh = self._get_mesh((1, self.n_devices))
+    # Notice that the function does not modify in-place.
+    xst1 = xs.mark_sharding_with_gradients(xt1, mesh, partition_spec)
+    output = xst1.sum()
+    output.backward()
+    self.assertTrue(xt1.grad is not None)
+    hlo_t = torch_xla._XLAC._get_xla_tensors_hlo([xst1])
+    hlo_grad = torch_xla._XLAC._get_xla_tensors_hlo([xst1.grad])
+    sharding_annotation_pattern = r"sharding={([^}]*)}"
```
bhavya01, line 1557: Thanks for the change! Can we check explicitly that the second dimension is sharded along self.n_devices?

Like this test https://github.com/pytorch/xla/blob/5beb65117dbba7dfd0b31155c5d5bc27488f540e/test/spmd/test_xla_sharding.py#L349

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -564,6 +564,40 @@ def mark_sharding(
...
+  """
+    Annotates the tensor provided and the gradient computed with XLA partition spec. Internally,  
+    it calls mark_sharding on both given tensor and its gradent with the same mesh setting and partition spec.
+
+    Args:
+        t (Union[torch.Tensor, XLAShardedTensor]): input tensor to be annotated with partition_spec.
+
+        mesh (Mesh): describes the logical XLA device topology and the underlying device IDs.
+
+        partition_spec (Tuple[Tuple, int, str, None]): A tuple of device_mesh dimension index or
```
tengyifei, line 580: We also recently improved the documentation of `partition_spec` in `mark_sharding`: https://github.com/pytorch/xla/blob/5beb65117dbba7dfd0b31155c5d5bc27488f540e/torch_xla/distributed/spmd/xla_sharding.py#L532

Could you copy the new version here as well:

```
        partition_spec (PartitionSpec): A tuple of one or more device mesh axes that describes how
            to shard the input tensor. Each element can be:

            - an int: refer to a mesh axis by index
            - str: refer to a mesh axis by name
            - a tuple of the above: refer to multiple mesh axes
            - None: the corresponding tensor dimension will be replicated over all devices

            This specifies how each input rank is sharded (index to mesh_shape) or replicated (None).
            When a tuple is specified, the corresponding input tensor axis will be sharded along all
            mesh axes in the tuple. Note that the order the mesh axes are specified in the tuple
            will impact the resulting sharding.
```

tengyifei:
> I am a little bit confused about what to expect here. in my test, i have
> 
> ```
> xt1 = torch.tensor([[1, 2, 3, 4, 5, 6, 7, 8]],
>                        dtype=torch.float,
>                        device=xm.xla_device(),
>                        requires_grad=True)
> xst1 = xs.mark_sharding_with_gradients(xt1, mesh, partition_spec)
> xst1.retain_grad()
> output = xst1.sum()
> output.retain_grad()
> output.backward()
> ```
> there are three tensors: xt1, xst1, output and their corresponding gradients. I expect xst1, xst1.grad, output,
> output.grad, and xt1.grad have "sharding" section in their hlo. However, my experiment shows that xt1, xt1.grad, xst1, 
> output have the "sharding" section in their hlo but xst1.grad and output.grad doesn't have the "sharding". is this 
> expected? or is that related to retain_grad (by default, these grads are not retained)? or do i miss something?

There's a difference between "having sharding in their hlo" vs "having sharding in their `torch_xla._XLAC._get_xla_sharding_spec(my_tensor)`.

If you check the HLO of a tensor, that will contain not just the HLO corresponding to the tensor itself, but also the HLO of any input tensor and their inputs, etc., recursively until you hit device data tensors. It's better to check `torch_xla._XLAC._get_xla_sharding_spec`, which will return the sharding spec of the tensor and nothing else. I think if you search for this function's usage in the code base you'd find how to write tests with it.

In our snippet above, I'd expect the following:

- `xst1` has sharding annotations if you look in `_get_xla_sharding_spec`
- `xst1.grad` has sharding annotations if you look in `_get_xla_sharding_spec`
- everything else won't have sharding annotations

That's because if we don't call `torch_xla.sync()`, the GSPMD sharding propagation is not run. Then only the tensors which we explicitly called `mark_sharding` on will have a sharding spec.

I do think `xst.retain_grad()` is required to keep the `xst.grad` node around though; otherwise it would be cleared by PyTorch after `output.backward()`.

tengyifei:
Ah, you're right

test/spmd/test_xla_sharding.py
iwknow: added.

torch_xla/distributed/spmd/xla_sharding.py
iwknow: added.

PR #8828, opened 2025-03-13
Add build triggers for 2.7-rc1

zpcore:


infra/tpu-pytorch-releases/artifacts.auto.tfvars
```diff
@@ -50,6 +50,118 @@ nightly_builds = [
 
 # Built on push to specific tag.
 versioned_builds = [
+  # Remove libtpu from PyPI builds, pre-C++11 ABI builds
```
pgmoka, line 53: Is there no way to use something templates, or vars to do this rather multiple copy and pastes?

zpcore:
I probably find out the issue why CUDA build was not triggered. We have a new option `cxx11_abi` intorduced, but the field is not added to the CUDA build. This results in the error when call `b.cxx11_abi == "1", false`. 

Apart from that, those .tf configs are quite out dated and not align with the currect GCP config. I have redo the `terraform apply` to keep everything up to date.

zpcore:
Hi @lsy323 , I have confirmed the syntax issue with `terraform plan` and already applied the update to the GCP. Let's see if those wheels can be successfully built or not before merge in. Thanks!

zpcore:
I come to notice that our build will fail if we set `BUNDLE_LIBTPU=1`. The fail is due to the builder script will look for libtpu from the following address:
```
https://storage.googleapis.com/libtpu-nightly-releases/wheels/libtpu/libtpu-0.0.11.dev20250303+nightly-py3-none-linux_x86_64.whl
```
However, the address is not valid anymore. Will fix this issue later.

ManfeiBai (approved):
Thanks, LGTM

Issue #8794, opened 2025-03-05
xb.call_jax always assumes the traced function makes use of all arguments


tengyifei:
If the jax function doesn't use one of the arguments (common in gradient calculations), then the HLO parameter count and function parameters will mismatch.

We need a way to get jax to tell us which input argument maps to which HLO parameter.

PR #8830, opened 2025-03-13
Support unused arguments in xb.call_jax

tengyifei:
Fixes https://github.com/pytorch/xla/issues/8794.

Turns out the answer is as simple as `keep_unused=True`.

I also took a test case in
https://github.com/pytorch/xla/pull/8800 and made it pass.


zpcore (approved):
This is nice!

PR #8844, opened 2025-03-17
Make cxx_abi enable by default

zpcore:
1. Make cxx_abi flag on by default but still give user option to build with the flag off. When the flag is off, the wheel will append `_precxx11` in the naming;
2. Minor naming fix in the CI;
3. Add cuda 12.8 in the nightly build since the default CUDA driver in the XLML test VM is 12.8 now.


README.md
```diff
@@ -51,8 +51,9 @@ pip install 'torch_xla[pallas]' \
 ```
 
 ### C++11 ABI builds
+**Note: As of 03/18/2025 and starting from Pytorch/XLA 2.7 release, we make C++11 ABI enable by default and stop building wheels with the flag turned off. Wheels with the flag turned off will append `_precxx11` in the naming.**
```
tengyifei, line 54: nit:

```
As of 03/18/2025 and starting from Pytorch/XLA 2.7 release, C++11 ABI builds are
the default and we no longer provide wheels built with pre-C++11 ABI.
```
to be clearer.

> Wheels with the flag turned off will append `_precxx11` in the naming.

If this is only an opt-in option, I think it doesn't need to be documented at the front of the README.
Maybe put it in the ansible `.yaml` if you think it's worth it.

Since this markdown has been reflowed to 80 characters column limit, it's good to match that style too.

infra/tpu-pytorch-releases/artifacts_builds.tf
```diff
@@ -40,40 +39,37 @@ locals {
     # Regular TPU builds (non-libtpu, pre-C++11 ABI)
```
tengyifei, line 39: This says "non-libtpu, pre-C++11 ABI". Is it still accurate?

tengyifei (approved):

README.md
zpcore: Sounds good, let's make it simple.

infra/tpu-pytorch-releases/artifacts_builds.tf
zpcore: Nice catch, should be `C++11 ABI`.

ManfeiBai (approved):
Thanks, LGTM

PR #8849, opened 2025-03-18
Avoid unnecessary copy in TensorSource

lsy323:
Avoid `at::Tensor` copy in `TensorSource` if it's not necessary.

The copy operations are needed under 2 cases:

1. On XLA:GPU path, if tensor is on CUDA device, need to copy to CPU then pass to PJRT runtime to transfer to GPU. This is because @ysiraichi found passing CUDA tensor to PJRT runtime doesn't work, so doing the roundtrip as a workaround.
2. On XLA:TPU path, if the tensor is not contiguous, need to use the copy to make the memory contiguous. Because PJRT takes raw data ptr which expects data to be contiguous.

The copy operation needs to be blocking, since the transfer operation depends on the copied tensor.


lsy323:
Hi @ysiraichi, just follow up on offline discussion on the copy operation. PTAL at the PR, thanks!

torch_xla/csrc/runtime/tensor_source.h
```diff
@@ -54,13 +54,19 @@ class AtenSource : public TensorSource {
     if (target_torch_type != tensor.type().scalarType()) {
       TORCH_LAZY_COUNTER("AtenSourceDowncasts", 1);
     }
+    bool need_copy = true;
+    // The purpose of copy is to ensure the memory is contiguous, which is
+    // expected by PJRT.
+    if (tensor.device() == at::kCPU && tensor.is_contiguous()) {
+      need_copy = false;
```
yaochengji, line 61: Do you think we can directly skip the `tensor.to` when `need_copy` is False?

lsy323: Good point, we can avoid calling `tensor.to` if target dtype is the same

lsy323: updated

torch_xla/csrc/runtime/tensor_source.h
```diff
@@ -54,13 +54,23 @@ class AtenSource : public TensorSource {
...
+      tensor_ = std::move(tensor);
+    } else {
+      // TODO(ysiraichi): check, first, if tensor lives in a device that the
+      // current PjRt client has access. If so, we don't need to go through the
+      // CPU.
+      tensor_ = std::move(tensor.to(
+          at::TensorOptions().device(at::kCPU).dtype(target_torch_type),
+          /*non_blocking=*/false,
+          /*copy=*/true, at::MemoryFormat::Contiguous));
+    }
```
ysiraichi, line 73: As far as I understand it, `tensor.to(...)` (without the `copy` argument) already checks whether it should actually copy or not. So, what do you think of reverting to the old `tensor.to(...)` usage, but removing the `copy` argument, instead?

ysiraichi (commented):
As a side note, we can use the DLPack machinery for doing the CUDA to XLA:CUDA transfer (that wasn't implemented at the time I worked on this). I will open an issue for this.

torch_xla/csrc/runtime/tensor_source.h
lsy323: Hi @ysiraichi, I didn't find a `tensor.to(...)` without the `copy` arg in C++, is it only in python?

torch_xla/csrc/runtime/tensor_source.h
```diff
@@ -54,13 +54,23 @@ class AtenSource : public TensorSource {
...
-    tensor_ = std::move(
-        tensor.to(at::TensorOptions().device(at::kCPU).dtype(target_torch_type),
-                  /*non_blocking=*/false,
-                  /*copy=*/true, at::MemoryFormat::Contiguous));
+    // The purposes of copy are:
+    // 1. Ensure the memory is contiguous, which is expected by PJRT.
+    // 2. Move CUDA tensor to CPU since we cannot pass CUDA memory to PJRT now.
+    // 3. Cast data type.
+    // We can avoid if copy is not needed.
+    if (tensor.device() == at::kCPU && tensor.is_contiguous() &&
```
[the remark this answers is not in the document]
tengyifei: I guess I'm not sure which copy is avoided by this PR. Does `tensor.to(at::TensorOptions().device(at::kCPU))` introduce a copy _even if_ `tensor` is already a CPU tensor?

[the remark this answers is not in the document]
tengyifei: Oh, I saw your internal message. I guess the ` /*copy=*/true, ` part will force a CPU->CPU copy.

tengyifei (changes requested):

torch_xla/csrc/runtime/tensor_source.h
```diff
@@ -54,13 +54,23 @@ class AtenSource : public TensorSource {
...
+        tensor.dtype() == target_torch_type) {
+      tensor_ = std::move(tensor);
```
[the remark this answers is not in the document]
lsy323: Done

torch_xla/csrc/runtime/tensor_source.h
```diff
@@ -54,13 +54,23 @@ class AtenSource : public TensorSource {
...
+    } else {
+      // TODO(ysiraichi): check, first, if tensor lives in a device that the
```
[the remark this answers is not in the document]
lsy323: Added a new tests to cover all 3 cases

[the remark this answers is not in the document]
tengyifei: nice!

tengyifei (approved):

torch_xla/csrc/runtime/tensor_source.h
ysiraichi: You are right. But, I think we can just `/* copy= */false`.

ysiraichi:
In the [`tensor.to(...)` implementation](https://github.com/pytorch/pytorch/blob/00a2c68f67adbd38847845016fd1ab9275cefbab/aten/src/ATen/native/TensorConversions.cpp#L425-L447), we can see that if `copy=false`, using the old implementation, that function will check whether we need to actually copy the tensor. No need to do it ourselves.

torch_xla/csrc/runtime/tensor_source.h
```diff
@@ -54,13 +54,25 @@ class AtenSource : public TensorSource {
...
+    // 2. Move CUDA tensor to CPU since we cannot pass CUDA memory to PJRT now.
+    // 3. Cast data type.
+    // We can avoid if copy is not needed.
+    if (tensor.device() == at::kCPU && tensor.is_contiguous() &&
+        tensor.dtype() == target_torch_type) {
+      // Skip copying a CPU tensor to CPU.
+      tensor_ = std::move(tensor);
+    } else {
+      TORCH_LAZY_COUNTER("AtenSourceTensorCopy", 1);
+      // TODO(ysiraichi): check, first, if tensor lives in a device that the
```
pgmoka, line 68: NIT: I personally prefer to have TODOs linked to issues, and then having those assigned to people. That way, things can be more easily followed-up if a contributor is no longer active

pgmoka (approved):
Left one small NIT. Otherwise, LGTM. If you could address the NIT before submission, that might be nice.

lsy323:
@ysiraichi @tengyifei Actually took a 2nd thought on this. Skipping the copy seems to be unsafe if the underlying PJRT transfer is async.

```
def gen_rand_on_xla():
  t = torch.rand(100000)
  t_xla = t.to('xla') # Async transfer start
  return t_xla

my_rand_on_xla = gen_rand_on_xla()
# After the above line, async transfer may not be done, but CPU tensor is already out of scope and may be gc'ed.
```

ysiraichi:
Hmm... I don't think this is the case. That's because PyTorch tensors are ref-counted in the C++ side (unless otherwise specified). So, if we hold a C++ `Tensor` type data, that won't be deleted.

ysiraichi:
By the way, I still don't think you need the if-else there. I believe you can just leave the old `tensor.to(...)`, while setting `copy=false`. PyTorch should verify whether a copy is needed for us.

ysiraichi (approved):
LGTM.

lsy323:
> By the way, I still don't think you need the if-else there. I believe you can just leave the old `tensor.to(...)`, while setting `copy=false`. PyTorch should verify whether a copy is needed for us.

Thank you so much @ysiraichi for the suggestion! Updated accordingly.

PR #8857, opened 2025-03-19
Increase TPU CI node count from 2 to 32

tengyifei:
Increase our TPU CI capacity by 16 times while also disabling the auto scale-down which makes TPU nodes slow to start.


lsy323 (approved):

zpcore:
Can you also paste the output of `terraform plan` in the PR? Just to double check what will be changed. Thanks

tengyifei:
Spot checked the terraform plan internally.

PR #8864, opened 2025-03-20
Avoid destorying the cluster

tengyifei:
By keeping the inital_node_count unchanged as before https://github.com/pytorch/xla/commit/6a93dcfbc4764f8f503eaa480222ddc10da0f04e, terraform should not attempt to destroy the cluster and fail.


zpcore (approved):
LGTM@

PR #8865, opened 2025-03-20
Add lowering for `bitwise_left_shift`.

ysiraichi:
This PR lowers `torch.bitwise_left_shift` operation. It introduces the following changes:

- Add `bitwise_left_shift.Tensor` to `full_codegen` section of _native_functions.yaml_
- Introduce the corresponding `*OuputShape()` and `*::Lower()` functions
- Add test confirming that we don't fallback

### DIFF
--- a/torch_xla/csrc/ops/ops_xla_shape_fn.cpp
+++ b/torch_xla/csrc/ops/ops_xla_shape_fn.cpp
@@ -363,7 +363,8 @@ xla::Shape BitwiseXorTensorOutputShape(const torch::lazy::Value& input,
 xla::Shape BitwiseLeftShiftTensorOutputShape(const torch::lazy::Value& input,
                                              const torch::lazy::Value& other) {
   return InferBinaryOpShape(input, other, [](xla::XlaOp one, xla::XlaOp two) {
-    return xla::ShiftLeft(one, two, XlaHelpers::getBroadcastDimensions(one, two));
+    return xla::ShiftLeft(one, two,
+                          XlaHelpers::getBroadcastDimensions(one, two));
   });
 }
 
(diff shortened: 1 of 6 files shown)


tengyifei (approved):

torch_xla/csrc/ops/ops_xla_shape_fn.cpp
```diff
@@ -360,6 +360,14 @@ xla::Shape BitwiseXorTensorOutputShape(const torch::lazy::Value& input,
   });
 }
 
+xla::Shape BitwiseLeftShiftTensorOutputShape(const torch::lazy::Value& input,
+                                             const torch::lazy::Value& other) {
+  return InferBinaryOpShape(input, other, [](xla::XlaOp one, xla::XlaOp two) {
+    return xla::ShiftLeft(one, two,
+                          XlaHelpers::getBroadcastDimensions(one, two));
```
pgmoka, line 367: If this is just a bitwise shift to the left, wouldn't the output shape be the same as the input? If yes, would `GetXlaShape(input)` work here?

pgmoka (commented):
Could you add a link to https://github.com/pytorch/xla/blob/master/docs/source/contribute/codegen_migration.md as potential reference for what was done here?

Also, I had one question I was curious about. Depending on that, there might be changes I would suggest.

torch_xla/csrc/ops/ops_xla_shape_fn.cpp
ysiraichi: Not necessarily. We are broadcasting the shapes of `one` and `two`, there. Even though it doesn't need to be exactly a `ShiftLeft` call there, we still need something that knows how to broadcast the shapes.

pgmoka: I see.

pgmoka (approved):

PR #8882, opened 2025-03-25
Encapsulate Mesh invariants

rpsilva-aws:
This PR improves the input validation in the Mesh class by making error messages more descriptive. The changes make it easier to debug mesh configuration issues and provide clearer feedback to users. In addition, it moves the invariant encapsulation to the Mesh constructor, as opposed to mark_sharding. This makes it so that we only validate the sharding annotations with respect to the specified Mesh, as opposed to having standalone checks against the number of global participating devices.

### DIFF
--- a/test/spmd/test_xla_sharding.py
+++ b/test/spmd/test_xla_sharding.py
@@ -1,9 +1,10 @@
 import copy
 
-import unittest
-from unittest.mock import patch
+from collections import OrderedDict
 import math
 import numpy as np
+import unittest
+from unittest.mock import patch
 import sys
 
 import torch
@@ -762,9 +763,12 @@ class BasicXlaShardingTest(test_xla_sharding_base.XlaShardingTest):
                    "Crash on TPU v2")
   @patch('torch_xla.runtime.global_runtime_device_attributes')
   @patch('torch_xla.core.xla_model.xla_device_hw')
-  def test_hybrid_mesh(self, xla_device_mock, device_attributes_mock):
+  @patch('torch_xla.runtime.global_runtime_device_count')
+  def test_hybrid_mesh(self, device_count_mock, xla_device_mock,
+                       device_attributes_mock):
     # mock device attributes for 2 slices of v4-8
     num_slices = 2
+    device_count_mock.return_value = 8
     xla_device_mock.return_value = "TPU"
     device_attributes_mock.return_value = [{
         'coords': [0, 0, 0],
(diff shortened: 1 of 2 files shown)


test/spmd/test_xla_sharding.py
```diff
@@ -1565,6 +1566,84 @@ def test_mark_sharding_with_gradients_annotation(self):
...
+  def test_mesh_shape_method(self):
+    mesh_shape = (1, self.n_devices)
+    axis_names = ('data', 'model')
+    mesh = xs.Mesh(self.device_ids, mesh_shape, axis_names)
+
+    expected_shape = OrderedDict([('data', 1), ('model', self.n_devices)])
+    self.assertEqual(mesh.shape(), expected_shape)
+
+  def test_get_logical_mesh(self):
+    mesh_shape = (2, 2)
```
tengyifei, line 1639: This will fail if there are not exactly 4 devices in the environment. Maybe `(2, self.n_devices // 2)`

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -555,16 +566,14 @@ def mark_sharding(t: Union[torch.Tensor, XLAShardedTensor], mesh: Mesh,
       >>> linear = nn.Linear(32, 10).to(xm.xla_device())
       >>> xs.mark_sharding(linear.weight, mesh, (None, 1)) # 2-way model parallel
   """
-  num_devices = xr.global_runtime_device_count()
```
tengyifei, line 558: IIUC right now the XLA compiler has a requirement that the SPMD graph must use all devices in the environment. If we don't have this check, then the user could still get into an assertion failure in XLA if their mesh is too small. @lsy323 's https://github.com/pytorch/xla/pull/8810 aims to loosen the constraint, but until that lands we still need to check that the number of devices in the mesh equals to `xr.global_runtime_device_count()`.

Does Trainium not use XLA/not have this problem somehow?

tengyifei (changes requested):

test/spmd/test_xla_sharding.py
rpsilva-aws: Indeed. However, we're only showcasing the resulting logical mesh, not actually doing any execution here.

torch_xla/distributed/spmd/xla_sharding.py
rpsilva-aws: It does, although this change is independent (minus L558), we would also need to modularize the needed changes to loosen the constraint. In this case, L621 is still helpful to maintain the requirement indirectly. If a user defines a Mesh over only a subset of the devices, then that is the set of devices that mark sharding can derive from the mesh. In the scope of maintaining the same constraint, do we want to move `xr.global_runtime_device_count()` to the Mesh initialization until that lands?

tengyifei: Yes. Let's move the `global_runtime_device_count` check over to Mesh constructor too. Thanks!

rpsilva-aws: Done. We are actually in need / doing the same changes as Siyuan's for PP, so I'll coordinate with Milad depending on the priorities. Thanks!

tengyifei: Thanks!

tengyifei (approved):

tengyifei:
cc @lsy323

rpsilva-aws:
Switched the mesh shape' size (np.prod) assert to the mesh shape' length, since partition spec (0, 1) is still accurate to represent a single device with (1,1) mesh shape.

rpsilva-aws:
> Switched the mesh shape' size (np.prod) assert to the mesh shape' length, since partition spec (0, 1) is still accurate to represent a single device with (1,1) mesh shape (size = 1, len = 2).

We also support an empty tuple for the partition spec for scalar values, as well as specifying ('x', None) for 1D mesh shapes - as long as the tensor shape matches it. @tengyifei in this case, do you see any concern with removing the added expected constraint in https://github.com/pytorch/xla/pull/8882/files#diff-3dcff2b7395bbf1f8a09170775388ef686a1e5f593b3c3889996d78c93a9c394R582 (L582)? Technically, it was meant as a way to enforce a constraint with the specified partition spec and the provided mesh - but that is really an added delta. We just moved the global device check to the mesh, since it's more suitable there.

tengyifei (changes requested):
```
  assert len(partition_spec) == len(mesh.shape()), \
    f"Partition spec length ({len(partition_spec)}) should be equal to the mesh shape dimensions ({len(mesh.shape())})."
```

This check wouldn't make sense. I can shard a 2D tensor over a 4D mesh by e.g. sharding each tensor dim over two mesh axes.

It sounds like you're removing it, and I don't see this check in the latest commit, so that sounds reasonable to me.

However, there's a failed `HybridMesh` test. That test failed because it didn't mock out the `global_runtime_device_count` method. We can probably fix it by mocking out this function to return the desired number of devices for the test.

rpsilva-aws:
Absolutely, agree - thanks. Added the mock, I'll re-request review once the CI succeeds.

tengyifei (approved):

PR #8894, opened 2025-03-27
Update cudnn dependency for newer cuda versions

zpcore:
Fix the issue that 
```
msg: No package matching 'cuda-libraries-12-8' is available
```


ManfeiBai (approved):
Thanks, LGTM

Issue #8831, opened 2025-03-13
Ship torchax inside torch_xla


tengyifei:
## 🚀 Feature

When users install `torch_xla`, they would get `torchax` as well.

## Motivation

The `xb.call_jax` [1] feature under heavy development uses `torchax` and we expect `torchax` usage to grow. In order for all users of torch_xla to use this feature, we need to include `torchax` in `torch_xla` wheels/dockers/etc.

## Pitch

I think there are two approaches:

1. Make `torchax` its own Python wheel/package, and have `torch_xla` depend on it always.

2. Literally link all `torchax` files inside the `torch_xla` wheel/package. This approach might need to deal with some packaging subtleties because some users may have already installed `torchax` standalone and not sure what will happen when they also install `torch_xla` with its own `torchax`.

Either way, this probably requires some setup.py changes.

cc @qihqi 

[1]: https://github.com/pytorch/xla/blob/0aac10e25d66b3d43cf83bc6ffa211eed34e6cc8/torch_xla/core/xla_builder.py#L879

tengyifei:
A big motivation is making atomic changes across torchax and torch_xla. This is only doable if torch_xla includes torchax by source. So the wheel option is probably out.

PR #8893, opened 2025-03-26
Add alternative dynamo backend

qihqi:

PR #8895, opened 2025-03-27
Include torchax in torch_xla

tengyifei:
Fixes https://github.com/pytorch/xla/issues/8831

This PR changes the `torch_xla` build such that `torchax` is automatically included, instead of being a separate install.

This motivation is to support features like https://github.com/pytorch/xla/pull/8893 and others that use `xb.call_jax` and requires `torchax`. Instead of yelling at the user to install `torchax`, it's easier to just install it for them.

Because `torchax` uses `jax`, we'll just automatically install `jax` all the time. As a result, I removed all the `pip install torch_xla[pallas]` and `pip install torchax` commands from our CI scripts in this repo.

This PR should be backwards compatible because any existing standalone `torchax` installs will be overwritten to use the copy from `torch_xla`, and any `pip install torch_xla[pallas]` commands will become no-op instead of failing.

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -115,9 +115,6 @@ function install_post_deps_pytorch_xla() {
   pip install 'torch_xla[pallas]' \
   -f https://storage.googleapis.com/jax-releases/jax_nightly_releases.html \
   -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
-
-  # TODO(https://github.com/pytorch/xla/issues/8831): Remove this when torchax is part of torch_xla.
-  pip install xla/torchax
 }
 
 function build_torch_xla() {
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -106,10 +106,10 @@ function install_pre_deps_pytorch_xla() {
 }
 
 
+# TODO(https://github.com/pytorch/xla/issues/8934): Remove PyTorch usage of this function, then
+# remove this function from the script.
 function install_post_deps_pytorch_xla() {
-  # Install dependencies after we built torch_xla. This is due to installing
-  # those packages can potentially trigger `pip install torch_xla` if torch_xla
-  # is not detected in the system.
+  true
 }
 
 function build_torch_xla() {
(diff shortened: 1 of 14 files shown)


.github/workflows/_test_requiring_torch_cuda.yml
```diff
@@ -91,9 +91,7 @@ jobs:
         shell: bash
         run: |
           set -x
-          pip install -U --pre jaxlib -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
-          pip install -U --pre jax-cuda12-pjrt jax-cuda12-plugin -f https://storage.googleapis.com/jax-releases/jax_cuda_plugin_nightly_releases.html
-          pip install -U --pre jax -f https://storage.googleapis.com/jax-releases/jax_nightly_releases.html
+          pip install -U --pre jax jaxlib "jax-cuda12-plugin[with_cuda]" jax-cuda12-pjrt -f https://storage.googleapis.com/jax-releases/jax_nightly_releases.html
```
zpcore, line 94: Are we missing the urls here?

qihqi (approved):

.circleci/common.sh
```diff
@@ -110,14 +110,6 @@ function install_post_deps_pytorch_xla() {
...
   # those packages can potentially trigger `pip install torch_xla` if torch_xla
   # is not detected in the system.
-
-  # Install JAX dependency since a few tests depend on it.
-  pip install 'torch_xla[pallas]' \
-  -f https://storage.googleapis.com/jax-releases/jax_nightly_releases.html \
-  -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html
-
-  # TODO(https://github.com/pytorch/xla/issues/8831): Remove this when torchax is part of torch_xla.
-  pip install xla/torchax
```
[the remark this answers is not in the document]
tengyifei: done: https://github.com/pytorch/xla/issues/8934

.github/workflows/_test_requiring_torch_cuda.yml
tengyifei: Actually the old URLs are obsolete. I followed exactly JAX's nightly install instruction from https://docs.jax.dev/en/latest/installation.html#jax-nightly-installation.

BTW IMO installing jax this way in CI is also wrong IMO. This means we're not pinning JAX and our CI will be non-deterministic. I was gonna fix this in https://github.com/pytorch/xla/pull/8919.

zpcore: I see, thanks for cleaning this up!

PR #8903, opened 2025-03-28
Fix cuda dependency not found

zpcore:
We are trying to align with PyTorch to support CUDA 11.8, 12.6 and 12.8 for 2.7 release. 

However, 12.8 only shows up in debian12: https://developer.download.nvidia.com/compute/cuda/repos/debian12/x86_64/. Originally we use debian11 for docker image the and here is the support list: https://developer.download.nvidia.com/compute/cuda/repos/debian11/x86_64/.

This PR update the docker image to debian12 and the corresponding CUDA library.

### DIFF
--- a/infra/tpu-pytorch-releases/artifacts_builds.tf
+++ b/infra/tpu-pytorch-releases/artifacts_builds.tf
@@ -8,7 +8,7 @@ locals {
   nightly_package_version = "2.8.0"
   cuda_versions = {
     "nightly": ["11.8", "12.1", "12.6", "12.8"],
-    "r2.7": ["11.8", "12.6", "12.8"] # align with PyTorch 2.7 release
+    "r2.7": ["11.8", "12.1", "12.6", "12.8"] # PyTorch 2.7 release only needs 11.8, 12.6, 12.8
   }
 
   # Built once a day from master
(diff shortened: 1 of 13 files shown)


zpcore:
Hi @ysiraichi , I updated to `debian12` as you suggested. Now the cuda version can be discovered. Thanks

ysiraichi:
Could you add a PR description with more details on what's the problem it's solving? And why this works?
It's good if we wonder why we did this in the future.

infra/tpu-pytorch-releases/artifacts_builds.tf
```diff
@@ -8,7 +8,7 @@ locals {
   nightly_package_version = "2.8.0"
   cuda_versions = {
     "nightly": ["11.8", "12.1", "12.6", "12.8"],
-    "r2.7": ["11.8", "12.6", "12.8"] # align with PyTorch 2.7 release
+    "r2.7": ["11.8", "12.1", "12.6", "12.8"] # PyTorch 2.7 release only needs 11.8, 12.6, 12.8
   }
```
ManfeiBai, line 12: why we need 12.1 here?

ManfeiBai (approved):
Thanks, LGTM

zpcore:
I think I will revert to debian11 for now. Update to debian12 is too risky not to mention the dependency is still failing.

PR #8905, opened 2025-03-28
[Draft] Add Experimental limited sparse embedding bag

### LINKED ISSUE #8719: Support for enabling sparse gradients in EmbeddingBag
## 🚀 Feature
Support for enabling sparse gradients in [EmbeddingBag](https://pytorch.org/docs/stable/generated/torch.nn.functional.embedding_bag.html). 

## Motivation

Adding support for sparse gradients will allow to fit larger embedding tables on the TPU. 

## Pitch

I encountered the following error when turning on the `sparse=True` in [EmbeddingBag](https://pytorch.org/docs/stable/generated/torch.nn.functional.embedding_bag.html) API. 

```
NotImplementedError: Could not run 'aten::_sparse_coo_tensor_with_dims_and_tensors' with arguments from the 'SparseXLA' backend. This could be because the operator doesn't exist for this backend, or was omitted during the selective/custom build process (if using custom build). If you are a Facebook employee using PyTorch on mobile, please visit https://fburl.com/ptmfixes for possible resolutions. 'aten::_sparse_coo_tensor_with_dims_and_tensors' is only available for these backends: [XLA, Meta, SparseCPU, SparseCUDA, SparseMeta, BackendSelect, Python, FuncTorchDynamicLayerBackMode, Functionalize, Named, Conjugate, Negative, ZeroTensor, ADInplaceOrView, AutogradOther, AutogradCPU, AutogradCUDA, AutogradHIP, AutogradXLA, AutogradMPS, AutogradIPU, AutogradXPU, AutogradHPU, AutogradVE, AutogradLazy, AutogradMTIA, AutogradPrivateUse1, AutogradPrivateUse2, AutogradPrivateUse3, AutogradMeta, AutogradNestedTensor, Tracer, AutocastCPU, AutocastXPU, AutocastXLA, AutocastMPS, AutocastCUDA, FuncTorchBatched, BatchedNestedTensor, FuncTorchVmapMode, Batched, VmapMode, FuncTorchGradWrapper, PythonTLSSnapshot, FuncTorchDynamicLayerFrontMode, PreDispatch, PythonDispatcher].
```

Without this flag, memory consumption doubles during training (Embedding table memory size + gradients for the same). If we implement the support for the sparse gradients we can almost double the embedding_dim of any model on the same hardware (provided it doesn't exceed the HBM).  

## Alternatives

Alternative is to just keep using EmbeddingBag with `sparse=False`.

amjames:
Users of `torch_xla` encounter an issue when using the `sparse=True` option with the `Embedding` or `EmbeddingBag` modules. 

The gradient for weight is created as a sparse tensor and there is no dispatch registered for the combination of sparse creation APIs w/ the `XLA` key, or the `Sparse` functionality key and the `XLA` backed key used in conjunction. 

This is a workaround that can be removed, ported to C++, or extended later:

- SparseCOOTensor: a tensor subclass implementing the optimization and semantics of upstream SparseTensor. it is Composabile with the XLA device.
- drop in replacements for `F.embedding` `F.embedding_bag`, `nn.Embedding`, and `nn.EmbeddingBag` which forward to a custom implementation of the backward and produce the above tensor subclass rather than a native torch sparse tensor. 

The tensor subclass may have component tensors `indices` and `values` which have xla device without issue. 

fixes #8719


qihqi:
Hi @amjames if this passes test and is finished feel free to publish as PR and merge.

test/run_tests.sh
```diff
@@ -261,6 +261,10 @@ function run_xla_op_tests3 {
   # Test examples
   run_test "$CDIR/../examples/scan/scan_examples.py"
 
+  # Experimental sparse
+  run_test "$CDIR/test_sparse_coo.py"
```
amjames, line 265: Need to add this file.

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,327 @@
...
+def _register_dispatch_func(op):
+
+  def wrapper(f):
+    _COO_DISPATCH_TABLE[op] = f
+    return f
+
+  return wrapper
+
+
+def _register_function_func(op):
```
amjames, line 23: Need to remove this, and the `__torch_function__` registrations. They are not a solution to the problem. Things are far more broken if we have a `__torch_function__` override.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -4364,4 +4364,19 @@ at::Tensor XLANativeFunctions::view_symint(const at::Tensor& self,
       bridge::GetXlaTensor(self), XlaHelpers::I64List(size)));
 }
 
+// Sparse mask and coalesce are sparse dispatch only. These exist to prevent
+// the functionalization layer from incorrectly applying wrappers when we
```
bhavya01, line 4368: Can you help with more context here? What does sparse_mask do and why is it incorrect to apply a functionalization wrapper?

ysiraichi: It would be incorrect to apply the functionalization wrapper because their return value should be a `SparseCOOTensor` which is not a functional tensor. In the functionalization layer (fallback function), PyTorch wraps the return value to a functional tensor, and that is not supposed to happen.

torch_xla/csrc/aten_xla_type.cpp
```diff
@@ -4364,4 +4364,19 @@ at::Tensor XLANativeFunctions::view_symint(const at::Tensor& self,
...
+// pass through before hitting python dispatch.
+at::Tensor XLANativeFunctions::sparse_mask(const at::Tensor& self,
+                                           const at::Tensor& mask) {
+  at::AutoDispatchSkipFunctionalize guard;
+  at::AutoDispatchBelowAutograd guard2;
+  return self.sparse_mask(mask);
```
ysiraichi, line 4374: Could you try using `at::_ops::sparse_mask::redispatch({XLA}, ...)`, instead? I think this would be better because:

1. We wouldn't have to worry about using view ops down the line.
2. I think that this call is invisible to autograd

torch_xla/experimental/sparse/coo.py
```diff
@@ -0,0 +1,103 @@
...
+    dtype = dtype if dtype is not None else values.dtype
+    # We must use the autograd function to create in an autograd context, see [_coo_ops.make_sparse, or SparesCOOTensor.create]
+    assert not requires_grad or not values.requires_grad or not torch.is_grad_enabled(
+    )
+    assert (device == values.device and device == indices.device and
+            dtype == values.dtype)
+
+    # Passing layout to _make_wrapper_subclass will trigger the Sparse modifier
+    # being added to the dispatch key set. We want to avoid this so that on
+    # dispatch we don't hit an overlaod resolution error if things try to hit
```
ysiraichi, line 44: typo: overload

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,316 @@
...
+  _check_no_kwargs(kwargs, "resize_as_")
+
+  new_nnz = other._nnz()
+  values_shape = (new_nnz,)
+  index_shape = (len(other.shape), new_nnz)
+
+  # attribute access to modify the tensor in-place, accessors return a clone
+  self._v.resize_(values_shape)
+  self._i.resize_(index_shape)
+  return self
```
ysiraichi, line 127: This implementation is a bit unexpected for me. I thought that calling `self.resize_as_(other)` would result in `self.shape == other.shape`.

However, it looks like this implementation is updating the number of `nnz` for `self`. Is this the expected behavior?

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,316 @@
...
+    # args will be none except values which we want to send through
+    return args
+
+
+def make_sparse(indices, values, size, dtype=None, device=None):
+  return _SparseMaker.apply(indices, values, size, dtype, device)
+
+
+@_register_dispatch_func(aten.sparse_mask)
+def coo_sparse_mask(args=(), kwargs=None):
```
ysiraichi, line 63: Is there any reason for not specifying the parameters, here (i.e. instead of `args, kwargs` parameters)? I feel like that would be more readable. Then, you could just call them as `fn(*args, **kwargs)`.

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,316 @@
...
+  self._v.resize_(values_shape)
+  self._i.resize_(index_shape)
+  return self
+
+
+@_register_dispatch_func(aten.is_coalesced)
+def coo_is_coalesced(args, kwargs=None):
+  _check_no_kwargs(kwargs, "is_coalesced")
+  self = args[0]
+  _check_sparse(self, "is_coalesced", "self")
```
ysiraichi, line 134: Do we need this check? If we got dispatched here, doesn't it mean that `self` is sparse?

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+@_register_dispatch_func(aten.add)
+def coo_add(args=(), kwargs=None):
+  alpha = kwargs.pop('alpha', 1)
+  torch._check(
+      len(args) == 2, lambda: f"add: expected two operands, got {len(args)}")
+  self, other = args
+  _check_sparse(other, "other", "add")
+
```
ysiraichi, line 196: Shouldn't we check if the sparsity pattern is the same here, too?

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,316 @@
...
+    ret = torch.to_dense(self)
+  else:
+    need_copy = dtype != self.dtype or device != self.device or layout != self.layout
+    copy = kwargs.get('copy', need_copy)
+    non_blocking = kwargs.get('non_blocking', False)
+    if copy:
+      # otherwise ignore non_blocking
+      torch._check(not non_blocking,
+                   lambda: f"to: sparse_coo does not support async copy")
+      ret = torch._to_copy(
```
ysiraichi, line 308: I don't think this works. I'm getting: `AttributeError: module 'torch' has no attribute '_to_copy'`

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,316 @@
...
+  self = args[0]
+  _check_sparse(self, "to", "self")
+  dtype = kwargs.get('dtype', self.dtype)
+  device = kwargs.get('device', self.device)
+  layout = kwargs.get('layout', self.layout)
+  if layout != self.layout:
+    torch._check(layout == torch.strided,
+                 lambda: f"to: only layout conversion to strided is supported")
+    # .to(dtype) is no op if dtype is not changed
+    ret = torch.to_dense(self)
```
ysiraichi, line 299: I think you are not returning `ret` on this execution path. Also, if we end up here, I don't think we are handling different `dtype` and `device`.

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,316 @@
...
+                   lambda: f"to: sparse_coo does not support async copy")
+      ret = torch._to_copy(
+          args=(self,), kwargs={
+              'dtype': dtype,
+              'device': device
+          })
+    else:
+      # if no conversion is needed to returns an alias to this
+      ret = self
+    return ret
```
ysiraichi, line 316: Maybe this should be less indented?

torch_xla/experimental/sparse/_coo_ops.py
```diff
@@ -0,0 +1,316 @@
...
+  return make_sparse(new_i, new_v, self.shape)
+
+
+@_register_dispatch_func(aten.to)
+def coo_to(args=(), kwargs=None):
+  self = args[0]
+  _check_sparse(self, "to", "self")
+  dtype = kwargs.get('dtype', self.dtype)
+  device = kwargs.get('device', self.device)
+  layout = kwargs.get('layout', self.layout)
```
ysiraichi, line 294: Suggestion: I think [PyTorch core `to`][1] has a nice implementation, where it:

- Checks whether we actually need to do anything
- Calls `_to_copy` otherwise

Question: can't we use PyTorch core `to()` implementation?

[1]: https://github.com/pytorch/pytorch/blob/523b637cbeb69665072a2cf489ec1c5313b57670/aten/src/ATen/native/TensorConversions.cpp#L425-L447

torch_xla/experimental/sparse/embedding_bag.py
```diff
@@ -0,0 +1,346 @@
...
+      index_grad = grad_.index_select(0, offset2bag)
+      index_grad = _apply_bag_size_backward(ctx.mode, index_grad, offset2bag,
+                                            bag_size)
+      if per_sample_weights is not None:
+        torch._check(ctx.mode == _EmbeddingBagMode.SUM)
+        index_grad.mul_(per_sample_weights.unsqueeze(1))
+        breakpoint()
+      grad_weight = _sparse_embedding_backward(index_grad, indices,
+                                               ctx.num_weights, ctx.padding_idx)
+      return grad_weight, None, None, None, None, None, None, None, None
```
ysiraichi, line 130: Shouldn't this be less indented? i.e. if `grad_` is None, should we return None or a tuple of Nones (like `SparseEmbedding`?

ysiraichi (commented):
Awesome PR, Andrew. Thank you.
I left a few comments. Let me know if you need more information.

ysiraichi:
I think it would be nice to have tests for each operation implemented. What do you think?

torch_xla/experimental/sparse/_coo_ops.py
amjames: This is a good catch! The shape of `self` is inferred from the shapes of values and indices in the upstream SparseImpl, but of course that logic does not apply here. I will need to fix this. Actually this should have broken something if we need it, so we might be able to remove it.

torch_xla/experimental/sparse/_coo_ops.py
amjames: This is for safety in the future if someone else updates this we could easily land in a situation where the wrong layout is placed in the SparseCOO subclass wrapper.

Following upstream convention sparse means sparse_coo so this would fail for other sparse layouts.

torch_xla/experimental/sparse/_coo_ops.py
amjames: In this case self is dense and other is sparse. That is the only combination needed for the optimizer update. 

This would be much more clear if I expanded the signatures like you suggested. I will do that.

PR #8911, opened 2025-03-31
Adapt Splash Attention from TorchPrime

zpcore:
Adapt the PR https://github.com/AI-Hypercomputer/torchprime/pull/145 from TorchPrime into PTXLA. Also simplified the code to use jit hashing from https://github.com/pytorch/xla/pull/8878.

In addition, fix a small bug in xla_builder.call_jax when the input arg contains both None and other hashable types in sequence.


torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+  sa_block_kv_dkv_compute: int = 2048
+  sa_block_q_dq: int = 2048
+  sa_block_kv_dq: int = 2048
+  sa_use_fused_bwd_kernel: bool = True
+  sa_q_layout: str = "HEAD_DIM_MINOR"
+  sa_k_layout: str = "HEAD_DIM_MINOR"
+  sa_v_layout: str = "HEAD_DIM_MINOR"
+  mesh: str | None = None
+  # Check more rules from MaxText config:
+  # https://github.com/AI-Hypercomputer/maxtext/blob/462087ed90a60485a145e909e047bacc28397f82/MaxText/configs/base.yml#L261.
```
tengyifei, line 35: What does this comment mean? We don't use logical axis rules anymore yah?

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+import torch
+import torch_xla.debug.profiler as xp
+from torch.library import custom_op
+from torch.utils._pytree import tree_flatten
+from torch_xla.core.xla_builder import call_jax
+from torch_xla.distributed.spmd import Mesh
+from torch_xla.experimental.custom_kernel import requires_jax
+
+
+@dataclasses.dataclass
```
tengyifei, line 16: We can add eq=True and hash=True to the dataclasses decorator call. This way any instance of this config will be hashable and can be used as an argument to call_jax, avoiding the lru cache.

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+  """
+  import jax
+  from jax.experimental import shard_map
+  from jax.experimental.pallas.ops.tpu.splash_attention import (
+      splash_attention_kernel,
+      splash_attention_mask,
+  )
+
+  mesh = config.maybe_convert_and_get_jax_mesh()
+  # input q,k,v shape: [batch, #head, seq_len, kv]
```
tengyifei, line 118: Should the last dim be called "head or "head_dim"?

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+      out_specs=axis_names,
+      check_rep=False,
+  )
+  def wrap_flash_attention(query, key, value, decoder_segment_ids):
+    if decoder_segment_ids is not None:
+      assert (
+          query.shape[2] == decoder_segment_ids.q.shape[1]
+      ), "Sharding along sequence dimension not allowed in tpu kernel attention"
+    block_sizes = splash_attention_kernel.BlockSizes(
+        block_q=min(global_block_q, query.shape[2]),
```
tengyifei, line 160: Could we factor out a seq_len variable

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+      ("data", "fsdp"),
+      None,
+      None,
+      None,
+  )
+  segment_ids_partition_spec: tuple[tuple[str] | str | None] = (
+      ("data", "fsdp"),
+      None,
+  )
+  AttentionType_LOCAL_SLIDING: bool = False
```
tengyifei, line 46: nit: field names should be simple snake case

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+  SLIDE_WINDOW_SIZE: int | None = None
```
tengyifei, line 47: ditto

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+  assert (query.shape[0] / devices_in_data_fsdp).is_integer(), (
+      "Batch dimension should be shardable among the devices in data and fsdp axis"
+  )
+  x = wrap_flash_attention(query, key, value, decoder_segment_ids)
+  # x.shape = [batch, heads, seq_length, head_dim]
+  return x
+
+
+@functools.lru_cache(maxsize=16)
+def _get_jax_forward_function(config_json: str, attn_logits_soft_cap,
```
tengyifei, line 213: I think we could get rid of this wrapper. What happens is that call_jax now can take both arguments and keyword arguments. Additionally, each argument can either be a tensor or a hashable type. If we can make the config object hashable, then that config can be an additional argument into the Jax function, and we don't need to use local closures to capture it. Similarly, I think the segment ID optionality could be handled by passing different things in the kwargs object to the call_jax.

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+def _get_jax_backward_function(config_json: str, attn_logits_soft_cap,
+                               has_segment_ids):
+  """Cached factory function to create JAX backward functions"""
+  jax_f = _get_jax_forward_function(config_json, attn_logits_soft_cap,
+                                    has_segment_ids)
+  import jax
+
+  if has_segment_ids:
+
+    def jax_grad_f_wrapper(query, key, value, decoder_segment_ids, grad_output):
```
tengyifei, line 242: Similar to the forward case, I think this could be made into a global function that takes additional arguments such as the config object and the decoder_segment_ids as a kwargs, this avoiding the need for an additional caching layer.

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+                                                decoder_segment_ids,
+                                                attn_logits_soft_cap)
+    return grad_q, grad_k, grad_v, None, None, None
+
+
+def splash_attention(
+    q: torch.Tensor,
+    k: torch.Tensor,
+    v: torch.Tensor,
+    config: str,
```
tengyifei, line 392: While custom_ops can only take strings, could we make the main API take a proper SplashAttentionConfig object? That has nicer/more type safety.

test/test_splash_attention_jax.py
```diff
@@ -0,0 +1,287 @@
...
+
+  def maybe_expend_kv(self, hidden_state):
+    if hidden_state.size(1) == self.NUM_Q_HEADS:
+      return hidden_state
+    num_kv_group = self.NUM_Q_HEADS // self.NUM_KV_HEADS
+    return hidden_state.repeat_interleave(num_kv_group, dim=1)
+
+  def maybe_reduce_kv_grad(self, hidden_state_grad):
+    # For GQA, the kv grad shape is [BATCH_SIZE, NUM_Q_HEADS, SEQ_LEN,
+    # HEAD_DIM]. We need to convert it back to [BATCH_SIZE, NUM_Q_HEADS,
```
tengyifei, line 93: Convert back to NUM_KV_HEADS?

test/test_splash_attention_jax.py
```diff
@@ -0,0 +1,287 @@
...
+    torch_xla.sync()
+    loss = torch.sum(self.o)
+    loss.backward()
+    torch_xla.sync()
+    self.q_grad, k_grad, v_grad = self.q.grad, self.k.grad, self.v.grad
+    with torch.no_grad():
+      self.k_grad = self.maybe_reduce_kv_grad(k_grad)
+      self.v_grad = self.maybe_reduce_kv_grad(v_grad)
+
+  def maybe_expend_kv(self, hidden_state):
```
tengyifei, line 85: nit: expand, or repeat

test/tpu/run_tests.sh
```diff
@@ -41,6 +41,7 @@ run_xla_hlo_debug python3 "$TEST_CDIR/scan/test_scan_debug.py"
 python3 "$TEST_CDIR/test_pallas.py" -v
 python3 "$TEST_CDIR/test_pallas_spmd.py"
 XLA_DISABLE_FUNCTIONALIZATION=1 python3 "$TEST_CDIR/test_pallas_spmd.py"
+python3 "$TEST_CDIR/test_splash_attention_jax.py"
```
tengyifei, line 44: nit: Jax is just an implementation detail. We could simply call this file test_splash_attention.py

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
+import dataclasses
```
tengyifei, line 1: nit: I think we could call this splash_attention.py to be more specific. If there are generic Jax utilities, those could be put in a custom_kernels_from_jax.py or similar

torch_xla/experimental/custom_kernel_from_jax.py
```diff
@@ -0,0 +1,397 @@
...
+                                                attn_logits_soft_cap)
+    return grad_q, grad_k, grad_v, None, None, None
+
+
+def splash_attention(
+    q: torch.Tensor,
+    k: torch.Tensor,
+    v: torch.Tensor,
+    config: str,
+    decoder_segment_ids: torch.Tensor | None = None,
```
tengyifei, line 393: Could we at least document the segment IDs thing and the soft cap in triple quote style doc comments? 

Maybe also document that splash attention vs flash attention, e.g. splash attention can be faster if the attention mask is sparse by skipping blocks, etc etc

tengyifei (changes requested):
Nice!

torch_xla/experimental/custom_kernel_from_jax.py
zpcore: Right, I think we can just remove this.

torch_xla/experimental/custom_kernel_from_jax.py
zpcore: The naming is actually from https://github.com/AI-Hypercomputer/maxtext/blob/d8ffb5c4fc65e6832976226a8053236c2fe3164e/MaxText/layers/attentions.py#L345, but I think `head_dim` should be better here. Updated all `kv` naming.

test/test_splash_attention_jax.py
zpcore: Nice catch!

test/tpu/run_tests.sh
zpcore: Renamed.

torch_xla/experimental/custom_kernel_from_jax.py
zpcore: SGTM

tengyifei:
Looks like some comments still need to be addressed -- LMK whenever I should TAL!

zpcore:
> Looks like some comments still need to be addressed -- LMK whenever I should TAL!

Yes, I am working on getting rid of the `lru_cache`. Need to fix some small issues before resolving the feedback. Thanks!

torch_xla/experimental/custom_kernel_from_jax.py
zpcore: Nice suggestion! Removed the lru_cache and added the cache miss test.

torch_xla/experimental/custom_kernel_from_jax.py
zpcore: Updated

torch_xla/experimental/splash_attention.py
```diff
@@ -0,0 +1,371 @@
...
+                                                decoder_segment_ids,
+                                                attn_logits_soft_cap)
+    return grad_q, grad_k, grad_v, None, None, None
+
+
+def splash_attention(
+    q: torch.Tensor,
+    k: torch.Tensor,
+    v: torch.Tensor,
+    config: str,
```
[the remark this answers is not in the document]
zpcore: Yes, this works fine when we call `splash_attention` directly. However, I notice when we do:
```
compiled_splash_attention = aot_function(
        splash_attention, fw_compiler=compiler)
```
`compiled_splash_attention` will not accept `SplashAttentionConfig` type as argument.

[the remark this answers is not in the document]
tengyifei: Ack. NVM then.

Sigh... why is AOTAutograd so hard to use

zpcore:
Oh, interesting that the test failed for the cache miss count. Looks like the HLO cache can be reused between test functions.

zpcore:
Hi @tengyifei , I created issue https://github.com/pytorch/xla/issues/8963 to track the hashing issue. Will follow up in a separate PR for the fix.

test/test_splash_attention.py
```diff
@@ -0,0 +1,327 @@
...
+                                 [q_grad_sa, k_grad_sa, v_grad_sa],
+                                 strict=False):
+      torch.testing.assert_close(
+          org_grad.cpu(), sa_grad.cpu(), rtol=1e-4, atol=1e-2)
+
+  @unittest.skipIf(xr.device_type() != "TPU" or tpu.version() < 3,
+                   "This test only works on TPUv3+.")
+  @with_jax_high_precision  # remove the decorator will cause failure in other tests :)
+  def test_splash_attention_cache_hit(self):
+    xb._JAX_TO_HLO_CACHE.clear()
```
tengyifei, line 279: I think we don't have to clear the cache then.

test/test_splash_attention.py
```diff
@@ -0,0 +1,327 @@
...
+    torch.testing.assert_close(self.o.cpu(), o_sa.cpu(), rtol=1e-3, atol=1e-5)
+    for org_grad, sa_grad in zip([self.q_grad, self.k_grad, self.v_grad],
+                                 [q_grad_sa, k_grad_sa, v_grad_sa],
+                                 strict=False):
+      torch.testing.assert_close(
+          org_grad.cpu(), sa_grad.cpu(), rtol=1e-4, atol=1e-2)
+
+  @unittest.skipIf(xr.device_type() != "TPU" or tpu.version() < 3,
+                   "This test only works on TPUv3+.")
+  @with_jax_high_precision  # remove the decorator will cause failure in other tests :)
```
tengyifei, line 277: I think this comment can be removed because once we fix #8963, it will be invalidated :)

tengyifei (approved):
SGTM / LGTM

test/test_splash_attention.py
tengyifei: Also _JAX_TO_HLO_CACHE was renamed to _JAX_TO_XLA_COMPUTATION_CACHE

test/test_splash_attention.py
zpcore: Right, I will add a test case for call_jax later.

test/test_splash_attention.py
zpcore: Still need to clear it to be safe, or `ending_cache_misses - starting_cache_misses` will be 0 due to the caching from other tests.

tengyifei: Oh, sorry, you are absolutely right.

Issue #8860, opened 2025-03-19
Replace upstream PyTorch GRU module


tengyifei:
Now that we have a GRU implemented with scan (https://github.com/pytorch/xla/pull/8777), we can figure out a way to patch or intercept the PyTorch upstream GRU module such that it just gets replaced with our own GRU logic. Then the user won't have to know and the upstream GRU just works.

iwknow:
i can take a look at this issue if no one else is assigned.

tengyifei:
Hi @iwknow , thanks for offering to take a look. Here are more details. If you take a look at https://github.com/pytorch/xla/pull/8777, it adds a GRU implementation that is interface-compatible with upstream PyTorch, but not weight compatible. Also it doesn't support batch_first and bidirectional.

My thoughts are:

- We'd need to make the experimental GRU weight key-compatible with PyTorch also. Otherwise existing checkpoints cannot be  loaded.
- Maybe we need to teach the experimental GRU to fallback to upstream GRU somehow if the user passes `batch_first` or `bidirectional`, or support these in the scan-based GRU
- Once we make our GRU fully compatible with upstream, we could add a patching logic to https://github.com/pytorch/xla/blob/master/torch_xla/_patched_functions.py#L60

iwknow:
this issue may be closed.

tengyifei:
Thanks @iwknow !

PR #8914, opened 2025-04-01
Rewrite scan-based GRU based on nn.GRU

iwknow:
This is a preparation step for https://github.com/pytorch/xla/issues/8860. specifically, this PR modified scan-based GRU implementation to derive from standard pytorch GRU implementation. It reuses all the components of the standard pytorch GRU implementation except the `forward` function. Reusing the standard GRU implementation makes the scan-based GRU weight key-compatible, enabling it to load checkpoints of the standard GRU.

The standard pytorch GRU implementation is similar to the scan-based GRU implementation except that it directly add parameters to module instead of using a `ParameterList`. It also uses the common performance optimization that concatenates the gate matrices into one matrix. For details on how it arranges and initialize the parameters, see: https://github.com/pytorch/pytorch/blob/340beb7f7cfe9c3c248819d2aae4170b8def3a8c/torch/nn/modules/rnn.py#L158-L201

Currently, setting `batch_first`, `bidirectional`, and `proj_size` are disabled. i will either make them work or allow them to fallback to the standard implementation in the following PRs.

### DIFF
--- a/test/test_gru.py
+++ b/test/test_gru.py
@@ -1,6 +1,5 @@
 import torch
 import torch.nn as nn
-
 import torch_xla
 from torch_xla.experimental.gru import GRU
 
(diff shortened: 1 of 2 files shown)


iwknow:
@tengyifei

iwknow:
@tengyifei please run the test workflows and take a look. thanks!

tengyifei:
Hi yes, just did. Thank you for the great work. I'll try to review it later today.

iwknow:
it seems that the failed test is not related to this change. should i pull the latest commits and run the tests again?

tengyifei (approved):

PR #8919, opened 2025-04-01
Make GPU CUDA plugin require JAX

tengyifei:
Some XLA GPU features require JAX. Rather than only installing the latest version of JAX in CI, we'll just make the CUDA plugin depend on a version of JAX that's the same as what's used by PyTorch/XLA on TPU. (Except the JAX CUDA wheels).


tengyifei:
This is not easily fixable because JAX 0.6.1 started requiring CuDNN 9.8 (see https://github.com/jax-ml/jax/blob/main/CHANGELOG.md?plain=1#L61), but CuDNN 9.8 requires Debian 12 (https://github.com/pytorch/xla/pull/8928).

PR #8922, opened 2025-04-02
Add heuristic default block sizes for different cases in ragged attention kernel

yaochengji:


test/test_pallas.py
```diff
@@ -856,6 +867,22 @@ def test_ragged_paged_attention_wrapper_without_dynamo(
         use_dynamo=False,
     )
```
bythew3i, line 868: Remove the old _test_ragged_paged_attention?

test/test_pallas.py
```diff
@@ -817,6 +812,22 @@ def test_ragged_paged_attention_wrapper_with_dynamo(
         use_dynamo=True,
```
bythew3i, line 812: Remove the old _test_ragged_paged_attention?

torch_xla/experimental/custom_kernel.py
```diff
@@ -955,6 +955,28 @@ def ragged_paged_attention(
...
+      num_kv_pages_per_block = 128
+      num_queries_per_block = 96
+
+  if vmem_limit_bytes is None:
+    tpu_version = torch_xla.tpu.version()
+    if tpu_version < 4:
+        raise NotImplementedError("TPU version must be 4 or higher.")
+    # NOTE: the TPU v4's vmem capacity is 16MB
+    if tpu_version == 4:
+        vmem_limit_bytes = 16 * 1024 * 1024
```
bythew3i, line 976: It is fine, even we set more than 16MB, it will still use 16 MB

torch_xla/experimental/custom_kernel.py
```diff
@@ -955,6 +955,28 @@ def ragged_paged_attention(
   # in the global scope which could cause problems for xmp.spawn.
   from torch_xla.experimental.pallas_kernels.ragged_paged_attention_v2 import ragged_paged_attention as ragged_attention
 
+  if num_kv_pages_per_block is None and num_queries_per_block is None:
+    token_num = q.shape[0]
+    # This heristic is based on the initial kernel micro benchmarking:
+    # When the token_num is small, there's no long request of prefill.
+    # While when it's larger, the block size is adjusted for it.
+    if token_num <= 128:
```
vanbasten23, line 963: I wonder if we should choose the block sizes in vLLM. If we choose to do in torch_xla, then we need to change it in torch_xla and wait for the wheel tmr. If we do in vLLM, it'd be more convenient. wdyt?

test/test_pallas.py
yaochengji: We'd better also test the non-None block size parameter, that's why it's still there.

test/test_pallas.py
yaochengji: We'd better also test the non-None block size parameter, that's why it's still there.

torch_xla/experimental/custom_kernel.py
yaochengji: Thanks, done.

test/test_pallas.py
```diff
@@ -817,6 +812,22 @@ def test_ragged_paged_attention_wrapper_with_dynamo(
...
+        head_dim,
+        page_size,
+        num_pages,
+        dtype,
+        sm_scale=sm_scale,
+        sliding_window=sliding_window,
+        soft_cap=soft_cap,
+        pad_tokens_and_seqs=pad_tokens_and_seqs,
+        use_dynamo=True,
+        num_kv_pages_per_block=None,
```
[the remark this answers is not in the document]
yaochengji: Thanks for the suggestion, done.

torch_xla/experimental/custom_kernel.py
yaochengji: If vLLM pass a non-None block size, the default value will not be used.

vanbasten23: I wonder why we couldn't do `_get_default_ragged_paged_attention_block_size` in vLLM..

yaochengji: Usually it's a good idea to put the tuned-parameter table in the kernel lib, not in the app lib.

vanbasten23 (approved):
Thanks Chengji!

PR #8938, opened 2025-04-04
Trigger build for 2.6.1 patch

zpcore:


tengyifei (approved):

zpcore:
Hmm, the GPU test fails from the HEAD. We have to wait for the GPU fix first.

tengyifei:
> Hmm, the GPU test fails from the HEAD. We have to wait for the GPU fix first.

There are no GPU wheels for 2.6 and in any case, the GPU tests are not required to pass to merge a PR because they break too much rn.

zpcore:
> > Hmm, the GPU test fails from the HEAD. We have to wait for the GPU fix first.
> 
> There are no GPU wheels for 2.6 and in any case, the GPU tests are not required to pass to merge a PR because they break too much rn.

Yes, I think it doesn't matter. The block is due to missing the review from code owner.

ManfeiBai (approved):
Thanks, LGTM

ManfeiBai:
would you mind update the description too?

PR #8944, opened 2025-04-05
Update to pytorch 2.6

qihqi:
Pytorch 2.6 introduced usage of decompositions via CompositeImplicitAutograd key: https://github.com/pytorch/pytorch/blob/60a45eb862d5e8b4ba2dd435d34ef04ae231e885/torch/_export/utils.py#L1249

These decompositions currently does not play nice with `torch_dispatch` based decomposition, resulting infinite recursion.
Likely we are using it wrong, I made a post here soliciting the right way to use decomposition https://dev-discuss.pytorch.org/t/what-is-the-right-way-to-use-decompositions-in-dispatch-mode/2888 soliciting advice on the topic.

Meanwhile the workaround is to explicitly list out decompositions we like without the CompositeImplicitAutograd decompositions. The explicit list is produced via keys in `core_aten_decompositions()` call using `torch 2.5.1`.

fixes https://github.com/pytorch/xla/issues/8952

### DIFF
--- a/torchax/pyproject.toml
+++ b/torchax/pyproject.toml
@@ -49,8 +49,8 @@ odml = ["jax[cpu]>=0.4.30", "jax[cpu]"]
 [tool.hatch.build.targets.wheel]
 packages = ["torchax"]
 
-[tool.pytest.ini_options]
-addopts="-n auto"
+#[tool.pytest.ini_options]
+#addopts="-n auto"
 
 [tool.ruff]
 line-length = 80
--- a/torchax/pyproject.toml
+++ b/torchax/pyproject.toml
@@ -49,8 +49,8 @@ odml = ["jax[cpu]>=0.4.30", "jax[cpu]"]
 [tool.hatch.build.targets.wheel]
 packages = ["torchax"]
 
-#[tool.pytest.ini_options]
-#addopts="-n auto"
+[tool.pytest.ini_options]
+addopts="-n auto"
 
 [tool.ruff]
 line-length = 80
--- a/torchax/test/test_ops.py
+++ b/torchax/test/test_ops.py
@@ -129,7 +129,6 @@ def run_export_and_compare(testcase,
                            equal_nan=True,
                            ignore_indices=False):
   atol, rtol = (1e-3, 1e-5)
-  #breakpoint()
   if func.name in atol_dict:
     atol, rtol = atol_dict[func.name]
 
--- a/torchax/test/test_ops.py
+++ b/torchax/test/test_ops.py
@@ -129,6 +129,7 @@ def run_export_and_compare(testcase,
                            equal_nan=True,
                            ignore_indices=False):
   atol, rtol = (1e-3, 1e-5)
+  #breakpoint()
   if func.name in atol_dict:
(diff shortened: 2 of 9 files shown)


torchax/pyproject.toml
```diff
@@ -49,8 +49,8 @@ odml = ["jax[cpu]>=0.4.30", "jax[cpu]"]
 [tool.hatch.build.targets.wheel]
 packages = ["torchax"]
 
-[tool.pytest.ini_options]
-addopts="-n auto"
+#[tool.pytest.ini_options]
+#addopts="-n auto"
```
tengyifei, line 53: Is this intended change?

torchax/test/test_ops.py
```diff
@@ -129,6 +129,7 @@ def run_export_and_compare(testcase,
                            equal_nan=True,
                            ignore_indices=False):
   atol, rtol = (1e-3, 1e-5)
+  #breakpoint()
```
tengyifei, line 132: Is this intended change?

tengyifei (approved):

tengyifei:
I wonder if it's possible to eliminate all autograd-related dispatch keys when calling these torch ops, since we just need decompositions in the forward direction and can rely on Jax autograd, I think..

It maybe runs into other PyTorch problems

torchax/torchax/ops/jaten.py
```diff
@@ -451,12 +453,6 @@ def repeat_interleave(self, repeats, dim=0):
   return jnp.repeat(self, repeats, dim, total_repeat_length=total_repeat_length)
 
 
-# aten.upsample_bilinear2d
-@op(torch.ops.aten.upsample_bilinear2d)
-def _aten_upsample_bilinear2d(x, output_size, align_corners=False, scale_h=None, scale_w=None):
-  return _aten_upsample_bilinear2d_aa(x, output_size=output_size, align_corners=align_corners, scale_factors=None, scales_h=scale_h, scales_w=scale_w)
-
```
ManfeiBai, line 458: why we skip/delete this implementation?

torchax/torchax/ops/jaten.py
```diff
@@ -987,39 +983,22 @@ def _aten_gelu(self, *, approximate="none"):
...
-    dim: The dimension to squeeze.
-
-  Returns:
-    The squeezed tensor with the specified dimension removed if it is 1,
-    otherwise the original tensor is returned.
-  """
-
-  # Validate input arguments
-  if not isinstance(self, jnp.ndarray):
-    raise TypeError(f"Expected a Jax tensor, got {type(self)}.")
```
ManfeiBai, line 1004: do we skip this type check due to 2.6 adoption?

ManfeiBai (approved):
Thanks, LGTM

torchax/torchax/ops/jaten.py
qihqi: there is a generic check in `tensor.py`

torchax/torchax/ops/jaten.py
qihqi: it's not called in practice. The decompositon one is the one getting called. It also produces wrong result.

PR #8950, opened 2025-04-08
Use pages_per_seq * page_size instead of directly passing max_model_len

yaochengji:


yaochengji:
Credit to @bythew3i for the suggestion!

vanbasten23 (approved):

PR #8955, opened 2025-04-10
Remove deprecated typing module in ansible flow

tengyifei:
Remove deprecated typing module (it's built into Python now).

### DIFF
--- a/infra/ansible/config/pip.yaml
+++ b/infra/ansible/config/pip.yaml
@@ -26,7 +26,6 @@ pip:
       - tensorboard
       - tensorboardX
       - tqdm
-      - typing
       - typing_extensions
       - sympy
       - yapf==0.30.0


zpcore:
Curious what is the build error you see?

zpcore (approved):

tengyifei:
> Curious what is the build error you see?

The error was

```
Step #2 - "build_xla_docker_image":   stdout: |-
Step #2 - "build_xla_docker_image":     Building wheel torch-2.8.0
Step #2 - "build_xla_docker_image":     invalid command 'bdist_wheel'
```

So the PyTorch upstream setup.py does not support `bdist_wheel`. The LLM told me it could be a problem in the wheel package. I didn't find what's the problem, but I saw this unrelated tech debt which is we are installing an unnecessary `typings` package. Hence fix here.

PR #8958, opened 2025-04-10
Add runtime check when using non-kernel for ragged paged attn

bythew3i:
I explored adding jax checkify directly into the kernel file (torch_xla/experimental/pallas_kernels/ragged_paged_attention_v2.py). **This does not work because the jax-checkified function will return a RuntimeError which is required to throw during runtime. So we can not wrap the jax-checkified function under a pytorch function that will be compiled as well.**

The best (probably the easiest) solution we can do is to add these runtime checks in non-kernel version. So instead of creating a Debug flag and toggling it, we just use `use_kernel=false` to check the input dynamicly during runtime.

Tested: (We need to reduce tests size... it takes 14 min to run)
```
```

### DIFF
--- a/torch_xla/experimental/custom_kernel.py
+++ b/torch_xla/experimental/custom_kernel.py
@@ -862,19 +862,53 @@ def flash_attention(
                               sm_scale, ab, partition_spec, mesh)
 
 
+# This function should only be called and excuted on runtime.
+def _ragged_paged_attention_runtime_check(
+    q,  # [max_num_batched_tokens, num_q_heads, head_dim]
+    kv_pages,  # [total_num_pages, page_size, num_combined_kv_heads, head_dim]
+    kv_lens,  # i32[max_num_seqs]
+    page_indices,  # i32[max_num_seqs, pages_per_seq]
+    cu_q_lens,  # i32[max_num_seqs + 1]
+    num_seqs,  # i32[1]
+):
+  max_num_batched_tokens = q.shape[0]
+  page_size = kv_pages.shape[1]
+  max_num_seqs, pages_per_seq = page_indices.shape
+  if num_seqs[0] > max_num_seqs:
+    raise ValueError(f"{num_seqs[0]=} must be less or equal to {max_num_seqs=}")
+  max_kv_len = torch.max(kv_lens)
+  min_pages_per_seq = (max_kv_len + page_size - 1) // page_size
+  if pages_per_seq < min_pages_per_seq:
+    raise ValueError(
+        f"{pages_per_seq=} must be greater or equal to"
+        f" {min_pages_per_seq=} given {max_kv_len=} and {page_size=}.")
+  if cu_q_lens[num_seqs[0]] > max_num_batched_tokens:
+    raise ValueError(
+        f"Total q tokens {cu_q_lens[num_seqs[0]]} must be less or equal to"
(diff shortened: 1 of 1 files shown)


bythew3i:
@yarongmu-google @vanbasten23 @yaochengji

bythew3i:
@vanbasten23 The goal of this check is to make sure "inputs are valid". I do not think we need to wait 50th iteration to figure out the inputs are invalid. The check will happen at very beginning. Once we wait long enough, this means the input should be valid.

yaochengji (approved):
LGTM, thanks Jevin!

vanbasten23:
> @vanbasten23 The goal of this check is to make sure "inputs are valid". I do not think we need to wait 50th iteration to figure out the inputs are invalid. The check will happen at very beginning. Once we wait long enough, this means the input should be valid. Otherwise, it will throw the error very quickly.

Yeah, I agree that the checkis to make sure the inputs are valid. But there is no guarantee that the issue happens at the very beginning. Taking the correctness issue in the kernel v1 as an example, it's at the ~20th example that we first observed the gibberish and I waited for ~20 schedule iterations to get a repro which was about ~30 min. Because in each scheduled iteration, the kernel input may be different. That's why I think that the invalid input may not happen at the very beginning and thus checking in the nonkernel will make it hard for us in case of a real issue.

bythew3i:
@vanbasten23 

I see... I understand what you mean now - the 20th refers to scheduled iteration not the seq idx. Yeah that makes sense. Then I think we should seriously consider how to add the runtime in debug mode in vLLM. But for this PR, I think it still has its value to add because the runtime check added for non-kernel does affect performance much but can help validate inputs for non-kernel. I think we should consider still check in this even we do not rely on this for inputs validation for kernel.

torch_xla/experimental/custom_kernel.py
```diff
@@ -862,27 +862,61 @@ def flash_attention(
                               sm_scale, ab, partition_spec, mesh)
 
 
+# This function should only be called and excuted on runtime.
+def _ragged_paged_attention_runtime_check(
+    q,  # [max_num_batched_tokens, num_q_heads, head_dim]
```
yaochengji, line 867: Could you directly use `max_num_batched_tokens` and `page_size` as the inputs of this check function, instead of using q and kv_pages? Then we can use it on the CPU side before the model execution.

bythew3i: What do you mean by CPU side? non-kernel or vllm side?

yaochengji: vllm side.

bythew3i: We can add a check function on vllm side directly and optionally gated by debug mode. This PR seems orthogonal?  I don't think we should reuse the helper function in custom_kernel.py.

yaochengji: I think we can. After you changing the arguments, it can be used for both cpu tensors and xla tensors.

bythew3i: Not about can or can't. I don't think we **should**... just add the function to vllm side, why would we call it all the way from torch_xla.

yaochengji: Thanks, sounds fair to me.

PR #8962, opened 2025-04-11
@assume_pure

### LINKED ISSUE #8805: Use jax autograd from PyTorch
## 🚀 Feature

Instead of using PyTorch autograd and checkpointing, we'll investigate using `jax.grad`, `jax.vjp`, `jax.remat` etc. to control the rematerialization of a PyTorch model.

## Motivation

JAX remat is more powerful than PyTorch autograd. For example, we can name individual tensors and selectively save/offload them. PyTorch does not support naming a tensor.

## Pitch

Something like https://github.com/tengyifei/playground/blob/master/torch-jax-autograd.ipynb combined with https://github.com/pytorch/xla/pull/8781 and torchax.

cc @qihqi

tengyifei:
Fixes #8805.

We introduce a decorator, `@assume_pure`, that can be placed on PyTorch/XLA functions and easily eliminate lazy tensor tracing overhead. If you have a pure function that only uses torch upstream ops, that function can be decorated with `@assume_pure` and will only be traced once for each unique input tensor shape combinations.

## Design

`@assume_pure` brings together three pieces of existing technologies:

- `jax.vjp`, which takes a JAX function and gives you the autograd forward and backward pass
- `torchax`, which converts a pure PyTorch function to a JAX function
- `xb.call_jax`, which can call any JAX function from PyTorch/XLA and integrate it into the HLO graph

It works by:

- Use `torchax.interop.jax_view` to obtain a JAX function from the input PyTorch function
- Use `jax.vjp` to get the forward and backward pass
- Return a `torch.autograd.Function` instance, where the forward implementation is `xb.call_jax(forward_pass)`, and the backward implementation is `xb.call_jax(backward_pass)`, respectively.

The core logic is actually just a single line:

```python
def assume_pure(fn):
  from torchax.interop import jax_view
  return j2t_autograd(jax_view(fn))
```

### How is the HLO cached

`xb.call_jax` caches the HLO if all the input shapes/dtypes and non-tensor arguments are the same.

Therefore, subsequent `xb.call_jax` will just reuse the cached HLO instead of retracing.

The same kind of caching happens in both the forward and backward pass.

Different from the jax wrapper we used in `splash_attention`, the `j2t_autograd` function saves the residuals (intermediate activations) during the forward pass and reuses them during the backward by plugging those into the `vjp_fun` again. This means it won't force a rematerialization (rerun the fwd) during the backward.

## Alternatives

Instead of `jax.vjp` we could also use AOTAutograd to get the forward and backward pass. However, AOTAutograd has a number of downsides:

- It does more than just getting the backward. It also forcefully decomposes all operations into the "aten" op set. Decomposing operations will negatively impact performance, especially in the case of einsum.
- There is no straightforward path to support profiler trace spans. In contrast, in the proposed approach we could translate `xp.Trace(...)` to `jax.named_scope(...)`.
- Supporting custom operations such as pallas kernels will be cumbersome. We'll need to wrap every kernel into a PyTorch custom operator in order for AOTAutograd to not crash on those functions. In contrast, in the proposed approach we could augment our pallas kernels to directly jump into JAX when the input tensor is a torchax tensor.

Instead of `assume_pure`, we could also use `torch.compile` to cache the XLA executable of the compiled function and skip the lazy tensor tracing. However, `torch.compile` has its own downsides:

- `torch.compile` itself uses AOTAutograd and will suffer from the decomposition and customer operations issues etc.
- `torch.compile` has a general perception of "either it works, or debugging will be complicated", which has been corroborated by experiments by people in the PyTorch/XLA team. See PyTorch team members' own recommendation [1]. In contrast, `@assume_pure` has very simple rules for determining if it will work: if your function is pure, then it works.
- `torch.compile` will graph break when entering and leaving the compiled region. In contrast, `@assume_pure` can avoid tracing overhead without even breaking the graph. The cached HLO is inlined into the overall HLO.

## Benchmarks

I tested tracing an example 100 layer decoder-only model:

```sh
~/pytorch/xla
❯ TESTBRIDGE_TEST_ONLY=test_trace_transformer_with_spda_attention python3 test/test_assume_pure.py --benchmark_iterations 100
[...]
No `@assume_pure` time: 140.1342 ms
`@assume_pure` time: 24.1658 ms
```

Importantly, the `@assume_pure` does not scale with increasing complexity inside the model. That's because we only trace the model once, paying a fixed up-front cost, and then later runs will reuse the cached XLA computation object.

Anecdotally, @bhavya01 reported saving >200ms tracing time in an SDXL experiment. That's very significant since each training step is sub-1 second.

[1]: https://docs.google.com/document/d/1y5CRfMLdwEoF1nTk9q8qEu1mgMUuUtvhklPKJ2emLU8/edit?tab=t.0

### DIFF
--- a/docs/source/perf/assume_pure.md
+++ b/docs/source/perf/assume_pure.md
@@ -1,36 +1,36 @@
 # Use `@assume_pure` to speed up lazy tensor tracing
 
-This document guides how to use `torch_xla.experimental.assume_pure` to eliminate
-lazy tensor tracing overhead. See [this blog post][lazy-tensor] for a primer on
-how lazy tensor tracing (operation recording) works.
+This document explains how to use `torch_xla.experimental.assume_pure` to
+eliminate lazy tensor tracing overhead. See [this blog post][lazy-tensor] for a
+primer on how lazy tensor tracing (operation recording) works.
 
 ## Background and motivation
 
 PyTorch/XLA's lazy tensor tracing ensures correct execution by recording an
-operation graph (lazy tensor IR) when running PyTorch operations. However, for
-complex models, this tracing overhead can exceed the execution time of the
-graph, leading to performance bottlenecks. When training a model, the layers in
-the model must be re-traced on every training step. That's because in the limit,
-there's no guarantee that the layers will do the same thing in different
-training steps. As an extreme example, a layer's `forward()` function may call
-`math.random()` and decide what code to run based on the roll of a dice.
+operation graph (lazy tensor IR) when running PyTorch operations. For complex
+models, this tracing overhead can exceed the execution time of the graph,
+leading to performance bottlenecks. When training a model, the layers in the
+model must be re-traced on every training step. That's because there's no
+guarantee that the layers will do the same thing in different training steps. As
+an extreme example, a layer's `forward()` function may call `math.random()` and
+decide what code to run based on a pseudo random number.
 
-However, this re-tracing errs on the conservative side too much. In many cases,
-the layers in your model do exactly the same thing when given the same input
-tensor shapes, and they will return the same output if given the same input.
+Re-tracing can introduce unnecessary overhead. In many cases, the layers in your
+model will do exactly the same thing when given the same input tensor shapes. In
+other words, given the same input, the function return the same output. Often,
+the layers also will not will not perform side-effects such as saving the tensor
+to a file or adding it to a global list. Such functions are called
+"[pure functions][pure-function]".
 
-Taking advantage of this, `@assume_pure` can be placed on PyTorch/XLA functions
-and easily eliminate lazy tensor tracing overhead. If you have a
-[pure function][pure-function] that operates on XLA tensors, that function can be
-decorated with `@assume_pure` and will only be traced once for each unique input
-tensor shape and dtype combinations. PyTorch/XLA will cache the traced
-computation instead of repeatedly tracing the same operations.
+Any PyTorch/XLA function decorated with `@assume_pure` will only be traced once
+for each unique input tensor shape and dtype combination. PyTorch/XLA will cache
+the traced computation instead of repeatedly tracing the same operations.
 
 ## How to use `@assume_pure`
 
 ### Using `@assume_pure` with a function
 
-If you know your function is pure, simply decorate it with `@assume_pure`:
+If you know your function is pure, decorate your function with `@assume_pure`:
 
 ```py
 import torch
@@ -113,7 +113,7 @@ No `@assume_pure` time: 140.1342 ms
 `@assume_pure` time: 24.1658 ms
 ```
 
-We can see that the version with `@assume_pure` is much faster.
+The version with `@assume_pure` is much faster.
 
 Importantly, the `@assume_pure` running time does not scale with increasing
 complexity inside the model. That's because we only trace the model once, paying
--- a/docs/source/perf/assume_pure.md
+++ b/docs/source/perf/assume_pure.md
@@ -1,0 +1,131 @@
+# Use `@assume_pure` to speed up lazy tensor tracing
+
+This document guides how to use `torch_xla.experimental.assume_pure` to eliminate
+lazy tensor tracing overhead. See [this blog post][lazy-tensor] for a primer on
+how lazy tensor tracing (operation recording) works.
+
+## Background and motivation
+
+PyTorch/XLA's lazy tensor tracing ensures correct execution by recording an
+operation graph (lazy tensor IR) when running PyTorch operations. However, for
+complex models, this tracing overhead can exceed the execution time of the
+graph, leading to performance bottlenecks. When training a model, the layers in
+the model must be re-traced on every training step. That's because in the limit,
+there's no guarantee that the layers will do the same thing in different
+training steps. As an extreme example, a layer's `forward()` function may call
+`math.random()` and decide what code to run based on the roll of a dice.
+
+However, this re-tracing errs on the conservative side too much. In many cases,
+the layers in your model do exactly the same thing when given the same input
+tensor shapes, and they will return the same output if given the same input.
+
+Taking advantage of this, `@assume_pure` can be placed on PyTorch/XLA functions
+and easily eliminate lazy tensor tracing overhead. If you have a
+[pure function][pure-function] that operates on XLA tensors, that function can be
+decorated with `@assume_pure` and will only be traced once for each unique input
+tensor shape and dtype combinations. PyTorch/XLA will cache the traced
+computation instead of repeatedly tracing the same operations.
+
+## How to use `@assume_pure`
+
+### Using `@assume_pure` with a function
+
+If you know your function is pure, simply decorate it with `@assume_pure`:
+
+```py
+import torch
+import torch_xla
+from torch_xla.experimental.assume_pure import assume_pure
+
+@assume_pure
+def do_some_math(
+    # You can pass any number of XLA tensors.
+    a: torch.Tensor,
+    b: torch.Tensor,
+
+    # Non-tensor arguments are also supported, and passing different values will
+    # trigger re-tracing and caching more computations.
+    c: int,
+):
+    # Evaluate some pure expressions.
+    return a @ b + c
+
+# Simulate a training loop.
+# Even if we run this function ten times, it will only be traced once.
+for i in range(10):
+    v = do_some_math(
+        torch.tensor([1.0], device='xla'),
+        torch.tensor([2.0], device='xla'),
+        c=42,
+    )
+    print(v)
+```
+
+### Using `@assume_pure` with a `nn.Module`
+
+If you have a pure `nn.Module` i.e. its `forward` behavior only depends on the
+input arguments and the model parameters, we can use `torch.func.functional_call`
+to convert the module into a pure function and pass that to `assume_pure`:
+
+```python
+import torch
+import torch.nn as nn
+from torch.func import functional_call
+from torch_xla.experimental.assume_pure import assume_pure
+
+class MyModule(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.linear = nn.Linear(10, 10)
+    def forward(self, x):
+        return self.linear(x)
+
+# Create module and move to XLA device
+module = MyModule()
+module = module.to('xla')
+
+# Convert module's forward pass into a pure function
+pure_forward = lambda params, buffers, x: functional_call(module, (params, buffers), (x,))
+
+# Wrap the pure function with @assume_pure
+cached_forward = assume_pure(pure_forward)
+
+# Simulate a training loop
+# Even if we run the model ten times, its forward function will only be traced once.
+params = dict(module.named_parameters())
+buffers = dict(module.named_buffers())
+for i in range(10):
+    x = torch.randn(5, 10, device='xla')
+    y = cached_forward(params, buffers, x)
+    print(y)
+```
+
+## Benchmarks
+
+The unit tests contain a benchmark that traces an example 100 layer decoder-only
+language model:
+
+```sh
+~/pytorch/xla
+❯ TESTBRIDGE_TEST_ONLY=test_trace_transformer_with_spda_attention python3 test/test_assume_pure.py --benchmark_iterations 100
+[...]
+No `@assume_pure` time: 140.1342 ms
+`@assume_pure` time: 24.1658 ms
+```
+
+We can see that the version with `@assume_pure` is much faster.
+
+Importantly, the `@assume_pure` running time does not scale with increasing
+complexity inside the model. That's because we only trace the model once, paying
+a fixed up-front cost, and then later runs will reuse the cached XLA computation.
+
+## Limitations
+
+Currently, all operations in a function wrapped with `@assume_pure` must be
+PyTorch upstream operations (e.g. `torch.einsum`, `torch.sin`, ...). More
+PyTorch/XLA operations (e.g. `mark_sharding`) will be supported in the future.
+
+<!-- xrefs -->
+
+[lazy-tensor]: https://pytorch.org/blog/understanding-lazytensor-system-performance-with-pytorch-xla-on-cloud-tpu/
+[pure-function]: https://en.wikipedia.org/wiki/Pure_function
--- a/test/test_assume_pure.py
+++ b/test/test_assume_pure.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from absl.testing import absltest
 
 import torch
@@ -60,6 +61,57 @@ class TestJaxInterop(absltest.TestCase):
 
     torch.testing.assert_close(
         o, model(torch.ones(3, 3).to('xla')), check_device=False)
+
+  def test_assume_pure_complex_module(self):
+    # Define a module comprising of some linear, conv, and relu layers
+    class MyModule(nn.Module):
+
+      def __init__(self):
+        super().__init__()
+        self.linear = nn.Linear(3, 3)
+        self.conv = nn.Conv2d(3, 3, kernel_size=(3, 3), stride=(1, 1))
+        self.relu = nn.ReLU()
+        self.flatten = nn.Flatten()
+        self.fc = nn.Linear(3, 3)
+
+      def forward(self, x):
+        x = self.linear(x)
+        x = self.conv(x)
+        x = self.relu(x)
+        x = self.flatten(x)
+        x = self.fc(x)
+        return x
+
+    # Prepare inputs
+    orig_model = MyModule().to('xla')
+    orig_params = dict(orig_model.named_parameters())
+    pure_model = deepcopy(orig_model)
+    pure_params = dict(pure_model.named_parameters())
+    orig_x = torch.ones((5, 3, 3, 3), device='xla',
+                        requires_grad=True).to('xla')
+    pure_x = orig_x.clone().detach().requires_grad_(True)
+    torch_xla.sync()
+
+    # Call module in a pure way.
+    def pure_call(params, x):
+      return torch.func.functional_call(pure_model, params, x)
+
+    orig_output = orig_model(orig_x)
+    pure_output = assume_pure(pure_call)(pure_params, pure_x)
+    torch_xla.sync()
+
+    # Check that the outputs are close
+    torch.testing.assert_close(orig_output, pure_output, check_device=False)
+
+    # Check that the gradients are close
+    orig_output.sum().backward()
+    pure_output.sum().backward()
+    torch_xla.sync()
+    assert_gradients_close(self, orig_x, pure_x)
+    for name, _ in orig_model.named_parameters():
+      orig_param = orig_params[name]
+      pure_param = pure_params[name]
+      assert_gradients_close(self, orig_param, pure_param)
 
   def test_assume_pure_avoid_retracing_avoid_rejit(self):
     starting_lowerings = xb._jax_to_hlo_cache_num_misses()
--- a/test/test_assume_pure.py
+++ b/test/test_assume_pure.py
@@ -49,7 +49,7 @@ class TestAssumePure(absltest.TestCase):
     # Assert
     expected = torch.sin(torch.ones(3, 3) @ torch.ones(3, 3))
     torch.testing.assert_close(actual, expected, check_device=False)
-    
+
   @unittest.skipUnless(xr.global_runtime_device_count() >= 2,
                        "Multiple devices required")
   def test_assume_pure_other_xla_devices(self):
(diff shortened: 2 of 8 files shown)


test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+import torch
+import torch.nn as nn
+import torch_xla
+import torch_xla.core.xla_builder as xb
+from torch_xla.experimental.assume_pure import assume_pure
+from torch_xla._internal.jax_workarounds import jax_import_guard
+
+
+# Helper function to compare gradients, handling None cases
+def assert_gradients_close(test_case, tensor1, tensor2):
```
yaoshiang, line 15: Would this function possibly return True even if the grads don't match, but the loss.backward() function was called without (retain_graph=False)?

test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+  def setUp(self):
+    self.device = torch_xla.device()
+
+  def test_assume_pure_basic(self):
+
+    @assume_pure
+    def simple_torch_function(a, b):
+      return torch.sin(a @ b)
+
+    a = torch.ones((3, 3), device='xla', requires_grad=True)
```
yaoshiang, line 43: would it be possible to reduce the mixing and matching between self.device and "xla"? The latter would be preferred I think. If we want a handle on a device, `torch.device("xla")` feels more canonical - LMK what you think.

test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+    o = simple_torch_function(a, a)
+    o.sum().backward()
+
+    torch_xla.sync()
+    torch.testing.assert_close(
```
yaoshiang, line 48: my fear about assert_close() possibly returning True simply because the grads were thrown away because loss.backwards was not called by "retain_graph" gets triggered here. Even if this is correct, I'd recommend we be explicit and remove the None case from assert_close.

test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+        o, torch.sin(torch.ones(3, 3) @ torch.ones(3, 3)), check_device=False)
+
+  def test_assume_pure_module(self):
+    model = nn.Linear(3, 3).to('xla')
```
yaoshiang, line 52: It can be helpful to include # Arrange # Act and # Assert on all unit tests so help the reader understand exactly what's where.

test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+    def simple_torch_function(a, b):
+      return torch.sin(a @ b)
+
+    a = torch.ones((3, 3), device='xla', requires_grad=True)
+    o = simple_torch_function(a, a)
+    o.sum().backward()
+
+    torch_xla.sync()
+    torch.testing.assert_close(
+        o, torch.sin(torch.ones(3, 3) @ torch.ones(3, 3)), check_device=False)
```
yaoshiang, line 49: this line mixes the assert and the arrange - the o and the torch.sin(...) are not parallel.

A good name here would be change `o` to actual and calculate an `expected = torch.sin...`. then the assert_close can be actual and expected. The order matters because of rtol. those names are idiomatic to the function: 

`torch.testing.assert_close(actual, expected, ` - https://pytorch.org/docs/stable/testing.html

test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+    # Call module in a pure way.
+    def pure_call(params, x):
+      return torch.func.functional_call(pure_model, params, x)
+
+    orig_output = orig_model(orig_x)
+    pure_output = assume_pure(pure_call)(pure_params, pure_x)
+    torch_xla.sync()
+
+    # Check that the outputs are close
+    torch.testing.assert_close(orig_output, pure_output, check_device=False)
```
yaoshiang, line 110: pedantically unit tests should only test one thing, but I think it's okay to flex that rule when you can easily test two or three things about a more complicated setup, like this one.

test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+    orig_x = torch.randn((5, 3, 9, 9), device='xla', requires_grad=True)
+    pure_x = orig_x.clone().detach().requires_grad_(True)
+    torch_xla.sync()
+
+    # Call module in a pure way.
+    def pure_call(params, x):
+      return torch.func.functional_call(pure_model, params, x)
+
+    orig_output = orig_model(orig_x)
+    pure_output = assume_pure(pure_call)(pure_params, pure_x)
```
yaoshiang, line 106: Adding a #Act would be very helpful here

test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+    torch_xla.sync()
```
yaoshiang, line 107: Is a sync necessary? I assumed the assert accessing the values would be a barrier.

torch_xla/core/xla_builder.py
```diff
@@ -872,15 +872,13 @@ def jax_func_to_xla_computation(jax_func, args, kwargs, name=None):
     import torchax.ops.mappings as mappings
 
     flattened_inputs, spec = jax.tree.flatten((args, kwargs))
+    xla_device = torch_xla.device()  # this is expensive so do it once
```
yaoshiang, line 875: Is there no caching on this call? Does the expensiveness mean that torch.tensor(1.0, device="xla") is an uncached, expensive call? 

https://github.com/pytorch/xla/blob/458305192e60d7210e2a861e43c47431c6c0e15d/torch_xla/runtime.py#L126

torch_xla/core/xla_builder.py
```diff
@@ -872,15 +872,13 @@ def jax_func_to_xla_computation(jax_func, args, kwargs, name=None):
...
 
     def abstractify(a):  # make a pytree leaf abstract
-      import jax
-      import torch_xla
       if a is None:
         return None
       if isinstance(a, torch.Tensor):
-        assert a.device == torch_xla.device(
-        ), f"Inputs must be XLA tensors. Got {a.device}"
+        assert a.device == xla_device, f"Inputs must be XLA tensors. Got {a.device}"
```
yaoshiang, line 881: what if the nester was `torch.tensor(1.0, device="xla:2")` but the xla_device was created as `torch_xla.device()`, meaning, default number?

torch_xla/experimental/assume_pure.py
```diff
@@ -0,0 +1,121 @@
+import torch
```
yaoshiang, line 1: What is the definition of experimental vs non-experimental?

yaoshiang (changes requested):

torch_xla/experimental/assume_pure.py
tengyifei: That is a great question. I don't think there's been any clear definition. For example, the pallas kernels exist for a year, they are used a ton and you basically always need it for LLMs, but they're all experimental.

My sense is that if an API is likely to change or has some potential gaps, then it should be in experimental. Based on this criteria, I put this in experimental.

I think proposing a clear definition is very welcome.

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
+# Use `@assume_pure` to speed up lazy tensor tracing
+
+This document guides how to use `torch_xla.experimental.assume_pure` to eliminate
```
mikegre-google, line 3: This document explains how to use ....

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
...
+lazy tensor tracing overhead. See [this blog post][lazy-tensor] for a primer on
+how lazy tensor tracing (operation recording) works.
+
+## Background and motivation
+
+PyTorch/XLA's lazy tensor tracing ensures correct execution by recording an
+operation graph (lazy tensor IR) when running PyTorch operations. However, for
```
mikegre-google, line 10: For complex models....

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
...
+complex models, this tracing overhead can exceed the execution time of the
+graph, leading to performance bottlenecks. When training a model, the layers in
+the model must be re-traced on every training step. That's because in the limit,
```
mikegre-google, line 13: That's because there's no guarantee that the layers will do the same thing in different training steps.

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
...
+there's no guarantee that the layers will do the same thing in different
+training steps. As an extreme example, a layer's `forward()` function may call
+`math.random()` and decide what code to run based on the roll of a dice.
```
mikegre-google, line 16: As an extreme example, a layer's `forward()` function may call
`math.random()` and decide what code to run based on a pseudo random number.

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
...
+
+However, this re-tracing errs on the conservative side too much. In many cases,
```
mikegre-google, line 18: Re-tracing can introduce unnecessary overhead. In many cases, the layers in your model will do exactly the same thing when given the same input tensor shapes.  In other words, given the same input, the function return the same output. Such functions are called "pure".

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
...
+the layers in your model do exactly the same thing when given the same input
+tensor shapes, and they will return the same output if given the same input.
+
+Taking advantage of this, `@assume_pure` can be placed on PyTorch/XLA functions
```
mikegre-google, line 22: Replace this paragraph with:

Any PyTorch/XLA function decorated with `@assume_pure` will only be traced once for each unique input tensor shape and dtype combination. PyTorch/XLA will cache the traced computation instead of repeatedly tracing the same operations.

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
...
+[pure function][pure-function] that operates on XLA tensors, that function can be
+decorated with `@assume_pure` and will only be traced once for each unique input
+tensor shape and dtype combinations. PyTorch/XLA will cache the traced
+computation instead of repeatedly tracing the same operations.
+
+## How to use `@assume_pure`
+
+### Using `@assume_pure` with a function
+
+If you know your function is pure, simply decorate it with `@assume_pure`:
```
mikegre-google, line 33: If you know your function is pure, decorate your function with `@assume_pure`:

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
...
+        torch.tensor([1.0], device='xla'),
+        torch.tensor([2.0], device='xla'),
+        c=42,
+    )
+    print(v)
+```
+
+### Using `@assume_pure` with a `nn.Module`
+
+If you have a pure `nn.Module` i.e. its `forward` behavior only depends on the
```
mikegre-google, line 66: From what I can tell, functional_call() calls an nn.Module's `forward` method passing in different parameters than what are stored within the nn.Module. Is this correct? If so the return value from calling functional_call() will be the result of the computation defined in 'forward'. You would then be passing that result to assume_pure(). How does assume_pure() effect the re-tracing behavior if the calculation has already occured?

docs/source/perf/assume_pure.md
```diff
@@ -0,0 +1,131 @@
...
+
+```sh
+~/pytorch/xla
+❯ TESTBRIDGE_TEST_ONLY=test_trace_transformer_with_spda_attention python3 test/test_assume_pure.py --benchmark_iterations 100
+[...]
+No `@assume_pure` time: 140.1342 ms
+`@assume_pure` time: 24.1658 ms
+```
+
+We can see that the version with `@assume_pure` is much faster.
```
mikegre-google, line 116: The version with `@assume_pure` is much faster.

test/test_assume_pure.py
```diff
@@ -0,0 +1,387 @@
...
+    assert_gradients_close(self, orig_x, pure_x)
+    for name, _ in orig_model.named_parameters():
+      orig_param = orig_params[name]
+      pure_param = pure_params[name]
+      assert_gradients_close(self, orig_param, pure_param)
+
+  def test_assume_pure_avoid_retracing_avoid_rejit(self):
+    """Tests that we avoid retracing and re-jitting when using assume_pure."""
+
```
zpcore, line 124: Suggest clear jit cache at the beginning to avoid new tests impact correctness accidentally.

torch_xla/experimental/assume_pure.py
```diff
@@ -0,0 +1,121 @@
...
+    class JaxFun(torch.autograd.Function):
+
+      @staticmethod
+      def forward(ctx, tree_def, *flat_inputs):
+        # Reconstruct the original args and kwargs
+        args, kwargs = tree_unflatten(tree_def, flat_inputs)
+
+        # Execute the JAX computation
+        # Pass the reconstructed args/kwargs tuple as the primal
+        y, fun_vjp = xb.call_jax(
```
zpcore, line 54: We are still tracing within the forward function, but we don't trace within `xb.call_jax` because it returns the caching directly for the second run. This is the magic to save the tracing time. Did I understand correctly?

tengyifei: `xb.call_jax` caches the HLO if all the input shapes/dtypes and non-tensor arguments are the same.

Therefore, the second `xb.call_jax` will just reuse the cached HLO instead of retracing.

The same kind of caching happens in both the forward and backward pass.

Different from the jax wrapper you used in `splash_attention`, the `j2t_autograd` function saves the residuals (intermediate activations) during the forward pass and reuses them during the backward by plugging those into the vjp_fun again. This means it won't force a rematerialization (rerun the fwd) during the backward.

Hope my explanation makes sense! LMK any Q

torch_xla/experimental/assume_pure.py
```diff
@@ -0,0 +1,121 @@
...
+                                         flat_input_grads):
+          final_grads.append(grad if needs_grad else None)
+
+        return tuple(final_grads)
+
+    sig = signature(fn)
+    bound = sig.bind(*args, **kwargs)
+    bound.apply_defaults()
+    flat_args_kwargs, tree_def = tree_flatten((bound.args, bound.kwargs))
+    return JaxFun.apply(tree_def, *flat_args_kwargs)
```
zpcore, line 96: Can we pass in the pytree as input directly to JaxFun so that we don't need to tree_flatten here and tree_unflatten again in the forward? I think autograd can automatically detect the leaf node requires grad.

zpcore:
This is GREAT! Looking foward to this feature!

yaoshiang:
Another issue to think about: the naming `@assume_pure`...

Some decorators change the function's behavior, like, @staticmethod. 

Others change the behavior of the function to the caller, like torch's `@torch.no_grad`. 

`@assume_pure` is closer to `@no_grad`... the function behavior itself isn't changing, but it's a directive to *some* callers to treat this function differently. `no_grad` is great because we immediately know the caller: the autograd system. 

`@assume_pure` doesn't really indicate which caller is being directed. Is there a name that includes the specific caller. Maybe something like `@torch_xla.**.compile(pure=True)`? Then it's really explicit who this decorator is for. 

It may be too hard to think through all the ways we expose lazy tensor and harmonize them in the timeframe of this PR, so consider this optional, and we may have to refactor all the ways we discuss lazy tensor and compilation at some point in the future. (e.g. mark step, xm.optimizer_step, torch_xla.compile, etc).

torch_xla/experimental/assume_pure.py
zpcore: Thanks, that's a great explanation. Maybe we can also mention this in the PR description.

torch_xla/core/xla_builder.py
tengyifei: Each `torch_xla.device()` call takes a few microseconds from my benchmarks. However, the `abstractify` is called in a hot loop for each weight/bias in the model. So this can still add up for large models. In the 100-layer llama model benchmark, the overhead was significant enough to be a few ms.

torch_xla/core/xla_builder.py
tengyifei: Eagle eyes! Updated to allow all kinds of XLA devices.

test/test_assume_pure.py
tengyifei: Yes. That makes a lot of sense. Sorry -- I generated a lot of the tests with Gemini, and it got stuck in the old patterns :)

test/test_assume_pure.py
tengyifei: Done

test/test_assume_pure.py
tengyifei: Nice suggestions.

test/test_assume_pure.py
tengyifei: It's not strictly necessary but can be beneficial for performance and seems like an existing convention to me:
- Without `torch_xla.sync()`, only the tensor you're accessing will be materialized. All other tensors still hold abstract IRs and materializing those would rerun a bunch of computations.
- With `torch_xla.sync()`, all in-scope tensors are materialized.

So if we're accessing 10 tensors in sequence, `torch_xla.sync()` turns a `O(N^2)` compute to `O(N)`.

test/test_assume_pure.py
tengyifei: Ack

test/test_assume_pure.py
tengyifei: Good idea.

torch_xla/experimental/assume_pure.py
tengyifei: Unfortunately the PyTorch autograd Function only supports a flat list of tensor inputs.

zpcore: I see, thanks!

bhavya01:
Thanks for the great work! Just confirming that when using `assume_pure` with gradient checkpointing, it will still store the residuals for the forward function wrapped by `assume_pure`, right? Otherwise, LGTM

tengyifei:
> when using assume_pure with gradient checkpointing, it will still store the residuals for the forward function wrapped by assume_pure

Depends on how those two are combined.

- `torch_xla.utils.checkpoint(assume_pure(fn))`: then the PyTorch autograd will discard the residuals and recompute them again during the backward pass. It just works like we wanted checkpointing to work.
- `assume_pure(torch_xla.utils.checkpoint(fn))`: this case doesn't work as `torch_xla.utils.checkpoint` will attempt to add an optimization barrier onto torchax tensors. But supporting this falls in the same bucket as supporting `xs.mark_sharding` etc: we just need to capture that function in torchax and handle it accordingly.

bhavya01 (approved):

docs/source/perf/assume_pure.md
tengyifei: The calculation is delayed until the function wrapped by `assume_pure` is called:

```
# First step we define a function (no calculation here)
pure_forward = lambda params, buffers, x: functional_call(module, (params, buffers), (x,))

# Then we wrap the function (still no calculation here)
cached_forward = assume_pure(pure_forward)

# Finally, the cached_forward is called, which causes `assume_pure` to trace `pure_forward` once,
# which runs `functional_call` once.
y = cached_forward(params, buffers, x)
```

tengyifei: LMK how I may clarify this point!

tengyifei:
> It may be too hard to think through all the ways we expose lazy tensor and harmonize them in the timeframe of this PR, so consider this optional, and we may have to refactor all the ways we discuss lazy tensor and compilation at some point in the future. (e.g. mark step, xm.optimizer_step, torch_xla.compile, etc).

Ack. I don't have a great immediate thought (maybe `@pure_function_avoid_retrace`? a mouthful..), but I have circulated an internal proposal about using torchax and JAX etc, that has a section with naming ideas.

tengyifei:
This is ready for another look

torch_xla/experimental/assume_pure.py
```diff
@@ -0,0 +1,121 @@
...
+    `torch.einsum`, `torch.nn.functional.layer_norm`. Custom PyTorch/XLA
+    operations such as `mark_sharding` are not supported. This limitation
+    may be lifted in the future.
+  """
+  from torchax.interop import jax_view
+  return j2t_autograd(jax_view(fn))
+
+
+@requires_jax
+def j2t_autograd(fn):
```
qihqi, line 32: actually I think this function makes sense to be inside of `torchax`.

This file can import it and call it there.

Torchax cannot import from this file because `@requires_jax` would disable the jax device?

zpcore (approved):

torch_xla/experimental/assume_pure.py
tengyifei: I'll move

tengyifei: We could leave the `@requires_jax` in torch_xla land to keep a uni-directional dependency graph

docs/source/perf/assume_pure.md
mikegre-google: Okay, make sense. Thanks

mikegre-google (approved):

yaoshiang (approved):
tests appear to confirm the expected behavior of the decorator.

PR #8965, opened 2025-04-11
Add an option for JittableModule to dedup parameters.

qihqi:
Add an option for JittableModule to dedup parameters.

This is needed when we want to use the `donate_argnums` feature of jax_jit. Because, if duplicated parameters are passed into the callable with donated argnums, then, it will show an argument is donated twice and error out.

### DIFF
--- a/torchax/test/test_interop.py
+++ b/torchax/test/test_interop.py
@@ -5,27 +5,6 @@ import torchax
 from torchax import interop
 import torchax
 
-class M1(torch.nn.Module):
-
-    def __init__(self):
-        super().__init__()
-        self.x = torch.ones(10, 10)
-
-class M(torch.nn.Module):
-
-    def __init__(self):
-        super().__init__()
-        self.a = torch.nn.Linear(100, 100)
-        self.b = torch.nn.Parameter(
-            torch.ones(10, 10)
-        )
-        c = torch.ones(10, 10)
-        self.register_buffer('c', c)
-        self.register_buffer('c2', c, persistent=False)
-        self.d = torch.ones(10, 10)
-        self.m1 = M1()
-
-
 
 class InteropTest(unittest.TestCase):
 
@@ -34,7 +13,28 @@ class InteropTest(unittest.TestCase):
 
 
     def test_mod_attr(self):
-        m = M()
+
+        class Child(torch.nn.Module):
+
+            def __init__(self):
+                super().__init__()
+                self.x = torch.ones(10, 10)
+
+        class ModuleWithUnregisteredTensor(torch.nn.Module):
+
+            def __init__(self):
+                super().__init__()
+                self.a = torch.nn.Linear(100, 100)
+                self.b = torch.nn.Parameter(
+                    torch.ones(10, 10)
+                )
+                c = torch.ones(10, 10)
+                self.register_buffer('c', c)
+                self.register_buffer('c2', c, persistent=False)
+                self.d = torch.ones(10, 10)
+                self.m1 = Child()
+
+        m = ModuleWithUnregisteredTensor()
         params, buffers = interop.extract_all_buffers(m)
         self.assertEqual(
             set(params.keys()), {'a.weight', 'a.bias', 'b'}
@@ -83,7 +83,8 @@ class InteropTest(unittest.TestCase):
 
     def test_module_with_shared_weights(self):
 
-        class M2(torch.nn.Module):
+        # arrange
+        class ModuleWithSharedWeights(torch.nn.Module):
 
             def __init__(self):
                 super().__init__()
@@ -93,23 +94,25 @@ class InteropTest(unittest.TestCase):
             def forward(self, x):
                 return self.a(self.b(x))
 
-        m = M2().to('jax')
+        m = ModuleWithSharedWeights().to('jax')
 
         m_jitted = interop.JittableModule(m, dedup_parameters=True)
-
 
         # a's weights and bias and b's weights and bias
         self.assertEqual(len(m.state_dict()), 4)
         
         # b's weights and bias are deduped
         self.assertEqual(len(m_jitted.params), 2)
-
         x = torch.randn(10, 10).to('jax')
-
         expected = m(x)
-        
-        torch.testing.assert_allclose(m_jitted(x), expected)
 
+        # act 
+        actual = m_jitted(x)
+        
+        # assert
+        torch.testing.assert_allclose(actual, expected)
+
+        # arrange
         # make sure buffer donation works
         functional_forward = interop.jax_jit(
             functools.partial(m_jitted.functional_call, 'forward'),
@@ -118,8 +121,10 @@ class InteropTest(unittest.TestCase):
             }
         )
 
-        torch.testing.assert_allclose(
-            functional_forward(m_jitted.params, m_jitted.buffers, x) , expected)
+        # act
+        actual = functional_forward(m_jitted.params, m_jitted.buffers, x)
+        # assert
+        torch.testing.assert_allclose(actual, expected)
 
 
         
--- a/torchax/test/test_interop.py
+++ b/torchax/test/test_interop.py
@@ -1,7 +1,9 @@
+import functools
 import torch
 import unittest
 import torchax
 from torchax import interop
+import torchax
 
 class M1(torch.nn.Module):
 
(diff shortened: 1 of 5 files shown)


torchax/test/test_interop.py
```diff
@@ -1,6 +1,8 @@
+import functools
 import torch
 import unittest
 from torchax import interop
+import torchax
 
 class M1(torch.nn.Module):
```
yaoshiang, line 7: I realize M1 and M predate the changes in this PR, but it took me a while to understand what they were - a descriptive name would have been helpful. The Google Python Style Guide is pedantic about this, I think it's okay to break that rule for idiomatic abbreviations, such as shape dimensions like BHWC. I'd suggest calling M something like "BuffersParamsAndUnregisteredTensors" since I think that's what it's trying to do. 

Also, if M1 is used solely to nest params inside M, you could consider declaring it as a nested class in M, with a private name like `class _UnregisteredTensorModule`, if that's your intent.

torchax/test/test_interop.py
```diff
@@ -41,6 +47,49 @@ def test_mod_attr(self):
...
+            
+            def forward(self, x):
+                return self.a(self.b(x))
+
+        m = M2().to('jax')
+
+        m_jitted = interop.JittableModule(m, dedup_parameters=True)
+
+
+        # a's weights and bias and b's weights and bias
```
yaoshiang, line 67: helpful comment

torchax/test/test_interop.py
```diff
@@ -41,6 +47,49 @@ def test_mod_attr(self):
...
+        # make sure buffer donation works
+        functional_forward = interop.jax_jit(
+            functools.partial(m_jitted.functional_call, 'forward'),
+            kwargs_for_jax_jit={
+                'donate_argnums': (0, )
+            }
+        )
+
+        torch.testing.assert_allclose(
+            functional_forward(m_jitted.params, m_jitted.buffers, x) , expected)
```
yaoshiang, line 88: Could you make the call to functional_forward and assign the value to "actual", so that the assert is clear about what is actual and expected? the sig for assert_allclose is `assert_allclose(actual, expected)`, so that matches the function directly, and it's clear what is expected and what is actual.

https://pytorch.org/docs/stable/testing.html#module-torch.testing

torchax/test/test_interop.py
```diff
@@ -41,6 +47,49 @@ def test_mod_attr(self):
         self.assertEqual(m.a.weight.item(), 0)
         self.assertEqual(m.m1.x.item(), 0)
 
+    def test_module_with_shared_weights(self):
+
```
yaoshiang, line 51: unit tests that test multiple things can really benefit from the # Arrange # Act # Assert comments, especially when there are multiple arrange blocks and assert blocks.

torchax/test/test_interop.py
```diff
@@ -23,8 +25,12 @@ def __init__(self):
         self.m1 = M1()
 
 
+
 class InteropTest(unittest.TestCase):
 
+    def setUp(self):
+        torchax.enable_globally()
+
```
yaoshiang, line 33: should d and m1 be buffers? I thought self.x = torch.tensor() is NOT a buffer unless you register it using torch.nn.Module.register_buffer(). 
Screenshot 2025-04-11 at 9 07 23 AM

torchax/torchax/interop.py
```diff
@@ -49,14 +50,27 @@ def set_one(module, prefix):
 
 class JittableModule(torch.nn.Module):
 
-    def __init__(self, m: torch.nn.Module, extra_jit_args={}):
+    def __init__(self, m: torch.nn.Module, extra_jit_args={}, dedup_parameters=True):
```
yaoshiang, line 53: Is it the intent that buffers do not need to be deduped?

torchax/torchax/ops/jaten.py
```diff
@@ -1590,12 +1590,14 @@ def _aten_bitwise_not(self):
 
 
 # aten.bitwise_left_shift
+@op(torch.ops.aten.__lshift__)
```
yaoshiang, line 1593: Are these part of this PR? I didn't see a use of lshift / rshift in the rest of the code.

yaoshiang (changes requested):
Cool stuff here, nice to see the donate args in action.

torchax/test/test_interop.py
qihqi: The test is to make sure that even unregistered tensors, as well as non-persistent buffers are captured by `interop.extract_all_buffers()`. Otherwise `module.state_dict()` would have been sufficient.

torchax/torchax/interop.py
qihqi: If you don't use donate_argnums then you don't really has to dedupe from the input arguments. Overall it doesn't hurt.

torchax/torchax/ops/jaten.py
qihqi: technically not. Both this and buffer deduping params are found in process of supporting reco model.

torchax/test/test_interop.py
```diff
@@ -23,8 +25,12 @@ def __init__(self):
         self.m1 = M1()
 
 
+
```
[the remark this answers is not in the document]
qihqi: we had yapf for python codes but traditionally torchax is excluded.

torchax/torchax/ops/jaten.py
yaoshiang: sounds okay - the pure readability reviewer would tell you to create a second PR for that, even if it's trivial :) no worries. this is approved.

torchax/test/test_interop.py
```diff
@@ -1,33 +1,39 @@
...
-class InteropTest(unittest.TestCase):
+        class Child(torch.nn.Module):
 
+            def __init__(self):
+                super().__init__()
+                self.x = torch.ones(10, 10)
 
-    def test_mod_attr(self):
-        m = M()
+        class ModWithUnregisteredTensor(torch.nn.Module):
```
yaoshiang, line 22: Optional nit: Mod is an abbreviation and there's a guideline against that. Defer to you on this one. https://google.github.io/styleguide/pyguide.html#316-naming

torchax/test/test_interop.py
```diff
@@ -41,6 +47,54 @@ def test_mod_attr(self):
         self.assertEqual(m.a.weight.item(), 0)
         self.assertEqual(m.m1.x.item(), 0)
 
+    def test_module_with_shared_weights(self):
+
+        # arrange
+        class Module(torch.nn.Module):
```
yaoshiang, line 53: Using Module could eventually confuse a reader on line 63, not realizing this is a local class `Module`, not a `from torch.nn import Module`. Could you update this to a descriptive name, perhaps, "ReuseWeights" or something like that?

torchax/test/test_interop.py
```diff
@@ -1,33 +1,39 @@
...
+                    torch.ones(10, 10)
+                )
+                c = torch.ones(10, 10)
+                self.register_buffer('c', c)
+                self.register_buffer('c2', c, persistent=False)
+                self.d = torch.ones(10, 10)
+                self.m1 = Child()
+
+        m = ModWithUnregisteredTensor()
         params, buffers = interop.extract_all_buffers(m)
```
yaoshiang, line 37: It still worries me that d and m1 are technically not buffers, but, they are returned by interim.extract_all_buffers. I demonstrated in a previous comment that a subclass of torch.nn.Module does not consider a member of type torch.Tensor a buffer unless it is then called with self.register_buffer. This has implications on both saving the model, and when an instance is called with `.to("cuda")` or `.to("xla")`. If you are sure this is what you want, I'll defer to you on the design.

yaoshiang (approved):
approved, with optional nits and optional comment on design.

torchax/test/test_interop.py
qihqi: More context on why we want this:

1. `m1` is not buffer but `torch.Tensor` attributes of `m1` are. We are not including `m1` itself there.
2. If we don't include `d` then that argument will not be part of input list, and it's value will hardcoded in the stablehlo graph. This is fine from a semantical point of view, but it will OOM the code cache in TPU.

torchax/test/test_interop.py
qihqi: done.

torchax/test/test_interop.py
qihqi: done.

PR #8967, opened 2025-04-12
Add a helper class to handle mesh and sharding

### LINKED ISSUE #8635: [torchax] jit compile the model constructor
## 🚀 Feature

It's worth providing some function to capture the model constructor (i.e., the torch ops that generates the random weights that make up the parameters of a model) as one StableHLO graph, and run that on accelerator devices.

## Motivation

The primary motivation is to more closely match PyTorch eager UX during SPMD training.

Today, in order to initialize a large model on e.g. 256 TPUs, we randomly initialize every layer, and then send that layer to TPUs following a sharding spec:

https://github.com/pytorch/xla/blob/8e6ca6000e83ccbc4365a9d9358e510504b71dea/torchax/examples/train_llama_torchtitan/train_llama.py#L193-L211

This works but has some drawbacks:

- We're initializing the weights with `torch.randn` but eager PyTorch initializes the weights with a variety of different distributions. When I tested training a Llama model with `randn` (Gaussian distributed) weights, the loss at step 0 is 10x larger than what eager PyTorch gives us to start with.
- We could probably workaround this in the near term by having the user specify a dictionary of `module: initializer_fn` mappings. But that's more code over eager PyTorch, and is a cost that users are paying without corresponding gains. In comparison, SPMD sharding annotations is a cost that users pay to get automatic collectives/sharding propagation.

## Pitch

In PyTorch/XLA, I could do this:

```python
with torch_xla.device():
  model = Model()
  xs.mark_sharding(model.some_weight, ...)
torch_xla.sync()
```

The above will compile down to a graph that outputs a bunch of model weights, and the outputs (weights) have sharding annotations in them. When this graph is executed, each TPU will initialize the shard of the model weight that they are responsible for. Note that in this case the only extra cost (besides the `torch_xla.sync()`) is the `mark_sharding` for SPMD sharding annotations, reflecting a principle of "only pay for what you use".

## Alternatives

An alternative is to maintain a comprehensive `module: initializer_fn` mapping inside `torchax` itself. However, that still won't cover cases where the user added custom initialization logic in their model constructor.

## Additional context

We found this while bringing up Llama 3.1 405B training, c.f. https://github.com/AI-Hypercomputer/torchprime/pull/25/files

cc @qihqi

qihqi:
The first functionality is to initialize a model
on sharded devices

fixes #8635

### DIFF
--- a/torchax/test_dist/test_mesh_util.py
+++ b/torchax/test_dist/test_mesh_util.py
@@ -1,12 +1,14 @@
 import unittest
 
 import jax
+from jax.sharding import PartitionSpec
 import torch
 import torchax
-from torchax.mesh_util import Mesh
+from torchax.mesh_util import Mesh, SingleAxisSharder
 
 
 class MeshUtilTest(unittest.TestCase):
+
   def setUp(self):
     torchax.enable_globally()
 
@@ -25,6 +27,25 @@ class MeshUtilTest(unittest.TestCase):
     self.assertEqual(
       len(model.a.bias.jax().addressable_shards), len(jax.devices())
     )
+
+  def test_sharder_call(self):
+    """Test the __call__ method produces the correct PartitionSpec."""
+    sharder = SingleAxisSharder(axis_name="fsdp", axis_size=4)
+    # Use a simple named tuple instead of MagicMock
+    shaped_type = torch.ones((5, 8, 12))  # Middle dim divisible by 4
+
+    spec = sharder("param_name", shaped_type)
+    self.assertEqual(spec, PartitionSpec(None, "fsdp", None))
+
+  def test_sharder_call_no_shardable(self):
+    """Test __call__ when no dimension is shardable."""
+    sharder = SingleAxisSharder(axis_name="fsdp", axis_size=4)
+    shaped_type = torch.ones((5, 7, 11))
+
+    with self.assertRaisesRegex(
+      AssertionError, "Unable to find a dim to shard"
+    ):
+      sharder("param_name", shaped_type)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


torchax/torchax/mesh_util.py
```diff
@@ -0,0 +1,69 @@
+import jax
+from jax.sharding import PartitionSpec, NamedSharding
+import torch
+import torchax
+from torchax import interop
+
+
+def _shard_first_multiple_of(axis_name, shape, multiple_of):
+  sharding = []
+  for size in shape:
```
tengyifei, line 10: As written, this for loop is a no-op

I think you meant

```
found = False
for size in shape:
  if size % multiple_of == 0 and not found:
    sharding.append(axis_name)
    found = True
  else:
    sharding.append(None)
```

or something like that?

It's unclear to me how the unit test passes

torchax/test_dist/test_mesh_util.py
```diff
@@ -0,0 +1,31 @@
+import unittest
```
tengyifei, line 1: Need to add this file into the unit test shell script and activate the CPU SPMD env var, I think

tengyifei (approved):
lgtm but need to address the comments

torchax/torchax/mesh_util.py
qihqi: good catch!

torchax/test_dist/test_mesh_util.py
qihqi: it already taken care of here: https://github.com/pytorch/xla/blob/master/.github/workflows/torchax.yml#L59

PR #8976, opened 2025-04-14
Avoid re-computing computation hashes

rpsilva-aws:
Currently, we are recomputing the hash of the underlying computation for every hash lookup, as a mere logging in two places. For small models where tracing is not negligible, this can have a small impact - particularly since we deserialize the protobuf deterministically (requiring the ordering of unordered dictionary/map entries). C++'s evaluates stream operators eagerly, so the impact is there with or without the added logging. We recently saw ~10% throughput impact for small BERT/Llama models.

Note that this is only used to provide an unique hash string for which a hash key maps to. The actual hash of the protobuf is only meaningful for UserComputation computations, where it is factored in the hash key. In all other cases, it is unnecessary, and the user can still verify the mapping for any given graph hash key when enabling `post_compilation_analysis`.

We see this during hash lookup, which is evaluated every time. We also see it in `Compile`, though it is there only for the very first computation (across all instances). The user can still access the computation proto hash by enabling `PT_XLA_DEBUG`.


tengyifei (approved):

tengyifei:
Looks like you need to update some tests that check for these kind of logs.

torch_xla/csrc/debug_util.cpp
```diff
@@ -431,9 +432,20 @@ void DebugUtil::post_compilation_analysis(
...
   }
 
+  std::stringstream ss;
+  // This can be used to verify the hash of the underlying computation proto.
+  // Note that for UserComputation computations, the protobuf is factored in
+  // the graph hash.
+  std::string serialized_computation =
+      ConsumeValue(runtime::util::GetDeterministicSerializedModuleProto(
+          computation->computation().proto()));
+  ss << "\n"
```
jeffhataws, line 442: How do we associate the compute hash with the graph hash?

jeffhataws (approved):

PR #8978, opened 2025-04-15
Fix splash attention test

zpcore:
1. Fix splash attention test;
2. Reduce paged attention number of test.


zpcore:
I completely get rid of flash attention and use vanilla attention in this case. To test again vanilla attention with segment id, we need to support `caucal = True` for splash attention kernel.

I still can't figure out why flash attention always fails in the backward sharding in the CI. Flash attention and Splash attention seem have a conflict related to in/out shard map.

torch_xla/experimental/splash_attention.py
```diff
@@ -170,8 +171,10 @@ def wrap_flash_attention(query, key, value, decoder_segment_ids):
         k_layout=splash_attention_kernel.QKVLayout[global_k_layout],
         v_layout=splash_attention_kernel.QKVLayout[global_v_layout],
     )
-
-    mask = splash_attention_mask.CausalMask(shape=(seq_len, seq_len))
+    if causal:
+      mask = splash_attention_mask.FullMask(_shape=(seq_len, seq_len))
```
tengyifei, line 175: Are these inverted? I thought we need `CausalMask` for `causal = True`

tengyifei (changes requested):

torch_xla/experimental/splash_attention.py
zpcore: Aha, nice catch, in fact I reversed all causal both in test and source code, which made the test pass. Fixed now!

tengyifei (approved):

test/test_multi_queries_paged_attention_kernel.py
```diff
@@ -109,10 +109,10 @@ def setUp(self):
 
   @parameterized.product(
```
tengyifei, line 110: Are these relevant to fixing Splash Attention? A good practice for PR is usually splitting out changes to different topics. Then you would also be able to directly request a review from Xiongfei

tengyifei (approved):

test/test_multi_queries_paged_attention_kernel.py
zpcore: Not relevant, the test just takes too much time to complete. Let me split this out.

PR #8979, opened 2025-04-15
Update to ubuntu latest

qihqi:
Also reapplied formatting changes becuase of new yapf version. The old version is no longer available in the new OS version.

### DIFF
--- a/.github/workflows/lintercheck.yml
+++ b/.github/workflows/lintercheck.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           python-version: '3.10'
           cache: 'pip'
-      - run: pip install yapf
+      - run: pip install yapf==0.40.2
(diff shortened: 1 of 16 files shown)


tengyifei (commented):
Rubberstamped.

However, upgrading yapf and clang-format broke the formatting tests. Consider pinning yapf and clang-format.

.github/workflows/lintercheck.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Run clang-format
         shell: bash
         env:
-          CLANG_FORMAT: clang-format-11
```
tengyifei, line 37: This causes C++ formatting to change.

tengyifei (approved):

PR #8981, opened 2025-04-16
Fix call_jax hashing

zpcore:
`xb.call_jax` now hashing with `jax._src.config`.

### DIFF
--- a/test/test_jax_interop.py
+++ b/test/test_jax_interop.py
@@ -8,6 +8,9 @@ import torch_xla.core.xla_builder as xb
 
 class TestJaxInterop(absltest.TestCase):
 
+  def setUp(self):
+    xb._JAX_TO_XLA_COMPUTATION_CACHE.clear()
+
   def test_call_jax(self):
     """Test that we can call a JAX function from PyTorch/XLA lazy tensor tracing."""
 
@@ -252,13 +255,12 @@ class TestJaxInterop(absltest.TestCase):
     import jax
     starting_cache_misses = xb._jax_to_xla_computation_cache_elements()
 
-    # Let's trace the same jax function with different static args.
     dev = xm.xla_device()
     a = torch.ones((3, 3), device=dev)
 
-    def f(a, num: float):
+    def f(a, b):
       import jax.numpy as jnp
-      return a + jnp.sin(num)
+      return a + jnp.sin(b)
 
     jax.config.update('jax_default_matmul_precision', "highest")
     xb.call_jax(f, (a, a))
--- a/test/test_jax_interop.py
+++ b/test/test_jax_interop.py
@@ -167,7 +167,7 @@ class TestJaxInterop(absltest.TestCase):
   def test_call_jax_cache_hlo(self):
     """Test that the HLO of a jax function should be cached."""
 
-    starting_cache_misses = xb._jax_to_xla_computation_cache_num_misses()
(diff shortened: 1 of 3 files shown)


test/test_jax_interop.py
```diff
@@ -245,9 +245,30 @@ def f(a, num: float):
     xb.call_jax(f, (a, 2.0))
     xb.call_jax(f, (a, 3.0))
 
-    ending_cache_misses = xb._jax_to_xla_computation_cache_num_misses()
+    ending_cache_misses = xb._jax_to_xla_computation_cache_elements()
     self.assertEqual(ending_cache_misses - starting_cache_misses, 3)
 
+  def test_call_jax_with_different_jax_config(self):
+    import jax
```
tengyifei, line 252: We need to clear the cache to be safe, I think

test/test_jax_interop.py
```diff
@@ -245,9 +245,30 @@ def f(a, num: float):
...
+    starting_cache_misses = xb._jax_to_xla_computation_cache_elements()
+
+    # Let's trace the same jax function with different static args.
```
tengyifei, line 255: nit: this test is not passing a static arg (like number 3). It's passing `a` twice

test/test_jax_interop.py
```diff
@@ -245,9 +245,30 @@ def f(a, num: float):
...
+    dev = xm.xla_device()
+    a = torch.ones((3, 3), device=dev)
+
+    def f(a, num: float):
```
tengyifei, line 259: This won't be `float` because the test ends up passing two copies of `a` into the test function

torch_xla/core/xla_builder.py
```diff
@@ -943,38 +943,36 @@ def _jax_to_xla_computation_cache_get_or_insert(jax_func,
...
-                                           input_tree_spec)] = hlo
-  return hlo
-
-
-def _jax_to_xla_computation_cache_num_misses() -> int:
+  # Use three layers of dictionary lookup.
+  # The first layer uses the `config.trace_context()`, which is strongly referenced.
+  # The second layer uses the `jax_func`, which is weakly referenced.
+  # The third layer uses the sample inputs and the tree spec, which is strongly referenced.
+  # jax_config = JaxConfigWrapper(config.trace_context())
```
tengyifei, line 952: nit: obsolete comment

tengyifei (approved):

tengyifei:
Just a heads up: I submitted `@assume_pure` in https://github.com/pytorch/xla/pull/8962/files#diff-088930d7c2ec653486390434bafbb93fed86eb38c39c14a997f116276976ba9eR154 and you might need to rebase then rename my few usages of the cache_miss function too!

test/test_jax_interop.py
zpcore: Added `setUp` to clear cache for all tests.

test/test_jax_interop.py
zpcore: Oh, good catch!

test/test_jax_interop.py
zpcore: Forgot to remove after copying the test...

tengyifei:
Unfortunately you'd have to rebase again because our CI broke today and just got fixed by https://github.com/pytorch/xla/pull/8979

PR #8982, opened 2025-04-16
Default explicit donation for step barriers

rpsilva-aws:
We currently rely on the user to enable `GetAliasWithBufferDonorConfig`, so that the explicitly marked tensors for donation can be included in the buffer donor indices. However, it is unnecessary to have this constraint during step barrier, where the graph is expected to break, requiring the donation to happen (force_ltc_data) for all live tensors. Instead, we choose to default explicit donation for the step barrier (mark step), and only make it optional, depending on the buffer donor config, for other graph synchronizations (e.g. `_xla_sync_live_tensors` or `_xla_warm_up_cache`), where it requires an opt-in by the user to donate at that point (knowingly that the tensors won't be accessible anymore).

Originally, the explicit donation config was mostly to decide whether to run LTC or explicit donations. Now that these can be used simultaneously, there is no added benefit in requiring the option, at least for the step barrier.

### DIFF
--- a/test/dynamo/test_dynamo_aliasing.py
+++ b/test/dynamo/test_dynamo_aliasing.py
@@ -185,41 +185,6 @@ class TestDynamoBufferDonationAliasing(unittest.TestCase):
     self.assertNotIn('XlaSetBufferDonation', met.counter_names())
 
 
-class TestNonDynamoBufferDonationAliasing(unittest.TestCase):
-
-  def dummy_fn(self, input):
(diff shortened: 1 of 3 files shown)


rpsilva-aws:
Hey @tengyifei, do you see concerns with this? I also aligned with @mcuiaws, who had done the [some of the refactorings](https://github.com/pytorch/xla/pull/8467/files) on the code path.

qihqi (approved):

PR #8984, opened 2025-04-16
scan-based GRU falls back to nn.GRU when `bidirectional` is true.

iwknow:
This change makes scan-based GRU implementation to fall back to the standard pytorch GRU when `bidirectional` is true. scan-based GRU will not support `bidirectional` unless there is enough demand. for details: https://github.com/pytorch/xla/issues/8860


iwknow:
@tengyifei please review

tengyifei (approved):
Thanks!

Our CI got some troubles yesterday. Once you rebase the PR on the latest `master` branch, it should be landable.

PR #8989, opened 2025-04-16
composibility of assume_pure and call_jax

qihqi:


torch_xla/core/xla_builder.py
```diff
@@ -869,16 +869,17 @@ def jax_func_to_xla_computation(jax_func, args, kwargs, name=None):
...
 
     def abstractify(a):  # make a pytree leaf abstract
       if a is None:
         return None
       if isinstance(a, torch.Tensor):
-        assert a.device.type == 'xla', f"Inputs must be XLA tensors. Got {a.device}"
-        return jax.ShapeDtypeStruct(a.shape, mappings.t2j_dtype(a.dtype))
+        assert a.device == torch_xla.device(
+        ), f"Inputs must be XLA tensors. Got {a.device}"
+        return jax.ShapeDtypeStruct(a.shape, tx.ops.mappings.t2j_dtype(a.dtype))
```
tengyifei, line 882: I think this came from an old version of the code. It wouldn't work if the graph was running on `xla:2` etc.

torch_xla/core/xla_builder.py
```diff
@@ -1016,9 +1017,12 @@ def call_jax(jax_func,
   works. If you get tracing overhead, check if `jax_func` is being redefined all the time.
   A common mistake is defining `jax_func` as a local function, e.g. during a training step.
   """
-  import jax
   kwargs = kwargs or {}
-  flattened, _spec = jax.tree.flatten((args, kwargs))
+  flattened, _spec = tree_flatten((args, kwargs))
```
tengyifei, line 1021: I think the `tree_flatten` also came from an old version of the code. It failed a test because the PyTorch interpretation of PyTree is different from JAX wrt None values.

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -584,6 +584,12 @@ def mark_sharding(t: Union[torch.Tensor, XLAShardedTensor], mesh: Mesh,
   assert len(t.shape) == len(partition_spec), \
     f"Partition spec length ({len(partition_spec)}) should be equal to the input rank ({len(t.shape)})."
 
+  tx = maybe_get_torchax()
```
tengyifei, line 587: Can we add a test for this case? I think maybe running a graph with `assume_pure(mark_sharding)` and then testing that the outputs were sharded correctly like the `test_scan_spmd.py` are doing

tengyifei (changes requested):
Nice! Thanks!

torch_xla/core/xla_builder.py
tengyifei: https://github.com/pytorch/xla/blob/cce7cc0782bbf507c96e1d81d705db8ec5099f0a/test/test_assume_pure.py#L55 should catch it

qihqi: done.

torch_xla/distributed/spmd/xla_sharding.py
qihqi: done

torch_xla/core/xla_builder.py
qihqi: done.

tengyifei (approved):

tengyifei (approved):

torch_xla/distributed/spmd/xla_sharding.py
```diff
@@ -584,6 +584,14 @@ def mark_sharding(t: Union[torch.Tensor, XLAShardedTensor], mesh: Mesh,
...
+  if tx is not None and (t, tx.tensor.Tensor):
+    from jax.sharding import PartitionSpec, Mesh as JMesh, NamedSharding
+    device_ids = mesh.device_ids.reshape(mesh.mesh_shape)
+    jax_mesh = JMesh(device_ids, mesh.axis_names)
```
tengyifei, line 591: Hmm, do you think the PyTorch/XLA device IDs will translate to JAX device IDs directly? I think the device IDs in the PyTorch/XLA mesh is PyTorch/XLA specific, like it's just an integer into some PyTorch/XLA global variable.

It should be safer to do what https://github.com/pytorch/xla/blob/cce7cc0782bbf507c96e1d81d705db8ec5099f0a/torch_xla/experimental/splash_attention.py#L67-L90 does

tengyifei (changes requested):

tengyifei (approved):

PR #8999, opened 2025-04-17
Fix the source param sharding for GradAcc API

rpsilva-aws:
This explains the recent nuances with the observed memory footprint model initialization, small fix.


rpsilva-aws:
@tengyifei PTAL, easiest PR :)

PR #9003, opened 2025-04-18
Replace upstream GRU implementation with scan-based GRU

iwknow:
This concludes https://github.com/pytorch/xla/issues/8860

### DIFF
--- a/.github/workflows/lintercheck.yml
+++ b/.github/workflows/lintercheck.yml
@@ -9,16 +9,16 @@ on:
 
 jobs:
   linter_check:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.7'
+          python-version: '3.10'
           cache: 'pip'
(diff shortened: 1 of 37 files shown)


iwknow:
@tengyifei please review

tengyifei (changes requested):
Nice. Hmm, if we do this patch directly, would that nullify the tests in https://github.com/pytorch/xla/blob/329be9e59e49d3b581f777bc17a6c345bfefc7aa/test/test_gru.py#L26 (since it's importing a patched GRU which is the same as ScanGRU)?

I wonder if we could something like:

- Store an original copy of `nn.GRU` in some globals in `_patched_functions`
- Have the test grab that `nn.GRU` copy so that we could actually compare upstream and PyTorch/XLA
- Maybe a simple test that verifies that `nn.GRU` actually equals `ScanGRU` now

Thanks again!

iwknow:
for an unknown reason, the commit (https://github.com/pytorch/xla/commit/ca210958dbf72b401998c9f5d70ac4ea9e5720d3) addresses the issue above cannot be properly recognized by github. see https://github.com/pytorch/xla/compare/master...iwknow:xla:fallback

i will close this PR and open a new one on a new branch

PR #9008, opened 2025-04-18
fix libtpu path

zpcore:
libtpu release wheel naming doesn't follow the trailing `-py3-none-linux_x86_64.whl`. It should be something like `-py3-none-manylinux_2_31_x86_64` and the number `2_31` is subject to change. So I encode the naming into `_libtpu_version`.

Check https://storage.googleapis.com/libtpu-wheels/index.html for detailed naming.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,10 @@ base_dir = os.path.dirname(os.path.abspath(__file__))
 USE_NIGHTLY = True  # whether to use nightly or stable libtpu and jax
 
 _date = '20250406'
-_libtpu_version = '0.0.13'
+# The postfix can be changed when the version is updated. Check
+# https://storage.googleapis.com/libtpu-wheels/index.html for detailed
+# versioning.
+_libtpu_version = '0.0.13-py3-none-manylinux_2_31_x86_64'
 _jax_version = '0.5.4'
 _jaxlib_version = '0.5.4'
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -68,21 +68,24 @@
...
+  _jaxlib_version += f'.dev{_date}'
+  _libtpu_wheel_name = f'libtpu-{_libtpu_version}.dev{_date}+nightly-py3-none-linux_x86_64'
   _libtpu_storage_directory = 'libtpu-nightly-releases'
-
-_libtpu_storage_path = f'https://storage.googleapis.com/{_libtpu_storage_directory}/wheels/libtpu/{_libtpu_wheel_name}-py3-none-linux_x86_64.whl'
+else:
+  # The postfix can be changed when the version is updated. Check
+  # https://storage.googleapis.com/libtpu-wheels/index.html for correct
+  # versioning.
+  _libtpu_wheel_name = 'libtpu-{_libtpu_version}-py3-none-manylinux_2_31_x86_64'
```
ManfeiBai, line 85: might need `f` to make sure `_libtpu_version` is caught:
```
_libtpu_wheel_name = f'libtpu-{_libtpu_version}-py3-none-manylinux_2_31_x86_64'
```

ManfeiBai (approved):
Thanks, LGTM

setup.py
zpcore: Oh, nice catch!

PR #9025, opened 2025-04-23
Propagate mutations of Tensor slices to the source Tensor.

wenxindongwork:
This PR introduces a View class which propagate mutations of Tensor slices to the source Tensor.  

Example: 
```
>>> torchax.enable_globally()
>>> x = torch.zeros(5, device = "jax")
>>> x1 = x[0:1].add_(1)
>>> print(x1)
View(Tensor(<class 'jaxlib.xla_extension.ArrayImpl'> [1.]))
>>> x1.add_(1)
View(Tensor(<class 'jaxlib.xla_extension.ArrayImpl'> [2.]))
>>> print(x)
Tensor(<class 'jaxlib.xla_extension.ArrayImpl'> [2. 0. 0. 0. 0.])
```

Previous behavior: x remains unchanged

**Background** 

Ops can be grouped into two categories 

1. Ones that return reference to original Tensors 
2. Ones that return fresh, history-free Tensors

The first category of ops include slicing (e.g. x[1:3]), transpose, reshape, and a few others. We call reference to original Tensors `View` and introduced this concept in this PR. 

A mutation done to a View will propagate to the source Tensor. E.g `x[1:3].add_(1)` will change `x`, similarly `x.transpose().add_(1)` should also change `x`. 

This PR supports mutations of the slicing op (e.g. x[1:3].add_(1)). 

**Future PR plan** 
- Add support for the remaining view generating ops (transpose, diagonal, reshape, etc) 
- Refactor the View class and merge it with the Tensor class

**PR details:** 

* Introduced `View` and `ViewInfo` class 
* Introduced `is_view_op` attribute to the `Operator` class to distinguish ops that expect View as inputs vs ops that don't.
* Added `MUTABLE_DECOMPOSITION` in `torchax/torchax/decompositions.py` to track inplace decomposition operations.

**Unit test**

* Added `torchax/test/test_view.py` to test view operations.


torchax/torchax/view.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        Initialize a ViewInfo object.
+
+        Args:
+            view_info_type: The type of view operation
+        """
+        self.view_info_type = view_info_type
+
+    @abstractmethod
+    def update_tensor(self, new_value: jax.Array, jax_array: jax.Array) -> jax.Array:
```
qihqi, line 43: jax_array should be parent?

torchax/torchax/view.py
```diff
@@ -0,0 +1,384 @@
...
+        self.parent = parent
+        self.view_info = view_info
+        self._env = env
+
+    def get_transformation_chain(self) -> List[ViewInfo]:
+        """
+        Get all view transformations from the source tensor to this view.
+        """
+        if isinstance(self.parent, View):
+            return self.parent.get_transformation_chain() + [self.view_info]
```
qihqi, line 253: this impl is quadratic on number of ancestors.
You can assign the result to a variable and just use `.append` instead.

torchax/torchax/view.py
```diff
@@ -0,0 +1,384 @@
...
+    __torch_function__ = torch._C._disabled_torch_function_impl
+
+    def source_jax(self) -> jax.Array:
+        """
+        Returns the source tensor.
+        """
+        if isinstance(self.parent, View):
+            return self.parent.source_jax()
+        else:
+            return self.parent._elem
```
qihqi, line 266: parent.jax()

torchax/torchax/view.py
```diff
@@ -0,0 +1,384 @@
...
+        # And store intermediate values
+        intermediate_values = [source_array]
+        for view_info in view_infos[:-1]:
+            intermediate_values.append(
+                view_info.transform_tensor(intermediate_values[-1])
+            )
+
+        # Update the source array with the new value by
+        # applying inverse transformations in reverse order
+        for view_info, parent_array in zip(
```
qihqi, line 318: Add a TODO here. I suspect this logic is slow.

qihqi: also this is only true for slice?

torchax/torchax/view.py
wenxindongwork: updated the inconsistent naming

torchax/torchax/view.py
wenxindongwork: maybe we can filter out some expensive edge cases like x[:] and merge sequential slicing ops to avoid some intermediate values. 

we also might not need additional intermediate values for ops like transpose, which could help freeing up some space.

qihqi (approved):

qihqi:
Thanks, this is great!

PR #9031, opened 2025-04-23
Update README for 2.7 release

zpcore:


tengyifei (approved):

README.md
```diff
@@ -332,9 +327,10 @@ The torch wheel version `2.8.0.dev20250318+cpu` can be found at https://download
 </details>
 
 ### Docker
-
+(C++11 ABI enable by default since 2.7 release)
```
tengyifei, line 330: nit: grammar: `NOTE: Since PyTorch/XLA 2.7, all builds will use the C++11 ABI by default`

README.md
```diff
@@ -26,9 +26,7 @@ started:
 To install PyTorch/XLA stable build in a new TPU VM:
 
 ```sh
-pip install torch~=2.6.0 'torch_xla[tpu]~=2.6.0' \
-  -f https://storage.googleapis.com/libtpu-releases/index.html \
-  -f https://storage.googleapis.com/libtpu-wheels/index.html
+pip install torch==2.7.0 'torch_xla[tpu]==2.7.0'
```
tengyifei, line 29: Why not use `~=2.7.0`? I thought that would make things easier if PyTorch or us ends up releasing a `2.7.1`

zpcore: There is a discussion with @yaoshiang about the change. Basically we want to strictly constrain the version. He proposes to add PyTorch version dependency in the setup.py in the future, which can make this more user friendly.

yaoshiang: Isn't it true that if torch releases a 2.7.1, our PTXLA 2.7 would not work, because the .so symbols would be different and not correctly link up anymore?

tengyifei: Actually, I'm not so sure. Ignore my comment. I think it could be better to err on the side of caution.

PR #9033, opened 2025-04-23
Migration from deprecated (@openxla/head) headers and build-targets

sdasgup3:
As part of HLO componentization [commns](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/L3kcf8b7F5c), we consolidated several components from `xla/` to `xla/hlo`. As  part of that several header files and build targets are moved from `xla/...` to `xla/hlo/...`. For sometime we maintained the mirror for those artifacts (to avoid breaking dependent projects), but recently deleted those from `xla/...`.   

Given that  currently PT/XLA is using an old pin Openxla/xla [pin](https://github.com/openxla/xla/tree/be767afe3f8749c6689b33323859d1a657893d14/xla/client) that still has refs of now deprecated aliases, this PR does the migration to the new header/targets in order to  avoid breakages at future pin bump.


qihqi (approved):
thanks! this is great!

PR #9038, opened 2025-04-25
Improve assume_pure docs and tests

tengyifei:
Beefed up the assume_pure tests and updated the docs to mention that mark_sharding is supported thanks to qihqi@'s https://github.com/pytorch/xla/pull/8989.

Also update yapf in the dev image to match CI.


qihqi (approved):

CONTRIBUTING.md
```diff
@@ -169,7 +169,7 @@ find -name '*.cpp' -o -name '*.h' -o -name '*.cc' | xargs clang-format-11 -i -st
 If your PR touches the Python source files, please run the following command before submitting a PR.
 
 ```Shell
-# How to install: pip install yapf==0.30.0
+# How to install: pip install yapf==0.40.2
```
zpcore, line 172: I have made the update. Need to rebase.

test/test_assume_pure_spmd.py
```diff
@@ -0,0 +1,79 @@
...
+import torch
+import torch_xla
+import torch_xla.runtime as xr
+from torch_xla.experimental.assume_pure import assume_pure
+from torch_xla.distributed.spmd import mark_sharding, set_global_mesh, get_1d_mesh, Mesh
+
+
+class AssumePureSpmdTest(unittest.TestCase):
+
+  def setUp(self):
```
zpcore, line 15: We can use
```
@classmethod
def setUpClass(cls):
```
since we only need to enable SPMD once.

test/test_jax_interop_spmd.py
```diff
@@ -0,0 +1,49 @@
...
+class TestJaxInteropSpmd(unittest.TestCase):
+
+  def setUp(self):
+    xb._JAX_TO_XLA_COMPUTATION_CACHE.clear()
+    # Activate SPMD
+    xr.use_spmd()
+
+    # Set up a simple SPMD mesh for these tests.
+    self.spmd_mesh = get_1d_mesh(axis_name="model")
+    set_global_mesh(self.spmd_mesh)
```
zpcore, line 21: Same here to place in `setUpClass `.

zpcore (approved):

torch_xla/experimental/splash_attention.py
```diff
@@ -86,7 +86,7 @@ def splash_attention_jax_wrapper(
       splash_attention_kernel,
       splash_attention_mask,
   )
-  mesh = Mesh.from_str(config.mesh).maybe_convert_and_get_jax_mesh()
+  mesh = Mesh.from_str(config.mesh).get_jax_mesh()
```
zpcore, line 89: Is it possible we can get of the mesh and  `@functools.partial(shard_map...` wrapper in the future?  Since we already have `mesh = xs.get_global_mesh()` under `call_jax`, once the global mesh is set and I just need to shard those input tensors for the `splash_attention_jax_wrapper` function.

zpcore: Or we can just read mesh from global mesh directly instead of decoding from `config.mesh`. Others can stay the same.

mikegre-google (approved):

qihqi:
I think what's happening with `ValueError: torch_xla device ID [1 2 3] not found in available JAX devices` is:

torch_xla.devices() has 4 devices but jax.devices() only have 1. One possibility is that jax[cuda] was not installed so jax.devices() returnts one device and that is the CPU device.

Last time I tried to add the install and hit a different error. I am OK with disabling the test for CUDA until later.

CONTRIBUTING.md
tengyifei: If there's a PR I didn't find it :)

Rebased anyhow.

torch_xla/experimental/splash_attention.py
tengyifei: Yes, it is

tengyifei:
> I think what's happening with ValueError: torch_xla device ID [1 2 3] not found in available JAX devices is [...]

That is a great point!

This made me realize that if we land this PR as-is, then not only will `call_jax` not work on GPUs, it also won't work under multi-slice TPUs. That in turn means `shard_as` won't work, and `scan` won't work either, preventing training e.g. Llama 3.1 405B.

I think I'll split out the call_jax part of this PR into a separate one, and unfortunately that one can't be landed unless we fix the PJRT client sharing between PyTorch/XLA and JAX.

torch_xla/experimental/splash_attention.py
tengyifei: Unfortunately I realized that this PR would break multi-slice LLM training, and making your suggested simplification is blocked on https://github.com/pytorch/xla/pull/9038#issuecomment-2831152384

tengyifei:
Draft for registering Jax contextual mesh in call_jax: https://github.com/pytorch/xla/pull/9043 (informational only)

PR #9044, opened 2025-04-25
reduce redundant graph in collective op

zpcore:
In all-reduce lowering, we create the redundant graph as below:

image

If we don't plan to scale on the final reduced output, the graph is unnecessary. This PR is to remove the redundant graph when scale=1.0.


tengyifei (approved):
LGTM. But note that a lot of collectives-related tests all failed.

PR #9048, opened 2025-04-28
Fix tensor factory when passing XLA device.

ysiraichi:


ysiraichi:
I'm having a hard time reproducing this issue on CI. Here, I'm getting:

```
test_rand_on_xla_cpu (__main__.TestAtenXlaTensor) ... ok
```

rpsilva-aws:
@ysiraichi Maybe we could try explicitly running the test with PJRT_DEVICE=CPU, but still within other test runs? I think when we say we repro locally, we are usually using the TPU docker (I assume in your case too). Perhaps there's some nuance there (e.g. pinned XLA versions).

ysiraichi:
Actually, I'm using the CUDA development docker. But, sure, let's try that.

PR #9053, opened 2025-04-29
Add `clang` link to `PATH`.

ysiraichi:
This PR install the `clang` and `clang++` alternative links using `update-alternatives`. This is so we don't need to specify a the Clang version inside `bazelrc` in #8665. Therefore, instead of setting `CC=/usr/lib/llvm-17/bin/clang`, we can just set `CC=clang`.

### DIFF
--- a/infra/ansible/config/apt.yaml
+++ b/infra/ansible/config/apt.yaml
@@ -14,7 +14,7 @@ apt:
       - vim
       - wget
       - clang-format
-      - clang-17
+      - clang-{{ clang_version }}
       - gcc-10
       - g++-10
       - lcov
--- a/infra/ansible/playbook.yaml
+++ b/infra/ansible/playbook.yaml
@@ -66,6 +66,8 @@
             pip.pkgs_nodeps[stage + '_' + arch] | default([], true) +
             pip.pkgs_nodeps[stage + '_' + accelerator] | default([], true)
           }}"
+
+        llvm_path: "/usr/lib/llvm-{{ clang_version }}/bin"
       tags: install_deps
 
     - role: fetch_srcs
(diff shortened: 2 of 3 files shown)


tengyifei (approved):

yaoshiang (changes requested):
Can you document this in an existing or new guide in the contributing section so that users can discover support for clang? And also speak to the value of doing so? (faster compile / better debugging symbols / etc)?

yaoshiang:
Sounds good. Would you mind filing an issue and making sure we do in fact document these flags in a future PR? I personally would have liked to use clang instead of GCc…

bhavya01:
Will this change also build cuda plugin using clang. I recently ran into an issue that was caused by our gcc use https://github.com/openxla/xla/issues/25801

ysiraichi:
@bhavya01 Yes.

ysiraichi:
@yaoshiang Could you stamp this PR?

yaoshiang (commented):
I'm not an expert on the PT/XLA and PT build systems, so I'll refrain from trying to be helpful, other than to say, let's just try to keep it as consistent with PT as possible, and to document all the controls a contributor gets.

ysiraichi:
@yaoshiang Right. I requested your review because I believe your request for changes is blocking this PR to land.

yaoshiang:
> @yaoshiang Right. I requested your review because I believe your request for changes is blocking this PR to land.

Yup, I dismissed my prior review so you should be clear - thanks.

tengyifei (approved):

bhavya01:
@ysiraichi This PR broke our nightly builds 
image

ysiraichi:
This is odd. I can't reproduce that.
I'm starting from `python:3.8-bullseye`, and running:

```
$ ansible-playbook playbook.yaml -e "stage=build" -e "accelerator=cuda" -e "arch=amd64" -e "python_version=3.8" -e "cuda_version=12.3" --tags="bazel,configure_env,install_deps"
```

Everything runs as expected. Could you share the full log?

ysiraichi:
@bhavya01 Do you have the full log?

bhavya01:
> @bhavya01 Do you have the full log?

Yes, can you access this: https://gist.github.com/bhavya01/c10a592d87a7b73596aa7ea3ce658615

PR #9054, opened 2025-04-29
Extend device data node binding API to not clone specified input tensors

rpsilva-aws:
In this PR, we extend the `_get_tensors_xla_device_data_node` binding API to return the same tensor values for a given set of specified unmutated tensor inputs. It currently returns a list of tensors that capture the XLATensor values of the graph inputs. However, these tensors end up creating new ATen tensors, which are effectively clones of the original tensors.

https://github.com/pytorch/xla/blob/c4b45a969789e93e9f2a3067e0d5a85f74f1216f/torch_xla/csrc/init_python_bindings.cpp#L2919

This makes it so that these are not eligible to be aliased at the step barrier. Instead, we extend the API to allow users to specify the list of known input tensors, such that if graph input matches one of those inputs, then the same ATen tensor is returned back to the user as instead of returning a clone.

```
>>> t4 = torch.tensor(50).to(device)
>>> t5 = t4 + 10
>>> t6 = t5 * 20
>>> [torch_xla._XLAC._xla_get_tensor_id(x) for x in [t4, t5, t6]]
[8, 10, 12]
>>> results_with_input = torch_xla._XLAC._get_tensors_xla_device_data_node([t4, t5, t6], [t4])
>>> [torch_xla._XLAC._xla_get_tensor_id(x) for x in results_with_input[1]]
[8, 18, 19]
>>> results_without_input = torch_xla._XLAC._get_tensors_xla_device_data_node([t4, t5, t6], [])
>>> [torch_xla._XLAC._xla_get_tensor_id(x) for x in results_without_input[1]]
[20, 21, 22]
```

The input tensors are kept as optional, ensuring that the API change is backwards compatible.

cc: @mcuiaws

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -2877,50 +2877,69 @@ void InitXlaModuleBindings(py::module m) {
   // -------------Dynamo Integration API Start-------------------------
   /*
    * Return tensor ids and at::tensors for all DeviceData nodes that is needed
-   * to compute the value of tensors.
+   * to compute the value of tensors. Note that all DeviceData nodes return a
+   * clone of the tensor. In case the user provides a list of uncloned tensors,
+   * then the API will ensure that any captured tensor that is included in this
+   * list does not clone the XLA tensor. This ensures that uses of the resulting
+   * tensor values can be correctly aliased, as the tensor retains the original
+   * tensor IDs.
    */
-  m.def("_get_tensors_xla_device_data_node",
-        [](const std::vector<at::Tensor>& tensors)
-            -> std::pair<std::vector<int64_t>, std::vector<at::IValue>> {
-          std::vector<int64_t> tensor_ids;
-          std::vector<at::IValue> ivalues;
-          std::vector<const torch::lazy::Node*> roots;
-          for (const at::Tensor& tensor : tensors) {
-            auto xtensor = bridge::TryGetXlaTensor(tensor);
-            if (xtensor) {
-              roots.push_back(xtensor->GetIrValue().node.get());
-            }
+  m.def(
+      "_get_tensors_xla_device_data_node",
+      [](const std::vector<at::Tensor>& output_tensors,
+         const std::vector<at::Tensor>& uncloned_tensors)
+          -> std::pair<std::vector<int64_t>, std::vector<at::IValue>> {
+        std::vector<const torch::lazy::Node*> roots;
+        for (const at::Tensor& tensor : output_tensors) {
+          auto xtensor = bridge::TryGetXlaTensor(tensor);
+          if (xtensor) {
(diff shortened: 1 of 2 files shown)


rpsilva-aws:
@tengyifei @qihqi PTAL, as this will help us move away from explicit aliasings for the gradient accumulation (needed for 2.7.1), and with some internal nightly experimental workstreams.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2877,50 +2877,67 @@ void InitXlaModuleBindings(py::module m) {
...
-          std::vector<at::IValue> ivalues;
-          std::vector<const torch::lazy::Node*> roots;
-          for (const at::Tensor& tensor : tensors) {
-            auto xtensor = bridge::TryGetXlaTensor(tensor);
-            if (xtensor) {
-              roots.push_back(xtensor->GetIrValue().node.get());
-            }
+  m.def(
+      "_get_tensors_xla_device_data_node",
+      [](const std::vector<at::Tensor>& output_tensors,
```
tengyifei, line 2888: Is the argument order swapped? Should `input_tensors` come first?

mcuiaws: we are trying not to break backward compatibility. The existing API only takes outputs. The inputs are just a hint.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -2877,50 +2877,67 @@ void InitXlaModuleBindings(py::module m) {
...
             at::Tensor tensor = bridge::AtenFromXlaTensor(
                 torch_xla::XLATensor::Create(backend_data));
             ivalues.emplace_back(tensor);
           }
-          return std::make_pair(tensor_ids, ivalues);
-        });
+        }
+        return std::make_pair(tensor_ids, ivalues);
+      },
+      py::arg("input_tensors"), py::arg("output_tensors") = py::list());
```
tengyifei, line 2940: @mcuiaws should this be `output_tensors` followed by `input_tensors` in that case?

tengyifei: I guess I'm not very sure what's the effect of putting `input_tensors` first in the Python argument list here, but putting `output_tensors` in the C++ argument definition. Is it going to match up?

mcuiaws: Yes, output_tensors should be first. Definitely a typo.

I think this affects pydoc, and when called from python using kwargs. So it should match up.

rpsilva-aws: That's a typo, probably highlighting the fact that we usually expect the order :)

We'll actually update it to make it explicit that we're specifying a set of tensors to not be cloned, instead of "input".

mcuiaws (approved):

tengyifei (approved):

PR #9057, opened 2025-04-29
add torch_xla_graph_execution_check_level (default disabled) flag that emits warning(1) or throw error(2) during tensor sync and output the python frame

aws-yyjau:
Revision from https://github.com/pytorch/xla/pull/9050

This PR introduces a new configuration flag `torch_xla_graph_execution_check_level` to provide better visibility into tensor synchronization operations during XLA graph execution. The AWS neuron team will use this flag during HLO conversion so that it can help developers catch the issues of evaluating the input tensor value during compilation.

### Key changes:
- Added new configuration flag `torch_xla_graph_execution_check_level` (default: disabled).
- Implemented warning/error logging during tensor synchronization events
- Added Python stack trace output for debugging tensor sync operations
- Log messages include relevant context about the tensor shape being performed
- Throw an error when we set to level 2

### The checking levels supported are:
- DISABLED (default): No logging
- WARNING (value: 1): check and log tensor sync operations as warnings
- ERROR (value: 2): check and log tensor sync operations as warnings and throw an XLA error

###  Example usage:
```python
import torch_xla
torch_xla._XLAC._set_torch_xla_graph_execution_check_level(1)
```

### This enhancement helps developers:
The team saw issues when we use the tensor value for if-else statement.
For example,
```python 
def forward(self, tensor):
  if tensor[0] == 1:
     return tensor
  else:
     return tensor * 2
```
The example above can compile and run. However, it may make the developers to believe the tensors can be evaluated on the fly, leading to unexpected behaviors.
With the change and some other future changes during graph tracing, we can
1. Identify potential code path issue in XLA graph execution and prevent the users from using tensor values in the graph
1. Debug and trace tensor synchronization issues more effectively

### Testing:

Added unit tests for different check levels
Verified log output format and stack trace information
Tested with tensor sync scenarios

### Documentation (To do):

Add usage examples and checking level descriptions
Include troubleshooting guide for common tensor sync issues


tengyifei (commented):
Hi, thanks for the contribution!

What is the relationship of this feature with `PT_XLA_DEBUG_LEVEL` c.f. https://github.com/pytorch/xla/blob/master/docs/source/learn/troubleshoot.md#pytorchxla-debugging-tool ?

Would it be possible to improve the existing feature and/or maybe introduce a Pythonic API for it?

aws-yyjau:
The performance for the commit was verified using r2.6_aws_neuron branch (https://github.com/pytorch/xla/pull/9077#issuecomment-2856318403).

aws-yyjau:
Hi reviewers (@tengyifei , @rpsilva-aws , @zpcore , @miladm  ), 
Greatly appreciate your review. Would like to see if it can get merged soon.

tengyifei:
Can you give either of those two features a try and see if that solves your problem?

aws-yyjau:
Hi @tengyifei 

Thanks for your comments. I'm reviewing the features coming along with PT_XLA_DEBUG_LEVEL which seems quite useful. Will come back and update soon. Thanks!

aws-yyjau:
Hi @tengyifei 

I checked the flags and the code around it.
Looks like the flag `torch_xla._XLAC._set_allow_execution` can achieve the thing we need in this PR.
When `torch_xla._XLAC._set_allow_execution(False)`, any tensor value sync will trigger an error.

We are thinking about adding tests around `_set_allow_execution` so that the feature can be guarded, like this: https://github.com/pytorch/xla/blob/7fd11d93359f37cba75b957669dda48c4fadb224/test/test_xla_graph_execution.py

Issue #9049, opened 2025-04-28
Placeholder tracing segmentation faults


rpsilva-aws:
## 🐛 Bug

There is a segmentation fault when attempting to stage the HLO dump from that involve placeholder tensors. The issue seems to originate from the PjRt computation client, as it attempts to execute the computation. However, for placeholder tensors, it requires the buffers to be present from a given PjRt data, which is not the case for placeholders. We should not be doing any materialization when tracing and returning the HLO, so this seems undesired behavior.

## To Reproduce

```
import torch
import torch_xla
from torch_xla.core.xla_builder import create_placeholder_tensor

shape = (32,32)
dtype = torch.float32
p = create_placeholder_tensor(shape, dtype)
result = p + 1.0
print(torch_xla._XLAC._get_xla_tensors_hlo(result))
```

## Environment

 - Reproducible on XLA backend [CPU/TPU/CUDA]: CPU
 - torch_xla version: 2.6+

rpsilva-aws:
@ysiraichi https://github.com/pytorch/xla/pull/9062 to at least prevent the seg fault, since the use case here is valid.

PR #9062, opened 2025-04-29
Assert on empty PJRT buffers

rpsilva-aws:
Simple CR to avoid a segmentation fault when there are placeholder tensors involved when attempting to de-reference the device from the buffer.

### DIFF
--- a/torch_xla/csrc/runtime/pjrt_computation_client.cc
+++ b/torch_xla/csrc/runtime/pjrt_computation_client.cc
@@ -738,6 +738,9 @@ PjRtComputationClient::ExecuteComputation(
   for (auto& argument : arguments) {
     const PjRtData* pjrt_data = dynamic_cast<PjRtData*>(argument.get());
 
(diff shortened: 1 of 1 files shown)


bhavya01:
Is it possible to add a test for this?

Looking at #9049 quickly, I didn't follow the path how calling `_get_xla_tensors_hlo` leads to `ExecuteComputation` call on the PjRtClient

rpsilva-aws:
> Looking at https://github.com/pytorch/xla/issues/9049 quickly, I didn't follow the path how calling _get_xla_tensors_hlo leads to ExecuteComputation call on the PjRtClient

Good question, that'll need to be explained as part of the issue above.

Usually, this sort of checks don't require a standalone test, as these invariants only exist to capture those particular undesired cases. In this case, it gives us a stacktrace and error message instead of a seg fault. Is the request to add a test to check that `_get_xla_tensors_hlo` with placeholders hits an assert?

PR #9065, opened 2025-04-30
Remove trailing whitespace from the repo

### LINKED ISSUE #9101: Remove trailing whitespace from the repo
## Motivation

I noticed that the repo currently contains a lot of trailing spaces across docs and code. We should remove them once and for all.

## Pitch

Nuking trailing whitespaces using regex s/\s+$//

ghpvnist:
Nuking trailing whitespaces using regex `s/\s+$//`

closes #9101


tengyifei (approved):

tengyifei:
@zpcore ptal, ty!

lsy323 (approved):

bhavya01 (approved):
Thanks!

qihqi (approved):

ghpvnist:
@tengyifei Do we need to increase tolerance for this particular GPU test? I think I've seen this test fail for multiple PRs.

tengyifei:
> @tengyifei Do we need to increase tolerance for this particular GPU test? I think I've seen this test fail for multiple PRs.

I don't think so. That looks like a legit bug to me. Because the delta is not small.

In any case, we can merge this without blocking on GPU tests. If we want to do something with the tolerance, that should be done in a separate PR.

tengyifei:
@ghpvnist feel free to squash and merge

tengyifei:
Specifically the number is very close to 113/125

PR #9072, opened 2025-05-01
Disable running slow CI for doc-only changes

### LINKED ISSUE #9073: Disable running slow CI for doc-only changes
## Motivation

I noticed that we're running expensive (and long) CI even for no-op changes like removing trailing whitespaces from markdown files.

## Pitch

* Saves valuable TPU resources for trivial changes.
* Faster PR merge times for quick doc fixes.

ghpvnist:
While getting #9065 reviewed, I noticed that we're running expensive (and long) CI even for no-op changes like removing trailing whitespaces from markdown files.

Note that this only skips CI when the entire PR only consists of markdown changes. Removing the CI for comments are more difficult. However, this is a step in the right direction to save some resources and enable merging PRs faster.

fixes #9073

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - master
       - r[0-9]+.[0-9]+
+    paths-ignore:
+      - '**.md' # Ignore any file ending with .md in any directory
   push:
     branches:
       - master
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4,8 +4,6 @@ on:
     branches:
       - master
       - r[0-9]+.[0-9]+
-    paths-ignore:
-      - '**.md' # Ignore any file ending with .md in any directory
   push:
     branches:
       - master
@@ -17,9 +15,20 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check_code_changes:
+    name: Check Code Changes
+    uses: ./.github/workflows/_check_code_changes.yml
+    with:
+      event_name: ${{ github.event_name }}
+      # For pull_request, use PR's base and head. For push, use event's before and sha.
+      base_sha: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.before }}
+      head_sha: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
 
   get-torch-commit:
     runs-on: ubuntu-24.04
+    needs: [check_code_changes]
+    # Only run if code changes detected
+    if: needs.check_code_changes.outputs.has_code_changes == 'true'
     outputs:
       torch_commit: ${{ steps.commit.outputs.torch_commit }}
     steps:
@@ -31,7 +40,9 @@ jobs:
   build-torch-xla:
     name: "Build PyTorch/XLA"
     uses: ./.github/workflows/_build_torch_xla.yml
-    needs: get-torch-commit
+    needs: [check_code_changes, get-torch-commit]
+    # Only run if code changes detected
+    if: needs.check_code_changes.outputs.has_code_changes == 'true'
     with:
       dev-image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.10_tpuvm
       torch-commit: ${{needs.get-torch-commit.outputs.torch_commit}}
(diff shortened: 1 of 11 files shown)


tengyifei:
Currently we require that the CPU and TPU tests must pass before a PR can be landed.

If these checks are skipped, will that prevent any markdown changes from landing, since GitHub checks would look like this:

screenshot-2025-04-30-17-46-33

ghpvnist:
Okay I've done some testing on my fork. It should trigger the workflows when there are code changes. If PR only has markdown changes, it should return success early and skip running the actual tests.

tengyifei (changes requested):
Thanks! The logic looks sensible to me, but looks like the workflow file has a syntax error:

screenshot-2025-04-30-23-59-53

.github/workflows/build_and_test.yml
```diff
@@ -15,99 +15,178 @@ concurrency:
...
     uses: ./.github/workflows/_build_torch_xla.yml
-    needs: get-torch-commit
-    with:
-      dev-image: us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/development:3.10_tpuvm
-      torch-commit: ${{needs.get-torch-commit.outputs.torch_commit}}
-      timeout-minutes: 240
-    secrets:
-      gcloud-service-key: ${{ secrets.GCLOUD_SERVICE_KEY }}
+    needs: [check_code_changes, get-torch-commit]
+    steps:
```
bhavya01, line 47: Probably just add an if statement here checking for code changes instead of steps and add a separate common stage for skip build notification instead of adding it to every step. I understand that we might need finer control to skip individual steps but for this PR, this might be sufficient

```
skip-build-notification:
    name: "Skip Build Notification"
    runs-on: ubuntu-latest
    needs: check-code-changes # Depends on the change detection job
    if: needs.check-code-changes.outputs.has_changes != 'true'
    steps:
      - name: Echo no changes message
        run: echo "No code changes detected. Skipping PyTorch/XLA build."
```

ghpvnist: My understanding is that each of these jobs are required to run. Having the if statement will bypass running them leaving the CI in pending state.

tengyifei: yeah, it's a github quirk i think

tengyifei:
@ghpvnist still not works:

```
Invalid workflow file: .github/workflows/build_and_test.yml#L54
The workflow is not valid. .github/workflows/build_and_test.yml (Line: 54, Col: 9): Unexpected value 'secrets' .github/workflows/build_and_test.yml (Line: 44, Col: 5): Required property is missing: runs-on
```

tengyifei:
@ghpvnist could you try rebasing? looks like there are still test failures

ghpvnist:
I did rebase this ~2 hrs ago. @tengyifei Does `nvidia-smi: not found` or `docker: command not found` ring a bell?

tengyifei:
Unfortunately I'm not sure why there would be an nvidia-smi failure on CPU CI. 
NVIDIA-SMI is used to check GPU chips status. 
If we're running that on CPU CI, it's expected for it to fail. 
Maybe the yaml rules got mixed up somehow? 
@ghpvnist

.github/workflows/build_and_test.yml
```diff
@@ -15,32 +15,47 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check_code_changes:
+    name: Check Code Changes
+    uses: ./.github/workflows/_check_code_changes.yml
+    with:
+      event_name: ${{ github.event_name }}
+      # For pull_request, use PR's base and head. For push, use event's before and sha.
+      base_sha: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.before }}
```
tengyifei, line 24: I think `github.event.pull_request.base.sha` isn't quite right. On the "Check Code Changes" step of this PR, I can see the following output:

screenshot-2025-05-09-19-19-36

IOW, GitHub thinks that this PR changes some `.cpp` file, even if that is not the case.

I think that's because `base.sha` just means the SHA of the main branch at the time of running the check. Instead, we probably want the SHA of the commit right before the first commit of the PR in the author's source branch.

tengyifei (changes requested):

tengyifei:
Nice work!

.github/workflows/build_and_test.yml
ghpvnist: Updated!

tengyifei (approved):

PR #9075, opened 2025-05-01
Fix + Run `DynamicShapeDetector` tests on CI.

ysiraichi:
`DynamicShapeDetector` was introduced on #7918. However, it has the following issues:

- The test never run on CI
- [Its last commit](https://github.com/pytorch/xla/pull/7918/commits/7618f5d48b23ac852b2e34a6e29c1a73acd7e481) changed the parameter name of `torch_xla.compile()` from `allowed_traces` to `num_different_graphs_allowed`. However, the tests weren't changes accordingly.
- The tests used the context manager `assertRaises(RuntimeError, msg=msg)`. However, `msg` was actually supposed to be the expected error.
- From the `torch_xla.compile()` parameter name to the names used in the implementation, there was no standard. "Traces" and "graphs" were used interchangeably.

This PR fixes the aforementioned issues by:

- Adding the test to the _run_tests.sh_ file
- Replace `assertRaises` by `assertRaisesRegex`
- Standardize the name to `max_different_graphs` everywhere
- Prefer "graphs" to "traces"


ysiraichi:
@tengyifei Could you take a look at this PR?

tengyifei (approved):

PR #9078, opened 2025-05-01
Refine the gradient accumulation API

rpsilva-aws:
In this PR, we refine the gradient accumulation API to include:
- Making the body function wrapper pure without side effects
- Enforcing train step pure requirements
- Simplifying the mapping logic
- Moving all train step non specific loop logic to the body wrapper
- Initializing local accumulated gradients on the device (prevent requiring a data transfer if not present)
- Change the API to return a tuple of carried tensors, instead of unpacking
- Remove the explicit buffer donation, given the function is pure and https://github.com/pytorch/xla/pull/9054
- RNG fix for all iterations

Testing:
- Validated the existing A/B testing for MLP with and without grad checkpointing
- Added a few basic sanity tests
- Validated the API with Llama 3.1 8B

cc: @mcuiaws

### DIFF
--- a/test/neuron/run_tests.sh
+++ b/test/neuron/run_tests.sh
@@ -167,6 +167,7 @@ function run_xla_op_tests1 {
   run_xla_hlo_debug "$CDIR/stablehlo/test_stablehlo_save_load.py"
   run_save_tensor_ir run_test "$CDIR/spmd/test_spmd_graph_dump.py"
   run_save_tensor_hlo run_test "$CDIR/spmd/test_spmd_graph_dump.py"
+  run_test "$CDIR/test_gradient_accumulation.py"
 }
 
 function run_xla_op_tests2 {
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -243,6 +243,7 @@ function run_xla_op_tests3 {
   run_test "$CDIR/spmd/test_spmd_parameter_wrapping.py"
   run_test "$CDIR/spmd/test_mp_input_sharding.py"
   run_test "$CDIR/spmd/test_train_spmd_linear_model.py" "$@" --skip-gradient-checkpointing
+  run_test "$CDIR/test_gradient_accumulation.py"
   run_save_tensor_hlo run_test "$CDIR/spmd/test_spmd_lowering_context.py"
   run_test "$CDIR/test_operations_hlo.py" "$@" --verbosity=$VERBOSITY
   run_test "$CDIR/test_input_output_aliases.py"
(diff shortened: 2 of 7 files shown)


tengyifei (approved):

rpsilva-aws:
Hmm, only PJRT_DEVICE=CUDA is having issues with the existing MLP A/B test: `SIG11 on torch_xla::runtime::PjRtComputationClient::PjRtShardedData::GetHandle()`. I'll look into it.

tengyifei:
@rpsilva-aws the GPU CI is not very stable. I would worry only about TPU CI and CPU CI for now, and make sure your tests are registered in those two environments!

tengyifei:
You're welcome to file a GPU-specific issue.

rpsilva-aws:
@tengyifei Thanks, perfect - will do! Do we have the means to disable a test for GPU?

rpsilva-aws:
https://github.com/pytorch/xla/issues/9128

tengyifei:
@rpsilva-aws you can disable a GPU test by marking it as "skipped" using the unittest API. When you disable a test you should attach the bug reference URL in the message. Use your judgement as to whether the test is truly due to some other broken thing in GPU vs caused by a bug in your PR.

rpsilva-aws:
Absolutely, had I not tried it to reproduce with A100, it'd be harder to judge - but given it succeeded with it, I don't think it is blocking.

In any case, I will take the slow approach and flush a few logs, and partially skip the test and make a better decision after a couple CI runs.

ysiraichi:
@rpsilva-aws Did you end up skipping the GPU test? Or did it start passing out of nowhere?

rpsilva-aws:
It is not currently skipped. We suspected a race condition with the [iterable tensor](https://github.com/pytorch/xla/pull/9078/files#diff-ce5f2ef27d0b95e9957a3f1f310e5c108c3d2cf2e10d3d9620e6886a5117b3adR168) when [syncing](https://github.com/pytorch/xla/pull/9078/files#diff-ce5f2ef27d0b95e9957a3f1f310e5c108c3d2cf2e10d3d9620e6886a5117b3adR176) these device data nodes. Hence, added a [synchronous](https://github.com/pytorch/xla/pull/9078/files#diff-ce5f2ef27d0b95e9957a3f1f310e5c108c3d2cf2e10d3d9620e6886a5117b3adR197) ops wait operation to confirm it. That might've done it. We're just pending on https://github.com/pytorch/xla/issues/9049 to remove the iterable tensors needed for staging the body function.

PR #9083, opened 2025-05-02
Check the scan op input for requires_grad

### LINKED ISSUE #8783: torch_xla scan forces inputs to have gradients
The snippet https://github.com/pytorch/xla/blob/00fac787e986a51e1d5d8b74603867cea4086477/torch_xla/experimental/scan.py#L217-L226 is probably wrong.

The most obvious example is that if one of the input is an integer, then it can't possibly have gradients.

haifeng-jin:
Resolves #8783
Still a draft PR.


Topologized:
FYI, although I know it's somewhat unlikely that this will actually be used, complex tensors can have gradients as well, and they're not supported by this commit.

Issue #9074, opened 2025-05-01
Unsupported unsigned (16, 32, 64) integer tensor types


rpsilva-aws:
## 🐛 Bug

We moved the mapping the unsigned integer (16, 32, 64 bits) XLA Builder's types to PyTorch's in https://github.com/pytorch/xla/pull/8873, which was previously only the case for 8-bit integers.

However, it might be the case that it matches the same set of supported tensor types in:

https://github.com/pytorch/xla/blob/64a676c0f79e1d8d2bfcc7379095231c2e5e0db7/torch_xla/csrc/tensor_util.cpp#L642

Hence, when we try to build an op with an (16, 32, 64 bits) unsigned integer with the XLA builder, we see an error:
```
RuntimeError: torch_xla/csrc/tensor_util.cpp:705 : Tensor type not supported: CPUUInt64Type
````

We need to investigate if we can close the bridge there as well, or if the limitation is expected - in which case, we need to revert the mapping changes to the earlier one (not ideal, since it's semantically incorrect, but wouldn't fail).

## To Reproduce

(Note that the same applies to 16 and 32 bits).
```
>>> import torch_xla
>>> import torch_xla.core.xla_builder as xb
>>> tensor = xb.Op.scalar(curr_iter.builder(), 1, dtype=xb.Type.U64)
>>> builder = xb.create_builder('builder')
>>> a = xb.Op.scalar(builder, 1, dtype=xb.Type.U64)
```

## Environment

 - Reproducible on XLA backend [CPU/TPU/CUDA]: Any
 - torch_xla version: 2.7

rpsilva-aws:
@iwknow @tengyifei

rpsilva-aws:
An use case is using the XLA builder to extend/work around seed states, which are u64 by default, with the XLA builder. This is currently blocking a current workstream to fork a local RNG seed in the gradient accumulation API.

tengyifei:
Thanks! @iwknow

iwknow:
change merged. you may close the issue @rpsilva-aws

PR #9090, opened 2025-05-05
Add missing tensor data types (unsigned int 16, 32, 64) to PopulateTensorBuffer

iwknow:
Add missing tensor data types (unsigned int 16, 32, 64) to PopulateTensorBuffer. The downstream function TensorToBufferSType already support support the missing data types.

This addresses https://github.com/pytorch/xla/issues/9074#issuecomment-2848324813

### DIFF
--- a/test/cpp/test_tensor.cpp
+++ b/test/cpp/test_tensor.cpp
@@ -63,6 +63,28 @@ TEST_F(TensorTest, TestConversions) {
     EXPECT_TRUE(CheckBidirectionalConversion(a, at::ScalarType::Long));
   }
   {
+    at::Tensor a = at::randint(std::numeric_limits<uint16_t>::min(),
+                               std::numeric_limits<uint16_t>::max(), {2, 2},
+                               at::TensorOptions(at::kUInt16));
+    EXPECT_TRUE(CheckBidirectionalConversion(a, at::ScalarType::UInt16));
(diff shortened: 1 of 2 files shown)


iwknow:
@rpsilva-aws please take a look.

rpsilva-aws:
Thanks @iwknow, great ownership! Running CI - do you know the historical reason to why we didn't have it earlier? Did PT have these limitations?

rpsilva-aws:
LGTM, @tengyifei double-check if you can confirm or if we need someone from PT to do so.

rpsilva-aws (approved):

iwknow:
historical reason to why we didn't have it earlier ==> i don't know

Did PT have these limitations? ==> i am not sure if i understand your question. can you elaborate with more details?

rpsilva-aws:
They are basically the same question - understanding what the limitation used to be, and confirming that it is no longer present on PyTorch. We can see if other folks know. In case of the XLABuilder, this fixes the issue - but I am not too sure about other uses of this function.

tengyifei (approved):

tengyifei:
Sorry, I also have no context to this.

rpsilva-aws:
I think it is valid, so merging to unblock some other work streams as I have tested it with CPU/TRN. We can revert if we identify any issues.

Issue #8877, opened 2025-03-24
Create a nightly torch_xla wheel without version name


bhavya01:
## 🚀 Feature
Create a torch_xla wheel with name `torch_xla.dev-cp310-cp310-linux_x86_64.whl` that always points to the latest wheel independent of the release version

## Motivation

The torch_xla nightly wheel is used for testing in other repositories like ml-auto-solutions and when the release version is updated, it breaks a bunch of tests till the wheel is updated. It would be nice to have a name independent of the release version to avoid the manual updates every time.

Relevant code: https://github.com/pytorch/xla/blob/master/infra/ansible/roles/build_srcs/tasks/main.yaml#L92

bfolie:
Fixed in #9091

PR #9091, opened 2025-05-05
Create copy of nightly wheel without version number

bfolie:
#8877

### DIFF
--- a/infra/ansible/roles/build_srcs/tasks/main.yaml
+++ b/infra/ansible/roles/build_srcs/tasks/main.yaml
@@ -126,7 +126,7 @@
         datedf="${f/.dev-/.dev${current_date}-}"
         # Remove the version number. For example:
         #   torch.dev-cp310-cp310-linux_x86_64.whl
-        unversionedf="${f/-(.*?)\.dev/.dev/1}"
+        unversionedf="$(echo $f | sed 's/-.*\.dev/.dev/1')"
 
         # Copy the file to the new filenames.
         cp "$f" "$datedf"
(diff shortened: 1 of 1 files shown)


bfolie:
@bhavya01 this creates copies of any wheels it finds, but re-reading the issue it sounds like you might only want a copy of the torch_xla wheel. LMK which is needed

infra/ansible/roles/build_srcs/tasks/main.yaml
```diff
@@ -122,10 +123,14 @@
...
         # becomes
         #   torch-2.6.0.dev20241206-cp310-cp310-linux_x86_64.whl
-        newf="${f/.dev-/.dev${current_date}-}"
-
-        # Copy the file to the new filename.
-        cp "$f" "$newf"
+        datedf="${f/.dev-/.dev${current_date}-}"
+        # Remove the version number. For example:
+        #   torch.dev-cp310-cp310-linux_x86_64.whl
+        unversionedf="${f/-(.*?)\.dev/.dev/1}"
```
bhavya01, line 129: Can we just use `unversionedf="${f/-[0-9].[0-9].[0-9].dev/.dev}"`?

The existing line of the script didn't work for me.

bhavya01 (changes requested):

bhavya01:
> @bhavya01 this creates copies of any wheels it finds, but re-reading the issue it sounds like you might only want a copy of the torch_xla wheel. LMK what is needed
No issue with creating a copy of all the wheels. The newest one just replaces the old one with the same name.

infra/ansible/roles/build_srcs/tasks/main.yaml
bfolie: Updated to use sed. Works for me now in local testing

bhavya01 (approved):

PR #9098, opened 2025-05-05
Move requires_jax to inner flash_attention functions

bhavya01:
Move `requires_jax` decorators to `_fa_custom_forward_single_device` and `_fa_custom_backward_single_device` to avoid dynamo graph breaks when using flash attention kernel.


tengyifei (changes requested):
Thanks. Is it possible to add a test to prevent regression? Otherwise, as we refactor the code, it's very easy to regress.

Issue #9068, opened 2025-04-30
distributed.all_to_all_single fails to compile when input split size is > 1


bfolie:
## 🐛 Bug

Torch.distributed's `all_to_all_single` collective fails when compiled with dynamo if the size of each split is > 1. The failure message comes from [here](https://github.com/pytorch/xla/blob/a7c3e85e5e219c6903a34f4ea349c86becbbf016/torch_xla/csrc/cross_replica_reduces.cpp#L332-L337), which implies that only split sizes of 1 are supported. I don't think this is the case -- as long as the split sizes are all equal it should be ok.

## To Reproduce

The following script fails.

```
import torch_xla
import torch_xla.core.xla_model as xm
from torch_xla import runtime as xr
import torch
import torch.distributed as dist

MULTIPLIER = 2

def main(rank: int):
  dist.init_process_group("xla", init_method="xla://")
  device = xm.xla_device()
  input_size = xr.world_size() * MULTIPLIER
  input_tensor = torch.full(
      size=(input_size,), fill_value=rank, dtype=torch.bfloat16, device=device)
  output_tensor = torch.zeros_like(input_tensor)

  def callable_fn(output_t: torch.Tensor, input_t: torch.Tensor):
    dist.all_to_all_single(output_t, input_t)
    return output_t

  fn = torch.compile(callable_fn, backend='openxla')

  output = fn(output_tensor, input_tensor)
  torch_xla.sync()

  print(f"rank {rank} output {output}")
  dist.destroy_process_group()


if __name__ == '__main__':
  torch_xla.launch(
      main,
      args=(),
  )
```

The exception is

```
RuntimeError: torch_xla does not support arbitrary split sizes for all_to_all

While executing %all_to_all_single : [num_users=1] = call_function[target=torch.ops._c10d_functional.all_to_all_single.default](args = (%arg1_1, [2, 2, 2, 2], [2, 2, 2, 2], 0), kwargs = {})
GraphModule: class <lambda>(torch.nn.Module):
    def forward(self, arg0_1: "bf16[8][1]", arg1_1: "bf16[8][1]"):
         # File: /usr/local/lib/python3.10/site-packages/torch/distributed/_functional_collectives.py:490 in all_to_all_single, code: tensor = torch.ops._c10d_functional.all_to_all_single(  # type: ignore[attr-defined]
[...]

PR #9100, opened 2025-05-06
change all_to_all check to allow for split sizes > 1

bfolie:
#9068 

Updates all_to_all test so that tensor size(0) can be greater than num_devices, and fixes the check so that it passes as long as all split sizes are equal to size(0) / num_devices

### DIFF
--- a/test/pjrt/test_collective_ops_tpu.py
+++ b/test/pjrt/test_collective_ops_tpu.py
@@ -1,11 +1,9 @@
 import numpy as np
-from typing import List
 import torch
 import torch.nn as nn
 import torch.distributed as dist
 import torch.utils._pytree as pytree
 from absl.testing import absltest, parameterized
-from unittest import mock
 import torch_xla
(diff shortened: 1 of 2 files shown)


tengyifei (approved):

bhavya01 (commented):
Just curious, what does a split mean in this context? Is it just a shard of tensor on a device  and where are the split sizes set?

bfolie:
> Just curious, what does a split mean in this context? Is it just a shard of tensor on a device and where are the split sizes set?

It's how the tensor is broken up before/after the all_to_all call. The pytorch distributed docs have an [example](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.all_to_all_single) of uneven split sizes. It can be set when calling `all_to_all_single`, and has match with the tensors themselves. But the XLA compiler doesn't support non-uniform splits

bhavya01 (approved):

Issue #9056, opened 2025-04-29
Fix the contribution instructions for creating PRs


zhanyong-wan:
## 📚 Documentation

https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md suggests to clone the original PyTorch/XLA repo directly. However, doing so makes it impossible to create PRs later unless the user has write permission to the repo. Instead, it should ask the users to fork the repo first, and then work against their fork. This allows creating PRs without having write access to the original repo.

While at this, we can also clarify the steps for creating PRs.

PR #9104, opened 2025-05-06
Enrich instructions for setting up dev environment.

zhanyong-wan:
https://github.com/pytorch/xla/issues/9056

#ptxla-usability

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,94 +11,155 @@ open an issue and discuss the feature with us. Sending a PR without discussion
 might result in a rejected PR, because we might be taking the repository in a
 different direction.
 
-## Building from source
+## Setting up the Workspace
+
+Please follow the steps below in order:
+
+### Prerequisite
+
+To work on PyTorch/XLA, you'll need a powerful Linux machine with plenty of
+CPUs and RAM. Make sure you have `git` and `docker` installed on this machine.
+If you don't, follow https://github.com/git-guides/install-git and
+https://docs.docker.com/engine/install/ to install them.
+
+### Forking the Git Repos
+
+In order to create PRs later, we need to first fork the Git repos we'll be
+working with:
+
+1.  Go to https://github.com/pytorch/pytorch and fork it as `pytorch`.
+1.  Go to https://github.com/pytorch/vision and fork it as `vision`.
+1.  Go to https://github.com/pytorch/xla and fork it as `pytorch-xla`. Note
+    the change of project name: we want to avoid confusion with the OpenXLA
+    project that PyTorch/XLA depends on, which is also named `xla`.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -11,95 +11,156 @@ open an issue and discuss the feature with us. Sending a PR without discussion
...
+### Forking the Git Repos
 
-```bash
-  git clone --recursive --depth=1 https://github.com/pytorch/pytorch.git
+In order to create PRs later, we need to first fork the Git repos we'll be
+working with:
 
-  # Install TorchVision if you need to run tests that involve vision modules
-  git clone --recursive --depth=1 https://github.com/pytorch/vision.git
+1.  Go to https://github.com/pytorch/pytorch and fork it as `pytorch`.
```
qihqi, line 30: you only need to fork torch and torchvision if you plan to make changes to it (and upstream).

Most of cases we don't need to do that.

CONTRIBUTING.md
```diff
@@ -11,95 +11,156 @@ open an issue and discuss the feature with us. Sending a PR without discussion
...
+# sandbox mode.
+umask 022
+
+# Create the workspace directory if you haven't.
+export WORKSPACE_DIR=<absolute-path-to-your-workspace>
+mkdir -p $WORKSPACE_DIR
+
+# Clone the repos.
+cd $WORKSPACE_DIR
+git clone --recursive git@github.com:<your-github-user-name>/pytorch.git
```
qihqi, line 56: I would just clone from pytorch/pytorch.git

CONTRIBUTING.md
qihqi: If one cloned pytorch/pytorch thinking there is no need to make changes upstream, and later they want to contribute to pytorch/pytorch, you can still fork then and add an upstream with git remote add. So we can keep the default path cloning directly pytorch/pytorch

qihqi (approved):
some nits

CONTRIBUTING.md
zhanyong-wan: Thanks for the advice. My thinking is to minimize the number of configurations we support. If some people fork pytorch and some don't, our instructions will have to cover for both cases, which can add more burden to both the readers and the maintainers. In the same spirit, we decided that internally we'll only support GCE VMs, even though it might not be strictly necessary. Let me know if this makes sense!

CONTRIBUTING.md
zhanyong-wan: See my reply in the other thread. Trying to provide a consolidated approach here, even though it may not be optimal.

zhanyong-wan:
I'll practice emergency merge with this PR as it's a safe, documentation-only change. :-)

PR #9113, opened 2025-05-07
Update numerical verification for SPMD Linear checkpointing

sdasgup3:
The current PR tracks a issue where an internal TPU CI is failing on v5p hardware. A specific test failing with assertion failure at [test_train_spmd_linear_model.py#L49](https://github.com/pytorch/xla/blob/f92228a137e457cfaad687f3a26c7fc8c9c83dbb/test/spmd/test_train_spmd_linear_model.py#L49) and  [test_train_spmd_linear_model.py#L51](https://github.com/pytorch/xla/blob/f92228a137e457cfaad687f3a26c7fc8c9c83dbb/test/spmd/test_train_spmd_linear_model.py#L51) with maximum absolute difference of `0.0042718649` and `0.0000191778` respectively. 

The fix here is to update the corresponding atols.

### DIFF
--- a/test/spmd/test_train_spmd_linear_model.py
+++ b/test/spmd/test_train_spmd_linear_model.py
@@ -46,9 +46,9 @@ class TestSPMDLinearModel(test_xla_sharding_base.XlaShardingTest):
       with extended_argv(['--use_gradient_checkpointing']):
         checkpointing_losses, checkpointing_result = train_and_evaluate()
         # Verify that the runs match with and without checkpointing.
-        assert torch.allclose(baseline_result, checkpointing_result)
+        assert torch.allclose(baseline_result, checkpointing_result, atol=0.005)
         assert all(
-            torch.allclose(baseline_loss, checkpointing_loss)
+            torch.allclose(baseline_loss, checkpointing_loss, atol=0.00002)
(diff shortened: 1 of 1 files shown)


lsy323 (approved):

bhavya01 (approved):

yaoshiang:
@bhavya01 have you changed your mind from your comment on May 8th, concerned about non-identical behavior from gradient checkpointing, versus your approval four hours ago?

yaoshiang:
This has been a long-lived bug and Bhavya initially expressed concern about some deviation between grad checkpointing and not. But more recently, he is okay with merging this fix and opening a new issue to further investigate numerical equiv. of checkpointing vs not. I have set this PR to automerge once the tests pass. Thanks guys.

yaoshiang:
As I look at this bug a bit further, I think it's actually okay to include some tol and not expect exact equivalence between grad checkpointing and not. 

With grad checkpointing vs not, you could imagine XLA reordering some operations, leading to some small rounding issues in the final bit of the weights. The best tolerance to measure would be on the weight and activation itself... by the time you get to loss, you can imagine a decent amount of movement.

yaoshiang:
This PR is stuck in cicd and the creator is not working on it. Closing, and replaced with #9404

PR #9115, opened 2025-05-08
[benchmarks] Fix run single config command error

haifeng-jin:
## Problem
The README in `benchmarks` have some errors in the command of running single configuration.

1. `enable_functionalization` is a required arg, but not present in the command.
2. `"false"` should not have the double quotes. `json.loads` would read it as a string.

## Fix

This PR adds `enable_functionalization` to the command and removed the double quotes for the `"false"` values in the args.


ysiraichi (changes requested):
Looks good. But, could add a test for this?

Issue #9096, opened 2025-05-05
PyTorch/XLA shouldn't crash on XLA errors


zhanyong-wan:
## 🐛 Bug

If XLA fails to compile a program and returns a Status error, PyTorch/XLA should handle it gracefully (e.g. by raising an appropriate Python exception) instead of crashing.

## To Reproduce

See  https://github.com/pytorch/xla/blob/a45a05ab8d7ac14d64ae5e2e696faddd12d9e6a0/torch_xla/csrc/runtime/pjrt_computation_client.cc#L635-L641 - `CompileAndLoad(...).value()` will crash if `CompileAndLoad(...)` returns an error. We should check for error before getting the value.

zhanyong-wan:
No, I mean to fail gracefully by raising a python exception. `ConsumeValue()` would still crash, although with a nicer message.

zhanyong-wan:
Closing this to focus on torchTPU

PR #9121, opened 2025-05-08
If compilation fails, throw a Python exception instead of crashing.

zhanyong-wan:
For https://github.com/pytorch/xla/issues/9096


torch_xla/csrc/runtime/pjrt_computation_client.cc
```diff
@@ -625,21 +626,26 @@ std::vector<ComputationClient::ComputationPtr> PjRtComputationClient::Compile(
...
+      maybe_executable =
+          client_->CompileAndLoad(instance.computation, compile_options);
     }
+    if (!maybe_executable.ok()) {
+      // This will automatically raise a Python ValueError exception.
+      // See https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html.
+      throw std::invalid_argument(
+          std::string(maybe_executable.status().message()));
+    }
+    auto executable = std::move(maybe_executable).value();
```
bhavya01, line 648: nit: `.value()` has a check for validity built in which is not need here.

Maybe just access using auto executable = std::move(*maybe_executable)

zhanyong-wan: Sorry I forgot to put this in draft mode. I plan to add tests and then ask for review.

There's one benefit of `std::move(maybe_executable).value()` - it enables warning on use of `maybe_executable` later. `std::move(*maybe_executable)` doesn't do that. I can switch to `*std::move(maybe_execuable)` though.

bhavya01: Sounds good. Thanks!

zhanyong-wan:
Messed up the git history. Will create a different PR.

Issue #8858, opened 2025-03-19
Document the difference between tracing time and execution time


tengyifei:
## 📚 Documentation

If we write a loop like

```
start = time.time()
for step in range(num_steps):
  run_model()
  xm.mark_step()
end = time.time()
```

Then `end - start` will only measure the tracing time. We'll need to do `torch_xla.sync(wait=True)` to block on device execution to measure the execution time.

We should document this in some "common FAQs/sharp edges" maybe

sdasgup3:
@tengyifei As `mark_step` is synonymous with `torxh_xla.sync` (https://github.com/pytorch/xla/blob/01b5408dded9bf5bdea3e59c387b3b201a2bdab9/torch_xla/core/xla_model.py#L1056). should this issue be deprecated.

tengyifei:
@sdasgup3 no. we can replace `mark_step` with `sync` and the same ideas still persist:

```
start = time.time()
for step in range(num_steps):
  run_model()
  torch_xla.sync()
end = time.time()
```

sdasgup3:
@tengyifei I have added a section `## Case Study: Correctly Profiling Loops with torch_xla.sync` in https://github.com/pytorch/xla/pull/9133 for the same. Please take a look.

tengyifei:
That looks great!

PR #9086, opened 2025-05-03
Replace implementation of `xla_torch.sync()` to `xm.mark_step()`

ghpvnist:
PyTorch/XLA team recommends using `xla_torch.sync()` over `xm.mark_step()` for newer releases and expect the implementation to be the same, so `sync` simply calls `xm.mark_step()`.

#8862

PR #9133, opened 2025-05-09
Document the difference between tracing time and execution time

sdasgup3:
fixes https://github.com/pytorch/xla/issues/8858

### DIFF
--- a/docs/source/learn/trace-vs-execution-time.md
+++ b/docs/source/learn/trace-vs-execution-time.md
@@ -1,0 +1,144 @@
+# Tracing Time vs. Execution Time in PyTorch/XLA
+
+When working with PyTorch/XLA, it's essential to understand that operations on
+XLA tensors are not typically executed immediately in the way they are with
+standard PyTorch tensors on CPU or CUDA devices (which operate in "eager mode").
+PyTorch/XLA employs a "lazy execution" model. This means that when you write
+PyTorch code using XLA tensors, you are primarily defining or tracing a
+computation graph. The actual execution of this graph on the XLA device (like a
+TPU) is deferred until a specific trigger point.
+
+This leads to two distinct types of "time" to consider:
+
+1. Tracing Time: The time spent by PyTorch/XLA to record operations, build the
+computation graph.
+1. Execution Time: The actual time the XLA device spends compiling (if
+necessary) and executing the computation graph.
+
+## Tracing: Defining What to Compute
+
+When you write PyTorch code using XLA tensors (e.g., tensors on a TPU),
+PyTorch/XLA doesn't execute each operation on the device right away. It traces
+these operations, adding them to an internal computation graph. If you measure
+the duration of code that only performs XLA operations without an explicit
+instruction to wait for the device, you are primarily measuring this tracing
+time plus Python overhead.
+
+### Illustrating Tracing-Dominant Measurement
+
+Consider the following conceptual code:
+
+```Python
+# Assume 'a' and 'b' are XLA tensors
+# start_time = time.perf_counter()
+
+# This operation is recorded in PyTorch/XLA's graph
+result = torch.matmul(a, b)
+
+# end_time = time.perf_counter()
+# elapsed_time = end_time - start_time
+```
+
+The `elapsed_time` here would predominantly reflect how long it took PyTorch/XLA
+to trace the matmul operation. The actual matrix multiplication on the XLA
+device is not started.
+
+## Execution: Device Computation
+
+To accurately measure XLA device computation time:
+
+1. Ensure the traced computational graph is sent to the device for execution.
+
+1. Make sure your Python script pauses its timer until the device finishes processing.
+
+### Triggering Execution and Ensuring Completion
+
+Several mechanisms trigger graph execution and/or ensure completion:
+
+1. `torch_xla.sync(wait=True)`: This is the most direct method for benchmarking.
+It ensures all pending XLA operations are launched and, crucially, blocks the
+Python script until the device finishes.
+
+1. `Data Access/Transfer`: Operations like `tensor.cpu()`, `tensor.item()`, or
+printing an XLA tensor require the actual data. To provide it, PyTorch/XLA must
+execute the graph that produces the tensor and wait for its completion.
+
+1. `torch_xla.core.xla_model.mark_step()`: Commonly used in loops, this signals
+to XLA that a logical unit of work (e.g., a training step) is defined and can be
+executed. It triggers execution but doesn't inherently block Python until device
+completion.
+
+1. `torch_xla.core.xla_model.optimizer_step(optimizer)`: In training, this
+usually implies `mark_step` and ensures relevant computations are done.
+
+1. `torch_xla.core.xla_model.unlazy(tensors)`: Blocks until specified tensors
+are materialized.
+
+### Measuring Full Execution Time
+
+To accurately measure the time including device computation:
+
+```Python
+# Assume 'a' and 'b' are XLA tensors
+# start_time = time.perf_counter()
+
+result = torch.matmul(a, b)
+
+# Explicitly wait for the XLA device to finish all pending operations
+xm.sync(wait=True)
+
+# end_time = time.perf_counter()
+# elapsed_time = end_time - start_time
+```
+
+Here, `elapsed_time` would include the tracing time, dispatch overhead, and the
+actual XLA device computation time for the matmul.
+
+## Case Study: Timing Loops with `mark_step`
+
+A common scenario involves loops, such as in model training, where
+`mark_step` is used. Consider this structure:
+
+```Python
+def run_model_step():
+    #... XLA tensor operations...
+    pass
+
+# start_loop_time = time.perf_counter()
+for step in range(num_steps):
+  run_model_step()  # Operations are traced
+  torch_xla.core.xla_model.mark_step()    # Graph for this step is submitted for execution
+# end_loop_time = time.perf_counter()
+# elapsed_loop_time = end_loop_time - start_loop_time
+```
+
+The `elapsed_loop_time` in this case primarily measures the cumulative host-side
+time. This includes:
+
+1. The time spent in `run_model_step()` for each iteration, which is largely
+tracing XLA operations.
+
+1. The time taken by `mark_step` in each iteration to process the graph for that
+step and dispatch it to the XLA device for execution.
+
+Crucially, `mark_step` typically allows the XLA device to execute
+asynchronously. The Python loop can proceed to trace the next step while the
+device is still working on the current or previous step's computations. Thus,
+`elapsed_loop_time` does not guarantee inclusion of the full device execution
+time for all `num_steps` if the device work lags behind the Python loop.
+
+In order to Measure total loop time (including all device execution),
+`torch_xla.sync(wait=True)` has to be added after the loop and before taking the
+final timestamp.
+
+```Python
+# start_loop_time = time.perf_counter()
+for step in range(num_steps):
+  run_model_step()
+  torch_xla.core.xla_model.mark_step()
+
+xm.sync(wait=True)
+
+end_loop_time = time.perf_counter()
+elapsed_loop_time = end_loop_time2 - start_loop_time
+```


docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,144 @@
...
+PyTorch/XLA employs a "lazy execution" model. This means that when you write
+PyTorch code using XLA tensors, you are primarily defining or tracing a
+computation graph. The actual execution of this graph on the XLA device (like a
+TPU) is deferred until a specific trigger point.
+
+This leads to two distinct types of "time" to consider:
+
+1. Tracing Time: The time spent by PyTorch/XLA to record operations, build the
+computation graph.
+1. Execution Time: The actual time the XLA device spends compiling (if
```
qihqi, line 15: do we want to split "execution time" to "compile time" + "execution time"?

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,144 @@
...
+
+```Python
+# Assume 'a' and 'b' are XLA tensors
+# start_time = time.perf_counter()
+
+# This operation is recorded in PyTorch/XLA's graph
+result = torch.matmul(a, b)
+
+# end_time = time.perf_counter()
+# elapsed_time = end_time - start_time
```
bfolie, line 39:
```suggestion
### INCORRECT ###
# elapsed_time = end_time - start_time
```

The primary goal of this document is to teach people how to correctly profile their code, right? I suggest making it abundantly clear, in an attention grabbing way, that this is not what they should do (which hopefully leads them to the correct way to do it, below).

I might also take the above sentence "If you measure the duration of code that only performs XLA operations without an explicit instruction to wait for the device, you are primarily measuring this tracing time plus Python overhead." and bold it

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+The `elapsed_time` here would predominantly reflect how long it took PyTorch/XLA
+to trace the matmul operation. The actual matrix multiplication on the XLA
+device is not started.
```
yaoshiang, line 44: This may or may not be a good idea, but there is something identical about eager cuda since tensors are futures, not materialized values. You need torch.cuda.synchronize to properly time ops.

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Execution: Device Computation
+
+To accurately measure XLA device computation time:
+
+1. Ensure the traced computational graph is sent to the device for execution.
+
+1. Make sure your Python script pauses its timer until the device finishes processing.
```
yaoshiang, line 52: I didn't quite follow this one - did you mean we don't launch the timer until after the tracing is done but we start the ops via a barrier?

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,144 @@
...
+
+1. `torch_xla.sync(wait=True)`: This is the most direct method for benchmarking.
+It ensures all pending XLA operations are launched and, crucially, blocks the
+Python script until the device finishes.
+
+1. `Data Access/Transfer`: Operations like `tensor.cpu()`, `tensor.item()`, or
+printing an XLA tensor require the actual data. To provide it, PyTorch/XLA must
+execute the graph that produces the tensor and wait for its completion.
+
+1. `torch_xla.core.xla_model.mark_step()`: Commonly used in loops, this signals
```
yaoshiang, line 66: typo: loops

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,144 @@
...
+to XLA that a logical unit of work (e.g., a training step) is defined and can be
+executed. It triggers execution but doesn't inherently block Python until device
+completion.
+
+1. `torch_xla.core.xla_model.optimizer_step(optimizer)`: In training, this
```
yaoshiang, line 71: Can we make a stronger statement that "this wraps mark_step", if that is true?

yaoshiang (commented):
Thanks for adding to the docs! Consider making this a runnable doc if you think it makes sense to do so.

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,144 @@
...
+`torch_xla.sync(wait=True)` has to be added after the loop and before taking the
+final timestamp.
+
+```Python
+# start_loop_time = time.perf_counter()
+for step in range(num_steps):
+  run_model_step()
+  torch_xla.core.xla_model.mark_step()
+
+xm.sync(wait=True)
```
tengyifei, line 140: nit: replace all `xm.sync` with `torch_xla.sync`, which is what we recommended in all the docs elsewhere

docs/source/learn/trace-vs-execution-time.md
sdasgup3: makes sense. Done.

docs/source/learn/trace-vs-execution-time.md
sdasgup3: Thanks! updated the text. PTAL.

docs/source/learn/trace-vs-execution-time.md
sdasgup3: Sorry for the confusion. Clarified using `1. Make sure the Python script wait until the XLA device has completed all its
assigned computations before taking the final timestamp.`

docs/source/learn/trace-vs-execution-time.md
sdasgup3: done.

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,175 @@
...
+
+This leads to two distinct types of "time" to consider:
+
+1. Tracing Time: The period during which PyTorch/XLA records your operations and
+builds this graph on the host (CPU).
+
+1. Device Time: The period during which the XLA device (e.g., TPU) actually
+performs the computations. This can be further broken down:
+
+   - Compilation Time: The time XLA takes to compile the traced graph into
```
tengyifei, line 19: The compilation runs on the host (inside the Python interpreter). It's not part of the device time. It _may_ be considered part of the tracing time, since `torch_xla.sync()` will synchronously block on the compilation.

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,175 @@
...
+
+1. `torch_xla.sync(wait=True)`: This is the most direct method for benchmarking.
+It ensures all pending XLA operations are launched and, crucially, blocks the
+Python script until the device finishes.
+
+1. `Data Access/Transfer`: Operations like `tensor.cpu()`, `tensor.item()`, or
+printing an XLA tensor require the actual data. To provide it, PyTorch/XLA must
+execute the graph that produces the tensor and wait for its completion.
+
+1. `torch_xla.core.xla_model.mark_step()`: Commonly used in loops, this signals
```
tengyifei, line 82: `mark_step()` is synonymous with `torch_xla.sync`. I think we can merge their descriptions.

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,175 @@
...
+executed. It triggers execution but doesn't inherently block Python until device
+completion.
+
+1. `torch_xla.core.xla_model.optimizer_step(optimizer)`: In training, this
+usually implies `mark_step` and ensures relevant computations are done.
+
+1. `torch_xla.core.xla_model.unlazy(tensors)`: Blocks until specified tensors
+are materialized.
+
+### Correctly Measuring Device Time (Compilation + Execution)
```
tengyifei, line 93: nit: Compilation is part of the Host-Side Activity

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,175 @@
...
+
+```Python
+def run_model_step():
+    #... XLA tensor operations...
+    pass
+
+# start_loop_time = time.perf_counter()
+for step in range(num_steps):
+  run_model_step()  # Operations are traced
+  torch_xla.core.xla_model.mark_step()    # Graph for this step is submitted for execution
```
tengyifei, line 139: nit: prefer to use `torch_xla.sync()` for all places we're using `mark_step()`

docs/source/learn/trace-vs-execution-time.md
sdasgup3: Yes, we should based on https://github.com/pytorch/xla/blob/8450607a307856f623ff35d6bfcec1494c8142f5/torch_xla/core/xla_model.py#L1241 and 
https://github.com/pytorch/xla/blob/8450607a307856f623ff35d6bfcec1494c8142f5/API_GUIDE.md?plain=1#L147

Updated the text to

```
1. `torch_xla.core.xla_model.optimizer_step(optimizer)`: Reduces gradients,
applies optimizer updates, and conditionally triggers `mark_step` via its
barrier argument (default False, as data loaders often handle `mark_step`).
```

docs/source/learn/trace-vs-execution-time.md
sdasgup3: @yaoshiang What is the typo here?

docs/source/learn/trace-vs-execution-time.md
yaoshiang: Make sure the Python script wait**s** until the XLA device has completed all its assigned computations before taking the final timestamp.

yaoshiang: nit: add an s to "wait" :)

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,176 @@
...
+   - Execution Time: The time the XLA device spends running the compiled code.
+
+**If you measure the duration of code that only performs XLA operations without an
+explicit instruction to wait for the device, you are primarily measuring this
+tracing time plus Python overhead. This is NOT an accurate measure of your
+model's performance on the XLA device.**
+
+## Tracing: Defining the Computation (Host-Side Activity)
+
+When you write PyTorch code using XLA tensors (e.g., tensors on a TPU),
```
mikegre-google, line 32: This paragraph is essentially the same as the one directly above. I think we can remove the paragraph above.

docs/source/learn/trace-vs-execution-time.md
sdasgup3: That is a good point! I have rephrased the host time to include tracing and cross compilation. With that execution is only attributed to device. PTAL

docs/source/learn/trace-vs-execution-time.md
sdasgup3: @tengyifei My motivation behind keeping the subtle difference  `mark_step` and `sync` 
```
mark_step: doesn't inherently block Python until device completion.

VS

sync: blocks the Python script until the device finishes.
```

to illustrate the need for `sync` in the example at https://github.com/pytorch/xla/issues/8858#issue-2933324391. 


wdyt?

docs/source/learn/trace-vs-execution-time.md
sdasgup3: fixed. PTAL.

docs/source/learn/trace-vs-execution-time.md
sdasgup3: Based on https://github.com/pytorch/xla/pull/9133#discussion_r2093767487, my proposal would be to keep the difference exposed?

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,176 @@
...
+time. This includes:
+
+1. The time spent in `run_model_step()` for each iteration (largly tracing).
+
+1. The time taken by `mark_step` in each iteration to process the graph for that
+step and dispatch it to the XLA device for execution.
+
+Crucially, `mark_step` typically allows the XLA device to execute
+asynchronously. The Python loop can proceed to trace the next step while the
+device is still working on the current or previous step's computations. Thus,
```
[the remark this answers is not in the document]
sdasgup3: Yes, I clarified it as
```
Crucially, `mark_step` typically works asynchronously: The Python loop proceed
to trace the next step while the device is still performing its execution for
the current or previous step.
```

docs/source/learn/trace-vs-execution-time.md
sdasgup3: I do not have a strong opinion is one or other way. But one thing I would like to avoid is for the reader to get confused on repeated back and forth between bad practice and good practices. Rephrased the doc keeping that into account. PTAL.

docs/source/learn/trace-vs-execution-time.md
sdasgup3: nvm. I got to know about this change https://github.com/pytorch/xla/pull/9086 recently. With that your suggestion makes sense.

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,105 @@
+# Tracing Time vs. Execution Time in PyTorch/XLA
+
+When working with PyTorch/XLA, it's essential to understand that operations on
+XLA tensors are not typically executed immediately in the way they are with
+standard PyTorch tensors on CPU or CUDA devices (which operate in "eager mode").
+PyTorch/XLA employs a "lazy execution" model. This means that when you write
+PyTorch code using XLA tensors, you are primarily defining or tracing a
+computation graph. The compilation of the currently traced graph and it's
+subsequent execution on the device is deferred until a specific trigger point.
```
bfolie, line 9:
```suggestion
subsequent execution on the device are deferred until a specific trigger point.
```

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,105 @@
...
+   and builds the computation graph.
+
+   - Compilation Time: The time the host-side XLA compiler takes to transform
+   the traced graph into optimized device code. This is most significant on the
+   first execution of a new graph or if the graph changes.
+
+2. Device Time: The period during which the XLA device (e.g., TPU) actually
+performs the computations. This is primarily:
+
+   - Execution Time: The time the XLA device spends running the compiled code.
```
bfolie, line 26: I don't understand why it's worth calling out "time spent running the compiled code" as a sub-part of "[time spent] perform[ing] the computation"

sdasgup3: merged the two into one. PTAL

qihqi (approved):

bfolie (approved):

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,103 @@
...
+
+```Python
+# Assume 'a' and 'b' are XLA tensors
+# start_time = time.perf_counter()
+
+# This operation is recorded in PyTorch/XLA's graph
+result = torch.matmul(a, b)
+
+#!!! INCORRECT PROFILING: compilation and execution are deferred !!!
+# end_time = time.perf_counter()
```
tengyifei, line 45: Why comment out these code? Seems fine to show a wrong example for demonstration purposes imo. Maybe we could surround it with an emoji like ❌❌❌

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,103 @@
...
+This is exemplified, using `torch_xla.sync(wait=True)`, in the following
+conceptual code:
+
+```Python
+# Assume 'a' and 'b' are XLA tensors
+# start_time = time.perf_counter()
+
+result = torch.matmul(a, b)
+
+# CORRECT PROFILING: Explicitly wait for the XLA device to finish.
```
tengyifei, line 74: Maybe add a ✅ emoji to be abundantly clear

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,103 @@
...
+torch_xla.sync(wait=True)
+
+# end_time = time.perf_counter()
+# elapsed_time = end_time - start_time
```
tengyifei, line 78: Why not uncomment these code?

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+By performing warm-up iterations (running the target code with
```
tengyifei, line 81: WDYT adding an example of warm-up onto the code snippet above? We could basically run the matmul once before, then run the same matmul operation again, only measuring the second time.

tengyifei (changes requested):

docs/source/learn/trace-vs-execution-time.md
sdasgup3: done. PTAL

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,110 @@
...
+
+1. Make sure the Python script waits until the XLA device has completed all its
+assigned computations before taking the final timestamp.
+
+This is exemplified, using `torch_xla.sync(wait=True)`, in the following
+conceptual code:
+
+```Python
+# Assume 'a' and 'b' are XLA tensors
+start_time = time.perf_counter()
```
tengyifei, line 70: nit: `start_time` should be measured after the first `sync(wait=True)`, and right before the second `sync(wait=True)`. This way we only measure the execution time and not tracing time.

tengyifei (approved):

docs/source/learn/trace-vs-execution-time.md
sdasgup3: Ah I missed that. fixed.

mikegre-google (approved):

docs/source/learn/trace-vs-execution-time.md
```diff
@@ -0,0 +1,165 @@
...
+The `elapsed_loop_time` in this case primarily measures the cumulative host-side
+time. This includes:
+
+1. The time spent in `run_model()` for each iteration (largly tracing).
+
+1. The time taken by `torch_xla.sync` in each iteration to trigger the host-side
+compilation (if the graph is new or changed) and dispatch the graph for that
+step to the XLA device for execution.
+
+Crucially, `torch_xla.sync` typically works asynchronously: The Python loop
```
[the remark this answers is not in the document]
sdasgup3: done.

docs/source/learn/trace-vs-execution-time.md
yaoshiang: This is just documenting existing behavior, so I don't want to hold it up.

Can you file an issue that torch_xla.sync(wait=True) is anti-parallel to torch.cuda.synchronize(), making it unnecessarily confusing. We should refactor these sync() and mark_step() APIs to better match torch... will cost us nothing but improve the devx.

sdasgup3: Sure.
Here is the tracking issue: 
https://github.com/pytorch/xla/issues/9276

PR #9138, opened 2025-05-10
Throw a Python exception if compilation fails.

zhanyong-wan:


torch_xla/csrc/runtime/util.h
```diff
@@ -128,6 +130,53 @@ T Multiply(const S& input) {
...
 }
 
+namespace internal {
+
+// ExtractValue<U>::type is T if U is absl::StatusOr<T>, and is undefined
+// otherwise.
+template <typename U>
+struct ExtractValue;
+template <typename T>
+struct ExtractValue<absl::StatusOr<T>> {
```
qihqi, line 140: Looks like `StatusOr<T>::value_type` provides this

torch_xla/csrc/runtime/util.h
```diff
@@ -128,6 +130,53 @@ T Multiply(const S& input) {
...
+//   - if `func()` successfully returns a value of type T, returns the value;
+//   - however, if `func()` crashes (e.g. due to a CHECK), we cannot
+//     catch it; therefore we should ensure that `func()` never
+//     crashes (and fix any crash as a bug).
+template <typename Func>
+typename internal::ExtractValue<decltype(std::declval<Func>()())>::type
+RaisePythonValueErrorOnFailure(const Func& func) {
+  decltype(std::declval<Func>()()) result;
+  try {
+    result = func();
```
qihqi, line 165: Can probably simply a bit by moving logics using result inside; so this can be

```
auto result = func();
 if (result.ok()) {
    return *std::move(result);
 }
 throw std::invalid_argument(std::string(result.status().message()));
```
and the decltype in 163 can be removed.

The invalid_argument will get caught in 166 and rethrow with the same message in 167.

torch_xla/csrc/runtime/util.h
zhanyong-wan: That won't be enough though. Here we want to enforce that the type is a `StatusOr<T>`.  If we just use `...::value_type`, there's no guarantee that `...` is a `StatusOr<>`, which will lead to people using the function in cases we want want to support.

torch_xla/csrc/runtime/util.h
zhanyong-wan: I considered this and actually think the re-throw makes the logic more complex (compared with the `decltype`) and slightly less efficient.

torch_xla/csrc/runtime/util.h
zhanyong-wan: Thanks for the review. I marked this PR as a draft as it triggers some test failures I still need to debug.

torch_xla/csrc/runtime/pjrt_computation_client.h
```diff
@@ -162,6 +162,14 @@ class PjRtComputationClient : public ComputationClient {
                        const std::function<void()>& callback) override;
 
  private:
+  friend class PjRtComputationClientTest;
+
+  // If `function` is not nullptr, makes the client call it instead of the real
+  // XLA compiler when compiling. Used for injecting fault for testing.
```
qihqi, line 168: Nice!

qihqi (approved):

PR #9139, opened 2025-05-10
Test torchax on Python 3.10 - 3.12

tengyifei:
Because why not.

Why not 3.13? Because TensorFlow.

Also pin test package versions to improve determinism.


qihqi (approved):

qihqi:
thanks!

We can make tf optional too.

PR #9153, opened 2025-05-13
Don't install wheels when building docs

### LINKED ISSUE #9158: Build docs CI fails for non-code change PRs
## 🐛 Bug

"build-docs" workflow depends on the torch_xla_wheels. Now that we're not building PyTorch/XLA wheels in every PR, that step now fails to build because the wheels don't exist.

## To Reproduce

CI breaking at HEAD.

ghpvnist:
I've done some debugging and it seems that the wheels are needed to get the latest changes. I'm not sure why the docs building needs the wheels, but let's unbreak CI at HEAD first.

It's a one line fix to only build at push, so that PRs can still skip running the CI. Once they are pushed to HEAD, it will run the full CI regardless of the diff.

Other notable changes:
* Removed `has_code_changes` variable that is always defaulted to true.
* Reorder requirements.txt alphabetically.

fixes #9158


.github/workflows/build_and_test.yml
```diff
@@ -129,6 +129,7 @@ jobs:
     name: "Build docs"
     uses: ./.github/workflows/_docs.yml
     needs: build-torch-xla
+    if: github.event_name == 'push'
```
tengyifei, line 132: Interesting find... I guess previously we were pushing docs even from pending PR branches? Seems risky

tengyifei (approved):
lgtm

.github/workflows/build_and_test.yml
bhavya01, line 132: Does this mean push to the `master` branch by default?

ghpvnist: Thats right

ghpvnist: This used to be inside the `_docs.yml` but I moved it here so we can skip the workflow altogether for non-push action because we are uploading docs on push commit anyway. Previously, all the steps up to the Deploy step from `_docs.yml` would be run.

Issue #8632, opened 2025-01-27
[scan] Avoid re-tracing the combine function on every call


tengyifei:
## 🚀 Feature

It should be possible to somehow cache the traced graphs in `torch_xla.experimental.scan` so we don't trace on every call.

## Motivation

Today `torch_xla.experimental.scan` traces the user function with both AOTAutograd (to get the backward) and with LazyTensor (to lower them to HLO). AOTAutograd is very slow and we can easily become tracing bound.

tengyifei:
A prototype is at https://github.com/pytorch/xla/compare/master...flash-attention-aot-autograd#diff-540e30d11e43afaaff4e4f89efc8b6ee396b222b8748d9eac70327ba0a7e0517R547

miladm:
Isn't this related to the cache we talked about with Rochard Zou earlier today @tengyifei?

tengyifei:
> Isn't this related to the cache we talked about with Rochard Zou earlier today

Yes. But if we tried to implement the general case, IMO we'll end up reimplementing the guard system in Dynamo.

iwknow:
Hi @tengyifei, Are you working on this? if not, i can take a look at this issue

tengyifei:
Hi @iwknow , thanks for all your contributions. It really helps a lot. 

Yes, this one is free for taking too. I linked a somewhat messy prototype in a previous comment thread. I think that one might have some gaps. The idea is that we could cache the XLA computation object based on the shape of the inputs and the function. 

I'm working on a xb.call_jax feature that has a similar caching mechanism. It should demonstrate the idea more clearly. 

A benefit of this is that once scan supports assume_pure=True, we can directly use that in GRU, because we know that our implementation is pure!

iwknow:
can you please provide the links for the related `xb.call_jax` feature so i can keep as a reference? thanks!

tengyifei:
@iwknow Hi yes, here's it is: https://github.com/pytorch/xla/blob/4eea5e133d9d02eb6c8257fa86acd576eeaf2b71/torch_xla/core/xla_builder.py#L981

`call_jax` calls `jax_func_to_xla_computation`. `jax_func_to_xla_computation` calls `_jax_to_xla_computation_cache_get_or_insert`. And `_jax_to_xla_computation_cache_get_or_insert` calls `get_xla_computation` when it realizes that it needs to add a new entry to the cache.

[...]

PR #9163, opened 2025-05-14
Add cache to `value_and_grad_partitioned`

iwknow:
Add cache to `value_and_grad_partitioned`, which perform AOTAutograd tracing that is time consuming.
The cache is controlled by `is_fn_pure`. the cache key is composed of all the inputs of `scan` (i.e. fn, init, xs, partition_fn). 

this addresses https://github.com/pytorch/xla/issues/8632

### DIFF
--- a/test/scan/test_scan.py
+++ b/test/scan/test_scan.py
@@ -10,6 +10,7 @@ from torch.utils._pytree import tree_map, tree_flatten, tree_iter, tree_leaves, 
 
 import torch_xla
 import torch_xla.debug.metrics as met
+import torch_xla.experimental.scan as scan_module
 from torch_xla.experimental.scan import scan, value_and_grad_partitioned, tree_flatten_none
 
 parent_folder = os.path.dirname(os.path.dirname(__file__))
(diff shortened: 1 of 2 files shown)


iwknow:
@tengyifei please review

iwknow:
@tengyifei a kindly ping

tengyifei:
Hi @iwknow  I was on vacation. I'll review this soon.

iwknow:
@tengyifei  a kindly ping

test/scan/test_scan.py
```diff
@@ -486,6 +487,98 @@ def unpack(x):
     # as opposed to just numerically identical but otherwise an extra copy.
     assert id(stored_xs) == id(xs)
 
+  def test_scan_computation_cache(self):
+    """
+    Test that the computation cache is populated correctly.
+    """
+
```
tengyifei, line 512: To avoid side-effect, let's clear the cache before each test case with a `setUp()` hook

torch_xla/experimental/scan.py
```diff
@@ -51,17 +51,26 @@
...
+# the carry and xs.
+_SCAN_COMPUTATION_CACHE = WeakKeyDictionary()
+
 
 def scan(
     fn: Callable[[Carry, X], tuple[Carry, Y]],
     init: Carry,
     xs: X,
     partition_fn=default_partition,
+    is_fn_pure: bool = True,
```
tengyifei: Since fn is not necessarily pure, this optimization should be opt-in.

test/scan/test_scan.py
```diff
@@ -486,6 +487,98 @@ def unpack(x):
...
+                      requires_grad=True)
+    scan(fn1, init, xs, is_fn_pure=False)
+
+    cache = scan_module._SCAN_COMPUTATION_CACHE
+
+    # Check if my_scan_fn is in the cache
+    assert fn1 not in cache, "fn1 should not be in the cache"
+
 
 class PyTreeTest(TestBase):
```
tengyifei, line 614: Could we parameterize a few of the existing test cases with both `is_fn_pure=True` and `is_fn_pure=False`?

test/scan/test_scan.py
```diff
@@ -486,6 +487,98 @@ def unpack(x):
...
+    assert len(second_level_cache) > 0, "Second-level cache should not be empty"
+
+    # You can further inspect the contents of the second-level cache if needed
+    for key, value in second_level_cache.items():
+      forward, alias_input, backward = value
+      # Add assertions or print statements to check the functions
+      assert callable(forward)
+      assert callable(alias_input)
+      assert callable(backward)
+
```
tengyifei, line 546: Could we add an incrementing counter in `fn1`, and call `scan` a few times? Then I'd expect the counter to be equal to `1`, indicating that the re-tracing is skipped.

tengyifei (changes requested):
Thanks again for the contribution! Sorry it took so long -- the past few weeks were especially busy.

torch_xla/experimental/scan.py
iwknow: I set it to `true` for two reasons: 1. my experience is that most of the fn is pure. 2. if disabled by default, not many people would use the cache. what do you think?

test/scan/test_scan.py
iwknow: done

test/scan/test_scan.py
iwknow: surprisingly, the `fn1` is called twice. i asked GPT why and here is the answer that i think is reasonable.
```
The first call to fn happens during the tracing of the forward graph. This is necessary to understand the operations performed by fn and to record them in the forward graph.

The second call to fn happens during the tracing of the backward graph. To construct the backward graph, the forward pass needs to be simulated again, as the backward graph depends on the intermediate activations and outputs of the forward pass.
```

test/scan/test_scan.py
iwknow: done

iwknow:
@tengyifei updated. please review again. thanks!

torch_xla/experimental/scan.py
tengyifei: Let's use `False` for safety. Even if most people don't use it as a start, this will prove very valuable in our internal usages. For example, we could make GRU use `scan(is_fn_pure=True)`. That will get rid of the significant tracing overhead in GRU.

If the `scan` defaults to assuming the function is pure, then stuff like this will fail in mysterious ways:

```
foo=False
def combine_fn():
  if foo:
    bar()
  else:
    baz()
```

tengyifei: Sorry for the last minute nit: could we `s/is_fn_pure/fn_is_pure`? `fn_is_pure` reads more naturally!

tengyifei (changes requested):

torch_xla/experimental/scan.py
iwknow: updated the default to `False`.

tengyifei: Still `True` as of https://github.com/pytorch/xla/pull/9163/commits/7313adcdad35cc4e9947a46c9e6efbd7fdd774e3

iwknow: i think it is a convention to have `is_` as a prefix for a boolean variables (and this is how most of the boolean variables are named). This makes "if is_fn_pure" reads better.

tengyifei:
Thanks!

tengyifei (approved):

PR #9164, opened 2025-05-14
fix: reload the param groups in optimizer

avizon-aws:
The code currently only reloads the "state" in the base optimizer, however if there is a change in the param_groups, that is not updated when reloading from the checkpoint. The base optimizer also needs to take in the "param_groups" from the loaded checkpoint.

### DIFF
--- a/test/test_zero1.py
+++ b/test/test_zero1.py
@@ -88,6 +88,54 @@ class XlaZeRO1Test(test_utils.XlaTestCase):
     opt2.step()
     xm.mark_step()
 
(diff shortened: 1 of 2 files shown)


bhavya01 (commented):
Is it possible to add a test in https://github.com/pytorch/xla/blob/master/test/test_zero1.py

avizon-aws:
@bhavya01 I think the failing test is not due to this PR.

bhavya01 (approved):
Great, thanks!

PR #9171, opened 2025-05-14
test: add test_xla_graph_execution to test flags (_set_allow_execution with PT_XLA_DEBUG_LEVEL)

aws-yyjau:
Following up on the https://github.com/pytorch/xla/pull/9057, the flag is useful to catch unexpected graph execution during model tracing.

###  Example usage:
```python
# 1) print warning messages when tensor sync happens
import os
os.environ['PT_XLA_DEBUG_LEVEL'] = '2'

# 2) error out when tensor sync happens
import torch_xla
torch_xla._XLAC._set_allow_execution(False)  # this flag disallows graph execution
```

### This enhancement helps developers:
The team saw issues when we use the tensor value for if-else statement.
For example,
```python 
def forward(self, tensor):
  if tensor[0] == 1:
     return tensor
  else:
     return tensor * 2
```
The example above can compile and run. However, it may make the developers to believe the tensors can be evaluated on the fly, leading to unexpected behaviors.
With the flags used during graph tracing, we can
1. Identify potential code path issue in XLA graph execution and prevent the users from using tensor values in the graph
1. Debug and trace tensor synchronization issues more effectively

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -151,6 +151,7 @@ function run_xla_op_tests1 {
   run_dynamic "$CDIR/ds/test_dynamic_shape_models.py" "$@" --verbosity=$VERBOSITY
   run_eager_debug  "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
   run_test "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
+  run_test "$CDIR/test_xla_graph_execution.py" "$@" --verbosity=$VERBOSITY
   run_test_without_functionalization "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
   run_pt_xla_debug "$CDIR/debug_tool/test_pt_xla_debug.py"
   run_pt_xla_debug_level1 "$CDIR/debug_tool/test_pt_xla_debug.py"
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -115,6 +115,11 @@ function run_pt_xla_debug_level1 {
   PT_XLA_DEBUG_LEVEL=1 PT_XLA_DEBUG_FILE="/tmp/pt_xla_debug.txt" run_test "$@"
 }
 
+function run_pt_xla_debug_level2 {
+  echo "Running in save tensor file mode: $@"
+  PT_XLA_DEBUG_LEVEL=2 PT_XLA_DEBUG_FILE="/tmp/pt_xla_debug.txt" run_test "$@"
+}
+
 function run_torchrun {
   if [ -x "$(command -v nvidia-smi)" ] && [ "$XLA_CUDA" != "0" ]; then
     echo "Running torchrun test for GPU $@"
@@ -151,7 +156,7 @@ function run_xla_op_tests1 {
   run_dynamic "$CDIR/ds/test_dynamic_shape_models.py" "$@" --verbosity=$VERBOSITY
   run_eager_debug  "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
   run_test "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
-  run_test "$CDIR/test_xla_graph_execution.py" "$@" --verbosity=$VERBOSITY
+  run_pt_xla_debug_level2 "$CDIR/test_xla_graph_execution.py" "$@" --verbosity=$VERBOSITY
   run_test_without_functionalization "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
   run_pt_xla_debug "$CDIR/debug_tool/test_pt_xla_debug.py"
   run_pt_xla_debug_level1 "$CDIR/debug_tool/test_pt_xla_debug.py"
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -156,6 +156,7 @@ function run_xla_op_tests1 {
   run_dynamic "$CDIR/ds/test_dynamic_shape_models.py" "$@" --verbosity=$VERBOSITY
   run_eager_debug  "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
   run_test "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
+  run_test "$CDIR/test_xla_graph_execution.py" "$@" --verbosity=$VERBOSITY
   run_pt_xla_debug_level2 "$CDIR/test_xla_graph_execution.py" "$@" --verbosity=$VERBOSITY
   run_test_without_functionalization "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
   run_pt_xla_debug "$CDIR/debug_tool/test_pt_xla_debug.py"
--- a/test/test_xla_graph_execution.py
+++ b/test/test_xla_graph_execution.py
@@ -1,0 +1,74 @@
+# Parse local options first, and rewrite the sys.argv[].
+# We need to do that before import "common", as otherwise we get an error for
+# unrecognized arguments.
+import argparse
+import os
+import sys
+import torch
+import torch_xla
+import torch_xla.core.xla_model as xm
+import torch_xla.utils.utils as xu
+import unittest
+import test_utils
+import time
+
+parser = argparse.ArgumentParser(add_help=False)
+parser.add_argument('--verbosity', type=int, default=0)
+FLAGS, leftovers = parser.parse_known_args()
+sys.argv = [sys.argv[0]] + leftovers
+print(FLAGS)
+
+XLA_DISABLE_FUNCTIONALIZATION = bool(
+    os.environ.get('XLA_DISABLE_FUNCTIONALIZATION', False))
+
+
+class TestXlaGraphExecution(test_utils.XlaTestCase):
+
+  def test_graph_execution_disabled(self):
+    # Test xla_debug_level = 0
+    print("Test xla debug level disabled.")
+    torch_xla._XLAC._set_allow_execution(True)
+    os.environ['PT_XLA_DEBUG_LEVEL'] = '0'
+    start_time = time.time()
+    x = torch.ones(2, device=xm.xla_device())
+    self.assertEqual(x[0], 1.0)  # This should trigger the checking
+
+    print("--- %s seconds ---" % (time.time() - start_time))
+    del x
+
+  def test_graph_execution_warning(self):
+    # Test xla_debug_level = 2 (with WARNING)
+    print("Test xla debug level enabled.")
+    torch_xla._XLAC._set_allow_execution(True)
+    os.environ['PT_XLA_DEBUG_LEVEL'] = '2'
+    start_time = time.time()
+    x = torch.ones(2, device=xm.xla_device())
+    self.assertEqual(x[0], 1.0)  # This should trigger the checking
+    print("--- %s seconds ---" % (time.time() - start_time))
+    del x
+
+  def test_graph_execution_error(self):
+    # Test ERROR level
+    print(
+        "Test check level as runtime error with warning messages before that.")
+    torch_xla._XLAC._set_allow_execution(
+        False)  # this flag disallows graph execution
+    start_time = time.time()
+    x = torch.ones(2, device=xm.xla_device())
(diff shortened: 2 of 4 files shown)


test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,74 @@
...
+import test_utils
+import time
+
+parser = argparse.ArgumentParser(add_help=False)
+parser.add_argument('--verbosity', type=int, default=0)
+FLAGS, leftovers = parser.parse_known_args()
+sys.argv = [sys.argv[0]] + leftovers
+print(FLAGS)
+
+XLA_DISABLE_FUNCTIONALIZATION = bool(
```
bhavya01, line 21: Seems like this is not used anywhere

test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,74 @@
...
+    del x
+    print(
+        "--- Timers are added for reference. However, the 1st test runs slower due to memory initialization ---"
+    )
+
+
+if __name__ == '__main__':
+  torch.set_default_dtype(torch.float32)
+  torch.manual_seed(42)
+  torch_xla._XLAC._xla_set_mat_mul_precision('highest')
```
bhavya01, line 70: nit: I feel this is not needed for the test set up here.

test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,74 @@
...
+        "Test check level as runtime error with warning messages before that.")
+    torch_xla._XLAC._set_allow_execution(
+        False)  # this flag disallows graph execution
+    start_time = time.time()
+    x = torch.ones(2, device=xm.xla_device())
+    with self.assertRaises(RuntimeError) as e:
+      self.assertEqual(x[0], 1.0)  # This should trigger the checking
+    print("--- %s seconds ---" % (time.time() - start_time))
+    del x
+    print(
```
bhavya01, line 62: nit: Any particular reason to add time information to tests? Unless these tests are needed for benchmarking, it is cleaner to not add unnecessary information.

test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,74 @@
...
+sys.argv = [sys.argv[0]] + leftovers
+print(FLAGS)
+
+XLA_DISABLE_FUNCTIONALIZATION = bool(
+    os.environ.get('XLA_DISABLE_FUNCTIONALIZATION', False))
+
+
+class TestXlaGraphExecution(test_utils.XlaTestCase):
+
+  def test_graph_execution_disabled(self):
```
bhavya01, line 27: nit: Should this be named test_graph_execution_allowed?

test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,74 @@
...
+    self.assertEqual(x[0], 1.0)  # This should trigger the checking
+
+    print("--- %s seconds ---" % (time.time() - start_time))
+    del x
+
+  def test_graph_execution_warning(self):
+    # Test xla_debug_level = 2 (with WARNING)
+    print("Test xla debug level enabled.")
+    torch_xla._XLAC._set_allow_execution(True)
+    os.environ['PT_XLA_DEBUG_LEVEL'] = '2'
```
bhavya01, line 43: nit: Since this test is not checking the presence of a warning automatically, I don't think that it is adding any more value than the first test. 

No one would manually look at the tests unless it's failing once it's merged.

So, probably better to remove `PT_XLA_DEBUG_LEVEL` from these tests. IIUC, these tests are only checking for `torch_xla._XLAC._set_allow_execution`

test/test_xla_graph_execution.py
aws-yyjau: will remove `XLA_DISABLE_FUNCTIONALIZATION`. The rest are common setup for sys args.

test/test_xla_graph_execution.py
aws-yyjau: will remove that. Thanks!

test/test_xla_graph_execution.py
aws-yyjau: Will update the test names for clarity.

test/test_xla_graph_execution.py
aws-yyjau: I see. I'll remove the `PT_XLA_DEBUG_LEVEL` inside the test for simplicity.
However, `PT_XLA_DEBUG_LEVEL=2` is critical to capture python frame when graph execution happens.

I suggest that we trigger the test using `PT_XLA_DEBUG_LEVEL=2` so that this feature is covered. There are some existing ways of triggering the tests using `PT_XLA_DEBUG_LEVEL=1`.

aws-yyjau:
Hi @bhavya01 
Thanks for your review and comments. Please see the response and updated code.

test/test_xla_graph_execution.py
bhavya01: Thanks for the changes! It looks so clean now. Just curious, are you using PT_XLA_DEBUG_LEVEL=2 just to output debugging information or is it actually needed for test functionality?

bhavya01 (approved):

test/run_tests.sh
```diff
@@ -115,6 +115,11 @@ function run_pt_xla_debug_level1 {
   PT_XLA_DEBUG_LEVEL=1 PT_XLA_DEBUG_FILE="/tmp/pt_xla_debug.txt" run_test "$@"
 }
 
+function run_pt_xla_debug_level2 {
+  echo "Running in save tensor file mode: $@"
+  PT_XLA_DEBUG_LEVEL=2 PT_XLA_DEBUG_FILE="/tmp/pt_xla_debug.txt" run_test "$@"
```
rpsilva-aws, line 120: Why do we need `PT_XLA_DEBUG_LEVEL=2`? Is the concern that with this level, we would not be capturing the error? I think we ought avoid branching such behavior on torch-xla, but at the same time, it's relatively more important to ensure that the failure occurs even without setting the debug level.

test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,45 @@
...
+    x = torch.ones(2, device=xm.xla_device())
+    with self.assertRaises(RuntimeError) as e:
+      self.assertEqual(x[0], 1.0)  # This should trigger the checking
+    del x
+
+
+if __name__ == '__main__':
+  test = unittest.main(verbosity=FLAGS.verbosity, exit=False)
+  if xu.getenv_as('METRICS_DEBUG', bool, defval=False):
+    print(met.metrics_report())
```
rpsilva-aws, line 44: I think we should keep verbose specific to what we're testing or asserting against. If we want to intercept these errors, we can do it as part of the test - asserting that we see the expected entries in the metrics report as separate files (ideally in https://github.com/pytorch/xla/blob/master/test/test_metrics.py). However, I don't see much benefit in printing this otherwise.

test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,45 @@
...
+    del x
+
+  def test_graph_execution_disallowed_with_error(self):
+    # Test ERROR when graph execution is not allowed
+    print("Trigger runtime error for unexpected graph execution")
+    torch_xla._XLAC._set_allow_execution(
+        False)  # this flag disallows graph execution
+    x = torch.ones(2, device=xm.xla_device())
+    with self.assertRaises(RuntimeError) as e:
+      self.assertEqual(x[0], 1.0)  # This should trigger the checking
```
rpsilva-aws, line 37: Can we assert that the exception is precisely on what we expect it to be? This would mask any other runtime error that isn't issued by the flag above.

test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,45 @@
...
+  def test_graph_execution_allowed(self):
+    # Test graph execution allowed
+    torch_xla._XLAC._set_allow_execution(True)
+    x = torch.ones(2, device=xm.xla_device())
+    self.assertEqual(x[0], 1.0)  # This should trigger the checking
+    del x
+
+  def test_graph_execution_disallowed_with_error(self):
+    # Test ERROR when graph execution is not allowed
+    print("Trigger runtime error for unexpected graph execution")
```
rpsilva-aws, line 32: Remove print, and just add any context to the comment above.

test/test_xla_graph_execution.py
```diff
@@ -0,0 +1,45 @@
...
+parser.add_argument('--verbosity', type=int, default=0)
+FLAGS, leftovers = parser.parse_known_args()
+sys.argv = [sys.argv[0]] + leftovers
+print(FLAGS)
+
+
+class TestXlaGraphExecution(test_utils.XlaTestCase):
+
+  def test_graph_execution_allowed(self):
+    # Test graph execution allowed
```
rpsilva-aws, line 24: Can remove the comment, I think the naming is quite clear.

test/run_tests.sh
```diff
@@ -151,6 +156,7 @@ function run_xla_op_tests1 {
   run_dynamic "$CDIR/ds/test_dynamic_shape_models.py" "$@" --verbosity=$VERBOSITY
   run_eager_debug  "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
   run_test "$CDIR/test_operations.py" "$@" --verbosity=$VERBOSITY
+  run_pt_xla_debug_level2 "$CDIR/test_xla_graph_execution.py" "$@" --verbosity=$VERBOSITY
```
rpsilva-aws, line 159: We need the test for the TPU, GPU and Neuron files too. See `test/{device}/`, e.g. https://github.com/pytorch/xla/pull/9078/files

rpsilva-aws (changes requested):
A few comments, thanks!

test/run_tests.sh
aws-yyjau: Thanks @rpsilva-aws .
The tensor sync check will be triggered Either OR
- `_set_allow_execution(False)`
- `PT_XLA_DEBUG_LEVEL=2`

The error will be thrown when
- `_set_allow_execution(False)` and tensor sync happens

Code ref: https://github.com/pytorch/xla/blob/master/torch_xla/csrc/debug_util.cpp#L279

aws-yyjau: To cover the case when debugging level is not set, we can have test file trigger both `run_pt_xla_debug_level2` and `run_test`.

test/test_xla_graph_execution.py
aws-yyjau: I see. Thanks for the comment. I don't see a benefit of the lines. Will remove that.

test/test_xla_graph_execution.py
aws-yyjau: will add the exception check. Thanks!

test/test_xla_graph_execution.py
aws-yyjau: Will do. Thanks!

test/test_xla_graph_execution.py
aws-yyjau: will do. thanks!

test/run_tests.sh
aws-yyjau: Adding to tpu and neuron test scripts. Thanks.

aws-yyjau:
Hi @rpsilva-aws ,

I made the change in the commits above. Could you take a look? Thanks!
- [*improve the xla graph test](https://github.com/pytorch/xla/pull/9171/commits/cf5c5069a0e919195f3fb1fb079945e7f0e0e33a)
- [add tests to tpu and neuron test script](https://github.com/pytorch/xla/pull/9171/commits/2e9b2b18689e9d584a676b68ac700a1009fa2c46)

rpsilva-aws (approved):

test/test_xla_graph_execution.py
aws-yyjau: Hi @bhavya01 

Thanks for the comments. I'm aligned with @rpsilva-aws with the concept of narrow test cases.
The current tests should provide enough guard around the flags and simple enough compared to creating a separate error class. Let me know what you think. Thanks!

bhavya01: Ack! Makes sense. Since the error messages come from our own functions, it shouldn't be a problem here.

bhavya01 (approved):

PR #9180, opened 2025-05-16
[Kernel] Use call_jax to simplify the gmm pallas kernel wrapper

yaochengji:


yaochengji:
The background is there's a PR https://github.com/vllm-project/vllm/pull/18025 trying to enable torch_xla gmm kernel but encountered correctness issue.

Simplifying the wrapper to use `call_jax` can fix it.

yaochengji:
also cc @bythew3i for visibility, seems I cannot add you as a reviewer.

torch_xla/core/xla_builder.py
```diff
@@ -873,6 +873,13 @@ def jax_func_to_xla_computation(jax_func, args, kwargs, name=None):
 
     flattened_inputs, spec = jax.tree.flatten((args, kwargs))
 
+    def convert_arg_to_jax(a):
```
vanbasten23, line 876: what are the common types of a?

torch_xla/experimental/custom_kernel.py
```diff
@@ -968,23 +969,6 @@ def ragged_paged_attention(
     num_queries_per_block=None,
     vmem_limit_bytes=None,
 ):
-  if mask_value is None:
```
vanbasten23, line 971: why remove this part?

torch_xla/core/xla_builder.py
```diff
@@ -881,6 +888,7 @@ def abstractify(a):  # make a pytree leaf abstract
         return jax.ShapeDtypeStruct(a.shape, tx.ops.mappings.t2j_dtype(a.dtype))
       return a
 
+    flattened_inputs = list(convert_arg_to_jax(a) for a in flattened_inputs)
```
vanbasten23, line 891: do you know why it worked before when we don't convert the flattened_inputs to jax type?

torch_xla/core/xla_builder.py
yaochengji: It depends on args the the custom op

torch_xla/experimental/custom_kernel.py
yaochengji: Thanks for catching this. It's removed by accident.

torch_xla/core/xla_builder.py
yaochengji: Because dtype is not used as arguments in the existing tests.

test/test_gmm.py
```diff
@@ -409,8 +255,9 @@ def test_gmm_backward(self):
             group_sizes.to("xla"))
         # same gmm/tgmm was run for the `test_cache=False` case so the
         # cache should be populated now
+        new_cnt = xr.get_num_cached_compilation_graph()
         if test_cache:
-          self.assertEqual(met.counter_value('trace_pallas_cache_hit'), 2)
+          self.assertEqual(new_cnt, old_cnt)
```
vanbasten23, line 260: based on the previous test, shouldn't you also need to test the case when test_cache=False: new_cnt==old_cnt+2?

vanbasten23 (approved):
Thanks Chengji. LGTM with one comment.

test/test_gmm.py
yaochengji: I think the original test code is to make sure that there's no recompilation, the new code did the same thing.

PR #9195, opened 2025-05-19
[torchax]: JittableModule statedict handling

zmelumian972:
torchax aims to improve seamless interoperability between torch and jax

one of the parts in torch training pipeline revolves around storing and loading statedict (checkpoints)

most of the objects revolving torch checkpoints expect a (non nested) dict containing weight name and it's value (in either CPU or GPU device)

since torchax tensors are held in jax container, torch checkpointers cannot easily handle it

this changes forces JittableModule to convert state_dict() functions (both load and get), making it seamless to the user when he wants to extract the statdict prior to saving it as a checkpoint

### DIFF
--- a/torchax/torchax/interop.py
+++ b/torchax/torchax/interop.py
@@ -125,6 +125,14 @@ class JittableModule(torch.nn.Module):
       return jitted(self.params, self.buffers, *args, **kwargs)
 
     self._jitted[key] = call
+    
+  def cpu_state_dict(self, *args, **kwargs):
+    state_dict = super().state_dict(*args, **kwargs)
+    state_dict = pytree.tree_map(
+      lambda t: t.cpu(),
+      state_dict
+    )
+    return state_dict
 
 
 class CompileMixin:
--- a/torchax/torchax/interop.py
+++ b/torchax/torchax/interop.py
@@ -1,3 +1,4 @@
+from typing import Mapping, Any
 import collections
 import copy
 import functools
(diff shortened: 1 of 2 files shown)


torchax/torchax/interop.py
```diff
@@ -125,6 +126,80 @@ def call(*args, **kwargs):
       return jitted(self.params, self.buffers, *args, **kwargs)
 
     self._jitted[key] = call
+    
+  def state_dict(self, *args, **kwargs):
```
qihqi, line 130: I think this function should return torchax.Tensors (and leave the users who want to move it to CPU call the move). 

Or maybe name it cpu_state_dict.

in regular `torch.nn.Module`'s `state_dict` method returns references to weights inside of the module, so there is an implicit understanding  that there is no copy.

torchax/torchax/interop.py
```diff
@@ -125,6 +126,80 @@ def call(*args, **kwargs):
...
+      lambda t: t.cpu(),
+      state_dict
+    )
+    return state_dict
+  
+  def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True, assign: bool = False):
+    """
+    Wrapper for load_state_dict
+    
+    This function assumes torch CPU state dict and will transfer the parameters to the correct device
```
qihqi, line 151: Maybe make this function work both with CPU tensors as well as torchax.Tensors?

torchax/torchax/interop.py
```diff
@@ -125,6 +126,80 @@ def call(*args, **kwargs):
...
+    state_dict = pytree.tree_map(
+      tensor.t2j,
+      state_dict
+    )
+    # Convert ordered dict to regular dict, pjit type-safety checks
+    state_dict = dict(state_dict)
+    jitted = jax_jit(
+        lambda t: t, kwargs_for_jax_jit={"out_shardings": output_shards}
+    )
+    state_dict = jitted(state_dict)
```
qihqi, line 198: Can probably make use of `  from jax.experimental.shard_alike import shard_alike`.

Something like:
```
sharded_dict_jax = pytree.tree_map(lambda cpu_tensor, orig: shard_alike(tensor.t2j(cpu_tensor), orig), state_dict, current_state_dict)
state_dict = torch_view(sharded_dict_jax)
...
```

torchax/torchax/interop.py
zmelumian972: I prefer cpu_state_dict and keep it torch native

keeping it in torchax.Tensors might make it harder for checkpointers who expect native torch tensors to work (because torch.save does not expect torchax.Tensors)

torchax/torchax/interop.py
zmelumian972: Sure, that's a good idea

I will simply ignore converting torch tensor to torchax.tensor.Tensor if it's already there, meaning both torchax.tensor.Tensor is acceptable and torch.Tensor is acceptable

torchax/torchax/interop.py
zmelumian972: It's a bit unnatural here because `state_dict` and `current_state_dict` are not promised to have the same tree structure (user can specify strict=False and call with partial state dict)

iirc, pytrees do not support that natively

qihqi: I see, makes sense.

qihqi (approved):
Thanks! Please fix the lint with 

`yapf -i -r *.py test/ scripts/ torch_xla/ benchmarks/ torchax/`

Thanks

zmelumian972:
how do I move forward? I am unfamiliar with pytorch/XLA CI and resources

qihqi:
Hi @zmelumian972 can you rebase? The GPU CIs are not merge blocking anymore on the new HEAD. I also enabled auto-merge so if CI passes it should merge automatically. Thanks!

qihqi (approved):

zmelumian972:
Done :)

PR #9196, opened 2025-05-19
[torchax] Added support for bicubic and billinear resampling

zmelumian972:
Currently torchax does not support image resampling with antialiasing enabled

that, and the current implantation for bicubic resampling does not match torch

this PR aims to fix that

### DIFF
--- a/torchax/torchax/ops/jaten.py
+++ b/torchax/torchax/ops/jaten.py
@@ -5181,17 +5181,16 @@ def _aten_max_unpoolxd(input, indices, output_size, stride=None, padding=0):
   return output
 
 
-@op(torch.ops.aten._upsample_bilinear2d_aa)
-def _aten_upsample_bilinear2d_aa(input,
-                                 output_size,
-                                 align_corners,
-                                 scale_factors=None,
-                                 scales_h=None,
-                                 scales_w=None):
+def _aten_upsample(input,
+                   output_size,
+                   align_corners,
+                   antialias,
+                   method,
+                   scale_factors=None,
+                   scales_h=None,
+                   scales_w=None):
   # input: is of type jaxlib.xla_extension.ArrayImpl
   image = input
-  method = "bilinear"
-  antialias = True  # ignored for upsampling
 
   # https://jax.readthedocs.io/en/latest/_autosummary/jax.image.resize.html
   # Resize does not distinguish batch, channel size.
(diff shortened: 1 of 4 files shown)


torchax/torchax/ops/jaten.py
```diff
@@ -5253,6 +5253,70 @@ def _aten_upsample_bilinear2d_aa(input,
...
+      output_size,
+      align_corners,
+      True, # antialias
+      "bilinear", # method
+      scale_factors,
+      scales_h,
+      scales_w
+  )
+
+@op(torch.nn.functional.interpolate)
```
qihqi, line 5274: Move it to `jtorch.py` with `register_*` this is actually a torch function and not an ATen op. (Note currently they actually share the registry so it doesn't matter, its a convention).

torchax/torchax/ops/jaten.py
```diff
@@ -5253,6 +5253,70 @@ def _aten_upsample_bilinear2d_aa(input,
...
+      "trilinear",
+      "cubic",
+      "bicubic",
+      "tricubic",
+      "lanczos3",
+      "lanczos5",
+  )
+  is_jax_supported = mode in supported_methods
+  if not is_jax_supported:
+    raise NotImplementedError(
```
qihqi, line 5297: you can import tensor here and raise `tensor.OperatorNotFound` here (https://github.com/pytorch/xla/blob/master/torchax/torchax/tensor.py#L22)

The reason is that because this is a torch function, and a torch function usually decomposes to ATen ops. So raising OperatorNotFound will make the dispatcher attempt to dicompose and run the individual ATen ops.

This logic is triggered here: https://github.com/pytorch/xla/blob/master/torchax/torchax/tensor.py#L264

torchax/torchax/ops/jaten.py
```diff
@@ -5253,6 +5253,70 @@ def _aten_upsample_bilinear2d_aa(input,
...
+  align_corners = align_corners or False
+  
+  if mode in ('cubic', 'bicubic', 'tricubic') and not antialias:
+    return jimage.interpolate_bicubic_no_aa(
+      input,
+      size[0],
+      size[1],
+      align_corners,
+    )
+  return _aten_upsample(
```
qihqi, line 5311: same here, you want to raise OperatorNotFound so it will go through the old code path (and hopefully pass the tests in test_ops.py for interpolate.

zmelumian972: I tried it, it does not pass the test I built at `torchax/test/test_image.py` (`torchax.tensor.OperatorNotFound: Operator with name aten::_upsample_bicubic2d_aa has no lowering`) , so either I did something wrong building this test or I did something wrong building the operator

torchax/torchax/ops/jaten.py
zmelumian972: elegant, I like it, done

torchax/torchax/ops/jaten.py
zmelumian972: Sounds good to me, done

PR #9205, opened 2025-05-19
Document test/neuron/run_tests.sh.

zhanyong-wan:


test/neuron/run_tests.sh
```diff
@@ -1,4 +1,13 @@
 #!/bin/bash
+
+# This script is used in AWS Neuron's nightly build to verify that PyTorch/XLA
+# works for them. It runs a subset of the tests in test/run_tests.sh.
```
rpsilva-aws, line 4: nit: "works on Trn/Inf."

test/neuron/run_tests.sh
```diff
@@ -1,4 +1,13 @@
...
+# Eventually (after we fix or skip the rest of the tests for Neuron) we will
+# remove this and use test/run_tests.sh for Neuron nightly instead.
```
rpsilva-aws, line 6: nit: "with Trn/Inf CI instead."

I think if we are phrasing where we want to be, we can include CI as our final objective. Bridging in the gaps between CPU and Trn/Inf is ideal, but technically a separate item as well (we can keep that as is in the sentence).

test/neuron/run_tests.sh
```diff
@@ -1,4 +1,13 @@
...
+#
+# For question on this script, please contact the AWS folks in
+# https://pytorch.slack.com/archives/C02GNBKM2HM.
```
rpsilva-aws, line 9: I think we can instead suggest folks to file a bug/issue directly.

test/neuron/run_tests.sh
zhanyong-wan: Thanks! Fixed.

aws-cph (approved):

test/neuron/run_tests.sh
zhanyong-wan: I was thinking about asking quick questions. Filing an issue seems too heavyweight, and is a well-known workflow and thus doesn't need to be called out.

rpsilva-aws (approved):

PR #9209, opened 2025-05-20
Improve CI build speed by letting bazel decide how many worker threads to have

zhanyong-wan:
We hard-code the number of bazel worker threads to 16 in the CI. However, our CI workers have 48 CPUs and 96 GB RAM. We are severely underutilizing our workers.

By removing the hard-coded value, we allow bazel to decide how many worker threads to launch, based on the available CPUs. This reduces the overall build time from 2h5m to 1h42m, a 23m reduction (22% speed-up).

In particular, it reduces the build times for the following jobs on the critical path:

- Build XLA CUDA plugin / build: 40m37s => 26m43s
- Build PyTorch/XLA / build: 1h18m42s => 56m10s

The other jobs are not noticeably affected by this change (some of them don't use bazel, and some may have a different bottleneck).

I also tried hard-coding the number of worker threads to 32 or 48, and the results are slightly slower than not setting the number of worker threads.

### DIFF
--- a/.github/workflows/_build_plugin.yml
+++ b/.github/workflows/_build_plugin.yml
@@ -33,7 +33,7 @@ jobs:
       image: ${{ inputs.dev-image }}
     env:
       GCLOUD_SERVICE_KEY: ${{ secrets.gcloud-service-key }}
-      BAZEL_JOBS: 16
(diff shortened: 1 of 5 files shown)


tengyifei (approved):
Nice!

PR #9212, opened 2025-05-20
Revert "Switch bazel default spawn strategy to `sandboxed`"

zhanyong-wan:
Reverts pytorch/xla#9168 as it slows down the local build by more than 2x.


tengyifei (approved):
Revert SGTM since I have experienced several hours of slow build time in the development docker image in a TPU VM without this revert, and applying this revert fixes the slow build for me.

PR #9216, opened 2025-05-20
update openxla pin

haifeng-jin:


qihqi (approved):
thansk this is great!

PR #9218, opened 2025-05-20
Disable AMP by default on CPU

### LINKED ISSUE #9119: The benchmarks does not work on CPU because of AMP
## 🐛 Bug

The `benchmarks` in `pytorch/xla` does not work on CPU because of it is set to use AMP by default.

## To Reproduce

1. Follow the instruction to run the benchmarks in the [`README.md`](https://github.com/pytorch/xla/blob/master/benchmarks/README.md).
2. Or run this command directly
```
python xla/benchmarks/experiment_runner.py --dynamo=openxla --xla=PJRT --test=eval --test=train --suite-name=torchbench --accelerator=cpu --output-dirname=/tmp/output --repeat=1 --print-subprocess --no-resume --dump-pytorch-xla-metrics
```

## Expected behavior

Add a new arg to the CLI `--amp`. Let user configure it (disable AMP) when run on CPU.

## Environment

 - Reproducible on XLA backend [CPU/TPU/CUDA]: CPU
 - torch_xla version: master or 2.7

haifeng-jin:
Resolves #9119.

The benchmarks in pytorch/xla does not work on CPU because of it is set to use AMP by default, which supports cuda and tpu only, but not cpus.

This PR fixes this issue by disable AMP by default when the accelerator is set to cpu.

### DIFF
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -373,6 +373,9 @@ class TorchBenchModel(BenchmarkModel):
     return self.benchmark_experiment.accelerator == "tpu"
 
   def use_amp(self):
+    # AMP is only supported on cuda and tpu, not on cpu.
+    if self.benchmark_experiment.accelerator == "cpu":
+      return False
     return self.is_training() or self.model_name in config(
     ).dtype.force_amp_for_fp16_bf16_models
 
--- a/benchmarks/torchbench_model.py
+++ b/benchmarks/torchbench_model.py
@@ -375,6 +375,7 @@ class TorchBenchModel(BenchmarkModel):
   def use_amp(self):
     # AMP is only supported on cuda and tpu, not on cpu.
     if self.benchmark_experiment.accelerator == "cpu":
+      logger.warning("AMP is not used due to running on CPU.")
       return False
     return self.is_training() or self.model_name in config(
     ).dtype.force_amp_for_fp16_bf16_models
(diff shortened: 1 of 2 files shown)


bhavya01 (approved):

haifeng-jin:
Merging the PR. The failed tests is irrelevant to the change in this PR.
The tests failure is being investigated and appeared on other PRs as well.
So, merging.

haifeng-jin:
Maybe wait for @ysiraichi 's comment before merging. : )

ysiraichi (changes requested):
Thank you for the fix. I think it would be nice to also have:

1. A warning, so that the user knows AMP is not being used
2. A test, checking that we don't error and AMP is not being used

haifeng-jin:
> Thank you for the fix. I think it would be nice to also have:
> 
> 1. A warning, so that the user knows AMP is not being used
> 2. A test, checking that we don't error and AMP is not being used

I added the warning.
However, for the test, I do not see any existing tests for the benchmarks.
Is there any such test? If not, I suggest we open a separate issue and implement some of these basic tests in another PR.
WDYT? @ysiraichi

haifeng-jin:
Great! working on it!

haifeng-jin:
@ysiraichi The tests are added and passed locally.

haifeng-jin:
Merging the pull request as the CI errors are irrelevant to the changes.

PR #9221, opened 2025-05-21
Revert "Fix a clang compilation error. (#9156)"

yaoshiang:
This reverts commit ffd4488c40afbacee8154206b6c5c24543698a26.

If this passes CICD especially the bug listed in the issue, then that PR caused this cicd regression.


yaoshiang:
This wasn't the regression and I should have known if I'd looked harder at the code changes. I'll keep searching for the regression.

PR #9222, opened 2025-05-21
[torchax] Added support for bicubic and billinear resampling

qihqi:
Fork of https://github.com/pytorch/xla/pull/9196
With changes to fix tests.

### DIFF
--- a/torchax/torchax/ops/jtorch.py
+++ b/torchax/torchax/ops/jtorch.py
@@ -3,7 +3,7 @@
 import math
 import collections.abc
 import functools
-from typing import Optional, Sequence
+from typing import Optional, Sequence, Tuple
 import numpy as np
 
 import jax
@@ -13,7 +13,7 @@ from jax.experimental.shard_map import shard_map
 
 import torch
 from torchax.ops.ops_registry import register_torch_function_op
-from torchax.ops import op_base, mappings, jaten
+from torchax.ops import op_base, mappings, jaten, jimage
 import torchax.tensor
 from torchax.view import View, NarrowInfo
 import torch.utils._pytree as pytree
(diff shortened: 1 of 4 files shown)


torchax/torchax/ops/jtorch.py
```diff
@@ -512,3 +512,48 @@ def functional_linear(self, weights, bias=None):
...
+  is_jax_supported = mode in supported_methods
+  if not is_jax_supported:
+    raise torchax.tensor.OperatorNotFound(
+        f"JAX does not support interpolation mode: {mode}. Supported modes are: {supported_methods}"
+    )
+  # None check
+  antialias = antialias or False
+  align_corners = align_corners or False
+
+  if mode in ('cubic', 'bicubic',
```
tengyifei, line 547: Do we support all of `cubic`, `bicubic`, `tricubic`? Looks like the mode is just ignored and it calls into `interpolate_bicubic_no_aa`

tengyifei (approved):

torchax/torchax/ops/jtorch.py
zmelumian972: you are right, but I think it's not so much work to make `interpolate_bicubic_no_aa` become cubic or tricubic

the `cubic_kernel` can be called multiple times to create N cubic kernels

```
def bicubic_kernel(dx, dy, a=-0.75):
    kx = cubic_kernel(dx, a)
    ky = cubic_kernel(dy, a) 
    return jnp.outer(ky, kx)
```

or

```
def tricubic_kernel(dx, dy, dz, a=-0.75):
    kx = cubic_kernel(dx, a)
    ky = cubic_kernel(dy, a)
    kz = cubic_kernel(dz, a)
    return jnp.einsum('i,j,k->ijk', kz, ky, kx)
```

do you want me to force push those changes to add support?

tengyifei: Sure. If we add support. Let's add corresponding unit tests too! I think it could also be fine to defer the support -- I have no knowledge on the requirements here.

Issue #7458, opened 2024-06-25
Op info test for `kthvalue`


qihqi:
## Fix the Op info test for `kthvalue`

1. Find the line 105 of [test_ops.py](test/test_ops.py) and remove
   `kthvalue` from `skip_list`
2. Run op_info test with `pytest test/test_ops.py`
3. Fix the failure.

Please refer to 
[this guide](https://github.com/pytorch/xla/blob/master/experimental/torch_xla2/docs/fixing_op_info_test.md) as guide to fix:

Also refer to these PRs:
* https://github.com/pytorch/xla/pull/7343
* https://github.com/pytorch/xla/pull/7312
* https://github.com/pytorch/xla/pull/7325

ManfeiBai:
Hi, @priyansi, this issue would be part of the first issues to try, please feel free to comment, so that we could assign this ticket to you

qianminj123:
Does torchax support kthvalue? I saw the error "torchax.tensor.OperatorNotFound: Operator with name aten::kthvalue has no lowering" when running test, and didn't find kthvalue implementation in the repository. Also found this thread https://discuss.pytorch.org/t/how-can-i-implement-aten-kthvalue-values/215490

Shall I add the implementation of kthvalue in torchax/torchax/ops/jaten.py?

qianminj123:
nvm, I found https://github.com/pytorch/xla/blob/master/torchax/docs/fixing_op_info_test.md#fix-target-op-failure Thanks!

PR #9223, opened 2025-05-21
[torchax] Fixes test for kthvalue pytorch/xla#7458

qianminj123:
Implemented kthvalue op by JAX and enabled the test.


qihqi (approved):
fix the lint, thanks!

PR #9224, opened 2025-05-21
Fixed reloading of sharded master weights

avizon-aws:
When reloading from checkpoint, the order of the params might change and hence the indexing as well. We should not be hardcoding the indexing, it should be taken as per the loaded state dict.

### DIFF
--- a/torch_xla/distributed/zero_redundancy_optimizer.py
+++ b/torch_xla/distributed/zero_redundancy_optimizer.py
@@ -523,12 +523,13 @@ class ZeroRedundancyOptimizer(Optimizer):
     self.base_optimizer.load_state_dict(tmp)
     if 'sharded_master_weights' in state_dict:
       master_weights = state_dict['sharded_master_weights']
-      index = 0
-      for param_group, sharded_param_group in zip(
-          self.param_groups, self.base_optimizer.param_groups):
-        for param, shard in zip(param_group['params'],
-                                sharded_param_group['params']):
-          shard.data.copy_(master_weights[index])
+      for param_group, sharded_param_group, loaded_param_groups in zip(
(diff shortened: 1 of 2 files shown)


avizon-aws:
Failing test is not due to this CR, seems to be flaky test

torch_xla/distributed/zero_redundancy_optimizer.py
```diff
@@ -523,12 +523,13 @@ def load_state_dict(self, state_dict):
...
       master_weights = state_dict['sharded_master_weights']
-      index = 0
-      for param_group, sharded_param_group in zip(
-          self.param_groups, self.base_optimizer.param_groups):
-        for param, shard in zip(param_group['params'],
-                                sharded_param_group['params']):
-          shard.data.copy_(master_weights[index])
+      for param_group, sharded_param_group, loaded_param_groups in zip(
+          self.param_groups, self.base_optimizer.param_groups,
+          state_dict['param_groups']):
```
bhavya01, line 528: It seems like `state_dict[param_groups]` and `base_optimizer.param_groups` are the same because we do `base_optimizer.load_state_dict` above. Can you help me understand this?

avizon-aws: The `params` inside `base_optimizer.param_groups` are not  indices, they are the sharded model parameter tensors. Whereas in `state_dict['param_groups']['params']`, the indices are present which is what we want to use here.

bhavya01: Sorry, I'm still confused. What do these lines do then:

```
   tmp['param_groups'] = state_dict['param_groups']
    self.base_optimizer.load_state_dict(tmp)
```

PR #9227, opened 2025-05-21
Run all torchax tests with `find`

tengyifei:
Replace the manual hardcoded list of test files with `find`


torchax/test/BUILD
```diff
@@ -13,8 +13,8 @@ package(
 )
 
 py_library(
-    name = "test_base",
-    srcs = ["test_base.py"],
+    name = "base_test_util",
```
qihqi, line 16: Let's delete this file.

I think it's unintended to have a BUILD file

qihqi (approved):
thanks!

PR #9230, opened 2025-05-21
Add description of prerequisites for running benchmarks

### LINKED ISSUE #9118: [benchmarks] Add installation instructions to `benchmarks/README.md`
## 📚 Documentation

The [`benchmarks/README.md`](https://github.com/pytorch/xla/blob/master/benchmarks/README.md) does not contain the installation instructions, which is crucial for running the benchmarks.

It requires installing the [`pytorch/benchmark`](https://github.com/pytorch/benchmark) repo and other libraries like `libGL` (required by Llava).

## Solution

Add the instructions to [`benchmarks/README.md`](https://github.com/pytorch/xla/blob/master/benchmarks/README.md).
Install [`pytorch/benchmark`](https://github.com/pytorch/benchmark) as a library.
Install `libGL`.
Install any other requirements.

To verify, make sure the instructions work with the devcontainer.

haifeng-jin:
Resolves #9118.

This PR adds a new section in `benchmarks/README.md` to describe the
prerequisites, including the required libraries and repo clones, to run the
pytorch/xla benchmarks.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -10,21 +10,33 @@ Follow the [contributing
 guide](https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md) to set up
 the repo for development. Using the devcontainer is recommended.
 
-Clone the [`pytorch/benchmark`](https://github.com/pytorch/benchmark) repo into
-your workspace directory, which is mounted to the devcontainer at
-`/workspaces/torch`.
+Install the required libraries by the models.  Depending on the specific
+models, you may need to install `torchvision`, `torchaudio`, `torchtext`.
+
+Follow [this guide](https://pytorch.org/get-started/locally/) to install the
+`torchvision` and `torchaudio` version you need. Use the following command to
+install `torchtext`.
+
+```shell
+pip install torchtext
+```
+
+The following LibGL library is
+required by the Llava model.
+
+```shell
+apt install libgl1-mesa-glx
+```
+
+For the scripts to find the models to benchmark, clone the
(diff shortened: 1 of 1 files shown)


benchmarks/README.md
```diff
@@ -4,6 +4,27 @@ The two main benchmarking scripts are
   - `experiment_runner.py` to run benchmark experiments, and
   - `result_analyzer.py` to aggregate the benchmark result in CSV form.
 
+## Prerequisites
+
+Follow the [contributing
+guide](https://github.com/pytorch/xla/blob/master/CONTRIBUTING.md) to set up
+the repo for development. Using the devcontainer is recommended.
+
+Clone the [`pytorch/benchmark`](https://github.com/pytorch/benchmark) repo into
```
bhavya01, line 13: nit: I feel this line is not needed. Just in case, someone is using a different setup.

bhavya01 (approved):

ysiraichi:
Maybe we should also mention torchvision, torchaudio, torchtext. What do you think?

haifeng-jin:
> Maybe we should also mention torchvision, torchaudio, torchtext. What do you think?

I did not install these libraries as I run the benchmarks. I only installed torchvision. Are they required by the models?

benchmarks/README.md
haifeng-jin: I revised it to be more general.

haifeng-jin:
I added the instructions for installing `torchvision`, `torchaudio`, and `torchtext`.

PR #9231, opened 2025-05-21
Separate pallas test into a separate test shard

ghpvnist:
Reduces total TPU test time by 7 minutes, but TPU test is no longer a bottleneck. It's still good to add this optimization in case it becomes a bottleneck again.


test/tpu/run_pallas_test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -xue
+CDIR="$(cd "$(dirname "$0")" ; pwd -P)"
```
zhanyong-wan, line 3: Please follow the convention of starting an internal env var with a `_`. See `run_tests.sh` for examples.

zhanyong-wan (approved):
Thanks!

PR #9232, opened 2025-05-21
Modify NoneRemover to return modified list/tuple

ghpvnist:
When the type of `value_list` is a tuple, it cannot be modified in place. Instead, return a copy of the modified value if the type is a tuple.


qihqi (approved):
Looks good! Please add a unit test.

ghpvnist:
I utilized Gemini to generate the unit tests, and I'm pretty impressed by the coverage tests!

PR #9239, opened 2025-05-22
Update README.md with supported Python version

zzzwen:
As of 05/22 we have build for 3.8 to 3.11, newer version will need to rollback

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ To install PyTorch/XLA stable build in a new TPU VM:
 pip install torch==2.7.0 'torch_xla[tpu]==2.7.0'
 ```
 
+Note: As of 05/22/2025, we have build for Python 3.8 to 3.11, if you are on a newer version, please follow https://cloud.google.com/python/docs/setup and change the Python version
+
 To install PyTorch/XLA nightly build in a new TPU VM:
 
 ```sh


README.md
```diff
@@ -29,6 +29,8 @@ To install PyTorch/XLA stable build in a new TPU VM:
 pip install torch==2.7.0 'torch_xla[tpu]==2.7.0'
 ```
 
+Note: As of 05/22/2025, we have build for Python 3.8 to 3.11, if you are on a newer version, please follow https://cloud.google.com/python/docs/setup and change the Python version
```
zhanyong-wan, line 32: Can we move this before the `pip install` command above? We don't want people to try the installation first and fail. We want them to see this first so that they can avoid the failure.

zhanyong-wan, line 32: Instead of fanning out to another website, which is not tailored for our case and takes more effort to digest, can we spell out the exact commands needed here?

1. How to check whether the user's python version is supported.
2. How to install a supported python version without removing the previously installed version (e.g. using `uv`).
3. How to make the supported python version the default.
4. How to revert to defaulting to the user's previous python version (e.g. after they are done with the PTXLA experiement).

The idea is to make it as easy as possible for the users. Thanks!

zhanyong-wan (changes requested):
Thanks!

README.md
zzzwen: We can do that, if you feel strongly about it. 
And I would suggest to keep the README on torch-xla itself, and reduce a the amount of back ground information. As change python version can be done in multiple ways and just a google search away.

zhanyong-wan: I understand the desire to stay focused on PTXLA itself. The reason we need to diverge from that is the fact that PTXLA requires a workaround for some users, which is a bug. Once we fix the bug, we can revert the documentation so that we can be focused on PTXLA again. I.e. I think the diversion is a necessary temporary thing.

While the user can search up how to change the python version, every such annoyance will lose a percentage of the users. This step is easy for someone who already knows how to do it, but can be confusing to someone inexperienced - the search result may have several hits, so which one should I follow? The external documentation has a lot of details that don't apply to our case, so why should we force a user to go through them and identify the part that's relevant?

This may be but a small annoyance, but poor user experience is the result of many such annoyances.

README.md
```diff
@@ -24,8 +24,14 @@ started:
 ### TPU
 
 To install PyTorch/XLA stable build in a new TPU VM:
+Note: As of 05/29/2025, we have build for Python 3.8 to 3.11, please use one of the support versions
```
zhanyong-wan, line 27: "we have built PyTorch/XLA using Python 3.8-3.11. Please use one of the supported Python versions."

zhanyong-wan (approved):

PR #9249, opened 2025-05-27
[Kernel] support kv cache quantization in ragged attention kernel

yaochengji:


torch_xla/experimental/custom_kernel.py
```diff
@@ -927,6 +935,12 @@ def _ragged_paged_attention_nonkernel(
                                               head_dim)[:kv_len]
     v = kv_pages[indices, :, 1::2, :].reshape(-1, num_kv_heads,
                                               head_dim)[:kv_len]
+    if k_scale is not None:
+      k = k.to(torch.float32) * k_scale
```
vanbasten23, line 939: Do you know the reason why we need to do `k.to(torch.float32)`?

test/test_pallas.py
```diff
@@ -130,9 +131,10 @@ def _ragged_pagedattention_generate_qkv(
                                       (0, max_num_seqs - kv_lens.shape[0]),
                                       "constant", 0)
     q = torch.randn((max_num_batched_tokens, num_q_heads, head_dim),
```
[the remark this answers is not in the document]
vanbasten23: oh i see, just saw the comment `# Use float32 for randn because it doesn't support some dtypes like float8`, maybe move it before the `q`?

test/test_pallas.py
```diff
@@ -654,9 +657,16 @@ def _test_ragged_paged_attention(
...
         page_size,
         num_pages,
-        dtype,
+        q_dtype,
+        kv_dtype,
         max_num_batched_tokens=max_num_batched_tokens,
         max_num_seqs=max_num_seqs)
+    k_scale = 0.5 if kv_dtype in [torch.float8_e5m2] else None
+    v_scale = 0.5 if kv_dtype in [torch.float8_e5m2] else None
+    if num_heads[1] == 1 and kv_dtype in [torch.float8_e5m2]:
```
vanbasten23, line 666: nit: could you use a var num_kv_heads=num_heads[1] as it's not obvious

Also, could you remind me why attn kernel cannot support this case?

test/test_pallas.py
```diff
@@ -654,9 +657,16 @@ def _test_ragged_paged_attention(
...
+      self.skipTest("attention kernel cannot support ")
+    if kv_dtype is torch.float8_e5m2 and tpu.version() <= 4:
```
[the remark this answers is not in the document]
yaochengji: The kernel can support other types of fp8. But we only test float8_e5m2 here.

torch_xla/experimental/custom_kernel.py
vanbasten23: curious if it's necessary to make the multiplication  k*k_scale in fp32. If so, do you know the reasoning behind it?

vanbasten23 (approved):
Thanks Chengji. Looks good with one question for https://github.com/pytorch/xla/pull/9249/files#r2113078862.

Issue #9015, opened 2025-04-21
[torch-xla 2.7] "TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'" on python 3.9


jeffhataws:
## 🐛 Bug

When running on Amazon Linux 2023 python 3.9, I get an error "TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'" very early during imports. I narrowed down to the following imports:

```
(test_venv) [ec2-user@ip-10-3-144-37 ~]$ python -c "import torch_xla; from torch_xla.core import xla_builder, xla_model"
WARNING:root:Defaulting to PJRT_DEVICE=CPU
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/ec2-user/test_venv/lib64/python3.9/site-packages/torch_xla/core/xla_builder.py", line 7, in <module>
    from torch_xla.experimental.custom_kernel import _jax_env_context, jax_import_guard
  File "/home/ec2-user/test_venv/lib64/python3.9/site-packages/torch_xla/experimental/custom_kernel.py", line 1089, in <module>
    attn_logits_soft_cap: float | None = None,
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

I doesn't affect python 3.10/3.11 installations.

I don't know if it is reproducible on ubuntu yet.

## To Reproduce

1. Setup a python 3.9 virtual environment
2. `pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/cpu`
3. `pip3 install 'torch_xla @ https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-2.7.0rc5-cp39-cp39-linux_x86_64.whl' -f https://storage.googleapis.com/libtpu-releases/index.html`
4. `python -c "import torch_xla; from torch_xla.core import xla_builder, xla_model"`

## Expected behavior

No error

## Environment

 - Reproducible on XLA backend [CPU/TPU/CUDA]: CPU
 - torch_xla version: 2.7.0rc5

rpsilva-aws:
https://peps.python.org/pep-0604/, and I see many results with the pipe operator type hint (60 results):

<details>

<summary>GREP entries</summary>

```
grep -r "| None" --exclude-dir=build/ -n | while read -r line; do
    file=$(echo "$line" | cut -d: -f1)
    linenum=$(echo "$line" | cut -d: -f2)
    echo "=== $file:$linenum ==="
[...]

PR #9251, opened 2025-05-27
[Kernel] add group_offset and transpose_rhs support in gmm kernel

yaochengji:


yaochengji:
> Could you add some context about this PR? I understand this PR exposes group_offset and tranpose_rhs in the gmm, I wonder why the change is needed in the vLLM level

Or there will be transpose operations on the weight. https://github.com/vllm-project/vllm/blob/fced756923e3b65dad0e1f4da202c4d2f82d2a4b/vllm/model_executor/layers/fused_moe/moe_pallas.py#L71

rpsilva-aws:
FYI @yaochengji: https://github.com/pytorch/xla/issues/9015

I don't think we have any CI that tests with 3.9, but see if this would be an issue for your users.

yaochengji:
@rpsilva-aws thanks for reminding, the minimum python version of vLLM TPU users should be 3.10

PR #9257, opened 2025-05-28
Run `test_collective_permute.py` on TPU CI.

ysiraichi:
Since its creation, `test_collective_permute.py` only runs on either TPU or NEURON. However, at some point, TPU tests were moved somewhere else without this one. Meanwhile, this test was left running on both CPU and GPU CI, even though it doesn't actually tests anything because, again, it only does so on TPU or NEURON devices.

This PR removes `test_collective_permute.py` from CPU and GPU CI, and moves it to TPU CI.

### DIFF
--- a/test/tpu/run_tests.sh
+++ b/test/tpu/run_tests.sh
@@ -49,6 +49,7 @@ run_test "$_TEST_DIR/test_operations.py" -v
 run_test "$_TEST_DIR/test_xla_graph_execution.py" -v
 run_test "$_TEST_DIR/pjrt/test_runtime_tpu.py"
 run_test "$_TEST_DIR/pjrt/test_collective_ops_tpu.py"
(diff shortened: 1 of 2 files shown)


test/tpu/run_tests.sh
```diff
@@ -49,6 +49,7 @@ run_test "$_TEST_DIR/test_operations.py" -v
 run_test "$_TEST_DIR/test_xla_graph_execution.py" -v
 run_test "$_TEST_DIR/pjrt/test_runtime_tpu.py"
 run_test "$_TEST_DIR/pjrt/test_collective_ops_tpu.py"
+run_test "$_TEST_DIR/test_mp_collective_permute.py"
```
zhanyong-wan, line 52: Do we need to add this to neuron as well?

zhanyong-wan (approved):
Thanks!

rpsilva-aws (approved):

PR #9282, opened 2025-06-02
Do not clear UncachedCompile between graph executions

### LINKED ISSUE #9117: Get the number of graphs with dynamo for the models in benchmarks
## Context
We would like to learn more about the performance gaps between using `pytorch/xla` with `torch.compile(backend='openxla')` and using the `LazyTensor` tracing.

The main source of performance gap is the graph breaks, which has two sources. One is during torch dynamo fx tracing. The other is lowering that to HLO.

## Task

The first step is to compare the number of graph breaks between fx graphs and the HLO graphs.
We will decide the next moves based on the numbers.

haifeng-jin:
Resolves #9117 

The `UncachedCompile` metric was cleared at every fx graph compile and execution, when used with `torch.compile`.
This PR ensures the values are not cleared between the graphs.

### DIFF
--- a/test/dynamo/test_bridge.py
+++ b/test/dynamo/test_bridge.py
@@ -303,6 +303,20 @@ class TorchXLAReuseGraphTest(torch._dynamo.test_case.TestCase):
     t = torch.randint(0, 5, (3,), device=device)
     self._compile_and_check(foo, (t,))
 
+  def test_metrics_preserved(self):
+    metrics.clear_counters()
+    torch_xla._XLAC._xla_increment_counter('UncachedCompile', 3)
+    torch_xla._XLAC._xla_increment_counter('DynamoExtractCompiledGraph', 4)
+
+    model = BasicModule()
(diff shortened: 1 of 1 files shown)


haifeng-jin:
@ghpvnist Would you mind help explain how the following functions related to the number of fx or HLO graphs?
I am not very familiar with the `dyanmo_bridge`.
I just followed the existing code to make the changes.

Why do we port over the `DynamoExtractCompiledGraph` in `UnsupportedNodesCollector.run_node()` and incremented it by 1 in `extract_compiled_graph()`?

Thank you!

haifeng-jin:
As I further inspect the code. The `UncachedCompile` was correctly counted.
We do not need to increment it in `dynamo_bridge.py`.
We only need to prevent it from being cleared between the graph executions.

ghpvnist (approved):
LGTM! Can you add an unit test?

haifeng-jin:
Unit test added.

PR #9287, opened 2025-06-03
Add DLRM DCN v2 with JAX Sparsecore

chandrasekhard2:


qihqi (approved):
Please run `yapf -i -r *.py test/ torchax` before merging. Thanks

vlad-karp:
@qihqi , do we still want this model implementation to put here instead or in addition to torchprime?

PR #9289, opened 2025-06-03
[Kernel] add heuristic gmm block sizes choosing logic

yaochengji:


yaochengji:
The e2e vLLM inference throughput increased from 8.23 req/s to 11.99 req/s in my local benchmark.

torch_xla/experimental/custom_kernel.py
```diff
@@ -1306,46 +1306,114 @@ def repeat_with_fixed_output_size(input: torch.Tensor, repeats: torch.Tensor,
...
+
+    Returns:
+        tuple[int, int, int]: A tuple (tm, tk, tn)
+    """
+  # In average each expert has m // g tokens, but as it might be unbalanced,
+  # here we doubled the token size when choosing tiling size of m.
+  # TODO(Chengji): increase the upper limit tiling size of m when we can set
+  # the vmem size to be used for gmm kernel.
+  tm = _round_up_to_multiple_of_128_within_limit(2 * m // g, 512)
+  tm = min(tm, m)  # there's a requirement that m % tm == 0
```
vanbasten23, line 1366: Besides "there's a requirement that m % tm == 0", is it needed because after `_round_up_to_multiple_of_128_within_limit`, tm can be greater than m?

torch_xla/experimental/custom_kernel.py
```diff
@@ -1306,46 +1306,114 @@ def repeat_with_fixed_output_size(input: torch.Tensor, repeats: torch.Tensor,
   return res
 
 
+def _round_up_to_multiple_of_128_within_limit(x: int, limit: int) -> int:
+  """
+    Rounds the given integer `x` up to the nearest multiple of 128, without exceeding
+    the specified `limit`.
+
+    If `x` is less than or equal to 128, returns 128.
+    If `x` is less than `limit`, returns the smallest multiple of 128 greater than or
```
vanbasten23, line 1315: if x=129 and limit=130, then it would return 256 which greater than `limit` so the name `_round_up_to_multiple_of_128_within_limit` doesn't seem accurate?

torch_xla/experimental/custom_kernel.py
yaochengji: Yes

torch_xla/experimental/custom_kernel.py
yaochengji: limit must be multiply of 128.

torch_xla/experimental/custom_kernel.py
```diff
@@ -1306,46 +1306,114 @@ def repeat_with_fixed_output_size(input: torch.Tensor, repeats: torch.Tensor,
...
+        g (int): The number of experts.
+
+    Returns:
+        tuple[int, int, int]: A tuple (tm, tk, tn)
+    """
+  # In average each expert has m // g tokens, but as it might be unbalanced,
+  # here we doubled the token size when choosing tiling size of m.
+  # TODO(Chengji): increase the upper limit tiling size of m when we can set
+  # the vmem size to be used for gmm kernel.
+  tm = _round_up_to_multiple_of_128_within_limit(2 * m // g, 512)
```
vanbasten23, line 1365: Is `2 * m // g` usually greater or less than 512? Similar for k/n and 2048?

Also, could you explain the reasoning behind this heuristic?

yaochengji: This is to make sure that one single block can contain all the tokens of a single expert for most of the cases.

I also benchmarked it locally and find this heuristic logic is not bad.

torch_xla/experimental/custom_kernel.py
vanbasten23: I meant the function name "_round_up_to_multiple_of_128_within_limit" is not accurate because when x=129 and limit=130 the result will be greater than `limit`.

Also, when x>limit, the function actually rounds down. That's why I find it confusing. Can you come up with a better name?

torch_xla/experimental/custom_kernel.py
vanbasten23: For tm, tk, tn, I'm curious if it usually round up or down.
I guess for tm, `2 * m // g` is usually greater than 512, so we usually round down.
And for tk and tn, k/n is usually less than 2048 so we usually round up. 
Is my understanding correct?

vanbasten23: discussed offline:
- 2*m//g can be either greater or less than 512. If there are 32 tokens and topk=2, m=topk*num_tokens=64, in this case, 2*m//g will be less than 512.
- k/n correspond to n_input_features/n_output_features in the matmul so they are normally greater than 2048, unless the num shards is large.

vanbasten23 (approved):
Thanks Chengji!

PR #9292, opened 2025-06-04
Split CPU tests so the shards have a more even test time

ghpvnist:
Recent [CI run](https://github.com/pytorch/xla/actions/runs/15452994620) shows that CPU has the longest test at 26m 39s while the longest TPU test took 23m 59s. This saves ~2 minutes of CI time.

At this point, we're near the optimal CI build & test times when considering the balance between more shards (which has overhead time and limited TPU resources) vs reducing number of tests per shard (reduces test time).


zhanyong-wan (approved):
Thanks!

Issue #9268, opened 2025-05-30
set up pre-push auto formatting


zhanyong-wan:
## 🚀 Feature

After a one-time set-up, the C++/Python file formatters should automatically run whenever a developer runs `git push`.

## Motivation

It's tedious to remember to format the code before creating/updating a PR. The commandline for the formatting is also hard to remember. We should automate this.

PR #9293, opened 2025-06-04
Add a script to automatically format C++/Python files upon `git push`

zhanyong-wan:
Closes: https://github.com/pytorch/xla/issues/9268

With this PR, one can run
```
scripts/git_fix.py --set_git_push_hook
```
once to set up auto pre-push file formatting. I.e. afterwards, whenever one runs `git push`, the changed/added C++ and Python files will be automatically formatted. The developer no longer needs to remember the command line for formatting files.

One can also manually format the changed/added C++ and Python files by running
```
scripts/git_fix.py
```
directly.

### DIFF
--- a/scripts/git_fix.py
+++ b/scripts/git_fix.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+"""Fixes the format of locally changed git files.
+"""
+
+import logging
+import os
+import re
+import sys
+from typing import Optional
+
+# Root of the PyTorch/XLA repo.
+_PTXLA_DIR = os.path.abspath(os.path.dirname(__file__) + '/..')
+
+logger = logging.getLogger(__name__)
+
+
+def get_uncommitted_changed_added_files() -> set[str]:
+  """Gets the list of changed or added files that are not committed yet.
+
+  This includes untracked files (i.e. newly created files not added to git yet).
+  """
+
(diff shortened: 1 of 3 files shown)


scripts/git_fix.py
```diff
@@ -0,0 +1,213 @@
...
+        f'Are you sure you want to overwrite {_GIT_PRE_PUSH_HOOK_PATH}? [y/N] ')
+    if overwrite.lower() != 'y':
+      print('Skipping git pre-push hook setup.', file=sys.stderr)
+      return False
+
+  # Create a git pre-push hook to run this script.
+  with open(_GIT_PRE_PUSH_HOOK_PATH, 'w') as f:
+    f.write(f'''#!/bin/bash
+# This hook is automatically set by `{_SCRIPT_PATH } --set_git_push_hook`.)
+# We ignore any errors, as file formatting is best effort.
```
tengyifei, line 177: Should we abort the push operation from inside the hook if the files are formatted? Seems useful since otherwise the user might forget that they need to re-commit the formatted files and push again.

tengyifei (approved):

scripts/git_fix.py
zhanyong-wan: I decided not to block on formatting failures as the formatting could fail due to the user not having the proper tools installed. In this case, we should let the user push and rely on feedback from CI to format the files manually.

As a future improvement, we can make the script check if the formatter made any changes, and abort if yes so that the user will know they need to commit the changes.

Thanks!

PR #9294, opened 2025-06-05
Misc torchax improvements

qihqi:
1. Ability to disable views
2. Removed some asserts inside of ops: asserts that access the data of array are not jit compilable.
3. repeat interleave changes.

### DIFF
--- a/torchax/torchax/config.py
+++ b/torchax/torchax/config.py
@@ -8,6 +8,7 @@ class Configuration:
   debug_mixed_tensor: bool = False
   debug_print_each_op_operands: bool = False
   use_int32_for_index: bool = False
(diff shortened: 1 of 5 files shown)


torchax/torchax/config.py
```diff
@@ -7,7 +7,9 @@ class Configuration:
   debug_accuracy_for_each_op: bool = False
   debug_mixed_tensor: bool = False
   debug_print_each_op_operands: bool = False
+
   use_int32_for_index: bool = False
+  force_materialize_views: bool = False
```
tengyifei, line 14: Can you add a doc comment below explaining what this does? Convert PyTorch view objects to PyTorch tensors?

PR #9295, opened 2025-06-05
Deprecate XLA:CUDA: add warning on initialization.

ysiraichi:
This PR starts the XLA:CUDA deprecation process by issuing a warning whenever we initialize the CUDA PjRt client.

```python
>>> torch_xla.device()
2025-06-05 15:31:31.225748: W torch_xla/csrc/runtime/pjrt_registry.cpp:151] The XLA:CUDA device is deprecated in release 2.8. Future releases might remove XLA:CUDA support entirely. Use the PyTorch native CUDA backend, instead.
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1749137491.307757   53448 se_gpu_pjrt_client.cc:1101] Using BFC allocator.
I0000 00:00:1749137491.307890   53448 gpu_helpers.cc:136] XLA backend allocating 12632997888 bytes on device 0 for BFCAllocator.
I0000 00:00:1749137491.308026   53448 gpu_helpers.cc:177] XLA backend will use up to 4210999296 bytes on device 0 for CollectiveBFCAllocator.
I0000 00:00:1749137491.319481   53448 cuda_dnn.cc:529] Loaded cuDNN version 90000
device(type='xla', index=0)
```

### DIFF
--- a/torch_xla/csrc/runtime/pjrt_registry.cpp
+++ b/torch_xla/csrc/runtime/pjrt_registry.cpp
@@ -148,6 +148,10 @@ InitializePjRt(const std::string& device_type) {
   } else if (device_type == "TPU_LEGACY") {
     XLA_ERROR() << "TPU_LEGACY client is no longer available.";
   } else if (device_type == "CUDA") {
+    TF_LOG(WARNING)
+        << "The XLA:CUDA device is deprecated in release 2.8. "
+        << "Future releases might remove XLA:CUDA support entirely. "
+        << "Use the PyTorch native CUDA backend, instead.";
     TF_VLOG(1) << "Initializing PjRt GPU client...";
     bool async = sys_util::GetEnvBool(env::kEnvPjrtAsyncGpuClient, true);
     int local_process_rank = sys_util::GetEnvInt(env::kEnvPjRtLocalRank, 0);


torch_xla/csrc/runtime/pjrt_registry.cpp
```diff
@@ -148,6 +148,10 @@ InitializePjRt(const std::string& device_type) {
   } else if (device_type == "TPU_LEGACY") {
     XLA_ERROR() << "TPU_LEGACY client is no longer available.";
   } else if (device_type == "CUDA") {
+    TF_LOG(WARNING)
```
zhanyong-wan, line 151: Thanks for the quick action!

Not sure if C++ warning logs are always visible to python users. Can we follow up with a python warning as well?

yaoshiang: Agree. The pattern PyTorch established is here - we should follow it if possible for consistent deprecation warnings. 

https://github.com/pytorch/pytorch/blob/134179474539648ba7dee1317959529fbd0e7f89/torch/testing/_comparison.py#L1522

zhanyong-wan: Thanks for the pointer, @yaoshiang !

The example is for deprecating an entire function, and I believe the warning will show up in the IDE or build log when the user uses that function. In our case, though, we are not deprecating any function - we are deprecating the use of a particular parameter value (`'CUDA'`) for some function. Therefore I think we need a different mechanism, one that prints a warning at run time (as opposed to build time). Do you know of such examples?

ysiraichi: The way PyTorch solves this is by using a `torch::PyWarningHandler` RAII combined with `TORCH_WARN`. In other words, we would have to do 2 things:

- Call `TORCH_WARN("...")` macro
- Create a `torch::PyWarningHandler` variable at each Python API entry-point available in _init_python_bindings.cpp_

What do you think?

ysiraichi: My suggestion is that we leave this warning as is for this PR, and work on making that into a Python warning in a separate PR.

zhanyong-wan: Agreed on the incremental approach. Thanks for illustrating how it can be done, Yaoshiang!

yaoshiang: Thanks for the clarification Zhanyong! Here is what I found re: deprecated args, but not the entire function. 

https://github.com/pytorch/pytorch/blob/134179474539648ba7dee1317959529fbd0e7f89/torch/nn/modules/module.py#L2201

jrfonseca:
@ysiraichi , I presume this deprecation of XLA:CUDA applies just to Torch -> XLA -> CUDA PjRt, and there are no plans to deprecate JAX -> XLA -> CUDA PjRt, right?

Does this effectively mean that the ability of having [Custom GPU Kernels via Triton](https://github.com/pytorch/xla/blob/37e5fd242f38b57fdded624980cc5914342cfa30/docs/triton.md) is consequently also deprecated given it's specific to CUDA plugin?

I ask because just yesterday I was playing with XLA + Triton and noticed that [`test_gpu_custom_call_triton_flash_attention`](https://github.com/pytorch/xla/blob/37e5fd242f38b57fdded624980cc5914342cfa30/test/test_triton.py#L269-L271) is not testing effectively, as it uses `torch.empty` instead of `torch.rand` for the input q/k/v tensors, and the tests fails if one uses the latter.

zhanyong-wan:
@bhavya01 , would you be able to take the questions? Thanks.

bhavya01:
> @ysiraichi , I presume this deprecation of XLA:CUDA applies just to Torch -> XLA -> CUDA PjRt, and there are no plans to deprecate JAX -> XLA -> CUDA PjRt, right?
> 
> Does this effectively mean that the ability of having [Custom GPU Kernels via Triton](https://github.com/pytorch/xla/blob/37e5fd242f38b57fdded624980cc5914342cfa30/docs/triton.md) is consequently also deprecated given it's specific to CUDA plugin?
> 
> I ask because just yesterday I was playing with XLA + Triton and noticed that [`test_gpu_custom_call_triton_flash_attention`](https://github.com/pytorch/xla/blob/37e5fd242f38b57fdded624980cc5914342cfa30/test/test_triton.py#L269-L271) is not testing effectively, as it uses `torch.empty` instead of `torch.rand` for the input q/k/v tensors, and the tests fails if one uses the latter.

@jrfonseca Yes, this deprecation just applies to torch->XLA -> Cuda PjRt. We don't own the JAX stack. Yes, support of custom triton kernels via PyTorch/XLA is also deprecated

PR #9298, opened 2025-06-05
Merge cpp_tests1 and cpp_tests2 into single shard

ghpvnist:
The extra split is not necessary since the combined test time is shorter than the longest running CPU tests.

### DIFF
--- a/.circleci/common.sh
+++ b/.circleci/common.sh
@@ -202,7 +202,7 @@ function run_torch_xla_tests() {
   export PYTORCH_TESTING_DEVICE_ONLY_FOR="xla"
   export CXX_ABI=$(python -c "import torch;print(int(torch._C._GLIBCXX_USE_CXX11_ABI))")
 
-  if [[ -z "$RUN_BENCHMARK_TESTS" && -z "$RUN_CPP_TESTS1" && -z "$RUN_CPP_TESTS2" && -z "$RUN_PYTHON_TESTS" ]]; then
+  if [[ -z "$RUN_BENCHMARK_TESTS" && -z "$RUN_CPP_TESTS1" && -z "$RUN_PYTHON_TESTS" ]]; then
     run_torch_xla_python_tests $PYTORCH_DIR $XLA_DIR $USE_COVERAGE
     run_torch_xla_cpp_tests $PYTORCH_DIR $XLA_DIR $USE_COVERAGE
     run_torch_xla_benchmark_tests $XLA_DIR
--- a/.github/scripts/run_tests.sh
+++ b/.github/scripts/run_tests.sh
@@ -44,10 +44,10 @@ function run_torch_xla_cpp_tests() {
                "test_aten_xla_tensor_2"
                "test_aten_xla_tensor_3"
                "test_aten_xla_tensor_4"
-               "pjrt_computation_client_test")
+               "pjrt_computation_client_test"
                # Disable IFRT test as it currently crashes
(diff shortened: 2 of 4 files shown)


ghpvnist:
Since I'm removing a required CI test, the obsolete CI `CPU tests / test (cpp_tests, cpp_tests2)` will hang. So the PR needs manual force merge.

.circleci/common.sh
```diff
@@ -202,7 +202,7 @@ function run_torch_xla_tests() {
   export PYTORCH_TESTING_DEVICE_ONLY_FOR="xla"
   export CXX_ABI=$(python -c "import torch;print(int(torch._C._GLIBCXX_USE_CXX11_ABI))")
 
-  if [[ -z "$RUN_BENCHMARK_TESTS" && -z "$RUN_CPP_TESTS1" && -z "$RUN_CPP_TESTS2" && -z "$RUN_PYTHON_TESTS" ]]; then
+  if [[ -z "$RUN_BENCHMARK_TESTS" && -z "$RUN_CPP_TESTS1" && -z "$RUN_PYTHON_TESTS" ]]; then
```
zhanyong-wan: Nit: call it `RUN_CPP_TESTS` as we only have one C++ test suite one?

.github/scripts/run_tests.sh
```diff
@@ -44,10 +44,10 @@ function run_torch_xla_cpp_tests() {
                "test_aten_xla_tensor_2"
```
zhanyong-wan, line 44: Nit: rename `test_names1` to `test_names`?

.github/scripts/run_tests.sh
```diff
@@ -58,10 +58,6 @@ function run_torch_xla_cpp_tests() {
                "test_xla_sharding")
   if [[ "$RUN_CPP_TESTS1" == "cpp_tests1" ]]; then
```
zhanyong-wan: Nit: `cpp_tests`?

.github/scripts/run_tests.sh
```diff
@@ -58,10 +58,6 @@ function run_torch_xla_cpp_tests() {
...
     test_names=("${test_names1[@]}")
-  elif [[ "$RUN_CPP_TESTS2" == "cpp_tests2" ]]; then
-    test_names=("${test_names2[@]}")
-  else
```
zhanyong-wan, line 63: Are we sure we want to remove the default branch? This would change the existing behavior. It seems that we should just unconditionally run all tests (i.e. always set `test_names` to all tests).

.circleci/common.sh
ghpvnist: Done.

.github/scripts/run_tests.sh
ghpvnist: Done.

.github/scripts/run_tests.sh
ghpvnist: Done.

.github/scripts/run_tests.sh
ghpvnist: Ah yes, we should just test all now that we combined all tests, thanks.

zhanyong-wan (approved):
Thanks! Would you monitor the CI build to double check that we are not skipping any tests after the merge?

ghpvnist:
@zhanyong-wan All checks passed other than the two we're changing. Can you help force merge this?

PR #9304, opened 2025-06-06
Add an option to not use dlpack.

qihqi:
DLpack introduces Jax arrays with wierd Layout.

### DIFF
--- a/torchax/test/test_context.py
+++ b/torchax/test/test_context.py
@@ -63,18 +63,19 @@ class TestContext(unittest.TestCase):
 
   def test_jit_with_rng(self):
 
-    @xla_env
-    def random_op():
-      x = torch.randn(3, 3)
-      y = torch.randn(3, 3)
-      return x @ y
+    with xla_env:
 
-    random_jit = torchax.interop.jax_jit(random_op)
-    self.assertIsInstance(random_jit(), tensor.Tensor)
+      def random_op():
+        x = torch.randn(3, 3)
+        y = torch.randn(3, 3)
+        return x @ y
 
-    # Result always expected to be the same for a jitted function because seeds
-    # are baked in
-    torch.testing.assert_close(random_jit(), random_jit(), atol=0, rtol=0)
+      random_jit = torchax.interop.jax_jit(random_op)
+      self.assertIsInstance(random_jit(), tensor.Tensor)
+
+      # Result always expected to be the same for a jitted function because seeds
+      # are baked in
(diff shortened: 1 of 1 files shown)


torchax/test/test_context.py
```diff
@@ -76,20 +74,15 @@ def random_op():
...
 
   def test_generator_seed(self):
     with xla_env:
       x = torch.randn(2, 3, generator=torch.Generator().manual_seed(0))
       y = torch.randn(2, 3, generator=torch.Generator().manual_seed(0))
 
-    # Values will be different, but still check device, layout, dtype, etc
-    torch.testing.assert_close(
-        torchax.tensor.j2t(x._elem), torchax.tensor.j2t(y._elem))
+      # Values will be different, but still check device, layout, dtype, etc
```
tengyifei: nit: Values will be the same in this case?

torchax/torchax/tensor.py
```diff
@@ -591,11 +574,20 @@ def to_tensor(x):
     return res
 
   def j2t_iso(self, jaxarray):
```
tengyifei, line 609: Technically not this PR, but would you like to document `j2t_iso` and `j2t_copy`? It seems that `j2t_iso` gives you a view, while `j2t_copy` gives you a copy? But then all Jax arrays are immutable so I'm not sure what value does `j2t_copy` provide on top of `j2t_iso`. Would be good to clarify to the user.

torchax/torchax/ops/mappings.py
```diff
@@ -37,15 +42,29 @@ def t2j(t):
...
+      try:
+        dl = jaxdl.to_dlpack(x)
+        res = torchdl.from_dlpack(dl)
+      except Exception:
+        res = None
+
+    orig_dtype = None
+    if res is None:
+      orig_dtype = None
+      if x.dtype == jnp.bfloat16.dtype:
```
tengyifei, line 58: Question: why do we have to special case for bf16?

tengyifei (changes requested):

torchax/test/test_context.py
qihqi: Right. The comment is wrong. Removed.

tengyifei (approved):

torchax/torchax/ops/mappings.py
qihqi: Because numpy doesnt have bf16.

torchax/torchax/tensor.py
qihqi: done.

PR #9311, opened 2025-06-09
Support trace_me and xp.Trace in assume_pure

tengyifei:
This way the HLO staged out by `@assume_pure` and torchax will contain name scopes pushed by `xp.Trace`.

Also reduce the flakiness of `test/test_profiler.py`.

### DIFF
--- a/torch_xla/debug/profiler.py
+++ b/torch_xla/debug/profiler.py
@@ -107,12 +107,8 @@ class Trace(torch_xla._XLAC.profiler.TraceMe):
 
   The traces generated can then be collected using the above profiling APIs.
   The profiling server first needs to be started up and then can be sampled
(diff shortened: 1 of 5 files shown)


torch_xla/debug/profiler.py
```diff
@@ -132,7 +128,13 @@ def __enter__(self):
     self.scope = torch_xla._XLAC.profiler.scope_pusher(self.name)
     super().__enter__()
 
+    # Also enter the JAX named scope, to support torchax lowering.
+    import jax
```
qihqi, line 132: do we need try-catch around the import?

tengyifei: We're always installing jax and torchax all the time so it shouldn't be needed anymore!

PR #9312, opened 2025-06-09
enable_python_dispatcher in some xla custom passes.

laithsakka:
Address failure in 
https://github.com/pytorch/pytorch/pulls/laithsakka


laithsakka:
@qihqi thanks for accepting we can merge seems good no errors.

PR #9313, opened 2025-06-09
Fix basic_training.py

zzzwen:
loss needs grad

```
  File "/Users/wzz/work/torch-xla/torchax/test/.venv/lib/python3.12/site-packages/torch/autograd/graph.py", line 823, in _engine_run_backward
    return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
```

### DIFF
--- a/torchax/examples/basic_training.py
+++ b/torchax/examples/basic_training.py
@@ -17,6 +17,12 @@ import torchvision.transforms as transforms
 # NOTE: add these lines to make it run on TPUs!
 import torchax
 
+import torch.nn as nn
+import torch.nn.functional as F
+
+import matplotlib.pyplot as plt
+import numpy as np
+
 torchax.enable_globally()
 
 transform = transforms.Compose(
@@ -43,9 +49,6 @@ classes = ('T-shirt/top', 'Trouser', 'Pullover', 'Dress', 'Coat', 'Sandal',
 print('Training set has {} instances'.format(len(training_set)))
 print('Validation set has {} instances'.format(len(validation_set)))
 
-import matplotlib.pyplot as plt
-import numpy as np
-
 
 # Helper function for inline image display
 def matplotlib_imshow(img, one_channel=False):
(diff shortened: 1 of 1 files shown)


torchax/examples/basic_training.py
```diff
@@ -133,6 +133,9 @@ def train_one_epoch(epoch_index, tb_writer=None):
 
     # Compute the loss and its gradients
     loss = loss_fn(outputs, labels)
+    # Manually set requires_grad to True
+    loss.requires_grad_(True)
```
yaoshiang: why is this necessary when it's not necessary in standard PT?

yaoshiang (changes requested):

torchax/examples/basic_training.py
zzzwen: Update the code, it is the output from the jax model that has requires_grad == False.
@qihqi, it will take me a while to figure out what's going on here, do you happen to know.

yaoshiang:
At first blush, I'd assume this is a workaround for a bug in the internals. The requires_grad should be on for most of the model weights, triggering a grad on the loss value. Setting it after the criterion(expected, actual) already populates the grads. loss.requires_grad at that point should have no meaning. If this code change triggers behavior changes, I think there is something deeper at work here.

torchax/examples/basic_training.py
qihqi: Yeah likely a bug.

qihqi:
> At first blush, I'd assume this is a workaround for a bug in the internals. The requires_grad should be on for most of the model weights, triggering a grad on the loss value. Setting it after the criterion(expected, actual) already populates the grads. loss.requires_grad at that point should have no meaning. If this code change triggers behavior changes, I think there is something deeper at work here.

Yeah, likely the requires_grad (which should be propagated automatically) did not for some reason. We can fix that in a separate PR.

zzzwen:
This wont work, the problem turn out to be a lot larger. the TL;DR is non of the torchax tensor has requires_grad, therefore autograd will not backprop any of these. 
Let' use the https://github.com/pytorch/xla/issues/9317 to track it

PR #9323, opened 2025-06-10
Split _ops and _decomps.

qihqi:
We use _ops as jax implementation of a aten op of torch function. And we use _decomps as aten op implementation written in terms of other aten ops.

Splitting this has implication on the order of op resolution:

For an ATen op (OpOverload):
1. if the same op itself is found, then it is used
2. if not, we lookup by `op.overloadpacket` for a catch all. 
decomps often registered for op overload, so 1. will find a decomp despite there is a jax impl for the overloadpacket.

So previously, many jax impl are not used at all. So we are not testing those.
This change make it prefer jax impl; and only if the op and overload packet are not found we go through decomposition.

### DIFF
--- a/torchax/test_dist/test_distributed.py
+++ b/torchax/test_dist/test_distributed.py
@@ -14,6 +14,8 @@ import torchax.distributed
 # TODO(wcromar): do something useful with group name
 GROUP_NAME = "process_group"
 
+torchax.enable_globally()
+
 
 @pytest.fixture(scope="module")
 def multi_cpu():
@@ -77,16 +79,17 @@ def test_all_gather_tensor_func(multi_cpu, process_group):
     ],
 )
 def test_all_reduce(op, expected, multi_cpu, process_group):
-  device_count = multi_cpu
+    device_count = multi_cpu
 
-  def f(index):
-    dist.all_reduce(index, op)
-    return index
+    def f(index):
+      with torchax.default_env():
+        dist.all_reduce(index, op)
+        return index
 
-  res = torchax.distributed.spawn(f)
+    res = torchax.distributed.spawn(f)
 
-  expected_tensors = [expected for _ in range(device_count)]
-  np.testing.assert_equal(res.numpy(), expected_tensors)
+    expected_tensors = [expected for _ in range(device_count)]
+    np.testing.assert_equal(res.numpy(), expected_tensors)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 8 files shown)


torchax/test_dist/test_distributed.py
```diff
@@ -80,8 +82,9 @@ def test_all_reduce(op, expected, multi_cpu, process_group):
   device_count = multi_cpu
 
   def f(index):
-    dist.all_reduce(index, op)
-    return index
+    with torchax.default_env():
```
tengyifei, line 85: Do we need to enable the env twice?

torchax/torchax/ops/jaten.py
```diff
@@ -16,67 +16,13 @@
 from torchax import interop
 from torchax.ops import jax_reimplement
 from torchax.view import View
-from torchax.tensor import Tensor
 # Keys are OpOverload, value is a callable that takes
 # Tensor
 all_ops = {}
 
 # list all Aten ops from pytorch that does mutation
```
tengyifei: Is this comment still relevant?

torchax/torchax/tensor.py
```diff
@@ -122,14 +123,13 @@ def type_as(self, other):
...
-        env = arg._env
-        break
-
-    with env:
-      return func(*args, **(kwargs or {}))
+    # TODO(hanq): figure out why is dispatch mode not sufficient
+    if func == torch.ops._c10d_functional.wait_tensor.default:
+      return args[0]._env.dispatch(func, types, args, kwargs)
+    raise AssertionError(
+        'torchax Tensors can only do math within torchax environment.'
```
tengyifei: nit: "torchax Tensors can only do math within a torchax environment".

torchax/torchax/tensor.py
```diff
@@ -122,14 +123,13 @@ def type_as(self, other):
...
+        'Please wrap your code with `with torchax.defautl_env()` or '
```
tengyifei: nit: `torchax.default_env()`

torchax/torchax/view.py
```diff
@@ -332,17 +332,7 @@ def __torch_dispatch__(
...
-    env = None
-    for arg in torch_pytree.arg_tree_leaves(*args, **(kwargs or {})):
-      if isinstance(arg, Tensor) or isinstance(arg, View):
-        env = arg._env
-        break
-
-    with env:
-      res = func(*args, **(kwargs or {}))
-      return res
+    assert False
```
tengyifei: Should we throw a more meaning error if you intend this code path to be unreachable?

tengyifei (approved):

torchax/test_dist/test_distributed.py
qihqi: Yeah, because this test actually uses multi-process. So the spawned process need to enable again.

torchax/torchax/ops/jaten.py
qihqi: removed.

Issue #9252, opened 2025-05-27
Migrate `torch_xla.device()` to `torch.device('xla')`


ghpvnist:
The PyTorch world is more familiar with the latter and the two are basically equivalent. Instead of reinventing, let's reuse what everyone else is familiar with.

PR #9324, opened 2025-06-10
Migrate `.to(torch_xla.device())` to `.to('xla')`

ghpvnist:
Related to https://github.com/pytorch/xla/issues/9252

### DIFF
--- a/API_GUIDE.md
+++ b/API_GUIDE.md
@@ -47,7 +47,7 @@ Or used with neural network modules:
 
 ```python
 l_in = torch.randn(10, device='xla')
-linear = torch.nn.Linear(10, 20).to("xla")
+linear = torch.nn.Linear(10, 20).to('xla')
 l_out = linear(l_in)
 print(l_out)
 ```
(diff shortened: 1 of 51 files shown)


yaoshiang (approved):
Thanks so much! 

FYI ruff wants you to do double quotes, but the bigger rule is, "keep it consistent". If the rest of the file / repo uses single, that's fine. Just a potential nit.

ghpvnist:
Is that rule built-in into our yapf? I can use double quotes instead if that's preferred. I actually prefer double quotes myself, but found more single quotes than double quotes.

yaoshiang:
> Is that rule built-in into our yapf? I can use double quotes instead if that's preferred. I actually prefer double quotes myself, but found more single quotes than double quotes.

I don't know TBH. torchprime lints with yapf so I'm used to it... but as I look at the source here, I see single, so ignore my comment... sorry for the confusion.

tengyifei (approved):

PR #9333, opened 2025-06-10
Surface CUDA deprecation warning.

ysiraichi:
Follow-up from: #9295

This PR surfaces the warning issued when PjRt is initialized with the CUDA device to Python. In order to do that, we apply the following changes:

- Replace `TF_LOG(WARNING)` with `TORCH_WARN()`
- Create a `torch::PyWarningHandler` variable in the beginning of each Python API function

**Rationale:** these changes are necessary so as to decouple the PjRt part of the codebase depend on Python libraries. In summary, `torch::PyWarningHandler` is an RAII that will issue the Python warning on its destructor.


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -76,6 +78,17 @@
...
+/**
+ * RAII for emitting Python warnings.
+ *
+ * This macro should be used at every Python API entry-point, so that internal
+ * `TORCH_WARN()` uses are translated to Python warnings.
+ *
+ * Even though it's a PyTorch dependency, it doesn't rely on PyTorch specific
+ * code, i.e. it's possible to implement it ourselves in this project.
+ */
+#define TORCH_XLA_PYTHON_HEADER torch::PyWarningHandler ___handler;
```
zhanyong-wan: Identifiers with `__` in the name are reserved for the C++ compiler and standard libraries.

zhanyong-wan: Style: macros should only be used as the last resort. We don't need to define a macro here, especially since the macro's syntax doesn't look like a statement.

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -926,6 +939,7 @@ void BuildProfilerSubmodule(py::module* m) {
   profiler.def(
       "start_server",
       [](int port) -> std::unique_ptr<runtime::profiler::ProfilerServer> {
+        TORCH_XLA_PYTHON_HEADER
```
zhanyong-wan: It's tedious and error-prone to do this for each function we bind. Let's define a helper `BindPythonFunction(module, name, func)` and define the warning handler inside the implementation of the helper, so that we can omit this for all function bindings.

torch_xla/csrc/runtime/pjrt_registry.cpp
```diff
@@ -17,6 +17,8 @@
 #include "xla/pjrt/pjrt_c_api_client.h"
 #include "xla/pjrt/tfrt_cpu_pjrt_client.h"
 
+#include <c10/util/Exception.h>
```
zhanyong-wan: Style: per https://google.github.io/styleguide/cppguide.html, this should be before line 3.

test/test_operations.py
```diff
@@ -55,6 +55,7 @@
 import torch_xla.core.functions as xf
 import torch_xla.debug.profiler as xp
 import unittest
+import warnings
```
zhanyong-wan: Why this change?

test/pjrt/test_cuda.py
```diff
@@ -0,0 +1,26 @@
+import os
+import torch_xla
+import unittest
+import warnings
+
+
+def onlyOnCUDA(fn):
```
zhanyong-wan: We already removed the CUDA test job from the CI, so I don't think this test will be run?

I think it's fine to skip the test.

test/test_operations.py
ysiraichi: I moved the test from this file. Forgot to remove it.

test/pjrt/test_cuda.py
ysiraichi: Ah, you're right! Forgot about that.

bhavya01:
I don't have any other comments. Please wait for Zhanyong's approval

torch_xla/csrc/init_python_bindings.cpp
ysiraichi: I have updated the PR. Let me know what you think.

PR #9338, opened 2025-06-11
Switch XLA_CHECKs to ABSL_CHECKs in lowering_context.

zhanyong-wan:


torch_xla/csrc/ir.h
```diff
@@ -100,7 +100,7 @@ class XlaNode : public torch::lazy::Node {
   XlaNode(torch::lazy::OpKind op, xla::Shape xla_shape, size_t num_outputs,
           torch::lazy::hash_t hash_seed);
 
-  virtual ~XlaNode();
+  ~XlaNode() override;
```
ysiraichi, line 103: Why remove `virtual` keyword?

torch_xla/csrc/lowering_context.cpp
```diff
@@ -2,66 +2,72 @@
...
-#include "torch_xla/csrc/unwrap_data.h"
 
 namespace torch_xla {
 
 namespace {
 
 class HloMetadataSetter {
  public:
-  HloMetadataSetter(LoweringContext* loctx, const torch::lazy::Node* node) {
+  HloMetadataSetter(LoweringContext* const low_ctx,
```
ysiraichi: Since we are at it, why don't we make this also a reference: `LoweringContext& low_ctx`? And, maybe, call it `lowering_context`?

torch_xla/csrc/lowering_context.cpp
```diff
@@ -78,66 +84,65 @@ class HloMetadataSetter {
...
+    const c10::ArrayRef<const torch::lazy::Node*> post_order,
     torch::lazy::Util::EmissionMap emit_status)
-    : torch::lazy::LoweringContext(name, device, {}, emit_status),
+    : torch::lazy::LoweringContext(name, std::move(device), {},
+                                   std::move(emit_status)),
       builder_(name),
       stack_frame_index_builder_(std::make_shared<StackFrameIndexBuilder>()) {
-  for (auto node : post_order) {
-    LowerNode(node);
+  for (const auto* node : post_order) {
```
ysiraichi, line 112: Isn't the `const` inferred by `auto`, here?

torch_xla/csrc/lowering_context.cpp
```diff
@@ -208,72 +214,71 @@ absl::StatusOr<xla::XlaComputation> LoweringContext::BuildXla(xla::XlaOp root) {
 }
 
 void LoweringContext::AssignOutputOp(const torch::lazy::Output& output,
-                                     xla::XlaOp op) {
-  emitted_outputs_[output] = std::move(op);
+                                     const xla::XlaOp op) {
+  emitted_outputs_[output] = op;
```
ysiraichi, line 218: Why are we removing `std::move()`?

torch_xla/csrc/lowering_context.cpp
```diff
@@ -208,72 +214,71 @@ absl::StatusOr<xla::XlaComputation> LoweringContext::BuildXla(xla::XlaOp root) {
...
 xla::XlaOp LoweringContext::GetOutputOp(const torch::lazy::Output& output) {
   auto it = emitted_outputs_.find(output);
 
   if (it == emitted_outputs_.end()) {
-    auto post_order =
+    const auto post_order =
         torch::lazy::Util::ComputePostOrder(output.node, &emit_status_);
-    for (auto node : post_order) {
-      LowerNode(node);
+    for (const auto* const node : post_order) {
```
ysiraichi, line 227: Isn't the first `const` is inferred by `auto`?

torch_xla/csrc/lowering_context.cpp
```diff
@@ -2,66 +2,72 @@
 
 #include <torch/csrc/lazy/core/ir_metadata.h>
 
-#include <iostream>
 #include <optional>
 #include <sstream>
 #include <stdexcept>
-#include <string_view>
+#include <utility>
```
ysiraichi, line 8: Why do we need this include?

zhanyong-wan:
> Overall, looks good to me. It is a lot of refactoring coming with the replacement. I think it would be better and easier to review if you split this PR into 2, next time.

Sounds good!

torch_xla/csrc/ir.h
zhanyong-wan: This is driven by a clangd warning.

`override` implies `virtual`. It has the benefit of catching the bug where the base class's dtor is not virtual. So `override` is strictly better than `virtual`.

torch_xla/csrc/lowering_context.cpp
zhanyong-wan: For `std::move`.

torch_xla/csrc/lowering_context.cpp
zhanyong-wan: It's for readability: we don't want the reader to have to spend effort finding out whether the pointee is `const`.

torch_xla/csrc/lowering_context.cpp
zhanyong-wan: clangd asked to remove this as it's a trivially copyable type.

torch_xla/csrc/lowering_context.cpp
zhanyong-wan: For readability. We are optimizing for the reader, not the writer.

torch_xla/csrc/lowering_context.cpp
zhanyong-wan: Done!

ysiraichi (approved):
LGTM. Thank you.

Issue #9314, opened 2025-06-09
[torch-xla v2.8] Error "Check failed: state Expected an array shape." when running test/test_mp_reduce_scatter.py


jeffhataws:
## 🐛 Bug

With torch-xla v2.8, the Neuron team is getting "Check failed: state Expected an array shape." errors when running many training tests that uses reduce-scatter. These errors were not there in v2.7. Furthermore, I have narrowed it down to commit that updated openxla pin https://github.com/pytorch/xla/pull/9045 , because using the torch-xla nightly from 4/30 works.

I have also narrowed down the testcase to the existing test/test_mp_reduce_scatter.py as seen in the next section.

## To Reproduce

On a CPU instance:

```
python -m venv test_venv_pt2.8
source test_venv_pt2.8/bin/activate
pip3 install -U pip
pip3 install torch torchvision --pre --extra-index-url https://download.pytorch.org/whl/nightly/cpu
pip3 install 'torch_xla @ https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-2.8.0.dev-cp310-cp310-linux_x86_64.whl' -f https://storage.googleapis.com/libtpu-releases/index.html
cd ~/
git clone https://github.com/pytorch/xla
cd ~/xla/test
```

Add 'CPU' to the device list in test_mp_reduce_scatter.py:

```
   if xm.xla_device_hw(device) in ['TPU', 'CUDA', 'NEURON', 'CPU']:
```

then run

```
PJRT_DEVICE=CPU python test_mp_reduce_scatter.py
```

```
WARNING:root:MASTER_ADDR environment variable is not set, defaulting to localhost
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR                                                      
F0000 00:00:1749493093.448876   48267 shape.cc:166] Check failed: state Expected an array shape. Got (f32[32,2,32], f32[32,2,32], f32[32,2,32], f32[32,2,32], f32[32,2,32])
This is a programmer error. Please read the Shape object's array properties (e.g. dimensions) only when it's an array shape.
*** Check failure stack trace: ***                                                                                                                                           
    @     0x7ea70c886e99  absl::lts_20230802::log_internal::LogMessageFatal::~LogMessageFatal()
    @     0x7ea7017dc8c7  xla::Shape::array_state()
    @     0x7ea701a3b1ce  torch_xla::BuildReduceScatterCoalesced()
[...]

PR #9347, opened 2025-06-12
Fix usage of dimensions_size() to check for tuple

bfolie:
#9314


bhavya01:
Thanks! This so much more clear now.

PR #9351, opened 2025-06-13
Improve the API for binding python functions in C++

zhanyong-wan:
Replace `BindPythonFunction()` and friends with a `PythonScope` class template so that we can leverage the daisy chain syntax to simplify the call sites.

Instead of
```
py::module m = ...;
BindPythonFunction(m, "foo", ...);
BindPythonFunction(m, "bar", ...);
```
we can now write
```
PythonScope<py::module>(...)
    .def("foo", ...)
    .def("bar", ...);
```

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1057,12 +1057,21 @@ class PythonScope : public Scope {
 };
 
 void BuildProfilerSubmodule(py::module* m) {
-  // Define the profiler module.
+  // Define the profiler submodule.
   PythonScope<py::module> profiler =
       m->def_submodule("profiler", "Profiler integration");
+
+  // Define the profiler.ProfilerServer class.
   PythonScope<py::class_<runtime::profiler::ProfilerServer,
                          std::unique_ptr<runtime::profiler::ProfilerServer>>>
       profiler_server_class(profiler, "ProfilerServer");
+
+  // Define the profiler.ScopePusher class.
+  py::class_<torch::lazy::ScopePusher,
+             std::unique_ptr<torch::lazy::ScopePusher>>
+      scope_pusher_class(profiler, "ScopePusher");
+
+  // Define functions in the profiler submodule.
   profiler
       .def(
           "start_server",
@@ -1130,7 +1139,7 @@ void BuildProfilerSubmodule(py::module* m) {
            })
       .def_static("is_enabled", &tsl::profiler::TraceMe::Active);
 
-  // Define the profiler.TslProfilerSessionWrapper class.ß
+  // Define the profiler.TslProfilerSessionWrapper class.
   PythonScope<py::class_<
       runtime::profiler::TslProfilerSessionWrapper,
       std::unique_ptr<runtime::profiler::TslProfilerSessionWrapper>>>(
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -985,32 +985,54 @@ std::vector<bool> check_materialization_helper(
   return need_materialization;
 }
 
-// Wraps a py::module to provide more convenient APIs.
-class PythonModule {
+// Wraps a python scope (e.g. py::module) to provide more convenient APIs.
+// It behaves like a Scope object but has enhanced behaviors for the def()
+// method. This class has reference semantics, just like the Scope class.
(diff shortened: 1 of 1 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -986,99 +998,111 @@ std::vector<bool> check_materialization_helper(
...
+              const py::object& ex_value,
+              const py::object& traceback) -> py::object {
+             py::cast<xla::profiler::TraceMeWrapper*>(self)->Stop();
+             return py::none();
+           })
+      .def("set_metadata",
+           [](xla::profiler::TraceMeWrapper& self, const py::kwargs& kwargs) {
+             self.SetMetadata(kwargs);
+           })
+      .def_static("is_enabled", &tsl::profiler::TraceMe::Active);
```
ysiraichi, line 1085: Is this function being wrapped by the defined `PythonScope` lambda?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -84,69 +84,81 @@ namespace {
...
-          // passed to `TORCH_WARN()`, and emits a Python warning.
-          torch::PyWarningHandler handler;
-          return (*f)(args...);
-        },
-        std::forward<Extra>(extra)...);
+// Wraps a python scope (e.g. py::module) to provide more convenient APIs.
+// It behaves like a Scope object but has enhanced behaviors for the def*()
+// methods. This class has reference semantics, just like the Scope class.
+template <typename Scope>
+class PythonScope : public Scope {
```
ysiraichi, line 91: Maybe it would be safer to make `PythonScope` not inherit from `Scope`, since that might make us think we are using, for example, the `PythonScope::def_static` method, where we are actually using the `Scope::def_static` method.

Instead, we could have a `Scope` member variable, and forward the methods we don't wish to wrap, explicitly. What do you think?

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -986,99 +998,111 @@ std::vector<bool> check_materialization_helper(
...
+              throw py::error_already_set();
+            }
+          },
+          py::arg("service_addr"),              //
+          py::arg("logdir"),                    //
+          py::arg("duration_ms") = 1000,        //
+          py::arg("num_tracing_attempts") = 3,  //
+          py::arg("timeout_s") = 120,           //
+          py::arg("interval_s") = 5,            //
+          py::arg("options"))
```
ysiraichi, line 1059: Are these trailing comments intentional?

torch_xla/csrc/init_python_bindings.cpp
ysiraichi, line 91: Since we are re-writing this, what do you think of just wrapping the lambda functions (e.g. with a `LambdaWrapper` function), instead. For example:

```c++
py::class_<A>(m, "A")
    .def(py::init(LambdaWrapper([](...) { ... }))
    .def("run", LambdaWrapper([](...) { ... })
```

This would keep the `pybind11` types, while explicitly only changing the lambda bound to the Python API.

torch_xla/csrc/init_python_bindings.cpp
zhanyong-wan: It's not. This PR is meant as a pure refactoring. I can wrap `def_static` in the next PR.

torch_xla/csrc/init_python_bindings.cpp
zhanyong-wan: My intention is to make `PythonScope<S>` behave like `S`, except with some enhanced behavior. I think this reduces the cognitive burden.

There is no `PythonScope::def_static` defined, so the reader shouldn't think that `def_static` refers to `PythonScope::def_static`.

torch_xla/csrc/init_python_bindings.cpp
zhanyong-wan: Yes, it's a common trick to force `clang-format` to break the line.

torch_xla/csrc/init_python_bindings.cpp
zhanyong-wan: That would be verbose as we need to add `LambdaWrapper()` for each `def`. More importantly, it's unsafe as one can easily forget to add it for some `def`.

ysiraichi (approved):
LGTM.

PR #9354, opened 2025-06-13
Refactor call_jax to allow implementing ops in python

qihqi:
This PR introduces FlattenedInputFunc abstraction, which is an refactor of `jax_func_to_xla_computation` exposing the closures insides of it as methods.

FlattenedInputFunc itself splits a function into 3 components. (preprocess, flat_call, postprocess); among them, flat_call will take a single flattened list/tuple as input ans returns a flattened list as output. The way to think of it is to treat it as a triplet of functions, that are expected to be called in this exact order. The inner function has no restrictions, it can be a jax function, an function that takes and returns XlaOps or any other.

The above splitting is useful because compiling a function to xla will have to be a flat function.

Then, I abstracted the cache management to be something that any FlattenedInputFunc can share, as long as we know how to "especialize" or create shape specialized functions.

With this refactor, JaxCallable will replace previous capability on call_jax, and XlaCallable can call into Python functions using xla_builder API directly.

### DIFF
--- a/torch_xla/core/xla_builder.py
+++ b/torch_xla/core/xla_builder.py
@@ -830,6 +830,7 @@ def get_computation_hlo(computation):
 
 
 def xla_computation_as_func(computation, name=None):
+  """Converts an XlaComputation object to a callable that takes XLATensors."""
   name = name or "xla::computation"
   if not '::' in name:
     name = 'xla::' + name
@@ -938,25 +939,15 @@ class CompiledCallableWithCache(abc.ABC):
     self._flat_input_func = flat_input_func
 
   def __call__(self, *args, **kwargs):
-
-    def abstractify(a):  # make a pytree leaf abstract
-      if a is None:
-        return None
-      if isinstance(a, torch.Tensor):
-        assert a.device.type == 'xla', f"Inputs must be XLA tensors. Got {a.device}"
-        return jax.ShapeDtypeStruct(a.shape, tx.ops.mappings.t2j_dtype(a.dtype))
-      return a
-
     tensor_input_flattened = self._flat_input_func.preprocess(args, kwargs)
 
     abstract_inputs = tuple((a.shape, a.dtype) if a is not None else a
                             for a in tensor_input_flattened)
 
-    cache_key = hash((abstract_inputs, tuple(self._flat_input_func.non_tensors),
-                      self._flat_input_func.in_spec))
-
+    cache_key = (abstract_inputs, tuple(self._flat_input_func.non_tensors),
+                 self._flat_input_func.in_spec)
     if cache_key not in self._cache:
-      self._cache[cache_key] = self.especialize(tensor_input_flattened)
+      self._cache[cache_key] = self.specialize(tensor_input_flattened)
 
     flat_callable = self._cache[cache_key]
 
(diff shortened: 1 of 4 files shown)


torch_xla/core/xla_builder.py
```diff
@@ -888,91 +947,84 @@ def abstractify(a):  # make a pytree leaf abstract
...
+    if cache_key not in self._cache:
+      self._cache[cache_key] = self.especialize(tensor_input_flattened)
+
+    flat_callable = self._cache[cache_key]
+
+    output = flat_callable(tensor_input_flattened)
+    return self._flat_input_func.postprocess(output)
+
+  @abc.abstractmethod
+  def especialize(self, sample_flat_args):
```
tengyifei: nit: I think the right English word is `specialize`

torch_xla/core/xla_builder.py
```diff
@@ -828,57 +829,115 @@ def get_computation_hlo(computation):
   return torch_xla._XLAC._xla_computation_text(computation)
 
 
-class XlaComputation:
+def xla_computation_as_func(computation, name=None):
```
tengyifei, line 832: Could we document this function: """Converts an `XlaComputation` object to a callable."""

torch_xla/core/xla_builder.py
```diff
@@ -888,91 +947,84 @@ def abstractify(a):  # make a pytree leaf abstract
...
+      # Get a protobuf representation of the HLO. `as_serialized_hlo_module_proto` is
+      # mentioned at https://github.com/jax-ml/jax/discussions/22266
+      hlo_module = hlo_ir.as_serialized_hlo_module_proto()  # type: ignore
+      computation = computation_from_module_proto('jax_callable', hlo_module)
+      return xla_computation_as_func(computation, 'jax_func')
+
+
+class XlaCallable(CompiledCallableWithCache):
+  """
+
```
tengyifei: """XlaCallable lets you implement LazyTensor callables using the Python XlaBuilder API."""

torch_xla/core/xla_builder.py
```diff
@@ -888,91 +947,84 @@ def abstractify(a):  # make a pytree leaf abstract
...
+    math using xla builder python API above
+    """
+    super().__init__(FlattenedInputFunc(xla_func))
+
+  def especialize(self, sample_flat_args):
+    sample_args_shapes = tuple(
+        Shape.create(Op.from_torch_type(a.dtype), a.shape)
+        for a in sample_flat_args)
+
+    with xp.Trace('jax_to_xla_computation'):
```
tengyifei: nit: `'build_xla_computation'`

torch_xla/core/xla_builder.py
```diff
@@ -983,7 +1035,8 @@ def _jax_to_xla_computation_cache_elements() -> int:
 def call_jax(jax_func,
              args: tuple[Any, ...],
              kwargs: Optional[dict[str, Any]] = None,
-             name=None):
+             name=None,
+             override_hash=None):
```
tengyifei, line 1045: Can we document `override_hash` in the doc comments

test/test_assume_pure.py
```diff
@@ -13,6 +13,7 @@
 import torch_xla.runtime as xr
 from torch_xla.experimental.assume_pure import assume_pure
 from torch_xla._internal.jax_workarounds import jax_import_guard
+import jax.numpy as jnp
```
tengyifei: nit: unnecessary import?

tengyifei (approved):
Neat!

Issue #9315, opened 2025-06-09
[RFC] Improved coverage for native distributed collective operations


bfolie:
## 🚀 Feature
Improve coverage of the [PyTorch collective operations](https://docs.pytorch.org/docs/stable/distributed.html#synchronous-and-asynchronous-collective-operations) so that native distributed code is more likely to work without any user modification required. Collective ops would ideally work both in LazyTensor and in compiled mode, but many do not have an upstream Dynamo path.

The following are in scope.
* [All Gather Into Tensor](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.all_gather_into_tensor)
* [Broadcast](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.broadcast)
* [Scatter](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.scatter)
* [Reduce Scatter](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.reduce_scatter)
* [Send and Recv](https://docs.pytorch.org/docs/stable/distributed.html#point-to-point-communication)
* [Gather](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.gather)
* [Reduce](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.reduce)
* [All to All](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.all_to_all)

Collectives that operate on objects instead of tensors are out of scope.

### All Gather Into Tensor

Gathers tensors into a single output tensor. Like the already-implemented [all_gather](https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.reduce_scatter_tensor) but outputs a single tensor instead of a list of tensors. It appears in FSDP ([example](https://github.com/pytorch/pytorch/blob/v2.7.0/torch/distributed/fsdp/_optim_utils.py#L231)), and other places.

This is implemented by [wrapping all_gather](https://github.com/pytorch/xla/blob/v2.7.0/torch_xla/distributed/xla_backend.py#L86), so it fails in some cases because it has not been set up to support both stacking and concatenation of the input tensors. [This zip](https://github.com/pytorch/xla/blob/v2.7.0/torch_xla/distributed/xla_backend.py#L89), which is meant for a list of tensors, happens to work in some cases by splitting the tensors along the 0th dimension.

[...]

PR #9365, opened 2025-06-16
Support torch.distributed.scatter collective

bfolie:
#9315 

XLA doesn't have a Scatter op but we can put dummy tensor lists on the non-source rank and use reduce_scatter

### DIFF
--- a/test/pjrt/test_collective_ops_tpu.py
+++ b/test/pjrt/test_collective_ops_tpu.py
@@ -104,6 +104,25 @@ class TestXMCollectiveOpsTpu(parameterized.TestCase):
       np.testing.assert_array_equal(value, [-ordinal])
 
   @staticmethod
+  def _scatter():
+    dist.init_process_group("xla", init_method='xla://')
+    device = torch_xla.device()
+    world_size = xr.world_size()
+    if xr.global_ordinal() == 0:
+      tensors = [torch.tensor([i], device=device, dtype=torch.float) for i in range(world_size)]
+    else:
+      tensors = None
+    output_tensor = torch.tensor([-1], device=device)
+    dist.scatter(output_tensor, tensors, src=0)
+    torch_xla.sync()
+    return output_tensor.cpu()
+
+  def test_scatter(self):
+    results = pjrt.run_multiprocess(self._scatter)
+    for ordinal, value in results.items():
+      np.testing.assert_array_equal(value, [ordinal])
+
+  @staticmethod
   def _all_to_all(pin_layout):
     device = torch_xla.device()
     world_size = xr.world_size()
--- a/test/pjrt/test_collective_ops_tpu.py
+++ b/test/pjrt/test_collective_ops_tpu.py
@@ -109,12 +109,14 @@ class TestXMCollectiveOpsTpu(parameterized.TestCase):
     device = torch_xla.device()
     world_size = xr.world_size()
     if xr.global_ordinal() == 0:
-      tensors = [torch.tensor([i], device=device, dtype=torch.float) for i in range(world_size)]
+      tensors = [
+          torch.tensor([i], device=device, dtype=torch.float)
+          for i in range(world_size)
+      ]
     else:
       tensors = None
-    output_tensor = torch.tensor([-1], device=device)
+    output_tensor = torch.tensor([-1], dtype=torch.float, device=device)
     dist.scatter(output_tensor, tensors, src=0)
-    torch_xla.sync()
     return output_tensor.cpu()
 
   def test_scatter(self):
--- a/torch_xla/distributed/xla_backend.py
+++ b/torch_xla/distributed/xla_backend.py
@@ -5,7 +5,7 @@ import torch_xla.runtime as xr
 from torch_xla._internal import rendezvous
 import logging
 import os
-from torch._C._distributed_c10d import ProcessGroup
+from torch._C._distributed_c10d import ProcessGroup, ScatterOptions, ReduceScatterOptions
 
 
 def _create_xla_process_group(prefix_store, rank, size, timeout):
@@ -232,8 +232,17 @@ class ProcessGroupXla(ProcessGroup):
   def gather(self, *args):
     raise NotImplementedError
 
-  def scatter(self, *args):
-    raise NotImplementedError
+  def scatter(self, output_tensor_list: list[torch.Tensor], input_tensors_list: list[list[torch.Tensor]], opts: ScatterOptions):
+    output_tensor = output_tensor_list[0]
+    if xr.global_ordinal() == opts.rootRank:
+      input_tensors = input_tensors_list[0]
+    else:
+      input_tensors = [torch.zeros_like(output_tensor)] * xr.world_size()
+
+    rs_opts = ReduceScatterOptions()
+    rs_opts.reduceOp = dist.ReduceOp.SUM
+    return self.reduce_scatter([output_tensor], [input_tensors], rs_opts)
+
 
   # Dummy channel id maker. Different backend (TPU, GPU, etc) should replace
   # the maker with their specific one. See unit test in
--- a/torch_xla/distributed/xla_backend.py
+++ b/torch_xla/distributed/xla_backend.py
@@ -232,7 +232,9 @@ class ProcessGroupXla(ProcessGroup):
   def gather(self, *args):
     raise NotImplementedError
 
-  def scatter(self, output_tensor_list: list[torch.Tensor], input_tensors_list: list[list[torch.Tensor]], opts: ScatterOptions):
+  def scatter(self, output_tensor_list: list[torch.Tensor],
+              input_tensors_list: list[list[torch.Tensor]],
+              opts: ScatterOptions):
     output_tensor = output_tensor_list[0]
     if xr.global_ordinal() == opts.rootRank:
       input_tensors = input_tensors_list[0]
@@ -242,7 +244,6 @@ class ProcessGroupXla(ProcessGroup):
     rs_opts = ReduceScatterOptions()
     rs_opts.reduceOp = dist.ReduceOp.SUM
     return self.reduce_scatter([output_tensor], [input_tensors], rs_opts)
-
 
   # Dummy channel id maker. Different backend (TPU, GPU, etc) should replace
   # the maker with their specific one. See unit test in


test/test_torch_distributed_xla_backend.py
```diff
@@ -360,7 +360,6 @@ def test_barrier(self):
       'allreduce_coalesced',
       'alltoall',
       'gather',
-      'scatter',
```
bfolie, line 363: I'm not sure if there's a reason to add a `scatter` test to this file. The test in `test/pjrt/test_collective_ops_tpu.py` is more robust in that it tests the actual result. The tests in this file just check that the IR looks correct, which can be misleading (as was the case for send/recv).

torch_xla/distributed/xla_backend.py
```diff
@@ -253,8 +253,18 @@ def alltoall_base(self, output, input, output_split_sizes, input_split_sizes,
...
+              input_tensors_list: list[list[torch.Tensor]],
+              opts: ScatterOptions):
+    if xr.global_ordinal() == opts.rootRank:
+      inputs = input_tensors_list
+    else:
+      inputs = [[torch.zeros_like(output_tensor)] * xr.world_size()
+                for output_tensor in output_tensor_list]
+
+    rs_opts = ReduceScatterOptions()
+    rs_opts.reduceOp = dist.ReduceOp.SUM
```
pgmoka, line 270: Is this by chance superfluous? I see https://github.com/pytorch/pytorch/blob/bb462a6237c163774c99e01462703ebce55f4589/torch/csrc/distributed/c10d/Types.hpp#L156, but I am not sure if it has a direct relationship with https://github.com/pytorch/pytorch/blob/3106a33e410db9d1c2b0e47f9e329611d344a827/torch/_C/_distributed_c10d.pyi#L129.

torch_xla/distributed/xla_backend.py
```diff
@@ -253,8 +253,18 @@ def alltoall_base(self, output, input, output_split_sizes, input_split_sizes,
   def gather(self, *args):
     raise NotImplementedError
 
-  def scatter(self, *args):
-    raise NotImplementedError
+  def scatter(self, output_tensor_list: list[torch.Tensor],
```
pgmoka, line 260: Could you add a comment pointing to either an explanation of scatter, or add a little bit here?

torch_xla/distributed/xla_backend.py
```diff
@@ -253,8 +253,18 @@ def alltoall_base(self, output, input, output_split_sizes, input_split_sizes,
...
+              input_tensors_list: list[list[torch.Tensor]],
+              opts: ScatterOptions):
+    if xr.global_ordinal() == opts.rootRank:
+      inputs = input_tensors_list
+    else:
+      inputs = [[torch.zeros_like(output_tensor)] * xr.world_size()
+                for output_tensor in output_tensor_list]
```
pgmoka, line 267: Learning question more than anything: What are the circumstances where the input would be empty?

test/test_torch_distributed_xla_backend.py
pgmoka: I see other operation calls to things like 'gather' and 'alltoall'. What is the reasoning to keep them and remove scatter?

Would it perhaps to improve test documentation on what it does rather than remove 'scatter'?

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -103,6 +103,27 @@ def test_reduce_scatter(self, pin_layout):
...
+    device = torch_xla.device()
+    world_size = xr.world_size()
+    if xr.global_ordinal() == 0:
+      tensors = [
+          torch.tensor([i], device=device, dtype=torch.float)
+          for i in range(world_size)
+      ]
+    else:
+      tensors = None
+    output_tensor = torch.tensor([-1], dtype=torch.float, device=device)
```
pgmoka: NIT: Add space between if statements and next logical block.

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -103,6 +103,27 @@ def test_reduce_scatter(self, pin_layout):
...
+    dist.scatter(output_tensor, tensors, src=0)
+    return output_tensor.cpu()
+
+  def test_scatter(self):
```
pgmoka, line 122: I think this test is complex enough to warrant a comment about what it is doing. My read is:

For one of the threads, a tensor is scattered. Each cell scattered is the equivalent of what the ordinal is. That way, you check here that the the ordinal and the value are the same. If they are, the scatter function worked correctly

pgmoka (commented):
Mostly questions, and a requests for extra documentation

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -103,6 +103,27 @@ def test_reduce_scatter(self, pin_layout):
...
+    results = pjrt.run_multiprocess(self._scatter)
+    for ordinal, value in results.items():
+      np.testing.assert_array_equal(value, [ordinal])
```
pgmoka, line 127: Do we need to worry about race conditions, or is `pjrt.run_multiprocess` blocking?

torch_xla/distributed/xla_backend.py
ghpvnist: +1 for documentation. This could be confusing for those familiar with OpenXLA because XLA also has gather and scatter ops (they're different semantics and not distributed).

Not expecting a full fledged doc here, but just a short description noting the difference.
ref: [scatter op documentation in StableHLO](https://openxla.org/stablehlo/spec#scatter)

test/pjrt/test_collective_ops_tpu.py
bfolie: Yes that is correct. I'll add a doc

test/pjrt/test_collective_ops_tpu.py
bfolie: It is blocking

test/test_torch_distributed_xla_backend.py
bfolie: This test checks that the methods are unimplemented for XLAProcessGroup. `scatter` is now implemented so it is being removed.

What I'm wondering if if there's a reason to add a new test to this file that checks if `group.scatter` outputs the expected HLO. That's what other tests in this file do, but it's not clear to me what value they add beyond the existing tests in test_collective_ops_tpu.

torch_xla/distributed/xla_backend.py
bfolie: Will do. Pytorch also has a non-distributed scatter op, so this can definitely be confusing

torch_xla/distributed/xla_backend.py
bfolie: Input is typically empty on all but the non-source process

torch_xla/distributed/xla_backend.py
bfolie: What do you mean by superfluous?

pgmoka: Unnecessary. If this is the same `ReduceScatterOptions` as https://github.com/pytorch/pytorch/blob/bb462a6237c163774c99e01462703ebce55f4589/torch/csrc/distributed/c10d/Types.hpp#L156, reduceOp  will already be set to `ReduceOp::SUM`

bfolie: Oh, I understand what you're asking now. No, I don't think so. `opts.reduceOp` gets extracted and converted to a string [here](https://github.com/pytorch/xla/blob/215484193f15b8bd7f89cd858625e1d423ab9d4c/torch_xla/distributed/xla_backend.py#L59). It then bounces through a few functions before being used to generate an XLAComputation, [here](https://github.com/pytorch/xla/blob/215484193f15b8bd7f89cd858625e1d423ab9d4c/torch_xla/csrc/cross_replica_reduces.cpp#L65). If we don't specify `opts.reduceOp` here then those translation layers will fail.

I don't think that torch_xla makes use of torch's `ReduceScatterOptions`

pgmoka: You are right. The link I had was to a `ReduceScatterOptions` struct; just a bit of confusing duplicate naming for different declarations.

test/test_torch_distributed_xla_backend.py
pgmoka: Which tests are calling this function in `test/pjrt/test_collective_ops_tpu.py`? The tests I see there are for [`reduce_scatter`](https://github.com/pytorch/xla/blob/master/torch_xla/core/xla_model.py#L839C5-L839C19) and `ReduceScatter` which are seem to me to be higher level abstractions with other things happening. Perhaps I am missing something.

bfolie: The test added in this PR, `test_collective_ops_tpu:test_scatter` calls `torch.dist.scatter` which calls `ProcessGroupXLA.scatter`.

Abstracting things a bit, we have a function A (torch.dist.scatter) that wraps function B (ProcessGroupXLA.scatter). We have a test for A. Should we have a test for B as well? In many cases the answer is yes, especially if B is used in multiple places, if the test for B is logically self-contained and informative, if A adds significant additional logic that we want to test without worrying having to think about B, etc. In this case I'm advocating for not testing B, because

* The test for B is unreliable (as we saw for send/recv, the IR might look reasonable but not work)
* A is a fairly thin wrapper around B
* The contents of A are in upstream PT, so we're not testing them independently of B
* B isn't used anywhere else, nor would it be directly called by other code

What do you think?

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -103,6 +103,30 @@ def test_reduce_scatter(self, pin_layout):
     for ordinal, value in results.items():
       np.testing.assert_array_equal(value, [-ordinal])
 
+  @staticmethod
+  def _scatter():
+    dist.init_process_group("xla", init_method='xla://')
+    device = torch_xla.device()
+    world_size = xr.world_size()
+    if xr.global_ordinal() == 0:
```
ghpvnist, line 112: Minor readability improvement:
```
tensors = None
if xr.global_ordinal() == 0:
      tensors = [
          torch.tensor([i], device=device, dtype=torch.float)
          for i in range(world_size)
      ]
```

ghpvnist (approved):

test/test_torch_distributed_xla_backend.py
pgmoka: Ah, I see. I misunderstood a couple things. The clarification here helps a lot. It is preferable to test the calling API to a method rather than the method itself. Given that `torch.dist.scatter` is serving as the external API to `ProcessGroupXLA.scatter`, it is reasonable to only test it.

I can see this popping up as an issue on some coverage tests, so I would add an explicit comment to the tests in addition to your comment on `torch.distributed.scatter`.

pgmoka (approved):
Follow-up seems good. Let me know if you have any questions on https://github.com/pytorch/xla/pull/9365/files#r2151351304.

Otherwise, LGTM

bfolie:
> One minor thing: I believe the tests failing are due to flakyness. Can you confirm?

The TPU test failure is a known flake and was cleared up by re-running.

The Torchprime e2e test failure is probably real but because PRs from forks aren't exercising that test it's very difficult to tell where the failure comes from

PR #9369, opened 2025-06-16
manual revert of 9346

bfolie:
Checking to see if #9346  is related to the failing torchprime tests


bfolie:
Disregard, fix was merged in #9367

Issue #8815, opened 2025-03-11
CollectivePermute ambiguous support


rpsilva-aws:
## 🐛 Bug

We to discourage the use of CollectivePermute in https://github.com/pytorch/xla/pull/2384. There is no context behind why this is the case, including whether the motivation was hardware specific. It seems that we have enabled it for All-to-All (https://github.com/pytorch/xla/pull/2472) - do we have the same guidance/information from the XLA team?

I can not find any relevant reference to 'all_to_all_emitter'.

cc: @miladm @ManfeiBai @JackCaoG

ManfeiBai:
Hi, @rpsilva-aws, thanks,

IIUC, `all_to_all_emitter` is defined internally for `all_to_all`, which helped fixed `all_to_all` unstable issue;

for `collective_permute`, we would need to sync with XLA team, cc @ddunl 

btw, what's the current failure did you met with `collective_permute` now? any context or reproduce material?

rpsilva-aws:
Any updates on whether there are still known XLA limitations with this op? cc: @yaochengji @ddunl

yaochengji:
Hi @rpsilva-aws , thanks for asking.

Currently I have a PoC script https://github.com/pytorch/xla/blob/chengji/cm/test/torch_distributed/cm_perf.py to demostrate the CM support.

And the current main blocker is that sometimes in real workload, `pin_layout` should be set to True. The CM optimization in XLA compiler cannot handle this at the moment.

yaochengji:
Yes, `ENABLE_COLLECTIVE_MATMUL_IN_MP` is TPU specific, it will turn on some optimization flag in the tpu compiler.

PR #9373, opened 2025-06-17
implement send and recv using collective_permute

bfolie:
#9315


torch_xla/core/xla_model.py
```diff
@@ -748,9 +748,6 @@ def collective_permute(value: torch.Tensor,
                        pairs: List[List[int]]) -> torch.Tensor:
   """Performs a XLA `CollectivePermute()` operation on the input tensor.
 
-  WARNING: This function is not very reliable, may produce wrong results under
-           certain inputs. Use it at your own risk.
-
```
bfolie, line 753: As discussed in #8815 there's no context for this ancient warning. Given the age, lack of details, and lack of any other reported bugs I think it's best to remove it. If we get a specific bug report then we can act on that.

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -326,6 +326,28 @@ def test_all_to_all_single(self, use_dynamo):
                          expected.sort().values),
           f"Got {val}, expected {expected}")
 
+  @staticmethod
+  def _send_recv():
+    dist.init_process_group("xla", init_method='xla://')
+    device = torch_xla.device()
+    world_size = xr.world_size()
+    cutoff = world_size // 2
```
benawilson, line 344: I think if the world size is not even, this test will hang. For example, if world size is 3, then index 0 will send to 1 and 1 will recv from 0, but index 2 will try to recv from 1 without an associated send.

bfolie: Good point. I'll update the test so that it is more defensive

torch_xla/distributed/xla_backend.py
```diff
@@ -235,41 +235,36 @@ def gather(self, *args):
...
-  # test/test_torch_distributed_xla_backend.py for an example.
-  def make_send_channel_id(self, dst_rank, tag):
-    raise NotImplementedError
-
   # Call site e.g.
   # https://github.com/pytorch/pytorch/blob/release/1.10/torch/distributed/distributed_c10d.py#L877
   def send(self, tensors, dst_rank, tag=0):
+    logging.warning(
+        "Individual send/recv ops are inefficient on an XLA device. Consider using xla_model.collective_permute()."
+    )
```
rpsilva-aws: Does it happen to print it everytime we trace?

torch_xla/distributed/xla_backend.py
```diff
@@ -235,41 +235,36 @@ def gather(self, *args):
...
 
-  # Dummy channel id maker. Different backend (TPU, GPU, etc) should replace
-  # the maker with their specific one. See unit test in
-  # test/test_torch_distributed_xla_backend.py for an example.
-  def make_recv_channel_id(self, src_rank, tag):
-    raise NotImplementedError
-
   # Call site e.g.
   # https://github.com/pytorch/pytorch/blob/release/1.10/torch/distributed/distributed_c10d.py#L913
   def recv(self, out_tensors, src_rank, tag=0):
```
rpsilva-aws, line 356: Do we need the warning on the recv end too, so each host has it?

torch_xla/distributed/xla_backend.py
```diff
@@ -235,41 +235,36 @@ def gather(self, *args):
...
 
-  # Dummy channel id maker. Different backend (TPU, GPU, etc) should replace
-  # the maker with their specific one. See unit test in
-  # test/test_torch_distributed_xla_backend.py for an example.
-  def make_send_channel_id(self, dst_rank, tag):
-    raise NotImplementedError
-
   # Call site e.g.
   # https://github.com/pytorch/pytorch/blob/release/1.10/torch/distributed/distributed_c10d.py#L877
   def send(self, tensors, dst_rank, tag=0):
```
rpsilva-aws, line 342: If we're warning to use collective_permute, but it still ends up using a collective permute, should the warning itself be clearer that this is happening under the hood?

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -326,6 +326,28 @@ def test_all_to_all_single(self, use_dynamo):
                          expected.sort().values),
           f"Got {val}, expected {expected}")
 
+  @staticmethod
```
rpsilva-aws, line 339: Last time we checked, we also noticed that https://github.com/pytorch/xla/blob/master/test/test_mp_collective_permute.py didn't work on the CPU, but send/recv did. We might want to double check it.

Is `test/test_torch_distributed_xla_backend.py` tested for CPU and Neuron? Would it be possible to test it and see if the change is compatible?

bfolie: > Is `test/test_torch_distributed_xla_backend.py` tested for CPU and Neuron? Would it be possible to test it and see if the change is compatible?

It is, but it just checks that the expected IR is emitted. It doesn't run anything. And in this case it wasn't a reliable test because, at least for TPU, that IR does not actually run.

test_mp_collective_permute is run for both TPU and Neuron. I don't think it works for CPU but neither do send/recv. The success of test_mp_collective_permute indicates this change should work for Neuron, but to be more certain I could add a test that covers a pipeline-like transfer in addition to the existing test of a permutation-like transfer.

The most direct test would be something like what's in test_collective_ops_tpu.py, which runs the ops to completion, for Neuron.

torch_xla/distributed/xla_backend.py
bfolie: I could word this better. The real advice is to restructure your code so that each process calls collective_permute *with all of the send-recv pairs*

torch_xla/distributed/xla_backend.py
bfolie: Probably. I'm not sure how to only make it print once -- will look into it

test/pjrt/test_collective_ops_tpu.py
```diff
@@ -326,6 +326,28 @@ def test_all_to_all_single(self, use_dynamo):
...
+    cutoff = world_size // 2
+    index = xr.global_ordinal()
+    tensor = torch.tensor([index + 1], dtype=torch.float, device=device)
+    if index < cutoff:
+      dist.send(tensor, index + cutoff)
+    else:
+      dist.recv(tensor, index - cutoff)
+    return tensor.cpu()
+
+  def test_send_recv(self):
```
pgmoka: The original test separated both send and receive. While this is more code efficient, it might be harder to debug as it will not be obvious what the issue is.

I think keeping a test for the total interaction is valid, but is there a way to replicate the other two tests that existed previously?

torch_xla/distributed/xla_backend.py
pgmoka: I checked around, and couldn't find a in built way to do this through `logging.warning`. Given this is at warning level and can be filtered out, is it worth to seek a solution?

torch_xla/distributed/xla_backend.py
```diff
@@ -235,41 +235,36 @@ def gather(self, *args):
...
-      input_as_result = xm.send(t, channel_id)
-      # Make the sent tensor depend on the token, such that the `send`
-      # op can actually be built into the computation graph.
+      result_t = xm.collective_permute(
+          t, pairs=[[xr.global_ordinal(), dst_rank]])
+      # Every process must have the same IR, otherwise they deadlock. But in
+      # the receiving process the provided tensor receives the result, while
+      # in the sending process it is unchanged. The solution used here is to
+      # have every process copy a linear combination of the two tensors, but
+      # send/recv use different coefficients to achieve different outcomes.
```
pgmoka: This took a couple reads until I understood what was going on here. My understanding is that by having both `result_t * X + t * Y` you are having both operation IRs be the same as X and Y are constants. That way when the IRs are compared they will be equivalent.

If this understanding is correct, could you add a little bit more here to make it more apparent?

torch_xla/distributed/xla_backend.py
pgmoka, line 356: We should not assume someone reading "recv" will have read the documentation for "send". I think we should add documentation here. I would then add a note specific about what the IR expectation will be for "send" and "recv" on each of their comments.

test/pjrt/test_collective_ops_tpu.py
bfolie: send and recv don't work independently. The original test was a "dry run" -- it checked the IR but didn't execute. If it did execute it would fail.

test/pjrt/test_collective_ops_tpu.py
rpsilva-aws: > The most direct test would be something like what's in test_collective_ops_tpu.py, which runs the ops to completion, for Neuron.

This would be great. Any chance we can move it outside of this file and make it general? I can help test it out if so. Otherwise, I'll need to follow up if we can port this entire file to Neuron. I see `tpu.num_expected_global_devices`, and `pjrt.run_multiprocess`, but haven't seen/used these before.

bfolie:
The approach implemented here works for a "pipeline" type operation but does not work for a "permutation" type operation. The way this is commonly done in native pytorch in order to avoid deadlocks is that half of the devices send and the other receive, then they switch roles. What this means is that the sending and receiving tensors must be different, and one half of the devices end up having a different IR than the other half, resulting in a deadlock. I'm still searching for a way around this.

bfolie:
The only way I was able to make a "permutation" type op (every device sends and every device receives) work is by inserting a sync after each set of send/recv. This is not ideal. It's better than the status quo for TPU, which is that send/recv don't work at all. But since Neuron does have something working I'll defer to you @rpsilva-aws . We can put this on ice until the Send/Recv XLA ops can be called directly.

rpsilva-aws:
Hm, that does complicate things... I have it working on TRN, though I deviated a bit with multi-operands to capture tokens. I'll end up creating a PR for this one, which would build upon the work you had in the prior commits. Actually, TRN has the same limitation for send/recv, requiring a graph break.

Do you think we can merge this PR without the sync since it's working for existing devices (e.g. TRN), and revisit as we figure out the underlying issues with TPU? If you want to defer until the new ops, or we re-raise the need as we bring in our work, both are ok with me.

bfolie:
> Do you think we can merge this PR without the sync since it's working for existing devices (e.g. TRN), and revisit as we figure out the underlying issues with TPU?

There are two tests in the PR, `test_send_recv_pipeline` and `test_send_recv_permute`. Without a sync, the former works on TPU (that's what I initially committed) but the latter does not. Does `test_send_recv_permute` work on TRN without a sync? I would expect it to have the same non-uniform-IR problems, so that would be a surprise and interesting to me if it did work.

> I have it working on TRN, though I deviated a bit with multi-operands to capture tokens. I'll end up creating a PR for this one, which would build upon the work you had in the prior commits.

I'd be interested in seeing that

kvshbg-aws:
@bfolie i tried your changes (on top of ToT) on neuron device and also wrote a small test specific to neuron device which uses 2 ranks and does `pipeline_cp_ops` and `permute_cp_ops` as they are in your test files, and i even added an additional test to verify if the computation on the receiving rank happens as expected after receiving the tensor. Writing the test functions over here if you want to take a look (the communication part is exactly same as your tests, just changed the way we do asserts on neuron device using the test file since we use torchrun for some neuron specific tests).

I also tried reproducing the cpu xla_op test failures from this PR, and was not able to reproduce those locally

```
    def test_pipeline_send_recv(self):
        """Test basic pipeline pattern: first half sends to second half"""
        cutoff = self.world_size // 2
        index = xr.global_ordinal()
        
        tensor = torch.tensor([index], dtype=torch.float, device=self.device)
        
        # Pipeline communication: first half sends to second half
        if index < cutoff:
            dist.send(tensor, index + cutoff)
        else:
            dist.recv(tensor, index - cutoff)
        
        # Verify communication correctness
        if index < cutoff:
            expected = torch.tensor([index], dtype=torch.float)
            self.assertTrue(torch.allclose(tensor.cpu(), expected))
        else:
            expected = torch.tensor([index - cutoff], dtype=torch.float)
            self.assertTrue(torch.allclose(tensor.cpu(), expected))

    def test_permute_communication(self):
        """Test permutation pattern: each device sends to next and receives from previous"""
        index = xr.global_ordinal()
        sending_tensor = torch.tensor([index], dtype=torch.float, device=self.device)
        receiving_tensor = torch.tensor([-1.0], dtype=torch.float, device=self.device)
        
        # Ring communication pattern with deadlock avoidance
        if index % 2 == 0:
            dist.send(sending_tensor, (index + 1) % self.world_size)
            dist.recv(receiving_tensor, (index - 1) % self.world_size)
        else:
            dist.recv(receiving_tensor, (index - 1) % self.world_size)
            dist.send(sending_tensor, (index + 1) % self.world_size)
        
        # Verify ring communication correctness
        expected = torch.tensor([(index - 1) % self.world_size], dtype=torch.float)
        self.assertTrue(torch.allclose(receiving_tensor.cpu(), expected))

    def test_pipeline_send_recv_with_computation(self):
        """Test pipeline pattern with computation on received data"""
        cutoff = self.world_size // 2
        index = xr.global_ordinal()
        
        tensor = torch.tensor([index], dtype=torch.float, device=self.device)
        
        # Pipeline with computation on receiver side
        if index < cutoff:
            dist.send(tensor, index + cutoff)
        else:
            dist.recv(tensor, index - cutoff)
            res = tensor + 3.0  # Compute on received data
        
        # Verify results
        if index < cutoff:
            expected = torch.tensor([index], dtype=torch.float)
            self.assertTrue(torch.allclose(tensor.cpu(), expected))
        else:
            expected = torch.tensor([3.0], dtype=torch.float)
            self.assertTrue(torch.allclose(res.cpu(), expected))
```

cc: @rpsilva-aws

bfolie:
> @bfolie i tried your changes (on top of ToT) on neuron device and also wrote a small test specific to neuron device which uses 2 ranks and does `pipeline_cp_ops` and `permute_cp_ops` as they are in your test files, and i even added an additional test to verify if the computation on the receiving rank happens as expected after receiving the tensor. 

Good to know. I would expect that this PR as it currently is, with the syncs on both send and receive, would work for all devices. It's not the most efficient but if this works for you then we can go ahead and merge it.

rpsilva-aws (approved):
We need tests under neuron/ too, but we can do that separately. LGTM.

pgmoka (commented):
LGTM

PR #9374, opened 2025-06-17
Torchax: Allow reuse of the jittable procedure with different functio…

zmelumian972:
Torchax exposes an API named JittableMoudle

JittableModule will accept an existing model and wrap it's forward function with torchax jittable call, making it seamless for the user to wrap torch models via torchax

this PR aims to expose not just the forward function, but any function the user chooses, with the help of inheritance


zmelumian972:
Guys what can I do to move this PR further?

qihqi (approved):

PR #9380, opened 2025-06-17
Wrap def_static to enable warning reporting in static python functions.

zhanyong-wan:
This makes `def_static`'s behavior consistent with `def` and `def_init` in `PythonScope`.

Also make a small style improvement: make `seed_info_id` `constexpr` and rename it to indicate that it's a constant.

### DIFF
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -82,7 +82,7 @@
 namespace torch_xla {
 namespace {
 
-static int64_t seed_info_id = -127389;
+constexpr int64_t kSeedInfoId = -127389;
 
 // Wraps a python scope (e.g. py::module) to provide more convenient APIs.
 // It behaves like a Scope object but has enhanced behaviors for the def*()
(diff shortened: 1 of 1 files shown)


torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -103,59 +103,78 @@ class PythonScope : public Scope {
...
+          // warnings.
+          torch::PyWarningHandler handler;
+          return f(args...);
+        };
+        if constexpr (kind == FunctionKind::kStatic) {
+          scope.def_static(name, std::move(lambda), extra...);
+        } else {
+          scope.def(name, std::move(lambda), extra...);
+        }
+      }
```
ysiraichi: What do you think of unconditionally creating a `lambda` variable first (since all execution paths require it), and then call `scope.def*()` as needed?

ysiraichi (approved):
I left a suggestion, but lgtm, otherwise.

torch_xla/csrc/init_python_bindings.cpp
zhanyong-wan: Great point. Done!

PR #9381, opened 2025-06-17
Deprecate ShapeOfXlaOp in favor of GetShape

zhanyong-wan:
For https://github.com/pytorch/xla/issues/9096

Deprecate `ShapeOfXlaOp()` as:

1. It ignores potential errors from XLA and thus is unsafe.
2. It violates the Google style guide: ["Do not create classes only to group static members"](https://google.github.io/styleguide/cppguide.html#Nonmember,_Static_Member,_and_Global_Functions).
3. The `OfXlaOp` part of the name is redundant as it's already obvious from the parameter type.

Replace it with a new function `GetShape()`.

### DIFF
--- a/torch_xla/csrc/BUILD
+++ b/torch_xla/csrc/BUILD
@@ -243,6 +243,8 @@ ptxla_cc_library(
     srcs = ["shape_builder.cpp"],
     hdrs = ["shape_builder.h"],
     deps = [
+        "@com_google_absl//absl/base:core_headers",
+        "@com_google_absl//absl/base:nullability",
         "@com_google_absl//absl/types:span",
         "@xla//xla:shape_util",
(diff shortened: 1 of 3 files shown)


zhanyong-wan:
@ghpvnist and @ysiraichi , somehow GH doesn't let me request a code review on this PR. Can you review it anyway? Thanks.

ghpvnist (approved):
Do you plan to do a follow up to change existing uses of `ShapeOfXlaOp`?

zhanyong-wan:
> Do you plan to do a follow up to change existing uses of `ShapeOfXlaOp`?

There are many uses of this. I don't plan to rewrite all of them, as that doesn't have a big ROI. Instead, I may replace some of them as needed for improving a user error message.

PR #9382, opened 2025-06-17
Add jax_device context manager to control the device target

zzzwen:
context manager to control the device/platform jax array is on


torchax/torchax/tensor.py
```diff
@@ -322,11 +325,22 @@ def __init__(self, configuration=None):
 
     self._manually_entered = False
     self.enabled = False
-    self._jax_devices = set(["jax", "jax_cpu", "xla"])
+
+    self.prng_key = jax.random.key(torch.initial_seed() % (1 << 63))
```
qihqi: qq. why need another prng_key?

zzzwen: Thanks for catching it, I was merging the conflict on github's webui and made a mistake

PR #9383, opened 2025-06-17
Include pytorch and torchxla revisions in the compilation cache key.

zhanyong-wan:


torch_xla/csrc/xla_graph_executor.cpp
```diff
@@ -525,8 +525,16 @@ torch::lazy::hash_t XLAGraphExecutor::GetGraphHash(
         static_cast<XlaDeviceType>(tensor->GetDevice().type())));
   }
   PostOrderData po_data = RunPostOrder(ir_values, &coll);
+
+  // Seed the hash with the pytorch git revision and the torch_xla git
+  // revision, so that differennt versions of the code can produce
+  // different hashes for the same graph.
   torch::lazy::hash_t res_hash = torch::lazy::HashCombine(
-      coll.hash, torch::lazy::Hash(po_data.parameter_sequence));
+      torch::lazy::Hash(std::string_view(TORCH_GITREV)),
```
tengyifei: nit: technically if I'm developing PTXLA locally, I could fix a bug without making a git commit, in which case the SHA won't change.

Maybe in the future we'd need to read the BuildID of the ELF binary to be really robust.

zhanyong-wan: Thanks! Sorry this was sent pre-maturely. I'm still tweaking the implementation. Somehow GH automatically added you as a reviewer and I didn't notice it.

Enjoy your vacation!

zhanyong-wan: Yes, I agree that this is not completely safe, but it only affects PTXLA developers, so I'm not so worried. I'll add a comment to call it out.

tengyifei (approved):

Issue #9339, opened 2025-06-11
normalize error handling in PyTorch/XLA


zhanyong-wan:
## 🐛 Bug

Error handling in PyTorch/XLA is inconsistent and leads to poor user experience. We ought to clearly distinguish between user errors and internal errors and handle them differently.

Once [the design](https://docs.google.com/document/d/1hlr_CFcB_oO_SINak8YXOqORKdxX1_PQq9Ku3TjLyD8/edit?usp=sharing) is approved, we should implement it.

ysiraichi:
I was actually thinking we could go forward without that design document. The reason is two-fold:

1. We have already reached a consensus on what to do: start using `Status` and `StatusOr` for user error handling; and crash on internal assertions
2. That design document was created mainly for aligning PyTorch/XLA error handling with PyTorch. Since that's not what we are looking for at the moment, I don't think that document is relevant, right now.

What do you think?

ysiraichi:
### `GetComputationClient()`

I'm currently working on getting the function `GetComputationClient()` to return `StatusOr<T>`. This is a function that is used a lot in lots of places. Therefore, I'm thinking of doing something similar to #9381: create a new function that returns `StatusOr<T>`, while leaving the old call with a `ABSL_DEPRECATED()` annotation.

Specifically, in order to make transitioning easier, I'm thinking of creating a new function with the same name, but on a new namespace `status` that resides in the same namespace that the old function resides. So, we would have the old `torch_xla::runtime::GetComputationClient()` and the new `torch_xla::runtime::status::GetComputationClient()` that returns a `StatusOr<T>` type. 

```c++
namespace torch_xla::runtime {
  namespace status {
    absl::StatusOr<ComputationClient*> GetComputationClient();
  }

  ComputationClient* GetComputationClient();
}
```

@zhanyong-wan @ghpvnist 
Let me know what you think.

zhanyong-wan:
Agreed on deprecating the old function and introducing a new.

We should make it more awkward to use the old function than to use the new. Therefore I propose to rename the old one to `GetComputationClientOrDie()`, and introduce the new as `GetComputationClient()` in `torch_xla::runtime`.

ysiraichi:
### init_python_bindings.cpp

[...]

PR #9384, opened 2025-06-18
ErrorHandling: make `GetComputationClient()` return `StatusOr<T>` type. 

ysiraichi:
Ref: #9339 

This PR creates a new version of `GetComputationClient()` that returns a `StatusOr<T>` type for better error handling. Since this function is frequently used in the codebase, its old version remains, but annotated with `ABSL_DEPRECATED()` (following #9381 example) and renamed to `GetComputationClientOrDie()`. The new version is created with the old name. Summarizing:

- Old `GetComputationClient()` is renamed to `GetComputationClientOrDie()`
- New `torch_xla::runtime::status::GetComputationClient()` function that returns `StatusOr<T>` type
- The old version remains, and calls the new function. It uses `XLA_CHECK()` for throwing an error on non-ok status, maintaining its old behavior
- New `ConsumeAndMaybeThrow()` function for throwing errors if non-ok status reach the Python API boundary

### DIFF
--- a/test/cpp/test_runtime.cpp
+++ b/test/cpp/test_runtime.cpp
@@ -5,7 +5,7 @@
 namespace torch_xla::runtime {
 
 TEST(RuntimeTest, NullComputationClient) {
-  auto client = GetComputationClientIfInitialized();
+  auto* client = GetComputationClientIfInitialized();
   EXPECT_EQ(client, nullptr);
 }
 
@@ -19,13 +19,13 @@ TEST(RuntimeTest, GetComputationClientSuccess) {
   // Check all the APIs return the same valid ComputationClient.
 
   client = GetComputationClientOrDie();
-  EXPECT_NE(client, nullptr);
+  ASSERT_NE(client, nullptr);
 
   auto status = GetComputationClient();
-  EXPECT_TRUE(status.ok());
-  EXPECT_EQ(client, status.value());
+  ASSERT_TRUE(status.ok());
 
-  EXPECT_EQ(client, GetComputationClientIfInitialized());
+  EXPECT_EQ(status.value(), client);
+  EXPECT_EQ(GetComputationClientIfInitialized(), client);
 }
 
 }  // namespace torch_xla::runtime
--- a/test/cpp/test_runtime.cpp
+++ b/test/cpp/test_runtime.cpp
@@ -1,6 +1,6 @@
-#include "torch_xla/csrc/runtime/runtime.h"
-
 #include <gtest/gtest.h>
+
+#include "torch_xla/csrc/runtime/runtime.h"
 
 namespace torch_xla::runtime {
 
--- a/test/cpp/test_runtime.cpp
+++ b/test/cpp/test_runtime.cpp
@@ -4,12 +4,7 @@
 
 namespace torch_xla::runtime {
 
-TEST(RuntimeTest, NullComputationClient) {
-  auto* client = GetComputationClientIfInitialized();
-  EXPECT_EQ(client, nullptr);
-}
-
-TEST(RuntimeTest, GetComputationClientSuccess) {
+TEST(RuntimeTest, ComputationClientInitialization) {
   ComputationClient* client;
 
   client = GetComputationClientIfInitialized();
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -1696,7 +1696,7 @@ void InitXlaModuleBindings(py::module m) {
            })
       .def("_init_computation_client",
            []() {
-             ConsumeAndMaybeThrow(runtime::status::GetComputationClient());
+             ConsumeAndMaybeThrow(runtime::GetComputationClient());
            })
       .def("_xla_get_device_hw_type",
            [](const at::Tensor& tensor) {
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -172,6 +172,22 @@ class PythonScope : public Scope {
   };
 };
 
+static void ConsumeAndMaybeThrow(absl::Status status) {
+  if (status.ok()) {
+    return;
+  }
+
+  throw std::runtime_error(std::string(status.message()));
+}
+
+template <class T>
+static T ConsumeAndMaybeThrow(absl::StatusOr<T> status) {
+  if (status.ok()) {
+    return std::move(status.value());
+  }
+  ConsumeAndMaybeThrow(status.status());
+}
+
 struct NoGilSection {
   NoGilSection() : state(PyEval_SaveThread()) {}
   ~NoGilSection() { PyEval_RestoreThread(state); }
@@ -1681,7 +1697,10 @@ void InitXlaModuleBindings(py::module m) {
             XLA_CHECK(UseVirtualDevice(/*force_spmd=*/true));
            })
       .def("_init_computation_client",
-           []() { runtime::GetComputationClient(); })
+           []() {
+             auto status = runtime::status::GetComputationClient();
+             ConsumeAndMaybeThrow(status);
+           })
       .def("_xla_get_device_hw_type",
            [](const at::Tensor& tensor) {
             XLATensorPtr xtensor = bridge::GetXlaTensor(tensor);
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -180,10 +180,7 @@ static void ConsumeAndMaybeThrow(absl::Status status) {
 
 template <class T>
 static T ConsumeAndMaybeThrow(absl::StatusOr<T> status) {
-  if (status.ok()) {
-    return std::move(status.value());
-  }
-  ConsumeAndMaybeThrow(status.status());
+  return std::move(status.value());
 }
 
 struct NoGilSection {
--- a/torch_xla/csrc/init_python_bindings.cpp
+++ b/torch_xla/csrc/init_python_bindings.cpp
@@ -173,11 +173,9 @@ class PythonScope : public Scope {
 };
 
 static void ConsumeAndMaybeThrow(absl::Status status) {
-  if (status.ok()) {
-    return;
+  if (!status.ok()) {
+    throw std::runtime_error(std::string(status.message()));
   }
-
-  throw std::runtime_error(std::string(status.message()));
 }
 
 template <class T>
@@ -233,7 +231,7 @@ torch::lazy::BackendDevice GetDeviceOrCurrent(const std::string& device_str) {
 
 void WaitDeviceOps(absl::Span<const std::string> devices = {}) {
   XLAGraphExecutor::Get()->WaitDeviceOps(devices);
-  runtime::GetComputationClient()->WaitDeviceOps(devices);
+  runtime::GetComputationClientOrDie()->WaitDeviceOps(devices);
 }
 
 void PrepareToExit() {
@@ -687,7 +685,7 @@ void StepMarker(const std::string& device_str,
   bool debug_mode = runtime::sys_util::GetEnvBool("PT_XLA_DEBUG", false);
   if (TF_PREDICT_FALSE(debug_mode)) {
     std::string report = runtime::metrics::CreatePerformanceReport(
-        runtime::GetComputationClient()->GetMetrics());
+        runtime::GetComputationClientOrDie()->GetMetrics());
     if (!report.empty()) {
       std::string fout =
           runtime::sys_util::GetEnvString("PT_XLA_DEBUG_FILE", "");
@@ -936,7 +934,7 @@ py::dict GetMemoryInfo(const std::string& device_str) {
     NoGilSection nogil;
     torch::lazy::BackendDevice device = GetDeviceOrCurrent(device_str);
     mem_info =
-        runtime::GetComputationClient()->GetMemoryInfo(device.toString());
+        runtime::GetComputationClientOrDie()->GetMemoryInfo(device.toString());
   }
   auto py_dict = py::dict();
   py_dict["bytes_used"] = mem_info.bytes_used;
@@ -1244,7 +1242,7 @@ class PyLoweringContext {
 
     // Fetch this parameter data
     std::vector<xla::Literal> literals =
-        runtime::GetComputationClient()->TransferFromDevice(
+        runtime::GetComputationClientOrDie()->TransferFromDevice(
             UnwrapXlaData(device_data));
 
     // Create a mapping from paramater id to the tensor data
@@ -1485,9 +1483,9 @@ void InitXlaModuleBindings(py::module m) {
             CreateComputationShapeFromTensor(tensor, nullptr);
         if (minibatch) {
           int num_local_devices =
-              runtime::GetComputationClient()->GetLocalDevices().size();
+              runtime::GetComputationClientOrDie()->GetLocalDevices().size();
           int num_global_devices =
-              runtime::GetComputationClient()->GetAllDevices().size();
+              runtime::GetComputationClientOrDie()->GetAllDevices().size();
           XLA_CHECK(tile_assignment.size() == num_global_devices)
               << "Minibatch sharding only supports sharding along the batch "
                  "dimension";
@@ -1713,16 +1711,16 @@ void InitXlaModuleBindings(py::module m) {
               // Under SPMD context, there is only one virtual devices from
               // user perspective.
               std::vector<std::string> all_devices =
-                  runtime::GetComputationClient()->GetAllDevices();
+                  runtime::GetComputationClientOrDie()->GetAllDevices();
               all_devices.resize(1);
               return all_devices;
             } else {
-              return runtime::GetComputationClient()->GetLocalDevices();
+              return runtime::GetComputationClientOrDie()->GetLocalDevices();
             }
            })
       .def("_get_stream_for_cuda_device",
            [](const int device_id) {
-            return runtime::GetComputationClient()->GetCudaStreamForDevice(
+            return runtime::GetComputationClientOrDie()->GetCudaStreamForDevice(
                 device_id);
            })
       .def("_xla_num_devices",
@@ -1730,17 +1728,17 @@ void InitXlaModuleBindings(py::module m) {
             if (UseVirtualDevice()) {
               return 1;
             } else {
-              return runtime::GetComputationClient()->GetNumLocalDevices();
+              return runtime::GetComputationClientOrDie()->GetNumLocalDevices();
             }
            })
       .def("_xla_num_global_devices",
            []() -> int64_t {
-            return runtime::GetComputationClient()->GetNumDevices();
+            return runtime::GetComputationClientOrDie()->GetNumDevices();
            })
       .def("_xla_get_all_devices",
            []() {
             std::vector<std::string> all_devices =
-                runtime::GetComputationClient()->GetAllDevices();
+                runtime::GetComputationClientOrDie()->GetAllDevices();
             if (UseVirtualDevice()) {
               // Under SPMD context, there is only one virtual devices from
               // user perspective.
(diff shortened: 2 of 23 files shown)


torch_xla/csrc/runtime/runtime.h
```diff
@@ -3,10 +3,19 @@
...
 
-namespace torch_xla {
-namespace runtime {
+namespace torch_xla::runtime {
+namespace status {
 
 // Returns the ComputationClient singleton.
+absl::StatusOr<ComputationClient*> GetComputationClient();
+
+}
```
ghpvnist: nit:
`}  // namespace status`

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,79 @@
...
+  return client;
+}
+
+absl::StatusOr<ComputationClient*> GetComputationClient() {
+  static std::unique_ptr<ComputationClient> client;
 
-    g_computation_client_initialized = true;
-    return client;
-  }();
+  if (client.get() == nullptr) {
```
ghpvnist: This check is always true so not needed right?

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,79 @@
...
+    }
+  }
 
   return client.get();
 }
 
+}  // namespace status
+
+ComputationClient* GetComputationClient() {
+  auto client = status::GetComputationClient();
```
ghpvnist: To keep the underlying implementation same, perhaps:
```
std::unique_ptr<ComputationClient> client = InitializeComputationClient().value();
XLA_CHECK(client);
g_computation_client_initialized = true;
return client.get();
```

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -160,6 +160,22 @@ class PythonScope : public Scope {
   };
 };
 
+static void ConsumeAndMaybeThrow(absl::Status status) {
+  if (status.ok()) {
```
ghpvnist:
```
static void ConsumeAndMaybeThrow(absl::Status status) {
  if (!status.ok()) {
    throw std::runtime_error(std::string(status.message()));
  }
}
```

torch_xla/csrc/runtime/runtime.cpp
ysiraichi: Not really. Note that `client` is actually a static variable.

torch_xla/csrc/runtime/runtime.cpp
ysiraichi: Problem is that we will crash if we call `Status::value()`, given a non-ok status.

ysiraichi: `XLA_CHECK()` will throw an exception.

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,79 @@
...
-    // static bool use_ifrt = sys_util::GetEnvBool("XLA_USE_IFRT", false);
-    static bool use_ifrt = false;
-    if (sys_util::GetEnvString(env::kEnvPjRtDevice, "") != "") {
-      if (use_ifrt) {
-        client = std::make_unique<IfrtComputationClient>();
-      } else {
-        client = std::make_unique<PjRtComputationClient>();
-      }
+  // Disable IFRT right now as it currently crashes.
+  // static bool use_ifrt = sys_util::GetEnvBool("XLA_USE_IFRT", false);
```
ghpvnist, line 26: Unrelated to PR, but can you prepend a `// TODO` here so we know to fix in the future?

torch_xla/csrc/runtime/runtime.cpp
ghpvnist: Got it, then the change lgtm.

ysiraichi:
Incorporated the changes suggested by @zhanyong-wan [comment][1].

[1]: https://github.com/pytorch/xla/issues/9339#issuecomment-2985172117

torch_xla/csrc/init_python_bindings.cpp
```diff
@@ -160,6 +160,22 @@ class PythonScope : public Scope {
...
+    return;
+  }
+
+  throw std::runtime_error(std::string(status.message()));
+}
+
+template <class T>
+static T ConsumeAndMaybeThrow(absl::StatusOr<T> status) {
```
zhanyong-wan, line 182: If you call `status.value()`, it will throw if the value is not present, so I'm not sure what value this adds.

torch_xla/csrc/runtime/runtime.h
```diff
@@ -3,10 +3,19 @@
...
+
+ABSL_DEPRECATED(
+    "Use status::GetComputationClient(), instead. "
+    "This function throws an exception on error, instead of "
+    "actually handling the StatusOr return value, which is "
+    "safer.")
 ComputationClient* GetComputationClient();
```
zhanyong-wan: Did you see my comments on the issue about renaming this to `*OrDie()` and having the new function in the `torch_xla::runtime` namespace?

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,79 @@
...
 std::atomic<bool> g_computation_client_initialized(false);
 
-ComputationClient* GetComputationClient() {
-  static std::unique_ptr<ComputationClient> client = []() {
-    if (sys_util::GetEnvBool("XLA_DUMP_FATAL_STACK", false)) {
-      tsl::testing::InstallStacktraceHandler();
-    }
+namespace status {
+
+static absl::StatusOr<std::unique_ptr<ComputationClient>> InitializeComputationClient() {
```
zhanyong-wan: Document this? In particular, use `absl_nonnull` to indicate that the returned unique_ptr cannot be null.

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,79 @@
...
-    static bool use_ifrt = false;
-    if (sys_util::GetEnvString(env::kEnvPjRtDevice, "") != "") {
-      if (use_ifrt) {
-        client = std::make_unique<IfrtComputationClient>();
-      } else {
-        client = std::make_unique<PjRtComputationClient>();
-      }
+  // Disable IFRT right now as it currently crashes.
+  // static bool use_ifrt = sys_util::GetEnvBool("XLA_USE_IFRT", false);
+  static bool use_ifrt = false;
```
zhanyong-wan: `static` => `const`?

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,79 @@
...
+  } else {
+    return absl::FailedPreconditionError("$PJRT_DEVICE is not set.");
+  }
 
-    XLA_CHECK(client);
+  return client;
+}
+
+absl::StatusOr<ComputationClient*> GetComputationClient() {
+  static std::unique_ptr<ComputationClient> client;
```
zhanyong-wan: The old code is threadsafe but the new code is not. Please keep the old pattern of initializing `client` where it's defined.

`static ... client = [] ...;`

zhanyong-wan (changes requested):

torch_xla/csrc/runtime/runtime.cpp
ysiraichi: Ah, I thought `status.value()` call would crash! It does throw an error as you and @zhanyong-wan pointed out.

torch_xla/csrc/runtime/runtime.h
ysiraichi: Yeah. I did update as you suggested. I think you reviewed an outdated version.

ghpvnist (changes requested):
Can you create a unit test for the new API? `runtime_test.cpp` in the same directory could work.

ysiraichi:
@ghpvnist What kind of tests did you have in mind? 
I think C++ tests usually go inside `test/cpp` directory.

ghpvnist:
> I think C++ tests usually go inside test/cpp directory.

This directory has precedent of tests within the same directory. I prefer to have tests close where the files are stored that way, future refactoring changes would be easier. Wdyt? @zhanyong-wan 

> What kind of tests did you have in mind?

I was thinking the typical unit tests for test coverage purposes. Something along the lines of:
* Check that calls to `GetComputationClient` return a valid client.
* `GetComputationClientIfInitialized` should return nullptr vs a valid client.
* etc.

torch_xla/csrc/init_python_bindings.cpp
ysiraichi: I think that it would be better to have one source of truth for raising exceptions from these status constructs. In other words, always call `ConsumeAndMaybeThrow` whenever we want to do that. What do you think?

zhanyong-wan:
> > I think C++ tests usually go inside test/cpp directory.
> 
> This directory has precedent of tests within the same directory. I prefer to have tests close where the files are stored that way, future refactoring changes would be easier. Wdyt? @zhanyong-wan 

While I prefer to have tests in the same directory as the code too, pytorch uses the `test/cpp/test_foo.cpp` convention. I think we should align with that to ease upstreaming.

torch_xla/csrc/runtime/runtime_test.cpp
```diff
@@ -0,0 +1,31 @@
...
+TEST(RuntimeTest, NullComputationClient) {
+  auto client = GetComputationClientIfInitialized();
+  EXPECT_EQ(client, nullptr);
+}
+
+TEST(RuntimeTest, GetComputationClientSuccess) {
+  ComputationClient* client;
+
+  client = GetComputationClientIfInitialized();
+  EXPECT_EQ(client, nullptr);
```
zhanyong-wan: This overlaps with the previous test case. Remove?

torch_xla/csrc/runtime/runtime_test.cpp
```diff
@@ -0,0 +1,31 @@
+#include "torch_xla/csrc/runtime/runtime.h"
+
+#include <gtest/gtest.h>
+
+namespace torch_xla::runtime {
+
+TEST(RuntimeTest, NullComputationClient) {
+  auto client = GetComputationClientIfInitialized();
```
zhanyong-wan: Style: `auto*`

torch_xla/csrc/runtime/runtime_test.cpp
```diff
@@ -0,0 +1,31 @@
...
+  EXPECT_EQ(client, nullptr);
+
+  // Initialize the ComputationClient.
+  // Check all the APIs return the same valid ComputationClient.
+
+  client = GetComputationClientOrDie();
+  EXPECT_NE(client, nullptr);
+
+  auto status = GetComputationClient();
+  EXPECT_TRUE(status.ok());
```
zhanyong-wan: `ASSERT_TRUE`. We don't want to continue if this is false.

torch_xla/csrc/runtime/runtime_test.cpp
```diff
@@ -0,0 +1,31 @@
...
+  EXPECT_EQ(client, status.value());
```
zhanyong-wan: Style: swap the arguments - we want `EXPECT_EQ(value_being_tested, expected_value)`.

torch_xla/csrc/runtime/runtime_test.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+  EXPECT_EQ(client, GetComputationClientIfInitialized());
```
zhanyong-wan: Ditto.

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,66 @@
...
+    return absl::FailedPreconditionError("$PJRT_DEVICE is not set.");
+  }
+
+  g_computation_client_initialized = true;
+  return client;
+}
+
+absl::StatusOr<ComputationClient*> GetComputationClient() {
+  static absl::StatusOr<absl_nonnull std::unique_ptr<ComputationClient>>
+      maybeClient = InitializeComputationClient();
```
zhanyong-wan: Style: `maybe_client`

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,66 @@
...
     }
+  } else {
+    return absl::FailedPreconditionError("$PJRT_DEVICE is not set.");
+  }
+
+  g_computation_client_initialized = true;
+  return client;
+}
+
+absl::StatusOr<ComputationClient*> GetComputationClient() {
```
zhanyong-wan: Use `abs_nonnull` here?

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,66 @@
...
+  static absl::StatusOr<absl_nonnull std::unique_ptr<ComputationClient>>
```
zhanyong-wan: Style: per https://google.github.io/styleguide/cppguide.html, static variable must trivially destructible. We'll need to change this into a `static const StatusOr<...>* const`.

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,66 @@
...
 
-    XLA_CHECK(client);
+  if (!maybeClient.ok()) {
+    return maybeClient.status();
+  }
 
-    g_computation_client_initialized = true;
-    return client;
-  }();
+  auto client = maybeClient.value().get();
```
zhanyong-wan: Style: always write `auto*` instead of `auto` when the type is a raw pointer.

torch_xla/csrc/runtime/runtime.h
```diff
@@ -3,19 +3,24 @@
 
 #include "torch_xla/csrc/runtime/computation_client.h"
 
-namespace torch_xla {
-namespace runtime {
+namespace torch_xla::runtime {
 
 // Returns the ComputationClient singleton.
-ComputationClient* GetComputationClient();
+absl::StatusOr<ComputationClient*> GetComputationClient();
```
zhanyong-wan: Annotate with `absl_nonnull`?

torch_xla/csrc/runtime/runtime.h
```diff
@@ -3,19 +3,24 @@
...
+
+ABSL_DEPRECATED(
+    "Use status::GetComputationClient(), instead. "
+    "This function throws an exception on error, instead of "
+    "actually handling the StatusOr return value, which is "
+    "safer.")
+ComputationClient* GetComputationClientOrDie();
```
zhanyong-wan: Ditto.

torch_xla/csrc/runtime/runtime.h
```diff
@@ -3,19 +3,24 @@
...
 
 ComputationClient* GetComputationClientIfInitialized();
```
zhanyong-wan, line 21: Document? In particular, clarify when this returns null.

torch_xla/csrc/init_python_bindings.cpp
zhanyong-wan: Why cannot `.value()` be that one source of truth?

zhanyong-wan (changes requested):

torch_xla/csrc/runtime/runtime_test.cpp
ysiraichi: I am calling this first, so as to show that the value returned by `GetComputationClientIfInitialized()` changes, after the initialization correctly takes place. Given that, do you still think it's better to remove it?

torch_xla/csrc/init_python_bindings.cpp
ysiraichi: My idea was to have one source of truth for both `Status` and `StatusOr<T>` types. Since `Status.value()` doesn't exist, I thought that it would be better to have a single `ConsumeAndMaybeThrow()` function for both of them. Then, it would be easier to spot whenever we are transitioning from `Status*` to throwing exceptions.

torch_xla/csrc/runtime/runtime.cpp
ysiraichi: What do you think of changing it into:

```c++
/* From: */
static absl::StatusOr<absl_nonnull std::unique_ptr<ComputationClient>>
    maybe_client = InitializeComputationClient();

/* To: */
static auto& maybe_client =
    *new absl::StatusOr<ComputationClient * absl_nonnull>(
        InitializeComputationClient());
```

Reasoning:
- References can only be initialized once (i.e. same as `T* const`)
- There are lots of non-const `ComputationClient` member functions we need to call (that's why no `const auto&`)

What do you think?

ysiraichi:
@zhanyong-wan @ghpvnist Let me know if there's anything else you'd like to see in this PR.

zhanyong-wan:
Hi @ysiraichi , I'll take a look soon. In the future, when the PR is ready for another review, could you click on the spinning-arrows icon next to the reviewer name? That'll let us know that you are done with editing and it's our turn. Thanks!

ysiraichi:
Ah, ok. Got it. I don't think I ever noticed that feature. Thanks for the tip.

test/cpp/test_runtime.cpp
```diff
@@ -0,0 +1,31 @@
+#include <gtest/gtest.h>
+
+#include "torch_xla/csrc/runtime/runtime.h"
+
+namespace torch_xla::runtime {
+
+TEST(RuntimeTest, NullComputationClient) {
```
zhanyong-wan: Test cases should be independent of each other and be able to run in any order. Otherwise it's hard to refactor and debug the test. This test case must be run before the other test case (otherwise it will fail). We should fix this dependency on test order.

I'd just remove this given that the logic is covered in the other test case.

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,68 @@
...
+InitializeComputationClient() {
+  if (sys_util::GetEnvBool("XLA_DUMP_FATAL_STACK", false)) {
+    tsl::testing::InstallStacktraceHandler();
+  }
 
+  // TODO: enable IFRT once it's not crashing anymore.
+  // Ref: https://github.com/pytorch/xla/pull/8267
+  //
+  // static bool use_ifrt = sys_util::GetEnvBool("XLA_USE_IFRT", false);
+  static const bool use_ifrt = false;
```
zhanyong-wan: No need for `static`.

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -1,50 +1,68 @@
...
+  // Reference to singleton Status-wrapped ComputationClient instance.
+  //
+  // Since we only allow a single initialization, as soon as this function is
+  // called, we store the initialization result in this trivially destructible
+  // reference.
+  static auto& maybe_client =
+      *new absl::StatusOr<ComputationClient * absl_nonnull>(
+          InitializeComputationClient());
+
+  if (!maybe_client.ok()) {
```
zhanyong-wan: You can replace the rest of this function with `return maybe_client;`.

zhanyong-wan (approved):
Thanks!

torch_xla/csrc/runtime/runtime.cpp
ysiraichi: You are right. Not sure what I was thinking, here.

torch_xla/csrc/runtime/runtime.cpp
```diff
@@ -24,13 +24,14 @@ InitializeComputationClient() {
...
   // static bool use_ifrt = sys_util::GetEnvBool("XLA_USE_IFRT", false);
-  static const bool use_ifrt = false;
+  const bool use_ifrt = false;
   if (sys_util::GetEnvString(env::kEnvPjRtDevice, "") != "") {
     auto* client =
         (use_ifrt)
             ? static_cast<ComputationClient*>(new IfrtComputationClient())
             : static_cast<ComputationClient*>(new PjRtComputationClient());
     g_computation_client_initialized = true;
+    ABSL_CHECK(client);
```
zhanyong-wan: FYI, this is unnecessary as `new` guarantees to either return a non-null pointer or crash.

zhanyong-wan (approved):

torch_xla/csrc/init_python_bindings.cpp
zhanyong-wan: SG

zhanyong-wan (approved):

